PR #753, opened 2014-12-19
Add a unit test for broken backslashes on MySQL in LIKE

nickvergessen:
There is an inconsistence between MySQL and other databases when it comes to backslashes in LIKE statements.

The correct queries are:
### MySQL

```
WHERE `id` = 'Te\\st'
WHERE `id` LIKE 'Te\\\\st'
```
### SQLite, PostgreSQL

```
WHERE `id` = 'Te\\st'
WHERE `id` LIKE 'Te\\t'
```

originally reported in https://github.com/owncloud/core/issues/12689


deeky666:
I see the requirement of portability here. Maybe we could achieve that by configuring the connection appropriately with the necessary options set for MySQL and maybe other just like we did for Oracle?

PR #755, opened 2014-12-23
Update MysqliStatement to support PDO::FETCH_OBJ

gohanman:
Return a simple stdClass object if that fetch type is requested.


deeky666:
Merged manually as of https://github.com/doctrine/dbal/commit/a8f66a29d06856356681ba8480e80015c7227f33 adding a test case.

PR #760, opened 2014-12-31
[DBAL-1097] Fix foreign key constraint referential action on Oracle

deeky666:
Oracle only supports `CASCADE` and `SET NULL` referential actions for foreign key constraints. It also supports `NO ACTION` but it cannot be declared with an explicit `ON DELETE NO ACTION` clause. The complete clause has to be omitted in order to make it work.
This PR fixes supported referential actions and omits the referential clause if either action `NO ACTION` or `RESTRICT` is set for the foreign key constraint to create.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
@@ -675,6 +675,44 @@ LEFT JOIN user_cons_columns r_cols
     }
 
     /**
+     * {@inheritdoc}
+     */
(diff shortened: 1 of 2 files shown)


Ocramius:
Merged!

`2.6`: doctrine/dbal@608d2a281b58209b8e2f69e5286c0108e254350e
`2.5`: doctrine/dbal@185b886e57e9557c4fad7a39d118000f652b72de
`2.4`: doctrine/dbal@81591fd41aee9ab3109041e530e104ff8c3cff11

PR #763, opened 2015-01-03
Add date formatting and from/to unix timestamp conversion

andig:
I'm proposing to add date formatting as well as from/to unix timestamp conversion to DBAL. If this PR is going to find resonance I'd gladly add implementations for SQlite and MySQL.

Please let me know if this is something worth to pursue.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -1290,6 +1290,47 @@ abstract class AbstractPlatform
     }
 
     /**
+     * Returns the SQL to format a date according to given format.
+     *
+     * @param string  $date
+     * @param string  $format
+     *
+     * @return string
+     *
+     * @throws \Doctrine\DBAL\DBALException If not supported on this platform.
+     */
+    public function getDateFormatExpression($date, $format) {
+        throw DBALException::notSupported(__METHOD__);
+    }
+
+    /**
+     * Returns the SQL to convert a date to a UNIX timestamp.
+     *
+     * @param string  $date
+     *
+     * @return string
+     *
+     * @throws \Doctrine\DBAL\DBALException If not supported on this platform.
+     */
+    public function getDateToUnixTimestampExpression($date) {
+        throw DBALException::notSupported(__METHOD__);
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


Ocramius:
@andig what is the exact use-case?

deeky666:
Yeah a use-case would be good. Also looking at this it seems we would need another abstraction level for the date format which requires us to implement a parser to be able to convert the DBAL date format into the database specific format. Not sure if that is worth the effort for 2.x...
I tend to rather not include something like this at this point even though it could be something for 3.0 when we potentially have other capabilities to realize things like this.

andig:
> Yeah a use-case would be good.

At http://www.volkszaehler.org we've developed a smart meter app. Millions of meter readings are stored with their timestamp in ms since 1970- unix timestamps. For analytical purposes the middleware has abilities to retrieve readings grouped by day, month, year etc which requires unix timestamp <-> date conversions as well as date arithmetics.

While we've originally only supported MySQL large, but not all parts of the application are already running on SQlite and PostgreSQL. Looking at what's missing and how to do this in the most portable way I've found the biggest gaps in the unix timestamp and date formatting functions. The missing date arithmetics have already been added in 2.5.0 (though unfortunately `getDateArithmeticIntervalExpression` is `protected`).

Ocramius:
@andig this sounds like a very custom use-case, whereas DBAL is a common abstraction layer. What would other users do with such a functionality, and why can't the `DateTime` type help them here instead?

andig:
> this sounds like a very custom use-case,

I don't know how you classify custom and non-custom use cases, but unix timestamp conversion is a recurring problem on various platforms (http://stackoverflow.com/questions/3693076/unix-timestamp-in-sqlite, http://stackoverflow.com/questions/20124862/timestamp-on-postgresql-to-unix-timestamp). MySQL has built-in functions for this purpose, other DBs do it via more general date formatting functions. 

What I would- on second though- agree with is that it might not be the right way to approach this problem by generalizing the MySQL solution of specific functions for that purpose.

Ocramius:
One thing that you might consider is that we don't consider timestamps as supported (because of how much they differ across platforms)

andig:
> One thing that you might consider is that we don't consider timestamps as supported

I wasn't aware of that. Thanks for bearing with me trough this discussion! I'll approach this by subclassing the relevant platforms for now and will see what 3.0 then brings.

PR #764, opened 2015-01-03
[DBAL-1063] Allow defining duplicate indexes on a table

deeky666:
This PR removes silent dropping of "duplicate" indexes on a table if two indexes fulfill each other or one overrules the other.
This is necessary to make detection of renamed indexes work properly (introduced in 2.5). Currently there are situations where false positive index renamings are detected because of race conditions.

**Example**

An ORM entity contains an association which adds a foreign key constraint to the underlying table. Foreign key constraints always create implicit indexes on the particular columns.
The entity also contains a mapping for a custom named index fulfilling the foreign key column(s).
The ORM schema tool first adds the foreign key constraint to the table, implicitly creating an index. Afterwards it adds the custom named index which is dropped silently by the table instance because it is considered a "duplicate".
So the table instance created by the schema now contains the auto generated index but not the custom named one.
When using the schema tool to update the schema, the DBAL schema manager is utilized to create the "online" schema. The schema manager detects two indexes in the database, the auto generated one and the custom named one (manually created by the user).
When the schema manger creates the table instance it eventually adds the custom named index first which results in the auto generated index being dropped silently.
When comparing both "offline" and "online" schema, the comparator detects an index rename which actually is none.
This leads to failing `DROP INDEX` and `CREATE INDEX` SQL.

Besides this issue "duplicate" indexes are totally legit and DBAL should not decide whether or not to create an index. This should be user responsibility.

This implementation is much more transparent and less error prone and opens up more flexibility for the user to define indexes on tables.

/cc @flack @Ocramius @beberlei

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,13 +1,32 @@
+# Upgrade to 2.5.1
+
+## MINOR BC BREAK: Doctrine\DBAL\Schema\Table
+
+When adding indexes to ``Doctrine\DBAL\Schema\Table`` via ``addIndex()`` or ``addUniqueIndex()``,
+duplicate indexes are not silently ignored/dropped anymore (based on semantics, not naming!).
+Duplicate indexes are considered indexes that pass ``isFullfilledBy()`` or ``overrules()``
+in ``Doctrine\DBAL\Schema\Index``.
+This is required to make the index renaming feature introduced in 2.5.0 work properly and avoid
+issues in the ORM schema tool / DBAL schema manager which pretends users from updating
+their schemas and migrate to DBAL 2.5.*.
+Additionally it offers more flexibility in declaring indexes for the user and potentially fixes
+related issues in the ORM.
+With this change, the responsibility to decide which index is a "duplicate" is completely deferred
+to the user.
+Please also note that adding foreign key constraints to a table via ``addForeignKeyConstraint()``,
+``addUnnamedForeignKeyConstraint()`` or ``addNamedForeignKeyConstraint()`` now first checks if an
+appropriate index is already present and avoids adding an additional auto-generated one eventually.
+
 # Upgrade to 2.5
 
 ## BC BREAK: time type resets date fields to UNIX epoch
 
-When mapping `time` type field to PHP's `DateTime` instance all unused date fields are 
-reset to UNIX epoch (i.e. 1970-01-01). This might break any logic which relies on comparing 
-`DateTime` instances with date fields set to the current date. 
+When mapping `time` type field to PHP's `DateTime` instance all unused date fields are
+reset to UNIX epoch (i.e. 1970-01-01). This might break any logic which relies on comparing
+`DateTime` instances with date fields set to the current date.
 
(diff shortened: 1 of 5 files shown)


lib/Doctrine/DBAL/Schema/Comparator.php
```diff
@@ -336,6 +336,42 @@ private function detectColumnRenamings(TableDiff $tableDifferences)
...
+
+        foreach ($tableDifferences->addedIndexes as $addedIndexName => $addedIndex) {
+            foreach ($tableDifferences->removedIndexes as $removedIndex) {
+                if (! $this->diffIndex($addedIndex, $removedIndex)) {
+                    $renameCandidates[$addedIndex->getName()][] = array($removedIndex, $addedIndex, $addedIndexName);
+                }
+            }
+        }
+
+        foreach ($renameCandidates as $candidateIndexes) {
```
Ocramius: Can you add inline comments to this file? Maybe I'm sleepy, but this is hard to read IMO

tests/Doctrine/Tests/DBAL/Schema/TableTest.php
```diff
@@ -354,19 +354,6 @@ public function testAllowImplicitSchemaTableInAutogeneratedIndexNames()
     /**
      * @group DBAL-50
      */
-    public function testAddIndexTwice_IgnoreSecond()
```
Ocramius: This is a minor BC break and needs documenting

UPGRADE.md
```diff
@@ -1,13 +1,32 @@
+# Upgrade to 2.5.1
+
+## MINOR BC BREAK: Doctrine\DBAL\Schema\Table
+
+When adding indexes to ``Doctrine\DBAL\Schema\Table`` via ``addIndex()`` or ``addUniqueIndex()``,
+duplicate indexes are not silently ignored/dropped anymore (based on semantics, not naming!).
+Duplicate indexes are considered indexes that pass ``isFullfilledBy()`` or ``overrules()``
+in ``Doctrine\DBAL\Schema\Index``.
+This is required to make the index renaming feature introduced in 2.5.0 work properly and avoid
+race conditions in the ORM schema tool / DBAL schema manager which pretends users from updating
```
Ocramius: They're not race conditions

Ocramius: Just "issues" is fine

UPGRADE.md
```diff
@@ -1,13 +1,32 @@
...
+their schemas.
```
Ocramius: and migrate to DBAL 2.5.*

UPGRADE.md
```diff
@@ -1,13 +1,32 @@
...
+Additionally it offers more flexibility in declaring indexes for the user and potentially fixes
+related issues in the ORM.
+With this change the responsibility to decide which index is a "duplicate" is completely deferred
```
Ocramius: comma after `change`

Ocramius:
Backported into `2.5` at doctrine/dbal@0d817986c293893d45a4ec4768558dc7a86b4643

PR #765, opened 2015-01-03
[WIP] Remove PHP 5.3 support

Ocramius:
Just a suggestion for master (2.6). We should move on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 5.3
   - 5.4
   - 5.5
   - 5.6
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,9 @@
+# Upgrade to 2.6
+
+## PHP Version requirement change
+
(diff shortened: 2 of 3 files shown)


deeky666:
:+1: will take some time anyways until 2.6 will be released so I'm fine with it.

stof:
Dropping it just for the matter of dropping it is not worth it IMO. This is useless if it does not allow to clean the codebase

Ocramius:
@stof this has actually a lot of avantages for us, even though we won't change the codebase in one shot (mainly to avoid merge conflicts).

@deeky666 is setting up a complex CI environment for internal use by the doctrine org, and 5.3.x is very problematic to deal with.

Yes, that was an excuse to drop it, but there's also no good excuse to keep supporting 5.3.x.

Considering the current PHP 5.4 lifetime, we will also be able to introduce generator usage by the end of the year by dropping 5.4 support in DBAL 2.7 or such.

Tobion:
This can be closed as #880 has been merged.

deeky666:
Implemented in #2419

PR #767, opened 2015-01-04
[3.0] Removed drizzle support.

kimhemsoe:


UPGRADE.md
```diff
@@ -1,3 +1,10 @@
+# Upgrade to 2.6
```
Ocramius: I think we need to keep this for 3.6, as we don't have a way to stop users from upgrading...

kimhemsoe: 3.0 ?

Ocramius: I think so...

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

stof: Removing it is a BC break, so removing it would force to go to 3.0 (or to stop following semver, which would be a pain for all Composer users)

stof:
it could be marked as deprecated though

SenseException:
@stof That makes sense. It would be a surprise for Drizzle users if the support would suddenly vanish.

Ocramius:
This will not land in 2.x anyway, adding a `[3.0]` to the title

SenseException:
So how about deprecation in 2.x versions? Maybe starting in version 2.6 (If there will be one of course).

Ocramius:
@SenseException the PR as-is is fine, it will just be merged once we have a schedule for 3.0

PR #768, opened 2015-01-04
DBAL-1106 Improve schema introspection performance on Oracle

deeky666:
This PR improves the overall schema introspection performance on Oracle by optimizing some of the `getList*()` method SQL queries in the platform.
Unfortunately Oracle system views are rather slow, especially when joining tables. Using subqueries can improve the execution time a lot.

**Improved queries**

(the times have been measured using an exemplary query from the functional test suite)

`getListTableForeignKeysSQL()`: Down from `~2.33s` to `~0.68s` per query
`getListTableColumnsSQL()`: Down from `~0.49s` to `~0.14s` per query
`getListTableIndexesSQL()`: Down from `~0.5s` to `~0.25s` per query

Overall improvement of `OracleSchemaManagerTest` from `~100s` to `~37.5s`.

I was quite impressed by the performance improvements :)

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
@@ -419,15 +419,33 @@ class OraclePlatform extends AbstractPlatform
     {
         $table = $this->normalizeIdentifier($table);
 
-        return "SELECT uind.index_name AS name, " .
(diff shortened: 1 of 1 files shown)


Ocramius:
Looks good, but will only land in `2.6` due to the risk of regressions

deeky666:
Thx. Did not intend to backport it either :)

PR #769, opened 2015-01-08
Allow overriding implicit indexes

deeky666:
This is a follow-up patch for the change introduced in https://github.com/doctrine/dbal/pull/764.
It makes the `Table` class act more intelligent when it comes to preferring explicit over implicit indexes.
Currently it is necessary to construct a `Table` instance in a specific order to not have unnecessary duplicate indexes because of foreign keys. If you first add a foreign key and afterwards and explicit index that is fulfilling the implicit one, both indexes are stored. Instead it is more intelligent to replace the implicit by the explicit one if and only if the explicit one fulfills the implicit one.
This should also fix possible issues with how ORM's schema tool constructs a schema. See [here](https://travis-ci.org/doctrine/doctrine2/jobs/46321581#L307) where for a OneToOne relation on a primary key an additional regular index for the foreign key is created.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Table.php
+++ b/lib/Doctrine/DBAL/Schema/Table.php
@@ -45,6 +45,11 @@ class Table extends AbstractAsset
     /**
      * @var Index[]
      */
+    private $implicitIndexes = array();
+
+    /**
+     * @var Index[]
+     */
     protected $_indexes = array();
 
     /**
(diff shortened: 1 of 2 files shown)


stof:
What about ``getIndex`` ? Should it take implicit indexes into account ? If it does not, it might be a BC break

deeky666:
@stof implicit indexes are also stored in the regular index array. So this is no problem.

Ocramius:
Merged, thanks @deeky666!

`master`: doctrine/dbal@32fae862d376eb4eed60db4bf4e82f1a675c6509
`2.5`: doctrine/dbal@4067c72515eac46c94aaf7db1c66389c8c75c9f2

webdevilopers:
:+1: 

I can confirm this works for `indexes` on `Table` now. I was trying the same for `uniqueConstraints`:
```php
/**
 * @ORM\Entity(repositoryClass="Application\Sonata\UserBundle\Entity\UserRepository")
 * @ORM\Table(name="users",
 *   indexes={
 *     @ORM\Index(name="compartment_idx", columns={"compartment_id"}),
 *   },
 *   uniqueConstraints={
 *     @ORM\UniqueConstraint(name="username_canonical_unique", columns={"username_canonical"})
 *   }
 * )
 */
```

but running `doctrine:migrations:diff` creates the same unique index twice:
```
class Version20150114110814 extends AbstractMigration
{
    public function up(Schema $schema)
    {
        $this->addSql('CREATE INDEX compartment_idx ON users (compartment_id)');
        $this->addSql('CREATE UNIQUE INDEX UNIQ_1483A5E992FC23A8 ON users (username_canonical)');
        $this->addSql('CREATE UNIQUE INDEX username_canonical_unique ON users (username_canonical)');
    }
}
```

deeky666:
The unique index `UNIQ_1483A5E992FC23A8` is probably created by a column mapping with option `unique=true`. As this one is auto-generated implicitly by ORM's schema tool, we might consider filtering those out if a fulfilling custom one is present.
@Ocramius what do you think? Btw besides creating a unique index, what does `@Column(unique=true)` else do?

PR #774, opened 2015-01-14
Added SET and ENUM types for MySQL and fix issue with schema update tool

dimak08:
Set and Enum data types are one of the most popular. However, addition of these data types in the project creates a problem when updating of the database structure.

### DIFF
--- a/docs/en/reference/types.rst
+++ b/docs/en/reference/types.rst
@@ -233,6 +233,38 @@ or ``null`` if no data is present.
     bit data type if necessary to ensure the least possible data storage
     requirements are met.
 
+enum
+^^^^
+
+Maps and convert enum type
+An ENUM value must be one of those listed in the column definition, or the internal
+numeric equivalent thereof.
+You can specify allowed values in ``values`` in ``options`` @Column annotation
+For example:
+
+.. code-block:: php
+
+    /**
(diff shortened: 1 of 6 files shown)


Ocramius:
This won't be implemented, sorry.

We only support portable functionality, and `SET` and `ENUM` are, honestly, a mess that we can't deal with in a cross-platform compliant way.

Also see http://stackoverflow.com/a/9057352/347063

stof:
you should submit them to https://github.com/beberlei/DoctrineExtensions with other MySQL-specific stuffs (not sure whether the repo is still maintained though).

And I'm not sure this will work properly in the schema tool when inspecting enums or when comparing schemas btw

Ocramius:
@dimak08 note that this is also a very old feature request that was marked as "won't fix" earlier too: http://www.doctrine-project.org/jira/browse/DBAL-4

deeky666:
You could make it a commented type to allow the schema manager to recognize it.

stof:
@deeky666 no you cannot. You can rely on commented types if you have a Doctrine type for ``ENUM('foo', 'bar')`` and a **different** Doctrine type for ``ENUM('foo', 'bar', 'baz')``. this is precisely what makes the ENUM type so painful

PR #775, opened 2015-01-16
fix diffColumn in text length is different

arima-ryunosuke:
Because mysql has multiple TEXT types.

```
    $column1 = new Column('clobfield1', Type::getType('text'), array('Length' => 128));
    $column2 = new Column('clobfield1', Type::getType('text'), array('Length' => 65536));

    $platform = new MySqlPlatform();
    var_dump($platform->getClobTypeDeclarationSQL($column1->toArray()));// this is "TINYTEXT"
    var_dump($platform->getClobTypeDeclarationSQL($column2->toArray()));// this is "MEDIUMTEXT"

    $c = new Comparator();
    var_dump($c->diffColumn($column1, $column2));// this is empty!
```

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Comparator.php
+++ b/lib/Doctrine/DBAL/Schema/Comparator.php
@@ -441,7 +441,7 @@ class Comparator
         }
 
         if (($properties1['type'] instanceof Types\StringType && ! $properties1['type'] instanceof Types\GuidType) ||
-            $properties1['type'] instanceof Types\BinaryType
+            $properties1['type'] instanceof Types\TextType || $properties1['type'] instanceof Types\BinaryType
         ) {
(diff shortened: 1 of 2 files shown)


deeky666:
@arima-ryunosuke the `TextType` type is not inteded to use the `length` option of a column and therefore should also not be compared. MySQL is the only platform that has certain variants of "text" types but in fact all of those except `LONGTEXT` are considered `StringType` in DBAL sense.
The `TextType` shall be used for a varying character column without length limit. Otherwise if you want to limit the amount of data to a certain length, please use `StringType` (DBAL will map to `TINYTEXT`, `TEXT`, `MEDIUMTEXT` and `LONGTEXT` accordingly as soon as the `length` option exceeds `255`).

arima-ryunosuke:
I see. Thank you.

PR #776, opened 2015-01-16
[WIP] Don't throw an exception during database platform guessing

### LINKED ISSUE #748: Doctrine DBAL 2.5 Breaks Symfony if DB Doesn't Exist
This is a problem with the most recent symfony-standard.

With the newest Doctrine DBAL 2.5, Doctrine introduced `Doctrine\DBAL\Connection::detectDatabasePlatform()` which is different than 2.4. Now in 2.5, any time a repository is instantiated (not even used) by the DI, we get a PDOException if the DB doesn't already exist.

Imagine trying to run `app/console doctrine:database:create` and getting a "Database doesn't exist!" error. Repositories as services are very common in most projects, so this is a huge issue.

To reproduce quickly, create a fresh Symfony project, letting composer install the AcmeDemoBundle. Use all default parameters. Now try to generate an entity. You'll get a PDOException. This is a new regression.
```
composer create-project symfony/framework-standard-edition symfony-standard
cd symfony-standard
app/console doctrine:generate:entity --entity=AcmeDemoBundle:Foo
```
You get error this on every OS/distro/platform:
```
  [Doctrine\DBAL\Exception\ConnectionException]
  An exception occured in driver: SQLSTATE[HY000] [1049] Unknown database 'symfony'

  [Doctrine\DBAL\Driver\PDOException]
  SQLSTATE[HY000] [1049] Unknown database 'symfony'

  [PDOException]
  SQLSTATE[HY000] [1049] Unknown database 'symfony'
```

rosier:
Fallback to the "default" database platform if connecting to the database fails.

Fixes https://github.com/symfony/symfony-standard/issues/748

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -32,6 +32,7 @@ use Doctrine\DBAL\Cache\QueryCacheProfile;
 use Doctrine\DBAL\Cache\ArrayStatement;
 use Doctrine\DBAL\Cache\CacheException;
 use Doctrine\DBAL\Driver\PingableConnection;
+use Doctrine\DBAL\Exception\DriverException;
 
 /**
  * A wrapper around a Doctrine\DBAL\Driver\Connection that adds features like
@@ -424,9 +425,14 @@ class Connection implements DriverConnection
             return $this->_params['serverVersion'];
         }
 
-        // If not connected, we need to connect now to determine the platform version.
+        // If not connected, try to connect now to determine the platform version.
         if (null === $this->_conn) {
-            $this->connect();
+            try {
+                $this->connect();
+            } catch (DriverException $e) {
+                // Unable to detect platform version.
+                return null;
+            }
         }
 
         // Automatic platform version detection.


Ocramius:
This fix seems incorrect to me (silent failure, not good).

See http://www.doctrine-project.org/jira/browse/DBAL-1057 for further discussion on the matter.

rosier:
I looked into this a bit more and may have another fix.

I will try to update this PR soon

deeky666:
No sorry but this is silencing all other possible connection failures also. Unfortunately it is not that easy :(
I tend to say that this issue has to be fixed in the calling API and that we should add documentation to `Doctrine\DBAL\Connection` that calling `getDatabasePlatform()` can only succeed if the db credentials are valid unless platform detection is turned off by either using `serverVersion` parameter or custom platform implementation.
I know that this suggestion may not be satisfying but otherwise we would try to add too much intelligence into DBAL connection which cannot be safe as we never know the exact context of the calling API. There may be use cases when a failure is expected...

Ocramius:
Closing this PR as inapplicable

PR #777, opened 2015-01-18
[DBAL-1121] Allow optional host and server connection parameters for sqlanywhere driver

deeky666:
This patch makes it possible to connect to a SQL Anywhere database without having to specify `host` and/or `server` connection parameter.
The original assumption that the `server` parameter is necessary to connect to the database was wrong. It is only required if the host is running multiple named server instances and a specific instance should be connected to.
Also if the `host` parameter is not specified, it will default to `localhost` now.
The `LINKS` DSN parameter was replaced by the simpler and encouraged `HOST` parameter as the `LINKS` parameter is only required if you want to pass specific TCP/IP options.


Ocramius:
Merged, thanks @deeky666!

`master`: doctrine/dbal@9e7a971c77c348bb501c16b58eabc2b1c7a37c10
`2.5`: doctrine/dbal@cf3afaedf94542e54cb19cff3e50b64465862467

PR #779, opened 2015-01-18
DBAL-1123 Initialize database schema only once per PHPUnit run

deeky666:
Currently the PHPUnit test suite recreates the database schema on each invocation of `TestUtil::getConnection()` which is highly inefficient and costly concerning performance, especially on platforms where schema recreation consumes a lot of time.
With this patch the database schema is initilaized once per PHPUnit run, increasing overall test suite runtime significantly.
For example using SQL Anywhere, database creation takes about 4-5 seconds which results in a complete test suite runtime of several minutes. With this patch the complete test suite runs in about 12-15 seconds **!**.

### DIFF
--- a/tests/Doctrine/Tests/TestUtil.php
+++ b/tests/Doctrine/Tests/TestUtil.php
@@ -13,6 +13,11 @@ use Doctrine\DBAL\DriverManager;
 class TestUtil
 {
     /**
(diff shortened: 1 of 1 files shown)


Ocramius:
Thanks @deeky666 !

PR #780, opened 2015-01-20
Add autoload to composer.json

ng-pgill:


stof:
The 2.0.x branch is not maintained anymore since approx 3 years. So this does not make much sense

PR #781, opened 2015-01-22
Call `detectDatabasePlatform` only once

rosier:
Database platform detection is triggered twice if `Doctrine/DBAL/Connection::getDatabasePlatform()` is called before `Doctrine/DBAL/Connection::connect()`

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -324,7 +324,7 @@ class Connection implements DriverConnection
      */
     public function getDatabasePlatform()
     {
-        if (null == $this->platform) {
+        if (null === $this->platform) {
             $this->detectDatabasePlatform();
         }
 
@@ -359,10 +359,6 @@ class Connection implements DriverConnection
 
         $this->_conn = $this->_driver->connect($this->_params, $user, $password, $driverOptions);
         $this->_isConnected = true;
-
-        if (null === $this->platform) {
-            $this->detectDatabasePlatform();
-        }
 
         if (false === $this->autoCommit) {
             $this->beginTransaction();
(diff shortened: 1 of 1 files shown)


Ocramius:
@rosier requires a test case

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1597,7 +1593,7 @@ public function ping()
         }
 
         try {
-            $this->query($this->platform->getDummySelectSQL());
+            $this->query($this->getDatabasePlatform()->getDummySelectSQL());
```
deeky666: Why is this change necessary? `Connection::ping()` calls `Connection::connect()` as the beginning, so it ensures the platform is set.

rosier: The platform isn't set anymore in `connect()` with this PR.

In the current version of `Connection` the following calls are triggered:
`getDatabasePlatform()` => `detectDatabasePlatform()` => `getDatabasePlatformVersion()`  => `connect()` at this point because the platform isn't set yet `connect()` calls `detectDatabasePlatform()` again.

This PR removes the call to `detectDatabasePlatform()` from `connect()` to stop that from happening.

rosier:
@Ocramius `Connection::detectDatabasePlatform()` is a private function. I'm not sure if there is a good way to test that it is called only once.

Ocramius:
@rosier try looking at the wrapped driver: that one can be mocked

deeky666:
Merged manually via https://github.com/doctrine/dbal/commit/de755610a46e7f3b91ad3070109364315e95fb73. Verified the issue by adding a test. Thanks @rosier !

PR #783, opened 2015-01-23
Fixes issue (DBAL-1057) use default platform when not connected

weaverryan:
Hi guys!

This is a fix for http://www.doctrine-project.org/jira/browse/DBAL-1057. It leaves the ability to set the platform and platform version from sha: 3176f51d1426022d305c6531a9b9bc93a868bddd, but does not try to connect if we're not already connected. This is consistent with the previous behavior (again see the linked sha) - before there was never a connection made to determine the platform.

The alternate solution is to connect, but surround this by a try-catch (`PDOException`, `DriverException`, `Exception`?) and return `null`. in case we have some situation where the database isn't created or the connection information is wrong.

I know this issue is causing a lot of problems around my world (I ran into myself yesterday), so thanks in advance for the attention.

Thanks!

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -20,6 +20,7 @@
 namespace Doctrine\DBAL;
 
 use Doctrine\DBAL\Driver\ServerInfoAwareConnection;
+use Doctrine\DBAL\Exception\DriverException;
 use Doctrine\DBAL\Exception\InvalidArgumentException;
 use PDO;
 use Closure;
@@ -429,7 +430,12 @@ class Connection implements DriverConnection
         // cause an error in our application when determining the platform
         // version is not mission-critical
         if (null === $this->_conn) {
-            return null;
+            try {
+                $this->connect();
+            } catch (DriverException $e) {
+                // i.e. connection params are invalid or db doesn't exist
+                return null;
+            }
         }
 
         // Automatic platform version detection.
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -424,9 +424,12 @@ class Connection implements DriverConnection
             return $this->_params['serverVersion'];
         }
 
-        // If not connected, we need to connect now to determine the platform version.
+        // If not connected, we can't figure out the server version. If
+        // we try to connect, it may fail (e.g. no database), which would
+        // cause an error in our application when determining the platform
+        // version is not mission-critical
         if (null === $this->_conn) {
-            $this->connect();
+            return null;
         }
 
         // Automatic platform version detection.
--- a/tests/Doctrine/Tests/DBAL/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/ConnectionTest.php
@@ -403,15 +403,6 @@ SQLSTATE[HY000]: General error: 1 near \"MUUHAAAAHAAAA\"");
         $this->assertSame($result, $conn->fetchColumn($statement, $params, $column, $types));
     }
 
-    public function testConnectionIsClosed()
-    {
-        $this->_conn->close();
-
-        $this->setExpectedException('Doctrine\\DBAL\\Exception\\DriverException');
-
-        $this->_conn->quoteIdentifier('Bug');
-    }
-
     public function testGetPlatformVersionDoesNotFailOnBadConnection()
     {
         // the connection uses bogus connection requirements, so connecting
--- a/tests/Doctrine/Tests/DBAL/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/ConnectionTest.php
@@ -412,6 +412,14 @@ SQLSTATE[HY000]: General error: 1 near \"MUUHAAAAHAAAA\"");
         $this->_conn->quoteIdentifier('Bug');
     }
 
+    public function testGetPlatformVersionDoesNotFailOnBadConnection()
+    {
+        // the connection uses bogus connection requirements, so connecting
+        // *will* fail. We should be able to call this without connection
+        // and causing that exception
+        $this->_conn->getDatabasePlatform();
+    }
+
     public function testFetchAll()
     {
         $statement = 'SELECT * FROM foo WHERE bar = ?';


Ocramius:
@weaverryan can you inspect testability of this change?

weaverryan:
@Ocramius yep, let me see about that now. Is this ok that it's against the 2.5 branch? My branch is based off of 2.5.

Ocramius:
@weaverryan we can handle merging back int master, yeah: we'd simply cherry pick.

weaverryan:
@Ocramius I've just added a test for this, it's very simple: calling `Connection::getDatabasePlatform()` in `ConnectionTest` WAS failing before this patch, because the params passed to `Connection` in the test are bogus. The fact that we can call this (with bogus params) and not get an exception shows the functionality.

I also removed another test that I think os bogus - see #784.

Thanks!

weaverryan:
I just changed the implementation to use try-catch at sha: 2db254d336a9d7cb87f4bc71cf20e050cd91d8e1. I can go either way but this makes more sense to me.

Thanks!

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -403,13 +403,12 @@ public function testFetchColumn()
         $this->assertSame($result, $conn->fetchColumn($statement, $params, $column, $types));
     }
 
-    public function testConnectionIsClosed()
```
Ocramius: @weaverryan why was this test actually removed?

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -403,13 +403,12 @@ public function testFetchColumn()
...
     {
-        $this->_conn->close();
-
-        $this->setExpectedException('Doctrine\\DBAL\\Exception\\DriverException');
-
-        $this->_conn->quoteIdentifier('Bug');
+        // the connection uses bogus connection requirements, so connecting
+        // *will* fail. We should be able to call this without connection
+        // and causing that exception
+        $this->_conn->getDatabasePlatform();
```
Ocramius: Missing assertion (should assert that the returned value is `null`

tests/Doctrine/Tests/DBAL/ConnectionTest.php
weaverryan: I don't believe this is a valid test, which is what I'm arguing here: https://github.com/doctrine/dbal/pull/784/files#r23499338. I would like to remove this test - I was using that PR simply to abuse Travis to show how the test was passing for the wrong reason.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
weaverryan: Actually, the database platform will *never* be null, and the point of the test is simply to show that asking for the database platform (with invalid connection credentials, which this test class uses) no longer causes an exception. If you removed my "fix" in this PR, calling `getDatabasePlatform()` would actually explode.

I can add an assertion that the database platform is in fact *not* null (it is *always* determine, it's just that it uses the default platform if it cannot connect to get the version), but the real point is that this doesn't cause an exception.

I feel like you and I aren't *quite* on the same page with this change and #784. I am vastly less qualified than you in this library, so perhaps it's my mistake, and if so - I appreciate the help clarifying :). Overall, I'm simply trying to fix the change in behavior where calling `getDatabasePlatform()` began throwing an exception if a connection couldn't be made (instead of using the default platform, like in earlier versions of DBAL).

Thanks!

Ocramius: > began throwing an exception if a connection couldn't be made (instead of using the default platform, like in earlier versions of DBAL).

Yeah, well, it's due to platform detection, as already discussed.

I'm also not very qualified to talk about the DBAL: it's mainly @deeky666's competence area.

weaverryan: Cool, I would like to hear what @deeky666 thinks then :).

I'm here because the exception has wide-reaching consequences that are affecting Symfony users - e.g. doctrine/DoctrineBundle#351 and symfony/symfony-standard#774. 

Thanks for the attention!

SoboLAN:
I think surrounding the connection statement with `try-catch` and returning `null` if it fails is more of a hack than a fix.

What I mean is... let's have an example:
- if I have 2 projects in Symfony: project A and project B
- project A depends on Doctrine/DBAL, project B does not

In this situation, if I only use project B, then there will always be a connection made to the database, even if I never use it. Not really ideal.

stof:
@SoboLAN if you have 2 projects, you should probably use 2 kernels, so that your project which does not need DBAL does not load it at all. It is not related to this fix at all

weaverryan:
Hi guys!

I just hit this in a room of trainees today as we started a new Symfony 2.6 project. We tried to generate an entity (not a DB-connecting operation), and it bombed out (we hadn't even talked DB credentials yet). This seems small, but in practical terms, it affects all new Symfony 2.6 projects the moment they first interact with Doctrine. I'm happy to make any changes necessary.

@SoboLAN I won't agree or disagree actually :). But regardless, a connection *is* made currently if you ask for the database platform - that's a fact of this DBAL version (I want to not discuss it here for reasons of focus). This PR simply says: if that existing "open a connection" fails while figuring out the DB platform, don't bomb out, just use the default platform (which is the behavior of the last DBAL version).

Thanks for everyone's time :)

Ocramius:
OT: please don't teach entity generation to trainees ;-)

lib/Doctrine/DBAL/Connection.php
```diff
@@ -424,9 +425,17 @@ private function getDatabasePlatformVersion()
...
-        // If not connected, we need to connect now to determine the platform version.
+        // If not connected, we can't figure out the server version. If
+        // we try to connect, it may fail (e.g. no database), which would
+        // cause an error in our application when determining the platform
+        // version is not mission-critical
         if (null === $this->_conn) {
-            $this->connect();
+            try {
+                $this->connect();
+            } catch (DriverException $e) {
```
deeky666: I'm still unhappy silencing all kind of connection errors as it also silences errors for valid credentials when there are other problems such as failing network or whatever.
Returning the default platform in such cases still might be fine 90% of the time but in IMO this is a hack to come around "invalid" (since 2.5) `getDatabasePlatform()` calls from userland. This really is breaking the feature that was implemented here and might be causing more trouble in the end than it solves.
The question is why do you need the platform at all if not connected? And what advantage do you have by retrieving just "some" platform that might be incompatible with your backend anyways? Before DBAL 2.5 that was just plain wrong IMO.
One might strongly disagree with my opinion here and "just don't care" but then all this platform version specific behaviour stuff that got implemented over the past few years is pretty useless.
I get the problems and the frustration this issue currently creates but we really please should not hack around that feature here. Instead we should fix calling libs to just use DBAL in the right way or revert the whole feature...
Just my 2 cents...

weaverryan:
@Ocramius but if I make them do it by hand the first time, they will kick me out of the room :p

weaverryan:
@deeky666 Thanks very much for the clarification! I don't love the BC break, but I also see that it's well-documented in the UPGRADE log :).

I have opened doctrine/doctrine2#1294, which i believe fixes the biggest "userland"  bad use of this method (indeed, it's calling `getDatabasePlatform()` in situations where it doesn't even need that information).

Thanks!

deeky666:
@weaverryan thank you for your patience and help! Fixing the ORM to avoid unneccesary `getDatabasePaltform()` calls is a good first step. I believe the bigger problem (Symfony-wise) is in the DoctrineBundle [here](https://github.com/doctrine/DoctrineBundle/blob/master/ConnectionFactory.php#L60-L68). Registering mapping types requires the retrieving the platform currently. We have to find a solution for that so that commands like `doctrine:database:create` work again. I believe the `ConnectionFactory` is the biggest problem as it is utilized nearly everywhere in Symfony userland AFAIK.

lib/Doctrine/DBAL/Connection.php
stof: @deeky666 the issue is that many places in Doctrine ORM (and in some places in DBAL too IIRC) are retrieving the platform eagerly in constructors to optimize performance by avoiding to call the platform getter multiple times.

stof:
@deeky666 for DoctrineBundle, anyone creating DBAL types with the uptodate ways to mark them as commented or as mapped to DB types (i.e. in the type itself) would not trigger this code anymore (it is the legacy way to register the type mapping). Thus newcomers won't be registering custom types. This is not the cause of their issue.
The bigger issue for Symfony users is because of the issues in the ORM.

PR #670, opened 2014-08-27
Handle default values for datetime and datetimetz columns in PostgreSQL

sarcher:
When dealing with legacy schema it would be nice to be able to map default values and not have the schema spit out `ALTER` statements each time. This works correctly today for basic integer and string columns, but does not handle columns with special types such as `boolean` or `datetime`.

For example, the following statement:

`ALTER TABLE test_table ALTER test_column SET DEFAULT CURRENT_TIMESTAMP;`

Results in a column definition like:

`test_column | timestamp with time zone | not null default now()`

However, repeating the same schema generation will result in the same `ALTER` statement each time, because it will always detect that the default value has changed. The same is true for boolean columns.

This simple change prevents this situation from happening and correctly detects that the column default has not changed. It is specific to PostgreSQL.

PR #785, opened 2015-01-25
Handle default values for datetime/datetimetz columns in PostgreSQL

sarcher:
(This was originally in #670 but am re-submitting due to a busted merge)

When dealing with legacy schema it would be nice to be able to map default values and not have the schema spit out ALTER statements each time. This works correctly today for basic integer and string columns, but does not handle columns with special types such as boolean or datetime.

For example, the following statement:

`ALTER TABLE test_table ALTER test_column SET DEFAULT CURRENT_TIMESTAMP;`

Results in a column definition like:

`test_column | timestamp with time zone | not null default now()`

However, repeating the same schema generation will result in the same ALTER statement each time, because it will always detect that the default value has changed. The same is true for boolean columns.

This simple change prevents this situation from happening and correctly detects that the column default has not changed. It is specific to PostgreSQL.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
@@ -426,6 +426,12 @@ class PostgreSqlSchemaManager extends AbstractSchemaManager
             case 'year':
                 $length = null;
                 break;
+            case 'timestamp':
+            case 'timestamptz':
+                if ($tableColumn['default'] === 'now()') {
+                    $tableColumn['default'] = $this->_platform->getCurrentTimestampSQL();
+                }
+                break;
 
             // PostgreSQL 9.4+ only
             case 'jsonb':
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
@@ -17,6 +17,7 @@ use function array_map;
 use function array_pop;
 use function count;
 use function strtolower;
+use function substr;
 
 class PostgreSqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
 {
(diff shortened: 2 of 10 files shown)


sarcher:
@Ocramius re-submitted as requested; thanks and sorry for the problems!

Ocramius:
> thanks and sorry for the problems!

What problems? ;-)

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -378,6 +378,12 @@ protected function _getPortableTableColumnDefinition($tableColumn)
             case 'year':
                 $length = null;
                 break;
+            case 'timestamp':
+            case 'timestamptz':
+                if ($tableColumn['default'] == 'now()') {
```
Ocramius: I don't see `now()` in the test cases: is it a missing test?

sarcher: It is kind of subtle; the tests will create a new table using the mapping of `CURRENT_TIMESTAMP`, and then the schema manager will call `listTableDetails()` on the database itself, which will report the column as having a default of `now()` (e.g. `not null default now()`). So the actual change in this PR is recognizing that "now()" is the same as the platform's `getCurrentTimestampSQL()`.

The reason the tests must use `CURRENT_TIMESTAMP` and not `NOW()` is that it must match what is in `AbstractPlatform::getCurrentTimestampSQL()`.

sarcher:
Just wanted to check up on this to see if there was anything else needed of me on it -- it's on my open to do list. :)

Ocramius:
@sarcher I think this just needs a rebase. Assigning to @deeky666 for merging when he has time

sarcher:
@deeky666 I have rebased this to the current master, and I think it is ready for merging. Note that I did need to add an explicit `$conncetion->close();` to an existing unrelated test in order for the PostgreSQL tests to work outside of process isolation, which is needed until #849 is merged.

During testing the rebase I also logged [DBAL-1274](http://www.doctrine-project.org/jira/browse/DBAL-1274) as I realized that all of the Travis builds seemed to be testing against PostgreSQL 9.4. I proposed a fix in there, but not sure of the impact.

morozov:
Closing as stale.

PR #786, opened 2015-01-26
Fix removing autoincrement column from a primary key

andig:
https://github.com/doctrine/dbal/pull/430 is only a partial fix for DBAL-464. This adds treating autoincrement columns that get removed when altering a primary key.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -659,7 +659,8 @@ class MySqlPlatform extends AbstractPlatform
             // Dropping primary keys requires to unset autoincrement attribute on the particular column first
             if ($chgIndex->isPrimary() && $diff->fromTable instanceof Table) {
 
-                foreach ($diff->fromTable->getIndex($chgIndex->getName())->getColumns() as $columnName) {
+                // when chgIndex->isPrimary we can be sure fromTable->hasPrimaryKey is true
+                foreach ($diff->fromTable->getPrimaryKeyColumns() as $columnName) {
                     $column = $diff->fromTable->getColumn($columnName);
 
                     if ($column->getAutoincrement() === true && in_array($columnName, $chgIndex->getColumns()) === false) {
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -654,6 +654,24 @@ class MySqlPlatform extends AbstractPlatform
             }
         }
 
+        // handle columns that are removed from changed indexes
+        foreach ($diff->changedIndexes as $chgKey => $chgIndex) {
+            // Dropping primary keys requires to unset autoincrement attribute on the particular column first
+            if ($chgIndex->isPrimary() && $diff->fromTable instanceof Table) {
+
+                foreach ($diff->fromTable->getIndex($chgIndex->getName())->getColumns() as $columnName) {
+                    $column = $diff->fromTable->getColumn($columnName);
+
+                    if ($column->getAutoincrement() === true && in_array($columnName, $chgIndex->getColumns()) === false) {
+                        $column->setAutoincrement(false);
+
+                        $sql[] = 'ALTER TABLE ' . $table . ' MODIFY ' .
+                            $this->getColumnDeclarationSQL($column->getQuotedName($this), $column->toArray());
+                    }
+                }
+            }
+        }
+
         $engine = 'INNODB';
 
         if ($diff->fromTable instanceof Table && $diff->fromTable->hasOption('engine')) {
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -625,6 +625,9 @@ class MySqlPlatform extends AbstractPlatform
 
                         $sql[] = 'ALTER TABLE ' . $table . ' MODIFY ' .
                             $this->getColumnDeclarationSQL($column->getQuotedName($this), $column->toArray());
+
+                        // original autoincrement information might be needed later on by other parts of the table alteration
+                        $column->setAutoincrement(true);
                     }
                 }
             }
@@ -668,6 +671,9 @@ class MySqlPlatform extends AbstractPlatform
 
                         $sql[] = 'ALTER TABLE ' . $table . ' MODIFY ' .
                             $this->getColumnDeclarationSQL($column->getQuotedName($this), $column->toArray());
+
+                        // original autoincrement information might be needed later on by other parts of the table alteration
+                        $column->setAutoincrement(true);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -654,6 +654,24 @@ protected function getPreAlterTableIndexForeignKeySQL(TableDiff $diff)
             }
         }
 
+        // handle columns that are removed from changed indexes
+        foreach ($diff->changedIndexes as $chgKey => $chgIndex) {
```
Ocramius: Can this block be moved to a private method?

Ocramius:
@deeky666 I think this change is fine.

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -654,6 +654,24 @@ protected function getPreAlterTableIndexForeignKeySQL(TableDiff $diff)
...
+            // Dropping primary keys requires to unset autoincrement attribute on the particular column first
+            if ($chgIndex->isPrimary() && $diff->fromTable instanceof Table) {
+
+                foreach ($diff->fromTable->getIndex($chgIndex->getName())->getColumns() as $columnName) {
+                    $column = $diff->fromTable->getColumn($columnName);
+
+                    if ($column->getAutoincrement() === true && in_array($columnName, $chgIndex->getColumns()) === false) {
+                        $column->setAutoincrement(false);
```
deeky666: I'm not sure if it's a good idea mutating the "from" table here as this piece of information might be needed later on by other parts of the table alteration (even if it was not the case at the moment). I'd better go a safe way here. So my suggestion is to either set it to `true` again after the SQL generation part or clone the column.
@Ocramius thoughts?

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -654,6 +654,24 @@ protected function getPreAlterTableIndexForeignKeySQL(TableDiff $diff)
             }
         }
 
+        // handle columns that are removed from changed indexes
+        foreach ($diff->changedIndexes as $chgKey => $chgIndex) {
+            // Dropping primary keys requires to unset autoincrement attribute on the particular column first
+            if ($chgIndex->isPrimary() && $diff->fromTable instanceof Table) {
+
+                foreach ($diff->fromTable->getIndex($chgIndex->getName())->getColumns() as $columnName) {
```
deeky666: Can you use `$diff->fromTable->getPrimaryKey()` here? It is more explicit and the correct API for retrieving it and does not rely on the name of a primary key.

deeky666: Or even better please use `$diff->fromTable->getPrimaryKeyColumns()` API

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
```diff
@@ -329,6 +329,32 @@ public function testAlterTableAddPrimaryKey()
     /**
      * @group DBAL-464
      */
+    public function testAlterPrimaryKeyWithAutoincrementColumn()
+    {
+        $table = new Table("drop_primary_key");
+        $table->addColumn('id', 'integer', array('primary' => true, 'autoincrement' => true));
```
deeky666: There is no column option called `primary`. Please remove.

deeky666:
Besides my nitpicking it is looking good!

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
andig: changedIndexes should imho be handled on similar level as removedIndexes

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
andig: Understood, will update PR.

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
andig: @deeky666 not sure why- the same is being done for the `removedIndexes` https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L621?

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
andig: @deeky666 can do, but the same happens in the original DBAL-464 fix: https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php#L342
Do you want me to update both?

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
deeky666: @andig I know that it is also done for `removedIndexes` and it might be causing the same issues there. So it was done wrong before and should not be done wrong again ;)

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
deeky666: @andig yes please. The option is just invalid.

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -654,6 +654,24 @@ protected function getPreAlterTableIndexForeignKeySQL(TableDiff $diff)
...
+                    $column = $diff->fromTable->getColumn($columnName);
+
+                    if ($column->getAutoincrement() === true && in_array($columnName, $chgIndex->getColumns()) === false) {
```
andig: Thinking out loud: is `in_array` a good idea here? `Index->hasColumnAtPosition` is much more complex.
I could either add a `indexHasColumn($index, $columnName)` private function mimicking the index function or could add an `hasColumn` function to `Index` (similar to `Table`) or allow `Index->hasColumnAtPosition` to allow `null` as position for `any`.

OR it is ok as it is ;)

deeky666: You could add `Index::hasColumn()` but that is not absolutely required. I could also live with `in_array()` here ;)

andig:
> Besides my nitpicking it is looking good!

No problem at all. If you're fine with the changes I'm happy to provide a squashed commit.

andig:
btw, do you want a reference to DBAL-1132, too?

andig:
as this seems pretty complete- any chance to get it committed?

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
Ocramius: @andig it's just a question of readability

deeky666:
Thanks @andig ! Merged manually (extracted a private method in the platform and fixed some minor glitches).
Also backported to 2.5 and 2.4.

PR #787, opened 2015-01-26
Add left-over console file to the mapped composer binaries

t0xicCode:
`bin/doctrine-dbal` includes `bin/doctrine-dbal.php`, but the later is not symlinked to the `vendor/bin` directory of projects as it is not declared in composer.json. This PR declares the file as a vendor binary.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     "suggest": {
         "symfony/console": "For helpful console commands such as SQL execution and import of files."
     },
-    "bin": ["bin/doctrine-dbal"],
+    "bin": ["bin/doctrine-dbal", "bin/doctrine-dbal.php"],
     "autoload": {
         "psr-0": { "Doctrine\\DBAL\\": "lib/" }
     },


Ocramius:
@t0xicCode thanks!

stof:
This does not make sense. The symlink in ``vendor/bin`` does not change the location of the ``bin/doctrine-dbal`` itself, so requiring the file should still work. there is no reason to copy ``doctrine-dbal.php`` in the ``vendor/bin`` folder

Ocramius:
@stof the cli script was actually broken for me on windows, heh.

t0xicCode:
@stof It depends on how the symlink is interpreted. On linux (and OSX I believe), _vendor/bin/doctrine-dbal_ is a symlink, so `__DIR__` resolves relative to the original file. AFAIK, on Windows, symlinks are not handled in the same way (I don't think it even does symlinks, but proxies instead), so `__DIR__` resolves to a different directory.

stof:
@t0xicCode I don't see how the bin script would resolve ``__DIR__`` to another folder than its own folder when it is called by a proxy script. And I never got any issue with other bin scripts using ``__DIR__`` when I was on windows.
The case where it would break is if you were *copying* the script to a different location. But this would force moving the whole codebase to keep all ``__DIR__`` links working. And this is **not** what Composer does (it creates a proxy file, it does not copy the file).

and btw, given that the ``bin/doctrine-dbal.php`` file is not executable, it causes Composer to report a changed file when using source installs and symlinks because of the fact that Composer chmods the file it creates in vendor/bin to make it executable (and for symlinks, it impacts the target file)

stof:
@Ocramius this is also breaking any non-interactive composer update btw because of these changes

PR #788, opened 2015-01-26
Improved failure resilience for array type

garwat82:
The array type is the only one that throws an error if the value can't be decoded and that causes problems if the database column is empty or invalid. The patch adjusts the behavior to the rest of the implemented types.


lib/Doctrine/DBAL/Types/ArrayType.php
```diff
@@ -57,7 +57,7 @@ public function convertToPHPValue($value, AbstractPlatform $platform)
         $value = (is_resource($value)) ? stream_get_contents($value) : $value;
         $val = unserialize($value);
         if ($val === false && $value != 'b:0;') {
-            throw ConversionException::conversionFailed($value, $this->getName());
+            return array();
```
Ocramius: The failure should really not be silent.

Ocramius:
Closing as invalid: I'd rather have all other types throw exceptions instead.

PR #789, opened 2015-01-28
Fix when finding arrays of binary strings

cerealean:
## What the problem was

When using the ORM findBy() method to select data in a binary column (compared against an array of binary strings), I would get exceptions. Code example:

``` php
    /**
     * @param string[] $email_addresses Encrypted email addresses
     * @return \Project\Entity\EmailAddresses[]
     */
    public function retrieveEmailAddressEntities(array $email_addresses){
        return $this->repository->findBy(['EmailAddress' => $email_addresses]);
    }
```

When I would only search for one binary string, however, everything would work correctly. I noticed I was getting a PHP Notice for array to string conversion in my error log:

```
PHP Notice:  Array to string conversion in/var/www/html/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php on line 91
```

In my MySQL table, the column is a varbinary(255) and the ORM entity was marked as type "binary".
## How I fixed

I noticed that the `$type` in `SQLParserUtils::expandListParameters` was marking the type as 103. When this method was checking array types, it was not checking for binary strings. By simply adding the extra check for type 103 in this method, everything started working properly. Line comments below.
## Extra

This worked to fix it for me and I haven't noticed any side effects of this, however I have not extensively tested all features. If there are any problems with this please let me know.
## Specific file commit comments:

Connection.php 
- Added constant PARAM_LOB_ARRAY for binary

SQLParserUtils.php
- Added check to see if the array type is binary
- Changed the way that the foreach checked if the type was not equal to int, string, or binary

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -84,6 +84,13 @@ class Connection implements DriverConnection
     const PARAM_STR_ARRAY = 102;
 
     /**
+     * Represents an array of binary strings to be expanded by Doctrine SQL parsing.
+     *
+     * @var integer
+     */
+    const PARAM_LOB_ARRAY = 103;
+
+    /**
      * Offset by which PARAM_* constants are detected as arrays of the param type.
      *
      * @var integer
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -84,7 +84,7 @@ class Connection implements DriverConnection
     const PARAM_STR_ARRAY = 102;
 
     /**
-     * Represents an array of binary strings to be expanded by Doctrine SQL parsing.
+     * Represents an array of large objects to be expanded by Doctrine SQL parsing.
      *
      * @var integer
      */
--- a/lib/Doctrine/DBAL/SQLParserUtils.php
+++ b/lib/Doctrine/DBAL/SQLParserUtils.php
@@ -89,11 +89,6 @@ class SQLParserUtils
         $isPositional   = is_int(key($params));
         $arrayPositions = array();
         $bindIndex      = -1;
-        $relevantArrayTypes = array(
-            Connection::PARAM_INT_ARRAY,
-            Connection::PARAM_STR_ARRAY,
-            Connection::PARAM_LOB_ARRAY,
-        );
 
         if ($isPositional) {
             ksort($params);
@@ -103,7 +98,9 @@ class SQLParserUtils
         foreach ($types as $name => $type) {
             ++$bindIndex;
 
-            if (!in_array($type, $relevantArrayTypes)) {
+            if ($type !== Connection::PARAM_INT_ARRAY
+                && $type !== Connection::PARAM_STR_ARRAY
+                && $type !== Connection::PARAM_LOB_ARRAY) {
                 continue;
             }
 
--- a/lib/Doctrine/DBAL/SQLParserUtils.php
+++ b/lib/Doctrine/DBAL/SQLParserUtils.php
@@ -89,6 +89,11 @@ class SQLParserUtils
         $isPositional   = is_int(key($params));
         $arrayPositions = array();
         $bindIndex      = -1;
+        $relevantArrayTypes = array(
+            Connection::PARAM_INT_ARRAY,
+            Connection::PARAM_STR_ARRAY,
+            Connection::PARAM_LOB_ARRAY,
+        );
 
         if ($isPositional) {
             ksort($params);
@@ -98,7 +103,7 @@ class SQLParserUtils
         foreach ($types as $name => $type) {
             ++$bindIndex;
 
-            if ($type !== Connection::PARAM_INT_ARRAY && $type !== Connection::PARAM_STR_ARRAY) {
+            if (!in_array($type, $relevantArrayTypes)) {
                 continue;
             }
 
--- a/tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
@@ -21,6 +21,7 @@ class DataAccessTest extends \Doctrine\Tests\DbalFunctionalTestCase
             $table->addColumn('test_int', 'integer');
             $table->addColumn('test_string', 'string');
             $table->addColumn('test_datetime', 'datetime', array('notnull' => false));
+            $table->addColumn('test_binary', 'binary', array('notnull' => false));
             $table->setPrimaryKey(array('test_int'));
 
             $sm = $this->_conn->getSchemaManager();
@@ -441,11 +442,11 @@ class DataAccessTest extends \Doctrine\Tests\DbalFunctionalTestCase
         $this->assertEquals(array(array(100), array(101), array(102), array(103), array(104)), $data);
 
         $stmt = $this->_conn->executeQuery('SELECT test_int FROM fetch_table WHERE test_binary IN (?)',
-            array(array($binaryValue)), array(Connection::PARAM_LOB_ARRAY));
+            array(array($binaryValue)), array(Connection::PARAM_STR_ARRAY));
 
         $data = $stmt->fetchAll(PDO::FETCH_NUM);
-        $this->assertEquals(5, count($data));
-        $this->assertEquals(array(array(100), array(101), array(102), array(103), array(104)), $data);
+        $this->assertEquals(10, count($data));
+        $this->assertEquals(array(array(100), array(101), array(102), array(103), array(104), array(105), array(106), array(107), array(108), array(109)), $data);
     }
 
     /**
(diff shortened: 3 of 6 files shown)


lib/Doctrine/DBAL/Connection.php
```diff
@@ -84,6 +84,13 @@ class Connection implements DriverConnection
     const PARAM_STR_ARRAY = 102;
 
     /**
+     * Represents an array of binary strings to be expanded by Doctrine SQL parsing.
+     *
+     * @var integer
+     */
+    const PARAM_LOB_ARRAY = 103;
```
cerealean: Added for binary type

lib/Doctrine/DBAL/SQLParserUtils.php
```diff
@@ -98,7 +103,7 @@ static public function expandListParameters($query, $params, $types)
         foreach ($types as $name => $type) {
             ++$bindIndex;
 
-            if ($type !== Connection::PARAM_INT_ARRAY && $type !== Connection::PARAM_STR_ARRAY) {
+            if (!in_array($type, $relevantArrayTypes)) {
```
cerealean: Instead of making this `if` extra long and checking 3 different times, I placed the types into an array and am doing a `!in_array()` call

deeky666:
Thanks for revealing this issue! Seems it was not on our radar while implementing binary types.
However there are some things to do before it can be merged. First of all it needs test cases. You should add tests [here](https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/SQLParserUtilsTest.php#L74), [here](https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/SQLParserUtilsTest.php#L358), [here](https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php#L416) and [here](https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/Functional/NamedParametersTest.php#L11).
Additionally you need to add the new constant to `Doctrine\DBAL\Connection` also.
Last but not least the new constant has to be added to the documentation [here](https://github.com/doctrine/dbal/blob/master/docs/en/reference/data-retrieval-and-manipulation.rst).
Otherwise the patch is good.

lib/Doctrine/DBAL/SQLParserUtils.php
deeky666: I agree that using `in_array()` might be cleaner code but unfortunately this method here is performance critical so it is better to add the third comparison instead.

cerealean:
Awesome @deeky666 thanks for pointing that out. I'll get those changes made and committed to this pull request today.

deeky666:
@cerealean ah sorry thought that constant was added in `SQLParserUtils`. Nevermind about that note then :)

cerealean:
Still in progress of adding tests. Just pushing in-between other work. I'll message here when finished.

deeky666:
@cerealean no rush :)

cerealean:
Hey @deeky666 I'm having some troubles. I was curious if you could provide another set of eyes and some suggestions for this.

### DataAccessTest.php
- I am having to use `Connection::PARAM_STR_ARRAY` instead of `Connection::PARAM_LOB_ARRAY`, otherwise I am getting errors. 
- https://github.com/doctrine/dbal/pull/789/files#diff-9409d2bac73b5a78e56ab0eb2dd628f5R445 

### NamedParametersTest.php
- Similar error to the DataAccessTest.php error, I am being forced to use string array type.
    - I can use `Connection::PARAM_LOB_ARRAY` with MySQL databases by using the BINARY keyword in the SQL
    - Example: `SELECT * FROM ddc1372_foobar f WHERE f.foo = :foo AND f.bar IN (:bar) AND BINARY f.baz = :baz` works for mysql, but without the `BINARY` keyword it does not (unless I use `Connection::PARAM_STR_ARRAY`). This also does not transfer to other databases since this is MySQL specific.
- I have also tried using resources with binary strings inside of them (instead of using the PHP native `pack()` method), however still no luck

Any feedback you have would be great. Thanks!

tests/Doctrine/Tests/DBAL/Functional/NamedParametersTest.php
```diff
@@ -10,18 +10,28 @@
...
+                    array('id'=>1,'foo'=>1,'bar'=>1,'baz'=>$binaryData),
+                    array('id'=>2,'foo'=>1,'bar'=>2,'baz'=>$otherBinary),
+                    array('id'=>3,'foo'=>1,'bar'=>3,'baz'=>$binaryData),
+                )
+            ),
+
+            array(
+                'SELECT * FROM ddc1372_foobar f WHERE f.foo = :foo AND f.bar IN (:bar) AND f.baz = :baz',
+                array('foo' => 1,'bar' => array(1, 2, 3), 'baz' => $otherBinary),
+                array('foo'=>PDO::PARAM_INT,'bar'=>Connection::PARAM_INT_ARRAY, 'baz'=>PDO::PARAM_STR),
```
cerealean: This is an example of where I'm currently having difficulties. This line should read:
```php
array('foo'=>PDO::PARAM_INT,'bar'=>Connection::PARAM_INT_ARRAY, 'baz'=>PDO::PARAM_LOB),
```
However that does not work. I am being forced to use STR instead, even though `$otherBinary` and `$binaryData` are both binary strings.

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
```diff
@@ -432,6 +440,13 @@ public function testNativeArrayListSupport()
         $data = $stmt->fetchAll(PDO::FETCH_NUM);
         $this->assertEquals(5, count($data));
         $this->assertEquals(array(array(100), array(101), array(102), array(103), array(104)), $data);
+
+        $stmt = $this->_conn->executeQuery('SELECT test_int FROM fetch_table WHERE test_binary IN (?)',
+            array(array($binaryValue)), array(Connection::PARAM_STR_ARRAY));
```
deeky666: Looks like you will have to insert the data with types beforehand, otherwise the binary data won't be inserted properly:

```php
for ($i = 100; $i < 110; $i++) {
    $this->_conn->insert('fetch_table', array(
        'test_int'      => $i,
        'test_string'   => 'foo' . $i,
        'test_datetime' => '2010-01-01 10:10:10',
        'test_binary'   => $binaryValue
    ), array(
        'test_binary' => \PDO::PARAM_LOB
    ));
}
```

tests/Doctrine/Tests/DBAL/Functional/NamedParametersTest.php
deeky666: Probably the same issue as with `DataAccessTest`. Try inserting the binary data with `\PDO::PARAM_LOB`.

cerealean: That's what I am wanting to do, however I was getting issues unless I used `\PDO::PARAM_STR` which did not make sense

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
cerealean: Ah now this makes a little more sense. Let me give this a try and see if it helps any.

cerealean:
Alright @deeky666, your extra help with the inserting as type `\PDO::PARAM_LOB` fixed most of my errors in testing. I think all is well now. Please let me know if you would like any other tests added or anything else done. Thanks!

tests/Doctrine/Tests/DBAL/Functional/NamedParametersTest.php
```diff
@@ -154,6 +177,9 @@ public function testTicket($query,$params,$types,$expected)
         $result = $stmt->fetchAll(\PDO::FETCH_ASSOC);
 
         foreach ($result as $k => $v) {
+            if(is_resource($v["baz"])){
```
cerealean: Have to add this check for postgresql. LOBs come back as a resource.

cerealean:
Just stating that this is ready from my perspective unless you need anything else added @deeky666 . Not trying to rush, just wanted to be sure that it was clear.

deeky666:
@cerealean no worried, I have it on my list. Just don't have much time at the moment to take care of PRs. I did not forget you ;)

cerealean:
:+1: awesome!

cerealean:
Hi @deeky666 . Once again not trying to rush you as I'm sure you're busy with a lot of other things, however I was wondering what the timeline for this was if you had an estimate? I actually made this fix so that we can use Doctrine 2 at our company for projects and one of those projects is about to go live. Currently I have that project pointing at my fork as a temporary solution, however I know they will want me to have a more permanent solution in the near future. Thanks!

Ocramius:
Just a note: seems like a 2.6.0 addition to me.

cerealean:
Hmm ok. Wasn't sure if this had to wait for that version since (from my understanding) this functionality was intended to be in the current release.

morozov:
Closing as stale.

resolving-ngi:
This is still relevant

morozov:
@resolving-ngi please feel free to resubmit against `2.11.x`.

PR #790, opened 2015-01-29
[DBAL-1137] Fix infinite recursion on non-unique table/join alias in QueryBuilder

JarJak:
What do you guys think about this solution? I found this problem when got not unique join alias - getSQLForJoins went for infinite recursion and apache crashed without promt..

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -1308,9 +1308,9 @@ class QueryBuilder
         $sql = '';
 
         if (isset($this->sqlParts['join'][$fromAlias])) {
-			if (array_key_exists($join['joinAlias'], $knownAliases)) {
-				throw QueryException::aliasNotUnique($join['joinAlias']);
-			}
+            if (array_key_exists($join['joinAlias'], $knownAliases)) {
+                throw QueryException::notUniqueAlias($join['joinAlias'], $knownAliases);
+            }
             foreach ($this->sqlParts['join'][$fromAlias] as $join) {
                 $sql .= ' ' . strtoupper($join['joinType'])
                     . ' JOIN ' . $join['joinTable'] . ' ' . $join['joinAlias']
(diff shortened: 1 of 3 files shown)


Ocramius:
@jarekj requires a test case

jarekj:
here it is: https://github.com/doctrine/dbal/pull/791

deeky666:
@jarekj the next time please just add your test to the existing PR by committing and pushing to the same branch again.

jarekj:
@derky666 this test just shows that the problem exists, running this test in my fix branch I get:
Fatal error: Method Doctrine\DBAL\Query\QueryBuilder::__toString() must not throw an exception

stof:
@jarekj and this shows you that your PR is wrong, which is actually a good thing too

deeky666:
@stof not really. In his PR he already fixed the recursion issue by throwing an exception which would explain the error message he gets...

stof:
@deeky666 but the exception is thrown in the wrong place, as throwing it inside ``__toString`` triggers a PHP Fatal Error. ``__toString`` is not allowed to throw exceptions in PHP

deeky666:
@stof you are right. But the QueryBuilder also throws other exceptions in `__toString()` / `getSQL()` I'm not even sure why one would use `__toString()` to generate the SQL from the QueryBuilder anyways. Always found that weird...

stof:
@deeky666 I also find it weird, and it should probably be removed in 3.0

jarekj:
test fails for some unrelated reason i think

deeky666:
@jarekj merged manually with some fixes and additions. Thanks!

jarekj:
I am not sure this is right solution, I have tested it locally and found some false positives

deeky666:
@jarekj what do you mean exactly by "false positives"? The test suite runs fine. Anything we have missed?

jarekj:
Maybe we don't have enough tests written...

deeky666:
Feel free to provide failing test cases in another PR if you have any.

jarekj:
I am not sure what $knownAliases array is for - it could be changed to count aliases and throw the Exception when any alias comes more than once.

jarekj:
@deeky666 see https://github.com/doctrine/dbal/pull/794 for more tests and throw exception in another place

PR #791, opened 2015-01-30
Failing test case for DBAL-1137

JarJak:
test shows out of memory

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
+++ b/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
@@ -863,4 +863,23 @@ class QueryBuilderTest extends \Doctrine\Tests\DbalTestCase
 
         $this->assertSame(array('name' => \PDO::PARAM_STR, 'isActive' => \PDO::PARAM_BOOL), $qb->getParameterTypes());
     }
+
+    /**
+     * @group DBAL-790
+     */
+    public function testSelectWithJoinsWithNotUniqueAliases()
+    {
+        $qb = new QueryBuilder($this->conn);
+
+        $qb->select('a.id, b.id')
+            ->from('table_a', 'a')
+            ->join('a', 'table_b', 'b', 'a.fk_b = b.id')
+            ->join('a', 'table_b', 'b', 'a.fk_b = b.id')
+            ->join('b', 'table_a', 'a', 'a.fk_b = b.id');
+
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
```diff
@@ -863,4 +863,23 @@ public function testGetParameterTypes()
...
+        $qb->select('a.id, b.id')
+            ->from('table_a', 'a')
+            ->join('a', 'table_b', 'b', 'a.fk_b = b.id')
+            ->join('a', 'table_b', 'b', 'a.fk_b = b.id')
+            ->join('b', 'table_a', 'a', 'a.fk_b = b.id');
+
+        $this->assertEquals(
+            'dumb string',
+            (string) $qb
+        ); //you will get out of memory anyway
```
stof: your assertion is wrong, because it will not pass once the bug is fixed

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
```diff
@@ -863,4 +863,23 @@ public function testGetParameterTypes()
 
         $this->assertSame(array('name' => \PDO::PARAM_STR, 'isActive' => \PDO::PARAM_BOOL), $qb->getParameterTypes());
     }
+
+    /**
+     * @group DBAL-790
+     */
+    public function testSelectWithJoinsWithNotUniqueAliases()
```
stof: is the issue about a not unique alias or about a circular reference between a and b ?

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
deeky666: Would be best to add the test to your patch PR and assert that the `QueryException`  is thrown.

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
jarekj: circular reference is caused by not unique alias

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
jarekj: ok I will change that

PR #792, opened 2015-01-30
getSQLForJoins - infinite recursion if join alias is not unique

JarJak:
fix for exception message

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryException.php
+++ b/lib/Doctrine/DBAL/Query/QueryException.php
@@ -49,6 +49,6 @@ class QueryException extends DBALException
     {
         return new self("The given alias '" . $alias . "' is not unique " .
             "in FROM and JOIN clause table. The currently registered " .
-            "aliases are: " . implode(", ", $registeredAliases) . ".");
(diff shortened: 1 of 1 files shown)


Ocramius:
Requires a test case, IMO

jarekj:
it just fixes exception message from my previous pull request, values of this array are all booean true

Ocramius:
Yes, but a test verifies the fix and prevents a regression later on.

deeky666:
I already applied that fix here https://github.com/doctrine/dbal/commit/7fec259f3fdfa86f3c5441b2583bd8c1a3f1d69a while merging your last PR.

jarekj:
Thanks @deeky666 but I'm still not sure this is the best solution. I've got false positive in my project. I'll try to prepare test case that wiould produce it too.

jarekj:
@deeky666 OK I got it. I was not false positive, I was doing something like:
$qb->from('table_a', 'a')->from('table_a', 'a')->from('table_a', 'a')->join('a', 'table_b', 'b')
Then it throwed the exception, which is quite accurate:
The given alias 'a' is not unique in FROM and JOIN clause table. The currently registered aliases are: a, b.
I can create test case for it.

PR #793, opened 2015-01-30
fix SequenceName for Oracle

Gemorroj:
I want to use a new feature, specify the oracle IDENITY generator (http://www.doctrine-project.org/jira/browse/DDC-2875) . But after inserting a new row, doctrine causes the wrong sequence to find out the id of the inserted row. My problem is solved by adding processing the column name.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
@@ -934,7 +934,7 @@ END;';
         $table = new Identifier($tableName);
 
         // No usage of column name to preserve BC compatibility with <2.5
-        $identitySequenceName = $table->getName() . '_SEQ';
+        $identitySequenceName = $table->getName() . ((string)$columnName !== '' ? '_' . $columnName : '') . '_SEQ';
 
         if ($table->isQuoted()) {
             $identitySequenceName = '"' . $identitySequenceName . '"';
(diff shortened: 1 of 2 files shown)


Ocramius:
@Gemorroj this change requires a test case: what is being fixed is also quite unclear

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -932,9 +932,12 @@ public function usesSequenceEmulatedIdentityColumns()
     public function getIdentitySequenceName($tableName, $columnName)
     {
         $table = new Identifier($tableName);
+        $column = new Identifier($columnName);
 
         // No usage of column name to preserve BC compatibility with <2.5
-        $identitySequenceName = $table->getName() . '_SEQ';
+        // ?????????
+
+        $identitySequenceName = $table->getName() . ($column->getName() !== '' ? '_' . $column->getName() : '') . '_SEQ';
```
deeky666: Unfortunately we cannot do this (see comment about BC above). We included the column name in the sequence naming during the development of 2.5 but decided to revert it to stay backwards compatible. Otherwise existing application will break when upgrading DBAL.

Gemorroj: That is, I can not count that will be supported titles sequence?

deeky666: @Gemorroj you have to clarify about the actual error you encounter. Identity sequences created by Doctrine should be detected and working. What is the name of the sequence in your DB and what sequence name is DBAL generating? What are the steps to reproduce your error here?

morozov:
Closing as stale.

PR #794, opened 2015-01-30
Added more tests to QueryBuilder

JarJak:
just testing if tests will fail

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -1140,6 +1140,9 @@ class QueryBuilder
                 $tableSql = $from['table'];
                 $tableReference = $from['table'];
             } else {
+                if (array_key_exists($from['alias'], $knownAliases)) {
+                    throw QueryException::nonUniqueAlias($from['alias'], array_keys($knownAliases));
+                }
                 $tableSql = $from['table'] . ' ' . $from['alias'];
                 $tableReference = $from['alias'];
             }
--- a/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
+++ b/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
@@ -904,10 +904,10 @@ class QueryBuilderTest extends \Doctrine\Tests\DbalTestCase
 
         $qb->select('*')
             ->from('table_a', 'a')
-            ->andFrom('table_b', 'b')
+            ->From('table_b', 'b')
             ->join('a', 'table_c', 'c', 'a.fk_c = c.id')
             ->leftJoin('a', 'table_b', 'b', 'a.fk_b = b.id')
-            ->where('a.fk_b = b.id');
+            ->andWhere('a.fk_b = b.id');
 
         $sql = (string) $qb;
         $this->assertFalse((bool) $sql);
--- a/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
+++ b/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
@@ -882,4 +882,34 @@ class QueryBuilderTest extends \Doctrine\Tests\DbalTestCase
 
         $qb->getSQL();
     }
+
+    public function testJoinsWithSameAliases()
+    {
+        $qb = new QueryBuilder($this->conn);
+
+        $qb->select('*')
+            ->from('table_a', 'a')
+            ->andFrom('table_b', 'b')
+            ->join('a', 'table_c', 'c', 'a.fk_c = c.id')
+            ->leftJoin('a', 'table_b', 'bb', 'a.fk_b = bb.id')
+            ->where('a.fk_b = b.id');
+
+        $sql = (string) $qb;
+        $this->assertTrue((bool) $sql);
+    }
+
+    public function testJoinsWithSameDuplicateAliases()
+    {
+        $qb = new QueryBuilder($this->conn);
+
+        $qb->select('*')
+            ->from('table_a', 'a')
+            ->andFrom('table_b', 'b')
+            ->join('a', 'table_c', 'c', 'a.fk_c = c.id')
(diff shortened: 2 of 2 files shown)


jarekj:
@deeky666 could you have a look at these tests? Looks like SQL output is wrong when there are more than 2 WHERE statements and a JOIN between them.

jarekj:
This is what is expected:
SELECT a.id, b.id, c.id FROM table_a a, **table_b b** INNER JOIN table_c c ON a.fk_c = c.id WHERE a.fk_b = b.id
This is what comes up:
SELECT a.id, b.id, c.id FROM table_a a INNER JOIN table_c c ON a.fk_c = c.id, **table_b b** WHERE a.fk_b = b.id

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
```diff
@@ -882,4 +882,73 @@ public function testJoinWithNonUniqueAliasThrowsException()
...
+    {
+        $qb = new QueryBuilder($this->conn);
+
+        $qb->select('a.id, b.id, c.id')
+            ->from('table_a', 'a')
+            ->from('table_b', 'b')
+            ->join('a', 'table_c', 'c', 'a.fk_c = c.id')
+            ->where('a.fk_b = b.id');
+
+        $this->assertSame('SELECT a.id, b.id, c.id FROM table_a a, table_b b INNER JOIN table_c c ON a.fk_c = c.id WHERE a.fk_b = b.id', $qb->getSQL());
```
deeky666: This SQL is wrong as table alias `a` is unknown in the `INNER JOIN` clause. You will get an error from the database. `table_a` is a different base table and therefore out of scope of the join clause as it uses `table_b` as base table.

jarekj: Shouldn't it be an error from QueryBuilder not database?

jarekj: From docs:
"A join always belongs to one part of the from clause. This is why you have to specify the alias of the FROM part the join belongs to as the first argument."
Why would the join cluase use table_b as base table if I specify 'a' alias?

Ocramius: > Shouldn't it be an error from QueryBuilder not database?

The query builder is a stupid string assembler: it doesn't nor has to check the validity of the SQL.

jarekj:
added tests mentioned in https://github.com/doctrine/dbal/pull/792

jarekj:
ready to merge

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
```diff
@@ -882,4 +882,41 @@ public function testJoinWithNonUniqueAliasThrowsException()
...
+        $qb = new QueryBuilder($this->conn);
+
+        $qb->select('a.id, b.id, c.id')
+            ->from('table_a', 'a')
+            ->from('table_b', 'b')
+            ->join('a', 'table_c', 'c', 'a.fk_c = c.id')
+            ->leftJoin('a', 'table_b', 'b', 'a.fk_b = b.id')
+            ->andWhere('a.fk_b = b.id');
+
+        $this->setExpectedException(
```
Ocramius: I think the expectation should be set before constructing the QueryBuilder, as we may change the point in time when the exception is thrown

jarekj: changed

jarekj:
@deeky666 is it going to be merged?

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
```diff
@@ -869,16 +869,53 @@ public function testGetParameterTypes()
...
+            ->from('table_a', 'a')
+            ->from('table_b', 'b')
+            ->join('a', 'table_c', 'c', 'a.fk_c = c.id')
+            ->leftJoin('a', 'table_b', 'b', 'a.fk_b = b.id')
+            ->andWhere('a.fk_b = b.id');
+
+        $qb->getSQL();
+    }
+
+    public function testDuplicateFromAliasThrowsException()
```
deeky666: @Ocramius do you know whether it was expected behaviour that `FROM` aliases may be overridden? Because changing the current behaviour to throw an exception in that case might break existing applications...

Ocramius: @deeky666 it may, yes, I'm just being proactive on the test approach here

deeky666: Sorry but I am not talking about the test here. I am talking about the change in QueryBuilder that throws an exception now (the comment you rather have been there).

Ocramius: @deeky666 that was not factored in, but it may be in future

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1140,6 +1140,9 @@ private function getFromClauses()
                 $tableSql = $from['table'];
                 $tableReference = $from['table'];
             } else {
+                if (array_key_exists($from['alias'], $knownAliases)) {
+                    throw QueryException::nonUniqueAlias($from['alias'], array_keys($knownAliases));
```
deeky666: You will have to revert this. This is a BC break for applications that rely on the current behaviour where defining a `FROM` clause with the same alias will overwrite the previous one. We cannot change this behaviour for now, even though it might have been a minor mistake in designing the query builder.

jarekj: But it could be included like in Symfony 3.x?

Ocramius: @jarekj we didn't plan anything for 3.x yet

deeky666:
Closing for now because of BC break. Can be reevaluated for 3.0.

PR #795, opened 2015-01-31
Add tsvector type support in PostgreSQL Platform

believer-ufa:
Just one simple string. Maybe add this to code?

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -1077,6 +1077,7 @@ class PostgreSqlPlatform extends AbstractPlatform
             'char'          => 'string',
             'bpchar'        => 'string',
             'inet'          => 'string',
+            'tsvector'      => 'string',
             'date'          => 'date',
             'datetime'      => 'datetime',
             'timestamp'     => 'datetime',


deeky666:
I am not familiar with the `tsvector` type. What are the semantics of that type? Isn't rather a `text` type instead of `string`? To be more precise: Is it limited to a certain amount of characters like `VARCHAR` or does it have length limitations like `TEXT`?
Also when retrieving data of such a column type via DBAL, will the driver return a plain PHP string or does it return some other type like a binary stream or something because I read it is stored as a vectorish type internally in the database?

believer-ufa:
Hello)

Its a special type in PgSQL for storing information about some strings for future searching in them through fulltext search technique. Read more about this you can in official docs: http://www.postgresql.org/docs/9.4/static/datatype-textsearch.html

But for me and you, simple humans, receiving data from this field uniquely retrieve data from a "string" or "text" field. Most likely, yes, there is more correct to set the type as "text", not "string", because data in this field can be long.

And sorry if I suggest that something is not right. I am not a great expert in Doctrine, I just need to specify this type of field in one of their projects for one of the libraries, which receives information from the database through Doctrine.

deeky666:
@believer-ufa thanks for clarification. No worries. Can you change the mapping to the more appropriate `text` type then?
Also we need a test case for this to avoid regressions in the future. You should add a test case to `Doctrine\Tests\DBAL\Platforms\AbstractPostgreSqlPlatformTestCase`. You can adopt it from [this test case](https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/Platforms/PostgreSQL92PlatformTest.php#L54-L58).
Afterwards I'll merge. Thanks.

PR #797, opened 2015-01-31
unsigned boolean

liutec:


deeky666:
This does not make a lot of sense to me. Why would you want to allow the user define an unsigned option for a boolean type column? I could however agree to force unsigned on MySQL's boolean type declaration (TINYINT) to disallow negative values on the database side.
Can you please clarify your use case / issue?

liutec:
If the column already exists in the table as unsigned, when importing the mapping (metadata) it will be imported with the option unsigned. If afterwards doctrine:schema:update is issued then the column will be altered to signed. If afterwards doctrine:schema:validate is issued it will fail because it will see the column as unsigned in the metadata and as signed in the table. This case will forever see the field as changed and attempt to alter it in doctrine:schema:update.

liutec:
This was not an issue before the mapping import was fixed: https://github.com/doctrine/doctrine2/commit/00eb0d3b67f19c98f5949bfceac713f0a454ff02 because the unsigned flag was not mapped correctly to the options key and thus completely ignored.

deeky666:
I get the problem now, however this is not the appropriate fix for it. This should rather be fixed in the Comparator I suppose, so that the comparator does not compare the unsigned option for BooleanType columns.

liutec:
If the comparator does not compare the unsigned option for boolean type then the schema will remain out of sync with the mapping - plus the comparator is meant to be type independent.

Integer representation of boolean is possible for both signed and unsigned integers. 

From what I know, there are two integer values that correspond to True based on whether the integer is signed or unsigned. For signed integers -1 is True because the sign bit is also negated, for unsigned integers is 1. (holds true in php as well http://php.net/manual/en/language.types.boolean.php)

skafandri:
What about?

````php
    public function getBooleanTypeDeclarationSQL(array $field)
    {
        return 'TINYINT(1)'.$this->_getCommonIntegerTypeDeclarationSQL($field);
    }
````
To reuse existing code, since the same check is made in `_getCommonIntegerTypeDeclarationSQL`
` $unsigned = (isset($columnDef['unsigned']) && $columnDef['unsigned']) ? ' UNSIGNED' : '';`

liutec:
@skafandri Thx. You're right

deeky666:
I'm sorry but that solution is not acceptable. `unsigned` option simply is not applicable for an abstracted BooleanType. It should be ignored by the comparator and that's it. Evaluating `unsigned` option for MySQL's BooleanType declaration SQL is weird. If MySQL had a native `BOOLEAN` type, we would not do that either. `TINYINT` really is just a fallback type here because of MySQL lacking a native `BOOLEAN` type.
The schema remaining unsynced is not an argument here IMO as you should not rely on the `unsigned` attribute anyways as DBAL is meant to be portable and thus `unsigned` is not applicable to all platforms. If you have `unsigned` option defined in your mapping for a `boolean` type column, you are using Doctrine in a wrong way...

liutec:
If you find it unacceptable then don't accept it. I'm really getting tired of people wasting my time just for the sake of pointless arguments. If it's meant to be portable then this solution does just that: takes boolean as it is and leaves it as it is.

deeky666:
No need to get offensive here. I totally get the problem, I am just saying that the proposed soultion is wrong in a DBAL sense IMO. I would be absolutely fine with fixing the comparator though...

liutec:
I do get offended because I gave you a very strong argument (see above) which I don't think you've read: I see no reason why boolean should be treated as a special case either in the comparator or here -- especially in the comparator which has to deal with ALL data types. It's an abstract type which may be mapped in certain engines both as signed as well as unsigned. An abstraction layer should just take it as-is as long as it is valid in the language of the DBAL (PHP).

PR #799, opened 2015-02-09
Added option to set connect_timeout setting in PDOPgSql dsn

bashofmann:
See http://php.net/manual/en/function.pg-connect.php


beberlei:
@bashofmann This is PDO PgSQL, does the same option work here that is for `pg_connect`?

bashofmann:
I would have thought since it is part of the dsn PDO just forwards it, but after a bit of more testing and reading up in the PHP src it does not but overwrites connect_timeout with PDO_ATTR_TIMEOUT. So there seems to be now way to set this independently with PDO :(

PR #801, opened 2015-02-10
Add deprecated annotation to renameColumn method

driesvints:
I added the deprecated annotation to warn a dev that the method isn't working anymore. This will allow a dev to spot the deprecated method before he actually attempts to use it.

screen shot 2015-02-10 at 09 55 50

Should there be some comment added or perhaps a referring to the exception message?

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Table.php
+++ b/lib/Doctrine/DBAL/Schema/Table.php
@@ -328,6 +328,8 @@ class Table extends AbstractAsset
     /**
      * Renames a Column.
      *
+     * @deprecated Please use a manual SQL ALTER TABLE query to rename a column.
+     *
      * @param string $oldColumnName
      * @param string $newColumnName
      *


lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -328,6 +328,8 @@ public function addColumn($columnName, $typeName, array $options=array())
     /**
      * Renames a Column.
      *
+     * @deprecated
```
stof: There should be an explanation of the deprecation pointing to what to do instead (just like we should never throw an exception with an empty message for instance)

deeky666:
I don't think this method is deprecated but rather outdated. Renaming columns is possible since quite some time now (the comparator can detect that).
So rather than marking this method as deprecated, it should be fixed to allow renamings again (just like we did for indexes).

lib/Doctrine/DBAL/Schema/Table.php
driesvints: @stof updated.

driesvints:
@deeky666 I agree but in the meantime this is a good warning for the developer to not use the method.

At the moment a dev will only notice that the `renameColumn` method isn't working when he has already written his migration and attempts to run it.

driesvints:
That being said, it would be nice if we had an `@outdated` annotation or something similar.

deeky666:
@driesvints you cannot make the method deprecated if it will not be removed. So IMO annotating it here like this is wrong.

driesvints:
@deeky666 ok, I understand your reasoning. I hope you guys can fix the method at some point.

Thanks for your feedback both :)

PR #802, opened 2015-02-19
Fix issue where schema diffs on sql server try and fail to drop nonexistent indexes

billschaller:
When running a schema diff on SQL Server, the diff generated includes commands to drop indexes that don't exist. This doesn't fix that problem.

This patch works around the problem by changing sql generated like this:

``` sql
IF EXISTS (SELECT * FROM sysobjects WHERE name = 'IDX_1234567890')
  ALTER TABLE sometable DROP CONSTRAINT IDX_1234567890
ELSE
  DROP INDEX IDX_1234567890 ON sometable
```

to this:

``` sql
IF EXISTS (SELECT * FROM sysobjects WHERE name = 'IDX_1234567890')
  ALTER TABLE sometable DROP CONSTRAINT IDX_1234567890
ELSE IF EXISTS (SELECT name FROM sysindexes WHERE name = 'IDX_1234567890')
  DROP INDEX IDX_1234567890 ON sometable
```

Checking for the existence of the index to be dropped before trying to drop it.

The root of the problem, however, is that when the "from" schema is hydrated from schema-details via Table::__construct, a unique index is added for @JoinColumns that are flagged as unique. This also happens for joined table inheritance child tables.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -192,7 +192,7 @@ class SQLServerPlatform extends AbstractPlatform
 
         return "IF EXISTS (SELECT * FROM sysobjects WHERE name = '$index')
                     ALTER TABLE " . $table . " DROP CONSTRAINT " . $index . "
-                ELSE
+                ELSE IF EXISTS (SELECT name FROM sysindexes WHERE name = '$index')
                     DROP INDEX " . $index . " ON " . $table;
     }
 


lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -192,7 +192,7 @@ public function getDropIndexSQL($index, $table = null)
 
         return "IF EXISTS (SELECT * FROM sysobjects WHERE name = '$index')
```
deeky666: What the heck about this `IF EXISTS` SQL oO wasn't aware that we have this in the platform... Why exactly is that necessary? Looks really hackish. My opinion is that if I call this method I already know that the index exists... Can you clear me up? =)

zeroedin-bill: It seems that when a Table object is instantiated and hydrated from the schema-details, there are some conditions that cause indexes that don't actually exist (implicit indexes?) to be recorded on the Table object. This is ok in some cases, as the generated "to" schema usually has the implicit indexes as well.

One scenario where this messes up is when the primary key of an entity is also a foreign key to the id value of another entity, but the primary key that exists in the database includes a second column. Ideally, the schema diff should convert the existing primary key to only reference the single column, to match the entity. Instead, it tries to drop a nonexistent index with a generated name.

The same thing happens when a property flagged is flagged in the entity as a OneToOne relation. If a FK constraint already exists in the database, but no index exists for the column, the generated schema has a unique index for the column that does not match up with the implicit index generated when the existing schema is loaded. Thus, the schema diff tries to drop a nonexistent implicit index, and create a new unique index for the column. 

In the above scenario, it would be pretty much OK if the generated SQL would check if the nonexistent index exists before trying to delete it.

deeky666: @zeroedin-bill still this is not an issue of the platform IMO as the platform more or less acts as an independent SQL generator (mostly). The platform does not know about the schema introspection / schema tool. Sounds to me as if this should rather be fixed in the schema manager. Using `IF EXISTS` clauses in the generated SQL is not desired IMO as it silences errors in situations where dropping an index might be expected to fail if it does not exist.
Unfortunately we already have that clause in here which I was not aware of...

zeroedin-bill: @deeky666 you're right, but the sql generation for schema changes on SQL server has had a pretty rocky history. Schemas originally generated by doctrine are usually OK, but there are many issues when you build doctrine mappings for an existing schema and then attempt to generate a schema update.

As it stands, the SQL generated in the scenarios I outlined fails reliably, but shouldn't. So either the schema tool should be fixed, or the platform should be fixed to fail gracefully when it is provided garbage input.

stof: I would prefer fixing the schema tool actually

zeroedin-bill: @stof Ditto - I'll see if I can get some of the failure scenarios fixed or at least tested for.

PR #805, opened 2015-02-22
Allow symfony 3.0 components

nicolas-grekas:
Tests should tell if any deprecated interfaces of Symfony are used. If not, then the bundle is defacto compatible with 3.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,8 +16,8 @@
         "doctrine/common": ">=2.4,<2.6-dev"
     },
     "require-dev": {
-        "phpunit/phpunit": "4.*",
-        "symfony/console": "2.*"
+        "phpunit/phpunit": "~4",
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Please note that without this patch, symfony can not be tested properly. Please tell we want can be done to merge soon.

deeky666:
@nicolas-grekas I don't fully get how that affects our testsuite. AFAIK the only component that is using the Symfony Console component is the DBAL cli tool but that is not unit tested anyways so what's the point here? Please clarify :)

nicolas-grekas:
I was wrong: since this is in require-dev, and require-dev is root only, it does not affect symfony/symfony

deeky666:
@nicolas-grekas so can it be closed then?

PR #806, opened 2015-02-23
Fix broken functional test for SQL server

billschaller:
Fixes default constraints test for SQL server -- table name was wrong.


deeky666:
@zeroedin-bill thanks!

PR #808, opened 2015-02-28
Compatibility layer for PHP installations without PDO support

adamziel:
Even though DBAL currently offers non-PDO drivers, it depends on a number of PDO constants which renders it unusable if PHP was explicitly compiled without PDO. This PR is an attempt to shim PDO constants as stated in ticket DBAL-1156 (http://www.doctrine-project.org/jira/browse/DBAL-1156)

### DIFF
--- a/lib/Doctrine/DBAL/Compat/PDO.php
+++ b/lib/Doctrine/DBAL/Compat/PDO.php
@@ -1,0 +1,592 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+/**
+ * PDO shim to keep DBAL compatible with non-PDO PHP installations.
+ *
+ * @link http://php.net/manual/en/pdo.constants.php
+ * @package Doctrine\DBAL
+ */
+class PDO
+{
+
+    /**
+     * Represents a boolean data type.
+     *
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Cache/ArrayStatement.php
```diff
@@ -20,7 +20,7 @@
 namespace Doctrine\DBAL\Cache;
 
 use Doctrine\DBAL\Driver\ResultStatement;
-use PDO;
+use Doctrine\DBAL\PDOCompat;
```
Ocramius: Why not simply providing a `PDO` class via something like following? Much simpler, much better compatibility:

```json
{
    "autoload": {
        "classmap": [
            "path/to/PDO.php"
        ]
    }
}
```

AZielinski:
I updated this pull request to match your suggestion

lib/Doctrine/DBAL/Compat/PDO.php
```diff
@@ -0,0 +1,592 @@
...
+ * <http://www.doctrine-project.org>.
+ */
+
+/**
+ * PDO shim to keep DBAL compatible with non-PDO PHP installations.
+ *
+ * @link http://php.net/manual/en/pdo.constants.php
+ * @package Doctrine\DBAL
+ */
+class PDO
```
stof: defining this class with only the constants is a very bad idea IMO. It means that any project detecting the availability of PDO through a class_exists check will now think that PDO is available while it is not.

Ocramius: @stof I think that's not a big problem for us.
While I agree with you on the risks, it is still for the little percentage of projects that want to use Doctrine DBAL *without* PDO, so the crowd being affected is actually pretty small.

stof: Well, it means that DBAL is now saying ``F*** you`` to any library trying to detect whether PDO is available to choose their implementation, thus forbidding users to use such library and DBAL together in a project. This seems a bad idea IMO, because the responsible for this incompatibility will be us

AZielinski: @stof actually I modified it to this form after @Ocramius specifically asked. Initially this PR was an additional class `Doctrine\DBAL\PDOCompat` and rest of the DBAL refactored to use it instead of `\PDO` - what do you think about this approach?

Ocramius: @stof the risk seems to be acceptable, tbh

deeky666: @stof @adamziel @Ocramius can't we somehow make the registration of the sim class conditional on whether PDO is installed or not? Maybe by bootstrapping composer autoloader somehow (don't have enough knowledge about this).

Ocramius: @deeky666 looking again at this, I think it isn't worth supporting this approach anymore: we should just make `ext-pdo` a hard dependency and call it "done".

While removing the dependency is a nice-to-have, `PDO` is probably the ugliest API we can polyfill, riddled with poor design decisions, strange method signatures and behaviors.

Ocramius (changes requested):

morozov:
> I think it isn't worth supporting this approach anymore: we should just make `ext-pdo` a hard dependency and call it "done".

Why? Doctrine doesn't use anything from PDO besides constants and even this dependency can be eliminated by getting rid of usage of `PDO::` constants in favor of Doctrine's own. If an app uses DBAL and doesn't support PDO it will have to require it even if it doesn't support it.

> While removing the dependency is a nice-to-have, PDO is probably the ugliest API we can polyfill, riddled with poor design decisions, strange method signatures and behaviors.

Why is it needed to polyfill anything else then constants from PDO? This shim works perfectly, but just getting rid of their usage will be the right fix IMO.

lib/Doctrine/DBAL/Compat/PDO.php
lcobucci: Creating this thing to only throw an exception is pretty much the same as the hard dependency.

stof: Creating such class could break any other library checking the existing of the `PDO` class and then trying to use something else than a constant from it when it exists. So I'm -1 on adding it.

Ocramius:
Given the discussion above, can we add ext-pdo to composer.json in this
patch instead?

Ocramius:
Closing: moving to https://github.com/doctrine/dbal/pull/2709

PR #811, opened 2015-03-05
BlobType without fopen & allow_url_fopen

ahurt2000:
The allow_url_fopen parameter  is disabled in many host providers by security. This commit introduces a small change to avoid the use of fopen function that this is affected by this parameter.

### DIFF
--- a/lib/Doctrine/DBAL/Types/BlobType.php
+++ b/lib/Doctrine/DBAL/Types/BlobType.php
@@ -46,7 +46,7 @@ class BlobType extends Type
         }
 
         if (is_string($value)) {
-            $fp = fopen('php://temp', 'rb+');
+            $fp = tmpfile();
             fwrite($fp, $value);
             fseek($fp, 0);
             $value = $fp;


lib/Doctrine/DBAL/Types/BlobType.php
```diff
@@ -46,7 +46,7 @@ public function convertToPHPValue($value, AbstractPlatform $platform)
         }
 
         if (is_string($value)) {
-            $fp = fopen('php://temp', 'rb+');
+            $fp = tmpfile();
```
Ocramius: This will cause actual disk I/O.

beberlei: Hm this creates a real file, php://temp only does when over 2MB. Not sure about this.

ahurt2000: Correct. But allow_url_fopen = Off causes the fopen not work. Perhaps an alternative is checked before the state allow_url_fopen.

deeky666:
@ahurt2000 this is invalid. The code works even if `allow_url_fopen` is disabled. Tested it locally and it works. Please also refer to the [manual](http://php.net/manual/en/function.fopen.php) that states that the ini setting only has to be enabled for network URL protocols:

> If PHP has decided that filename specifies a registered protocol, and that protocol is registered as a network URL, PHP will check to make sure that allow_url_fopen is enabled. If it is switched off, PHP will emit a warning and the fopen call will fail. 

There might be some other problem with your PHP setup. If you get what's going wrong, please provide a failing test case.

PR #812, opened 2015-03-07
Treat SQLite Connection URLs Differently in DriverManager

chrisguitarguy:
In addition to setting `dbname`, which is ignored by the SQLite Driver, set the
`path` and `memory` params based on the database URL.

See http://www.doctrine-project.org/jira/browse/DBAL-1164

Not sure if this is a good solution, but it seemed more okay (less risky to BC) than changing the driver itself.


chrisguitarguy:
Well, never mind then.

PR #813, opened 2015-03-07
Treat SQLite Connection URLs Differently in DriverManager

chrisguitarguy:
In addition to setting `dbname`, which is ignored by the SQLite Driver, set the
`path` and `memory` params based on the database URL.

See DBAL-1164

Not sure if this is a good solution, but it seemed more okay (less risky to BC) than changing the driver itself.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/DriverManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/DriverManagerTest.php
@@ -151,27 +151,27 @@ class DriverManagerTest extends \Doctrine\Tests\DbalTestCase
             ),
             'sqlite relative URL with host' => array(
                 'sqlite://localhost/foo/dbname.sqlite',
-                array('dbname' => 'foo/dbname.sqlite', 'driver' => 'Doctrine\DBAL\Driver\PDOSqlite\Driver'),
+                array('dbname' => 'foo/dbname.sqlite', 'path' => 'foo/dbname.sqlite', 'driver' => 'Doctrine\DBAL\Driver\PDOSqlite\Driver'),
             ),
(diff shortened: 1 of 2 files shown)


chrisguitarguy:
Test failure is due to HHVM: https://travis-ci.org/doctrine/dbal/jobs/53478163

Maybe that one should go in allowed failures?

Ocramius:
It really shouldn't. We are actually waiting for an HHVM upgrade by travis-ci.

tests/Doctrine/Tests/DBAL/DriverManagerTest.php
```diff
@@ -151,27 +151,27 @@ public function databaseUrls()
             ),
             'sqlite relative URL with host' => array(
                 'sqlite://localhost/foo/dbname.sqlite',
-                array('dbname' => 'foo/dbname.sqlite', 'driver' => 'Doctrine\DBAL\Driver\PDOSqlite\Driver'),
+                array('dbname' => 'foo/dbname.sqlite', 'path' => 'foo/dbname.sqlite', 'driver' => 'Doctrine\DBAL\Driver\PDOSqlite\Driver'),
```
Ocramius: Please use different data-provider entries: don't change the existing ones.

chrisguitarguy: The original data provider entries test things that don't matter: `dbname` does nothing for SQLite driver. I can create new data provider entries if you'd like, but it seemed that changing the existing ones to test something that mattered made more sense.

deeky666:
This has been fixed in https://github.com/doctrine/dbal/pull/2267

PR #814, opened 2015-03-09
Allow `serverVersion` to be explicitly unspecified (`null`)

BreiteSeite:
Hi,

this PR allows `serverVersion`to be nullable.

We use doctrine/dbal in Integration-Tests and to prevent the unit-tests from connecting to a database, we specify `serverVersion` with something made-up (like `5.6`).

However, i'm not comfortable set `serverVersion` to a made-up server-version to prevent connections from being made, when there even isn't a server to connect to. With this PR merged, we could specify `serverVersion` to be `null` instead of something like `5.6` and still prevent connections from being made. `null` is a valid return value for `getDatabasePlatformVersion()`.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -418,7 +418,7 @@ class Connection implements DriverConnection
         }
 
         // Explicit platform version requested (supersedes auto-detection).
-        if (isset($this->_params['serverVersion'])) {
+        if (array_key_exists('serverVersion', $this->_params)) {
             return $this->_params['serverVersion'];
         }
 
--- a/tests/Doctrine/Tests/DBAL/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/ConnectionTest.php
@@ -8,6 +8,7 @@ use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Events;
 use Doctrine\Tests\Mocks\DriverConnectionMock;
 use Doctrine\Tests\Mocks\DriverMock;
+use Doctrine\Tests\Mocks\VersionAwareDriverMock;
 
 class ConnectionTest extends \Doctrine\Tests\DbalTestCase
 {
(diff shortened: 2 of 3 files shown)


BreiteSeite:
Can somebody please explain to me, why the PHP 5.3 travis builds for commit e74caf9 doesn't fail, even though i used short-array-syntax in my unit-test?

Thank you.

deeky666:
@BreiteSeite can you point me to some example code and or use case to evaluate whether this change makes sense? Thanks.

BreiteSeite:
@deeky666 yes.. the point is config mutability. So we read our config via environment variables, which leads to a config like this:

```php
<?php

return [
    // ...
    'serverVersion' => false !==getenv('SQL_SERVER_VERSION') ? getenv('SQL_SERVER_VERSION) : null
];
```

We now have a config per environment. For example, for phpunit, we do not set SQL_SERVER_VERSION as our test suite obv. should/will not make queries to the database (we have also an integration suite which loads the whole application and mocks the persistence layer/services).

Till now, if we initialize the application and serverVersion is null, doctrine will make queries to the database to figure out, which mysql version is running. This crashes the app/test as we do not have a connection configured.

With this change, null would be allowed and would prevent doctrine from trying to fetch the sql server version from the server.

An easy workaround of course is to specify as fake version in the phpunit environment (we make this at the moment) but it doesn't feel right to load SQL_SERVER_VERSION=5.6 even if we do not connect to the database at all or even have one available.

Hope this makes things a bit more clear, if you still have any questions or feedback i would be happy to hear from you.

BreiteSeite:
Rebase onto master.

deeky666:
@BreiteSeite so if I get you right you are facing issues when running unit tests and mocking DBAL? I agree that in such a scenario, DBAL should not try to connect implicity, but it sounds to me as if you were not mocking out DBAL correctly then. I'd still like to see some kind of snippet or pseudo-code that triggers the issue so that I can see if it might be misuage in the tests.

Ocramius:
Handled in https://github.com/doctrine/dbal/pull/2671

leofeyer:
This is actually not handled in #2671 and should be reopened.

```yml
doctrine:
    dbal:
        default_connection: default
        connections:
            default:
                driver: pdo_mysql
                host: "%database_host%"
                port: "%database_port%"
                user: "%database_user%"
                password: "%database_password%"
                dbname: "%database_name%"
                charset: UTF8
                server_version: ~
```

This should bypass the auto detection in case username and password have not been set yet.

PR #816, opened 2015-03-13
Fix uses Statement::setFetchMode function in class Connection

splokhikh:
I have a problem because interface function Connection::setFetchMode doesnt match PDOStatement::setFetchMode. Through Connection::setFetchMode I can send only first param to function PDOStatement::setFetchMode. 
With this fix in Connection class, function Statement::setFetchMode used with call_user_func_array and this allow send more than one argument.
With this fix, I can use native PDO::FETCH_COLUMN like this:
        $conn = $this->get('database_connection');
        $conn->setFetchMode(\PDO::FETCH_COLUMN, 0);
        $result = $conn->fetchAll("SELECT ..........");
and get full result array.
With function Connection::fetchColumn, I can take only the first row in result array.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -186,7 +186,7 @@ class Connection implements DriverConnection
     /**
      * @var integer
      */
-    protected $defaultFetchMode = PDO::FETCH_ASSOC;
+    protected $defaultFetchMode = [PDO::FETCH_ASSOC];
 
     /**
      * Initializes a new instance of the Connection class.
@@ -492,9 +492,9 @@ class Connection implements DriverConnection
      *
      * @return void
      */
-    public function setFetchMode($fetchMode)
+    public function setFetchMode($fetchMode, $arg2 = null, $arg3 = null)
     {
-        $this->defaultFetchMode = $fetchMode;
+        $this->defaultFetchMode = [$fetchMode, $arg2, $arg3];
     }
 
     /**
@@ -781,7 +781,7 @@ class Connection implements DriverConnection
             throw DBALException::driverExceptionDuringQuery($this->_driver, $ex, $statement);
         }
 
-        $stmt->setFetchMode($this->defaultFetchMode);
+        call_user_func_array([$stmt, 'setFetchMode'], $this->defaultFetchMode);
 
         return $stmt;
     }
@@ -832,7 +832,7 @@ class Connection implements DriverConnection
             throw DBALException::driverExceptionDuringQuery($this->_driver, $ex, $query, $this->resolveParams($params, $types));
         }
 
-        $stmt->setFetchMode($this->defaultFetchMode);
+        call_user_func_array([$stmt, 'setFetchMode'], $this->defaultFetchMode);
 
         if ($logger) {
             $logger->stopQuery();
@@ -876,7 +876,7 @@ class Connection implements DriverConnection
             $stmt = new ResultCacheStatement($this->executeQuery($query, $params, $types), $resultCache, $cacheKey, $realKey, $qcp->getLifetime());
         }
 
-        $stmt->setFetchMode($this->defaultFetchMode);
+        call_user_func_array([$stmt, 'setFetchMode'], $this->defaultFetchMode);
 
         return $stmt;
     }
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Connection.php
```diff
@@ -186,7 +186,7 @@ class Connection implements DriverConnection
     /**
      * @var integer
      */
-    protected $defaultFetchMode = PDO::FETCH_ASSOC;
+    protected $defaultFetchMode = [PDO::FETCH_ASSOC];
```
Ocramius: This is a BC break

lib/Doctrine/DBAL/Connection.php
```diff
@@ -489,12 +489,14 @@ public function setAutoCommit($autoCommit)
...
+     * @param mixed $arg2
+     * @param mixed $arg3
+     * 
      * @return void
      */
-    public function setFetchMode($fetchMode)
+    public function setFetchMode($fetchMode, $arg2 = null, $arg3 = null)
     {
-        $this->defaultFetchMode = $fetchMode;
+        $this->defaultFetchMode = [$fetchMode, $arg2, $arg3];
```
Ocramius: This syntax is incompatible with PHP 5.3

lib/Doctrine/DBAL/Connection.php
```diff
@@ -489,12 +489,14 @@ public function setAutoCommit($autoCommit)
      * Sets the fetch mode.
      *
      * @param integer $fetchMode
-     *
+     * @param mixed $arg2
```
Ocramius: What are `$arg2` and `$arg3`?

plokhikh:
Yes, this syntax incompatible with PHP 5.3, but it easy to change brackets [] to array(...). In my case, I use $arg2 - this is column number for fetching.
More examples can find in PHP PDO statement documentation (http://php.net/manual/en/pdostatement.setfetchmode.php):
public bool PDOStatement::setFetchMode ( int $PDO::FETCH_COLUMN , int $colno )
public bool PDOStatement::setFetchMode ( int $PDO::FETCH_INTO , object $object )
public bool PDOStatement::setFetchMode ( int $PDO::FETCH_CLASS , string $classname , array $ctorargs )

$arg2 can be string, object, int. $arg3 only array.

lib/Doctrine/DBAL/Connection.php
plokhikh: Yes, this syntax incompatible with PHP 5.3, but it easy to change brackets [] to array(...). In my case, I use $arg2 - this is column number for fetching.
More examples can find in PHP PDO statement documentation (http://php.net/manual/en/pdostatement.setfetchmode.php):
public bool PDOStatement::setFetchMode ( int $PDO::FETCH_COLUMN , int $colno )
public bool PDOStatement::setFetchMode ( int $PDO::FETCH_INTO , object $object )
public bool PDOStatement::setFetchMode ( int $PDO::FETCH_CLASS , string $classname , array $ctorargs )

$arg2 can be string, object, int. $arg3 only array.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -186,7 +186,7 @@ class Connection implements DriverConnection
     /**
      * @var integer
```
stof: this is not an integer anymore in your code

lib/Doctrine/DBAL/Connection.php
```diff
@@ -781,7 +783,7 @@ public function prepare($statement)
             throw DBALException::driverExceptionDuringQuery($this->_driver, $ex, $statement);
         }
 
-        $stmt->setFetchMode($this->defaultFetchMode);
+        call_user_func_array([$stmt, 'setFetchMode'], $this->defaultFetchMode);
```
stof: it would be much better to store the information in a way which would not need to rely on ``call_user_func_array``, which is slower than a direct method call

lib/Doctrine/DBAL/Connection.php
Ocramius: Please rename them accordingly for clarity. Tests are also required to  clarify the use-case scenario (and validate it)

Ocramius: Also: not sure if I'd ever want these PDO horrors in here (it's something that we've deliberately left out also when building stuff like https://github.com/crate/crate-pdo, as this sort of API is kinda useless :-\ )

plokhikh:
Ok, close merge request as unuseful.

PR #818, opened 2015-03-17
Rebuild SQLServerPlatform::doModifyLimitQuery again to use a CTE

billschaller:
Also only uses 1 regex.

This method is vastly simpler and more reliable than the old method of parsing a whole lot of stuff.

A caveat: if a query is passed that contains a subquery with an ORDER BY clause, the inner ORDER BY clause will be dropped. SQL server doesn't support using ORDER BY inside subqueries. In the future I think we should consider throwing an exception when these are found.

~~For now, hold off on merging this. There is a PR open for the ORM Paginator that prevents it from sending queries with multiple ORDER BY clauses. The Paginator is the only place in the ORM where this occurs. Until that PR is merged, this one should probably stay open.~~

I'll be adding a set of functional tests for the Paginator later this week.


zeroedin-bill:
@deeky666 review pls?

kokspflanze:
thanks for the PR, it also fix a problem with the "where in ('foo', 'bar')" clause in the "main-select".

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1182,77 +1182,63 @@ protected function doModifyLimitQuery($query, $limit, $offset = null)
...
-                    'hasTable'  => ( ! empty($matches[2])),
-                    'sort'      => isset($matches[4]) ? $matches[4] : null,
-                    'table'     => empty($matches[2]) ? '[^\.\s]*' : $matches[2]
-                );
-            }
+        // We'll find a SELECT or SELECT distinct and append TOP n to it
+        // Even if the TOP n is very large, the use of a CTE will
+        // allow the SQL Server query planner to optimize it so it doesn't
+        // actually scan the entire range covered by the TOP clause.
+        $selectPattern = '/^(\s*SELECT\s+(?:DISTINCT|)\s*)(.*)$/i';
```
deeky666: @zeroedin-bill the `(?:DISTINCT|)` is weird. Shoudln't ir rather be `(?:DISTINCT)?` ?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1182,77 +1182,63 @@ protected function doModifyLimitQuery($query, $limit, $offset = null)
...
+        $replacePattern = sprintf('$1%s $2', "TOP $end");
```
deeky666: I think `SELECT DISTINCT TOP ...` is not supported by SQL Server, is it?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1182,77 +1182,63 @@ protected function doModifyLimitQuery($query, $limit, $offset = null)
...
+        $query = preg_replace($selectPattern, $replacePattern, $query);
+
+        if (stristr($query, "ORDER BY")) {
+            // Inner order by is not valid in SQL Server for our purposes
+            // TODO throw DBALException here?
```
deeky666: IIRC inner `ORDER BY` clauses don't have any effect, do they? If not I think it's legit to just get rid of them.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
zeroedin-bill: That is weird. I'll change it.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
zeroedin-bill: Sure is.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
zeroedin-bill: That is what ```scrubInnerOrderBy``` does - if it finds an orderby clause inside a subquery, it strips it out.

zeroedin-bill:
@deeky666 pingie, also @Ocramius

afranioce:
Please add the query to test

```php
public function testModifyLimitQueryWithFromConditionalSubqueryAndOrder()
{
    $querySql = 'SELECT * FROM test t WHERE t.data = (SELECT MAX(t2.data) FROM test t2 WHERE t.id = t2.id)';
    $alteredSql = 'SELECT * FROM (SELECT *, ROW_NUMBER() OVER (ORDER BY (SELECT 0)) AS doctrine_rownum FROM test t WHERE t.data = (SELECT MAX(t2.data) FROM test t2 WHERE t.id = t2.id)) AS doctrine_tbl WHERE doctrine_rownum BETWEEN 1 AND 10 ORDER BY doctrine_rownum';
    $sql = $this->_platform->modifyLimitQuery($querySql, 10);
    $this->assertEquals(sprintf(self::$selectFromCtePattern, $alteredSql, 1, 10), $sql);

    $querySql = 'SELECT * FROM test t WHERE t.id = (SELECT TOP 1 t2.id FROM test t2 WHERE t.id = t2.id ORDER BY t2.data DESC)';
    $alteredSql = 'SELECT * FROM (SELECT *, ROW_NUMBER() OVER (ORDER BY (SELECT 0)) AS doctrine_rownum FROM test t WHERE t.id = (SELECT TOP 1 t2.id FROM test t2 WHERE t.id = t2.id ORDER BY t2.data DESC)) AS doctrine_tbl WHERE doctrine_rownum BETWEEN 1 AND 10 ORDER BY doctrine_rownum';
    $sql = $this->_platform->modifyLimitQuery($querySql, 10);
    $this->assertEquals(sprintf(self::$selectFromCtePattern, $alteredSql, 1, 10), $sql);

    $querySql = 'SELECT * FROM test t WHERE t.id = (SELECT TOP 1 t2.id FROM test t2 WHERE t.id = t2.id ORDER BY t2.data DESC) ORDER BY t.id';
    $alteredSql = 'SELECT * FROM (SELECT *, ROW_NUMBER() OVER (ORDER BY t.id) AS doctrine_rownum FROM test t WHERE t.id = (SELECT TOP 1 t2.id FROM test t2 WHERE t.id = t2.id ORDER BY t2.data DESC)) AS doctrine_tbl WHERE doctrine_rownum BETWEEN 1 AND 10 ORDER BY doctrine_rownum';
    $sql = $this->_platform->modifyLimitQuery($querySql, 10);
    $this->assertEquals(sprintf(self::$selectFromCtePattern, $alteredSql, 1, 10), $sql);
}
```

zeroedin-bill:
@deeky666 got a SO bug about this - any chance we can merge this?

http://stackoverflow.com/questions/34274008/doctrine2-with-ms-server-pagination-and-order-by-not-working-under-zf2/34280402#34280402

zeroedin-bill:
@afranioce In two of those assertions, the queries don't really make sense.

This will return all rows, because the WHERE predicate in the correlated subquery will cause it to be evaluated in a nested loop with the outside predicate.

``` SELECT * FROM test t WHERE t.id = (SELECT TOP 1 t2.id FROM test t2 WHERE t.id = t2.id ORDER BY t2.data DESC) ```
``` SELECT * FROM test t WHERE t.id = (SELECT TOP 1 t2.id FROM test t2 WHERE t.id = t2.id ORDER BY t2.data DESC) ORDER BY t.id DESC```

I think what you're getting at is that the new doModifyLimitQuery is stripping the ORDER BY clause in a subquery using TOP N.

This one will show the issue better:

``` SELECT * FROM test t WHERE t.id = (SELECT TOP 1 t2.id FROM test t2 ORDER BY t2.data DESC) ```
``` SELECT * FROM test t WHERE t.id = (SELECT TOP 1 t2.id FROM test t2 ORDER BY t2.data DESC)  ORDER BY t.date DESC```

Does that sound about right?

**Comment by zeroedin-bill** [comment]
@deeky666 ping - I think this PR is ready to merge.

File: lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php [hunk 4]
```diff
@@ -1182,77 +1182,113 @@ protected function doModifyLimitQuery($query, $limit, $offset = null)
...
+     *
+     * @param string $query The query
+     * @param int $currentPosition Start position of ORDER BY clause
+     * @return bool true if ORDER BY is in a TOP N query, false otherwise
+     */
+    private function isOrderByInTopNSubquery($query, $currentPosition)
+    {
+        // Grab query text on the same nesting level as the ORDER BY clause we're examining.
+        $subQueryBuffer = '';
+        $parenCount = 0;
```
deeky666: Typo `$parentCount`

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1182,77 +1182,113 @@ protected function doModifyLimitQuery($query, $limit, $offset = null)
...
 
-                $overColumns[] = $overColumn;
+        // If $parentCount goes negative, we've exited the subquery we're examining.
+        // If $currentPosition goes negative, we've reached the beginning of the query.
+        while ($parenCount >= 0 && $currentPosition >= 0) {
+            if ($query[$currentPosition] == '(') {
```
deeky666: Strict comparison please.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1182,77 +1182,113 @@ protected function doModifyLimitQuery($query, $limit, $offset = null)
...
+                $parenCount--;
+            }
+            if ($query[$currentPosition] == ')') {
```
deeky666: Strict comparison please.

deeky666: Shouldn't we have an `elseif` here to avoid the unnecessary additional check?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1182,77 +1182,113 @@ protected function doModifyLimitQuery($query, $limit, $offset = null)
...
+                $parenCount++;
+            }
+            if ($parenCount === 0) {
+                // Only yank query text on the same nesting level as the ORDER BY clause.
+                $subQueryBuffer = $query[$currentPosition] . $subQueryBuffer;
+            } else {
+                $subQueryBuffer = " " . $subQueryBuffer;
             }
```
deeky666: Can we make this `if` condition a ternary?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1182,77 +1182,113 @@ protected function doModifyLimitQuery($query, $limit, $offset = null)
...
+        // If $parentCount goes negative, we've exited the subquery we're examining.
+        // If $currentPosition goes negative, we've reached the beginning of the query.
+        while ($parenCount >= 0 && $currentPosition >= 0) {
+            if ($query[$currentPosition] == '(') {
+                $parenCount--;
+            }
+            if ($query[$currentPosition] == ')') {
+                $parenCount++;
+            }
+            if ($parenCount === 0) {
```
deeky666: Looks like we could get rid of this check inside the loop, too as it is an abortion condition and could be checked after the loop has run.

deeky666: Ah no nvm.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1182,77 +1182,113 @@ protected function doModifyLimitQuery($query, $limit, $offset = null)
...
-        foreach ($orderByParts as &$part) {
+    /**
+     * Remove ORDER BY clauses in subqueries - they're not supported by SQL Server.
+     * Caveat: will leave ORDER BY in TOP N subqueries.
+     *
+     * @param $query
+     * @return string
+     *
+     */
+    private function scrubInnerOrderBy($query) {
```
deeky666: Opening bracket on a new line

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1182,77 +1182,113 @@ protected function doModifyLimitQuery($query, $limit, $offset = null)
...
+        $count = substr_count(strtoupper($query), "ORDER BY");
+        $offset = 0;
+        while ($count-- > 0) {
```
deeky666: New line before

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1182,77 +1182,113 @@ protected function doModifyLimitQuery($query, $limit, $offset = null)
...
-                    $overColumn = preg_match($pattern, $query, $matches)
-                        ? ($column['hasTable'] ? $column['table']  . '.' : '') . $column['column']
-                        : $column['column'];
-                }
+            if ($currentPosition < $qLen - 1) {
+                $query = substr($query, 0, $orderByPos) . substr($query, $currentPosition - 1);
+                $offset = $orderByPos;
+            }
+        }
+        return $query;
```
deeky666: New line before please

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1182,77 +1182,113 @@ protected function doModifyLimitQuery($query, $limit, $offset = null)
...
 
-        //Split ORDER BY into parts
-        foreach ($orderByParts as &$part) {
+    /**
+     * Remove ORDER BY clauses in subqueries - they're not supported by SQL Server.
+     * Caveat: will leave ORDER BY in TOP N subqueries.
+     *
+     * @param $query
+     * @return string
+     *
```
deeky666: Unnecessary new line

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
zeroedin-bill: There's a nl before already

zeroedin-bill:
merge dis shit

Good to go.

tests/Doctrine/Tests/DBAL/Platforms/AbstractSQLServerPlatformTestCase.php
```diff
@@ -165,14 +167,18 @@ public function getGenerateForeignKeySql()
 
     public function testModifyLimitQuery()
     {
-        $sql = $this->_platform->modifyLimitQuery('SELECT * FROM user', 10, 0);
-        $this->assertEquals('SELECT * FROM (SELECT *, ROW_NUMBER() OVER (ORDER BY (SELECT 0)) AS doctrine_rownum FROM user) AS doctrine_tbl WHERE doctrine_rownum BETWEEN 1 AND 10 ORDER BY doctrine_rownum', $sql);
+        $querySql = 'SELECT * FROM user';
+        $alteredSql = 'SELECT TOP 10 * FROM user';
+        $sql = $this->_platform->modifyLimitQuery($querySql, 10, 0);
+        $this->assertEquals(sprintf(self::$selectFromCtePattern, $alteredSql, 1, 10), $sql);
```
deeky666: Theses tests are rather hard to read. I wonder if it would make sense to add our own assertion here like:

```php

protected function assertModifiesLimitQuery($query, $modifiedQuery, $limit, $offset)
{
    $selectFromCtePattern = "WITH dctrn_cte AS (%s) SELECT * FROM (SELECT *, ROW_NUMBER() OVER (ORDER BY (SELECT 0)) AS doctrine_rownum FROM dctrn_cte) AS doctrine_tbl WHERE doctrine_rownum BETWEEN %d AND %d ORDER BY doctrine_rownum ASC";

    $this->assertEquals(sprintf($selectFromCtePattern, $modifiedQuery, $offset + 1, $limit), $query);
}
```

deeky666:
Awesome work @zeroedin-bill ! Thanks! Let's hope the best ;)

zeroedin-bill:
:+1:

:thought_balloon: SQL Server Sux XD

afranioce:
@zeroedin-bill I have a table with the following architecture

| id |    movement_date    | vec |
| :--: | :-------------------: | :---: |
|  1 | 2015-02-01 12:00:00 | 123 |
|  2 | 2015-02-02 12:00:00 | 123 |
|  3 | 2015-02-01 12:00:00 | 456 |
|  4 | 2015-02-02 12:00:00 | 456 |

I just want to catch the latest movements, the sql query would be as follows

```sql
SELECT * FROM AutoNetVehicleBundle:Vehicle v WHERE v.movement_date = (SELECT MAX(v2.movement_date) FROM AutoNetVehicleBundle:Vehicle v2 WHERE v2.vec = v.vec)
```

That does not work, the problem is that when we place parentheses in a sub query, within the "WHERE" does not work,
the query returns an error

**An exception occurred while executing 'SELECT * FROM (SELECT DISTINCT a0_.id AS id0 FROM an_vehicle a0_ WHERE a0_.movement_date = (SELECT MAX(a1_.movement_date) AS dctrn__1, ROW_NUMBER() OVER (ORDER BY (SELECT 0)) AS doctrine_rownum FROM an_vehicle a1_ WHERE a1_.vec = a0_.vec)) AS doctrine_tbl WHERE doctrine_rownum BETWEEN 1 AND 50':**

**SQLSTATE[42S22]: [Microsoft][ODBC Driver 11 for SQL Server][SQL Server]Invalid column name 'doctrine_rownum'.**

When i change the core doctrine of the line 813 in the file lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php

before: $selectFromPattern = '/^(\s*SELECT\s+(?:\((?>[^()]+)|(?:R)*\)|[^(])+)\sFROM\s/i';
after: $selectFromPattern = '/\sFROM\s/i';

The query correctly returns as expected

zeroedin-bill:
@afranioce can you try with 2.5@dev?

afranioce:
@zeroedin-bill I downloaded now from 2.5@dev and it worked as expected, thanks!

zeroedin-bill:
@afranioce Great!

PR #820, opened 2015-03-19
SchemaManager quoting fixes

billschaller:
The SchemaManager, when used with SQL Server, fails to quote incoming table names and column names that should be quoted. That means that when a table that exists in the database called 'quote-address' needs to be dropped, the drop table statement will fail because the identifer is not marked as quoted when the schema diff is created.

This patch adds a method isValidIdentifier to the AbstractPlatform API - the purpose of this is to check if an identifier is valid to be used in SQL on the platform. 

This is used by a new protected method in AbstractSchemaManager, quoteIncomingIdentifier. This method checks if the passed string literal is a valid identifier using isValidIdentifier. If the identifier is not valid, it quotes it for the platform and returns it. If it is valid, or is already quoted, it returns the identifier unchanged.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -824,7 +824,7 @@ class SQLServerPlatform extends AbstractPlatform
         END AS name
 FROM sysobjects t
 LEFT JOIN sys.schemas s ON s.schema_id = uid
-LEFT JOIN sys.database_principals p ON p.principal_id = s.principal_id
+CROSS JOIN sys.database_principals p
 WHERE t.type = 'U'
     AND t.NAME != 'sysdiagrams'
     AND t.category != 2
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -815,7 +815,21 @@ class SQLServerPlatform extends AbstractPlatform
     {
         // "sysdiagrams" table must be ignored as it's internal SQL Server table for Database Diagrams
         // Category 2 must be ignored as it is "MS SQL Server 'pseudo-system' object[s]" for replication
-        return "SELECT name FROM sysobjects WHERE type = 'U' AND name != 'sysdiagrams' AND category != 2 ORDER BY name";
+        // Table names are returned with the schema name IF and only if the schema
+        // is not the default schema for the user accessing the database.
+        return "SELECT CASE 
+        WHEN s.NAME <> p.default_schema_name
+            THEN '[' + s.NAME + '].[' + t.NAME + ']'
+        ELSE t.NAME
+        END AS name
+FROM sysobjects t
+LEFT JOIN sys.schemas s ON s.schema_id = uid
+LEFT JOIN sys.database_principals p ON p.principal_id = s.principal_id
+WHERE t.type = 'U'
+    AND t.NAME != 'sysdiagrams'
+    AND t.category != 2
+    AND p.principal_id = DATABASE_PRINCIPAL_ID()
+ORDER BY t.NAME";
     }
 
     /**
@@ -1449,7 +1463,7 @@ class SQLServerPlatform extends AbstractPlatform
      */
     public function quoteSingleIdentifier($str)
     {
-        return "[" . str_replace("]", "][", $str) . "]";
+        return "[" . str_replace("]", "]]", $str) . "]";
     }
 
     /**
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -817,19 +817,14 @@ class SQLServerPlatform extends AbstractPlatform
         // Category 2 must be ignored as it is "MS SQL Server 'pseudo-system' object[s]" for replication
         // Table names are returned with the schema name IF and only if the schema
         // is not the default schema for the user accessing the database.
-        return "SELECT CASE 
-        WHEN s.NAME <> p.default_schema_name
-            THEN '[' + s.NAME + '].[' + t.NAME + ']'
-        ELSE t.NAME
-        END AS name
+        return "SELECT t.NAME AS table_name,
+        s.NAME AS schema_name
 FROM sysobjects t
 LEFT JOIN sys.schemas s ON s.schema_id = uid
-CROSS JOIN sys.database_principals p
 WHERE t.type = 'U'
     AND t.NAME != 'sysdiagrams'
     AND t.category != 2
-    AND p.principal_id = DATABASE_PRINCIPAL_ID()
-ORDER BY t.NAME";
+ORDER BY s.NAME, t.NAME";
     }
 
     /**
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -311,7 +311,7 @@ class SQLServerPlatform extends AbstractPlatform
      * Returns the SQL snippet for declaring a default constraint.
      *
      * @param string $table  Name of the table to return the default constraint declaration for.
-     * @param array  $column Column definition.
+     * @param array  $column Column definition. Note: The column name in the column definition should be quoted.
      *
      * @return string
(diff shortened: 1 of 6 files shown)


lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1551,4 +1565,12 @@ private function generateIdentifierName($identifier)
 
         return strtoupper(dechex(crc32($identifier->getName())));
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isValidIdentifier($name)
+    {
+        return !!preg_match('/^(?:(?:\p{L}|[_@#])(?:\p{L}|\p{Nd}|[$@#_])*|".*"|\[.*\])$/ui', $name);
```
Ocramius: Where was this voodoo picked from? Can we also avoid the `!!` and use `(bool)`?

billschaller: Sure thing, I can make that change tomorrow. As for the regex, it's checking for either a quoted string ([] or "") or a valid identifier as specified in the TSQL spec. The \p{L} is for unicode letters, etc.

Ocramius: Note that `\p{L}` works only if the regex stuff was compiled with enabled unicode: that may or may not be a problem (I don't think it is).

I was more referring to where this reference regex comes from, if it comes from somewhere :-)

billschaller: Wrote it meself :p

Ocramius: @zeroedin-bill ok, that may or may not be a problem too ;-)

Anyway, can you put comments on what you assume an example (complex) matched identifier would look like? Some things such as the `.*` are a bit fishy imo.

billschaller: Sure thing- I can write a few quick tests for this tomorrow as well.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3525,4 +3525,17 @@ public function getStringLiteralQuoteCharacter()
...
+
+    /**
+     * Determines if a string literal is valid as an identifier in this
+     * platform.
+     *
+     * @param string $name The string literal to test
+     *
+     * @return bool
+     */
+    public function isValidIdentifier($name)
```
deeky666: I remember that I was trying to implement something like this some time ago and we had a long discussion about that internally. The problem here are two things:
1. AFAIK know there is no secure way in PHP to be 100% sure that a given value is a valid identifier with regards to the underyling platform. A complex regex might catch 99% but I doubt it can provide 100% safety. @beberlei was asking PHP internals about that topic as far as I remember and they gave a similar answer.
2. If we cannot be 100% sure that a given identifier is valid, such a method will be misleading to the consumers as they will rely on that and in a worst case scenario use it with unfiltered user input which opens up for SQL injection.

The only option I see to put such a method in here is to document security issue well (in the code and also in the documentation).

@beberlei @Ocramius thoughts?

billschaller: This is a good point -- My intent here is only to provide a way to conditionally quote identifiers being extracted from the database. Maybe instead of a wide-ranging isValidIdentifier method here, we could name the method differently and note its intended use.

The alternative is to quote every identifier coming from the DB. This is more or less what SQL server's internal script generation does. The trick there would be to ensure things don't get double-escaped.

Right now, identifiers with illegal characters are not being properly handled when they are being retrieved from the database on SQL Server. This causes the ORM functional test suite to error out during cleanup.

deeky666: @zeroedin-bill the issue is totally valid and I also agree that identifiers should be quoted intelligently by the schema manager when building the schema objects. Just not sure what the proper approach is here. Maybe we should leave that functionality completely internal.
I think I have done something similar for Oracle already. See [here](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/OracleSchemaManager.php#L339-L356). Although this was more about casing, it is somewhat related as it "normalizes" identifiers coming from the DB to be usable by Doctrine and to correlate better with mappings and such...

billschaller: Yes, an internal approach is fine here - the only requirement is for the DBAL to produce a valid schema. This functionality can just as easily be rolled back into SQLServerSchemaManager.

If that seems like a good course of action I can take care of that in short order.

deeky666: :+1:

zeroedin-bill:
@deeky666 Cleaner now - a slight change to quoting in the platform -- it was actually generating invalid quoted names. 

SQL Server would actually quote

```
foo]
```

as 

```
[foo]]]
```

But the platform was quoting it as

```
[foo][]
```

which is wrong.

zeroedin-bill:
@deeky666 is there a functional testcase somewhere for platform quoting?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -815,7 +815,21 @@ public function getListTablesSQL()
     {
         // "sysdiagrams" table must be ignored as it's internal SQL Server table for Database Diagrams
         // Category 2 must be ignored as it is "MS SQL Server 'pseudo-system' object[s]" for replication
-        return "SELECT name FROM sysobjects WHERE type = 'U' AND name != 'sysdiagrams' AND category != 2 ORDER BY name";
+        // Table names are returned with the schema name IF and only if the schema
+        // is not the default schema for the user accessing the database.
+        return "SELECT CASE 
+        WHEN s.NAME <> p.default_schema_name
+            THEN '[' + s.NAME + '].[' + t.NAME + ']'
```
deeky666: @zeroedin-bill I don't think this is a good idea to always quote here. Those identifiers should go through `quoteIncomingIdentifier()` aswell IMO as you might otherwise get problems with casing and such.
Maybe it would be a good idea to retrieve the schema name and table name seperately here and treat them accordingly in the schema manager. Thoughts?

billschaller: So I suppose I would return the table and schema name separately here as in postgres, and update SQLServerSchemaManager::_getPortableTableDefinition to handle that...

billschaller: Resolved this using a similar infrastructure to that which is used in PostgreSQL.

zeroedin-bill:
@deeky666 I think this fixes your issues and is closer to postgres impl now.

lib/Doctrine/DBAL/Schema/SQLServerSchemaManager.php
```diff
@@ -169,7 +176,22 @@ protected function _getPortableTableForeignKeyDefinition($tableForeignKey)
      */
     protected function _getPortableTableDefinition($table)
     {
-        return $table['name'];
+        $defaultSchema = $this->getDefaultSchemaName();
+        if ($table['schema_name'] == $defaultSchema) {
```
deeky666, line 183: Please surround control blocks with newlines for readability

tests/Doctrine/Tests/DBAL/Functional/Schema/SQLServerSchemaManagerTest.php
```diff
@@ -355,4 +355,38 @@ public function testPkOrdering()
         $this->assertEquals('colB', $columns[0]);
         $this->assertEquals('colA', $columns[1]);
     }
+
+    /**
+     * @group DBAL-75
+     */
+    public function testTableWithSchema()
+    {
+        $this->_conn->exec('CREATE SCHEMA nested');
```
deeky666: Can you please use `$this->_platform->getCreateSchemaSQL()` for this?

morozov:
Closing as stale in favor of the approach developed in https://github.com/doctrine/dbal/issues/4772.

> This patch adds a method isValidIdentifier to the AbstractPlatform API - the purpose of this is to check if an identifier is valid to be used in SQL on the platform.

This approach is invalid IMO. All (well, most) identifiers are valid if quoted.

PR #821, opened 2015-03-25
Added method for switching support for foreign keys for SQLite

hason:


lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -41,6 +41,11 @@
 class SqlitePlatform extends AbstractPlatform
 {
     /**
+     * @var bool
+     */
+    protected $supportsForeignKeyConstraints = false;
```
Ocramius, line 46: I'd rather make detect this, like with other platforms.

deeky666:
What's the background for a switch here? Is it version specific?

Ocramius:
> This document describes the support for SQL foreign key constraints introduced in SQLite version 3.6.19. 

From https://www.sqlite.org/foreignkeys.html

deeky666:
So it would make sense to introduce a new platform class and have the driver detect the appropriate platform accordingly (as done with other vendors already)

Ocramius:
I think so, though it doesn't just work "out of the box" (can't remember why: @lsmith worked on it last time)

lsmith77:
if I could only remember .. maybe the answer is hidden in some comment in some ticket :-/

hason:
I'm preparing an automatic detection in another PR.

> Foreign key constraints are disabled by default (for backwards compatibility), so must be enabled separately for each database connection. (Note, however, that future releases of SQLite might change so that foreign key constraints enabled by default. Careful developers will not make any assumptions about whether or not foreign keys are enabled by default but will instead enable or disable them as necessary.)

https://www.sqlite.org/foreignkeys.html

Adirelle:
@deeky666:  According to SQLite documentation, this isn't completely version-dependant though it can be detected using some PRAGMA queries:
* If the results of "PRAGMA compile_options" contains "OMIT_FOREIGN_KEYS", foreign key syntax are disabled and will cause errors.
* If the results of "PRAGMA compile_options" contains "OMIT_TRIGGERS", foreign key syntax is enabled but the constraint won't be honored.
* If "PRAGMA foreign_keys" returns no rows, foreign keys aren't supported, they may cause an error or not (see above).
* If "PRAGMA foreign_keys" returns a row with value "0", foreign keys are supported but the constraints are not enforced. They can be enforced using "PRAGMA foreign_keys = on".
* If "PRAGMA foreign_keys" returns a row with value "1", foreign keys are supported and honored. They can be disabled using "PRAGMA foreign_keys = off".

@Ocramius: even though foreign key constraints are supported in "CREATE TABLE" statements, they aren't in "ALTER TABLE". This causes issue on table creation when supportsForeignKeyConstraints() returns true, as the schema manager will try to call getCreateForeignKeySQL or getDropForeignKeySQL which both throw an exception. See related issue: [DBAL-1253](http://www.doctrine-project.org/jira/browse/DBAL-1253).

hason:
@deeky666 @Ocramius @Adirelle This PR solves issue: [DBAL-1253](http://www.doctrine-project.org/jira/browse/DBAL-1253)

deeky666:
@Adirelle @hason @Ocramius so if foreign keys are not natively supported in table alterations, can we work around that somehow to get it still working? Thinking of that, SQLite doesn't have `ALTER TABLE` anyways, does it? I remeber that the platform is recreating the complete table each time on table alteration. If there is no way way to make foreign keys compatible with table alterations, we should not try to implement it "partially" IMO as we do not benefit from that, do we?

deeky666:
Reference: https://github.com/doctrine/dbal/issues/1204

Adirelle:
@deeky666: SQLite does have an `ALTER TABLE` command but it is very limited, see: https://www.sqlite.org/lang_altertable.html.

Disabling foreigns key in SQLitePlatform is an easy way to avoid some errors and discrepancy.

Now, I think proper foreign key support would be possible but would require some work and I'm not sure how Doctrine should handle them: automatically use and enable them when possible ? or, considering recreating a table can be quite costly, make them optional and raise an error (or a warning) if they cannot be enabled ?

deeky666:
@Adirelle yeah remembering now that we are always recreating tables because alter table limitations. My implementation suggestion would be the following:

`SqlitePlatform` (current one, should act as pre `3.6.19` version platform)
- leave foreign key support disabled
- do not offer possibility to enable foreign keys explicitly (as they are not supported anyways)
- investigate the current schema diff problem and fix it to not generate foreign key SQL at all

`Sqlite3619Platform` (new platform, should act as >= `3.6.19` version platform)
- enable foreign key support by default (is it actually enabled and working in a default setup or do we have to enforce via `PRAGMA` here?)
- add `enableForeignKeySupport()` and `disableForeignKeySupport()` (or alike)
- implement foreign key support properly

`SpliteSchemaManager`
- check for foreign key support when introspecting the schema to either include or dismiss foreign keys in the introspected schema

`AbstractSQLiteDriver`
- implement `VersionAwarePlatformDriver` interface
- do `PRAGMA` magic to evaluate and eventually force foreign key support, then set support in the platform

What do you think? Does that cover everything said so far?

webmozart:
For reference, comment by @Ocramius on Twitter: https://twitter.com/Ocramius/status/730022221809389569

> @webmozart @sulu btw, I think that deeky666 summed it up quite nicely

webmozart:
@deeky666 Do you think you could work on your suggestion?

morozov:
Closing as stale.

PR #822, opened 2015-03-26
Fix for bad profiling data, showing an indefinitely long query

DHager:
This fix solves a problem where a failed transaction generates inaccurate profiling data.

In Symfony's "Timeline" profiler view, the bug makes it appear as if Symfony is taking a lot of time to process something, up until the very end of the request.


Ocramius:
Requires a test case, I'd say.

DHager:
Test added.

deeky666:
@DHager thanks!

PR #823, opened 2015-03-27
fix client_encoding setting to support pgbouncer

rocksfrow:
The current release doesn't allow connecting to pgbouncer when specifying the charset due to the lack of support for the 'options' parameter. The fix is to use the client_encoding option directly instead of passing it through the options parameter.

This exact same bug was fixed in node.js's PG driver the exact same way.

https://github.com/brianc/node-postgres/pull/356

I've confirmed this fix works with pgbouncer.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
@@ -83,7 +83,7 @@ class Driver extends AbstractPostgreSQLDriver
         }
 
         if (isset($params['charset'])) {
-            $dsn .= "options='--client_encoding=" . $params['charset'] . "'";
+            $dsn .= 'client_encoding=' . $params['charset'] . ' ';
         }
 
         if (isset($params['sslmode'])) {


rocksfrow:
FYI -- the error you'll get when trying to connect to pgbouncer using the current release:
````
Uncaught exception 'PDOException' with message 
'SQLSTATE[08006] [7] ERROR:  Unsupported startup parameter: options' 
in /home/xxx/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php:43
````
Again the fix is simply to pass client_encoding option directly.

deeky666:
Thanks @rocksfrow !

deeky666:
backported to 2.5 in https://github.com/doctrine/dbal/commit/e25ccaeae7d3fb8c7756c603361857695c80b42b

rocksfrow:
@deeky666 no prob -- thanks for the quick merge! Would 2.5.2 be the first release that'll include this fix?

rocksfrow:
I might just have to revert this "fix" and basically just not set the charset in my connection string in order to use pgbouncer. Based on the current logic -- that might be the only option.

Give me a few minutes to confirm.

zeroedin-bill:
@rocksfrow maybe a test for these options would be a good idea

rocksfrow:
@zeroedin-bill there isn't already a postgres connection test?

riccardonar:
@rocksfrow I'm using postgres 9.3.4 on slackware..

zeroedin-bill:
@rocksfrow yes, but it clearly is not covering completely...

rocksfrow:
@zeroedin-bill I'll check out the test and see if it's not using the parameters properly.

rocksfrow:
@riccardonar I don't see how you're getting an invalid option connecting to 9.3. Can you please provide more details? Are you connecting to a connectiong manager/pooler of any sort (pg-pool,pgbouncer) or just connecting directly to tcp 5432 -- or are you connecting via socket maybe?

rocksfrow:
I want to do a manual test and actually confirm client_encoding doesn't work on 8.x -- it might just not be in the doc. I'm definitely connecting to 9.3 with client_encoding -- but via TCP.

riccardonar:
@rocksfrow you are right.. my postgres client version is 9.0.4

rocksfrow:
@riccardonar -- lol, okay I was going to say!

Alright so I am wrapping up a fix right now that will set the client encoding via the sql standard `SET NAMES` instead of via PDO dsn.

This will ensure the encoding is set properly in ALL versions. I just confirmed it works on both.

rocksfrow:
@riccardonar I just pushed up changes to my fork. Would you be a saint and confirm it works well for you with your environment? I've confirmed it's properly setting the charset set via a manual SHOW client_encoding after connection.

@deeky666 Here are my changes: https://github.com/rocksfrow/dbal/commit/b8131d1f3b7443a2a47ef8fcf65839b9d9a244af

I'll create a pull request.

zeroedin-bill:
@rocksfrow can you add a comment alongside that change hat explains which
versions don't support the dsn option?

rocksfrow:
Here is my pull request:

https://github.com/doctrine/dbal/pull/828

rocksfrow:
@zeroedin-bill please see my pull request -- I just pushed up another commit adding details RE: client_encoding support. Is that sufficient?

@riccardonar would you mind forking my repo to test these changes on your environment? rocksfrow/dbal

rocksfrow:
@riccardonar I'd greatly appreciate you replying to my NEW pull request (not this one) -- and confirm the fix provided there resolves the issue you are running into. That way they can expedite the merge upstream.

Basically just comment out where client_encoding was being set, and add the set names query in the connect() method).
https://github.com/rocksfrow/dbal/commit/5a4c0f2f65a2858a5d5dae0fb36a7f1af200b113

riccardonar:
@rocksfrow your repo works fine for me! Thanks

rocksfrow:
@riccardonar great! I am glad I was able to resolve that for you quickly --
sorry about that.

@deeky666 let me know if there is anything else you need to happen in order
to get this pull request merged asap.

rocksfrow:
@riccardonar -- would you mind commenting on the new pull request - #825
confirming that the fix works? That way whoever reviews that pull request
will hopefully get it merged in quickly so we don't have to hack our
versions :)

rocksfrow:
Thanks guys! Happy Coding.

PR #828, opened 2015-04-02
rehashed charset implementation to support old versions of postgresql

rocksfrow:
My previous client_encoding fix turned out not to work on older postgres versions -- BUT, the options param isn't supported by pgbouncer at all. SO, because there isn't consistent behavior with setting the character set via DSN - this change utilizes `SET NAMES`. I used SET NAMES since that is SQL standard and supported by PG.

I confirmed `SET NAMES` is supported all the way back to the older doc PG has, 7.1: http://www.postgresql.org/docs/7.1/static/multibyte.html

### DIFF
--- a/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
@@ -53,7 +53,10 @@ class Driver extends AbstractPostgreSQLDriver
                 $pdo->setAttribute(PDO::PGSQL_ATTR_DISABLE_PREPARES, true);
             }
 
-            /* define charset via SET NAMES to avoid inconsistent DSN support */
+            /* defining client_encoding via SET NAMES to avoid inconsistent DSN support
+             * - the 'client_encoding' connection param only works with postgres >= 9.1
+             * - also, padding client_encoding via the 'options' param breaks pgbouncer support
+             */
             if (isset($params['charset'])) {
               $pdo->query('SET NAMES \''.$params['charset'].'\'');
             }
--- a/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
@@ -53,6 +53,11 @@ class Driver extends AbstractPostgreSQLDriver
                 $pdo->setAttribute(PDO::PGSQL_ATTR_DISABLE_PREPARES, true);
             }
 
+            /* define charset via SET NAMES to avoid inconsistent DSN support */
+            if (isset($params['charset'])) {
+              $pdo->query('SET NAMES \''.$params['charset'].'\'');
+            }
+
             return $pdo;
         } catch (PDOException $e) {
             throw DBALException::driverException($this, $e);
@@ -80,10 +85,6 @@ class Driver extends AbstractPostgreSQLDriver
 
         if (isset($params['dbname'])) {
             $dsn .= 'dbname=' . $params['dbname'] . ' ';
-        }
-
-        if (isset($params['charset'])) {
-            $dsn .= 'client_encoding=' . $params['charset'] . ' ';
         }
 
         if (isset($params['sslmode'])) {
--- a/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
@@ -55,7 +55,7 @@ class Driver extends AbstractPostgreSQLDriver
 
             /* defining client_encoding via SET NAMES to avoid inconsistent DSN support
              * - the 'client_encoding' connection param only works with postgres >= 9.1
-             * - also, padding client_encoding via the 'options' param breaks pgbouncer support
+             * - passing client_encoding via the 'options' param breaks pgbouncer support
              */
             if (isset($params['charset'])) {
               $pdo->query('SET NAMES \''.$params['charset'].'\'');


zeroedin-bill:
Looks like travis is happy, @riccardonar is happy, my tests run, all seems OK. I think we need to add a functional test for pgsql connections though - Esp. given that we have people connecting via bouncers and whatnot.

riccardonar:
Yes, for me it works!

rocksfrow:
@zeroedin-bill when it comes to postgres, it's extremely common to use poolers. PG has a lot more overhead when creating connections than other DBs so they're almost required for any high-volume environment.

RE: travis checks

I noticed you only have travis checks setup for 9.1+ anyway. The previous fix would have passed travis checks too :)

RE: functional test

Can you elaborate on the critiera and/or point me to some references? I wouldn't mind contributing a test. I just wrote one basically when I made these recent changes -- in that, it opens a connection used the configuration and then uses `SHOW client_encoding` and determines the previous SET NAMES usage works. Is that basically what you mean?

zeroedin-bill:
@rocksfrow I'm aware - I use postgres :)

I'm writing a quick functional test for the charset stuff.

rocksfrow:
@zeroedin-bill well you had mentioned the fact it passed the travis tests deemed it worth to be merged in -- so I was clarifying that shouldn't really matter in this case anyway.

Thanks for taking care of writing the test and merging my fix. Have a good day.

lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
```diff
@@ -53,6 +53,14 @@ public function connect(array $params, $username = null, $password = null, array
                 $pdo->setAttribute(PDO::PGSQL_ATTR_DISABLE_PREPARES, true);
             }
 
+            /* defining client_encoding via SET NAMES to avoid inconsistent DSN support
+             * - the 'client_encoding' connection param only works with postgres >= 9.1
+             * - passing client_encoding via the 'options' param breaks pgbouncer support
+             */
+            if (isset($params['charset'])) {
+              $pdo->query('SET NAMES \''.$params['charset'].'\'');
```
deeky666: I am not sure whether this is an acceptable solution. Having one extra query on each connect is affecting performance I suppose... I remember we had `SET NAMES` back in old Doctrine 1 days and I believe it was avoided in 2.x for a good reason.
@Ocramius @beberlei thoughts?

rocksfrow: @deeky666 I definitely agree this isn't the cleanest fix -- BUT, with that said, I also disagree with it going back how it was -- because you're then making the class unusable with pgbouncer (or other poolers) when it worked fine before.

I would think the only other option would be to simply rely on AUTO, and not set the client_encoding at all.

rocksfrow:
@deeky666 what we _COULD_ do is use my previous fix for PG >= 9.1, and for backwards compatibility -- use `SET NAMES`. These versions that don't support client_encoding aren't supported by postgres anyway. I noticed you guys have "platforms" to add support for new stuff in 9.1 and 9.2 -- is that something we could take advantage of?

Curious to hear your thoughts.

lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
deeky666: I totally get your point. It's just a question of portability vs performance. Gosh why can't PostgreSQL be consistent on this -.-

deeky666:
@rocksfrow yep we could do that if we knew the version of the underlying server before actually connecting. This is totally driver specific, nothing that needs to be in the platform...

lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
rocksfrow: I know right! So frustrating!

Well FYI -- this wasn't even being set before at all (like you said I guess it was removed). The reason this came up is because somebody added it -- so when I upgraded it totally broke on PG.

So maybe the BEST solution would be to simply totally revert these changes AND the original change:

https://github.com/rocksfrow/dbal/commit/36fa4101f51d48506dce295214dca148e17a6671

Basically, that means DBAL doesn't set the charset at all and leaves it up to the server "auto" mode.

TBH, I was a little surprised it was being set explicity anyway as I never have in the past (instead I use AUTO (default) and it uses the systems LOCALE).

Thoughts?

rocksfrow: Haha -- I just noticed you're actually the one who made that change @deeky666 heh

deeky666: Yep I added it back then but I wasn't aware of the pgbouncer thingy... Well other drivers support setting the charset explicitly, too so supporting this for pgsql is reasonable. TBH I COULD live with `SET NAMES` I am just curious on other opinions about this...

rocksfrow:
@deeky666 yeah I mean honestly I was just throwing that out there -- see my more recent comments to you. 

This wasn't a problem at all until you introduced the charset param a bit ago -- why was that added to begin with?

I've tested leaving it out all together connects fine too -- of course the limitation there is that the character set it always whatever your LOCALE is... but that may very well be sufficient.

On the other hand, is SET NAMES usage really that much overhead?

rocksfrow:
TO BE CLEAR -- I would be fine with just removing your original commit:

https://github.com/rocksfrow/dbal/commit/36fa4101f51d48506dce295214dca148e17a6671

Everything worked fine for me before that change was introduced. I am only messing with this because the fact that change broke pgbouncer connections.

lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
rocksfrow: I feel ya. So we are in agreement.

Ie, set names is the best way to set the encoding BUT, is it worth the
overhead?

Note: we could leave it as all the user has to do is NOT SET the charset in
their connection params and it's skipped.

Note, this email was sent from my cell phone.

rocksfrow: PS - you said you use PG, but don't use pgbouncer? You need to get on that!
haha -- trust me, you'll love it. The performance increase is insane. :1:

deeky666: No I don't use that in any application. Just for development of DBAL/ORM

zeroedin-bill: @rocksfrow @deeky666 we could just use the 9.1+ method of connecting inside a try block - if it fails, check the message for 'client_encoding' and connect again using the fallback method with SET NAMES.

We could also check first if the pgsql extension is loaded -- pgsql has a constant PGSQL_LIBPQ_VERSION which tells what client version is in use. If PGSQL_LIBPQ_VERSION >= 9.1, use client_encoding in the dsn. If not, use the fallback method. 

This would be really easy to fix if PDO would allow us to get more info about drivers...

Another alternative would be to run phpinfo and find "PostgreSQL\(libpq\) Version => ([0-9\.]*)"... But that's just dumb.

rocksfrow: @zeroedin-bill I wouldn't imagine most would like the idea of requiring double-connections.

What PHP version was `PGSQL_LIBPQ_VERSION ` introduced? That's exactly what I was asking for earlier (to use client_encoding for PG >= 9.1 or fallback to SET NAMES). BUT, only if that constant is reliably available.

I'll look into it... that would be an easy addition to my current state.

Originally I had asked if there was a way to do a version compare -- and only use this method for 9.1+. I wasn't aware of the PGSQL_LIBPQ_VERSION constant. I like this.

rocksfrow: Actually, @zeroedin-bill that PGSQL_LIBPQ_VERSION suggestion I don't think is valid either. This is a PHP variable. My PHP (web servers) are not on the same as PGSQL (db servers).

The PGSQL_LIBPQ_VERSION constant on my web servers is going to reflect the installed pgsql client -- NOT the version of my actual server.

For example, if I have a RHEL 5 web server -- it's going to have postgresq 8.1 by default. BUT, I have PG 9.3 on my db server. Using that constant is going to make the script think I am connecting to 8.1 which is totally false.

You can't assume the client on the web server matches the database server version... right?

I'm not confused am I?

riccardonar: The problem is postgres client, not server. For me PGSQL_LIBPQ_VERSION is
the best way
Il 02/apr/2015 20:09 "Kyle Renfrow" <notifications@github.com> ha scritto:

rocksfrow: Yeah I'm def. not confused. @zeroedin-bill those suggestions aren't valid unfortunately.

IE, PGSQL_LIBPQ_VERSION gives me 8.4 on my web server (default version from a centos 6 install) -- but of course I have 9.3 on my db server. Most of us aren't running on a single-server environment!

The first one out of the three is the only one that would actually work.

@deeky666 what do you think about the suggestion of trying with client_encoding, and falling back to SET NAMES on failure? I can make those modifications easy enough. That way -- you only have the double connection if you're using an old (unsupported mind you) version of PG.

rocksfrow: @riccardonar I don't see how that's true. The server is rejecting the option -- not the client

IE, if you're connecting from a centos 6 install out of the box... it's going to have the 8.4 client. PGSQL_LIBPQ_VERSION = 8.4.

I can't use that value to determine what options to use, as I need to know what options the SERVER supports.

IE, i'd be checking logic againstn 8.4 when the server is running 9.3

The only way using PGSQL_LIBPQ_VERSION  would work is if you ensure you are manually upgrading the CLIENT on your web severs to match the same release as your SERVER. That seems like a bad idea.

Please clarify?

rocksfrow: PGSQL_LIBPQ_VERSION _could_ work -- but only if you ensure ALL of your servers have the same client version (including your db servers). But most of the time this isn't true (not in my environment at least).

You guys not following me? Are you guys using multi-server environments or single?

rocksfrow: @riccardonar pretty sure you're confused.

@deeky666 I can totally implement your first suggestion. That way you're not going to run into the inefficiency if you're running on a recent/supported PG version -- but we'd still have backwards compatibility for older versions, with a small inefficiency.

I think we can agree there is no PERFECT solution unfortunately.

riccardonar: Sorry,
This is my case:
9.0.1 client
9.3 server
fails without "--option=" to set client encoding
 Il 02/apr/2015 20:24 "Kyle Renfrow" <notifications@github.com> ha scritto:

zeroedin-bill: @rocksfrow From what I can tell, the DSN is handled by libpq - the client_encoding option was [added in libpq v9.1](http://www.postgresql.org/docs/9.1/static/release-9-1.html).

What does your phpinfo say about the libpq version used by pdo_pgsql?

rocksfrow: @riccardonar you're throwing me for loops going back/forth with differing information.

You said earlier that you tested my fork -- and that it worked. Is that true or not?

riccardonar: Sure, it works because you don't use client encoding in connection..
Il 02/apr/2015 20:42 "Kyle Renfrow" <notifications@github.com> ha scritto:

riccardonar: With set names it works, but it make a query..
Il 02/apr/2015 20:45 "Riccardo Narduzzi" <riccardonar@gmail.com> ha scritto:

rocksfrow: @riccardonar EXACTLY. You just said:

````fails without "--option=" to set client encoding"````

My current fork doesn't pass the options param AT ALL. So if my fork is working for you -- then how are you saying it's failing without --option? That's false. Did you make a typo?

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: I will play around with PGSQL_LIBPQ_VERSION and how it behaves. My worry is relying on the versions when they can/will differ on multi-server environments.

I'll follow-up.

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @deeky666 do you get what I am saying about PGSQL_LIBPQ_VERSION?

Please let me know if I am confused -- but that only tells us what CLIENT is being used -- NOT what server is being connected to.

The point is -- just because you are connecting from a libpq client that supports client_encoding -- that doesn't mean it's going to work. The server has to suppport it of course. The client_encoding option wasn't added until 9.1 (both client and server).

If we make the logic based on the client version it means two things:

1) we introduce a bug where if you're connecting to, say an 8.4 server from a 9.1 client -- we'll end up passing client_encoding (because PGSQL_LIBPQ_VERSION tells us 9.1) -- but the 8.4 server is going to reject it.

Can somebody else chime in here? I want to make sure I am not the one understanding this wrong -- but I am pretty sure I am understanding this properly.

@riccardonar you're using 9.0 client anyway -- so regardless of what we did you're going to see the same behavior (IE, SET NAMES 'xxx' is the only way to set the character encoding reliably)

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @riccardonar remember -- the reason this all came up is because pgbouncer does not support the 'options' param. The options param is out of the question. We either need to use client_encoding, or SET NAMES.

I am all for building out support for both. I'll move forward with the fallback feature (on rejection of client_encoding) and reconnect AFTER you guys confirm PGSQL_LIBPQ_VERSION is not feasible (as it reflects the libpq version of the CONNECTING server, not the database server).

Comment (deeky666) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @rocksfrow I doubt that the server is checking the DSN option. I am pretty sure that it is the client that accepts or denies that option so working with the PHP constant would be the best option here. Catching an exception for BC on connect is a really bad idea as I think it causes more overhead than just performing `SET NAMES` (building the stack trace for exceptions is really expensive).

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @deeky666 not too sure about that based on my tests.

I am running final tests now to get this all straightened out. The code that's already been merged WORKS atleast (doesn't break pgbouncer, or postgres). 

I'll work on _reliably_ detecting when we should pass client_encoding, or when we should use SET NAMES.

I'm not convinced using PGSQL_LIBPQ_VERSION is sufficent for the reasons I just explained. I am running some tests now to confirm this for sure.

Comment (riccardonar) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: I don't say to use option param, i only say my case.
Also i think that the problem is the client..
Il 02/apr/2015 21:06 "Kyle Renfrow" <notifications@github.com> ha scritto:

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @deeky666 have you done any tests -- are you just saying this? I've done tests to prove otherwise.

I just did another test from other servers man -- and I am able to use my ORIGINAL fix (using client-encoding), from a client using 8.4!

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @deeky666 the point being -- checking PGSQL_LIBPQ_VERSION if it's >= 9.1 to decide whether or not to use client_encoding is obviously pointless... considering it works from 8.4

@riccardonar I am afraid you're giving us bad info... you've told me 3 different versions you are running on now. Can you please give us good/full info? Can you connect to your DB server from CLI, and copy/paste the client/server versions from there?

client_encoding is working fine from an 8.4 client -> 9.3 server.

This tells me it's the SERVER version that matters (like I've been saying) -- not the client.

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: From my test script (using doctrine with my ORIGINAL fix (simply using client_encoding instead of options).

client version: 8.4.20
Array
(
    [?column?] => 1
)
connected to 9.3

That array is just a test query I hit to make sure it was actually connected.

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @deeky666 I'd love it if you could confirm what your'e saying -- otherwise I want to move forward based on the facts I have (not guesses). If I am missing something I'd love for you guys to correct me.

Comment (riccardonar) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: I don't understand..
I've this:
9.0.1 client
9.3 server
connecting with --option=client_encoding works
connecting with client_encoding fails
connecting witout client_encoding (and with set names) works

Where i've say something wrong?

2015-04-02 21:43 GMT+02:00 Kyle Renfrow <notifications@github.com>:

> In lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
> <https://github.com/doctrine/dbal/pull/828#discussion_r27690085>:
>
> > @@ -53,6 +53,14 @@ public function connect(array $params, $username = null, $password = null, array
> >                  $pdo->setAttribute(PDO::PGSQL_ATTR_DISABLE_PREPARES, true);
> >              }
> >
> > +            /* defining client_encoding via SET NAMES to avoid inconsistent DSN support
> > +             * - the 'client_encoding' connection param only works with postgres >= 9.1
> > +             * - passing client_encoding via the 'options' param breaks pgbouncer support
> > +             */
> > +            if (isset($params['charset'])) {
> > +              $pdo->query('SET NAMES \''.$params['charset'].'\'');
>
> @deeky666 <https://github.com/deeky666> I'd love it if you could confirm
> what your'e saying -- otherwise I want to move forward based on the facts I
> have (not guesses). If I am missing something I'd love for you guys to
> correct me.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/doctrine/dbal/pull/828/files#r27690085>.
>



-- 
Riccardo Narduzzi

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @riccardonar you told me you had pg 9.3.4 in a previous comment, then you said no you had 9.0 -- then 9.3 haha -- that's all I meant.

Can you please provide more information/tests? Do you see why I am confused now? I have an OLDER client (8.4) -- and client_encoding is working with pg 9.3.

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @riccardonar can you COPY/PASTE the output from `psql -V` pretty please?

Comment (zeroedin-bill) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @rocksfrow @riccardonar the psql -V output is not relevant - run phpinfo(), and look at the pdo_pgsql section. The libpq listed there is what matters. It is possible to have multiple versions of libpq on your client...

Comment (deeky666) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @rocksfrow no I haven't done any tests I was just assuming that considering the fact that the DSN is client side. Seems you proved the opposite so forget about that. Also I think the PHP constant is not PDO specific as it is defined in the native `pgsql` extension. So let's just stick to `SET NAMES` for now if you don't have a better solution and stop making politics here ;)

Comment (riccardonar) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: Ok, the server is 9.3.4, and not 9.3 (i'm with cell and i think not important because the problem isn't between 9.3.*)

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @deeky666 , PGSQL_LIBPQ_VERSION is showing 8.4.20 -- and client_encoding is working connecting to 9.3. So obviously checking if PGSQL_LIBPQ_VERSION is >= 9.1 in order to decide to use client_encoding (or SET NAMES) is not a reliable solution (or we'll be making the query when we don't even need to!).

I found an article earlier with somebody complaining that PGSQL_LIBPQ_VERSION wasn't properly refecting the accurate version -- and the comments mentioned that PGSQL_LIBPQ_VERSION won't reflect upgrades that are made later.

So in my case, I think PGSQL_LIBPQ_VERSION is showing 8.4 (the out of the box version) -- even though I've replaced it with 9.x.

Moral of the story: PGSQL_LIBPQ_VERSION is not a reliable constant to use for this purpose.

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @zeroedin-bill I was asking him for the psql -V output to confirm he was actually using PG 9.3 (because he's gone back/forth on what version he's using). The libpq version isn't really what matters either -- it's the value that PHP has for the PGSQL_LIBPQ_VERSION constant (considering they were suggesting that's the best solution to the problem, that's the value that matters).

@deeky666 has confirmed he was just guessing -- so yes, like I said -- using PGSQL_LIBPQ_VERSION to check what client version isn't a valid solution.


@riccardonar yes you are right -- the minor version doesn't matter.

@deeky666

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: No need to continue this discussion @zeroedin-bill -- I had to explain to @deeky666 and @riccardonar that the suggested use of PGSQL_LIBPQ_VERSION is not a valid solution. They both disagreed so I had to prove why it's not a valid solution, hehe.

SO, at this point, the best solution would be to try to connect using client_encoding -- catch exception and detect client_encoding rejection. Reattempt connection without client_encoding and use SET NAMES instead if so.

That way -- for any recent installs, client_encoding will work as expected (and pgbouncer will be happy) -- and we'll retain support for older systems (with the SLIGHT drawback that SET NAMES is called after connection.

@deeky666 at this point -- you agree right?

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @zeroedin-bill and just FYI -- phpinfo() outputs the same thing as PGSQL_LIBPQ_VERSION

8.4.20

YET, client_encoding works fine for me. @riccardonar is saying it doesn't work 9.0 client but I haven't confirmed that. I am surprised it works with 8.4 but not 9.0 though.

Comment (zeroedin-bill) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @rocksfrow - deeky666 said,

> So let's just stick to SET NAMES for now

Doing the exception stuff is not a good option...

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: At least right now the class is doing what it should be doing, and works across the board.

We could improve it and use client_encoding only if we have a RELIABLE condition to match on (and fallback to set names)

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @riccardonar forget the psql -V -- what does this give you:

````php -r "echo PGSQL_LIBPQ_VERSION;````

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @riccardonar nevermind don't worry about it I just saw @deeky666 reply -- no that he realized PGSQL_LIBPQ_VERSION isn't a viable solution (like I said from the get-go), he's happy with how it is now.

thx.

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @deeky666 trust me -- I didn't just volunteer a few hours from my work-day to resolve this issue for 'politics' ha. Glad it's resolved. Take it ez.

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: For the record -- I prefer the solution that's in place over requiring double connections anyway.

Comment (riccardonar) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: My phpinfo img
and server postgres 9.3.4 is on another server

With
$dsn .= 'client_encoding=' . $params['charset'] . ' ';
i confirm that don't work.

**Comment by rocksfrow** [comment]
NOTE: if you don't want the overhead of `SET NAMES` and are OK with AUTO (your LOCALE is the same as your DB) -- just leave charset out of your config.

File: lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php (hunk 1 above)
Comment (deeky666) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @rocksfrow no offense! I just realized that this discussion is getting rather heavy for a rather trivial issue. I was just asking for additional opinions about the `STE NAMES` fix. I can live with that and obviously there is no better solution yet. So let's just keep it. Thanks for all your investigation on this!

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @riccardonar yeah so check it out -- I have a LOWER version (8.4.20) -- and client_encoding works fine for me.

screen shot 2015-04-02 at 4 30 18 pm

This just confirms what I was trying to tell you and @deeky666 -- this is not reliable data to make a decision off of.

@deeky666 confirmed he was just guessing anyway -- so we're leaving the current implementation (my fork). You can use my fork meanwhile until they release the next minor if you want.

Comment (deeky666) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: Backported to 2.5 branch.

Comment (rocksfrow) [inline lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php]: @deeky666 the convo got crazy because people jumping in giving inaccurate data haha.

I agree this is the best solution for now -- maybe eventually we'll have something we can rely on (like PGSQL_LIBPQ_VERSION being accurate).

I think the issue is that PGSQL_LIBPQ_VERSION does not update -- even after you upgrade (I read that somewhere).

Anyway -- thank you guys for an awesome class! I am glad to contribute.

We should be good to move forward now at least.

Have a good day.

**Comment by rocksfrow** [comment]
@zeroedin-bill I just saw you're from Maryland. Baltimore here!

**Comment by rocksfrow** [comment]
MD represent! haha

**Comment by zeroedin-bill** [comment]
@rocksfrow how bout dem O's

**Comment by zeroedin-bill** [comment]
@deeky666 should we backport this?

**Comment by rocksfrow** [comment]
The previous fix this is enhancing was backported so this one should be
otherwise that previous fix is going to remain flawed I'd think.

Note, this email was sent from my cell phone.

**Comment by rocksfrow** [comment]
(The previous fix of mine that deeky already backported wasn't working w
older versions of PG)

Note, this email was sent from my cell phone.

**Comment by rocksfrow** [comment]
Just out of curiosity - what are you guys doing when you backport? When you
do that does it mean you're making it available in previous 2.5.x releases?
I'm just am curious for my own dev processes.

Note, this email was sent from my cell phone.

**Comment by zeroedin-bill** [comment]
@rocksfrow deeky did backport it to 2.5 - i missed that. Take a look at the 2.5 branch.

**Comment by rocksfrow** [comment]
Yeah OK -- I thought he already did. So backporting it just means it's
going to be available in the NEXT 2.5.x release.

**Comment by rocksfrow** [comment]
Is that true? IE -- I can expect to see this fix in 2.5.2 I suppose?

**Comment by rocksfrow** [comment]
I just recently started using DBAL because it wasn't working with pgbouncer
so I just am using my own branch for now -- but I want to make sure I am
using the official branch again once it hits the next release.

Thanks for any clarification you may give. I am not sure about DBAL's
release schedule/practice.

**Comment by zeroedin-bill** [comment]
@rocksfrow if you're targeting the 2.5 *branch* in your composer.json, you will get the update - if you're targeting a specific point version, you'll have to wait for 2.5.2 and then target that point version...

**Comment by rocksfrow** [comment]
@zeroedin-bill I am using packagist, so I suppose the "2.5.x-dev" is what you're talking about. These are production servers so I'm definitely not pointing a dev branch I am pointing to 2.5.1. I will keep an eye out for the 2.5.2 release -- and meanwhile I'll just point to my own branch.

Thanks!

**Comment by rocksfrow** [comment]
@zeroedin-bill does the dbal have any standard release schedule -- or is it pretty random? Are there any processes that could lead to a predictable release date or should I not bother and just check in every now and then?

**Comment by rixbeck** [comment]
Hey @rocksfrow,

I'm also hitting this problem with pgbouncer and if you don't want to patch DBAL or use your branch - that is not always possible - there's a workaround here.
In `pgbouncer.ini` set 
```
ignore_startup_parameters=application_name,options
```
Then after connection made you may call `SET NAMES` query directly or you can build up a `postConnect` event listener that will be fired by DBAL.

rocksfrow:
@rixbeck that's a cool trick but I would imagine DBAL managers would agree -- we don't want to require config changes to your database stack to work with DBAL -- it should work out of the box.

I am simply using the 2.5-x branch which includes my merged in fix -- I am not sure if you noticed but we ended up merging in a second/better fix which is doing exactly that -- setting the charset via SET NAMES after connection.

I have been running 2.5-x branch in production within any issues -- just waiting on them to cut 2.5.2 so I can start using that (to avoid any bugs introduced) in the dev branch.

rixbeck:
Yeah, agree. Although it is really just a workaround until fix comes out with stable.
We are working on the Bolt CMS project [here](https://github.com/bolt/bolt) and DBAL is just a depending component under with `stable` required branches so currently I'm glad with this 'cause I can use pgbouncer in my projects without patching. :)

PR #829, opened 2015-04-02
Add PostgreSQL connection test with the new `charset` parameter

billschaller:
This PR adds 2 tests for connecting to pgsql via PDOPgSql, while using a charset parameter.

Related to #828, #823

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOPgsqlConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOPgsqlConnectionTest.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Functional\Driver;
+
+use Doctrine\DBAL\DriverManager;
+use Doctrine\DBAL\Platforms\PostgreSqlPlatform;
+use Doctrine\Tests\DbalFunctionalTestCase;
+use Doctrine\Tests\TestUtil;
+
+class PDOPgsqlConnectionTest extends DbalFunctionalTestCase
+{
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/DBAL/Functional/Driver/PDOPgsqlConnectionTest.php
```diff
@@ -0,0 +1,42 @@
+<?php
+/**
+ * PDOPgsqlConnectionTest.php
+ * Created by William Schaller
+ * Date: 4/2/2015
+ * Time: 12:08 PM
+ */
```
deeky666: Please remove this file docblock

tests/Doctrine/Tests/DBAL/Functional/Driver/PDOPgsqlConnectionTest.php
```diff
@@ -0,0 +1,42 @@
...
+        parent::setUp();
+
+        if ( ! $this->_conn->getDatabasePlatform() instanceof PostgreSqlPlatform) {
+            $this->markTestSkipped('PDOPgsql only test.');
+        }
+    }
+
+    public function testPdoPgsqlConnectionWithUTF8Charset()
+    {
+        $GLOBALS['db_charset'] = 'UTF8';
```
deeky666: I don't think it is a good idea to change global state here as all following tests will be affected by this.

zeroedin-bill: Ah, I meant to put an unset afterwards - would that make this acceptable?

PR #830, opened 2015-04-03
`README.md` nicer badges, cleanup, 2.3 dropped

TomasVotruba:
Ref https://github.com/doctrine/doctrine2/pull/1362

Is there any reason why coverage is not in here? Would be much more useful than dependency badge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,12 @@
 # Doctrine DBAL
 
+| [Master][Master] | [2.5][2.5] | [2.4][2.4] |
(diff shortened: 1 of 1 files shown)


deeky666:
@TomasVotruba I guess coverage is not available as a batch as it looks like DBAL is not yet configured properly for coveralls.io. Merging it as is for now. Thanks!

TomasVotruba:
@deeky666 Thanks!

PR #833, opened 2015-04-09
[DX] Added bool and int types

malarzm:
Lately we have done similar thing for MongoDB ODM in https://github.com/doctrine/mongodb-odm/pull/1073 so I thought it will add value to DBAL as well. I've also changed `integer` and `boolean` to `bool` and `int` when used in PHP context. If this will be merged I can submit a PR to ORM docs as well


guilhermeblanco:
AI TROAE

Ocramius:
Note: this is a backwards incompatible change for users that defined an `'int'` custom type

malarzm:
Got the reasoning, closing PR then

PR #837, opened 2015-04-15
Revert the addition of the wrong `bin` script to `composer.json`

stof:
The bin/doctrine-dbal.php file is not an executable file. Adding them as a bin in composer.json means that any composer install will trigger changes in the source when using symlinks because of the chmod. This makes things a pain when installing from source.
Thus, there is no valid reason to add it. It is absolutely not necessary when using a composer install. The issue requesting it previously is actually an issue in Laravel which replaces the proxy file/symlink generated by Composer with a copy of the original file, which of course cannot work because of paths used in require. But copying a second file does not help for that (unless in very specific cases). It only moves the issue until the next require call.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     "suggest": {
         "symfony/console": "For helpful console commands such as SQL execution and import of files."
     },
-    "bin": ["bin/doctrine-dbal", "bin/doctrine-dbal.php"],
+    "bin": ["bin/doctrine-dbal"],
     "autoload": {
         "psr-0": { "Doctrine\\DBAL\\": "lib/" }
     },


Ocramius:
>  an issue in Laravel which replaces the proxy file/symlink generated by Composer with a copy of the original file

#WTF

stof:
@Ocramius yeah, after seeing similar reports on several CLI tools, I finally managed to find that it was in the Laralevel community:

deeky666:
@stof thanks!

deeky666:
@stof does it need backport?

stof:
@deeky666 yes please

deeky666:
@stof just realized that it already is correct in 2.5 and 2.4 does not even have the `bin` property. So everything is fine I guess.

PR #840, opened 2015-04-25
Driver for SQLite3

BenMorel:
This PR adds an additional driver using the [SQLite3](http://php.net/manual/fr/book.sqlite3.php) class, in addition to the PDO_SQLITE driver.
## Why this driver?

Even though the PDO SQLite driver is already available to interact with SQLite databases, PDO currently has a big limitation: it [does not allow to load SQLite extensions](https://bugs.php.net/bug.php?id=64810).

This functionality is provided by the `SQLite3` class, which becomes your only option if you need to use your PHP application with an SQLite extension such as [SpatiaLite](http://www.gaia-gis.it/gaia-sins/).

### DIFF
--- a/lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
@@ -1,0 +1,78 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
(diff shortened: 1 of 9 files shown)


lib/Doctrine/DBAL/Types/DecimalType.php
```diff
@@ -49,6 +49,6 @@ public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $pla
      */
     public function convertToPHPValue($value, AbstractPlatform $platform)
     {
-        return (null === $value) ? null : $value;
+        return (null === $value) ? null : (string) $value;
```
BenMorel: I had to modify this type because [this test](https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/Functional/TypeConversionTest.php#L66) was still failing. This is because `SQLite3` returns numeric values as `float` for numbers that can be represented as a floating-point value without loosing precision, and as `string` for others.

I'm not sure if this line should be updated or if the test should be modified instead, but I suppose this should be all right as the Decimal type is always supposed to return a string.

BenMorel:
Any feedback on this PR?

lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
```diff
@@ -0,0 +1,78 @@
...
+ * The SQLite3 driver.
+ *
+ * @since 2.0
+ */
+class Driver extends AbstractSQLiteDriver
+{
+    /**
+     * @var array
+     */
+    protected $_userDefinedFunctions = array(
```
stof: no leading underscore and private visibility please.

lib/Doctrine/DBAL/Types/DecimalType.php
kimhemsoe: You should fix this in the driver for now. It is a known problem that we do not like anything else then strings. 
I think beberlei have some plans to fix it in the future. If we lucky someone will do it when they get annoyed enough by it :)

kimhemsoe:
:+1: when the underscore issue have been resolved

lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
BenMorel: Sorry, just copy/pasted that from another driver. Fixed.

BenMorel:
Good for me!

BenMorel:
Ping @kimhemsoe , can this PR be merged?

BenMorel:
Ping @stof @kimhemsoe , what's preventing this PR from being merged, 7 months later?

BenMorel:
Hi guys, I've worked hard on this PR (and others), and it's a pity no one seems to care about them.
Could you please review and at least tell me if there is a problem preventing you from merging this?

stof:
@BenMorel I'm not a merger on the DBAL repo

lib/Doctrine/DBAL/Types/DecimalType.php
zeroedin-bill: I remove this change and the test passes. Not sure why. I'm going to branch and rebase this PR with this change reverted for now and see if the tests pass on travis.

BenMorel: @zeroedin-bill That's weird, please keep me updated!

zeroedin-bill:
See #2264

PR #841, opened 2015-04-25
Documentation and code styling fixes

BenMorel:
This fixes:
- incorrect / incomplete / missing PHPdoc
- wrong case for method names
- unused imports

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1104,7 +1104,7 @@ class Connection implements DriverConnection
             $func($this);
             $this->commit();
         } catch (Exception $e) {
-            $this->rollback();
+            $this->rollBack();
             throw $e;
         }
     }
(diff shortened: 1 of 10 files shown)


deeky666:
@BenMorel thanks!

BenMorel:
@lastzero This PR has been [merged into master](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Connection.php#L1108), but [not into the 2.5 branch](https://github.com/doctrine/dbal/blob/2.5/lib/Doctrine/DBAL/Connection.php#L1111), not sure why. Maybe  @deeky666 can explain how/when merges between branches happen?

lastzero:
@BenMorel Thank you for the quick feedback! It's a bit confusing, even though PHP is not case sensitive.

Ocramius:
> not sure why

We backport only bugfixes that may not affect behavior for current users.
This will get into 2.6

PR #842, opened 2015-04-26
Fixed incorrect handling of single quotes in SQL-Strings

ancpru:
escaped by repeated single-quote (DBAL-1205)

### DIFF
--- a/tests/Doctrine/Tests/DBAL/SQLParserUtilsTest.php
+++ b/tests/Doctrine/Tests/DBAL/SQLParserUtilsTest.php
@@ -59,6 +59,19 @@ SQLDATA
             array('SELECT foo::date as date FROM Foo WHERE bar > :start_date AND baz > :start_date', false, array(46 => 'start_date', 68 =>  'start_date')), // Ticket GH-259
             array('SELECT `d.ns:col_name` FROM my_table d WHERE `d.date` >= :param1', false, array(57 => 'param1')), // Ticket DBAL-552
             array('SELECT [d.ns:col_name] FROM my_table d WHERE [d.date] >= :param1', false, array(57 => 'param1')), // Ticket DBAL-552
+            array(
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/DBAL/SQLParserUtilsTest.php
```diff
@@ -59,6 +59,19 @@ public function dataGetPlaceholderPositions()
...
             array('SELECT [d.ns:col_name] FROM my_table d WHERE [d.date] >= :param1', false, array(57 => 'param1')), // Ticket DBAL-552
+            array(
+<<<'SQLDATA'
+SELECT * FROM foo WHERE 
+bar = ':not_a_param1 ''":not_a_param2"'''
+OR bar=:a_param1
+OR bar=:a_param2||':not_a_param3'
+OR bar=':not_a_param4 '':not_a_param5'' :not_a_param6'
+OR bar=''
+OR bar=':a_param3
```
deeky666: This looks wrong. `OR bar=''` is correct and comparing to an empty string so the subsequent `OR bar=':a_param3` is invalid SQL

ancpru: Yep, you are right. In fact it should not have found this parameter, but I guess the regex considered it as outside quotes because it did not end with a quote. But I am happy enough if it does not find parameters in a "correct" string.

deeky666:
The original issue is #1151

deeky666:
@ancpru thanks!

deeky666:
Backported to 2.5 branch via https://github.com/doctrine/dbal/commit/8330aa4f0a62b48c072b652fb8457be55708dfd4, https://github.com/doctrine/dbal/commit/1a251129a951a7efa37bbc1b62018f60e1585f36

PR #843, opened 2015-04-28
fix TypeHinting for method parameter 'convertToPHPValueSQL'

kranikitao:


Ocramius:
Can't merge this, as it is a BC break.

PR #844, opened 2015-04-29
template1 as default database for PostgreSQL

pdesgarets:
Fixes (including but not limited to) https://github.com/doctrine/DoctrineBundle/issues/402 by connecting by default to 'template1' instead of the database with the same name as the user (PostgreSQL default in case of no dbname).

### DIFF
--- a/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
@@ -89,6 +89,9 @@ class Driver extends AbstractPostgreSQLDriver
         if (isset($params['dbname'])) {
             $dsn .= 'dbname=' . $params['dbname'] . ' ';
         }
+        else {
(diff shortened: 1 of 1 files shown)


sghribi:
:+1:

kiwidg:
I finally wonder if it is the good way, to "fake" à behavior from PostgreSQL. 
Please let me know what you think.

deeky666:
@kiwidg I think this is a fair solution to the problem. As we do not have an abstraction to the concept of connecting to "no database", this solution goes along nicely with other databases supporting connections without a database specified.
Will merge shortly.

deeky666:
@kiwidg merged manually with some cs fixes, documentation and test. Thanks!

kiwidg:
Thanks for the responsiveness and the additions @deeky666 !

acasademont:
BTW the patch works like a charm, thanks!!

PR #850, opened 2015-05-03
Allow to specify a charset and collation per column for Mysql

Freeaqingme:


zerocrates:
The MysqlPlatform already supports per-column collations. See:

https://github.com/doctrine/dbal/blob/v2.5.1/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php#L2548

and

https://github.com/doctrine/dbal/blob/v2.5.1/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L344

PR #317, opened 2013-05-15
Conditionaly upgrade utf8 to utf8mb4 for MySQL 5.5.3

nicolas-grekas:
See http://dev.mysql.com/doc/refman/5.5/en/charset-unicode-utf8mb4.html

As utf8mb4 is a superset of utf8, this should be transparent and backward compatible.
For those really requiring the "utf8" meant by MySQL, they can use explicitely the utf8mb3 charset.
But IMHO by default, Doctrine should really use utf8mb4, which is what everybody expect from a charset named "utf8".

PR #851, opened 2015-05-04
Change MySQL defaults from broken utf8 to fixed utf8mb4

DHager:
This is a conservative echo of https://github.com/doctrine/dbal/pull/317 .

Essentially MySQL's `uft8` character set is broken and does not support full UTF-8, and a better alternative, `utf8mb4` has existed for about five years now. When a 4-byte UTF-8 character comes in for a `utf8` table, by default MySQL will _truncate_ the string and log a warning. 

Insofar as Doctrine has any default configuration values for MySQL, I think `utf8mb4` is a better, safer choice. If anybody is running MySQL <5.5.3 and has problems, they should have a fairly easy time figuring out what's going wrong due to how distinctive the string is. (The other way around, searching for `utf8`, is not.)

### DIFF
--- a/docs/en/reference/security.rst
+++ b/docs/en/reference/security.rst
@@ -169,6 +169,6 @@ by passing the driver option "charset" to Doctrine PDO MySQL driver. Using SET N
     <?php    
     $conn = DriverManager::getConnection(array(
         'driver' => 'pdo_mysql',
-        'charset' => 'UTF8',
+        'charset' => 'utf8mb4',  // May need 'utf8' for MySQL < 5.5.3
     ));
 
--- a/lib/Doctrine/DBAL/Event/Listeners/MysqlSessionInit.php
+++ b/lib/Doctrine/DBAL/Event/Listeners/MysqlSessionInit.php
@@ -53,7 +53,7 @@ class MysqlSessionInit implements EventSubscriber
      * @param string         $charset   The charset.
      * @param string|boolean $collation The collation, or FALSE if no collation.
      */
-    public function __construct($charset = 'utf8', $collation = false)
+    public function __construct($charset = 'utf8mb4', $collation = false)
     {
         $this->_charset = $charset;
         $this->_collation = $collation;
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -476,14 +476,14 @@ class MySqlPlatform extends AbstractPlatform
 
         // Charset
         if ( ! isset($options['charset'])) {
-            $options['charset'] = 'utf8';
+            $options['charset'] = 'utf8mb4';
         }
 
         $tableOptions[] = sprintf('DEFAULT CHARACTER SET %s', $options['charset']);
 
         // Collate
         if ( ! isset($options['collate'])) {
-            $options['collate'] = 'utf8_unicode_ci';
+            $options['collate'] = 'utf8mb4_unicode_ci';
         }
 
         $tableOptions[] = sprintf('COLLATE %s', $options['collate']);
--- a/tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
@@ -287,4 +287,36 @@ class WriteTest extends \Doctrine\Tests\DbalFunctionalTestCase
 
     }
 
+    public function testMysqlUtf8Truncates()
+    {
+        $platform = $this->_conn->getDatabasePlatform();
+
+        if (!in_array($platform->getName(), array('mysql'))) {
+            $this->markTestSkipped('Related to MySQL only');
+        }
+
+        $table = new \Doctrine\DBAL\Schema\Table('truncation_test');
+        $table->addColumn('bad', 'string', array('charset' => 'utf8'));
+        $table->addColumn('good', 'string', array('charset' => 'utf8mb4'));
+
+        try {
+            $this->_conn->getSchemaManager()->dropTable($table);
+        } catch(\Exception $e) { }
+
+        foreach ($platform->getCreateTableSQL($table) as $sql) {
+            $this->_conn->exec($sql);
+        }
+        // F09D8C86 = UTF8 bytes for codepoint U+1D306, Center Tetragram
+        $fragment = "CONCAT('foo',UNHEX('F09D8C86'),'bar')";
+        $this->_conn->exec("INSERT INTO truncation_test (`bad`,`good`) VALUES ($fragment,$fragment);");
+
+        $lengths = $this->_conn->fetchAssoc('SELECT LENGTH(good) as good, LENGTH(bad) as bad FROM truncation_test');
+
+
+        $this->assertEquals(7, $lengths['good']);
+        // This is due to MySQL truncating the 4-byte character
+        $this->assertEquals(3, $lengths['bad']);
+
+    }
+
 }


Ocramius:
This seems reasonable to me, but only if an example string in a functional test is also being provided (in order to prove the validity of the change), especially around the data truncation you reported in the description

fprochazka:
Doctrine is already parsing the database version in order to create Platform objects. I would preffer a condition for this, like Nette did https://github.com/nette/http/issues/28

DHager:
@Ocramius: I don't think a test is worthwhile in this case, because there's no new behavior of Doctrine to test, nor any particular way to "solve" or react the issue in Doctrine. (Yet?) We'd just be writing tests to assert how third-party non-PHP code behaves.

Here's a simple way to validate the issue, with thanks to @mathiasbynens for the example problem-character:

    -- Create table using utf8 with two columns, one problematic and one OK
    CREATE TABLE `trunctest` (
        `utf8text` VARCHAR(255) CHARACTER SET 'utf8' COLLATE 'utf8_general_ci' NOT NULL,
        `mb4text` VARCHAR(255) CHARACTER SET 'utf8mb4' COLLATE 'utf8mb4_general_ci' NOT NULL
    );
        
    -- Hex is equivalent to 'foo𝌆bar' and makes things simpler 
    SELECT UNHEX('666F6FF09D8C86626172') INTO @astralstring;
    
    -- Will generate a warning, but still insert "foo" into the utf8 field
    INSERT INTO trunctest (`utf8text`,`mb4text`) VALUE (@astralstring,@astralstring);
    
    -- Now the two fields don't match up!
    SELECT utf8text, mb4text, (char_length(utf8text) = char_length(mb4text)) as success FROM trunctest;

    -- Cleanup
    DROP TABLE `trunctest`;

DHager:
@fprochazka: I notice there's a `MySQL57Platform` class, do you mean creating a new one like  `MySQL553Platform` that "layers on" the use of `utf8mb4`?

Ocramius:
> I don't think a test is worthwhile in this case

A change is only justified by a test here, IMO: we show why this default is
being changed and the test should fail (and decisions should be taken) if
the change is reverted.

You are not really testing the functionality, you are just hardening
support by preventing regressions ;-)

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

DHager:
@fprochazka: I've got a side-branch that uses a `MySQL553Platform` where each platform can override the default collation/charset, but unfortunately it causes a bunch of problems with the `AbstractMySQLPlatformTestCase`, because it assumes all MySQL platforms will always use one and only one collation and charset between them.

DHager:
Ah, yes, now the build-server is telling me something I should've remembered from my own work: `utf8mb4` introduces a new quirk, where `varchar(255)` cannot be indexed because it's (potentially) too long for `InnoDB`. 

It seems like there are three different choices, turning on MySQL-specific behavior:

* Continue hardcoding the charset/collation in a way which is difficult for people to override except by adding the right options to every single entity. (Or setting up an event-listener to do it for them.) 
 * Keep it as `utf8`, with the lurking incompatibility issue compared the UTF8 used by the rest of the tech-stack.
 * Change it to `utf8mb4`, and make people using `varchar(255)` indexes either specify a charset in the options for that field, or specify a lower length.
* Create a system that inspects the database-state on startup to determine which to use, which may bite users who never changed their settings from `latin1` and were depending on Doctrine to quietly always override it for them.

DHager:
@Ocramius, @fprochazka : At least for my immediate use-case, I think I've found a better solution with https://github.com/doctrine/DoctrineBundle/pull/420 instead.

Ocramius:
Reopening as it is worth discussing the upgrade.

MisatoTremor:
As a follow up on the post about ``varchar(255)``:

This is because the maximum size of an InnoDB index is 767 bytes. Thus in 3 byte utf8 a varchar type field can be at max 255 chars long (255 * 3 -> 765), while in 4 byte utf8 it can only be 191 chars long (191 * 4 -> 764).

Starting with MySQL 5.6.3 you can circumvent this by configuring it with the setting ``innodb_large_prefix=ON`` (defaults is ``OFF``) which itself needs ``innodb_file_format=Barracuda`` (default is ``Antelope``) _AND_ creating the table with the option ``ROW_FORMAT`` set to either ``DYNAMIC`` or ``COMPRESSED``. The maximum index size will then be 3072 bytes, so varchar can be  longer again.
Starting with MyQL 5.7.7 however ``innodb_file_format=Barracuda`` and ``innodb_large_prefix=ON`` will be the default values and even be removed as configurable options in a later release.

Maybe these things should be considered to build a stable solution for this.
I think creating a platform classes for at least 5.5.3 and 5.7.7 should be suitable.

What do you think @Ocramius?

deeky666:
This one is rather tricky and considering the amount of risks I doubt we can easily change this in 2.x. Also I wonder IF we changed it *somehow*, how we can keep BC and consistency here. When users would start to upgrade they'd have new tables with charset `utf8mb4` and old ones being stuck at `utf8`. So manual upgrade would be required here then. IMO the best option for Doctrine would be to not define a default charset and collation at all and let the user or database server decide about it. This might surely raise some other issues (like the index problem discussed) if a particular charset is used but the can of worms is open already anyways as you can define the default table charset and collation via table options.

Ocramius:
> IMO the best option for Doctrine would be to not define a default charset and collation at all and let the user or database server decide about it.

Disagree with this, according to history (remember `latin1`?)

deeky666:
@Ocramius yeah what about it? Is it wrong to use `latin1`? For example in my company we are dealing with german customers only and everything is bound to `latin1` from the customers side. We are sometimes having a hard job converting from `latin1` to `utf8` (our application and database default) forth and back. Also we could gain performance and disk space when using `latin1` instead of `utf8` not to mention sorting/comparing issues `latin1_german_ci` is what we actually need for germanized applications.

deeky666:
So what can we do in 2.x go get at least an intermediate solution? I tend to prefer the idea of introducing a `MySql553Platform`, making the default charset depend on the platform used (`utf8mb4` for MySQL > 5.5.3) and have a decent upgrade instruction. As a change of the default charset only affects newly created tables where a custom charset is not defined by the user, this could maybe acceptable?

realityking:
@deeky666 Even if all your customers in Germany, they might have a name that doesn't fit into latin1.

Storing non-legacy data in anything but Unicode really shouldn't be encouraged.

mcfedr:
I am very much in favour of this, there is small backwards compatibility issue that is worth talking about.

The default length of a `string` type is 255 in Doctrine, which fits in the maximum key length for InnoDB. If the default charset is going to be utf8mb the max key length is 191 characters.

This will cause the generated index SQL for many `string` columns to fail as it doesn't include key length.

MisatoTremor:
@mcfedr See https://github.com/doctrine/dbal/pull/851#issuecomment-104269536 for details about this.

ste93cry:
any news on this?

ThaDafinser:
Just also wanted to switch "easily", but got trapped by this issue.

Since there is already a Mysql57Platform...what about changing the default only there?
https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/MySQL57Platform.php

mnapoli:
@ThaDafinser thank you for that piece of code, that's perfect!

Ocramius:
FWIW, this would be an awesome addition if it wasn't for mysql's awful index size limits :-\

If we provide it out of the box, then the varchar size should go down by default.

mcfedr:
@Ocramius Or generate index declarations with size specified, although that might be another can of worms

mcfedr:
Also adding to this conversation the 'correct' collate to use with `utf8mb4` is `utf8mb4_unicode_520_ci`

afoeder:
@mcfedr do you have a source for the `utf8mb4_unicode_520_ci` statement?

mcfedr:
Unfortunately there isn't good information on the MySQL docs, but the best source I have found is the Wordpress bug tracker: https://core.trac.wordpress.org/ticket/32105#comment:3

Wordpress recently changed to using utf8mb4_unicode_520_ci by default. The key reason
>The problem stems from MySQL's collation behaviour - it treats all Unicode Supplementary Characters (which emoji fall under) as being equivalent. It's not until MySQL 5.6, wich the addition of the utf8mb4_unicode_520_ci collation that this changes.

pgnd:
What's the status on this 'defaults' merge, and utf8mb4 **_support_** in general?

Currently, in a new Symfony4 project
```
bin/console doctrine:phpcr:repository:init
```

returns
```
	Successfully registered system node types.
	Executing initializer: CmfRoutingBundle

	In ObjectManager.php line 859:
	  Error inside the transport layer: An exception occurred while executing 'SELECT id FROM phpcr_nodes WHERE path COLLATE utf8mb4_bin = ? AND workspace_name = ?' with params ["\/", "default"]:
	  SQLSTATE[42000]: Syntax error or access violation: 1253 COLLATION 'utf8mb4_bin' is not valid for CHARACTER SET 'utf8'

	In AbstractMySQLDriver.php line 121:
	  An exception occurred while executing 'SELECT id FROM phpcr_nodes WHERE path COLLATE utf8mb4_bin = ? AND workspace_name = ?' with params ["\/", "default"]:
	  SQLSTATE[42000]: Syntax error or access violation: 1253 COLLATION 'utf8mb4_bin' is not valid for CHARACTER SET 'utf8'

	In PDOStatement.php line 107:
	  SQLSTATE[42000]: Syntax error or access violation: 1253 COLLATION 'utf8mb4_bin' is not valid for CHARACTER SET 'utf8'

	In PDOStatement.php line 105:
	  SQLSTATE[42000]: Syntax error or access violation: 1253 COLLATION 'utf8mb4_bin' is not valid for CHARACTER SET 'utf8'
```

also noting
```
vendor/doctrine/dbal/UPGRADE.md

	## Creating MySQL Tables now defaults to UTF-8

	  If you are creating a new MySQL Table through the Doctrine API, charset/collate are now set to 'utf8'/'utf8_unicode_ci' by default. Previously the MySQL server defaults were used.
```

the current config includes,
```
config/packages/doctrine.yaml 
	...
	doctrine:
	    dbal:
	        default_connection: dev
	        connections:
	            dev:
	                ...
	                driver: 'pdo_mysql'
	                server_version: '5.7'
	                charset: utf8mb4
	                default_table_options:
	                    charset: utf8mb4
	                    collate: utf8mb4_unicode_ci
	            phpcr_dev:
	                ...
	                driver: 'pdo_mysql'
	                server_version: '5.7'
	                charset: utf8mb4
	                default_table_options:
	                    charset: utf8mb4
	                    collate: utf8mb4_unicode_ci                                                                                                                                              
	...

	doctrine_phpcr:
	    session:
	        backend:
	            type: doctrinedbal
	            connection: phpcr_dev
	...
```

mcfedr:
You seem to be mixing the two, might need to change your table charset to
utf8mb

pgnd:
@mcfedr 
mixing how/where?
both DBs have
```
	                charset: utf8mb4
	                default_table_options:
	                    charset: utf8mb4
	                    collate: utf8mb4_unicode_ci
```
consistent with MariaDB defaults ...

mcfedr:
Hard to say, the error message shows that the column or table is set to use
utf8, you need to look at the generated `create` queries to work out where
you've gone wrong

pgnd:
@mcfedr 
These are 'clean installed' DBs ... so far nothing other than install & (attempts to) initialize, etc.

e.g., that installation,

for the doctrine_phpcr clean installed tables,
```
MariaDB [test_sf4_phpcr]> show tables;
	+----------------------------+
	| Tables_in_test_sf4_phpcr   |
	+----------------------------+
	| phpcr_binarydata           |
	| phpcr_internal_index_types |
	| phpcr_namespaces           |
	| phpcr_nodes                |
	| phpcr_nodes_references     |
	| phpcr_nodes_weakreferences |
	| phpcr_type_childs          |
	| phpcr_type_nodes           |
	| phpcr_type_props           |
	| phpcr_workspaces           |
	+----------------------------+
	10 rows in set (0.00 sec)

MariaDB [test_sf4_phpcr]> SHOW FULL COLUMNS FROM phpcr_binarydata;
	+----------------+--------------+-----------------+------+-----+---------+----------------+---------------------------------+---------+
	| Field          | Type         | Collation       | Null | Key | Default | Extra          | Privileges                      | Comment |
	+----------------+--------------+-----------------+------+-----+---------+----------------+---------------------------------+---------+
	| id             | int(11)      | NULL            | NO   | PRI | NULL    | auto_increment | select,insert,update,references |         |
	| node_id        | int(11)      | NULL            | NO   | MUL | NULL    |                | select,insert,update,references |         |
	| property_name  | varchar(255) | utf8_unicode_ci | NO   |     | NULL    |                | select,insert,update,references |         |
	| workspace_name | varchar(255) | utf8_unicode_ci | NO   |     | NULL    |                | select,insert,update,references |         |
	| idx            | int(11)      | NULL            | NO   |     | 0       |                | select,insert,update,references |         |
	| data           | longblob     | NULL            | NO   |     | NULL    |                | select,insert,update,references |         |
	+----------------+--------------+-----------------+------+-----+---------+----------------+---------------------------------+---------+
	6 rows in set (0.00 sec)
```

for the 'main' ORM tables
```
MariaDB [test_sf4]> show tables;
	+-------------------------------+
	| Tables_in_test_sf4            |
	+-------------------------------+
	| acl_classes                   |
	| acl_entries                   |
	| acl_object_identities         |
	| acl_object_identity_ancestors |
	| acl_security_identities       |
	| fos_group                     |
	| fos_user                      |
	| fos_user_group                |
	| fos_user_user                 |
	| fos_user_user_group           |
	+-------------------------------+
	10 rows in set (0.00 sec)

MariaDB [test_sf4]> SHOW FULL COLUMNS FROM acl_classes;
	+------------+------------------+--------------------+------+-----+---------+----------------+---------------------------------+---------+
	| Field      | Type             | Collation          | Null | Key | Default | Extra          | Privileges                      | Comment |
	+------------+------------------+--------------------+------+-----+---------+----------------+---------------------------------+---------+
	| id         | int(10) unsigned | NULL               | NO   | PRI | NULL    | auto_increment | select,insert,update,references |         |
	| class_type | varchar(200)     | utf8mb4_unicode_ci | NO   | UNI | NULL    |                | select,insert,update,references |         |
	+------------+------------------+--------------------+------+-----+---------+----------------+---------------------------------+---------+
	2 rows in set (0.00 sec)

	MariaDB [test_sf4]>
```

morozov:
The change is obsolete as of https://github.com/doctrine/dbal/pull/4644.

PR #852, opened 2015-05-06
Add SQL Anywhere builds to Travis

deeky666:
This is an attempt to enable functional testing for SQL Anywhere 12 + 16 on Travis.
It uses precompiled `sqlanywhere` PHP extensions for 5.3, 5.4, 5.5 and 5.6. Server-wise it uses preinstalled, lightweight packages (around 15 MB). This allows for a fast environment setup.
The packages reside on my Google Drive account, the URLs for the server packages are secured by Travis encryption.
The only downside of using secured environment variables is, that builds won't work on forks because of different SSH keys but that is a fair trade-off IMO.

/cc @zeroedin-bill @Ocramius @beberlei thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,18 +9,29 @@ php:
   - hhvm
 
 env:
-  - DB=mysql
-  - DB=pgsql POSTGRESQL_VERSION=9.1
-  - DB=pgsql POSTGRESQL_VERSION=9.2
-  - DB=pgsql POSTGRESQL_VERSION=9.3
-  - DB=pgsql POSTGRESQL_VERSION=9.4
-  - DB=sqlite
-  - DB=mysqli
+  global:
+    # $SQLANYWHERE_12_GDRIVE_ID
+    - secure: "TaIWN0OTS1fGCdvpDpRIX/svYb3pjwBqISeN6KMspqn8QB2bkqRWn+A1s9lAZw/1Ou27Qep2sRvo3lYBEssz2ZC84/qlt520bga5P/DRBLzJWjNowFh1c8jti6WiAHCnhoEO+HoUa/4ZFdeNA71RSL+8Fb8slKjC/InrwrYTmQo="
+    # $SQLANYWHERE_16_GDRIVE_ID
+    - secure: "jMGH+6QisWj5XDLq5seTdJjQa8IXqqjjggufVOM0iqK3yqsCm/3D+XgXiREOABvV2iWhbR2j0jvrOGqYTaCX2mp/HSj4w86zFl+IDrsC7IcozoaJg2I4fBYn56jBEsk/wsMpGKBdH7TW1bMWyvJ1+3MCUVsEzTHYdpauVydv7Ek="
+    - LD_LIBRARY_PATH=`pwd`/sqlanywhere/lib64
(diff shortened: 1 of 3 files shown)


deeky666:
Meh seems it does not work on PRs either...

tests/travis/install_sqlanywhere.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+wget -qO- "https://drive.google.com/uc?export=download&id=$SQLANYWHERE_GDRIVE_ID" | tar xz # SQL Anywhere server
```
Ocramius: Uhm... So we depend on a file maintained by you there?

stof:
@deeky666 it works only for PRs coming from the main repo. Secure variables are not available for PRs coming from forks as the .travis.yml file cannot be trusted in this case (a PR could add a line displaying the content of the variable)

.travis.yml
```diff
@@ -64,6 +79,14 @@ matrix:
       env: DB=pgsql POSTGRESQL_VERSION=9.3 # driver currently unsupported by HHVM
     - php: hhvm-nightly
       env: DB=pgsql POSTGRESQL_VERSION=9.4 # driver currently unsupported by HHVM
+    - php: hhvm-nightly
+      env: DB=sqlanywhere SQLANYWHERE_GDRIVE_ID=$SQLANYWHERE_12_GDRIVE_ID # driver currently unsupported by HHVM
+    - php: hhvm-nightly
```
stof: hhvm-nightly is gone

deeky666:
@stof yep noticed that already. Any other idea?

.travis.yml
stof: ah no. I rmeoved it for the ORM, not for DBAL actually.

tests/travis/install_sqlanywhere.sh
deeky666: @Ocramius well doesn't have to be me but that's the only option to enable builds that don't require a long setup time. Basically what I did is:

- Download server package from SAP
- Install with developer license key (which you get after registration)
- Remove unnecessary files to shrink size
- tar folder
- upload to GDrive

"Maintenance" is a bit of an exaggeration here IMO because the package won't change. It simply works...

Better ideas are always welcome :)

stof:
@deeky666 we would need somethign which does not depend on secure variables IMO

.travis.yml
stof: https://github.com/doctrine/dbal/pull/853 simplifies this config now

deeky666: I know but it's currently still present in the `travis.yml` so I adopted it for now. Can be removed completely later with another PR IMO.

stof: you can rebase now to fix conflicts and this will be simpler here

zeroedin-bill:
@stof @deeky666 @Ocramius after some discussion with SAP folks, there is really no other solution.

Unit testing the platform only on PRs originating from the doctrine repositories is quite a lot better than not unit testing against the platform at all. It appears as though this will be the absolute best we can do with this vendor.

deeky666:
Closing this as `Won't fix` for now. This can be addressed in a future external Doctrine CI.

PR #853, opened 2015-05-06
Remove HHVM-nightly builds

stof:
This is similar to https://github.com/doctrine/doctrine2/pull/1401

This also simplies the allowed failures for HHVM


deeky666:
@stof thanks!

PR #854, opened 2015-05-09
add `DateInterval` type

vbartusevicius:
Added DateInterval Type to store/receive PHP DateInterval objects.
Internally DateInterval is converted to string to store in VARCHAR field.
For maximum performance full DateInterval format is used when converting to database value.

### DIFF
--- a/lib/Doctrine/DBAL/Types/DateIntervalType.php
+++ b/lib/Doctrine/DBAL/Types/DateIntervalType.php
@@ -9,8 +9,6 @@ use Doctrine\DBAL\Platforms\AbstractPlatform;
  */
 class DateIntervalType extends Type
 {
-    const DATEINTERVAL_PATTERN = '#(?P<date>\d{4}-\d{2}-\d{2}).(?P<time>\d{2}:\d{2}:\d{2})#';
-
     /**
      * {@inheritdoc}
      */
@@ -35,9 +33,10 @@ class DateIntervalType extends Type
         $spec = null;
         if ($value !== null) {
             /** @var \DateInterval $value */
-            $spec = str_pad($value->y, 4, '0', STR_PAD_LEFT) . '-'
+            $spec = 'P'
+                . str_pad($value->y, 4, '0', STR_PAD_LEFT) . '-'
                 . $value->format('%M') . '-'
-                . $value->format('%D') . ' '
+                . $value->format('%D') . 'T'
                 . $value->format('%H') . ':'
                 . $value->format('%I') . ':'
                 . $value->format('%S')
@@ -56,11 +55,8 @@ class DateIntervalType extends Type
             return $value;
         }
 
-        if (preg_match(self::DATEINTERVAL_PATTERN, $value, $parts) !== 1) {
-            throw ConversionException::conversionFailedFormat($value, $this->getName(), 'Y-m-d H:i:s');
-        }
         try {
-            $interval = new \DateInterval('P' . $parts['date'] . 'T' . $parts['time']);
+            $interval = new \DateInterval($value);
         } catch (\Exception $e) {
             throw ConversionException::conversionFailedFormat($value, $this->getName(), 'PY-m-dTH:i:s');
 
--- a/lib/Doctrine/DBAL/Types/DateIntervalType.php
+++ b/lib/Doctrine/DBAL/Types/DateIntervalType.php
@@ -1,0 +1,12 @@
+<?php
+
+
+namespace Doctrine\DBAL\Types;
+
+/**
+ * 
+ */
+class DateIntervalType 
+{
+
+}
--- a/lib/Doctrine/DBAL/Types/DateIntervalType.php
+++ b/lib/Doctrine/DBAL/Types/DateIntervalType.php
@@ -10,8 +10,6 @@ use Doctrine\DBAL\Platforms\AbstractPlatform;
  */
 class DateIntervalType extends Type
 {
-    const DATEINTERVAL_PATTERN = '#(?P<date>\d{4}-\d{2}-\d{2}).(?P<time>\d{2}:\d{2}:\d{2})#';
-
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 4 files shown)


v-bartusevicius:
Any comments or decision to merge this feature?

portals12:
Maybe it is better to store this as int? Your solution does not allow to use most of database features like sorting.

fprochazka:
@portals12 well, I'd argue you can sort by it, if it has always constant length. But you have a valid point. Is there any SQL standard for storing date intervals, or is this only a PHP thing?

v-bartusevicius:
`DateInterval`'s serialized value is `PxYxMxDTxHxMxS` where `x` represents any positive integer - this is the only one possible full representation of `DateInterval` in PHP, so converting it to `integer` would impact it's readability, manual editing and common understanding. At least I would think so.

@fprochazka  - I suggest You to compare this type to dbal's `ArrayType` or `JsonArrayType` - there is no SQL standard to store `arrays` as there is no standard for `DateInterval` too.

zeroedin-bill:
If you change this to use the date time format (P[YYYY]-[MM]-[DD]T[hh]:[mm]:[ss]), sorting will work.

That format is supported by PHP and will store and sort nicely as a string.

See http://php.net/manual/en/dateinterval.construct.php at the end of the section about the interval_spec parameter.

v-bartusevicius:
Sorry for a shower of commits - accidently rebased, so tried to undo...

v-bartusevicius:
Any news on this change?

v-bartusevicius:
?

v-bartusevicius:
Anyone willing to give some feedback?

zeroedin-bill:
@v-bartusevicius this looks OK to me - it should work fine as a type, and it should present no problems with sorting.

lib/Doctrine/DBAL/Types/DateIntervalType.php
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToPHPValue($value, AbstractPlatform $platform)
+    {
+        if ($value === null || $value instanceof \DateInterval) {
+            return $value;
+        }
+
+        if (preg_match(self::DATEINTERVAL_PATTERN, $value, $parts) !== 1) {
```
zeroedin-bill: This is not needed - if the DateInterval constructor cannot recognize the interval string, it will throw an exception. No need to run the regex.

beberlei: if the constructor fails anyways on invalid format, why this additional step here? this just costs performance.

lib/Doctrine/DBAL/Types/DateIntervalType.php
```diff
@@ -0,0 +1,71 @@
...
+namespace Doctrine\DBAL\Types;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+
+/**
+ * Type that maps interval string to a PHP DateInterval Object.
+ */
+class DateIntervalType extends Type
+{
+    const DATEINTERVAL_PATTERN = '#(?P<date>\d{4}-\d{2}-\d{2}).(?P<time>\d{2}:\d{2}:\d{2})#';
```
zeroedin-bill: Not needed - see below

lib/Doctrine/DBAL/Types/DateIntervalType.php
v-bartusevicius: In db the DateTime is stored as string `0002-00-01 01:02:03`, so regex here primary separates `date` and `time` parts in given string. I could use `explode(' ', $str)` here if it's needed.

zeroedin-bill: instead of storing it as 0002-00-01 01:02:03, store it as P0002-00-01T01:02:03 and then you won't need to convert like this...

v-bartusevicius:
@zeroedin-bill, could you please give some feedback?

lib/Doctrine/DBAL/Types/DateIntervalType.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        return Type::DATEINTERVAL;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform)
+    {
+        return $platform->getVarcharTypeDeclarationSQL($fieldDeclaration);
```
zeroedin-bill: Since the data format's length will never change, the generated column should have a fixed length. See DBAL\Types\GuidType for example of a similar situation.

zeroedin-bill:
@deeky666 I think this is almost ready to merge, what are your thoughts on inclusion? I think this makes sense as a DBAL type, given that the PHP type exists, and the database representation as implemented is sortable/filterable/stable for all intents and purposes.

zeroedin-bill:
@v-bartusevicius please document the new type appropriately in docs/en/reference/types.rst as well.

v-bartusevicius:
@zeroedin-bill, please take a look at recent changes you requested.

zeroedin-bill:
@v-bartusevicius Great work :+1:

stof:
The form type should return true from ``requiresSQLCommentHint`` otherwise it will not work nicely with the schema tool

v-bartusevicius:
@stof, could You please explain more clearly?

deeky666:
@v-bartusevicius because otherwise the new type won't be mapped back from existing database correctly when the schema manager introspects the database. It will detect a `VARCHAR` column and map it as `StringType` as it is not commented. So @stof is right this has to be fixed.

v-bartusevicius:
Created new commit for this: https://github.com/doctrine/dbal/pull/899

mickadoo:
@vbartusevicius I tried to use the `DateIntervalType` but encountered a problem based on how php DateInterval creation behaves.

If I did this:

```
$interval = \DateInterval::createFromDateString('32 days');
```

I works and the data is saved to the database, but on retrieval it will fail because 

```
$interval = new \DateInterval(P0000-00-32T00:00:00);
```

I'm not sure what's standard practice for Types, but maybe there could be validation before persisting it.

vbartusevicius:
Please read http://php.net/manual/en/dateinterval.construct.php - there is a note to be careful with date time interval representation:

> The specification can also be represented as a date time. A sample of one year and four days would > be P0001-00-04T00:00:00. But the values in this format can not exceed a given period's > roll-over-point (e.g. 25 hours is invalid). 

`32 days` is invalid day number.
i.e. `new \DateInterval('P0000-00-31T00:00:00');` would work as expected.

mickadoo:
Exactly, I don't think it's a problem with anything in doctrine, just that it's strange PHP will allow you to construct it with 32 days if it's using the `createFromDateString` method and wanted to point it out.

PR #855, opened 2015-05-16
Connection::ping() no longer produces warnings on connection timeout

fprochazka:
With PDOMySql (I haven't tested other drivers), the ping call produces warnings, that IMHO should be handled by the library.

Because the test suite is ignoring warnings and notices, I assume nobody noticed this before. But our application has strict no-errors policy.
I've created similar temporary-hotfix https://github.com/Kdyby/Doctrine/commit/f7250e5b771eb1ba6c0abe23cb2dc689247d1b4c for my integration lib with Nette, but I believe that it's best to handle this directly in doctrine.

I was considering adding a global error handler to bootstrap file, but that broke several other tests so that should be IMHO taken care of in separate pullrq.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1589,11 +1589,17 @@ class Connection implements DriverConnection
             return $this->_conn->ping();
         }
 
+        set_error_handler(function ($severity, $message) {
+            throw new \PDOException($message, $severity);
+        });
+
         try {
             $this->query($this->platform->getDummySelectSQL());
+            restore_error_handler();
 
             return true;
         } catch (DBALException $e) {
+            restore_error_handler();
             return false;
         }
     }
--- a/tests/Doctrine/Tests/DBAL/Functional/Driver/Mysql/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Driver/Mysql/ConnectionTest.php
@@ -1,0 +1,85 @@
+<?php
+namespace Doctrine\Tests\DBAL\Functional\Driver\Mysql;
+
+use Doctrine\DBAL\Driver\PDOConnection;
+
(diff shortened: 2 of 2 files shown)


lib/Doctrine/DBAL/Connection.php
```diff
@@ -1589,11 +1589,17 @@ public function ping()
             return $this->_conn->ping();
         }
 
+        set_error_handler(function ($severity, $message) {
+            throw new \PDOException($message, $severity);
```
deeky666: I doubt it is the correct way to set an error handler here that throws a `\PDOException` as not all of the drivers are PDO drivers. Furthermore `Doctrine\DBAL\Connection` should only throw `Doctrine\DBAL\DBALException` based exceptions that wrap the original driver exception.
This is more of a general PDOMySQL driver issue which should be solved in the driver implementation itself (if possible).
In the end it looks like a [PHP bug](https://bugs.php.net/bug.php?id=63812) because when setting `\PDO::ERRMODE_EXCEPTION` on a PDO connection, no warnings are expected to be risen.

fprochazka: You're right, it should throw an exception, but it's triggering a warning. Anyway, Doctrine should contain a fix for the driver.

I was thinking of implementing it directly in the `Doctrine\DBAL\Driver\PDOConnection::query()`, but changing error handler for every query wouldn't be very fast.

zeroedin-bill: Given that this is a PHP bug, I think it would be better to work around this in PDOMySQL driver, and revert when the PHP bug is fixed (probably a long time)

fprochazka: I was gonna implement this in `PDOMySQL` but I wasn't sure about the performance impact of attaching and reseting an error handler on every query.

morozov:
Obsolete as of https://github.com/doctrine/dbal/pull/4128.

PR #858, opened 2015-05-21
Update DateType.php

xoeoro:


Ocramius:
A silent failure is not acceptable

stof:
Thus this also removes the support of immutable datetimes

Ocramius:
> Thus this also removes the support of immutable datetimes

Very good point indeed :-)

PR #862, opened 2015-05-21
Quote table and constraint names in drop foreign key / drop constraint SQL

mpoiriert:
On the MySQLPlatform::getDropForeignKeySQL table name will not be escaped if the name is passed instead of the object table itself.

Since the getLocalTableName use the localTable property the object is always available, there is no reason not to use it.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/SchemaDiff.php
+++ b/lib/Doctrine/DBAL/Schema/SchemaDiff.php
@@ -154,7 +154,7 @@ class SchemaDiff
 
         if ($platform->supportsForeignKeyConstraints() && $saveMode == false) {
             foreach ($this->orphanedForeignKeys as $orphanedForeignKey) {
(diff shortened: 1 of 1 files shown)


deeky666:
Requires a test case.

deeky666:
Merge manually. Added quotation fixed in the platforms, too and fixed failing test due to this patch.

deeky666:
Backported to branch 2.5 via https://github.com/doctrine/dbal/commit/acad5d6e2fd359a39f5a9e51afb0280141139ae4, https://github.com/doctrine/dbal/commit/a9e08c8c17e7d6e8aca944c93c15d1e451c665a8, https://github.com/doctrine/dbal/commit/6460e4aa2df2b3d0dc833a88926656b30ace6872

mdular:
Hey, this also affects Postgres. I was a bit surprised to see that the fix has existed for 9 months already - is there a rough estimate on when it will be tagged?

Keep up the great work! :)

PR #863, opened 2015-05-21
Strip leading slash of databasename from URL

kingcrunch:
This is actuall the commit for [DBAL-1234](http://www.doctrine-project.org/jira/browse/DBAL-1234)

When using a custom driver via `driverClass` in all (some?) cases one cannot set
the scheme properly. In the earlier implementation when the scheme is missing the
`DriverManager` leaves the leading slash in the path, because it silently assumes,
that this is a SQLite-connection. With custom drivers this leads to invalid database
names.

Additionally this takes care, that if one specifies the driver via configuration key
`driver`, but the connection with scheme-less URL it ends up in an invalid database
name too    

```
driver: pdo_mysql
url: //user:pass@localhost/database
```

Another solution is to introduce a special `custom`-scheme, that doesn't point to a driver, but declares, that `driverClass` is required

```
custom://foo:bar@localhost:123/my_db
```

However, this would not take care of the other use-case,

### DIFF
--- a/lib/Doctrine/DBAL/DriverManager.php
+++ b/lib/Doctrine/DBAL/DriverManager.php
@@ -231,8 +231,15 @@ final class DriverManager
         
         // (pdo_)?sqlite3?:///... => (pdo_)?sqlite3?://localhost/... or else the URL will be invalid
         $url = preg_replace('#^((?:pdo_)?sqlite3?):///#', '$1://localhost/', $params['url']);
-        
-        $url = parse_url($url);
+
+        // PHP<5.4.8 doesn't parse schemeless urls properly
+        if (strpos($url, '//') === 0 && version_compare(PHP_VERSION, '5.4.8') < 0) {
+            $url = parse_url("fake:$url");
+            unset($url['scheme']);
+        } else {
+            $url = parse_url($url);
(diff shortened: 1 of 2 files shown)


zeroedin-bill:
This change needs tests. Please also remove the whitespace changes from the changeset.

KingCrunch:
Hi @zeroedin-bill 
Thanks for the feedback. I'd made the changes.

I've change the order of the `if`-statement, because this way it is easier to avoid the `undefined index 'driver'` of the `strpos()`-call.

KingCrunch:
Hi @zeroedin-bill 

The failing tests are unrelated and mostly timeout-issues. There is one other issue with a mock regarding DB2-driver or something like that.

However, there is also a different fix regarding `parse_url()`, which was broken before PHP 5.4.8.

deeky666:
Reference: #1183

deeky666:
@KingCrunch taking this over. Working on a minor overall improvement of the URL parsing. Will include your fixes there. Thanks!

KingCrunch:
Sure. For reference: The other PR is #2287 :)

Ocramius:
Ported into #2287

PR #864, opened 2015-05-22
Fix undefined notices within MasterSlaveConnection

randomonkey:
To fix the following notice-level errors:

Notice: Undefined property: Doctrine\DBAL\Connections\MasterSlaveConnection::$_conn in vendor/doctrine/dbal/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php line 154
Notice: Undefined index: slave in vendor/doctrine/dbal/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php line 165

### DIFF
--- a/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
+++ b/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
@@ -151,7 +151,7 @@ class MasterSlaveConnection extends Connection
         // If we have a connection open, and this is not an explicit connection
         // change request, then abort right here, because we are already done.
         // This prevents writes to the slave in case of "keepSlave" option enabled.
-        if ($this->_conn && !$requestedConnectionChange) {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -151,7 +151,7 @@ public function connect($connectionName = null)
         // If we have a connection open, and this is not an explicit connection
         // change request, then abort right here, because we are already done.
         // This prevents writes to the slave in case of "keepSlave" option enabled.
-        if ($this->_conn && !$requestedConnectionChange) {
+        if (isset($this->_conn) && $this->_conn && !$requestedConnectionChange) {
```
stof: this change is wrong IMO. The property exists in the class (we already fixed the bug unsetting the property rather than the value previously), and the case of null is already covered by the condition before the patch

PR #865, opened 2015-05-22
Added random number generation support

ghost:
This adds generic support for the various random number generation mechanisms on all the platforms.  This is a precursor commit to adding a DQL RANDOM() expression.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -2661,6 +2661,16 @@ abstract class AbstractPlatform
     }
 
     /**
+     * Returns the SQL specifier for the platform to get a random number
+     *
+     * @return string
+     */
+    public function getRandomFunctionSQL()
+    {
+        return 'RAND()';
+    }
+
+    /**
      * Returns the SQL for a given transaction isolation level Connection constant.
      *
      * @param integer $level
--- a/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
@@ -481,6 +481,14 @@ class OraclePlatform extends AbstractPlatform
     }
 
     /**
+     * {@inheritDoc}
+     */
+    public function getRandomFunctionSQL()
+    {
+        return 'DBMS_RANDOM.VALUE';
+    }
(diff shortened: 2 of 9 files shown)


lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2661,6 +2661,16 @@ public function getCurrentTimestampSQL()
     }
 
     /**
+     * Returns the SQL specifier for the platform to get a random number
+     *
+     * @return string
+     */
+    public function getRandomFunctionSQL()
+    {
+        return 'RAND()';
```
deeky666: I doubt this is SQL-92 standard. Maybe better to throw an exception here in the first place.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -1054,6 +1054,14 @@ public function getReadLockSQL()
     /**
      * {@inheritDoc}
      */
+    public function getRandomFunctionSQL()
+    {
+        return 'random()';
```
deeky666: The documentation states it produces a number like this:

`random value in the range 0.0 <= x < 1.0`

Not really intended, is it?

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -481,6 +481,14 @@ public function getDropViewSQL($name)
     }
 
     /**
+     * {@inheritDoc}
+     */
+    public function getRandomFunctionSQL()
+    {
+        return 'DBMS_RANDOM.VALUE';
```
deeky666: Documentation states it produces a number like this:

`DBMS_RANDOM.VALUE produces numbers in [0,1) with 38 digits of precision.`

Not really inteneded, is it?

deeky666:
You should clarify on your exact purpose. "random number" is somewhat unclear. Seems like most DB vendors support random decimal numbers between `0` and `1` and integers values between their specific integer boundaries...

zeroedin-bill:
I'm not sure we should include this at all. Using DBMS random number features in a query is almost never a good idea.

deeky666:
Closing accodring to https://github.com/doctrine/doctrine2/issues/1412

PR #866, opened 2015-06-07
Fix issue with  zero date for DateTime Type

Fedik:
Fix PHP issue with  zero date for DateTime Type,
when PHP return `-0001-11-30 00:00:00` for `0000-00-00 00:00:00` value

### DIFF
--- a/lib/Doctrine/DBAL/Types/DateTimeType.php
+++ b/lib/Doctrine/DBAL/Types/DateTimeType.php
@@ -49,8 +49,12 @@ class DateTimeType extends Type
      */
     public function convertToDatabaseValue($value, AbstractPlatform $platform)
     {
-        return ($value !== null)
-            ? $value->format($platform->getDateTimeFormatString()) : null;
+        if($value === null) {
+            return null;
+        }
+
+        $value = $value->format($platform->getDateTimeFormatString());
+        return $value === '-0001-11-30 00:00:00' ? '0000-00-00 00:00:00' : $value;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


kimhemsoe:
I do not believe that we should handle a mysql and schema specific issue in our types.
If you do have the issue i think a better approach would be that you created a new type and overwriting the default one by registering it with the same name.

Ocramius:
This is a broken mysqlism indeed. 0000-00-00 00:00:00 is not a valid date,
and will also cause errors depending on mysql settings.

Fedik:
> created a new type and overwriting the default one 

yes I alredy use it,
but thought would be good to have this fixed by in the library

I do not know better way to fix it,
maybe something like $platform->getZeroDate() wold be better

kimhemsoe:
You can hack it until you get your data fixed by overwriting the datetime type.

Read more at http://doctrine-dbal.readthedocs.org/en/latest/reference/types.html#custom-mapping-types

Instead of Type::addType('datetime', YouClass) you would need to call Type::overrideType('datetime', YouClass')

Fedik:
@kimhemsoe  yes I use it  :wink:

Some more example, for this issue:

Example exist entity with `endDate` field:
```xml
...
<field name="endDate" type="datetime" nullable="true" />
....
```
you save it with empty value in `endDate`

Then tries search all items with empty `endDate`:
```php
...
$criteris->where(Criteria::expr()->eq('endDate', new \DateTime('0000-00-00 00:00:00'));
...
```
and in result you will get nothing, because final SQL Query looks like: `... WHERE endDate = -0001-11-30 00:00:00 ...`

Ocramius:
@Fedik why are you saving `0000-00-00 00:00:00` instead of `null`?

Also, `0000-00-00 00:00:00` is not valid: `0000-01-01 00:00:00` is ;-)

Fedik:
@Ocramius good point :smile:

PR #867, opened 2015-06-12
Add test for schema diffing on a table with a renamed foreign key column referencing a renamed table

billschaller:
~~In situations where SQL is generated with SchemaDiff::toSaveSql, foreign keys for columns that have been renamed or removed were not being dropped if the referenced table was orphaned.~~

This PR adds a test to ensure that when a table should be renamed, and the FK columns referencing it should also be renamed, the FKs are properly removed and readded. This PR formerly fixed an issue where this was not successful, but it has since been fixed upstream, so this PR has just been pared back to the added test.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
+++ b/tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
@@ -1232,4 +1232,44 @@ class ComparatorTest extends \PHPUnit_Framework_TestCase
             array('0', 'foo', false),
         );
     }
+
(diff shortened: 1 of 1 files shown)


deeky666:
@zeroedin-bill I'm sorry but this requires a rebase now.

zeroedin-bill:
@deeky666 Somewhere upstream this was fixed, so I just pared down this pr to the added test and squashed commits.

deeky666:
@zeroedin-bill I think this has been addressed here: https://github.com/doctrine/dbal/pull/861

deeky666:
@zeroedin-bill Thanks!

PR #868, opened 2015-06-14
Make SQLite honor the "memory" option for in-memory db

mikeSimonson:
Shouldn't this be the default behavior ?

### DIFF
--- a/lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
@@ -52,9 +52,13 @@ class SqliteSchemaManager extends AbstractSchemaManager
         $params = $this->_conn->getParams();
         $driver = $params['driver'];
         $options = array(
-            'driver' => $driver,
-            'path' => $database
+            'driver' => $driver
         );
+        if(isset($params['memory']) && $params['memory']){
+            $options['memory'] = true;
+        }else{
+            $options['path'] = $database;
+        }
         $conn = \Doctrine\DBAL\DriverManager::getConnection($options);
         $conn->connect();
         $conn->close();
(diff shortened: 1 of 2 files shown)


mikeSimonson:
Especialy regarding this documentation http://doctrine-dbal.readthedocs.org/en/latest/reference/configuration.html#pdo-sqlite

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -52,9 +52,13 @@ public function createDatabase($database)
         $params = $this->_conn->getParams();
         $driver = $params['driver'];
         $options = array(
-            'driver' => $driver,
-            'path' => $database
+            'driver' => $driver
         );
+        if(isset($params['memory']) && $params['memory']){
```
deeky666: @mikeSimonson shouldn't `createDatabase()` rather be a noop in this case? It doesn't make sense to connect afterwards and immediately close again, as in-memory database only exists during the connection, right?

deeky666: Sorry got something messed up here. I think your patch is wrong anyways because you could create a persistent database from a current in-memory connection:

1. connect to in-memory DB
2. with the schema manager and that connection create another database that is persistent.

So in this case it is totally unrelated whether you are currently connected to a in-memory database. But what we could do i check whether the given `$database` parameter is `:memory:` and then perform a noop.

deeky666:
@mikeSimonson closed as `invalid`?

mikeSimonson:
@deeky666 Yes, I am pretty sure it's invalid as I am using a memory connection in my test right now and it works. But it's hard to remember exactly why I ended up wanting that change.

deeky666:
@mikeSimonson okay nvm then. Thanks for the feedback :)

PR #869, opened 2015-06-17
Make date and time types throw exception when invalid values are passed to `convertToDatabaseValue`

billschaller:


Ocramius:
Kinda disagree with this one, mainly due to how performance intensive these classes are...

zeroedin-bill:
@Ocramius as of now these conversions throw fatal errors when a non-null value that isn't a DateTime is passed. I think the tradeoff of adding a conditional identical test ANDed with instanceof is worth it in this case. A benchmark would probably show us quantitatively how it affects performance.

Ocramius:
I think we actually need that benchmark at this point...

lib/Doctrine/DBAL/Types/DateTimeType.php
```diff
@@ -49,6 +49,10 @@ public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $pla
      */
     public function convertToDatabaseValue($value, AbstractPlatform $platform)
     {
+        if ($value !== null && !$value instanceof \DateTime) {
```
deeky666: I think you can at least avoid one `$value !== null` check here.

lib/Doctrine/DBAL/Types/DateTimeTzType.php
```diff
@@ -67,6 +67,10 @@ public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $pla
      */
     public function convertToDatabaseValue($value, AbstractPlatform $platform)
     {
+        if ($value !== null && !$value instanceof \DateTime) {
```
deeky666: I think you can at least avoid one `$value !== null` check here.

lib/Doctrine/DBAL/Types/TimeType.php
```diff
@@ -49,6 +49,10 @@ public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $pla
      */
     public function convertToDatabaseValue($value, AbstractPlatform $platform)
     {
+        if ($value !== null && !$value instanceof \DateTime) {
```
deeky666: I think you can at least avoid one `$value !== null` check here.

lib/Doctrine/DBAL/Types/DateTimeType.php
Ocramius: This is actually correct: `null` is an allowed value. Maybe applying De Morgan would make it more readable tho

lib/Doctrine/DBAL/Types/DateTimeTzType.php
Ocramius: This is actually correct: `null` is an allowed value. Maybe applying De Morgan would make it more readable tho

lib/Doctrine/DBAL/Types/DateTimeType.php
deeky666:
```php
if (null === $value) {
    return null;
}

if (! $value instanceof \DateTime) {
    throw ConversionException::conversionFailedInvalidType($value, $this->getName(), "DateTime");
}

return $value->format($platform->getDateTimeFormatString());
```

lib/Doctrine/DBAL/Types/ConversionException.php
```diff
@@ -65,4 +65,29 @@ static public function conversionFailedFormat($value, $toType, $expectedFormat)
...
+    static public function conversionFailedInvalidType($value, $toType, $fromType)
+    {
+        $actualType = gettype($value);
+        if ($actualType === 'object') {
+            $actualType .= " (" . get_class($value) . ")";
+            if (!method_exists($value, '__toString')) {
+                $value = 'object';
+            }
+        }
+        $value = (string)$value;
```
Ocramius: Argh, I just realised that this logic is too convoluted :-(

@zeroedin-bill:

 - arrays will crash this
 - testing is needed
 - I simply wouldn't care about abbreviating the string
 - I wouldn't rely on `__toString`

lib/Doctrine/DBAL/Types/ConversionException.php
```diff
@@ -65,4 +65,29 @@ static public function conversionFailedFormat($value, $toType, $expectedFormat)
...
+
+    /**
+     * Thrown when the PHP value passed to the converter was not of the expected type.
+     *
+     * @param mixed $value
+     * @param string $fromType
+     *
+     * @return \Doctrine\DBAL\Types\ConversionException
+     */
+    static public function conversionFailedInvalidType($value, $toType, $fromType)
```
Ocramius: `public static` rather than `static public`

lib/Doctrine/DBAL/Types/ConversionException.php
```diff
@@ -65,4 +65,29 @@ static public function conversionFailedFormat($value, $toType, $expectedFormat)
             $toType . '. Expected format: ' . $expectedFormat
         );
     }
+
+    /**
+     * Thrown when the PHP value passed to the converter was not of the expected type.
+     *
+     * @param mixed $value
+     * @param string $fromType
```
Ocramius: `$toType` is missing

Ocramius:
Tracked via http://www.doctrine-project.org/jira/browse/DBAL-1293

PR #871, opened 2015-06-18
`SqlConsoleCommand` should show results of queries containing `RETURNING`

bountin:
PostgreSQL supports returning values after they are inserted or updated which is especially handy if one wants to get the value of an used sequence. (See http://www.postgresql.org/docs/9.4/static/sql-insert.html) Those returned values are currently omitted but could be displayed.

Currently (with symfony):

```
vagrant@packer-parallels-iso:/vagrant⟫ app/console doctrine:query:sql "INSERT INTO account (id) VALUES (uuid_generate_v4()) RETURNING id"
int 0
```

After the patch the output is the following:

```
array (size=1)
  0 =>
    array (size=1)
      'id' => string 'ad7a7c8f-72fd-48b9-b216-568ac204649d' (length=36)
```

Looking directly for 'returning' is a bit direct in my eyes but as there is no sql parser present, there is no other easy way to do so. If someone has a better implementation or any suggestion, feel free to comment :)

### DIFF
--- a/lib/Doctrine/DBAL/Tools/Console/Command/RunSqlCommand.php
+++ b/lib/Doctrine/DBAL/Tools/Console/Command/RunSqlCommand.php
@@ -48,7 +48,8 @@ class RunSqlCommand extends Command
         ->setDescription('Executes arbitrary SQL directly from the command line.')
         ->setDefinition(array(
             new InputArgument('sql', InputArgument::REQUIRED, 'The SQL statement to execute.'),
-            new InputOption('depth', null, InputOption::VALUE_REQUIRED, 'Dumping depth of result set.', 7)
+            new InputOption('depth', null, InputOption::VALUE_REQUIRED, 'Dumping depth of result set.', 7),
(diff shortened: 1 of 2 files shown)


bountin:
Also detecting a SELECT statement by looking at the first SELECT does not always work. E.g. if one is using CTE: 
```
vagrant@packer-parallels-iso:/vagrant⟫ app/console doctrine:query:sql "WITH bar as (SELECT 1) SELECT * FROM bar"
int 0
```

zeroedin-bill:
Rather than this nonsense of detecting strings inside queries, I'd rather just create an option that specifies this behavior. 

Maybe --force-fetch?

bountin:
I've rewritten the code to your suggestion!

bountin:
And the tests are only failing because of composer timeouts :(

deeky666:
Merged manually, renamed `fetch-result` to `force-fetch` as suggested by @zeroedin-bill (to be more explicit about intention). Thanks @bountin

zeroedin-bill:
:+1:

PR #872, opened 2015-06-19
return result from Doctrine\DBAL\Connection::transactional()

mathroc:
otherwise we have to import a variable from the parent scope into the closure scope like that

``` php
$conn->transactional(function(Connection $conn) use ($res) {
    $res = ...
});
// use $res here
```

instead with the PR it can be written this way:

``` php
$res = $conn->transactional(function(Connection $conn) {
    return ...;
});
// use $res here
```

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1093,7 +1093,7 @@ class Connection implements DriverConnection
      *
      * @param \Closure $func The function to execute transactionally.
      *
-     * @return void
+     * @return mixed The value returned by $func
      *
      * @throws \Exception
      */
@@ -1101,8 +1101,9 @@ class Connection implements DriverConnection
     {
         $this->beginTransaction();
         try {
-            $func($this);
+            $res = $func($this);
             $this->commit();
+            return $res;
         } catch (Exception $e) {
             $this->rollBack();
             throw $e;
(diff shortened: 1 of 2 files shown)


stof:
this looks wrong to me returning the return value of the callback would make more sense than returning the return value of ``$this->commit()``

Ocramius:
Should indeed return the value of the callback instead.

zeroedin-bill:
@mathroc Please also add a functional test for this to Doctrine\Tests\DBAL\Functional\ConnectionTest.php.

mathroc:
@zeroedin-bill I'll do that yes
@stof that what I intended to do... I guess I should have wrote a test in the first place that's :/

mathroc:
I've added a test and fixed the code, let me know if it's seems ok

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1093,7 +1093,7 @@ public function lastInsertId($seqName = null)
      *
      * @param \Closure $func The function to execute transactionally.
      *
-     * @return void
+     * @return mixed
```
deeky666: Maybe a little description about what is returned here might make sense so that the API user knows it returns the result from the callback

mathroc: absolutely

roddolf:
Have this been merged?
When I download the code with Composer, this feature isn't in it.

stof:
it is merged in the master branch, which is the dev version for Doctrine DBAL 2.6

PR #873, opened 2015-06-21
use travis_retry to avoid network timeout with composer

mathroc:
let's see if it actually works

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,10 @@ env:
   - DB=mysqli
 
 before_install:
-  - composer self-update
+  - travis_retry composer self-update
 
 install:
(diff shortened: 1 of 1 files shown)


mathroc:
it does fixes this problem however : https://travis-ci.org/doctrine/dbal/jobs/67740438. [travis_wait](http://docs.travis-ci.com/user/build-timeouts/) could be used to extends the timeout to 10 minutes but I'm not sure that it would solve the problem. I'd like it better if composer self-update aborted on its own sooner

zeroedin-bill:
I think the workaround in this PR will fix the issue - https://github.com/doctrine/dbal/pull/879

PR #874, opened 2015-06-21
convertToDatabaseValueSQL with $columnName

mihai-stancu:
## Goal:

I want to be able to atomically increment a property such that `$stock->setQuantityDelta(2);` will render into an SQL such as `UPDATE stock SET quantity = quantity+? WHERE id = ?;`.

I would like to accomplish this without using `DQL` every time it is necessary hence I implemented a custom Doctrine2 type which can accomplish this -- with support from this PR.
## Changes:

Added a new `$columnName` parameter to `\Doctrine\DBAL\Types\Type::convertToDatabaseValueSQL` which would be passed by `\Doctrine\ORM\Persisters\BasicEntityPersister::updateTable` ([PR](https://github.com/doctrine/doctrine2/pull/1434)) and used by the concrete type instance (ex.: **@mihai-stancu/doctrine-types-extra**:`\MS\Doctrine\DBAL\Types\DeltaType`).

### DIFF
--- a/lib/Doctrine/DBAL/Types/Type.php
+++ b/lib/Doctrine/DBAL/Types/Type.php
@@ -291,10 +291,11 @@ abstract class Type
      *
      * @param string                                    $sqlExpr
      * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform
+     * @param string                                    $columnName
      *
      * @return string
      */
-    public function convertToDatabaseValueSQL($sqlExpr, AbstractPlatform $platform)
+    public function convertToDatabaseValueSQL($sqlExpr, AbstractPlatform $platform, $columnName = null)
     {
         return $sqlExpr;
     }


Majkl578:
This change is a nasty BC break for existing custom types [1]. It can't be easily workarounded in user-land code (assuming the code wants to keep BC with previous Doctrine versions - e.g. any library using Doctrine).

[1] http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html#custom-mapping-types

mihai-stancu:
I thout it would be...

OTOH we could create a separate interface for types such as:

```
Interface ColumnNameAwareType
{
    Public function convertToDatabaseValueSQL($sqlExpr, $platform, $columnName);
}
```

If the type is column aware we feed it the column in the persister otherwise we don't.

I think that should cover BC.

mihai-stancu:
Another (better) way to implement the increment/decrement functionality would be to extend the ORM mapping of integers to allow you to specify a custom property name for the delta property and use the specified delta property (if present and non-null) to issue the appropriate `UPDATE stock SET quantity = quantity + :delta;`.

Example:
```
/**
 * @Column(type="integer", delta="quantityDelta")
 */
protected $quantity;
protected $quantityDelta;
```

Which would imply any & all integers (and possibly other numerics) can benefit from delta properties.

Ocramius:
I don't understand what kind of problem is being solved.

You are trying to make delta types, but this is not the responsibility of the ORM nor of the DBAL types. DBAL types are for storing/loading a value in a idempotent way, and this breaks this assumption.

mihai-stancu:
Agreed. If idempotency is a goal of the DBAL types then this approach is bad.

How about the alternative proposal?

mihai-stancu:
@Ocramius if you have no objections I think I'll close this PR and experiment with the other idea (custom property with delta behavior), see where that leads to.

PR #875, opened 2015-06-28
Generating a ColumnDiff led to unquoted Columns for Postgres

to-kn:
Postgres needs (or understands) quoted column-names all the Time, especially if one is using mix cased column names.
While generating Diffs, the old column name got no quotation.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/ColumnDiff.php
+++ b/lib/Doctrine/DBAL/Schema/ColumnDiff.php
@@ -77,6 +77,10 @@ class ColumnDiff
      */
     public function getOldColumnName()
     {
-        return new Identifier($this->oldColumnName);
+        if ($this->column->isQuoted()) {
+            return new Identifier('"' . $this->oldColumnName . '"');
+        } else {
+            return new Identifier($this->oldColumnName);
+        }
     }
 }
--- a/lib/Doctrine/DBAL/Schema/ColumnDiff.php
+++ b/lib/Doctrine/DBAL/Schema/ColumnDiff.php
@@ -79,8 +79,7 @@ class ColumnDiff
     {
         if ($this->column->isQuoted()) {
             return new Identifier('"' . $this->oldColumnName . '"');
-        } else {
-            return new Identifier($this->oldColumnName);
         }
+        return new Identifier($this->oldColumnName);
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Schema/ColumnDiffTest.php
+++ b/tests/Doctrine/Tests/DBAL/Schema/ColumnDiffTest.php
@@ -9,31 +9,54 @@ use Doctrine\DBAL\Types\Type;
 
 class ColumnDiffTest extends \PHPUnit_Framework_TestCase
 {
-    public function testColumnDiffQuotationForOldAndNewName()
+    public function testColumnDiffQuotationForOldAndNewNameSqlite()
     {
-        $comparator = new Comparator();
+        list($columnA, $columnB, $diff) = $this->getColumnDiff();
 
-        $string = Type::getType('string');
-        $columnA = new Column("`bar`", $string, array());
-        $columnB = new Column("`foo`", $string, array());
-
-        $diff = new ColumnDiff($columnA->getName(), $columnB, $comparator->diffColumn($columnA, $columnB));
-
-
-        $mysqlPlatform = new \Doctrine\DBAL\Platforms\MySqlPlatform();
         $sqlitePlatform = new \Doctrine\DBAL\Platforms\SqlitePlatform();
 
         $this->assertEquals($columnA->getName(), $diff->getOldColumnName()->getName());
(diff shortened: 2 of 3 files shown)


zeroedin-bill:
@tk-s Can you add a test for this please?

tk-s:
Shure, i will look at it these days!

Am 15.07.2015 um 16:29 schrieb Bill Schaller <notifications@github.com<mailto:notifications@github.com>>:


@tk-s<https://github.com/tk-s> Can you add a test for this please?

tk-s:
Sorry for the delay - can you give me a hint, where i should put the test?

Thanks a lot in advance!

zeroedin-bill:
@tk-s Make a ColumnDiffTest in tests/Doctrine/Tests/DBAL/Schema? Since there's no existing ColumnDiffTest (there should be...) It will be a lonely test, but we're trying to at least cover logic changes moving forward. Schema diffs in general need more tests in DBAL.

tk-s:
finally the unit-tests are in place. i used the column-test for quotation as a boilerplate

tests/Doctrine/Tests/DBAL/Schema/ColumnDiffTest.php
```diff
@@ -0,0 +1,47 @@
...
+
+        $sqlServerPlatform = new \Doctrine\DBAL\Platforms\SQLServerPlatform();
+
+        $this->assertEquals($columnA->getName(), $diff->getOldColumnName()->getName());
+        $this->assertEquals($columnA->getQuotedName($sqlServerPlatform), $diff->getOldColumnName()->getQuotedName($sqlServerPlatform));
+
+        $this->assertEquals($columnB->getName(), $diff->column->getName());
+        $this->assertEquals($columnB->getQuotedName($sqlServerPlatform), $diff->column->getQuotedName($sqlServerPlatform));
+    }
+
```
TomasVotruba: This line is redundant.

tests/Doctrine/Tests/DBAL/Schema/ColumnDiffTest.php
```diff
@@ -0,0 +1,47 @@
...
+        $comparator = new Comparator();
+
+        $string = Type::getType('string');
+        $columnA = new Column("`bar`", $string, array());
+        $columnB = new Column("`foo`", $string, array());
+
+        $diff = new ColumnDiff($columnA->getName(), $columnB, $comparator->diffColumn($columnA, $columnB));
+
+
+        $mysqlPlatform = new \Doctrine\DBAL\Platforms\MySqlPlatform();
```
Ocramius: Split the tests for the different platforms into separate test methods ;-)

lib/Doctrine/DBAL/Schema/ColumnDiff.php
```diff
@@ -77,6 +77,10 @@ public function hasChanged($propertyName)
      */
     public function getOldColumnName()
     {
-        return new Identifier($this->oldColumnName);
+        if ($this->column->isQuoted()) {
+            return new Identifier('"' . $this->oldColumnName . '"');
+        } else {
```
Ocramius: Since you are returning early, the `else` keyword is not needed

tk-s:
better like that?

lib/Doctrine/DBAL/Schema/ColumnDiff.php
```diff
@@ -77,6 +77,9 @@ public function hasChanged($propertyName)
      */
     public function getOldColumnName()
     {
+        if ($this->column->isQuoted()) {
```
zeroedin-bill: This logic doesn't quite make sense. You're saying that if the 'to' column name is quoted, the 'from' column name should be as well. 

The issue is really that in Comparator, the ColumnDiff is created like this:
```php
new ColumnDiff($column->getName(), $table2->getColumn($columnName), $changedProperties)
```

Which results in the 'from' column name always being unquoted, because even though the from column may originally have been properly marked as quoted, that information is not being passed into the ColumnDiff constructor. I see the same sort of issue in a few other places. It needs a more comprehensive fix than this, as this simply hides the issue.

tests/Doctrine/Tests/DBAL/Schema/ColumnDiffTest.php
TomasVotruba: Ahh, sorry, I thought it's under the last method.

tk-s:
hi @zeroedin-bill that is somehow true, but i can't think of a situation where the from column shouldn't be quoted if the to column is, since its in any case the same platform?!

Since there is no $platform-Object in the Comparator something like:

    new ColumnDiff($column->getQuotedName(), $table2->getColumn($columnName), $changedProperties)

won't work. What do you suggest?

martinlindhe:
Any update on this? Hit the issue in Laravel, which is using doctrine/dbal and tracked it down to this PR. Would happily help out if something is needed.

pavanpatharde:
I too face the same issue as @martinlindhe . Can you please provide us the update ?

deeky666:
Closing in favour of #2288 which contains a correct solution.

PR #877, opened 2015-07-13
Fixed simple_array type if array is empty

bigfoot90:
I'm using `"doctrine/orm": v2.5.0` and have `simple_array` type.
If I persist the class where the property with type `simple_array` is empty `array()` the query fails.

```
[Doctrine\DBAL\Exception\NotNullConstraintViolationException]                                                                                
  An exception occurred while executing 'INSERT INTO option (value, indicators) VALUES (?, ?)' with params ["Y", null]:  
  SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'question_id' cannot be null
```

``` php
class Option
{
    /**
     * @Column(type="simple_array")
     */
    public $indicators = array();
}
```

Found the problem in `/lib/Doctrine/DBAL/Types/JsonArrayType.php` line 38
https://github.com/doctrine/dbal/pull/143/files#diff-505f8625ef2628cdf37564b6bb499bdcR40

This PR allow pesisting empty arrays.

### DIFF
--- a/lib/Doctrine/DBAL/Types/SimpleArrayType.php
+++ b/lib/Doctrine/DBAL/Types/SimpleArrayType.php
@@ -44,9 +44,8 @@ class SimpleArrayType extends Type
      */
     public function convertToDatabaseValue($value, AbstractPlatform $platform)
     {
-        if (!$value) {
-            return null;
-        }
+        if ($value === null) return null;
+        if (!$value) return '';
 
         return implode(',', $value);
     }
@@ -56,9 +55,8 @@ class SimpleArrayType extends Type
      */
     public function convertToPHPValue($value, AbstractPlatform $platform)
     {
-        if ($value === null) {
-            return array();
-        }
+        if ($value === null) return null;
+        if (!$value) return array();
 
         $value = (is_resource($value)) ? stream_get_contents($value) : $value;
 


lib/Doctrine/DBAL/Types/SimpleArrayType.php
```diff
@@ -44,9 +44,8 @@ public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $pla
      */
     public function convertToDatabaseValue($value, AbstractPlatform $platform)
     {
-        if (!$value) {
-            return null;
-        }
+        if ($value === null) return null;
+        if (!$value) return '';
```
zeroedin-bill: if ```$value``` is not null, but is an empty array, ```return implode(',', $value)``` will return ```''```

bigfoot90: I think thank this is more performant than implode

deeky666: @zeroedin-bill no because `if (!$value)` check will also be true for an empty array and thus returning `null` (currently).
But unfortunately I think this is a BC break because users are relying on the current behaviour I suppose. We had similar discussion with `JsonArray` and could agree on a change because of keeping BC (I know this sounds weird but IMO we cannot change the assumption that an empty array will evaluate to `null`). Existing code with checks against `null` will break with this PR....

zeroedin-bill: @deeky666 yup...

Ocramius:
BC Break, closing. Sorry!

bigfoot90:
The field is not mapped as nullable, so it should never be ```null```.
If you cannot merge this quickly, you should at least add an warning in the docs saying that the field must be mapped as nullable.

abePdIta:
I think it would be good to at least mention it in the documentation.

PR #878, opened 2015-07-14
Fix call on non-object in ping() with PDO wrapper

Virakal:
Fixes an issue when calling `ping()` where `$this->platform` is null. Happened to me because using the `'pdo'` param bypasses the `connect()` method and the platform is never set without explicitly calling the method.

Not sure if this is the most elegant way to deal with this but it worked enough for my purposes so I figured I should submit a PR.


zeroedin-bill:
There are a few other spots that should probably be calling getDatabasePlatform as well. This is a simple and straightforward fix, so I'll merge.

PR #879, opened 2015-07-15
workaround for travis composer self-update ipv6 timeout issue

billschaller:
Testing to see if making ipv6 have lower precedence than ipv4 on travis boxes fixes the composer timeout issue.

Workaround from @Seldaek in this issue thread: https://github.com/composer/composer/issues/4142

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ env:
   - DB=mysqli
 
 before_install:
+  - sudo sh -c "echo 'precedence ::ffff:0:0/96 100' >> /etc/gai.conf"
   - composer self-update
 
 install:


.travis.yml
```diff
@@ -18,6 +18,7 @@ env:
   - DB=mysqli
 
 before_install:
+  - sudo sh -c "echo 'precedence ::ffff:0:0/96 100' >> /etc/gai.conf"
```
deeky666: So another `sudo` based workaround just to come around travis issues? =( I was hoping that we could get rid of `sudo` commands totally sometime in the future to make use of docker-based builds and caching...

zeroedin-bill: I'm not sure how else to prevent composer from using ipv6 to update itself... We could maybe use wget to download composer.phar, but that's even uglier I think... Docker builds would be quite nice.

Seldaek: What needs to happen is to make whichever network is losing v6 packets in
the middle fix their shit. Not disabling things..

zeroedin-bill: Other projects have been having the same issue. Anyone know how to contact
Travis' network team?

zeroedin-bill:
This seems to be not an issue anymore - closing PR.

PR #880, opened 2015-07-15
Drop PHP 5.3 from Travis-CI build matrix

billschaller:
Drops php 5.3 from travis build matrix and updates composer.json to require php >= 5.4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 5.3
   - 5.4
   - 5.5
   - 5.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,10 +29,6 @@ script: ./vendor/bin/phpunit --configuration tests/travis/$DB.travis.xml
 
 matrix:
   include:
-    - php: 5.3
-      env: DB=mariadb
-      addons:
-        mariadb: 5.5
     - php: 5.4
       env: DB=mariadb
       addons:
@@ -54,10 +50,6 @@ matrix:
       addons:
         mariadb: 5.5
 
-    - php: 5.3
-      env: DB=mariadb
-      addons:
-        mariadb: 10.0
     - php: 5.4
       env: DB=mariadb
       addons:
@@ -79,10 +71,6 @@ matrix:
       addons:
         mariadb: 10.0
 
-    - php: 5.3
-      env: DB=mariadb
-      addons:
-        mariadb: 10.1
     - php: 5.4
       env: DB=mariadb
       addons:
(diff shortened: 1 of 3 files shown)


deeky666:
:+1: as long as you cleaned all 5.3 stuff I'm okay with it. Just for a reference:

https://github.com/doctrine/doctrine2/pull/1255

this PR also optimized some things like `$this` in closures and stuff.

zeroedin-bill:
@deeky666 as far as I can tell there's no similar changes needed to dbal - no issues with $this in closures, no conditional blocks checking PHP_VERSION_ID other than 50600, etc.

Tobion:
@zeroedin-bill there is a reference to PHP 5.3 in https://github.com/doctrine/dbal/blob/master/docs/en/reference/security.rst#non-ascii-compatible-charsets-in-mysql
I guess this paragraph can be removed. Otherwise this PR looks good. :+1:

zeroedin-bill:
@Tobion This means that people who expect to use non-ascii charsets with MySQL MUST use the charset parameter, right? So essentially the paragraph should remain, but the bit about PHP < 5.3.6 should be removed?

Tobion:
No, from what I understand this is only relevant for PHP <= 5.3.6. Using a newer version should fix the charset problem.

zeroedin-bill:
@Tobion 

> Up until PHP 5.3.6 PDO has a security problem when using non ascii compatible charsets. Even if specifying the charset using "SET NAMES", emulated prepared statements and PDO#quote could not reliably escape values, opening up to potential SQL injections. **If you are running PHP 5.3.6 you can solve this issue by passing the driver option "charset" to Doctrine PDO MySQL driver. Using SET NAMES does not suffice!**

To me this seems to imply that if you are running PHP > 5.3.6, and you do not set the charset, you are still open to injection. I don't use MySQL regularly, so I'm not up on the standard practice here... If I don't set the charset, but I am using emulated prepared statements, is it possible to inject using UTF-8 characters in parameter values?

Tobion:
What this describes is the workaround for `Up until PHP 5.3.6 PDO has a security problem when using non ascii compatible charsets.`. So you don't need it if you use a newer php version.

zeroedin-bill:
@Tobion OK, thanks.

deeky666:
@Ocramius ready to merge from my POV. Anything missing here? Otherwise you can mergie :)

Ocramius:
@deeky666 done :)

Ocramius:
Note: created http://www.doctrine-project.org/jira/browse/DBAL-1292 to track this.

deeky666:
@Ocramius good idea. thx!

PR #883, opened 2015-07-16
Fixed problem when having subqueries in where clause.

mreuter:
Hello,

I had some problems when using a query like

``` sql
SELECT a, b, c FROM t1 WHERE (SELECT COUNT(*) FROM t2 WHERE t1.a = t2.a) > 0
```

because the FROM in the WHERE-clause got matched as the "main" FROM. Changing part of the regex to non-greedy seems to fix the problem (without affecting SELECT ... FROM subqueries in the SELECT-part.


zeroedin-bill:
This needs a test case.

mreuter:
Tests added.

mreuter:
I'm sorry for the broken commits. Somehow I managed to include errors after success running the test locally.

I had to get rid of one of the regex, because I tend to have very large queries and had to set pcre.backtrack_limit and pcre.recursion_limit to very high values (about 100 times bigger than the defaults).

The "bracket counting" solution without regex might not be totally optimized, but it passes all tests and in case of my large queries is more than 100 times faster than the regex (regex took 0.4s, bracket counting about 0.003s).

Ocramius:
@mreuter checkout a new branch and cherry-pick only the relevant commits from the list above, then send a new PR :-)

zeroedin-bill:
@Ocramius @deeky666 I already have a PR open that I think makes this irrelevant... https://github.com/doctrine/dbal/pull/818

deeky666:
@zeroedin-bill can you please check if this might already be solved by your recent changes?

zeroedin-bill:
@deeky666 Yeah, I think this is covered now but I'll merge the patches.

PR #884, opened 2015-07-16
Travis-CI - Switch to container-based infrastructure

TomasVotruba:
Ref: https://github.com/doctrine/doctrine2/pull/1466

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,3 +44,5 @@ matrix:
 
 addons:
   postgresql: '9.4'
+
+sudo: false


Ocramius:
:+1:

TomasVotruba:
:+1:

.travis.yml
```diff
@@ -44,3 +44,5 @@ matrix:
 
 addons:
   postgresql: '9.4'
+
+sudo: false
```
stof: not suited as the script currently uses sudo

stof: hmm, actually, it never uses it because the code switching the Postgres version based on the ENV variable is broken and never switches it (meaning we don't actually test against multiple versions)

TomasVotruba:
@stof What/where to fix then?

.travis.yml
TomasVotruba: @stof What/where to fix then?

PR #885, opened 2015-07-20
parameter $database for listTableIndexes, listTableDetails

pmaxs:


Ocramius:
@podbeltsev this requires a test case

podbeltsev:
I've made tests, but they didn't pass
For testing I need tables in different databases, so I use main (db_***) and tmp (tmpdb_***). There is no tmpdb_name for  tmp connection in config, so I get error SQLSTATE[3D000]: Invalid catalog name: 1046 No database selected
I just need parameter $database for listTableIndexes, listTableDetails to request tables info from databases that are not assigned to current connection. Functions listTableColumns and listTableForeignKeys allready supports this parameter

PR #886, opened 2015-07-23
Add test for MariaDB 5.5, 10.0 and 10.1 on Travis

PowerKiKi:
This use the brand new supported addon `mariadb` (not yet officially announced). This is unfortunately a bit verbose, but I don't think there is any alternative because we cannot install the addon when testing against mysql otherwise it would overwrite mysql install.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,81 @@ before_script:
 script: ./vendor/bin/phpunit --configuration tests/travis/$DB.travis.xml
 
 matrix:
+  include:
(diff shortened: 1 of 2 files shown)


deeky666:
@PowerKiKi thanks that truly makes sense.

PowerKiKi:
Thanks for merging @deeky666, you might also be interested in a similar PR for doctrine2: https://github.com/doctrine/doctrine2/pull/1472

PR #887, opened 2015-07-23
use REGEXP operator for sqlite

janschoenherr:
I believe this should be changed, as REGEXP can actually be used as an operator in SQLite. 

https://www.sqlite.org/lang_expr.html#regexp


deeky666:
@janschoenherr the documentation states that using `REGEXP` expression will result in an error message as long as there is no user function defined for `regexp()` (which it isn't by default). But it seems `RLIKE` expression isn't supported at all so replacing it by `REGEXP` might at least be the better option (for those that define a custom `regexp()`) function.

deeky666:
@janschoenherr thanks for pointing this out!

PR #888, opened 2015-07-29
Allow testing against `doctrine/common` `2.6`

nicolas-grekas:
By forbidding 2.6-dev, dbal 2.5 is untestable, because when composer tries to resolve deps to their highest, it has two choices:
- doctrine/common@dev-master + dbal@2.4
- or doctrine/common@2.5 + dbal@dev-master

and it chooses the first option.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     ],
     "require": {
         "php": ">=5.3.2",
-        "doctrine/common": ">=2.4,<2.6-dev"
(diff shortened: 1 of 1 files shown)


deeky666:
@nicolas-grekas thanks!

deeky666:
@nicolas-grekas I suppose this needs to be backported to 2.5???

nicolas-grekas:
Not sure it's required but why not yes

PR #889, opened 2015-07-30
#1234 On OracleDB, a "Table has no primary key" error is raised when the index and the constraint name don't match

nitso:
When reverse-engineering oracle database, "Table has no primary key" error is triggered. That is caused by a mistake in index selection from database.


deeky666:
@nitso this needs a failing test case (functional).

deeky666:
@nitso ping. We need at least information about how to reproduce the error. @DeepDiver1975 maybe you have a clue?

athos99 (approved):

athos99:
For reproduce this bug, in the constraint dfintion, the CONSTRAINT_NAME must be different from the INDEX_NAME.

Ocramius:
Handled in #2686

PR #892, opened 2015-08-04
DDC-3863 add a `json` type that doesn't have the flaws of `json_array`

Taluu:
As discussed in #891, the `JsonArrayType` was deprecated in favor of a new `JsonType`. :}

### DIFF
--- a/lib/Doctrine/DBAL/Types/ConversionException.php
+++ b/lib/Doctrine/DBAL/Types/ConversionException.php
@@ -99,4 +99,16 @@ class ConversionException extends \Doctrine\DBAL\DBALException
             implode(', ', $possibleTypes)
         ));
     }
+
+    static public function conversionFailedSerialization($value, $format, $error, \Exception $previous = null)
+    {
+        $actualType = is_object($value) ? get_class($value) : gettype($value);
+
+        return new self(sprintf(
+            "Could not convert PHP type '%s' to '%s', as an '%s' error was triggered by the serialization",
+            $actualType,
+            $format,
+            $error
+        ));
+    }
 }
--- a/lib/Doctrine/DBAL/Types/JsonArrayType.php
+++ b/lib/Doctrine/DBAL/Types/JsonArrayType.php
@@ -25,30 +25,11 @@ use Doctrine\DBAL\Platforms\AbstractPlatform;
  * Array Type which can be used to generate json arrays.
  *
  * @since  2.3
+ * @deprecated Use JsonType instead
  * @author Johannes M. Schmitt <schmittjoh@gmail.com>
  */
-class JsonArrayType extends Type
+class JsonArrayType extends JsonType
 {
-    /**
-     * {@inheritdoc}
-     */
-    public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform)
-    {
-        return $platform->getJsonTypeDeclarationSQL($fieldDeclaration);
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function convertToDatabaseValue($value, AbstractPlatform $platform)
-    {
-        if (null === $value) {
-            return null;
-        }
-
-        return json_encode($value);
-    }
-
     /**
      * {@inheritdoc}
      */
--- a/lib/Doctrine/DBAL/Types/JsonType.php
+++ b/lib/Doctrine/DBAL/Types/JsonType.php
@@ -46,7 +46,13 @@ class JsonType extends Type
             return null;
         }
 
-        return json_encode($value);
+        $encoded = json_encode($value);
+
+        if (JSON_ERROR_NONE !== json_last_error()) {
+            throw ConversionException::conversionFailedSerialization($value, 'json', $this->getLastErrorMessage());
+        }
+
+        return $encoded;
     }
 
     /**
@@ -90,5 +96,45 @@ class JsonType extends Type
          */
         //return ! $platform->hasNativeJsonType();
         return true;
+    }
+
+    /**
+     * Get the latest json error message
+     *
+     * This method declaration has been extracted from symfony's php 5.5 polyfill
+     *
+     * @link https://github.com/symfony/polyfill-php55/blob/master/Php55.php
+     * @link http://nl1.php.net/manual/en/function.json-last-error-msg.php
+     *
+     * @return string
+     */
+    private function getLastErrorMessage()
+    {
+        if (function_exists('json_last_error_msg')) {
+            return json_last_error_msg();
+        }
+
+        switch (json_last_error()) {
+            case JSON_ERROR_NONE:
+                return 'No error';
+
+            case JSON_ERROR_DEPTH:
+                return 'Maximum stack depth exceeded';
+
+            case JSON_ERROR_STATE_MISMATCH:
+                return 'State mismatch (invalid or malformed JSON)';
+
+            case JSON_ERROR_CTRL_CHAR:
+                return 'Control character error, possibly incorrectly encoded';
+
+            case JSON_ERROR_SYNTAX:
+                return 'Syntax error';
+
+            case JSON_ERROR_UTF8:
+                return 'Malformed UTF-8 characters, possibly incorrectly encoded';
+
+            default:
+                return 'Unknown error';
+        }
     }
 }
--- a/lib/Doctrine/DBAL/Types/JsonType.php
+++ b/lib/Doctrine/DBAL/Types/JsonType.php
@@ -1,0 +1,94 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Types;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+
+/**
+ * Type generating json objects values
+ *
+ * @since  2.6
+ * @author Baptiste Clavié <clavie.b@gmail.com>
+ */
+class JsonType extends Type
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform)
+    {
+        return $platform->getJsonTypeDeclarationSQL($fieldDeclaration);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToDatabaseValue($value, AbstractPlatform $platform)
+    {
+        if (null === $value) {
+            return null;
+        }
+
+        return json_encode($value);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToPHPValue($value, AbstractPlatform $platform)
+    {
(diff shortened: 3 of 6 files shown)


lib/Doctrine/DBAL/Types/JsonArrayType.php
```diff
@@ -70,12 +51,5 @@ public function getName()
...
-
-    /**
-     * {@inheritdoc}
-     */
-    public function requiresSQLCommentHint(AbstractPlatform $platform)
-    {
-        return ! $platform->hasNativeJsonType();
-    }
 }
+
```
stof: the extra empty line should be removed

lib/Doctrine/DBAL/Types/JsonArrayType.php
```diff
@@ -70,12 +51,5 @@ public function getName()
     {
         return Type::JSON_ARRAY;
     }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function requiresSQLCommentHint(AbstractPlatform $platform)
```
stof: the issue is that both the json_array and the json type are now mapped to the native JSON type, meaning it can break schema changes for platforms using the native type. You need to distinguish them to avoid endless schema changes for them

lib/Doctrine/DBAL/Types/Type.php
```diff
@@ -36,6 +36,7 @@
     const TARRAY = 'array';
     const SIMPLE_ARRAY = 'simple_array';
     const JSON_ARRAY = 'json_array';
+    const JSON_UNFLAWED = 'json';
```
stof: this should just be ``JSON``, not ``JSON_UNFLAWED``. Type name constant are matching the type names (unless it is not possible because of reserved keywords like ``array``)

tests/Doctrine/Tests/DBAL/Types/JsonTest.php
```diff
@@ -0,0 +1,89 @@
...
+        $databaseValue = json_encode($value);
+        $phpValue      = $this->type->convertToPHPValue($databaseValue, $this->platform);
+
+        $this->assertEquals($value, $phpValue);
+    }
+
+    /** @dataProvider providerFailure */
+    public function testConversionFailure($data)
+    {
+        error_reporting( (E_ALL | E_STRICT) - \E_NOTICE );
```
stof: this looks wrong to me. Conversion failures should not trigger a notice before the real exception.

Taluu: I copied that from the `ArrayType`, but I can remove it, as it is true that `json_decode` does not trigger such errors

lib/Doctrine/DBAL/Types/Type.php
Taluu: This was more a easter egg more than anything else, but `JSON` works too. :}

lib/Doctrine/DBAL/Types/JsonArrayType.php
Taluu: But these are both json structures, so basically what should I use instead for this deprecation ?

lib/Doctrine/DBAL/Types/Type.php
stof: the issue is that this is the public API to access type names

lib/Doctrine/DBAL/Types/JsonArrayType.php
stof: I'm not sure about the right one to keep uncommented. It should probably be the old one for BC reasons.

But the native type can only be mapped to a single doctrine type, and all other doctrine types reusing the same native type must use a comment hint to make the schema retrieval work properly.
Otherwise the native field will always be considered as a ``json_array`` and any of your ``json`` field will trigger an ``ALTER TABLE`` query in each schema update run or migration generation. this is precisely why we introduced comment hints

Taluu: So basically, until 3.0, we should keep the new one always commented, and then switch it out when the old one will be dropped ?

tests/Doctrine/Tests/DBAL/Types/JsonTest.php
stof: Well, I would remove it then, to be sure we don't get a notice by mistake and uncaught by the testsuite

Taluu: Yup, I removed it

lib/Doctrine/DBAL/Types/Type.php
Taluu: Renamed to `JSON`. The commit will keep the easter egg name then :}

lib/Doctrine/DBAL/Types/JsonArrayType.php
Taluu: fixed

lib/Doctrine/DBAL/Types/JsonType.php
```diff
@@ -0,0 +1,94 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToDatabaseValue($value, AbstractPlatform $platform)
+    {
+        if (null === $value) {
+            return null;
+        }
+
+        return json_encode($value);
```
deeky666: Shouldn't we throw an exception here, too if encoding fails?

zeroedin-bill: Should throw some variant of ConversionException with the return value of json_last_error().

Taluu: Should I do it too for the `convertToPHPValue` ?

lib/Doctrine/DBAL/Types/JsonType.php
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToDatabaseValue($value, AbstractPlatform $platform)
+    {
+        if (null === $value) {
+            return null;
+        }
```
stof: should this actually be the case ? shouldn't it be converted to a json null instead ?

Taluu: Not quite sure if it should be encoded too, if the field can be nullable ? Otherwise, it gives a `"null"` string when encoded in json...

zeroedin-bill: @stof @Taluu Kind of debatable - 
```
php > var_dump(json_decode('null') === json_decode(null));
bool(true)
```

If the platform has native json support (like postgresql) it will be stored as null. If the platform doesn't have native json support, then storing it as a SQL null will achieve much the same result, as deserializing the value to PHP value will be the same whether it's ```'null'``` or ```NULL```

lib/Doctrine/DBAL/Types/JsonType.php
zeroedin-bill: No - the logic there is fine - ConversionException is missing a static constructor for this PHP->database scenario - there should probably be a ConversionException::serializationFailed method or something like that.

lib/Doctrine/DBAL/Types/JsonType.php
Taluu: Maybe, but on `json_encode`, it differs ;

```
Psy Shell v0.6.1 (PHP 7.0.2 — cli) by Justin Hileman
>>> json_encode('null');
=> ""null""
>>> json_encode(null);
=> "null"
```

And then on a `json_decode` call :

```
Psy Shell v0.6.1 (PHP 7.0.2 — cli) by Justin Hileman
>>> json_decode(null);
=> null
>>> json_decode('null');
=> null
>>> json_decode('"null"');
=> "null"
```

So I think returning a null value when the value is actually null should be more appropriate (Sorry for the late answer...)

deeky666: I tend to prefer SQL `NULL` whenever possible as it seems more convenient than having a JSON `null` string stored (might also be more efficient storage wise...). Also it avoids unnecessary `json_decode()` calls for converting back to PHP which might improve performance. If nobody disagrees, I'll accept the current implementation.

lib/Doctrine/DBAL/Types/JsonType.php
deeky666: @Taluu the PR seems fine so far but still throwing a `ConversionException` if `json_encode()` fails is missing.

Taluu: Yep, I was on it, and I just updated it (and rebased it with the master branch)

Taluu:
The PR was updated with the latest comment and remarks. cc @deeky666 @zeroedin-bill @stof (sorry for the delay)

lib/Doctrine/DBAL/Types/JsonType.php
```diff
@@ -0,0 +1,120 @@
...
+         * JsonArrayType will be dropped
+         */
+        //return ! $platform->hasNativeJsonType();
+        return true;
+    }
+
+    // extracted from symfony's php 5.5 polyfill
+    // @link https://github.com/symfony/polyfill-php55/blob/master/Php55.php
+    // @link http://nl1.php.net/manual/en/function.json-last-error-msg.php
+    private static function json_last_error_msg()
```
deeky666: Why `static`? This method is not used in static context so we should make it non-static. Also please don't use underscore naming. I'd propose something like `getLastErrorMessage()`.

lib/Doctrine/DBAL/Types/JsonType.php
```diff
@@ -0,0 +1,120 @@
...
+    {
+        if (function_exists('json_last_error_msg')) {
+            return json_last_error_msg();
+        }
+
+        switch (json_last_error()) {
+            case JSON_ERROR_NONE: return 'No error';
```
deeky666: Please put a new line after each colon and indent according to PSR-2.

lib/Doctrine/DBAL/Types/JsonType.php
```diff
@@ -0,0 +1,120 @@
...
+         * should be switched back to the platform detection at 3.0, when
+         * JsonArrayType will be dropped
+         */
+        //return ! $platform->hasNativeJsonType();
+        return true;
+    }
+
+    // extracted from symfony's php 5.5 polyfill
+    // @link https://github.com/symfony/polyfill-php55/blob/master/Php55.php
+    // @link http://nl1.php.net/manual/en/function.json-last-error-msg.php
```
deeky666: Please use DocBlock here instead of line comments

lib/Doctrine/DBAL/Types/JsonType.php
Taluu: changed. I kind of copied the polyfill, and static seemed ok to be used here as it does not depend on any context. But I removed the static, as it doesn't really matter

lib/Doctrine/DBAL/Types/JsonType.php
Taluu: fixed

lib/Doctrine/DBAL/Types/JsonType.php
Taluu: fixed

lib/Doctrine/DBAL/Types/ConversionException.php
```diff
@@ -99,4 +99,16 @@ static public function conversionFailedInvalidType($value, $toType, array $possi
             implode(', ', $possibleTypes)
         ));
     }
+
+    static public function conversionFailedSerialization($value, $format, $error, \Exception $previous = null)
```
deeky666: Not sure if we really need to introduce another method here. Maybe it would suffice the purpose by sticking to `conversionFailedFormat()`

deeky666: /cc @Ocramius @zeroedin-bill

Taluu: That is also what I thought but as @zeroedin-bill suggested it, I went with it

deeky666: @Taluu yeah sorry then for the extra effort. Could you be so kind to just use `conversionFailedFormat()` then? Also you might just put the error message into `$expectedFormat`. Like:

```php
<?php

// snippet

if (JSON_ERROR_NONE !== json_last_error()) {
    throw ConversionException::conversionFailedFormat(
        $value,
        Type::JSON,
        'json (' . $this->getLastErrorMessage() . ')'
    );
}
```

deeky666: @Taluu ah no wait.

deeky666: The exception message says `Could not convert database value to Doctrine type` which would be wrong then.

Taluu: I'm leaving this as is then, unless something better can be found ?

Taluu: I'm not quite sure the message format is good too though

Taluu:
Ping. Any news around here ?

Thanks

Taluu:
Sorry to stir it up again, but any news around here ? /cc @deeky666 @zeroedin-bill @Ocramius

lib/Doctrine/DBAL/Types/JsonType.php
deeky666: What is also coming to my mind is database queries against JSON columns where you want to filter "empty" (NULL) values which would not be possible via `IS NULL` if we would put JSON NULL into DB. So returning `null` here is the better approach IMO.

deeky666:
LGTM. /cc @zeroedin-bill @Ocramius merge?

Ocramius:
LGTM. @zeroedin-bill, can you check it and press the big green scary button? :-)

zeroedin-bill:
Driving all day, sry

Ocramius:
Assigned it to you - whenever you got time :-)

Doesn't need merging back into 2.5, so it's as quick as the UI interaction,
after review.

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

guilhermeblanco:
"Even though I walk through the valley of the shadow of death, I fear no evil"

Merging... =)

PR #894, opened 2015-08-13
Fixed correct value in "Doctrine\Tests\DBAL\Schema\Visitor\SchemaSqlCollectorTest"

LeoOnTheEarth:
In the `Doctrine\DBAL\Platforms\AbstractPlatform`, those three methods' return value are `string`
- getCreateSchemaSQL
- getCreateForeignKeySQL
- getCreateSequenceSQL

The `$platformMock` should return string value, not array

I also refactored `Doctrine\DBAL\Schema\Visitor\CreateSchemaSqlCollector`
Since these three methods' return value are string, we can simply add the SQL string into an array with this style `$foo[] = $sql`


deeky666:
@LeoOnTheEarth thanks!

PR #896, opened 2015-08-18
Override methods for sharding connection

vincentchalamon:
Required by https://github.com/doctrine/DoctrineBundle/pull/455
- [x] Override `getHost`, `getUsername`, `getPassword`, `getPort` & `getParams` methods on [PoolingShardConnection](http://www.doctrine-project.org/api/dbal/2.4/class-Doctrine.DBAL.Sharding.PoolingShardConnection.html)
- [x] Add unit tests

### DIFF
--- a/lib/Doctrine/DBAL/Sharding/PoolingShardManager.php
+++ b/lib/Doctrine/DBAL/Sharding/PoolingShardManager.php
@@ -18,6 +18,7 @@
  */
 
 namespace Doctrine\DBAL\Sharding;
+use Doctrine\DBAL\Sharding\ShardChoser\ShardChoser;
 
 /**
  * Shard Manager for the Connection Pooling Shard Strategy
@@ -27,12 +28,12 @@ namespace Doctrine\DBAL\Sharding;
 class PoolingShardManager implements ShardManager
 {
     /**
-     * @var \Doctrine\DBAL\Sharding\PoolingShardConnection
+     * @var PoolingShardConnection
      */
     private $conn;
 
     /**
-     * @var \Doctrine\DBAL\Sharding\ShardChoser\ShardChoser
+     * @var ShardChoser
      */
     private $choser;
 
@@ -42,7 +43,7 @@ class PoolingShardManager implements ShardManager
     private $currentDistributionValue;
 
     /**
-     * @param \Doctrine\DBAL\Sharding\PoolingShardConnection $conn
+     * @param PoolingShardConnection $conn
      */
     public function __construct(PoolingShardConnection $conn)
     {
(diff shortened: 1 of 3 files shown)


lib/Doctrine/DBAL/Connection.php
```diff
@@ -96,6 +96,13 @@ class Connection implements DriverConnection
     protected $_conn;
 
     /**
+     * The connection name.
+     *
+     * @var string
+     */
+    protected $_name;
```
stof: why adding a name in the connection ? This does not make sense IMO. there is nothing requiring a connection name in Doctrine

lib/Doctrine/DBAL/Sharding/PoolingShardManager.php
```diff
@@ -52,6 +53,14 @@ public function __construct(PoolingShardConnection $conn)
     }
 
     /**
+     * @return PoolingShardConnection
+     */
+    public function getConnection()
```
stof: why do you need to expose the connection here ?

lib/Doctrine/DBAL/Connection.php
vincentchalamon: I need to get the connection name to retrieve the related shard manager in https://github.com/doctrine/DoctrineMigrationsBundle/pull/132

stof: This does not make any sense. Naming connections is only needed when you have multiple connections, but the connection has strictly no reason to be aware of it.
The implementation looking for the shard manager should be changed in the bundle to avoid this meaningless change in DBAL

vincentchalamon:
On `pgsql`, I got following error:
>1) Doctrine\Tests\DBAL\Functional\Schema\PostgreSqlSchemaManagerTest::testListDatabases
Exception: [Doctrine\DBAL\DBALException] An exception occurred while executing 'CREATE DATABASE test_create_database':

>SQLSTATE[55006]: Object in use: 7 ERROR:  source database "template1" is being accessed by other users
DETAIL:  There are 1 other session(s) using the database.

>With queries:
3. SQL: 'CREATE DATABASE test_create_database' Params: 
2. SQL: 'DROP DATABASE test_create_database' Params: 

Any idea ?

MLKiiwy:
Is the test working without the change ?

vincentchalamon:
@MLKiiwy Nop: https://travis-ci.org/doctrine/dbal/builds/60729009

deeky666:
@vincentchalamon is there a reason why you don't PR against master here? You want to merge this into 2.4 but we are only accepting backports here.
Also while this looks mergeable so far, could you please also add a unit test for `PoolingShardConnection::getActiveShardId()` as you introduced that method and it should be covered for completeness and regression purposes.

vincentchalamon:
@deeky666 You're right, I just create it on master: https://github.com/doctrine/dbal/pull/903

PR #898, opened 2015-08-26
Add PDO-based dblib driver

livkiss:
This driver may be necessary if you are trying to connect to a MS SQL Server
database from Linux based OS. The pdo-sqlsrv driver is only supported when PHP
itself is running on a Windows machine.

This driver enables access to SQL Server via the [freetds](http://www.freetds.org/) library.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/PDODblib/Connection.php
+++ b/lib/Doctrine/DBAL/Driver/PDODblib/Connection.php
@@ -1,0 +1,54 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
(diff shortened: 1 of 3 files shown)


zeroedin-bill:
@felixkiss I've actually written this before, and used it. However, due to a lot of finickiness with bound parameters and PDO attributes, SQL Server via pdo_dblib isn't really suitable for general use. 

I don't think this is a good candidate for inclusion in DBAL. MS have stated that they are working on Linux support for their sqlsrv driver (https://github.com/Azure/msphpsql/issues/43). Once that's released, the issue of connectivity from Linux becomes moot.

felixkiss:
I am using it in a project right now (in combination with `doctrine/doctrine2` and have no issues with it.

I just wanted to get rid of my fork to speed up `composer install` (since it doesn't have to parse my fork via Github API if it is available in `doctrine/dbal`).

Linux support for the sqlsrv driver is IMO not a good argument, because we can't say how long it will take them.

zeroedin-bill:
@felixkiss You can actually keep your driver in your own namespace, and simply specify it in the configuration parameters. See http://doctrine-dbal.readthedocs.org/en/latest/reference/configuration.html#driver

deeky666:
Closing for the reasons explained. Let's wait for a stable native Linux implementation of a SQL Server driver. For now you can use the workaround suggested by @zeroedin-bill

PR #899, opened 2015-08-28
add `requiresSQLCommentHint` in `DateIntervalType`

vbartusevicius:
According to comment https://github.com/doctrine/dbal/pull/854#issuecomment-126935212

### DIFF
--- a/lib/Doctrine/DBAL/Types/DateIntervalType.php
+++ b/lib/Doctrine/DBAL/Types/DateIntervalType.php
@@ -67,4 +67,12 @@ class DateIntervalType extends Type
 
         return $interval;
     }
+    
+    /**
+     * {@inheritdoc}
+     */
+    public function requiresSQLCommentHint(AbstractPlatform $platform)
+    {
+        return true;
+    }
 }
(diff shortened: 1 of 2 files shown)


deeky666:
@v-bartusevicius thanks for the addition. We need at least a unit test case for the newly added method. Also I would like to have the functional test suite extended to verify that the mapping is working correctly now. A test case should be added [here](https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php)

v-bartusevicius:
@deeky666 - I did something here... not sure what as tests are skipped locally.
If You could provide me some more information on what You want me to extend in `DbalFunctionalTestCase` (or elsewhere), I would try to implement that.

zeroedin-bill:
@v-bartusevicius I think he's referring to https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/Functional/TypeConversionTest.php

deeky666:
@v-bartusevicius merged manually with an additional test added. Thanks!

PR #900, opened 2015-08-30
treat enum as string

shivergard:
fix for https://laracasts.com/discuss/channels/general-discussion/got-fatal-error-when-running-migrate

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -1005,6 +1005,7 @@ class MySqlPlatform extends AbstractPlatform
             'binary'        => 'binary',
             'varbinary'     => 'binary',
             'set'           => 'simple_array',
+            'enum'           => 'string'
         );
     }
 


kimhemsoe:
I do not think it is a good idea to treat enum as a string for every user. We have this behavior and possible solutions well documented. http://doctrine-orm.readthedocs.org/en/latest/cookbook/mysql-enums.html

shivergard:
Might be you are right. Just there is an issue on Laravel Migrations with field updates. I'll use this forked versions and will look at the issue overall. Cause , not fully competent in dbal code yet.

kimhemsoe:
I think the problem needs to be fixed in Laravel migrations with either adding the type to dbal or verbose error messages. I wonder if this is not an issue for all types we do not know. Migrations should not care about if a type can be mapped to a string or not, it should just use the specified type. 

For doctrine migrations we even recommend people should not use anything else then addSql(), as dbal schema will break in this usecase (I am guessing Artisan is using dbal Schema). That is one of many reason why doctrine migrations is still in alpha and not in active development.

Btw the issue you link is about missing dependencies, nothing about mysql enums.

kimhemsoe:
Ok, just read some of the laravel migrations code. The best fix in my humble opinion (I do not know schema that well) is to get rid of the dbal requirement. Dbal schema 2.x will never work for this usecase and support can not be added without breaking BC. This also seems to align with the types of issues that migrations get reported the most.

I may be wrong about this, as i said i do know the schema that well.

kimhemsoe:
I am closing this for reasons explained.

PR #903, opened 2015-09-08
Override methods for sharding connection

vincentchalamon:
(Originally https://github.com/doctrine/dbal/pull/896)
Required by https://github.com/doctrine/DoctrineBundle/pull/456
- [x] Override methods on [PoolingShardConnection](http://www.doctrine-project.org/api/dbal/2.4/class-Doctrine.DBAL.Sharding.PoolingShardConnection.html)
- [x] Add unit tests

### DIFF
--- a/lib/Doctrine/DBAL/Sharding/PoolingShardManager.php
+++ b/lib/Doctrine/DBAL/Sharding/PoolingShardManager.php
@@ -19,6 +19,8 @@
 
 namespace Doctrine\DBAL\Sharding;
 
+use Doctrine\DBAL\Sharding\ShardChoser\ShardChoser;
+
 /**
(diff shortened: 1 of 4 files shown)


lib/Doctrine/DBAL/Sharding/PoolingShardManager.php
```diff
@@ -18,6 +18,7 @@
  */
 
 namespace Doctrine\DBAL\Sharding;
+use Doctrine\DBAL\Sharding\ShardChoser\ShardChoser;
```
stof: please add an empty line between the namespace and the use statements

deeky666:
@vincentchalamon thanks!

PR #904, opened 2015-09-10
Fix `tearDown` of some functional test cases

deeky666:
The manual tear down of some test cases doesn't work if an exception is thrown before expecting it, thus making following test fail also. See [here](https://travis-ci.org/doctrine/dbal/jobs/79671519#L398)


zeroedin-bill:
:+1:

PR #908, opened 2015-09-17
Fix Oracle connection string without dbname

mRoca:
When using the DoctrineBundle, the`CreateDatabaseDoctrineCommand` and `DropDatabaseDoctrineCommand` commands unset the `dbname` parameter before getting the connection, in order to get a connection to the instance and not to the database which has to be created or removed : https://github.com/doctrine/DoctrineBundle/blob/master/Command/CreateDatabaseDoctrineCommand.php#L76 , and https://github.com/doctrine/DoctrineBundle/blob/master/Command/DropDatabaseDoctrineCommand.php#L78 .

Without this parameter, the connect string can't be created. This PR checks if a `dbname` parameter is available before setting the CONNECT_DATA value.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
@@ -118,19 +118,26 @@ abstract class AbstractOracleDriver implements Driver, ExceptionConverterDriver
                 $params['port'] = 1521;
             }
 
-            $serviceName = $params['dbname'];
+            $serviceName = '';
 
             if ( ! empty($params['servicename'])) {
                 $serviceName = $params['servicename'];
+            } elseif(isset($params['dbname'])) {
+                $serviceName = $params['dbname'];
             }
 
-            $service = 'SID=' . $serviceName;
-            $pooled  = '';
-            $instance = '';
+            if (empty($serviceName)) {
+                throw new \InvalidArgumentException('No service name provided for Easy Connect String.');
+            }
 
             if (isset($params['service']) && $params['service'] == true) {
-                $service = 'SERVICE_NAME=' . $serviceName;
+                $service = '(SERVICE_NAME=' . $serviceName . ')';
+            } else {
+                $service = '(SID=' . $serviceName . ')';
             }
+
+            $pooled = '';
+            $instance = '';
 
             if (isset($params['instancename']) && ! empty($params['instancename'])) {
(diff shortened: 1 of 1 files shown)


deeky666:
@mRoca AFAIK connecting to an Oracle database without a database instance reference (SID, SERVICE_NAME). So without the `$params['dbname']` parameter it is not possible to connect to Oracle which in return means that we cannot open up "temporary" connections for tasks like those used in DoctrineBundle. Oracle does not support these operations via SQL anyways. We should rather require the parameter and throw an InvalidArgumentException instead to make things clear. Then in the DoctrineBundle commands those kind of platforms should be checked and a reasonable Exception thrown there too, to make clear that those commands cannot be used with platforms that don't support it.

mRoca:
@deeky666 : I've updated the PR, an InvalidArgumentException is now thrown without valid SERVICE_NAME or SID.

lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
```diff
@@ -118,20 +118,27 @@ protected function getEasyConnectString(array $params)
...
-            $instance = '';
+            if (empty($serviceName)) {
+                throw new \InvalidArgumentException('No service name provided for Easy Connect String.');
+            }
 
             if (isset($params['service']) && $params['service'] == true) {
-                $service = 'SERVICE_NAME=' . $serviceName;
+                $service = '(SERVICE_NAME=' . $serviceName . ')';
+            } else {
+                $service = '(SID=' . $serviceName . ')';
```
deeky666: No need to add the parentheses here. Having to add them in multiple statements is cumbersome, when you can add them later on. Please leave it the way it was before.

lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
```diff
@@ -118,20 +118,27 @@ protected function getEasyConnectString(array $params)
                 $params['port'] = 1521;
             }
 
-            $serviceName = $params['dbname'];
+            $serviceName = '';
 
             if ( ! empty($params['servicename'])) {
                 $serviceName = $params['servicename'];
+            } elseif(isset($params['dbname'])) {
```
deeky666: Space missing between `elseif` and opening parenthesis

deeky666:
Besides nit picking, looks good to me, but is missing a test. Afterwards can be merged IMO.

mRoca:
Closing this PR in favor of https://github.com/doctrine/dbal/pull/3297

PR #912, opened 2015-10-06
Fix bug about get default params on PoolingShardConnection

vincentchalamon:


Ocramius:
@vincentchalamon please *add* a test, don't modify existing ones, unless they're wrong.

vincentchalamon:
@Ocramius Existing tests were wrong. `getParams` method was supposed to return the `global` section of params merged with all params, instead of params only

deeky666:
@vincentchalamon thanks!

vincentchalamon:
@deeky666 Thanks for merging. Do you know when a new release of `doctrine/dbal` will be published ?

deeky666:
@vincentchalamon this fix is part of 2.5.3 which is published already quite some time.

PR #913, opened 2015-10-10
Add enum type to mysql platform

windok:
Added enum type to mysql platform, because doctrine-migrations could not generate database schema with enum fields in table.


windok:
New pull request
https://github.com/doctrine/dbal/pull/914

PR #914, opened 2015-10-10
Add enum type to mysql platform

windok:
Added enum type to mysql platform, because doctrine-migrations could not generate database schema with enum fields in table.


deeky666:
This has been suggested several time now. We cannot support `ENUM` in the core, as those column types contain state and DBAL's type system is based on flyweights. Please also see the documentation: http://doctrine-orm.readthedocs.org/projects/doctrine-orm/en/latest/cookbook/mysql-enums.html

PR #916, opened 2015-10-12
Fix return type of `Table::renameColumn()` and mark it as deprecated

aivus:
What do you think, I should use `@deprecated` or `@internal` here?


Ocramius:
@aivus `@deprecated` is correct, thanks!

PR #917, opened 2015-10-17
Remove GIT submodule entries and use `autoload-dev` for the test suite

Tobion:
1. remove obsolete entries from gitmodules which was forgotten in #876
2. composer.lock is not part of the repo, so its useless in gitattributes
3. DoctrineTestCase is not used and is a remnant pre-split
4. use autoload-dev completely instead of an init script, complement to #770 and doctrine/cache#98
5. add travis cache. before: ran for 56 min 40 sec. after: ran for 46 min 58 sec
6. setUp and tearDown are meant to be protected


tests/Doctrine/Tests/TestInit.php
```diff
@@ -1,17 +0,0 @@
-<?php
-/*
- * This file bootstraps the test environment.
- */
-
-error_reporting(E_ALL | E_STRICT);
-date_default_timezone_set('UTC');
```
Tobion: this workaround is fixed: https://github.com/travis-ci/travis-ci/issues/2523

Ocramius: It's still relevant for folks testing locally

Tobion: that doesn't belong to the test setup. it's random a random value that has nothing to to with the tests. for testing locally its part of the php setup otherwise it would fail for every project of those people anyway. not just dbal.

Ocramius: Agreed, but the php defaults are... a mess :-P

deeky666:
@Tobion awesome, thanks!

PR #919, opened 2015-10-19
add `sslrootcert` connection option to the `pdo_pgsql` driver

peterjmit:
Right now it is pretty hard to trust certificates for a server with doctrine. Postgres requires you to put the `root.crt` in the home directory for a user which becomes messy rather quickly.


deeky666:
@peterjmit thanks!

PR #920, opened 2015-11-05
Unknown database type geometry requested

brandfocus:


deeky666:
I'm sorry to disappoint you but this type isn't portable across different vendors and thus we cannot add it to the core. Please add a custom type in your application yourself for that purpose.

See here: http://doctrine-orm.readthedocs.org/projects/doctrine-dbal/en/latest/reference/types.html#custom-mapping-types

and here: http://doctrine-orm.readthedocs.org/projects/doctrine-orm/en/latest/cookbook/custom-mapping-types.html

PR #922, opened 2015-11-20
Allow usage of getNowExpression() value in default

geelweb:
This update allow the usage of the `AbstractPlatform::getNowExpression()` return value for the default.


morozov:
Closing as stale. It should be implemented via the support of default expressions.

PR #2259, opened 2015-12-12
fixed getListTableColumnsSQL for PostgreSQL

### LINKED ISSUE #2258: error when table namespace (schema) contains a hyphen (-)
I'm just building a project based on Laravel, test environment had schema was named "web-test" in postgresql. I have set database search_path to "web-test", everything worked in Laravel except changing column definitions. Migration code contains this line:

	$table->text('notes')->nullable()->change();

Migration gives error:

	[Doctrine\DBAL\Schema\SchemaException]
	There is no column with name 'notes' on table 'reports'.

There is a column named notes in the table. I have checked Doctrine\DBAL\Schema\Table why does it throw Doctrine\DBAL\Schema\SchemaException. Dumped $_columns property in method hasColumn but it was an empty array.

I have changed schema name to "webtest" and renaming column worked.

First I have reported this issue at https://github.com/doctrine/doctrine2/issues/5545 where I was asked to give details on how to reproduce it outside the context of laravel.

I have made a test for this situation:

	public function testTableWithSchemaContainingHyphen()
	{
		$this->createTableAndTestForColumns('hyphentest');
		$this->createTableAndTestForColumns('hyphen-test');
	}

	public function createTableAndTestForColumns ($schema) {
		$this->_conn->exec('CREATE SCHEMA "' . $schema . '"');
		$this->_conn->exec('SET SEARCH_PATH="' . $schema . '"');
		$table = new \Doctrine\DBAL\Schema\Table('tablename');
		$column = $table->addColumn('id', 'integer');
		$this->_sm->createTable($table);
		$table = $this->_sm->listTableDetails('tablename');
		$this->assertTrue($table->hasColumn('id'));
		$this->_sm->dropTable($table);
		$this->_conn->exec('DROP SCHEMA "' . $schema . '"');
	}

I have placed these at the end of file named below and ran phpunit:

	phpunit --filter testTableWithSchemaContainingHyphen tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php

It fails if there is a hyphen in schema name. It seems listTableDetails misses the table, returns the same empty table object that I have seen in Laravel.

kolesar-andras:
Fixes #2258, tests added.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -361,16 +361,8 @@ class PostgreSqlPlatform extends AbstractPlatform
      */
     private function getTableWhereClause($table, $classAlias = 'c', $namespaceAlias = 'n')
     {
-        $whereClause = $namespaceAlias.".nspname NOT IN ('pg_catalog', 'information_schema', 'pg_toast') AND ";
-        if (strpos($table, ".") !== false) {
-            list($schema, $table) = explode(".", $table);
-            $schema = "'" . $schema . "'";
-        } else {
-            $schema = "ANY(string_to_array((select replace(replace(setting,'\"\$user\"',user),' ','') from pg_catalog.pg_settings where name = 'search_path'),','))";
-        }
-
-        $table = new Identifier($table);
-        $whereClause .= "$classAlias.relname = '" . $table->getName() . "' AND $namespaceAlias.nspname = $schema";
+        $tableNameEscaped = str_replace("'", "''", $table);
+        $whereClause = $classAlias.".oid='".$tableNameEscaped."'::regclass::oid";
 
         return $whereClause;
     }
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
@@ -390,6 +390,53 @@ class PostgreSqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
         $this->assertEquals('json_array', $columns['foo']->getType()->getName());
         $this->assertEquals(true, $columns['foo']->getPlatformOption('jsonb'));
     }
+
+    /**
+     * @group DBAL-75
+     */
+    public function testTableWithSchemaContainingSpecialCharacters()
+    {
+        $this->createTableAndTestForColumns('charactertest');
+        $this->createTableAndTestForColumns('character-test');
+        $this->createTableAndTestForColumns('character"test');
+        $this->createTableAndTestForColumns('character\'test');
+        $this->createTableAndTestForColumns('character,test');
+        $this->createTableAndTestForColumns('character.test');
+    }
+
+    public function createTableAndTestForColumns ($schemaName) {
+        $pathBackup = $this->_conn->fetchColumn('SHOW search_path');
+        $tableName = 'dummy';
+        $schemaNameQuoted = '"' . str_replace('"', '""', $schemaName) . '"';
+        $this->_conn->exec('CREATE SCHEMA '. $schemaNameQuoted);
+        $this->_conn->exec('SET SEARCH_PATH=' . $schemaNameQuoted);
+        $table = new \Doctrine\DBAL\Schema\Table($tableName);
+        $column = $table->addColumn('id', 'integer');
+        $this->_sm->createTable($table);
+        $tableTest = $this->_sm->listTableDetails($schemaNameQuoted . '.' . $tableName);
+        $this->assertTrue($tableTest->hasColumn('id'));
+        $this->_sm->dropTable($table);
+        $this->_conn->exec('DROP SCHEMA ' . $schemaNameQuoted);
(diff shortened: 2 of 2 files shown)


lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -361,16 +361,8 @@ public function getListTableIndexesSQL($table, $currentDatabase = null)
...
     {
-        $whereClause = $namespaceAlias.".nspname NOT IN ('pg_catalog', 'information_schema', 'pg_toast') AND ";
-        if (strpos($table, ".") !== false) {
-            list($schema, $table) = explode(".", $table);
-            $schema = "'" . $schema . "'";
-        } else {
-            $schema = "ANY(string_to_array((select replace(replace(setting,'\"\$user\"',user),' ','') from pg_catalog.pg_settings where name = 'search_path'),','))";
-        }
-
-        $table = new Identifier($table);
```
deeky666: I think we'll have to leave this logic in for incoming already quoted identifiers. Example:

```php
// "foo'bar"

$tableIdentifier = new Identifier($table);
$tableName = $table->getName(); // foo'bar
$tableName = $this->quoteStringLiteral($tableName); // 'foo''bar'
```

This should be correct.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -361,16 +361,8 @@ public function getListTableIndexesSQL($table, $currentDatabase = null)
      */
     private function getTableWhereClause($table, $classAlias = 'c', $namespaceAlias = 'n')
     {
-        $whereClause = $namespaceAlias.".nspname NOT IN ('pg_catalog', 'information_schema', 'pg_toast') AND ";
-        if (strpos($table, ".") !== false) {
```
deeky666: IIRC we require this logic for full qualified table names. You can't simply leave this out.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -361,16 +361,8 @@ public function getListTableIndexesSQL($table, $currentDatabase = null)
...
-            list($schema, $table) = explode(".", $table);
-            $schema = "'" . $schema . "'";
-        } else {
-            $schema = "ANY(string_to_array((select replace(replace(setting,'\"\$user\"',user),' ','') from pg_catalog.pg_settings where name = 'search_path'),','))";
-        }
-
-        $table = new Identifier($table);
-        $whereClause .= "$classAlias.relname = '" . $table->getName() . "' AND $namespaceAlias.nspname = $schema";
+        $tableNameEscaped = str_replace("'", "''", $table);
+        $whereClause = $classAlias.".oid='".$tableNameEscaped."'::regclass::oid";
```
deeky666: Can you clear me up on this? What does it exactly do? Does it work for full qualified table names? How does it know which namespace to look on for unqualified table names?

tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
```diff
@@ -390,6 +390,53 @@ public function testJsonbColumn()
         $this->assertEquals('json_array', $columns['foo']->getType()->getName());
         $this->assertEquals(true, $columns['foo']->getPlatformOption('jsonb'));
     }
+
+    /**
+     * @group DBAL-75
+     */
+    public function testTableWithSchemaContainingSpecialCharacters()
```
deeky666: Please transform this into a dataProvider instead.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
kolesar-andras: Takes table name and asks PostgreSQL for oid. Filter matches oid instead of relname and nspname.

deeky666: I see. And if the table name is unqualified? Does it use the search path to evaluate the schema from which to retrieve the information?

kolesar-andras: Yes, this method uses search path. Works well with qualified and unqualified table names.

deeky666: Neat! So much simpler then :) The only thing we'll have to take care of then is handling incoming quoted identifiers:

```php
private function getTableWhereClause($table, $classAlias = 'c', $namespaceAlias = 'n')
{
    $table = new Identifier($table);
    $table = $this->quoteStringLiteral($table->getName());
    
    return sprintf('%s.oid = %s::regclass::oid', $classAlias, $table);
}
```

Something like that. Tests should also cover some of these cases. Quoted/unquoted, qualified/unqualified, special characters/regular characters

zeroedin-bill: @deeky666 Incoming quoted names aren't handled well in Identifier... the full and proper fix needs to occur in 3.0 I think.

deeky666: @zeroedin-bill agreed. Just wanted to make clear, that `Identifier` has to be used here.

Issue #2261, opened 2015-12-15
OCI8 - bindValue overwrite previous values


pdgcvs:
Simple successful insert example, to show that all columns get overwritten:
https://github.com/pdgcvs/DoctrineInsertPHP7

When trying to use persist method with OCI8 driver on PHP 7 all the column values got overwritten by one.
This also leads to many format validation errors preventing the data from being inserted into table.

The problem seems to be on OCI8Statement->bindValue function when the local variable $value is passed as reference to bindParam.

I tried to make a simple fix by storing values of bind call in an array (instance member) before passing it as reference to bindParam.

So far it seems to be working.

https://github.com/pdgcvs/dbal/commit/4906c6518e09d56063c6c1e9594987467071bc35

arcostasi:
This fix solved the problem that I was having with doctrine+oci8 on php7. Thanks my friend!

deeky666:
@pdgcvs is that actually the "new" expected behaviour of OCI8 on PHP7 or rather a bug in PHP7? We should try to avoid fixing PHP bugs in userland where possible. If it is a PHP bug, we could still fix that in DBAL as a workaround, but should then try to get that fixed in PHP itself in the long term.

arcostasi:
@deeky666 and if this bug was in earlier versions of php 7? The simple way that is fixed, if it's a bug of php 7, the code yet will keep running. Good that's now working on php 7.

Ocramius:
> @deeky666 and if this bug was in earlier versions of php 7? 

That's why they're called "patch releases": you'd simply upgrade

deeky666:
I'll leave this issue open for reference until it is fixed in the driver. See [PHP ticket](https://bugs.php.net/bug.php?id=71148). Let's see what's happening at PHP side before taking any action in Doctrine.

eisberg:
In [PHP ticket](https://bugs.php.net/bug.php?id=71148):
[2016-02-22 04:41 UTC] sixd@php.net
Looks like an expected change with PHP 7.

OCI8 needs the zval of the bound variable to be available when OCIExecute is called.
Try using pass by reference in your bindValue() function, or don't bind in a function.
This is similar to example #3 in http://php.net/manual/en/function.oci-bind-by-name.php

From the dev who analyzed this:

"Reason for this new behavior in PHP 7 is that, unlike in PHP 5.6, zval (_zend_value) in PHP 7.0 is no more have reference count field. Only its values field (zend_value) has reference count field.

[...]

Issue #1237, opened 2015-08-20
DBAL-1283: Oracle parameter binding broken


doctrinebot:
Jira issue originally created by user deepdiver:

On insert statements parameters are all bound to the last parameter value.

Please test with following test script

<?php
require 'vendor/autoload.php';

$config = new \Doctrine\DBAL\Configuration();
//..
$connectionParams = array(
        'dbname' => 'XE',
        'user' => 'autotest',
        'password' => 'owncloud',
    'host' => '10.0.0.7',
    'port' => 1521,
        'driver' => 'oci8',
);
$conn = \Doctrine\DBAL\DriverManager::getConnection($connectionParams, $config);

$conn->connect();

$conn->executeUpdate('DROP TABLE TEST');
$conn->executeUpdate('CREATE TABLE TEST(A VARCHAR2(4000) NOT NULL, B VARCHAR2(4000) NOT NULL, C VARCHAR2(4000) NOT NULL)');

$conn->insert('TEST', [
    'A' => '1',
    'B' => '2',
    'C' => '3']);

$all = $conn->fetchAll('select \* from TEST');
var_dump($all);

see https://github.com/owncloud/core/issues/18425

morozov:
Fixed by https://github.com/doctrine/dbal/pull/2434.

PR #2262, opened 2015-12-15
Issue #2261 - OCI8 Driver PHP 7 - Fix bindValue overwriting other params. 

pdgcvs:
https://github.com/doctrine/dbal/issues/2261

### DIFF
--- a/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
+++ b/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
@@ -72,6 +72,11 @@ class OCI8Statement implements \IteratorAggregate, Statement
     protected $_paramMap = array();
 
     /**
+     * @var array
+     */
+    protected $_bindings = array();
+
+    /**
      * Creates a new OCI8Statement that uses the given connection handle and SQL statement.
      *
      * @param resource                                  $dbh       The connection handle.
@@ -134,7 +139,8 @@ class OCI8Statement implements \IteratorAggregate, Statement
      */
     public function bindValue($param, $value, $type = null)
     {
-        return $this->bindParam($param, $value, $type, null);
+        $this->_bindings[$param] = $value;
+        return $this->bindParam($param, $this->_bindings[$param], $type, null);
     }
 
     /**


Ocramius:
Requires a test case scenario that validates the fix.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -72,6 +72,11 @@ class OCI8Statement implements \IteratorAggregate, Statement
     protected $_paramMap = array();
 
     /**
+     * @var array
+     */
+    protected $_bindings = array();
```
deeky666: please do not introduce a proptected member here, unless completely necessary. Otherwise we will have to keep BC for that afterwards. Also please do not use the old underscore prefix naming here. `private $bindings` is fine.

deeky666:
Delayed for now as I'd like to wait for the PHP guys to come up with a fix before we start working around here.

nickvergessen:
Considered not a bug by [PHP](https://bugs.php.net/bug.php?id=71148):

>  [2016-02-22 04:41 UTC] sixd@php.net
>Looks like an expected change with PHP 7.
>
>OCI8 needs the zval of the bound variable to be available when OCIExecute is called.
>Try using pass by reference in your bindValue() function, or don't bind in a function.
>This is similar to example #3 in http://php.net/manual/en/function.oci-bind-by-name.php
>
>From the dev who analyzed this:
>
>"Reason for this new behavior in PHP 7 is that, unlike in PHP 5.6, zval (_zend_value) in PHP 7.0 is no more have reference count field. Only its values field (zend_value) has reference count field.
>
>Since we store the placeholder (i.e., zval variable) [not its content (zend_value)] in php_oci_bind
structure, it gets overwritten when oci_bind_by_name() is called with same variable and different content in it."

nickvergessen:
Or to translate it, and give more insite why this PR works:
`$value` is a local variable, in php7 it's always the same local variable. So you always pass the same variable placeholder to `bindParam` (by reference).
So at the end all placeholders point to the same variable.

By using the array, you always pass a different reference and therefor the values are not overwritten.
As a test you can use https://github.com/doctrine/dbal/issues/1237

Ocramius:
So basically what is happening is that the values are GC'd before we can ever use them.
How the heck are people supposed to use `oci_bind_by_name()` if this is not a bug? :-\

arcostasi:
Ok, it's a bug! how can we try to solve in the doctrine?
Apparently don't have solution in oci8 with php7. :(

Actually, we solved the problem after applied the patch that use solution in the issue “https://github.com/doctrine/dbal/issues/2261”

soyuka:
So, I've tried this pull-request and I ran the test suite on an oracle-12 instance (https://hub.docker.com/r/sath89/oracle-12c/):

- There is an issue with blobs, this should store a reference to the lob directly from the bindValue function. I've fixed this and I'm willing to do another pull-request if needed.
- I've one failing test and I'm wondering if it's related or if it's a misconfiguration:
```
1) Doctrine\Tests\DBAL\Functional\Schema\OracleSchemaManagerTest::testListDatabases
Doctrine\DBAL\Exception\DriverException: An exception occurred while executing 'CREATE USER c##test_create_database IDENTIFIED BY oracle':

ORA-65094: invalid local user or role name
```
My test configuration is available in [this gist](https://gist.github.com/soyuka/d8af834377228fe24687).

Btw, without this patch none of the OCI8 tests will pass because this breaks the `executeUpdate` when there are more than one binding. With only one binding, the value is still available, can be misleading when debugging. 
For my information, what would be a valid test-case for this to be merged?

eisberg:
We expect look !!!

lynx91:
If someone got a solution for this problem without manipulating the doctrine sources locally i would be interesseted in this solution.

soyuka:
No solution on my end, this also adds issues regarding CLOB/BLOB to me. 

We just need to determine whether it's a php bug or a feature, in which case we have to fix the OCI8Statement library.

lynx91:
As written before, its NOT a bug by PHP, its a "new behavior in PHP 7"

soyuka:
@lynx91 I agree, just saying this because it's still labeled as "Delayed".

> How the heck are people supposed to use oci_bind_by_name() if this is not a bug? :-\

He has a point, so I'm not sure everyone agrees to this behavior...

lynx91:
Hm..it seems to be a difficult decision...hope, that there will be any solution in the near future...thx for your response

mmucklo:
@soyuka Have you done another PR with your CLOB/BLOB fix?

If your PR works, maybe the solution is just to clone off your fork for awhile until the Doctrine team stops playing ostrich on this issue -- since apparently the PHP team is not considering it a bug but a feature (as has been repeatedly pointed out)...

soyuka:
@mmucklo I need time to check this further, I have no clean fix yet. 

However, you can check [this commit](https://github.com/soyuka/dbal/commit/61755786ffb628f19a50b2a78b022bd92d21baf5) on my fork. It's almost there, I remember there was a small issue on top of it (maybe unrelated). 
Don't use my fork as is because I replaced this with a crap hotfix where I comment every lob conditions!

Garfield-fr:
It's possible to merge this commit, because Statement on Oracle doesn't work on PHP7 :(

soyuka:
Nope this one should be closed in favor of #2434

deeky666:
Closing in favour of #2434.

Ocramius:
Handled in https://github.com/doctrine/dbal/pull/2434

PR #2264, opened 2015-12-16
Add sqlite3 driver (BenMorel)

billschaller:
This PR supplements and closes #840.

I made some changes to the exception handling in the driver. SQLite3 is one of the only non-PDO drivers that throws exceptions instead of errors. The changes I made allow the exceptions thrown by the underlying driver to be chained into the SQLite3Exception.

I also added some phpdoc @ since and @ author annotations.

/ping @BenMorel @deeky666 @stof
## From BenMorel's Original PR #840:

This PR adds an additional driver using the [SQLite3](http://php.net/manual/fr/book.sqlite3.php) class, in addition to the PDO_SQLITE driver.
## Why this driver?

Even though the PDO SQLite driver is already available to interact with SQLite databases, PDO currently has a big limitation: it [does not allow to load SQLite extensions](https://bugs.php.net/bug.php?id=64810).

This functionality is provided by the `SQLite3` class, which becomes your only option if you need to use your PHP application with an SQLite extension such as [SpatiaLite](http://www.gaia-gis.it/gaia-sins/).

### DIFF
--- a/lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
@@ -25,7 +25,9 @@ use Doctrine\DBAL\Driver\AbstractSQLiteDriver;
 /**
  * The SQLite3 driver.
  *
- * @since 2.0
+ * @since 2.6
+ * @author Ben Morel <ben@benjaminmorel.com>
+ * @author Bill Schaller <bill@zeroedin.com>
  */
 class Driver extends AbstractSQLiteDriver
 {
--- a/lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
@@ -53,7 +53,7 @@ class Driver extends AbstractSQLiteDriver
 
         try {
             $connection = new SQLite3Connection(
-                isset($params['path']) ? $params['path'] : ':memory:',
+                !empty($params['path']) ? $params['path'] : ':memory:',
                 isset($params['flags']) ? $params['flags'] : null,
                 isset($params['encryption_key']) ? $params['encryption_key'] : null
             );
--- a/lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
@@ -1,0 +1,78 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Driver\SQLite3;
+
+use Doctrine\DBAL\DBALException;
+use Doctrine\DBAL\Driver\AbstractSQLiteDriver;
+
+/**
+ * The SQLite3 driver.
+ *
+ * @since 2.0
+ */
+class Driver extends AbstractSQLiteDriver
+{
+    /**
+     * @var array
+     */
+    protected $_userDefinedFunctions = array(
+        'sqrt' => array('callback' => array('Doctrine\DBAL\Platforms\SqlitePlatform', 'udfSqrt'), 'numArgs' => 1),
+        'mod'  => array('callback' => array('Doctrine\DBAL\Platforms\SqlitePlatform', 'udfMod'), 'numArgs' => 2),
+        'locate'  => array('callback' => array('Doctrine\DBAL\Platforms\SqlitePlatform', 'udfLocate'), 'numArgs' => -1),
+    );
+
+    /**
+     * {@inheritdoc}
+     */
+    public function connect(array $params, $username = null, $password = null, array $driverOptions = array())
+    {
+        if (isset($driverOptions['userDefinedFunctions'])) {
+            $this->_userDefinedFunctions = array_merge(
+                $this->_userDefinedFunctions, $driverOptions['userDefinedFunctions']);
+            unset($driverOptions['userDefinedFunctions']);
+        }
+
+        try {
+            $connection = new SQLite3Connection(
+                isset($params['path']) ? $params['path'] : ':memory:',
+                isset($params['flags']) ? $params['flags'] : null,
+                isset($params['encryption_key']) ? $params['encryption_key'] : null
+            );
+        } catch (SQLite3Exception $e) {
+            throw DBALException::driverException($this, $e);
+        }
+
+        $sqlite3 = $connection->getSQLite3();
+
+        foreach ($this->_userDefinedFunctions as $fn => $data) {
+            $sqlite3->createFunction($fn, $data['callback'], $data['numArgs']);
+        }
+
+        return $connection;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        return 'sqlite3';
+    }
+}
--- a/lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
@@ -26,7 +26,7 @@ use Doctrine\DBAL\Driver\AbstractSQLiteDriver;
  * The SQLite3 driver.
  *
  * @since 2.6
- * @author Ben Morel <ben@benjaminmorel.com>
+ * @author Ben Morel <benjamin.morel@gmail.com>
  * @author Bill Schaller <bill@zeroedin.com>
  */
 class Driver extends AbstractSQLiteDriver
--- a/lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
@@ -32,7 +32,7 @@ class Driver extends AbstractSQLiteDriver
     /**
      * @var array
      */
-    protected $_userDefinedFunctions = array(
+    private $userDefinedFunctions = array(
         'sqrt' => array('callback' => array('Doctrine\DBAL\Platforms\SqlitePlatform', 'udfSqrt'), 'numArgs' => 1),
         'mod'  => array('callback' => array('Doctrine\DBAL\Platforms\SqlitePlatform', 'udfMod'), 'numArgs' => 2),
         'locate'  => array('callback' => array('Doctrine\DBAL\Platforms\SqlitePlatform', 'udfLocate'), 'numArgs' => -1),
@@ -44,8 +44,8 @@ class Driver extends AbstractSQLiteDriver
     public function connect(array $params, $username = null, $password = null, array $driverOptions = array())
     {
         if (isset($driverOptions['userDefinedFunctions'])) {
-            $this->_userDefinedFunctions = array_merge(
-                $this->_userDefinedFunctions, $driverOptions['userDefinedFunctions']);
+            $this->userDefinedFunctions = array_merge(
+                $this->userDefinedFunctions, $driverOptions['userDefinedFunctions']);
             unset($driverOptions['userDefinedFunctions']);
         }
 
@@ -61,7 +61,7 @@ class Driver extends AbstractSQLiteDriver
 
         $sqlite3 = $connection->getSQLite3();
 
-        foreach ($this->_userDefinedFunctions as $fn => $data) {
+        foreach ($this->userDefinedFunctions as $fn => $data) {
             $sqlite3->createFunction($fn, $data['callback'], $data['numArgs']);
         }
 
--- a/lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Abstract.php
+++ b/lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Abstract.php
@@ -1,0 +1,62 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Driver\SQLite3;
+
+/**
+ * Base functionality for SQLite3Connection and SQLite3Statement.
+ */
+abstract class SQLite3Abstract
+{
+    /**
+     * @var \SQLite3
+     */
+    protected $sqlite3;
+
+    /**
+     * Runs code interacting with native SQLite3 objects, and catches errors to throw a proper exception.
+     *
+     * @param \Closure $function A function calling native SQLite3 objects.
+     *
+     * @return mixed The return value of the given function.
+     *
+     * @throws SQLite3Exception If an error occurs.
+     */
+    protected function call(\Closure $function)
+    {
+        // Temporary set the error reporting level to 0 to avoid any warning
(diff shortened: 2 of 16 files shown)


BenMorel:
Thanks for taking over on this one!

zeroedin-bill:
@BenMorel Driver now throws exceptions instead of warnings. As far as I can tell, the driver will never throw warnings once enableExceptions is called.

https://github.com/php/php-src/blob/master/ext/sqlite3/sqlite3.c

BenMorel:
@zeroedin-bill I did not know about `SQLite3::enableExceptions()`. In fact, this [method](http://nl1.php.net/manual/en/sqlite3.enableexceptions.php) is [not documented](http://bohwaz.net/p/Undocumented-PHP-SQLite3-features-openBlob-readOnly-enableExceptions), so can we really rely on it? Undocumented features can be removed without warning as this would not be a BC break.

Plus, I would advise against enabling this feature by default. I agree that exceptions are better than warnings, no doubt about that. But if the developer passes an instance of `SQLite3` to the DBAL layer, he would be very surprised if the DBAL changed the behaviour of his instance. He might be expecting the warnings for some code he uses that calls the `SQLite3` instance directly, and might have an error handling mechanism configured for that.

In the light of this `enableExceptions()` method though, what I would advise is keeping what you've done (`try {} catch {}` + calling `throwExceptionOnError()` right after), but leaving the `@` in front of the SQLite3 method calls, to mute any warnings that would be triggered inside the DBAL layer.

That's the best of both worlds: whether or not the `SQLite3` instance given by the developer is configured to throw exceptions, the behaviour of the DBAL layer is consistent, by throwing exceptions and not emitting warnings.

zeroedin-bill:
@BenMorel The method is not documented, but it has been in the driver for a very long time, and the driver is quite stable.

There are a few places in the driver on codepaths used by the DBAL driver where a warning can be triggered without setting the error code in the driver. If the driver is not in exception mode, and we're using @ on all the calls, the error could simply be swallowed by the DBAL driver. There are some places in other code paths not used by DBAL where warnings can be thrown without setting the error code also. 

In general, I would say it is a bad idea to use this driver without setting it to throw exceptions. As the driver is written, it's not possible to pass in a initialized SQLite3 instance, but you could get the SQLite3 instance from the driver and then call enableExceptions(false) to set it back to warning mode. I can't say I'd like to support that use case.

I think putting back the @ is probably the best compromise, but when a driver actually supports throwing exceptions for all error conditions, I'm really hesitant to support using it in the warning mode.

lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
```diff
@@ -0,0 +1,80 @@
...
+ * @since 2.6
+ * @author Ben Morel <benjamin.morel@gmail.com>
+ * @author Bill Schaller <bill@zeroedin.com>
+ */
+class Driver extends AbstractSQLiteDriver
+{
+    /**
+     * @var array
+     */
+    private $userDefinedFunctions = array(
```
deeky666: Maybe I am missing something, but we already have those defined in the abstract class. So why duplicating it here?

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
```diff
@@ -0,0 +1,216 @@
...
+class SQLite3Connection extends SQLite3Abstract implements Connection, ServerInfoAwareConnection
+{
+    /**
+     * @param string      $filename
+     * @param int|null    $flags
+     * @param string|null $encryption_key
+     *
+     * @throws SQLite3Exception
+     */
+    public function __construct($filename, $flags = null, $encryption_key = null)
```
deeky666: Please use camelCase for variable names.

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
```diff
@@ -0,0 +1,216 @@
...
+    {
+        try {
+            if ($flags === null) {
+                $this->sqlite3 = new \SQLite3($filename);
+            } elseif ($encryption_key === null) {
+                $this->sqlite3 = new \SQLite3($filename, $flags);
+            } else {
+                $this->sqlite3 = new \SQLite3($filename, $flags, $encryption_key);
+            }
```
deeky666: Why do we require this condition? Can't we pass `null`s into the constructor?

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Abstract.php
```diff
@@ -0,0 +1,72 @@
...
+        throw new SQLite3Exception($errorMessage, null, $errorCode);
+    }
+
+    /**
+     * Wraps an exception thrown by SQLite3 in a SQLite3Exception.
+     *
+     * @param \Exception $driverException
+     * @return SQLite3Exception
+     */
+    protected function wrapDriverException(\Exception $driverException)
```
deeky666: I think we should rather use a static factory method in SQLite3Exception like:

``` php
public static function fromNativeException(\Exception $nativeException, \SQLite3 $connection = null)
{
}
```

We do that in other driver exceptions, too and it feels like the better place to do that IMO.

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Abstract.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        $errorMessage = $driverException->getMessage();
+        $errorCode = $driverException->getCode();
+
+        if ($this->sqlite3 instanceof \SQLite3) {
+            $errorMessage = $this->sqlite3->lastErrorMsg();
+            $errorCode = $this->sqlite3->lastErrorCode();
+        }
```
deeky666: Isn't the SQLite3 exception already populated with these?

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Abstract.php
```diff
@@ -0,0 +1,72 @@
...
+     * @var \SQLite3
+     */
+    protected $sqlite3;
+
+    /**
+     * @return void
+     *
+     * @throws SQLite3Exception
+     */
+    protected function throwExceptionOnError()
```
deeky666: Is this method still necessary? With exceptions enabled on SQLite3, we should always get an exception, no?

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
```diff
@@ -0,0 +1,216 @@
...
+            throw $this->wrapDriverException($e);
+        }
+    }
+
+    /**
+     * Returns the underlying SQLite3 object.
+     *
+     * @return \SQLite3
+     */
+    public function getSQLite3()
```
deeky666: I'm really against exposing the internal connection ressource to the outside world as it opens a can of worms. Why do we even need this besides registering custom functions (which could still be done without exposing the connection here)?

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
```diff
@@ -0,0 +1,216 @@
...
+        return $this->sqlite3;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function prepare($prepareString)
+    {
+        try {
+            $statement = @ $this->sqlite3->prepare($prepareString);
```
deeky666: Can't we omit the `@` if exceptions are enabled?

deeky666: We should defer this to the constructor of the statement as it is done in other driver implementations, so that the statement resource is encapsulated by `SQLite3Statement`

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function beginTransaction()
+    {
+        try {
+            $this->sqlite3->exec('BEGIN');
+        } catch (\Exception $e) {
```
deeky666: Hmm this is something I never thought about. The contract here is to return `false` on error but we are actually throwing exception here (in other drivers too) :S

deeky666: Gosh looks like we have this issue in a numerous of methods... Exceptions are not really part of the connection interface unfortunately :( But as we are doing similar things in other drivers, I could like with that here I guess.

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Exception.php
```diff
@@ -0,0 +1,33 @@
...
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Driver\SQLite3;
+
+use Doctrine\DBAL\Driver\AbstractDriverException;
+
+/**
+ * Exception thrown by SQLite3Connection.
```
deeky666: I would reword here, as not only the connection is throwing this exception. I would prefer something about what this actually represents like `SQLite3 driver exception.`.

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
BenMorel: I'm afraid not:

    new Sqlite3(':memory:'); // OK
    new Sqlite3(':memory:', null, null); // Exception: Unable to open database: out of memory

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Abstract.php
BenMorel: The exception thrown is the base `Exception` class (an heresy if you ask me).

The error message is the same, but the error code differs:

```
$s = new SQLite3(':memory:');
$s->enableExceptions(true);

try {
    $s->exec('invalid query');
} catch (\Exception $e) {
    echo $e->getMessage(), PHP_EOL; // near "invalid": syntax error
    echo $e->getCode(), PHP_EOL; // 0
}

echo $s->lastErrorMsg(), PHP_EOL; // near "invalid": syntax error
echo $s->lastErrorCode(), PHP_EOL; // 1
```

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Statement.php
```diff
@@ -0,0 +1,394 @@
...
+                    $this->bindValue(is_int($key) ? $key + 1 : $key, $value);
+                }
+            }
+
+            $result = $this->stmt->execute();
+
+            $this->lastErrorCode = $this->sqlite3->lastErrorCode();
+            $this->lastErrorMessage = $this->sqlite3->lastErrorMsg();
+
+            $this->result = $result;
```
deeky666: You can assign the result directly when calling `$this->stmt->execute()`

BenMorel:
@zeroedin-bill If we cannot give the driver an existing instance of `SQLite3`, then creating the instance pre-configured to throw exceptions is fine to me (as you say, getting the instance from the driver and calling `enableExceptions(false)` would be... weird (although still possible).

However, if I look at the PDO driver, [it is possible to pass an existing `PDO` instance](http://doctrine-orm.readthedocs.org/projects/doctrine-dbal/en/latest/reference/configuration.html#driver):

- `pdo`: Specifies an existing PDO instance to use.

If we somehow want to support the same for SQLite3, then we have to support exceptions *and* warnings. Do we want this, by the way?

> There are a few places in the driver on codepaths used by the DBAL driver where a warning can be triggered without setting the error code in the driver. If the driver is not in exception mode, and we're using @ on all the calls, the error could simply be swallowed by the DBAL driver. There are some places in other code paths not used by DBAL where warnings can be thrown without setting the error code also.

What code are we talking about here? To me, the only errors "swallowed" (to be converted to exceptions anyway) are in the methods implementing the `Connection` interface, so there is no interference with any other code as far as I can see. You will never get a swallowed warning without an exception.

I don't like adding unnecessary complexity either, but due to PHP's historical support for several different error handling mechanism, I think that the driver layer handling both use cases is the correct way to go. Unless you want to jail the `SQLite3` instance and make it inaccessible to the developer.

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Statement.php
```diff
@@ -0,0 +1,394 @@
...
+    {
+        return $this->rowCount;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function closeCursor()
+    {
+        return $this->result->finalize();
```
deeky666: we should check if we have a result before calling `finalize()` on it.

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
BenMorel: Isn't it still time to "fix" the other drivers, if this is just a matter of changing the docblock for the method? If the behaviour stays the same, this is not a BC break, right?

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Abstract.php
BenMorel: It depends on what's decided. See my comments [here](https://github.com/doctrine/dbal/pull/2264#issuecomment-165269382) and [here](https://github.com/doctrine/dbal/pull/2264#issuecomment-165407106).

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Statement.php
```diff
@@ -0,0 +1,394 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function columnCount()
+    {
+        if ($this->result) {
+            try {
+                return $this->result->numColumns();
```
deeky666: Doesn't look like this method is ever going to throw an exception. Guess we do not need to catch here.

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
BenMorel: Agree, *unless* we want to [allow to pass an existing instance](https://github.com/doctrine/dbal/pull/2264#issuecomment-165407106).

Otherwise, adding a specific method to the driver to load an extension is good to me.

That being said, I'm not sure about this whole "jailed SQLite3 instance" thing. I sometimes had to share a (PDO) connection between Doctrine and third-party code, and this worked well.

The same goes with SQLite3. Why create two instances of `SQLite3` to connect to the same database at the same time, when the Doctrine driver could use the same instance as third-party code, just working consistently with however it's configured? (which is not such a big deal, IMHO).

lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
zeroedin-bill: Already have what defined in the abstract class? The UDFs? The UDFs were already defined in PDOSqlite\Driver, not AbstractSQLiteDriver. Would make sense to pull them up though.

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
zeroedin-bill: Yup, will do.

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
zeroedin-bill: This is because internally, SQLite3::__construct is an alias of SQLite3::open -
```php
public void SQLite3::open ( string $filename [, int $flags = SQLITE3_OPEN_READWRITE | SQLITE3_OPEN_CREATE [, string $encryption_key ]] )
```

I'll rejigger this to call it with open's default parameters if the $flags param isn't set. $encryptionKey can safely be null.

zeroedin-bill:
Opened patch https://edit.php.net/?patchID=1868&project=PHP to address documentation issues.

Opened bug https://bugs.php.net/bug.php?id=71150 to address issues with enableExceptions.

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Abstract.php
billschaller: Yes. If you look at the driver source code, it looks like the exception error code is pretty much always 0.

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
zeroedin-bill: The only other driver that allows this is mysqli. I'm really strongly against returning the SQLite3 instance as well. If we want to add specific support methods for SQLite3 features, I'm just fine with that.

If you have external code that uses SQLite3 directly, we could simply change the driver so that it extends \SQLite3 instead of completely wrapping it in a non-compatible interface. Methods we don't want callable by userland (such as enableExceptions) we could simply add a stub that throws something like a NotImplementedException.

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
billschaller: Yes.

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
zeroedin-bill: That's a great point, @deeky666 - the drivers are extremely inconsistent. Even DBAL\Connection has void return types for commit, rollback, and beginTransaction, where the contract has return true on success, false on error.

The question is: what is the better behavior? Return true/false on error or raise an exception? I'd argue that raising exceptions is the expected behavior, especially within the ORM. If you look at usages of beginTransaction, commit, and rollback, the ORM doesn't use the return values on those methods at all. It simply expects an exception to be thrown if there's a problem.

Using true/false return values is pretty fragile IMO. It relies on the value being returned throughout the call chain, and it also forces consumers to either create code paths to marshal error information via special getErrorInfo methods throughout all consumers, or via direct access to the underlying statement (bad).

I'd say throwing an exception is much cleaner, because even if the exception is wrapped multiple times, if they're properly handled, the $previous exception will chain back to the originally thrown one. It also allows us to abstract over the oddities of driver return values. Some functions may return true or void on error where others return false. Using exceptions throughout the DBAL driver stack lets us reason a bit more clearly about what is actually happening.

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Statement.php
zeroedin-bill: This method can throw an exception if the SQLite3Stmt initialization check fails. 

See: https://github.com/php/php-src/blob/master/ext/sqlite3/sqlite3.c#L1654

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
BenMorel: Inheritance instead of encapsulation is a no-no to me, and I'm not sure what problem it would solve anyway.

Now I've exposed my opinion: to me, the best (less bad) approach is to accept, and deal with, any `SQLite3` instance, however it's configured. But creating internally, and then forbidding external access to the `SQLite3` instance, is my second best choice, so I can live with that (you might be forcing a few developers to have 2 concurrent connections to the same database, though).

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
BenMorel: Fully agree with you on this one, @zeroedin-bill :+1:

lib/Doctrine/DBAL/Driver/SQLite3/Driver.php
deeky666: Ah yes you're right. Indeed I would prefer moving them to the abstract driver then to avoid duplication.

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
deeky666: Maybe it would make sense to disallow `null` then in the constructor for `$flags` and have the default flags already be defined as defualt constructor value...

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Connection.php
deeky666: Fully agree @zeroedin-bill. The only problem is we have contracts for those kind of things which we *SHOULD* not violate. Otherwise having contracts at all is rather useless. On the other hand as you mentioned, I guess if even ORM is relying on the false behaviour, fixing all the drivers to adhere to the contract will probably break more things than it solves.
I really have no idea which direction to go :> maybe @Ocramius has an opinion here...

lib/Doctrine/DBAL/Driver/SQLite3/SQLite3Statement.php
deeky666: agreed

morozov:
Closing as stale.

PR #2266, opened 2015-12-18
Add native JSON type support for MySQL >=5.7.8

ismailbaskin:
> As of MySQL 5.7.8, MySQL supports a native JSON data type that enables efficient access to data in  JSON (JavaScript Object Notation) documents. The JSON data type provides these advantages over storing JSON-format strings in a string column:

http://dev.mysql.com/doc/refman/5.7/en/json.html

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/MySQL578Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySQL578Platform.php
@@ -1,0 +1,56 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Platforms;
+
+/**
+ * Provides the behavior, features and SQL dialect of the MySQL >=5.7.8 database platform.
+ * Extends for json support >=5.7.8 version
+ *
+ * @author İsmail BASKIN <ismailbaskin1@gmail.com>
+ * @link   www.doctrine-project.org
+ * @since  2.5
+ */
+class MySQL578Platform extends MySQL57Platform
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function initializeDoctrineTypeMappings()
+    {
+        parent::initializeDoctrineTypeMappings();
+        $this->doctrineTypeMapping['json'] = 'json_array';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function hasNativeJsonType()
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getJsonTypeDeclarationSQL(array $field)
+    {
+        return 'JSON';
+    }
+}
--- a/lib/Doctrine/DBAL/Platforms/MySQL578Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySQL578Platform.php
@@ -53,4 +53,12 @@ class MySQL578Platform extends MySQL57Platform
     {
         return 'JSON';
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function getReservedKeywordsClass()
+    {
+        return 'Doctrine\DBAL\Platforms\Keywords\MySQL578Keywords';
+    }
 }
--- a/lib/Doctrine/DBAL/Platforms/MySQL578Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySQL578Platform.php
@@ -25,7 +25,7 @@ namespace Doctrine\DBAL\Platforms;
  *
  * @author İsmail BASKIN <ismailbaskin1@gmail.com>
  * @link   www.doctrine-project.org
- * @since  2.5
+ * @since  2.6
  */
 class MySQL578Platform extends MySQL57Platform
 {
--- a/docs/en/reference/platforms.rst
+++ b/docs/en/reference/platforms.rst
@@ -35,6 +35,7 @@ MySQL
 
 -  ``MySqlPlatform`` for version 5.0 and above.
 -  ``MySQL57Platform`` for version 5.7 and above.
+-  ``MySQL578Platform`` for version 5.7.8 and above.
 
 Oracle
 ^^^^^^
--- a/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
@@ -22,6 +22,7 @@ namespace Doctrine\DBAL\Driver;
 use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Driver;
 use Doctrine\DBAL\Exception;
+use Doctrine\DBAL\Platforms\MySQL578Platform;
 use Doctrine\DBAL\Platforms\MySQL57Platform;
 use Doctrine\DBAL\Platforms\MySqlPlatform;
 use Doctrine\DBAL\Schema\MySqlSchemaManager;
@@ -135,6 +136,10 @@ abstract class AbstractMySQLDriver implements Driver, ExceptionConverterDriver, 
         $minorVersion = isset($versionParts['minor']) ? $versionParts['minor'] : 0;
         $patchVersion = isset($versionParts['patch']) ? $versionParts['patch'] : 0;
         $version      = $majorVersion . '.' . $minorVersion . '.' . $patchVersion;
+
+        if (version_compare($version, '5.7.8', '>=')) {
+            return new MySQL578Platform();
+        }
 
         if (version_compare($version, '5.7', '>=')) {
             return new MySQL57Platform();
--- a/lib/Doctrine/DBAL/Platforms/Keywords/MySQL578Keywords.php
+++ b/lib/Doctrine/DBAL/Platforms/Keywords/MySQL578Keywords.php
@@ -1,0 +1,57 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Platforms\Keywords;
+
+/**
+ * MySQL 5.7.8 reserved keywords list.
+ *
+ * @author İsmail BASKIN <ismailbaskin1@gmail.com>
+ * @link   www.doctrine-project.org
+ * @since  2.6
+ */
+class MySQL578Keywords extends MySQL57Keywords
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        return 'MySQL578';
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @link https://dev.mysql.com/doc/refman/5.7/en/keywords.html
+     */
+    protected function getKeywords()
+    {
+        $parentKeywords = array_diff(parent::getKeywords(), array(
+            'NONBLOCKING',
+        ));
+
+        return array_merge($parentKeywords, array(
+            'GENERATED',
+            'OPTIMIZER_COSTS',
+            'STORED',
+            'VIRTUAL'
+        ));
+    }
+}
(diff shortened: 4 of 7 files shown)


lib/Doctrine/DBAL/Platforms/MySQL578Platform.php
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Doctrine\DBAL\Platforms;
+
+/**
+ * Provides the behavior, features and SQL dialect of the MySQL >=5.7.8 database platform.
+ * Extends for json support >=5.7.8 version
+ *
+ * @author İsmail BASKIN <ismailbaskin1@gmail.com>
+ * @link   www.doctrine-project.org
+ * @since  2.5
```
deeky666: `@since 2.6`

lib/Doctrine/DBAL/Platforms/MySQL578Platform.php
```diff
@@ -0,0 +1,56 @@
...
+ */
+class MySQL578Platform extends MySQL57Platform
```
deeky666: If we introduce a new platform, we also need to introduce a new reserved keywords class. See `MySQL57Platform` for reference. I had the time to get a diff of keywords between 5.7.0 and 5.7.8 (according to the [documentation](http://dev.mysql.com/doc/refman/5.7/en/keywords.html):

*New reserved keywords*
- `GENERATED`
- `OPTIMIZER_COSTS`
- `STORED`
- `VIRTUAL`

*Removed reserved keywords*
- `NONBLOCKING`

The new keywords class can simply extend `MySQL57Keywords` class and then you can `array_merge` and `array_diff` the 5.7.0 keywords from parent.

ddctd143:
Any progress on this ?

lib/Doctrine/DBAL/Platforms/MySQL578Platform.php
ismailbaskin: https://github.com/ismailbaskin/dbal/commit/aa594405c4aa79f4237ebbb03a16901fcc12e114

lib/Doctrine/DBAL/Platforms/MySQL578Platform.php
ismailbaskin: https://github.com/doctrine/dbal/commit/fc86ebddfcce5e0bf1277fb82a630571b8902288
https://github.com/doctrine/dbal/commit/cd9d330f6259e6db1366bfcd9365917b41ac0ba2

sbusch:
We need this for barryvdh/laravel-ide-helper#295

Is there anything blocking this merge? If yes I could try to help out.

TNAJanssen:
+1

danielkeil:
+1

tmaneri:
+1

Ocramius:
+1-ing this doesn't change things, folks. If @deeky666 has no time, that's how it is.

tmaneri:
Pretty sure +1 is a github standard to dictate interest, I don't see anything wrong with people expressing that they are interested in this getting fixed.

Ocramius:
It just adds to an already infinite inbox. Github gave you like buttons exactly to avoid this: use them, please.

AymanAlMalki:
Not sure but I think that registering json type is still required to avoid 'Unknown database type json requested..' exception in same cases like generating migration files.

nicodmf:
+1

ddctd143:
@Ocramius  Package that has 17 million downloads and is so popular, but has problems to accept important pull request, that is passing unit tests. Everything is done by contributor, just accept, where is problem ? Or I don't understand something ? Is this project abandoned ?

nicodmf:
This seems ready to merge, do you need help to review or something else ?

basz:
> Not sure but I think that registering json type is still required to avoid 'Unknown database type json requested..' exception in some cases like generating migration files.

@AymanAlMalki FYI I have successfully tested migrations. I ALTERED columns without problems.

ismailbaskin:
@Ocramius Is possible to assign another member for this PR. @deeky666 's last activity in github at 23 february

guilhermeblanco:
Personally speaking, MySQL 5.7.8 is GA release. We shouldn't have a < 5.7.8 since there're not public releases (unless you compiled from source).

I'd suggest merge the MySQL578Platform into MySQL57Platform and we should be good to go.

Crell:
What is currently holding back this PR? Testing? More code? Just time?  How can we non-maintainers help move this forward?

beberlei:
The two MySQL platforms 5.7 need to be merged into one class, since everything before the GA version (5.7.8) is not considered stable, so we can ignore it.

Additionally this Pull Request can be considered a BC break, because it will lead the SchemaTool to generate a new column that is not 100% compatible with a `LONGTEXT`. `JsonArrayType` is not strict with data and invalids are allowed, but changing to `JSON` MySQL type leads to an error:

```
mysql> CREATE TABLE foo (id INT PRIMARY KEY, bar LONGTEXT);
Query OK, 0 rows affected (0,02 sec)

mysql> INSERT INTO foo (id, bar) VALUES (1, '{}'), (2, '[]'), (3, '{"foo": "bar"}');                                                                                                                        Query OK, 3 rows affected (0,00 sec)
Records: 3  Duplicates: 0  Warnings: 0

mysql> INSERT INTO foo (id, bar) VALUES (4, 'foo'), (5, 'invalid " json'), (6, '{]');
Query OK, 3 rows affected (0,00 sec)
Records: 3  Duplicates: 0  Warnings: 0

mysql> SELECT * FROM foo;
+----+----------------+
| id | bar            |
+----+----------------+
|  1 | {}             |
|  2 | []             |
|  3 | {"foo": "bar"} |
|  4 | foo            |
|  5 | invalid " json |
|  6 | {]             |
+----+----------------+
6 rows in set (0,00 sec)

mysql> ALTER TABLE foo MODIFY bar JSON;                                                                                                                                                                     ERROR 3140 (22032): Invalid JSON text: "Invalid value." at position 1 in value for column '#sql-504_5.bar'.
mysql> ALTER TABLE foo MODIFY bar JSON;                                                                                                                                                                     ERROR 3140 (22032): Invalid JSON text: "Invalid value." at position 1 in value for column '#sql-504_5.bar'.
```

thexpand:
When is this planed to be merged?

saschaglo:
+1

micln:
+1

Ocramius:
See concerns above. Stop +1-ing, please.

blackfyre:
The current version of the Oracle MySQL Community server is now 5.7.13, and is downloadable through their site and supports the JSON column type...

Sooner or later more and more people will come requesting support for it.

rh389:
In @ismailbaskin's absence I've submitted https://github.com/doctrine/dbal/issues/2455 which implements this directly into the MySQL57 platform as per suggestions above.

sameg14:
If you need to setup a custom JSON type for Doctrine in Symfony here is how you can achieve it

Ensure that `app/config/config.yml` has this line for doctrine
```yml
doctrine:
    dbal:
        driver:   pdo_mysql
        host:     "%database_host%"
        port:     "%database_port%"
        dbname:   "%database_name%"
        user:     "%database_user%"
        password: "%database_password%"
        charset:  UTF8
        server_version: 5.7
        mapping_types:
            json: string
        types:
            custom_json:
                class: CoreBundle\DBAL\Type\JsonType
                commented: true
```

Then ensure that you create the custom type as such
```php
<?php

namespace CoreBundle\DBAL\Type;

use Doctrine\DBAL\Types\JsonArrayType;

/**
 * Class JsonType
 * @package CoreBundle\DBAL
 */
class JsonType extends JsonArrayType
{
    /**
     * @param array $fieldDeclaration
     * @param AbstractPlatform $platform
     * @return string
     */
    {
        return 'json';
    }

    /**
     * Convert the value of the field into a PHP data type
     * @param mixed $value
     * @param AbstractPlatform $platform
     * @return array|mixed|null
     */
    {
        return $value === null ? null : parent::convertToPHPValue($value, $platform);
    }

    /**
     * Name of the db type
     * @return string
     */
    public function getName()
    {
        return 'json';
    }
}
```

Then your entity property should use the `custom_json` type to store data in mysql using json
```php
<?php

namespace CoreBundle\Entity;

use Doctrine\ORM\Mapping as ORM;

/**
 * @package CoreBundle\ExampleJsonEntity
 * @ORM\Table(name="example_json_table")
 * @ORM\Entity(repositoryClass="CoreBundle\Repository\ExampleJsonRepository")
 */
class ExampleJsonEntity
{
    /**
     * @var int
     * @ORM\Column(name="id", type="integer")
     * @ORM\Id
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    private $id;

    /**
     * @var string
     * @ORM\Column(name="json_data", type="custom_json", nullable=true)
     */
    private $jsonData;

    /**
     * @return int
     */
    public function getId()
    {
        return $this->id;
    }

    /**
     * @param int $id
     */
    public function setId($id)
    {
        $this->id = $id;
    }

    /**
     * @return string
     */
    public function getJsonData()
    {
        return $this->jsonData;
    }

    /**
     * @param string $jsonData
     */
    public function setJsonData($jsonData)
    {
        $this->jsonData = $jsonData;
    }
}
```

charmpitz:
Will this ever be merged? :)

dunglas:
@Ocramius @beberlei can we help in any way to get this merged? It would be nice to be able to use [Doctrine JSON ODM](https://github.com/dunglas/doctrine-json-odm) with MySQL too.

RicardoJ:
We also like to see this PR merged to enable support with it. If we can help in any way, please let me know.

StephenOTT:
+1

uniring:
+1!

mendicm:
+1000

Lydnasty:
+2

kimhemsoe:
There is still open [issues/concerns](https://github.com/doctrine/dbal/pull/2266#issuecomment-225466372) about these changes. 

Posting "+1" comments to does not help anyone. If you wish to express your wish for this feature there is +1 buttons for the same.

rh389:
@kimhemsoe - Are there any issues not addressed by https://github.com/doctrine/dbal/pull/2455 ? If so, please tell us what they are.

kimhemsoe:
@rh389 Yep still one issue. It could be considered a BC break as [mentioned](https://github.com/doctrine/dbal/pull/2266#issuecomment-241816915) by beberlei.

dkmuir:
@beberlei, I don't understand how it could be considered a BC break? Doctrine shouldn't be storing invalid json, and it can't load invalid json either. How is it storing invalid data for a datatype that Doctrine currently claims the database doesn't support? What am I missing?

RicardoJ:
@dkmuir, I'd also like to see this feature to be implemented but the BC break is as follow:

> Additionally this Pull Request can be considered a BC break, because it will lead the SchemaTool to generate a new column that is not 100% compatible with a LONGTEXT. JsonArrayType is not strict with data and invalids are allowed, but changing to JSON MySQL type leads to an error:

```sql
mysql> CREATE TABLE foo (id INT PRIMARY KEY, bar LONGTEXT);
Query OK, 0 rows affected (0,02 sec)

mysql> INSERT INTO foo (id, bar) VALUES (1, '{}'), (2, '[]'), (3, '{"foo": "bar"}');                                                                                                                        Query OK, 3 rows affected (0,00 sec)
Records: 3  Duplicates: 0  Warnings: 0

mysql> INSERT INTO foo (id, bar) VALUES (4, 'foo'), (5, 'invalid " json'), (6, '{]');
Query OK, 3 rows affected (0,00 sec)
Records: 3  Duplicates: 0  Warnings: 0

mysql> SELECT * FROM foo;
+----+----------------+
| id | bar            |
+----+----------------+
|  1 | {}             |
|  2 | []             |
|  3 | {"foo": "bar"} |
|  4 | foo            |
|  5 | invalid " json |
|  6 | {]             |
+----+----------------+
6 rows in set (0,00 sec)

mysql> ALTER TABLE foo MODIFY bar JSON;                                                                                                                                                                     ERROR 3140 (22032): Invalid JSON text: "Invalid value." at position 1 in value for column '#sql-504_5.bar'.
mysql> ALTER TABLE foo MODIFY bar JSON;       
```

[Source](https://github.com/doctrine/dbal/pull/2266#issuecomment-225466372)

dunglas:
Storing invalid JSON in a `json_array` column is not documented as a supported use case, is weird, requires to bypass DBAL to access to the DB directly and is probably the sign of an underlying app problem.

I would say that it's an acceptable change for a new minor version as soon as it is documented in the changelog.

A possible workaround to limit this BC break is to map MySQL's `json` type to the DBAL's `json` type instead of `json_array` because the `json` type already returns an error if the data from the database is invalid JSON.

basz:
Bumping the major version point to signify a bc seems an option, but i dont think doctrine follows semver. What is the forward then?

CawaKharkov:
:+1:

dkmuir:
@RicardoJ, @beberlei, JsonArrayType doesn't allow storing invalid json:

```php
    {
        if (null === $value) {
            return null;
        }

        return json_encode($value);
    }

    {
        if ($value === null || $value === '') {
            return array();
        }

        return json_decode($value, true);
    }
```

If the DB has invalid json, all you'll get back is `false`.
Same when saving the value. Invalid json will be saved as `false`. The only way you're going to end up with invalid json is by screwing with the data directly.

kimhemsoe:
@dkmuir 

> The only way you're going to end up with invalid json is by screwing with the data directly.
Which easily can happen if have more then one system using the database or even a dbal update statement in the in the doctrine orm application.

teohhanhui:
> Which easily can happen if have more then one system using the database or even a dbal update statement in the in the doctrine orm application.

That cannot and should not be covered by Doctrine ORM's BC promise. There are infinite ways to screw up your data anyway.

kimhemsoe:
It is a bit worse then just corrupt json. Without having tested it, but how do we handle empty strings values? 
To me it seems that json array type supports it, but mysql json does not. 

A simple test for this would be:
> SELECT CAST('' AS JSON);

I know an empty string is not valid json, but that does not change the fact that it seems that json array supports it.

teohhanhui:
As @guilhermeblanco and @beberlei have mentioned above, this should be merged into `MySQL57Platform`.

So shouldn't this be closed in favour of #2455?

teohhanhui:
Anyway, I still don't understand the BC problem. If you take a look at [`PostgreSQL92Platform`](https://github.com/doctrine/dbal/blob/ec8fc16604915110eea8046e3ea594c133aee880/lib/Doctrine/DBAL/Platforms/PostgreSQL92Platform.php) it does the exact same thing.

kimhemsoe:
PostgreSQL92Platform was [merged](https://github.com/doctrine/dbal/pull/469/files) before #720.

It could be considered a BC issue. Question is if it is and if yes, is it an acceptable BC?

> So shouldn't this be closed in favour of #2455?
I see no need to move discussing to another PR. The main issue is still the same.

StephenOTT:
Can we get a restatement of the remaining issue?

I seem to see a few issue being discussed:
1. `json_array` vs `json` data type usage
1. empty string values
1. invalid json handling

Am i missing any?

rh389:
A maintainer needs to make a decision here since there's nothing more the community can do to move this along. The so-called BC break is unavoidable if we're ever to use MySQL's native JSON support for Doctrine's `json` type (since it's inherently more strict) so this change is either fundamentally acceptable (perhaps with a version bump and/or BC note) or both PRs should be closed and we should give up on ever using native JSON as the `json` default. Do correct me if I've missed something.

Personally I agree with @teohhanhui and @dunglas - no one should be surprised that mapping arbitrary text to a `json` column is a bad idea even if it's currently permitted, it's surely not a supported use case. The [docs](http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html) for both `json` and (deprecated) `json_array` already state 

> Some vendors have a native JSON type and Doctrine will use it if possible

and 

> If you know that the data to be stored always is in a valid UTF-8 encoded JSON format string, you should consider using this type

Which clearly imply 1) that you'd be wrong to rely on the type for non-JSON values (empty strings included) and 2) that we positively should be mapping to the native type for >=5.7.

IMO it's not worth blocking the project from evolving for the sake of a very weird and marginal (ab)use-case. @beberlei is the only one to mention that this is "can be considered a BC break" and even he didn't sound so sure! Does anyone think it's a blocking problem?

rh389:
Urgh - this is problematic and I think we'll need some input from @beberlei. `Doctrine\DBAL\Schema\Column` has no knowledge of the database column type so 1) the `Doctrine\DBAL\Schema\Comparator` fails to notice that column needs modifying at all and 2) `MySqlPlatform::getAlterTableSQL` can't do anything clever to convert empty strings to valid JSON, since it doesn't know that the original type is text.

I'd propose adding `dbType` to `Column` and inspecting that in the `Comparator` alongside other changed properties, but I think that would merit a separate PR and I'd appreciate a maintainer's opinion on whether that's the best approach.

StephenOTT:
any feedback from @beberlei  on this?

fmonts:
+1. Any news?

rh389:
Nothing yet. I guess @beberlei isn't active on this project any more. Maybe @deeky666 or @Ocramius could have a look?

I think basically the outstanding issue is that we need to make changes to `Column` and `Comparator` to distinguish target SQL type from existing SQL type. Then we can use `getAlterTableSql` to deal with upgrading a text column which might contain empty strings to a JSON column that can't. That addresses the BC break (although I think the same BC break existed when postgres was given native JSON support and no one seemed to notice/care).

I'm happy to make a PR but it's a bit more work than this one and it'd be nice to think there's a chance of it being merged within six months!

blackfyre:
Well... this has been floating here more then a year now... it would be good to have some kind of resolution...

Ricknox:
Symfony is ready too for these changes. I agree with @blackfyre, lets merge this pull request.

StephenOTT:
So, is there someone working on merging this?  Or are there issues to resolve?

uniring:
+1

deeky666:
I will have a look into this as soon as I get time. There are currently some higher priority issues that have to be resolved first.

deeky666:
Aye, a lot of input here :D. First decision of the day: Closing this in favour of #2455 because as said before, this is part of the first GA release. Please let's continue discussions there. Thank you all for your input an patience, we will get this baby merged soon!

Ocramius:
Handled in #2653

PR #2267, opened 2015-12-18
Pass "path" to pdo-sqlite drivers from DriverManager instead of "dbname"

### LINKED ISSUE #1106: DBAL-1164: Creating SQLite Connections via a URL Does Not Work
Jira issue originally created by user chrisguitarguy:

When creating a SQL light connection put the URL path into the `dbname` param. But the SQLite driver doesn't using the `dbname` param: it uses the `path` parameter.

So doing this:

  'url' => 'sqlite:////some/path/here',
]);

Generates a PDO DSN like this: `sqlite:`. The path is completely ignored.

See the driver code here: https://github.com/doctrine/dbal/blob/6b6143ba16e5f17242835910173c033a8f73f845/lib/Doctrine/DBAL/Driver/PDOSqlite/Driver.php#L81-L88

`DriverManager` either needs some logic to use the path when it sees a sqlite URL or the Driver itself should use `dbname` (eg. check path, check dbname, check memory).

erikjwaxx:
This fixes https://github.com/doctrine/dbal/issues/1106. Any input on the correct way to do this per Doctrine coding standards is welcome.

### DIFF
--- a/lib/Doctrine/DBAL/DriverManager.php
+++ b/lib/Doctrine/DBAL/DriverManager.php
@@ -259,10 +259,14 @@ final class DriverManager
         }
         
         if (isset($url['path'])) {
+            $nameKey = isset($url['scheme']) && strpos($url['scheme'], 'sqlite') !== false
+                ? 'path' // pdo_sqlite driver uses 'path' instead of 'dbname' key
+                : 'dbname';
+
             if (!isset($url['scheme']) || (strpos($url['scheme'], 'sqlite') !== false && $url['path'] == ':memory:')) {
-                $params['dbname'] = $url['path']; // if the URL was just "sqlite::memory:", which parses to scheme and path only
+                $params[$nameKey] = $url['path']; // if the URL was just "sqlite::memory:", which parses to scheme and path only
             } else {
-                $params['dbname'] = substr($url['path'], 1); // strip the leading slash from the URL
+                $params[$nameKey] = substr($url['path'], 1); // strip the leading slash from the URL
             }
         }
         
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/DriverManager.php
```diff
@@ -259,10 +259,14 @@ private static function parseDatabaseUrl(array $params)
         }
         
         if (isset($url['path'])) {
+            $nameKey = isset($url['scheme']) && strpos($url['scheme'], 'sqlite') !== false
+                ? 'path' // pdo_sqlite driver uses 'path' instead of 'dbname' key
+                : 'dbname';
+
             if (!isset($url['scheme']) || (strpos($url['scheme'], 'sqlite') !== false && $url['path'] == ':memory:')) {
```
deeky666: Can we somehow try to avoid duplicate checks on whether the path is a sqlite path here?
Also I think it would be a good idea to consider setting `memory` parameter instead of `path` parameter, if the URL path refers to in-memory (it makes no difference for the driver but could be useful in userland when checking for specific parameters IMO).

Otherwise this patch looks good to me.

zeroedin-bill:
:+1:

deeky666:
Merged manually, added evaluation of `memory` connection parameter and refactored logic. @erikjwaxx thanks!

deeky666:
Backported to 2.5 branch via https://github.com/doctrine/dbal/commit/85e9508e9e650bc5dca2136d97ee97a077bb2d81, https://github.com/doctrine/dbal/commit/096f59dcba335f781857610f34f9ff33ac7dcfe7, https://github.com/doctrine/dbal/commit/b1cbe8c776cb1fdd0d1d3d0c27e679fd95f82f86

stof:
@deeky666 do you have an ETA for a 2.5.5 release including this fix ? This would allow https://github.com/symfony/symfony-demo/pull/297#issuecomment-172958354 to work.

deeky666:
@stof depends on review and merge capabilities of @Ocramius and @zeroedin-bill. Would like to get the open PRs targeted for `2.5.5` get merged first but it is not a MUST. We could also release earlier if desired.

laoneo:
We need it as well as soon as possible :smile:  Thanks.

PR #2273, opened 2016-01-05
Fix usage of PDO::PGSQL_ATTR_DISABLE_PREPARES for edge case pdo_pgsql setups

### LINKED ISSUE #2249: Compatibility mis-match? PGSQL_ATTR_DISABLE_PREPARES in v2.5.2 (in a PHP 5.6.16 envt)
Greetings,

I'm coming up empty on the following observation. I'm not sure if it's an issue or not, and perhaps someone here could advise.

I went ahead and installed doctrine/dbal 2.5.2 (via composer) into a project running atop of PHP 5.6.16, and I'm using pgsql as my backend database. I was previously running 2.5.1 on a PHP 5.3.* without issue. 

In connecting via dbal to the database, I am seeing this now:
` Fatal error: Undefined class constant 'PGSQL_ATTR_DISABLE_PREPARES' in /path-to-files/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php on line 49`

That block is shown:
```php
if (PHP_VERSION_ID >= 50600
                && (! isset($driverOptions[PDO::PGSQL_ATTR_DISABLE_PREPARES])
                    || true === $driverOptions[PDO::PGSQL_ATTR_DISABLE_PREPARES]
                )
            ) {
                $pdo->setAttribute(PDO::PGSQL_ATTR_DISABLE_PREPARES, true);
            }
```

If I comment out this entire block, the fatal error goes away and things *seem* to work normally. I'm not comfortable just commenting this out, since I don't know the full ramifications of such an action.

Any advice is much appreciated and I hope I'm not breaking any conventions by reporting this here. Thanks!

JP

deeky666:
The constant `PDO::PGSQL_ATTR_DISABLE_PREPARES` is available since PHP 5.6.0, however this is only true if the extension is installed/built from the PHP core sources. Looks like in rare situations it is possible to run PHP 5.6 with a heavily outdated `pdo_pgsql` extension (coming from PECL?!?). In this case DBAL crashes.
This PR makes the check for the constant's availability more robust to also work in those setups.

fixes #2249

### DIFF
--- a/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
@@ -45,7 +45,7 @@ class Driver extends AbstractPostgreSQLDriver
                 $driverOptions
             );
 
-            if (PHP_VERSION_ID >= 50600
+            if (defined('PDO::PGSQL_ATTR_DISABLE_PREPARES')
                 && (! isset($driverOptions[PDO::PGSQL_ATTR_DISABLE_PREPARES])
                     || true === $driverOptions[PDO::PGSQL_ATTR_DISABLE_PREPARES]
                 )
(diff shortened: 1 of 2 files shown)


Ocramius:
Backported into 2.5 via doctrine/dbal@2bb8c06644caa4b27e44f7eee663806a4033cf28

PR #2278, opened 2016-01-06
Allow PHP 7 failure on Travis-CI

deeky666:
Now that PHP7 is stable, we should allow failures again in Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,6 +98,13 @@ matrix:
         mariadb: 10.1
   allow_failures:
     - php: 7.0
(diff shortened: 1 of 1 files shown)


deeky666:
Again, the postgres errors are unrelated. This seems to be a problem with parallel Travis builds.

deeky666:
Okay allowed failures for postgres + PHP7 for now again until the issue is fixed.

Ocramius:
Meh @ the failures, but I guess we still can't fix everyone's problems :\

deeky666:
@Ocramius thanks. Still not sure whether this is related to PHP and/or Travis.

PR #2279, opened 2016-01-07
Stop using template1 as default database for postgres drivers

kimhemsoe:
The "postgresql" should be used as the default database and not "template1". Postgresql creates databases by copying an existing database and "template1" is the default if none is specified, if any connection to the template database exists when a "CREATE DATABASE" is issued postgresql is unable to create the new database.

http://www.postgresql.org/docs/9.4/static/manage-ag-templatedbs.html

### DIFF
--- a/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
@@ -90,9 +90,9 @@ class Driver extends AbstractPostgreSQLDriver
             $dsn .= 'dbname=' . $params['dbname'] . ' ';
         } else {
             // Used for temporary connections to allow operations like dropping the database currently connected to.
-            // Connecting without an explicit database does not work, therefore "template1" database is used
-            // as it is certainly present in every server setup.
-            $dsn .= 'dbname=template1' . ' ';
+            // Connecting without an explicit database does not work, therefore "postgres" database is used
+            // as it is certainly present in every server setup. 
+            $dsn .= 'dbname=postgres' . ' ';
         }
 
         if (isset($params['sslmode'])) {
--- a/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOPgSql/DriverTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOPgSql/DriverTest.php
@@ -33,6 +33,6 @@ class DriverTest extends AbstractDriverTest
      */
     protected function getDatabaseNameForConnectionWithoutDatabaseNameParameter()
     {
-        return 'template1';
+        return 'postgres';
     }
 }


kimhemsoe:
Ohh yeah.. and ... "fixes" #2278

deeky666:
@kimhemsoe IIRC the reason we decided against using `postgresql` as temporary database was that it is not guaranteed that the database exists in every database setup as you could theoretically delete it.

See https://github.com/doctrine/DoctrineBundle/issues/402#issuecomment-92557417

Ocramius:
Can we somehow query to check whether the schema is there?

deeky666:
@Ocramius I doubt this helps a lot. If the schema isn't there we still can't establish a temp connection. Also relying on a "default" database that has no special purpose and can be dopped by anyone. It's basically like a demo database to start with.
I might have found another solution, will check this no and report back.

kimhemsoe:
You can drop template1 too, postgres database is the "recommended" database to use as default af applications (Read the bottom note in the link), using template1 is not an option for various  reasons. There is numerous stackoverflow explaining this. 

I do not think you can drop template0 but that is not a good idea either for the same as template1.

I do strongly believe that using he postgres database is the lesser evil of the 3 options that we have.
If people have dropped postgres they can either recreate it or specify a schema.

"Unknown database "postgres" is alot better error message then random ""SQLSTATE[55006]: Object in use: 7 ERROR:  source database "template1" is being accessed by other users"" which may not get throwed from dbal but from other random applications using the pg instance because we are using template1 as the default schema.

deeky666:
Closing in favour of #2280

deeky666:
@kimhemsoe in a default database setup you cannot drop template databases as they are protected via the `` configuration directive.

```
SQLSTATE[42809]: Wrong object type: 7 ERROR:  cannot drop a template database
```

See #2280 for a complete explanation. Also `postgresql` database is the default database that is owned by the `postgresql` user and it is very unreliable that user and thus the database is present.

deeky666:
Reopening as the discussion in #2280 revealed this might be the better approach.

deeky666:
@kimhemsoe thanks :)

deeky666:
Backported to 2.5 branch in af38e30ab9c5828b20312d796651d192020d18c1

kimhemsoe:
We better provide the driver param fast to catch the issue flooding from doctrine bundle.

deeky666:
@kimhemsoe do you think that many people are not having the postgres database? Otherwise yeah go ahead do a PR :)

kimhemsoe:
The template1 change came from a few doctrine bundle issues, i will provide PR's for the parameter.

https://github.com/doctrine/DoctrineBundle/issues/461
https://github.com/doctrine/DoctrineBundle/issues/351

Btw i think we should merge #2280 too, not the template1 active connection reason, but to make sure we always have clean a database for tests, without any local changes.

deeky666:
@kimhemsoe I don't see how your referenced PRs are related to this issue here. Those PRs deal with a completely different problem.
About merging #2280: I don't think we should do that. When running the tests it is not Doctrine's task to ensure that the environment is set up correctly. Travis always spawns a fresh default environment so we won't have any problems there. Is there any other reason why we should use `template0`? I think `template1` might be better because then environments that want to have default objects in every database can still make use of a modified `template1` database.

PR #2280, opened 2016-01-07
Fix creating database on PostgreSQL with active connections on "template1" database

deeky666:
Fixes:

```
Exception: [Doctrine\DBAL\Exception\DriverException] An exception occurred while executing 'CREATE DATABASE test_drop_database':

SQLSTATE[55006]: Object in use: 7 ERROR:  source database "template1" is being accessed by other users

DETAIL:  There are 19 other sessions using the database.
```

The problem here is that if we want to create a database in PostgreSQL, we temporarily connect to `template1` database in order to be able to create a database safely. Now if there are other active connections on `template1` already, PostgreSQL refuses to create the database as it internally just uses `template1` as a template for the database to be created.
To come around this problem we can use `template0` as the template database when creating database. `template0` doesn't accept connections in a default server setup, so we know this won't issue any errors. If I understood correctly there is no noticeable difference between `template0` and `template1` so it doesn't matter for us which template to use for creating new databases. By default when omitting the `TEMPLATE <template_database>` expression in the `CREATE DATABASE` statement, PostgreSQL uses `template1`.
There is a pretty good explanation of this in the [documentation](http://www.postgresql.org/docs/9.1/static/manage-ag-templatedbs.html).

To sum up in a default database setup we can be sure to have the following environment:
- `template0` database which is protected (cannot be dropped) and **does not** accept connections
- `template1` database which is protected (cannot be dropped) and **does** accept connections
- `postgres` database which is **not** protected (can be dropped)

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -417,7 +417,7 @@ class PostgreSqlPlatform extends AbstractPlatform
      */
     public function getCreateDatabaseSQL($name)
     {
-        return 'CREATE DATABASE ' . $name;
+        return 'CREATE DATABASE ' . $name . ' TEMPLATE template0';
     }
 
     /**
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
@@ -130,7 +130,7 @@ abstract class AbstractPostgreSqlPlatformTestCase extends AbstractPlatformTestCa
 
     public function testGeneratesDDLSnippets()
     {
-        $this->assertEquals('CREATE DATABASE foobar', $this->_platform->getCreateDatabaseSQL('foobar'));
+        $this->assertEquals('CREATE DATABASE foobar TEMPLATE template0', $this->_platform->getCreateDatabaseSQL('foobar'));
         $this->assertEquals('DROP DATABASE foobar', $this->_platform->getDropDatabaseSQL('foobar'));
         $this->assertEquals('DROP TABLE foobar', $this->_platform->getDropTableSQL('foobar'));
     }


deeky666:
@kimhemsoe I know that using the template databases is also not 100% safe BUT it is IMO the best we can get when dealing with default database setups. We can't and won't support funky custom setups anyways so this is the environment we have to expect.

kimhemsoe:
@deeky666 I do not agree.

To sum up.
template1 can be dropped as explained in he documentation if one really want to and other users can not create databases using template1 as the template. And if we really screw up we may end creating stuff in there.

Using "template0", result in that other users can not create databases using template0 as the template while we are connected.

Using "postgres" we end up annoying a few people who have dropped postgres. It seems that most other drivers out there is using postgres as the default as recommend by the documentation.

deeky666:
@kimhemsoe I think there are still some misunderstandings on your side.

1. By default the template databases are protected against deletion:

```php
$schemaManager->dropDatabase('template1'); // PDOException: SQLSTATE[42809]: Wrong object type: 7 ERROR:  cannot drop a template database
```

2. We don't use `template0` for temporary connection as by default PostgreSQL refuses connection to `template0`:

```php
$params = $connection->getParams();
$params['dbname'] = 'template0';
$user = isset($params['user']) ? $params['user'] : null;
$password = isset($params['password']) ? $params['password'] : null;

$connection->getDriver()->connect($params, $user, $password); // Doctrine\DBAL\Exception\ConnectionException: An exception occurred in driver: SQLSTATE[08006] [7] FATAL:  database "template0" is not currently accepting connections
```

3. Connectiong to `postgres` does not work if the database was dropped (as you pointed out). So what is the advantage using this over the other option then? I don't get it. Just saying "because everyone else is doing it like this" is no argument ;)

kimhemsoe:
@deeky666 
You can drop template1 if want too (http://stackoverflow.com/questions/11388786/how-does-one-drop-a-template-database-from-postgresql)

template0 is the clean slate schema and other users may be using it that reason.

postgres is better option because:
  It does not interfere with other applications.
  It is **recommend** schema by the docummentation.
  It is the default in every other driver. My argument here is that, if every other out there including the team postgres behind is using postgres as the default. It is properly not the best idea for to be the odd child in the class.

For migrations and etc one can change to use whatever they have available and if people choose to any of the template database we did not hand them the gun or load it.

kimhemsoe:
Sorry my mistake with template0, just ignore it.

deeky666:
@kimhemsoe sorry but I am not sure if I completely get you. What do you mean by `It does not interfere with other applications.`? I see that `postgres` is *recommended* but a recommendation is not a *must*. Our concern here is not to go all the way standard as the vendor suggests but to enforce safety and robustness in the application as much as possible. We have a lot of things in DBAL that are not *standart* because we need that for good abstraction. If using a (by default) protected database like `template1` as temporary connection for creating new databases gives us better reliability than a non-protected default database (where the potential of being unavailable is greater than using a database that cannot be dropped without further actions taken), I would always prefer that (as long as there are no other decent side-effect, which I can't see here).

In the end I am just looking at the pros and cons for evaluating which option might be better (knowing that both are not 100% safe). When comparing both I can't see where `postgres` has better arguments than `template1`. Using `postgres` just for the sake of going the *recommended* *standard* way is no argument for me TBH. But please clear me up if I am missing something here :)

Maybe a third opinion by @Ocramius might help, too.

kimhemsoe:
@deeky666 Postgresql can not use a schema as a template if any connection to that template exists. We will be interfering with other applications if any of these try to create a database while we are connected to template1. 
The issue i have with this is that the error we get is not throw from the offender (us) but from the other application 

I believe is this one of those cases where it is better to document it and maybe even add a driver parameter to override the default database.

deeky666:
@kimhemsoe ah now I get what you mean. Yeah that *would* indeed be a problem. That is more or less convincing me to go for `postgres` then. Besides that the idea of having a driver option for "default database" as a last resort for those who don't have `postgres` was coming to my mind, too. We should deal with that in a seperate PR though.

Sorry if I created confusion here ;)

Closing then and reopening your PR.

PR #2287, opened 2016-01-09
Fix parsing schemeless connection URLs

### LINKED ISSUE #1183: DBAL-1234: Additional slash in dbname when providing settings as URL without scheme
Jira issue originally created by user KingCrunch:

Hi,

I use https://github.com/realestateconz/MssqlBundle to connect to an MSSQL-database and I'd like to provide the connection parameters as URL. Because `dblib` is not a supported driver I setup the `driverClass` instead

```
driver_class:   \Realestate\MssqlBundle\Driver\PDODblib\Driver
```

So the corresponding URL would look like

```
//user:pass@127.0.0.1/dabasename
```

But now it tries to connect to the database `/databasename` instead of `databasename`. I can set an arbitrary scheme here as long as it exists and is supported (and is not SQLite) 

```
mysqli://user:pass@127.0.0.1/dabasename
```

Now it works, but it's a hack.

It seems, that the issue is here
https://github.com/doctrine/dbal/blob/32b1a4f85a078f67752851c27be4065071db1f8b/lib/Doctrine/DBAL/DriverManager.php#L262
As long as there is no scheme the leading slash remains. I'd guess, that it should also take into account, that there might be no driver name, but a concrete `driverClass` instead

```
(!isset($url['scheme']) && !isset(isset($url['driverClass']))
```

?

deeky666:
Taken over from and inspired by #863. Fixes #1183.

Also includes proper handling of the default driver connection parameters `pdo`, `driver` and `driverClass` when used together with connection URLs.
Throws exception now if URL is given but neither URL scheme nor `driver` parameter nor `driverClass` parameter is given. If also given a default PDO instance, DBAL would currently use that one instead and silently discard connection URL information.

Finally I have done a minor refactoring of the connection URL parsing by simplifying the logic and splitting into smaller private methods. It is much more comprehensible now IMO. Also added a lot of comments to keep track of all the weird things that happen there :D

### DIFF
--- a/lib/Doctrine/DBAL/DBALException.php
+++ b/lib/Doctrine/DBAL/DBALException.php
@@ -77,10 +77,22 @@ class DBALException extends \Exception
     }
 
     /**
+     * @param bool|null $url The URL that was provided in the connection parameters (if any).
+     *
      * @return \Doctrine\DBAL\DBALException
      */
-    public static function driverRequired()
+    public static function driverRequired($url = null)
     {
+        if ($url) {
+            return new self(
+                sprintf(
+                    "The options 'driver' or 'driverClass' are mandatory if a connection URL without scheme " .
+                    "is given to DriverManager::getConnection(). Given URL: %s",
+                    $url
+                )
+            );
+        }
+
         return new self("The options 'driver' or 'driverClass' are mandatory if no PDO ".
             "instance is given to DriverManager::getConnection().");
     }
--- a/lib/Doctrine/DBAL/DriverManager.php
+++ b/lib/Doctrine/DBAL/DriverManager.php
@@ -245,7 +245,15 @@ final class DriverManager
         // (pdo_)?sqlite3?:///... => (pdo_)?sqlite3?://localhost/... or else the URL will be invalid
         $url = preg_replace('#^((?:pdo_)?sqlite3?):///#', '$1://localhost/', $params['url']);
 
-        $url = parse_url($url);
+        // PHP < 5.4.8 doesn't parse schemeless urls properly.
+        // See: https://php.net/parse-url#refsect1-function.parse-url-changelog
+        if (PHP_VERSION_ID < 50408 && strpos($url, '//') === 0) {
+            $url = parse_url('fake:' . $url);
+
+            unset($url['scheme']);
+        } else {
+            $url = parse_url($url);
+        }
 
         if ($url === false) {
             throw new DBALException('Malformed parameter "url".');
--- a/lib/Doctrine/DBAL/DriverManager.php
+++ b/lib/Doctrine/DBAL/DriverManager.php
@@ -213,6 +213,19 @@ final class DriverManager
     }
 
     /**
+     * Normalizes the given connection URL path.
+     *
+     * @param string $urlPath
+     *
+     * @return string The normalized connection URL path
+     */
+    private static function normalizeDatabaseUrlPath($urlPath)
+    {
+        // Trim leading slash from URL path.
+        return substr($urlPath, 1);
+    }
+
+    /**
      * Extracts parts from a database URL, if present, and returns an
      * updated list of parameters.
      *
@@ -238,12 +251,11 @@ final class DriverManager
             throw new DBALException('Malformed parameter "url".');
         }
 
-        if (isset($url['scheme'])) {
-            $params['driver'] = str_replace('-', '_', $url['scheme']); // URL schemes must not contain underscores, but dashes are ok
-            if (isset(self::$driverSchemeAliases[$params['driver']])) {
-                $params['driver'] = self::$driverSchemeAliases[$params['driver']]; // use alias like "postgres", else we just let checkParams decide later if the driver exists (for literal "pdo-pgsql" etc)
-            }
-        }
+        // If we have a connection URL, we have to unset the default PDO instance connection parameter (if any)
+        // as we cannot merge connection details from the URL into the PDO instance (URL takes precedence).
+        unset($params['pdo']);
+
+        $params = self::parseDatabaseUrlScheme($url, $params);
 
         if (isset($url['host'])) {
             $params['host'] = $url['host'];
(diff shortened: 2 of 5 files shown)


deeky666:
Fixed mocking zeh PDO

lib/Doctrine/DBAL/DBALException.php
```diff
@@ -77,10 +77,22 @@ public static function invalidPdoInstance()
...
     {
+        if ($url) {
+            return new self(
+                sprintf(
+                    "The options 'driver' or 'driverClass' are mandatory if a connection URL without scheme " .
+                    "is given to DriverManager::getConnection(). Given URL: %s",
+                    $url
+                )
+            );
+        }
```
KingCrunch: I'd put this into a separate static method. To me only looking at this code it is not clear, that when `$url` is given here, it misses one special part. Maybe `missingSchemeRequiredDriver()` or something like that?

lib/Doctrine/DBAL/DriverManager.php
```diff
@@ -258,53 +278,97 @@ private static function parseDatabaseUrl(array $params)
...
     {
-        if (!isset($url['scheme'])) {
-            $params['dbname'] = $url['path'];
-
+        if (! isset($url['path'])) {
             return $params;
         }
 
-        $url['path'] = substr($url['path'], 1);
+        $url['path'] = self::normalizeDatabaseUrlPath($url['path']);
```
KingCrunch: It is only used here. It doesn't look like a benefit to extract this `substr()` call

tests/Doctrine/Tests/DBAL/DBALExceptionTest.php
```diff
@@ -21,4 +21,20 @@ public function testAvoidOverWrappingOnDriverException()
...
+        $url = 'mysql://localhost';
+        $exception = DBALException::driverRequired($url);
+
+        $this->assertInstanceOf('Doctrine\DBAL\DBALException', $exception);
+        $this->assertSame(
+            sprintf(
+                "The options 'driver' or 'driverClass' are mandatory if a connection URL without scheme " .
+                "is given to DriverManager::getConnection(). Given URL: %s",
+                $url
+            ),
```
KingCrunch: This could be just a string. Test should contain as less logic as possible :)

    "The options 'driver' or 'driverClass' are mandatory if a connection URL without scheme is given to DriverManager::getConnection(). Given URL: mysql://localhost"

lib/Doctrine/DBAL/DriverManager.php
```diff
@@ -232,18 +245,25 @@ private static function parseDatabaseUrl(array $params)
...
 
-        if (isset($url['scheme'])) {
-            $params['driver'] = str_replace('-', '_', $url['scheme']); // URL schemes must not contain underscores, but dashes are ok
-            if (isset(self::$driverSchemeAliases[$params['driver']])) {
-                $params['driver'] = self::$driverSchemeAliases[$params['driver']]; // use alias like "postgres", else we just let checkParams decide later if the driver exists (for literal "pdo-pgsql" etc)
-            }
-        }
+        // If we have a connection URL, we have to unset the default PDO instance connection parameter (if any)
+        // as we cannot merge connection details from the URL into the PDO instance (URL takes precedence).
+        unset($params['pdo']);
```
KingCrunch: My 2 cent: When I pass a concrete connection instance, I'd expect, that this instance is used. So either the   connection url should be dropped, instead of the connection instance, or an error should occur. Else a new connection will be made, even if I already pass one. That may be unexpected. What do you think?

KingCrunch:
Since this was my approach originally I left some comments :) Still somebody with a better feeling for the big picture should have a look.

lib/Doctrine/DBAL/DriverManager.php
deeky666: @KingCrunch yeah I get what you mean. TBH it also took me some time to gather what actually the expected behaviour is/should be. The thing is (and after thinking a little bit about it also makes a lot of sense) the documentation propagates that if a URL is specified, it's information takes precedence over "default" value being set already in the connection parameters. This makes sense because as a service provider you could for example provide connection details for a default connection while the client can fully or partially ovewrite those settings. I think this was also part of the idea of your original PR. Telling that, a "default" PDO makes no differences in semantics towards providing a `driver` or `driverClass` because all of those end up in a connection instance using default values. So IMO making an exception for PDO here is rather confusing and defeats the original message of an URL taking precedence. Also silently discarding a connection URL may be dangerous to the client as he might expect to be connected to somewhere else than the default PDO connection. So IMO when providing a URL we have to error as soon as we have connection parameter conflicts.

tests/Doctrine/Tests/DBAL/DBALExceptionTest.php
deeky666: Arguable ;) But as we didn't practice something like that in the testsuite I'll not start here. Also the test case is quite simple so it should be understandable enough.

lib/Doctrine/DBAL/DriverManager.php
deeky666: True that. I don't have a strong opinion about that, my preference was to make things clear by defining each step that is part of the parsing process as individual method.

lib/Doctrine/DBAL/DBALException.php
deeky666: Also no strong opinion here. @Ocramius @zeroedin-bill thoughts?

lib/Doctrine/DBAL/DriverManager.php
dzuelke: Yeah so since I originally wrote this URL stuff... the idea is definitely that a URL can override any individual defaults that are set. The reason is that were it the other way round, you'd have to repeat all individual args to unset the URL instructions. Also, the idea is kind of to have a few defaults for e.g. all dev envs set individually through a framework or whatever, and then a URL can "fill" or "override" those. Think charset, server version and SSL mode set "globally" through individual parameters, and then each developer/environment just has a single `mysql://user:pass@host:port/dbname` URL parameter set, but they could always attach `?sslmode=none` to that URL for testing to override the default.

lib/Doctrine/DBAL/DBALException.php
Ocramius: Would have done a separate method as well, yep.

Ocramius: Errata: seen now that `'url'` may be passed in from multiple contexts. It is indeed better to let the exception class decide here.

Ocramius:
Merged, thanks @deeky666 and @KingCrunch!

`master`: doctrine/dbal@4a85a587d0fcd6f7c46c4185dd65fb13c042a834
`2.5`: doctrine/dbal@dcce6f8637a80c536791a640d841858b6ff627db

PR #2288, opened 2016-01-09
Preserve quotation of old column name in ColumnDiff

### LINKED ISSUE #1206: DBAL-1255: [GH-875] Generating a ColumnDiff led to unquoted Columns for Postgres
Jira issue originally created by user @doctrinebot:

This issue is created automatically through a Github pull request on behalf of tk-s:

Url: https://github.com/doctrine/dbal/pull/875

Message:

Postgres needs (or understands) quoted column-names all the Time, especially if one is using mix cased column names.
While generating Diffs, the old column name got no quotation.

deeky666:
fixes #1206

Takes over #875 with a correct solution.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Identifier.php
+++ b/lib/Doctrine/DBAL/Schema/Identifier.php
@@ -35,9 +35,14 @@ class Identifier extends AbstractAsset
      * Constructor.
      *
      * @param string $identifier Identifier name to wrap.
+     * @param bool   $quote      Whether to force quoting the given identifier.
      */
-    public function __construct($identifier)
+    public function __construct($identifier, $quote = false)
     {
         $this->_setName($identifier);
+
+        if ($quote && ! $this->_quoted) {
+            $this->_setName('"' . $this->getName() . '"');
+        }
     }
 }
--- a/lib/Doctrine/DBAL/Schema/ColumnDiff.php
+++ b/lib/Doctrine/DBAL/Schema/ColumnDiff.php
@@ -77,6 +77,8 @@ class ColumnDiff
      */
     public function getOldColumnName()
     {
-        return new Identifier($this->oldColumnName);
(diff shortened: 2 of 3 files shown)


lib/Doctrine/DBAL/Schema/Identifier.php
```diff
@@ -35,9 +35,14 @@ class Identifier extends AbstractAsset
      * Constructor.
      *
      * @param string $identifier Identifier name to wrap.
+     * @param bool   $quote      Whether to force quoting the given identifier.
      */
-    public function __construct($identifier)
+    public function __construct($identifier, $quote = false)
```
zeroedin-bill: YASSSSSSSSSS

lib/Doctrine/DBAL/Schema/Identifier.php
```diff
@@ -35,9 +35,14 @@ class Identifier extends AbstractAsset
...
     {
         $this->_setName($identifier);
+
+        if ($quote && ! $this->_quoted) {
+            $this->_setName('"' . $this->getName() . '"');
```
zeroedin-bill: The problem is always that if a name existing in the db contains an unmatched quote character in it, the item essentially becomes untouchable and breaks DBAL's schema system completely. I don't care that much if people have broken crazy table names in their databases, but the way AbstractAsset::trimQuotes works is incorrect in an academic sense.

I think in dbal 3 we should probably just quote everything and call it a day.

deeky666: @zeroedin-bill yes I am completely aware that quotation identification is much too simplistic and fragile. Just didn't want to change behaviour here, only fixing the bug for now. We will fix that in `3.0`

deeky666:
Backport to `2.5.` branch via https://github.com/doctrine/dbal/commit/9567eaaa084c155710c56316f3fc4dc140d64691

PR #2292, opened 2016-01-11
Do not generate FROM clause in QueryBuilder if no tables specified

### LINKED ISSUE #2291: The QueryBuilder::getSQLForSelect() always appends a FROM clause for select queries
Running queries without `FROM` clause is not possible currently with the `Doctrine\DBAL\Query\QueryBuilder`.

For instance we have these queries: 

`SELECT EXISTS(...)`
`SELECT some_function(1, 1)`

We would probably write something like this: 
`$qb->select('EXISTS (...)')`
And, `$qb->select('my_function(1, 1)')`

But, when we do `$qb->execute()` it would fail because, internally the `FROM` clause is being appended to the SQL every time we use `QueryBuilder::select()`. 

Here's, the part of `QueryBuilder::getSQLForSelect()` that's responsible for this. 
```php
    /**
     * @return string
     *
     * @throws \Doctrine\DBAL\Query\QueryException
     */
    private function getSQLForSelect()
    {
        $query = 'SELECT ' . implode(', ', $this->sqlParts['select']) . ' FROM ';

        $query .= implode(', ', $this->getFromClauses())
```
https://github.com/doctrine/dbal/blob/a3a86aa50ebfe180ac5943cce90a40262041bbdf/lib/Doctrine/DBAL/Query/QueryBuilder.php#L1108-L1111

I have come across a lot of scenarios, where we need to run `SELECT` without `FROM`. But, it's not possible right now. 

Eg: If I want to check if the user exists in the database, I would want to do: 
`SELECT EXISTS(SELECT 1 FROM user WHERE id = :userId)` as this is the fastest way

But, I have been using this type of query as a work around
`SELECT COUNT(id) FROM user WHERE id = :userId` which is comparatively slower than the first one. 
http://stackoverflow.com/questions/7471625/fastest-check-if-row-exists-in-postgresql#answer-16467634

deeky666:
fixes #2291 

`SELECT` statements without `FROM` clause can be valid and legit in some scenarios. See issue.

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -1107,9 +1107,9 @@ class QueryBuilder
      */
     private function getSQLForSelect()
     {
-        $query = 'SELECT ' . implode(', ', $this->sqlParts['select']) . ' FROM ';
+        $query = 'SELECT ' . implode(', ', $this->sqlParts['select']);
 
-        $query .= implode(', ', $this->getFromClauses())
+        $query .= ($this->sqlParts['from'] ? ' FROM ' . implode(', ', $this->getFromClauses()) : '')
             . ($this->sqlParts['where'] !== null ? ' WHERE ' . ((string) $this->sqlParts['where']) : '')
             . ($this->sqlParts['groupBy'] ? ' GROUP BY ' . implode(', ', $this->sqlParts['groupBy']) : '')
             . ($this->sqlParts['having'] !== null ? ' HAVING ' . ((string) $this->sqlParts['having']) : '')
--- a/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
+++ b/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
@@ -23,6 +23,18 @@ class QueryBuilderTest extends \Doctrine\Tests\DbalTestCase
                    ->will($this->returnValue($expressionBuilder));
     }
 
+    /**
+     * @group DBAL-2291
+     */
+    public function testSimpleSelectWithoutFrom()
+    {
+        $qb = new QueryBuilder($this->conn);
+
+        $qb->select('some_function()');
+
+        $this->assertEquals('SELECT some_function()', (string) $qb);
+    }
+
     public function testSimpleSelect()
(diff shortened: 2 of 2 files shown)


Ocramius:
@deeky666 :+1: 

But doesn't this count as improvement?

deeky666:
@Ocramius not sure. I'm fine having it as improvement. Guess it depends on what was to expect before, whether the `QueryBuilder` was intended to work without `FROM` specified. I guess nobody can answer that. But currently it just creates borked SQL if `FROM` is not specified, even though this can be valid SQL. We are doing conditions on other clauses like `WHERE` and `ORDER BY`, too so it's basically the same thing IMO. If we didn't generate those clauses conditonally, the resulting SQL would be borked just as it is now with `FROM`. I'll let you decide :)

Ocramius:
@deeky666 ok, then I'll take it as bugfix. Restoring milestone

deeky666:
@Ocramius fine. You can merge then as soon as you want.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1107,9 +1107,9 @@ public function resetQueryPart($queryPartName)
      */
     private function getSQLForSelect()
     {
-        $query = 'SELECT ' . implode(', ', $this->sqlParts['select']) . ' FROM ';
+        $query = 'SELECT ' . implode(', ', $this->sqlParts['select']);
 
-        $query .= implode(', ', $this->getFromClauses())
+        $query .= ($this->sqlParts['from'] ? ' FROM ' . implode(', ', $this->getFromClauses()) : '')
```
zeroedin-bill: Consider that on Oracle you can't select without FROM. Needs select from dual in that case.

deeky666: Yeah it is neither meant to be platform aware nor is it meant to be able to validate SQL..

zeroedin-bill: Good call

deeky666:
@Ocramius poke

Ocramius:
Backported to `2.5` via doctrine/dbal@e69d1cf3bddd269e76d68eb81823c01f02b230d0

PR #2293, opened 2016-01-11
Fixed some broken links in the types documentation

jeanCarloMachado:
I tested on the html and pdf versions of the documentation and now its working. I may do the same for the ORM in the near future if you like.


deeky666:
@jeanCarloMachado thanks! It would be very welcome if you want to fix links in ORM, too :) Also will backport this, as I find it rather important that the links in the docs work.

deeky666:
Hmmm looks like we do not generate docs for other branches than master. So no backport then.

PR #2296, opened 2016-01-12
Fix parameter list expansion by making SQLParserUtils platform aware

### LINKED ISSUE #2295: SQLParserUtils treats brackets in Postgres like quoting characters
After this commit (https://github.com/doctrine/dbal/commit/646f479a6a816044e2116e17416113b33de380e3#diff-e7dd540cfb86c74ce7361e2bd00053cf
  added because of SQL Server) SQLParserUtils::getUnquotedStatementFragments() treats brackets like quoting character. This is OFC fine in case of SQL Server. But Postgres does not use brackets for strings or escaping but for example as an array constructor like `ARRAY[20000, 25000, 25000, 25000]`.

Let's say I have an SQL like this: `SELECT * FROM a WHERE jsonb_exists_any(a.b, ARRAY[:parameter]) = false` and i want to set an array parameter. Then SQLParserUtils treats `[?]` as a quoted statement and does not expand the array parameter like it does in other cases,.

This class cannot be universal as the behavior should depend on the platform. So maybe the platform should be the place where this is set.

Another, quick way is to change the regular expression to not treat `[...]` like `ARRAY[...]`... That would still be probably better than it's now.

deeky666:
fixes #2295

@Ocramius @zeroedin-bill Ideas about how to test this properly without having to completely rewrite `Doctrine\Tests\DBAL\SQLParserUtilsTest` are welcome. Also I don't see a reasonable way how to test the changes in the connection as `SQLParserUtils` is called statically.

@maryo also please test if that fixes your problem if you have time to.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -812,7 +812,12 @@ class Connection implements DriverConnection
 
         try {
             if ($params) {
-                list($query, $params, $types) = SQLParserUtils::expandListParameters($query, $params, $types);
+                list($query, $params, $types) = SQLParserUtils::expandListParameters(
+                    $query,
+                    $params,
+                    $types,
+                    $this->getDatabasePlatform()
+                );
 
                 $stmt = $this->_conn->prepare($query);
                 if ($types) {
@@ -971,7 +976,12 @@ class Connection implements DriverConnection
 
         try {
             if ($params) {
-                list($query, $params, $types) = SQLParserUtils::expandListParameters($query, $params, $types);
+                list($query, $params, $types) = SQLParserUtils::expandListParameters(
+                    $query,
+                    $params,
+                    $types,
+                    $this->getDatabasePlatform()
+                );
 
                 $stmt = $this->_conn->prepare($query);
                 if ($types) {
--- a/lib/Doctrine/DBAL/SQLParserUtils.php
+++ b/lib/Doctrine/DBAL/SQLParserUtils.php
@@ -19,6 +19,9 @@
 
 namespace Doctrine\DBAL;
 
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Platforms\SQLServerPlatform;
+
 /**
  * Utility class that parses sql statements with regard to types and parameters.
  *
@@ -43,13 +46,17 @@ class SQLParserUtils
      * Returns an integer => integer pair (indexed from zero) for a positional statement
      * and a string => int[] pair for a named statement.
      *
-     * @param string  $statement
-     * @param boolean $isPositional
+     * @param string                $statement
+     * @param boolean               $isPositional
+     * @param AbstractPlatform|null $platform     The platform to use for retrieving the placeholder positions.
      *
      * @return array
      */
-    static public function getPlaceholderPositions($statement, $isPositional = true)
-    {
+    static public function getPlaceholderPositions(
+        $statement,
+        $isPositional = true,
+        AbstractPlatform $platform = null
+    ) {
         $match = ($isPositional) ? '?' : ':';
         if (strpos($statement, $match) === false) {
             return array();
@@ -58,7 +65,7 @@ class SQLParserUtils
         $token = ($isPositional) ? self::POSITIONAL_TOKEN : self::NAMED_TOKEN;
         $paramMap = array();
 
-        foreach (self::getUnquotedStatementFragments($statement) as $fragment) {
+        foreach (self::getUnquotedStatementFragments($statement, $platform) as $fragment) {
             preg_match_all("/$token/", $fragment[0], $matches, PREG_OFFSET_CAPTURE);
             foreach ($matches[0] as $placeholder) {
                 if ($isPositional) {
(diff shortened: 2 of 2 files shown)


maryo:
That's really great you are working on it! :+1: 
I made a quick workarround to satisfy the issue i described. But OFC your solution is the solution.

https://github.com/vaniocz/dbal/commit/0dcfbbab9b232a138b477a3a7d24b743cc869937

At least i added very simple test scenario.

deeky666:
@maryo thanks for investigation on this. Theoretically I need to test everything that already is present with every available platform. This would be quite some work to achieve so I am first asking for ideas of a simpler solution before doing the effort :)

lib/Doctrine/DBAL/SQLParserUtils.php
```diff
@@ -196,16 +204,46 @@ static public function expandListParameters($query, $params, $types)
...
-                   self::ESCAPED_BACKTICK_QUOTED_TEXT . '|' .
-                   self::ESCAPED_BRACKET_QUOTED_TEXT;
-        preg_match_all("/([^'\"`\[]+)(?:$literal)?/s", $statement, $fragments, PREG_OFFSET_CAPTURE);
+        $quoteCharacterClass = "'";
+        $literal = self::ESCAPED_SINGLE_QUOTED_TEXT . '|';
+        $identifierQuoteCharacter = null !== $platform ? $platform->getIdentifierQuoteCharacter() : null;
+
+        switch ($identifierQuoteCharacter) {
+            case '"':
+                if ($platform instanceof SQLServerPlatform) {
```
deeky666: Btw another misconception in the platform. SQL Server has different quote characters for start and end, therefore the platform returns `"` here as a workaround :>

maryo: Looks like `SQLServerPlatform` does not override `getIdentifierQuoteCharacter` so the `AbstractPlatform` implementation is used which is `return '"'`. Does it really makes sense for `SQLServerPlatform`? Could't it be modified to return `[`? It's still not ideal but it makes more sense. Then the $quoteCharacterClass could be set to `preg_quote($platform->getIdentifierQuoteCharacter())`

deeky666: @maryo I don't think it makes much sense as we cannot profit from it. Applications not knowing that it's SQL Server will assume it is also the ending quote character and still generate invalid SQL. IIRC it is possible to enable ANSI quotes in SQL Server so in that case `"` would work.
Won't open another can here. For now it is only about the platform awareness that's all. Everything else will be addressed in `3.0`

maryo: If `"` can be valid then you're right :)

maryo:
"also please test if that fixes your problem if you have time to."
@deeky666 I've just replaced my workarround with your fix uxing "doctrine/dbal":"dev-DBAL-2295 as 2.6.x-dev". It works fine.

deeky666:
@maryo thank you for testing this. Now the only things that's left is figuring out how to add tests properly...

maryo:
Since nobody contributed since that time I wanted to complete the tests but you are right - it's tricky to test this implementation properly since it expects actual implementations of AbstractPlatform and contains those conditions and so the test would need to reflect it also. And since there is no possibility in PHPUnit to chain/merge data providers it would be very ugly.

So I created another PR https://github.com/doctrine/dbal/pull/2496 which solves just the exact problem I had just using negative look behind.

Ocramius:
@maryo since @deeky666 didn't get to check your diff yet, I will have to delay this (removed milestone)

lib/Doctrine/DBAL/Connection.php
```diff
@@ -812,7 +812,12 @@ public function executeQuery($query, array $params = array(), $types = array(),
 
         try {
             if ($params) {
-                list($query, $params, $types) = SQLParserUtils::expandListParameters($query, $params, $types);
+                list($query, $params, $types) = SQLParserUtils::expandListParameters(
```
lcobucci: `[$query, $params, $types]`

lib/Doctrine/DBAL/Connection.php
```diff
@@ -971,7 +976,12 @@ public function executeUpdate($query, array $params = array(), array $types = ar
...
```
lcobucci: `[$query, $params, $types]`

lib/Doctrine/DBAL/SQLParserUtils.php
```diff
@@ -43,13 +46,17 @@ class SQLParserUtils
...
-     * @param boolean $isPositional
+     * @param string                $statement
+     * @param boolean               $isPositional
+     * @param AbstractPlatform|null $platform     The platform to use for retrieving the placeholder positions.
      *
      * @return array
      */
-    static public function getPlaceholderPositions($statement, $isPositional = true)
-    {
+    static public function getPlaceholderPositions(
```
lcobucci: `public static`

lib/Doctrine/DBAL/SQLParserUtils.php
```diff
@@ -76,15 +83,16 @@ static public function getPlaceholderPositions($statement, $isPositional = true)
...
+     * @param array  $params                  The parameters to bind to the query.
+     * @param array  $types                   The types the previous parameters are in.
+     * @param AbstractPlatform $platform|null The platform to make parameter list expansion aware of.
      *
      * @return array
      *
      * @throws SQLParserUtilsException
      */
-    static public function expandListParameters($query, $params, $types)
+    static public function expandListParameters($query, $params, $types, AbstractPlatform $platform = null)
```
lcobucci: `public static`

lib/Doctrine/DBAL/SQLParserUtils.php
```diff
@@ -196,16 +204,46 @@ static public function expandListParameters($query, $params, $types)
...
      * 1 => offset of fragment in $statement
      *
-     * @param string $statement
+     * @param string                $statement
+     * @param AbstractPlatform|null $platform The platform to use for unquoting.
+     *
      * @return array
      */
-    static private function getUnquotedStatementFragments($statement)
+    static private function getUnquotedStatementFragments($statement, AbstractPlatform $platform = null)
```
lcobucci: `private static`

lcobucci: Since it's a private method we can add the type declarations and use `?AbstractPlatform $platform` instead of `AbstractPlatform $platform = null`

lcobucci (changes requested):
Patch looks good, just some small things

PR #2307, opened 2016-01-21
Add support for Easy Connect string as connection parameter in OracleDB

### LINKED ISSUE #2306: Feature request: Easy connect string support for Oci8Driver
I am using the [Oci8Driver](https://github.com/doctrine/dbal/blob/02e5b61baedbeb1832702d0feac3174e70c32c74/lib/Doctrine/DBAL/Driver/OCI8/Driver.php) currently with static connect information, such as host, port and dbname. These parameters are used in the [AbstractOracleDriver#L114](https://github.com/doctrine/dbal/blob/02e5b61baedbeb1832702d0feac3174e70c32c74/lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php#L114) class to generate an easy connect string, which is used to connect to the server/database. 

However, I want to switch to LDAP resolving for the database host and port. The LDAP server already returns an easy connect string, which currently can not be used directly. I parse the string to retrieve the first mentioned hostname and port, which are set in the connection using a Symfony Compiler Pass (see the shortened snippet below).

```php
/** @var DefinitionDecorator $dbal */
$matches        = array();
preg_match('/HOST=([^)]*)/ix', $connectionInfo, $matches);
$host = $matches[1];
preg_match('/PORT=([^)]*)/ix', $connectionInfo, $matches);
$port = $matches[1];
$dbal                 = $container->getDefinition('doctrine.dbal.osiris_connection');
$arguments            = $dbal->getArguments();
$arguments[0]['host'] = $host;
$arguments[0]['port'] = $port;
$dbal->setArguments($arguments);
```

This creates a problem when the first server does not have the database available: an error is thrown (in this case 'ORA-12505: TNS:listener does not currently know of SID given in connect descriptor') and there is no possibility to switch to the other database, although there is a connection list available during the Compiller Pass (but not during runtime). 

```
(DESCRIPTION=
  (ADDRESS_LIST=
    (ADDRESS=
      (PROTOCOL=TCP)
      (Host=server1.host.nl)
      (Port=1521)
    )
    (ADDRESS=
      (PROTOCOL=TCP)
      (Host=server2.host.nl)
      (Port=1521)
    )
  )
  (CONNECT_DATA=
    (SERVER=DEDICATED)
    (SERVICE_NAME=PROD.host.nl)
  )
)
```

I believe that when the connect string can be set in the driver, that handling of such database errors can be done by the lower driver (correct me if I'm wrong). 

So, it would be convenient to be able to set the connect string directly, which can be used if set. I can make a PR for this if wanted. Any opinions?

P.S.: The link in [AbstractOracleDriver#L112](https://github.com/doctrine/dbal/blob/02e5b61baedbeb1832702d0feac3174e70c32c74/lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php#L112) is no longer valid, I can fix that as well.

bobvandevijver:
As discussed in #2306, the standard for sharing Oracle database parameters is using a connect description. This PR adds support for this with a simple parameter.

I've updated the code, added a line in the doc and added a test (not sure if I did the last one correct). The extra tests pass here.

Note: when merged, I will also add this to the DoctrineBundle, such that it can easily be used in Symfony.

### DIFF
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -295,9 +295,11 @@ pdo\_oci / oci8
 -  ``instancename`` (string): Optional parameter, complete whether to
    add the INSTANCE_NAME parameter in the connection. It is generally used
    to connect to an Oracle RAC server to select the name of a particular instance.
-
-Complete connection descriptors are also supported. You only need to provide the following parameter, and none of the above:
-- ``connectstring`` (string): Complete Easy Connect connection descriptor, see https://docs.oracle.com/cd/E11882_01/network.112/e41945/naming.htm.
+-  ``connectstring`` (string): Complete Easy Connect connection descriptor,
+   see https://docs.oracle.com/database/121/NETAG/naming.htm. When using this option,
+   you will still need to provide the ``user`` and ``password` parameters, but the other
+   parameters will no longer be used. Note that when using this parameter, the `getHost`
+   and `getPort` methods from `Doctrine\DBAL\Connection` will no longer function as expected.
 
 pdo\_sqlsrv / sqlsrv
 ^^^^^^^^^^^^^^^^^^^^
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -296,6 +296,8 @@ pdo\_oci / oci8
    add the INSTANCE_NAME parameter in the connection. It is generally used
    to connect to an Oracle RAC server to select the name of a particular instance.
 
+Complete connection descriptors are also supported. You only need to provide the following parameter, and none of the above:
+- ``connectstring`` (string): Complete Easy Connect connection descriptor, see https://docs.oracle.com/cd/E11882_01/network.112/e41945/naming.htm.
 
 pdo\_sqlsrv / sqlsrv
 ^^^^^^^^^^^^^^^^^^^^
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -297,9 +297,9 @@ pdo\_oci / oci8
    to connect to an Oracle RAC server to select the name of a particular instance.
 -  ``connectstring`` (string): Complete Easy Connect connection descriptor,
    see https://docs.oracle.com/database/121/NETAG/naming.htm. When using this option,
-   you will still need to provide the ``user`` and ``password` parameters, but the other
-   parameters will no longer be used. Note that when using this parameter, the `getHost`
-   and `getPort` methods from `Doctrine\DBAL\Connection` will no longer function as expected.
+   you will still need to provide the ``user`` and ``password`` parameters, but the other
+   parameters will no longer be used. Note that when using this parameter, the ``getHost``
+   and ``getPort`` methods from ``Doctrine\DBAL\Connection`` will no longer function as expected.
 
 pdo\_sqlsrv / sqlsrv
 ^^^^^^^^^^^^^^^^^^^^
--- a/lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
@@ -109,11 +109,10 @@ abstract class AbstractOracleDriver implements Driver, ExceptionConverterDriver
      *
      * @return string
      *
-     * @link https://docs.oracle.com/cd/E11882_01/network.112/e41945/naming.htm
+     * @link https://docs.oracle.com/database/121/NETAG/naming.htm
      */
     protected function getEasyConnectString(array $params)
     {
-
         if ( ! empty($params['connectstring'])) {
             return $params['connectstring'];
         }
--- a/lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
@@ -109,10 +109,15 @@ abstract class AbstractOracleDriver implements Driver, ExceptionConverterDriver
      *
      * @return string
      *
-     * @link http://download.oracle.com/docs/cd/E11882_01/network.112/e10836/naming.htm
+     * @link https://docs.oracle.com/cd/E11882_01/network.112/e41945/naming.htm
      */
     protected function getEasyConnectString(array $params)
     {
+
+        if ( ! empty($params['connectstring'])) {
+            return $params['connectstring'];
+        }
+
         if ( ! empty($params['host'])) {
             if ( ! isset($params['port'])) {
                 $params['port'] = 1521;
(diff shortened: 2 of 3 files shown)


deeky666:
I see another problem with this. `Doctrine\DBAL\Connection` provides methods for inspecting the connection parameters like `getHost()`, `getPort()` which is not possible anymore when only providing the easy connect string. There might be applications that rely on those values, maybe even inside Doctrine packages. Connecting like this would break those applications in this case. I tend to better not support this in DBAL directly. @Ocramius @zeroedin-bill thoughts?

bobvandevijver:
I guess you're right on that one, although I can not find any usage of the methods in the Doctrine package I use. I do not experience any problems in my large applications with these adjustments, maybe we can fix this with documentation?

Otherwise I will simply overwrite the class using the Symfony inheritance method, adding this for my project only. Your call to make :+1:

Ocramius:
@deeky666 very good point. A simple workaround (for now) is to force the connection string AND the other parameters to be passed in. This is counter-intuitive, but BC compliant.

bobvandevijver:
I've been looking into this host/port issue, but the reason I want to use it is because of the support for multiple servers, as described by the easy connect string example below:

```
(DESCRIPTION=
  (ADDRESS_LIST=
    (ADDRESS=
      (PROTOCOL=TCP)
      (Host=server1.host.nl)
      (Port=1521)
    )
    (ADDRESS=
      (PROTOCOL=TCP)
      (Host=server2.host.nl)
      (Port=1521)
    )
  )
  (CONNECT_DATA=
    (SERVER=DEDICATED)
    (SERVICE_NAME=PROD.host.nl)
  )
)
```

This means that my host/port in the parameters, even when providing it specifically, could result in the wrong information when using those calls. I checked for methods to retrieve the current connected host from the PHP OCI driver, but it does not appear to have such functionality.

How about adding a `getConnectString` method to the `Doctrine\DBAL\Connection` such that user wanting to use these easy connect strings can retrieve them from the parameters and check them themselves? That way there is no BC-break and an option to still get the information. Only users who switch to the string will be affected and by adding the correct documentation, they can notified about the consequences.

deeky666:
@bobvandevijver what exactly do you mean by "support for multiple servers"? What is defining multiple hosts doing in fact when connecting? It looks like this is some use case that is not intended for the default connection implementation of DBAL (not sure yet). Also we won't add a `getConnectString()` method to the high level `Doctrine\DBAL\Connection` API as it must server the common needs of all vendors and should not contain vendor specific API.
Please clear me up on the multiple hosts thingy :)

bobvandevijver:
@deeky666 The multiple hosts can have two reasons: fail over and load balancing. According to the [Oracle documentation](https://docs.oracle.com/cd/E11882_01/network.112/e10835/tnsnames.htm), you can set the exact behavior with the connect string when including options such as `FAILOVER` and `LOAD_BALANCE`.

The example I provided is a simple random load balancing description, but I actually need this for fail over implementation (the description I get from the organization through LDAP is currently incorrect..). 

So, in short, by defining multiple hosts, the internal Oracle C library chooses the best possible viable connection based on the information provided in the connect descriptor.

deeky666:
Gosh Oracle connection parameters are mess :(
It is really hard to put that into the abstraction level in a reasonable way. Considering the newly introduced possibility to connect via a connection URL, I wonder how we could even put all those possibilities into the URI syntax at all. As `host` in
```
scheme:[//[user:password@]host[:port]][/]path[?query][#fragment]
```
is a required part if you provide `user` and `password`, this will never work out without putting in some "dummy" value there (which would cause headaches afterwards when relying on that parameter).

I get your use-case now and it is quite reasonable, but I really don't know how to put that into DBAL abstraction in a clean way yet. Will need some more time to think about it...

bobvandevijver:
I would say that it is impossible to put **all** information in the URI syntax without making stuff too difficult and error prone... Note that together with the connect descriptor you are still required to supply the credentials. 
I guess a clean solution is impossible in this case, so for my project I will overwrite the class for my needs.

docs/en/reference/configuration.rst
```diff
@@ -296,6 +296,8 @@ pdo\_oci / oci8
    add the INSTANCE_NAME parameter in the connection. It is generally used
    to connect to an Oracle RAC server to select the name of a particular instance.
 
+Complete connection descriptors are also supported. You only need to provide the following parameter, and none of the above:
```
deeky666: This breaks the list in the markup. Please start with `- connectstring (string)` (as with the other parameters) and then have the complete description follow after. Also the statement is not quite true. `user` and `password` are still required, right? Please consider that in the description.

lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
```diff
@@ -109,10 +109,15 @@ public function getSchemaManager(\Doctrine\DBAL\Connection $conn)
      *
      * @return string
      *
-     * @link http://download.oracle.com/docs/cd/E11882_01/network.112/e10836/naming.htm
+     * @link https://docs.oracle.com/cd/E11882_01/network.112/e41945/naming.htm
      */
     protected function getEasyConnectString(array $params)
     {
+
```
deeky666: Please remove the empty line.

deeky666:
@bobvandevijver okay I think I'll accept that for `2.6`. As URI compatibility is probably not given anyways as of yet, I can accept that (as I see the need for it). For `3.0` we might consider that from the beginning when designing URL connections etc.

bobvandevijver:
I've updated it according to your comments. Also changed the URL in the docs to the same URL PHP uses in the `oci_connect` docs.

docs/en/reference/configuration.rst
```diff
@@ -295,7 +295,11 @@ pdo\_oci / oci8
 -  ``instancename`` (string): Optional parameter, complete whether to
    add the INSTANCE_NAME parameter in the connection. It is generally used
    to connect to an Oracle RAC server to select the name of a particular instance.
-
+-  ``connectstring`` (string): Complete Easy Connect connection descriptor,
+   see https://docs.oracle.com/database/121/NETAG/naming.htm. When using this option,
+   you will still need to provide the ``user`` and ``password` parameters, but the other
```
deeky666: You messed up some backticks here :D There is missing one after `password` and I suppose `getHost` and `getPort` will require double backticks, too ;) Otherwise fine.

bobvandevijver: Oh, dang it. The backtick is a little bit buggy on my laptop keyboard, and rst != md. Will fix!

deeky666:
@bobvandevijver thanks for your work!

gaving:
Just throwing in my appreciation for this feature, lets me specify a connection string with multiple hosts for fail over purposes! :+1:

bobvandevijver:
@deeky666 Being more than a year since this PR was merged: when do I we get an tag on it? We do not want to use any dev-dependencies in our product, so we're kinda waiting for it :)

PR #2308, opened 2016-01-21
Add missing args to fetchAll/fetch

Seldaek:
Without this stuff static analysis with tools like phan fails on `$statement->fetchAll(PDO::FETCH_CLASS, 'classname');` for example with `PhanParamTooMany Call with 2 arg(s) to \Doctrine\DBAL\Driver\resultstatement::fetchall() which only takes 1 arg(s)`

The second commit is similar, allowing ints for QueryBuilder::setParameter (`PhanTypeMismatchArgument Argument 3 (type) is int but \Doctrine\DBAL\Query\querybuilder::setparameter() takes null|string)` because all the `PDO::FETCH_*` are ints.


Ocramius:
Note: marked for 3.x because implementing classes may break after this change (see https://3v4l.org/hcv3L )

Seldaek:
Yup I realize it's not really feasible now but would be nice to fix eventually :)

mathroc:
`Doctrine\DBAL\Connection::query()` have the same problem

Majkl578:
This has been handled in #2527.

PR #2309, opened 2016-01-22
Remove useless ternary in the `DecimalType`

ghost:
see https://github.com/doctrine/doctrine2/pull/1104#issuecomment-51755834
Thanks to @stof

### DIFF
--- a/lib/Doctrine/DBAL/Types/DecimalType.php
+++ b/lib/Doctrine/DBAL/Types/DecimalType.php
@@ -49,6 +49,6 @@ class DecimalType extends Type
      */
     public function convertToPHPValue($value, AbstractPlatform $platform)
     {
-        return (null === $value) ? null : $value;
+        return $value;
     }
 }


Ocramius:
```php
if ($val === true) {
    return true;
} elseif ($val === false) {
    return false;
} else {
    return $val;
}
```

#facepalm 

Thanks for sending the patch, @JHGitty!

JHGitty:
@Ocramius As I understand this fix could also be merged to 2.1, 2.2, 2.3, 2.4 and 2.5
because it is just a minor code improvement. Are these versions still supported and should get this fix?

Ocramius:
@JHGitty anything that isn't an actual bugfix won't make it into stable versions

JHGitty:
@Ocramius Okay. Thanks :+1:

deeky666:
lol crying tears over here :D thanks @JHGitty @Ocramius

PR #2310, opened 2016-01-22
[DB2] Move row number to the end of the field list in query limit/offset modification

asgrim:
Change order of parameters in doModifyLimitQuery to work with Migrations.

Specifically, when retrieving the `migrations:status`, it expects the first cell to be returned to be the "current version" etc., but in this, the `DC_ROWNUM` is returned, causing it to break.

Sidenote: I did already fix this in #910, but we could do with this fixed in existing code anyway.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/DB2Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/DB2Platform.php
@@ -749,7 +749,7 @@ class DB2Platform extends AbstractPlatform
         $offset = (int) (($offset)?:0);
 
         // Todo OVER() needs ORDER BY data!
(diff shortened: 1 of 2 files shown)


deeky666:
LGTM :+1:

deeky666:
@asgrim thanks!

deeky666:
Backported to `2.5` branch via https://github.com/doctrine/dbal/commit/84a1841c10b3cee2b24da33d471ea53640c31dfb, https://github.com/doctrine/dbal/commit/9e1300bffb9cae048705bb4ed39c19f33e6896d6

Ocramius:
@deeky666 awesum!

asgrim:
thanks @deeky666 :)

Issue #2314, opened 2016-02-11
DateIntervalType supports conversion of invalid values


MisatoTremor:
The current implementation allows to convert a DateInterval to DB values that can not be converted back to PHP values. As a side node the DB field length of the type of 20 could not hold these values.

Intervals with values that overflow their moduli like `P53W` or `P100M` will yield these DB values:  `P0000-00-371T00:00:00` or `P0000-100-00T00:00:00`.
While these exceed the declared 20 character field length they are also invalid as per ISO 8601 and the PHP manual. If they were tried to be converted back to a DateInterval an exception would be thrown (e.g. 'DateInterval::__construct(): Unknown or bad format (P0000-00-371T00:00:00)')

Maybe DateIntervals should be saved in the `PnYnMnDTnHnMnS` format which allows overflowing.

Ping @vbartusevicius

vbartusevicius:
Initially `PnYnMnDTnHnMnS` was in my original implementation, but later @zeroedin-bill suggested to use datetime format (https://github.com/doctrine/dbal/pull/854#issuecomment-112848030) as it's sortable in db.
So if @zeroedin-bill would suggest to use `PnYnMnDTnHnMnS` - I will refactor current implementation

zeroedin-bill:
@MisatoTremor @vbartusevicius It's a very good point, but I'm not sure how to satisfy both conditions that the database representation be sortable and the format is comprehensive.

Ocramius:
Note that changing a format that landed in a stable version is not possible

MisatoTremor:
@vbartusevicius @zeroedin-bill When saving it normalized like ``PnnYnnMnnDTnnHnnMnnS`` with ``'P'.$value->format('%YY%MM%DDT%HH%IM%SS')`` it would be almost as sortable as it is now. Only exception is that the year will have 2 instead of 4 digits

@Ocramius Afaik there is no stable release with the DateIntervalType in it, yet.

vbartusevicius:
@MisatoTremor - you're right, there is no `DateIntervalType` in https://github.com/doctrine/dbal/tree/v2.5.4/lib/Doctrine/DBAL/Types

So if @zeroedin-bill will agree, I could quickly make some changes to `DateIntervalType` to internally store in @MisatoTremor's suggested format `'P' . $value->format('%YY%MM%DDT%HH%IM%SS')`

vbartusevicius:
Created a PR: https://github.com/doctrine/dbal/pull/2316

[...]

PR #2316, opened 2016-02-15
Fix `DateInterval` database value truncation (string overflow)

vbartusevicius:
`DateIntervalType` now stores data in `P%YY%MM%DDT%HH%IM%SS` format as `%M-%DT%H:%I:%S` brings a lot of problems with large intervals, i.e.: `P100DT500M`

### DIFF
--- a/lib/Doctrine/DBAL/Types/DateIntervalType.php
+++ b/lib/Doctrine/DBAL/Types/DateIntervalType.php
@@ -22,7 +22,7 @@ class DateIntervalType extends Type
      */
     public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform)
     {
-        $fieldDeclaration['length'] = 20;
+        $fieldDeclaration['length'] = 255;
         $fieldDeclaration['fixed']  = true;
 
         return $platform->getVarcharTypeDeclarationSQL($fieldDeclaration);
@@ -38,9 +38,8 @@ class DateIntervalType extends Type
         }
 
         if ($value instanceof \DateInterval) {
-            return 'P'
-                . str_pad($value->y, 4, '0', STR_PAD_LEFT) . '-'
-                . $value->format('%M-%DT%H:%I:%S');
+            /** @var \DateInterval $value */
+            return $value->format('P%YY%MM%DDT%HH%IM%SS');
         }
 
         throw ConversionException::conversionFailedInvalidType($value, $this->getName(), ['null', 'DateInterval']);
@@ -58,10 +57,10 @@ class DateIntervalType extends Type
         try {
             return new \DateInterval($value);
         } catch (\Exception $exception) {
-            throw ConversionException::conversionFailedFormat($value, $this->getName(), 'PY-m-dTH:i:s', $exception);
+            throw ConversionException::conversionFailedFormat($value, $this->getName(), 'P%YY%MM%DDT%HH%IM%SS', $exception);
         }
     }
-    
+
     /**
      * {@inheritdoc}
      */
--- a/lib/Doctrine/DBAL/Types/DateIntervalType.php
+++ b/lib/Doctrine/DBAL/Types/DateIntervalType.php
@@ -38,7 +38,6 @@ class DateIntervalType extends Type
         }
 
         if ($value instanceof \DateInterval) {
-            /** @var \DateInterval $value */
             return $value->format('P%YY%MM%DDT%HH%IM%SS');
         }
 
--- a/tests/Doctrine/Tests/DBAL/Types/DateIntervalTest.php
+++ b/tests/Doctrine/Tests/DBAL/Types/DateIntervalTest.php
@@ -32,7 +32,7 @@ class DateIntervalTest  extends \Doctrine\Tests\DbalTestCase
     {
         $interval = new \DateInterval('P2Y1DT1H2M3S');
 
-        $expected = 'P0002-00-01T01:02:03';
+        $expected = 'P02Y00M01DT01H02M03S';
         $actual = $this->type->convertToDatabaseValue($interval, $this->platform);
 
         $this->assertEquals($expected, $actual);
@@ -40,9 +40,9 @@ class DateIntervalTest  extends \Doctrine\Tests\DbalTestCase
 
     public function testDateIntervalConvertsToPHPValue()
     {
-        $date = $this->type->convertToPHPValue('P0002-00-01T01:02:03', $this->platform);
+        $date = $this->type->convertToPHPValue('P02Y00M01DT01H02M03S', $this->platform);
         $this->assertInstanceOf('DateInterval', $date);
-        $this->assertEquals('P2Y0M1DT1H2M3S', $date->format('P%yY%mM%dDT%hH%iM%sS'));
+        $this->assertEquals('P02Y00M01DT01H02M03S', $date->format('P%YY%MM%DDT%HH%IM%SS'));
     }
 
     public function testInvalidDateIntervalFormatConversion()


lib/Doctrine/DBAL/Types/DateIntervalType.php
```diff
@@ -22,7 +22,7 @@ public function getName()
      */
     public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform)
     {
-        $fieldDeclaration['length'] = 20;
+        $fieldDeclaration['length'] = 255;
```
MisatoTremor: Despite my remark that 20 can be to short for large intervals, 255 might be a bit over the top. :)
Question is what a sensible limit would be ...
To have an interval spanning the seconds past since the "beginning" of the visible universe would be P00Y00M00DT00H00M435400000000000000S = 66 characters. I can't imagine right now any higher possible overflowing, but with adding some "buffer" length I think any value between 70 or 80 should be safe.

vbartusevicius: Yeah, but theoretically you can add as many numbers in any interval part as you want, so i.e. MySQL 64, 128 or other length lower than 255 (varchar) reservers same amount of memory, so is it really needed to calculate potential minimum?

Ocramius:
Linking #2314

lib/Doctrine/DBAL/Types/DateIntervalType.php
MisatoTremor: Afaik that only applies to the memory storage type.
This being DBAL the length should not consider only one specific platform. And if I'm not mistaken a developer could adjust the length if needed when defining the schema.
So this is mainly about finding a sensible default length.

juliangut:
@Ocramius Do you plan to include this new type is next release? 2.4.6? or is it too soon?

Ocramius:
New features can only go to minor versions, not to patch releases

vbartusevicius:
@zeroedin-bill, @Ocramius - could you please review the code?

juliangut:
You are right @Ocramius, then plans for 2.6 :wink: ?

Ocramius:
@vbartusevicius code looks ok to me. Assigned to @deeky666 for review.

@juliangut @deeky666 is the gateway keeper for DBAL: he's in command

lib/Doctrine/DBAL/Types/DateIntervalType.php
```diff
@@ -38,9 +38,8 @@ public function convertToDatabaseValue($value, AbstractPlatform $platform)
         }
 
         if ($value instanceof \DateInterval) {
-            return 'P'
-                . str_pad($value->y, 4, '0', STR_PAD_LEFT) . '-'
-                . $value->format('%M-%DT%H:%I:%S');
+            /** @var \DateInterval $value */
```
deeky666: This comment is not needed, please remove.

vbartusevicius:
@deeky666 - any luck on this?

galeaspablo:
@vbartusevicius For more seconds, more minutes, etc. I would pad everything, to keep sortability. Just like it was done for years. 

```
//convertToDatabaseValue
if ($value instanceof \DateInterval) {
             return 'P'
                 . str_pad($value->y, 10, '0', STR_PAD_LEFT) . 'Y'
                 . str_pad($value->m, 10, '0', STR_PAD_LEFT) . 'M'
                 . str_pad($value->d, 10, '0', STR_PAD_LEFT) . 'DT'
                 . str_pad($value->h, 10, '0', STR_PAD_LEFT) . 'H'
                 . str_pad($value->i, 10, '0', STR_PAD_LEFT) . 'I'
                 . str_pad($value->s, 10, '0', STR_PAD_LEFT) , 'S';
          }
```

We would still need some arbitrary number, 10 in the case above. But you get the idea.

vbartusevicius:
@galeaspablo - what do you mean by increasing the number of years? this PR already solves the problem you're addressing here:
```
(new \DateInterval('P123456789Y12M4DT00H00M00S'))->format('P%YY%MM%DDT%HH%IM%SS');
// "P123456789Y12M04DT00H00M00S"
```

In addition, why `new \DateInterval('PT61S');` looks invalid to you? You can easily manipulate such intervals:
```
$a = new \DateInterval('PT61S');
$b = new \DateTime('2016-12-15 12:00:00');
var_dump($b->add($a));

class DateTime#3 (3) {
  public $date =>
  string(26) "2016-12-15 12:01:01.000000"
  public $timezone_type =>
  int(3)
  public $timezone =>
  string(15) "Europe/Helsinki"
}
```

galeaspablo:
@vbartusevicius Like I said,

Invalid according to \DateInterval::__construct
From http://php.net/manual/en/dateinterval.construct.php

>The specification can also be represented as a date time. A sample of one year and four days would be P0001-00-04T00:00:00. But the values in this format can not exceed a given period's roll-over-point (e.g. 25 hours is invalid).

-----

Plus, the PR does not keep sortability, which was considered for this feature. And should remain so. Your PR would mean the database would sort 10, 111, and 12 seconds as  `10`, `111`, `12`.

If you'd like to keep your approach, we'd need to `str_pad()` seconds, minutes, etc.

By increase the number of year digits, I mean that we store 10 digits of a year instead of 4 (as before), or 2 as you propose.

vbartusevicius:
@galeaspablo - your case with invalid spec only applies to datetime representation (please read again):

> The specification can also be represented as a date time. [...] But the values in this format can not exceed a given period's roll-over-point (e.g. 25 hours is invalid).

If you use a regular `DateInterval` spec, you can easily create valid date intervals with 30 hours or so:
```
new \DateInterval('PT30H61S');
```
just try this.


As of 2 digits for year - where do you see that? `'P%YY%MM%DDT%HH%IM%SS'` - here `%Y` is a placeholder and `Y` is just a letter.

galeaspablo:
> @vbartusevicius As of 2 digits year...
You are right.

> @vbartusevicius You can easily create valid date intervals with 30 hours.
True.

>@galeaspablo The PR does not keep sortability, which was considered for this feature. And should remain so. Your PR would mean the database would sort 10, 111, and 12 seconds as 10, 111, 12.
You cannot address this by using `str_pad()`

>@galeaspablo store valid (according to the specification intervals), to keep sortability. Otherwise 1 minute 10 seconds < 0 minutes 80 seconds
So we're back to, throw an exception if we have an "invalid" (according to datetime) interval.
Or converting, e.g. 61 seconds is stored as 1 minute 1 second.

vbartusevicius:
@galeaspablo - thanks for information, but I like to hear the opinion of @deeky666 or @Ocramius on this.

galeaspablo:
Agreed. :)

MisatoTremor:
@galeaspablo 
> Or converting, e.g. 61 seconds is stored as 1 minute 1 second.
This could lead to invalid behavior, as it actually changes the meaning of that interval.
Consider this years leap second as an example. At 2016-12-31 18:59:59 there will be an additional second. So when using your example interval in a calculation 61 seconds from 18:59:00 will be 19:00:00. After converting it to 1 minute 1 second the same calculation would yield 19:00:01.
The same would apply to leap years and a possible day to month and day conversion.

galeaspablo:
@MisatoTremor I think you meant `a calculation 61 seconds from 18:59:59*`, not `18:59:00` right? 😃 Anyway, what's your take? Happy to hear more.

**So far, I see these options**
1. Store DateIntervals, as they come. But lose enforced sort-ability at the database level (a database would tell you that `00001months00090days` < `00002months00002days`).
2. Only store DateIntervals, where values don't exceed their moduli (i.e. not 61 seconds, 61 minutes, 25 hours, etc). But force users who want to use something like `2 minutes, 1000 seconds`, to use separate int columns.
3. ~~Convert DateIntervals, where values exceed their moduli (i.e. 61 seconds becomes 1 minute, and 1 second). But, this changes the meaning of the interval.... Plus, how do we convert (for instance how many days and months is 35 days).~~ Even though I proposed this, I can now see how this is a bad idea. 🗡

MisatoTremor:
As I see it, DateIntervals can't be reliably sorted anyway so I'd choose 1).
2) Seems to restricting to the user to be worth it.

deeky666:
I really can't get my head around that weird interval stuff, sorry. Will leave that one up for @Ocramius to decide :P Please just make sure to keep BC. Maybe only possible by introducing a new type (ugh).

galeaspablo:
Thanks for the taking the time to chime in @deeky666 . As I mentioned in #2579, 
> Is this really a BC break? This isn't on 2.5.

lib/Doctrine/DBAL/Types/DateIntervalType.php
Ocramius: Tweaking is not really up to us - we just need to accommodate a generic use-case scenario, and that's it.

Ocramius (approved):

Ocramius:
This is ready to merge. Sorry for the delay, @vbartusevicius!

Also, for readers, this is not a BC break, as this class was not in any tagged release.

MisatoTremor:
Thanks @Ocramius for merging and ofc thanks @vbartusevicius for your work on this!

PR #2318, opened 2016-02-17
[Oracle] Fix list table columns when using external or OS authentication with Oracle

butonic:
To use an external authentication mechanism, ef. Oracle Wallet you specify "/" us the db user and "" as the host when calling oci_connect. See http://php.net/manual/de/function.oci-connect.php#refsect1-function.oci-connect-parameters

This works as expected until you try to read the table columns with `getListTableColumnsSQL()`.

Since `$database` is "/" the code that should limit the listed columns to the given user actually prevents listing anything because a user "/" never exists.

I did not find a test for `getListTableColumnsSQL()` but am documenting the debug session that lead to this PR at https://github.com/owncloud/core/wiki/%5BWIP%5D-Oracle-Wallet

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
@@ -697,6 +697,60 @@ EOD;
     }
 
     /**
+     * @dataProvider getReturnsGetListTableColumnsSQL
+     * @group DBAL-831
+     */
+    public function testReturnsGetListTableColumnsSQL($database, $expectedSql)
+    {
+        $this->assertEquals($expectedSql, $this->_platform->getListTableColumnsSQL('"test"', $database));
+    }
+
+    public function getReturnsGetListTableColumnsSQL()
+    {
+        return array(
+            array(
+                null,
+                "SELECT   c.*,
+                         (
+                             SELECT d.comments
+                             FROM   user_col_comments d
+                             WHERE  d.TABLE_NAME = c.TABLE_NAME
+                             AND    d.COLUMN_NAME = c.COLUMN_NAME
+                         ) AS comments
+                FROM     user_tab_columns c
+                WHERE    c.table_name = 'test' 
+                ORDER BY c.column_name"
+            ),
+            array(
+                '/',
+                "SELECT   c.*,
+                         (
+                             SELECT d.comments
+                             FROM   user_col_comments d
+                             WHERE  d.TABLE_NAME = c.TABLE_NAME
+                             AND    d.COLUMN_NAME = c.COLUMN_NAME
+                         ) AS comments
+                FROM     user_tab_columns c
+                WHERE    c.table_name = 'test' 
+                ORDER BY c.column_name"
+            ),
+            array(
+                'scott',
+                "SELECT   c.*,
+                         (
+                             SELECT d.comments
+                             FROM   all_col_comments d
+                             WHERE  d.TABLE_NAME = c.TABLE_NAME
+                             AND    d.COLUMN_NAME = c.COLUMN_NAME
+                         ) AS comments
+                FROM     all_tab_columns c
+                WHERE    c.table_name = 'test' AND c.owner = 'SCOTT'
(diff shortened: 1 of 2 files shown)


Ocramius:
Requires tests

butonic:
I am on it

butonic:
@Ocramius I copied the `@group DBAL-831` ... is that correct?

Ocramius:
@butonic seems good. Assigning to @deeky666 for review.

tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
```diff
@@ -697,6 +697,60 @@ public function testQuotedTableNames()
...
+                             WHERE  d.TABLE_NAME = c.TABLE_NAME
+                             AND    d.COLUMN_NAME = c.COLUMN_NAME
+                         ) AS comments
+                FROM     all_tab_columns c
+                WHERE    c.table_name = 'test' AND c.owner = 'SCOTT'
+                ORDER BY c.column_name"
+            ),
+        );
+    }
+    /**
```
Ocramius: Empty newline to be introduced here

tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
```diff
@@ -697,6 +697,60 @@ public function testQuotedTableNames()
     }
 
     /**
+     * @dataProvider getReturnsGetListTableColumnsSQL
+     * @group DBAL-831
+     */
+    public function testReturnsGetListTableColumnsSQL($database, $expectedSql)
+    {
+        $sql = $this->_platform->getListTableColumnsSQL('"test"', $database);
+        $this->assertEquals($expectedSql, $sql);
```
Ocramius: Can write the entire thing in one line

deeky666:
@butonic @Ocramius hmm another funny Oracle story :D The fix seems a bit akward and tends to lead to additional checks against `/` database name throughout the code base which sucks. As an alternative fix we could maybe simply adjust the abstract Oracle driver to return `null` if `/` is encountered in the connection parameters. This seems also more convenient than coercing `/` to `null` from given method arguments. What do you think?

tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
```diff
@@ -697,6 +697,60 @@ public function testQuotedTableNames()
...
     /**
+     * @dataProvider getReturnsGetListTableColumnsSQL
+     * @group DBAL-831
+     */
+    public function testReturnsGetListTableColumnsSQL($database, $expectedSql)
+    {
+        $this->assertEquals($expectedSql, $this->_platform->getListTableColumnsSQL('"test"', $database));
+    }
+
+    public function getReturnsGetListTableColumnsSQL()
```
deeky666: We tend to avoid testing against complete SQL strings when it comes to testing `getList*()` Methods as those tests are fragile and need to be adjusted on every change in SQL. I would rather do a contains check in the string to check whether `all_tab_columns` or `user_tab_columns` is present (which is the interesting difference).

deeky666:
Btw this would also automatically fix other `getList*()` methods which are affected by this issue, too.

MorrisJobke:
@deeky666 Are there any steps from our side(@butonic and mine ;)) to help out here?

butonic:
@deeky666 could you comment hero or or on the alternate approach in https://github.com/doctrine/dbal/issues/2325. What is missing?

tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
Ocramius: Merging this regardless of the fragile test. We may simply rework the test, should some changes break this.

Ocramius:
Merged, thanks @butonic!

`master`: doctrine/dbal@a3577e49dc3897410f0059b701f5bccbbc646407
`2.5`: doctrine/dbal@9f8c05cd5225a320d56d4bfdb4772f10d045a0c9

PR #2324, opened 2016-02-20
Allow use own exceptions in 3d-side db-connectors.

chernomor:
DBALException::driverExceptionDuringQuery wrap exceptions from our db-connector into self because our exceptions not inherited from DriverException. We use external db-connector (it is standalone project with minimal dependencies) and cannot inherit it's exceptions from DriverException.

### DIFF
--- a/lib/Doctrine/DBAL/DBALException.php
+++ b/lib/Doctrine/DBAL/DBALException.php
@@ -112,7 +112,7 @@ class DBALException extends \Exception
         }
         $msg .= ":\n\n".$driverEx->getMessage();
 
-        if ($driver instanceof ExceptionConverterDriver && $driverEx instanceof DriverException) {
+        if ($driver instanceof ExceptionConverterDriver) {
             return $driver->convertException($msg, $driverEx);
         }
 
@@ -129,7 +129,7 @@ class DBALException extends \Exception
     {
         $msg = "An exception occured in driver: " . $driverEx->getMessage();
 
-        if ($driver instanceof ExceptionConverterDriver && $driverEx instanceof DriverException) {
+        if ($driver instanceof ExceptionConverterDriver) {
             return $driver->convertException($msg, $driverEx);
         }
 
--- a/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
@@ -42,7 +42,7 @@ abstract class AbstractMySQLDriver implements Driver, ExceptionConverterDriver, 
      * @link http://dev.mysql.com/doc/refman/5.7/en/error-messages-client.html
      * @link http://dev.mysql.com/doc/refman/5.7/en/error-messages-server.html
      */
-    public function convertException($message, DriverException $exception)
+    public function convertException($message, \Exception $exception)
     {
         switch ($exception->getErrorCode()) {
             case '1050':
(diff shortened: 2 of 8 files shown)


Ocramius:
If you implement your own friver, exceptions should still be driver exceptions. Why do you need non-driver-exceptions to be thrown?

chernomor:
Our driver is external project which not use doctrine. In our main project we use that driver (more than 10 years) and try to use doctrine in some cases, but doctrine DriverException masked our old warm tube exceptions. :)

What we can do to use native exceptions from external DBAL not dependent from doctrine?

Ocramius:
You'd re-throw custom exceptions when catching in your adapter...

chernomor:
Ocramius, let's see on flow:
1. our main project execute sql query via Doctrine\DBAL\Driver\Statement::execute()
2. Doctrine\DBAL\Driver\Statement::execute() call statement::execute() via out adapter (external db-connector)
3. adapter throw it's own EMyException based on \Exception
4. Doctrine\DBAL\Driver\Statement::execute() catch this EMyException and call DBALException::driverExceptionDuringQuery()
5. DBALException::driverExceptionDuringQuery() check is EMyException instanceof ExceptionConverterDriver? It not, so $driver->convertException will be not call, and driverExceptionDuringQuery()  return itself
6.Doctrine\DBAL\Driver\Statement::execute() throw exception object from DBALException::driverExceptionDuringQuery()
7. our main project catch exception Doctrine\\DBAL\\DBALException instead EMyException

Our db-connector is on level under doctrine, so we cannot catch and re-throw exception from doctrine.

It's enough if doctrine will not mask our exceptions from db-connector. My commit allow transparently pass ANY exception from db-connector through doctrine to upper level.

Ocramius:
> pass ANY exception from db-connector through doctrine to upper level.

Feature on which we worked a lot (converting exceptions into meaningful stuff) :-P

A statement should only throw [`Doctrine\DBAL\Exception`](https://github.com/doctrine/dbal/blob/61e1f860a304c3fb6961ecbec36927b5f20b71c6/lib/Doctrine/DBAL/Statement.php#L151) by design. If you need to see the previous exception, then re-convert in the consumer of the statement (and re-throw there if you rely on it).

If the problem with the conversion comes from the fact that the driver is not throwing driver exceptions (weird), then you should wrap that driver or patch it, IMO

Ocramius:
Assigning to @deeky666, as he wrote most of the conversion code

chernomor:
>Feature on which we worked a lot (converting exceptions into meaningful stuff) :-P
Nice feature! But why it limited by Doctrine\DBAL\Exception only? :)

>If the problem with the conversion comes from the fact that the driver is not throwing driver exceptions (weird), then you should wrap that driver or patch it, IMO
We use own old stable external driver for PDO with big hierarchy of exceptions. It use in several our projects without doctrine. In one project we decide try doctrine for some functional and found what doctrine masks exceptions from our driver.

Exceptions handling in our projects designed for this hierarchy and we cannot change it for various libraries (even for doctrine). 

On the over side: we set general messages for all our exceptions and set event details to specific fields, i.e.:
class MyDBExceptionDublicateKey {
   'message' => 'Duplicate key',
   'query' => 'INSERT INTO table VALUES("a")',
   'connection_name' => 'master',
   'originator' => '/path/to/query/originator', # the first customer caller from stack track (we except all dbal layer and other library functions when search originator)
}
Information structured and may be aggregated automatically.

What offers doctrine for us? Some like this:
'message' => 'An exception occurred while executing \'SELECT h0_.hash AS hash0.....WHERE h0_.hash = ? LIMIT 1\' with params ["69905e21c07ba0119ebc683eb6fcfde3"]:
Cannot connect to database 
\'mysql:host=10.10.10.10;port=3306;dbname=maindb
SQLSTATE[HY000] [1226] User \'main\' has exceeded the \'max_user_connections\' resource (current value: 200)'

Very large and difficult for analize message. No automation.

So I think it's not good behaviour for library to change exceptions which it's not throw: just let it go up!

PR #2325, opened 2016-02-22
database is unknown on os based auth

butonic:
Alternate approach to https://github.com/doctrine/dbal/pull/2318

@Ocramius @deeky666 should I port over the test (or rather only test for the difference)? This is a totally different solution. It works for my case ( Yay! ), but adding that case is kinda testing this at the wrong place, isn't it?

Let's see what travis thinks.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
@@ -83,6 +83,10 @@ abstract class AbstractOracleDriver implements Driver, ExceptionConverterDriver
     {
         $params = $conn->getParams();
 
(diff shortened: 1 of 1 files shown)


butonic:
ping @Ocramius @deeky666 for a review on the different approach to fixing os based auth

butonic:
@Ocramius @deeky666 any feedback?

butonic:
@Ocramius @deeky666 could you comment? Anything we can do?

DeepDiver1975:
@deeky666 @Ocramius anything I can help to get this in?

Ocramius:
Requires tests, as far as I can see...

morozov:
Fixed via https://github.com/doctrine/dbal/pull/3606.

PR #2328, opened 2016-02-28
[FIX] - Fix PHP original PDO_PGSQL::lastInsertId

phackwer:
Fix PHP original PDO_PGSQL::lastInsertId which sometimes complains

about a sequence that has not been used it on the transaction (error):

Object not in prerequisite state: 7 ERROR:  currval of sequence sequence_name is not yet defined in this session

### DIFF
--- a/lib/Doctrine/DBAL/Driver/PDOPgSql/Exception.php
+++ b/lib/Doctrine/DBAL/Driver/PDOPgSql/Exception.php
@@ -1,0 +1,39 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Driver\PDOPgSql;
+
+/**
+ * Class Exception
+ * @package Doctrine\DBAL\Driver\PDOPgSql
+ * @author Pablo Santiago Sanchez <phackwer@hotmail.com>
+ * @since 2.4
+ */
+class Exception extends \Exception
+{
+    /**
+     * @param array $error
+     *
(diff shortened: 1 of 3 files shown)


Ocramius:
@deeky666 is 2.4 still maintained?

lib/Doctrine/DBAL/Driver/PDOPgSql/Exception.php
```diff
@@ -0,0 +1,39 @@
...
+ * @since 2.4
+ */
+class Exception extends \Exception
+{
+    /**
+     * @param array $error
+     *
+     * @return \Doctrine\DBAL\Driver\PDOPgSql\Exception
+     */
+    static public function fromErrorInfo($error)
```
Ocramius: unused?

Ocramius:
@phackwer what kind of situation triggers the problem? Can you reproduce it in a test case?

lib/Doctrine/DBAL/Driver/PDOPgSql/Exception.php
phackwer: No way!!! Please, note the namespace.

https://github.com/doctrine/dbal/pull/2328/files#diff-bb13d524a0e6064b2a578f84c62c2ff5R66

phackwer:
Marco,

How to trigger:
When you try to execute a "lastInsertId" on any sequence that has not yet executed an insert, you get this error, since currval not avaliable (this is how you trigger it by forcing the error). Somewhere on Doctrine code, there may be a bad call on lastInsertId (I don't know where exactly), making it to be called before the insert is actually made on the database. I just couldn't backtrace it to force the error on Doctrine code. 

This is a PHP bug (which I will also and still send the fix - here's the cause (https://github.com/php/php-src/blob/master/ext/pdo_pgsql/pgsql_driver.c#L375), and this is a Doctrine workaround to allow Doctrine to still work with earlier PHP versions, since bugfix will be avaliable only on 5.6.x and 7.x.

Answering about branch 2.4, I don't know if it's still maintened, but you (specifically you) have made the 2.4.5 release earlier these year and it is marked as the latest release. Also this is the one we are using here in Brazil with some government clients that insists on force us to make code compatible with PHP 5.3.3, just because this is the one supported by RedHat "Enterprice" Linux 6, which is the one they have support contracts. Maybe I should upgrade to 2.5.x, but still, this fix should be made.

About the PHP bug, you can google it just to find out that this is a big that affects thousands of people around the world. The correct solution PostgreSQL solution since version 8.1 is to call the LASTVAL() function instead of using the currval old option.

Releases (latest is yours and on 2.4 branch):
https://github.com/doctrine/dbal/releases/tag/v2.4.5

PHP Bug: 
https://www.google.com.br/search?sourceid=chrome-psyapi2&ion=1&espv=2&es_th=1&ie=UTF-8&q=Object%20not%20in%20prerequisite%20state%3A%207%20ERROR%3A%20currval%20of%20sequence&oq=Object%20not%20in%20prerequisite%20state%3A%207%20ERROR%3A%20currval%20of%20sequence&aqs=chrome..69i57j69i59.510j0j7

phackwer:
About tests: I will try to write them. So far, the bug seems a critical one and fixes doctrine for thousands of systems.

phackwer:
As you can see here, also phalcon has this problem, since it is PHP's bug:

https://github.com/phalcon/cphalcon/issues/1134

phackwer:
Any idea of when will your database be avaliable again?

This build error has no relation to my code:

https://travis-ci.org/doctrine/dbal/jobs/112441592#L319

lib/Doctrine/DBAL/Driver/PDOPgSql/Exception.php
Ocramius: Yeah, I see that you use the constructor, not this named constructor...

Ocramius:
Ah yes, we do not support php <5.4 anymore. Unless there is a security issue, this patch will only land in 2.5

phackwer:
:-(

Can't you at least make a final release to overwrite this PHP problem for those who must use 2.4.x?

Just this last pull request? XD

daniloesser:
I've got the same situation here. 
+1 vote for this pull request.

fabioluciano:
+1 vote on thia pull request.


Same issue here.

GustavoMagalhaess:
+1 vote for this pull request.

Ocramius:
> Can't you at least make a final release to overwrite this PHP problem for those who must use 2.4.x?

This is not a security issue. Don't think it's gonna happen. The merge will happen for 2.5.x. As you pointed out to me, for 2.4.x, the workaround is to simply use your own custom platform.

@ the other 3 that "upvoted" - upvoting has nothing to do with release schedules, sorry :-\

celorodovalho:
I've got the same situation here. 
+1 vote for this pull request.

phackwer:
This has been fixed directly on PHP source by myself.

https://github.com/php/php-src/pull/2014

PR #2329, opened 2016-02-28
[FIX] - Fix PHP original PDO_PGSQL::lastInsertId

phackwer:
[FIX] - Fix PHP original PDO_PGSQL::lastInsertId which sometimes complains about a sequence that has not been used it on the transaction (error):

```
Object not in prerequisite state: 7 ERROR:  currval of sequence sequence_name is not yet defined in this session
```


phackwer:
So, no news?

Ocramius:
@phackwer the fix needs testing

phackwer:
Weill, I fixed it on https://github.com/php/php-src/pull/2014. So, this code will be unnecessary from no on.

Ocramius:
@phackwer I assume that we should introduce a `conflict` with anything `7.0.0 - 7.0.9`, since this will be fixed in `7.0.10`?

PR #2334, opened 2016-03-13
Add Polygon type to DBAL types

roroxide:
I needed doctrine to support polygon type, but this type does not implemented in DBAL/Types.
So, I wrote the PolygonType to be more complete.


mvrhov:
IMO, this is mysql specific, so it doesn't belong into core.

Ocramius:
This is indeed MySQL specific and cannot be merged in core. Please try sending a patch to https://github.com/beberlei/doctrineextensions instead (which is MySQL specific)

PR #2336, opened 2016-03-16
[Types] Make unknown type exception a bit friendlier

kix:
This adds a suggested type for cases when someone could have badly referenced a mapping type, yet there is a type that's named closely to what the user originally wanted.

### DIFF
--- a/lib/Doctrine/DBAL/DBALException.php
+++ b/lib/Doctrine/DBAL/DBALException.php
@@ -22,6 +22,7 @@ namespace Doctrine\DBAL;
 use Doctrine\DBAL\Exception;
 use Doctrine\DBAL\Driver;
 use Doctrine\DBAL\Driver\ExceptionConverterDriver;
+use Doctrine\DBAL\Types\Type;
 
 class DBALException extends \Exception
 {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/DBALException.php
```diff
@@ -234,13 +234,21 @@ public static function typeExists($name)
...
-            'to be registered with \Doctrine\DBAL\Types\Type::addType(). You can get a list of all the ' .
-            'known types with \Doctrine\DBAL\Types\Type::getTypesMap(). If this error occurs during database ' .
-            'introspection then you might have forgot to register all database types for a Doctrine Type. Use ' .
-            'AbstractPlatform#registerDoctrineTypeMapping() or have your custom types implement ' .
-            'Type#getMappedDatabaseTypes(). If the type name is empty you might ' .
-            'have a problem with the cache or forgot some mapping information.'
+        $bestGuess = '';
+
+        foreach (array_keys(Type::getTypesMap()) as $typeName) {
+            if (levenshtein($typeName, $name) < 3) {
```
Ocramius: While the intent is good here, shouldn't you look for the closest match? The loop will overwrite the value with non-closest matches here.

In addition to that, test case needs to be added, since we start having some actual non-trivial stuff going on here :-)

kix: Sure, sorry for the lack of tests, maybe I should have opened a PR once the code is in a bit more stabilised/tested state. Anyways, will update this soon.

kix:
Updated. Perhaps, handling a case for `int` → `integer` would be nice as well, but this is a lot less trivial than just using `levenshtein()`.

lib/Doctrine/DBAL/DBALException.php
```diff
@@ -234,13 +235,24 @@ public static function typeExists($name)
...
+
+        foreach (array_keys(Type::getTypesMap()) as $typeName) {
+            $score = levenshtein($typeName, $name, 5, 2, 1);
+            if ($score < $bestScore) {
+                $bestScore = $score;
+                $bestGuess = 'Did you mean "'.$typeName.'"? ';
+            }
+        }
+        
+        return new self('Unknown column type "'.$name.'" requested. '. $bestGuess . PHP_EOL .
```
kix: Is `PHP_EOL` okay here? The string is single-quoted, so no `\n`'s, right?

morozov:
Closing as stale.

PR #2340, opened 2016-03-25
Fix SQL queries numbering in test failure output

guilliamxavier:
For example if the DebugStack SQL logger contains 3 queries, then we want an output like this:

```
3. <query 3>
2. <query 2>
1. <query 1>
```

not like that:

```
4. <query 3>
3. <query 2>
2. <query 1>
```


deeky666:
Merged manually via b4e15f258ac6e95b9f584f520fb5bea94254ff53

Thanks!

PR #2341, opened 2016-03-26
infinite loop in SQLServerPlatform

kokspflanze:
the substr_count is without a space and the stripos is with a space in the check, this ends in a infinite loop.

ErrorLog:
PHP Notice:  String offset cast occurred

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -1222,7 +1222,7 @@ class SQLServerPlatform extends AbstractPlatform
      */
     private function scrubInnerOrderBy($query)
     {
-        $count = substr_count(strtoupper($query), "ORDER BY");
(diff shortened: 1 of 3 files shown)


lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1222,7 +1222,7 @@ protected function doModifyLimitQuery($query, $limit, $offset = null)
      */
     private function scrubInnerOrderBy($query)
     {
-        $count = substr_count(strtoupper($query), "ORDER BY");
+        $count = substr_count(strtoupper($query), " ORDER BY");
```
deeky666: This patch doesn't help if you put a whitespace in front of `ORDER BY` in your query.
Then this clause`OVER( ORDER BY t0_.test2 DESC)` will get converted to this `OVER()` :(

I guess we are hitting limits here without a real SQL parser...

kokspflanze:
https://github.com/doctrine/dbal/pull/2408 is merged so no need for that one

Issue #2330, opened 2016-03-01
doctrine \Doctrine\DBAL\Query\QueryBuilder getSQL crash


shengjie:
Hello, I've a problem while using QueryBuilder.

Following code crashed and nginx served a 502 bad gateway.

```php
$qb = $sqlConnection->createQueryBuilder();
$qb->select('u')->from('User', 'u')->where('u.id = 1')->andWhere('u.email is not null');
echo $qb->getSQL();
```

Doctrine/DBAL: v2.4.5
PHP: 7.0.3

I've investigate a little, found crash happens in `(string) $this->sqlParts['where']` (\Doctrine\DBAL\Query\QueryBuilder:1022)
which calls:

`\Doctrine\DBAL\Query\Expression\CompositeExpression::__toString`

I've put following code
```php
/**
     * Retrieves the string representation of this composite expression.
     *
     * @return string
     */
    public function __toString()
    {
        if (count($this->parts) === 1) {
            return (string) $this->parts[0];
        }
        //MY TEST CODE
        echo $this->parts[0];
        echo $this->parts[1]==$this ? "true": "false";
        die();
        //END MY TEST CODE

        return '(' . implode(') ' . $this->type . ' (', $this->parts) . ')';
    }
```

result: `u.id = 1true`

so i think this is a circular reference problem.

Did i used the query builder wrong ? or is it fixed in a newer version ?

Thanks for any help :)

Ocramius:
Can you please replicate the issue against the [`QueryBuilderTest`](https://github.com/doctrine/dbal/blob/61e1f860a304c3fb6961ecbec36927b5f20b71c6/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php)?

shengjie:
```
shengjie@MacBook-Pro-de-Shengjie:~/dbal$ git status
HEAD detached at v2.4.5
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	mysql.phpunit.xml
	phpunit

nothing added to commit but untracked files present (use "git add" to track)
```

```
shengjie@MacBook-Pro-de-Shengjie:~/dbal$ ./phpunit -c mysql.phpunit.xml tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php 
PHPUnit 3.7.24 by Sebastian Bergmann.

Configuration read from /Users/shengjie/dbal/mysql.phpunit.xml

......Segmentation fault: 11
```

[...]

PR #2343, opened 2016-03-27
doctrine \Doctrine\DBAL\Query\QueryBuilder getSQL crash

jeanCarloMachado:
Refers to https://github.com/doctrine/dbal/issues/2330

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Query/Expression/CompositeExpressionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Query/Expression/CompositeExpressionTest.php
@@ -22,6 +22,13 @@ class CompositeExpressionTest extends \Doctrine\Tests\DbalTestCase
         $this->assertEquals(2, count($expr));
     }
 
+    public function testToString()
+    {
+        $expr = new CompositeExpression(CompositeExpression::TYPE_AND, array('u.username = "foo"', 'u.name = "bar"'));
+
+        static::assertEquals('(u.username = "foo") AND (u.name = "bar")', (string) $expr);
+    }
+
     /**
      * @dataProvider provideDataForConvertToString
      */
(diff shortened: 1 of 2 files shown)


deeky666:
Sorry but where is the difference? I can't see any change in the logic.... It doesn't matter whether you fetch the args or the existing where clause first?

jeanCarloMachado:
Its related to this issue https://github.com/facebook/hhvm/issues/1027. And a fix was already introduced on newer tags with the commit: 7540295 ("HHVM compatibility: func_get_args", 2013-09-15).

tests/Doctrine/Tests/DBAL/Query/Expression/CompositeExpressionTest.php
```diff
@@ -22,6 +22,13 @@ public function testCount()
         $this->assertEquals(2, count($expr));
     }
 
+    public function testToString()
+    {
+        $expr = new CompositeExpression(CompositeExpression::TYPE_AND, array('u.username = "foo"', 'u.name = "bar"'));
+
+        static::assertEquals('(u.username = "foo") AND (u.name = "bar")', (string) $expr);
```
deeky666: Why `static` ?

deeky666:
@jeanCarloMachado can you please redo the PR to target against `master`? We are taking care of backports ourselves then. Thanks.

tests/Doctrine/Tests/DBAL/Query/Expression/CompositeExpressionTest.php
Ocramius: Should indeed be `self::`

deeky666: Yeah but why not `$this` like in every other test, too? Even though it might work, it still looks weird and makes you think something special is going on here... Or am I missing something?

Ocramius:
I just ported this to `master`, but the change was already applied there. Closing as `duplicate`, although I don't know the exact reference of the duplication (although I remember discussing HHVM compatibility with `func_get_args()`)

PR #2346, opened 2016-03-31
WIP Parameter names are normalized to not contain colons

mrclay:
Fetched array keys do not contain colons.
For BC, params/types can be still be accessed via names with colons.
Invalid names result in QueryException.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,15 @@
+## MINOR BC BREAK: Doctrine\DBAL\Query\QueryBuilder
+
+When using ``Doctrine\DBAL\Query\QueryBuilder``, the names of parameters are no longer stored
+with leading colons (``:``), hence the methods ``getParameters()`` and ``getParameterTypes()``
+always return array keys without colons. For back compatibility, parameter names can still be
+given, and accessed individually, with or without a leading colon.
+
+Parameter names are now limited to ``[a-zA-Z0-9_]`` and may not being with a digit. Invalid names
+now throw ``Doctrine\DBAL\Query\QueryException``.
+
+Positional parameters (with integer keys) are not affected.
+
 # Upgrade to 2.5.1
 
 ## MINOR BC BREAK: Doctrine\DBAL\Schema\Table
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -267,9 +267,13 @@ class QueryBuilder
      * @param string|null    $type  One of the PDO::PARAM_* constants.
      *
      * @return \Doctrine\DBAL\Query\QueryBuilder This QueryBuilder instance.
+     *
+     * @throws QueryException
      */
     public function setParameter($key, $value, $type = null)
     {
+        $key = $this->normalizeKey($key);
+
         if ($type !== null) {
             $this->paramTypes[$key] = $type;
         }
@@ -297,11 +301,20 @@ class QueryBuilder
      * @param array $types  The query parameters types to set.
      *
      * @return \Doctrine\DBAL\Query\QueryBuilder This QueryBuilder instance.
+     *
+     * @throws QueryException
      */
     public function setParameters(array $params, array $types = array())
     {
-        $this->paramTypes = $types;
-        $this->params = $params;
+        $this->params = array_combine(
+            array_map([$this, 'normalizeKey'], array_keys($params)),
+            array_values($params)
+        );
+
+        $this->paramTypes = array_combine(
+            array_map([$this, 'normalizeKey'], array_keys($types)),
+            array_values($types)
+        );
 
         return $this;
     }
@@ -322,9 +335,13 @@ class QueryBuilder
      * @param mixed $key The key (index or name) of the bound parameter.
      *
      * @return mixed The value of the bound parameter.
+     *
+     * @throws QueryException
      */
     public function getParameter($key)
     {
+        $key = $this->normalizeKey($key);
+
         return isset($this->params[$key]) ? $this->params[$key] : null;
     }
 
@@ -347,6 +364,8 @@ class QueryBuilder
      */
     public function getParameterType($key)
     {
+        $key = $this->normalizeKey($key);
+
         return isset($this->paramTypes[$key]) ? $this->paramTypes[$key] : null;
     }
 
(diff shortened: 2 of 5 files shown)


tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
```diff
@@ -882,4 +882,55 @@ public function testJoinWithNonUniqueAliasThrowsException()
...
+        $this->assertSame('foo', $qb->getParameter('foo'));
+        $this->assertSame('foo', $qb->getParameter(':foo'));
+
+        $qb->setParameter(':bar', 2, \PDO::PARAM_INT);
+        $this->assertSame(2, $qb->getParameter('bar'));
+        $this->assertSame(2, $qb->getParameter(':bar'));
+
+        $this->assertSame(array(
+            'foo' => 'foo',
+            'bar' => 2,
```
mrclay, line 900: The big question is whether this is BC. If user code _expects_ a `:bar` key, it will fail. execute() handles it of course.

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
```diff
@@ -882,4 +882,55 @@ public function testJoinWithNonUniqueAliasThrowsException()
...
+    public function testRejectsInvalidParamNames()
+    {
+        $qb = new QueryBuilder($this->conn);
+
+        $this->setExpectedException(
+            'Doctrine\DBAL\Query\QueryException',
+            "The given name '1a' is invalid."
+        );
+
+        $qb->setParameter('1a', 'one a');
```
mrclay: The plan here is to use PDO's validation strategy http://stackoverflow.com/a/5810058/3779 but allow string integers like "1" to pass through.

.gitignore
```diff
@@ -4,5 +4,6 @@ reports/
 dist/
 download/
 vendor/
+.idea/
```
Ocramius: This should be in your global `.gitignore` instead

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1329,6 +1348,36 @@ private function getSQLForJoins($fromAlias, array &$knownAliases)
...
+     * @throws QueryException
+     */
+    private function normalizeKey($key)
+    {
+        if (is_int($key)) {
+            return $key;
+        }
+
+        // int as string
+        if (preg_match('~^(0|[1-9]\\d*)\\z~', $key)) {
```
Ocramius: Can be simplified to `$key === (string) (int) $key`

lib/Doctrine/DBAL/Query/QueryException.php
```diff
@@ -51,4 +51,14 @@ static public function nonUniqueAlias($alias, $registeredAliases)
             "in FROM and JOIN clause table. The currently registered " .
             "aliases are: " . implode(", ", $registeredAliases) . ".");
     }
+
+    /**
+     * @param string $name
+     *
+     * @return \Doctrine\DBAL\Query\QueryException
+     */
+    static public function invalidParamName($name)
```
Ocramius, line 60: don't abbreviate `Parameter`

lib/Doctrine/DBAL/Query/QueryException.php
```diff
@@ -51,4 +51,14 @@ static public function nonUniqueAlias($alias, $registeredAliases)
             "in FROM and JOIN clause table. The currently registered " .
             "aliases are: " . implode(", ", $registeredAliases) . ".");
     }
+
+    /**
+     * @param string $name
+     *
+     * @return \Doctrine\DBAL\Query\QueryException
```
Ocramius: `@return self` is also sufficient

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
Ocramius, line 900: Getting parameters is a rare use-case, but we can document the breakage here. This one seems to be the only affected endpoint

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
```diff
@@ -882,4 +882,77 @@ public function testJoinWithNonUniqueAliasThrowsException()
...
+
+        $qb->setParameter('foo', 'foo');
+        $this->assertSame('foo', $qb->getParameter('foo'));
+        $this->assertSame('foo', $qb->getParameter(':foo'));
+
+        $qb->setParameter(':bar', 2, \PDO::PARAM_INT);
+        $this->assertSame(2, $qb->getParameter('bar'));
+        $this->assertSame(2, $qb->getParameter(':bar'));
+
+        $this->assertSame(array(
```
Ocramius: You can use short arrays: we dropped 5.3 support ages ago :-)

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
```diff
@@ -882,4 +882,77 @@ public function testJoinWithNonUniqueAliasThrowsException()
...
+            'foo' => 'foo',
+            'bar' => 2,
+        ), $qb->getParameters());
```
Ocramius: Note that this formatting is not PSR-2 compliant

lib/Doctrine/DBAL/Query/QueryException.php
mrclay: Mind if I send another PR to change all the fluent methods to use `self`?

Ocramius: Not worth the effort, imo, but appreciated :-)

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
mrclay, line 900: FYI I'm already normalizing in my subclass so this isn't a high priority for me. If you'd rather me just update the docs accordingly to avoid BC changes I can do that instead.

Ocramius, line 900: > If you'd rather me just update the docs accordingly to avoid BC changes I can do that instead.

Updating [`UPGRADE.md`](https://github.com/doctrine/dbal/blob/61e1f860a304c3fb6961ecbec36927b5f20b71c6/UPGRADE.md) is sufficient

UPGRADE.md
```diff
@@ -1,3 +1,15 @@
+## MINOR BC BREAK: Doctrine\DBAL\Query\QueryBuilder
```
mrclay: What version do you think this will be in? I figured I'd just leave this out for now.

Ocramius: @mrclay 2.6.0

mrclay:
Technically PDO should accept parameters not only starting with digits, but containing only digits, like `:0` or `:00`. I kinda don't know what to do with this fact. Such keys would probably cause problems in DBAL's execution strategy, but a key like `1a` shouldn't cause a problem (and might be used with this class today).

Strawman: If given a key containing only digits, assume it's positional (like this PR). If that key is prepended with `:` or if PHP will interpret the string as an integer key (e.g. `1e2`), throw.

Positional keys: `1`, `"1"`
Named keys: `"foo"`, `"9lives"`
Rejected: `":1"`, `"1e9"`

PR #2349, opened 2016-04-03
Fluent methods in QueryBuilder are now documented as returning `self`

mrclay:
This allows a subclass to be used fluently without confusing static analysis tools.


Ocramius:
@mrclay thanks!

PR #2357, opened 2016-04-14
Support array binding for custom types

lkm:
At present you are only able to have string and integer arrays as a binding parameter. If you needed any custom parsing of variables using the `Type::convertToDatabaseValue` method that needs to be executed manually on the array before, this also meant that the binding type could not be overridden, in my case of implementing binary guid `\PDO::PARAM_LOB;` needed to be set before sqlite would accept it. 

This PR should hopefully take care of that by testing whether the incoming param is an array and then extract it into separate params.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/SQLParserUtilsTest.php
+++ b/tests/Doctrine/Tests/DBAL/SQLParserUtilsTest.php
@@ -356,6 +356,15 @@ SQLDATA
                 array(1, 2),
                 array(\PDO::PARAM_INT, \PDO::PARAM_INT)
             ),
+            // Test parameters that require a different binding type provided by Type::getBindingType()
+            array(
+                "SELECT * FROM Foo WHERE foo IN (?)",
+                array(array('oof', 'rab')),
+                array('customReverseType'),
+                'SELECT * FROM Foo WHERE foo IN (?, ?)',
+                array('oof', 'rab'),
+                array('customReverseType', 'customReverseType')
+            )
         );
     }
 
--- a/lib/Doctrine/DBAL/SQLParserUtils.php
+++ b/lib/Doctrine/DBAL/SQLParserUtils.php
@@ -98,7 +98,7 @@ class SQLParserUtils
         foreach ($types as $name => $type) {
             ++$bindIndex;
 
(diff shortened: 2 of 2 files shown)


RossMBoswell330:
Any update on this? I'm having the same issue.

deeky666:
@lkm I'm sorry but I don't quite get the issue. Please also provide a testcase (that also might clear things up). Otherwise we can't merge.

lkm:
Sorry @deeky666, I was going to provide a test, it somehow slipped my mind. Hopefully this casts some light on what the change is doing. I assume this would be an issue for any type (even the bundled ones), that requires `convertToPHPValue`

tests/Doctrine/Tests/DBAL/SQLParserUtilsTest.php
```diff
@@ -356,6 +356,15 @@ public function dataExpandListParameters()
                 array(1, 2),
                 array(\PDO::PARAM_INT, \PDO::PARAM_INT)
             ),
+            // Test parameters that require a different binding type provided by Type::getBindingType()
+            array(
+                "SELECT * FROM Foo WHERE foo IN (?)",
+                array(array('oof', 'rab')),
+                array('customReverseType'),
```
deeky666: Excuse me but I still don't get the binding type here. We only have limited types that we map to PDO types, and this one is not supported... How should we map that one to the driver? Or am I missing something here?

lkm: CustomReverseType isn't a binding type, but a custom type that extends `Doctrine\DBAL\Types\Type` and is added using `Type::addType('customReverseType', 'Namespace\To\Types\CustomReverseType');`.  
It's PDO binding type is retrieved by calling `getBindingType()` this happens in `Doctrine\DBAL\Connection->getBindingInfo().  

This is especially important if you both rely on a binding type of PARAM_LOB and have implemented `convertToDatabaseValue`, I'm using it like this:
```
...
 ->setParameter("id", array(binary1, binary2, binaryN), CustomBinaryType::CUSTOMBINARYTYPE);
```

lkm:
@deeky666 Have you had time to look at this?

morozov:
Closing as stale.

PR #2358, opened 2016-04-14
New driver enabling Doctrine to use REST APIs as databases

TobiasHauck:
Just have a lookg at https://github.com/CircleOfNice/DoctrineRestDriver .
It's working fine and adding some new features to Doctrine allowing Doctrine to act as a REST client.

I'd like to add this driver as an official Doctrine driver so that you can configure it by 
'driver: "rest"' instead of 'driver_class: "Circle\DoctrineRestDriver\Driver"'

What do you think about it?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
     ],
     "require": {
         "php": ">=5.4",
-        "doctrine/common": "~2.4"
+        "doctrine/common": "~2.4",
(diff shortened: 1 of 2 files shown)


beberlei:
This should be an external extension to Doctrine

beberlei:
If we were to add all external drivers to the map will be hard for us to maintain, because we must check on the status of external projects and we don't want to do that at this point.

PR #2347, opened 2016-03-31
Add SAP Adaptive Server Enterprise database vendor

Deltachaos:
This PR adds the driver and DBAL for SAP Sybase Adaptive Server Enterprise databases. It distinguishes between versions 12.5 and below (base `ASEPlatform`), 15.0 (`ASE150Platform`), 15.5 (`ASE155Platform`), 15.7 (`ASE157Platform`), and 16.0 (`ASE160Platform`), similar to Microsoft SQL Server.
Because Microsoft SQL Server originally where the same product and then later have diverged i have oriented myself at the Microsoft SQL Server platform. A bit of the code could be copied over one by one, some other is only slightly different and even other is completely different.
The differences are most in the system tables, the basic SQL syntax is the same.

I had to move the conversion for date and time into the platform as well because in Sybase ASE the dates have the same string length every time. For example the date `Jan  1 2012` and `Jan 10 2012`. Both have the same length, for the date with the one digit day there is a space in front (GitHub trim's multiple spaces here). This is something where we are not able to represent such a logic only with the PHP date format methods anymore. We need advanced logic to add this additional space if needed.
Another problem with the date and time handling is that the `sybase_ct` driver only returns dates and times with the precision of minutes and not with seconds (even if ASE can store date times up to milliseconds precision).
The tests are not able to handle this as well. Therefore, i have adjusted them to handle this case correctly.

Currently only a driver for the `sybase_ct` extension is implemented. There is a PDO driver DBLib available that is currently already used by the `PDOSqlsrv` as well. This PDO driver can also be used to connect to a ASE server.

There are currently some TODO's on my list:
- Adding a implementation of the PDO DBLib driver for ASE as well (may this would solve the problem with the datetime precision).
- Implementing the `getAlterTableSQL` method in the platform
- Writing more tests for the platform itself

There are two other PR's related to this PR in other doctrine packages:
- doctrine/DoctrineBundle#532 Add SAP Adaptive Server Enterprise support for explain queries
- doctrine/doctrine2#5752 Changes required for pagination

I hope this addition is appreciated and any feedback/help on the above issues would be welcomed.

PR #2359, opened 2016-04-15
Add selectAliasColumn to AbstractPlatform

Deltachaos:

This change is related to doctrine/doctrine2#5788 as well.
It is required in the platform if the "AS" keyword may differ in platforms (which is probably not very common) and for creating workarounds when using deferred tables to simulate ROWNUM() by identity columns.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -2655,6 +2655,20 @@ abstract class AbstractPlatform
     }
 
     /**
+     * Some platforms need to convert aliases to workaround identity columns in deferred tables.
+     *
+     * @param string $column
+     * @param string $alias
+     * @param array  $mapping
+     *
+     * @return string
+     */
+    public function selectAliasColumn($column, $alias, array $mapping = array())
+    {
+        return $column . ' AS ' . $alias;
+    }
+
+    /**
      * Returns the SQL specific for the platform to get the current date.
      *
      * @return string
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
@@ -1382,4 +1382,9 @@ abstract class AbstractPlatformTestCase extends \Doctrine\Tests\DbalTestCase
             array(array('precision' => 8, 'scale' => 2), 'DOUBLE PRECISION'),
         );
     }
+
+    public function testSelectAliasColumn()
+    {
+        $this->assertSame('foo AS bar', $this->_platform->selectAliasColumn('foo', 'bar'));
+    }
 }


deeky666:
@Deltachaos where is the "use case"?

deeky666:
90% of the grammar we are expecting from SQL-92 so we are only deriving the "special" cases

Deltachaos:
Hi,  please have a look at https://github.com/doctrine/dbal/blob/72478b13a1cf8969d432cd3d6a55bca02eee7137/lib/Doctrine/DBAL/Platforms/ASEPlatform.php#L1161-L1174

The problem is that similar to SQL Server we need to insert into a temporary table and then select from this to emulate limit/offset.
Sadly ASE has no support for ROWNUM(), therefore i havr to use autoincrement identity columns for generating "ROWNUM's" in the temporary table.
ASE is very restrictive in terms of implicit data type conversations. So you cannot insert data of type integer identity into a integer column.
Because of the fact that i have to create a additional identity column for emulating ROWNUM the other columns of the select should be no identity columns.
To ensure this i need to wrap the identity columns into a CONVERT function.
The only possibility i have seen to achieve this (halfway clean) was to call this additional function on the platform in the ORM.
May if the name matches  not realy what the purpose is we could also rename it to something like modifySelectColumn.

Deltachaos:
@deeky666 ping

deeky666:
@Deltachaos I get your problem but this is not at all a solution to your problem I suppose. Sybase ASE does not have a `OFFSET` conecpt, how would you tell the query which records to skip when retrieving "page X"? The only thing I see you could do is retrieve all rows, iterate and paginate at application level (which is is highly inefficient of course) but this is nothing Doctrine supports out of the box I think. Using a temptable + some kind of column type conversion won't give you the support for `OFFSET` or am I missing something?

Deltachaos:
Hi @deeky666 please have a look into ``OraclePlatform`` and ``SQLServerPlatform`` ``doModifyLimitQuery``. Both solve the problem with LIMIT/OFFSET in a similar way like I do here in ASE.
Both do not have native support for LIMIT/OFFSET.

For example lets take this example:

id | color
---------- | ----------
5 | red
6 | orange
7 | yellow
8 | green
9 | blue
10 | purple

Assuming ``SELECT id, color FROM rainbows`` with a LIMIT of 2 and OFFSET of 3. SQL Server Platform transforms it into the following:

```sql
WITH dctrn_cte AS (SELECT id, color FROM rainbows)
  SELECT * FROM (
    SELECT *, ROW_NUMBER() OVER (ORDER BY (SELECT 0)) AS doctrine_rownum FROM dctrn_cte
  ) AS doctrine_tbl
  WHERE doctrine_rownum BETWEEN 4 AND 5 ORDER BY doctrine_rownum ASC
```

So basicly we have a subselect that adds a additional field here called ``doctrine_rownum`` that is computed from the function ``ROW_NUMBER()`` resulting in the follwing derived table:

id | color | doctrine_rownum
---------- | ---------- | ----------
5 | red | 1
6 | orange | 2
7 | yellow | 3
8 | green | 4
9 | blue | 5
10 | purple | 6

Now the wrapping select adds a where condition selecting only rows with ``doctrine_rownum >= 4 and doctrine_rownum <= 5`` which results in the following result set:

id | color | doctrine_rownum
---------- | ---------- | ----------
8 | green | 4
9 | blue | 5

Now lets apply the same example to the ASE platform. Again we take the same query, ``SELECT id, color FROM rainbows`` with a LIMIT of 2 and OFFSET of 3. ASE Platform now would translate this into the following:

```sql
SELECT id, color, doctrine_rownum=identity(10) INTO #dctrn_cte FROM rainbows
SELECT * FROM #dctrn_cte WHERE doctrine_rownum BETWEEN 4 AND 5
DROP TABLE #dctrn_cte
```

Because ASE dont has a ``ROW_NUMBER()`` function that could then be used in a subselect we need to emulate the ``ROW_NUMBER()`` with a identity field that counts up for each row.
So the first part of the query selects the contents from rainbows into the derived table ``#dctrn_cte`` adding the field ``doctrine_rownum`` as identity field. So the table looks like this after this first select:

id | color | doctrine_rownum
---------- | ---------- | ----------
5 | red | 1
6 | orange | 2
7 | yellow | 3
8 | green | 4
9 | blue | 5
10 | purple | 6

Now the second part is easy. We select only the rows with ``doctrine_rownum >= 4 and doctrine_rownum <= 5``. Then we get exactly the same result as for SQL Server.

Now if this is the solution why do we need this PR? Because the above query in this form would not work if the field ``id`` is a identity column as well. Lets have a look why:

In the query we have used to emulate the ``ROW_NUMBER()`` the types are defined implicitly. ``doctrine_rownum`` is a ``INT IDENTITY`` because ``identity(10)`` returns a value of this type. The other columns of course inherit the type defined in the rainbow table.
```sql
SELECT id, color, doctrine_rownum=identity(10) INTO #dctrn_cte FROM rainbows
```
Now lets assume the rainbows table would have the follwing types:

column | type
---------- | ----------
id | UNSIGNED INT IDENTITY
color | VARCHAR(50)

If we now add the ``doctrine_rownum`` field here this would result in the following types for the derived table.

column | type
---------- | ----------
id | UNSIGNED INT IDENTITY
color | VARCHAR(50)
doctrine_rownum | INT IDENTITY

The table now would have two identity columns while it is only allowed to have one identity column per table. This is what the sever then complains.
Because we are only interested in the data of the ``id`` column in the select we can workaround this problem by changing the type of the ``id`` column to ``UNSIGNED INT`` instead of ``UNSIGNED INT IDENTITY``. The only way to do this is to apply a ``CONVERT`` function to the identity columns of the select.
```sql
SELECT CONVERT(id, UNSIGNED INT) AS id, color, doctrine_rownum=identity(10) INTO #dctrn_cte FROM rainbows
```
So the final query to solve the LIMIT/OFFSET problem in ASE would look like this:

```sql
SELECT CONVERT(id, UNSIGNED INT) AS id, color, doctrine_rownum=identity(10) INTO #dctrn_cte FROM rainbows
SELECT * FROM #dctrn_cte WHERE doctrine_rownum BETWEEN 4 AND 5
DROP TABLE #dctrn_cte
```
So the reason for this PR is that i need some kind of method to add the ``CONVERT`` function to the query. The cleanest way of doing it that i have found was to add the ``AS`` keyword to the platform so that i can wrap the columns that are identity columns into a ``CONVERT`` in the platform class.

deeky666:
@Deltachaos thanks for the detailed explanation. I know understand how it is supposed to work (I think). This is tricky but what you are proposing here in this PR is misleading and in fact you are using it for something that the method does not describe. What you need is more of a specialized convert method sort of. This has nothing to do with aliasing in the first place. I doubt this is something we want to have in the abstract platform especially as the method requires mapping information for "aliasing". What kind of mapping information is this even, it looks like something ORM has built together?
Unfortunately I don't have a good solution for this.
Btw your example involves executing 3 queries if I understand correctly: `SELECT`, `SELECT`, `DROP`. How is that supposed to work? You will need to generate 1 query that returns data otherwise Doctrine can't deal with it.

Deltachaos:
@deeky666 Year you are right. The method name is missleading in this case. May we should name it something like `convertToPHPValueSQL` because it is similar to what is done for types that have enabled ``requireSQLConversion`` only that this convert is not only needed for special types but for the whole platform.
The mapping information passed here is in the standard mapping for a single field that was build by ORM.
The example contains 3 queries thats true. But execution multiple queries with is is a valid syntax and very common in ASE (especially without seperation). For the ORM it is one query, even if it contains multiple ones.

Deltachaos:
@deeky666 here again. Do you think naming the method ``convertToPHPValueSQL`` would be ok?

Deltachaos:
Seems not to be required anymore.

PR #2361, opened 2016-04-15
Add getMaxIndexColumns to AbstractPlatform

Deltachaos:

Some platforms have a maximum number of columns that can be used in one index.
This feature can be used later on in the ORM to create better error messages when a index on more then allowed fields should be created.
One the other hand this value may changes in different versions of a platform and then can be used in other classes to receive the maximum number of index columns and do whatever stuff with it.

In the ASE implementation i use it in the platform `getColumnConstraintSQL`.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -3401,6 +3401,16 @@ abstract class AbstractPlatform
     }
 
     /**
+     * Maximum number of columns in a index. If unlimited return 0
+     *
+     * @return integer
+     */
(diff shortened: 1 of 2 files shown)


deeky666:
I don't think we should add this information to the code and rather let the database vendor handle that. The thing here is that there is no real benefit in validating the max number of index columns in the code vs letting the database server validate it. We cannot provide a "workaround" to come around the particular vendor limitation anyways. Furthermore this boundary is not only highly vendor specific but also dependant on other metadata like column length of string columns in MySQL for example.
So I'm closing this for now until there is a use case where we gain real benefit and a way to abstract it properly.

PR #2362, opened 2016-04-15
Move DateTime conversion to AbstractPlatform to be more flexible

Deltachaos:

For the most platforms the standard PHP date format patterns are enough for creating compliant date strings. But some platforms have date format's that have a fixed length and filled up with spaces between parts where the length of the replacement differ (like for patterns: n, F or z).
To achieve a correct conversion of those formats we need more logic then just a PHP format pattern.

In my case ASE has very strange date formats. It uses `Aug 24 1998  5:36PM` or `Aug  4 1998 11:36PM` where the length of this date is always the same. It is filled up with spaces to stretch the string to the defined length.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -2655,6 +2655,124 @@ abstract class AbstractPlatform
     }
 
     /**
+     * Some platforms have datetime strings that needs to be correctly converted
+     *
+     * The default conversion tries to convert value into string with the defined time format
+     *
+     * @param \DateTime $value
+     *
+     * @return string
+     */
+    public function convertDateTimeToDatabaseValue(\DateTime $value)
+    {
+        return $value->format($this->getDateTimeFormatString());
+    }
+
+    /**
+     * Some platforms have datetime strings that needs to be correctly converted
+     *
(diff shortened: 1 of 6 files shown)


Ocramius:
@deeky666 overall, change looks good, and surely better than what we have now. Not sure if we want more methods on the abstract platform though...

Deltachaos:
@Ocramius the tests `Doctrine\Tests\DBAL\Types\DateTest`, `Doctrine\Tests\DBAL\Types\DateTimeTest`, `Doctrine\Tests\DBAL\Types\DateTimeTzTest`, `Doctrine\Tests\DBAL\Types\TimeTest` covers the changes in this PR so the code is tested.

Where sould the conversions be places instead when not into the `AbstractPlatform`? I have oriented at the `Doctrine\DBAL\Types\BooleanType` which does it the same way.

Ocramius:
> covers the changes in this PR so the code is tested.

A few tests need to be in place to verify that the abstract platform actually has control over the conversion

Deltachaos:
@Ocramius i have added the tests

Deltachaos:
@deeky666 @Ocramius any updates on this? Are the tests sufficient?

Deltachaos:
@deeky666 ping

Deltachaos:
@deeky666 thanks to merging master again the tests went back to green.

Deltachaos:
Solved it using Type::overrideType.

PR #2363, opened 2016-04-15
Add test case scenario for `LIMIT`/`OFFSET` when selecting from a sub-`SELECT`

Deltachaos:

Some platforms like ASE and SQL Server need to do a lot of modifications to the queries to simulate LIMIT/OFFSET. Selecting from a subselect is one more case that can cause problems here.
Therefore a tests should cover this case.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/ModifyLimitQueryTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/ModifyLimitQueryTest.php
@@ -110,6 +110,20 @@ class ModifyLimitQueryTest extends \Doctrine\Tests\DbalFunctionalTestCase
         $this->assertLimitResult(array(2, 1), $sql, 2, 2, false);
     }
 
+    public function testModifyLimitQueryFromSubSelect()
+    {
+        $this->_conn->insert('modify_limit_table', array('test_int' => 1));
+        $this->_conn->insert('modify_limit_table', array('test_int' => 2));
+        $this->_conn->insert('modify_limit_table', array('test_int' => 3));
+        $this->_conn->insert('modify_limit_table', array('test_int' => 4));
+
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/DBAL/Functional/ModifyLimitQueryTest.php
```diff
@@ -110,6 +110,20 @@ public function testModifyLimitQuerySubSelect()
...
+    public function testModifyLimitQueryFromSubSelect()
+    {
+        $this->_conn->insert('modify_limit_table', array('test_int' => 1));
+        $this->_conn->insert('modify_limit_table', array('test_int' => 2));
+        $this->_conn->insert('modify_limit_table', array('test_int' => 3));
+        $this->_conn->insert('modify_limit_table', array('test_int' => 4));
+
+        $sql = "SELECT * FROM (SELECT * FROM modify_limit_table) sub";
+
+        $this->assertLimitResult(array(4, 3, 2, 1), $sql, 10, 0, false);
```
deeky666: No, no, no this doesn't make sense without and order by. While I appreciate your efforts, this is just plain undeterministic. I know the "old" tests are not the best, but please let's not introduce more....

Deltachaos: @deeky666 i will add an order by

deeky666: @Deltachaos thanks :) Just complaining because we had some trouble with those tests on some platforms when there was no order by present (which makes a lot of sense actually).

Deltachaos:
@deeky666 i have changed the query and assetLimitResult call to use a ordered result

deeky666:
@Deltachaos thanks!

Deltachaos:
@deeky666 thanks for merging

PR #2364, opened 2016-04-15
Display further contextual information on failed functional tests

Deltachaos:

This is useful if you have arrays as parameter to a query to see actual value.


deeky666:
@Deltachaos thanks!

Deltachaos:
@deeky666 thanks for merging

PR #2365, opened 2016-04-15
Use platform agnostic date time format in access and write tests

Deltachaos:

Some platforms cannot handle the date format that is specified in the tests at all and have a very different one.
The tests should only validate if the persistence and receival of the data works. For this purpose they dont need fixed date time formats and can rely on the platform to convert the data to the correct format.

This depends on doctrine/dbal#2362


Deltachaos:
@deeky666 now the tests are not failing anymore

Deltachaos:
Solved it using ``Type::overrideType``.

PR #2366, opened 2016-04-18
Add more options to TestUtil

Deltachaos:

Some drivers (for example sybase_ct or sqlanywhere) can use a servername instead of a hostname to connect to a database server.
Also some platforms need aditional configuration for working with the database server correctly. Like they need to know about given devices etc.

### DIFF
--- a/tests/Doctrine/Tests/TestUtil.php
+++ b/tests/Doctrine/Tests/TestUtil.php
@@ -61,16 +61,28 @@ class TestUtil
             $GLOBALS['db_type'],
             $GLOBALS['db_username'],
             $GLOBALS['db_password'],
-            $GLOBALS['db_host'],
-            $GLOBALS['db_name'],
-            $GLOBALS['db_port']
+            $GLOBALS['db_name']
+        )
+        && (
+            isset(
+                $GLOBALS['db_host'],
+                $GLOBALS['db_port']
+            ) || isset(
+                $GLOBALS['db_server']
+            )
         )
         && isset(
             $GLOBALS['tmpdb_type'],
             $GLOBALS['tmpdb_username'],
-            $GLOBALS['tmpdb_password'],
-            $GLOBALS['tmpdb_host'],
-            $GLOBALS['tmpdb_port']
+            $GLOBALS['tmpdb_password']
+        )
+        && (
+            isset(
+                $GLOBALS['tmpdb_host'],
+                $GLOBALS['tmpdb_port']
+            ) || isset(
+                $GLOBALS['tmpdb_server']
+            )
         );
     }
 
@@ -140,10 +152,16 @@ class TestUtil
             'driver' => $GLOBALS['tmpdb_type'],
             'user' => $GLOBALS['tmpdb_username'],
             'password' => $GLOBALS['tmpdb_password'],
-            'host' => $GLOBALS['tmpdb_host'],
-            'dbname' => null,
-            'port' => $GLOBALS['tmpdb_port']
+            'dbname' => null
         );
+
+        if (isset($GLOBALS['tmpdb_host'])) {
+            $connectionParams['host'] = $GLOBALS['tmpdb_host'];
+        }
+
+        if (isset($GLOBALS['tmpdb_port'])) {
+            $connectionParams['port'] = $GLOBALS['tmpdb_port'];
+        }
 
         if (isset($GLOBALS['tmpdb_name'])) {
             $connectionParams['dbname'] = $GLOBALS['tmpdb_name'];
@@ -157,6 +175,10 @@ class TestUtil
             $connectionParams['unix_socket'] = $GLOBALS['tmpdb_unix_socket'];
         }
 
+        if (isset($GLOBALS['tmpdb_platform_options'])) {
+            $connectionParams['platform_options'] = json_decode($GLOBALS['tmpdb_platform_options'], true);
+        }
+
         return $connectionParams;
     }
 
@@ -166,10 +188,16 @@ class TestUtil
             'driver' => $GLOBALS['db_type'],
             'user' => $GLOBALS['db_username'],
             'password' => $GLOBALS['db_password'],
-            'host' => $GLOBALS['db_host'],
-            'dbname' => $GLOBALS['db_name'],
-            'port' => $GLOBALS['db_port']
+            'dbname' => $GLOBALS['db_name']
         );
+
+        if (isset($GLOBALS['db_host'])) {
+            $connectionParams['host'] = $GLOBALS['db_host'];
+        }
+
+        if (isset($GLOBALS['db_port'])) {
+            $connectionParams['port'] = $GLOBALS['db_port'];
+        }
 
         if (isset($GLOBALS['db_server'])) {
             $connectionParams['server'] = $GLOBALS['db_server'];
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/TestUtil.php
```diff
@@ -61,16 +61,28 @@ private static function hasRequiredConnectionParams()
...
-            $GLOBALS['db_name'],
-            $GLOBALS['db_port']
+            $GLOBALS['db_name']
+        )
+        && (
+            isset(
+                $GLOBALS['db_host'],
+                $GLOBALS['db_port']
+            ) || isset(
+                $GLOBALS['db_server']
```
deeky666: I think this is not true. `server` alone is not sufficient. I remember when implementing SQL Anywhere, the `server` parameter was sometimes required but is not sufficient without `host` and `port`. Because it is another abstraction beneath `host`. Not sure why `server` parameter is not in here though because I know that I used it for testing purposes and it worked,

deeky666: @Deltachaos ping

Deltachaos: @deeky666 Sorry, for the delay. The sybase_ct `sybase_connect` function only accepts a server name. The further information to connect is then read from a file called "interfaces" where you can also specify information for connection encryption etc.
This is the standard behavior of working with Sybase ASE and Sybase IQ servers (when using the Sybase CT library provided by the database vendor)

deeky666: @Deltachaos I get that but this class is meant to be able to deal with all vendor connection. To me it sounds as if `server` is actually `host` (Doctrine wise; abstracted) because how can you simply connect to a database server only providing an arbitrary "server name" without the client knowing the host address? Or is the concept like ODBC where you define a data source name (`server`) which then resolves params like `host`, `port` etc.? Because then I would map Sybase's `server` param to Doctrine's `host` param in the driver. This is the most convenient way IMO (see Doctrine 1 with `dblib` driver for example). We cannot manage all the custom vendor implementation connection parameters and need to find common denominator for the most crucial ones. Just have a look at the Oracle drivers and you'll get the idea.

Hope this helps (tending to close this as this seems to be a misunderstanding of how DBAL works and need driver fixing).

Deltachaos: @deeky666 yes it is like ODBC. I will change this to use the host parameter instead.

Deltachaos:
@deeky666 now only the platform_options parameter is left. The problem is that for ASE i need to pass information about devices to the driver/platform. I did not have seen any other way then adding a additional paramter here in the TestUtil

deeky666:
@Deltachaos can you please clarify what the information exactly is that you need to pass through and what impact/purpose it has for DBAL driver/platform?

Deltachaos:
Hi @deeky666 please have a look into ``OraclePlatform`` and ``SQLServerPlatform`` ``doModifyLimitQuery``. Both solve the problem with LIMIT/OFFSET in a similar like I do here in ASE.
Both do not have support for LIMIT/OFFSET.

For example lets take this example:

id | color
---------- | ----------
5 | red
6 | orange
7 | yellow
8 | green
9 | blue
10 | purple

Assuming ``SELECT id, color FROM rainbows`` with a LIMIT of 2 and OFFSET of 3. SQL Server Platform transforms it into the following:

```sql
WITH dctrn_cte AS (SELECT id, color FROM rainbows)
  SELECT * FROM (
    SELECT *, ROW_NUMBER() OVER (ORDER BY (SELECT 0)) AS doctrine_rownum FROM dctrn_cte
  ) AS doctrine_tbl
  WHERE doctrine_rownum BETWEEN 4 AND 5 ORDER BY doctrine_rownum ASC
```

So basicly we have a subselect that adds a additional field here called ``doctrine_rownum`` that is computed from the function ``ROW_NUMBER()`` resulting in the follwing derived table:

id | color | doctrine_rownum
---------- | ---------- | ----------
5 | red | 1
6 | orange | 2
7 | yellow | 3
8 | green | 4
9 | blue | 5
10 | purple | 6

Now the wrapping select adds a where condition selecting only rows with ``doctrine_rownum >= 4 and doctrine_rownum <= 5`` which results in the following result set:

id | color | doctrine_rownum
---------- | ---------- | ----------
8 | green | 4
9 | blue | 5

Now lets apply the same example to the ASE platform. Again we take the same query, ``SELECT id, color FROM rainbows`` with a LIMIT of 2 and OFFSET of 3. ASE Platform now would translate this into the following:

```sql
SELECT id, color, doctrine_rownum=identity(10) INTO #dctrn_cte FROM rainbows
SELECT * FROM #dctrn_cte WHERE doctrine_rownum BETWEEN 4 AND 5
DROP TABLE #dctrn_cte
```

Because ASE dont has a ``ROW_NUMBER()`` function that could then be used in a subselect we need to emulate the ``ROW_NUMBER()`` with a identity field that counts up for each row.
So the first part of the query selects the contents from rainbows into the derived table ``#dctrn_cte`` adding the field ``doctrine_rownum`` as identity field. So the table looks like this after this first select:

id | color | doctrine_rownum
---------- | ---------- | ----------
5 | red | 1
6 | orange | 2
7 | yellow | 3
8 | green | 4
9 | blue | 5
10 | purple | 6

Now the second part is easy. We select only the rows with ``doctrine_rownum >= 4 and doctrine_rownum <= 5``. Then we get exactly the same result as for SQL Server.

Now if this is the solution why do we need this PR? Because the above query in this form would not work if the field ``id`` is a identity column as well. Lets have a look why:

In the query we have used to emulate the ``ROW_NUMBER()`` we only specified the type for the additional field ``doctrine_rownum`` because for the other columns the type is already defined by the original select.
```sql
SELECT id, color, doctrine_rownum=identity(10) INTO #dctrn_cte FROM rainbows
```
Now lets assume the rainbows table would have the follwing types:

column | type
---------- | ----------
id | UNSIGNED INT IDENTITY
color | VARCHAR(50)

If we now add the ``doctrine_rownum`` field here this would result in the following types for the derived table.

column | type
---------- | ----------
id | UNSIGNED INT IDENTITY
color | VARCHAR(50)
doctrine_rownum | INT IDENTITY

The table now would have two identity columns while it is only allowed to have one identity column per table. This is what the sever then complains.
Because we are only interested in the data of the ``id`` column in the select we can workaround this problem by changing the type of the ``id`` column to ``UNSIGNED INT`` instead of ``UNSIGNED INT IDENTITY``. The only way to do this is to apply a ``CONVERT`` function to the identity columns of the select.
```sql
SELECT CONVERT(id, UNSIGNED INT) AS id, color, doctrine_rownum=identity(10) INTO #dctrn_cte FROM rainbows
```
So the final query to solve the LIMIT/OFFSET problem in ASE would look like this:

```sql
SELECT CONVERT(id, UNSIGNED INT) AS id, color, doctrine_rownum=identity(10) INTO #dctrn_cte FROM rainbows
SELECT * FROM #dctrn_cte WHERE doctrine_rownum BETWEEN 4 AND 5
DROP TABLE #dctrn_cte
```
So the reason for this PR is that i need some kind of method to add the ``CONVERT`` function to the query. The cleanest way of doing it that i have found was to add the ``AS`` keyword to the platform so that i can wrap the columns that are identity columns into a ``CONVERT`` in the platform class.

Deltachaos:
@deeky666 for ASE we have to specify which the devices for a CREATE DATABASE. The tests need to recreate a database again and again. To be able to run them we need to pass this options somehow.

Deltachaos:
Not required anymore

PR #2367, opened 2016-04-18
Add fix for broken decimals

Deltachaos:

Some drivers (for example sybase_ct) return decimal types as float and not as string. The tests currently expect a decimal to be a string.

 I did not had any other idea to work around this, then converting the float to a string in the decimal type. Any better solutions are very appreciated.

### DIFF
--- a/lib/Doctrine/DBAL/Types/DecimalType.php
+++ b/lib/Doctrine/DBAL/Types/DecimalType.php
@@ -49,6 +49,10 @@ class DecimalType extends Type
      */
     public function convertToPHPValue($value, AbstractPlatform $platform)
     {
+        if (is_float($value)) {
+            return (string) $value;
+        }
+
         return $value;
     }
 }


kimhemsoe:
I think the driver should try to fix this. 

Changing the type does not fix that the driver still return the wrong type. Sounds like we are missing a few tests here :)

Deltachaos:
Has the type of a decimal to be a string in PHP? I think this is also the more interesting question here.

kimhemsoe:
It needs to be a string to keep precision 

ex. float(1.1) is far from exact "1.1"

Deltachaos:
But as soon as you try to use the value for calculation it will be converted to a float anyway?

kimhemsoe:
There is ways to get around that like gmp.
But that is not our problem, we just need to keep the data exact.

deeky666:
@kimhemsoe said already what it's about. Decimal type covers the exact precision and is not considered `float`. As PHP and other languages don't have that kind of type, we just pass the value through. You need to fix it in the driver by specifying `string` type.

PR #2378, opened 2016-04-29
Do not skip indkey with zero value

dxops:
This is an array of indnatts values that indicate which table columns this index indexes. For example a value of 1 3 would mean that the first and the third table columns make up the index key. **A zero in this array indicates that the corresponding index attribute is an expression over the table columns, rather than a simple column reference.**

Column name used in index should match to index name, as we dont have any other values for column name and we require it https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/Index.php#L98

Use case is

```
CREATE INDEX value ON oro_search_index_text USING gin(to_tsvector('english', 'value'))
```


deeky666:
This requires a testcase.

PR #2382, opened 2016-05-05
Remove arbitrary line number from github link in the caching documentation

jacobhenke:


jacobhenke:
Actually, I think I may be misunderstanding what the documentation is saying?

Ocramius:
The issue here is that the linking was done against `master`, and not a fixed commit. The line may have moved by now...

I think the patch is fine/sufficient tho, merging.

jacobhenke:
Yay! My first official open source commit! haha

PR #2383, opened 2016-05-09
Add method to set change real cache key prefix

superdav42:
This makes it possible to invalidate the cache when changes outside of parameters or sql happen, such as schema changes.


superdav42:
Turns out this is not a bug in doctrine but with our own custom QueryCacheProfile we were using.

PR #2385, opened 2016-05-10
Allow unsigned boolean declarations

succinct:
Update getBooleanTypeDeclarationSQL() to allow UNSIGNED

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -289,7 +289,9 @@ class MySqlPlatform extends AbstractPlatform
      */
     public function getBooleanTypeDeclarationSQL(array $field)
     {
-        return 'TINYINT(1)' . $this->_getCommonIntegerTypeDeclarationSQL($field);;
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -289,7 +289,7 @@ public function getTimeTypeDeclarationSQL(array $fieldDeclaration)
      */
     public function getBooleanTypeDeclarationSQL(array $field)
     {
-        return 'TINYINT(1)';
+        return 'TINYINT(1)' . $this->_getCommonIntegerTypeDeclarationSQL($field);;
```
deeky666: This is akward on MySQL. In the past we used to decline this kind of request. But I get the desire. Please just use `'UNSIGNED'` here. No need to evaluate any other properties of integer type columns here. Also declaring `unsigned` property on boolean type columns should not be allowed. Also please add a testcase for this. Thanks.

succinct:
I decided it's better to fix our schema to make all boolean signed, rather than bother with this change.

PR #2386, opened 2016-05-11
Fix oci driver bindValue overwrite with php7

soyuka:
Follow up on #2262


deeky666:
@soyuka I'm sorry but with PHP7 when I'm executing the tests I get the following for both master and your PR:

```
deeky@98N-MUELLER-STEVE:~/dev/doctrine/dbal-remote$ phpunit -c oci8.phpunit.xml 
PHPUnit 4.8.24 by Sebastian Bergmann and contributors.

...............................................SSS...........   61 / 3858 (  1%)
.SS..............S.......SS...S.......SSS...............SS...  122 / 3858 (  3%)
........SSS..............SS...SSS.........SS...S.............  183 / 3858 (  4%)
.S.....S........................................SSSS.......S.  244 / 3858 (  6%)
............................................................F  305 / 3858 (  7%)
.....S.SS......SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS............SS.  366 / 3858 (  9%)
...S.SSSSSS..........................SSSSSSSSSSSSSSSSSSSSSSSS  427 / 3858 ( 11%)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS  488 / 3858 ( 12%)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS.....E.  549 / 3858 ( 14%)
.S.....S.......SS.........S...........SSSSSSSSSSSSSSSSSSSSSSS  610 / 3858 ( 15%)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS  671 / 3858 ( 17%)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS  732 / 3858 ( 18%)
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS..SS  793 / 3858 ( 20%)
S..SS.SSSSSSSSSSS.........................S..................  854 / 3858 ( 22%)
.............................................................  915 / 3858 ( 23%)
..............I........SSS...................................  976 / 3858 ( 25%)
............................................................. 1037 / 3858 ( 26%)
................S............................................ 1098 / 3858 ( 28%)
............................................................. 1159 / 3858 ( 30%)
......S...................................................... 1220 / 3858 ( 31%)
............................................................. 1281 / 3858 ( 33%)
..I........SSS............................................... 1342 / 3858 ( 34%)
............................................................. 1403 / 3858 ( 36%)
..............................................SSS............ 1464 / 3858 ( 37%)
............................................................. 1525 / 3858 ( 39%)
............................................................. 1586 / 3858 ( 41%)
.....................SSS..................................... 1647 / 3858 ( 42%)
............................................................. 1708 / 3858 ( 44%)
..........................................................SSS 1769 / 3858 ( 45%)
............................................................. 1830 / 3858 ( 47%)
............................................................. 1891 / 3858 ( 49%)
..................................SSS........................ 1952 / 3858 ( 50%)
............................................................. 2013 / 3858 ( 52%)
............................................................. 2074 / 3858 ( 53%)
........SSS....................S............................. 2135 / 3858 ( 55%)
............................................................. 2196 / 3858 ( 56%)
.............................I........SSS..................S. 2257 / 3858 ( 58%)
.....S....................................................... 2318 / 3858 ( 60%)
............................................................. 2379 / 3858 ( 61%)
I........SSS....................S.....S...................... 2440 / 3858 ( 63%)
............................................................. 2501 / 3858 ( 64%)
.................................I........SSS................ 2562 / 3858 ( 66%)
............................................................. 2623 / 3858 ( 67%)
............................................................I 2684 / 3858 ( 69%)
........SSS.................................................. 2745 / 3858 ( 71%)
............................................................. 2806 / 3858 ( 72%)
..............I........SSS................................... 2867 / 3858 ( 74%)
............................................................. 2928 / 3858 ( 75%)
.................................I........SSS................ 2989 / 3858 ( 77%)
.....................S......S..S.S........................... 3050 / 3858 ( 79%)
.....................................................I....... 3111 / 3858 ( 80%)
.SSS.......................................II................ 3172 / 3858 ( 82%)
...........SSS......................S........S............... 3233 / 3858 ( 83%)
............................................................. 3294 / 3858 ( 85%)
............................................................. 3355 / 3858 ( 86%)
............................................................. 3416 / 3858 ( 88%)
............................................S................ 3477 / 3858 ( 90%)
............................................................. 3538 / 3858 ( 91%)
............................................................. 3599 / 3858 ( 93%)
.....................................................S..SSS.. 3660 / 3858 ( 94%)
............................................................. 3721 / 3858 ( 96%)
............................................................. 3782 / 3858 ( 98%)
............................................................. 3843 / 3858 ( 99%)
...............

Time: 2.04 minutes, Memory: 96.50Mb

There was 1 error:

1) Doctrine\Tests\DBAL\Functional\Schema\OracleSchemaManagerTest::testDropsDatabaseWithActiveConnections
Exception: [Doctrine\DBAL\Exception\DriverException] An exception occurred while executing 'CREATE USER test_drop_database IDENTIFIED BY sql':

ORA-65096: invalid common user or role name

With queries:
3. SQL: 'CREATE USER test_drop_database IDENTIFIED BY sql' Params: 
2. SQL: 'DROP USER test_drop_database CASCADE' Params: 

Trace:
/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/DBALException.php:142
/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/DBALException.php:116
/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/Connection.php:988
/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/Schema/OracleSchemaManager.php:306
/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php:596
/home/deeky/dev/doctrine/dbal-remote/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php:53
/home/deeky/dev/doctrine/dbal-remote/vendor/phpunit/phpunit/src/Framework/TestCase.php:909
/home/deeky/dev/doctrine/dbal-remote/vendor/phpunit/phpunit/src/Framework/TestCase.php:768
/home/deeky/dev/doctrine/dbal-remote/vendor/phpunit/phpunit/src/Framework/TestResult.php:612
/home/deeky/dev/doctrine/dbal-remote/vendor/phpunit/phpunit/src/Framework/TestCase.php:724
/home/deeky/dev/doctrine/dbal-remote/vendor/phpunit/phpunit/src/Framework/TestSuite.php:747
/home/deeky/dev/doctrine/dbal-remote/vendor/phpunit/phpunit/src/Framework/TestSuite.php:747
/home/deeky/dev/doctrine/dbal-remote/vendor/phpunit/phpunit/src/TextUI/TestRunner.php:440


/home/deeky/dev/doctrine/dbal-remote/tests/Doctrine/Tests/DbalFunctionalTestCase.php:73

Caused by
Doctrine\DBAL\Exception\DriverException: An exception occurred while executing 'CREATE USER test_drop_database IDENTIFIED BY sql':

ORA-65096: invalid common user or role name

/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php:76
/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/DBALException.php:142
/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/DBALException.php:116
/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/Connection.php:988
/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/Schema/OracleSchemaManager.php:306
/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php:596
/home/deeky/dev/doctrine/dbal-remote/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php:53

Caused by
Doctrine\DBAL\Driver\OCI8\OCI8Exception: ORA-65096: invalid common user or role name

/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/Driver/OCI8/OCI8Exception.php:33
/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php:214
/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/Driver/OCI8/OCI8Connection.php:140
/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/Connection.php:985
/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/Schema/OracleSchemaManager.php:306
/home/deeky/dev/doctrine/dbal-remote/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php:596
/home/deeky/dev/doctrine/dbal-remote/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php:53

--

There was 1 failure:

1) Doctrine\Tests\DBAL\Functional\DataAccessTest::testDateArithmetics
Date difference should be approx. -2324 days.
Failed asserting that -2322 matches expected -2324.0.

/home/deeky/dev/doctrine/dbal-remote/tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php:524

FAILURES!
Tests: 3858, Assertions: 7704, Errors: 1, Failures: 1, Skipped: 501, Incomplete: 11.
```

The failures are not related. Looks like the tests don't reveal the issue? Maybe I am wrong here....

deeky666:
Also please don't add a travis build config file as long as we don't have the possibility to test Oracle on travis. Thanks you.

soyuka:
Failures are not related IMO, I'll remove the travis file indeed, wasn't supposed to commit it. However I'm still seeing issues (#2387), but I haven't managed to reproduce them through tests yet!

Won't it be possible to add a docker service to travis so that Oracle can also be tested?

deeky666:
@soyuka  we tried things like that but it's not that easy unfortunately and even if we could get it going we would have licensing issues :( so for this PR please try to reproduce it on your machine and add a testcase. I can test it locally and will merge as soon as I can see the actual issue is solved. Thanks!

soyuka:
I've removed the travis oracle configuration, and I've managed to reproduce #2387 which is related to how LOB null values are handled. Current error for [this test](https://github.com/doctrine/dbal/pull/2386/files#diff-26d11fc87bf8c177daa3204315d96683R77) is:

```
Doctrine\DBAL\Exception\DriverException: An exception occurred while executing 'INSERT INTO DBAL2386 VALUES (:id, :title, :address, :geo, :name)' with params [1, null, null, null, null]:
ORA-24816: Expanded non LONG bind data supplied after actual LONG or LOB column
```

 However, I'm not sure where I should fix this:
- should `convertToDatabaseValue` handle null values differently when it comes to LOB types?
- I assume null values can be inserted correctly because when executed directly in plsql the query (binded or non-binded) works just fine
- is it related to the fact that the `bindParam` `&$variable` references NULL, and so the bug might come from the oci8 extension or another php7 behavior that I'm not aware of?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -145,10 +158,7 @@ public function bindParam($column, &$variable, $type = null, $length = null)
         $column = isset($this->_paramMap[$column]) ? $this->_paramMap[$column] : $column;
 
         if ($type == \PDO::PARAM_LOB) {
-            $lob = oci_new_descriptor($this->_dbh, OCI_D_LOB);
-            $lob->writeTemporary($variable, OCI_TEMP_BLOB);
-
```
soyuka: If I try to write temporary from the reference, it'll fail with the following: `ORA-22275: invalid LOB locator specified`.

WDYT about throwing an error if the `$variable` isn't an OCI-Lob instance?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -134,7 +139,15 @@ static public function convertPositionalToNamedPlaceholders($statement)
      */
     public function bindValue($param, $value, $type = null)
     {
-        return $this->bindParam($param, $value, $type, null);
+        if ($type === \PDO::PARAM_LOB || -99 === $type) {
```
Ocramius: This `-99` can't stay this way. If it is a magic number provided by the platform, then it needs to be documented and made a class constant instead

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -144,12 +157,9 @@ public function bindParam($column, &$variable, $type = null, $length = null)
     {
         $column = isset($this->_paramMap[$column]) ? $this->_paramMap[$column] : $column;
 
-        if ($type == \PDO::PARAM_LOB) {
-            $lob = oci_new_descriptor($this->_dbh, OCI_D_LOB);
-            $lob->writeTemporary($variable, OCI_TEMP_BLOB);
-
-            return oci_bind_by_name($this->_sth, $column, $lob, -1, OCI_B_BLOB);
-        } elseif ($length !== null) {
+        if (is_a($variable, 'Oci-Lob')) {
```
Ocramius: `Oci-Lob` can't stay as magic constant. Also, seems weird that the OCI extension would provide a type with a dash in it.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -72,6 +72,11 @@ class OCI8Statement implements \IteratorAggregate, Statement
     protected $_paramMap = array();
 
     /**
+     * @var array
```
Ocramius: This docblock needs additional documentation, describing why we are actually keeping a reference to the bound data (this is an oracle bug yadda yadda, horrible PHP7 migration path yadda yadda)

deeky666:
@soyuka it looks really promising but still the "-99" part is quite "magic" and without and explanation and comment I doubt we'll get this thing merged... :(

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
soyuka: I added this number to differ CLOB from BLOB. It's hard coded in the test and has nothing to do with oci8. I want to move this as a constant, where `getBindingType` knows (when using oracle) that this difference should be handled. Still, `getBindingType` is not platform aware if I'm not mistaken?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
soyuka: Sadly it does, http://php.net/manual/en/class.OCI-Lob.php, https://bugs.php.net/bug.php?id=45853

soyuka:
Indeed. I need a way to differ CLOB from BLOB when it comes to data binding. However, PDO constants does not have such differentiation. How/where should I add this constant? Any idea on how `Types` could be platform-aware?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
deeky666: I don't think we can add another constant for CLOB without breaking BC (because we would have to change the binding types in DBAL's Type classes). Can't we maybe detect whether we need `\PDO::PARAM_STR` or `\OCI_TEMP_CLOB` according to the given value? Then we could get rid of the pseudo binding type `-99` and don't need to introduce a new type constant.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -134,7 +139,15 @@ static public function convertPositionalToNamedPlaceholders($statement)
      */
     public function bindValue($param, $value, $type = null)
     {
-        return $this->bindParam($param, $value, $type, null);
+        if ($type === \PDO::PARAM_LOB || -99 === $type) {
+            $descriptor = oci_new_descriptor($this->_dbh, OCI_D_LOB);
+            $descriptor->writeTemporary($value, $type === -99 ? OCI_TEMP_CLOB : OCI_TEMP_BLOB);
```
deeky666: The [PHP docs](http://php.net/manual/en/oci-lob.writetemporary.php) are encouraging us to close the descriptor when finished. Maybe we should do that?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
soyuka: I assumed this would lead to BC break. I thought about that but the given value will always be a `string`, we still need to choose whether it's supposed to be stored as a CLOB or as a BLOB.

We could however send an oci descriptor to the `bindValue` directly. This would allow the following test to work:

```
public function bindParam($column, &$variable, $type = null, $length = null)
    if (is_a($variable, 'Oci-Lob')) {
        return oci_bind_by_name($this->_sth, $column, $variable, -1, $type === PDO::PARAM_LOB ? : OCI_B_BLOB : OCI_B_CLOB);
    }
    //...
}
```

I may give this a shot, I need to make sure that the `PDO::PARAM_LOB` always correspond to a `BLOB` and not a `CLOB` (not sure about this). Also, I need to find where I could create the oci descriptor if not in the statement.

deeky666: @soyuka `PDO::PARAM_LOB` is only used in `BinaryType` and `BlobType` in DBAL so yeah it always corresponds to binary values in the core. I also guess that DBAL 2.x had in mind to map `PDO::PARAM_LOB` only to binary types (given that even `TextType`, `ObjectType` etc. map to `PDO::PARAM_STR`).
So I think we can assume safely assume `PDO::PARAM_LOB` always to be bound to binary.

soyuka: Indeed, I made the same assumption while going through the dbal code, I'll update this PR and let you know! Thanks for the confirmation!

soyuka:
I've updated my implementation, this one should be better. I've added an `OCI8Descriptor` class that allows me to change the binding type from DBAL Types. 

Still, there is something that isn't 100% functional when it comes to inserting CLOB directly from the connection (ie without using DBAL Types). See [this condition](https://github.com/doctrine/dbal/pull/2386/commits/053d54ce2b9b1e02a7cb85327acd1d2c28fe7d34#diff-7b1e94fb7dd408985f7b7973330dc2edR1502), it forces the type to be `BLOB`, but if `\PDO::PARAM_LOB` is meant to bind a `CLOB` it might throw the `Invalid Lob` error.

I just can't find a condition that would lead to the correct LOB type because the only thing we have there is the PDO type.

Tests pass, but still it migth lead to issues in the future, maybe we could mention this in the documentation while we haven't figured a way to differentiate LOB's based on PDO types. 

About the ` OCI-Lob::close`, I think that it has to be done after the query has been executed or it'll throw `ORA-22275: invalid LOB locator specified` as the LOB doesn't exist after calling `close`. If you have any clue/idea where I could do this, please let me know. I don't think it matters much, maybe @cjbj could answer this.

Are you using yoda style for conditions?

cjbj:
@soyuka if you have a standalone OCI8 sample I can take a look.

soyuka:
Ofc, about closing the descriptor:  https://gist.github.com/soyuka/dc502a072cb9f70ebbdb07cf5aaa38aa

cjbj:
Noted.

soyuka:
@deeky666 WDYT about the latest implementation?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -134,7 +141,14 @@ static public function convertPositionalToNamedPlaceholders($statement)
      */
     public function bindValue($param, $value, $type = null)
     {
-        return $this->bindParam($param, $value, $type, null);
+        if ($value instanceof OCI8Descriptor) {
```
deeky666: This is somewhat against the `Statement::bindValue()` contract which says that `$value` is the actual value to be bound. I see that `OCI8Descriptor` "represents" this particular value but this will break the usage of the DBAL OCI8 driver as standalone component. Furthermore in such case the `$type` parameter is not declaring anymore HOW the value is to be bound (because it is ignored). How should the user know how to properly use this API? It does not conform to the given contract anymore.
We need to find a solution that conforms to the contract and encapsulates the logic inside this method. I still wonder why we need to use descriptors for `CLOB` bindings at all. Can't we simply use the value + `\PDO::PARAM_STR`? Is there a failing test case? If that would be the case how does the driver determine whether a descriptor is necessary or not? Does it evaluate the given `$value` or is it bound to the actual table column type? I we cannot use `PARAM_STR` without descriptors at all anymore (which would be really weird), we should stick to always using `CLOB` descriptors in that case. `\PDO::PARAM_LOB` should then as already said always use a `BLOB` descriptor.

soyuka: > This is somewhat against the Statement::bindValue() contract which says that $value is the actual value to be bound. I see that OCI8Descriptor "represents" this particular value but this will break the usage of the DBAL OCI8 driver as standalone component. 

Indeed, that's the solution I used because the `Type::getBindingType` method isn't platform aware. If it was, we could easily give a `PDO::PARAM_CLOB` or `PDO::PARAM_LOB`.  

> Furthermore in such case the $type parameter is not declaring anymore HOW the value is to be bound (because it is ignored). How should the user know how to properly use this API? It does not conform to the given contract anymore.

Right, the type is here `OCI8Descriptor::$_type` (`OCI_BLOB` or `OCI_CLOB`). Still, the default PDO behavior is unchanged. For example, `->bindValue(':param', $blobdescriptor, OCI_BLOB)` is a valid call but  `->bindValue(':param', $blobdescriptor, \PDO::PARAM_LOB)` is not with this implementation (can be fixed).

> I still wonder why we need to use descriptors for CLOB bindings at all. Can't we simply use the value + \PDO::PARAM_STR? Is there a failing test case?

Because, that's how Oracle stores big text fields. We could change the `TextType` to give a `varchar2` with oracle but it would be length-restricted.
For a failing test case, see the `DBAL2386Test` below. It will fail on the #2262 PR because of CLOBS.

>  If that would be the case how does the driver determine whether a descriptor is necessary or not?

If it's a BLOB or a CLOB the driver wants a descriptor.

> If we cannot use PARAM_STR without descriptors at all anymore (which would be really weird)

No issues with this, it's was left unchanged for `StringType` for example.

> \PDO::PARAM_LOB should then as already said always use a BLOB descriptor.

Indeed, I may add a fix for this.

deeky666: I would try to keep it simple then and map `PDO::PARAM_STR` to OCI descriptor + `OCI_B_CLOB` and `PDO::PARAM_LOB` to OCI descriptor + `OCI_BLOB` if that works for all cases. We should able to keep complete BC and do not need to touch other components of DBAL in this case. What do you think?

soyuka: We are not able to bind clobs descriptor to varchar columns. This was the
first thing I tried to keep BC.
Le mer. 29 juin 2016 à 17:19, Steve Müller <notifications@github.com> a
écrit :

deeky666: Meh I am out of ideas. Maybe @cjbj can clarify why it was possible before PHP7 to bind values to `SQLT_CHR` and `CLOB` type columns. It's strange that was possible before and isn't anymore now...

cjbj: Binding should work the same in PHP 5 & 7 as long as the bound zval still exists when `oci_execute()` is later called.  This has always been the case.  But the PHP 5 VM held an extra refcount on zvals which PHP 7 doesn't.  This means that cleanup can occur in PHP 7 before `oci_execute()` is called.  The OCI8 driver can't fake the refcount since it doesn't know what the app wants to do with the var.  Incrementing the refcount can fix some IN bind cases but will break OUT binds.  The bottom line is that the app has to manage the refcount and end-of-scope cleanup.  If you are seeing any bind issues where bind & execute are done in the same scope, log a bug.

deeky666: @cjbj yes we got that refcount thing already, what we are talking about here is a different thing. I'll try to give some more details.
Until now we were always binding a value that corresponds to a `CLOB` type column in the database to OCI's `SQLT_CHR` binding type (without using a descriptor) and it perfectly worked in PHP5.
Now in PHP7 using `SQLT_CHR` for binding `CLOB` type column values does not work anymore and requires a descriptor + binding type `OCI_B_CLOB`.
I thought you might have a bit more background information why this is the case and if this is intended.

soyuka: Here is a full test case using oci8 only

https://gist.github.com/soyuka/16273fde9b00c3c47147b537fc7f27fa

See results without descriptor, binding works but column isn't filled with the value:

```
/home/abluchet/forks/dbal/oci8test.php:50:
array(4) {
  [0] =>
  array(3) {
    'ID' =>
    string(1) "2"
    'ADDRESS' =>
    string(0) ""
    'OTHER' =>
    string(47) "Is using a descriptor with address value "null""
  }
  [1] =>
  array(3) {
    'ID' =>
    string(1) "3"
    'ADDRESS' =>
    NULL
    'OTHER' =>
    string(39) "No descriptor with address value "null""
  }
  [2] =>
  array(3) {
    'ID' =>
    string(1) "0"
    'ADDRESS' =>
    string(5) "tests"
    'OTHER' =>
    string(48) "Is using a descriptor with address value "tests""
  }
  [3] =>
  array(3) {
    'ID' =>
    string(1) "1"
    'ADDRESS' =>
    NULL
    'OTHER' =>
    string(40) "No descriptor with address value "tests""
  }
}
```

soyuka: My bad, it works:

https://gist.github.com/soyuka/16273fde9b00c3c47147b537fc7f27fa

soyuka: So, this works, but Doctrine is still broken with the initial code.

Please try the `master` branch with the test case attached here, you'll get:

```
ORA-24816: Expanded non LONG bind data supplied after actual LONG or LOB column
```

Which I explained in previous comments. I'm all ears if you have a solution.

deeky666: First of all I'm glad that we could potentially still use `SQLT_CHR` :)
Second I think that the `ORA-24816` is something different that is not directly a bug of Doctrine. It looks like it is not possible to bind a non LONG column after having bound a LOB column with discriptor (sort of). I am still not completely sure what is happening here and why but I would say that is something different and should be dealt with in another PR.
Let's first focus on making DBAL OCI8 compatible with PHP7. So the only problem that is left (which I see) is the refcount thingy which we have to fix. I do not see where we have to change anything about binding types, am I right? If that is the case, could you revert everything that is not specific to keeping references to the bound values and see if our test suite succeeds? Thanks.

deeky666: Btw as a reference to the `ORA-24816` issue: http://stackoverflow.com/questions/23341391/ora-24816-expanded-non-long-bind-data-supplied-after-actual-long-or-lob-column

soyuka: Updated with the workaround for null values

deeky666: And here the solution for `ORA-24816`: http://psoug.org/oraerror/ORA-24816.htm
YOLO ONO :D

soyuka: HAha yes but in fact the value is NULL and has no length, so not helpfull \o/.

How can we prevent CLOB null values?

soyuka: Btw I've read the whole internet about this stupid undocumented error :laughing:

deeky666: @soyuka yeah but your gist is not reflecting what Doctrine will be doing in the end because you are binding a descriptor to the `CLOB` column whereas Doctrine will not be using a descriptor.

soyuka: Nope, I'm also binding without descriptors (on $i % 2 !== 0) ;)

I'm preparing a new PR, waiting for the tests to complete (~2 min).

deeky666: @soyuka yes but the problem is that you are mixing bindings of descriptor + CLOB and non descriptor + `SQLT_CHR` which does not happen in Doctrine because Doctrine doesn not use descriptors for `CLOB`

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -145,11 +159,12 @@ public function bindParam($column, &$variable, $type = null, $length = null)
         $column = isset($this->_paramMap[$column]) ? $this->_paramMap[$column] : $column;
 
         if ($type == \PDO::PARAM_LOB) {
-            $lob = oci_new_descriptor($this->_dbh, OCI_D_LOB);
-            $lob->writeTemporary($variable, OCI_TEMP_BLOB);
+            if (!is_a($variable, 'Oci-Lob')) {
+                throw new OCI8Exception('Can not bind this value to the BLOB column, please use a oci8 descriptor');
+            }
```
soyuka: I don't like this but it's to avoid misuse of the `bindParam` function (when not using `bindValue`). I can't instantiate the descriptor here because `writeTemporary` does not accept the reference (`&$variable) (hence the `ORA-22275: invalid LOB locator specified` error).

lib/Doctrine/DBAL/Types/JsonType.php
```diff
@@ -43,6 +43,10 @@ public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $pla
     public function convertToDatabaseValue($value, AbstractPlatform $platform)
     {
         if (null === $value) {
+            if ($platform->getName() == 'oracle') {
+                return '';
+            }
+
```
soyuka: If you have another idea for this, I'm all ears. It's the smallest change I can do to avoid binding `null` to a CLOB.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
soyuka: See the new commit, if you don't change types for CLOB stored values when a `null` is issued, you'll get that error. We're not using descriptors for CLOB in doctrine. My gist illustrate the fact that:

- by using CLOB with a correct descriptor, you can bind `null` values and everything works
- without descriptor, you can't bind `null` and so you have to use an empty string

Maybe that's not how it's supposed to work, but it is.

soyuka:
Anyway, to follow up on [our discussion](https://github.com/doctrine/dbal/pull/2386#discussion_r69099012), I think that it would be a lot cleaner if we use descriptors for both `BLOB` and `CLOB` bindings. As this breaks BC, it won't be part of this PR, but I think it should be considered for doctrine `3.x`. It'll be my pleasure to help with this in the future, so don't hesitate to ping me.

Btw we should also keep track of open descriptors to properly close them after `oci_execute()`.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
deeky666: Is there a failing test case in the Doctrine suite? Because I don't get any errors with my implementation...

lib/Doctrine/DBAL/Types/JsonType.php
deeky666: @soyuka the issue here that triggers `ORA-24816` is not that we cannot bind `null` to a `CLOB` type column (which works) but in fact this appears if do an insert like this:

```
Doctrine\DBAL\Exception\DriverException: An exception occurred while executing 'INSERT INTO DBAL2386 VALUES (:id, :title, :address, :geo, :name)' with params [1, null, null, null, null]:

ORA-24816: Expanded non LONG bind data supplied after actual LONG or LOB column
```

What works would be the following:

```
INSERT INTO DBAL2386 VALUES (:id, :title, :address, :geo, :name)' with params [1, <anything_including_null>, <anything_including_null>, <anything_including_null>, "test"]
```

So what PHP7 oci8 does not "like" (or I would even say it's a bug because it worked in PHP5) is having a `null` value bound to a column that comes after a LONG/LOB column (the order of the insert list is important here, not the order of binding).
I am not sure what we could possibly do about this.

soyuka: Take my gist https://gist.github.com/soyuka/336e211077de41da6b28928f9050919d#file-ocitestnulls-php-L13

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
soyuka: +1 it's a nice implementation btw I wasn't aware that we could reference this way around to `boundValues`.

lib/Doctrine/DBAL/Types/JsonType.php
deeky666: @cjbj done
https://bugs.php.net/bug.php?id=72524

Hope that is clear and simple enough.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
deeky666: It is done exactly the same in the mysqli driver and it also deals with `bindParam()` references which can potentially get lost, too. Thanks for taking over. I had in mind to do an alternativ approach PR but we can leave it here too :)

soyuka: I'm can close this PR in favor of your implementation. I'm opening a new one regarding the phpbug with my failing doctrine test case. WDYT?

deeky666: Sounds reasonable especially as I am still of the opinion that the bug you are revealing is seperate issue that might even be a PHP bug. So we could at least solve the binding reference issue for now.

soyuka: +1 exactly, I let you do your own PR about that then ;)

soyuka:
Closing this in favor of https://github.com/deeky666/dbal/commit/a35cc8aa86dac44cbe7e70a964b936a79b483d5d

lib/Doctrine/DBAL/Types/JsonType.php
cjbj: @deeky666 thanks

Ocramius:
Handled in https://github.com/doctrine/dbal/pull/2434

PR #2389, opened 2016-05-12
Prevent adding empty CompositeExpression objects

### LINKED ISSUE #2388: CompositeExpression can't handle empty CompositeExpression objects
Normally the \Doctrine\DBAL\Query\Expression\CompositeExpression::add should prevent to add empty parts. But if a part is an object of type CompositeExpression the empty check is always false, because the object itself is never empty. Currently the second count() test for an object of type CompositeExpression is never reached.

IchHabRecht:
Fixes #2388

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Query/Expression/CompositeExpressionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Query/Expression/CompositeExpressionTest.php
@@ -77,4 +77,12 @@ class CompositeExpressionTest extends \Doctrine\Tests\DbalTestCase
             ),
         );
     }
+
+    public function testAddPreventsAddingEmptyCompositeExpressionObjects()
+    {
+        $emptyExpr = new CompositeExpression(CompositeExpression::TYPE_AND, []);
+        $expr = new CompositeExpression(CompositeExpression::TYPE_AND, [$emptyExpr]);
+
+        $this->assertEquals(0, count($expr));
+    }
 }
--- a/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
+++ b/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
@@ -91,7 +91,8 @@ class CompositeExpression implements \Countable
      */
     public function add($part)
     {
-        if ( ! empty($part) || ($part instanceof self && $part->count() > 0)) {
+        $isEmpty = $part instanceof self ? $part->count() === 0 : empty($part);
+        if (!$isEmpty) {
             $this->parts[] = $part;
         }
 


tests/Doctrine/Tests/DBAL/Query/Expression/CompositeExpressionTest.php
```diff
@@ -77,4 +77,12 @@ public function provideDataForConvertToString()
             ),
         );
     }
+
+    public function testAddPreventsAddingEmptyCompositeExpressionObjects()
+    {
+        $emptyExpr = new CompositeExpression(CompositeExpression::TYPE_AND, []);
+        $expr = new CompositeExpression(CompositeExpression::TYPE_AND, [$emptyExpr]);
+
+        $this->assertEquals(0, count($expr));
```
deeky666: I don't quite get it. The object is mutable, so you can add any more parts as you want as long as you do not cast it as `string`. Then you might want to check for number of parts >= 2 and eventually throw an exception. But your test is not revealing any issue. Also I am not sure if this is not considered part of the internal API /cc @Ocramius

IchHabRecht: Sorry, but I don't get your comment at all. I can try to explain the issue a bit more, but maybe there is any way to get in contact with you in any chat?

The problem which I ran into:
- reading the (old) code, it should be possible to pass an object of type CompositeExpression (object2) as a new part in another CompositeExpression object (object1)
- object2 should only be added it it contains parts on its own
- so creating object2 with empty parts and passing it to object1 shouldn't change parts of object1

Currently this is broken and an empty object2 is added to the parts of object1. IMHO this is not intended. I created the patch and a unit test which fails before and is fixed with the patch.

deeky666: Hehe sorry if I was not clear enough. What I want to point out is that if passing an empty instance of `CompositeExpression` it should not be rejected at all IMO because it is passed as an object by reference and you can still add parts to that later on. Example:

```php
$expression1 = new CompositeExpression(CompositeExpression::TYPE_AND, []);
$expression2 = new CompositeExpression(CompositeExpression::TYPE_AND, [$expression1]);

$expression1->add(array('u.group_id = 1')); // now $expression1 is not empty anymore
```

This is what I was referring to as "the API is mutable".

IchHabRecht: Ir that would be a way how the api can be used, why was there a check 

> ($part instanceof self && $part->count() > 0)

in the code? I think it shouldn't be possible to add constraints after adding the object to the CompositeExpression

deeky666: I don't have a strong opinion on this TBH and I don't know what the original intention was /cc @Ocramius @guilhermeblanco

Ocramius: It would be much easier to prevent an empty operands list to make it into a `CompositeExpression`, but I suppose that we can't change the API now.

helhum: @Ocramius just to be sure: what you're saying is that the fix here is correct, but as it affects public API, it can't be fixed?

Ocramius: No, it was just wishful thinking: if we didn't allow mutation of expressions, this problem would not exist in first place.

Anyway, the fix doesn't seem correct to me: I think the correct way of dealing with the problem is to remove unnecessary composite expression building within `__toString`, which is called at the very end of the entire procedure. This way, even if state changed in-between, we'd still get consistent behavior.

morozov:
Fixed in https://github.com/doctrine/dbal/pull/3864.

Issue #2353, opened 2016-04-08
PostgreSqlPlatform tstzrange unknown database type


EspadaV8:
Trying to alter a postgresql table that has a `tstzrange` which is causing an exception to be thrown

```
[Doctrine\DBAL\DBALException]
Unknown database type tstzrange requested, Doctrine\DBAL\Platforms\PostgreSqlPlatform may not support it.
```

Ranges were added in PostgreSQL 9.2 (http://www.postgresql.org/docs/9.2/static/release-9-2.html) and I couldn't see anything listed in `PostgreSQL92Platform` for it either.

morozov:
This type doesn't look portable across supported databases.

PR #2398, opened 2016-05-18
Added PostgreSQL 9.2 range types

Stardog:
Fixes doctrine/dbal#2353

See http://www.postgresql.org/docs/9.2/static/rangetypes.html

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/PostgreSQL92Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSQL92Platform.php
@@ -71,6 +71,12 @@ class PostgreSQL92Platform extends PostgreSQL91Platform
     {
         parent::initializeDoctrineTypeMappings();
         $this->doctrineTypeMapping['json'] = 'json_array';
+        $this->doctrineTypeMapping['int4range'] = 'integer';
+        $this->doctrineTypeMapping['int8range'] = 'integer';
+        $this->doctrineTypeMapping['numrange'] = 'decimal';
+        $this->doctrineTypeMapping['tsrange'] = 'datetime';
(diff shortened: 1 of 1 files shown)


deeky666:
@Stardog well, those native types are representing ranges of types, not atomic values (if I get the docs correctly), so how can those map to atomic values (literals)?

Stardog:
Honestly, I was hoping it was a quick fix that would allow you to merge and move. I agree, it doesn't really fit the method's purpose. Can you suggest an alternative coding direction to approach the error that ranges cause?
Please close this if not the direction you want to go.
Thanks!

deeky666:
@Stardog this seems highly PostgreSQL specific so I am not sure we will/can add this to DBAL core. What do you actually get from the driver when you query those columns? I mean the type system is also responsible for converting PHP types from and to SQL types. I wonder what the driver returns for those SQL types. That said, I wonder how a range (having two values, a lower and upper bound) can be converted to an atomic PHP type like `int` as your PR suggests...

EspadaV8:
The ranges are returned as string values. We are currently making use of them extensively in an application (currently using Laravel's Eloquent) and we manually parse the string into an array of values

```php
// original from DB
$tstzrange = "[2016-01-01 00:00:00+00,2016-02-01 00:00:00+00)";

// after some string manipulation
$parsed = [
    'lower' => Carbon('2016-01-01 00:00:00+00'),
    'upper' => Carbon('2016-02-01 00:00:00+00'),
];
```

This isn't really enough information though for a standard library since the upper and lower values could either be open or closed which may be important information for the consumer of the data (we don't care because all of our `tstzrange`s are the same).

We then do some formatting from the parsed array back into the correctly formatted string before saving it into the column.

morozov:
Closing as stale.

Issue #2400, opened 2016-05-20
PostgreSQL's SchemaDiff doesn't change collumn collation


j-d:
Hello, 

The method `getAlterTableSQL` on `PostgreSQLPlatform` does not perform changes on the column collation. This is because there are two bugs: 
- First, `Comparator::diffColumn()` doesn't see the change due to a bug on the login when comparing the platform options. I have prepared a Pull request for this and I am going to send it in a minute. 
- Second, whilst the `addedColumns` section of the `getAlterTableSQL` method on the `PostgresSQLPlatform` class calls `getColumnDeclarationSQL` which in turn returns the right collation, the `changedColumns` section of the same method doesn't. I am not sure how best to fix this, so I would appreciate if someone could give me an indication or fix it. 

Thanks a lot.

mishal:
I tried to fix the issue, see the code.. not sure if it is the best approach.
Postgresql collation on the column can be changed only with type, see:

https://www.postgresql.org/docs/current/static/sql-altertable.html

When reading the docs again, the "default" collation setting should be removed..

PR #2401, opened 2016-05-20
Detect changes on the column collation (Issue #2400)

j-d:
I cannot run the unit tests on my current environment so I would appreciate if someone could make sure that it is ok. 

I have replicated the logic on the Comparator that was already there for the platform options.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Comparator.php
+++ b/lib/Doctrine/DBAL/Schema/Comparator.php
@@ -488,8 +488,10 @@ class Comparator
         $platformOptions1 = $column1->getPlatformOptions();
         $platformOptions2 = $column2->getPlatformOptions();
 
-        foreach (array_keys(array_intersect_key($platformOptions1, $platformOptions2)) as $key) {
-            if ($properties1[$key] !== $properties2[$key]) {
+        foreach (array_merge(array_keys($platformOptions1), array_keys($platformOptions2)) as $key) {
+            if ( ! array_key_exists($key, $properties1) || ! array_key_exists($key, $properties2)) {
+                $changedProperties[] = $key;
+            } elseif ($properties1[$key] !== $properties2[$key]) {
                 $changedProperties[] = $key;
             }
         }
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Schema/Comparator.php
```diff
@@ -488,8 +488,10 @@ public function diffColumn(Column $column1, Column $column2)
         $platformOptions1 = $column1->getPlatformOptions();
         $platformOptions2 = $column2->getPlatformOptions();
 
-        foreach (array_keys(array_intersect_key($platformOptions1, $platformOptions2)) as $key) {
-            if ($properties1[$key] !== $properties2[$key]) {
+        foreach (array_merge(array_keys($platformOptions1), array_keys($platformOptions2)) as $key) {
```
deeky666: Hmmm this does not seem to be working (see Travis). IMO the logic should be as follows: Do an `array_diffkey()` and mark all key differences as changed, do an `array_intersect_key()` and mark all keys as changed where the values do not match. However I am not sure how this will behave with "default values" like collation which DBAL sets implicitly in the platform (MySQL's default collation for tables for example) if not explicitly set. This could lead to undesired diffs, too. So you might have to play around a bit and get a working test environment going on your machine.

j-d: Thanks for your feedback @deeky666 . I will try to get it working in my environment. Do you think there is also a problem here https://github.com/j-d/dbal/blob/fba875af209772b5ce0446d4adaa59da8c4fe6c2/lib/Doctrine/DBAL/Schema/Comparator.php#L480-L486 ? I basically copy/pasted from there and adjusted it. 
Thanks

deeky666: Looking at this again, it seems that it should be doing already what I was suggesting, even though a lit bit more complicated. So you might just add some more tests to reveal any issues.

morozov:

Issue #2374, opened 2016-04-22
SQL Server platform cannot apply LIMIT to a UNION query with ORDER BY


morozov:
The cases below were reproduced on Doctrine DBAL 2.5.4 and SQL Server 11.0.2100.60.

1. UNION with sub-queries surrounded by parentheses

    ```sql
    (SELECT id FROM accounts) UNION ALL (SELECT id FROM contacts) ORDER BY id
    ```
    Then `TOP` keyword is not added to the query:
    ```sql
    WITH dctrn_cte AS ((SELECT id
                        FROM accounts)
                       UNION ALL (SELECT id
                                  FROM contacts)
                       ORDER BY id) SELECT *
                                    FROM (SELECT
                                            *,
                                            ROW_NUMBER()
                                            OVER (
                                              ORDER BY (SELECT 0)) AS doctrine_rownum
                                          FROM dctrn_cte) AS doctrine_tbl
                                    WHERE doctrine_rownum BETWEEN 1 AND 20
                                    ORDER BY doctrine_rownum ASC
    ```
    Which results into SQL error:
    ```
    [156] Incorrect syntax near the keyword 'ORDER'.
    ```

2. UNION without parentheses

    ```sql
    SELECT id FROM accounts UNION ALL SELECT id FROM contacts ORDER BY id
    ```
Then `TOP` keyword is added to the first sub-query instead of the resulting set.
    ```sql
    WITH dctrn_cte AS (SELECT TOP 20 id
                       FROM accounts
                       UNION ALL SELECT id
                                 FROM contacts
                       ORDER BY id) SELECT *
                                    FROM (SELECT
                                            *,
                                            ROW_NUMBER()
                                            OVER (
                                              ORDER BY (SELECT 0)) AS doctrine_rownum
                                          FROM dctrn_cte) AS doctrine_tbl
                                    WHERE doctrine_rownum BETWEEN 1 AND 20
                                    ORDER BY doctrine_rownum ASC
    ```
    The SQL server still seems to not like the `ORDER BY` in the CTE:
    ```
    [1033] The ORDER BY clause is invalid in views, inline functions, derived tables, subqueries, and common table expressions, unless TOP, OFFSET or FOR XML is also specified.
    ```

3. UNION with line breaks.

    ```sql
SELECT id FROM accounts
UNION ALL
SELECT id FROM contacts
ORDER BY id
    ```
Unlike #2372 where only an invalid query is produced, in this case PHP goes into infinite loop:
    ```
[...]

Issue #2313, opened 2016-02-05
SQLServerPlatform::scrubInnerOrderBy() enters infinite loop


BenjaminNolan:
Not sure if this is an issue specific to 5.6, but that's all I've had time to test it in. (Sorry!) Wouldn't have noticed it were it not for the `String offset cast occurred` notice appearing hundreds of thousands of times in my PHP logs. :)

`Doctrine\DBAL\Platforms\SQLServerPlatform::scrubInnerOrderBy()` has a bug in it where it enters an infinite loop after `stripos()` returns `false` if the string `ORDER BY` is not present after the last `)` found at `$currentPosition`.

When it runs on the following query, it enters an infinite loop and never exits:

> `SELECT DISTINCT TOP 30 id_1, MIN(sclr_6) AS dctrn_minrownum FROM (SELECT q0_.available_at AS available_at_0, q0_.id AS id_1, q0_.is_active AS is_active_2, q0_.is_deleted AS is_deleted_3, q0_.created_at AS created_at_4, q0_.last_updated_at AS last_updated_at_5, ROW_NUMBER() OVER(ORDER BY q0_.created_at ASC) AS sclr_6 FROM ThisIsASquirrel q0_ WHERE (q0_.is_deleted = 0 AND q0_.is_active = 1) AND (q0_.is_deleted = 0)) dctrn_result GROUP BY id_1 ORDER BY dctrn_minrownum ASC`

Adding a check for $orderByPos === false and then breaking out allows the code to continue successfully.

kokspflanze:
hi,

i had the bug too, i fix it in https://github.com/doctrine/dbal/pull/2341

the problem is "OVER(ORDER BY q0_.created_at ASC)" in your query.

https://github.com/doctrine/dbal/blob/v2.5.4/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php#L1222 count the number of "ORDER BY" and https://github.com/doctrine/dbal/blob/v2.5.4/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php#L1227 search for " ORDER BY"

morozov:
Closing as irrelevant as of https://github.com/doctrine/dbal/pull/3756.

PR #2408, opened 2016-06-07
Hotfix 2313 infinite loop

BenjaminNolan:
This fixes the `String offset cast occurred` due to poor parsing of the SQL query being checked by checking if the string being searched for was actually found and then breaking if it is not.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -1226,8 +1226,12 @@ class SQLServerPlatform extends AbstractPlatform
         $offset = 0;
 
         while ($count-- > 0) {
-            $qLen = strlen($query);
             $orderByPos = stripos($query, " ORDER BY", $offset);
+            if ($orderByPos === false) {
+                break;
+            }
+
+            $qLen = strlen($query);
             $parenCount = 0;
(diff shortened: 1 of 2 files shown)


Ocramius:
Test case required

TwoWholeWorms:
Thought I'd submitted this PR back in February, but clearly I didn't, and the error is still there.

TwoWholeWorms:
Tests added (one for a query with a clause that needs to be scrubbed, and one that just gets reordered without scrubbing anything). Also, this should fix issue #2374 as well as it breaks out of the infinite loop which their query also triggered.

deeky666:
@TwoWholeWorms isn't this a duplicate of https://github.com/doctrine/dbal/pull/2341 ?

TwoWholeWorms:
@deeky666 I'd say that's a duplicate of #2313. :p

kathastaden:
Hi, I just wanted to know what the status of this PR is? There hasn't been any progress in the last six months but the error still exists.

billschaller:
LGTM

deeky666:
Backporte to `2.5` via https://github.com/doctrine/dbal/commit/9eb014a571105d2e80a8f945362bba6f676adb46 and https://github.com/doctrine/dbal/commit/6314d55f60f85f0295c90688397b75c13faf9f0e

PR #2410, opened 2016-06-10
Add method to get affected column name in constraint exception

juniwalk:
I sometimes need to get the column name causing this exception and I don't want to parse exception message in the catch clause.

### DIFF
--- a/lib/Doctrine/DBAL/Exception/ConstraintViolationException.php
+++ b/lib/Doctrine/DBAL/Exception/ConstraintViolationException.php
@@ -28,4 +28,17 @@ namespace Doctrine\DBAL\Exception;
  */
 class ConstraintViolationException extends ServerException
 {
+	/**
+	 * Get the name of affected column.
+	 *
+	 * @return string|NULL
+	 */
+	public function getAffectedColumnName()
+	{
+		if (preg_match_all('/^DETAIL\:\s+Key\s+\((\w+)\)/im', $this->getMessage(), $matches)) {
+			return $matches[1][0];
+		}
+
+		return NULL;
+	}
 }


lib/Doctrine/DBAL/Exception/ConstraintViolationException.php
```diff
@@ -28,4 +28,17 @@
...
 class ConstraintViolationException extends ServerException
 {
+	/**
+	 * Get the name of affected column.
+	 *
+	 * @return string|NULL
+	 */
+	public function getAffectedColumnName()
+	{
+		if (preg_match_all('/^DETAIL\:\s+Key\s+\((\w+)\)/im', $this->getMessage(), $matches)) {
```
deeky666: This will not work for all vendors and drivers as error messages are different for each.

juniwalk: Ah, dammit! I haven't thought about that!

juniwalk: Hey @deeky666,
so I have collected error messages from `PostgreSQL`, `MySQL`, `SQLite` and `SQL Server`. They are all different and I cannot reliably get column name, they tend to use index name. How shall I proceed?

Is this PR bad idea?

deeky666: @juniwalk relying on the exception message for such information is a bad idea (also consider different locale settings for exception messages etc.), thus I don't think we should add something like this to the core (especially as not all vendors provide that information).
This should be dealt with in userland with some kind of error handler class or similar if you really need it.

juniwalk: @deeky666 You are right, too much holes in my idea. Thanks for consideration anyway! :)

Issue #2368, opened 2016-04-18
MySQL: Create indexes with columns length


DeepDiver1975:
With MySQL (and other DBMS as well :question:) is is possible to define the column length to be used within an index

``` sql
CREATE INDEX part_of_name ON customer (name(10));
```

Is there a way to respect this with the schema classes when generating the sql statements?

Jean85:
Documentation refers an `option` array for the `@Index` annotation, but I don't know if `lenght` is supported...

http://doctrine-orm.readthedocs.org/projects/doctrine-orm/en/latest/reference/annotations-reference.html#annref-index

morozov:
Fixed by #2412.

PR #2412, opened 2016-06-13
Add mysql specific indexes with lengths

### LINKED ISSUE #1597: DBAL-404: Support of index length for text fields
Jira issue originally created by user maxlun86:

For table with text field description with index on it generated query is:

CREATE TABLE t (id INT UNSIGNED AUTO_INCREMENT NOT NULL, description LONGTEXT DEFAULT NULL, number INT DEFAULT NULL, INDEX index2 (description), INDEX index3 (description, number), PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB

This query lead to error, because for index on text field absented key length.

bburnichon:
First step to fix #1597 & #2368

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -1408,7 +1408,9 @@ abstract class AbstractPlatform
 
         if ($table instanceof Table) {
             $table = $table->getQuotedName($this);
-        } elseif (! is_string($table)) {
+        }
+
+        if (! is_string($table)) {
             throw new InvalidArgumentException('getDropTableSQL() expects $table parameter to be string or \Doctrine\DBAL\Schema\Table.');
         }
 
@@ -1784,7 +1786,7 @@ abstract class AbstractPlatform
             $table = $table->getQuotedName($this);
         }
         $name    = $index->getQuotedName($this);
-        $columns = $index->getQuotedColumns($this);
+        $columns = $index->getColumns();
 
         if (count($columns) === 0) {
             throw new InvalidArgumentException("Incomplete definition. 'columns' required.");
@@ -1795,7 +1797,7 @@ abstract class AbstractPlatform
         }
 
         $query  = 'CREATE ' . $this->getCreateIndexSQLFlags($index) . 'INDEX ' . $name . ' ON ' . $table;
-        $query .= ' (' . $this->getIndexFieldDeclarationListSQL($columns) . ')' . $this->getPartialIndexSQL($index);
+        $query .= ' (' . $this->getIndexFieldDeclarationListSQL($index) . ')' . $this->getPartialIndexSQL($index);
 
         return $query;
     }
@@ -1833,7 +1835,7 @@ abstract class AbstractPlatform
      */
     public function getCreatePrimaryKeySQL(Index $index, $table)
     {
-        return 'ALTER TABLE ' . $table . ' ADD PRIMARY KEY (' . $this->getIndexFieldDeclarationListSQL($index->getQuotedColumns($this)) . ')';
+        return 'ALTER TABLE ' . $table . ' ADD PRIMARY KEY (' . $this->getIndexFieldDeclarationListSQL($index) . ')';
     }
 
     /**
@@ -2337,7 +2339,7 @@ abstract class AbstractPlatform
      */
     public function getUniqueConstraintDeclarationSQL($name, Index $index)
     {
-        $columns = $index->getQuotedColumns($this);
+        $columns = $index->getColumns();
         $name    = new Identifier($name);
 
         if (count($columns) === 0) {
@@ -2345,7 +2347,7 @@ abstract class AbstractPlatform
         }
 
         return 'CONSTRAINT ' . $name->getQuotedName($this) . ' UNIQUE ('
-            . $this->getIndexFieldDeclarationListSQL($columns)
+            . $this->getIndexFieldDeclarationListSQL($index)
             . ')' . $this->getPartialIndexSQL($index);
     }
 
@@ -2362,7 +2364,7 @@ abstract class AbstractPlatform
      */
     public function getIndexDeclarationSQL($name, Index $index)
     {
-        $columns = $index->getQuotedColumns($this);
+        $columns = $index->getColumns();
         $name    = new Identifier($name);
 
         if (count($columns) === 0) {
@@ -2370,7 +2372,7 @@ abstract class AbstractPlatform
         }
 
         return $this->getCreateIndexSQLFlags($index) . 'INDEX ' . $name->getQuotedName($this) . ' ('
-            . $this->getIndexFieldDeclarationListSQL($columns)
+            . $this->getIndexFieldDeclarationListSQL($index)
             . ')' . $this->getPartialIndexSQL($index);
     }
 
@@ -2392,17 +2394,23 @@ abstract class AbstractPlatform
      * Obtains DBMS specific SQL code portion needed to set an index
      * declaration to be used in statements like CREATE TABLE.
      *
-     * @param mixed[][] $fields
-     *
-     * @return string
+     * @param mixed[]|Index $columnsOrIndex array declaration is deprecated, prefer passing Index to this method
      */
-    public function getIndexFieldDeclarationListSQL(array $fields)
+    public function getIndexFieldDeclarationListSQL($columnsOrIndex) : string
     {
+        if ($columnsOrIndex instanceof Index) {
+            return implode(', ', $columnsOrIndex->getQuotedColumns($this));
+        }
+
+        if (! is_array($columnsOrIndex)) {
+            throw new InvalidArgumentException('Fields argument should be an Index or array.');
+        }
+
         $ret = [];
 
-        foreach ($fields as $field => $definition) {
+        foreach ($columnsOrIndex as $column => $definition) {
             if (is_array($definition)) {
-                $ret[] = $field;
+                $ret[] = $column;
             } else {
                 $ret[] = $definition;
             }
@@ -3069,6 +3077,14 @@ abstract class AbstractPlatform
      * @return bool
      */
     public function supportsPartialIndexes()
+    {
+        return false;
+    }
+
+    /**
+     * Whether the platform supports indexes with column length definitions.
+     */
+    public function supportsColumnLengthIndexes() : bool
     {
         return false;
     }
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -645,7 +645,7 @@ class MySqlPlatform extends AbstractPlatform
 
                     $query  = 'ALTER TABLE ' . $table . ' DROP INDEX ' . $remIndex->getName() . ', ';
                     $query .= 'ADD ' . $indexClause;
-                    $query .= ' (' . $this->getIndexFieldDeclarationListSQL($addIndex->getQuotedColumns($this)) . ')';
+                    $query .= ' (' . $this->getIndexFieldDeclarationListSQL($addIndex) . ')';
 
                     $sql[] = $query;
 
@@ -1131,5 +1131,13 @@ class MySqlPlatform extends AbstractPlatform
     public function getDefaultTransactionIsolationLevel()
     {
         return TransactionIsolationLevel::REPEATABLE_READ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supportsColumnLengthIndexes() : bool
+    {
+        return true;
     }
 }
--- a/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
@@ -843,17 +843,26 @@ abstract class AbstractSchemaManager
             $keyName = strtolower($keyName);
 
             if (! isset($result[$keyName])) {
+                $options = [
+                    'lengths' => [],
+                ];
+
+                if (isset($tableIndex['where'])) {
+                    $options['where'] = $tableIndex['where'];
+                }
+
                 $result[$keyName] = [
                     'name' => $indexName,
-                    'columns' => [$tableIndex['column_name']],
+                    'columns' => [],
                     'unique' => $tableIndex['non_unique'] ? false : true,
                     'primary' => $tableIndex['primary'],
                     'flags' => $tableIndex['flags'] ?? [],
-                    'options' => isset($tableIndex['where']) ? ['where' => $tableIndex['where']] : [],
+                    'options' => $options,
                 ];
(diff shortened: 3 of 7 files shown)


mnapoli:
I think `length` (singular)  might be a better name than `lengths` (plural). Funny I ran into that issue just today.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2386,12 +2386,18 @@ public function getCustomTypeDeclarationSQL(array $columnDef)
      * Obtains DBMS specific SQL code portion needed to set an index
      * declaration to be used in statements like CREATE TABLE.
      *
-     * @param array $fields
+     * @param array|Index $fields
```
deeky666: This is weird and we should not do this. This API accepts an array of field definitions and not an index. I see why you are doing this but this is more of a workaround/hack rather than a good solution. Also having `lengths` option on the index as an array of lengths for all of the index columns isn't better either. What we are missing is an `IndexColumn` class that has options itself dedicated to the index column but I am not sure we should/can add this in 2.x.

bburnichon: I know the API currently accepts an array of field names. My first intention was to change this to only accept an `Index` parameter. I agree that an `IndexColumn` could be a solution but it seems heavier to implement this class.

I did not wanted to have a simple array of lengths in the API but I currently do not see another better way to avoid this while keeping BC.

bburnichon:
@mnapoli I made `lengths` plural to be consistent with `columns`, `flags` and so on.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2386,12 +2386,18 @@ public function getCustomTypeDeclarationSQL(array $columnDef)
...
      *
      * @return string
      */
-    public function getIndexFieldDeclarationListSQL(array $fields)
+    public function getIndexFieldDeclarationListSQL($fields)
     {
+        if ($fields instanceof Index) {
+            return implode(', ', $fields->getQuotedColumns($this));
+        } elseif (!is_array($fields)) {
```
Ocramius: no need for `else`

bburnichon: Don't you want to check `$fields` is of proper type before continuing? If I remove the else part, developers could call method with improper data type. Or do you just prefer to have a separate `if` statement instead of `elseif`?

Ocramius: You'd just make an `if` instead of an `elseif` ;-)

aimeos:
Would be great to get this feature into DBAL because it would simplify managing indexes on text columns very much!

tmaroschik:
What is necessary to get this pull request merged? Fixing the conflicts only or is the general approach in question?

morozov (changes requested):
@tmaroschik please provide a functional test which would fail without the fix. Additionally, as far as I understand the design decision made in `AbstractPlatform::getIndexFieldDeclarationListSQL()`, the usage of `array $fields` should be deprecated in favor of `Index $index`.

bburnichon:
@morozov Is functional test provided sufficient?

morozov:
@bburnichon yes, the test looks good. As far as I understand, only MySQL (and probably MariaDB) support.

Would it make sense to move the test up to `SchemaManagerFunctionalTestCase` and explicitly disable in all classes except `MySqlSchemaManagerTest`? It looks so to me because it covers the new `'length'` flag of the index definition which belongs to the DBAL itself, not just to the MySQL schema manager.

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -868,16 +868,25 @@ protected function _getPortableTableIndexesList($tableIndexRows, $tableName=null
...
                     'columns' => [$tableIndex['column_name']],
                     'unique' => $tableIndex['non_unique'] ? false : true,
                     'primary' => $tableIndex['primary'],
                     'flags' => $tableIndex['flags'] ?? [],
-                    'options' => isset($tableIndex['where']) ? ['where' => $tableIndex['where']] : [],
+                    'options' => $options,
                 ];
             } else {
                 $result[$keyName]['columns'][] = $tableIndex['column_name'];
+                $result[$keyName]['options']['lengths'][] = $tableIndex['sub_part'] ?? null;
```
morozov: The same logic seems duplicated in `if` and `else ` branches. Can it be moved outside? Probably, once it's moved out of the base class, it may become irrelevant.

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
```diff
@@ -478,4 +478,22 @@ public function testEnsureDefaultsAreUnescapedFromSchemaIntrospection() : void
         $onlineTable = $this->_sm->listTableDetails("test_column_defaults_with_create");
         self::assertSame($default, $onlineTable->getColumn('col1')->getDefault());
     }
+
+    public function testGenerateAnIndexWithPartialColumnLength() : void
+    {
+        $this->_conn->query('DROP TABLE IF EXISTS test_partial_column_index');
```
morozov: This is not needed. `dropAndCreateTable()` will drop it.

morozov (changes requested):

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
bburnichon: I removed duplication in branches in a single commit. It also removes `'columns'` duplication. Not sure it makes code more understandable though. Could be reverted easily if needed.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2465,17 +2467,25 @@ public function getCustomTypeDeclarationSQL(array $columnDef)
...
      */
-    public function getIndexFieldDeclarationListSQL(array $fields)
+    public function getIndexFieldDeclarationListSQL($columnsOrIndex) : string
     {
+        if ($columnsOrIndex instanceof Index) {
+            return implode(', ', $columnsOrIndex->getQuotedColumns($this));
+        }
+
+        if (! is_iterable($columnsOrIndex)) {
+            throw new \InvalidArgumentException('Fields argument should be an Index or iterable.');
```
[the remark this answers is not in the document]
bburnichon: I'm not sure also but PHPCS rules enforce `mixed[]` type hinting as it does not accept simple `array`. It looks to me that `mixed[]` is type hinting for an iterable, not an array.

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -1150,4 +1151,35 @@ public function getDefaultTransactionIsolationLevel()
...
+        }
+
+        $subParts = $columnsOrIndex->hasOption('lengths') ? $columnsOrIndex->getOption('lengths') : [];
+
+        $ret = [];
+
+        foreach ($columnsOrIndex->getQuotedColumns($this) as $quotedColumn) {
+            $length = array_shift($subParts);
+
+            if ($length !== null) {
```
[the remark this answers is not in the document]
bburnichon: hum, although I do agree that it would be nice to have a base implementation in base class, current way of declaring SQL indexes with partial columns is MySQL specific and does not belong to base class

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -868,17 +868,26 @@ protected function _getPortableTableIndexesList($tableIndexRows, $tableName=null
...
                     'flags' => $tableIndex['flags'] ?? [],
-                    'options' => isset($tableIndex['where']) ? ['where' => $tableIndex['where']] : [],
+                    'options' => $options,
                 ];
-            } else {
-                $result[$keyName]['columns'][] = $tableIndex['column_name'];
             }
+
+            $result[$keyName]['columns'][]            = $tableIndex['column_name'];
+            $result[$keyName]['options']['lengths'][] = $tableIndex['sub_part'] ?? null;
```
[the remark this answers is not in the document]
bburnichon: So is the `where` clause which belongs to PostGre, isn't it? The current way it is done prevent using `options` index to add platform specific code. Will try finding a better way

morozov:
@bburnichon please squash commits and I'll merge it.

morozov (dismissed):

bburnichon:
@morozov Squashed and all tests green! \o/

morozov (approved):

morozov:
Thank you @bburnichon, merged.

exptom:
This doesn't seem to work on primary key definitions. e.g. I have defined this:
```
<id name="publicKeyCredentialId" type="base64">
            <options>
                <option name="lengths">255</option>
            </options>
</id>
```
The lengths option is silently ignored.

bburnichon:
@exptom I did not check real behavior but I think that it is not valid anyway. A primary key cannot be partial as it need to ensure all values are unique.

exptom:
@bburnichon Yes, you are right, sorry I need to rethink my schema!

exptom:
@bburnichon Sorry to bring this up again, but I have refactored my entity and I am now just trying to apply the index lengths option to a normal index, like this:

```
<field name="publicKeyCredentialId" type="base64" />
<indexes>
            <index name="publicKeyCredentialId" columns="publicKeyCredentialId">
                <options>
                    <option name="lengths">200</option>
                </options>
            </index>
</indexes>
```

However this still doesn't work because I get this error:
> PHP Warning:  array_shift() expects parameter 1 to be array, string given in /home/tom/Workspaces/PHP_Libs/zf-mvc-users/vendor/doctrine/dbal/lib/Doctrine/DBAL/Schema/Index.php on line 109

which is because a string "200" is passed in rather than an array containing 200. How would I specify the index lengths correctly in the xml config?

bburnichon:
I am afraid that it is because this feature was not implemented to XmlMapping. This is out of the scope of this PR. Have a look at https://github.com/doctrine/orm/blob/8c259ea5cb632dbb57001b2262048ae7fa52b102/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php#L993

gggeek:
@exptom @bburnichon I just tested creation of a PK(191) on a varchar(255) column and it does work (at least in a mysql 5.6 db configured so that creation of the PK on the full length of the same col fails).

So to have support allowing partial-lenght PKs on Mysql shall I open a separate ticket and PR ?

bburnichon:
@gggeek Even though MySQL allows it does not mean it should. A primary key must ensure all values are different, a partial key leaves room to add duplicates. Anyway, This ticket was merged around 2 years ago, if you wish to add support for partial-length primary key, you shall open a new ticket.

gggeek:
@bburnichon I don't disagree with you - this is possibly one of those features of MySQL that come from those dark times when it was developed according to the 'make it stick' philosophy, regardless of correctness of data and query results ;-)
However it is also a useful feature to work around the dead stupid limitation of maximum index length on varchar columns - especially useful when the db config is such that this limit is as low as 190 characters...
Given the fact that both MySQL and DBAL already allow the definition of a Unique index that checks only a part of the data - a blatant logical nonsense - it stands to reason that we could allow as well the same 'relaxed' constraint in the definition of a PK.
I have thus created issue #4405

devnix:
> @bburnichon Sorry to bring this up again, but I have refactored my entity and I am now just trying to apply the index lengths option to a normal index, like this:
> 
> ```
> <field name="publicKeyCredentialId" type="base64" />
> <indexes>
>             <index name="publicKeyCredentialId" columns="publicKeyCredentialId">
>                 <options>
>                     <option name="lengths">200</option>
>                 </options>
>             </index>
> </indexes>
> ```
> 
> However this still doesn't work because I get this error:
> 
> > PHP Warning:  array_shift() expects parameter 1 to be array, string given in /home/tom/Workspaces/PHP_Libs/zf-mvc-users/vendor/doctrine/dbal/lib/Doctrine/DBAL/Schema/Index.php on line 109
> 
> which is because a string "200" is passed in rather than an array containing 200. How would I specify the index lengths correctly in the xml config?



> I am afraid that it is because this feature was not implemented to XmlMapping. This is out of the scope of this PR. Have a look at https://github.com/doctrine/orm/blob/8c259ea5cb632dbb57001b2262048ae7fa52b102/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php#L993

Hi @exptom, @bburnichon, sorry for reviving this thread after so long. I guess this comment could be out of scope, but I would like to note that I've been able to use this in XML:

```xml
<index name="test" columns="foo,bar,baz">
    <options>
        <option name="lengths">
            <value>100</value>
            <value>100</value>
        </option>
    </options>
</index>
```

generates the following diff:
```sql
CREATE INDEX test ON eventos_rol (foo(100), bar(100), baz);
```

I've still not found how to generate this diff:
```sql
CREATE INDEX test ON eventos_rol (foo, bar(100), baz);
```
but I guess this is a big step.

I guess that this possibility is not documented in the schema:
image

bburnichon:
Hello.

I guess it expects a `null` value. So maybe it could work with a `<value/>`
as first element of list. If it does not work, you will have to reorder the
columns so that first ones have their length given.

Ie:
```

<index name="test" columns="foo,bar,baz">

    <options>

        <option name="lengths">

            <value/>

            <value>100</value>

        </option>

    </options>

</index>

```

Or


```
<index name="test" columns="bar,foo,baz">
    <options>
        <option name="lengths">
            <value>100</value>
        </option>
    </options>
</index>
```

A third way is to give the actual length of the column. Driver should
detect that a complete column is indexed.

Hope this helps.

devnix:
Hi @bburnichon, thank you for your fast answer!

In my case, all these options
```xml
<option name="lengths">
    <value></value>
    <value>100</value>
</option>
```
```xml
<option name="lengths">
    <value/>
    <value>100</value>
</option>
```
```xml
<option name="lengths">
    <value xsi:nil="true"/>
    <value>100</value>
</option>
```

are throwing me an `Argument 1 passed to Doctrine\DBAL\Schema\Index::Doctrine\DBAL\Schema\{closure}() must be of the type int or null, string given` exception, as I guess it's being casted to string in https://github.com/doctrine/orm/blob/8c259ea5cb632dbb57001b2262048ae7fa52b102/lib/Doctrine/ORM/Mapping/Driver/XmlDriver.php#L1002

I have doubts if there is a current mechanism here to achieve this. I'm not inverting the order of the params, as I'm trying to map an already existing legacy database, and I'm trying to null the schema diff. I've tested the third approach and it generates a diff too.

bburnichon:
I guess you will have to update `parseOptions` recursion and detect attribute name like it is done for `fixed` and `unsigned`. Add a special case for `lengths` and add a special evaluation method (`evaluateIntegerOrNull` in the same way `evaluateBoolean`).

PR #2413, opened 2016-06-14
Ensure deterministic results in the `ModifyLimitQueryTest`, which may otherwise fail depending on database settings and provisioning

Deltachaos:
As changed in PR doctrine/dbal#2363 it would be better if the tests in ModifyLimitQueryTest return more deterministic results.


deeky666:
@Deltachaos thanks!

Deltachaos:
@deeky666 thanks for merging!

PR #2414, opened 2016-06-14
Add enum type

Deltachaos:
Depends on doctrine/common#723

### DIFF
--- a/lib/Doctrine/DBAL/Types/Type.php
+++ b/lib/Doctrine/DBAL/Types/Type.php
@@ -53,6 +53,7 @@ abstract class Type
     const FLOAT = 'float';
     const GUID = 'guid';
     const DATEINTERVAL = 'dateinterval';
+    const ENUM = 'enum';
 
     /**
      * Map of already instantiated type objects. One instance per type (flyweight).
@@ -87,6 +88,7 @@ abstract class Type
         self::BLOB => 'Doctrine\DBAL\Types\BlobType',
         self::GUID => 'Doctrine\DBAL\Types\GuidType',
         self::DATEINTERVAL => 'Doctrine\DBAL\Types\DateIntervalType',
+        self::ENUM => 'Doctrine\DBAL\Types\EnumType',
     );
 
     /**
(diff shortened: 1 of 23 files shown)


lib/Doctrine/DBAL/Types/Type.php
```diff
@@ -114,12 +116,13 @@ public function convertToDatabaseValue($value, AbstractPlatform $platform)
...
-     * @param mixed                                     $value    The value to convert.
-     * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform The currently used database platform.
+     * @param mixed                                     $value            The value to convert.
+     * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform         The currently used database platform.
+     * @param array|null                                $fieldDeclaration The field declaration.
      *
      * @return mixed The PHP representation of the value.
      */
-    public function convertToPHPValue($value, AbstractPlatform $platform)
+    public function convertToPHPValue($value, AbstractPlatform $platform, array $fieldDeclaration = null)
```
stof: Such BC break is a no-go in Doctrine DBAL 2.x. It would break all projects using a custom type class

stof:
the schema comparator does not seem to handle enums either in your PR

stof:
Btw, it is clear that you haven't even tried to run tests (not even talking about testing your new feature...), as the tests are broken precisely because of the BC break you introduced

Deltachaos:
@stof yep thats true.

First of all i would like to say sorry for creating such a incomplete PR.
The main purpose of this was to see the reaction to introducing a Enum type.
Handling enums seems to be one of the most frequently asked questions. One the other hand,  there is no standard solution for PHP to define enums (not talking about a native one).
Enums would really increase the stability and readability of code.
The question is: What has to happen that we get a well accepted and supported solution for this all over the php community?

We probably need to show a need for a common solution. The only organization that could introduce such a common solution for enums is probably the PHP FIG. But waiting for a PSR to solve the enum issue is probably not sufficient. On the other hand could a doctrine implementation for enums that integrates with other enum libraries speed up the process to find a general solution.

Im am now interested about your thoughts and ideas.

Ocramius:
Closing here - remarks similar to https://github.com/doctrine/common/pull/723

This should be brought up again once ENUMs in userland (before DB) are a solved problem.

PR #2416, opened 2016-06-16
Fix OracleSchemaManager to return the corrent primary key name

liberostelios:
For some odd reason (to my eye), OracleSchemaManager forces all primary keys to be called as 'primary', no matter what the original name is. I believe that this serves no purpose and therefore should be changed.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
@@ -78,7 +78,6 @@ class OracleSchemaManager extends AbstractSchemaManager
             $keyName = strtolower($tableIndex['name']);
 
             if (strtolower($tableIndex['is_primary']) == "p") {
-                $keyName = 'primary';
                 $buffer['primary'] = true;
                 $buffer['non_unique'] = false;
             } else {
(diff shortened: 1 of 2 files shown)


deeky666:
This seems to be an old leftover from pre 2.0 release (https://github.com/doctrine/dbal/commit/cfea7883e07c1497920eec2e8eca8417a2ffacb5). Looks like this change is valid (also considering that other vendor implementations use use the index name for primary keys, too). But still needs a test of course as @Ocramius pointed out.

deeky666:
Also not sure if this is really a bug. We do not use primary key names in DBAL, not even the comparator is diffing it. So I guess this rather is an improvement... @Ocramius

liberostelios:
All right, then. I will try to implement some test for it.

liberostelios:
Let me know if this is considered a legitimate test (i really new at testing).

soyuka:
Indeed, I've seen this with something like `DROP INDEX primary`, where it doesn't refer to anything.

About the test, you could maybe test that dropping the id on oracle gives the correct sql result?

morozov:
Closing as stale.

morozov:
Closing as stale and missing functional tests.

PR #2418, opened 2016-06-17
Update PostgreSqlPlatform.php

xujif:
if use getOldColumnName will create a new instance of  Identifier and will lost the "$_quoted" info.  
so if the colume name contains upper letter will not be quoted.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -525,7 +525,7 @@ class PostgreSqlPlatform extends AbstractPlatform
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


Ocramius:
@xujif while the change seems legit, it needs to be backed by a test case.

xujif:
@Ocramius  good job. but the packagist has not been updated

deeky666:
@xujif looking at the code of `ColumnDiff::getOldColumnName()`, your statement is not true. May it be you missed [the recent patch](https://github.com/doctrine/dbal/commit/d6317d35f990d05c1f07013be71924ae2519f498) to quotation? Please have a look if you still face the issue with our current master.

This looks like a duplicate.

Ocramius:
Linking related #1206 and #2288, which are the possible duplication targets

deeky666:
@xujif ping

PR #2422, opened 2016-06-23
Map custom exceptions for the "no default value" (1364) error in the MySQL drivers

MorrisJobke:
When following error message is shown, then it will get converted to the
correct Exception classs of NotNullConstraintViolationException:

  PDOException: SQLSTATE[HY000]: General error: 1364 Field '...' doesn't have a
  default value

The mentioned field has a Not Null constraint but no default value. I tested this with the latest mysql and mariadb docker container today (and with the mysql and mysqlnd PHP extension of the current Debian stable).

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
@@ -112,6 +112,7 @@ abstract class AbstractMySQLDriver implements Driver, ExceptionConverterDriver, 
             case '1171':
             case '1252':
             case '1263':
+            case '1364':
             case '1566':
                 return new Exception\NotNullConstraintViolationException($message, $exception);
         }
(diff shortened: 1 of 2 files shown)


kimhemsoe:
Tests?

MorrisJobke:
> Sorry my mistake, did not see that you changed the test file. I will take a look at later and report back.

Okay. I can also try to add some sort of integration test. Where should they then be added?

MorrisJobke:
> I think the error only applies if the STRICT_TRANS_TABLES SQL mode is enabled in MySQL.

Thanks a lot for this information and for accepting this :)

MorrisJobke:
@deeky666 Should I prepare a backport for this to the 2.5 version?

deeky666:
@MorrisJobke this is not a bug, this won't land in 2.5

MorrisJobke:
> @MorrisJobke this is not a bug, this won't land in 2.5

Okay :) Thanks anyways

PR #2424, opened 2016-06-24
change DB2Connection _conn property from private to protected

cassvail:
Allow classes extending DB2Connection to access _conn resource, i.e. necessary in IBMi systems to use XMLToolkit library

### DIFF
--- a/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
+++ b/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
@@ -27,7 +27,7 @@ class DB2Connection implements Connection, ServerInfoAwareConnection
     /**
      * @var resource
      */
-    private $_conn = null;
+    protected $_conn = null;
 
     /**
      * @param array  $params


kimhemsoe:
We would rather want you to just copy the whole driver and implement whatever you need. It is nice for us that we do not have to keep BC internal in drivers, if we wish to make future changes.

Closing as wont fix.

cassvail:
I understand your point, but this edit is not about BC of the Driver logic, because the ibm_db2 in use is the same driver which is provided by IBM as is, with all the methods and behaviour. If this driver will change the same will happen to the whole driver class. 
The goal here is to simply obtain access to the connection property without modification. The _conn is a resource. I hardly think that a resource could change in time, but accessing it is needed in some particular situations.
IMO reimplementing the whole driver logic would be breaking the principle of OOP and code reusability, because the driver itself is the same and so is the driver logic, without concerns for BC.
Maybe it could be interesting to have a second thought on this subject.
In alternative it would be possible to implement a getter getWrappedResourceHandle for the _conn property as implemented in https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php ?
Thanks for the attention

Ocramius:
Not gonna happen: if you need to do ugly/custom things with the underlying
DB connection, then please do so in your custom connection object instead,
or allow passing in the resource from the outside. Doctrine won't make it
more accessible to you.

PR #2432, opened 2016-06-29
Removed nearly all `call_user_*` usages

kimhemsoe:


deeky666:
LGTM although I am not sure about HHVM compat because we had issues with variable numbers of arguments calls...

kimhemsoe:
I did not change any method signatures and travis hhvm jobs did not fail in a new way.

deeky666:
Yeah the related failures were not related to method signatures but to method calls. Not sure if the tests do cover those cases.

deeky666:
Okay merging, thanks @kimhemsoe !

PR #2433, opened 2016-06-30
Failing test regarding ORA-24816, php 7 and CLOB

soyuka:
Ref bug discovery https://github.com/doctrine/dbal/pull/2386
Ref php bug https://bugs.php.net/bug.php?id=72524

Because @deeky666 just had the right words for this issue, I'm quoting him:

> There seems to be a difference between PHP5 and PHP7 in oci8 when binding `NULL` values to `LONG/LOB` type columns.
> In PHP5 it was possible to execute the following `INSERT` statement without any error (sample):
> 
> ```
> INSERT INTO mytable VALUES (:clob_col, :varchar2_col)
> ```
> 
> If the value bound to parameter `:varchar2_col` is `NULL`, the following error is triggered (no matter what value the parameter `:clob_col` is bound to):
> 
> ```
> ORA-24816: Expanded non LONG bind data supplied after actual LONG or LOB column
> ```
> 
> If you switch the parameter order and both parameter values are bound to `NULL`, the same error occurs:
> 
> ```
> INSERT INTO mytable VALUES (:varchar2_col, :clob_col)
> ```
> 
> So no matter what order, not matter what value is bound to `:clob_col`, if `:varchar2_col` is bound to NULL, the error occurs.
> If you have only one `VARCHAR2` column or two (without a `CLOB` column) there also is no error.

This Pull request adds a test case for this scenario by using the DBAL, it's obviously failing for now.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/Ticket/PHP72524Test.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Ticket/PHP72524Test.php
@@ -1,0 +1,96 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Functional\Ticket;
+
+use Doctrine\DBAL\Types\Type;
+
+/**
+ * @group DBAL-2386
+ */
+class DBAL2386Test extends \Doctrine\Tests\DbalFunctionalTestCase
+{
+    protected function setUp()
+    {
+        parent::setUp();
+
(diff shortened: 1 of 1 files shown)


deeky666:
@soyuka thanks for your investigation and help. Much appreciated!

soyuka:
@deeky666 It was a pleasure, thanks to you for contributing to the investigation!

deeky666:
Handled in #2434.

PR #2434, opened 2016-06-30
Keep references to bound parameter values in oci8 driver

### LINKED ISSUE #2261: OCI8 - bindValue overwrite previous values

deeky666:
For PHP7 compatibility. The previously failing tests pass now.

Closes #2261
Closes #2262
Closes #2386

### DIFF
--- a/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
+++ b/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
@@ -72,6 +72,15 @@ class OCI8Statement implements \IteratorAggregate, Statement
     protected $_paramMap = array();
 
     /**
+     * Holds references to bound parameter values.
+     *
+     * This is a new requirement for PHP7's oci8 extension that prevents bound values from being garbage collected.
+     *
+     * @var array
+     */
+    private $boundValues = array();
+
+    /**
      * Creates a new OCI8Statement that uses the given connection handle and SQL statement.
      *
      * @param resource                                  $dbh       The connection handle.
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -72,6 +72,15 @@ class OCI8Statement implements \IteratorAggregate, Statement
     protected $_paramMap = array();
 
     /**
+     * Holds references to bound parameter values.
+     *
+     * This is a new requirement for PHP7's oci8 extension.
```
Ocramius: You should mention here that this is just to prevent GC to kick in

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -72,6 +72,15 @@ class OCI8Statement implements \IteratorAggregate, Statement
...
+     *
+     * @var array
+     */
+    private $boundValues = array();
```
Ocramius: Can safely use `[]`

Ocramius: Nvm, I see that 2.5.x still supports PHP 5.3.x

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
deeky666: @Ocramius better like this?

Ocramius:
LGTM! 👍 

Thanks @deeky666!

`master`: doctrine/dbal@69b57859c8674f2c8ba3798a2541ac18128a0bdb
`2.5`: doctrine/dbal@e05f26d87d4132c73365a48c696b9ff9f823237e

deeky666:
@Ocramius thanks for merging!

Garfield-fr:
Thanks

mmoll:
will there be a 2.5.5 soon?

Ocramius:
@mmoll I'm probably allocating time next week to do a release.

cjbj:
@Ocramius can you see if the proposed patch for PHP 7s' OCI8 in https://bugs.php.net/patch-display.php?bug_id=71148&patch=oci8-php7-bind&revision=latest removes the need for this workaround?

Ocramius:
Possibly, but it needs testing on that patch too (PHPT). The only negative side of this particular patch is that GC is shifted one level up (DBAL statement)

PR #2435, opened 2016-06-30
Replace `getMock()` with `createMock()`

deeky666:
As a result of bumping PHPUnit requirement to `^5.4.6` in https://github.com/doctrine/dbal/pull/2419 we get a lot of deprecation warnings for `PHPUnit_Framework_TestCase::getMock()` in our test suite.

This PR fixes all of them. Also two tests had to be adjusted, because they were doing expectations on non-existing methods which PHPUnit now warns about.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/ConnectionTest.php
@@ -102,13 +102,15 @@ class ConnectionTest extends \Doctrine\Tests\DbalTestCase
 
     public function testConnectDispatchEvent()
     {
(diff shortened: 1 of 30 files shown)


Ocramius:
@deeky666 patch needs bumping PHPUnit dependency to at least `^5.4.0` (pick latest, please)

deeky666:
@Ocramius you did that already in #2419 or am I missing something? =)

deeky666:
Actually the issues fixed here were introduced because of that PR ;)

Ocramius:
@deeky666 indeed, thanks for pointing it out :-)

PR #2437, opened 2016-07-03
Improve the phpdoc on `Connection`

mnapoli:
Improve return type's phpdoc because it wasn't clear from the phpdoc or the online documentation what would happen if no row was found.


mnapoli:
For reference, the online docs: http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/data-retrieval-and-manipulation.html#fetcharray

Ocramius:
@mnapoli thanks!

PR #2439, opened 2016-07-05
WIP: Fix HHVM pdo_mysql exceptions

Deltachaos:


kimhemsoe:
You can hook travis up to your own repos. Less spam here :D

Deltachaos:
@kimhemsoe sorry for that. But i noticed that this seems to be a bug in HHVM anyway and the already have a issue there opened by @deeky666.

One the other side older versions of HHVM only pass the correct ``errorCode`` only as part of the exception message (``SQLSTATE[HY000] [1008] Can't drop database 'doctrine_tests'; database doesn't exist``) so the code i have added here would workaround this problem. One the other hand some messages seem to miss the ``errorCode`` even in the message text.
Therefore i am not shure if we should address this problem anyway.

deeky666:
@Deltachaos as you noticed this is a HHVM bug as it does not adhere to the PHP implementation of error codes correctly. Therefore nothing to do here, we also won't workaround something that is not our fault. They have to fix it.

PR #2442, opened 2016-07-07
Escape identifiers in metadata SQL properly when used as string literal

### LINKED ISSUE #2436: MySqlPlatform::getListTableColumnsSQL() is not escaping names
Hi !

When creating by accident a table that contain a `\` in the name, `abc\` The migration fail to read and escape correctly the table names.

$ php app/console oro:platform:update --env dev --force
Process migrations...

```
[Doctrine\DBAL\Exception\SyntaxErrorException]
An exception occurred while executing 'SELECT COLUMN_NAME AS Field, COLUMN_TYPE AS Type, IS_NULLABLE AS `Null`, COLUMN_KEY AS `Key`, COLUMN_DEFAULT AS `Default`, EXTRA
 AS E  xtra, COLUMN_COMMENT AS Comment, CHARACTER_SET_NAME AS CharacterSet,  OLLATION_NAME AS Collation FROM information_schema.COLUMNS WHERE TABLE_SCHEMA = 'asdf' AND TABLE   _NAME = ''abc\'':

SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the ri
ght syntax to use near ''oro_tag_taxonomy\'' at line 1
```

deeky666:
Closes #2436

@Ocramius I tried to write tests but neither of my approaches felt right. First I tried to `assertContains()` the resulting SQL but that is not applicable to all platforms and methods as the given identifiers to escape are not always used as string literals (heavily depends on the platform's SQL statement) or are not even used at all. Then I tried to write functional tests but that is problematic, too as I cannot test the `$database` identifiers properly on all platforms because that involves creating a database (which is heavy and not supported on all platforms) with escape characters in the name.

Any idea what test to write here or should we take it "as is"?

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/DB2Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/DB2Platform.php
@@ -250,6 +250,8 @@ class DB2Platform extends AbstractPlatform
      */
     public function getListTableColumnsSQL($table, $database = null)
     {
+        $table = $this->quoteStringLiteral($table);
+
         // We do the funky subquery and join syscat.columns.default this crazy way because
         // as of db2 v10, the column is CLOB(64k) and the distinct operator won't allow a CLOB,
         // it wants shorter stuff like a varchar.
@@ -282,7 +284,7 @@ class DB2Platform extends AbstractPlatform
                    ON (c.tabschema = k.tabschema
                        AND c.tabname = k.tabname
                        AND c.colname = k.colname)
-               WHERE UPPER(c.tabname) = UPPER('" . $table . "')
+               WHERE UPPER(c.tabname) = UPPER(" . $table . ")
                ORDER BY c.colno
              ) subq
           JOIN syscat.columns cols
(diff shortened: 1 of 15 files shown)


kimhemsoe:
I agree that it is a bug, but this very much looks like a BC break for me. If i remember correct i even wrote a few scripts many years ago where i hit this and worked around it by calling quote manually. If ive have done that possible others have as well.

deeky666:
@kimhemsoe yeah but it is wrong to do a `quoteStringLiteral()` for an identifier before passing it to the `getList*SQL()` methods because whether or not the identifier is treated as string literal is implementation detail and platform specific. The user should not know how the underlying SQL is built. Also the platform API does not make any statements about whether a given identifier should be quoted/escaped or the like. So I don't think this is a BC break but rather a misuse of the API then. It would be wrong if the (mis)usage of our API prevents us from fixing bugs.

Ocramius:
@deeky666 I don't think that there is a sensible way to test this. I'd still apply some minimal assertions around the identifiers inside the SQL though, as that would show BC breaks in the quoting immediately, even though the tests may seem fragile.

deeky666:
@Ocramius done.

Ocramius:
Merged!

`master`: doctrine/dbal@b6b247a3c6039f9364d061f15ef6c7aaef08d9c3
`2.5`: doctrine/dbal@4bca2264cc657ea18ff147e23e09adb97f56df83

deeky666:
@Ocramius thanks

PR #2447, opened 2016-07-12
connect without host param or with 2 hosts

andipohl:
buildDsn has to recognize, that no host parameter is given. A tcpip-broadcast has to be made to find the right server/database.
Furthermore if I have a db-mirror-system, I have to get both hosts comma-separated as host parameter with or without ports (HOST=host1:1111,host2:2345)

### DIFF
--- a/lib/Doctrine/DBAL/Driver/SQLAnywhere/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/SQLAnywhere/Driver.php
@@ -81,19 +81,37 @@ class Driver extends AbstractSQLAnywhereDriver
      */
     private function buildDsn($host, $port, $server, $dbname, $username = null, $password = null, array $driverOptions = array())
     {
-        $host = $host ?: 'localhost';
-        $port = $port ?: 2638;
+        $hostPart = "";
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Driver/SQLAnywhere/Driver.php
```diff
@@ -81,19 +81,37 @@ public function getName()
...
+
+        //if host is empty, "LINKS=tcpip()" has to be set and SQLAnywhere did broadcasts to find the right DB
+        //port-parameter is only used if host has no colon and has no comma (db-mirror configuration)
+        if (! empty($host)) {
+            $hostPart = 'HOST='.$host;
+            if (false === strpos($host,':') && false === strpos($host,',')) {
+                $hostPart .= isset($port)?':'.$port:':2638';
+            }
+        } else {
+            $linksPart = ';LINKS=tcpip()';
```
Ocramius: Can set it as default, no need for the `else` clause

morozov:
Obsolete as of https://github.com/doctrine/dbal/pull/4077.

PR #2448, opened 2016-07-13
Fallback to other slaves if connection is not possible

Deltachaos:


lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -213,7 +213,24 @@ protected function connectTo($connectionName)
         $user = isset($connectionParams['user']) ? $connectionParams['user'] : null;
         $password = isset($connectionParams['password']) ? $connectionParams['password'] : null;
 
-        return $this->_driver->connect($connectionParams, $user, $password, $driverOptions);
+        if ($connectionName == 'slave') {
+            while(true) {
```
Ocramius: This will try connecting forever...

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -213,7 +213,24 @@ protected function connectTo($connectionName)
         $user = isset($connectionParams['user']) ? $connectionParams['user'] : null;
         $password = isset($connectionParams['password']) ? $connectionParams['password'] : null;
 
-        return $this->_driver->connect($connectionParams, $user, $password, $driverOptions);
+        if ($connectionName == 'slave') {
```
Ocramius: This entire block should be moved to a private method

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -213,7 +213,24 @@ protected function connectTo($connectionName)
...
+                        throw new ConnectionException('No slaves are available', $e->getPrevious());
+                    }
+
+                    $connectionParams = array_pop($params['slaves']);
+
+                    $user = isset($connectionParams['user']) ? $connectionParams['user'] : null;
+                    $password = isset($connectionParams['password']) ? $connectionParams['password'] : null;
+                }
+            }
+        } else {
```
Ocramius: No need for `else`

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
Deltachaos: If the connection in line 219 is successfull it exits the loop by the return. If it fails it leaves the loop by throwing an exception as soon as no slaves are left. Did i miss any case? Or do you just prefer to not have a ``while(true)`` in the code to preserve faults in futere changes?

Ocramius: Ah, I missed the `array_pop()`. Should be `do { } while ($connectionParams = array_pop($params['slaves']))` then

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -204,48 +204,67 @@ public function connect($connectionName = null)
...
-        if ($connectionName === 'master') {
-            return $params['master'];
+        $params = $this->getParams();
+
+        $driverOptions = isset($params['driverOptions']) ? $params['driverOptions'] : array();
+
+        shuffle($params['slaves']);
+
+        do {
+            $connectionParams = array_pop($params['slaves']);
```
Ocramius: If you design it with this approach, then the loop can be `while ($connectionParams = array_pop($params['slaves'])) {`

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -217,18 +234,37 @@ protected function connectTo($connectionName)
...
+        }
+
+        $previous = null;
+
+        if (isset($e) && $e->getPrevious() instanceof \Doctrine\DBAL\Driver\DriverException) {
+            $previous = $e->getPrevious();
         }
 
-        return $params['slaves'][array_rand($params['slaves'])];
+        throw new ConnectionException('No slaves are available', $previous);
```
Ocramius: Wondering if recording all previous exceptions would make sense here.

Deltachaos: I was not shure about this as well. Thought that it is always better to keep the history for debugging reasons.

Ocramius:
@deeky666 besides missing tests, do you have any opinion on this? Looks sane to me...

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -217,18 +235,37 @@ protected function connectTo($connectionName)
...
      *
-     * @return mixed
+     * @param array $params
+     * @param array $driverOptions
+     *
+     * @return Driver\Connection
+     *
+     * @throws ConnectionException
      */
-    protected function chooseConnectionConfiguration($connectionName, $params)
```
deeky666: This is a BC break. Can we avoid that?

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -217,18 +235,37 @@ protected function connectTo($connectionName)
...
+    private function connectToSlave($params, $driverOptions)
     {
-        if ($connectionName === 'master') {
-            return $params['master'];
+        shuffle($params['slaves']);
```
deeky666: Why do we need to shuffle here?

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -217,18 +235,37 @@ protected function connectTo($connectionName)
...
+
+        while($connectionParams = array_pop($params['slaves'])) {
+            $user = isset($connectionParams['user']) ? $connectionParams['user'] : null;
+            $password = isset($connectionParams['password']) ? $connectionParams['password'] : null;
+
+            try {
+                return $this->_driver->connect($connectionParams, $user, $password, $driverOptions);
```
deeky666: Can we try to avoid the code duplication here (this block)?

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -217,18 +235,37 @@ protected function connectTo($connectionName)
...
+     * @throws ConnectionException
      */
-    protected function chooseConnectionConfiguration($connectionName, $params)
+    private function connectToSlave($params, $driverOptions)
     {
-        if ($connectionName === 'master') {
-            return $params['master'];
+        shuffle($params['slaves']);
+
+        while($connectionParams = array_pop($params['slaves'])) {
```
deeky666: Whitespace after `while`

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
deeky666: Agree with @Ocramius

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
Ocramius: This is a replacement of the `array_rand()` stuff that was done before. See removed diffs

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -217,18 +235,37 @@ protected function connectTo($connectionName)
     }
 
     /**
-     * @param string $connectionName
-     * @param array  $params
+     * Connets to a random slave and tries others until if found a server that is alive
```
Ocramius: Typo in `Connets`

deeky666:
@Ocramius looks sane to me but I am not quite sure if there might be use cases where it matters for the user if one slave connection is not avaliable. Because then this information would be suppressed. Might be better to have some kind of strategy object for this...

Ocramius:
> Might be better to have some kind of strategy object for this...

Something like these?

 * `FirstAvailableConnector`
 * `RandomConnector`

Makes sense, but is it worth asking @Deltachaos to build it like that right now?

deeky666:
@Ocramius yep this is what I was thinking about.

Deltachaos:
@deeky666 @Ocramius ``FirstAvailableConnector`` sounds better to me. ``RandomConnector`` is not really unambiguous, because the current implementation is random as well but fails if the picked slave is not available.
My use case was that i have defined the master in the slaves array as well to achieve that my application is still alive (at least for reading opterations) is the master server goes done.
One the other hand, it feels natural to me that if there are multiple slaves available, that it should not harm the application if one of them goes down because a other slave is as good as the first one that was picked.

I will change the code to add an new ``FirstAvailableConnector`` in the next days if this is ok for you. I also will try to add some tests.
One question is still open: How do we make the new strategy selectable in the DoctrineBundle?

Deltachaos:
@Ocramius @deeky666 I have added the tests for this PR

Deltachaos:
@deeky666 @Ocramius still something that we could merge? I have rebased the branch and created a backport version for 2.x in #4004

Issue #1609, opened 2013-01-14
DBAL-415: [GH-248] Enable multiple postgresql datetime formats


doctrinebot:
Jira issue originally created by user @beberlei:

This issue is created automatically through a Github pull request on behalf of cedriclombardot:

  Url: https://github.com/doctrine/dbal/pull/248

Message:

Postgre use ISO 8601 date formats so some times can be Y-m-d H:i:s or sometime Y-m-d H:i:s.u

PR #2450, opened 2016-07-15
Add immutable date types support

### LINKED ISSUE #1882: DBAL-662: Supporting PHP 5.5 DateTimeImmutable
Jira issue originally created by user stof:

Introducing new types converting dates into a `DateTimeImmutable` rather than a DateTime could be useful for applications prefering the use of immutable datetimes (and relying on PHP 5.5+).

The existing types already support setting a `DateTimeImmutable` in a field mapped with the datetime type, as it does not check the value against DateTime but only expects a format method. but the conversion from DB to PHP will always produce a mutable DateTime instance, thus preventing to use the immutable flavour consistently.

Not that this is a low priority issue as any code interacting with third party packages will probably need to use DateTime anyway because of the typehints (typehinting DateTimeInterface would not work if you need to keep compatibility with 5.4)

If DBAL 3.0 bumps the minimal supported version to 5.5 (which might be possible depending of the release date of 3.0), we could decide to break BC and use the immutable flavour in the datetime type directly.

deeky666:
As we have bumped PHP requirement to 5.6+ in 2.6, we now can finally add immutable variants of date types.

Closes #1882

### DIFF
--- a/lib/Doctrine/DBAL/Types/DateImmutableType.php
+++ b/lib/Doctrine/DBAL/Types/DateImmutableType.php
@@ -1,0 +1,89 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Types;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+
+/**
+ * Immutable type of {@see DateType}.
+ *
+ * @since  2.6
+ * @author Steve Müller <deeky666@googlemail.com>
+ */
+class DateImmutableType extends DateType
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        return Type::DATE_IMMUTABLE;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToDatabaseValue($value, AbstractPlatform $platform)
+    {
+        if (null === $value) {
+            return $value;
+        }
+
+        if ($value instanceof \DateTimeImmutable) {
+            return $value->format($platform->getDateFormatString());
+        }
+
+        throw ConversionException::conversionFailedInvalidType(
+            $value,
+            $this->getName(),
+            ['null', \DateTimeImmutable::class]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToPHPValue($value, AbstractPlatform $platform)
+    {
+        if ($value === null || $value instanceof \DateTimeImmutable) {
+            return $value;
+        }
+
+        $dateTime = \DateTimeImmutable::createFromFormat('!' . $platform->getDateFormatString(), $value);
+
+        if (! $dateTime) {
+            throw ConversionException::conversionFailedFormat(
+                $value,
+                $this->getName(),
+                $platform->getDateFormatString()
+            );
+        }
+
+        return $dateTime;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function requiresSQLCommentHint(AbstractPlatform $platform)
+    {
+        return true;
+    }
+}
--- a/lib/Doctrine/DBAL/Types/DateTimeImmutableType.php
+++ b/lib/Doctrine/DBAL/Types/DateTimeImmutableType.php
@@ -1,0 +1,93 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Types;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+
+/**
+ * Immutable type of {@see DateTimeType}.
+ *
+ * @since  2.6
+ * @author Steve Müller <deeky666@googlemail.com>
+ */
+class DateTimeImmutableType extends DateTimeType
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        return Type::DATETIME_IMMUTABLE;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToDatabaseValue($value, AbstractPlatform $platform)
+    {
+        if (null === $value) {
+            return $value;
+        }
+
+        if ($value instanceof \DateTimeImmutable) {
+            return $value->format($platform->getDateTimeFormatString());
+        }
+
+        throw ConversionException::conversionFailedInvalidType(
+            $value,
+            $this->getName(),
+            ['null', \DateTimeImmutable::class]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToPHPValue($value, AbstractPlatform $platform)
+    {
+        if ($value === null || $value instanceof \DateTimeImmutable) {
+            return $value;
+        }
+
+        $dateTime = \DateTimeImmutable::createFromFormat($platform->getDateTimeFormatString(), $value);
+
+        if (! $dateTime) {
+            $dateTime = \date_create_immutable($value);
+        }
+
+        if (! $dateTime) {
+            throw ConversionException::conversionFailedFormat(
+                $value,
+                $this->getName(),
+                $platform->getDateTimeFormatString()
+            );
+        }
+
+        return $dateTime;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function requiresSQLCommentHint(AbstractPlatform $platform)
+    {
+        return true;
+    }
+}
--- a/lib/Doctrine/DBAL/Types/DateTimeTzImmutableType.php
+++ b/lib/Doctrine/DBAL/Types/DateTimeTzImmutableType.php
@@ -1,0 +1,89 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Types;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+
+/**
+ * Immutable type of {@see DateTimeTzType}.
+ *
+ * @since  2.6
+ * @author Steve Müller <deeky666@googlemail.com>
+ */
+class DateTimeTzImmutableType extends DateTimeTzType
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        return Type::DATETIMETZ_IMMUTABLE;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToDatabaseValue($value, AbstractPlatform $platform)
+    {
+        if (null === $value) {
+            return $value;
+        }
+
+        if ($value instanceof \DateTimeImmutable) {
+            return $value->format($platform->getDateTimeTzFormatString());
+        }
+
+        throw ConversionException::conversionFailedInvalidType(
+            $value,
+            $this->getName(),
+            ['null', \DateTimeImmutable::class]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToPHPValue($value, AbstractPlatform $platform)
+    {
+        if ($value === null || $value instanceof \DateTimeImmutable) {
+            return $value;
+        }
+
+        $dateTime = \DateTimeImmutable::createFromFormat($platform->getDateTimeTzFormatString(), $value);
+
+        if (! $dateTime) {
+            throw ConversionException::conversionFailedFormat(
+                $value,
+                $this->getName(),
+                $platform->getDateTimeTzFormatString()
+            );
+        }
+
+        return $dateTime;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function requiresSQLCommentHint(AbstractPlatform $platform)
+    {
+        return true;
+    }
+}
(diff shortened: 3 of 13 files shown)


Deltachaos:
If we would merge #2362 and #2365 we could try to re use the platform methods to to generate the immutable objects. Currently we have a lot of code duplication here. This could be reduced in my opinion. Additionaly i dont see anything that holds #2362 from getting merged.

deeky666:
@Deltachaos I don't see how merging those PRs helps with reusability as those deal with mutable `\DateTime` objects. We would have to introduced another 8! methods to `AbstractPlatform` which makes it 16! new methods overall already. While I see the advantage with your other PRs, I am still not happy with adding some many new methods to the platform (which is public API and has to be maintained afterwards!).

Deltachaos:
@deeky666 you could reuse the methods in ``AbstractPlatform`` and convert the result into a immutable object by ``DateTimeImmutable::createFromMutable``.
By using this method we would not duplicate the code for converting a database date type to a php type just because we have two different type of date objects we would like to convert to.
On the database level there is no difference between mutable and immutable ones.

deeky666:
@Deltachaos unfortunately this is not an option because of performance implications. The type system is highly performance sensitive.

Deltachaos:
@deeky666 ok i see. But there is the question what kind of solution we should use. Because fact is that the current date time conversion is to unflexible for some platforms.
I meet @beberlei and he said one of the team members is currently working on a draft for a new type system that is more flexible. He wanted to bring me in touch with him because i am interested to work on this topic.
But anyway this would be the solution for the long run. In my opinion we should have a solution for this problem also in the 2.x branch. But it would be a good idea to synchronize concepts with the 3.x solution.

May introducing a new ``TypeConverter`` class that can be defined by the ``AbstractPlatform`` similar to the ``getReservedKeywordsClass`` method would be a starting point.
This new class could inherit from a ``AbstractTypeConverter`` and override only the methods where special handling for this type is required for the specific platform.
The instance of ``TypeConverter`` could be revived in the specific types by calling something like ``getTypeConverter`` on the platform. In a new type system the platform and the type converter could be injected into the types at constructing time of the objects.
Now that the type has the ``TypeConverter`` it could simply bypass the convert calls to this ``TypeConverter`` (to seperate all the conversions we could use traits if we would like to eliminate one more call into a other class, or we could have specific type convers like ``DateTimeTypeConverter`` which are created by a factory in the platform using ``getTypeConverter(Type::DATETIME)``).

In this ``TypeConverter`` we could then also merge the conversions for mutable and immutable ones by having something like ``convertFromDateTime(bool $immutable = true)``. And if this is not perfomant enough having a ``if`` switch in the method for creating immutable or mutable ones, at least ``convertDateTimeToDatabaseValue(\DateTimeInterface)`` could be unified by rely on ``\DateTimeInterface`` instead of ``\DateTime``.

Using such an approach we would not have any BC break but a more flexible type system that can considerate all the platform specific stuff whitout bloating everything again.
We could then deprecate a few methods in the ``AbstractPlatform`` like ``convertFromBoolean``, ``convertBooleansToDatabaseValue``, ``convertBooleans``.
We could in a second step also try to move the declaration SQL's into a sperate class.

If you like the idea i could implement this.

Deltachaos:
@deeky666 any thoughts?

lib/Doctrine/DBAL/Types/DateTimeImmutableType.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        if (null === $value) {
+            return $value;
+        }
+
+        if ($value instanceof \DateTimeImmutable) {
+            return $value->format($platform->getDateTimeFormatString());
+        }
+
+        throw ConversionException::conversionFailedInvalidType($value, $this->getName(), ['null', 'DateTimeImmutable']);
```
Ocramius: `DateTimeImmutable::class`

lib/Doctrine/DBAL/Types/DateImmutableType.php
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToPHPValue($value, AbstractPlatform $platform)
+    {
+        if ($value === null || $value instanceof \DateTimeImmutable) {
+            return $value;
+        }
+
+        $val = \DateTimeImmutable::createFromFormat('!'.$platform->getDateFormatString(), $value);
```
Ocramius: Don't abbreviate. Also, spacing around `.`

lib/Doctrine/DBAL/Types/DateTimeImmutableType.php
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToPHPValue($value, AbstractPlatform $platform)
+    {
+        if ($value === null || $value instanceof \DateTimeImmutable) {
+            return $value;
+        }
+
+        $val = \DateTimeImmutable::createFromFormat($platform->getDateTimeFormatString(), $value);
```
Ocramius: Same as above regarding abbreviations

lib/Doctrine/DBAL/Types/DateTimeImmutableType.php
```diff
@@ -0,0 +1,89 @@
...
+
+        if (! $val) {
+            $val = date_create_immutable($value);
```
Ocramius: This block looks really shady. While it may have been discussed before, could you clarify in a comment why it is needed?

Ocramius: Also, use the FQN version of it (`\` prefix)

lib/Doctrine/DBAL/Types/DateTimeTzImmutableType.php
```diff
@@ -0,0 +1,85 @@
...
+    {
+        if (null === $value) {
+            return $value;
+        }
+
+        if ($value instanceof \DateTimeImmutable) {
+            return $value->format($platform->getDateTimeTzFormatString());
+        }
+
+        throw ConversionException::conversionFailedInvalidType($value, $this->getName(), ['null', 'DateTimeImmutable']);
```
Ocramius: `::class`

lib/Doctrine/DBAL/Types/DateTimeTzImmutableType.php
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToPHPValue($value, AbstractPlatform $platform)
+    {
+        if ($value === null || $value instanceof \DateTimeImmutable) {
+            return $value;
+        }
+
+        $val = \DateTimeImmutable::createFromFormat($platform->getDateTimeTzFormatString(), $value);
```
Ocramius: Same as above regarding abbreviations

lib/Doctrine/DBAL/Types/TimeImmutableType.php
```diff
@@ -0,0 +1,85 @@
...
+    {
+        if (null === $value) {
+            return $value;
+        }
+
+        if ($value instanceof \DateTimeImmutable) {
+            return $value->format($platform->getTimeFormatString());
+        }
+
+        throw ConversionException::conversionFailedInvalidType($value, $this->getName(), ['null', 'DateTimeImmutable']);
```
Ocramius: `::class`

lib/Doctrine/DBAL/Types/TimeImmutableType.php
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToPHPValue($value, AbstractPlatform $platform)
+    {
+        if ($value === null || $value instanceof \DateTimeImmutable) {
+            return $value;
+        }
+
+        $val = \DateTimeImmutable::createFromFormat('!' . $platform->getTimeFormatString(), $value);
```
Ocramius: Spacing and abbreviations

lib/Doctrine/DBAL/Types/Type.php
```diff
@@ -80,9 +84,13 @@
         self::STRING => 'Doctrine\DBAL\Types\StringType',
         self::TEXT => 'Doctrine\DBAL\Types\TextType',
         self::DATETIME => 'Doctrine\DBAL\Types\DateTimeType',
+        self::DATETIME_IMMUTABLE => 'Doctrine\DBAL\Types\DateTimeImmutableType',
```
Ocramius: We can safely use `::class` here now (also for all other constants)

lib/Doctrine/DBAL/Types/VarDateTimeImmutableType.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        if (null === $value) {
+            return $value;
+        }
+
+        if ($value instanceof \DateTimeImmutable) {
+            return $value->format($platform->getDateTimeFormatString());
+        }
+
+        throw ConversionException::conversionFailedInvalidType($value, $this->getName(), ['null', 'DateTimeImmutable']);
```
Ocramius: `::class`

lib/Doctrine/DBAL/Types/VarDateTimeImmutableType.php
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function convertToPHPValue($value, AbstractPlatform $platform)
+    {
+        if ($value === null || $value instanceof \DateTimeImmutable) {
+            return $value;
+        }
+
+        $val = date_create_immutable($value);
```
Ocramius: Same remarks as above:

 * why this instead of a constructor (comment needed)?
 * abbreviations?
 * FQN (`\` prefix)

tests/Doctrine/Tests/DBAL/Types/DateImmutableTypeTest.php
```diff
@@ -0,0 +1,111 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Types;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\DateImmutableType;
+use Doctrine\DBAL\Types\Type;
+
+class DateImmutableTypeTest extends \PHPUnit_Framework_TestCase
```
Ocramius: Add `@covers` annotation

tests/Doctrine/Tests/DBAL/Types/DateTimeImmutableTypeTest.php
```diff
@@ -0,0 +1,114 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Types;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\DateTimeImmutableType;
+use Doctrine\DBAL\Types\Type;
+
+class DateTimeImmutableTypeTest extends \PHPUnit_Framework_TestCase
```
Ocramius: Add `@covers` annotation

tests/Doctrine/Tests/DBAL/Types/DateTimeTzImmutableTypeTest.php
```diff
@@ -0,0 +1,102 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Types;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\DateTimeTzImmutableType;
+use Doctrine\DBAL\Types\Type;
+
+class DateTimeTzImmutableTypeTest extends \PHPUnit_Framework_TestCase
```
Ocramius: Add `@covers` annotation

tests/Doctrine/Tests/DBAL/Types/TimeImmutableTypeTest.php
```diff
@@ -0,0 +1,111 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Types;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\TimeImmutableType;
+use Doctrine\DBAL\Types\Type;
+
+class TimeImmutableTypeTest extends \PHPUnit_Framework_TestCase
```
Ocramius: Add `@covers` annotation

tests/Doctrine/Tests/DBAL/Types/VarDateTimeImmutableTypeTest.php
```diff
@@ -0,0 +1,101 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Types;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\VarDateTimeImmutableType;
+use Doctrine\DBAL\Types\Type;
+
+class VarDateTimeImmutableTypeTest extends \PHPUnit_Framework_TestCase
```
Ocramius: Add `@covers` annotation

tests/Doctrine/Tests/DBAL/Types/DateImmutableTypeTest.php
```diff
@@ -0,0 +1,111 @@
...
+        );
+    }
+
+    public function testConvertsNullToDatabaseValue()
+    {
+        $this->assertNull($this->type->convertToDatabaseValue(null, $this->platform->reveal()));
+    }
+
+    /**
+     * @expectedException \Doctrine\DBAL\Types\ConversionException
```
Ocramius: Use `$this->expectException()` in the m method, also use `::class` there

tests/Doctrine/Tests/DBAL/Types/DateImmutableTypeTest.php
```diff
@@ -0,0 +1,111 @@
...
+    {
+        $this->platform->getDateFormatString()->willReturn('Y-m-d');
+
+        $date = $this->type->convertToPHPValue('2016-01-01', $this->platform->reveal());
+
+        $this->assertSame('2016-01-01 00:00:00.000000', $date->format('Y-m-d H:i:s.u'));
+    }
+
+    /**
+     * @expectedException \Doctrine\DBAL\Types\ConversionException
```
Ocramius: Use `$this->expectException()` in the m method, also use `::class` there

tests/Doctrine/Tests/DBAL/Types/DateTimeImmutableTypeTest.php
```diff
@@ -0,0 +1,114 @@
...
+        );
+    }
+
+    public function testConvertsNullToDatabaseValue()
+    {
+        $this->assertNull($this->type->convertToDatabaseValue(null, $this->platform->reveal()));
+    }
+
+    /**
+     * @expectedException \Doctrine\DBAL\Types\ConversionException
```
Ocramius: Use `$this->expectException()` in the m method, also use `::class` there

tests/Doctrine/Tests/DBAL/Types/DateTimeImmutableTypeTest.php
```diff
@@ -0,0 +1,114 @@
...
+    {
+        $this->platform->getDateTimeFormatString()->willReturn('Y-m-d H:i:s');
+
+        $date = $this->type->convertToPHPValue('2016-01-01 15:58:59.123456', $this->platform->reveal());
+
+        $this->assertSame('2016-01-01 15:58:59', $date->format('Y-m-d H:i:s'));
+    }
+
+    /**
+     * @expectedException \Doctrine\DBAL\Types\ConversionException
```
Ocramius: Use `$this->expectException()` in the m method, also use `::class` there

tests/Doctrine/Tests/DBAL/Types/DateTimeTzImmutableTypeTest.php
```diff
@@ -0,0 +1,102 @@
...
+        );
+    }
+
+    public function testConvertsNullToDatabaseValue()
+    {
+        $this->assertNull($this->type->convertToDatabaseValue(null, $this->platform->reveal()));
+    }
+
+    /**
+     * @expectedException \Doctrine\DBAL\Types\ConversionException
```
Ocramius: Use `$this->expectException()` in the m method, also use `::class` there

tests/Doctrine/Tests/DBAL/Types/DateTimeTzImmutableTypeTest.php
```diff
@@ -0,0 +1,102 @@
...
+        $this->platform->getDateTimeTzFormatString()->willReturn('Y-m-d H:i:s T')->shouldBeCalled();
+
+        $date = $this->type->convertToPHPValue('2016-01-01 15:58:59 UTC', $this->platform->reveal());
+
+        $this->assertInstanceOf(\DateTimeImmutable::class, $date);
+        $this->assertSame('2016-01-01 15:58:59 UTC', $date->format('Y-m-d H:i:s T'));
+    }
+
+    /**
+     * @expectedException \Doctrine\DBAL\Types\ConversionException
```
Ocramius: Use `$this->expectException()` in the m method, also use `::class` there

tests/Doctrine/Tests/DBAL/Types/TimeImmutableTypeTest.php
```diff
@@ -0,0 +1,111 @@
...
+        );
+    }
+
+    public function testConvertsNullToDatabaseValue()
+    {
+        $this->assertNull($this->type->convertToDatabaseValue(null, $this->platform->reveal()));
+    }
+
+    /**
+     * @expectedException \Doctrine\DBAL\Types\ConversionException
```
Ocramius: Use `$this->expectException()` in the m method, also use `::class` there

tests/Doctrine/Tests/DBAL/Types/TimeImmutableTypeTest.php
```diff
@@ -0,0 +1,111 @@
...
+    {
+        $this->platform->getTimeFormatString()->willReturn('H:i:s');
+
+        $date = $this->type->convertToPHPValue('15:58:59', $this->platform->reveal());
+
+        $this->assertSame('1970-01-01 15:58:59', $date->format('Y-m-d H:i:s'));
+    }
+
+    /**
+     * @expectedException \Doctrine\DBAL\Types\ConversionException
```
Ocramius: Use `$this->expectException()` in the m method, also use `::class` there

tests/Doctrine/Tests/DBAL/Types/VarDateTimeImmutableTypeTest.php
```diff
@@ -0,0 +1,101 @@
...
+        );
+    }
+
+    public function testConvertsNullToDatabaseValue()
+    {
+        $this->assertNull($this->type->convertToDatabaseValue(null, $this->platform->reveal()));
+    }
+
+    /**
+     * @expectedException \Doctrine\DBAL\Types\ConversionException
```
Ocramius: Use `$this->expectException()` in the m method, also use `::class` there

tests/Doctrine/Tests/DBAL/Types/VarDateTimeImmutableTypeTest.php
```diff
@@ -0,0 +1,101 @@
...
+        $this->platform->getDateTimeFormatString()->shouldNotBeCalled();
+
+        $date = $this->type->convertToPHPValue('2016-01-01 15:58:59.123456 UTC', $this->platform->reveal());
+
+        $this->assertInstanceOf(\DateTimeImmutable::class, $date);
+        $this->assertSame('2016-01-01 15:58:59.123456 UTC', $date->format('Y-m-d H:i:s.u T'));
+    }
+
+    /**
+     * @expectedException \Doctrine\DBAL\Types\ConversionException
```
Ocramius: Use `$this->expectException()` in the m method, also use `::class` there

Ocramius (changes requested):

enumag:
ping @deeky666

eigan:
Looking at the other types in DBAL while trying to fix up @Ocramius comments, and I see several of his comments also applies for them.

----

[$val](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Types/DateTimeType.php#L72) is used in almost all other classes.
`\Class` is used in other Date types
[Not using ::class](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Types/DateTimeType.php#L60), is also applied there.
using [method instead of constructor](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Types/DateTimeType.php#L75).

And with tests:
`@covers` is almost never used, I see it in two tests(?)
expectException is not used, but [setExpectedException](https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/Types/TimeTest.php#L34). _(isnt this method deprecated anyway?)_ `::class` is not used in them.

----

While I agree with your comments @Ocramius, I think its better to have it same as the other types for now, and then make another pull request that addresses your points on all the other types aswell.

Anyway, I opened a [pull request](https://github.com/deeky666/dbal/pull/5) on deeky666 that addresses some of the comments.

Looking forward to having immutable dates. Thanks for your work @deeky666

kimhemsoe:
Know it is out of the scope in this PR. But would it make sense to support DatetimeInterface when converting to sql value?

I see why it would i a bit.. !wrong! possible. So properly not a good idea, but i just wanted to throw the idea out here.

lib/Doctrine/DBAL/Types/DateTimeImmutableType.php
deeky666: @Ocramius the logic is more or less a 1:1 copy from the original DateTime types. As far as I recall it is kind of a fallback option if the first attempt to create the DateTime object fails. For more information see: https://github.com/doctrine/dbal/issues/1609
Anything I should change here?

Concerning the FQN I don't quite get why I should use it here for the function call. If that was the case, wouldn't we have to FQN ALL internal function calls aswell? This seems rather wird to me.

lib/Doctrine/DBAL/Types/VarDateTimeImmutableType.php
deeky666: See above. Also the class DocBlock should be enough. No need to duplicate comments IMO.

tests/Doctrine/Tests/DBAL/Types/DateImmutableTypeTest.php
deeky666: Since when are we doing that? Never used it before. Please give an example here if I should.

deeky666:
@Ocramius fixed CS concerns and the other stuff.

Ocramius:
All green :ship:

lib/Doctrine/DBAL/Types/DateTimeImmutableType.php
```diff
@@ -0,0 +1,93 @@
...
+    public function convertToPHPValue($value, AbstractPlatform $platform)
+    {
+        if ($value === null || $value instanceof \DateTimeImmutable) {
+            return $value;
+        }
+
+        $dateTime = \DateTimeImmutable::createFromFormat($platform->getDateTimeFormatString(), $value);
+
+        if (! $dateTime) {
+            $dateTime = \date_create_immutable($value);
```
Majkl578: Any particular reason why functional interface is preferred rather than `new DateTimeImmutable`?

jvasseur: Because it returns false on failure instead of throwing an exception.

PR #2458, opened 2016-07-26
Fixed camel case in method name.

zomberg:


phansys (changes requested):
:-1: [`Column::setNotnull()`](https://github.com/doctrine/dbal/blob/ba4561f74feb4f8060f0eb9de77daac5e1a901ce/lib/Doctrine/DBAL/Schema/Column.php#L218) doesn't match the requested change.

phansys (commented):
Sorry for my previous negative vote, I didn't noticed that you was updating the method definition too. Just for consistency, IMO, the argument (`$notnull`) and the property (`$_notnull`) cases can be updated too.

Ocramius:
Indeed, please update the argument as well :-) The property cannot be changed though, as it is protected API, and we don't know if anybody is using `get_object_vars()` in a subclass.

zomberg:
Sorry, I can't fix this because earlier I removed my fork for this repository.

Ocramius:
Closing as incomplete then.

PR #2459, opened 2016-07-27
Mysql Unknown database type geometry requested

yukato:
Hi, there :)
I'm using mysql and laravel framework, and want to use geometry type column.
But, when I try to migrate, I had an error below.

```
  [Doctrine\DBAL\DBALException]
  Unknown database type geometry requested, Doctrine\DBAL\Platforms\MySqlPlatform may not support it.
```

So, I wanted to add type of geometry type to Mysql platform.

And I tried to find people who is on a same kind of situation, and found PR below.
https://github.com/doctrine/dbal/pull/920

he was talking about postgreSQL, and mine is Mysql, but same issue.
I read a review, and it does't make sense, because...

on `lib/Doctrine/DBAL/Types/Type.php` , '$_typesMap' is just define supported doctrine mapping types

on review, it says

```
I'm sorry to disappoint you but this type isn't portable across different vendors and thus we cannot add it to the core. Please add a custom type in your application yourself for that purpose.
```

but each platform ( mysql, postgreSQL... etc) can determine which type they need to use by `initializeDoctrineTypeMappings` method.
To add type to `lib/Doctrine/DBAL/Types/Type.php` , '$_typesMap' does not affect to the all platform.

Thank you for your code review, and your support.
thank you. ;)

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -1018,6 +1018,7 @@ class MySqlPlatform extends AbstractPlatform
             'binary'        => 'binary',
             'varbinary'     => 'binary',
             'set'           => 'simple_array',
+            'geometry'      => 'geometry',
         );
     }
 
--- a/lib/Doctrine/DBAL/Types/GeometryType.php
+++ b/lib/Doctrine/DBAL/Types/GeometryType.php
@@ -1,0 +1,80 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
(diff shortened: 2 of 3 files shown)


Ocramius:
Hey @yukato, sorry, but we don't support platform-specific types unless cross-platform compliance can be guaranteed...

yukato:
@Ocramius thank you for your reply.
Why do you think it is not cross-platformed?

I think my code is cross-platformed.
Each platform can determine which type use by using `initializeDoctrineTypeMappings` method.

My code just added the type of `geometory type`, each platform can use geometory type using `initializeDoctrineTypeMappings`.
And this means `cross-platformed` isn't it?

Ocramius:
The `GeometryType` presented here can't work on non-mysql systems

yukato:
@Ocramius
You mean `GeometryType` should work on every platform, even if platform which does not support geometry define geometry type on `initializeDoctrineTypeMappings` function?

Ocramius:
Correct, or else it should live in a mysql-specific repository, such as
beberlei/doctrineextensions

yukato:
@Ocramius thank you for your reply, and thank you for tell me the library.
I read the library code, thank you for your help.

so, now I think the problem is spatial extension tables does not work well.
Like postgresql platform system, I add the code to exclude spatial extension tables with mysql-systems.
and i sent a new PR.
https://github.com/doctrine/dbal/pull/2461

PR #2460, opened 2016-07-27
Oracle fix binding null values to long/lob type columns

soyuka:
Php bug ref: https://bugs.php.net/bug.php?id=72524

According to _sixd_, to avoid this issue we should either reorder columns to put lobs at the end or always specify a length greater than 0 when using `oci_bind`.

This uses the second alternative because reordering insert columns would be a lot more work.

A test is available here: #2433 

ping @deeky666


soyuka:
http://git.php.net/?p=php-src.git;a=commit;h=b601dc5b29147bb1402d78e7f33a90981a2f94f5

Has been fixed with oci8 2.1.2. This workaround works well with oci8 2.1.1.

PR #2461, opened 2016-07-27
Exclude spatial extension tables

yukato:
Hi, there :)
I'm using mysql and laravel framework, and want to use geometry type column.
But, when I try to migrate, I had an error below.

```
  [Doctrine\DBAL\DBALException]
  Unknown database type geometry requested, Doctrine\DBAL\Platforms\MySqlPlatform may not support it.
```

I sent PR https://github.com/doctrine/dbal/pull/2459 , but it couldn't be accepted.
because `The GeometryType presented here can't work on non-mysql systems`.

so, now I think the problem is spatial extension tables does not work well.
Like postgresql platform system, I add the code to exclude spatial extension tables with mysql-systems.

Thank you for your review ;)

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -386,7 +386,7 @@ class MySqlPlatform extends AbstractPlatform
         return "SELECT COLUMN_NAME AS Field, COLUMN_TYPE AS Type, IS_NULLABLE AS `Null`, ".
                "COLUMN_KEY AS `Key`, COLUMN_DEFAULT AS `Default`, EXTRA AS Extra, COLUMN_COMMENT AS Comment, " .
                "CHARACTER_SET_NAME AS CharacterSet, COLLATION_NAME AS Collation ".
-               "FROM information_schema.COLUMNS WHERE TABLE_SCHEMA = " . $database . " AND TABLE_NAME = " . $table;
+               "FROM information_schema.COLUMNS WHERE TABLE_SCHEMA = " . $database . " AND TABLE_NAME = " . $table . " AND DATA_TYPE != 'geometry'";
     }
 
     /**


yukato:
d66789502217efd77943ad81caf2be31a5fd99a5
```
commit d66789502217efd77943ad81caf2be31a5fd99a5
Author: jsor <jan@sorgalla.com>
Date:   Tue Dec 6 16:41:19 2011 +0100

    Exclude spatial extension tables

diff --git a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
index 928da62..4e872c0 100644
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -194,7 +194,7 @@ class PostgreSqlPlatform extends AbstractPlatform
     public function getListTablesSQL()
     {
         return "SELECT tablename AS table_name, schemaname AS schema_name
-                FROM pg_tables WHERE schemaname NOT LIKE 'pg_%' AND schemaname != 'information_schema'";
+                FROM pg_tables WHERE schemaname NOT LIKE 'pg_%' AND schemaname != 'information_schema' AND tablename != 'geometry_columns' AND tablename != 'spatial_ref_sys'";
     }

     public function getListViewsSQL($database)
diff --git a/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php b/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
index da61583..701cb4b 100644
--- a/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
@@ -357,7 +357,7 @@ class SqlitePlatform extends AbstractPlatform

     public function getListTablesSQL()
     {
-        return "SELECT name FROM sqlite_master WHERE type = 'table' AND name != 'sqlite_sequence' "
+        return "SELECT name FROM sqlite_master WHERE type = 'table' AND name != 'sqlite_sequence' AND name != 'geometry_columns' AND name != 'spatial_ref_sys' "
              . "UNION ALL SELECT name FROM sqlite_temp_master "
              . "WHERE type = 'table' ORDER BY name";
```

And there is a commit for postgreSql, sqlite platforms already.

PR #2372, opened 2016-04-21
Handle arbitrary whitespaces when parsing SQL in order to apply `LIMIT` for MS SQL Server

morozov:
Given that the DBAL and QueryBuilder allow using plain SQL queries or their chunks, the DBAL should properly parse queries containing line breaks or other arbitrary formatting. Currently `SQLServerPlatform::doModifyLimitQuery()` (and `SQLServer2012Platform::doModifyLimitQuery()` which is not released yet) produce invalid SQL in case if the original query contains line breaks or other whitespaces.

PR #2463, opened 2016-08-03
IBM_DB2 platform doesn't handle offset without limit

morozov:
There's a non-empty table `users`. The following script should return all existing record IDs.

``` php
$builder = $conn->createQueryBuilder();
$builder->select('id')
    ->from('users')
    ->setFirstResult(0);
$data = $builder->execute()->fetchAll();
var_dump($data);
```

It does so on Oracle, SQL Server and MySQL, but returns empty result on DB2. The reason is that resulting SQL has a condition like `WHERE db22.DC_ROWNUM BETWEEN 1 AND 0`.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/DB2Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/DB2Platform.php
@@ -771,18 +771,26 @@ class DB2Platform extends AbstractPlatform
      */
     protected function doModifyLimitQuery($query, $limit, $offset = null)
     {
-        if ($limit === null && $offset === null) {
+        $where = array();
+
+        if ($offset > 0) {
+            $where[] = sprintf('db22.DC_ROWNUM >= %d', $offset + 1);
+        }
+
+        if ($limit !== null) {
+            $where[] = sprintf('db22.DC_ROWNUM <= %d', $offset + $limit);
+        }
+
+        if (empty($where)) {
             return $query;
         }
 
-        $limit = (int) $limit;
-        $offset = (int) (($offset)?:0);
-
         // Todo OVER() needs ORDER BY data!
-        $sql = 'SELECT db22.* FROM (SELECT db21.*, ROW_NUMBER() OVER() AS DC_ROWNUM '.
-               'FROM (' . $query . ') db21) db22 WHERE db22.DC_ROWNUM BETWEEN ' . ($offset+1) .' AND ' . ($offset+$limit);
-
-        return $sql;
+        return sprintf(
+            'SELECT db22.* FROM (SELECT db21.*, ROW_NUMBER() OVER() AS DC_ROWNUM FROM (%s) db21) db22 WHERE %s',
+            $query,
+            implode(' AND ', $where)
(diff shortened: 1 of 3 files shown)


morozov:
Resolved merge conflict with #2372.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -771,16 +771,26 @@ public function getTemporaryTableName($tableName)
...
+            $where[] = sprintf('db22.DC_ROWNUM >= %d', $offset + 1);
         }
 
-        $limit = (int) $limit;
-        $offset = (int) (($offset)?:0);
+        if ($limit !== null) {
+            $where[] = sprintf('db22.DC_ROWNUM <= %d', $offset + $limit);
+        }
+
+        if (count($where) === 0) {
```
deeky666: Can you use `empty()` here?

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -771,16 +771,26 @@ public function getTemporaryTableName($tableName)
...
+            return $query;
+        }
 
         // Todo OVER() needs ORDER BY data!
-        $sql = 'SELECT db22.* FROM (SELECT db21.*, ROW_NUMBER() OVER() AS DC_ROWNUM '.
-               'FROM (' . $query . ') db21) db22 WHERE db22.DC_ROWNUM BETWEEN ' . ($offset+1) .' AND ' . ($offset+$limit);
+        $sql = sprintf(
```
deeky666: Can you return early here like `return sprintf()`?

deeky666 (changes requested):
Patch is good, just some minor CS adjustments required.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
morozov: Sure. Just trying to understand the requirements, is `empty($array)` somehow better than `count($array)`? IMO, `empty()` is prone to typos like `empty($were)` which one may not notice w/o running tests.

deeky666: @morozov this is just about minor optimizations. `empty()` is a language construct and much more efficient than using the function `count()`. The point about typos is a good point, but in DBAL we try to code as much performance efficient as possible even if it seems neglible. Thanks for changing.

deeky666 (approved):

deeky666:
@morozov thanks for your effort!

deeky666:
Backported to `2.5` via https://github.com/doctrine/dbal/commit/0f7bf0ce3f061fa94912318f2d762ccaf78491ac

lib/Doctrine/DBAL/Platforms/DB2Platform.php
morozov: @deeky666 gotcha, thanks, will keep that in mind in the future.

PR #2464, opened 2016-08-04
Fix orderBy  : Notice: String offset cast occurred

pierre-H:
I had the error :

> Notice: String offset cast occurred

When I used orderBy in my repository


Ocramius:
Requires test case.

morozov:
Closing as stale and missing tests.

PR #2470, opened 2016-08-13
Fix problem with non-existing implicit indexes

aimeos:
On tables which contains a foreign key constraint with no explicit index, _addForeignKeyConstraint() in \Doctrine\DBAL\Schema\Table creates an implicit index. If that schema is used to create a SchemaDiff object, the diff creates a SQL statement to drop that implicit index that doesn't exist in the database, so DROP INDEX fails.

https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/Table.php#L546

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Index.php
+++ b/lib/Doctrine/DBAL/Schema/Index.php
@@ -202,7 +202,7 @@ class Index extends AbstractAsset implements Constraint
     {
         // allow the other index to be equally large only. It being larger is an option
         // but it creates a problem with scenarios of the kind PRIMARY KEY(foo,bar) UNIQUE(foo)
-        if (count($other->getColumns()) != count($this->getColumns())) {
+        if (count($other->getColumns()) < count($this->getColumns())) {
             return false;
         }
 
--- a/tests/Doctrine/Tests/DBAL/Platforms/SqlitePlatformTest.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/SqlitePlatformTest.php
@@ -407,7 +407,6 @@ class SqlitePlatformTest extends AbstractPlatformTestCase
         $diff->removedIndexes['index1'] = $table->getIndex('index1');
 
         $sql = array(
-            'DROP INDEX IDX_8D93D64923A0E66',
             'DROP INDEX IDX_8D93D6495A8A6C8D',
             'DROP INDEX IDX_8D93D6493D8E604F',
             'DROP INDEX index1',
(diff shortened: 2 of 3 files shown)


aimeos:
Any objections to this PR? It would be great to get it into the next release because the bug causes a lot failed migrations in our case.

tmbdrogba:
👍

tests/Doctrine/Tests/DBAL/Schema/TableTest.php
```diff
@@ -401,7 +401,7 @@ public function testAddForeignKeyAddsImplicitIndexIfIndexColumnsDoNotSpan()
         $table->addColumn('baz', 'string');
         $table->addColumn('bloo', 'string');
         $table->addIndex(array('baz', 'bar'), 'composite_idx');
-        $table->addIndex(array('bar', 'baz', 'bloo'), 'full_idx');
+        $table->addIndex(array('bloo', 'bar', 'baz'), 'full_idx');
```
Ocramius: Please *add* tests: removing or changing tests will lead to BC breaks.

tests/Doctrine/Tests/DBAL/Platforms/SqlitePlatformTest.php
```diff
@@ -422,7 +421,6 @@ public function testAlterTable()
             'INSERT INTO user ("key", article, comment) SELECT id, article, post FROM __temp__user',
             'DROP TABLE __temp__user',
             'ALTER TABLE user RENAME TO client',
-            'CREATE INDEX IDX_8D93D64923A0E66 ON client (article)',
```
Ocramius: Is this test supposed to be removed?

lib/Doctrine/DBAL/Schema/Index.php
```diff
@@ -202,7 +202,7 @@ public function isFullfilledBy(Index $other)
     {
         // allow the other index to be equally large only. It being larger is an option
         // but it creates a problem with scenarios of the kind PRIMARY KEY(foo,bar) UNIQUE(foo)
-        if (count($other->getColumns()) != count($this->getColumns())) {
+        if (count($other->getColumns()) < count($this->getColumns())) {
```
Ocramius: Not all databases are necessarily able to use larger indexes as a replacement for indexes with a subset of their covered columns. In addition to that, see the comment right above (about PK + UNIQUE)

Ocramius (changes requested):

Ocramius:
@aimeos also, really sorry for not seeing this PR earlier: it completely went off my radar :(

tests/Doctrine/Tests/DBAL/Platforms/SqlitePlatformTest.php
tmbdrogba: I supposed this is a non existing implicit index, right @aimeos ?

aimeos: @tmbdrogba Yes, it is. It's created to be added to the database at the end

maximecrtmc:
any update about this issue ?

loconox:
Hi,
I'm facing the issue.

PR #2473, opened 2016-08-22
Let the user provide a previous exception

greg0ire:
Previous exceptions might be very helpful when troubleshooting a
problem. Let's have a way to provide them.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^5.4.6",
+        "phpunit/phpunit-mock-objects": "!=3.2.4",
         "symfony/console": "2.*"
     },
     "suggest": {
--- a/lib/Doctrine/DBAL/Types/ConversionException.php
+++ b/lib/Doctrine/DBAL/Types/ConversionException.php
@@ -36,14 +36,19 @@ class ConversionException extends \Doctrine\DBAL\DBALException
      *
      * @param string $value
      * @param string $toType
+     * @param \Exception|null $previous
      *
      * @return \Doctrine\DBAL\Types\ConversionException
      */
-    static public function conversionFailed($value, $toType)
+    static public function conversionFailed($value, $toType, \Exception $previous = null)
     {
         $value = (strlen($value) > 32) ? substr($value, 0, 20) . '...' : $value;
 
-        return new self('Could not convert database value "' . $value . '" to Doctrine Type ' . $toType);
+        return new self(
+            'Could not convert database value "' . $value . '" to Doctrine Type ' . $toType,
+            0,
+            $previous
+        );
     }
 
     /**
@@ -75,10 +80,11 @@ class ConversionException extends \Doctrine\DBAL\DBALException
      * @param mixed    $value
      * @param string   $toType
      * @param string[] $possibleTypes
+     * @param \Exception|null $previous
      *
      * @return \Doctrine\DBAL\Types\ConversionException
      */
-    static public function conversionFailedInvalidType($value, $toType, array $possibleTypes)
+    static public function conversionFailedInvalidType($value, $toType, array $possibleTypes, \Exception $previous = null)
     {
         $actualType = is_object($value) ? get_class($value) : gettype($value);
 
@@ -89,7 +95,7 @@ class ConversionException extends \Doctrine\DBAL\DBALException
                 $actualType,
                 $toType,
                 implode(', ', $possibleTypes)
-            ));
+            ), 0, $previous);
         }
 
         return new self(sprintf(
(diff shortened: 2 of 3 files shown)


lib/Doctrine/DBAL/Types/ConversionException.php
```diff
@@ -109,6 +115,6 @@ static public function conversionFailedSerialization($value, $format, $error, \E
             $actualType,
             $format,
             $error
-        ));
+        ), 0, $previous);
```
greg0ire: lol the `$previous` argument was unused in this method @Taluu

greg0ire:
Looks like the build has to be fixed… I had similar errors on the Sonata project recently, there must have been a new release of phpunit.

greg0ire:
Just rebased on #2480 to see if the build is green

composer.json
```diff
@@ -17,6 +17,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^5.4.6",
+        "phpunit/phpunit-mock-objects": "!=3.2.4",
```
Ocramius: To be removed before merge

lib/Doctrine/DBAL/Types/ConversionException.php
```diff
@@ -36,14 +36,19 @@ class ConversionException extends \Doctrine\DBAL\DBALException
      *
      * @param string $value
      * @param string $toType
+     * @param \Exception|null $previous
      *
      * @return \Doctrine\DBAL\Types\ConversionException
      */
-    static public function conversionFailed($value, $toType)
+    static public function conversionFailed($value, $toType, \Exception $previous = null)
```
Ocramius: Doesn't seem to be used within the DBAL codebase?

lib/Doctrine/DBAL/Types/ConversionException.php
```diff
@@ -75,10 +80,11 @@ static public function conversionFailedFormat($value, $toType, $expectedFormat,
      * @param mixed    $value
      * @param string   $toType
      * @param string[] $possibleTypes
+     * @param \Exception|null $previous
      *
      * @return \Doctrine\DBAL\Types\ConversionException
      */
-    static public function conversionFailedInvalidType($value, $toType, array $possibleTypes)
+    static public function conversionFailedInvalidType($value, $toType, array $possibleTypes, \Exception $previous = null)
```
Ocramius: Doesn't seem to be used within the DBAL codebase?

lib/Doctrine/DBAL/Types/ConversionException.php
greg0ire: I found a need for it when creating a custom doctrine type for PSR7 uris : I would like to get the message from the library that parses the string into an UriInterface object if it fails.

lib/Doctrine/DBAL/Types/ConversionException.php
greg0ire: Not sure there is a need for this one, I did that one for the sake of consistency

lib/Doctrine/DBAL/Types/ConversionException.php
Ocramius: @greg0ire subclassing and implementing there then? `throw CouldNotCreatePSR7Uri::fromUriException($dbValue, $exception)`, where `CouldNotCreatePSR7Uri` extends `ConversionException`

greg0ire: This would work I guess. Feel free to close this PR if you feel this is feature creep, I would be happy enough with this kind of solution. I just saw something similar in this very class, and thought this would be ok.

Ocramius: @greg0ire yeh, that's what I'll do if this solution works for you. It would indeed be different if these exception classes were `final`

greg0ire: Ok, but then there is the unused variable (end of the diff). This is worth merging maybe?

lib/Doctrine/DBAL/Types/ConversionException.php
greg0ire: This might be worth merging @Ocramius

lib/Doctrine/DBAL/Types/ConversionException.php
Ocramius: I'd rather remove the parameter, since it's not used in doctrine either.

lib/Doctrine/DBAL/Types/ConversionException.php
Ocramius: This is unused inside the ORM either, and seems to be `master` only?

lib/Doctrine/DBAL/Types/ConversionException.php
greg0ire: I'll do that then

greg0ire:
Closing, otherwise the branch name won't be very consistent with the content of the PR

Issue #2475, opened 2016-08-22
The build is broken on master


greg0ire:
In #2473 , I noticed many errors in the build that seem unrelated to my changes.

I fixed similar problems with Sonata recently : 

- https://github.com/sonata-project/SonataMediaBundle/pull/1106/files
- https://github.com/sonata-project/SonataAdminBundle/pull/4079/files

It looks the mock system is a bit more strict now.

[Example error message](https://travis-ci.org/doctrine/dbal/jobs/154096041#L384) : 

> Doctrine\Tests\DBAL\ConnectionTest::testConnectDispatchEvent
>
> Undefined index: postConnect

The same message appears many times, with different method names.

Related : https://github.com/sebastianbergmann/phpunit-mock-objects/issues/321

greg0ire:
@Ocramius : should this be contributed on master or on 2.5 (or on both? I don't see 2.5 being merged in master or vice versa in the git graph).

Ocramius:
@greg0ire send patches to `master`, backporting is up to us.

PR #2476, opened 2016-08-22
Avoid latest version of phpunit mock object

greg0ire:
That way, people can still work on their PR until the build is fixed.

Workaround for #2475 (not a proper fix)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^5.4.6",
+        "phpunit/phpunit-mock-objects": "<3.2.4",
         "symfony/console": "2.*"
     },
     "suggest": {


Ocramius:
What is the issue leading to this? `createMock`?

greg0ire:
Not sure yet… in Sonata we use `getMock`, and have this problem too.

Ocramius:
`getMock` will not work with PHPUnit 5.4+ (deprecation errors all over the place). The solution is to replace the calls with `createMock` calls

greg0ire:
> getMock will not work with PHPUnit 5.4+ (deprecation errors all over the place). The solution is to replace the calls with createMock calls

I think it is a different issues : these are php warnings, not phpunit warnings.

greg0ire:
In Sonata it came from the fact we did not use `getMock` properly (not using FQCN, misspelling stdClass)

greg0ire:
> not using FQCN

Which seems to be the problem too in the case of this lib, see https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/ConnectionTest.php#L105 , for instance

Ocramius:
Then let's fix the test rather than blocking the dep (which will lead to more issues down the line)

greg0ire:
I don't have the time right now, that's why I'm suggesting this temporary workaround, but sure, the tests (plural, that's why I think it will be long) have to be fixed properly and the commit in this PR reverted. Meanwhile, having green builds will allow other contributions to be merged (like #2473 , which I wrote this morning).

greg0ire:
I can commit to fix it properly one evening this week though.

Ocramius:
Closing here meanwhile then. Sorry, workarounds are not something that I will merge, unless we are really cornered inside a tight solution space.

Ocramius:
@greg0ire thanks if you can submit further patches though!

greg0ire:
No problem, I understand. I should be able to fix that properly tomorrow night

Issue #949, opened 2011-03-19
DBAL-102: SQLSRV PDO doesn't support DateTime as a PHP Object


doctrinebot:
Jira issue originally created by user aarondm:

I have created an entity with has a datetime column

```

    /****
     * @orm:Column(name="deleteAt", type="datetime")
     */
    protected $deletedAt;

    /****
     * Get DeletedAt
     * @return  \DateTime   
     */
    public function getDeletedAt() {
        return $this->deletedAt;
    }

    /****
     * Set DeletedAt
     * @param   \DateTime       $deletedAt
     */
    public function setDeletedAt(\DateTime $deletedAt) {
        $this->deletedAt = $deletedAt;
    }
```

This should work, however I get this error when trying to set a "deleteAt" date

```
$this->setDeletedAt(new \DateTime());
```

```
SQLSTATE[22007]: [Microsoft][SQL Server Native Client 10.0][SQL Server]Conversion failed when converting date and/or time from character string.
```

This should work afaik.

pierre-H:
With Doctrine 2.5 I still got the error when I updat.

pierre-H:
Here is a solution which need to be implemented : 
[https://www.fusonic.net/en/blog/php-datetime-and-mssql/](https://www.fusonic.net/en/blog/php-datetime-and-mssql/)

dlegatt:
@pierre-H, here was the solution I found for this problem: http://stackoverflow.com/a/42678780/1952021

pierre-H:
@dlegatt thank you so much !

PR #2477, opened 2016-08-23
Fix Update datetime field for sqlsrv

pierre-H:
https://github.com/doctrine/dbal/issues/949

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php
@@ -66,7 +66,7 @@ class SQLServer2008Platform extends SQLServer2005Platform
      */
     public function getDateTimeFormatString()
     {
-        return 'Y-m-d H:i:s.u';
+        return 'Y-m-d\TH:i:s';
     }
 
     /**


billschaller:
This is not a fix, there are no tests, and it will break existing behavior because of time zone issues...

Ocramius:
Closing as per @billschaller's review

billschaller (commented):
AFAIK because this doesn't specify the timezone parameter for this time format, SQL Server will fill it with +00:00, breaking existing behavior. Provide a test suite ensuring behavior is consistent before and after and this would be a safe merge. Otherwise no.

pierre-H:
@billschaller I just used this solution : https://www.fusonic.net/en/blog/php-datetime-and-mssql/ because I had this same problem.
 I don't know exactly what do you need : a code which use that ?

billschaller:
pierre-H - at a bare minimum we need a failing test to demonstrate the error you're seeing. The error you're experiencing would show up if you were using datetime columns instead of datetime2. Can you post a gist of the entity and a schema dump of the table in the database you're trying to access?

billschaller:
@pierre-H - at a bare minimum we need a failing test to demonstrate the error you're seeing. The error you're experiencing would show up if you were using datetime columns instead of datetime2. Can you post a gist of the entity and a schema dump of the table in the database you're trying to access?

pierre-H:
This isn't needed anymore.
See https://github.com/doctrine/dbal/issues/949#issuecomment-285123074

PR #2479, opened 2016-08-23
Mock classes that do exist

greg0ire:
Phpunit seems to be unable to build a mock from a class that does not
exist.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/ConnectionTest.php
@@ -102,7 +102,7 @@ class ConnectionTest extends \Doctrine\Tests\DbalTestCase
 
     public function testConnectDispatchEvent()
     {
-        $listenerMock = $this->getMockBuilder('ConnectDispatchEventListener')
+        $listenerMock = $this->getMockBuilder('stdClass')
             ->setMethods(array('postConnect'))
             ->getMock();
         $listenerMock->expects($this->once())->method('postConnect');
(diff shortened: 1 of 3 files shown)


greg0ire:
This is obviously a work in progress, to show you the kind of fix that is needed. @Ocramius , does can you give me some feedback, does this fix look right to you ?

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -102,7 +102,7 @@ public function testGetEventManager()
 
     public function testConnectDispatchEvent()
     {
-        $listenerMock = $this->getMockBuilder('ConnectDispatchEventListener')
```
greg0ire: This class does not seem to exist at all.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -102,7 +102,7 @@ public function testGetEventManager()
...
+        $listenerMock = $this->getMockBuilder('stdClass')
```
greg0ire: I can replace stdClass with a dummy class if it makes things any clearer.

```php
namespace Doctrine\Dbal\Tests\Dummy;

class EventListener
{
}
```

greg0ire:
Wow that was not so long in fact, there was many messages for a single line. Anyway happy birthday @Ocramius :birthday: :gift:

greg0ire:
Sorry, it turns out this isn't a proper solution either, because the phpunit behavior is buggy and will be fixed asap. I think the best fix would be to avoid 3.2.4

PR #2482, opened 2016-08-25
Sqlite support both integer and field reference intervals in date functions

vhraban:
I stumbled upon a problem that my query worked in MySQL but not SQLite. As it turns out, when using SQLite it is not possible to perform date arithmetic on field references, only on actual integers.

This works

```
$qb = $this->repo->createQueryBuilder('foo');
$qb->join("foo.bar", "bar")
       ->where("CURRENT_TIMESTAMP() BETWEEN
                            DATE_SUB(foo.time, 1, 'hour')
                     AND
                            DATE_SUB(foo.time, 2, 'hour')
                     ");
```

generating (roughly)

> SELECT foo.time
> INNER JOIN bar ON foo.bar_id = bar.id
> WHERE CURRENT_TIMESTAMP() BETWEEN
>    DATETIME(foo.time,"-1 HOUR") 
> AND 
>    DATETIME(p0_.time,"-2 HOUR");

This does not

```
$qb = $this->repo->createQueryBuilder('foo');
$qb->join("foo.bar", "bar")
       ->where("CURRENT_TIMESTAMP() BETWEEN
                            DATE_SUB(foo.time, bar.start, 'hour')
                     AND
                            DATE_SUB(foo.time, bar.end, 'hour')
                     ");
```

Because this will get generated

> SELECT foo.time
> INNER JOIN bar ON foo.bar_id = bar.id
> WHERE CURRENT_TIMESTAMP() BETWEEN
>    DATETIME(foo.time,"-bar.start HOUR") 
> AND 
>    DATETIME(p0_.time,"-bar.end HOUR");

My fix adds support to reference other fields in the query, so that the first code example works the same way, but the second one uses SQLite concat symbol and generates the following:

> SELECT foo.time
> INNER JOIN bar ON foo.bar_id = bar.id
> WHERE CURRENT_TIMESTAMP() BETWEEN
>    DATETIME(foo.time,"-" || bar.start || " HOUR") 
> AND 
>    DATETIME(p0_.time,"-" || bar.end || " HOUR");

I've got only two concerns regarding this:
- Currently functional tests do not check for this case at all, but I added an SQLite platform test. Is this enough?
- I had to modify PHPDocs in AbstractPlatform to reflect that some methods now accept string or integer as $interval for date arithmetics. I know for sure this is the case for MySQL and, given this fix, SQLite, but other platforms may not support it. This brings inconsistency in what different platforms support

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -1034,8 +1034,8 @@ abstract class AbstractPlatform
     /**
      * Returns the SQL to add the number of given seconds to a date.
      *
-     * @param string  $date
-     * @param integer $seconds
+     * @param string         $date
+     * @param integer|string $seconds
      *
      * @return string
      *
@@ -1049,8 +1049,8 @@ abstract class AbstractPlatform
     /**
      * Returns the SQL to subtract the number of given seconds from a date.
      *
-     * @param string  $date
-     * @param integer $seconds
+     * @param string         $date
+     * @param integer|string $seconds
      *
      * @return string
      *
@@ -1064,8 +1064,8 @@ abstract class AbstractPlatform
     /**
      * Returns the SQL to add the number of given minutes to a date.
      *
-     * @param string  $date
-     * @param integer $minutes
+     * @param string         $date
+     * @param integer|string $minutes
      *
      * @return string
      *
@@ -1079,8 +1079,8 @@ abstract class AbstractPlatform
     /**
      * Returns the SQL to subtract the number of given minutes from a date.
      *
-     * @param string  $date
-     * @param integer $minutes
+     * @param string         $date
+     * @param integer|string $minutes
      *
      * @return string
      *
@@ -1094,8 +1094,8 @@ abstract class AbstractPlatform
     /**
      * Returns the SQL to add the number of given hours to a date.
      *
-     * @param string  $date
-     * @param integer $hours
+     * @param string         $date
+     * @param integer|string $hours
      *
      * @return string
      *
@@ -1109,8 +1109,8 @@ abstract class AbstractPlatform
     /**
      * Returns the SQL to subtract the number of given hours to a date.
      *
-     * @param string  $date
-     * @param integer $hours
+     * @param string         $date
+     * @param integer|string $hours
      *
      * @return string
      *
@@ -1124,8 +1124,8 @@ abstract class AbstractPlatform
     /**
      * Returns the SQL to add the number of given days to a date.
      *
-     * @param string  $date
-     * @param integer $days
+     * @param string         date
+     * @param integer|string $days
      *
      * @return string
      *
@@ -1139,8 +1139,8 @@ abstract class AbstractPlatform
     /**
      * Returns the SQL to subtract the number of given days to a date.
      *
-     * @param string  $date
-     * @param integer $days
+     * @param string         $date
+     * @param integer|string $days
      *
      * @return string
      *
@@ -1154,8 +1154,8 @@ abstract class AbstractPlatform
     /**
      * Returns the SQL to add the number of given weeks to a date.
      *
-     * @param string  $date
-     * @param integer $weeks
+     * @param string         $date
+     * @param integer|string $weeks
      *
      * @return string
      *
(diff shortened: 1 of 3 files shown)


lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1034,8 +1034,8 @@ public function getDateDiffExpression($date1, $date2)
     /**
      * Returns the SQL to add the number of given seconds to a date.
      *
-     * @param string  $date
-     * @param integer $seconds
+     * @param string         $date
+     * @param integer|string $seconds
```
Ocramius: All of these should be reverted: we shouldn't accept strings, and should be very strict about this due to SQL injection concerns

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -129,11 +130,29 @@ public function getLocateExpression($str, $substr, $startPos = false)
...
+        switch (gettype($interval))
+        {
+            case 'integer':
+                $formatString = '%s(%s,"%s%s %s")';
+                break;
+            case 'string':
+                $formatString = '%s(%s,"%s" || %s || " %s")';
+                break;
+            default:
+                throw new DBALException("Interval is of an unsupported type");
```
Ocramius: The `$interval` should be part of the exception, or else the user may go mad.

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -129,11 +130,29 @@ public function getLocateExpression($str, $substr, $startPos = false)
      */
     protected function getDateArithmeticIntervalExpression($date, $operator, $interval, $unit)
     {
+        // If $interval is a string (i.e. reference to other entity value), then it
+        // has to be concatenated using || operator
+
+        // Calculate the format string based in whether || concatenation has to be done
+        // or not
+        switch (gettype($interval))
```
Ocramius: Move this entire block to a private method, including explanation

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -129,11 +130,29 @@ public function getLocateExpression($str, $substr, $startPos = false)
...
+        {
+            case 'integer':
+                $formatString = '%s(%s,"%s%s %s")';
+                break;
+            case 'string':
```
Ocramius: This should not happen. To prevent that, cast the given values to `(int)`

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
Ocramius: I see what you mean by this, but `string` is not valid here. Must be something like an expression object, or a field reference. /cc @deeky666

vhraban: As far as I know, there is no Expr to represent datetime arithmetics. Field reference can not be represented as a data type either. Leaving this just an integer does not reflect how this method can be used. I can not see any good way of doing this PHPDoc at all apart of introducing datetime Expr that goes far beyond this issue. What would you suggest?

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
vhraban: It could be a string if $interval was a field reference

tests/Doctrine/Tests/DBAL/Platforms/SqlitePlatformTest.php
```diff
@@ -744,4 +744,19 @@ public function testQuotesTableNameInListTableForeignKeysSQL()
     {
         $this->assertContains("'Foo''Bar\\'", $this->_platform->getListTableForeignKeysSQL("Foo'Bar\\"), '', true);
     }
+
+    public function testSupportsIntAndStringInDateArithmeticExpressions()
```
vhraban: Are these tests enough of proper functional tests should be written?

vhraban: Just a note: probably it's a good idea to rename this to testSupportsIntAnd**FieldReference**InDateArithmeticExpressions

vhraban:
Please do not close this ticket due to inactivity, I will have a proper look into it the next week

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
Ocramius: No, what I'm saying is creating an `Expr` object or such. Passing around strings here is just going to be causing more and more problems in the future :(

Ocramius: @deeky666 do you have suggestions about how to design this API, eventually?

Issue #1110, opened 2015-03-11
DBAL-1168: Schema's getMigrateFromSql always adds CREATE SCHEMA


doctrinebot:
Jira issue originally created by user vbence:

I originally posted this to Migrations; noticing that all the generated down() methods start with a "CREATE SCHEMA public" line.

Inspecting the return from Schema#getMigrateFromSql it indeed contains the create statement.

cleentfaar:
Any updates on this? Would be nice to get rid of this in our version files

josensanchez:
As @doctrinebot commented, the problem seems to be here:

```
        $fromSchema = $conn->getSchemaManager()->createSchema();
        $toSchema = $this->getSchemaProvider()->createSchema();
```
the first line builds the schema querying the database 

```
SELECT schema_name AS nspname
          FROM   information_schema.schemata
...
```

 and the second one builds it from the ORM metadata in your application

`@ORM\Table(name="table_1" ...`

there is a "public" schema in `$fromSchema` but since there is no `@ORM\Table(name="table_1", schema="public"...)`  also there is no "public" schema in `$toSchema`

trying the @doctrinebot workaround `@ORM\Table(name="table_1", schema="public"...)` it's also useless as he already commented because table_1 and public.table_1 are two different things to the dbal, witch is logic.
 but it will try to perform the next code every time.
```
create table public.table_1 ...  -- table already exists exception
drop table table_1
```
the drop command is without the "public." and I assume that it's to add support to tables without namespaces...

I don't see an easy solution for this issue. Either finding a way to define table_1 as an alias for public.table_1 (maybe in postgresql if a database object doesn't have a namespace defined, then the 'public' namespace is attached to it) or removing support in postgresql for no namespace objects, forcing always to define schema="public" 

for the moment if anybody is working with postgresql and doctrine I suggest to use only the public schema or not use it at all.

idchlife:
Yep, waiting issue to be resolved

garex:
Can't use `diff` commands as schema always differs by those `CREATE SCHEMA public`.

Melkij:
My workaround without modify library and waiting patch.
Add class in any namespace in application:

```PHP
<?php

namespace EngineBundle\Doctrine;

use Doctrine\Common\EventSubscriber;
use Doctrine\ORM\Tools\Event\GenerateSchemaEventArgs;

class MigrationEventSubscriber implements EventSubscriber
{
    public function getSubscribedEvents()
    {
        return array(
[...]

PR #2490, opened 2016-08-26
Adds default namespace to schema.

garex:
Adds default namespace to schema

Some databases supports namespaces and has default one. For example Postgres has default "public".

When schema created from classes, which has table names without namespaces, but this namespace is assumed, we should add it explicitly, if it not exists yet.

It will help in diffs between real DB and config shemas.

The visitor is skipping platforms that don't support schemas.

This is 1st part of fix DBAL-1168 #1110

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
+++ b/lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
@@ -25,21 +25,31 @@ use Doctrine\DBAL\Platforms\AbstractPlatform;
 /**
  * Adds default namespace to schema
  *
- * Some databases supports namespaces and has default one. For example Postgres has default "public".
+ * Some databases supports namespaces and has default one.
+ * For example Postgres has default "public".
  *
- * When schema created from classes, which has table names without namespaces, but this namespace is assumed,
- * we should add it explicitly, if it not exists yet.
+ * When schema created from classes, which has table names without namespaces,
+ * but this namespace is assumed, we should add it explicitly, if it not exists
+ * yet.
  *
  * It will help in diffs between real DB and config shemas.
  *
+ * The visitor is skipping platforms that don't support schemas.
+ *
  * @author Alexander Ustimenko <a@ustimen.co>
- * @since  2.5
+ * @since  2.6
  */
 class AddDefaultNamespace extends AbstractVisitor
 {
 
+    /**
+     * @var \Doctrine\DBAL\Platforms\AbstractPlatform
+     */
     private $platform;
 
+    /**
+     * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform
+     */
     public function __construct(AbstractPlatform $platform)
     {
         $this->platform = $platform;
--- a/lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
+++ b/lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
@@ -1,0 +1,58 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Schema\Visitor;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+
+/**
+ * Adds default namespace to schema
+ *
+ * Some databases supports namespaces and has default one. For example Postgres has default "public".
+ *
+ * When schema created from classes, which has table names without namespaces, but this namespace is assumed,
+ * we should add it explicitly, if it not exists yet.
+ *
+ * It will help in diffs between real DB and config shemas.
+ *
+ * @author Alexander Ustimenko <a@ustimen.co>
+ * @since  2.5
+ */
+class AddDefaultNamespace extends AbstractVisitor
+{
+
+    private $platform;
+
+    public function __construct(AbstractPlatform $platform)
+    {
+        $this->platform = $platform;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function acceptSchema(Schema $schema)
+    {
+        $namespaceName = $this->platform->getDefaultSchemaName();
+        if (!$schema->hasNamespace($namespaceName)) {
+            $schema->createNamespace($namespaceName);
+        }
+    }
+}
--- a/lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
+++ b/lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
@@ -41,7 +41,6 @@ use Doctrine\DBAL\Platforms\AbstractPlatform;
  */
 class AddDefaultNamespace extends AbstractVisitor
 {
-
     /**
      * @var \Doctrine\DBAL\Platforms\AbstractPlatform
      */
--- a/lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
+++ b/lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
@@ -60,6 +60,9 @@ class AddDefaultNamespace extends AbstractVisitor
      */
     public function acceptSchema(Schema $schema)
     {
+        if (!$this->platform->supportsSchemas()) {
+            return;
+        }
         $namespaceName = $this->platform->getDefaultSchemaName();
         if (!$schema->hasNamespace($namespaceName)) {
             $schema->createNamespace($namespaceName);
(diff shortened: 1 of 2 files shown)


garex:
Next step will be to add this visitor to the end of `Doctrine\ORM\Tools\SchemaTool::getSchemaFromMetadata`

    if ($this->platform->supportsSchemas()) {
        $schema->visit(new AddDefaultNamespace($this->platform));
    }

lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
```diff
@@ -0,0 +1,58 @@
...
+ * we should add it explicitly, if it not exists yet.
+ *
+ * It will help in diffs between real DB and config shemas.
+ *
+ * @author Alexander Ustimenko <a@ustimen.co>
+ * @since  2.5
+ */
+class AddDefaultNamespace extends AbstractVisitor
+{
+
```
Ocramius: Docblock

tests/Doctrine/Tests/DBAL/Schema/Visitor/AddDefaultNamespaceTest.php
```diff
@@ -0,0 +1,82 @@
...
+
+        $this->assertNotEmpty($schema->getNamespaces());
+        $this->assertCount(1, $schema->getNamespaces());
+    }
+
+    public function namespacedPlatforms()
+    {
+        return [
+            [new PostgreSqlPlatform()],
+            [new SQLServerPlatform()],
```
Ocramius: What about platforms with no schema support?

Ocramius:
> Next step will be to add this visitor to the end of `Doctrine\ORM\Tools\SchemaTool::getSchemaFromMetadata`

Could be added by default, but then act as no-op if the platform doesn't support schemas?

lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
garex: What wrong with it?

tests/Doctrine/Tests/DBAL/Schema/Visitor/AddDefaultNamespaceTest.php
garex: It will not work there as it should be added only if platform support default namespace. You think it should be added in ORM without such check and do this check inside itself?

garex:
> Could be added by default, but then act as no-op if the platform doesn't support schemas?

I also thought about this option first, but then deside to do the check outside to give more readability at the place where it should be added at `getSchemaFromMetadata`.

But if your experience tells you to move it inside -- let's move it inside. It will become more reusable then.

Ocramius:
If it's not a no-op when the platform does not support schemas, then it will become a setup issue for most users, making working apps not portable. IMO, it should be done inside the visitor.

garex:
Ok, I will rework it. Pls not merge many commits, I will rebase em after approve. 

Regarding docblock. What bad there? Bad `@since` or description?

lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
Ocramius: It's missing :-)

lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Some databases supports namespaces and has default one. For example Postgres has default "public".
+ *
+ * When schema created from classes, which has table names without namespaces, but this namespace is assumed,
+ * we should add it explicitly, if it not exists yet.
+ *
+ * It will help in diffs between real DB and config shemas.
+ *
+ * @author Alexander Ustimenko <a@ustimen.co>
+ * @since  2.5
```
Ocramius: 2.6

tests/Doctrine/Tests/DBAL/Schema/Visitor/AddDefaultNamespaceTest.php
Ocramius: As discussed later, the check should probably be done by the visitor, skipping platforms that don't support schemas.

lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
garex: > It's missing :-)

What is missing?? There is top comment and class comment. What is "docblock" then? Can you provide an example?

teohhanhui:
``` php
/**
 * @var AbstractPlatform
 */
private $platform;
```

garex:
Pls give me know when you think it will be ready for merge -- I will rebase it. Or you can rebase it on master and squash all in one commit.

garex:
@deeky666 when approximately will you plan to review?

garex:
@Ocramius please assing someone else or review by yourself.

Two weeks passed already and @deeky666 is not reponds.

lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
```diff
@@ -0,0 +1,71 @@
...
+ * The visitor is skipping platforms that don't support schemas.
+ *
+ * @author Alexander Ustimenko <a@ustimen.co>
+ * @since  2.6
+ */
+class AddDefaultNamespace extends AbstractVisitor
+{
+
+    /**
+     * @var \Doctrine\DBAL\Platforms\AbstractPlatform
```
phansys, line 45: This class is already imported, so you should use `@var AbstractPlatform` instead.

lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
```diff
@@ -0,0 +1,71 @@
...
+ * It will help in diffs between real DB and config shemas.
+ *
+ * The visitor is skipping platforms that don't support schemas.
+ *
+ * @author Alexander Ustimenko <a@ustimen.co>
+ * @since  2.6
+ */
+class AddDefaultNamespace extends AbstractVisitor
+{
+
```
phansys: Please, remove surplus LF.

phansys (changes requested):

lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
garex: Not sure. In other classes in same directory also fully qualified names used. So I decided to leave it as it was here before.

garex:
This PR is covered with cobwebs. Betting, it will be merged may be after New Year.

Cobwebs

Ocramius:
There's waaaaaay older stuff around here. If @deeky666 didn't have time to check it, he's evidently busy at the moment.

garex:
@Ocramius As I see, @deeky666 has not any activity on his github profile after 2016, July.

Do you know him? Is he alive/healthy?

Can you assign someone else from 31 people that are in Doctrine's members team?

Ocramius:
No, I cannot. He's pretty much the gatekeeper, and I talk with him from
time to time, he just is that busy IRL ATM.

On 8 Nov 2016 8:25 a.m., "Alexander Ustimenko" <notifications@github.com>
wrote:

@Ocramius <https://github.com/Ocramius> As I see, @deeky666
<https://github.com/deeky666> has not any activity on his github profile
after 2016, July.

Do you know him? Is he alive/healthy?

Can you assign someone else from 31 people that are in Doctrine's members
team?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/doctrine/dbal/pull/2490#issuecomment-259064881>, or mute
the thread

lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        $this->platform = $platform;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function acceptSchema(Schema $schema)
+    {
+        if (!$this->platform->supportsSchemas()) {
```
phansys: Please, respect Doctrine CS.

phansys (changes requested):
Status: Needs work.

lib/Doctrine/DBAL/Schema/Visitor/AddDefaultNamespace.php
teohhanhui: Ii'm guessing it's this one? https://github.com/doctrine/coding-standard/pull/7

> Logical NOT operators (!) MUST have leading and trailing spaces

garex: Holy shit, pettifoggers )) I will README https://github.com/doctrine/coding-standard and will come back to you soon.

deeky666:
@garex sorry for the delay. Haven't been around here last year very much. The concern I have about the "default schema" topic how it is implemented in DBAL (via platforms) is that there really is no "fixed" default schema for probably all the vendors supporting schemas. In PostgreSQL for example you can configure the default schema via its search path, Oracle uses the user that connects as the default schema and I suppose others like SQL Server let the user define the default schema, too. All that happens at runtime so having the platforms return a default schema name is quite of a misconception IMO and should even be deprecated. Also the term "schema" and "namespace" as used in Doctrine is rather ambiguous and not clearly defined. IMO "schema" / "namespace" is a concept that groups database objects like tables, indexes etc. So in MySQL for example this concept would translate to "database".

So what we have to do (rather than relying on the platforms default schema name) is evaluate the runtime default schema name and add it to the table name when creating the schema form entity classes.

The problem IMO is [this line](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/Schema.php#L96). While assuming that `public` is the default namespace for all vendors (even those not supporting schemas) is completely wrong, the runtime default name [should already be evaluated in the schema manager](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php#L1055-L1058).

So maybe this needs reevaluation. I don't think this solution is the best we can achieve.

garex:
Hi @deeky666 , nice to see you here.

I think it will be overcomplicated fix that will be merged may be in next century. Current solution is not best, but it will fix issue in 99% of cases. 

How often you see users, who change default schema names or default root name? I think it's a rare case.

Anyway if you want more ideal solution for 1% of users, it could be later improved by evaluating from live connection during schema diff. But as I assume it will be low-priority task that will be never done.

deeky666:
@garex I get your point and you are possibly completely right about that it fixes the issue for "99%" of the users but I think we already have the possibility to fix this more properly without a lot of effort. The schema manager already has an API for evaluating the runtime schema. The reason I don't like the solution is that it pushes a misconception even further into the wrong direction. Also adding a special visitor for this makes us having to maintain that API which in fact IMO is just a workaround that doesn't even fix all edge cases. Speaking of edge cases, there is murphy's law and from my experience the "1%" edge cases will be requested to be fixed soon after that. So introducing this solution will only put us more into the corner.
I could probably live with a solution like this if in fact we do not have another better way of fixing this issue. I am only asking for evaluating if we can use the schema managers `getSchemaSearchPaths()` for this (because this is exactly the reasons why we implemented it IIRC). And looking at the code it should already do its job. IMO the logic should be:

1. evaluate the default schema via `getSchemaSearchPaths()`
2. if that does not return anything, use the default schema defined in the platform

That would save us an additional visitor, if I am not completely wrong.

garex:
@deeky666 I will research it and ping back here.

antonmedv:
@deeky666 any plans on solving this?

garex:
@antonmedv you can try to fix it yourself. @deeky666 is only as a reviewer here, not implementer/fixer.

greg0ire:
`getSchemaSearchPaths()` is no longer an option because of https://github.com/doctrine/dbal/pull/4821 `PostgreSQLSchemaManager::getCurrentSchema()` should be used instead.

PR #2494, opened 2016-09-01
Use the same statement resource for repeated execution of the same statement on SQL Server

### LINKED ISSUE #2493: The SQL Server driver doesn't reuse existing prepared statements
Internally `SQLSrvStatement::execute()` [calls `sqlsrv_query()`](https://github.com/doctrine/dbal/blob/2.5/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php#L200) which means that even if the same instance of `SQLSrvStatement` is executed multiple times, it creates a new one-off statement.

Performance-wise, the combination of `sqlsrv_prepare()` + `sqlsrv_execute()` [is preferred](https://msdn.microsoft.com/en-us/library/cc296181(v=sql.105).aspx) to `sqlsrv_query()`:
> The combination of `sqlsrv_prepare` and `sqlsrv_execute` separates statement preparation and statement execution in to two function calls and can be used to execute parameterized queries. This function is ideal to execute a statement multiple times with different parameter values for each execution.

However, in case if the statement is not going to be reused, `sqlsrv_query()` [is preferrable](https://msdn.microsoft.com/en-us/library/cc296184(v=sql.105).aspx):
> The `sqlsrv_query` function is well-suited for one-time queries and should be the default choice to execute queries unless special circumstances apply.

It would be nice to have a way to specify whether the statement is going to be reused so that the most optimal API could be used.

morozov:
Fixes #2493.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
+++ b/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
@@ -197,8 +197,17 @@ class SQLSrvStatement implements IteratorAggregate, Statement
             }
         }
 
-        $this->stmt = sqlsrv_query($this->conn, $this->sql, $this->params);
         if ( ! $this->stmt) {
+            $stmt = sqlsrv_prepare($this->conn, $this->sql, $this->params);
+
+            if (!$stmt) {
+                throw SQLSrvException::fromSqlSrvErrors();
+            }
+
+            $this->stmt = $stmt;
+        }
+
+        if (!sqlsrv_execute($this->stmt)) {
             throw SQLSrvException::fromSqlSrvErrors();
         }
 
--- a/tests/Doctrine/Tests/DBAL/Functional/Driver/SQLSrv/StatementTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Driver/SQLSrv/StatementTest.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Functional\Driver\SQLSrv;
+
+use Doctrine\DBAL\Driver\SQLSrv\Driver;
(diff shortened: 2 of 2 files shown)


lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
```diff
@@ -197,8 +197,15 @@ public function execute($params = null)
             }
         }
 
-        $this->stmt = sqlsrv_query($this->conn, $this->sql, $this->params);
         if ( ! $this->stmt) {
+            $stmt = sqlsrv_prepare($this->conn, $this->sql, $this->params);
+            if (!$stmt) {
```
Ocramius: This branch should be tested.

Also, add empty lines around control structures

tests/Doctrine/Tests/DBAL/Functional/Driver/SQLSrv/StatementTest.php
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public function testFailureToPrepareResultsInException()
+    {
+        // use the driver connection directly to avoid having exception wrapped
+        $stmt = $this->_conn->getWrappedConnection()->prepare(null);
+
+        // it's impossible to prepare the statement without bound variables for SQL Server,
+        // so the preparation happens before the first execution when variables are already in place
+        $this->setExpectedException('Doctrine\\DBAL\\Driver\\SQLSrv\\SQLSrvException');
```
Ocramius: You can use `SQLSrvException::class` here, since I think that we bumped version requirements to minimum PHP 5.5

Ocramius:
Other than that `::class`, seems good!

Ocramius:
@morozov merged, thanks!

PR #2495, opened 2016-09-02
Optimize and improve parameter conversion in OCI8Statement

morozov:
Current implementation of `OCI8Statement::convertPositionalToNamedPlaceholders()` is relatively slow. In a worst case scenario I observed during profiling the app I'm working on, parsing takes up to 10% of the execution time.

It could be sped up by getting rid of a character-by-character string traversal in PHP and using regular expressions.

Besides the performance problem, the current implementation incorrectly handles escaped quotes and quotes of another kind in string literals. Given that in many cases string literals contain user input, it may lead to all sort of issues including security ones:

``` php
$query = <<<SQL
SELECT * FROM table WHERE name = 'How is O\'Reilly?'
SQL;
list($converted) = OCI8Statement::convertPositionalToNamedPlaceholders($query);
echo $converted, PHP_EOL;
// SELECT * FROM table WHERE name = 'How is O\'Reilly:param1'

$query = <<<SQL
SELECT * FROM table WHERE name = 'What about a 65" 4K TV?'
SQL;
list($converted) = OCI8Statement::convertPositionalToNamedPlaceholders($query);
echo $converted, PHP_EOL;
// SELECT * FROM table WHERE name = 'What about a 65" 4K TV:param1'
```

The suggested implementation is times faster and handles the described above literals correctly.

Additionally, it introduces throwing exception in case of parsing errors. It will be mostly useful in cases if the converter is unable to parse a correct query and thus produces a knowingly invalid result.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
+++ b/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
@@ -151,10 +151,7 @@ class OCI8Statement implements \IteratorAggregate, Statement
             ));
         }
 
-        if ($fragmentOffset < strlen($statement)) {
-            $fragments[] = substr($statement, $fragmentOffset);
-        }
-
+        $fragments[] = substr($statement, $fragmentOffset);
         $statement = implode('', $fragments);
 
         return array($statement, $paramMap);
--- a/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
+++ b/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
@@ -218,7 +218,7 @@ class OCI8Statement implements \IteratorAggregate, Statement
         $token = self::findToken(
             $statement,
             $tokenOffset,
-            '/(?<!\\\\)' . preg_quote($currentLiteralDelimiter, '/') . '/'
+            '/' . preg_quote($currentLiteralDelimiter, '/') . '/'
         );
 
         if (!$token) {
--- a/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
+++ b/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
@@ -163,9 +163,10 @@ class OCI8Statement implements \IteratorAggregate, Statement
      * @param string $statement The SQL statement to parse
      * @param string $tokenOffset The offset to start searching from
      * @param int $fragmentOffset The offset to build the next fragment from
-     * @param array $fragments Resulting query fragments
-     * @param string|null $currentLiteralDelimiter Current literal delimiter
-     * @param array $paramMap Parameter map
+     * @param string[] $fragments Fragments of the original statement not containing placeholders
+     * @param string|null $currentLiteralDelimiter The delimiter of the current string literal
+     *                                             or NULL if not currently in a literal
+     * @param array<int, string> $paramMap Mapping of the original parameter positions to their named replacements
      * @return bool Whether the token was found
      */
     private static function findPlaceholderOrOpeningQuote(
@@ -182,17 +183,20 @@ class OCI8Statement implements \IteratorAggregate, Statement
             return false;
         }
 
-        if ($token == '?') {
+        if ($token === '?') {
             $position = count($paramMap) + 1;
             $param = ':param' . $position;
             $fragments[] = substr($statement, $fragmentOffset, $tokenOffset - $fragmentOffset);
             $fragments[] = $param;
             $paramMap[$position] = $param;
-            $fragmentOffset = ++$tokenOffset;
-        } else {
-            $currentLiteralDelimiter = $token;
-            ++$tokenOffset;
+            $tokenOffset += 1;
+            $fragmentOffset = $tokenOffset;
+
+            return true;
         }
+
+        $currentLiteralDelimiter = $token;
+        ++$tokenOffset;
 
         return true;
     }
@@ -202,7 +206,8 @@ class OCI8Statement implements \IteratorAggregate, Statement
      *
      * @param string $statement The SQL statement to parse
      * @param string $tokenOffset The offset to start searching from
-     * @param string|null $currentLiteralDelimiter Current literal delimiter
+     * @param string|null $currentLiteralDelimiter The delimiter of the current string literal
+     *                                             or NULL if not currently in a literal
      * @return bool Whether the token was found
      */
     private static function findClosingQuote(
@@ -210,7 +215,11 @@ class OCI8Statement implements \IteratorAggregate, Statement
         &$tokenOffset,
         &$currentLiteralDelimiter
     ) {
-        $token = self::findToken($statement, $tokenOffset, '/(?<!\\\\)' . $currentLiteralDelimiter . '/');
+        $token = self::findToken(
+            $statement,
+            $tokenOffset,
+            '/(?<!\\\\)' . preg_quote($currentLiteralDelimiter, '/') . '/'
+        );
 
         if (!$token) {
             return false;
--- a/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
+++ b/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
@@ -125,65 +125,123 @@ class OCI8Statement implements \IteratorAggregate, Statement
      */
     static public function convertPositionalToNamedPlaceholders($statement)
     {
-        $paramMap = array();
-        $count = 1;
-        $start = $offset = 0;
-        $fragments = array();
-        $quote = false;
-        $capture = function ($regex, $callback) use ($statement, $start, &$offset) {
-            if (preg_match($regex, $statement, $matches, PREG_OFFSET_CAPTURE, $offset)) {
-                $offset = $matches[0][1];
-                $callback($matches[0][0]);
-                return true;
-            }
-            return false;
-        };
+        $fragmentOffset = $tokenOffset = 0;
+        $fragments = $paramMap = array();
+        $currentLiteralDelimiter = null;
 
         do {
-            if (!$quote) {
-                $result = $capture('/[?\'"]/', function ($token) use (
+            if (!$currentLiteralDelimiter) {
+                $result = self::findPlaceholderOrOpeningQuote(
                     $statement,
-                    &$fragments,
-                    &$start,
-                    &$offset,
-                    &$quote,
-                    &$count,
-                    &$paramMap
-                ) {
-                    if ($token == '?') {
-                        $param = ':param' . $count;
-                        $fragments[] = substr($statement, $start, $offset - $start);
-                        $fragments[] = $param;
-                        $paramMap[$count] = $param;
-                        $start = ++$offset;
-                        ++$count;
-                    } else {
-                        $quote = $token;
-                        ++$offset;
-                    }
-                });
+                    $tokenOffset,
+                    $fragmentOffset,
+                    $fragments,
+                    $currentLiteralDelimiter,
+                    $paramMap
+                );
             } else {
-                $result = $capture('/(?<!\\\\)' . $quote . '/', function () use (&$offset, &$quote) {
-                    $quote = false;
-                    ++$offset;
-                });
+                $result = self::findClosingQuote($statement, $tokenOffset, $currentLiteralDelimiter);
             }
         } while ($result);
 
-        if ($quote) {
+        if ($currentLiteralDelimiter) {
             throw new OCI8Exception(sprintf(
                 'The statement contains non-terminated string literal starting at offset %d',
-                $offset - 1
+                $tokenOffset - 1
             ));
         }
 
-        if ($start < strlen($statement)) {
-            $fragments[] = substr($statement, $start);
+        if ($fragmentOffset < strlen($statement)) {
+            $fragments[] = substr($statement, $fragmentOffset);
         }
 
         $statement = implode('', $fragments);
 
         return array($statement, $paramMap);
+    }
+
+    /**
+     * Finds next placeholder or opening quote.
+     *
+     * @param string $statement The SQL statement to parse
+     * @param string $tokenOffset The offset to start searching from
+     * @param int $fragmentOffset The offset to build the next fragment from
+     * @param array $fragments Resulting query fragments
+     * @param string|null $currentLiteralDelimiter Current literal delimiter
+     * @param array $paramMap Parameter map
+     * @return bool Whether the token was found
+     */
+    private static function findPlaceholderOrOpeningQuote(
+        $statement,
+        &$tokenOffset,
+        &$fragmentOffset,
+        &$fragments,
+        &$currentLiteralDelimiter,
+        &$paramMap
+    ) {
+        $token = self::findToken($statement, $tokenOffset, '/[?\'"]/');
+
+        if (!$token) {
+            return false;
+        }
+
+        if ($token == '?') {
+            $position = count($paramMap) + 1;
+            $param = ':param' . $position;
+            $fragments[] = substr($statement, $fragmentOffset, $tokenOffset - $fragmentOffset);
+            $fragments[] = $param;
+            $paramMap[$position] = $param;
+            $fragmentOffset = ++$tokenOffset;
+        } else {
+            $currentLiteralDelimiter = $token;
+            ++$tokenOffset;
+        }
+
+        return true;
+    }
+
+    /**
+     * Finds closing quote
+     *
+     * @param string $statement The SQL statement to parse
+     * @param string $tokenOffset The offset to start searching from
+     * @param string|null $currentLiteralDelimiter Current literal delimiter
+     * @return bool Whether the token was found
+     */
+    private static function findClosingQuote(
+        $statement,
+        &$tokenOffset,
+        &$currentLiteralDelimiter
+    ) {
+        $token = self::findToken($statement, $tokenOffset, '/(?<!\\\\)' . $currentLiteralDelimiter . '/');
+
+        if (!$token) {
+            return false;
+        }
+
+        $currentLiteralDelimiter = false;
+        ++$tokenOffset;
+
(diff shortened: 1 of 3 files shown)


lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -114,27 +114,68 @@ public function __construct($dbh, $statement, OCI8Connection $conn)
...
+                $offset = $matches[0][1];
+                $callback($matches[0][0]);
+                return true;
+            }
+            return false;
+        };
+
+        do {
+            if (!$quote) {
+                $result = $capture('/[?\'"]/', function ($token) use (
```
Ocramius: Please move the closure to a private method, and give it a specific name

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -114,27 +114,68 @@ public function __construct($dbh, $statement, OCI8Connection $conn)
...
+            if (preg_match($regex, $statement, $matches, PREG_OFFSET_CAPTURE, $offset)) {
+                $offset = $matches[0][1];
+                $callback($matches[0][0]);
+                return true;
+            }
+            return false;
+        };
+
+        do {
+            if (!$quote) {
```
Ocramius: `$quote` is a weird variable here. You are basically tracking if you are in a quoted value or not, so maybe `$inQuotedValue`?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -114,27 +114,68 @@ public function __construct($dbh, $statement, OCI8Connection $conn)
...
-                $i += $len-1; // jump ahead
-                $stmtLen = strlen($statement); // adjust statement length
-                ++$count;
-            } elseif ($statement[$i] == "'" || $statement[$i] == '"') {
-                $inLiteral = ! $inLiteral; // switch state!
+        $count = 1;
+        $start = $offset = 0;
+        $fragments = array();
+        $quote = false;
+        $capture = function ($regex, $callback) use ($statement, $start, &$offset) {
```
Ocramius: This closure is to be moved to a private method.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -114,27 +114,68 @@ public function __construct($dbh, $statement, OCI8Connection $conn)
...
+            if (preg_match($regex, $statement, $matches, PREG_OFFSET_CAPTURE, $offset)) {
+                $offset = $matches[0][1];
+                $callback($matches[0][0]);
+                return true;
+            }
+            return false;
+        };
+
+        do {
```
Ocramius: The entire matching logic here is really complex, and has a huge cyclomatic complexity. It is worth checking coverage, as well as mutation-testing this class.

I feel like a lot of the concepts being introduced have too generic naming, and that makes for hard to read/understand code.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
morozov: It also tracks which quote you're in, so that a quote of another kind is not considered closing one.

Ocramius: I can't think of scenarios where multiple quote tokens are present, but still, naming according to what you mean would be best.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
morozov: IIRC, from the line coverage standpoint all the code should be covered (need to double check, though).

Is there a way to see the path coverage using PHPUnit/Xdebug? I think we could reduce/manage the complexity given some numbers at hand.

Is there any choice of preference regarding mutation-testing in Doctrine?

Ocramius: > Is there a way to see the path coverage using PHPUnit/Xdebug? I think we could reduce/manage the complexity given some numbers at hand.

Fairly sure that patch coverage was added recently, but isn't supported by phpcov.

As for mutation-testing, just testing this single scenario with humbug could work ( https://github.com/padraic/humbug ).

I am just being very cautious about parsing SQL here, heh...

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
morozov: It can be a query which contains mixed parametrized and inline values coming from user input. I agree it's unlikely but not possible, especially in legacy applications.

Ocramius: Yes, that's why I'm being cautious - anything that fiddles with SQL modification is a security nightmare :)

morozov:
@Ocramius please check the reworked version. It contains:

1. More meaningful variable names,
2. Closures reworked as private named methods,
3. Some better test coverage. At the moment Humbug doesn't report any escaped mutants in the code being modified. It does report a couple of fatals and timeouts, but it's should be fine given the nature of the code.

> I am just being very cautious about parsing SQL here, heh...

I totally agree. Luckily, unlike `SQLServerPlatform` we don't have to deal with real SQL syntax here. So it should be relatively safe. However, if you can think of any other ways to prove the safety, I'm open for suggestions.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -114,31 +114,127 @@ public function __construct($dbh, $statement, OCI8Connection $conn)
...
+     * @param string $tokenOffset The offset to start searching from
+     * @param int $fragmentOffset The offset to build the next fragment from
+     * @param array $fragments Resulting query fragments
+     * @param string|null $currentLiteralDelimiter Current literal delimiter
+     * @param array $paramMap Parameter map
+     * @return bool Whether the token was found
+     */
+    private static function findPlaceholderOrOpeningQuote(
+        $statement,
+        &$tokenOffset,
```
Ocramius: Having by-ref parameters is not something I'd want to maintain/debug later on. I see why it's made, but is there a way to put them together instead? Can we create an object that is passed down instead?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -114,31 +114,127 @@ public function __construct($dbh, $statement, OCI8Connection $conn)
...
+        &$currentLiteralDelimiter,
+        &$paramMap
+    ) {
+        $token = self::findToken($statement, $tokenOffset, '/[?\'"]/');
+
+        if (!$token) {
+            return false;
+        }
+
+        if ($token == '?') {
```
Ocramius: strict comparison please

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -114,31 +114,127 @@ public function __construct($dbh, $statement, OCI8Connection $conn)
...
 
     /**
+     * Finds next placeholder or opening quote.
+     *
+     * @param string $statement The SQL statement to parse
+     * @param string $tokenOffset The offset to start searching from
+     * @param int $fragmentOffset The offset to build the next fragment from
+     * @param array $fragments Resulting query fragments
+     * @param string|null $currentLiteralDelimiter Current literal delimiter
+     * @param array $paramMap Parameter map
```
Ocramius: A bit unclear what kind of map this is. Consider using a syntax such as `string[]` for the type, and explaining what key and value would be

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -114,31 +114,127 @@ public function __construct($dbh, $statement, OCI8Connection $conn)
...
+            return false;
+        }
+
+        if ($token == '?') {
+            $position = count($paramMap) + 1;
+            $param = ':param' . $position;
+            $fragments[] = substr($statement, $fragmentOffset, $tokenOffset - $fragmentOffset);
+            $fragments[] = $param;
+            $paramMap[$position] = $param;
+            $fragmentOffset = ++$tokenOffset;
```
Ocramius: Split this into two statements: `$tokenOffset += 1; $fragmentOffset = $tokenOffset;`

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
Ocramius: If I understand it correctly, this is an `array<int, string>`, where the keys are numeric (positional), and the strings are the parameter names

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -114,31 +114,127 @@ public function __construct($dbh, $statement, OCI8Connection $conn)
...
+        &$paramMap
+    ) {
+        $token = self::findToken($statement, $tokenOffset, '/[?\'"]/');
+
+        if (!$token) {
+            return false;
+        }
+
+        if ($token == '?') {
+            $position = count($paramMap) + 1;
```
Ocramius: Counting shouldn't be repeated: if the parameters of this method are refectored into a tiny object, this can be moved out.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -114,31 +114,127 @@ public function __construct($dbh, $statement, OCI8Connection $conn)
...
-                $stmtLen = strlen($statement); // adjust statement length
-                ++$count;
-            } elseif ($statement[$i] == "'" || $statement[$i] == '"') {
-                $inLiteral = ! $inLiteral; // switch state!
+        $fragmentOffset = $tokenOffset = 0;
+        $fragments = $paramMap = array();
+        $currentLiteralDelimiter = null;
+
+        do {
+            if (!$currentLiteralDelimiter) {
```
Ocramius: If I understand the code correctly, you are currently looping over the structure, and adding to the `$paramMap` as you go, checking for the `"` starting and closing delimiters. Is that correct?

Ocramius: How does this logic currently work with multiline literals?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -114,31 +114,127 @@ public function __construct($dbh, $statement, OCI8Connection $conn)
...
     }
 
     /**
+     * Finds next placeholder or opening quote.
+     *
+     * @param string $statement The SQL statement to parse
+     * @param string $tokenOffset The offset to start searching from
+     * @param int $fragmentOffset The offset to build the next fragment from
+     * @param array $fragments Resulting query fragments
+     * @param string|null $currentLiteralDelimiter Current literal delimiter
```
Ocramius: Description is a bit unclear/redundant

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -114,31 +114,127 @@ public function __construct($dbh, $statement, OCI8Connection $conn)
...
         return array($statement, $paramMap);
     }
 
     /**
+     * Finds next placeholder or opening quote.
+     *
+     * @param string $statement The SQL statement to parse
+     * @param string $tokenOffset The offset to start searching from
+     * @param int $fragmentOffset The offset to build the next fragment from
+     * @param array $fragments Resulting query fragments
```
Ocramius: Description is a bit unclear/redundant - what does determine a fragment?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -114,31 +114,127 @@ public function __construct($dbh, $statement, OCI8Connection $conn)
...
+        }
+
+        if ($token == '?') {
+            $position = count($paramMap) + 1;
+            $param = ':param' . $position;
+            $fragments[] = substr($statement, $fragmentOffset, $tokenOffset - $fragmentOffset);
+            $fragments[] = $param;
+            $paramMap[$position] = $param;
+            $fragmentOffset = ++$tokenOffset;
+        } else {
```
Ocramius: Return early instead of using `else`

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -114,31 +114,127 @@ public function __construct($dbh, $statement, OCI8Connection $conn)
...
+     * @param string $tokenOffset The offset to start searching from
+     * @param string|null $currentLiteralDelimiter Current literal delimiter
+     * @return bool Whether the token was found
+     */
+    private static function findClosingQuote(
+        $statement,
+        &$tokenOffset,
+        &$currentLiteralDelimiter
+    ) {
+        $token = self::findToken($statement, $tokenOffset, '/(?<!\\\\)' . $currentLiteralDelimiter . '/');
```
Ocramius: `$currentLiteralDelimiter` should be passed to `preg_quote()` here

Ocramius: Also, I don't really understand the `?<!` sequence here: can you clarify it? (couldn't make it work on JS regex parsing tools)

Ocramius: Probably need to have a load of security-sensitive string escape sequences to test against here. As good as your intentions are, this is reeeeeally playing around with fire

Ocramius (changes requested):
As much as I like the attempt to improve performance here, I am honestly frightened by possible security implications. I'll have to think about those - meanwhile, consider looking at the comments above.

While it might be slower, I really wish to move away from explicit by-ref parameters, and instead rely on a tiny object containing the arguments being passed around. Unsure if that makes any sense, but please let me know if you can envision that, or if it seems like a complication.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
morozov: It's a negative lookbehind: we're searching for the current literal delimiter which is not preceded by a backslash (i.e. not escaped). The backslash is escaped twice: once to avoid special meaning of the backslash a PHP string and once to avoid its special meaning in the regex.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
morozov: Yes. It works with multiline literals same as with others (I'll add tests for that). In regexes, we don't search for any character classes which do or do not match the newline characters. We only search for `"`, `'` and `?`.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
morozov: A fragment is a part of the original statement not containing placeholders. When re-constructing the query, fragments are joined back together with new parameters between them. For example, for `"SELECT * FROM users WHERE id = ? OR name = ?"`, the fragments are `"SELECT * FROM users WHERE id = "`, `" OR name = "` and `""`.

Probably, this much implementation details in a type description means there should be a dedicated type for that.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
morozov: Yes, it's `array<int, string>`.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
morozov: I think we can move all the parsing logic (not only the by-ref variables) into a different class. It might look like:
```php
$foo = new Foo($sql);
$newSql = $foo->getSql();
$paramMap = $foo->getParameterMap();
```

But I don't like that it only does `get*()` but doesn't do `do*()`. Maybe it could be a proxy (or sub-) class which would translate all messages with positional parameters (`__construct()`, `bind*()`) to the other object which only understands the named ones?

Ocramius:
@morozov sorry to poke you again about this - think it's good to go? There was a lot of discussion, but @deeky666 didn't look at it yet.

morozov:
@Ocramius let me try running some real app tests with this patch and seeing if there are any anomalies. Shame on me, according to both [documentation](http://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements003.htm) and [how it's implemented in the DBAL](https://github.com/doctrine/dbal/blob/v2.5.12/lib/Doctrine/DBAL/Driver/OCI8/OCI8Connection.php#L129), `\'` is not even a valid escaping syntax for Oracle, I should've used `''` instead. I will probably convert unit tests to the functional ones to have the DB prove the resulting query is syntactically and logically correct.

Currently unresolved concerns:
1. By-ref variables. I doubt I can come up with something essentially better without implementing something overcomplicated.
2. Security implications. Not sure where to start. We probably should be looking for some correctly escaped special characters which become unescaped after rebuilding the SQL? What else?
3. Potentially breaking changes. Given that Oracle supports other escaping syntaxes like `q'!name LIKE '%DBMS_%%'!'` we at least want to make sure we're not breaking the queries which happened to work previously. Not sure if we should care much about them given they are not portable.

Will be glad to hear from @deeky666 as well.

morozov:
Please see the update. I ran a large set of mixed unit/integration tests using this branch on Oracle and didn't find any issues. (TBH, I found that the empty string literal was not handled properly in the reworked version and fixed it, added a test).

Ocramius (approved):
@morozov this is good to go: :ship:-ing

PR #2496, opened 2016-09-04
Fix parameter list expansion inside PostgreSQL arrays

### LINKED ISSUE #2295: SQLParserUtils treats brackets in Postgres like quoting characters

maryo:
fixes https://github.com/doctrine/dbal/issues/2295

There is already this PR https://github.com/doctrine/dbal/pull/2296
But it misses tests wich are quite tricky to write for that implementation.

### DIFF
--- a/lib/Doctrine/DBAL/SQLParserUtils.php
+++ b/lib/Doctrine/DBAL/SQLParserUtils.php
@@ -35,7 +35,7 @@ class SQLParserUtils
     const ESCAPED_SINGLE_QUOTED_TEXT = "'(?:[^'\\\\]|\\\\'?|'')*'";
     const ESCAPED_DOUBLE_QUOTED_TEXT = '"(?:[^"\\\\]|\\\\"?)*"';
     const ESCAPED_BACKTICK_QUOTED_TEXT = '`(?:[^`\\\\]|\\\\`?)*`';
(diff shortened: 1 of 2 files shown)


deeky666 (approved):

deeky666:
@maryo thx! Will accept that for now until #2296 is decided over.

deeky666:
Backported to `2.5` via https://github.com/doctrine/dbal/commit/3a351369582dade60c750e2cef540eb7b568e6b3

PR #2498, opened 2016-09-07
MysqliStatement::fetch change return null to return false

### LINKED ISSUE #2497: ResultStatement::fetch() returns null instead of false with mysqli
ResultStatement::fetch() returns null with mysqli, false with sqlsrv. Documented is that it always returns false http://www.doctrine-project.org/api/dbal/2.0/class-Doctrine.DBAL.Driver.ResultStatement.html

http://www.doctrine-project.org/api/dbal/2.3/source-class-Doctrine.DBAL.Driver.Mysqli.MysqliStatement.html#243

I guess the code here should return false instead of null?

PatrickLouys:
Resolves #2497


Ocramius:
Requires a test case.

PR #2500, opened 2016-09-08
Making sure that fetch returns false if nothing is found

### LINKED ISSUE #2497: ResultStatement::fetch() returns null instead of false with mysqli

PatrickLouys:
added test to make sure that fetch returns false if nothing is found

fixed mysqli

Resolves #2497

### DIFF
--- a/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
+++ b/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
@@ -242,7 +242,7 @@ class MysqliStatement implements \IteratorAggregate, Statement
     {
         $values = $this->_fetch();
         if (null === $values) {
-            return null;
+            return false;
         }
 
         if (false === $values) {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -242,7 +242,7 @@ public function fetch($fetchMode = null)
     {
         $values = $this->_fetch();
         if (null === $values) {
-            return null;
```
Ocramius: What happens when a `null` is actually retrieved? Is it supposed to always retrieve an `array`?

PatrickLouys: I'm not sure that I understand your question.
It's supposed to return false or an array (boolean|array).

Now that I think about it, _fetch is not even supposed to return null in the first place, so that should be fixed as well I suppose.

Ocramius:
@PatrickLouys merged, thanks!

`master`: doctrine/dbal@a6d3d162caf815fa954ae5192981688f6b1a1f02
`2.5`: doctrine/dbal@75ac2f6cc7893b9570edc1dfb61d1d1fd1466213

Issue #2502, opened 2016-09-11
Schema add support for a many to many relationship join table as entity


raul782:
Adding this new method createOrUpdateTable would help to fix a very old bug that dates since 2013.
https://github.com/doctrine/doctrine2/issues/3536

raul782:
Closed since PR https://github.com/doctrine/doctrine2/pull/6027 doesn't need dbal library to be modified

PR #2503, opened 2016-09-11
Added updateTable that doesn't throw an exception on m2m if join table already exists

raul782:
This PR fixes the following issue: https://github.com/doctrine/dbal/issues/2502
It is aimed at support the ORM cli tool: doctrine:schema:update reported here.

https://github.com/doctrine/doctrine2/issues/6016


raul782:
PR https://github.com/doctrine/doctrine2/pull/6027 made the current pull request irrelevant.

PR #2504, opened 2016-09-12
URL-decode URL-style DSN

c960657:
The URL-style DSN does not support escaping, so if the username/password/database name contains certain characters, the connection options simply cannot be represented this way.

This patch URL-decodes the different parts. It introduces a small BC break for usernames/passwords/database names containing the character `%`.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,13 @@
+# Upgrade to 2.6
+
+## MINOR BC BREAK: URL-style DSN with percentage sign in password
+
+URL-style DSNs (e.g. ``mysql://foo@bar:localhost/db``) are now assumed to be percent-encoded
+in order to allow certain special characters in usernames, paswords and database names. If
+you are using a URL-style DSN and have a username, password or database name containing a
+percentage sign, you need to update your DSN. If your password is, say, ``foo%foo``, it
+should be encoded as ``foo%25foo``.
+
 # Upgrade to 2.5.1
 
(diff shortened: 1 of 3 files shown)


deeky666:
@Ocramius this looks like a reasonable improvement. The BC question arises of course, especially significant for passwords IMO. What do you think? Shall we accept the slight BC and target this patch for `2.6` only? Also I think this needs documentation.

Ocramius:
This is sane, and can be landed in `2.6`, but it needs documenting in [`UPGRADE.md`](https://github.com/doctrine/dbal/blob/7963f131bef97caa54f7e765334abc5d79562473/UPGRADE.md) before it can be merged. 

@c960657 can you add notes for how an upgrade would be performed, and what the BC break for `%` characters would be? I'd like to also have the `%` character as part of the test suite data-providers (additional entry to what you added), so that the expected output can be viewed by users hitting the problem, and further regressions prevented.

Ocramius (changes requested):
Requesting changes as per https://github.com/doctrine/dbal/pull/2504#issuecomment-272659210

deeky666:
@c960657 also an entry about URL econding/deconding should make its way into the DBAL documentation.

c960657:
I have added a test and some upgrade information.

Ocramius:
@c960657 awesome, thanks!

Ocramius (approved):

PR #2509, opened 2016-09-16
Oracle platform ignores OFFSET in case if LIMIT is not specified

morozov:
There's a non-empty table `users`. The following script should return all existing record IDs except the 1st one.

``` php
$builder = $conn->createQueryBuilder();
$builder->select('id')
    ->from('users')
    ->setFirstResult(1);
$data = $builder->execute()->fetchAll();
var_dump($data);
```

On Oracle, it returns all records ignoring the value set by calling `setFirstResult()`.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
@@ -977,24 +977,29 @@ END;';
      */
     protected function doModifyLimitQuery($query, $limit, $offset = null)
     {
-        $limit = (int) $limit;
-        $offset = (int) $offset;
+        if ($limit === null && $offset === null) {
+            return $query;
+        }
 
         if (preg_match('/^\s*SELECT/i', $query)) {
             if (!preg_match('/\sFROM\s/i', $query)) {
                 $query .= " FROM dual";
             }
-            if ($limit > 0) {
-                $max = $offset + $limit;
-                $column = '*';
-                if ($offset > 0) {
-                    $min = $offset + 1;
-                    $query = 'SELECT * FROM (SELECT a.' . $column . ', rownum AS doctrine_rownum FROM (' .
-                            $query .
-                            ') a WHERE rownum <= ' . $max . ') WHERE doctrine_rownum >= ' . $min;
-                } else {
-                    $query = 'SELECT a.' . $column . ' FROM (' . $query . ') a WHERE ROWNUM <= ' . $max;
-                }
+
+            $columns = array('a.*');
+
+            if ($offset > 0) {
+                $columns[] = 'ROWNUM AS doctrine_rownum';
+            }
+
+            $query = sprintf('SELECT %s FROM (%s) a', implode(', ', $columns), $query);
+
+            if ($limit !== null) {
+                $query .= sprintf(' WHERE ROWNUM <= %d', $offset + $limit);
+            }
+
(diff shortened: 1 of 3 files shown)


lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -975,24 +975,30 @@ public function getName()
      */
     protected function doModifyLimitQuery($query, $limit, $offset = null)
     {
-        $limit = (int) $limit;
```
deeky666: Can you please remove the integer casting entirely and just rely on the method signature `integer|null` for both `$limit` and `$offset`? Then just check everywhere against `null`. Also please remove the `$has*` variables as they are unnecessary.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -975,24 +975,30 @@ public function getName()
...
-                if ($offset > 0) {
-                    $min = $offset + 1;
-                    $query = 'SELECT * FROM (SELECT a.' . $column . ', rownum AS doctrine_rownum FROM (' .
-                            $query .
-                            ') a WHERE rownum <= ' . $max . ') WHERE doctrine_rownum >= ' . $min;
-                } else {
-                    $query = 'SELECT a.' . $column . ' FROM (' . $query . ') a WHERE ROWNUM <= ' . $max;
-                }
+
+            $query = 'SELECT a.*'
```
deeky666: Can you please use `sprintf()` here? Guess that makes it more readable.

deeky666 (changes requested):

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
morozov: @deeky666 The confusing part here is how is `$offset = null` different from `$offset = 0` from the SQL semantics standpoint? Should the DBAL produce additional sub-query in the latter case? Right not it doesn't which makes more sense to me.

deeky666: @morozov yeah good point. But then leave it as before casting both parameters to `int` and checking for `0` instead.

morozov: @deeky666 but unlike `OFFSET`, for `LIMIT` `0` and `null` have different meanings — "0 records" and "all records" respectively. There's no complete consistency between existing adapters but all of them except SQL Anywhere only omit `$limit` when it's `null`. Technically, it is possible to execute a query and retrieve 0 records from it, so the DBAL shouldn't be a limiting factor here. Also from the API consumer standpoint, if there's a bug in the calculation of `$limit` on the application end and it's calculated as `0`, I'd better expect 0 records to be returned instead of all of them.

deeky666: @morozov sounds reasonable. Thanks for clearing that up. I'll accept it as is now.

deeky666 (approved):

deeky666:
@morozov thank you very much!

deeky666:
Backported to `2.5` via https://github.com/doctrine/dbal/commit/b9a3b0876b72ae39191d95bfb37d876a81665a6d

morozov:
@deeky666 thank you for your patience. Can you also review #2463 which is a similar one?

PR #2510, opened 2016-09-17
Schemas with views

Bashka:
comparison schemas with views


Bashka:
After removed unused namespaces TravisCI failed? WTF?

Issue #2515, opened 2016-09-21
PHP Warning when listing table columns in Oracle


nitso:
```
PHP Warning:  oci_fetch_all(): ORA-01427: single-row subquery returns more than one row in /opt/kvk/WB/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php on line 290
```

Synopsis: 
1. Create two databases accessible by current user
2. Add identical tables with same name in both databases
3. Try to get database schema (in my case I was running a migration)

Marius2805:
This bug is already fixed (including test) in pull reqeust:
https://github.com/doctrine/dbal/pull/2518

deeky666:
Duplicate of #2517

PR #2516, opened 2016-09-21
#2515 fix schema listing when having identical tables in databases

nitso:
Subquery may return more than one row, so additional owner condition was added


Ocramius:
Requires test case.

deeky666:
Fixed in #2518

PR #2518, opened 2016-09-22
Fixed ORA-01427 when listing columns

marius-meissner:
Fixes #2517

Issue #2517, opened 2016-09-22
ORA-01427 when listing table columns


Marius2805:
Example:
`[SchemaManager]->listTableColumns('TABLE01', 'SCHEMA01')`

Under the following criteria a ORA-01427 is thrown:
- $database parameter != null
- Table exists in another schema with the same name
- Both tables have at least one column with the same name

Pull request + test will follow.

PR #2522, opened 2016-09-26
OraclePlatform - ORA-01427: single-row subquery returns more than one row

astepin:
The "all_col_comments" table contains all column comments for all database tables in all tablespaces. If the connection user has access to multiple tablespaces, this query fails with the following error:
ORA-01427: single-row subquery returns more than one row


nitso:
One more dupe #2515 #2517

Deamon:
Sorry for an other duplicate #2584 with associated PR #2585.
See this thread only now.

deeky666:
Fixed in #2518

Issue #2523, opened 2016-09-28
Truncate Doctrine\DBAL\Exception\DriverException message if too long


shtrom:
Hi have an issue---triggered by ownCloud and an erroneous vCard with a duplicated `PHOTO` field---where a database insertion request into MySQL via the Doctrine DBAL fails.

Essentially, the blob inserted is too big for MySQL, and it issues  a `Syntax error or access violation: 1118 The size of BLOB/TEXT data inserted in one transaction is greater than 10% of redo log size.”` which gets transformed into a `Doctrine\DBAL\Exception\DriverException`.

The message for the exception is `An exception occurred while executing [QUERY] with params [PARAMS]`. This is where the issue with Doctrine is: the blob is about 9MiB at this stage, and it is dumped verbatim as part of the `PARAMS` bit in the error message. This makes the Exception a bit hard to understand, as it is not concise enough, and this makes log files grow at a very very high rate.

A solution to this problem would be to either
- truncate or ellipsise `PARAMS` longer than a given threshold in the error message, e.g., `BEGIN:VCARD\r ... END:VCARD\r`(line ending might or might not be a good idea to rely on)
- not reproduce blobs, and replace them altogether with something like `BLOB of 9MiB`, or
- a combination of both:  `BEGIN:VCARD\r [about 9MiB of blob] END:VCARD\r`.

Note that my example is with ownCloud and vCards, but the problem is more generic, with any blob too big for comfortable display.

kimhemsoe:
Sounds to me that it is the exception handler which prints out the exception. Any reason why this could not be fixed in the presentation of the exception?

shtrom:
This could work. However the DriverException has a better view into which `$param` is the longest, and can cut the cruft more accurately, without touching the others.

In my example above, there is still valuable information at the end of the exception (in the output of `->getMessage()`), which would otherwise by removed at the logging site by, e.g., an `mb_strimwidth`, as I've used in my test above.

shtrom:
Yup, the fix in the commit above fixes the issue nicely, while retaining useful information. It truncates the long blob properly, but retains useful things such as the stacktrace after that.

Ocramius:
Looking at #2524, I think this is up to the exception handler printing the error. It is indeed a lot of data, but imagine having an incorrect character at position 2500 (invalid encoding), and then going mad :-P

shtrom:
>  It is indeed a lot of data, but imagine having an incorrect character at position 2500 (invalid encoding)

[...]

PR #2524, opened 2016-09-28
Truncate params in DBALException::formatParameters

shtrom:
This is a fix for #2523.

This simply uses `mb_strimwidth` to add an ellipsis at the end of string parameters longer than 240  characters. Doing so allows to remove information overload from long parameters, while retaining details from other parameters and subsequent information in the exception message (e.g., stacktrace).

Signed-off-by: Olivier Mehani shtrom@ssji.net

### DIFF
--- a/lib/Doctrine/DBAL/DBALException.php
+++ b/lib/Doctrine/DBAL/DBALException.php
@@ -25,6 +25,9 @@ use Doctrine\DBAL\Driver\ExceptionConverterDriver;
 
 class DBALException extends \Exception
 {
+    /** Length after which a parameter argument is truncated in the formatted error message */
+    const MAX_PARAM_LENGTH = 240;
+
     /**
      * @param string $method
      *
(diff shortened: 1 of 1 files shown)


shtrom:
Fix for #2523

Ocramius:
Requires a test case. Also, we don't have a dependency on `ext/mbstring`: https://github.com/doctrine/dbal/blob/ec8fc16604915110eea8046e3ea594c133aee880/composer.json#L14-L17

lib/Doctrine/DBAL/DBALException.php
```diff
@@ -168,6 +168,9 @@ private static function wrapException(Driver $driver, \Exception $driverEx, $msg
     private static function formatParameters(array $params)
     {
         return '[' . implode(', ', array_map(function ($param) {
+            if (is_string ($param)) {
+                $param = mb_strimwidth($param, 0, 240, "...");
```
Ocramius: If you really want to truncate, then then length should also be part of the string here.

morozov:
This change is obsolete as of https://github.com/doctrine/dbal/pull/4387.

PR #2527, opened 2016-10-01
Normalize method signatures for `fetch()` and `fetchAll()`, ensuring compatibility with the `PDOStatement` signature

### LINKED ISSUE #2519: [RFC] `ResultStatement::fetchAll()` must define 3 arguments
[`ResultStatement::fetchAll()`](https://github.com/doctrine/dbal/blob/ec8fc16604915110eea8046e3ea594c133aee880/lib/Doctrine/DBAL/Driver/ResultStatement.php#L83) must define 3 arguments, like [`\PDOStatement::fetchAll()`](http://php.net/manual/en/pdostatement.fetchall.php).
Even, there is a signature violation at [`PDOStatement::fetchAll()`](https://github.com/doctrine/dbal/blob/ec8fc16604915110eea8046e3ea594c133aee880/lib/Doctrine/DBAL/Driver/PDOStatement.php#L124), where `\PDOStatement` is extended and [`Statement`](https://github.com/doctrine/dbal/blob/ec8fc16604915110eea8046e3ea594c133aee880/lib/Doctrine/DBAL/Driver/Statement.php#L33) interface is implemeted; this finally extends `ResultStatement`.

I know this change is a BC break, but IMO the current implementation is wrong given what I'm exposing here.

phansys:
| Q | A |
| --- | --- |
| Branch | 2.5 |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | yes |
| Deprecations? | no |
| Tests pass? | yes |
| Fixed tickets | #2519 |
| License | MIT |
| Doc PR | n/a |

**TODO**
- [x] Check usage of `func_get_args()` at `DB2Statement`, `SQLAnywhereStatement` and `SQLSrvStatement`.

Close #2519.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,30 @@
 # Upgrade to 2.6
 
+## MINOR BC BREAK: `fetch()` and `fetchAll()` method signatures in `Doctrine\DBAL\Driver\ResultStatement`
+
+1. ``Doctrine\DBAL\Driver\ResultStatement::fetch()`` now has 3 arguments instead of 1, respecting
+``PDO::fetch()`` signature.
+
+Before:
+
+    Doctrine\DBAL\Driver\ResultStatement::fetch($fetchMode);
+
+After:
+
+    Doctrine\DBAL\Driver\ResultStatement::fetch($fetchMode, $cursorOrientation, $cursorOffset);
+
+2. ``Doctrine\DBAL\Driver\ResultStatement::fetchAll()`` now has 3 arguments instead of 1, respecting
+``PDO::fetchAll()`` signature.
+
+Before:
+
+    Doctrine\DBAL\Driver\ResultStatement::fetchAll($fetchMode);
+
+After:
+
+    Doctrine\DBAL\Driver\ResultStatement::fetch($fetchMode, $fetchArgument, $ctorArgs);
+
+
 ## MINOR BC BREAK: URL-style DSN with percentage sign in password
 
 URL-style DSNs (e.g. ``mysql://foo@bar:localhost/db``) are now assumed to be percent-encoded
--- a/lib/Doctrine/DBAL/Driver/ResultStatement.php
+++ b/lib/Doctrine/DBAL/Driver/ResultStatement.php
@@ -58,29 +58,53 @@ interface ResultStatement extends \Traversable
     /**
      * Returns the next row of a result set.
      *
-     * @param integer|null $fetchMode Controls how the next row will be returned to the caller.
-     *                                The value must be one of the PDO::FETCH_* constants,
-     *                                defaulting to PDO::FETCH_BOTH.
+     * * @param int|null $fetchMode  Controls how the next row will be returned to the caller.
+     *                               The value must be one of the \PDO::FETCH_* constants,
+     *                               defaulting to \PDO::FETCH_BOTH.
+     * @param int $cursorOrientation For a ResultStatement object representing a scrollable cursor,
+     *                               this value determines which row will be returned to the caller.
+     *                               This value must be one of the \PDO::FETCH_ORI_* constants,
+     *                               defaulting to \PDO::FETCH_ORI_NEXT. To request a scrollable
+     *                               cursor for your ResultStatement object, you must set the \PDO::ATTR_CURSOR
+     *                               attribute to \PDO::CURSOR_SCROLL when you prepare the SQL statement with
+     *                               \PDO::prepare().
+     * @param int $cursorOffset      For a ResultStatement object representing a scrollable cursor for which the
+     *                               cursorOrientation parameter is set to \PDO::FETCH_ORI_ABS, this value
+     *                               specifies the absolute number of the row in the result set that shall be
+     *                               fetched.
+     *                               For a ResultStatement object representing a scrollable cursor for which the
+     *                               cursorOrientation parameter is set to \PDO::FETCH_ORI_REL, this value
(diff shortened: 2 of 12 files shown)


Ocramius:
Please PR against master. Backporting is up to us.

mikeSimonson:
@phansys What about the usage of func_get_args ?

deeky666:
@phansys yes please do as @mikeSimonson suggested. We already follow this approach [here](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Driver/PDOConnection.php#L86-L108)

phansys:
@mikeSimonson, @deeky666; the reason for the comment in my **TODO** is that I'm trying to normalize the method signatures, making them as explicitly as possible. Using `func_get_args()` to unpack method arguments is precisely doing the opposite.

mikeSimonson:
@phansys If it's possible to make the signature more explicit in those cases go for it.
As long as it doesn't break backward compatibility.
If it does put it in another PR.

phansys:
The BC is respected internally, since `ResultStatement::fetch()` and `ResultStatement::fetchAll()` have default values for the new arguments. But this could lead to issues if somebody was using `ResultStatement` interface for their own classes.

phansys:
Which is the status of this PR? Ping @Ocramius, @deeky666, @mikeSimonson.

Ocramius (approved):

lib/Doctrine/DBAL/Driver/ResultStatement.php
```diff
@@ -58,29 +58,53 @@ public function setFetchMode($fetchMode, $arg2 = null, $arg3 = null);
...
+     *                               specifies the row to fetch relative to the cursor position before
+     *                               ResultStatement::fetch() was called.
      *
      * @return mixed The return value of this method on success depends on the fetch mode. In all cases, FALSE is
      *               returned on failure.
      *
      * @see PDO::FETCH_* constants.
      */
-    public function fetch($fetchMode = null);
+    public function fetch($fetchMode = null, $cursorOrientation = \PDO::FETCH_ORI_NEXT, $cursorOffset = 0);
```
Ocramius: This signature change is a BC break and needs upgrade/migration path in `UPGRADE.md`

lib/Doctrine/DBAL/Driver/ResultStatement.php
```diff
@@ -58,29 +58,53 @@ public function setFetchMode($fetchMode, $arg2 = null, $arg3 = null);
...
+     *                                The value must be one of the \PDO::FETCH_* constants,
+     *                                defaulting to \PDO::FETCH_BOTH.
      *
      * @return array
      *
-     * @see PDO::FETCH_* constants.
+     * @see \PDO::FETCH_* constants.
      */
-    public function fetchAll($fetchMode = null);
+    public function fetchAll($fetchMode = null, $fetchArgument = null, $ctorArgs = null);
```
Ocramius: This signature change is a BC break and needs upgrade/migration path in `UPGRADE.md`

Ocramius (changes requested):
@phansys excellent work, but we need to document the migration approach for existing implementations in `UPGRADE.md`

lib/Doctrine/DBAL/Driver/ResultStatement.php
phansys: Thank you @Ocramius, I've added a note in the upgrade path for 2.6 regarding these changes.

Ocramius:
Thanks @phansys!

Ocramius (approved):

UPGRADE.md
```diff
@@ -1,5 +1,30 @@
...
+2. ``Doctrine\DBAL\Driver\ResultStatement::fetchAll()`` now has 3 arguments instead of 1, respecting
+``PDO::fetchAll()`` signature.
+
+Before:
+
+    Doctrine\DBAL\Driver\ResultStatement::fetchAll($fetchMode);
+
+After:
+
+    Doctrine\DBAL\Driver\ResultStatement::fetch($fetchMode, $fetchArgument, $ctorArgs);
```
morozov: It probably should've been `Doctrine\DBAL\Driver\ResultStatement::fetchAll($fetchMode, $fetchArgument, $ctorArgs);`.

morozov:
After #2953 gets resolved, we may want to revert the API changes made here. Or maybe we could revert `fetchAll()` as part of #2953 (it already reverts the `fetch()` part)?

/cc @Ocramius

Ocramius:
As part of #2953 would be best, since it already deals with moving away from PDO

morozov:
Apparently, some drivers like [IBM DB2](https://github.com/doctrine/dbal/blob/e3ab127672fd72e5e64a17f20b11ab2d12b8bfc5/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php#L228), [SQLServer](https://github.com/doctrine/dbal/blob/e3ab127672fd72e5e64a17f20b11ab2d12b8bfc5/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php#L315)  and others implement handling additional arguments in `fetch()` and `fetchAll()` using `func_get_args()`, so removing all arguments except `$fetchMode` from the `fetch*()` methods is incorrect. I thought those arguments were introduced solely for compatibility with `PDOStatement`. Thanks to `DataAccessTest::testFetchAllSupportFetchClass()`.

Instead, we can change their signatures to `fetch*($fetchMode = null, ...$args)` to make the optional arguments explicit and at the same time independent on the `PDO` constants.

PR #2534, opened 2016-10-12
Handle errors during DB2 fetch operations

chriseskow:
I recently ran into a situation where one of our DB2 tables got into an error state due to a failed/interrupted LOAD operation. While the table was in this state, running a SELECT query against it from the DB2 command line returned an error message. However, running the same query through Doctrine DBAL produced no errors - just an empty result set. This was misleading because the query should have failed completely if the table was in the error state, or returned a valid, non-empty result set otherwise.

It turns out that this particular error only occurs when the `db2_fetch_xxx()` functions are called - the connect, prepare, and execute steps all run fine without errors. DBAL currently doesn't check for errors during the fetch. This pull request updates `DB2Statement::fetch()` to check if the result is `false`, and if so throws an exception if `db2_stmt_errormsg()` returns a non-empty result. I also `@`-silenced each of the `db2_fetch_xxx()` calls, since they emit PHP warnings when the fetch fails.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
+++ b/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
@@ -231,9 +231,9 @@ class DB2Statement implements \IteratorAggregate, Statement
                 throw new DB2Exception("Given Fetch-Style " . $fetchMode . " is not supported.");
         }
 
-        $error = db2_stmt_errormsg();
-        if (false === $result && $error) {
-            throw new DB2Exception("Error during fetch: " . $error);
+        list($errorMessage, $errorCode) = $this->errorInfo();
+        if (false === $result && !empty($errorMessage)) {
+            throw new DB2Exception("Error during fetch: " . $errorMessage);
         }
 
         return $result;
--- a/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
+++ b/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
@@ -233,7 +233,7 @@ class DB2Statement implements \IteratorAggregate, Statement
 
         $error = db2_stmt_errormsg();
         if (false === $result && $error) {
-            throw new DB2Exception($error);
+            throw new DB2Exception("Error during fetch: " . $error);
         }
 
         return $result;
--- a/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
+++ b/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
@@ -232,7 +232,7 @@ class DB2Statement implements \IteratorAggregate, Statement
         }
 
         $error = db2_stmt_errormsg();
-        if ($result === false && $error) {
+        if (false === $result && $error) {
             throw new DB2Exception($error);
         }
 
--- a/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
+++ b/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
@@ -200,10 +200,10 @@ class DB2Statement implements \IteratorAggregate, Statement
         $fetchMode = $fetchMode ?: $this->_defaultFetchMode;
         switch ($fetchMode) {
             case \PDO::FETCH_BOTH:
-                $result = @db2_fetch_both($this->_stmt);
+                $result = db2_fetch_both($this->_stmt);
                 break;
             case \PDO::FETCH_ASSOC:
-                $result = @db2_fetch_assoc($this->_stmt);
+                $result = db2_fetch_assoc($this->_stmt);
                 break;
             case \PDO::FETCH_CLASS:
                 $className = $this->defaultFetchClass;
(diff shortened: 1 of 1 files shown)


kimhemsoe:
Is it possible to write a test this error?

chriseskow:
I'm not sure... Our DBAs were able to reproduce it by running a NONRECOVERABLE LOAD operation on a table, then hitting Ctrl+C partway through to interrupt it. I'm familiar with DBAL's test setup so I'm not sure how to write a test for it...

kimhemsoe:
It is pretty straight forward.

Create a copy of phpunit.xml.dist config file and edit it.
You can run the suite with "./vendor/bin/phpunit -c [configfile]"
Tests can be found in "/tests/Doctrine/Tests/Dbal/Driver".

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
```diff
@@ -200,9 +200,11 @@ public function fetch($fetchMode = null)
         $fetchMode = $fetchMode ?: $this->_defaultFetchMode;
         switch ($fetchMode) {
             case \PDO::FETCH_BOTH:
-                return db2_fetch_both($this->_stmt);
+                $result = @db2_fetch_both($this->_stmt);
```
Ocramius: Silencing is absolutely a no-go

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
```diff
@@ -213,20 +215,28 @@ public function fetch($fetchMode = null)
...
+                break;
             case \PDO::FETCH_OBJ:
-                return db2_fetch_object($this->_stmt);
+                $result = @db2_fetch_object($this->_stmt);
+                break;
             default:
                 throw new DB2Exception("Given Fetch-Style " . $fetchMode . " is not supported.");
         }
+
+        $error = db2_stmt_errormsg();
```
Ocramius: Can we move this block to a private method?

Ocramius: Also, since DB2 only provides a procedural, singleton-ish API, the only way to test this is to mock the function.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
```diff
@@ -213,20 +215,28 @@ public function fetch($fetchMode = null)
...
+        if ($result === false && $error) {
+            throw new DB2Exception($error);
```
Ocramius: Can you add some details here? It is beneficial for consumers to know how we got to this failure (failure during fetch)

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
```diff
@@ -213,20 +215,28 @@ public function fetch($fetchMode = null)
...
             case \PDO::FETCH_OBJ:
-                return db2_fetch_object($this->_stmt);
+                $result = @db2_fetch_object($this->_stmt);
+                break;
             default:
                 throw new DB2Exception("Given Fetch-Style " . $fetchMode . " is not supported.");
         }
+
+        $error = db2_stmt_errormsg();
+        if ($result === false && $error) {
```
Ocramius: Please yoda this comparison

Ocramius (changes requested):

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
chriseskow: The other DB2 operations that emit warnings are already be silenced (db2_execute, db2_prepare, etc.). Same with the Oracle driver.

Ocramius: There should be an error handler then, not a silence operator. This just makes it terrible to find out if something goes wrong.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
chriseskow: In order to unit test it, we'll also need to move the rest of the procedural `db2_fetch_xxx()` calls into separate method(s).... Shall I go ahead and do that?

chriseskow: What if I simply move the original contents of the `fetch` method to a private method `doFetch`, and make `fetch` call `doFetch` and check the result?

``` php
    public function fetch($fetchMode = null)
    {
        $result = call_user_func_array(array($this, 'doFetch'), func_get_args());

        list($errorMessage, $errorCode) = $this->errorInfo();
        if (false === $result && !empty($errorMessage)) {
            throw [...]
        }

        return $result;
    }
```

That way, the original code (which wasn't unit tested anyway) is left alone, and the new, testable code wraps it.

chriseskow: Actually that won't work because we can't mock private functions. @Ocramius how would you suggest testing this? In my company's projects we use [php-mock/php-mock-phpunit](https://github.com/php-mock/php-mock-phpunit) to mock built-in functions, but Doctrine doesn't use this. Otherwise, could create a `DB2StatementMock` subclass which mocks `doFetch()`, and make that the SUT, but then we're not really testing `DB2Statement` directly.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
```diff
@@ -218,15 +220,23 @@ public function fetch($fetchMode = null)
...
-                return db2_fetch_object($this->_stmt);
+                $result = db2_fetch_object($this->_stmt);
+                break;
             default:
                 throw new DB2Exception("Given Fetch-Style " . $fetchMode . " is not supported.");
         }
+
+        list($errorMessage, $errorCode) = $this->errorInfo();
+        if (false === $result && !empty($errorMessage)) {
+            throw new DB2Exception("Error during fetch: " . $errorMessage);
```
deeky666: I'm not quite sure about this. The interface does not allow such exceptions here. I suppose the correct way would be to check the result for `false` and then fetch the error information via `errorInfo()` manually. Otherwise it does not fulfill the contract.

PR #2536, opened 2016-10-13
Bind result variables after each execution in order to prevent string truncation

morozov:
The issue is _only_ reproducible when `ext-mysqli` is compiled against `libmysqlclient` (for instance on MAMP or on Linux when compiled as `./configure --with-mysqli=/usr/bin/mysql_config`) and is _not_ reproducible in case of `mysqlnd`.

If a `SELECT` statement is reused, and the results of a subsequent execution contain longer strings than the result of the first execution, strings get truncated to the length of the originally selected strings.

According to [user contributed notes for `mysqli_stmt::store_result()`](http://php.net/manual/en/mysqli-stmt.store-result.php#99159), the methods on the statement must be called in the order of `prepare()` → `execute()` → `store_result()` → `bind_result()` → `fetch()`, but current implementation calls `bind_result()` _only_ after the first execution, so PHP doesn't reallocate memory segments for the new results, so they may not fit the previously allocated ones.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
+++ b/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
@@ -168,9 +168,6 @@ class MysqliStatement implements \IteratorAggregate, Statement
         if (null === $this->_columnNames) {
             $meta = $this->_stmt->result_metadata();
             if (false !== $meta) {
-                // We have a result.
-                $this->_stmt->store_result();
-
                 $columnNames = array();
                 foreach ($meta->fetch_fields() as $col) {
                     $columnNames[] = $col->name;
@@ -178,18 +175,37 @@ class MysqliStatement implements \IteratorAggregate, Statement
                 $meta->free();
 
                 $this->_columnNames = $columnNames;
-                $this->_rowBindedValues = array_fill(0, count($columnNames), null);
-
-                $refs = array();
-                foreach ($this->_rowBindedValues as $key => &$value) {
-                    $refs[$key] =& $value;
-                }
-
-                if (!call_user_func_array(array($this->_stmt, 'bind_result'), $refs)) {
-                    throw new MysqliException($this->_stmt->error, $this->_stmt->sqlstate, $this->_stmt->errno);
-                }
             } else {
                 $this->_columnNames = false;
+            }
+        }
+
+        if (false !== $this->_columnNames) {
+            // Store result of every execution which has it. Otherwise it will be impossible
+            // to execute a new statement in case if the previous one has non-fetched rows
+            // @link http://dev.mysql.com/doc/refman/5.7/en/commands-out-of-sync.html
+            $this->_stmt->store_result();
+
+            // Bind row values _after_ storing the result. Otherwise, if mysqli is compiled with libmysql,
+            // it will have to allocate as much memory as it may be needed for the given column type
+            // (e.g. for a LONGBLOB field it's 4 gigabytes)
+            // @link https://bugs.php.net/bug.php?id=51386#1270673122
+            //
+            // Make sure that the values are bound after each execution. Otherwise, if closeCursor() has been
+            // previously called on the statement, the values are unbound making the statement unusable.
+            //
+            // It's also important that row values are bound after _each_ call to store_result(). Otherwise,
+            // if mysqli is compiled with libmysql, subsequently fetched string values will get truncated
+            // to the length of the ones fetched during the previous execution.
+            $this->_rowBindedValues = array_fill(0, count($this->_columnNames), null);
+
+            $refs = array();
+            foreach ($this->_rowBindedValues as $key => &$value) {
+                $refs[$key] =& $value;
+            }
+
+            if (!call_user_func_array(array($this->_stmt, 'bind_result'), $refs)) {
+                throw new MysqliException($this->_stmt->error, $this->_stmt->sqlstate, $this->_stmt->errno);
             }
         }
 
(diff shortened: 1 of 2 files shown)


kimhemsoe:
I will try to spin up a debian box and verify in the next few days, if i remember correct they do not use mysqlnd.

Meanwhile, it is possible for you to test that we do not reintroduce #433

kimhemsoe:
Does this close #2487?

morozov:
@kimhemsoe it resolves the MySQLi related part of #2487 but apparently, `closeCursor()` doesn't work properly on other platforms, so I'd leave it as is for now and update/reopen after/if this one gets resolved.

kimhemsoe:
One can only love the php mysqli and all its lovely thorns.

Wonder if it is worth to have 2 code paths, one for libmysqli and another for mysqlnd

morozov:
> Meanwhile, it is possible for you to test that we do not reintroduce #433?

AFAIK, the memory issue with LOBs occurs on mysqli+libmysqlclient in case if `mysqli_stmt::store_result()` is not called at all or is called in an improper place (see https://bugs.php.net/bug.php?id=51386). According to the history, prior to #433 it was called after `bind_result()`, and the patch was to call it before. With my patch it's still called before `bind_result()`, so it shouldn't be a problem (however I'll check).

> Wonder if it is worth to have 2 code paths, one for libmysqli and another for mysqlnd

If it's only needed to prevent regression of #433, then probably no. However it does make sense to run tests against mysqli+libmysqlclient on Travis CI.

kimhemsoe:
I dont believe it will be an issue either.

> If it's only needed to prevent regression of #433, then probably no. However it does make sense to run tests against mysqli+libmysqlclient on Travis CI.

It would give us access to use mysqli_stmt::get_result() but that is about that. Any idea how we can test both on travis?

morozov:
> Any idea how we can test both on travis?

According to [default_configure_options.erb](https://github.com/travis-ci/travis-cookbooks/blob/precise-stable/ci_environment/phpbuild/templates/default/default_configure_options.erb), mysqli is compiled as a built-in extension, so there's no to way to disable it and recompile it with libmysqliclient. It should be possible to create a [custom docker image](https://docs.travis-ci.com/user/docker/) (ideally, clonned from some standard Travis one for PHP) and use it instead of the OOTB one but it looks like a project on its own.

morozov:
> Does this close #2487?

>> @kimhemsoe it resolves the MySQLi related part of #2487 but apparently, closeCursor() doesn't work properly on other platforms, so I'd leave it as is for now and update/reopen after/if this one gets resolved.

Sorry, I confused #2487 with #2489 which will be partially resolved by this one. #2487 is resolved entirely. I can move its unit tests here after the patch is approved.

morozov:
@kimhemsoe any update regarding the review?

I've added a regression test for #433 in the meanwhile:
```
↪  git checkout v2.4.1
↪  git checkout mysqli-always-bind-result -- tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
↪  php -r 'echo mysqli_get_client_info(), PHP_EOL;'
5.7.15
↪  phpunit --filter testFetchLongBlob -c mysql.phpunit.xml tests/Doctrine/Tests/DBAL/Functional/StatementTest.php

   Fatal error: Allowed memory size of 4294967296 bytes exhausted (tried to allocate 4294967296 bytes) in lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php on line 186
↪  git checkout mysqli-always-bind-result
↪  phpunit --filter testFetchLongBlob -c mysql.phpunit.xml tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
PHPUnit 5.6.1 by Sebastian Bergmann and contributors.

.                                                                   1 / 1 (100%)

Time: 253 ms, Memory: 16.75MB

OK (1 test, 1 assertion)
```

kimhemsoe:
Looks good to me.

@Ocramius Should i just merge this or wait for @deeky666?

Ocramius:
I'd wait for @deeky666.

Also, this sounds like a mysqli issue? Is this intended behavior? If so, who do we chase in order to become more sane?

morozov:
> Also, this sounds like a mysqli issue?

It's a [bug](https://bugs.php.net/bug.php?id=51386#1270673122) marked as won't fix with proposed workaround, however not documented anywhere in official PHP documentation.

Ocramius:
Wow, that sounds bad :S

I really dislike fixing other people's bugs, so we at least need comments
in the code explaining and linking the reasoning in a very detailed way.

morozov:
@Ocramius please see the updated pull request. I cannot find documentation for all mysqli quirks described in code comments, but they all are reflected in tests.

kimhemsoe:
Rubbing salt into the wound, the libmysql [documentation](http://dev.mysql.com/doc/refman/5.6/en/mysql-stmt-store-result.html) says that the order should be "bind, store, fetch" and not "store, bind, fetch"

morozov:
@kimhemsoe it was like this prior to fixing #433 :-)

kimhemsoe:
hehe i know, i wrote it :D I do remember that i had many of the same issues back when i wrote the driver. Cant remember why i settled on that specific call order. 

A few links to show how php handle this internally.

https://github.com/php/php-src/blob/master/ext/pdo_mysql/mysql_statement.c#L255
https://github.com/php/php-src/blob/master/ext/mysqli/mysqli_api.c#L2523
https://github.com/php/php-src/blob/master/ext/mysqli/mysqli_api.c#L504

I know little about libmysql and its quirks, so i am not trying to be clever about this. 

I do believe that we should do as advised in the php bug and comments in mysqli ext src

deeky666:
@morozov looks reasonable to me. Though, can you integrate this one into #2546 because they are sort of depending on each other and then you can omit the skippings in the tests. Thanks!

morozov:
Closing as being taken care of by #2546.

PR #2537, opened 2016-10-15
Fixed bug with detection of changed datetime/datetimetz

patrickbussmann:
I have a bug that after updating my fields from datetime to datetimetz that everytime I want to update my database I see the same changes:

```
ALTER TABLE cart CHANGE created_at created_at DATETIME NOT NULL, CHANGE updated_at updated_at DATETIME NOT NULL;
```

After executing:

```
doctrine:schema:update --force
```

and (again)

```
doctrine:schema:update --dump-sql
```

I got again

```
ALTER TABLE cart CHANGE created_at created_at DATETIME NOT NULL, CHANGE updated_at updated_at DATETIME NOT NULL
```

The problem is that the comparator not see the equality of DateTime and DateTimeTz in **MySQL databases**.
Or there is a problem with MySQL that they havent a special type for this and so doctrine think its different.

You can check this and merge if its useful.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Comparator.php
+++ b/lib/Doctrine/DBAL/Schema/Comparator.php
@@ -431,7 +431,14 @@ class Comparator
 
         foreach (array('type', 'notnull', 'unsigned', 'autoincrement') as $property) {
             if ($properties1[$property] != $properties2[$property]) {
-                $changedProperties[] = $property;
+            	if(!(
+	                ($properties1[$property] instanceof Types\DateTimeType || $properties1[$property] instanceof Types\DateTimeTzType)
+		            &&
+	                ($properties2[$property] instanceof Types\DateTimeType || $properties2[$property] instanceof Types\DateTimeTzType)
+	            ))
+	            {
+		            $changedProperties[] = $property;
+	            }
             }
         }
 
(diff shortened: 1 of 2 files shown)


deeky666:
@patrickbussmann [please have a look into the documentation](http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html#id100), especially Note 15 and 16. This is a limitation that some vendors like MySQL have. MySQL won't be able to store your Timezone information anyways, because it is lacking a native type that could do that. So not only are you having troubles with schema comparison but also with data loss.
Can you please reconsider that and eventually close this PR if you agree with that? Thanks

patrickbussmann:
Hmm you're right - BUT this is the reason why I did this pull request.
It solves the problem :-)

But if you say that this is not usefu I'll close this. No problem.

deeky666:
@patrickbussmann this PR does not solve the data loss issue though. Also the correct fix to distinguish ambiguous data types (that cannot be distinguished during schema introspection because of the lack of a native type) would be to make the ambiguous data type be commented via the column so that Doctrine can infer the correct type nevertheless. So I guess it could be possible to fix this issue entirely but not like suggested in this PR. To fix it correctly, we would have to change the datetimetz type to be eventually stored commented in the database and have the platform evaluate the proper fallback type and whether it has to be commented or not (whether a native type is available or not) just like we do for the json type for example.

PR #2539, opened 2016-10-21
retrieve version from query. 

catalinux:
This is needed when you expecially when you get Dbal Connection from a laravel connection

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -435,14 +435,12 @@ class Connection implements DriverConnection
             return $this->_conn->getServerVersion();
         }
 
-        if ($this->_conn instanceof \PDO) {
+        if ($this->_conn instanceof \PDO && $this->_conn->getAttribute(PDO::ATTR_DRIVER_NAME)==='mysql') {
             try {
                 return $this->_conn->query('select version()')->fetchColumn();
             } catch (\PDOException $e) {
                 //version query not supported
             }
-
-
         }
         // Unable to detect platform version.
         return null;
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -435,6 +435,15 @@ class Connection implements DriverConnection
             return $this->_conn->getServerVersion();
         }
 
+        if ($this->_conn instanceof \PDO) {
+            try {
+                return $this->_conn->query('select version()')->fetchColumn();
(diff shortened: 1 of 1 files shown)


kimhemsoe:
You can explain the usecase a more for those of us whom have never touched laravel?

Also i am guessing "version()" does not exist on all platforms and/or agree on the return format, so your proposed solution is a nogo.

catalinux:
Ok. When the connection is native PHP \PDO,  it won't detect the version.

catalinux:
I could add mysql testing so, I could have

```php
        if ($this->_conn instanceof \PDO && $this->_conn->getAttribute(PDO::ATTR_DRIVER_NAME)==='mysql') {
            try {
                return $this->_conn->query('select version()')->fetchColumn();
            } catch (\PDOException $e) {
                //version query not supported
            }
        }

**Comment by Ocramius** [comment]
> Ok. When the connection is native PHP \PDO, it won't detect the version. 

If you provide doctrine with a `PDO` instance on your own, then it should be up to you to also provide the correct platform to the connection, since you may have done a lot of custom stuff on it anyway.

Can you eventually demonstrate the issue with a failing test case? We don't merge anything that is not coming with corresponding test cases, sorry...

File: lib/Doctrine/DBAL/Connection.php [hunk 1]
```diff
@@ -435,6 +435,13 @@ private function getDatabasePlatformVersion()
             return $this->_conn->getServerVersion();
         }
 
+        if ($this->_conn instanceof \PDO && $this->_conn->getAttribute(PDO::ATTR_DRIVER_NAME)==='mysql') {
+            try {
+                return $this->_conn->query('select version()')->fetchColumn();
+            } catch (\PDOException $e) {
```
Ocramius: I'd rather have a failure here, since it's mysql, and yet something is going horribly wrong. This `catch` block should go away.

catalinux: It makes sense to make tests for my own code.  My problem was that when wrapping a PDO Connection of MySql , `getDatabasePlatformVersion` returns null. Looking more into the code I see there is a `requiresQueryForServerVersion` mechanism but for PDO this method return false. 

I'll dig more into this problem.

Ocramius: What you just described is what should be formalized into a test case.

morozov:
Probably obsolete as of https://github.com/doctrine/dbal/pull/4115.

PR #2543, opened 2016-10-25
Hotfix/changing column from notnull to null

pedrosik:


Ocramius:
Please only PR against `master`. Backporting is up to us.

PR #2546, opened 2016-10-30
Prepared statements become unusable after calling closeCursor() on IBM DB2, Oracle and MS SQL Server.

morozov:
The adapters above [free](https://github.com/doctrine/dbal/blob/v2.5.4/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php#L164) [their](https://github.com/doctrine/dbal/blob/v2.5.4/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php#L108) [statement resources](https://github.com/doctrine/dbal/blob/v2.5.4/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php#L154) as part of `closeCursor()`, so the statement cannot be reused after that. The rest of adapters (`mysqli`,  `sasql` and `PDO`) keep their statement resources in place during closing cursor and allow reuse after.

The fact that the statement is closed during `closeCursor()` contradicts [the purpose of the method](https://github.com/doctrine/dbal/blob/v2.5.4/lib/Doctrine/DBAL/Driver/ResultStatement.php#L30):

> Closes the cursor, enabling the statement to be executed again.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
+++ b/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
@@ -176,7 +176,13 @@ class OCI8Statement implements \IteratorAggregate, Statement
      */
     public function closeCursor()
     {
-        return oci_free_statement($this->_sth);
+        // emulate it by fetching and discarding rows, similarly to what PDO does in this case
+        // @link http://php.net/manual/en/pdostatement.closecursor.php
+        // @link https://github.com/php/php-src/blob/php-7.0.11/ext/pdo/pdo_stmt.c#L2075
+        // deliberately do not consider multiple result sets, since doctrine/dbal doesn't support them
+        while (oci_fetch($this->_sth));
+
+        return true;
     }
 
     /**
--- a/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
+++ b/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
@@ -81,6 +81,13 @@ class OCI8Statement implements \IteratorAggregate, Statement
     private $boundValues = array();
 
     /**
+     * Indicates whether the statement is in the state when fetching results is possible
+     *
+     * @var bool
+     */
+    private $result = false;
+
+    /**
      * Creates a new OCI8Statement that uses the given connection handle and SQL statement.
      *
      * @param resource                                  $dbh       The connection handle.
@@ -176,11 +183,18 @@ class OCI8Statement implements \IteratorAggregate, Statement
      */
     public function closeCursor()
     {
+        // not having the result means there's nothing to close
+        if (!$this->result) {
+            return true;
+        }
+
         // emulate it by fetching and discarding rows, similarly to what PDO does in this case
         // @link http://php.net/manual/en/pdostatement.closecursor.php
         // @link https://github.com/php/php-src/blob/php-7.0.11/ext/pdo/pdo_stmt.c#L2075
         // deliberately do not consider multiple result sets, since doctrine/dbal doesn't support them
         while (oci_fetch($this->_sth));
+
+        $this->result = false;
 
         return true;
     }
@@ -235,6 +249,8 @@ class OCI8Statement implements \IteratorAggregate, Statement
             throw OCI8Exception::fromErrorInfo($this->errorInfo());
         }
 
+        $this->result = true;
+
         return $ret;
     }
 
@@ -263,6 +279,12 @@ class OCI8Statement implements \IteratorAggregate, Statement
      */
     public function fetch($fetchMode = null)
     {
+        // do not try fetching from the statement if it's not expected to contain result
+        // in order to prevent exceptional situation
+        if (!$this->result) {
+            return false;
+        }
+
         $fetchMode = $fetchMode ?: $this->_defaultFetchMode;
 
         if (PDO::FETCH_OBJ == $fetchMode) {
@@ -310,6 +332,12 @@ class OCI8Statement implements \IteratorAggregate, Statement
                 $fetchStructure = OCI_FETCHSTATEMENT_BY_COLUMN;
             }
 
+            // do not try fetching from the statement if it's not expected to contain result
+            // in order to prevent exceptional situation
+            if (!$this->result) {
+                return array();
+            }
+
             oci_fetch_all($this->_sth, $result, 0, -1,
                 self::$fetchModeMap[$fetchMode] | OCI_RETURN_NULLS | $fetchStructure | OCI_RETURN_LOBS);
 
@@ -326,6 +354,12 @@ class OCI8Statement implements \IteratorAggregate, Statement
      */
     public function fetchColumn($columnIndex = 0)
     {
+        // do not try fetching from the statement if it's not expected to contain result
+        // in order to prevent exceptional situation
+        if (!$this->result) {
+            return false;
+        }
+
         $row = oci_fetch_array($this->_sth, OCI_NUM | OCI_RETURN_NULLS | OCI_RETURN_LOBS);
 
         if (false === $row) {
--- a/lib/Doctrine/DBAL/Driver/PDOSqlsrv/Statement.php
+++ b/lib/Doctrine/DBAL/Driver/PDOSqlsrv/Statement.php
@@ -19,36 +19,31 @@
 
 namespace Doctrine\DBAL\Driver\PDOSqlsrv;
 
-use Doctrine\DBAL\Driver\PDOConnection;
+use Doctrine\DBAL\Driver\PDOStatement;
+use PDO;
 
 /**
- * Sqlsrv Connection implementation.
- *
- * @since 2.0
+ * PDO SQL Server Statement
  */
-class Connection extends PDOConnection implements \Doctrine\DBAL\Driver\Connection
+class Statement extends PDOStatement
 {
     /**
      * {@inheritdoc}
      */
-    public function __construct($dsn, $user = null, $password = null, array $options = null)
+    public function bindParam($column, &$variable, $type = PDO::PARAM_STR, $length = null, $driverOptions = null)
     {
-        parent::__construct($dsn, $user, $password, $options);
-        $this->setAttribute(\PDO::ATTR_STATEMENT_CLASS, array(Statement::class, array()));
+        if ($type === PDO::PARAM_LOB && $driverOptions === null) {
+            $driverOptions = PDO::SQLSRV_ENCODING_BINARY;
+        }
+
+        return parent::bindParam($column, $variable, $type, $length, $driverOptions);
     }
 
     /**
-     * {@inheritDoc}
+     * {@inheritdoc}
      */
-    public function quote($value, $type=\PDO::PARAM_STR)
+    public function bindValue($param, $value, $type = PDO::PARAM_STR)
     {
-        $val = parent::quote($value, $type);
-
-        // Fix for a driver version terminating all values with null byte
-        if (strpos($val, "\0") !== false) {
-            $val = substr($val, 0, -1);
-        }
-
-        return $val;
+        return $this->bindParam($param, $value, $type);
     }
 }
--- a/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
+++ b/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
@@ -150,9 +150,13 @@ class SQLSrvStatement implements IteratorAggregate, Statement
      */
     public function closeCursor()
     {
-        if ($this->stmt) {
-            sqlsrv_free_stmt($this->stmt);
-        }
+        // emulate it by fetching and discarding rows, similarly to what PDO does in this case
+        // @link http://php.net/manual/en/pdostatement.closecursor.php
+        // @link https://github.com/php/php-src/blob/php-7.0.11/ext/pdo/pdo_stmt.c#L2075
+        // deliberately do not consider multiple result sets, since doctrine/dbal doesn't support them
+        while (sqlsrv_fetch($this->stmt));
+
+        return true;
     }
 
     /**
--- a/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
+++ b/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
@@ -53,11 +53,18 @@ class SQLSrvStatement implements IteratorAggregate, Statement
     private $stmt;
 
     /**
-     * Parameters to bind.
+     * References to the variables bound as statement parameters.
      *
      * @var array
      */
-    private $params = array();
+    private $variables = array();
+
+    /**
+     * Bound parameter types.
+     *
+     * @var array
+     */
+    private $types = array();
 
     /**
      * Translations.
@@ -145,11 +152,8 @@ class SQLSrvStatement implements IteratorAggregate, Statement
             throw new SQLSrvException("sqlsrv does not support named parameters to queries, use question mark (?) placeholders instead.");
         }
 
-        if ($type === \PDO::PARAM_LOB) {
-            $this->params[$column-1] = array($variable, SQLSRV_PARAM_IN, SQLSRV_PHPTYPE_STREAM(SQLSRV_ENC_BINARY), SQLSRV_SQLTYPE_VARBINARY('max'));
-        } else {
-            $this->params[$column-1] = $variable;
-        }
+        $this->variables[$column] =& $variable;
+        $this->types[$column] = $type;
     }
 
     /**
@@ -211,18 +215,13 @@ class SQLSrvStatement implements IteratorAggregate, Statement
             $hasZeroIndex = array_key_exists(0, $params);
             foreach ($params as $key => $val) {
                 $key = ($hasZeroIndex && is_numeric($key)) ? $key + 1 : $key;
-                $this->bindValue($key, $val);
+                $this->variables[$key] = $val;
+                $this->types[$key] = null;
             }
         }
 
         if ( ! $this->stmt) {
-            $stmt = sqlsrv_prepare($this->conn, $this->sql, $this->params);
-
-            if (!$stmt) {
-                throw SQLSrvException::fromSqlSrvErrors();
-            }
-
-            $this->stmt = $stmt;
+            $this->stmt = $this->prepare();
         }
 
         if (!sqlsrv_execute($this->stmt)) {
@@ -236,6 +235,38 @@ class SQLSrvStatement implements IteratorAggregate, Statement
         }
 
         $this->result = true;
+    }
+
+    /**
+     * Prepares SQL Server statement resource
+     *
+     * @return resource
+     * @throws SQLSrvException
+     */
+    private function prepare()
+    {
+        $params = array();
+
+        foreach ($this->variables as $column => &$variable) {
+            if ($this->types[$column] === \PDO::PARAM_LOB) {
+                $params[$column - 1] = array(
+                    &$variable,
+                    SQLSRV_PARAM_IN,
+                    SQLSRV_PHPTYPE_STREAM(SQLSRV_ENC_BINARY),
+                    SQLSRV_SQLTYPE_VARBINARY('max'),
+                );
+            } else {
+                $params[$column - 1] =& $variable;
+            }
+        }
+
+        $stmt = sqlsrv_prepare($this->conn, $this->sql, $params);
+
+        if (!$stmt) {
+            throw SQLSrvException::fromSqlSrvErrors();
+        }
+
+        return $stmt;
     }
 
     /**
--- a/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
+++ b/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
@@ -99,6 +99,13 @@ class SQLSrvStatement implements IteratorAggregate, Statement
     private $lastInsertId;
 
     /**
+     * Indicates whether the statement is in the state when fetching results is possible
+     *
+     * @var bool
+     */
+    private $result = false;
+
+    /**
      * Append to any INSERT query to retrieve the last insert id.
      *
      * @var string
@@ -150,11 +157,18 @@ class SQLSrvStatement implements IteratorAggregate, Statement
      */
     public function closeCursor()
     {
+        // not having the result means there's nothing to close
+        if (!$this->result) {
+            return true;
+        }
+
         // emulate it by fetching and discarding rows, similarly to what PDO does in this case
         // @link http://php.net/manual/en/pdostatement.closecursor.php
         // @link https://github.com/php/php-src/blob/php-7.0.11/ext/pdo/pdo_stmt.c#L2075
         // deliberately do not consider multiple result sets, since doctrine/dbal doesn't support them
         while (sqlsrv_fetch($this->stmt));
+
+        $this->result = false;
 
         return true;
     }
@@ -220,6 +234,8 @@ class SQLSrvStatement implements IteratorAggregate, Statement
             sqlsrv_fetch($this->stmt);
             $this->lastInsertId->setId(sqlsrv_get_field($this->stmt, 0));
         }
+
+        $this->result = true;
     }
 
     /**
@@ -249,6 +265,12 @@ class SQLSrvStatement implements IteratorAggregate, Statement
      */
     public function fetch($fetchMode = null)
     {
+        // do not try fetching from the statement if it's not expected to contain result
+        // in order to prevent exceptional situation
+        if (!$this->result) {
+            return false;
+        }
+
         $args      = func_get_args();
         $fetchMode = $fetchMode ?: $this->defaultFetchMode;
 
--- a/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
@@ -132,4 +132,45 @@ EOF
         );
         $this->assertSame($contents, stream_get_contents($stream));
     }
+
+    public function testIncompletelyFetchedStatementDoesNotBlockConnection()
+    {
+        $table = new Table('stmt_test_non_fetched');
+        $table->addColumn('id', 'integer');
+        $this->_conn->getSchemaManager()->createTable($table);
+        $this->_conn->insert('stmt_test_non_fetched', array('id' => 1));
+        $this->_conn->insert('stmt_test_non_fetched', array('id' => 2));
+
+        $stmt1 = $this->_conn->prepare('SELECT id FROM stmt_test_non_fetched');
+        $stmt1->execute();
+        $stmt1->fetch();
+        $stmt1->execute();
+        // fetching only one record out of two
+        $stmt1->fetch();
+
+        $stmt2 = $this->_conn->prepare('SELECT id FROM stmt_test_non_fetched WHERE id = ?');
+        $stmt2->execute(array(1));
+        $this->assertEquals(1, $stmt2->fetchColumn());
+    }
+
+    public function testReuseStatementAfterClosingCursor()
+    {
+        $table = new Table('stmt_test_close_cursor');
+        $table->addColumn('id', 'integer');
+        $this->_conn->getSchemaManager()->createTable($table);
+        $this->_conn->insert('stmt_test_close_cursor', array('id' => 1));
+        $this->_conn->insert('stmt_test_close_cursor', array('id' => 2));
+
+        $stmt = $this->_conn->prepare('SELECT id FROM stmt_test_close_cursor WHERE id = ?');
+
+        $stmt->execute(array(1));
+        $id = $stmt->fetchColumn();
+        $this->assertEquals(1, $id);
+
+        $stmt->closeCursor();
+
+        $stmt->execute(array(2));
+        $id = $stmt->fetchColumn();
+        $this->assertEquals(2, $id);
+    }
 }
--- a/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
@@ -44,15 +44,7 @@ class StatementTest extends \Doctrine\Tests\DbalFunctionalTestCase
         $stmt->execute();
         $stmt->closeCursor();
 
-        try {
-            $value = $stmt->fetchColumn();
-        } catch (\Exception $e) {
-            // some adapters trigger PHP error or throw adapter-specific exception in case of fetching
-            // from a closed cursor, which still proves that it has been closed
-            return;
-        }
-
-        $this->assertFalse($value);
+        $this->assertFalse($stmt->fetchColumn());
     }
 
     public function testReuseStatementWithLongerResults()
--- a/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
@@ -53,4 +53,37 @@ class StatementTest extends \Doctrine\Tests\DbalFunctionalTestCase
 
         $this->assertFalse($value);
     }
+
+    public function testReuseStatementWithLongerResults()
+    {
+        $sm = $this->_conn->getSchemaManager();
+        $table = new Table('stmt_test_longer_results');
+        $table->addColumn('param', 'string');
+        $table->addColumn('val', 'text');
+        $sm->createTable($table);
+
+        $row1 = array(
+            'param' => 'param1',
+            'val' => 'X',
+        );
+        $this->_conn->insert('stmt_test_longer_results', $row1);
+
+        $stmt = $this->_conn->prepare('SELECT param, val FROM stmt_test_longer_results ORDER BY param');
+        $stmt->execute();
+        $this->assertArraySubset(array(
+            array('param1', 'X'),
+        ), $stmt->fetchAll(\PDO::FETCH_NUM));
+
+        $row2 = array(
+            'param' => 'param2',
+            'val' => 'A bit longer value',
+        );
+        $this->_conn->insert('stmt_test_longer_results', $row2);
+
+        $stmt->execute();
+        $this->assertArraySubset(array(
+            array('param1', 'X'),
+            array('param2', 'A bit longer value'),
+        ), $stmt->fetchAll(\PDO::FETCH_NUM));
+    }
 }
--- a/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
@@ -2,21 +2,27 @@
 
 namespace Doctrine\Tests\DBAL\Functional;
 
+use Doctrine\DBAL\Driver\Statement;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Types\Type;
 
 class StatementTest extends \Doctrine\Tests\DbalFunctionalTestCase
 {
+    protected function setUp()
+    {
+        parent::setUp();
+
+        $table = new Table('stmt_test');
+        $table->addColumn('id', 'integer');
+        $this->_conn->getSchemaManager()->dropAndCreateTable($table);
+    }
+
     public function testStatementIsReusableAfterClosingCursor()
     {
-        $sm = $this->_conn->getSchemaManager();
-        $table = new Table('stmt_test_reusable');
-        $table->addColumn('id', 'integer');
-        $sm->createTable($table);
-        $this->_conn->insert('stmt_test_reusable', array('id' => 1));
-        $this->_conn->insert('stmt_test_reusable', array('id' => 2));
+        $this->_conn->insert('stmt_test', array('id' => 1));
+        $this->_conn->insert('stmt_test', array('id' => 2));
 
-        $stmt = $this->_conn->prepare('SELECT id FROM stmt_test_reusable ORDER BY id');
+        $stmt = $this->_conn->prepare('SELECT id FROM stmt_test ORDER BY id');
 
         $stmt->execute();
 
@@ -32,25 +38,10 @@ class StatementTest extends \Doctrine\Tests\DbalFunctionalTestCase
         $this->assertEquals(2, $id);
     }
 
-    public function testClosedCursorDoesNotContainResults()
-    {
-        $sm = $this->_conn->getSchemaManager();
-        $table = new Table('stmt_test_no_results');
-        $table->addColumn('id', 'integer');
-        $sm->createTable($table);
-        $this->_conn->insert('stmt_test_no_results', array('id' => 1));
-
-        $stmt = $this->_conn->prepare('SELECT id FROM stmt_test_no_results');
-        $stmt->execute();
-        $stmt->closeCursor();
-
-        $this->assertFalse($stmt->fetchColumn());
-    }
-
     public function testReuseStatementWithLongerResults()
     {
         $sm = $this->_conn->getSchemaManager();
-        $table = new Table('stmt_test_longer_results');
+        $table = new Table('stmt_longer_results');
         $table->addColumn('param', 'string');
         $table->addColumn('val', 'text');
         $sm->createTable($table);
@@ -59,9 +50,9 @@ class StatementTest extends \Doctrine\Tests\DbalFunctionalTestCase
             'param' => 'param1',
             'val' => 'X',
         );
-        $this->_conn->insert('stmt_test_longer_results', $row1);
+        $this->_conn->insert('stmt_longer_results', $row1);
 
-        $stmt = $this->_conn->prepare('SELECT param, val FROM stmt_test_longer_results ORDER BY param');
+        $stmt = $this->_conn->prepare('SELECT param, val FROM stmt_longer_results ORDER BY param');
         $stmt->execute();
         $this->assertArraySubset(array(
             array('param1', 'X'),
@@ -71,7 +62,7 @@ class StatementTest extends \Doctrine\Tests\DbalFunctionalTestCase
             'param' => 'param2',
             'val' => 'A bit longer value',
         );
-        $this->_conn->insert('stmt_test_longer_results', $row2);
+        $this->_conn->insert('stmt_longer_results', $row2);
 
         $stmt->execute();
         $this->assertArraySubset(array(
@@ -87,7 +78,7 @@ class StatementTest extends \Doctrine\Tests\DbalFunctionalTestCase
         $this->iniSet('memory_limit', '4G');
 
         $sm = $this->_conn->getSchemaManager();
-        $table = new Table('stmt_test_long_blob');
+        $table = new Table('stmt_long_blob');
         $table->addColumn('contents', 'blob', array(
             'length' => 0xFFFFFFFF,
         ));
@@ -110,11 +101,11 @@ d+N0hqezcjblboJ3Bj8ARJilHX4FAAA=
 EOF
     );
 
-        $this->_conn->insert('stmt_test_long_blob', array(
+        $this->_conn->insert('stmt_long_blob', array(
             'contents' => $contents,
         ), array(\PDO::PARAM_LOB));
 
-        $stmt = $this->_conn->prepare('SELECT contents FROM stmt_test_long_blob');
+        $stmt = $this->_conn->prepare('SELECT contents FROM stmt_long_blob');
         $stmt->execute();
 
         $stream = Type::getType('blob')
@@ -127,33 +118,27 @@ EOF
 
     public function testIncompletelyFetchedStatementDoesNotBlockConnection()
     {
-        $table = new Table('stmt_test_non_fetched');
-        $table->addColumn('id', 'integer');
-        $this->_conn->getSchemaManager()->createTable($table);
-        $this->_conn->insert('stmt_test_non_fetched', array('id' => 1));
-        $this->_conn->insert('stmt_test_non_fetched', array('id' => 2));
+        $this->_conn->insert('stmt_test', array('id' => 1));
+        $this->_conn->insert('stmt_test', array('id' => 2));
 
-        $stmt1 = $this->_conn->prepare('SELECT id FROM stmt_test_non_fetched');
+        $stmt1 = $this->_conn->prepare('SELECT id FROM stmt_test');
         $stmt1->execute();
         $stmt1->fetch();
         $stmt1->execute();
         // fetching only one record out of two
         $stmt1->fetch();
 
-        $stmt2 = $this->_conn->prepare('SELECT id FROM stmt_test_non_fetched WHERE id = ?');
+        $stmt2 = $this->_conn->prepare('SELECT id FROM stmt_test WHERE id = ?');
         $stmt2->execute(array(1));
         $this->assertEquals(1, $stmt2->fetchColumn());
     }
 
     public function testReuseStatementAfterClosingCursor()
     {
-        $table = new Table('stmt_test_close_cursor');
-        $table->addColumn('id', 'integer');
-        $this->_conn->getSchemaManager()->createTable($table);
-        $this->_conn->insert('stmt_test_close_cursor', array('id' => 1));
-        $this->_conn->insert('stmt_test_close_cursor', array('id' => 2));
+        $this->_conn->insert('stmt_test', array('id' => 1));
+        $this->_conn->insert('stmt_test', array('id' => 2));
 
-        $stmt = $this->_conn->prepare('SELECT id FROM stmt_test_close_cursor WHERE id = ?');
+        $stmt = $this->_conn->prepare('SELECT id FROM stmt_test WHERE id = ?');
 
         $stmt->execute(array(1));
         $id = $stmt->fetchColumn();
@@ -164,5 +149,71 @@ EOF
         $stmt->execute(array(2));
         $id = $stmt->fetchColumn();
         $this->assertEquals(2, $id);
+    }
+
+    /**
+     * @dataProvider emptyFetchProvider
+     */
+    public function testFetchFromNonExecutedStatement(callable $fetch, $expected)
+    {
+        $stmt = $this->_conn->prepare('SELECT id FROM stmt_test');
+
+        $this->assertSame($expected, $fetch($stmt));
+    }
+
+    public function testCloseCursorOnNonExecutedStatement()
+    {
+        $stmt = $this->_conn->prepare('SELECT id FROM stmt_test');
+
+        $this->assertTrue($stmt->closeCursor());
+    }
+
+    /**
+     * @dataProvider emptyFetchProvider
+     */
+    public function testFetchFromNonExecutedStatementWithClosedCursor(callable $fetch, $expected)
+    {
+        $stmt = $this->_conn->prepare('SELECT id FROM stmt_test');
+        $stmt->closeCursor();
+
+        $this->assertSame($expected, $fetch($stmt));
+    }
+
+    /**
+     * @dataProvider emptyFetchProvider
+     */
+    public function testFetchFromExecutedStatementWithClosedCursor(callable $fetch, $expected)
+    {
+        $this->_conn->insert('stmt_test', array('id' => 1));
+
+        $stmt = $this->_conn->prepare('SELECT id FROM stmt_test');
+        $stmt->execute();
+        $stmt->closeCursor();
+
+        $this->assertSame($expected, $fetch($stmt));
+    }
+
+    public static function emptyFetchProvider()
+    {
+        return array(
+            'fetch' => array(
+                function (Statement $stmt) {
+                    return $stmt->fetch();
+                },
+                false,
+            ),
+            'fetch-column' => array(
+                function (Statement $stmt) {
+                    return $stmt->fetchColumn();
+                },
+                false,
+            ),
+            'fetch-all' => array(
+                function (Statement $stmt) {
+                    return $stmt->fetchAll();
+                },
+                array(),
+            ),
+        );
     }
 }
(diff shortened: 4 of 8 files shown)


deeky666:
@morozov the patch looks legit. What is the matter with mysqli? Why is it not working? Can we get it behaving correctly somehow?

morozov:
@deeky666 the test will fail on mysqli until #2536 is resolved. Basically, mysqli driver currently cannot reuse statements for a different reason. That pull request is also waiting for your input.

morozov:
@deeky666 please see the update.

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
```diff
@@ -0,0 +1,182 @@
...
+        $id = $stmt->fetchColumn();
+        $this->assertNotEmpty($id);
+
+        $stmt->closeCursor();
+
+        try {
+            $value = $stmt->fetchColumn();
+        } catch (\Exception $e) {
+            // some adapters trigger PHP error or throw adapter-specific exception in case of fetching
+            // from a closed cursor, which still proves that it has been closed
```
deeky666: This is an inconsistency then. Maybe we should fix that too. The interface states that `false` is returned in that case. Having the drivers behave consistently finally would be great. Can you investigate? Thanks!

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
```diff
@@ -0,0 +1,182 @@
...
+        $sm->createTable($table);
+        $this->_conn->insert('stmt_test_no_results', array('id' => 1));
+        $this->_conn->insert('stmt_test_no_results', array('id' => 2));
+
+        $stmt = $this->_conn->prepare('SELECT id FROM stmt_test_no_results');
+
+        $stmt->execute();
+
+        $id = $stmt->fetchColumn();
+        $this->assertNotEmpty($id);
```
deeky666: I don't think we need to assert here again as it already is covered by the test before. So simply executing and immediately closing should be enough.

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    public function testClosedCursorDoesNotContainResults()
+    {
+        $sm = $this->_conn->getSchemaManager();
+        $table = new Table('stmt_test_no_results');
+        $table->addColumn('id', 'integer');
+        $sm->createTable($table);
+        $this->_conn->insert('stmt_test_no_results', array('id' => 1));
+        $this->_conn->insert('stmt_test_no_results', array('id' => 2));
```
deeky666: We only need one row, so please remove this one.

deeky666 (changes requested):

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
morozov: The current state is:

1. PDO returns `false`,
2. mysqli throws `Doctrine\DBAL\Driver\Mysqli\MysqliException` "Commands out of sync; you can't run this command now",
3. oci8 returns `false`,
4. sqlsrv returns `false`,
5. ibm_db2 triggers warning "db2_fetch_array(): Fetch Failure".

deeky666: @morozov it is indeed an exceptional case but we should stick to our interfaces as those are the contracts we provide. In the `3.0` we will rework the interfaces to throw meaningful exceptions as returning `false` is not a good behaviour in exceptional cases. But for now we should stick to returning `false` I think. @Ocramius thoughts?

morozov: @deeky666 if I understand correctly, there's no clear expectation of how cursors should behave after being closed given the current implementation. The problem with returning `false` is that it's hard to identify why exactly fetch failed on the application end. "Fetch Failure" and "Commands out of sync;" are too generic and may mean some other exception which we don't want to silently convert to `false`.

Another question is, do we really have to fix consistency as part of this ticket. It's already big enough and judging by the number resolves ~5 different issues (the number of test cases).

deeky666: @morozov you are completely right. I feel bad about returning `false` too but unfortunately we have made up that contract when designing the API.

I'm also aware that this PR is growing big already but in the end it is about fixing the close cursor behaviour across all the drivers. Unfortunately it's quite a rabbit hole but it feels wrong to leave some parts buggy when trying to fix a general problem (which this test case exposes). I am ok with a rather big change here if we get it fixed consistently.

Also relying on exceptions from some drivers but not from others in this scenario is not an issue for us because it is undefined behaviour (as not being defined in the interface).

morozov: @deeky666 I see your point. How should we implement handling these warnings/exceptions? Just do string/error code comparison? Or maybe manually track calls to `closeCursor()` in the drivers in question and return `false` without calling the actual `fetch()` to meet the contract.

deeky666: @morozov We could use flag `$result` for those problematic drivers, which is set in `execute` as soon as we have a result and have `fetch()` methods check for that flag and return `false` early. `closeCursor` will then reset the flag. Does that make sense or am I missing something here?

morozov: @deeky666 I was thinking about the opposite: having an `$isCursorClosed` flag which is set in `closeCursor()`, reset in `execute()` and is checked before calling the underlying `fetch()` implementation. Here, we only want to manually handle the case when the cursor is closed and then fetched. Not the cases when for example the statement is constructed but not yet executed (which would also be covered with your approach). In general, both are about the same. What would you suggest?

deeky666: @morozov I don't have a strong opinion about either of both options. I am just concerned about what the actual behaviour should be when fetching without having executed the query before. Because IMO the whole API defined in `Doctrine\DBAL\Driver\ResultStatement` is in theory only available after executing the prepared statement. I think of it as if one would return a `Result` object (which is in fact a cursor) form the `execute()` method. So to say an encapsulation of the result cursor. I think APIs like JDBC have similar constucts. So in this case the `$result` flag would make more sense to me as it is explicitly set when having a result rather than having as `$isCursorClosed` which does not have a meaning as long as we the prepared statment is not executed. I think I saw something like the `$result` flag in some driver already...

morozov:
@deeky666 please see the update.

morozov:
@deeky666, ping

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
```diff
@@ -150,9 +150,13 @@ public function bindParam($column, &$variable, $type = null, $length = null)
      */
     public function closeCursor()
     {
-        if ($this->stmt) {
```
deeky666: Also reusing prepared statements does currently not work as the binding by reference does not work. We need to modify `bindParam()` to this:

```php
public function bindParam($column, &$variable, $type = null, $length = null)
{
    if (!is_numeric($column)) {
        throw new SQLSrvException("sqlsrv does not support named parameters to queries, use question mark (?) placeholders instead.");
    }

    $this->values[$column] = $variable;

    if ($type === \PDO::PARAM_LOB) {
        $this->params[$column-1] = array(&$this->values[$column], SQLSRV_PARAM_IN, SQLSRV_PHPTYPE_STREAM(SQLSRV_ENC_BINARY), SQLSRV_SQLTYPE_VARBINARY('max'));
    } else {
        $this->params[$column-1] =& $this->values[$column];
    }
}
```
Also introduces a new private property `$values` that keeps track of parameter references. Otherwise using different parameters won't work and the driver will always use those from the first binding.

deeky666 (changes requested):
We also need to change `sasql_stmt_free_result` to `sasql_stmt_reset` in `SQLAnywhereStatement::closeCursor()` as reusing the prepared statement won't work otherwise.

Furthermore I would suggest that we introduce more tests for edge case scenarios:

- `fetch*()`
- `closeCursor()`
- `closeCursor()` -> `fetch*()`
- `execute()` -> `closeCursor()` -> `fetch*()`

What do you think? Anything else that comes to mind?

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
morozov: @deeky666 is it about the case like:
```php
public function testReuseStatementWithParameterBoundByReference()
{
    $this->_conn->insert('stmt_test', array('id' => 1));
    $this->_conn->insert('stmt_test', array('id' => 2));

    $stmt = $this->_conn->prepare('SELECT id FROM stmt_test WHERE id = ?');
    $stmt->bindParam(1, $id);

    $id = 1;
    $stmt->execute();
    $this->assertEquals(1, $stmt->fetchColumn());

    $id = 2;
    $stmt->execute();
    $this->assertEquals(2, $stmt->fetchColumn());
}
```

Then it should be ` $this->values[$column] =& $variable`, same as in `MysqliStatement::bindParam()`. In any case, DB2 is unhappy about this case because the DBAL does both `db2_bind_param()` and passes the values of `$this->_bindParam` to `db2_execute()`:
```
db2_execute(): Binding Error 2
```

morozov:
@deeky666 thank you for the detailed review.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -176,7 +176,13 @@ public function bindParam($column, &$variable, $type = null, $length = null)
      */
     public function closeCursor()
     {
-        return oci_free_statement($this->_sth);
+        // emulate it by fetching and discarding rows, similarly to what PDO does in this case
+        // @link http://php.net/manual/en/pdostatement.closecursor.php
+        // @link https://github.com/php/php-src/blob/php-7.0.11/ext/pdo/pdo_stmt.c#L2075
+        // deliberately do not consider multiple result sets, since doctrine/dbal doesn't support them
+        while (oci_fetch($this->_sth));
```
morozov: Done.

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
morozov: Done.

deeky666: @morozov basically it is about that yes. But the test case I used was this:

```
1) Doctrine\Tests\DBAL\Functional\StatementTest::testReuseStatementAfterClosingCursor
Failed asserting that 1 matches expected 2.

/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:151
```

But also we are talking about SQL Server here, not DB2.

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
```diff
@@ -0,0 +1,219 @@
...
+    {
+        $this->_conn->insert('stmt_test', array('id' => 1));
+        $this->_conn->insert('stmt_test', array('id' => 2));
+
+        $stmt1 = $this->_conn->prepare('SELECT id FROM stmt_test');
+        $stmt1->execute();
+        $stmt1->fetch();
+        $stmt1->execute();
+        // fetching only one record out of two
+        $stmt1->fetch();
```
deeky666: Do you need to execute and fetch twice for this test case? I think this is not intented and could have side effects that potentially let the test fail even though the error is unrelated to the test case.

deeky666:
@morozov we are getting closer :) Awesome test cases btw!

deeky666 (changes requested):

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
morozov: It's needed to reproduce the case described in #2487. Currently, there's different execution flow for the first and subsequent executions of a statement in `MysqliStatement`. So it's more of a regression test.

deeky666: Ok I see.

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
morozov: @deeky666 with your fix `testReuseStatementAfterClosingCursor()` passes, thank you. Not sure if this is relevant to this issue (I wish it wasn't), but `testReuseStatementWithParameterBoundByReference()` above still doesn't pass on SQL Server and IBM DB2. It's made up and I don't know if it should pass in the first place (however it does on other DBs).

The difference between the two is that the first binds param by value, and the second binds them by reference.

morozov: UPD: I've modified your solution to make `testReuseStatementWithParameterBoundByReference()` also pass.

deeky666:
Unfortunately there are still some errors.

**pdo_mysql + libmysql**
```
There was 1 error:

1) Doctrine\Tests\DBAL\Functional\StatementTest::testIncompletelyFetchedStatementDoesNotBlockConnection
Exception: [Doctrine\DBAL\Exception\DriverException] An exception occurred while executing 'SELECT id FROM stmt_test WHERE id = ?':

SQLSTATE[HY000]: General error: 2014 Cannot execute queries while other unbuffered queries are active.  Consider using PDOStatement::fetchAll().  Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute.

With queries:
6. SQL: 'SELECT id FROM stmt_test' Params: 
5. SQL: 'SELECT id FROM stmt_test' Params: 
4. SQL: 'INSERT INTO stmt_test (id) VALUES (?)' Params: '2'
3. SQL: 'INSERT INTO stmt_test (id) VALUES (?)' Params: '1'
2. SQL: 'CREATE TABLE stmt_test (id INT NOT NULL) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB' Params: 
1. SQL: 'DROP TABLE stmt_test' Params: 

Trace:
/php/srv/lib/Doctrine/DBAL/DBALException.php:154
/php/srv/lib/Doctrine/DBAL/DBALException.php:128
/php/srv/lib/Doctrine/DBAL/Connection.php:800
/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:131
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:1121
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:972
/php/srv/vendor/phpunit/phpunit/src/Framework/TestResult.php:709
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:927
/php/srv/vendor/phpunit/phpunit/src/Framework/TestSuite.php:728
/php/srv/vendor/phpunit/phpunit/src/Framework/TestSuite.php:728
/php/srv/vendor/phpunit/phpunit/src/TextUI/TestRunner.php:491
/php/srv/vendor/phpunit/phpunit/src/TextUI/Command.php:188
/php/srv/vendor/phpunit/phpunit/src/TextUI/Command.php:118
/php/srv/vendor/phpunit/phpunit/phpunit:52


/php/srv/tests/Doctrine/Tests/DbalFunctionalTestCase.php:81

Caused by
Doctrine\DBAL\Exception\DriverException: An exception occurred while executing 'SELECT id FROM stmt_test WHERE id = ?':

SQLSTATE[HY000]: General error: 2014 Cannot execute queries while other unbuffered queries are active.  Consider using PDOStatement::fetchAll().  Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute.

/php/srv/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php:120
/php/srv/lib/Doctrine/DBAL/DBALException.php:154
/php/srv/lib/Doctrine/DBAL/DBALException.php:128
/php/srv/lib/Doctrine/DBAL/Connection.php:800
/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:131

Caused by
Doctrine\DBAL\Driver\PDOException: SQLSTATE[HY000]: General error: 2014 Cannot execute queries while other unbuffered queries are active.  Consider using PDOStatement::fetchAll().  Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute.

/php/srv/lib/Doctrine/DBAL/Driver/PDOConnection.php:79
/php/srv/lib/Doctrine/DBAL/Statement.php:86
/php/srv/lib/Doctrine/DBAL/Connection.php:798
/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:131

Caused by
PDOException: SQLSTATE[HY000]: General error: 2014 Cannot execute queries while other unbuffered queries are active.  Consider using PDOStatement::fetchAll().  Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute.

/php/srv/lib/Doctrine/DBAL/Driver/PDOConnection.php:77
/php/srv/lib/Doctrine/DBAL/Statement.php:86
/php/srv/lib/Doctrine/DBAL/Connection.php:798
/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:131

--
```

**pdo_sqlsrv** (12, 13, 14, 15 should be fixed by Microsoft in the driver, I'll file an issue soon)
```
11) Doctrine\Tests\DBAL\Functional\StatementTest::testFetchLongBlob
Exception: [Doctrine\DBAL\DBALException] An exception occurred while executing 'INSERT INTO stmt_long_blob (contents) VALUES (?)' with params ["\x1f\x8b\x08\x08\x94\x40\x09\x58\x02\x03\x64\x6f\x63\x74\x72\x69\x6e\x65\x2e\x69\x63\x6f\x00\xed\x54\xdb\x4b\x14\x71\x14\xfe\x26\xb5\x8b\xd7\xd5\xcd\x55\xb1\xbc\x94\xab\x66\x6a\x17\x4b\xaa\x5d\xd3\x32\xbb\x42\x17\x4d\x73\x4d\xf3\x96\xba\x69\xa6\xad\x69\x5a\xa2\x42\x14\x44\x59\x91\x41\x44\x0f\x26\xd8\x05\x0a\x89\x84\x1e\xa2\x50\xea\xa1\x87\xec\x21\x82\x5e\x2a\x90\x5e\x8c\x7c\x68\x7e\xb3\xce\xce\xec\x65\x66\x3b\xa3\xa5\x05\xf5\x1f\x74\x86\x6f\xce\x7c\x67\x7e\x9c\xf3\x9d\x6f\x60\x00\x8e\x2e\x9d\x0e\x74\x5f\x08\xab\x1f\x60\x00\x90\x4c\xa0\x12\xe2\x30\x53\x9f\x0e\x0e\x73\xc1\xcd\x71\xaf\xd7\x8b\x04\xab\x80\x6b\xc3\x32\x8a\x2f\x8b\x18\xfb\xe4\x86\x4f\x3e\x8f\xa1\xd7\x2e\xb4\xde\x71\x20\xfb\xec\x14\x26\xbe\x2b\xd0\x97\x31\xf4\xbf\x70\xe2\xf9\x3b\x17\xac\x37\x45\x9c\x7b\x28\x21\xa3\x45\x80\x5d\x52\xf0\x71\x42\x41\x74\x35\x43\x6a\xa3\x80\x8a\x3e\x11\xaf\x3e\xb8\xd1\x79\x57\x42\x60\x31\x43\x5e\x8f\x1d\x4e\xb7\x8a\xa7\x6f\x5d\xb8\x38\x24\xe1\xcb\xa4\x82\xc1\x51\x27\x4c\xed\x76\x44\x54\x30\xdc\x7e\xe6\x44\x6c\x8d\x80\x49\xa6\x20\xb9\x41\x80\xae\x84\xa1\xf7\xb1\x8c\x9e\x07\x12\x78\x51\x45\x7a\xb3\x1d\x23\xef\xdd\xc8\x6c\xb5\xa3\xe1\x96\x03\x1e\x45\xc5\xf0\x1b\x17\x0a\x2f\x89\xa8\xee\x73\xe0\xc2\x23\x09\xfe\x45\x3c\x06\x46\x64\x48\x4e\x15\xf1\x75\x02\x72\xbb\xec\xf8\xfc\xd5\x03\x63\x3d\x83\x28\x2b\x08\x2e\xe1\x11\x5a\xca\x10\x7e\x84\x21\xaa\x92\xe1\xca\x13\x19\x5d\xf7\x24\x9c\x19\x94\xd0\xd2\xef\x40\x0a\xe9\x4e\x6d\x12\xb0\x8e\x66\xa8\xaa\x8a\xb8\x5a\x06\x73\xc7\x14\x72\x3a\xed\xa8\xba\x21\xa2\x6d\x40\x82\xdb\xa3\x42\x76\xa9\x98\x92\x54\xf8\x16\xf0\x08\xb2\x30\x84\x95\xf2\x88\xa9\x61\x30\x1e\x13\x90\x58\x2f\x20\xad\xc9\x8e\xb4\x66\x01\xe7\xc9\x9b\xee\xfb\x12\x2a\xaf\x8b\xe4\xaf\x0a\xaf\xea\x25\xa8\x50\xc9\x6b\x95\xb8\x4a\x5c\x21\xae\xd0\x2e\x1e\xca\x6e\xca\x2e\xad\xbf\x5b\xeb\xaf\xe9\x65\xd0\x1d\xd6\xfa\x33\x7c\x23\x5f\xc2\xca\x78\xe8\xe9\xd9\x40\xfa\x23\x49\x7f\x54\x15\xc3\x72\xfa\x66\x09\x34\x37\x89\x3c\x7b\x49\x7e\xa7\x91\xfe\x74\x9a\x3d\x3c\xe6\xc2\x55\xda\xaf\x97\x50\xd5\xa7\x69\x77\xe0\x7f\xfc\x8f\x7f\x86\x2d\x31\xb8\x23\xd0\x36\x4b\xdb\xe6\xe9\xb7\xed\x6b\x5c\xf9\xab\x90\x5d\x9b\x15\xc4\x45\x58\x0f\x98\x67\x0a\x29\x71\x4d\xa6\x18\x8e\xcb\xb1\x2e\x5b\x31\x5d\x48\xad\xcd\xdd\xa1\xb3\x71\x96\x1c\xb3\x39\x50\xe3\x3b\x17\x2f\x2d\xdf\x18\x9a\xb7\x87\xdb\xb2\x77\x8d\xc6\x63\x5b\xfc\xf3\x8e\xd6\x94\x57\x73\xbb\xe7\x17\x69\xdc\xd7\x74\x6a\xd1\xc1\x56\x8b\x25\xe4\xb8\x79\xb3\xc6\xe3\xad\xcd\x46\x1f\x0b\xb7\xc4\x98\xaf\xf7\x9b\x1e\x10\x65\xde\x6e\xe0\x4e\xd7\x25\x6c\xfa\xf9\xd7\xb2\x65\xae\x2d\x5c\x75\x22\x3f\x20\x7c\x56\x6e\xfc\xd6\x86\x93\xfb\x77\xcd\xe9\x8f\xde\x50\x50\xdf\x3e\x47\x0d\x59\xeb\x4d\xc5\xbf\xad\x97\x11\x19\xb0\xfa\x8f\x75\x93\x0e\xcd\x2e\x2b\xc7\x02\x7c\x08\x30\xbe\x00\x18\xf5\xf9\x3b\xb4\x77\xe3\x74\x86\xa7\xb3\x72\x36\xe5\x6e\x82\x77\x06\x3f\x00\x44\x98\xa5\x1d\x7e\x05\x00\x00"]:

SQLSTATE[IMSSP]: An error occurred translating string for input param 1 to UCS-2: Error code 0x54

With queries:
4. SQL: 'INSERT INTO stmt_long_blob (contents) VALUES (?)' Params: '�@	Xdoctrine.ico�T�Kq�&�����U����fjK�]�2�BMsM���i��iZ�BDY�AD&�
���Pꡇ�!�^*�^�|h~����ef;���t�o�|g~���o`�t�`�L��0S�s��q�׋��k�2�/����O>�.��q ��&�+З1��p��;�7E�{(!�E�]R�qBAt5Cj���>�>��yWB`1C^N���o]�8$�ˤ��Q'L�vDT0�~�Dl��I� �A��������xQEz�#���l����E�
/���s��#	�E<FdHN�ur�����c=�(.�Z�~�!����]�$�����@
�Nm�����Zs�r:���!�m@�ۣBv���T��0�����a0�X/ �Ɏ�f�ɛ��*��
��%�P�k��J\!��.�n�.��[���e���3|#_��x���@�#IT�r�f	47�<{I~���t�=<��Uگ�Pէiw���-1�#�6K�����k\���]��EX�g
)qM��.[1]H��ݡ�q��9P�;/-����۲w��c[���Ws��i��tj��V�%��y�����F
Y�Mſ������.+�0���;�w�t���r6�n�w?D��~'                      �Ę���e�n�N�%l��ײe�-\u"? |Vn�ֆ��w��PP�>G
3. SQL: 'CREATE TABLE stmt_long_blob (contents VARBINARY(MAX) NOT NULL)' Params: 
2. SQL: 'CREATE TABLE stmt_test (id INT NOT NULL)' Params: 
1. SQL: 'DROP TABLE stmt_test' Params: 

Trace:
/php/srv/lib/Doctrine/DBAL/DBALException.php:128
/php/srv/lib/Doctrine/DBAL/Connection.php:1001
/php/srv/lib/Doctrine/DBAL/Connection.php:711
/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:106
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:1121
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:972
/php/srv/vendor/phpunit/phpunit/src/Framework/TestResult.php:709
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:927
/php/srv/vendor/phpunit/phpunit/src/Framework/TestSuite.php:728
/php/srv/vendor/phpunit/phpunit/src/Framework/TestSuite.php:728
/php/srv/vendor/phpunit/phpunit/src/TextUI/TestRunner.php:491
/php/srv/vendor/phpunit/phpunit/src/TextUI/Command.php:188
/php/srv/vendor/phpunit/phpunit/src/TextUI/Command.php:118
/php/srv/vendor/phpunit/phpunit/phpunit:52


/php/srv/tests/Doctrine/Tests/DbalFunctionalTestCase.php:81

Caused by
Doctrine\DBAL\DBALException: An exception occurred while executing 'INSERT INTO stmt_long_blob (contents) VALUES (?)' with params ["\x1f\x8b\x08\x08\x94\x40\x09\x58\x02\x03\x64\x6f\x63\x74\x72\x69\x6e\x65\x2e\x69\x63\x6f\x00\xed\x54\xdb\x4b\x14\x71\x14\xfe\x26\xb5\x8b\xd7\xd5\xcd\x55\xb1\xbc\x94\xab\x66\x6a\x17\x4b\xaa\x5d\xd3\x32\xbb\x42\x17\x4d\x73\x4d\xf3\x96\xba\x69\xa6\xad\x69\x5a\xa2\x42\x14\x44\x59\x91\x41\x44\x0f\x26\xd8\x05\x0a\x89\x84\x1e\xa2\x50\xea\xa1\x87\xec\x21\x82\x5e\x2a\x90\x5e\x8c\x7c\x68\x7e\xb3\xce\xce\xec\x65\x66\x3b\xa3\xa5\x05\xf5\x1f\x74\x86\x6f\xce\x7c\x67\x7e\x9c\xf3\x9d\x6f\x60\x00\x8e\x2e\x9d\x0e\x74\x5f\x08\xab\x1f\x60\x00\x90\x4c\xa0\x12\xe2\x30\x53\x9f\x0e\x0e\x73\xc1\xcd\x71\xaf\xd7\x8b\x04\xab\x80\x6b\xc3\x32\x8a\x2f\x8b\x18\xfb\xe4\x86\x4f\x3e\x8f\xa1\xd7\x2e\xb4\xde\x71\x20\xfb\xec\x14\x26\xbe\x2b\xd0\x97\x31\xf4\xbf\x70\xe2\xf9\x3b\x17\xac\x37\x45\x9c\x7b\x28\x21\xa3\x45\x80\x5d\x52\xf0\x71\x42\x41\x74\x35\x43\x6a\xa3\x80\x8a\x3e\x11\xaf\x3e\xb8\xd1\x79\x57\x42\x60\x31\x43\x5e\x8f\x1d\x4e\xb7\x8a\xa7\x6f\x5d\xb8\x38\x24\xe1\xcb\xa4\x82\xc1\x51\x27\x4c\xed\x76\x44\x54\x30\xdc\x7e\xe6\x44\x6c\x8d\x80\x49\xa6\x20\xb9\x41\x80\xae\x84\xa1\xf7\xb1\x8c\x9e\x07\x12\x78\x51\x45\x7a\xb3\x1d\x23\xef\xdd\xc8\x6c\xb5\xa3\xe1\x96\x03\x1e\x45\xc5\xf0\x1b\x17\x0a\x2f\x89\xa8\xee\x73\xe0\xc2\x23\x09\xfe\x45\x3c\x06\x46\x64\x48\x4e\x15\xf1\x75\x02\x72\xbb\xec\xf8\xfc\xd5\x03\x63\x3d\x83\x28\x2b\x08\x2e\xe1\x11\x5a\xca\x10\x7e\x84\x21\xaa\x92\xe1\xca\x13\x19\x5d\xf7\x24\x9c\x19\x94\xd0\xd2\xef\x40\x0a\xe9\x4e\x6d\x12\xb0\x8e\x66\xa8\xaa\x8a\xb8\x5a\x06\x73\xc7\x14\x72\x3a\xed\xa8\xba\x21\xa2\x6d\x40\x82\xdb\xa3\x42\x76\xa9\x98\x92\x54\xf8\x16\xf0\x08\xb2\x30\x84\x95\xf2\x88\xa9\x61\x30\x1e\x13\x90\x58\x2f\x20\xad\xc9\x8e\xb4\x66\x01\xe7\xc9\x9b\xee\xfb\x12\x2a\xaf\x8b\xe4\xaf\x0a\xaf\xea\x25\xa8\x50\xc9\x6b\x95\xb8\x4a\x5c\x21\xae\xd0\x2e\x1e\xca\x6e\xca\x2e\xad\xbf\x5b\xeb\xaf\xe9\x65\xd0\x1d\xd6\xfa\x33\x7c\x23\x5f\xc2\xca\x78\xe8\xe9\xd9\x40\xfa\x23\x49\x7f\x54\x15\xc3\x72\xfa\x66\x09\x34\x37\x89\x3c\x7b\x49\x7e\xa7\x91\xfe\x74\x9a\x3d\x3c\xe6\xc2\x55\xda\xaf\x97\x50\xd5\xa7\x69\x77\xe0\x7f\xfc\x8f\x7f\x86\x2d\x31\xb8\x23\xd0\x36\x4b\xdb\xe6\xe9\xb7\xed\x6b\x5c\xf9\xab\x90\x5d\x9b\x15\xc4\x45\x58\x0f\x98\x67\x0a\x29\x71\x4d\xa6\x18\x8e\xcb\xb1\x2e\x5b\x31\x5d\x48\xad\xcd\xdd\xa1\xb3\x71\x96\x1c\xb3\x39\x50\xe3\x3b\x17\x2f\x2d\xdf\x18\x9a\xb7\x87\xdb\xb2\x77\x8d\xc6\x63\x5b\xfc\xf3\x8e\xd6\x94\x57\x73\xbb\xe7\x17\x69\xdc\xd7\x74\x6a\xd1\xc1\x56\x8b\x25\xe4\xb8\x79\xb3\xc6\xe3\xad\xcd\x46\x1f\x0b\xb7\xc4\x98\xaf\xf7\x9b\x1e\x10\x65\xde\x6e\xe0\x4e\xd7\x25\x6c\xfa\xf9\xd7\xb2\x65\xae\x2d\x5c\x75\x22\x3f\x20\x7c\x56\x6e\xfc\xd6\x86\x93\xfb\x77\xcd\xe9\x8f\xde\x50\x50\xdf\x3e\x47\x0d\x59\xeb\x4d\xc5\xbf\xad\x97\x11\x19\xb0\xfa\x8f\x75\x93\x0e\xcd\x2e\x2b\xc7\x02\x7c\x08\x30\xbe\x00\x18\xf5\xf9\x3b\xb4\x77\xe3\x74\x86\xa7\xb3\x72\x36\xe5\x6e\x82\x77\x06\x3f\x00\x44\x98\xa5\x1d\x7e\x05\x00\x00"]:

SQLSTATE[IMSSP]: An error occurred translating string for input param 1 to UCS-2: Error code 0x54

/php/srv/lib/Doctrine/DBAL/DBALException.php:157
/php/srv/lib/Doctrine/DBAL/DBALException.php:128
/php/srv/lib/Doctrine/DBAL/Connection.php:1001
/php/srv/lib/Doctrine/DBAL/Connection.php:711
/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:106

Caused by
Doctrine\DBAL\Driver\PDOException: SQLSTATE[IMSSP]: An error occurred translating string for input param 1 to UCS-2: Error code 0x54

/php/srv/lib/Doctrine/DBAL/Driver/PDOStatement.php:93
/php/srv/lib/Doctrine/DBAL/Connection.php:992
/php/srv/lib/Doctrine/DBAL/Connection.php:711
/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:106

Caused by
PDOException: SQLSTATE[IMSSP]: An error occurred translating string for input param 1 to UCS-2: Error code 0x54

/php/srv/lib/Doctrine/DBAL/Driver/PDOStatement.php:91
/php/srv/lib/Doctrine/DBAL/Connection.php:992
/php/srv/lib/Doctrine/DBAL/Connection.php:711
/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:106

12) Doctrine\Tests\DBAL\Functional\StatementTest::testCloseCursorOnNonExecutedStatement
Exception: [PDOException] SQLSTATE[IMSSP]: The statement must be executed before results can be retrieved.

With queries:
2. SQL: 'CREATE TABLE stmt_test (id INT NOT NULL)' Params: 
1. SQL: 'DROP TABLE stmt_test' Params: 

Trace:
/php/srv/lib/Doctrine/DBAL/Statement.php:197
/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:185
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:1121
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:972
/php/srv/vendor/phpunit/phpunit/src/Framework/TestResult.php:709
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:927
/php/srv/vendor/phpunit/phpunit/src/Framework/TestSuite.php:728
/php/srv/vendor/phpunit/phpunit/src/Framework/TestSuite.php:728
/php/srv/vendor/phpunit/phpunit/src/TextUI/TestRunner.php:491
/php/srv/vendor/phpunit/phpunit/src/TextUI/Command.php:188
/php/srv/vendor/phpunit/phpunit/src/TextUI/Command.php:118
/php/srv/vendor/phpunit/phpunit/phpunit:52


/php/srv/tests/Doctrine/Tests/DbalFunctionalTestCase.php:81

Caused by
PDOException: SQLSTATE[IMSSP]: The statement must be executed before results can be retrieved.

/php/srv/lib/Doctrine/DBAL/Statement.php:197
/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:185

13) Doctrine\Tests\DBAL\Functional\StatementTest::testFetchFromNonExecutedStatementWithClosedCursor with data set "fetch" (Closure Object (...), false)
Exception: [PDOException] SQLSTATE[IMSSP]: The statement must be executed before results can be retrieved.

With queries:
2. SQL: 'CREATE TABLE stmt_test (id INT NOT NULL)' Params: 
1. SQL: 'DROP TABLE stmt_test' Params: 

Trace:
/php/srv/lib/Doctrine/DBAL/Statement.php:197
/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:194
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:1121
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:972
/php/srv/vendor/phpunit/phpunit/src/Framework/TestResult.php:709
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:927
/php/srv/vendor/phpunit/phpunit/src/Framework/TestSuite.php:728
/php/srv/vendor/phpunit/phpunit/src/Framework/TestSuite.php:728
/php/srv/vendor/phpunit/phpunit/src/Framework/TestSuite.php:728
/php/srv/vendor/phpunit/phpunit/src/TextUI/TestRunner.php:491
/php/srv/vendor/phpunit/phpunit/src/TextUI/Command.php:188
/php/srv/vendor/phpunit/phpunit/src/TextUI/Command.php:118
/php/srv/vendor/phpunit/phpunit/phpunit:52


/php/srv/tests/Doctrine/Tests/DbalFunctionalTestCase.php:81

Caused by
PDOException: SQLSTATE[IMSSP]: The statement must be executed before results can be retrieved.

/php/srv/lib/Doctrine/DBAL/Statement.php:197
/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:194

14) Doctrine\Tests\DBAL\Functional\StatementTest::testFetchFromNonExecutedStatementWithClosedCursor with data set "fetch-column" (Closure Object (...), false)
Exception: [PDOException] SQLSTATE[IMSSP]: The statement must be executed before results can be retrieved.

With queries:
2. SQL: 'CREATE TABLE stmt_test (id INT NOT NULL)' Params: 
1. SQL: 'DROP TABLE stmt_test' Params: 

Trace:
/php/srv/lib/Doctrine/DBAL/Statement.php:197
/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:194
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:1121
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:972
/php/srv/vendor/phpunit/phpunit/src/Framework/TestResult.php:709
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:927
/php/srv/vendor/phpunit/phpunit/src/Framework/TestSuite.php:728
/php/srv/vendor/phpunit/phpunit/src/Framework/TestSuite.php:728
/php/srv/vendor/phpunit/phpunit/src/Framework/TestSuite.php:728
/php/srv/vendor/phpunit/phpunit/src/TextUI/TestRunner.php:491
/php/srv/vendor/phpunit/phpunit/src/TextUI/Command.php:188
/php/srv/vendor/phpunit/phpunit/src/TextUI/Command.php:118
/php/srv/vendor/phpunit/phpunit/phpunit:52


/php/srv/tests/Doctrine/Tests/DbalFunctionalTestCase.php:81

Caused by
PDOException: SQLSTATE[IMSSP]: The statement must be executed before results can be retrieved.

/php/srv/lib/Doctrine/DBAL/Statement.php:197
/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:194

15) Doctrine\Tests\DBAL\Functional\StatementTest::testFetchFromNonExecutedStatementWithClosedCursor with data set "fetch-all" (Closure Object (...), array())
Exception: [PDOException] SQLSTATE[IMSSP]: The statement must be executed before results can be retrieved.

With queries:
2. SQL: 'CREATE TABLE stmt_test (id INT NOT NULL)' Params: 
1. SQL: 'DROP TABLE stmt_test' Params: 

Trace:
/php/srv/lib/Doctrine/DBAL/Statement.php:197
/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:194
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:1121
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:972
/php/srv/vendor/phpunit/phpunit/src/Framework/TestResult.php:709
/php/srv/vendor/phpunit/phpunit/src/Framework/TestCase.php:927
/php/srv/vendor/phpunit/phpunit/src/Framework/TestSuite.php:728
/php/srv/vendor/phpunit/phpunit/src/Framework/TestSuite.php:728
/php/srv/vendor/phpunit/phpunit/src/Framework/TestSuite.php:728
/php/srv/vendor/phpunit/phpunit/src/TextUI/TestRunner.php:491
/php/srv/vendor/phpunit/phpunit/src/TextUI/Command.php:188
/php/srv/vendor/phpunit/phpunit/src/TextUI/Command.php:118
/php/srv/vendor/phpunit/phpunit/phpunit:52


/php/srv/tests/Doctrine/Tests/DbalFunctionalTestCase.php:81

Caused by
PDOException: SQLSTATE[IMSSP]: The statement must be executed before results can be retrieved.

/php/srv/lib/Doctrine/DBAL/Statement.php:197
/php/srv/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php:194
```

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
deeky666: @morozov I now figured out the issue with DB2 binding and it is quite nasty -.-. The binding error seems to be related due to `$this->_bindParam` having referenced variable values so the fix that worked for me is to change the portion in `execute()` to:

```php
        if ($params === null) {
            ksort($this->_bindParam);

            $params = array();

            foreach ($this->_bindParam as $column => $value) {
                $params[] = $value;
            }
        }
```

This fixes the binding error but in fact we cannot get `db_bind_param()` to work at all because it is not actually binding immediately at call time. Instead when calling `db2_execute()`, the function looks up variables by the names passed to `db2_bind_param` lazily. And even worse, it expects thos variables to be declared in global scope (WTF). In theory we could create global variables on binding with `global` keyword but that does not seem to work properly either as I am getting `db2_execute(): Binding Error 1` then. So I would suggest changing the code as shown above, because real parameter binding does not work atm anyways.

morozov:
@deeky666 could you provide some more details on reproducing the failure on **pdo_mysql + libmysql**? I know how I would fix it but I don't want to send the patch before seeing it actually fixes the problem.

Here's my environment:
```
↪  php --version
PHP 5.6.30 (cli) (built: Feb  1 2017 11:06:37) 
Copyright (c) 1997-2016 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2016 Zend Technologies
    with Xdebug v2.5.0, Copyright (c) 2002-2016, by Derick Rethans

↪  php -r 'echo mysqli_get_client_info(), PHP_EOL;'
5.7.17

↪  grep 'name="db_' pdo-mysql.phpunit.xml
        <var name="db_type" value="pdo_mysql" />
        <var name="db_host" value="localhost" />
        <var name="db_username" value="root" />
        <var name="db_password" value="" />
        <var name="db_name" value="doctrine" />
        <var name="db_port" value="3306" />

↪  phpunit -c pdo-mysql.phpunit.xml tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
PHPUnit 5.5.4 by Sebastian Bergmann and contributors.

................                                                                  16 / 16 (100%)

Time: 1.82 seconds, Memory: 6.75MB

OK (16 tests, 21 assertions)
```

deeky666:
@morozov I have built some docker images that I use for testing which might be officially published in the near future.

https://hub.docker.com/r/deeky666/mysql/
https://hub.docker.com/r/deeky666/php/

(Dockerfiles are here on GitHub in my repositories)

Then you can do this:

```
$ cd /path/to/doctrine/dbal
$ docker run --rm -d --name mysql -p 3306:3306 deeky666/mysql
$ docker run --rm -it --name doctrine --link mysql -v $(pwd):/php/srv deeky666/php -d extension=pdo.so -d extension=pdo_sqlite.so -d extension=pdo_mysql-libmysql.so vendor/bin/phpunit -c mysql.phpunit.xml
```

You will need the PHPUnit configuration `mysql.phpunit.xml` in your DBAL folder:

```xml
<?xml version="1.0" encoding="utf-8"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.8/phpunit.xsd"
         backupGlobals="false"
         colors="true"
         bootstrap="vendor/autoload.php"
         stopOnFailure="false"
>
  <php>
    <ini name="error_reporting" value="-1" />
    <var name="db_type" value="pdo_mysql"/>
    <var name="db_host" value="mysql" />
    <var name="db_username" value="root" />
    <var name="db_password" value="" />
    <var name="db_name" value="doctrine" />
    <var name="db_port" value=""/>

    <var name="tmpdb_type" value="pdo_mysql"/>
    <var name="tmpdb_host" value="mysql" />
    <var name="tmpdb_username" value="root" />
    <var name="tmpdb_password" value="" />
    <var name="tmpdb_port" value=""/>
  </php>

    <testsuites>
        <testsuite name="Doctrine DBAL Test Suite">
            <directory>./tests/Doctrine/Tests/DBAL</directory>
        </testsuite>
    </testsuites>
</phpunit>
```

deeky666:
I filed an issue for the `pdo_sqlsrv` `closeCursor()` issue: https://github.com/Microsoft/msphpsql/issues/267

morozov:
@deeky666 please see the update. A few key moments:

1. Fixing the failure on **pdo_mysql + libmysql** using `PDO::MYSQL_ATTR_USE_BUFFERED_QUERY` as I originally intended didn't work despite what the error message suggested:

   >  Cannot execute queries while other unbuffered queries are active.  Consider using PDOStatement::fetchAll().  Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute.

   On the other hand, it was only reproducible during the full suite run, but not in the isolated test case. Apparently, it depended on whether `PDOConnectionTest::testThrowsWrappedExceptionOnPrepare()` was executed prior to the failing test as the former modified shared connection attributes. Resetting the connection fixed the failure.
2. It's currently impossible to store/retrieve blobs as binary strings using the DBAL API with **pdo_sqlsrv**. See the details in the commit message. The test is fixed with a partial workaround.
3. The failure on **ibm_db2** is fixed as suggested.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Statement.php
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * PDO SQL Server Statement
+ */
+class Statement extends PDOStatement
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function bindParam($column, &$variable, $type = PDO::PARAM_STR, $length = null, $driverOptions = null)
```
deeky666: > 1. Binding binary string variable as a blob parameter with pdo_sqlsrv requires using `PDOStatement::bindParam()`, but `Connection::executeUpdate()` uses `PDOStatement::bindValue()`, so it's impossible to `INSERT`/`UPDATE` using `Connection::insert()`/`Connection::update()` in a portable way.

Can't you also override `bindValue()` to use `bindParam()` internally? Like this:

```php
public function bindValue($param, $value, $type = \PDO::PARAM_STR)
{
    return $this->bindParam($param, $value, $type);
}
```

That might fix the issue.

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
```diff
@@ -0,0 +1,242 @@
...
+        $stmt = $this->_conn->prepare('SELECT contents FROM stmt_long_blob');
+        $stmt->execute();
+
+        $value = $stmt->fetchColumn();
+
+        // this is a temporary workaround to make the test pass on pdo_sqlsrv as the Statement interface
+        // currently doesn't allow binding variables to the result columns which is required to specify blob encoding
+        // @see https://msdn.microsoft.com/en-us/library/ff754357(v=sql.105).aspx#Anchor_3
+        if ($this->_conn->getDriver()->getName() === 'pdo_sqlsrv') {
+            $value = hex2bin($value);
```
deeky666: I would suggest marking the test as incomplete here because it could be fixed in the future by either Microsoft or us (if we can figure out how to). Can you mark the test as incomplete here with a reference to the issue I filed for that?

https://github.com/Microsoft/msphpsql/issues/270

morozov: To your question:
> I am not quite sure if this is a bug or I just cannot figure out how to do this.

It can be done by using:
```php
$stmt->bindColumn(1, $image, PDO::PARAM_LOB, 0, PDO::SQLSRV_ENCODING_BINARY);  
$stmt->fetch(PDO::FETCH_BOUND);
```
instead of `$stmt->fetchColumn()`, but I agree it's a bug from the standpoint of consistency with `pdo_*` and `sqlsrv`.

lib/Doctrine/DBAL/Driver/Statement.php
```diff
@@ -19,6 +19,8 @@
 
 namespace Doctrine\DBAL\Driver;
 
+use Doctrine\DBAL\DBALException;
```
deeky666: You have to revert all of the changes in this file. We cannot change the interface as of `2.x`. The point of this PR is to make the code comply with the current contract.

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
morozov: > Can you mark the test as incomplete

I believe it needs to be marked as skipped for pdo_sqlsrv, not incomplete. Right?

lib/Doctrine/DBAL/Driver/Statement.php
morozov: You're saying that even if some of the methods throw exceptions, the contract implies they shouldn't? I thought the opposite.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Statement.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        if ($type === PDO::PARAM_LOB && $driverOptions === null) {
+            $driverOptions = PDO::SQLSRV_ENCODING_BINARY;
```
deeky666: @meet-bhagdev I wonder if that could be the default in the driver itself? Why should the driver use any other encoding than `PDO::SQLSRV_ENCODING_BINARY` when binding `PDO::PARAM_LOB` parameters?

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
deeky666: @morozov I know you can use `bindColumn()` but it seems wrong that this is required because there simply is no reason why it shouldn't work without that. I somewhat studied the driver source code today and it seems the driver is introspecting the column type on fetching anyways and evaluating the appropriate types but it simply does not seem to work. Also applying the situation to `sqlsrv` it works out of the box, too. There is no need for binding anything there. Considering that both drivers share a large part of the codebase, it should be possible for `pdo_sqlsrv` to achieve the desired result.

If we markthe test as skipped, it will get lost. I suppose we will have a solution for this in the future (either by driver bugfix or workaround) and I would like to be able to track some kind of "todo list" to check through once in a while. We did similar things already. I agree it is not the best solution but PHPUnit doesn't offer anything else for this use case. What do you think?

morozov: One more thought: if we skip the test for **pdo_sqlsrv**, we'll be not testing `PDOSqlsrv\Statement` (the `INSERT` part). Should we keep the changes untested, or we write another `INSERT`-only test (how then we know the data is properly inserted?), or we leave `hexbin()` to have at least the `INSERT` logic covered by the test?

lib/Doctrine/DBAL/Driver/Statement.php
deeky666: Basically we adopted the PDO interface back then. PDO heavily relies on returning `false` in case of an error and does not throw exceptions under normal circumstances. The problem is that our implementations are not completely respecting this fact and the only way to go IMO is to try to adhere to the contract. All of those thrown exceptions are bugs so to say. Also it's weird to have a contract that says "in case of an error I return `false` but I could also throw an exception every now and then" :D

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
deeky666: @morozov I think you can simply mark the test incomplete at this exact line. So `INSERT` is still tested because we know that should work but abort any further testing at this point. Using `hexbin()` to make the tests pass is wrong because then we are modifying assumptions.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Statement.php
morozov: Done.

Ocramius:
Random comment, but it has to be said: @morozov thanks for the patience, impressive work!

morozov:
@Ocramius my pleasure. I want to say you guys the same but I'll wait until this one gets merged :-)

deeky666 (approved):
simply: AWESOME!

deeky666:
LGTM! I'll let have @Ocramius have a second look on that and then we can merge. :tada:

morozov:
@deeky666 does it make sense to introduce a **libmysql** label as it's also an environment attribute (often not taken into account) which the DBAL behavior may depend on?

deeky666:
@morozov good point. Done. Added `libmysql` and `mysqlnd` labels.

Ocramius (approved):

Ocramius:
Awesome work! Some minor CS problems, but I'm going to close an eye and fix them post-merge.

Thanks @morozov and @deeky666!

Ocramius:
Backported to `2.5` in doctrine/dbal@b713ba7480905875aca74903c9e7cd2f0e38b490

PR #2549, opened 2016-11-03
Oracle column metadata import fix

ghost:
Added a check for table owner equivalence in the OraclePlatform class that will prevent metadata imports failing if two schemas (owners) have two tables with the exact same structure (name and columns).

doctrine error


deeky666:
@francoisdev this PR looks borked as it has a lot of unrelated changes in it. Also this is a duplicate of #2585

deeky666:
FIxed in #2518

Issue #2538, opened 2016-10-16
Diff on column with collation problem


MassiveHiggsField:
Copied from https://github.com/doctrine/migrations/issues/495

Used versions:

- os: Debian GNU/Linux 8 
- mysql: Ver 14.14 Distrib 5.7.8-rc, for Linux (x86_64) using  EditLine **wrapper**
- doctrine/orm v2.5.4 
- doctrine/migrations 1.4.1 

Entity:

    use Doctrine\ORM\Mapping as ORM;
    /**
     * @ORM\Table(name="Contest")
     * @ORM\Entity
     */
    class Contest
    {
        /**
         * @var string
         *
         * @ORM\Column(name="id", type="string", length=32, nullable=false, options={"collation":"utf8_unicode_ci"})
         * @ORM\Id
         * @ORM\GeneratedValue(strategy="IDENTITY")
         */
        private $id;
    }

Current table:

    CREATE TABLE `Contest` (
      `id` int(11) NOT NULL AUTO_INCREMENT
    ) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci 

Migration generated from diff:

    class Version20161014011453 extends AbstractMigration
    {
        /**
         * @param Schema $schema
         */
        public function up(Schema $schema)
        {
            // this up() migration is auto-generated, please modify it to your needs
            $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
    
            $this->addSql('ALTER TABLE Contest CHANGE id id VARCHAR(32) NOT NULL COLLATE utf8_unicode_ci');
        }
    
        /**
         * @param Schema $schema
         */
        public function down(Schema $schema)
        {
            // this down() migration is auto-generated, please modify it to your needs
            $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
    
            $this->addSql('ALTER TABLE Contest CHANGE id id VARCHAR(32) NOT NULL COLLATE utf8_unicode_ci');
        }
    }

So up() and down() do the same thing and if i execute this migration and rerun diff, it will generate the same migration again. 

Any help appreciated.

Ocramius:
@mikeSimonson can you check this?

mikeSimonson:
I will look at it

mikeSimonson:
[...]

PR #2551, opened 2016-11-06
Implement the column collation for Mysql

mikeSimonson:
Related to #2538

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
@@ -1429,4 +1429,18 @@ abstract class AbstractPlatformTestCase extends \Doctrine\Tests\DbalTestCase
             array(array('precision' => 8, 'scale' => 2), 'DOUBLE PRECISION'),
         );
     }
+
+    public function testFixGithub2538Collation()
+    {
+        $field = [
+            'charset' => 'utf8',
+            'collation' => 'utf8_unicode-ci',
+            'type' => Type::getType(Type::STRING),
+            'length' => 32,
+        ];
+        $this->assertEquals(
+            'foo VARCHAR(32)  DEFAULT NULL COLLATE utf8_unicode-ci',
+            $this->_platform->getColumnDeclarationSQL('foo', $field)
+        );
+    }
 }
--- a/tests/Doctrine/Tests/DBAL/Schema/SchemaTest.php
+++ b/tests/Doctrine/Tests/DBAL/Schema/SchemaTest.php
@@ -2,9 +2,11 @@
 
 namespace Doctrine\Tests\DBAL\Schema;
 
+use Doctrine\DBAL\Platforms\MySqlPlatform;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Schema\Sequence;
 use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\Types\Type;
 
 class SchemaTest extends \PHPUnit_Framework_TestCase
 {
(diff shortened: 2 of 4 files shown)


Ocramius:
@mikeSimonson should this fix (since it fixes #2538, if I understand it correctly) come with a test case then?

tristanbes:
Any updates on this @mikeSimonson ? I can confirm that this fix should fix my issue about Doctrine Migrations generating a diff without the collate information set on `Table` option

` * @ORM\Table(name="fos_user", options={"collate"="utf8mb4_unicode_ci", "charset"="utf8mb4"}) `

When I put a new field and I generate the diff it outputs
`$this->addSql('ALTER TABLE fos_user ADD lessNoise TINYINT(1) NOT NULL');`

resulting in the field created in `utf8_unicode` and `utf8_unicode_ci` for collation instead of the settings set on the table.

mikeSimonson:
@Ocramius Do you mean that the test should be removed or that it needs a higher level test case ?

Ocramius:
@mikeSimonson higher level test case reflecting #2538

mikeSimonson:
@Ocramius What is the right approach here for the test ?

tests/Doctrine/Tests/DBAL/Schema/SchemaTest.php
```diff
@@ -449,4 +451,48 @@ public function testVisitsNamespaceVisitor()
...
+            ['CREATE TABLE public.foo (test VARCHAR(255) NOT NULL COLLATE utf8_unicode_ci) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB'],
+            $schema->toSql($platform),
+            $platform->getColumnDeclarationSQL('test')
+        );
+    }
+
+    /**
+     * https://github.com/doctrine/dbal/issues/2538
+     */
+    public function testFixGithub2538_autoincrement()
```
Ocramius: This test seems unrelated to the changes, or am I missing something?

tests/Doctrine/Tests/DBAL/Schema/SchemaTest.php
```diff
@@ -449,4 +451,48 @@ public function testVisitsNamespaceVisitor()
...
+    {
+        $this->markTestSkipped();
+
+        $tableName = "public.foo";
```
Ocramius: Single quotes

tests/Doctrine/Tests/DBAL/Schema/SchemaTest.php
Ocramius: Please use a method name describing the issue being covered. Also, use `@group #2538` for labeling it

tests/Doctrine/Tests/DBAL/Schema/SchemaTest.php
```diff
@@ -449,4 +451,48 @@ public function testVisitsNamespaceVisitor()
 
         $this->assertNull($schema->visit($visitor));
     }
+
+    /**
+     * https://github.com/doctrine/dbal/issues/2538
+     */
+    public function testFixGithub2538Collation()
```
Ocramius: Please use a method name describing the issue being covered. Also, use `@group #2538` for labeling it

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
```diff
@@ -1429,4 +1429,18 @@ public function getGeneratesFloatDeclarationSQL()
             array(array('precision' => 8, 'scale' => 2), 'DOUBLE PRECISION'),
         );
     }
+
+    public function testFixGithub2538Collation()
```
Ocramius: Please use a method name describing the issue being covered. Also, use `@group #2538` for labeling it

Ocramius (changes requested):
One last change probably needed: verifying that the comparator will pick up the collation (unless that is already covered)

yukoff:
What is the progress with this one? And maybe there should be column charset implemented ass well?

Tobion:
I don't see the point of this PR. Column collations are working in mysql already in general.

`@ORM\Column(type="string", options={"collation"="ascii_general_ci"})` will correctly generate a mysql column like `myfield VARCHAR(255) NOT NULL COLLATE ascii_general_ci`

The requested change seems useless. MysqlPlatform already implements 

```
public function supportsColumnCollation()
{
        return true;
}
```
and the rest is handled by the default implementation of AbstractPlatform.

There are two bugs with column collations:
1. schema update / migrations doesn't recognize collation changes, see #2401 similar issues
2. When you have a column collation overwritten for an enity primary key that is used as forgeign key in another entity, then the collation is MISSING in the foreign key definition. This makes the generated schema invalid because the collations can be different which is not allowed in foreign key constraints. This is what is stumbled upon. I'll create an issue for this.

Ocramius:
> Column collations are working in mysql already in general.

Tests seem to say otherwise, but feel free to provide more tests or comment on the ones provided in this PR.

PR #2552, opened 2016-11-07
Update Type.php

eyepin:


Ocramius:
While this change is correct, we can't apply it without breaking BC compliance at the moment, sorry.

The reason is mostly the subclasses extending `Type`, which are to be found also outside of `doctrine/dbal`.

PR #2553, opened 2016-11-16
MS-SQL namespace

vinise:
On MS-SQL full name is not simply "database.table" but "database.dbo.table" so namespace should be full text till last dot

### DIFF
--- a/lib/Doctrine/DBAL/Schema/AbstractAsset.php
+++ b/lib/Doctrine/DBAL/Schema/AbstractAsset.php
@@ -64,9 +64,9 @@ abstract class AbstractAsset
             $name = $this->trimQuotes($name);
         }
         if (strpos($name, ".") !== false) {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Schema/AbstractAsset.php
```diff
@@ -64,9 +64,9 @@ protected function _setName($name)
             $name = $this->trimQuotes($name);
         }
         if (strpos($name, ".") !== false) {
```
phansys: It could be simplified as:
```php
if (false !== $pos = strrpos($name, '.')) {
    $this->_namespace = substr($name, 0, $pos);
    $name = substr($name, $pos + 1);
}
```
saving one call to `strpos()`.

phansys (changes requested):
Improvement suggested.

PR #2561, opened 2016-11-27
unit tests for oracle type consistency.

brettmc:
add unit tests that try to generate a table with a column type for every
available doctrine data type. inject the data type information from the
platform's CreateTableSQL back through the schema manager to confirm that
the types are consistent. Any types that are not implemented by the platform
are marked as skipped.

This only tests Oracle types for now, but I've implemented an
AbstractSchemaManagerTest as I think that the principle can be applied to
other platforms as well.


deeky666:
Closing in favour of #2612 which fixes the original issue.

PR #2570, opened 2016-12-02
Allow secure connections using SSL on mysqli

pgrau:
The current driver has no option to allow the usage of SSL connections on mysql. I've used the same name params as mysql client. 
An example of mysql client usage would be: 

mysql --ssl-ca='path_to_ca.pem' --ssl-cert='path_to_client_cert.pem' --ssl-key='path_to_client_key.pem'

### DIFF
--- a/lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
+++ b/lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
@@ -63,6 +63,7 @@ class MysqliConnection implements Connection, PingableConnection, ServerInfoAwar
 
         $this->_conn = mysqli_init();
 
+        $this->setSecureConnection($params);
         $this->setDriverOptions($driverOptions);
 
         set_error_handler(function () {});
@@ -262,5 +263,36 @@ class MysqliConnection implements Connection, PingableConnection, ServerInfoAwar
     public function ping()
     {
         return $this->_conn->ping();
+    }
+
+    /**
+     * Establish a secure connection
+     *
+     * @param array $params
+     * @throws MysqliException
+     */
+    private function setSecureConnection(array $params)
+    {
+        if (! isset($params['ssl_key']) &&
+            ! isset($params['ssl_cert']) &&
+            ! isset($params['ssl_ca']) &&
+            ! isset($params['ssl_capath']) &&
+            ! isset($params['ssl_cipher'])
+        ) {
+            return;
+        }
+
+        if (! isset($params['ssl_key']) || ! isset($params['ssl_cert'])) {
+            $msg = '"ssl_key" and "ssl_cert" parameters are mandatory when using secure connection parameters.';
+            throw new MysqliException($msg);
+        }
+
+        $this->_conn->ssl_set(
+            $params['ssl_key'],
+            $params['ssl_cert'],
+            $params['ssl_ca']     ?? null,
+            $params['ssl_capath'] ?? null,
+            $params['ssl_cipher'] ?? null
+        );
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Driver/Mysqli/MysqliConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Driver/Mysqli/MysqliConnectionTest.php
@@ -2,6 +2,8 @@
 
 namespace Doctrine\Tests\DBAL\Driver\Mysqli;
 
+use Doctrine\DBAL\Driver\Mysqli\MysqliConnection;
+use Doctrine\DBAL\Driver\Mysqli\MysqliException;
 use Doctrine\Tests\DbalTestCase;
 
 class MysqliConnectionTest extends DbalTestCase
@@ -21,7 +23,7 @@ class MysqliConnectionTest extends DbalTestCase
 
         parent::setUp();
 
-        $this->connectionMock = $this->getMockBuilder('Doctrine\DBAL\Driver\Mysqli\MysqliConnection')
+        $this->connectionMock = $this->getMockBuilder(MysqliConnection::class)
             ->disableOriginalConstructor()
             ->getMockForAbstractClass();
     }
@@ -30,4 +32,35 @@ class MysqliConnectionTest extends DbalTestCase
     {
         $this->assertFalse($this->connectionMock->requiresQueryForServerVersion());
     }
+
+    /**
+     * @dataProvider secureMissingParamsProvider
+     */
+    public function testThrowsExceptionWhenMissingMandatorySecureParams(array $secureParams)
+    {
+        $this->expectException(MysqliException::class);
+        $msg = '"ssl_key" and "ssl_cert" parameters are mandatory when using secure connection parameters.';
+        $this->expectExceptionMessage($msg);
+
+        new MysqliConnection($secureParams, 'xxx', 'xxx');
+    }
+
+    public function secureMissingParamsProvider()
+    {
+        return [
+            [
+                ['ssl_cert' => 'cert.pem']
+            ],
+            [
+                ['ssl_key' => 'key.pem']
+            ],
+            [
+                ['ssl_key' => 'key.pem', 'ssl_ca' => 'ca.pem', 'ssl_capath' => 'xxx', 'ssl_cipher' => 'xxx']
+            ],
+            [
+                ['ssl_ca' => 'ca.pem', 'ssl_capath' => 'xxx', 'ssl_cipher' => 'xxx']
+            ]
+        ];
+    }
 }
+
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -235,6 +235,11 @@ mysqli
    the database.
 -  ``charset`` (string): The charset used when connecting to the
    database.
+-  ``ssl_key`` (string): The path name to the key file to use for SSL encryption.
+-  ``ssl_cert`` (string): The path name to the certificate file to use for SSL encryption.
+-  ``ssl_ca`` (string): The path name to the certificate authority file to use for SSL encryption.
+-  ``ssl_capath`` (string): The pathname to a directory that contains trusted SSL CA certificates in PEM format.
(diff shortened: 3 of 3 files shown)


lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
```diff
@@ -63,6 +63,19 @@ public function __construct(array $params, $username, $password, array $driverOp
 
         $this->_conn = mysqli_init();
 
+        if (isset($driverOptions['ssl_key']) &&
+            isset($driverOptions['ssl_cert']) &&
+            isset($driverOptions['ssl_ca'])
```
deeky666: Can you tell which of those 5 options are required at minimum to use this feature? I would like to have those evaluated and a `MysqliException` being thrown if one of the required options is missing, so that misconfiguration is exposed to the user. Unfortunately the PHP docs only state `Any unused SSL parameters may be given as NULL`. It doesn't give a word about which are definitely required.

Also we need test cases for all of the possibilities.

deeky666: Also please make them connection parameters instead of driver options. We have SSL related connection parameters in other drivers too. Those are more appropriate for connection parameters.

Then finally please document them in our docs (also which are required).

deeky666 (changes requested):

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
pgrau: As you say unfortunately PHP and MySQL docs are not clear respect about what minimum of  parameters are mandatory about secure connections.

I think the common use is with key, cert and ca as mandatory parameter, but I've talk with various colleagues and we think is possible stablish a secure connection with only key and cert.

In conclusion mandatory fields must be key and cert.
@deeky666 Are you agree?

deeky666: @pgrau agreed. MySQL documentation also states that the `ca` is only mandatory if also used on the server side so we have to make it optional. Making `ssl_key` and `ssl_cert` required is a good choice then.

pgrau: @deeky666 I think connection ssl parameters must be driver options as PDO library. Can you  put an example of driver with ssl parameters instead driver options?

deeky666: @pgrau simply put them into `$params` argument instead of `$driverOptions`. `$driverOptions` is exclusively used for `\MYSQLI_*` constants so the SSL related parameters do not fit in there.

**Example**

```php
$connection = \Doctrine\DBAL\DriverManager::getConnection(
    array(
        'dbname' => 'mydb',
        'user' => 'user',
        'password' => 'secret',
        'host' => 'localhost',
        'driver' => 'mysqli',
        'ssl_key' => '/path/to/ssl_key',
        'ssl_cert' => '/path/to/ssl_key',
        'ssl_ca' => '/path/to/ssl_key',
        // ...
    )
);

Comment (pgrau) [inline lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php]: If we put ssl parameters as connection parameters against driver options will not run ok automatically in some php frameworks as Symfony.

Example:
 [Symfony\Component\Config\Definition\Exception\InvalidConfigurationException]           
  Unrecognized options "ssl_key, ssl_cert, ssl_ca" under "doctrine.dbal.connections.bd1"

Are you sure do you prefer ssl parameters as connection parameters against driver options?

Comment (deeky666) [inline lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php]: The reason it does not work with Symfony is because the available connection parameters are defined in the DoctrineBundle. Changes would have to be done in the bundle, too then to allow using those parameters. So yes new connection parameters are still preferred over driver options.

Comment (pgrau) [inline lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php]: Ok! Now, it's time to develop it : )

**Comment by pgrau** [comment]
@deeky666 I've done some changes. Can you review, please?

**Review by deeky666 — CHANGES_REQUESTED** [review]
This also requires documentation of the newly supported connection parameters for `mysqli`. Please see [here](https://github.com/doctrine/dbal/blob/master/docs/en/reference/configuration.rst#mysqli)

File: tests/Doctrine/Tests/DBAL/Driver/Mysqli/MysqliConnectionTest.php [hunk 2]
```diff
@@ -30,4 +32,34 @@ public function testDoesNotRequireQueryForServerVersion()
     {
         $this->assertFalse($this->connectionMock->requiresQueryForServerVersion());
     }
+
+    /**
+     * @dataProvider secureMissingParamsProvider
+     */
+    public function testItShouldReturnAnExceptionWhenMissingMandatorySecureParams(array $secureParams)
```
pgrau: done

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
```diff
@@ -263,4 +264,35 @@ public function ping()
...
+    /**
+     * @param array $params
+     * @return bool
+     * @throws MysqliException
+     */
+    private function areSecureConnectionParams(array $params)
+    {
+        $matches = preg_grep('/^(ssl_)+(key|cert|ca|capath|cipher)$/', array_keys($params));
+        if ($matches && (false === array_search('ssl_key', $matches) || false === array_search('ssl_cert', $matches))) {
+            throw new MysqliException('ssl_key and ssl_cert are mandatory for secure connections');
```
pgrau: done

pgrau: done

pgrau:
@deeky666 You can see the doc in https://github.com/doctrine/dbal/pull/2570/commits/8788fd463b14c1fde09323778d41eff882dc8f54

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
```diff
@@ -263,4 +264,35 @@ public function ping()
...
+            ! isset($params['ssl_ca']) &&
+            ! isset($params['ssl_capath']) &&
+            ! isset($params['ssl_cipher'])
+        ) {
+            return;
+        }
+
+        if (! isset($params['ssl_key']) || ! isset($params['ssl_cert'])) {
+            $msg = "\"ssl_key\" and \"ssl_cert\" parameters are mandatory when using secure connection parameters.";
+            throw new MysqliException($msg);
```
deeky666: Please don't do unnecessary string escaping. You can easily avoid that by using single quotes:

```php
throw new MysqliException(
    '"ssl_key" and "ssl_cert" parameters are mandatory when using secure connection parameters.'
);
```

tests/Doctrine/Tests/DBAL/Driver/Mysqli/MysqliConnectionTest.php
```diff
@@ -30,4 +32,35 @@ public function testDoesNotRequireQueryForServerVersion()
...
         $this->assertFalse($this->connectionMock->requiresQueryForServerVersion());
     }
+
+    /**
+     * @dataProvider secureMissingParamsProvider
+     */
+    public function testThrowsExceptionWhenMissingMandatorySecureParams(array $secureParams)
+    {
+        $this->expectException(MysqliException::class);
+        $msg = "\"ssl_key\" and \"ssl_cert\" parameters are mandatory when using secure connection parameters.";
```
deeky666: See above.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
```diff
@@ -263,4 +264,35 @@ public function ping()
     {
         return $this->_conn->ping();
     }
+
+    /**
+     * Stablish a secure connection
```
deeky666: Typo: `Establish`

deeky666 (changes requested):
Nearly done. Just some last nitpicks.

pgrau (commented):
@deeky666 It's all ok?

Do you want I move php5 code to php7
Example:
```
php5:
       $port = isset($params['port']) ? $params['port'] : ini_get('mysqli.default_port');

        // Fallback to default MySQL port if not given.
        if ( ! $port) {
            $port = 3306;
        }

        $socket = isset($params['unix_socket']) ? $params['unix_socket'] : ini_get('mysqli.default_socket');
        $dbname = isset($params['dbname']) ? $params['dbname'] : null;

        $flags = isset($driverOptions[static::OPTION_FLAGS]) ? $driverOptions[static::OPTION_FLAGS] : null;

php7:
     
        $port = $params['port'] ?? ini_get('mysqli.default_port') ?? 3306;
        $dbname = $params['dbname'] ?? null;
        $flags = $driverOptions[static::OPTION_FLAGS] ?? null;

```

deeky666:
@pgrau no please don't convert any code to PHP7 that could potentially also be backported to DBAL `2.5` (we are still supporting PHP 5.3 there). Doing so might cause a real hassle when backporting code.

deeky666:
@pgrau the patch looks good to me now. Can you please squash your commits? I'll merge afterwards. :+1:

pgrau:
@deeky666 I squash all commits. Can you re-build  https://travis-ci.org/doctrine/dbal/builds/197293269, please? Sometimes this job fail https://travis-ci.org/doctrine/dbal/jobs/197293304

deeky666:
@pgrau seems you did some kind of merge instead of a squash. There are still 15 commits. You need to:

```
git rebase -i origin/master
git push -f origin patch-1
```

While rebasing you can select the commits to squash.

pgrau:
@deeky666 please, see now.

deeky666 (approved):

deeky666:
@pgrau thanks! :)

PR #2571, opened 2016-12-06
Add support for PDO::FETCH_OBJ in result cache

hason:


Ocramius:
This is not going to be accepted, sorry.

`FETCH_OBJ` is generally a bad idea (along with most usages of `stdClass`).

Just do an `array_map(function (array $row) { return (object) $row; }, $query->fetchAll())` if you need it.

hason:
@Ocramius Please reopen.

> FETCH_OBJ is generally a bad idea
Yes, but this behaviour is necessary for some legacy applications.

This PR solves incosistence between documentation of `(https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Driver/ResultStatement.php#L61-L63)[ResultStatement]` and cache implementation:

> @param integer|null $fetchMode Controls how the next row will be returned to the caller.
> The value must be one of the PDO::FETCH_* constants, defaulting to PDO::FETCH_BOTH.

```php
$result = $connection->executeQuery($query);
$result->fetchAll(\PDO::FETCH_OBJ); // OK

$result = $connection->executeQuery($query, [], [], new QueryCacheProfile(30, $query));
$result->fetchAll(\PDO::FETCH_OBJ); // throws exception !
```

Ocramius:
I'd rather fix the documentation by stating that `\PDO::FETCH_OBJ` and `\PDO::FETCH_CLASS` will not work

Issue #2360, opened 2016-04-15
Comparator does not detect renamed foreign keys


yakobe:
The origin of this problem is here: https://github.com/doctrine/doctrine2/issues/4783
The comparator does not detect a change for names of foreign keys. It works fine for indexes, but foreign keys are just ignored.

A test for this change might look something like this (to match the index test):
https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php#L628
``` PHP
public function testCompareForeignKeyBasedOnPropertiesNotName()
    {
        $tableA = new Table("foo");
        $tableA->addColumn('id', 'integer');
        $tableA->addNamedForeignKeyConstraint('foo_constraint', 'bar', array('id'), array('id'));

        $tableB = new Table("foo");
        $tableB->addColumn('ID', 'integer');
        $tableB->addNamedForeignKeyConstraint('bar_constraint', 'bar', array('id'), array('id'));

        $c = new Comparator();
        $tableDiff = $c->diffTable($tableA, $tableB);

        $tableDiff = new TableDiff('foo');
        $tableDiff->fromTable = $tableA;
        $tableDiff->renamedForeignKeys['foo_constraint'] = new ForeignKeyConstraint(array('id'),'bar', array('id'), 'bar_constraint');

        $this->assertEquals(
            $tableDiff,
            $c->diffTable($tableA, $tableB)
        );
    }
```

senaria:
Is there any news about this? This possible solution looks like it would really help us out. Would be great if you could have a closer look.

yakobe:
Is it possible to get some feedback about this issue? Is it a confirmed bug? Is there any possibility of it being resolved? Can I or other members of the community help in any way or is it not a desired change?
It would be really useful to get some sort of status / opinion here.

Ocramius:
As far as I know, this would currently just drop and recreate a FK. Not sure if worth fixing atm.

yakobe:
It causes a problem because it does not detect if a foreign key has been renamed. This means that it does not show up when creating automatic migrations, so the key keeps the old name after migrations. 

If something more major then changes and want to drop this key, the new name is used in the generated migration. This new name is then not recognised when running the migration in production and everything crashes. **Hard.**

Normal indexes are handled correctly, but these are not. It's casuing us all manor of problems when preparing migrations. 😞

Ocramius:
> when running the migration in production and everything crashes. Hard.

[...]

PR #2573, opened 2016-12-08
DDC-3924 Comparator can detect a renamed foreign key

senaria:
Q | A
------ | ------
|Branch? 	| "master"
|Bug fix? 	| yes
|BC breaks? 	| no
|Deprecations?  |no
|Tests pass? 	|yes
|Referres to | #2360

As it appears, the Comparator was not able to detect changed foreign key names. Now it is...

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Comparator.php
+++ b/lib/Doctrine/DBAL/Schema/Comparator.php
@@ -408,6 +408,10 @@ class Comparator
             return true;
         }
 
+        if ($key1->getName() !== $key2->getName()){
+            return true;
+        }
+
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Schema/Comparator.php
```diff
@@ -408,6 +408,10 @@ public function diffForeignKey(ForeignKeyConstraint $key1, ForeignKeyConstraint
             return true;
         }
 
+        if ($key1->getName() !== $key2->getName()){
```
phansys: Why not simply `return $key1->getName() !== $key2->getName()`;`?

phansys (changes requested):
Status: Needs work.

Issue #2578, opened 2016-12-14
DateIntervalType (negative support)


galeaspablo:
*DateIntervalType does not support negative intervals*(http://php.net/manual/en/class.dateinterval.php#dateinterval.props.invert) :cry:

Is this an oversight? 
It's not in a stable version, so it can still be changed.

galeaspablo:
Pr is ready at #2579

galeaspablo:
@Ocramius Thanks for your feedback in #2579 

How is this a BC break if the type isn't there in 2.5?

Ocramius:
Urgh, re-opening @galeaspablo. Thanks for making us notice this fast :-)

PR #2579, opened 2016-12-14
DateIntervalType (negative support) resolves doctrine/dbal#2578

galeaspablo:


galeaspablo:
If #2316 is merged before this, we'll need to refactor this PR.

lib/Doctrine/DBAL/Types/DateIntervalType.php
```diff
@@ -56,9 +58,13 @@ public function convertToPHPValue($value, AbstractPlatform $platform)
         }
 
         try {
-            return new \DateInterval($value);
+            $interval = new \DateInterval(substr($value, 1));
```
lcobucci: With this you assume that `$value` always have the sign, which is not the case for existing data.
We should make sure that the type would still work for the "old" format.

lcobucci (changes requested):

lib/Doctrine/DBAL/Types/DateIntervalType.php
lcobucci: > We should make sure that the type would still work for the "old" format.

Testing it too :wink:

galeaspablo: I don't think `\DateInterval` accepts a leading `+` or `-`.
How about?
```php
try {
            $firstCharacter = substr($value, 0, 1);
            if ($firstCharacter !== '+' && $firstCharacter !== '-') {
                  return new \DateInterval($value);
            }
            $interval = new \DateInterval(substr($value, 1));
            if ($firstCharacter === '-') {
                 $interval->invert = 1;
             }
             return $interval;
}
```

deeky666: My 2 cents on this. As already pointed out, this is a BC break that cause schema changes and format changes, which we cannot accept for a minor release. We could however introduce a new type like `SignedDateIntervalType` to keep BC.

galeaspablo: Is this really a BC break? This isn't on 2.5.

deeky666: @galeaspablo oh you are right I thought it was part of 2.5 already. Thanks for pointing that out, then we won't have a problem with BC indeeed :)

foaly-nr1:
Any update on this? Would you go ahead with the merge with the solution from https://github.com/doctrine/dbal/pull/2579#discussion_r92590246, @lcobucci?

Ocramius:
After looking at it with @sebastianfeldmann at the DPC code night, I'd say that this is NOT ready to be deployed. Instead, a new type is needed, because this kind of change would mean that users of the library would have half the data WITH the sign, and half the data without.

In addition to that, binding with the sign would change behavior: queries using parameter binding would be broken.

That's the way to go IMO, whereas the existing type would have to stay unchanged.

Ocramius:
Closing and pushing discussion back to #2578

Ocramius:
Re-opening as per https://github.com/doctrine/dbal/issues/2579#issuecomment-312047665 - sorry @galeaspablo!

galeaspablo:
#2316 was merged. Through it, the database value format was changed. As I said before, this means that I had to refactor my PR.

I've refactored my PR to work with the new master branch (i.e. handle negative support with the new format).

lcobucci (dismissed):
Just synced with `master`. LGTM

Ocramius:
@lcobucci as mentioned before, the fact that I released early is my fuckup (technical terminology).

We discussed it earlier tonight, and while it is technically possible for us to introduce a quick `2.6.1` with the change in it, nothing denies folks to install `2.6.0` now, then upgrade and have it broken later.

This PR is now a BC break, and I don't think we can fix it without introducing a `SignedDateIntervalType`.

@galeaspablo I apologise, this will need some rework :-(

Ocramius:
I think so. @deeky666 @lcobucci thoughts?

lcobucci:
#2578 is labeled as a `bug` and the PR is an improvement. If we tackle it as an improvement I with a new type (maybe even deprecating the existing one), however if we see it as a bug I'd rather change the PR in a way that doesn't break BC.

We also need to keep in mind that new features will land just on `2.7`.

Majkl578:
Hi, shall we consider this a blocker for 2.7? If so, it needs a rebase & fixing the conflicts, otherwise we could push this back to 2.8/3.0 and decide later.

Ocramius (approved):

morozov (approved):

PR #2581, opened 2016-12-20
Fetch table options to improve dumping of table structure

digilist:
Currently, information like the collation, charset and the engine (in MySQL) get lost when I use Doctrine to analyze the data and to create "CREATE TABLE" statements.  When I have a table, that has for example a utf8_general_ci collation, Doctrine does not fetch this information. However, the collation of a column is fetched correctly.

Consider the following table:

<pre>
CREATE TABLE foo
(name VARCHAR(255) NOT NULL COLLATE utf8_general_ci)
DEFAULT CHARACTER SET utf8 COLLATE <b>utf8_general_ci</b> ENGINE = InnoDB;
</pre>

By dumping the table structure, we get the following result:

<pre>
CREATE TABLE foo
(name VARCHAR(255) NOT NULL COLLATE utf8_general_ci)
DEFAULT CHARACTER SET utf8 COLLATE <b>utf8_unicode_ci</b> ENGINE = InnoDB;
</pre>

As you can see, the collation information on the table got lost. And it is the same with the character set and the engine and probably other options I don't know about.

I'd like to propose a feature that analyses the tables and adds those options into the Table object.

Before I start to implement this feature for all platforms and build tests, I'd like to ask for your feedback. Are there any objections not to implement this feature?

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -2837,6 +2837,20 @@ abstract class AbstractPlatform
     }
 
     /**
+     * Returns the SQL Query that fetches a table's options like collation, charset or engine.
+     *
+     * @param string $table
+     *
+     * @return string
+     *
+     * @throws \Doctrine\DBAL\DBALException If not supported on this platform.
+     */
+    public function getTableOptionsSQL($table, $database = null)
+    {
+        throw DBALException::notSupported(__METHOD__);
+    }
+
+    /**
      * @param string $name
      * @param string $sql
      *
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -209,6 +209,32 @@ class MySqlPlatform extends AbstractPlatform
     }
 
     /**
+     * @param string $table
+     *
+     * @return string
+     *
+     * @throws \Doctrine\DBAL\DBALException If not supported on this platform.
+     */
+    public function getTableOptionsSQL($table, $database = null)
+    {
+        $table = $this->quoteStringLiteral($table);
+
+        if (null !== $database) {
+            $database = $this->quoteStringLiteral($database);
+        }
+
+        $sql = "SELECT t.ENGINE AS `engine`, t.table_collation AS `collate`, ccsa.character_set_name AS `charset`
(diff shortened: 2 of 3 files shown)


lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2837,6 +2837,20 @@ public function getListTableForeignKeysSQL($table)
...
     /**
+     * Returns the SQL Query that fetches a table's options like collation, charset or engine.
+     *
+     * @param string $table
+     *
+     * @return string
+     *
+     * @throws \Doctrine\DBAL\DBALException If not supported on this platform.
+     */
+    public function getTableOptionsSQL($table, $database = null)
```
phansys, line 2848: Docblock for argument 2 is missing.

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -209,6 +209,32 @@ public function getListTableForeignKeysSQL($table, $database = null)
...
+        if (null !== $database) {
+            $database = $this->quoteStringLiteral($database);
+        }
+
+        $sql = "SELECT t.ENGINE AS `engine`, t.table_collation AS `collate`, ccsa.character_set_name AS `charset`
+                FROM information_schema.`TABLES` t
+                JOIN information_schema.`COLLATION_CHARACTER_SET_APPLICABILITY` ccsa ON ccsa.collation_name = t.table_collation
+                WHERE t.table_name = $table";
+
+        $databaseNameSql = null === $database ? 'DATABASE()' : $database;
```
phansys, line 231: `$databaseNameSql = $database ?: 'DATABASE()';`

deeky666:
@digilist generally speaking I am not against this feature. You have have to be aware of the fact that this feature only covers dumping table options, it will not be used by the comparator for comparing schemas (and I would not suggest adding that in the future as this is too vendor specific). Despite that, this change introduced a BC break for custom platform/schema implementation because of the exception being thrown in the abstract platform. We would have to turn this into a non breaking change I suppose.

digilist:
I can remove the exception, I just reflected that behavior from another method. Returning an empty array is fine.

If we just return an empty array, do we need to implement this for all database platforms or is it sufficient to implement it only for some at first?

phansys (changes requested):
Status: Needs work.

morozov:
Closing as stale and missing tests.

PR #2585, opened 2016-12-22
OraclePlatform : Add an owner condition to the list table column comment subquery

### LINKED ISSUE #2584: Doctrine:Schema:Validate leeds to ORA-01427(single-row subquery returns more than one row)
Hi,
I have a problem on Oracle platform with the command doctrine:schema:validate with doctrine/dbal 2.5.5.

When I lauch the command I get the following error : 

```
  [Symfony\Component\Debug\Exception\ContextErrorException]
  Warning: oci_fetch_all(): ORA-01427: single-row subquery returns more than
  one row
```

I recently upgraded from doctrine/dbal 2.5.4 to doctrine/dbal 2.5.5 and I finally managed to found where the regretion(?) comes from.

It apparently comes from the change of (new) line 668 of the file OraclePlatform.php on this commit : https://github.com/doctrine/dbal/commit/a23a47115f0fa35cf349735dfe342e2715f97f30

It not only change style of the query but transform an innerJoin to a subquery.
The innerJoin allowed several line as subquery doesn't.

This would work perfectly fine if no column have more than on comment.
In my Database, I have several users and each user have a comment on my column. 
And the subquery fail.

This can be corrected By adding an other condition on the subquery to get current user comments only.

I'm working on a PR to show the fix and iterate.

Deamon:
This PR is to fix #2584.

It adds a complementary where condition on the request in the `getListTableColumnsSQL` of `OraclePlatform`class.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
@@ -672,8 +672,7 @@ LEFT JOIN user_cons_columns r_cols
                              SELECT d.comments
                              FROM   $colCommentsTableName d
                              WHERE  d.TABLE_NAME = c.TABLE_NAME
-                             AND    d.COLUMN_NAME = c.COLUMN_NAME
-                             $ownerCommentsCondition
+                             AND    d.COLUMN_NAME = c.COLUMN_NAME $ownerCommentsCondition
                          ) AS comments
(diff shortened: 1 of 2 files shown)


deeky666:
@Deamon that patch looks good but please target the PR against `master`. We will do backport ourselves then. Thanks.

deeky666:
Fixed in #2518

PR #2590, opened 2017-01-02
Remove support for PHP 5.x

mrailton:
Removes support for PHP 5.x from composer.json and .travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,10 +40,6 @@ matrix:
         mariadb: 10.0
       env: DB=mariadb MARIADB_VERSION=10.0
 
-    - php: 5.6
-      env: DB=mariadb MARIADB_VERSION=10.1
-      addons:
-        mariadb: 10.1
     - php: 7.0
       env: DB=mariadb MARIADB_VERSION=10.1
       addons:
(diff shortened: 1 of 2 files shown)


Ocramius:
@railto check the build matrix - some 5.x stuff still running.

Ocramius:
LGTM! 👍

photodude:
Personally, I would have suggested against this change. PHP 5.6 is a major supported version by basically every host out there and will be until at least the end of PHP 5.6 security support on 31 Dec 2018 (about 2 years) 

php 7 will end security support about 28 days before php 5.6
See http://php.net/supported-versions.php

Ocramius:
You'll have security support from us on DBAL 2.5.x.
There's no point in limiting development to older php versions when aiming
at cleanup and feature improvement.

photodude:
I could understand that considering a b/c break for a 3.x release representing the current master branch 
and the 2.5 branch is only supporting security fixes and is no longer back porting the master branch changes.

Ocramius:
Bumping dependencies is not a BC break. Also: build new software on new
foundations, maintain existing one on stable ones. We are not touching the
2.5.* dependencies, but we need to move on to be able to do anything at
all, and to discourage building new stuff on oldstable.

photodude:
understood. I see how you are working the same verifications I would expect (just in a slightly different way)

PR #2591, opened 2017-01-03
Allow empty where for update and delete

enumag:


Ocramius:
I'm inclined to disallow this, and throw an exception on empty identifiers instead.

An empty identifier in this API is high risk, IMO.

kimhemsoe:
@Ocramius Agree. It is a hidden foot gun.

deeky666:
Agreed. This API is not intended for your use-case. Also the method signature does not define empty identifiers to be allowed. Please use `executeUpdate()` instead.

PR #2592, opened 2017-01-06
Ensure the database name is always provided when trying to use pgsql in the test suite

mikaelkael:
See https://travis-ci.org/doctrine/dbal/jobs/181071893#L265 for example


Ocramius:
LGTM, thanks!

PR #2600, opened 2017-01-15
Build column

marktopper:
Makes it easier for developers to extend only the Column object.


Ocramius:
Column definitions are not designed for inheritance, nor are the concrete platforms, which are simply to be replaced when customization is needed.

Closing as `won't fix`

PR #600, opened 2014-05-07
Support for Partial Indexes for PostgreSql and Sqlite

PowerKiKi:
Support for Partial Indexes was available in Doctrine 1 following
http://www.doctrine-project.org/jira/browse/DC-82. This commit
reintroduce support for Doctrine 2. We use the same syntax with an
optionnal "where" attribute for Index and UniqueConstraint.

It is unit-tests covered and documented in manual.

PR #2607, opened 2017-01-18
DBAL-2565 Remove fragile test that cannot be abstracted across all PostgreSQL versions

deeky666:
See https://github.com/doctrine/dbal/issues/2565

As discussed in https://github.com/doctrine/dbal/pull/600#issuecomment-273343134 we have to remove a functional test about partial index definition comparison in PostgreSQL, as since version `9.5` the definition is decompiled in a different format by the database server, which makes the original definition uncomparable to the introspected one without a complex expression parser.

As partial indexes are not portable anyways, we accept this incompatibility to happen.


Ocramius:
LGTM 👍

Issue #2511, opened 2016-09-20
DBAL Connection::update() - types list does not match the parameter list if fields are used in SET and WHERE clause


Dormilich:
I think the issue is best explained with an example:

``` php
$conn->update('TestTable', [
    'text' => 'some text',
    'is_edited' => true,
], [
    'id' => 1,
    'is_edited' => false,
], [
    'text' => 'string',
    'is_edited' => 'boolean',
    'id' => 'integer',
]);
```

What is then passed to `Connection::executeUpdate()`:
SQL: (as expected)

``` sql
UPDATE TestTable SET text = ?, is_edited = ? WHERE id = ? AND is_edited = ?
```

Params: (as expected)

```
array(4) {
  [0]=>  string(9) "some text"
  [1]=>  bool(true)
  [2]=>  int(1)
  [3]=>  bool(false)
}
```

Types: (one type definition short!)

```
array(3) {
  [0]=>  string(6) "string"
  [1]=>  string(7) "boolean"
  [2]=>  string(7) "integer"
}
```

deeky666:
Patch provided in #2609

Ocramius:
Handled in #2609

PR #2609, opened 2017-01-18
[DBAL-2511] Fix type list extraction in connection update method

deeky666:
See #2511 

Using `Doctrine\DBAL\Connection::update()` method, extracting parameter types from `$data` and `$identifiers` altogether does not work if a column name exists both in `$data` and `$identifier` ans `$types` is given as a column => type map and not as an ordered list that reflects all given parameters.

As a positive side effect we got rid of a lot of `array_*()` function callse.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -579,16 +579,20 @@ class Connection implements DriverConnection
             throw InvalidArgumentException::fromEmptyCriteria();
         }
 
+        $columnList = array();
         $criteria = array();
+        $paramValues = array();
 
-        foreach (array_keys($identifier) as $columnName) {
+        foreach ($identifier as $columnName => $value) {
+            $columnList[] = $columnName;
             $criteria[] = $columnName . ' = ?';
+            $paramValues[] = $value;
         }
 
         return $this->executeUpdate(
             'DELETE FROM ' . $tableExpression . ' WHERE ' . implode(' AND ', $criteria),
-            array_values($identifier),
-            is_string(key($types)) ? $this->extractTypeValues($identifier, $types) : $types
+            $paramValues,
+            is_string(key($types)) ? $this->extractTypeValues($columnList, $types) : $types
         );
     }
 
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Connection.php
```diff
@@ -646,23 +650,31 @@ public function getTransactionIsolation()
...
 
         foreach ($data as $columnName => $value) {
+            $columnList[] = $columnName;
             $set[] = $columnName . ' = ?';
+            $paramValues[] = $value;
         }
 
-        if (is_string(key($types))) {
-            $types = $this->extractTypeValues(array_merge($data, $identifier), $types);
+        foreach ($identifier as $columnName => $value) {
```
Ocramius: This `foreach` matches 1:1 the previous one. Same goes for the  subsequent one, besides for one line diff. Is it possible to give this a name, and throw it into a private method?

lib/Doctrine/DBAL/Connection.php
```diff
@@ -682,29 +694,39 @@ public function insert($tableExpression, array $data, array $types = array())
...
      */
-    private function extractTypeValues(array $data, array $types)
+    private function extractTypeValues(array $columnList, array $types)
     {
         $typeValues = array();
 
-        foreach ($data as $k => $_) {
-            $typeValues[] = isset($types[$k])
-                ? $types[$k]
+        foreach ($columnList as $columnIndex => $columnName) {
```
Ocramius: Much better :-)

lib/Doctrine/DBAL/Connection.php
Ocramius: Hmm, no, never mind, this appends to the data....

Ocramius (approved):

Ocramius:
After checking locally, I'd say this is good to go! 👍

Ocramius:
Backported into `2.5` via doctrine/dbal@0d2f8187d4b4c7b72d8e2acba359e25c36feaf5e

PR #2611, opened 2017-01-20
[DBAL-2508] Pass through index options when renaming index on table

### LINKED ISSUE #2508: Renaming index does not rename the index, but creates new one
When I try to rename the index using the API, the old options are not passed to the new index, which causes that the index is not renamed, but _different_ index is created.

```
$table = Table();
$table->addIndex(['name'], 'idx_name', [], [            
    'where' => 'name IS NOT NULL'
]);
$table->renameIndex('idx_name', 'idy_foo');
```

The method renameIndex() does not pass the options from the old index.

deeky666:
Fixes #2508 

Renaming an index on a table object in fact creates a new index object from the old one internally but did not pass index options from the old index to the new one.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Table.php
+++ b/lib/Doctrine/DBAL/Schema/Table.php
@@ -253,10 +253,10 @@ class Table extends AbstractAsset
         unset($this->_indexes[$oldIndexName]);
 
         if ($oldIndex->isUnique()) {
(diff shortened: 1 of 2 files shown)


Ocramius (approved):

Ocramius:
👍

Ocramius:
Backported via doctrine/dbal@fc376f7a61498e18520cd6fa083752a4ca08072b

PR #2612, opened 2017-01-21
[DBAL-2555] Fix date and datetimetz type mapping on Oracle

### LINKED ISSUE #2555: schema generation maps oracle DATE fields to datetime
I am using the schema generation tool to generate create/update scripts for my database from yaml entities.
Using http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html#datetime as a guide, I have a date column of type 'date', which correctly generates an oracle column with type DATE.
However, when I try to generate update scripts, OraclePlatform::doctrineTypeMappings maps the oracle DATE to a doctrine DateTimeType. The columns are seen as different because DateType != DateTimeType.
If I change the mapping date->datetime to date->date, it correctly sees that the database column matches the entity definition.

I think this is a bug, although minor since the update sql is generated from the entity definition and ends up being a no-op.

deeky666:
Fixes #2555 

Oracle's `DATE` type was erroneously mapped to Doctrine's `DateTime` type. Although I wonder why nobody ever complained about that until now, the fix looks straight forward and also complies to what we have documented.

Additionally the instrospection of `DateTimeTzType` was wrong and has been fixed (wrong case in type comparison).

Tests pass locally.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
@@ -102,7 +102,7 @@ class OracleSchemaManager extends AbstractSchemaManager
 
         $dbType = strtolower($tableColumn['data_type']);
         if (strpos($dbType, "timestamp(") === 0) {
-            if (strpos($dbType, "WITH TIME ZONE")) {
+            if (strpos($dbType, "with time zone")) {
                 $dbType = "timestamptz";
             } else {
                 $dbType = "timestamp";
--- a/tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
@@ -412,6 +412,9 @@ class OraclePlatformTest extends AbstractPlatformTestCase
         $this->assertEquals($expectedSql, $this->_platform->getAlterTableSQL($tableDiff));
     }
 
+    /**
+     * @group DBAK-2555
+     */
     public function testInitializesDoctrineTypeMappings()
     {
         $this->assertTrue($this->_platform->hasDoctrineTypeMappingFor('long raw'));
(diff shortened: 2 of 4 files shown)


lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
```diff
@@ -102,7 +102,7 @@ protected function _getPortableTableColumnDefinition($tableColumn)
 
         $dbType = strtolower($tableColumn['data_type']);
         if (strpos($dbType, "timestamp(") === 0) {
-            if (strpos($dbType, "WITH TIME ZONE")) {
+            if (strpos($dbType, "with time zone")) {
```
Ocramius: Since this seems to be something that the DB engine doesn't guarantee, `strtolower()` should be used (or at least case insensitive matching).

Ocramius: Test probably also needed with it, but unsure if practical

Ocramius (changes requested):

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
deeky666: Have a look at line 193 ;)
But besides that you reminded me that I forgot a unit test case for the mapping issue in the platform which I added now.

deeky666: This is is covered by the functional test case btw. Not sure how else to test this :>

Ocramius: Dammit, didn't read that, thanks :-)

Ocramius (approved):

tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
```diff
@@ -412,13 +412,19 @@ public function testAlterTableNotNULL()
         $this->assertEquals($expectedSql, $this->_platform->getAlterTableSQL($tableDiff));
     }
 
+    /**
+     * @group DBAK-2555
```
DeepDiver1975: DBAL :wink:

deeky666: thx, fixed

Ocramius:
:shipit:

Ocramius:
Backported via doctrine/dbal@8a13144aa046f4c218dc12ed48167de7b184416c

PR #2618, opened 2017-01-26
Add Connection::fetchAllArray() method to allow a convenience method …

jackcholt:
…for getting all records as a two-dimensional numerically indexed array.

Add PHPunit test for fetchAllArray()

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -784,6 +784,20 @@ class Connection implements DriverConnection
     }
 
     /**
+     * Prepares and executes an SQL query and returns the result as a numerically indexed array.
+     *
+     * @param string $sql    The SQL query.
+     * @param array  $params The query parameters.
+     * @param array  $types  The query parameter types.
+     *
+     * @return array All rows returned by the SQL query as a two dimensional numerically indexed array.
+     */
+    public function fetchAllArray($sql, array $params = array(), $types = array())
+    {
+        return $this->executeQuery($sql, $params, $types)->fetchAll(PDO::FETCH_NUM);
+    }
+
+    /**
      * Prepares an SQL statement.
      *
      * @param string $statement The SQL statement to prepare.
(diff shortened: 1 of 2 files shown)


deeky666:
@Ocramius thoughts? I tend to reject this to not bloat the API with additional `fetch*()` methods (potentially having requests for more convenience methods in the future). Not sure if the connection should act as a full-blown facade or just as a "simple" wrapper around driver connections.

jackcholt:
In a sense fetchArray and all the other fetch* methods are convenience methods as well. To me it seems like a glaring ommission to not have fetchAllArray. The set of fetch methods don't seem complete without it.

Ocramius:
@deeky666 agree here: wouldn't add more convenience methods, and would instead have a separate "finder" or such object where we start stashing this functionality. Also: we really need to use iterators with this stuff.

deeky666:
@Ocramius agreed. That "finder" stack is theoretically the right thing to do but won't give any benefit over `executeQuery()->fetchAll(\PDO::FETCH_NUM)`.

@jackcholt I see your point here but if we do so, we would have to add methods for the other fetch modes, too (for consistency) and this is too much bloating. We should have stacked the `fetch*()` in the first place when designing the class.

An idea I have would be to add a fourth optional parameter to `fetchAll()` that allows to pass a fetch mode to use. Thoughts?

jackcholt:
I like the fourth optional parameter on fetchAll() if my current impl of
fetchAllArray is too much bloat.

deeky666:
@Ocramius do you see any issues with this compromise?

jackcholt:
If you agree to the compromise I would be happy to implement.

Ocramius:
@deeky666 @jackcholt I'm throwing my veto on this: the public API does indeed just provide a shortcut for `$this->executeQuery($sql, $params, $types)->fetchAll(PDO::FETCH_NUM);`.

Since the public API is already quite bloated, we'd just introduce further BC concerns for the future, when/if we want to move to finder/repository objects.

In addition to that, this is really as simple as `array_map('array_values', $connection->fetchAll($sql))`, and fetching by numerical index is below the 80/20 scenario.

deeky666:
Ok closing as won't fix then. It really is no big deal to use the provided alternatives.

jackcholt:
@deeky666 Thanks. I'm already using your "shortcut" in my code but thought
it slightly opaque and not too self-documenting for long term use. I may
extend the Connection class and add the fetchAllArray convenience method
for internal use.

PR #2619, opened 2017-01-27
Add Table->toArray() and Table->toJson()

abdgad:


Ocramius:
@abdgad this is just a serialization change: if you need it in your specific domain, code it there. There's no need to add it to the DBAL though.

Closing, since this API will not be used internally.

PR #2620, opened 2017-01-27
Added Interface for Drivers for Platforms that support deactivation of foreign key checks

stollr:
MySQL supports the deactivation of foreign key checks. This is helpful if one wants to purge the database. Usually Doctrine tries to figure out in which order tables have to be cleared. This must not be done if the platform supports the deactivation of foreign key checks.

This feature may be used in the ORMPurger of [doctrine/data-fixtures](https://github.com/doctrine/data-fixtures). This would fix the issue with purging tables that have self-referencing associations (see doctrine/data-fixtures#147, doctrine/data-fixtures#159 or doctrine/data-fixtures#251)

I'll create a PR that uses this feature in the ORMPurger of [doctrine/data-fixtures](https://github.com/doctrine/data-fixtures).

I hope that this could be merge.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
@@ -34,7 +34,11 @@ use Doctrine\DBAL\VersionAwarePlatformDriver;
  * @link   www.doctrine-project.org
  * @since  2.5
  */
-abstract class AbstractMySQLDriver implements Driver, ExceptionConverterDriver, VersionAwarePlatformDriver
+abstract class AbstractMySQLDriver implements
+    Driver,
+    ExceptionConverterDriver,
+    VersionAwarePlatformDriver,
+    ForeignKeyCheckDeactivatableDriver
 {
     /**
      * {@inheritdoc}
@@ -146,6 +150,22 @@ abstract class AbstractMySQLDriver implements Driver, ExceptionConverterDriver, 
         }
 
         return $this->getDatabasePlatform();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function activateForeignKeyChecks(\Doctrine\DBAL\Connection $conn)
+    {
+        $conn->query('SET FOREIGN_KEY_CHECKS=1');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function deactivateForeignKeyChecks(\Doctrine\DBAL\Connection $conn)
+    {
+        $conn->query('SET FOREIGN_KEY_CHECKS=0');
     }
 
     /**
--- a/lib/Doctrine/DBAL/Driver/ForeignKeyCheckDeactivatableDriver.php
+++ b/lib/Doctrine/DBAL/Driver/ForeignKeyCheckDeactivatableDriver.php
@@ -1,0 +1,46 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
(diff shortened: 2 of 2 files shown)


deeky666:
Unfortunately this is not a solution we can implement in DBAL this way. There are 2 issues here:

1. It is not the driver that defines the capability of being able to disable foreign key constraints but the platform. It is a platform feature, not driver specific.
2. The more important one: Such a solution as proposed here might help with MySQL but does not apply to other platforms (concept wise). The different platforms have different possibilities to solve the foreign key constraint problem. We need to find a way to potentially enable such a feature for every platform without being too vendor specific as for example assuming there is such common thing as "let's disable all foreign key checks for a moment". PostgreSQL for example is not able to do that but instead can only disable ALL kinds of constraints on a certain table. This might not be a problem in the context of migrations, but implementing it as "supports disable foreign keys" in the platform is wrong and dangerous. Using the API for other purposes let's the user think "ok now I have only foreign keys disabled" but in fact all other kinds of constraints are disabled, too. PostgreSQL has another concept of cascading truncates accross relations (which is part of the `TRUNCATE` statement). Other platforms have similar porpblems and different concepts to achieve the goal of truncation of tables without checking referential constraints.

We had multiple discussions about that already and it always comes down to the same issues. IMO we should not try to abstract it in the platform itself but rather introduce a new `TruncateTableStatement` or thelike that dispatches to the platform like `AbstractPlatform::getForcedTruncateTableSQL()` which then returns the SQL statements required to execute the operation (can be multiple then). With this approach the platform does not have to expose public methods like `disabledForeignKeyChecks()` which are too vendor specific but can use internal knowledge to achieve the solution.

Hope that helps.

deeky666:
@naitsirch I did some tests on this issue and my conclusion is that there are possibilities to achieve a "forced" truncate on probably all of the platforms but all of them but MySQL have side effects that can be pretty severe and thus not an option. I'll try to sum that up:

MySQL can disable foreign key checks per session and thus is considered quite safe. Disabling foreign key checks does not affect other connections and if something goes wrong, foreign key checks will be enabled again automatically on the next connection.

PostgreSQL does not need to disable foreign key checks as it supports a cascade option for `TRUNCATE TABLE` which first truncates childs so that integrity is not an issue at all.

Oracle also supports a cascade option but fails to execute if the referencing foreign key constraints have a `ON DELETE RESTRICT` clause (so this is not an option). Oracle also supports disabling constraints but doing so requires a `ALTER TABLE` statement and thus is permanent, not session based. If something goes wrong before reenabling them, the database is in an inconsistent state.

SQL Server and other vendors also support disabling constraints but as with Oracle on a permanent basis.

So the only addition we can make in DBAL would be MySQL specific and I doubt this is a good idea. So to put this topic to a conclusion, I would suggest not trying to bake anything into DBAL concerning "forced truncates" as it simply does not make any sense here. The only compromise I could accept is to provide methods for generating the SQL for disabling/enabling foreign key checks on the MySQL platform and have third party code deal with it on their own (no abstraction).
The cross-platform solution would be to drop and recreate foreign keys when purging data. But here too, there is the risk of leaving the database schema in an inconsistent state, if something goes wrong.

@guilhermeblanco @mikeSimonson @Ocramius thoughts?

lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
```diff
@@ -34,7 +34,11 @@
  * @link   www.doctrine-project.org
  * @since  2.5
  */
-abstract class AbstractMySQLDriver implements Driver, ExceptionConverterDriver, VersionAwarePlatformDriver
+abstract class AbstractMySQLDriver implements
+    Driver,
+    ExceptionConverterDriver,
+    VersionAwarePlatformDriver,
+    ForeignKeyCheckDeactivatableDriver
```
Ocramius: This is something that we really shouldn't expose to the outside world, as it opens also a series of cans of worms.

I would rather have the intent of the actual affected operations to be either correctly have workarounds built into DBAL (truncation, in this case), or just not using DBAL facilities at all (since we're going into veeeeeery specific details of horrible hacks like MySQL's)

Ocramius (changes requested):

Ocramius:
@deeky666 idea is nice, IMO a no-go. We're just going too deep into what a single platform can do, and honestly that's not what an abstraction layer should be for.

deeky666:
@Ocramius I think I have summed up the possibilities we have here quite well and I agree with you, we should not try to achieve something here in DBAL. We simply cannot provide something that is considered "safe" accross platforms. So IMO third party code has to built its own implementation around this because it is highly use-case specific. If database integrity does not matter, they can execute their statement to enforce unsafe truncation. The only thing we could provide here in DBAL is SQL generation for vendor specific stuff to "disable foreign keys" without making it an abstraction that ensures a certain behaviour.

naitsirch:
@deeky666 and @Ocramius thanks for looking into this. I unterstand your objections and will close this PR.

PR #2622, opened 2017-01-27
Remove dead code from `MasterSlaveConnection::connect()`

jnvsor:
Currently connect unsets the slave connection right before overwriting it with the master anyway.

### DIFF
--- a/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
+++ b/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
@@ -173,13 +173,9 @@ class MasterSlaveConnection extends Connection
         }
 
         if ($connectionName === 'master') {
-            // Set slave connection to master to avoid invalid reads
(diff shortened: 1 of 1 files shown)


deeky666:
@jnvsor can you please explain why this is dead code?

jnvsor:
Actually you're right, it's not dead it's duplicate.

Specifically, it unsets `$this->connections['slave']` under the condition that `!$this->keepSlave`

2 lines later it sets `$this->connections['slave']` to the newly acquired `$this->connections['master']` when `!$this->keepSlave`: If the first line executed, so will the second. Either way If the connection object no longer has any references it will be gc'd

Unless `$this->connectTo('master')` does something as unintuitive as querying the slave database there won't be any difference. (I checked, It doesn't)

deeky666 (approved):

deeky666:
Got it! Thanks for clearing that up. I'll accept it as is, as this code path is rather hard to test.

PR #2630, opened 2017-02-04
DBAL-2626 Add PHPUnit config for ContinuousPHP OracleDB testing

### LINKED ISSUE #2626: [RFC] testing Oracle DB on travis ci
Apparently, someone has a workaround for testing Oracle DB on travis ci as well https://github.com/cbandy/travis-oracle

an Oracle account is required and the Oracle username and password has to be in the build environment variables either as hidden repository settings or encrypted variables

deeky666:
Closes #2626.

### DIFF
--- a/tests/continuousphp/oci8.phpunit.continuousphp.xml
+++ b/tests/continuousphp/oci8.phpunit.continuousphp.xml
@@ -1,0 +1,38 @@
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.8/phpunit.xsd"
+         backupGlobals="false"
+         colors="true"
+         bootstrap="../../vendor/autoload.php"
+>
+  <php>
+    <ini name="error_reporting" value="-1" />
+
+    <var name="db_type" value="oci8"/>
+    <var name="db_host" value="oracle-xe-11" />
+    <var name="db_username" value="C##doctrine" />
+    <var name="db_password" value="ORACLE" />
+    <var name="db_name" value="XE" />
+    <var name="db_port" value="1521"/>
+    <var name="db_event_subscribers" value="Doctrine\DBAL\Event\Listeners\OracleSessionInit"/>
+
+    <var name="tmpdb_type" value="oci8"/>
+    <var name="tmpdb_host" value="oracle-xe-11" />
+    <var name="tmpdb_username" value="ORACLE" />
+    <var name="tmpdb_password" value="ORACLE" />
+    <var name="tmpdb_name" value="XE" />
+    <var name="tmpdb_port" value="1521"/>
+  </php>
+
+  <testsuites>
+    <testsuite name="Doctrine DBAL Test Suite">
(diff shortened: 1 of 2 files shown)


deeky666:
Seems to work. The error was expected and has to be fixed (it is a test error on Oracle 11g).

photodude:
Is there a reason for not using the new convention we set up for the config file naming?
- expected `tests/<service>/<db_type>.<db_version>.<dbal_driver_name>.<service>.xml`
- found `tests/continuousphp/oci8.phpunit.continuousphp.xml`

Does continuousphp not use a yml config file?

deeky666:
@photodude I did not care about convention here because there is only 1 DB version and only 1 driver we want to test here.
Continuousphp does not have a config file, no. It is all configured via their web frontend.

deeky666:
Though I can rename the phpunit config file is that is desired.

Ocramius:
Gotta love OracleDB's expressiveness:

```
1) Doctrine\Tests\DBAL\Functional\Driver\OCI8\OCI8ConnectionTest::testLastInsertIdAcceptsFqn 
Exception: [Doctrine\DBAL\Exception\DriverException] An exception occurred while executing 'CREATE TABLE DBAL2595(id NUMBER DEFAULT dbal2595_seq.NEXTVAL, foo NUMBER)': 
 
ORA-00984: column not allowed here 
 ```

File: tests/continuousphp/oci8.phpunit.continuousphp.xml [hunk 1]
```diff
@@ -0,0 +1,38 @@
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.8/phpunit.xsd"
+         backupGlobals="false"
+         colors="true"
+         bootstrap="../../vendor/autoload.php"
+>
+<php>
```
Ocramius: Alignment is a bit borked all over the file ;-)

Ocramius (changes requested):
Need it to be green before we merge it, obviously.

photodude:
Just a consistency item, if that is the convention that we want to try to stick to.

Also, isn't there two drivers to test, `PDO_OCI` and `OCI8`?

deeky666:
@Ocramius I wanted to add a followup PR that fixes the test but I can do it right away with this PR if you want to.
@photodude you are right but the `PDO_OCI` PHP is very buggy since years and even was marked "experimental" until some time ago (no clue why they changed this). With `PDO_OCI` our testsuite is even segfaulting. Maybe we should remove support for it altogether. Also continuousphp does not have the driver installed so we won't be able to test against it out of the box anyways.

tests/continuousphp/oci8.phpunit.continuousphp.xml
deeky666: Fixed.

deeky666:
@Ocramius fixed the failing test. It's green now :)

Ocramius:
LGTM! Finally some CI for OracleDB as well. I know we'll regret it, but at least we can enjoy a few moments of glory :-)

PR #2631, opened 2017-02-05
Fix `CREATE`/`DROP DATABASE` support on SQL Server

deeky666:
It is actually possible to use `CREATE DATABASE` and `DROP DATABASE` on SQL Server. The platform did not reflect that. The issue arised when testing SQL Server where the testsuite failed, if the database used by the testsuite didn't exist. With this patch, it works as expected.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -173,7 +173,7 @@ class SQLServerPlatform extends AbstractPlatform
      */
     public function supportsCreateDropDatabase()
     {
-        return false;
+        return true;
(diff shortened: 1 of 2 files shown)


Ocramius:
@deeky666 given that this changes behavior of the platform, wouldn't it better to push it to `2.6` only?

Ocramius:
Merging to `2.6` only, but feel free to fight over it with me :-)

deeky666:
@Ocramius well in my opinion it is a bug because the the platform does implement all that is necessary (the create/drop SQL) but does not tell the client that this feature is supported. I don't care really, but it will make impossible to test without explicitly creating the database before testing or creating a fresh database via ORM's schema tool.

PR #2632, opened 2017-02-05
Fix `"application_name"` connection parameter tests for PostgreSQL < 9.2

deeky666:
Testing against PostgreSQL < 9.2 revealed a test issue, where a column selected from `pg_stat_activity` does not exist. The column is named `current_query` instead of `query` on PostgreSQL < 9.2.


Ocramius:
LGTM 👍

PR #2633, opened 2017-02-05
Run SQL Azure tests conditionally

deeky666:
SQL Azure related tests should not run, if the SQL Server edition does not support Azure related features. This patch marks the tests in question as skipped, if the database server edition is not `SQL Azure`.


Ocramius:
LGTM, but not backporting it.

PR #2634, opened 2017-02-05
Fix fetching last insert ID for sequences on SQL Server

deeky666:
Fixes fetching last insert ID for sequences on both `sqlsrv` and `pdo_sqlsrv` drivers. The implementation on `sqlsrv` was wrong and `pdo_sqlsrv` does not support this out of the box.

The issue was revealed in #2617. See:

- https://ci.appveyor.com/project/photodude/dbal/build/1.0.138/job/r1twtafm8vsaswy4#L441
- https://ci.appveyor.com/project/photodude/dbal/build/1.0.140/job/878bxlaoeiiajsqs

### DIFF
--- a/lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
+++ b/lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
@@ -40,6 +40,21 @@ class Connection extends PDOConnection implements \Doctrine\DBAL\Driver\Connecti
     /**
      * {@inheritDoc}
      */
+    public function lastInsertId($name = null)
+    {
+        if (null === $name) {
+            return parent::lastInsertId($name);
+        }
+
+        $stmt = $this->prepare('SELECT CONVERT(VARCHAR(MAX), current_value) FROM sys.sequences WHERE name = ?');
+        $stmt->execute(array($name));
+
+        return $stmt->fetchColumn();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
     public function quote($value, $type=\PDO::PARAM_STR)
     {
         $val = parent::quote($value, $type);
--- a/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php
+++ b/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php
@@ -130,9 +130,8 @@ class SQLSrvConnection implements Connection, ServerInfoAwareConnection
     public function lastInsertId($name = null)
     {
         if ($name !== null) {
-            $sql = "SELECT IDENT_CURRENT(".$this->quote($name).") AS LastInsertId";
-            $stmt = $this->prepare($sql);
-            $stmt->execute();
+            $stmt = $this->prepare('SELECT CONVERT(VARCHAR(MAX), current_value) FROM sys.sequences WHERE name = ?');
+            $stmt->execute(array($name));
 
             return $stmt->fetchColumn();
         }


lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
```diff
@@ -40,6 +40,22 @@ public function __construct($dsn, $user = null, $password = null, array $options
     /**
      * {@inheritDoc}
      */
+    public function lastInsertId($name = null)
+    {
+        if (null === $name) {
+            return parent::lastInsertId($name);
+        }
+
+        $sql = 'SELECT CONVERT(VARCHAR(MAX), current_value) FROM sys.sequences WHERE name = ?';
```
Ocramius: Variable redundant

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
```diff
@@ -40,6 +40,22 @@ public function __construct($dsn, $user = null, $password = null, array $options
     /**
      * {@inheritDoc}
      */
+    public function lastInsertId($name = null)
+    {
+        if (null === $name) {
```
Ocramius: Can a name be empty?

Ocramius: As in `''`

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
```diff
@@ -40,6 +40,22 @@ public function __construct($dsn, $user = null, $password = null, array $options
...
+            return parent::lastInsertId($name);
+        }
+
+        $sql = 'SELECT CONVERT(VARCHAR(MAX), current_value) FROM sys.sequences WHERE name = ?';
+        $stmt = $this->prepare($sql);
+        $stmt->execute(array($name));
+
+        return $stmt->fetchColumn();
```
Ocramius: Can `current_value` ever be `NULL`?

Ocramius (changes requested):

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
deeky666: I don't think so. What should we expect in this case?

Ocramius: That's why I asked. Probably needs a new issue though, where we validate that the passed in value is not empty or such.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
deeky666: Given to [the documentation](https://msdn.microsoft.com/en-us/en-en/library/ff877934.aspx) no.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
deeky666: Redundant? You mean I should inject it into `prepare()` directly?

Ocramius: Correct

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
deeky666: @Ocramius the problem is the interface does not specify such case. Neither does PDO's interface. No exception, not non-string return value specified.

Ocramius: Then all ok: had to ask 👍

deeky666: Sure, not happy with the situation either but that's what it is right now :(

deeky666:
Btw the Oracle test failure is obviously unrelated. That test in question seems fragile. Think I have encountered that one on Oracle sometime before. Need to have a look at that.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
deeky666: Fixed.

deeky666:
Test failure fixed in #2635

Ocramius:
Backported to `2.5` via doctrine/dbal@31c41ee0d5d09bd02ee3ef2e93f647d0ceed4363

PR #2635, opened 2017-02-06
Fix date diff test

deeky666:
The date diff test was broken because it did not truncate the time part for the expected value. This could lead to test failures if running the test at a certain time of the day.

See: https://continuousphp.com/git-hub/doctrine/dbal/build/aaa77206-8e4a-40d7-994e-506c9b41862c/output


Ocramius:
Backported in doctrine/dbal@77b764ad8742feec5813df849f98d860ab08899d

PR #2636, opened 2017-02-06
Fix dropping in use databases on SQL Server and Oracle

deeky666:
This patch follows the same approach as for PostgreSQL. If trying to drop a database with active connections, the database server will refuse with:

```
SQLSTATE [42000, 3702]: [Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Cannot drop database "%s" because it is currently in use.
```

The schema manager catches that specific error code now, closes all active connections on that database and executes a `DROP DATABASE` again.

I had to fix the `SQLSrvException` to carry around the SQLSTATE and error code information.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/SQLServerSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/SQLServerSchemaManager.php
@@ -19,7 +19,8 @@
 
 namespace Doctrine\DBAL\Schema;
 
-use Doctrine\DBAL\Driver\SQLSrv\SQLSrvException;
+use Doctrine\DBAL\DBALException;
+use Doctrine\DBAL\Driver\DriverException;
 use Doctrine\DBAL\Types\Type;
 
 /**
@@ -34,6 +35,34 @@ use Doctrine\DBAL\Types\Type;
  */
 class SQLServerSchemaManager extends AbstractSchemaManager
 {
+    /**
+     * {@inheritdoc}
+     */
+    public function dropDatabase($database)
+    {
+        try {
+            parent::dropDatabase($database);
+        } catch (DBALException $exception) {
+            $exception = $exception->getPrevious();
+
+            if (! $exception instanceof DriverException) {
+                throw $exception;
+            }
+
+            // If we have a error code 3702, the drop database operation failed
+            // because of active connections on the database.
+            // To force dropping the database, we first have to close all active connections
+            // on that database and issue the drop database operation again.
+            if ($exception->getErrorCode() !== 3702) {
+                throw $exception;
+            }
+
+            $this->closeActiveDatabaseConnections($database);
+
+            parent::dropDatabase($database);
+        }
+    }
+
     /**
      * {@inheritdoc}
      */
@@ -212,7 +241,7 @@ class SQLServerSchemaManager extends AbstractSchemaManager
             } else {
                 throw $e;
             }
-        } catch (SQLSrvException $e) {
+        } catch (DBALException $e) {
             if (strpos($e->getMessage(), 'SQLSTATE [01000, 15472]') === 0) {
                 return array();
             } else {
@@ -257,5 +286,26 @@ class SQLServerSchemaManager extends AbstractSchemaManager
             INNER JOIN SysColumns Col ON Col.[ColID] = DefCons.[parent_column_id] AND Col.[ID] = Tab.[ID]
             WHERE Col.[Name] = " . $this->_conn->quote($column) ." AND Tab.[Name] = " . $this->_conn->quote($table) . "
             ORDER BY Col.[Name]";
+    }
+
+    /**
+     * Closes currently active connections on the given database.
+     *
+     * This is useful to force DROP DATABASE operations which could fail because of active connections.
+     *
+     * @param string $database The name of the database to close currently active connections for.
+     *
+     * @return void
+     */
+    private function closeActiveDatabaseConnections($database)
+    {
+        $database = new Identifier($database);
+
+        $this->_execSql(
+            sprintf(
+                'ALTER DATABASE %s SET SINGLE_USER WITH ROLLBACK IMMEDIATE',
+                $database->getQuotedName($this->_platform)
+            )
+        );
     }
 }
--- a/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvException.php
+++ b/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvException.php
@@ -19,9 +19,10 @@
 
 namespace Doctrine\DBAL\Driver\SQLSrv;
 
-use Doctrine\DBAL\DBALException;
 
-class SQLSrvException extends DBALException
+use Doctrine\DBAL\Driver\AbstractDriverException;
+
+class SQLSrvException extends AbstractDriverException
 {
     /**
      * Helper method to turn sql server errors into exception.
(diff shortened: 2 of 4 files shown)


lib/Doctrine/DBAL/Schema/SQLServerSchemaManager.php
```diff
@@ -37,6 +37,31 @@ class SQLServerSchemaManager extends AbstractSchemaManager
...
+
+            // If we have a error code 3702, the drop database operation failed
+            // because of active connections on the database.
+            // To force dropping the database, we first have to close all active connections
+            // on that database and issue the drop database operation again.
+            if ($exception->getErrorCode() !== 3702) {
+                throw $exception;
+            }
+
+            $this->_execSql($this->_platform->getCloseActiveDatabaseConnectionsSQL($database));
```
Ocramius: Call on a method not defined on the abstract class

lib/Doctrine/DBAL/Schema/SQLServerSchemaManager.php
```diff
@@ -37,6 +37,31 @@ class SQLServerSchemaManager extends AbstractSchemaManager
...
+            parent::dropDatabase($database);
+        } catch (DBALException $exception) {
+            /** @var \Doctrine\DBAL\Driver\DriverException $exception */
+            $exception = $exception->getPrevious();
+
+            // If we have a error code 3702, the drop database operation failed
+            // because of active connections on the database.
+            // To force dropping the database, we first have to close all active connections
+            // on that database and issue the drop database operation again.
+            if ($exception->getErrorCode() !== 3702) {
```
Ocramius: Move it to a constant, please (and document the constant)

lib/Doctrine/DBAL/Schema/SQLServerSchemaManager.php
```diff
@@ -37,6 +37,31 @@ class SQLServerSchemaManager extends AbstractSchemaManager
     /**
      * {@inheritdoc}
      */
+    public function dropDatabase($database)
```
Ocramius: The code additions here are untested. Can the scenario be replicated?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -169,6 +169,20 @@ public function getDropDatabaseSQL($name)
...
     /**
+     * Returns the SQL statement for closing currently active connections on the given database.
+     *
+     * This is useful to force DROP DATABASE operations which could fail because of active connections.
+     *
+     * @param string $database The name of the database to close currently active connections for.
+     *
+     * @return string
+     */
+    public function getCloseActiveDatabaseConnectionsSQL($database)
```
Ocramius: This method should probably land in the parent classes too then (sadly) :-\

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -169,6 +169,20 @@ public function getDropDatabaseSQL($name)
...
+    {
+        return 'ALTER DATABASE ' . $database . ' SET SINGLE_USER WITH ROLLBACK IMMEDIATE';
```
Ocramius: Should `$database` be quoted?

Ocramius (changes requested):

lib/Doctrine/DBAL/Schema/SQLServerSchemaManager.php
deeky666: There is already a test case that covers this issue in `SchemaManagerFunctionalTestCase::testDropsDatabaseWithActiveConnections()`

lib/Doctrine/DBAL/Schema/SQLServerSchemaManager.php
deeky666: I disagree here. It doesn't help anyone and if we start to introduce constants for all error codes and SQL states, we will go crazy. We started with this approach when we implemented driver exception converters and discarded that approach again. This check will disappear as soon as we implement SQL Server driver as driver exception converter drivers (hopefully) as we will be able to catch a dedicated exception type then. Also it is implemented that way for PostgreSQL already and I don't want to do everything that is necessary to make it "perfect" with this PR. First I want to get the SQL Server test suite to be green. We can still improve afterwards. What do you think?

lib/Doctrine/DBAL/Schema/SQLServerSchemaManager.php
deeky666: Same as with PostgreSQL. What shall I do here?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
deeky666: Using `Identifier` now.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
deeky666: I get your point but I do not want to enforce an abstraction here because it probably will be too vendor specific. Another option would be to put the SQL directly into the `dropDatabase()` method in the schema manager.

lib/Doctrine/DBAL/Schema/SQLServerSchemaManager.php
Ocramius: Works for me, since there is inline documentation. Also, didn't expect you to fix every instance in here.

If/when we move to PHP 7.1, then we should consider private constants.

lib/Doctrine/DBAL/Schema/SQLServerSchemaManager.php
Ocramius: @deeky666 one possible solution is to turn the return type into a `string[]`, then have a base method always returning an empty `array`. `array_walk`-ing over the result is sufficient to ensure consistency then.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
Ocramius: Possibly a better solution if the schema manager is platform specific (and won't rely on the platform abstraction anyway).

deeky666: So this is a "yes" let's put the SQL directly into the schema manager?

lib/Doctrine/DBAL/Schema/SQLServerSchemaManager.php
deeky666: :+1:

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
Ocramius: Yes, let's stuff it in the schema manager, and add a comment about the decision

deeky666: :+1:

deeky666:
@Ocramius marking this as WIP for now as I still need to figure out how to make nasty Oracle pass here. I am testing against version 12 locally and there seem to be some difference to version 11 (CI).

deeky666:
Documented an issue revealed while testing with Orcale 12c in #2643 
The issue is unrelated and should not be fixed here.

deeky666:
@Ocramius ok patch is finished from my side now. You can review again. Some remarks:

Obviously I had to fix the exact same issue on Oracle to make the tests pass. Approach is the same. I also had to change the privileges when creating a new database (user) on Oracle so that the user is able to see the views used in `killUserSessions()` when force dropping a database. That should not be an issue though, as users created by DBAL are not used as users but as databases (that is still confusing after all). Please blame Oracle. Thanks. BYEEEEE!

Ocramius (approved):

Ocramius:
LGTM! 👍

Ocramius:
Backported into `2.5` via doctrine/dbal@c13ee0cc94d12b40ad246b80144d52226fefd4bb

PR #2638, opened 2017-02-06
Tear down active transactions in functional test cases

deeky666:
I recently discovered a hard to debug issue where I was writing functional tests and suddenly the testsuite was hanging. After hours of debugging I found out that a previous test, involving transactions, failed inside a transaction and was affecting all subsequent tests as the transaction was marked as "rollback only" and still active. As we are mainly using a shared connection for performance reasons, this can get quite nasty.
I added a `tearDown` method to the base functional test case that checks, whether a transaction is active and rolls it back if that is the case.

### DIFF
--- a/tests/Doctrine/Tests/DbalFunctionalTestCase.php
+++ b/tests/Doctrine/Tests/DbalFunctionalTestCase.php
@@ -40,6 +40,13 @@ class DbalFunctionalTestCase extends DbalTestCase
         $this->_conn->getConfiguration()->setSQLLogger($this->_sqlLoggerStack);
     }
 
+    protected function tearDown()
+    {
+        while ($this->_conn->isTransactionActive()) {
+            $this->_conn->rollBack();
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


tests/Doctrine/Tests/DbalFunctionalTestCase.php
```diff
@@ -40,6 +40,13 @@ protected function setUp()
         $this->_conn->getConfiguration()->setSQLLogger($this->_sqlLoggerStack);
     }
 
+    protected function tearDown()
+    {
+        if ($this->_conn->isTransactionActive()) {
```
kimhemsoe: What happens if we have nested transactions?

deeky666: Fixed :+1:

Ocramius (approved):

Ocramius:
Backported to `2.5` via doctrine/dbal@eb0a49c203b5cb3b0fafb090e0b0b7dc36f6c695

👍

Issue #2613, opened 2017-01-24
Outdated VERSION constant value in 2.5 branch


mondrake:
Hi, the value of the VERSION constant in the 2.5 branch is still at 2.5.5-DEV while 2.5.10 is out.

```
    /**
     * Current Doctrine Version.
     */
    const VERSION = '2.5.5-DEV';

```

Ocramius:
Yarp, will try remembering it in the next release, sorry.

Closing here, can't do anything about past releases.

PR #2639, opened 2017-02-06
Update Version.php

mondrake:
The VERSION const is still outdated, after release 2.5.11 it is still at 2.5.5-DEV.

Was raised already in https://github.com/doctrine/dbal/issues/2613.


Ocramius:
LGTM. Also, I need more coffee when doing releases.

Ocramius (approved):

PR #2640, opened 2017-02-06
Fix drivers' `exec()` method to not execute queries via prepared statements

deeky666:
Even though the driver connection interface is not quite clear about this, the `exec()` method is obviously intended for DML/DDL statements to be directly executed without preparing the statement beforehand. It is not possible to bind parameters here and it will only ever return affected rows.
Now while nevertheless using prepared statements internally might not seem like such a big deal, there are some cons about this approach:

- SQL Server requires some DDL statements like creating/dropping temporary tables to be executed directly without prepared statements because of database internals (See: https://github.com/Microsoft/msphpsql/issues/276)
- Using prepared statements if not necessary creates unwanted overhead
- The behaviour is not consistent with the PDO implementations, which do not use prepared statement either if the driver in question is capable of doing so

I also updated the testsuite where invalid `executeQuery()` calls for DML/DDL statement where done.

This bugfix is required for the SQL Server testsuite to succeed. It makes 16 errors succeed on this branch with `sqlsrv`.

I did not add any additonal tests because it is not possible to mock whether prepared statements are used or not and also all of the various test cases are already covered by functional tests.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
+++ b/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
@@ -113,10 +113,13 @@ class DB2Connection implements Connection, ServerInfoAwareConnection
      */
     public function exec($statement)
     {
-        $stmt = $this->prepare($statement);
-        $stmt->execute();
+        $stmt = @db2_exec($this->_conn, $statement);
 
-        return $stmt->rowCount();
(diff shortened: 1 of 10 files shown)


Ocramius:
After going back and forth a few times, especially considering the test case changes, I'm going to merge this only into `2.6`

Ocramius:
@deeky666 👍

PR #2646, opened 2017-02-08
Fix the reuse of a statement on sqlsrv with explicit bindValue()

morozov:
If a value is bound with `bindValue()` between statement executions, the value used on a subsequent execution is ignored. The same happens with `bindParam()`. See the tests.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
+++ b/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
@@ -140,7 +140,14 @@ class SQLSrvStatement implements IteratorAggregate, Statement
      */
     public function bindValue($param, $value, $type = null)
     {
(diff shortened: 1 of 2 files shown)


morozov:
The proposed solution is more of a POC as it duplicates code and doesn't solve the second issue. I believe the proper one will not have these downsides.

deeky666:
@morozov would you be interested in joining the Doctrine core team? You did some very good quality additions to DBAL lately and we are still looking for support here :)

morozov:
@deeky666 it'd be my honor. Hopefully, my further contributions will not be that damaging :-).

deeky666:
@morozov hehe no worries, we're all humans ;) Can you join on IRC #doctrine-dev (freenode network) and throw a private message at me? I will tell you more then.

Ocramius:
@morozov welcome to the team! Meanwhile, this is being merged :-)

Ocramius (approved):

PR #2647, opened 2017-02-08
[DBAL-2644] Fix fetchAll(PDO::FETCH_COLUMN) on DB2 with Portability wrapper

### LINKED ISSUE #2644: Portability\Statement incorrectly handles fetchAll(PDO::FETCH_COLUMN)
The `Portability\Statement::fixRow()` implementation doesn't take into account that `mixed $row` is not always iterable (particularly in the case if `$fetchMode = PDO::FETCH_COLUMN`).

```php
use Doctrine\DBAL\Portability\Connection;

$conn = DriverManager::getConnection(array(
    'driver' => 'ibm_db2',
    'wrapperClass' => Connection::class,
    'portability' => Connection::PORTABILITY_FIX_CASE,
    'fetch_case' => PDO::CASE_LOWER,
));

$conn->executeQuery('SELECT id FROM users')
    ->fetchAll(\PDO::FETCH_COLUMN);
```
Produces:
> Warning: Invalid argument supplied for foreach() in vendor/doctrine/dbal/lib/Doctrine/DBAL/Portability/Statement.php on line 203

Besides that, the `Portability\Statement::connect()` implementation forcedly [replaces](https://github.com/doctrine/dbal/commit/daf4f16f597d4b358315ad0099ab200457e0e550#diff-a59fd02c89788e9db9c34b9b2b12b5e6R81) the `portability` value from whatever provided in the connection parameters with `PORTABILITY_DB2` which is inconsistent with the rest of the supported platforms.

morozov:
Fixes #2644:

1. Do not enforce platform-dependent portability settings
2. Properly handle the result of fetchAll(PDO::FETCH_COLUMN)

### DIFF
--- a/lib/Doctrine/DBAL/Portability/Connection.php
+++ b/lib/Doctrine/DBAL/Portability/Connection.php
@@ -71,11 +71,11 @@ class Connection extends \Doctrine\DBAL\Connection
                 } elseif ($this->getDatabasePlatform()->getName() === "sqlite") {
                     $params['portability'] = $params['portability'] & self::PORTABILITY_SQLITE;
                 } elseif ($this->getDatabasePlatform()->getName() === "drizzle") {
-                    $params['portability'] = self::PORTABILITY_DRIZZLE;
+                    $params['portability'] = $params['portability'] & self::PORTABILITY_DRIZZLE;
                 } elseif ($this->getDatabasePlatform()->getName() === 'sqlanywhere') {
-                    $params['portability'] = self::PORTABILITY_SQLANYWHERE;
+                    $params['portability'] = $params['portability'] & self::PORTABILITY_SQLANYWHERE;
                 } elseif ($this->getDatabasePlatform()->getName() === 'db2') {
-                    $params['portability'] = self::PORTABILITY_DB2;
+                    $params['portability'] = $params['portability'] & self::PORTABILITY_DB2;
                 } elseif ($this->getDatabasePlatform()->getName() === 'mssql') {
                     $params['portability'] = $params['portability'] & self::PORTABILITY_SQLSRV;
                 } else {
--- a/lib/Doctrine/DBAL/Portability/Statement.php
+++ b/lib/Doctrine/DBAL/Portability/Statement.php
@@ -175,8 +175,20 @@ class Statement implements \IteratorAggregate, \Doctrine\DBAL\Driver\Statement
             return $rows;
         }
 
+        if ($fetchMode === PDO::FETCH_COLUMN) {
+            foreach ($rows as $num => $row) {
+                $rows[$num] = array($row);
+            }
+        }
+
         foreach ($rows as $num => $row) {
             $rows[$num] = $this->fixRow($row, $iterateRow, $fixCase);
+        }
+
+        if ($fetchMode === PDO::FETCH_COLUMN) {
+            foreach ($rows as $num => $row) {
+                $rows[$num] = $row[0];
+            }
         }
 
         return $rows;
--- a/tests/Doctrine/Tests/DBAL/Functional/PortabilityTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/PortabilityTest.php
@@ -144,4 +144,39 @@ class PortabilityTest extends \Doctrine\Tests\DbalFunctionalTestCase
 
         $this->assertEquals($portability, $connection->getPortability());
     }
+
+    /**
+     * @dataProvider fetchAllColumnProvider
+     */
+    public function testFetchAllColumn($field, array $expected)
+    {
+        $conn = $this->getPortableConnection();
+        $stmt = $conn->query('SELECT ' . $field . ' FROM portability_table');
+
+        $column = $stmt->fetchAll(PDO::FETCH_COLUMN);
+        $this->assertEquals($expected, $column);
+    }
+
+    public static function fetchAllColumnProvider()
+    {
+        return array(
+            'int' => array(
+                'Test_Int',
+                array(1, 2),
+            ),
+            'string' => array(
+                'Test_String',
+                array('foo', 'foo'),
+            ),
+        );
+    }
+
+    public function testFetchAllNullColumn()
+    {
+        $conn = $this->getPortableConnection();
+        $stmt = $conn->query('SELECT Test_Null FROM portability_table');
+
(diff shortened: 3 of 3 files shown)


lib/Doctrine/DBAL/Portability/Statement.php
```diff
@@ -175,8 +175,20 @@ public function fetchAll($fetchMode = null, $columnIndex = 0)
             return $rows;
         }
 
-        foreach ($rows as $num => $row) {
-            $rows[$num] = $this->fixRow($row, $iterateRow, $fixCase);
+        if ($fetchMode === PDO::FETCH_COLUMN) {
```
Ocramius: @morozov can you isolate the contents of the `if` and `else` statements into private methods? Early return preferred

lib/Doctrine/DBAL/Portability/Statement.php
```diff
@@ -175,8 +175,20 @@ public function fetchAll($fetchMode = null, $columnIndex = 0)
...
+            if ($iterateRow) {
+                foreach ($rows as $num => $value) {
+                    if (($this->portability & Connection::PORTABILITY_EMPTY_TO_NULL) && $value === '') {
```
Ocramius: The first bitwise expression does not need re-computing within the loop, and should be moved to a variable

lib/Doctrine/DBAL/Portability/Statement.php
```diff
@@ -175,8 +175,20 @@ public function fetchAll($fetchMode = null, $columnIndex = 0)
...
+                        $rows[$num] = null;
+                    } elseif (($this->portability & Connection::PORTABILITY_RTRIM) && is_string($value)) {
```
Ocramius: The first bitwise expression does not need re-computing within the loop, and should be moved to a variable

Ocramius (changes requested):

lib/Doctrine/DBAL/Portability/Statement.php
```diff
@@ -169,14 +169,14 @@ public function fetchAll($fetchMode = null, $columnIndex = 0)
...
+        $rTrim = (bool) ($this->portability & Connection::PORTABILITY_RTRIM);
         $fixCase = !is_null($this->case) && ($fetchMode == PDO::FETCH_ASSOC || $fetchMode == PDO::FETCH_BOTH) && ($this->portability & Connection::PORTABILITY_FIX_CASE);
-        if ( ! $iterateRow && !$fixCase) {
-            return $rows;
-        }
 
-        foreach ($rows as $num => $row) {
-            $rows[$num] = $this->fixRow($row, $iterateRow, $fixCase);
+        if ($fetchMode === PDO::FETCH_COLUMN) {
+            $rows = $this->fixColumn($rows, $emptyToNull, $rTrim);
```
Ocramius: return early please

lib/Doctrine/DBAL/Portability/Statement.php
```diff
@@ -169,14 +169,14 @@ public function fetchAll($fetchMode = null, $columnIndex = 0)
...
+        } else {
```
Ocramius: No need for `else`

lib/Doctrine/DBAL/Portability/Statement.php
```diff
@@ -213,6 +213,55 @@ protected function fixRow($row, $iterateRow, $fixCase)
...
+     * Fixes column values according to portability settings
+     *
+     * @param mixed[] $values Column values indexed by row number
+     * @param bool $emptyToNull Should empty strings be converted to NULL
+     * @param bool $rTrim Should strings be right-trimmed
+     * @return mixed[]
+     */
+    private function fixColumn(array $values, $emptyToNull, $rTrim)
+    {
+        if (!$emptyToNull && !$rTrim) {
```
Ocramius: Isn't it simpler to loop twice, and call two separate methods that do one the `emptyToNull` fixing, and the other the `rTrim` fixing?

Ocramius (changes requested):

lib/Doctrine/DBAL/Portability/Statement.php
morozov: I think ideally (if [performance allows](http://stackoverflow.com/questions/18144782/performance-of-foreach-array-map-with-lambda-and-array-map-with-static-function/26527704#26527704)), we could have just 2-3 lower level functions which convert keys and values, and combine them together in reusable combinations based on `fetch*(PDO::FETCH_*)`. But let's try your suggestion first.

deeky666: Another idea that comes to mind is to normalize the rows at the beginning if `PDO::FETCH_COLUMN` is given and denormalize again at the end? Like expand the rows to a multi-dimensional array at the beginning and flatten again after portability adjustments? Not sure if that might be a better solution.

morozov: I think it'd look hacky but definitely save some lines of code.

deeky666: The most important thing is it saves a lot of complexity and function calls which only serve for 1 use-case. But as I said, not sure about that, it is just an idea that came to my mind.

morozov:
@Ocramius are you okay with the current approach?

Off-topic: I tried putting together an [alternative implementation](https://github.com/doctrine/dbal/compare/master...morozov:portability-fun) which uses a functional approach to converting data and avoids code duplication, but… for some reason, there's even more code there than in the original one 😕

Ocramius:
@morozov I'm fine with it, and merging :-)

lib/Doctrine/DBAL/Portability/Connection.php
```diff
@@ -71,11 +71,11 @@ public function connect()
                 } elseif ($this->getDatabasePlatform()->getName() === "sqlite") {
                     $params['portability'] = $params['portability'] & self::PORTABILITY_SQLITE;
                 } elseif ($this->getDatabasePlatform()->getName() === "drizzle") {
```
Ocramius: Note for the next person that has to maintain this: we're sorry about this massive chain of doom and `elseif` :-(

Ocramius:
Backported to `2.5` via doctrine/dbal@5a4ff4673c98f18fc034c7222ce2d4b1481d4565

PR #2648, opened 2017-02-08
Fix last insert ID consistency accross drivers

deeky666:
This one came while trying to fix the last remaining testsuite failure on `sqlsrv`.
Retrieving the last insert ID for non-sequences is highly unreliable accross the driver implementations (as also stated in the PDO docs). This PR makes all drivers behave the same (which support this feature) with two exceptions (shown later on).

To demonstrate some inconsistency examples:

- MySQL drivers reset the last insert ID to `0` after each non-insert query
- `mysqli` and `sqlanywhere` return `int` instead of `string`
- `sqlsrv` returns `null` instead of `0` if no INSERT was been executed, yet
- `ibm_db2` returns `` instead of `0` if no INSERT was been executed, yet
- `sqlsrv` only records the last insert ID on prepared statements (not on exec) because of custom implementation via object
- `pdo_pgsql` resets the last insert ID to `0` after dropping the particular table the last insert ID was retrieved from
- `oci8` always returns `false` instead of `0` (not supported on Oracle)
- Surely others I forgot right now...

Looking at our interface for `lastInsertId()` drivers always have to return `string`, no word about exceptions or what to return if not supported or no ID available.
To stick to the contract and finding a reasonable common denominator, I chose MySQL as a base for returning "I don't know the value". MySQL returns `0` in this case. Otherwise MySQL behaves different when it comes to keeping the last insert ID for the connection (resets to `0` while the others don't).

Two exceptions that cannot be made consistent:

- SQLite uses a different concept when it comes to reusing rolled back insert IDs (see tests)
- `pdo_pgsql` returns `4294967295` if no insert ID is available yet (weird thing, don't know what PHP intending here)

I tried to cover the most important situations in the tests, hope I did not forget anything.


lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php
```diff
@@ -139,7 +139,10 @@ public function lastInsertId($name = null)
             return $stmt->fetchColumn();
         }
 
-        return $this->lastInsertId->getId();
```
deeky666: I wanted to remove the custom object altogether but unfortunately the property is `protected` on the connection so I kept it here for BC.

lib/Doctrine/DBAL/Driver/PDOConnection.php
```diff
@@ -120,7 +125,19 @@ public function quote($input, $type = \PDO::PARAM_STR)
...
-        return parent::lastInsertId($name);
+        if (null !== $name) {
+            return parent::lastInsertId($name);
+        }
+
+        $lastInsertId = parent::lastInsertId($name);
+
+        // The last insert ID is reset to "0" in certain situations by some implementations,
+        // therefore we keep the previously set insert ID locally.
+        if ($lastInsertId) {
```
Ocramius: Is an empty string a valid generated identifier? I think this conditional should be checking for `'0' !== $lastInsertId`

Ocramius: Also introduces the issue in scenarios where sequences actually contain `'0'` as valid insert ID at some point.

Ocramius (approved):

lib/Doctrine/DBAL/Driver/PDOConnection.php
Ocramius: These are just edge cases that should really not be considered, IMO, but it's good to comment on them, at least.

Ocramius:
While the fix is perfectly valid, and our API documents the behavior correctly, people may (!!!) currently rely on the last insert id being an `int`. 

Therefore, this will go only to `2.6` due to regression risks in consumer code.

Ocramius:
I restarted the CI run, but postgresql + PHP nightly seem to have some issues with [fetching the `lastval`](https://travis-ci.org/doctrine/dbal/builds/199552265)

lib/Doctrine/DBAL/Driver/PDOConnection.php
deeky666: @Ocramius this PR does not touch sequence ID retrieval (which requires passing `$name`). So this is all about regular "autoincrement" via identity columns. I am not aware of the possibility to have an identity column generate an empty string or `0` (they start at `1`, don't they)?

Ocramius: Yes, sequences are indeed not affected, but we still need to consider scenarios where something like an `AUTO_INCREMENT` started at `-10`, and then at some point reaches `0`. We can't handle that, but we have to at least specify that to users (we only support positive insert identifiers, starting from `1`, while all, other scenarios lead to edge cases)

deeky666:
@Ocramius could you do me the favour and have a look at the current approach (no nitpicking, it's not polished yet) and tell me what you think about it? Concerns, optimizations, different ideas welcome. The only related failing test ist [this one](https://travis-ci.org/doctrine/dbal/jobs/199950370#L314) which is expected (see remarks in the code). Don't know how to fix this. I guess we need to skip the test?

lib/Doctrine/DBAL/Driver/PDOConnection.php
```diff
@@ -30,6 +30,11 @@
 class PDOConnection extends PDO implements Connection, ServerInfoAwareConnection
 {
     /**
+     * @var string
+     */
+    private $lastInsertId = '0';
```
Ocramius: Why not reuse the `LastInsertId` object that we created?

lib/Doctrine/DBAL/Driver/PDOConnection.php
```diff
@@ -130,4 +168,47 @@ public function requiresQueryForServerVersion()
...
+     * Returns whether this PDO driver is able to fetch the last insert ID for identity columns
+     * after each executed statement(even if it might not have created an ID)
+     * without influencing connection state and transaction state.
+     *
+     * Drivers like "pdo_pgsql" for example cannot safely try to retrieve the last insert ID without causing an error
+     * that might cause active transactions to fail.
+     *
+     * @return bool
+     */
+    protected function supportsTrackingLastInsertId()
```
Ocramius: This can most probably be a constant or a protected property. Having it as a method seems excessive, and can lead to (although minor) overhead

Ocramius: Could also be a constant, btw.

Ocramius: Another (better) approach is to make `trackLastInsertId` empty in subclasses not supporting it. Thoughts?

lib/Doctrine/DBAL/Driver/PDOPgSql/Connection.php
```diff
@@ -0,0 +1,101 @@
...
+     * @var string
+     */
+    private $lastInsertId = '0';
+
+    /**
+     * @var bool
+     */
+    private $lastInsertIdCallNested = false;
+
+    /**
```
Ocramius: Please don't keep commented code

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -88,7 +98,13 @@ public function bindParam($column, &$variable, $type = \PDO::PARAM_STR, $length
     public function execute($params = null)
     {
         try {
-            return parent::execute($params);
+            $result = parent::execute($params);
+
+            if (null !== $this->connection) {
```
Ocramius: `if ($this->connection) {`

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php
```diff
@@ -124,7 +125,19 @@ public function exec($statement)
             throw SQLSrvException::fromSqlSrvErrors();
         }
 
-        return sqlsrv_rows_affected($stmt);
+        $affectedRowCount = sqlsrv_rows_affected($stmt);
+
+        $stmt = sqlsrv_query($this->conn, 'SELECT @@IDENTITY');
```
Ocramius: This seems excessive, as it duplicates queries to the DB. Since we can just `SELECT @@IDENTITY` anyway, do we need to update the last insert ID?

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
```diff
@@ -381,4 +373,21 @@ public function rowCount()
...
         return sqlsrv_rows_affected($this->stmt);
     }
+
+    private function trackLastInsertId()
+    {
+        if (! $this->lastInsertId) {
+            return;
+        }
+
+        $statement = sqlsrv_query($this->conn, 'SELECT @@IDENTITY');
```
Ocramius: Same as above

Ocramius (changes requested):

lib/Doctrine/DBAL/Driver/LastInsertId.php
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+namespace Doctrine\DBAL\Driver;
+
+/**
+ * Last insert ID container.
+ *
+ * @author Steve Müller <st.mueller@dzh-online.de>
+ */
+class LastInsertId
```
deeky666: :+1:

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -106,6 +113,8 @@ public function __construct(\mysqli $conn, $prepareString)
             $this->types = str_repeat('s', $paramCount);
             $this->_bindedValues = array_fill(1, $paramCount, null);
         }
+
+        $this->lastInsertId = $lastInsertId;
```
deeky666: That won't help as we need the reference in the connection. I just wanted to keep BC here, but not sure if that is necessary.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Connection.php
```diff
@@ -148,18 +148,20 @@ public function exec($statement)
     public function lastInsertId($name = null)
     {
         if ($name === null) {
-            return false;
+            return '0';
         }
 
         $sql    = 'SELECT ' . $name . '.CURRVAL FROM DUAL';
```
deeky666: :+1:

lib/Doctrine/DBAL/Driver/PDOConnection.php
deeky666: Because I need to know how to retrieve the last insert ID in the statement object? Simply passing the LastInsertId container won't help in `PDOStatement::execute()`

deeky666: Ah I see what you mean. Yeah true I can do that.

lib/Doctrine/DBAL/Driver/PDOConnection.php
deeky666: YES! The last suggestion sounds perfect. I will see if that works.

lib/Doctrine/DBAL/Driver/PDOPgSql/Connection.php
deeky666: I left it in here to demonstrate you what the problem with postgres is. I will of course remove it in the final version ;)

lib/Doctrine/DBAL/Driver/PDOStatement.php
deeky666: :+1: BTW I am still not using either style consistently. Do you have any reason why to prefer one over the other?

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php
deeky666: This is what I was talking about before showing off this approach ("performance impact"). If you have a look at how the different PDO drivers implement `lastInsertId()` you will see that each of them actually trigger a query. `pdo_sqlsrv` uses [the exact same](https://github.com/Microsoft/msphpsql/blob/a02691e277d5804c3d8b28fcdfee31e6fd2df8ab/source/pdo_sqlsrv/pdo_dbh.cpp#L31), `pdo_pgsql` uses `SELECT LASTVAL()`, `pdo_mysql` uses `SELECT LAST_INSERT_ID()`. However I am not sure but I believe those are session related information calls which do not go through the server at all. Would need to investigate this.

deeky666: And yes we need to update the ID in `exec()`, `query()` and `prepare()->execute()` as those are the methods you can use to execute queries on each driver.

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
deeky666: As I said, this in fact is what would normally be in `lastInsertId()`

lib/Doctrine/DBAL/Driver/PDOConnection.php
```diff
@@ -130,4 +166,49 @@ public function requiresQueryForServerVersion()
...
+    public function trackLastInsertId()
+    {
+        // We need to avoid unnecessary exception generation for drivers not supporting this feature,
+        // by temporarily disabling exception mode.
+        $this->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_SILENT);
+
+        $lastInsertId = $this->fetchLastInsertId();
+
+        // Reactivate exception mode.
+        $this->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_EXCEPTION);
```
deeky666: @Ocramius should we rather evaluate and set the previous error mode here? because in userland you could potentially change out default error mode.

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -28,10 +28,20 @@
...
+     * @var PDOConnection|null
+     */
+    private $connection;
+
+    /**
      * Protected constructor.
+     *
+     * @param PDOConnection|null $connection
+     *
+     * @todo do we need to keep BC here? Or can we make $connection mandatory?
```
deeky666: @Ocramius no decision done here so far. What do you think?

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
```diff
@@ -128,11 +128,7 @@ public function __construct($conn, $sql, LastInsertId $lastInsertId = null)
     {
         $this->conn = $conn;
         $this->sql = $sql;
-
-        if (stripos($sql, 'INSERT INTO ') === 0) {
-            $this->sql .= self::LAST_INSERT_ID_SQL;
```
deeky666: @Ocramius do we need to keep the constant for BC?

lib/Doctrine/DBAL/Driver/LastInsertId.php
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * @param string $value
+     */
+    public function set($value)
+    {
+        // The last insert ID is reset to "0" in certain situations by some implementations,
+        // therefore we keep the previously set insert ID locally.
+        if ('0' !== $value) {
+            $this->value = $value;
```
Ocramius: Should probably apply a string cast, since we are not using PHP7's type hints.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
Ocramius: Yes, that's necessary here, sadly.

lib/Doctrine/DBAL/Driver/PDOConnection.php
Ocramius:
~~~We don't support changing error mode in DBAL anyway. Everything breaks down on our side if the mode is not `\PDO::ERRMODE_EXCEPTION`, so I wouldn't bother here.~~~

Scratch that - using `PDO#getAttribute()` is simple/easy here - anything against using it?

Ocramius: Can `fetchLastInsertId` ever fail? If so, this block should be in a `finally {}`

lib/Doctrine/DBAL/Driver/PDOConnection.php
```diff
@@ -130,4 +166,49 @@ public function requiresQueryForServerVersion()
     {
         return false;
     }
+
+    /**
+     * Tracks the last insert ID at the current state.
```
Ocramius: This method should be annotated with `@internal`

lib/Doctrine/DBAL/Driver/PDOStatement.php
Ocramius: We have to maintain BC in my opinion, otherwise we may break existing manual DIC calls.

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -102,7 +112,13 @@ public function closeCursor()
     public function execute($params = null)
     {
         try {
-            return parent::execute($params);
+            $result = parent::execute($params);
+
+            if (! $this->connection) {
```
Ocramius: This conditional seems wrong: you only track if there's no connection?

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php
Ocramius: > However I am not sure but I believe those are session related information calls which do not go through the server at all. Would need to investigate this.

This might cause an uproar of DBAs if they actually hit the backend. If only some exotic drivers (*cough* sqlsrv *cough*) actually hit the backend, then we should be quite OK.

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
Ocramius: The constant has to stay, but should be deprecated if we don't use it. This is where private constants will be vital once we switch to PHP 7.1

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
Ocramius: Looks OK then.

Ocramius (changes requested):

lib/Doctrine/DBAL/Driver/LastInsertId.php
Ocramius: No longer needed, as we can rely on PHP 7.1 now.

lib/Doctrine/DBAL/Driver/PDOConnection.php
Ocramius: Done

lib/Doctrine/DBAL/Driver/LastInsertId.php
Ocramius: No need: handled via addition of a type hint

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
Ocramius: Done

lib/Doctrine/DBAL/Driver/PDOConnection.php
Ocramius: Done

lib/Doctrine/DBAL/Driver/PDOConnection.php
Ocramius: Done both (restoring + `finally`)

lib/Doctrine/DBAL/Driver/PDOStatement.php
Ocramius: This comment is invalid - the connection passed in as constructor is a design decision, and nullability is just a side-effect of BC compliance. This code should never be skipped inside DBAL, but it could be skipped if the class is used directly in userland.

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php
Ocramius: Since this is an SQLSrv-only problem, this comment can be ignored.

IMO: their problem for relying on SQLSrv.

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
Ocramius: Done

Ocramius:
Moved to #2765

PR #2650, opened 2017-02-09
Fix result cache and PDO connection test on pdo_sqlsrv

deeky666:
The story continues...

This fixes two test cases where `pdo_sqlsrv` fails.

**ResultCacheTest**
```
Exception: [Doctrine\DBAL\DBALException] An exception occurred while executing 'DELETE FROM caching':

SQLSTATE[HY010]: [unixODBC][Driver Manager]Function sequence error
```

Each test errors with that message. It is caused by the `setUp()` which tries to create the test table and silences exceptions. There seems to be a bug in the driver, that leaves the connection in an erroneous state when managing exceptions in userland: https://github.com/Microsoft/msphpsql/issues/49#issuecomment-137774580

To make the tests pass for now, we correctly create the table on setup and drop it on teardown now.

**PDOConnectionTest**
```
1) Doctrine\Tests\DBAL\Functional\Driver\PDOConnectionTest::testThrowsWrappedExceptionOnPrepare
Failed asserting that exception of type "PDOException" matches expected exception "\Doctrine\DBAL\Driver\PDOException". Message was: "SQLSTATE[IMSSP]: The given attribute is only supported on the PDOStatement object." at
/php/srv/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOConnectionTest.php:69
```

`pdo_sqlsrv` for some reason only supports setting `PDO::ATTR_EMULATE_PREPARES` on the statement object but not on the connection object, which makes the test useless (so skipping that one).

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOConnectionTest.php
@@ -64,6 +64,10 @@ class PDOConnectionTest extends DbalFunctionalTestCase
      */
     public function testThrowsWrappedExceptionOnPrepare()
     {
+        if ($this->_conn->getDriver()->getName() === 'pdo_sqlsrv') {
+            $this->markTestSkipped('pdo_sqlsrv does not allow setting PDO::ATTR_EMULATE_PREPARES at connection level.');
+        }
+
         // Emulated prepared statements have to be disabled for this test
         // so that PDO actually communicates with the database server to check the query.
         $this->driverConnection->setAttribute(\PDO::ATTR_EMULATE_PREPARES, false);
--- a/tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
@@ -16,19 +16,14 @@ class ResultCacheTest extends \Doctrine\Tests\DbalFunctionalTestCase
     {
         parent::setUp();
 
-        try {
-            /* @var $sm \Doctrine\DBAL\Schema\AbstractSchemaManager */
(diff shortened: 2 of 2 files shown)


tests/Doctrine/Tests/DBAL/Functional/Driver/PDOConnectionTest.php
```diff
@@ -64,9 +64,13 @@ public function testThrowsWrappedExceptionOnExec()
...
     public function testThrowsWrappedExceptionOnPrepare()
     {
+        if ($this->_conn->getDriver()->getName() === 'pdo_sqlsrv') {
+            $this->markTestSkipped('pdo_sqlsrv does not allow setting PDO::ATTR_EMULATE_PREPARES at connection level.');
+        }
+
         // Emulated prepared statements have to be disabled for this test
         // so that PDO actually communicates with the database server to check the query.
-        $this->driverConnection->setAttribute(\PDO::ATTR_EMULATE_PREPARES, false);
+        //$this->driverConnection->setAttribute(\PDO::ATTR_EMULATE_PREPARES, false);
```
Ocramius: Is this supposed to be commented out?

Ocramius (changes requested):

tests/Doctrine/Tests/DBAL/Functional/Driver/PDOConnectionTest.php
deeky666: No, good catch! Thanks you

deeky666: Fixed.

Ocramius (approved):

Ocramius:
Backported to `2.5` via doctrine/dbal@0c6eda3cab3e0701a7b68bd3e7545f6f4d2372f0

PR #2651, opened 2017-02-09
Fix closing statement cursor on pdo_sqlsrv if not executed yet

deeky666:
`pdo_sqlsrv` does not strictly stick to the PDO interface, yet and throws and exception when trying to close a statement cursor if it has not been execute yet:

```
Exception: [PDOException] SQLSTATE[IMSSP]: The statement must be executed before results can be retrieved.
```

A patch is in the works, though: https://github.com/Microsoft/msphpsql/issues/267

Despite that, we should try to make the behaviour consistent for all kinds of PDO drivers, even those that we might not support in the core. Therefore silencing execptions in `closeCursor()`.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/PDOStatement.php
+++ b/lib/Doctrine/DBAL/Driver/PDOStatement.php
@@ -85,6 +85,20 @@ class PDOStatement extends \PDOStatement implements Statement
     /**
      * {@inheritdoc}
      */
+    public function closeCursor()
+    {
+        try {
+            return parent::closeCursor();
+        } catch (\PDOException $exception) {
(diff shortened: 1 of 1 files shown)


deeky666:
Btw there are already tests that reveal this issue: `Doctrine\Tests\DBAL\Functional\StatementTest`

Ocramius (approved):

Ocramius:
Backported to `2.5` via doctrine/dbal@9d633dadfae26d0a45ce432c86a74e6badff8b8b

morozov:
Should be fixed by https://github.com/Microsoft/msphpsql/issues/267.

PR #2655, opened 2017-02-14
Simplify Travis-CI build matrix

photodude:
New simpler Travis build matrix with a few selective addons

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,22 +16,21 @@ env:
   - DB=sqlite
   - DB=mysql
   - DB=mysqli
+  - DB=mariadb MARIADB_VERSION=10.0
+  - DB=mariadb MARIADB_VERSION=10.1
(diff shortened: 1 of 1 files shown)


Ocramius:
LGTM 👍

Ocramius (approved):

deeky666:
@photodude are you sure this is still working? Having a look at [this line](https://travis-ci.org/doctrine/dbal/jobs/201605852#L203), it does not seem that Travis is actually installing anything. The same goes for MariaDB. Also why is `sudo` suddenly allowed in non-sudo (docker) environments on Travis?

photodude:
@deeky666 You are correct it's not working. This will need to be reverted as sudo doesn't work (bad suggestion from travis). I've contacted travis to see what they suggest now.

(sometimes a green light is an unfortunate appearance of a passing build)

PR #2658, opened 2017-02-21
Making the `DBALException` implement `Throwable`

svycka:
`Doctrine\DBAL\DBALException` does not implement `\Exception`

### DIFF
--- a/lib/Doctrine/DBAL/Driver/DriverException.php
+++ b/lib/Doctrine/DBAL/Driver/DriverException.php
@@ -29,7 +29,7 @@ namespace Doctrine\DBAL\Driver;
  * @link   www.doctrine-project.org
  * @since  2.5
  */
-interface DriverException
+interface DriverException extends \Throwable
 {
     /**
      * Returns the driver specific error code if available.


lib/Doctrine/DBAL/Exception/DriverException.php
```diff
@@ -45,13 +45,7 @@ class DriverException extends DBALException
      */
     public function __construct($message, \Doctrine\DBAL\Driver\DriverException $driverException)
     {
-        $exception = null;
-
-        if ($driverException instanceof \Exception) {
```
deeky666: This makes perfect sense because we are type hinting against an interface in the constructor and need to make sure that we really pass an exception to the parent constructor (IIRC).

Ocramius: @deeky666 we should probably make `DriverException` inherit from `Throwable`

deeky666: Sounds reasonable, yes. I guess it was just a sanity check here for PHP < 7

Ocramius: @svycka can you add the `extends` on `DriverException` then?

svycka: hmm I read it wrong :D

svycka: @Ocramius but a bit don't udnerstand if I add interface DriverException extends \Throwable It would still be not possible to pass `parent::__construct($message, 0, $driverException);` because cosntructor expects `\Exception` not `\Throwable`

svycka:
omg I will have to clone it now :D

svycka:
ok not sure what happened :D @Ocramius isn't this what you wanted?

svycka:
is this phpunit limitation?

svycka:
ok it looks tat php does not allow implement \Throwable so closing

Ocramius:
@svycka patch looked correct: you can only implement `Throwable` by extending an `Exception` though. Our codebase does that, but mocking doesn't. We'd probably drop tests mocking this interface.

photodude:
FYI you should rebase so travis will correctly test your PR

Ocramius:
This change goes in as-is: patch is correct and we now enforce PHP 7.1 anyway. :+1:

PR #2660, opened 2017-02-21
Revert PgSQL and MariaDB Travis-CI build matrix simplification

photodude:
Removes all sudo items (except hhvm which currently still requires sudo)
reverting the other simpler matrix attempt at #2655

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: php
-
 sudo: false
 
 cache:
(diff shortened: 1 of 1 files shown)


deeky666:
@photodude LGTM. One question though: Is there a reason we allow failures for PHP-nightly + PostgreSQL 9.5 / 9.6? Because it seems they are succeeding.

Ocramius:
@deeky666 we shouldn't, but @mbeccati is helping with `master` tracking of doctrine support, and reports to php-internals if something breaks badly.

deeky666:
@Ocramius so we might disallow failures here, right? If so, can you change this here @photodude ? We can merge afterwards.

photodude:
Nightly is set correctly now

deeky666 (approved):

photodude:
you are welcome, I hope to figure out the whole conditional addons in container at some point (if it is even possible) Maybe after trusty container is publicly released (still waiting for them to fix the mysql socket issues).

deeky666:
I agree it's not the most elegant setup atm but at least it is working for now. Let's see what Travis offers us in the future ;)

PR #2664, opened 2017-02-25
Add oci8 persistent connection support to the configuration documentation

mathieubouchard:
The oci8 driver can connect using oci_pconnect to make persistent connections. Doctrine has support for this if the persistent boolean value is set to true in the configuration. However, this information was not documented.

### DIFF
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -305,6 +305,7 @@ pdo\_oci / oci8
    you will still need to provide the ``user`` and ``password`` parameters, but the other
    parameters will no longer be used. Note that when using this parameter, the ``getHost``
    and ``getPort`` methods from ``Doctrine\DBAL\Connection`` will no longer function as expected.
+-  ``persistent`` (boolean): Whether to establish a persistent connection.
 
 pdo\_sqlsrv / sqlsrv
 ^^^^^^^^^^^^^^^^^^^^


docs/en/reference/configuration.rst
```diff
@@ -305,6 +305,7 @@ pdo\_oci / oci8
    you will still need to provide the ``user`` and ``password`` parameters, but the other
    parameters will no longer be used. Note that when using this parameter, the ``getHost``
    and ``getPort`` methods from ``Doctrine\DBAL\Connection`` will no longer function as expected.
+-  ``persistent`` (boolean): Whether to establish a persistent connection.
```
Ocramius: Please indent as in previous bullet points

Ocramius (changes requested):

docs/en/reference/configuration.rst
mathieubouchard: I don't understand, it has the same indentation here. And the generated HTML has exactly the same structure: https://github.com/mathieubouchard/dbal/blob/master/docs/en/reference/configuration.rst

Ocramius: Urgh, sorry for that, the spacing is shown differently on github for mobile devices - my bad :-(

Ocramius (approved):

Ocramius:
@mathieubouchard merging, thanks! 👍

Issue #2576, opened 2016-12-13
Update schema has 2 time the same "drop foreign key"


Nikoms:
Hello!

On mysql, I have this issue with 2.5.5.

When I removed a Foreign key to a simple int, I get the "drop" twice.

ALTER TABLE my_table DROP FOREIGN KEY FK_CE1415A19EB6921;
ALTER TABLE my_table DROP FOREIGN KEY FK_CE1415A19EB6921;

I already checked a bit.
And I see here that we merge some sqls: doctrine/dbal/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php:674

$this->getPreAlterTableAlterIndexForeignKeySQL($diff)
AND
parent::getPreAlterTableIndexForeignKeySQL($diff),

both return the "ALTER TABLE ... DROP FOREIGN FEY"

Maybe one of them should not return the "ALTER TABLE", but which one? or I guess we should add a array_unique? 

If someone tell me, i can make a PR :)

Ocramius:
@Nikoms is this specific to 2.5.5, as in "worked fine on 2.5.4"?

Nikoms:
We were in 2.5.1. I'll try to test from version to version to know which version introduces this bug.

Nikoms:
I can say that it's from 2.4.5 to 2.5.0

Like I said, it worked with 2.4.5 which is the last 2.4 version. And I checked with 2.5.0 and it does not work

Ocramius:
Argh, sorry, my confusion. I thought this was the ORM repository 😒

Nikoms:
Np :)

Ocramius:
> Maybe one of them should not return the "ALTER TABLE", but which one? or I guess we should add a array_unique?

Before looking at a solution, we need to have a test case reproducing the issue.

Could you try creating one?

websmurf:
Hi @Ocramius 

Running into the same issue, have set up a repo for it here: https://github.com/websmurf/doctrine-foreign-key-duplicate

Let me know if you need more.

Ocramius:
@websmurf needs a test case running in our test suite.

websmurf:
Hi @Ocramius, took me a while to rebuild the exact TableDiff structure, but I've committed it in a fork here: https://github.com/websmurf/dbal/blob/master/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php#L894

Start it using:
[...]

PR #2667, opened 2017-03-01
Test and fix for duplicate foreign key entries

websmurf:
In relation to issue #2576 I've created a test that covers this scenario

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
@@ -3,12 +3,14 @@
 namespace Doctrine\Tests\DBAL\Platforms;
 
 use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Schema\Column;
 use Doctrine\DBAL\Schema\Comparator;
 use Doctrine\DBAL\Schema\ForeignKeyConstraint;
 use Doctrine\DBAL\Schema\Index;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Schema\TableDiff;
 use Doctrine\DBAL\TransactionIsolationLevel;
+use Doctrine\DBAL\Types\Type;
 use function array_shift;
 
 abstract class AbstractMySQLPlatformTestCase extends AbstractPlatformTestCase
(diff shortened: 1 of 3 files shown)


websmurf:
@Ocramius don't know why the tests have failed exactly btw, but it does not seem to relate to code change errors based on what I can find in travis logs

tests/Doctrine/Tests/DBAL/Platforms/MySQL57PlatformTest.php
```diff
@@ -82,4 +82,22 @@ protected function getGeneratesAlterTableRenameIndexUsedByForeignKeySQL()
...
+    public function testNotDuplicateDropForeignKeySql()
+    {
+        $diff = $this->createNotDuplicateDropForeignKeySql();
+
+        // Run through alter table method
+        $sql = $this->_platform->getAlterTableSQL($diff);
+
+        // Assert that there are no duplicates in the results
+        $this->assertSame([
+            'ALTER TABLE documents DROP FOREIGN KEY documents_ibfk_1',
```
morozov: Given that the set of expected statements is different depending on the platform, I believe this feature should be tested in a functional way (statements are executed on a real DB) rather on the SQL level. This way, every DB platform will _prove_ the correctness of the generated statements.

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
```diff
@@ -946,4 +948,75 @@ public function testGetCreateTableSQLWithColumnCollation() : void
...
+
+        // Create documents table
+        $documents = new Table('documents');
+        $documents->addColumn('id', 'integer');
+        $documents->addColumn('office_id', 'integer');
+        $documents->addForeignKeyConstraint('offices', [ 'office_id' ], [ 'id' ], [ 'onDelete' => 'CASCADE' ], 'FK_a788692ffa0c224');
+        $documents->setPrimaryKey([ 'id' ]);
+
+        // Added foreign key
+        $diff->addedForeignKeys['FK_a788692ffa0c224'] = $documents->getForeignKey('FK_a788692ffa0c224');
```
morozov: This is not what real DBAL usage looks like. The comparator should be used to produce the diff of two given schemas, the diff shouldn't be built in parallel with the schema(s).

morozov (changes requested):

PR #2670, opened 2017-03-09
Slash before '_' in postgresql like condition with 'pg_' prefix to prevent confusion with the wildcard character

blackbjorn:
There are conditions `schema not like 'pg_%'` for exclude system tables, sequence etc. (system prefix `pg_`)  in some queries for postgresql platform. 
But in like condition symbol `_` means "any single character" ([docs](https://www.postgresql.org/docs/current/static/functions-matching.html#FUNCTIONS-LIKE))
It's the cause of problems when schema name starts with `pg<non underscore character>` (`pgp` for example), but it's valid schema name.


deeky666:
@blackbjorn thx!

deeky666:
Backported to `2.5` via https://github.com/doctrine/dbal/commit/31db22630f925088a58a1d33bb4e401d3913f2d1

PR #2671, opened 2017-03-09
DBAL-990 Attempt platform detection even when the database name is not set

### LINKED ISSUE #351: Doctrine DBAL 2.5 Breaks Symfony if DB Doesn't Exist
After updating to DBAL 2.5, Doctrine introduced `Doctrine\DBAL\Connection::detectDatabasePlatform()` which is different than 2.4. Now in 2.5, any time a repository is instantiated (not even used) by the DI, we get a PDOException if the DB doesn't already exist.

Imagine trying to run `app/console doctrine:database:create` and getting a "Database doesn't exist!" error. Repositories as services are very common in most projects, so this is a huge issue.

To illustrate quickly, create a fresh Symfony project, letting composer install the AcmeDemoBundle. Then, while trying to generate an entity, you'll get the PDOException.

```
composer create-project symfony/framework-standard-edition symfony-standard
cd symfony-standard
app/console doctrine:generate:entity --entity=AcmeDemoBundle:Foo
```

```
  [Doctrine\DBAL\Exception\ConnectionException]
  An exception occured in driver: SQLSTATE[HY000] [1049] Unknown database 'testdb'

  [Doctrine\DBAL\Driver\PDOException]
  SQLSTATE[HY000] [1049] Unknown database 'testdb'

  [PDOException]
  SQLSTATE[HY000] [1049] Unknown database 'testdb'
```

### LINKED ISSUE #990: DBAL-1057: Connection is not lazy anymore when guessing the platform is necessary
Jira issue originally created by user stof:

In DBAL 2.5, many driver can rely on different versions of the platform. Unless the version is explicitly provided, the driver will guess it at instantiation time, killing the lazyness of the connection.
This is a critical issue in any context using DI as it means that injecting the connection into anything else will connect to the server.

### LINKED ISSUE #1109: DBAL-1167: [GH-814] allow serverVersion to be unspecified
Jira issue originally created by user @doctrinebot:

This issue is created automatically through a Github pull request on behalf of BreiteSeite:

Url: https://github.com/doctrine/dbal/pull/814

Message:

Hi,

this PR allows `serverVersion`to be nullable.

We use doctrine/dbal in Integration-Tests and to prevent the unit-tests from connecting to a database, we specify `serverVersion` with something made-up (like `5.6`).

However, i'm not comfortable set `serverVersion` to a made-up server-version to prevent connections from being made, when there even isn't a server to connect to. With this PR merged, we could specify `serverVersion` to be `null` instead of something like `5.6` and still prevent connections from being made. `null` is a valid return value for `getDatabasePlatformVersion()`.

deeky666:
Fixes #990 
Fixes https://github.com/doctrine/DoctrineBundle/issues/351
Closes #814 
Closes #1109 
and maybe more...

The long lasting problem with the connection not being lazy anymore and the database to connect to not existing (yet) will be fixed by this patch.

We now try to retrieve the platform version as usual. If connecting fails we try a fallback detection by unspecifying the database name and connecting again. If that works we fetch the platform version and disconnect again.

I think this extra connect operation is acceptable as it only occurs in case the database does not exist yet (edge case). So there should not be any noticeable performance impact.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -425,9 +425,51 @@ class Connection implements DriverConnection
 
         // If not connected, we need to connect now to determine the platform version.
         if (null === $this->_conn) {
-            $this->connect();
+            try {
+                $this->connect();
+            } catch (\Exception $originalException) {
+                if (empty($this->_params['dbname'])) {
+                    throw $originalException;
+                }
+
+                // The database to connect to might not yet exist.
+                // Retry detection without database name connection parameter.
+                $databaseName = $this->_params['dbname'];
+                $this->_params['dbname'] = null;
+
+                try {
+                    $this->connect();
+                } catch (\Exception $fallbackException) {
+                    // Either the platform does not support database-less connections
+                    // or something else went wrong.
+                    // Reset connection parameters and rethrow the original exception.
+                    $this->_params['dbname'] = $databaseName;
+
+                    throw $originalException;
+                }
+
+                // Reset connection parameters.
+                $this->_params['dbname'] = $databaseName;
+                $serverVersion = $this->getServerVersion();
+
+                // Close "temporary" connection to allow connecting to the real database again.
+                $this->close();
+
+                return $serverVersion;
+            }
+
         }
 
+        return $this->getServerVersion();
+    }
+
+    /**
+     * Returns the database server version if the underlying driver supports it.
+     *
+     * @return string|null
+     */
+    private function getServerVersion()
+    {
         // Automatic platform version detection.
         if ($this->_conn instanceof ServerInfoAwareConnection &&
             ! $this->_conn->requiresQueryForServerVersion()
--- a/tests/Doctrine/Tests/DBAL/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/ConnectionTest.php
@@ -12,6 +12,7 @@ use Doctrine\DBAL\Events;
 use Doctrine\Tests\Mocks\DriverConnectionMock;
 use Doctrine\Tests\Mocks\DriverMock;
 use Doctrine\DBAL\Cache\ArrayStatement;
+use Doctrine\Tests\Mocks\VersionAwarePlatformDriverMock;
 
 class ConnectionTest extends \Doctrine\Tests\DbalTestCase
 {
@@ -711,5 +712,30 @@ class ConnectionTest extends \Doctrine\Tests\DbalTestCase
             ArrayStatement::class,
             (new Connection($this->params, $driver))->executeCacheQuery($query, $params, $types, $queryCacheProfileMock)
         );
+    }
+
+    /**
+     * @group DBAL-990
+     */
+    public function testRethrowsOriginalExceptionOnDeterminingPlatformWhenConnectingToNonExistentDatabase()
+    {
+        /** @var \Doctrine\Tests\Mocks\VersionAwarePlatformDriverMock|\PHPUnit_Framework_MockObject_MockObject $driverMock */
+        $driverMock = $this->createMock(VersionAwarePlatformDriverMock::class);
+
+        $connection = new Connection(array('dbname' => 'foo'), $driverMock);
+        $originalException = new \Exception('Original exception');
+        $fallbackException = new \Exception('Fallback exception');
+
+        $driverMock->expects($this->at(0))
+            ->method('connect')
+            ->willThrowException($originalException);
+
+        $driverMock->expects($this->at(1))
+            ->method('connect')
+            ->willThrowException($fallbackException);
+
+        $this->expectExceptionMessage($originalException->getMessage());
+
+        $connection->getDatabasePlatform();
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
@@ -4,6 +4,7 @@ namespace Doctrine\Tests\DBAL\Functional;
 
 use Doctrine\DBAL\ConnectionException;
 use Doctrine\DBAL\DriverManager;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
 
 class ConnectionTest extends \Doctrine\Tests\DbalFunctionalTestCase
@@ -269,6 +270,31 @@ class ConnectionTest extends \Doctrine\Tests\DbalFunctionalTestCase
         );
 
         $this->assertTrue($connection->connect());
+
+        $connection->close();
+    }
+
+    /**
+     * @group DBAL-990
+     */
+    public function testDeterminesDatabasePlatformWhenConnectingToNonExistentDatabase()
+    {
+        if (in_array($this->_conn->getDatabasePlatform()->getName(), ['oracle', 'db2'], true)) {
+            $this->markTestSkipped('Platform does not support connecting without database name.');
+        }
+
+        $params = $this->_conn->getParams();
+        $params['dbname'] = 'foo_bar';
+
+        $connection = DriverManager::getConnection(
+            $params,
+            $this->_conn->getConfiguration(),
+            $this->_conn->getEventManager()
+        );
+
+        $this->assertInstanceOf(AbstractPlatform::class, $connection->getDatabasePlatform());
+        $this->assertFalse($connection->isConnected());
+        $this->assertSame($params, $connection->getParams());
 
         $connection->close();
     }


lib/Doctrine/DBAL/Connection.php
```diff
@@ -425,9 +425,46 @@ private function getDatabasePlatformVersion()
...
+                $databaseName = $this->_params['dbname'];
+                $this->_params['dbname'] = null;
+
+                try {
+                    $this->connect();
+                } catch (\Exception $fallbackException) {
+                    // Either the platform does not support database-less connections
+                    // or something else went wrong.
+                    // Reset connection parameters and rethrow the original exception.
+                    $this->_params['dbname'] = $databaseName;
```
mattjanssen: Could this and the identical line below be put into a `finally` block?

deeky666: @mattjanssen I thought about that but as we need to backport this to `2.5` which still supports PHP `5.3`, this unfortunately is not an option

dkarlovi:
Just to check as I was reading this patch: does Doctrine cache the server version after having detected it? I ask because @Ocramius [mentioned specifying it saves one query per request, as if it's not cached](https://github.com/doctrine/DoctrineBundle/issues/351#issuecomment-285357395), which seems weird.

deeky666:
@dkarlovi yes it kind of does because it is only required to create the correct platform instance which is then stored in the connection object. [See here](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Connection.php#L329)

Also I am not quite sure what @Ocramius meant by saying "it saves one request" because we only retrieve the server version if the driver supports it without having to do another query. Most driver can provide it once connected without firing a query. In fact the only driver that does not is SQL Anywhere (currently) and you have to either specify the correct version during connection or it will fall back to a "default" platform.

dkarlovi:
@deeky666 sorry, I meant across requests. So, if no version is specified and it gets auto-detected, is that data somehow available to the next request or does it redo the detection?

Maybe that's what @Ocramius meant?

deeky666:
@dkarlovi no it is not cached. This is out of scope here. One could implement something like this in userland. But I really don't see why you should do that as platform detection as it is implemented does not cause any overhead. You will have to establish a new connection on a new request anyways, detecting the platform version is nothing more than simply connecting and retrieving a piece of information store in the connection resource. What this PR deals with is edge cases where the platform is required even though the database to connect does not exist yet.

dkarlovi:
@deeky666 sorry for spamming, I know it's outside the scope, just asked here as you seemed knowledgeable about the topic. Thanks for the answer.

deeky666:
@dkarlovi no worries. Glad to clear that up :) Well, in the end I was the one that caused all the trouble because of implementing the feature :P

dkarlovi:
@deeky666 branch name `atonement` :)

deeky666:
@dkarlovi hehe actually I was thinking about possible solutions for a long time but nothing reasonable came to my mind what we could do in DBAL. I think this one here is a fair compromise.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -425,9 +425,46 @@ private function getDatabasePlatformVersion()
 
         // If not connected, we need to connect now to determine the platform version.
         if (null === $this->_conn) {
-            $this->connect();
+            try {
+                $this->connect();
+            } catch (\Exception $originalException) {
+                if (empty($this->_params['dbname'])) {
```
phansys: You could use `! $this->_params['dbname']`.

phansys (changes requested):
Status: Needs work.

lib/Doctrine/DBAL/Connection.php
deeky666: No I can't because it is possible to instantiate a connection object without this parameter. I even fell into this trap during writing the tests.

weaverryan:
👍 This makes sense to me.

Are we waiting for anything? This would close several issues, and help stop confusing users :). Looks like the last 1 comment from `phansys` was replied to.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -425,9 +425,46 @@ private function getDatabasePlatformVersion()
...
+
+                return $serverVersion;
+            }
+
         }
 
+        return $this->getServerVersion();
+    }
+
+    private function getServerVersion()
```
Ocramius: Documentation and return type for this function is missing

lib/Doctrine/DBAL/Connection.php
```diff
@@ -425,9 +425,46 @@ private function getDatabasePlatformVersion()
 
         // If not connected, we need to connect now to determine the platform version.
         if (null === $this->_conn) {
-            $this->connect();
+            try {
+                $this->connect();
+            } catch (\Exception $originalException) {
```
Ocramius: Is there a more specific exception that we can catch here?

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
```diff
@@ -272,4 +272,29 @@ public function testConnectWithoutExplicitDatabaseName()
...
+        $params = $this->_conn->getParams();
+        $params['dbname'] = 'foo_bar';
+
+        $connection = DriverManager::getConnection(
+            $params,
+            $this->_conn->getConfiguration(),
+            $this->_conn->getEventManager()
+        );
+
+        $this->assertInstanceOf('Doctrine\DBAL\Platforms\AbstractPlatform', $connection->getDatabasePlatform());
```
Ocramius: `::class`

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
```diff
@@ -272,4 +272,29 @@ public function testConnectWithoutExplicitDatabaseName()
 
         $connection->close();
     }
+
+    /**
+     * @group DBAL-990
+     */
+    public function testDeterminesDatabasePlatformWhenConnectingToNonExistentDatabase()
+    {
+        if (in_array($this->_conn->getDatabasePlatform()->getName(), array('oracle', 'db2'), true)) {
```
Ocramius: Short array syntax

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -670,4 +670,29 @@ public function testPlatformDetectionIsTriggerOnlyOnceOnRetrievingPlatform()
...
+
+    /**
+     * @group DBAL-990
+     */
+    public function testRethrowsOriginalExceptionOnDeterminingPlatformWhenConnectingToNonExistentDatabase()
+    {
+        /** @var \Doctrine\Tests\Mocks\VersionAwarePlatformDriverMock|\PHPUnit_Framework_MockObject_MockObject $driverMock */
+        $driverMock = $this->createMock('Doctrine\Tests\Mocks\VersionAwarePlatformDriverMock');
+
+        $connection = new Connection(array('dbname' => 'foo'), $driverMock);
```
Ocramius: Short array

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -670,4 +670,29 @@ public function testPlatformDetectionIsTriggerOnlyOnceOnRetrievingPlatform()
...
         $this->assertSame($platformMock, $connection->getDatabasePlatform());
     }
+
+    /**
+     * @group DBAL-990
+     */
+    public function testRethrowsOriginalExceptionOnDeterminingPlatformWhenConnectingToNonExistentDatabase()
+    {
+        /** @var \Doctrine\Tests\Mocks\VersionAwarePlatformDriverMock|\PHPUnit_Framework_MockObject_MockObject $driverMock */
+        $driverMock = $this->createMock('Doctrine\Tests\Mocks\VersionAwarePlatformDriverMock');
```
Ocramius: `::class`

Ocramius (changes requested):
@deeky666 moved milestone to 2.6, since this is an addition.

kimhemsoe:
Only big concern i have is the catch for \Exception else LGTM after Ocramius comments have been addressed.

weaverryan:
+1 for the comments from Ocramius. Hopefully we can catch a more specific exception... but even if we can't, the code *is* at least setup to throw that original exception eventually.

/cc @deeky666 updatie? ;)

lib/Doctrine/DBAL/Connection.php
deeky666: @Ocramius I do not know what else we could catch here as the driver interface does not specify what to throw in case `$driver->connect()` fails. Any ideas welcome, though.

Ocramius: That's a good-enough explanation for me

Ocramius (approved):

Ocramius:
All green! :ship:

weaverryan:
Woohoo! You guys ROCK!!! Thanks to you @deeky666 and @Ocramius!

leofeyer:
Unfortunately, this PR does not fix all issues.

If username and password have not been set (for whatever reason), the code still throws an exception (unless anonymous database connections are allowed).

IMHO, we should either return `null` instead of throwing an exception [here](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Connection.php#L432) or implement https://github.com/doctrine/dbal/pull/814 so we can explicitly configure `null`.

mattjanssen:
What is the use case for not having valid credentials being OK?

The original issue was if you have valid credentials the system is able to create a database for you. That is unless it first tries to fetch the server's version, in which case it threw a "Missing Database" exception. Which is a silly exception if you're just asking the DBAL to create said database.

But if you don't even have valid credentials, how would it create or query a database for you?

leofeyer:
We have a web-based GUI for creating the `parameters.yml` file in Symfony, which worked perfectly before this change.

lib/Doctrine/DBAL/Connection.php
catchamonkey: On this note, the merged fixes don't appear to have made it back to the 2.5 branch.

https://github.com/doctrine/dbal/blob/2.5/lib/Doctrine/DBAL/Connection.php#L405-L441

Ocramius:
This was only merged to 2.6

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

lxg:
@Ocramius: I’m not sure if this is entirely fixed in DBAL > 2.6 or if I have a different issue.

I’m using DBAL+ORM with Symfony 2.8, my Doctrine packages are:

```
$ composer show | grep doctrine
doctrine/annotations                 v1.5.0             Docblock Annotations Parser
doctrine/cache                       v1.7.1             Caching library offering an object-oriented API for many cache backends
doctrine/collections                 v1.5.0             Collections Abstraction library
doctrine/common                      v2.8.1             Common Library for Doctrine projects
doctrine/dbal                        v2.6.2             Database Abstraction Layer
doctrine/doctrine-bundle             1.7.0              Symfony DoctrineBundle
doctrine/doctrine-cache-bundle       1.3.0              Symfony Bundle for Doctrine Cache
doctrine/inflector                   v1.2.0             Common String Manipulations with regard to casing and singular/plural rules.
doctrine/instantiator                1.0.5              A small, lightweight utility to instantiate objects in PHP without invoking their constructors
doctrine/lexer                       v1.0.1             Base library for a lexer that can be used in Top-Down, Recursive Descent Parsers.
doctrine/orm                         v2.5.10            Object-Relational-Mapper for PHP
```

When I run `app/console cache:clear --no-optional-warmers --env=prod -v` with `database_driver: pdo_mysql` and `server_version` *not* being set, I get the following exception trace:

```
  [Doctrine\DBAL\Exception\ConnectionException]                                                                       
  An exception occurred in driver: SQLSTATE[HY000] [1045] Access denied for user ''@'localhost' (using password: NO)  
                                                                                                                      

Exception trace:
 () at ./vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php:108
 Doctrine\DBAL\Driver\AbstractMySQLDriver->convertException() at ./vendor/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php:176
 Doctrine\DBAL\DBALException::wrapException() at ./vendor/doctrine/dbal/lib/Doctrine/DBAL/DBALException.php:161
 Doctrine\DBAL\DBALException::driverException() at ./vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOMySql/Driver.php:47
 Doctrine\DBAL\Driver\PDOMySql\Driver->connect() at ./vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php:372
 Doctrine\DBAL\Connection->connect() at ./vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php:432
 Doctrine\DBAL\Connection->getDatabasePlatformVersion() at ./vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php:396
 Doctrine\DBAL\Connection->detectDatabasePlatform() at ./vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php:338
 Doctrine\DBAL\Connection->getDatabasePlatform() at ./vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php:763
 Doctrine\ORM\Mapping\ClassMetadataFactory->getTargetPlatform() at ./vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php:616
 Doctrine\ORM\Mapping\ClassMetadataFactory->completeIdGeneratorMapping() at ./vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php:174
 Doctrine\ORM\Mapping\ClassMetadataFactory->doLoadMetadata() at ./vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php:333
 Doctrine\Common\Persistence\Mapping\AbstractClassMetadataFactory->loadMetadata() at ./vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/ClassMetadataFactory.php:78
 Doctrine\ORM\Mapping\ClassMetadataFactory->loadMetadata() at ./vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php:217
 Doctrine\Common\Persistence\Mapping\AbstractClassMetadataFactory->getMetadataFor() at ./vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/AbstractClassMetadataFactory.php:115
 Doctrine\Common\Persistence\Mapping\AbstractClassMetadataFactory->getAllMetadata() at ./vendor/symfony/symfony/src/Symfony/Bridge/Doctrine/CacheWarmer/ProxyCacheWarmer.php:69
 Symfony\Bridge\Doctrine\CacheWarmer\ProxyCacheWarmer->warmUp() at ./vendor/symfony/symfony/src/Symfony/Component/HttpKernel/CacheWarmer/CacheWarmerAggregate.php:48
 Symfony\Component\HttpKernel\CacheWarmer\CacheWarmerAggregate->warmUp() at ./vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:526
 Symfony\Component\HttpKernel\Kernel->initializeContainer() at ./vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:133
 Symfony\Component\HttpKernel\Kernel->boot() at ./vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Command/CacheClearCommand.php:134
 Symfony\Bundle\FrameworkBundle\Command\CacheClearCommand->warmup() at ./vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Command/CacheClearCommand.php:96
 Symfony\Bundle\FrameworkBundle\Command\CacheClearCommand->execute() at ./vendor/symfony/symfony/src/Symfony/Component/Console/Command/Command.php:266
 Symfony\Component\Console\Command\Command->run() at ./vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:893
 Symfony\Component\Console\Application->doRunCommand() at ./vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:204
 Symfony\Component\Console\Application->doRun() at ./vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Console/Application.php:93
 Symfony\Bundle\FrameworkBundle\Console\Application->doRun() at ./vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:122
 Symfony\Component\Console\Application->run() at ./app/console:29
```

Apparently this has to do with ORM trying to determine the ID generator strategy for my entities.

However, as soon as I set `server_version` in my config, I can clear the cache just fine.

Ocramius:
Don't think this issue is valid once taken out of the symfony cache
context: can you clean up the issue description to clarify what the problem
is, without including any symfony specifics?

lxg:
You’re right, this has nothing to do with Symfony.

I think the core of the problem is that `Doctrine\ORM\Mapping\ClassMetadataFactory->completeIdGeneratorMapping` does a full connect with user/pass to the database server to determine the mapping strategy. I’m not sure if this is really necessary, or, more specific, how the server version helps in determining the mapping strategy.

yivi:
I know this is supposedly fixed, but I still have trouble on DBAL 2.9.2 **with** `server_version` set, and I still get this errors when building the application image if the DB is not available...

mpdude:
Regarding @lxg's [comment from above](https://github.com/doctrine/dbal/pull/2671#issuecomment-328553295): 

What you're seeing is that once you need to access ORM metadata – be it because you ask the `EntityManager` for a particular repository, or because you want to warm caches by creating ORM proxies – you may run into the situation that you need to know which database platform you're working against.

In your case (which I've run into as well) that is because the way the default [Identifier Generation Strategy](https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/reference/basic-mapping.html#identifier-generation-strategies) is chosen depends on the preferred mechanism for the platform.

What this PR fixed is that the platform can be determined by connecting to the database _without_ a particular schema being available. This allows to use DBAL to actually _create_ the schema.

Either way, it is still necessary that the database (server) can be connected to, and the username/password given/chosen need to at least allow connections.

This also means that when you want to do things like warming caches (for example, when building a Docker image) or [linting the Symfony Container](https://symfony.com/blog/new-in-symfony-4-4-service-container-linter), you need to have a database running _or_ specify the `server_version` explicitly.

yusefauto1:
adding `server_version` to  `doctrine.yaml` file fixed my issue. thanks

yuseferi:
adding `server_version` to `doctrine.yaml` file fixed my issue. thanks

Issue #1131, opened 2015-04-01
DBAL-1188: Oracle Platform: List table columns are returned in the wrong order


doctrinebot:
Jira issue originally created by user martijn:

The function:

```
Doctrine\DBAL\Platforms\OraclePlatform::getListTableColumnsSQL
```

returns the columns sorted by name. This causes a problem in my application. Other platforms sort them on position. So when I build a model for a HTML table (placing datatype formatters for each column) the order of formatters gets messed up.

The equivalent method in other platforms like MySQL or PostgreSQL all make sure the order is preserved (the same as in the database).

To fix I changed

```
...ORDER BY c.column_name
```

to

```
...ORDER BY c.column_id
```

deeky666:
Fixed via #2673

PR #2673, opened 2017-03-10
#1131 On OracleDB, `getListTableColumnsSQL` returns columns ordered by position instead of by name

mdwheele:
Related: #1131

Hello!

This is my first contribution to DBAL, so I would appreciate any feedback on how I could improve this PR and future ones, if necessary. To the meat of it!

Other platforms (than Oracle) return column listings by position (implicitly). The `OraclePlatform` implementation explicitly orders by `column_name`. Order matters in my application of DBAL. I am building tooling for scheduled data migrations between platforms and would like to maintain ordering between migrations.

I have tested this to the best of my ability, locally, following the instructions for specifying `phpunit.xml` configuration. I am running Oracle XE 12c via [this docker container](https://hub.docker.com/r/sath89/oracle-12c). That said, tests are passing (with skipped / risky) but I **would appreciate extra scrutiny** to make sure I've made all necessary changes.

Thanks!

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
@@ -199,10 +199,10 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
         $this->_sm->dropAndCreateTable($table);
 
         $columns = $this->_sm->listTableColumns('list_table_columns');
-        $columns_keys = array_keys($columns);
+        $columnsKeys = array_keys($columns);
 
         $this->assertArrayHasKey('id', $columns);
-        $this->assertEquals(0, array_search('id', $columns_keys));
+        $this->assertEquals(0, array_search('id', $columnsKeys));
         $this->assertEquals('id',   strtolower($columns['id']->getname()));
         $this->assertInstanceOf('Doctrine\DBAL\Types\IntegerType', $columns['id']->gettype());
         $this->assertEquals(false,  $columns['id']->getunsigned());
@@ -211,7 +211,7 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
         $this->assertInternalType('array',  $columns['id']->getPlatformOptions());
 
         $this->assertArrayHasKey('test', $columns);
-        $this->assertEquals(1, array_search('test', $columns_keys));
+        $this->assertEquals(1, array_search('test', $columnsKeys));
         $this->assertEquals('test', strtolower($columns['test']->getname()));
         $this->assertInstanceOf('Doctrine\DBAL\Types\StringType', $columns['test']->gettype());
         $this->assertEquals(255,    $columns['test']->getlength());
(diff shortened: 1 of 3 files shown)


morozov:
@mdwheele I'd say it makes sense to add a functional test which would actually prove that the columns are now returned in the right order. It would also help seeing if the same is really true for the rest of supported platforms.

mdwheele:
I agree. Would you say that maintaining column positioning is the intended behaviour across platforms? If so, I can write up tests for some of the other platforms that verify current behaviour and we might also consider updating the queries to explicitly order by column position. That said, as long as there are functional tests that assert intended behaviour, we should be good. In fact, the default behaviour for Oracle seems to be to maintain column ordering (if you just leave out the order by clause altogether).

morozov:
> Would you say that maintaining column positioning is the intended behaviour across platforms?

I'm not that well familiar with the Doctrine. What I'm saying is any behavior should be the same for all supported platforms if possible.

> I can write up tests for some of the other platforms

I think it should be a single functional test which would execute regardles what the current platform is (see *tests/Doctrine/Tests/DBAL/Functional*).

mdwheele:
> I'm not that well familiar with the Doctrine. What I'm saying is any behavior should be the same for all supported platforms if possible.

Same here, haha. Agree behaviour should be same. This issue created some confusion for me this morning.

> I think it should be a single functional test which would execute regardless of what the current platform is (see tests/Doctrine/Tests/DBAL/Functional).

I agree. It looks like adding assertions on ordering [could be done here](https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php#L195-L256) and would apply to all schema managers (which delegate to the platform). I'll go forward with adding extra assertions here and let CI handle setting up all the databases for functional tests. If things fail miserably, I'll do the work of getting everything running locally in containers. 

I'll probably stop for a bit after that until there's an "official" declaration of intended behaviour.

mdwheele:
After studying a little more and failing to reproduce CI results locally, I determined that it was because I am using Oracle 12c instead of Oracle XE 11g. I have pulled [sath89/oracle-xe-11g](https://hub.docker.com/r/sath89/oracle-xe-11g) and can now reliably run tests for Oracle locally. 

To be honest, I should have reverted my implementation change before committing 94df8b0 to show the failed assertion **only** for `OracleProvider`. That said, here are the results I get locally with `master`'s `OracleProvider`:

capture

Arranging the key names in this test alphabetically, we get: 

0. `bar`
1. `baz1`
2. `baz2`
3. `baz3`
4. `foo`
5. `id`
6. `test`

`id` is in the 5th position while we assert it should remain the first column. When I use the patched `OracleProvider`: 

capture

This is where 94df8b0 stands; passing.

I would like to pursue resolving the issues with Oracle 12c but think that is outside the scope of this PR. As far as I can tell, as long as CI is passing, the change in 94df8b0 should be asserting that all *other* providers maintain column position and that the `OracleProvider` now satisfies that assertion.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -199,8 +199,10 @@ public function testListTableColumns()
         $this->_sm->dropAndCreateTable($table);
 
         $columns = $this->_sm->listTableColumns('list_table_columns');
+        $columns_keys = array_keys($columns);
```
deeky666: Only nitpick I have: Please use camelCase throughout our codebase as we stick to PSR-2 coding standards.

deeky666 (changes requested):
Besides my nitpick, the patch looks good!

mdwheele:
@deeky666 Haha, thanks for the review! Looks like CI failed 7.0 MariaDB 10.1 again. Is this an issue on Travis' side? Seems intermittent.

deeky666 (approved):

deeky666:
@mdwheele yes it was a Travis problem. Restarted the job now it's green :) Thank you!

deeky666:
Backported to `2.5` via https://github.com/doctrine/dbal/commit/fadc21b3b03e60ae78172ff6e2cb7d737895aa74 https://github.com/doctrine/dbal/commit/97eefb83660ecd29b4846924fa54aeff6a9701ba https://github.com/doctrine/dbal/commit/a8c4e5c9d95177e4f59f14717b05d19ead51fbc2 and https://github.com/doctrine/dbal/commit/3cc78d64199ff1bf9b9b40e24ba5dd7445fd01b5

PR #2674, opened 2017-03-10
Correct `testListTableColumns` checked keys

mdwheele:
While working on #2673, I noticed an apparent error in a test assertion and wanted to keep it separate from my other PR. In a repetitive assertion block for `SchemaManagerFunctionalTestCase#testListTableColumns`, a column named `baz3` is being tested while a single assertion was testing `baz2` from the block before.


deeky666 (approved):

deeky666:
@mdwheele thx! Good catch!

Issue #2596, opened 2017-01-11
Schema diff keeps thinking custom mapping types have changed


vicdelfant:
I'm working on a relatively large project with a lot of custom mapping types (primarily value objects with UUIDs such as `OrderId`, `InvoiceId` etc.). For some reason, `schema:update` insists on re-creating all columns that contain such a custom type:

```bash
$ bin/console doctrine:schema:update --dump-sql
ALTER TABLE orders_order CHANGE order_id order_id CHAR(36) NOT NULL COMMENT '(DC2Type:orders.order_id)';
# ...

$ bin/console doctrine:schema:update --force
Updating database schema...
Database schema updated successfully! "33" queries were executed

$ bin/console doctrine:schema:update --dump-sql
ALTER TABLE orders_order CHANGE order_id order_id CHAR(36) NOT NULL COMMENT '(DC2Type:orders.order_id)';
# ...
```

Mapping types are properly registered, `requiresSQLCommentHint` is set to return `true` for all types, the `DC2Type` comment is correctly being added to the field definition (confirmed in the MySQL tables), storing and retrieving data works perfectly... everything but the schema comparison.

The types are registered through a dedicated service, but in a nutshell the process is as follows:
```php
foreach ($types as $name => $class) {
    Type::addType($name, $class);

    $platform->registerDoctrineTypeMapping($name, $name);

    // Added explicit call to markDoctrineTypeCommented(), makes no difference
    $platform->markDoctrineTypeCommented($name);
}
```

`AbstractPlatform::$doctrineTypeComments` contains all types so they're being processed just fine. Re-creating the entire database doesn't help either, and this obviously also affects migrations and `schema:validate`.

What am I missing here?

I came across the following similar issues but none of them solved my issue:
* https://github.com/sonata-project/SonataUserBundle/issues/103#issuecomment-10083314
* https://github.com/doctrine/dbal/issues/1614
* https://blog.vandenbrand.org/2015/06/25/creating-a-custom-doctrine-dbal-type-the-right-way/

Ocramius:
Possibly strictly related to https://github.com/doctrine/dbal/issues/2594

vicdelfant:
I'll try to find some time to write a test case so you can reproduce the issue more easily.

deeky666:
[...]

PR #2679, opened 2017-03-17
Fix: Doctrine custom types now accepts anything but ')'

PedroTroller:
Hi Doctrine Community.

This pull request fixes the issue #2596.

So if you try to create a type named like `my.type` without to follow [the regex](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php#L1096), everything works. You can use it and create new columns using this type.

But the problem appears when you try to update your schema or to generate a migration (using DoctrineMigrations for example). An "ALTER table" SQL command is always generated.

So why not allow any content but ")" ?


Ocramius:
Requires multiple additional test cases, especially around unbalanced
parentheses.

lcobucci:
Replaced by #2905

PR #2684, opened 2017-03-22
Use slave connection in executeQuery

enleur:


tests/Doctrine/Tests/DBAL/Functional/MasterSlaveConnectionTest.php
```diff
@@ -139,4 +139,19 @@ public function testMasterSlaveConnectionCloseAndReconnect()
...
     }
+
+    public function testKeepSlaveNoMasterOnExecuteQueryAfterWriteOperation()
+    {
+        $conn = $this->createMasterSlaveConnection(true);
+        $conn->insert('master_slave_table', array('test_int' => 30));
+
+        $this->assertTrue($conn->isConnectedToMaster());
+
+        $sql = "SELECT count(*) as num FROM master_slave_table";
```
Ocramius: What about `executeQuery` done with an `INSERT`, for example?

Ocramius (changes requested):

tests/Doctrine/Tests/DBAL/Functional/MasterSlaveConnectionTest.php
enleur: Is this a different case https://github.com/doctrine/dbal/blob/ec8fc16604915110eea8046e3ea594c133aee880/tests/Doctrine/Tests/DBAL/Functional/MasterSlaveConnectionTest.php#L67 ?

enleur: @Ocramius so?

Ocramius: `INSERT INTO ... SELECT ...` is what I meant.

Also, many RDBMS allow arbitrary subqueries that are not necessarily just `SELECT` queries. This includes running stored procedures as part of a `SELECT`

Ocramius (changes requested):
Please rebase instead of merging `master` into your branch

tests/Doctrine/Tests/DBAL/Functional/MasterSlaveConnectionTest.php
enleur: By default a slave connection used for `executeQuery`, until called any of `executeUpdate`,`beginTransaction` eg methods which switch a connection back to the master and keep it. So we will have the issue anyway.

This PR helps only switch a connection back to the slaves, so if we do queries like
`SELECT ...` (slave)
`UPDATE ...` (master)
`SELECT ...` (master, but why? should be slave again)
It's cool if you have async replication and want to avoid eventual consistency with monotonic queries, but it is not really useful if you want to do load balancing on read-only queries

10111282: Why `executeUpdate()` uses  a slave connection instead of master?

PR #2685, opened 2017-03-22
Keep track of full qualified type name when dealing with collections

sandvige:
This pull request should be considered with this one :
https://github.com/doctrine/doctrine2/pull/6354

We're currently facing an issue when trying to execute that kind of code:
``` php
$ids = array(new ComplexType(1), new ComplexType(2));
$manager
    ->getRepository('SomeEntity')
    ->findBy(array(
    'id' => $ids
));
```

This code should perform a query that looks like that:
SELECT ... FROM SomeEntityTable WHERE id IN (?, ?)

But it looks like the type name is lost when dealing with array. This code tries to bring a solution, but there's probably a better approach. Any idea?


Ocramius (changes requested):
Requires tests

sandvige:
@Ocramius Before trying harder to make unit tests, what do you think about the issue described above, and the solution we're trying to implement?

Ocramius:
I don't know what you are fixing without tests showing what is broken...

tests/Doctrine/Tests/DBAL/SQLParserUtilsTest.php
```diff
@@ -92,6 +94,24 @@ public function dataExpandListParameters()
...
+                array('integer', 'integer', 'integer')
+            ),
+            // DBAL-2685 - Positional: Very simple with one needle, with full type information, and a complex type
+            array(
+                "SELECT * FROM Foo WHERE foo IN (?)",
+                array(array($dt, $dt, $dt)),
+                array('[datetime]'),
+                'SELECT * FROM Foo WHERE foo IN (?, ?, ?)',
+                array($dt, $dt, $dt),
+                array('datetime', 'datetime', 'datetime')
```
sandvige: @Ocramius I added tests, I hope my tests are clear: we're trying to bring support for complex types into IN statements, like datetime, or any other custom types

sandvige:
@Ocramius?

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
```diff
@@ -419,11 +419,12 @@ public function testPrepareQueryBindValueDateTimeType()
 
     /**
      * @group DBAL-78
+     * @group DBAL-2685
      */
     public function testNativeArrayListSupport()
     {
         for ($i = 100; $i < 110; $i++) {
-            $this->_conn->insert('fetch_table', array('test_int' => $i, 'test_string' => 'foo' . $i, 'test_datetime' => '2010-01-01 10:10:10'));
+            $this->_conn->insert('fetch_table', array('test_int' => $i, 'test_string' => 'foo' . $i, 'test_datetime' => '2010-01-01 10:10:' . str_pad($i - 100, 2, "0", STR_PAD_LEFT)));
```
Ocramius: New tests should be written - existing tests should not be touched

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
```diff
@@ -433,12 +434,33 @@ public function testNativeArrayListSupport()
         $this->assertEquals(5, count($data));
         $this->assertEquals(array(array(100), array(101), array(102), array(103), array(104)), $data);
 
+        $stmt = $this->_conn->executeQuery('SELECT test_int FROM fetch_table WHERE test_int IN (?)',
```
Ocramius: Split out to a new test method

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
```diff
@@ -433,12 +434,33 @@ public function testNativeArrayListSupport()
...
+        $this->assertEquals(array(array(100), array(101), array(102), array(103), array(104)), $data);
+
         $stmt = $this->_conn->executeQuery('SELECT test_int FROM fetch_table WHERE test_string IN (?)',
             array(array('foo100', 'foo101', 'foo102', 'foo103', 'foo104')), array(Connection::PARAM_STR_ARRAY));
 
         $data = $stmt->fetchAll(PDO::FETCH_NUM);
         $this->assertEquals(5, count($data));
         $this->assertEquals(array(array(100), array(101), array(102), array(103), array(104)), $data);
+
+        $stmt = $this->_conn->executeQuery('SELECT test_int FROM fetch_table WHERE test_string IN (?)',
```
Ocramius: Split out to a new test method

lib/Doctrine/DBAL/SQLParserUtils.php
```diff
@@ -73,6 +73,42 @@ static public function getPlaceholderPositions($statement, $isPositional = true)
...
+
+    static private function removeCollectionMarker($type)
+    {
+        if (!static::isCollectionType($type))
+            throw SQLParserUtilsException::notCollectionType($type);
+        
+        if (is_numeric($type))
+            return $type - Connection::ARRAY_PARAM_OFFSET;
+
+        // [my_custom_type] -> my_custom_type
```
Ocramius: Ah, this is the trick: you are trying to use `[foo]` as an array mapping type of instances of `foo`.

The idea is good, but using strings for this is terrible, and we shouldn't let this land, as that is going to just complicate things much further.

A different approach could be needed here to define the type of a parameter: parsing the type is not viable...

Ocramius (changes requested):

Ocramius:
As mentioned in the review above, using strings for types is not viable for complex types. We should instead pass in a `ParameterType` object or such...

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
sandvige: I had to change this because the value wasn't based on $i, this is just improving the way the test is working

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
sandvige: Ok I could, but this is also a good way to prove the code is fully backward compatible

lib/Doctrine/DBAL/SQLParserUtils.php
sandvige: The DBAL project is already dealing with multiple types:
- string based types (looked up with Type::getType())
- PDO based types
- DBAL types (which are mixed in Connection::getBindingInfo() method)

We have to deal with strings without a major and full of BC breaks refacto ; this is how it currently works with Connection::ARRAY_PARAM_OFFSET. 

I understand this part is a major piece of the cake, maybe someone else could come with a better idea? We're already using this on our own project, and this is working well :)

sandvige:
@Ocramius Keep in mind the key feature is here: https://github.com/doctrine/doctrine2/pull/6354/files this is making your custom types being queriable through relationships \o/

lib/Doctrine/DBAL/SQLParserUtils.php
Ocramius: `"[foo]"` being considered `"special scenario in which the type represents an array of foo"` is not viable. What you are doing here is moving away from a simple map type, and converting type discovery into a function that actually creates more complex types.

I'd rather not have the feature at all than have it as magic strings.

 > We're already using this on our own project, and this is working well :)

The fact that it is working has nothing to do with correctness, sorry :-\

PR #2688, opened 2017-03-25
Support `IS NULL` checking in `Connection#delete()` and `Connection#update()` generated criteria, allowing for `null` column searches

jnvsor:
If you pass a null to the where array of update or delete it's
pretty obvious what you want.

Most (all?) databases won't match anything if you actually try
to compare with NULL making this a silent failure without it.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -603,6 +603,36 @@ class Connection implements DriverConnection
     }
 
     /**
+     * Gathers criteria for an update or delete call.
+     *
+     * @param array $identifiers Input array of columns to values
+     *
+     * @return string[][] a triplet with:
+     *                    - the first key being the column names
+     *                    - the second key being the values
+     *                    - the third key being the criteria strings
+     */
+    private function gatherCriteria(array $identifiers)
+    {
+        $columns = [];
+        $values = [];
+        $criteria = [];
+
+        foreach ($identifiers as $columnName => $value) {
+            if (null === $value) {
+                $criteria[] = $this->getDatabasePlatform()->getIsNullExpression($columnName);
+                continue;
+            }
+
+            $columns[] = $columnName;
+            $values[] = $value;
+            $criteria[] = $columnName . ' = ?';
+        }
+
+        return [$columns, $values, $criteria];
+    }
+
+    /**
      * Executes an SQL DELETE statement on a table.
      *
      * Table expression and columns are not escaped and are not safe for user-input.
@@ -621,15 +651,7 @@ class Connection implements DriverConnection
             throw InvalidArgumentException::fromEmptyCriteria();
         }
 
-        $columnList = array();
-        $criteria = array();
-        $paramValues = array();
-
-        foreach ($identifier as $columnName => $value) {
-            $columnList[] = $columnName;
-            $criteria[] = $columnName . ' = ?';
-            $paramValues[] = $value;
-        }
+        list($columnList, $paramValues, $criteria) = $this->gatherCriteria($identifier);
 
         return $this->executeUpdate(
             'DELETE FROM ' . $tableExpression . ' WHERE ' . implode(' AND ', $criteria),
@@ -694,7 +716,6 @@ class Connection implements DriverConnection
     {
         $columnList = array();
         $set = array();
-        $criteria = array();
         $paramValues = array();
 
         foreach ($data as $columnName => $value) {
@@ -703,11 +724,9 @@ class Connection implements DriverConnection
             $paramValues[] = $value;
         }
 
-        foreach ($identifier as $columnName => $value) {
-            $columnList[] = $columnName;
-            $criteria[] = $columnName . ' = ?';
-            $paramValues[] = $value;
-        }
+        list($whereColumns, $whereValues, $criteria) = $this->gatherCriteria($identifier);
+        $columnList = array_merge($columnList, $whereColumns);
+        $paramValues = array_merge($paramValues, $whereValues);
 
         if (is_string(key($types))) {
             $types = $this->extractTypeValues($columnList, $types);
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -603,33 +603,33 @@ class Connection implements DriverConnection
     }
 
     /**
-     * Gathers criteria for an update or delete call.
+     * Gathers conditions for an update or delete call.
      *
      * @param array $identifiers Input array of columns to values
      *
      * @return string[][] a triplet with:
-     *                    - the first key being the column names
+     *                    - the first key being the columns
      *                    - the second key being the values
-     *                    - the third key being the criteria strings
+     *                    - the third key being the conditions
      */
-    private function gatherCriteria(array $identifiers)
+    private function gatherConditions(array $identifiers)
     {
         $columns = [];
         $values = [];
-        $criteria = [];
+        $conditions = [];
 
         foreach ($identifiers as $columnName => $value) {
             if (null === $value) {
-                $criteria[] = $this->getDatabasePlatform()->getIsNullExpression($columnName);
+                $conditions[] = $this->getDatabasePlatform()->getIsNullExpression($columnName);
                 continue;
             }
 
             $columns[] = $columnName;
             $values[] = $value;
-            $criteria[] = $columnName . ' = ?';
+            $conditions[] = $columnName . ' = ?';
         }
 
-        return [$columns, $values, $criteria];
+        return [$columns, $values, $conditions];
     }
 
     /**
@@ -651,12 +651,12 @@ class Connection implements DriverConnection
             throw InvalidArgumentException::fromEmptyCriteria();
         }
 
-        list($columnList, $paramValues, $criteria) = $this->gatherCriteria($identifier);
+        list($columns, $values, $conditions) = $this->gatherConditions($identifier);
 
         return $this->executeUpdate(
-            'DELETE FROM ' . $tableExpression . ' WHERE ' . implode(' AND ', $criteria),
-            $paramValues,
-            is_string(key($types)) ? $this->extractTypeValues($columnList, $types) : $types
+            'DELETE FROM ' . $tableExpression . ' WHERE ' . implode(' AND ', $conditions),
+            $values,
+            is_string(key($types)) ? $this->extractTypeValues($columns, $types) : $types
         );
     }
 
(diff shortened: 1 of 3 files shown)


Ocramius:
Pick the number for this PR for now (2688)

jnvsor:
Thanks! The test is added, so feel free to merge if my points on backwards compatibility are ok

lib/Doctrine/DBAL/Connection.php
```diff
@@ -584,9 +584,13 @@ public function delete($tableExpression, array $identifier, array $types = array
         $paramValues = array();
 
         foreach ($identifier as $columnName => $value) {
-            $columnList[] = $columnName;
-            $criteria[] = $columnName . ' = ?';
-            $paramValues[] = $value;
+            if (is_null($value)) {
```
lcobucci: Can we not duplicate this on the two different methods?

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -403,6 +403,98 @@ public function testUpdateWithSameColumnInDataAndIdentifiers()
         );
     }
 
+    /**
+     * @group DBAL-2688
+     */
+    public function testUpdateWithIsNull()
+    {
+        $driverMock = $this->createMock('Doctrine\DBAL\Driver');
```
lcobucci: Please use the `::class` syntax

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -403,6 +403,98 @@ public function testUpdateWithSameColumnInDataAndIdentifiers()
...
+
+        $driverMock->expects($this->any())
+            ->method('connect')
+            ->will($this->returnValue(new DriverConnectionMock()));
+
+        $conn = $this->getMockBuilder('Doctrine\DBAL\Connection')
```
lcobucci: Same as https://github.com/doctrine/dbal/pull/2688/files#r108123891

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -403,6 +403,98 @@ public function testUpdateWithSameColumnInDataAndIdentifiers()
...
+            ->setMethods(array('executeUpdate'))
```
lcobucci: Please use the short array syntax

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -403,6 +403,98 @@ public function testUpdateWithSameColumnInDataAndIdentifiers()
...
+            ->setConstructorArgs(array(array('platform' => new Mocks\MockPlatform()), $driverMock))
```
lcobucci: Same as https://github.com/doctrine/dbal/pull/2688/files#r108124008

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -403,6 +403,98 @@ public function testUpdateWithSameColumnInDataAndIdentifiers()
...
+            ]
+        );
+    }
+
+    /**
+     * @group DBAL-2688
+     */
+    public function testDeleteWithIsNull()
+    {
+        $driverMock = $this->createMock('Doctrine\DBAL\Driver');
```
lcobucci: Same as https://github.com/doctrine/dbal/pull/2688/files#r108123891

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -403,6 +403,98 @@ public function testUpdateWithSameColumnInDataAndIdentifiers()
...
+
+        $driverMock->expects($this->any())
+            ->method('connect')
+            ->will($this->returnValue(new DriverConnectionMock()));
+
+        $conn = $this->getMockBuilder('Doctrine\DBAL\Connection')
```
lcobucci: Same as https://github.com/doctrine/dbal/pull/2688/files#r108123891

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -403,6 +403,98 @@ public function testUpdateWithSameColumnInDataAndIdentifiers()
...
+            ->setMethods(array('executeUpdate'))
```
lcobucci: Same as https://github.com/doctrine/dbal/pull/2688/files#r108124008

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -403,6 +403,98 @@ public function testUpdateWithSameColumnInDataAndIdentifiers()
...
+            ->setConstructorArgs(array(array('platform' => new Mocks\MockPlatform()), $driverMock))
```
lcobucci: Same as https://github.com/doctrine/dbal/pull/2688/files#r108124008

tests/Doctrine/Tests/DBAL/ConnectionTest.php
lcobucci: Can you extract this to a method (you can also encapsulate `$driverMock` there) so that it can be reused on `delete()`?

lcobucci (changes requested):

tests/Doctrine/Tests/DBAL/ConnectionTest.php
jnvsor: I copy pasted these from existing test `testUpdateWithSameColumnInDataAndIdentifiers` - I guess the code style has changed since then? I'll get the changes done

lib/Doctrine/DBAL/Connection.php
jnvsor: I don't see how without passing columnList, criteria and paramValues as references to the helper method. (Which is probably why they're duplicated in the first place) Is that ok?

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -403,6 +392,80 @@ public function testUpdateWithSameColumnInDataAndIdentifiers()
...
+                'is_edited' => null,
+            ],
+            [
+                'id' => null,
+                'name' => 'foo',
+            ],
+            [
+                'text' => 'string',
+                'is_edited' => 'null',
+                'id' => 'null',
```
deeky666: Specifying `'null'` type is a bit weird because Doctrine does not have such type. As the testsuite somewhat also serves as API usage documentation I would suggest not promoting misusage here. I am pretty sure that code won't work when actually executed as Doctrine should complain about not knowing `null` type.
IMO the best would be to not specify any type for `null` values and let Doctrine decide what to do with it. As an alternative, if you want to complete type specification here, you should use the approriate types that reflect the column type like for `id` you would probably pass `integer` as type.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -403,6 +392,80 @@ public function testUpdateWithSameColumnInDataAndIdentifiers()
...
+            );
+
+        $conn->delete(
+            'TestTable',
+            [
+                'id' => null,
+                'name' => 'foo',
+            ],
+            [
+                'id' => 'null',
```
deeky666: Same as above. Also specifying a type for `null` values here does make even less sense as those values won't be turned into bound parameters.

deeky666 (changes requested):
Concerning my comments about the tests, it might be a good idea to add a simple functional test for each operation, too.

jnvsor:
@deeky666 Addressed your points and added tests to `WriteTest`. Is there an up to date code style guide somewhere? Lots of the older code doesn't look anything like your suggestions

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
```diff
@@ -285,4 +285,44 @@ public function testEmptyIdentityInsert()
...
+    {
+        $this->_conn->insert(
+            'write_table',
+            ['test_int' => '30', 'test_string' => null],
+            ['test_string' => 'string', 'test_int' => 'integer']
+        );
+
+        $data = $this->_conn->fetchColumn('SELECT COUNT(*) FROM write_table WHERE test_int = 30');
+
+        $this->assertEquals(1, $data);
```
phansys: Please, use `assertCount()` instead.

phansys (changes requested):
Status: Needs work.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -561,6 +561,28 @@ public function isTransactionActive()
...
 
     /**
+     * Gathers criteria for an update or delete call.
+     *
+     * @param array $identifiers  Input array of columns to values
+     * @param array &$columnList  Output list of bound columns for extractTypeValues
+     * @param array &$criteria    Output array of criteria
+     * @param array &$paramValues Output bound values
+     */
+    private function gatherCriteria($identifiers, &$columnList, &$criteria, &$paramValues)
```
phansys: Can these arguments be declared?

phansys (changes requested):
Status: Needs work.

lib/Doctrine/DBAL/Connection.php
jnvsor: How do you mean?

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
jnvsor: assertCount isn't used anywhere in the WriteTest file - are you sure?

phansys: IMO, no matter if the right assertion was used in other tests or not, this must not prevent you to make a proper check.

lib/Doctrine/DBAL/Connection.php
phansys:
```php
private function gatherCriteria(array $identifiers, array &$columnList, array &$criteria, array &$paramValues)
```

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
jnvsor: Ok, I'll use count then. As I asked before:

> Is there an up to date code style guide somewhere?

phansys: Which type of assert choose isn't related to CS, it depends on the type of data you need to check.

jnvsor: Right, but the data was an int. Choosing between doing the count in the assert or in the query is CS

jnvsor:
Any chance I could get a new review?

Ocramius (approved):
Patch looks awesome, but I will need to go through it to remove that by-ref `gatherCriteria` signature (returning a triplet instead)

jnvsor:
I'll get on that right now then, thanks for the feedback!

Ocramius:
@jnvsor ideally, the result would be `list($columnList, $criteria, $parameterValues) = $this->theFunction(...)`

jnvsor:
How should I write up the docstring? Just `@returns array [x, y, z]`?

Ocramius:
@jnvsor 

```php
/**
 * ...
 * @return string[][] a triplet with:
 *                    - the first key being the column list
 *                    - the second key being the criteria string
 *                    - the third key being the bound parameter types
 */
```

jnvsor:
Ok, I've cleaned it up a bit:

```php
    /**
     * Gathers criteria for an update or delete call.
     *
     * @param array $identifiers Input array of columns to values
     * @param array $columns     List of assigned column names
     * @param array $values      List of assigned values
     *
     * @return string[][] a triplet with:
     *                    - the first key being the column names
     *                    - the second key being the values
     *                    - the third key being the criteria strings
     */
    private function gatherCriteria(array $identifiers, array $columns = [], array $values = [])

[...]

list($columnList, $paramValues, $criteria) = $this->gatherCriteria($identifier, $columnList, $paramValues);
```

How's that?

Ocramius:
Why do you need `, $columnList, $paramValues);`?

jnvsor:
Because the `SET` in an `UPDATE` query comes before the `WHERE` - if we don't append to them the parameters will be bound out of order. That said, they're optional because you don't have a `SET` in a `DELETE` (only in an `UPDATE`)

Ocramius:
> if we don't append

`array_merge()` the results ;-)

jnvsor:
Yeah but then you can't destructure can you?

Ocramius:
```php
list($columnList, $criteria, $parameterValues) = $this->theFunction(...);
$columns = array_merge($previouslyDefinedColumns, $columnList);
$filter = array_merge($previouslyDefinedCriteria, $criteria);
```

jnvsor:
Done. Though looking at it the variables in `update`/`delete` could stand a renaming. `columns`, `values`, `conditions`?

Ocramius:
@jnvsor very good improvement! The last bit to improve is to avoid overwriting variables at all, and then we're done. This bit specifically:

```php
$columnList = array_merge($columnList, $whereColumns);
$paramValues = array_merge($paramValues, $whereValues);
```

Indeed, a variable rename is needed

jnvsor:
Added a variable rename and stopped the overwriting

Ocramius:
@jnvsor awesome! Waiting for travis and then this can be merged :-)

Ocramius:
:ship:

Ocramius:
Thanks @jnvsor!

PR #2690, opened 2017-03-28
ExpressionBuilder: Support array arguments for orX, andX

jnvsor:


jnvsor:
Would a ternary if be preferred? It does go over 100 char line length if I do that though.

morozov:
Starting PHP 5.6 you can use the `andX(...$array)` syntax with the existing implementation.

jnvsor:
I'm aware of that but DBAL supports down to 5.3 (Should I have made this PR against a different branch?)

morozov:
As far as I can tell, only DBAL v2.5 supports PHP 5.3, and this PR proposes a new feature which should go to the new minor release (like v2.6), but it will not support PHP 5.3.

jnvsor:
Ah, I wasn't aware 2.6 would drop 5.3 support - thought that was for the next major version. Never mind!

PR #2692, opened 2017-03-30
Fix default isolation level for the MySqlPlatform

jnvsor:
Default isolation level on mysql is REPEATABLE READ

https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_tx_isolation


deeky666 (approved):

deeky666:
@jnvsor thx!
@Ocramius I leave it up to you to decide whether this is something we need to backport to `2.5`. I am not sure about this.

Ocramius:
@deeky666 this changes a default, so I wouldn't want to backport it.

deeky666:
Agreed.

PR #2694, opened 2017-04-02
Fix primary key alteration when adding new columns on MySQL platforms.

drieschel:
Trying to get a column in a table which is not added yet, will result in throwing a SchemaException. Also altering not yet added columns does not make sense.


drieschel:
I will open a new pr it with a feature branch.

PR #2696, opened 2017-04-04
MySQL: Fix primary key alteration when adding new columns

drieschel:
Trying to get a column in a table which is not added yet, will result in throwing a SchemaException. Also altering not yet added columns does not make sense.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -708,6 +708,10 @@ class MySqlPlatform extends AbstractPlatform
 
         // Dropping primary keys requires to unset autoincrement attribute on the particular column first.
         foreach ($index->getColumns() as $columnName) {
+            if (! $diff->fromTable->hasColumn($columnName)) {
+                continue;
+            }
+          
             $column = $diff->fromTable->getColumn($columnName);
 
             if ($column->getAutoincrement() === true) {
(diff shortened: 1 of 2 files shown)


drieschel:
A feedback from someone of the doctrine/dbal team would be great.

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -708,6 +709,10 @@ private function getPreAlterTableAlterPrimaryKeySQL(TableDiff $diff, Index $inde
 
         // Dropping primary keys requires to unset autoincrement attribute on the particular column first.
         foreach ($index->getColumns() as $columnName) {
+            if(isset($diff->addedColumns[$columnName]) && $diff->addedColumns[$columnName] instanceof Column) {
+                continue;
+            }
```
deeky666: I think using the following check is more convenient and safer concerning the following `getColumn()` call:

```php
if (! $diff->fromTable->hasColumn($columnName)) {
    continue;
}
```

deeky666 (changes requested):

drieschel:
@deeky666 Done :)

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -708,6 +708,10 @@ private function getPreAlterTableAlterPrimaryKeySQL(TableDiff $diff, Index $inde
 
         // Dropping primary keys requires to unset autoincrement attribute on the particular column first.
         foreach ($index->getColumns() as $columnName) {
+            if(!$diff->fromTable->hasColumn($columnName)) {
```
deeky666: sorry to nitpick but we use the following code style `if (! $diff` can you please adjust that? afterwards I'll merge.

deeky666 (changes requested):

deeky666 (approved):

deeky666:
@drieschel thanks! Backported to `2.5` via c41cc5a6f03081ea5139050cb20901c0d663cd0c

PR #2699, opened 2017-04-17
DBAL-858: Workaround for Oracle's limit in IN clause items

### LINKED ISSUE #2095: DBAL-858: oracle IN statement with more than 1000 values
Jira issue originally created by user mdrolet:

If I have a query with a IN statement with more tahn 1000 values I get an sql error.

I've try IN with implode:
select \* from test where id IN(' . implode(',', $values) . ')
and I've also try with executeQuery:
select \* from test where id IN(:test)
  executeQuery($sql, array($values), array(\Doctrine\DBAL\Connection::PARAM_INT_ARRAY))

tomaszmadeyski:
Oracle does not allow queries with more than 1000 elements in IN clause

Query:
`SELECT ... WHERE foo IN (1,..., 1001)` 

will result in `ORA-01795 maximum number of expressions in a list is 1000`

This PR introduces method to convert such queries from
`SELECT ... WHERE foo IN (1,..., 1001)` 
to
`SELECT ... WHERE (foo IN (1,..., 1000) OR foo IN (1001))` 
and does this conversion before parsing query

It fixes https://github.com/doctrine/dbal/issues/2095

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -19,6 +19,7 @@
 
 namespace Doctrine\DBAL;
 
+use Doctrine\DBAL\Driver\AbstractOracleDriver;
 use Doctrine\DBAL\Driver\ServerInfoAwareConnection;
 use Doctrine\DBAL\Exception\InvalidArgumentException;
 use PDO;
@@ -837,6 +838,11 @@ class Connection implements DriverConnection
             if ($params) {
                 list($query, $params, $types) = SQLParserUtils::expandListParameters($query, $params, $types);
 
+                //Oracle does not allow more than 1000 elements in IN clause, we need to split it
+                if ($this->getDriver() instanceof AbstractOracleDriver) {
+                    $query = SQLParserUtils::splitInClause($query, 1000);
+                }
+
                 $stmt = $this->_conn->prepare($query);
                 if ($types) {
                     $this->_bindTypedValues($stmt, $params, $types);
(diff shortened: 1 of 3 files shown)


morozov:
In my understanding, Oracle enforces the limit on the number of parameters in order to prevent huge queries and too many bound variables. As long as these values are not taken from the input, the query above can be rewritten using subquery. Otherwise, a temporary table can be used.

I don't think the DBAL should workaround this limitation.

Ocramius:
> I don't think the DBAL should workaround this limitation.

And especially not with hacks/workarounds. If there were a proper solution, then we could use it, but this really just needs to be escalated to the underlying driver's maintainer.

tomaszmadeyski:
> Otherwise, a temporary table can be used.

That's exact issue I'm trying to solve. Using temporary table means writing a lot of boilerplate code and it's a hassle
> I don't think the DBAL should workaround this limitation.

This is a port of solution for .NET's Entity Framework in their Oracle's driver, it is done in DBAL layer and I think it is very convenient http://blog.devart.com/entity-framework-sql-generation-enhancements-for-in-clause.html (see `Oracle-specific SQL` paragraph)
 
> but this really just needs to be escalated to the underlying driver's maintainer

Do you mean oci8 driver's maintainer? If so this is not an issue of a driver, it is a feature/limitation of a DB engine itself and I don't think there's any chance Oracle will change this

Ocramius:
> Do you mean oci8 driver's maintainer? If so this is not an issue of a driver, it is a feature/limitation of a DB engine itself and I don't think there's any chance Oracle will change this

Well aware of it, but this really shouldn't lead to this massive amount of complication for an edge case. 

The fact that a tool has an imposed limitation doesn't mean that everyone else should work hard to provide a messy workaround for it: it should be fixed at the source of the problem or not at all, sorry :-\

If anyone knows Oracle devs that can influence the decisions there (I doubt it, since it's corporate world, which is basically Vogon poetry ^ 2), then please poke them and ask what the correct approach would be here. Adding complexity to the DBAL is something I'd not really want, as it already spawned massive amounts of headache for SQL Server (limit subquery mess).

PR #2704, opened 2017-04-21
Add error as well as exception handling to Mysqli drivers

develancer:
This patch restores the error handler if mysqli_real_connect throws an exception.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
+++ b/lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
@@ -67,14 +67,13 @@ class MysqliConnection implements Connection, PingableConnection, ServerInfoAwar
         $this->setDriverOptions($driverOptions);
 
         set_error_handler(function () {});
-
-        if ( ! $this->_conn->real_connect($params['host'], $username, $password, $dbname, $port, $socket, $flags)) {
+        try {
+            if ( ! $this->_conn->real_connect($params['host'], $username, $password, $dbname, $port, $socket, $flags)) {
+                throw new MysqliException($this->_conn->connect_error, $this->_conn->sqlstate ?? 'HY000', $this->_conn->connect_errno);
+            }
+        } finally {
             restore_error_handler();
-
-            throw new MysqliException($this->_conn->connect_error, @$this->_conn->sqlstate ?: 'HY000', $this->_conn->connect_errno);
         }
-
-        restore_error_handler();
 
         if (isset($params['charset'])) {
             $this->_conn->set_charset($params['charset']);
--- a/tests/Doctrine/Tests/DBAL/Driver/Mysqli/MysqliConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Driver/Mysqli/MysqliConnectionTest.php
@@ -28,9 +28,31 @@ class MysqliConnectionTest extends DbalTestCase
             ->getMockForAbstractClass();
     }
 
+    protected function tearDown()
+    {
+        if ($this->getName() == "testRestoresErrorHandlerOnException") {
+            restore_error_handler();
+        }
+    }
+
     public function testDoesNotRequireQueryForServerVersion()
     {
         $this->assertFalse($this->connectionMock->requiresQueryForServerVersion());
+    }
+
+    public function testRestoresErrorHandlerOnException()
+    {
+        $handler = function () { self::fail('Never expected this to be called'); };
+        $default_handler = set_error_handler($handler);
+
+        try {
+            new MysqliConnection(['host' => '255.255.255.255'], 'user', 'pass');
+            self::fail('An exception was supposed to be raised');
+        } catch (MysqliException $e) {
+            self::assertSame('Network is unreachable', $e->getMessage());
+        }
+
+        self::assertSame($handler, set_error_handler($default_handler), 'Restoring error handler failed.');
     }
 
     /**


Ocramius:
Overall, patch looks good, but the try-catch error handler restoring should be tested.

A test would do something like:

```php
$handler = function () { self::fail('Never expected this to be called');}
set_error_handler($handler);

try {
    $doTheThingThatFails();
    self::fail('An exception was supposed to be raised');
} catch (MysqliException $e) {
    // assert on the exception message
}

self::assertSame($handler, restore_error_handler(), 'handler stack was restored');
```

Ocramius (changes requested):
Tests needed, as per https://github.com/doctrine/dbal/pull/2704#issuecomment-296439989

develancer:
@Ocramius I have added the missing test.

Ocramius:
A bunch of failures popped up - see https://travis-ci.org/doctrine/dbal/jobs/225133976#L276-L303

develancer:
@Ocramius Right, sorry about that. It seems to be OK now.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
```diff
@@ -67,15 +67,14 @@ public function __construct(array $params, $username, $password, array $driverOp
         $this->setDriverOptions($driverOptions);
 
         set_error_handler(function () {});
-
-        if ( ! $this->_conn->real_connect($params['host'], $username, $password, $dbname, $port, $socket, $flags)) {
+        try {
+            if ( ! $this->_conn->real_connect($params['host'], $username, $password, $dbname, $port, $socket, $flags)) {
+                throw new MysqliException($this->_conn->connect_error, @$this->_conn->sqlstate ?: 'HY000', $this->_conn->connect_errno);
```
phansys: You could use `$this->_conn->sqlstate ?? 'HY000'` instead.

phansys (changes requested):
Status: Needs work.

develancer:
I could, but it would break compatibility with PHP 5.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
develancer: I could, but it would break compatibility with PHP 5.

phansys: You're targeting `master` branch, which already have [dropped the support for PHP 5](https://github.com/doctrine/dbal/blob/master/composer.json#L15).

develancer: OK, done.

phansys (approved):

tests/Doctrine/Tests/DBAL/Driver/Mysqli/MysqliConnectionTest.php
```diff
@@ -28,11 +28,31 @@ protected function setUp()
             ->getMockForAbstractClass();
     }
 
+    protected function tearDown()
+    {
+        restore_error_handler();
```
Ocramius: This will cause `restore_error_handler()` to be called for every test in this class - that is a mistake, since any error handler registered by phpunit may be replaced. Only the actual test overriding the error handler should restore the error handler (by calling `restore_error_handler()` at the end)

Ocramius (changes requested):

tests/Doctrine/Tests/DBAL/Driver/Mysqli/MysqliConnectionTest.php
develancer: As you can see, at the end of the test there is a call to `set_error_handler($default_handler)`, which should essentially restore the error handler to the pre-test state. However, without `restore_error_handler` in tearDown, it causes 7 other tests to fail from some reason.

I have changed the tearDown to restore the error handler only for this test.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
ostrolucky: Replacing `?:` with `??` in this case doesn't make sense. Why do we suddenly need to ignore missing property `sqlstate` inside `$this->_conn`? I for one would like code generate error and not silently continue in such a case.

tests/Doctrine/Tests/DBAL/Driver/Mysqli/MysqliConnectionTest.php
```diff
@@ -28,11 +28,33 @@ protected function setUp()
             ->getMockForAbstractClass();
     }
 
+    protected function tearDown()
+    {
+        if ($this->getName() == "testRestoresErrorHandlerOnException") {
```
Ocramius: The issue here is that you have to do `restore_error_handler()` twice in the `testRestoresErrorHandlerOnException` test :-)

Ocramius (changes requested):

tests/Doctrine/Tests/DBAL/Driver/Mysqli/MysqliConnectionTest.php
Ocramius: I'll apply this change locally

Ocramius:
@develancer I manually merged this into `master`: won't be backported to `2.5`, since it uses `finally`, which isn't available in older PHP versions.

PR #2708, opened 2017-05-03
Added access to native close() method of mysqli

develancer:
I noticed that the `close()` method in class Connection does not close the driver connection, but only sets the reference to NULL. I added a CloseableConnection interface (implemented by MysqliConnection, for now) which allows to close the connection, if supported.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/CloseableConnection.php
+++ b/lib/Doctrine/DBAL/Driver/CloseableConnection.php
@@ -1,0 +1,35 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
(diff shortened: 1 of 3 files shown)


lib/Doctrine/DBAL/Driver/CloseableConnection.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+namespace Doctrine\DBAL\Driver;
+
+/**
+ * An interface for connections which support a "native" close method.
+ *
+ * @link   www.doctrine-project.org
+ */
+interface CloseableConnection
```
Ocramius: This should rather be handled in the `__destroy` of the `MysqliConnection` if it isn't able to handle closing a connection on GC

Ocramius (changes requested):

lib/Doctrine/DBAL/Driver/CloseableConnection.php
develancer: We rather thought of a way to explicitly close a connection without waiting for a GC.

Ocramius: @develancer call `__destruct` manually on the `MysqliConnection` then

develancer: @Ocramius We may as well call `$connection->getWrappedConnection()->getWrappedResourceHandle()->close()` but this patch seemed for us like a more civilized way to do it.

Ocramius: Let's not expose stuff that shouldn't be exposed in first place, please :-\

develancer: All right. Feel free to close this pull request.

Ocramius:
Closing as `invalid`: the proper approach would be to handle this operation correctly in `__destruct`, like with all other platforms (which handle GC correctly)

PR #2718, opened 2017-05-08
Add iterators for non pdo drivers.

jenkoian:
Allows result sets to be iterated over without running out of memory on
large data sets.

--

**Before:** Iterating (e.g. foreaching) over a large data set would often run out of memory.
**After:** Iterating over a large data set uses iterators using the cursor so that memory usage remains low.

That's the aim anyway, tested (manually) the SQLSrv driver and it has the desired effect, could do with some help testing the other drivers if anyone is able? I would have liked to add some unit tests but none of the existing stuff is unit tested and I'm not really sure where I would even start, again any help here appreciated also.

**Small update**: I have reverted the changes for the mysqli driver as I don't quite understand that driver enough, it seems to behave unlike any of the others, again any help here much appreciated. Also came across the DataAccessTest which should cover most of this, it certainly brought out the issues with the mysqli driver, so that's good.

**Another update**: I have removed all the stuff with iterators, favouring generators. This makes things a lot easier as we can just re-use the existing fetch functionality. Generators are not without their issues though as they cannot be re-used, see the discussion below for more info, in particular [this comment](https://github.com/doctrine/dbal/pull/2718#issuecomment-302127776).

**Update (20170724)**: Due to concerns over consistency both PDO and non PDO can now be iterated but *not* rewindable. A rewindable generator has been included if users wish to use this (possible this will be removed though?).

**Update (20170914)**: StatementIterator class created which is now used by all non PDO statements. Extracting to own class makes it easier to test.


jvasseur:
You could use a generator instead of creating classes for the iterators, it would simplify the code needed for the feature.

jenkoian:
@jvasseur agreed that would be easier...as `getIterator` needs to return a `Traversable` though not sure it would work? Did consider an anonymous class here and indeed started with that but (can't remember why now) couldn't quite get it to work, although admittedly I didn't spend very long looking into why.

jvasseur:
The `Generator` class implements the `Iterator` interface so the return type of `getIterator` is not a problem.

jenkoian:
@jvasseur wow, I did not know that! Thanks, will update PR, this makes it way easier.

jenkoian:
@jvasseur updated PR, thanks! Will likely give it some time for feedback before rebasing so's we can all forget that whole iterator thing ever happened 😆  Have manually tested the SQL server implementation and it works.

deeky666:
While the idea is good, using generators is a BC break when doing the following:

```php
$result = $statement->getIterator();

foreach ($result as $row) {
}

// will error because generators cannot be reused
foreach ($result as $row) {
}
```

Not sure if we should care though. /cc @Ocramius

Ocramius:
We could provide a rewindable iterator somehow: tricky but feasible

morozov:
@Ocramius could you elaborate a bit more, besides the technical possibility and BC, what could it be needed for?

Generators do not allow to be rewound on purpose, and I think the same is applicable to the statements: as a developer, I'd better be informed that I'm re-fetching data from the same statement twice and redesign my logic instead of just fetching the same data again.

Ocramius:
@morozov we could accumulate the data that was already iterated upon

deeky666:
@morozov I tend to agree with your argumentation but TBH I still find it confusing that generators implement `\Iterator` although `rewind()` is definitely something that generators do not support. It might be a matter of taste but when I want an iterator from a statement and it tells me I get `\Iterator`, I expect it to be an object that implements the interface correctly. There is no word in the documentation about allowing objects not to be rewindable. IMO PHP has brought a lot of confusion with that to the userland and personally I am not a friend of pushing it further.

@Ocramius do you mean "caching" fetched data in the object while iterating? What is wrong with simply executing the statement again on rewind?

Ocramius:
> What is wrong with simply executing the statement again on rewind?

Side effects

deeky666:
@Ocramius can you give an example? Funfact: I am actually working on `3.0` version of the driver component since two weeks or so and have come up with [this](https://gist.github.com/deeky666/0f82517e2af2c4c7ec61d99f7b005f44) (draft, mysqli).

Ocramius:
@deeky666 for example, a query that selects values from a sequence will allocate new sequence values, instead of retrieving the same ones over multiple calls.

deeky666:
@Ocramius good point. That's reason enough for me. So then iterators should probably really be forward only here. I don't see how we could make them rewindable then without storing fetched rows in memory.

jenkoian:
I've added a failing test case for the re-usable iterator problem.

jenkoian:
I added an example of a `RewindableGenerator` to the mysqli driver after I noticed [Symfony used a `RewindableGenerator`](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/DependencyInjection/Argument/RewindableGenerator.php), feedback welcome. If this works I will look to roll out to the other drivers.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -406,8 +407,12 @@ public function setFetchMode($fetchMode, $arg2 = null, $arg3 = null)
      */
     public function getIterator()
     {
-        $data = $this->fetchAll();
-
-        return new \ArrayIterator($data);
+        $that = $this;
```
jvasseur: You can use `$this` inside the closure instead of doing that.

lib/Doctrine/DBAL/RewindableGenerator.php
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    public function getIterator()
+    {
+        $g = $this->generator;
+
+        return $g();
+    }
+
+    public function count()
```
jvasseur: I don't think we need to implement the countable interface here.

jenkoian: Yep, you're right, will remove 👍

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
jenkoian: Been working on a legacy codebase for too long...thanks

lib/Doctrine/DBAL/RewindableGenerator.php
```diff
@@ -0,0 +1,23 @@
...
+     * @param callable $generator
+     */
+    public function __construct(callable $generator)
+    {
+        $this->generator = $generator;
+    }
+
+    public function getIterator()
+    {
+        $g = $this->generator;
```
Ocramius: `return ($this->generator)();`

Ocramius: This would simply call the generator callable at each `foreach () {}` (because each `foreach` will cause a `->getIterator()` call).

That's most likely what we do NOT want, as a query will be repeated.

The test case is as following:

```php
$result = someQuery();

foreach ($result as $value) {
}

foreach ($result as $value) {
}

self::assertCount(1, $executedQueries);
```

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
```diff
@@ -863,6 +863,25 @@ private function skipOci8AndMysqli()
...
+        foreach ($stmt as $row) {
+            $i++;
+        }
+
+        $j = 0;
+        foreach ($stmt as $row2) {
+            $j++;
+        }
+
+        $this->assertEquals($i, $j);
```
Ocramius: The amount of executed queries is to be checked here - you can use something like `SELECT CURRENT_TIME_IN_MICROSECONDS()` (or equivalent - I made this SQL up) to verify that the result doesn't change over multiple calls.

Ocramius (changes requested):

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
jenkoian: I've added a way of counting the number of executed queries using `DebugStack` take a look, if that doesn't quite get what we need I will do the microseconds thing.

Ocramius: Argh, I just checked: the logger is only used outside statements, not inside them :-(

This means that the logger will only log the creation of the statement and its first execution.

A stateful query would most likely be better. Sorry for not noticing this before!

Ocramius (changes requested):
Requires testing against stateful queries

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
jenkoian: Thanks for your reply and sorry for my delay. I've pushed something which I'm pretty sure is what you mean. However, I don't think this works. I can't seem to get microseconds. Only thing I can do to get around this is to add a sleep call 😞 . Any ideas?

jenkoian: I think this is likely because of the MySQL version. 5.6 and up seems to support it. Travis is on a pre 5.6 version of MySQL IIRC although I think the new trusty boxes may be newer. So I guess we could update travis to build against the newer versions if we want (if this is indeed the issue!).

Ocramius: We can do that, yes (using trusty on travis-ci). Ideally, we'd test this feature with SQLite in-memory though, unless we want something mysql-specific

jenkoian: afaict, travis is using mariadb which I think supports microseconds, so think it should be fine?

On another note, I noticed the build is failing and it's because pdo doesn't actually support rewinding the iterator (well not by default is my understanding, may be able to with unbuffered queries) so made me questions if we even needed to support this for non pdo drivers, or if we should, should we wrap the pdo statements so they can be rewound too?

Ocramius:
> On another note, I noticed the build is failing and it's because pdo doesn't actually support rewinding the iterator (well not by default is my understanding, may be able to with unbuffered queries) so made me questions if we even needed to support this for non pdo drivers, or if we should, should we wrap the pdo statements so they can be rewound too?

If an iterator cannot be rewinded, we'll just need to build an accumulator and rewind ourselves (at the cost of memory): providing inconsistent features is worse than providing features at all :-)

jenkoian:
@Ocramius to clarify, if we're going to provider rewindable iterators for non pdo drivers we should add something to do the same for pdo drivers? If so, I can take a look at doing that.

Ocramius:
> we should add something to do the same for pdo drivers?

At least looking at what is feasible first.

Rewindable iterator with emulation means memory usage.
Rewindable iterator should NOT perform a query a second time.

jenkoian:
@Ocramius I've committed an example of a PDO iterator where we store a cache and can be rewound. Just wanted to remind ourselves of what the PR intentions were and where we're at now.

**Initial:** Non PDO drivers were running out of memory when iterating over large data sets due to `fetchAll()` being used and essentially just shoving all results into memory. A generator was added to `MysqliStatement` (with a view to roll out to the other non PDO drivers too) to solve this issue.

**Iterator couldn't be reiterated** or couldn't be rewound. So a `RewindableGenerator` was added so that the iterator could be rewound and thus iterated over multiple times. 

**PDO couldn't be reiterated** noticed after adding an iterator that could be reiterated (rewound) for non PDO drivers that PDO drivers could not actually be reiterated, not sure if this is a bug or by design. 

**Latest** I've added an example of a `PDOStatementIterator` which basically wraps `PDOStatement` and makes it reiterable. This is done via a static cache (basically memoization)  which means that if large data sets are iterated over they will likely fill up memory. 

So...we're basically gone from non PDO drivers running out of memory when iterating over large data to PDO drivers now running out of memory when iterating over large data sets 😆 

Anyway, let me know how you wish to proceed 😄

Ocramius:
@jenkoian the alternative is that we make them ALL non-rewindable, and that's it. That means that the user is left the decision to call `iterator_to_array()` or equivalent, and potentially having the application run out of memory.

Thoughts?

jenkoian:
Options as I can see it are:

**a) Consistent, BC Break**
Make both PDO and NPDO (non PDO) rewindable, this is consistent but breaks BC for PDO as previously large data sets iterated over wouldn't suffer OOM. 

**b) Consistent, doesn't break BC**
Make NPDO iterable without risk of OOM but not rewindable. Leave PDO as non rewindable.

**c) Inconsistent, doesn't break BC**
Make NPDO iterable without risk of OOM and rewindable. Leave PDO as non rewindable.

If you're looking for consistency I would say option b would be the best choice. IMO I think option c is the best and inconsistency here can be tolerated due to shortcoming of PDO.

Ocramius:
> **b) Consistent, doesn't break BC**
 > Make NPDO iterable without risk of OOM but not rewindable. Leave PDO as non rewindable.

Fairly sure we need this scenario then: can't afford OOM

jenkoian:
@Ocramius sorry for the long delay but have removed the rewindable stuff (have left the `RewindableGenerator` in place in case users wish to use it from user land but can remove if you'd prefer). 

To sum up:

Using either PDO or non PDO drivers should now work just fine with large data sets, however neither can be iterated over multiple times without resetting the cursor somehow. 

Do you think it would be worth trying to add a test for this? Perhaps set the php memory limit and then construct a query which will exceed this limit if put into an array or something?

lib/Doctrine/DBAL/RewindableGenerator.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Doctrine\DBAL;
+
+class RewindableGenerator implements \IteratorAggregate
```
Ocramius: Do we still need this then?

Ocramius (approved):
Looking good :+1:

lib/Doctrine/DBAL/RewindableGenerator.php
jenkoian: Left the RewindableGenerator in place in case users wish to use it from user land but can remove if you'd prefer?

Ocramius:
Let's drop it

jenkoian:
@Ocramius dropped, let me know about my questions about an OOM test in https://github.com/doctrine/dbal/pull/2718#issuecomment-317352894 and if you'd like anything documented or anything. I will also squash the commits once everything is finalised also. 👍

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
```diff
@@ -863,6 +863,49 @@ private function skipOci8AndMysqli()
...
+    $this->_stmt->data_seek(0);
+    while ($row = $this->fetch()) {
+        yield $row;
+    }
+});
+
+If this has been done you can include this test. 
+MSG;
+
+        $this->markTestSkipped($skippedMessage);
```
Ocramius: Hmm, this test is skipped all the time now. Can we get a better test for the introduced change? Maybe memory-usage related?

Ocramius (changes requested):

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
morozov: We can wrap the underlying statement in a test double and make sure that none of the `fetch*()` methods are invoked before iteration, and then `fetch()` is invoked at each iteration step. This way we'll enforce the messaging protocol between the iterator and the statement.

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
```diff
@@ -863,6 +863,49 @@ private function skipOci8AndMysqli()
...
+        $sql = 'SELECT CURTIME(6) AS time_started, test_int, test_string FROM fetch_table';
+        $stmt = $this->_conn->query($sql);
+        $stmt->setFetchMode(\PDO::FETCH_ASSOC);
+
+        $timeStartedMsFirstLoop = null;
+        $timeStartedMsSecondLoop = null;
+
+        $i = 0;
+        foreach ($stmt as $row) {
+            $timeStartedMsFirstLoop = $row['time_started'];
```
morozov: This would probably fail on Oracle and DB2 without the portability layer. They will return the column name in the upper case.

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
Ocramius: That's a brilliant approach, good one :+1:

jenkoian: Sorry, not sure I totally understand. 

I think you're suggesting to decorate the statement with a spy of some kind so that we can examine that `fetch()` is being called as we'd expect? 

If that is correct, I'm still not 100% sure how to implement that. Statements, as far as I can tell, [are new'd up in the relevant connection class](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php#L120), meaning that we'd have to mock the connection or rewrite to allow a statement to be injected. Is this the path you were thinking of or am I way off?

morozov: @jenkoian I was thinking about mocking the `Connection::prepare()` implementations and returning a `Statement` mock instead of real statement. The statement mock will spy on the method invocations. I don't think we need to make statement classes configurable at least for this test. Not sure how much boilerplate PHPUnit code is required to achieve that.

Do you see any roadblocks with this approach?

jenkoian: Thanks for the additional info, have pushed a commit which I *think* does what you are talking about, would you mind taking a look and letting me know? https://github.com/doctrine/dbal/pull/2718/commits/07d5f6f8bf5fff49eb1f4f4efd5bf3e59ad377d0

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
```diff
@@ -863,6 +864,84 @@ private function skipOci8AndMysqli()
...
+        self::assertEquals(0, $stmtSpy->getFetchCalls());
+
+        $i = 0;
+        foreach ($stmt as $row) {
+            $i++;
+        }
+
+        // Assert expected number of fetches
+        $this->assertGreaterThan(0, $stmtSpy->getFetchCalls());
+        $this->assertEquals($i, $stmtSpy->getFetchCalls());
```
morozov: It doesn't guarantee that `fetch()` is called once per step. It would also pass if the code did `while (fetch()); return new ArrayIterator($rows)`;

morozov (commented):
@jenkoian please see pseudocode in review comments.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
```diff
@@ -199,9 +199,9 @@ public function setFetchMode($fetchMode, $arg2 = null, $arg3 = null)
      */
     public function getIterator()
     {
-        $data = $this->fetchAll();
-
-        return new \ArrayIterator($data);
+        while ($row = $this->fetch()) {
```
morozov: What if the fetch mode is `PDO::FETCH_COLUMN` and one of the values is empty? Probably should strictly compare with `FALSE` and add this to the test.

morozov (changes requested):

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
morozov: I'd replace `$row` with something more generic like `$result`. Otherwise, `$row` creates a false impression that it's always an array, and an empty value means the end of the result set.

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
```diff
@@ -863,6 +864,84 @@ private function skipOci8AndMysqli()
...
+    {
+        $sql = 'SELECT CURTIME(6) AS time_started, test_int, test_string FROM fetch_table';
+        $originalStmt = $this->_conn->prepare($sql);
+
+        // Mock the connection
+        $connectionMock = $this->getMockBuilder('Doctrine\\DBAL\\Driver\\Connection')
+            ->setMethods(['prepare', 'query'])
+            ->getMockForAbstractClass();
+
+        $stmtSpy = new StatementSpy($originalStmt);
```
jenkoian: Cheers, struggling to find time to look at this right now, but hopefully will be able to pick it back up either later today or tomorrow. Appreciate your help though 👍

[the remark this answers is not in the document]
jenkoian: > But we can test all of them in the same suite using data provider

I'm not sure about this, as not every driver will get loaded you could end up with errors regarding missing class constants etc.

> I don't know how to invoke getIterator() implicitly without calling current()/next() and without depending on the actual implementation

This is why I created a spy manually. I tried the inner iterator thing you mentioned but couldn't get it to work.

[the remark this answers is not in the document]
morozov: > I'm not sure about this, as not every driver will get loaded you could end up with errors regarding missing class constants etc.

On the mocked statement classes, the test will only call the `getIterator()` methods which do not use any driver-specific constants.

> This is why I created a spy manually. I tried the inner iterator thing you mentioned but couldn't get it to work.

But you still can call  `getIterator()` explicitly as in the example above. I think it's not a big problem.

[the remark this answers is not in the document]
jenkoian: The following test throws up the error `1) Doctrine\Tests\DBAL\Functional\DataAccessTest::testGettingIteratorDoesNotCallFetch with data set #1 ('Doctrine\DBAL\Driver\OCI8\OCI...ection')
Use of undefined constant OCI_COMMIT_ON_SUCCESS - assumed 'OCI_COMMIT_ON_SUCCESS'`

Admittedly this is likely because I don't have the OCI8 extension loaded, but then seems a shame to have to have all extensions loaded to run unit tests against just one driver.

Running the test with `vendor/bin/phpunit --filter DataAccessTest::testGettingIteratorDoesNotCallFetch --configuration tests/travis/mysqli.travis.xml`

```php
    public function getConnections()
    {
        return [
            [MysqliConnection::class],
            [OCI8Connection::class],
        ];
    }

    /**
     * @dataProvider getConnections
     */
    public function testGettingIteratorDoesNotCallFetch($connection)
    {
        $stmt = $this->createMock(Statement::class);
        $stmt->expects($this->never())->method('fetch');

        $conn = $this->createMock($connection);
        $conn->expects($this->any())->method('prepare')->willReturn($stmt);

        $ii = new IteratorIterator($stmt);
        $ii->getInnerIterator();
    }
```

> But you still can call getIterator() explicitly as in the example above. I think it's not a big problem.

Will need some extra clarification about this one. You can call `getIterator()` but it's on a mock, so doesn't return anything? Only way I can see of doing it is by basically re-writing the implementation within the mock?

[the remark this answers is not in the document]
morozov: > The following test throws up the error `1) Doctrine\Tests\DBAL\Functional\DataAccessTest::testGettingIteratorDoesNotCallFetch with data set #1 ('Doctrine\DBAL\Driver\OCI8\OCI...ection')
Use of undefined constant OCI_COMMIT_ON_SUCCESS - assumed 'OCI_COMMIT_ON_SUCCESS'`

Could you try moving this to constructor?
 https://github.com/doctrine/dbal/blob/4da603c97002c4a09f271d96bf808e62d8a3b9dc/lib/Doctrine/DBAL/Driver/OCI8/OCI8Connection.php#L41

> Will need some extra clarification about this one. You can call `getIterator()` but it's on a mock, so doesn't return anything? Only way I can see of doing it is by basically re-writing the implementation within the mock?

`getIterator()` shouldn't be mocked since we're testing it. Please try using something else than `createMock()` — this one will mock all methods. `createPartialMock()` may be a good alternative.

Please feel free to commit the failing test so that I could offer a fix if you're tired of rewriting it.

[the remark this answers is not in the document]
jenkoian: How did I not know about partial mocks?! 😩 

Ok this helps a lot, I'm still struggling a bit with the second test. Have pushed it up.

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
```diff
@@ -865,86 +875,54 @@ private function skipOci8AndMysqli()
...
-        }
-    }
+        $conn = $this->createMock($connection);
+        $conn->expects($this->any())->method('prepare')->willReturn($stmt);
 
-    public function getFetchCalls()
-    {
-        return $this->fetchCalls;
+        foreach ($stmt as $i => $_) {
+            $_->next();
```
morozov: This is not something I'd expect the code doing. Why should `next()` be called on the value returned by iterator? If I understand it correctly, `$stmt` is the iterator, and `foreach` will call `next()` on it.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
```diff
@@ -60,16 +60,18 @@ class DB2Statement implements \IteratorAggregate, Statement
...
+    static private $_typeMap;
 
     /**
      * @param resource $stmt
      */
     public function __construct($stmt)
     {
+        self::$_typeMap = [
+            \PDO::PARAM_INT => DB2_LONG,
+            \PDO::PARAM_STR => DB2_CHAR,
```
morozov: That's not something I anticipated and it really sucks (my bad). The static type map shouldn't be updated upon each object instantiation.

What if we take another route: right now, `getIterator()` is implemented exactly the same in all statements and we test all implementations with one test. What if instead, we move it out of the statements into a separate class and test it with a statement mock. In this case we'll be able to mock the statement interface, not test all implementations.

```php
class StatementIterator implements IteratorAggregate
{
    public function getIterator()
    {
        while (false !== ($result = $this->stmt->fetch())) {
            yield $result;
        }
    }
}

class WhateverStatement
{
    public function getIterator()
    {
        return new StatementIterator($this);
    }
}
```

The test logic will remain the same with the only exception that we'll be testing the statement iterator, not the statement classes.

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
jenkoian: Yeah agreed, it didn't seem to iterate over without calling next though. Hopefully the changes you mention below about extracting the iterator will help here.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
jenkoian: Ok I've made this change. The testIterationCallsFetchOncePerStep test doesn't pass. Without calling next (which I agree isn't right) it doesn't ever get into the callback of the statement mock. Any ideas?

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
```diff
@@ -863,9 +876,55 @@ private function skipOci8AndMysqli()
...
     }
-}
 
-class MyFetchClass
-{
-    public $test_int, $test_string, $test_datetime;
+    public function getConnections()
+    {
+        return [
+            [MysqliConnection::class],
```
morozov: If we extract the iteration logic into a separate class, we shouldn't be testing the connection/statement classes anymore. This can be moved out to a brand new `StatementIteratorTest` class and tested with a `Statement` mock. With this approach, we'll not need to instantiate connection/statement classes and deal with their runtime dependencies.

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereStatement.php
```diff
@@ -278,7 +278,9 @@ public function fetchColumn($columnIndex = 0)
      */
     public function getIterator()
     {
-        return new \ArrayIterator($this->fetchAll());
+        while ($row = $this->fetch()) {
```
morozov: Should it also use the new `StatementIterator`?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Connection.php
```diff
@@ -58,6 +57,8 @@ public function __construct($username, $password, $db, $charset = null, $session
             define('OCI_NO_AUTO_COMMIT', 0);
         }
 
+        $this->executeMode = OCI_COMMIT_ON_SUCCESS;
+
```
morozov: This can be reverted.

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
jenkoian: Ok, did some tidy up, moved tests, had to change the mock a little bit, the yield didn't quite work. Cheers for your help.

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
```diff
@@ -0,0 +1,38 @@
...
+
+use Doctrine\DBAL\Driver\Statement;
+use Doctrine\DBAL\Driver\StatementIterator;
+
+class StatementIteratorTest extends \Doctrine\Tests\DbalTestCase
+{
+    public function testGettingIteratorDoesNotCallFetch()
+    {
+        $stmt = $this->createMock(Statement::class);
+        $stmt->expects($this->never())->method('fetch');
```
morozov: I'd also add all other `fetch*` methods here (`fetchAll()`, `fetchColumn()`).

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -20,6 +20,8 @@
 namespace Doctrine\DBAL\Driver\OCI8;
 
 use Doctrine\DBAL\Driver\Statement;
+use Doctrine\DBAL\Driver\StatementIterator;
+use \IteratorAggregate;
```
morozov: The leading namespace separator is redundant.

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
```diff
@@ -0,0 +1,38 @@
...
+
+        $stmtIterator = new StatementIterator($stmt);
+        $stmtIterator->getIterator();
+    }
+
+    public function testIterationCallsFetchOncePerStep()
+    {
+        $values = ['foo', '', 'bar', '0', 'baz', 0, 'qux', null, 'quz', false];
```
morozov: We need to add one more value after `false` to prove that fetching `false` terminates iteration, while other empty values don't (see original example).

morozov (commented):
Only a few minor issues. Looks good overall.

jenkoian:
Thanks @morozov - minor issues resolved. Anything else I need to do, or can we get this in? @Ocramius

morozov (approved):

Ocramius (approved):
Excellent work! It took a while, but we moved from a complex implementation to a really simple, clean and maintainable one. Thanks @jenkoian and @morozov!

Ocramius:
:ship:

PR #2724, opened 2017-05-14
Improve `Table#getColumns()` performance by reducing its runtime complexity

### LINKED ISSUE #2722: Table->getColumns() performance
I was profiling a (Symfony2) application and found that `array_search` was called over 17.000 times during one unit test. Almost all of these calls come from [Table->getColumns()](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/Table.php#L637) (`uksort` sorting function).

Replacing this `uksort` statement with the following won me almost a second (and got rid of the 17.000 array_search calls):
```
return array_replace(array_flip($colNames), $columns);
```

Not entirely sure if this does produce any side effects. I read that `array_replace` and `array_merge` are equivalent in this scenario, but `array_replace` seemed faster in my experience.

Anyone sees a problem with this improvement?

evgpisarchik:
Resolves #2722 
Seems resonable proposal by @verheyenkoen
Could you review these changes please?

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -742,7 +742,7 @@ class MySqlPlatform extends AbstractPlatform
         foreach ($diff->changedIndexes as $changedIndex) {
             // Changed primary key
             if ($changedIndex->isPrimary() && $diff->fromTable instanceof Table) {
-                foreach ($diff->fromTable->getPrimaryKeyColumnNames() as $columnName) {
+                foreach ($diff->fromTable->getPrimaryKeyColumns() as $columnName) {
                     $column = $diff->fromTable->getColumn($columnName);
 
                     // Check if an autoincrement column was dropped from the primary key.
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -742,7 +742,7 @@ class MySqlPlatform extends AbstractPlatform
         foreach ($diff->changedIndexes as $changedIndex) {
             // Changed primary key
             if ($changedIndex->isPrimary() && $diff->fromTable instanceof Table) {
-                foreach ($diff->fromTable->getPrimaryKeyColumns() as $columnName) {
+                foreach ($diff->fromTable->getPrimaryKeyColumnNames() as $columnName) {
                     $column = $diff->fromTable->getColumn($columnName);
 
                     // Check if an autoincrement column was dropped from the primary key.
--- a/lib/Doctrine/DBAL/Schema/Table.php
+++ b/lib/Doctrine/DBAL/Schema/Table.php
@@ -621,58 +621,38 @@ class Table extends AbstractAsset
      */
     public function getColumns()
     {
-        return array_merge($this->getPrimaryKeyColumns(), $this->getForeignKeyColumns(), $this->_columns);
-    }
+        $primaryKeyColumns = [];
+        if ($this->hasPrimaryKey()) {
+            $primaryKeyColumns = $this->filterColumns($this->getPrimaryKey()->getColumns());
+        }
 
-    /**
-     * Returns primary key columns
-     * @return Column[]
-     */
-    public function getPrimaryKeyColumns()
-    {
-        $primaryKeyColumnNames = $this->getPrimaryKeyColumnNames();
-        return array_filter($this->_columns, function ($key) use ($primaryKeyColumnNames) {
-            return in_array($key, $primaryKeyColumnNames);
-        }, ARRAY_FILTER_USE_KEY);
+        return array_merge($primaryKeyColumns, $this->getForeignKeyColumns(), $this->_columns);
     }
 
     /**
      * Returns foreign key columns
      * @return Column[]
      */
-    public function getForeignKeyColumns()
+    private function getForeignKeyColumns()
     {
-        $foreignKeyColumnNames = $this->getForeignKeyColumnNames();
-        return array_filter($this->_columns, function ($key) use ($foreignKeyColumnNames) {
-            return in_array($key, $foreignKeyColumnNames);
-        }, ARRAY_FILTER_USE_KEY);
-    }
-
-    /**
-     * Returns primary key column names
-     * @return array
-     */
-    public function getPrimaryKeyColumnNames()
-    {
-        $primaryKeyColumnNames = [];
-        if ($this->hasPrimaryKey()) {
-            $primaryKeyColumnNames = $this->getPrimaryKey()->getColumns();
-        }
-        return $primaryKeyColumnNames;
-    }
-
-    /**
-     * Returns foreign key column names
-     * @return array
-     */
-    public function getForeignKeyColumnNames()
-    {
-        $foreignKeyColumnNames = [];
+        $foreignKeyColumns = [];
         foreach ($this->getForeignKeys() as $foreignKey) {
             /* @var $foreignKey ForeignKeyConstraint */
-            $foreignKeyColumnNames = array_merge($foreignKeyColumnNames, $foreignKey->getColumns());
+            $foreignKeyColumns = array_merge($foreignKeyColumns, $foreignKey->getColumns());
         }
-        return $foreignKeyColumnNames;
+        return $this->filterColumns($foreignKeyColumns);
+    }
+
+    /**
+     * Returns only columns that have specified names
+     * @param array $columnNames
+     * @return Column[]
+     */
+    private function filterColumns(array $columnNames)
+    {
+        return array_filter($this->_columns, function ($columnName) use ($columnNames) {
+            return in_array($columnName, $columnNames, true);
+        }, ARRAY_FILTER_USE_KEY);
     }
 
     /**
@@ -720,6 +700,21 @@ class Table extends AbstractAsset
         }
 
         return $this->getIndex($this->_primaryKeyName);
+    }
+
+    /**
+     * Returns the primary key columns.
+     *
+     * @return array
+     *
+     * @throws DBALException
+     */
+    public function getPrimaryKeyColumns()
+    {
+        if ( ! $this->hasPrimaryKey()) {
+            throw new DBALException("Table " . $this->getName() . " has no primary key.");
+        }
+        return $this->getPrimaryKey()->getColumns();
     }
 
     /**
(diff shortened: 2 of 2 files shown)


lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -634,11 +634,7 @@ public function getColumns()
         }
         $colNames = array_unique(array_merge($pkCols, $fkCols, array_keys($columns)));
 
-        uksort($columns, function ($a, $b) use ($colNames) {
-            return (array_search($a, $colNames) >= array_search($b, $colNames));
-        });
-
-        return $columns;
+        return array_replace(array_flip($colNames), $columns);
```
Ocramius: I must say that I neither understand this nor the original code. What is the intent: sorting columns? Is there a corresponding test somewhere?

Ocramius (changes requested):

lib/Doctrine/DBAL/Schema/Table.php
verheyenkoen: For some reasons PK columns are put first, the FK columns, then the other ones. No idea why that was implemented like this. No idea if anyone relies on that...

Ocramius: That is indeed the "standard approach" when writing DDL, as it otherwise is really hard to figure out what is going on. Still, is there no `O(n * log(n))` or similar solution that doesn't make the code unreadable? Original code was heavy due to the repeated `array_search()` calls, which make this `O(n * log(n) * n * log(n))`, so your approach is good, just not understandable at all.

verheyenkoen: You mean this may be better if the `array_replace` line is refactored in a separate private method with a meaningful name?

evgpisarchik: Is it more comprehensible?
```
  /**
     * Returns ordered list of columns (primary keys are first, then foreign keys, then the rest)
     * @return Column[]
     */
    public function getColumns()
    {
        return array_merge($this->getPrimaryKeyColumns(), $this->getForeignKeyColumns(), $this->_columns);
    }

    /**
     * Returns primary key columns
     * @return array
     */
    public function getPrimaryKeyColumns()
    {
        return array_filter($this->_columns, function ($key) {
            return in_array($key, $this->getPrimaryKeyColumnNames());
        }, ARRAY_FILTER_USE_KEY);
    }

    /**
     * Returns foreign key columns
     * @return array
     */
    public function getForeignKeyColumns()
    {
        return array_filter($this->_columns, function ($key) {
            return in_array($key, $this->getForeignKeyColumnNames());
        }, ARRAY_FILTER_USE_KEY);
    }

    /**
     * Returns primary key column names
     * @return array
     */
    public function getPrimaryKeyColumnNames()
    {
        $primaryKeyColumnNames = [];
        if ($this->hasPrimaryKey()) {
            $primaryKeyColumnNames = $this->getPrimaryKey()->getColumns();
        }
        return $primaryKeyColumnNames;
    }

    /**
     * Returns foreign key column names
     * @return array
     */
    public function getForeignKeyColumnNames()
    {
        $foreignKeyColumnNames = [];
        foreach ($this->getForeignKeys() as $foreignKey) {
            /* @var $foreignKey ForeignKeyConstraint */
            $foreignKeyColumnNames = array_merge($foreignKeyColumnNames, $foreignKey->getColumns());
        }
        return $foreignKeyColumnNames;
    }
```

verheyenkoen: @evgpisarchik That seems not too efficient as calls to `getPrimaryKeyColumnNames()` and `getForeignKeyColumnNames()` are called multiple times. Am I right?

evgpisarchik: I'm using php 5.6+ syntax since [composer.json](https://github.com/doctrine/dbal/blob/master/composer.json#L15) has
```
"require": {
        "php": "^7.0",
```

Ocramius: You can use PHP 7 syntax, as `master` requires PHP 7.x

evgpisarchik: @verheyenkoen both mehods are called 1 time per getColumns() invocation (as many times as in proposal you made)

evgpisarchik: @verheyenkoen you're right, I've updated the code

evgpisarchik:
Updated previously pushed code, squashed both commits

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -742,7 +742,7 @@ private function getPreAlterTableAlterIndexForeignKeySQL(TableDiff $diff)
         foreach ($diff->changedIndexes as $changedIndex) {
             // Changed primary key
             if ($changedIndex->isPrimary() && $diff->fromTable instanceof Table) {
-                foreach ($diff->fromTable->getPrimaryKeyColumns() as $columnName) {
+                foreach ($diff->fromTable->getPrimaryKeyColumnNames() as $columnName) {
```
Ocramius: Why is this not simply `$diff->fromTable->getPrimaryKeyColumns()` and then on line [`746`](https://github.com/doctrine/dbal/pull/2724/files#diff-23f1dda4df961964343780b794ea8095R746) `$columnName->getName()` ?

lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -616,29 +616,63 @@ public function removeForeignKey($constraintName)
     }
 
     /**
+     * Returns ordered list of columns (primary keys are first, then foreign keys, then the rest)
      * @return Column[]
      */
     public function getColumns()
     {
-        $columns = $this->_columns;
+        return array_merge($this->getPrimaryKeyColumns(), $this->getForeignKeyColumns(), $this->_columns);
```
Ocramius: This looks much cleaner :+1:

lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -616,29 +616,63 @@ public function removeForeignKey($constraintName)
...
-        $fkCols = array();
+    /**
+     * Returns primary key columns
+     * @return Column[]
+     */
+    public function getPrimaryKeyColumns()
+    {
+        $primaryKeyColumnNames = $this->getPrimaryKeyColumnNames();
+        return array_filter($this->_columns, function ($key) use ($primaryKeyColumnNames) {
+            return in_array($key, $primaryKeyColumnNames);
```
Ocramius: Use strict comparisons at all times

lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -689,22 +723,6 @@ public function getPrimaryKey()
...
-     * Returns the primary key columns.
-     *
-     * @return array
-     *
-     * @throws DBALException
-     */
-    public function getPrimaryKeyColumns()
-    {
-        if ( ! $this->hasPrimaryKey()) {
-            throw new DBALException("Table " . $this->getName() . " has no primary key.");
```
Ocramius: This exception is gone - that is a BC break, as now an empty list of primary key columns is returned

lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -616,29 +616,63 @@ public function removeForeignKey($constraintName)
...
-        });
+        return $primaryKeyColumnNames;
+    }
 
-        return $columns;
+    /**
+     * Returns foreign key column names
+     * @return array
+     */
+    public function getForeignKeyColumnNames()
```
Ocramius: Doesn't need to be `public`

lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -616,29 +616,63 @@ public function removeForeignKey($constraintName)
...
+        return array_filter($this->_columns, function ($key) use ($primaryKeyColumnNames) {
+            return in_array($key, $primaryKeyColumnNames);
+        }, ARRAY_FILTER_USE_KEY);
+    }
 
+    /**
+     * Returns foreign key columns
+     * @return Column[]
+     */
+    public function getForeignKeyColumns()
```
Ocramius: Doesn't need to be `public`

lib/Doctrine/DBAL/Schema/Table.php
Ocramius: Using `$this->getPrimaryKeyColumns()` in this method will fail if there are no primary keys, and the BC break in `getPrimaryKeyColumns` is reverted (exception throwing on empty PK)

lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -616,29 +616,63 @@ public function removeForeignKey($constraintName)
...
+        return array_filter($this->_columns, function ($key) use ($foreignKeyColumnNames) {
+            return in_array($key, $foreignKeyColumnNames);
+        }, ARRAY_FILTER_USE_KEY);
+    }
+
+    /**
+     * Returns primary key column names
+     * @return array
+     */
+    public function getPrimaryKeyColumnNames()
```
Ocramius: Doesn't need to be `public`

Ocramius (changes requested):
> squashed both commits

Urgh, now I can't compare it with the previous solution. Please don't squash commits next time :S

As far as I can see, this patch added `Table#getForeignKeyColumns()`, `Table#getPrimaryKeyColumnNames()`, `Table#getForeignKeyColumnNames()`, which are all now to be tested to prevent BC regressions.

I appreciate your effort, I really do, but the newly exposed public API is a problem.

evgpisarchik:
Reverted changes to public Api there were made before (now getPrimaryKeyColumns returns array of strings (not Column[]).

getColumns method is cleaner that it was before, but has the same behavior as the [original code](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/Table.php#L618-L642)  (it doesn't throw exception if table doesn't have primary keys )

evgpisarchik:
Reverted changes to public Api there were made before (now getPrimaryKeyColumns returns array of strings (not Column[]).

getColumns method is cleaner that it was before, but has the same behavior as the original code (it doesn't throw exception if table doesn't have primary keys )

Ocramius:
@evgpisarchik perfect, thanks for all the effort! \o/

:ship:

evgpisarchik:
my pleasure!

PR #2726, opened 2017-05-15
Warning: implode(): Invalid arguments passed in SimpleArrayType.php

JarJak:
Warning: implode(): Invalid arguments passed in SimpleArrayType.php line 52

The warning happens when you set field value to something different than array, ie. string.

I will provide test later.

### DIFF
--- a/lib/Doctrine/DBAL/Types/SimpleArrayType.php
+++ b/lib/Doctrine/DBAL/Types/SimpleArrayType.php
@@ -47,6 +47,10 @@ class SimpleArrayType extends Type
         if (!$value) {
             return null;
         }
+        
+        if (!is_array($value)) {
+            $value = array($value);
+        }
 
         return implode(',', $value);
     }


lib/Doctrine/DBAL/Types/SimpleArrayType.php
```diff
@@ -47,6 +47,10 @@ public function convertToDatabaseValue($value, AbstractPlatform $platform)
         if (!$value) {
             return null;
         }
+        
+        if (!is_array($value)) {
```
deeky666: I personally don't think we should allow non-array values for an array type. This looks like a misusage of the type.

kimhemsoe: @deeky666 Agree

deeky666:
Closing as per the comments provided. @JarJak please make sure your calling code passes an array or if you really need this type of semantics, you can still implement your own custom type.

JarJak:
Then the type should be checked and throw an Exception if it's not an array.

Ocramius:
The same comments apply to #2727 IMO, although I can't close it because WIFI is terrible here ATM.

PR #2727, opened 2017-05-16
Warning: implode(): Invalid arguments passed in SimpleArrayType.php

JarJak:
Updated version of #2726

### DIFF
--- a/lib/Doctrine/DBAL/Types/SimpleArrayType.php
+++ b/lib/Doctrine/DBAL/Types/SimpleArrayType.php
@@ -48,9 +48,7 @@ class SimpleArrayType extends Type
             return null;
         }
         
-        if (!is_array($value)) {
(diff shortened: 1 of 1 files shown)


Ocramius:
As much as I'd like this to be in every type, the added overhead is always a big issue, especially in types.

In addition to that, the type assumes that every value is stringable, yet this patch doesn't check for that.

The problem is exactly that: use a type checker like `phpstan/phpstan` and document your object properties accordingly, since runtime type checking is expensive (when not delegated to the engine), and it's a rabbit hole that is pretty much infinite.

jvasseur:
Maybe we could use `assert` to add some runtime checks that could be disabled on production servers to remove the overhead.

JarJak:
@jvasseur now it uses assert(), thanks for suggestion

JarJak:
@morozov why closed?

PR #2731, opened 2017-05-18
Adds missing return statement on SQLSrvStatement.execute()

fischerman:
The documentation states that the execute function requires a boolean return value. Based on `MysqliStatement` the return statement is missing on `SqlSrvStatement`.

The prevents `dbal:import` to run properly because it evaluates null to false although the statement ran successfully.


Ocramius:
Requires a test case

fischerman:
@Ocramius I won't have time to create a test case

Ocramius:
Closing as `incomplete` then.

PR #2733, opened 2017-05-22
Correct documented parameter and return types for `Connection#quote()`

lolli42:
Single drivers get and return various types in quote(), the
interface docs should reflect that.


Ocramius:
LGTM! :+1:

PR #2736, opened 2017-06-06
Correcting references to `Driver\Connection` in the documentation

b0nd0:
At least it looks like one to me.

### DIFF
--- a/docs/en/reference/architecture.rst
+++ b/docs/en/reference/architecture.rst
@@ -46,7 +46,7 @@ The drivers abstract a PHP specific database API by enforcing two
 interfaces:
 
 
--  ``\Doctrine\DBAL\Driver\Driver``
+-  ``\Doctrine\DBAL\Driver\Connection``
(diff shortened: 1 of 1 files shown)


Ocramius (changes requested):
Should probably be https://github.com/doctrine/dbal/blob/v2.5.0/lib/Doctrine/DBAL/Driver.php

Also, please make a more descriptive commit/issue title 👍

b0nd0:
Thanks for such quick response, i'll try to name pull requests better next time.
But speaking about changes, as far as i understand:
```
The drivers abstract a PHP specific database API by enforcing two interfaces:
...
The above two interfaces require exactly the same methods as PDO.
```
it should be `connection` and `statement`, isn't it? DBAL\Driver doesn't look like PHP PDO class

Ocramius:
@b0nd0 ah, yep, my bad!

Ocramius (approved):

Ocramius:
@b0nd0 thanks! \o/

PR #2739, opened 2017-06-07
Fix: executionMS should be milliseconds not seconds

Korri:
Looks like this is a mistake.


Ocramius:
@Korri possibly a naming mistake, but the tooling already relies on this scale....

PR #2742, opened 2017-06-13
Use short array declarations in the `Driver` namespace

AlessandroMinoccheri:
Upgraded all declaration and using of array inside Driver directory


Ocramius (approved):
@AlessandroMinoccheri thanks!

Issue #2706, opened 2017-04-27
[Question] PostgreSQL's pgcrypto uuid generator


Simperfit:
The postgreSQL Driver does not support [pgcrypto](https://www.postgresql.org/docs/9.6/static/pgcrypto.html)'s function to generate uuid (`gen_random_uuid() returns uuid`, it does support uuid-ossp but it is currently unavailable in some provider like Google Cloud SQL (postgres).

Should dbal supports pgcrypto too ? 

I'm willing to help contributing to this if it's the case

Ocramius:
> Should dbal supports pgcrypto too ?

Depends on whether it's provided OOTB, and if so, which version implemented it

Simperfit:
@Ocramius It is, just like uuid-ossp, I think it exist since a long time ago (maybe pgsql  8.2)

Ocramius:
@Simperfit so I suppose you just want to change https://github.com/doctrine/dbal/blob/fbed76eb8395d9a8c159cfaeceb775dfc6be7d9f/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php#L999 ?

Simperfit:
@Ocramius I think we should add a new platform and detect if it's either one or either the other in order to not do a BC Break with existing apps that use uuid-ossp.

WDYT ?

Ocramius:
@Simperfit if we're moving to something that is more widely supported, and the output is the same, then it's fine to just change it as-is

morozov:
The UUID functionality is deprecated in DBAL by https://github.com/doctrine/dbal/pull/3212.

PR #2743, opened 2017-06-13
feature: support pgcrypto

Simperfit:
As said in #2706 Google Cloud SQL postgres supports only pgcrypto


Ocramius:
This looks good to me, but do we have a way to verify that the change is valid via testing?

Simperfit:
@Ocramius I'm using this since the issue in production with symfony and we are generating uuid correctly. Do you what me to add tests ?

Ocramius:
If possible, yes. Identifying existing tests that already cover this would
also be enough.

morozov:
Obsolete as of https://github.com/doctrine/dbal/pull/4082.

PR #2746, opened 2017-06-20
Using short array declarations in the `MasterSlaveConnection` namespace

AlessandroMinoccheri:
fixed array declarations inside MasterSlaveConnection file


Ocramius (approved):
LGTM thanks!

Issue #2745, opened 2017-06-19
Wrong param comment 


helsner:
vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php
Line 762
The comment for the parameter is wrong in this case
$type is null or a \PDO::PARAM which is type integer

Ocramius:
There is nothing at [line `762`](https://github.com/doctrine/dbal/blob/899c62bb579027c146b73f6b580c871eb669d310/lib/Doctrine/DBAL/Connection.php#L762)

Consider sending a patch with the changed file: you can do that directly [from the UI](https://github.com/doctrine/dbal/edit/master/lib/Doctrine/DBAL/Connection.php)

Ocramius:
Handled in #2747

PR #2747, opened 2017-06-20
Correct `Connection#quote()` parameter hinting, now compatible with `PDO::PARAM_*` constants

helsner:
The param comment is currently set to string, which is wrong.
The correct hint would be integer/int

Issue: #2745


Ocramius:
@helsner thanks!

PR #2749, opened 2017-06-20
Add number of maximum allowed placeholders in prepared statements

### LINKED ISSUE #2748: Add maximum number of prepared statements to Platform
The platform should provide an information how many values can be bound in a prepared statement.

See also
https://stackoverflow.com/questions/6581573/what-are-the-max-number-of-allowable-parameters-per-database-provider-type

sascha-egerer:
The platform should provide an information how many values can be bound
in a prepared statement.

This adds the possibility to split up big quries if the limit of placeholders
is reached.

Closes #2748

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -601,6 +601,16 @@ abstract class AbstractPlatform
     }
 
     /**
+     * Gets the maximum number of placeholders for prepared statements
+     *
+     * @return integer
+     */
+    public function getMaximumNumberOfPreparedStatementPlaceholders()
+    {
+        return 950;
+    }
+
+    /**
      * Gets all SQL wildcard characters of the platform.
      *
      * @return array
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -1066,6 +1066,14 @@ class MySqlPlatform extends AbstractPlatform
     }
 
     /**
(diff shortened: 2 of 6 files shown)


lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -600,6 +600,16 @@ public function getBinaryDefaultLength()
         return 255;
     }
 
+    /**
+     * Gets the maximum number of placeholders for prepared statements
+     *
+     * @return integer
+     */
+    public function getMaximumNumberOfPreparedStatementPlaceholders()
```
Majkl578: Can you please express the type natively and remove `@return`?

Majkl578 (changes requested):
Please add some tests as well.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -600,6 +600,16 @@ public function getBinaryDefaultLength()
...
+    {
+        return 950;
```
morozov: How can an abstract platform have a concrete value defined?

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -1065,6 +1065,14 @@ public function getBinaryMaxLength()
         return 65535;
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function getMaximumNumberOfPreparedStatementPlaceholders()
+    {
+        return 65535;
```
morozov: In non-abstract platforms, please add links to the platform documentation which the numbers are taken from,

PR #2754, opened 2017-06-28
problem with GeneratedValue (uuid type) column query mapping

luked3v3l0p3r:


Ocramius:
Requires a failing test case first

Ocramius:
Also closing as `incomplete`: please target `master` for a PR with tests

luked3v3l0p3r:
Hi @Ocramius ,
i made pull request to 2.5 branch because master requires php 7.0.

Please let me give You background of the problem which i want to fix.
Query builder for sqlsrv driver has an issue with binding parameters:

$qb = $this->createQueryBuilder('g');
...
$qb->expr()->eq('g.parent', ':group')
$qb->setParameter('group', $group);

Definition of id column in Group Entity is:

    /**
     * @ORM\Id
     * @ORM\Column(type="uuid")
     * @ORM\GeneratedValue(strategy="CUSTOM")
     * @ORM\CustomIdGenerator(class="DoctrineUuidGenerator")
     */

Doctrine parameter "binder" causes a problem because getId() method returns object of instance Uuid instead of string. The simple workaround which is working properly without changing code of the sqlsrv driver: 

$qb->setParameter('group', $group->getId()->toString());

could You give me an example how such test could looks like? Should I create an Entity, mock it and write a test to show the problem doesn't exist anymore?

Ocramius:
> i made pull request to 2.5 branch because master requires php 7.0.

We backport it if it's worth backporting - please open a PR against `master` only ;-)

 > Doctrine parameter "binder" causes a problem because getId() method returns object of instance Uuid instead of string. The simple workaround which is working properly without changing code of the sqlsrv driver:

This is probably missing type conversion to be applied, which is where the test is needed.

See https://github.com/doctrine/doctrine2/tree/v2.5.6/tests/Doctrine/Tests/ORM/Functional/Ticket for test examples

PR #2756, opened 2017-06-29
Add Scrutinizer-CI to the build pipeline

malukenho:
- [x] Add badges to the README

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,0 +1,24 @@
+build:
+    environment:
+        php:
+            version: 7.1
+
+before_commands:
+    - "composer install --no-dev --prefer-source"
+
+tools:
+    external_code_coverage:
+        timeout: 3600
+
+filter:
+    excluded_paths:
+        - docs
+        - tools
+
+build_failure_conditions:
+    - 'elements.rating(<= C).new.exists'                        # No new classes/methods with a rating of C or worse allowed
+    - 'issues.label("coding-style").new.exists'                 # No new coding style issues allowed
+    - 'issues.severity(>= MAJOR).new.exists'                    # New issues of major or higher severity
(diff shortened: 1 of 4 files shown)


.scrutinizer.yml
```diff
@@ -0,0 +1,24 @@
...
+    - "composer install --no-dev --prefer-source"
+
+tools:
+    external_code_coverage:
+        timeout: 3600
+
+filter:
+    excluded_paths:
+        - docs
+        - tools
```
lcobucci: This folder doesn't exist here

Ocramius (approved):
Extra points for the :beer:!

Ocramius (approved):
LGTM :+1:

.travis.yml
```diff
@@ -112,6 +112,9 @@ matrix:
   allow_failures:
     - php: nightly
 
+after_script:
+  - sh .travis.coverage.sh
```
lcobucci: I'd rather keep the config here so we won't have to maintain multiple files (kind of what we have in the ORM)

.travis.coverage.sh
```diff
@@ -0,0 +1,5 @@
+set -x
+if [ "$TRAVIS_PHP_VERSION" = '7.1' ] ; then
```
lcobucci: This will run the code for ALL PHP 7.1 builds, it would be better to have it only for specific platform as well.

.travis.coverage.sh
```diff
@@ -0,0 +1,5 @@
...
+    wget https://scrutinizer-ci.com/ocular.phar
+    php ocular.phar code-coverage:upload --format=php-clover ./clover.xml
```
lcobucci: `clover.xml` will not be magically generated, we need to pass the flags to PHPUnit

Issue #1392, opened 2012-02-05
DBAL-218: Add Object for BulkInsert Abstraction


doctrinebot:
Jira issue originally created by user @beberlei:

darthf1:
What needs to be done to implement this functionality? A pull request has been [made](https://github.com/doctrine/dbal/pull/682) but the issue [related](https://github.com/doctrine/dbal/issues/2242) to the pull request has been closed as being duplicate of this issue.

Ocramius:
Linking https://github.com/doctrine/dbal/pull/682

Closing here - the pull request tracks this

PR #682, opened 2014-09-30
[WIP] [DBAL-218] Add bulk insert query

deeky666:
This is an approach to make use of database vendors' bulk row insert query syntax.
As the nature of bulk inserts usually is to insert a LOT of rows into a table (primarily from an external source), the implementation tries to focus on good performance and low memory consumption. It is NOT intended for `INSERT INTO ... SELECT ...` statement queries.

TODO:
- Add an "executor" class that encapsulates the `BulkInsertQuery` object, adds options like bulk size and internally automatically evaluates the underlying platform's limits for a single `INSERT` statement and splits queries accordingly.
- Evaluate platforms' max insert rows per `INSERT` statement and implement in platforms.
- Add unit tests for quoted identifiers.
- Add functional tests.

Future additions:
- Add support for expressions.
- Query builder (if there will be more bulk insert queries like `INSERT INTO ... SELECT ...`)?

Open for discussion. Ideas welcome!

PR #2762, opened 2017-06-29
Bulk inserts

edefimov:
This is an updated version of PR #682 solving issue #1392  refactored for php 7.1 and with small improvements like bulk insert execution even when the number of rows exceeds database limits.

### DIFF
--- a/lib/Doctrine/DBAL/Query/BulkInsertQuery.php
+++ b/lib/Doctrine/DBAL/Query/BulkInsertQuery.php
@@ -1,0 +1,271 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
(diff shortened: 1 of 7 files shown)


Ocramius:
@deeky666 this patch introduces some fixups on your existing bulk-insert patch.

Specifically, it will force transactions when an `insert` operation is acting on a range of values bigger than what the database supports.

We should consider further scenarios too:

 * when a query string is bigger than what is endorsed by the DB engine
 * when the number of bound parameters is bigger than what is allowed by the DB engine (for example, that's 1000 on OracleDB)

lib/Doctrine/DBAL/Query/BulkInsertQuery.php
```diff
@@ -0,0 +1,292 @@
...
+
+        if ($insertMaxRows > 0 && $totalValues > $insertMaxRows && !$this->connection->isTransactionActive()) {
+            throw QueryException::transactionRequired();
+        }
+
+        $offset = 0;
+        $result = 0;
+        while ($offset < $totalValues) {
+            $values = array_slice($this->values, $offset, $insertMaxRows);
+            $parameters = array_slice($this->parameters, $offset, $insertMaxRows);
```
LisTo4ek: Here should be something like:
$parameters = array_slice($this->parameters, $offset * count($this->values), $insertMaxRows * count($this->values));

morozov (changes requested):
@edefimov thank you for the patch. Please rebase it and make sure the most generic cases are tested in a functional way.

karousn:
any news about this ? available in next major version(3) ?

claudiobizzotto:
Until this gets incorporated into the next major release, here's a temporary option: https://github.com/pharako/mysql-dbal

takahashiyuya:
i hope can use it.

morozov:
Closing as stale.

PR #2765, opened 2017-07-02
#2648 fix last insert id consistency

Ocramius:
Replaces #2648

This patch is rebased, includes PHP 7.1 hints, and shows the failures that we already saw in #2648.

@deeky666 are those failures expected?


lib/Doctrine/DBAL/Driver/LastInsertId.php
```diff
@@ -0,0 +1,47 @@
...
+     * @var string
+     */
+    private $value = '0';
+
+    public function get() : string
+    {
+        return $this->value;
+    }
+
+    public function set(string $value) : void
```
morozov: Based on the logic inside, can we name it `register()` instead? The `set()` name may be misleading in the way that setting a `'0'` value doesn't actually set it.

morozov: I see the current contract in 2.x is `Connection::lastInsertId() : string`, but besides BC, is there a reason the ID is a string? Does the DBAL support any sequences which produce non-numeric value? If not, then it would make sense to me to store the value as `int` and return a `string` from the function call. Otherwise, most adapters have to cast their values to `(string)`.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Connection.php
```diff
@@ -145,21 +144,22 @@ public function exec($statement)
...
-
-        if ($result === false || !isset($result['CURRVAL'])) {
-            throw new OCI8Exception("lastInsertId failed: Query was executed but no result was returned.");
+        try {
+            $stmt = $this->prepare('SELECT ' . $name . '.CURRVAL FROM DUAL');
+            $stmt->execute();
+        } catch (OCI8Exception $exception) {
+            // In case no sequence value is available yet, we get the error:
+            // ORA-08002: sequence $name.CURRVAL is not yet defined in this session
+            return '0';
```
morozov: Would it make sense to also check the error code? Otherwise, the DBAL may be suppressing irrelevant exceptions.

morozov (commented):
Should we also deprecate the `SQLSrv\LastInsertId` in favor of the new one?

morozov:
@Ocramius @deeky666 Let me know if you need help rebasing this. Looks like the rebase will be less painful if squashed first. Please also see my questions above.

lib/Doctrine/DBAL/Driver/LastInsertId.php
```diff
@@ -0,0 +1,47 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
```
morozov: The license header is no longer needed.

lib/Doctrine/DBAL/Driver/PDOConnection.php
```diff
@@ -90,18 +101,35 @@ public function query()
...
+                $this->trackLastInsertId();
+
+                return $stmt;
             }
 
             if ($argsCount == 2) {
-                return parent::query($args[0], $args[1]);
+                $stmt = parent::query($args[0], $args[1]);
+
+                $this->trackLastInsertId();
```
morozov: Do these lines need to be duplicated? If the case with `$args[0]` is wrapped in the `else` block, you can use a single return statement and call to `$this->trackLastInsertId();` prior to that.

lib/Doctrine/DBAL/Driver/LastInsertId.php
Ocramius: I don't remember the specific details here, but we settled for a string identifier because auto-generation may really produce anything. I don't think we built a test case for it though.

Ocramius: :+1: for renaming this to `register()`

lib/Doctrine/DBAL/Driver/OCI8/OCI8Connection.php
Ocramius: Not sure if we can do that in a reliable way without starting to hit false-positives. The API is a bit too squishy to rely on precise codes.

lib/Doctrine/DBAL/Driver/LastInsertId.php
Ocramius: I'll run the entire thing through diffcs anyway

lib/Doctrine/DBAL/Driver/PDOConnection.php
Ocramius: I think our CS doesn't even allow that anymore, but I'll check

morozov:
A pull request from the `doctrine` repository triggers two builds per build service: one for the pull request and the other for the branch itself. @Ocramius can you disable building arbitrary branches on Travis and Appveyor?

morozov:
The PDO failures and deadlocks are fixed. However, the newly introduced test is now failing in _some_ configurations (Postges, MariaDB) with:
```
1) Doctrine\Tests\DBAL\Functional\LastInsertIdTest::testLastInsertIdAfterSelectQuery
Exception: [Doctrine\DBAL\Exception\ConnectionException] An exception occurred in driver: SQLSTATE[08006] [7] FATAL:  sorry, too many clients already
FATAL:  sorry, too many clients already
```
Looks like a connection resource leak in the test.

morozov:
Closing in favor of #3074. Travis stopped building this PR as PR and only builds it as a branch.

Issue #2676, opened 2017-03-14
Optimize Oracle SchemaManager


mathieubouchard:
On my system, building the list of tables on a schema with 61 tables result on a timeout after 480 seconds on an Oracle 12c database. The problem is that each call to get the list of foreign keys or indexes result in requests that take several seconds each.

The problem lies with requests on tables user_constraints and/or all_constraints. As an example, getting the list of foreign keys on a database that has no foreign key takes 5 seconds for each call (305 seconds total only for that part).

mathieubouchard:
This patch optimize the creation of a Schema from an Oracle database by overriding the OracleSchemaManager->listTables method to get the list of columns, foreign keys and indexes one time for all the tables in the current database. That information is then used to create the table schemas.

After the patch, the schema of the database is detected in 5 seconds instead of more than 8 minutes.

It has been tested on Doctrine DBAL version 2.5.5-DEV, as shipped with TYPO3 8.6.0.

[Doctrine-optimize-OracleSchemaManager.txt](https://github.com/doctrine/dbal/files/840225/Doctrine-optimize-OracleSchemaManager.txt)

Ocramius:
Could you please send it as a PR, so it can be reviewed? Also, what is the
bottleneck, exactly?

mondrake:
I am also seeing this, developing a db driver for Drupal (https://github.com/mondrake/drudbal). More than 3 minutes to introspect a schema of about 50 tables, without any FK. Will try to open a PR with @mathieubouchard patch.

morozov:
Closing as stale. Please send the above as a PR if still relevant.

greg0ire:
A PR has been sent, reopening.

PR #2766, opened 2017-07-03
Optimize OracleSchemaManager::listTables

mondrake:
A PR replicating the patch included in https://github.com/doctrine/dbal/issues/2676 by @mathieubouchard .

Tests executed on continuousphp show that running `createSchema` on an Oracle db with about 170 tables takes over a minute. See https://github.com/doctrine/dbal/pull/2767.

```
Time: 4.18 minutes, Memory: 80.00MB 
 
There was 1 failure: 
 
1) Doctrine\Tests\DBAL\Functional\Schema\OracleSchemaManagerTest::testCreateSchemaOnLargeNumberOfTables 
createSchema() executed in less than 15 sec. 
Failed asserting that 64.17334699630737 is less than 15. 
 
tests/Doctrine/Tests/DBAL/Functional/Schema/OracleSchemaManagerTest.php:292 
 
FAILURES! 
Tests: 4613, Assertions: 8763, Failures: 1, Skipped: 519, Incomplete: 11. 
```


Ocramius:
That's a lot of additional code with no additional tests covering all the newly introduced decision paths: if this patch is not a functional change, I'm inclined to rejecting it, sorry.

mondrake:
@Ocramius can you please leave it open for a while and see if I or anyone else can add appropriate tests, The underlying issue is relevant, you can't really use productively schema manager's `createSchema()` over an Oracle database that has some tables in it.

Ocramius:
That's fine by me: I realise that you put a lot of effort in it, but any
code path that isn't tested is a possible bug if not a security issue, so
it won't be merged until carefully tested.

Since this code runs oracledb specific queries, the tests should be
integration ones.

mondrake:
Added a functional test to create 150 test tables, then run `createSchema` on the database. The test fails, apparently there are some issues with the original patch. I will create a separate PR just with the test, which is relevant anyway in order to measure the time `createSchema` needs to run.

mathieubouchard:
Looking at the code of the Oracle test file, it seems there are some tables created for other tests that are not deleted before your new method create it's 150 new tables. So the total number of tables in the database will be greater than 150.

mondrake:
@mathieubouchard yes, I changed the test in https://github.com/doctrine/dbal/issues/2767 to test for more than 150 tables, but also most importantly to measure the execution time. 

Any idea why we have this failure in your original code

```
Caused by 
PHPUnit_Framework_Error_Notice: Undefined index: table_name 
 
lib/Doctrine/DBAL/Schema/OracleSchemaManager.php:438 
lib/Doctrine/DBAL/Schema/OracleSchemaManager.php:379 
tests/Doctrine/Tests/DBAL/Functional/Schema/OracleSchemaManagerTest.php:36 
```

which affects also other tests?

mathieubouchard:
The problem is with the use of ***array_key_exists($foreignKeyRow['table_name'], $tablesForeignKeys)*** in the method ***listTablesForeignKeys***. the ***table_name*** column is not in ***foreignKeyRow*** the first time a column is parsed for a given table.

```
    private function listTablesForeignKeys($database = null)
    {
        $sql = $this->_platform->getListTablesForeignKeysSQL($database);
        $tablesForeignKeysRows = $this->_conn->fetchAll($sql);

        $tablesForeignKeys = array();
        foreach ($tablesForeignKeysRows as $foreignKeyRow) {
            $foreignKeyRow = \array_change_key_case($foreignKeyRow, CASE_LOWER);
            
            if (!array_key_exists($foreignKeyRow['table_name'], $tablesForeignKeys)) {
                $tablesForeignKeys[$foreignKeyRow['table_name']] = array($foreignKeyRow);
            } else {
                $tablesForeignKeys[$foreignKeyRow['table_name']][] = $foreignKeyRow;
            }
        }
```

There should be an additional check for ***array_key_exists('table_name', $foreignKeyRow)*** before the assignment.

Per the PHP documentation, this comportement is :

_Attempting to access an array key which has not been defined is the same as accessing any other undefined variable: an ***E_NOTICE***-level error message will be issued, and the result will be ***NULL***._

I suppose that the PHPUnit test considers an ***E_NOTICE*** as an error and throw the ***PHPUnit_Framework_Error_Notice*** exception in that case.

mondrake:
Thanks @mathieubouchard , is your comment already based on the latest commit https://github.com/doctrine/dbal/pull/2766/commits/2ffcec0791ee5c719ee309849ea127fd64368df4?

mathieubouchard:
Based on the error detailed in the following build: https://app.continuousphp.com/git-hub/doctrine/dbal/build/41d54cad-99e4-4e4c-be18-4c06fb7fa26c

mondrake:
OK, so I think that this comment from #2676 
> ... optimize the creation of a Schema from an Oracle database by overriding the OracleSchemaManager->listTables method to get the list of columns, foreign keys and indexes one time for all the tables in the current database. That information is then used to create the table schemas.

is the way to go - that means 4 SQL queries regardless of the number of tables in the database, instead of [1 (get the tables) + #tables * 3 (to get for each table its columns, indexes and foreign keys separately)]. For 150 tables that means 4 queries instead of 451.

However, the changes so far need further work. My suggestion for now is to focus on the test - make it more robust by adding indexes and foreign keys to the test case, and make it pass on currrent code, with the current performance. With that done, then turn to the approach above.

mondrake:
The test now fails only on the last assert, that measures the performance of `createSchema`. We have tables, indexes, foreign keys created in a large number (150), including a table with quoted identifiers. The test checks the consistency of the objects returned by  `createSchema`, so that now we can go back adding the code of the original patch and override the ::listTables method with one specific for Oracle, and see that it keeps returning same results.

mondrake:
Maybe we do not need to add separate methods to OraclePlatform to return SQL for the all tables' columns / indexes /  foreign keys.

If we accept a convention that passing `null` as `$table` parameter to `getListTableColumnsSQL`, `getListTableIndexesSQL`, `getListTableForeignKeysSQL`, we return the SQL for enquiring the entire database, we can fit in the current methods.

Doing that for `getListTableColumnsSQL` in next commit. Just checking that we do not break anything, no implementation yet.

mathieubouchard:
The first approach I had was to have only one method to support both use cases. This will reduce the number of lines of code changed and eliminate duplicate SQL code.

But this is a breaking change on the AbstractPlatform API. That's why I created other public methods, specific to Oracle.

mondrake:
Thanks @mathieubouchard 

>  this is a breaking change on the AbstractPlatform API

Why?

mathieubouchard:
The method signatures doesn't accept null for the table parameter, for example :
http://www.doctrine-project.org/api/dbal/2.0/class-Doctrine.DBAL.Platforms.AbstractPlatform.html#_getListTableColumnsSQL

mondrake:
Ah OK, but there's doesn't seem to be any code checking or preventing passing `null` there, neither in the abstract class nor in its implementations. Would it be a major problem to allow that?

Maybe this is a question for @Ocramius .

mondrake:
Albeit small, there's already a reduction in timing of execution of `createSchema`, from 64 sec to 58 sec.

mondrake:
With columns and indexes changed for the one single SQL query vs per table queries, we are now down to 41 sec.

```
Time: 3.11 minutes, Memory: 80.00MB 
 
There was 1 failure: 
 
1) Doctrine\Tests\DBAL\Functional\Schema\OracleSchemaManagerTest::testCreateSchemaOnLargeNumberOfTables 
createSchema() executed in less than 15 sec. 
Failed asserting that 41.360909938812256 is less than 15. 
```

Now, will go back to columns to reflect some learnings while doing the fetching of indexes, then will tackle the final one about foreign keys.

mondrake:
Test on continuousphp now passes, meaning that with about 170 tables, `createSchema` runs in less than 15 sec vs initial 64. I'll stretch the test now to much more tables and less time, see what happens.

mondrake:
So now with 650 tables, `createSchema` runs in approx. 5 sec on continuousphp

```
Time: 1.51 minutes, Memory: 88.00MB 
 
There was 1 failure: 
 
1) Doctrine\Tests\DBAL\Functional\Schema\OracleSchemaManagerTest::testCreateSchemaOnLargeNumberOfTables 
createSchema() executed in less than 3 sec. 
Failed asserting that 4.7398459911346436 is less than 3. 
 
tests/Doctrine/Tests/DBAL/Functional/Schema/OracleSchemaManagerTest.php:339 
 
```

That's an achievement! I'll return back the timing check to 15 secs to be on the safe side, then this is ready for code review.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -418,9 +418,32 @@ protected function _getCreateTableSQL($table, array $columns, array $options = a
...
     public function getListTableIndexesSQL($table, $currentDatabase = null)
     {
-        $table = $this->normalizeIdentifier($table);
-        $table = $this->quoteStringLiteral($table->getName());
+        if (null !== $table) {
+            $table = $this->normalizeIdentifier($table);
+            $table = $this->quoteStringLiteral($table->getName());
+            $tableWhereClause = "AND ind_col.table_name = " . $table;
+        } else {
+            $tableWhereClause = '';
```
Ocramius: Use a default value, instead of an `else`

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -418,9 +418,32 @@ protected function _getCreateTableSQL($table, array $columns, array $options = a
      */
     public function getListTableIndexesSQL($table, $currentDatabase = null)
     {
-        $table = $this->normalizeIdentifier($table);
-        $table = $this->quoteStringLiteral($table->getName());
+        if (null !== $table) {
+            $table = $this->normalizeIdentifier($table);
```
Ocramius: Don't overwite the `$table` variable - create new variables every time the semantic meaning changes

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -418,9 +418,32 @@ protected function _getCreateTableSQL($table, array $columns, array $options = a
...
+            $table = $this->quoteStringLiteral($table->getName());
+            $tableWhereClause = "AND ind_col.table_name = " . $table;
+        } else {
+            $tableWhereClause = '';
+        }
 
+        if (null !== $currentDatabase && '/' !== $currentDatabase) {
```
Ocramius: Move to a private method if this means "a database is selected"

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -418,9 +418,32 @@ protected function _getCreateTableSQL($table, array $columns, array $options = a
...
+            $currentDatabase = $this->normalizeIdentifier($currentDatabase);
+            $currentDatabase = $this->quoteStringLiteral($currentDatabase->getName());
```
Ocramius: Same as above - don't overwrite variables

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -418,9 +418,32 @@ protected function _getCreateTableSQL($table, array $columns, array $options = a
...
+                           ind_col.column_position AS column_pos,
+                           con.constraint_type AS is_primary
+                     FROM  all_ind_columns ind_col
+                LEFT JOIN  all_indexes ind
+                       ON  ind.owner = ind_col.index_owner AND ind.index_name = ind_col.index_name
+                LEFT JOIN  all_constraints con
+                       ON  con.owner = ind_col.index_owner AND con.index_name = ind_col.index_name
+                    WHERE  ind_col.index_owner = $currentDatabase $tableWhereClause
+                 ORDER BY  ind_col.table_name, ind_col.index_name, ind_col.column_position";
+        } else {
```
Ocramius: Since you are using an early return above, an `else` is not needed here

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -608,11 +632,34 @@ private function getAutoincrementIdentifierName(Identifier $table)
     /**
      * {@inheritDoc}
      */
-    public function getListTableForeignKeysSQL($table)
+    public function getListTableForeignKeysSQL($table, $database = null)
```
Ocramius: This escapes the interface

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -608,11 +632,34 @@ private function getAutoincrementIdentifierName(Identifier $table)
...
     {
-        $table = $this->normalizeIdentifier($table);
-        $table = $this->quoteStringLiteral($table->getName());
+        if (null !== $table) {
+            $table = $this->normalizeIdentifier($table);
+            $table = $this->quoteStringLiteral($table->getName());
+            $tableWhereClause = "AND cols.table_name = " . $table;
+        } else {
```
Ocramius: `else` not needed: use a default value

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -608,11 +632,34 @@ private function getAutoincrementIdentifierName(Identifier $table)
...
      * {@inheritDoc}
      */
-    public function getListTableForeignKeysSQL($table)
+    public function getListTableForeignKeysSQL($table, $database = null)
     {
-        $table = $this->normalizeIdentifier($table);
-        $table = $this->quoteStringLiteral($table->getName());
+        if (null !== $table) {
+            $table = $this->normalizeIdentifier($table);
+            $table = $this->quoteStringLiteral($table->getName());
```
Ocramius: Don't overwrite variables

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -608,11 +632,34 @@ private function getAutoincrementIdentifierName(Identifier $table)
...
+            $tableWhereClause = "AND cols.table_name = " . $table;
+        } else {
+            $tableWhereClause = '';
+        }
 
+        if (null !== $database && '/' !== $database) {
```
Ocramius: Same as above - move to private method

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -608,11 +632,34 @@ private function getAutoincrementIdentifierName(Identifier $table)
...
+                   r_cols.table_name \"references_table\",
+                   r_cols.column_name \"foreign_column\"
+              FROM all_cons_columns cols
+         LEFT JOIN all_constraints alc
+                ON alc.owner = cols.owner AND alc.constraint_name = cols.constraint_name
+         LEFT JOIN all_cons_columns r_cols
+                ON r_cols.owner = alc.r_owner AND r_cols.constraint_name = alc.r_constraint_name AND r_cols.position = cols.position
+             WHERE cols.owner = $database $tableWhereClause AND alc.constraint_type = 'R'
+             ORDER BY cols.table_name, cols.constraint_name, cols.position";
+        } else {
```
Ocramius: `else` not needed

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -653,33 +701,35 @@ public function getListTableConstraintsSQL($table)
...
-        $tabColumnsTableName = "user_tab_columns";
-        $colCommentsTableName = "user_col_comments";
-        $tabColumnsOwnerCondition = '';
-        $colCommentsOwnerCondition = '';
+        if (null !== $table) {
+            $table = $this->normalizeIdentifier($table);
+            $table = $this->quoteStringLiteral($table->getName());
+            $tableWhereClause = "AND c.table_name = " . $table;
+        } else {
+            $tableWhereClause = '';
```
Ocramius: same remarks as above

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -653,33 +701,35 @@ public function getListTableConstraintsSQL($table)
...
-        $table = $this->normalizeIdentifier($table);
-        $table = $this->quoteStringLiteral($table->getName());
-
-        $tabColumnsTableName = "user_tab_columns";
-        $colCommentsTableName = "user_col_comments";
-        $tabColumnsOwnerCondition = '';
-        $colCommentsOwnerCondition = '';
+        if (null !== $table) {
+            $table = $this->normalizeIdentifier($table);
+            $table = $this->quoteStringLiteral($table->getName());
```
Ocramius: same remarks as above

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -653,33 +701,35 @@ public function getListTableConstraintsSQL($table)
...
-        }
-
-        return "SELECT   c.*,
+            return "SELECT   c.*, d.comments AS comments
+                    FROM    all_tab_columns c
+               LEFT JOIN    all_col_comments d
+                      ON    d.OWNER = c.OWNER AND d.TABLE_NAME = c.TABLE_NAME AND d.COLUMN_NAME = c.COLUMN_NAME
+                   WHERE    c.owner = $database $tableWhereClause
+                   ORDER BY c.table_name, c.column_id";
+        } else {
```
Ocramius: same remarks as above

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
```diff
@@ -36,6 +36,52 @@ class OracleSchemaManager extends AbstractSchemaManager
...
+        $allColumns = $this->_conn->fetchAll($this->_platform->getListTableColumnsSQL(null, $currentDatabase));
+
+        // Get all foreign keys definitions in one database call.
+        $allForeignKeys = $this->_conn->fetchAll($this->_platform->getListTableForeignKeysSQL(null, $currentDatabase));
+
+        // Get all indexes definitions in one database call.
+        $allIndexes = $this->_conn->fetchAll($this->_platform->getListTableIndexesSQL(null, $currentDatabase));
+
+        $tables = array();
+        foreach ($tableNames as $tableName) {
```
Ocramius: This entire block seems to be an elaborate `array_filter()`?

tests/Doctrine/Tests/DBAL/Functional/Schema/OracleSchemaManagerTest.php
```diff
@@ -273,4 +273,81 @@ public function testListTableDateTypeColumns()
         $this->assertSame('datetime', $columns['col_datetime']->getType()->getName());
         $this->assertSame('datetimetz', $columns['col_datetimetz']->getType()->getName());
     }
+
+    public function testCreateSchemaOnLargeNumberOfTables()
```
Ocramius: Can this test be made generic for multiple schema managers?

tests/Doctrine/Tests/DBAL/Functional/Schema/OracleSchemaManagerTest.php
```diff
@@ -273,4 +273,81 @@ public function testListTableDateTypeColumns()
...
+        $this->assertSame([$testForeignKey], $this->resolveAssetsNames($schema->getTable($testTable)->getForeignKeys()));
+        $this->assertSame($testTable, $schema->getTable($testTable)->getForeignKey($testForeignKey)->getLocalTable()->getQuotedName($this->_conn->getDatabasePlatform()));
+        $this->assertSame(['"Q_parent_id"'], $schema->getTable($testTable)->getForeignKey($testForeignKey)->getLocalColumns());
+        $this->assertSame('TBL_TEST_2766_0', $schema->getTable($testTable)->getForeignKey($testForeignKey)->getForeignTableName());
+        $this->assertSame(['X_ID'], $schema->getTable($testTable)->getForeignKey($testForeignKey)->getForeignColumns());
+
+        $this->assertLessThan(15, $endTime - $startTime, 'createSchema() executed in less than 15 sec.');
+    }
+
+    protected function resolveAssetsNames(array $assets)
```
Ocramius: To be documented. Also, make it `private` and add a return type

tests/Doctrine/Tests/DBAL/Functional/Schema/OracleSchemaManagerTest.php
Ocramius: ` : void`

Ocramius (changes requested):

mondrake:
Thanks for feedback @Ocramius . Will work on that. 

> This escapes the interface

So here we need a separate method?

Ocramius:
> So here we need a separate method?

As far as I can see, `getListTableForeignKeysSQL` is never called with a `$database`, so why was the parameter used?

mondrake:
> As far as I can see, getListTableForeignKeysSQL is never called with a $database, so why was the parameter used?

Because in the new concept we use the `ALL_*` views that provide information about Oracle DBMS objects, instead of the slower `USER_*` views that do a lot of more checking about user being authorised to access an object (checked with EXPLAIN PLAN). 

The other two methods (for columns and for indexes) have a $database parameter... in current state we have an asymmetry. But OK, in this case to avoid changing the signature I will add a separate method.

tests/Doctrine/Tests/DBAL/Functional/Schema/OracleSchemaManagerTest.php
mondrake: Uhmm this is tricky - both the 'setup' part and the 'assertion' part are kind of db specific

1) the 'setup' is with db specific queries, to simulate tables and assets generated out of DBAL
2) the 'assertion' part is also specific because it ened to be tailored around what is actually introspected - for instance, lowercase column names in Oracle are converted to uppercase, and these are returned as column names by `$schema->getColumn(...)->getName`. MySql will be different. This may be a bug if we want to ensure portability, but not in scope of addressing here IMHO.

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
mondrake: Actually no, we are iterating through table 'names' and building an array of Table 'objects', AFAIK `arrayFilter` would just return the names again.

mondrake:
@Ocramius I made the changes requested.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -418,9 +418,41 @@ protected function _getCreateTableSQL($table, array $columns, array $options = a
      */
     public function getListTableIndexesSQL($table, $currentDatabase = null)
     {
-        $table = $this->normalizeIdentifier($table);
-        $table = $this->quoteStringLiteral($table->getName());
+        if (null === $table && !$this->isDatabaseSelected($currentDatabase)) {
+            throw new \InvalidArgumentException('Table name must be specified if no database is specified');
```
Ocramius: This block is not currently covered by tests

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -1176,4 +1263,16 @@ public function quoteStringLiteral($str)
...
     }
+
+    /**
+     * Determines if the input variable identifies a specific database.
+     *
+     * @param string $database
+     *
+     * @return bool
+     */
+    private function isDatabaseSelected($database)
```
Ocramius: Since this is an addition that will land in `master`, you can use PHP 7.1 hints

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -610,9 +642,54 @@ private function getAutoincrementIdentifierName(Identifier $table)
...
+     *
+     * @param string $table
+     * @param string $database
+     *
+     * @return string
+     */
+    public function getListForeignKeysSQL($table = null, $database = null)
+    {
+        if (null === $table && !$this->isDatabaseSelected($database)) {
+            throw new \InvalidArgumentException('Table name must be specified if no database is specified');
```
Ocramius: This block is not currently covered by tests

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -610,9 +642,54 @@ private function getAutoincrementIdentifierName(Identifier $table)
...
+
+        // If database is specified, return SQL for all foreign keys in
+        // database, optionally filtered by table.
+        if ($this->isDatabaseSelected($database)) {
+            $databaseIdentifier = $this->normalizeIdentifier($database);
+            $quotedDatabaseIdentifier = $this->quoteStringLiteral($databaseIdentifier->getName());
+                return "SELECT cols.table_name,
+                   alc.constraint_name,
+                   alc.DELETE_RULE,
+                   cols.column_name \"local_column\",
```
Ocramius: You can probably use `HEREDOC` to avoid escaping quotes: 

```php
return <<<SQL
SQL;
```

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
Ocramius: Alright :+1:

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
```diff
@@ -36,6 +36,52 @@ class OracleSchemaManager extends AbstractSchemaManager
...
+            $tableColumns = array_filter($allColumns, function ($column) use ($unquotedTableName) {
+                return $column['TABLE_NAME'] === $unquotedTableName;
+            });
+            $columns = $this->_getPortableTableColumnList($tableName, null, $tableColumns);
+
+            // Process foreign keys for this table.
+            $tableForeignKeys = array_filter($allForeignKeys, function ($foreignKey) use ($unquotedTableName) {
+                return $foreignKey['TABLE_NAME'] === $unquotedTableName;
+            });
+            $foreignKeys = !empty($tableForeignKeys) ? $this->_getPortableTableForeignKeysList($tableForeignKeys) : array();
```
Ocramius: You can use the short array syntax overall

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1165,4 +1173,43 @@ public function testDoesNotListIndexesImplicitlyCreatedByForeignKeys()
...
+    /**
+     * Create a large number of tables with specific database syntax.
+     *
+     * @return int The number of tables created.
+     *
+     * @see ::testCreateSchemaOnLargeNumberOfTables
+     */
+    protected function createLargeNumberOfTables(): int
+    {
+        $this->markTestIncomplete($this->_conn->getDriver()->getName().' does not implement ' . __METHOD__);
```
Ocramius: `__METHOD__` points at the wrong symbol here, I suppose?

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1165,4 +1173,43 @@ public function testDoesNotListIndexesImplicitlyCreatedByForeignKeys()
...
+    /**
+     * Checks schema against a large number of tables created.
+     *
+     * @param \Doctrine\DBAL\Schema\Schema The schema created from the database.
+     *
+     * @see ::testCreateSchemaOnLargeNumberOfTables
+     */
+    protected function checkLargeNumberOfTables(Schema $schema): void
+    {
+        $this->markTestIncomplete($this->_conn->getDriver()->getName().' does not implement ' . __METHOD__);
```
Ocramius: `__METHOD__` points at the wrong symbol here, I suppose?

Ocramius (changes requested):
Other than nitpicks about unhappy path testing (required) and other details, this looks good to me.

I'd still want to get @deeky666's review, which might take some time, since he's busy.

@morozov can you also check this one?

mondrake:
Added tests for OracleSchemaManager::getListTableIndexesSQL, ::getListForeignKeysSQL, ::getListTableColumnsSQL

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -610,31 +646,80 @@ private function getAutoincrementIdentifierName(Identifier $table)
...
+
+    /**
+     * Returns the list of foreign keys for the current database.
+     *
+     * @param string $table
+     * @param string $database
+     *
+     * @return string
+     */
+    public function getListForeignKeysSQL($table = null, $database = null)
```
morozov: There should be two (sets of) methods: `getTableSomething($table)` and `getAllSomething()`. Otherwise, the single function has two paths which don't share much, and a mistakenly defined value of `$table` may change the resulting function logic.

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
```diff
@@ -36,6 +36,52 @@ class OracleSchemaManager extends AbstractSchemaManager
...
+
+        // Get all foreign keys definitions in one database call.
+        $allForeignKeys = $this->_conn->fetchAll($this->_platform->getListForeignKeysSQL(null, $currentDatabase));
+
+        // Get all indexes definitions in one database call.
+        $allIndexes = $this->_conn->fetchAll($this->_platform->getListTableIndexesSQL(null, $currentDatabase));
+
+        $tables = [];
+        foreach ($tableNames as $tableName) {
+            $unquotedTableName = rtrim(ltrim($tableName, '"'), '"');
```
morozov: Can we `fetchAll()` from `getListTablesSQL()` instead of dealing with quoted table names?

tests/Doctrine/Tests/DBAL/Functional/Schema/OracleSchemaManagerTest.php
```diff
@@ -273,4 +275,90 @@ public function testListTableDateTypeColumns()
...
+     * {@inheritdoc}
+     */
+    protected function createLargeNumberOfTables(): int
+    {
+        // Create a base table for test.
+        $sql = "CREATE TABLE tbl_test_2766_0 (x_id VARCHAR2(255) DEFAULT 'x' NOT NULL, x_data CLOB DEFAULT NULL NULL, x_number NUMBER(10) DEFAULT 0 NOT NULL, PRIMARY KEY(x_id))";
+        $this->_conn->executeUpdate($sql);
+
+        // Create a large number of tables with indexes and foreign keys.
+        for ($i = 1; $i < 649; $i++) {
```
morozov: I don't see much value in having tests like this. Even if it performs slow for any reason, we won't see anything on CI as long as the test passes. It would make more sense to create a test which verifies that the number of executed SQL queries doesn't depend on the number of tables in the DB (if that's the case).

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1165,4 +1173,43 @@ public function testDoesNotListIndexesImplicitlyCreatedByForeignKeys()
...
+    {
+        $numberOfTablesCreated = $this->createLargeNumberOfTables();
+
+        // Introspect the db schema.
+        $startTime = microtime(TRUE);
+        $schema = $this->_sm->createSchema();
+        $endTime = microtime(TRUE);
+
+        $this->assertGreaterThanOrEqual($numberOfTablesCreated, count($schema->getTables()));
+        $this->assertLessThan($this->timeThresholdForCreateSchema, $endTime - $startTime, 'createSchema() executed in less than ' . $this->timeThresholdForCreateSchema . ' sec.');
```
morozov: This test is by design brittle since the time spent in userland on the execution depends on a lot of factors we're not in control of. Be ready to have this test randomly failing.

morozov (changes requested):
The change itself makes a lot of sense to me but the code requires some polishing.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -418,10 +418,45 @@ protected function _getCreateTableSQL($table, array $columns, array $options = a
      */
     public function getListTableIndexesSQL($table, $currentDatabase = null)
     {
-        $table = $this->normalizeIdentifier($table);
-        $table = $this->quoteStringLiteral($table->getName());
+        if (null === $table && !$this->isDatabaseSelected($currentDatabase)) {
```
mondrake: `Connection` is not known within `AbstractPlatform` and children classes, so we cannot use `getParams` here to get the schema name. Getting from USERENV would mean a separate SQL call to the DB which IMHO we should avoid if possible. Accessing `USER*`  views is there to address that.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
mondrake: Added tests.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -418,10 +418,45 @@ protected function _getCreateTableSQL($table, array $columns, array $options = a
...
+                       ON  ind.owner = ind_col.index_owner AND ind.index_name = ind_col.index_name
+                LEFT JOIN  all_constraints con
+                       ON  con.owner = ind_col.index_owner AND con.index_name = ind_col.index_name
+                    WHERE  ind_col.index_owner = $quotedDatabaseIdentifier $tableWhereClause
+                 ORDER BY  ind_col.table_name, ind_col.index_name, ind_col.column_position
+SQL;
+        }
+
+        // If database is not specified, return SQL for the indexes of the
+        // specified table from the current database.
```
mondrake: See comments above about using ALL* vs USER* views.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
mondrake: This method needs to be called by OracleSchemaManager, hence the need to have it public. Are you suggesting to increase the public API to add  `getAllSomething()` methods? I'd be fine with that, we would have to add those to the `AbstractPlatform` class though... need @Ocramius input here I suppose.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
mondrake: Added tests. See comments above about it needs be public.

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
mondrake: We need both unquoted table names (to fetch information from arrays returned by the `getListWhatever()` methods) and quoted table names that are used to create DBAL's objects names. So yes we could avoid the trim here by fetchAll from db, but we would need to re-quote the identifiers later. Since `listTableNames()` does all the checks/validations upfront, I'd prefer to keep this as is.

tests/Doctrine/Tests/DBAL/Functional/Schema/OracleSchemaManagerTest.php
mondrake: I see your point, but at the end of the day this entire issue is about performance, not about a bug.  Current code can build the required DBAL schema, if you only accept waiting looong time for it. How about increasing the accepted threshold time to, say, one  minute? Or any value that we can still consider being a failure in any possible environment? (Can anybody reasonably accept the schema is built in more than one minute, provided PHP does not throught timeouts in the meantime)

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
Ocramius: > Be ready to have this test randomly failing.

That's... a bit of an issue :-\

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
morozov: This way, this code depends on the implementation details of the `OraclePlatform::listTableNames()` — only `OraclePlatform` returns table names quoted (not sure why) while the others don't.

> Since listTableNames() does all the checks/validations upfront, I'd prefer to keep this as is.

That's a valid point which brings up the question: if `AbstractSchemaManager::_getPortableTableDefinition()` returns quoted or non-quoted identifiers based on the platform, how is the behavior of `filterSchemaAssetsExpression` (I assume this is what was meant by checks/validations above) is expected to be portable? Or it's not? /cc @Ocramius

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
morozov: Agree.

morozov (changes requested):

mondrake:
Trying to summarize here the open points of this PR. I'd appreciate feedback before I jump into coding. Thanks.

1) The overall point of this PR is to try and get all columns, indexes, and foreign keys defined for all tables in a schema with as less calls to server as possible.

In the current HEAD state, `listTables` iterates over each single table in the schema and makes 3 SQL calls per table, via `getListTableColumnsSQL`, `getListTableForeignKeysSQL`, and `getListTableIndexesSQL`.

In the PR so far, I have been exploiting a backdoor of `getListTableColumnsSQL` and `getListTableIndexesSQL` that allows passing a `$database` leaving the `$table` parameter to NULL. Since 
`getListTableForeignKeysSQL` does not allow passing a `$database` parameter, I wrote a separate `getListForeignKeysSQL` method with both `$database` and `$table` parameters, and redirected the original `getListTableForeignKeysSQL` to call it without `$database` parameter.

Probably this is just generating confusion. 

My suggestion at this point is to revert all changes to `getListTableColumnsSQL`, `getListTableForeignKeysSQL`, and `getListTableIndexesSQL`, that can stay as-is to address the request 'give me details for a specific table (in a specified database optionally)'.

Instead, let's have 3 new methods:
```
getListAllColumnsSQL
getListAllForeignKeysSQL
getListAllIndexesSQL

```
each one with just the `$database` parameter (mandatory, not optional), returning all columns, fks and indexes addressing the request 'give me details for all the tables in the database at once'.

The question is: do we want these methods to be specific for `OraclePlatform`, or generally in `AbstractPlatform` (raising an exception in the abstract implementation), thus introducing an API addition?

I would suggest to make them abstract, so that later other platforms can use the same pattern if necessary (personally I would like to see if MySql could benefit from a similar approach in terms of performance).

2) The tests: I agree that performance testing should not be done via unit testing. However -- the entire point of this PR is about trying to improve performance. I do not see much benefit to just count the SQL calls, either -- what if they are few but with very low performance.
So, my suggestions: a) we just drop that test, or b) we set the test incomplete if it does not perform in the expected threshold time, instead of failing it with the assert. In any case, it's likely that testing environments would timeout anyway if tests take too long (Travis has a 50 mins overall max time, and 10 mins timeout if no output is received).

morozov:
> each one with just the `$database` parameter (mandatory, not optional), returning all columns, fks and indexes addressing the request 'give me details for all the tables in the database at once'.

Right now there's no consistent API around which database(s) can be inspected for objects. `getListTableForeignKeysSQL()` doesn't accept `$database` at all and uses the `USER_*` views, while `getListTableColumnsSQL()` accepts optional `$database` and uses either of `ALL_*` and `USER_*` views depending on the value.

I believe, in most cases, there's one database per application, so it makes sense to not require the `$database` and use the `USER_*` views by default. Otherwise, the caller will have to figure it out from the connection parameters.

> The question is: do we want these methods to be specific for `OraclePlatform`, or generally in `AbstractPlatform` (raising an exception in the abstract implementation), thus introducing an API addition?

It depends on how much sense it makes for other platforms and the use cases.

1. For Oracle (and maybe DB2), it does make sense to use the optimized APIs because otherwise, it will rebuild the views on every call.
2. If there are use cases in schema manager which iterate over all tables' objects, it also makes sense to retrieve them all at once instead of retrieving the list of tables first and then retrieving all their objects in a loop.

> The tests: I agree that performance testing should not be done via unit testing. However — the entire point of this PR is about trying to improve performance. I do not see much benefit to just count the SQL calls, either — what if they are few but with very low performance.

The point in counting the SQL calls is that we figured out that the execution time linearly depends on the number of calls, and we don't want the number of calls to depend on the number of tables. We want to have a test which guarantees that. We don't need a test which would fail if this code works slower than we expect because it can happen for any reason.

> So, my suggestions: a) we just drop that test, or b) we set the test incomplete if it does not perform in the expected threshold time, instead of failing it with the assert. In any case, it's likely that testing environments would timeout anyway if tests take too long (Travis has a 50 mins overall max time, and 10 mins timeout if no output is received).

Dropping the test is fine given that the code being changed is covered with other tests. The problem is, without the test it's not that clear why the change was needed in the first place. Reverting the changes and running the test often helps understand the changes. The second option is not really useful to me as it doesn't provide any continuous feedback.

mondrake:
@morozov , @Ocramius  I made some changes according to latest comments. Thanks for review.

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
```diff
@@ -36,6 +36,52 @@ class OracleSchemaManager extends AbstractSchemaManager
...
+
+        // Get all indexes definitions in one database call.
+        $allIndexes = $this->_conn->fetchAll($this->_platform->getListAllIndexesSQL($currentDatabase));
+
+        $tables = [];
+        foreach ($tableNames as $tableName) {
+            $unquotedTableName = rtrim(ltrim($tableName, '"'), '"');
+
+            // Process columns for this table.
+            $tableColumns = array_filter($allColumns, function ($column) use ($unquotedTableName) {
```
morozov: The list of columns is iterated/filtered once for every single table which is suboptimal. Instead, it could be re-grouped by table name once, and then every table could pick its own group if it exists.

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
```diff
@@ -36,6 +36,52 @@ class OracleSchemaManager extends AbstractSchemaManager
...
+                return $column['TABLE_NAME'] === $unquotedTableName;
+            });
+            $columns = $this->_getPortableTableColumnList($tableName, null, $tableColumns);
+
+            // Process foreign keys for this table.
+            $tableForeignKeys = array_filter($allForeignKeys, function ($foreignKey) use ($unquotedTableName) {
+                return $foreignKey['TABLE_NAME'] === $unquotedTableName;
+            });
+            $foreignKeys = !empty($tableForeignKeys) ? $this->_getPortableTableForeignKeysList($tableForeignKeys) : [];
```
morozov: Not really worth optimization IMO. Calling `_getPortableTableForeignKeysList()` with an empty array will produce an empty array which is fine. Especially if called once before grouping by table name.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -608,6 +633,31 @@ private function getAutoincrementIdentifierName(Identifier $table)
...
+                 cols.position,
+                 r_cols.table_name "references_table",
+                 r_cols.column_name "foreign_column"
+            FROM all_cons_columns cols
+       LEFT JOIN all_constraints alc
+              ON alc.owner = cols.owner AND alc.constraint_name = cols.constraint_name
+       LEFT JOIN all_cons_columns r_cols
+              ON r_cols.owner = alc.r_owner AND r_cols.constraint_name = alc.r_constraint_name AND r_cols.position = cols.position
+           WHERE cols.owner = $quotedDatabaseIdentifier AND alc.constraint_type = 'R'
+           ORDER BY cols.table_name, cols.constraint_name, cols.position
```
morozov: Please format all new SQL consistently (like this query, keyword on the left, one space, the rest on the righ).

morozov (commented):
@mondrake please see a few more comments on the implementation. The code looks mostly good besides a couple of issues. The tests still need to be taken care of.

mondrake:
@morozov your latest comments should be addressed now. Travis build failures are unrelated AFAICS. The open point is tests, I will look if there are ways to count the executed SQL; time checking is removed anyway.

mondrake:
OK, so this now has a test on the number of expected SQL queries that createSchema() will execute.

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
```diff
@@ -36,6 +36,54 @@ class OracleSchemaManager extends AbstractSchemaManager
...
+
+        // Get all indexes definitions in one database call.
+        $indexesByTable = $this->getNestedArrayByTableName($this->_conn->fetchAll($this->_platform->getListAllIndexesSQL($currentDatabase)));
+
+        $tables = [];
+        foreach ($tableNames as $tableName) {
+            $unquotedTableName = trim($tableName, '"');
+            $columns = $this->_getPortableTableColumnList($tableName, null, $columnsByTable[$unquotedTableName]);
+            $foreignKeys = $this->_getPortableTableForeignKeysList(isset($foreignKeysByTable[$unquotedTableName]) ? $foreignKeysByTable[$unquotedTableName] : []);
+            $indexes = $this->_getPortableTableIndexesList(isset($indexesByTable[$unquotedTableName]) ? $indexesByTable[$unquotedTableName] : [], $tableName);
```
morozov: This looks a bit bloated. What about:
```php
$foreignKeys = $indexes = [];

if (isset($foreignKeysByTable[$unquotedTableName])) {
    $foreignKeys = $this->_getPortableTableForeignKeysList($foreignKeysByTable[$unquotedTableName]);
}

if (isset($indexesByTable[$unquotedTableName])) {
    $indexes = $this->_getPortableTableForeignKeysList($indexesByTable[$unquotedTableName]);
}

```

tests/Doctrine/Tests/DBAL/Functional/Schema/OracleSchemaManagerTest.php
```diff
@@ -2,13 +2,29 @@
...
+     * For Oracle, we expect 5 queries:
+     *   1. fetch sequences
+     *   2. fetch tables
+     *   3. fetch all columns
+     *   4. fetch all foreign keys
+     *   5. fetch all indexes
+     *
+     * @var int
+     */
+    protected $expectedQueryCountToCreateSchema = 5;
```
morozov: It looks kind of okay but I still would prefer to not hard-code any concrete numbers. Instead, I'd create two schemas: for instance, one with one table, and one with two. And then made sure that the number of queries required to inspect them is the same.

tests/Doctrine/Tests/DBAL/Functional/Schema/OracleSchemaManagerTest.php
```diff
@@ -273,4 +289,90 @@ public function testListTableDateTypeColumns()
         $this->assertSame('datetime', $columns['col_datetime']->getType()->getName());
         $this->assertSame('datetimetz', $columns['col_datetimetz']->getType()->getName());
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createLargeNumberOfTables(): int
```
morozov: We don't need a large number of tables to test the fix.

tests/Doctrine/Tests/DBAL/Functional/Schema/OracleSchemaManagerTest.php
```diff
@@ -273,4 +289,90 @@ public function testListTableDateTypeColumns()
...
+        $sql = "CREATE UNIQUE INDEX \"tbl_testQ_2766_uix_649\" ON \"tbl_testQ_2766_649\" (\"Q_number\")";
+        $this->_conn->executeUpdate($sql);
+
+        return 650;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function checkLargeNumberOfTables(Schema $schema): void
```
morozov: Not sure which requirements these checks cover. If it's about handling different types, then it should be a set of smaller tests per type. Otherwise, a failing assertion on one column will prevent the ones for the rest of the columns to be performed.

In any event, if it's not directly about performance optimization but you think having them is important, please move them to a separate issue.

mondrake:
@morozov all your points addressed.

morozov:
@mondrake the code in the schema manager and the test looks fine to me but I'm still not sold on how fetching schema metadata is implemented on the platform level. I still dislike the fact that we have two entirely different sets of queries for getting metadata filtered and not filtered by a table.

The practical problem with that is we can never be sure that for instance `getListTableColumnsSQL()` and `getListAllColumnsSQL()` will always produce exactly the same result for a given table.

For instance, if `getListTableColumnsSQL()` is called with `$database = null` or `$database = '/'`, it will fetch metadata from the `USER*` views. But `getListAllColumnsSQL()` cannot even be called with `$database = null`, and if called with `$database = '/'`, it will produce an empty data set.

mondrake:
@morozov 

> The practical problem with that is we can never be sure that for instance getListTableColumnsSQL() and getListAllColumnsSQL() will always produce exactly the same result for a given table.

Would a test here be enough?

morozov:
It would solve the issue I mentioned above but still would leave the possibility for other similar issues to exist. To me, it would make more sense to remove code duplication instead of adding more tests which cover the same cases.

Why do we need two implementations per use case?

mondrake:
@morozov this is now taking away the code duplication. `getListTableColumnsSQL`, `getListTableForeignKeysSQL`, and `getListTableIndexesSQL` remain as one-liners calling the new methods that can all manage to fetch info for either the full db or a single table.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -411,43 +411,53 @@ protected function _getCreateTableSQL($table, array $columns, array $options = a
...
+        else {
+            $quotedDatabaseIdentifier = "(SELECT SYS_CONTEXT('userenv', 'current_schema') FROM DUAL)";
+        }
+        $tableCondition = NULL;
+        if ($table !== NULL) {
+            $tableIdentifier = $this->normalizeIdentifier($table);
+            $quotedTableIdentifier = $this->quoteStringLiteral($tableIdentifier->getName());
+            $tableCondition = "AND ind_col.table_name = $quotedTableIdentifier";
+        }
+        return <<<SQL
```
morozov: Not sure if it's enforced by the style guide, but it's preferable to have empty lines around control structures (`if`, `return`, etc).

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
```diff
@@ -36,6 +36,65 @@ class OracleSchemaManager extends AbstractSchemaManager
...
+
+        // Get all foreign keys definitions in one database call.
+        $foreignKeysByTable = $this->getAssetRecordsByTable($this->_platform->getListForeignKeysSQL($currentDatabase, null));
+
+        // Get all indexes definitions in one database call.
+        $indexesByTable = $this->getAssetRecordsByTable($this->_platform->getListIndexesSQL($currentDatabase, null));
+
+        $tables = [];
+        foreach ($tableNames as $tableName) {
+            $unquotedTableName = trim($tableName, '"');
```
morozov: The quote should be replaced with `$this->_platform->getIdentifierQuoteCharacter()`. It's not an unquoted table name, it's just a table name. And the original `$tableName` is the quoted table name.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -411,43 +411,53 @@ protected function _getCreateTableSQL($table, array $columns, array $options = a
...
+          SELECT ind_col.table_name as table_name,
+                 ind_col.index_name AS name,
+                 ind.index_type AS type,
+                 decode(ind.uniqueness, 'NONUNIQUE', 0, 'UNIQUE', 1) AS is_unique,
+                 ind_col.column_name AS column_name,
+                 ind_col.column_position AS column_pos,
+                 con.constraint_type AS is_primary
+            FROM all_ind_columns ind_col
+       LEFT JOIN all_indexes ind ON ind.owner = ind_col.index_owner AND ind.index_name = ind_col.index_name
+       LEFT JOIN all_constraints con ON  con.owner = ind_col.index_owner AND con.index_name = ind_col.index_name
```
morozov: Extra whitespace in `ON  con.owner`.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -411,43 +411,53 @@ protected function _getCreateTableSQL($table, array $columns, array $options = a
     }
 
     /**
+     * Returns the SQL for a list of indexes in the database.
+     *
+     * @param string $database
+     * @param string $table
```
morozov: It would be nice to have some explanation of the meaning of the parameters given they are nullable.

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
```diff
@@ -36,6 +36,65 @@ class OracleSchemaManager extends AbstractSchemaManager
...
+            $columns = $this->_getPortableTableColumnList($tableName, null, $columnsByTable[$unquotedTableName]);
+
+            $foreignKeys = [];
+            if (isset($foreignKeysByTable[$unquotedTableName])) {
+                $foreignKeys = $this->_getPortableTableForeignKeysList($foreignKeysByTable[$unquotedTableName]);
+            }
+
+            $indexes = [];
+            if (isset($indexesByTable[$unquotedTableName])) {
+                $indexes = $this->_getPortableTableIndexesList($indexesByTable[$unquotedTableName], $tableName);
```
morozov: Does `_getPortableTableIndexesList()` expect the quoted table name as the 2nd parameter? I cannot see it used anywhere in the implementation of `_getPortableTableIndexesList()` besides dispatching some event. Could you please explain what it is for here?

morozov (commented):
Structure-wise looks good to me.

@mondrake please see some more comments. 

Someone from the team, please take a look at the usage of quoted/unquoted table names here. I'm not really familiar with the APIs which produce the quoted ones.

morozov:
The type of `OracleSchemaManager#$_platform` should be changed to `OraclePlatform` since it uses the newly introduced Oracle-specific methods.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -411,43 +411,53 @@ protected function _getCreateTableSQL($table, array $columns, array $options = a
...
+     * @return string
+     */
+    public function getListIndexesSQL(?string $database, ?string $table): string
+    {
+        if ($database !== NULL && $database !== '/') {
+            $databaseIdentifier = $this->normalizeIdentifier($database);
+            $quotedDatabaseIdentifier = $this->quoteStringLiteral($databaseIdentifier->getName());
+        }
+        else {
+            $quotedDatabaseIdentifier = "(SELECT SYS_CONTEXT('userenv', 'current_schema') FROM DUAL)";
```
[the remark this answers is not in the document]
mondrake: The point here is that we do not have a connection to query against in the Platform object.

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
mondrake: Just passing over to honour the arguments and the chain or parent calls. The code here does not do anything with that.

mondrake:
@morozov  made the changes you suggested. Some comments from my side added to yours that now show having been outdated.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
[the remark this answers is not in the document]
morozov: Then we should use `ALL_*` objects for the non-current schema and `USER_*` ones for the current one.

> Because in the new concept we use the `ALL_*` views that provide information about Oracle DBMS objects, instead of the slower `USER_*` views that do a lot of more checking about user being authorised to access an object (checked with EXPLAIN PLAN).

Is using `USER_*` slower than using a sub-query on each row of `ALL_*`?

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -418,36 +470,19 @@ protected function _getCreateTableSQL($table, array $columns, array $options = a
...
-             WHERE     uind_col.table_name = " . $table . "
-             ORDER BY  uind_col.column_position ASC";
+    /**
+     * Returns the SQL for a list of all indexes in the database.
+     *
+     * @param string $database The database schema. If NULL or '/', the currently active schema will be queried.
+     *
+     * @return string
+     */
+    public function getListAllIndexesSQL(?string $database = null): string
```
morozov: It should be either nullable or optional, not both. I think the former is preferable.

Majkl578: I'll strongly disagree, it should either defaulting to null AND nullable or none. Optionality and nullablility are two things (and for null, the former implies the latter, but not the other way). Missing nullability sign for parameters defaulting to null is PHP's design flaw introduced only because 7.0 had no nullables.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -361,6 +361,24 @@ public function getClobTypeDeclarationSQL(array $field)
         return 'CLOB';
     }
 
+    /**
+     * Returns the database condition for querying the schema.
+     *
+     * @param string $database The database schema. If NULL or '/', the currently active schema will be queried.
+     *
+     * @return string
```
Majkl578: Please remove useless (duplicated) `@return`, it doesn't add any extra information.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -361,6 +361,24 @@ public function getClobTypeDeclarationSQL(array $field)
...
+     */
+    private function getDatabaseCondition(?string $database): string
+    {
+        if ($database !== null && $database !== '/') {
+            $databaseIdentifier = $this->normalizeIdentifier($database);
+            return $this->quoteStringLiteral($databaseIdentifier->getName());
+        }
+        else {
```
Majkl578: no need for else, we have return above

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -410,6 +428,40 @@ protected function _getCreateTableSQL($table, array $columns, array $options = a
...
+    {
+        $databaseCondition = $this->getDatabaseCondition($database);
+        $tableCondition = '';
+        if ($table !== null) {
+            $tableIdentifier = $this->normalizeIdentifier($table);
+            $quotedTableIdentifier = $this->quoteStringLiteral($tableIdentifier->getName());
+            $tableCondition = " AND ind_col.table_name = $quotedTableIdentifier";
+        }
+        return
+<<<SQL
```
Majkl578: Should be on the same line as return, but I'm not sure ATM how this is handled by CS.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -410,6 +428,40 @@ protected function _getCreateTableSQL($table, array $columns, array $options = a
         return $sql;
     }
 
+    /**
+     * Returns the SQL for a list of indexes in the database.
+     *
+     * @param string $database The database schema. If NULL or '/', the currently active schema will be queried.
+     * @param string $table    The table. If left NULL, the SQL will return all the indexes in the database.
+     *
+     * @return string
```
Majkl578: this `@return` is also useless

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -418,36 +470,19 @@ protected function _getCreateTableSQL($table, array $columns, array $options = a
...
-                       ) AS is_primary
-             FROM      user_ind_columns uind_col
-             WHERE     uind_col.table_name = " . $table . "
-             ORDER BY  uind_col.column_position ASC";
+    /**
+     * Returns the SQL for a list of all indexes in the database.
+     *
+     * @param string $database The database schema. If NULL or '/', the currently active schema will be queried.
+     *
+     * @return string
```
Majkl578: this `@return` is also useless

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -648,38 +705,52 @@ public function getListTableConstraintsSQL($table)
...
-                         ) AS comments
-                FROM     $tabColumnsTableName c
-                WHERE    c.table_name = " . $table . " $tabColumnsOwnerCondition
-                ORDER BY c.column_id";
+    /**
+     * Returns the SQL for a list of all columns in the database.
+     *
+     * @param string $database The database schema. If NULL or '/', the currently active schema will be queried.
+     *
+     * @return string
```
Majkl578: this `@return` is also useless

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -648,38 +705,52 @@ public function getListTableConstraintsSQL($table)
         return "SELECT * FROM user_constraints WHERE table_name = " . $table;
     }
 
+    /**
+     * Returns the SQL for a list of columns in the database.
+     *
+     * @param string $database The database schema. If NULL or '/', the currently active schema will be queried.
+     * @param string $table    The table. If left NULL, the SQL will return all the indexes in the database.
+     *
+     * @return string
```
Majkl578: this `@return` is also useless

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -605,36 +640,58 @@ private function getAutoincrementIdentifierName(Identifier $table)
...
-       ON alc.constraint_name = cols.constraint_name
-      AND alc.constraint_type = 'R'
-      AND alc.table_name = " . $table . "
-    ORDER BY cols.constraint_name ASC, cols.position ASC";
+    /**
+     * Returns the SQL for a list of all foreign keys in the database.
+     *
+     * @param string $database The database schema. If NULL or '/', the currently active schema will be queried.
+     *
+     * @return string
```
Majkl578: this `@return` is also useless

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -605,36 +640,58 @@ private function getAutoincrementIdentifierName(Identifier $table)
             : $identifierName;
     }
 
+    /**
+     * Returns the SQL for a list of foreign keys in the database.
+     *
+     * @param string $database The database schema. If NULL or '/', the currently active schema will be queried.
+     * @param string $table    The table. If left NULL, the SQL will return all the indexes in the database.
+     *
+     * @return string
```
Majkl578: this `@return` is also useless

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -605,36 +640,58 @@ private function getAutoincrementIdentifierName(Identifier $table)
...
+     */
+    private function getListForeignKeysSQL(?string $database, ?string $table): string
+    {
+        $databaseCondition = $this->getDatabaseCondition($database);
+        $tableCondition = '';
+        if ($table !== null) {
```
Majkl578: please add newlines around if to improve readability

Majkl578 (dismissed):
Just some CS nitpicking, mostly.

mondrake:
@Ocramius can we remove the Missing Tests tag? That's been addressed based on @morozov directions.

mondrake:
The problem is still there, schema introspection is practically unusable on an Oracle database. If anyone interested, https://github.com/mondrake/dbal/tree/oracle-schema-list-tables has the fix for DBAL 3.

greg0ire:
Sorry, I didn't mean to close this, I just meant to rename `master` to `4.0.x`. Why does it say `unknown repository` as source repository?

mondrake:
@greg0ire because in the meantime I actually dropped and re-forked the repo, so the originally referenced fork branch is gone. Probably the best course is to keep this PR closed now, and open a new one instead.

Issue #1592, opened 2010-08-05
DBAL-40: Transparent table&column names escaping


doctrinebot:
Jira issue originally created by user jantichy:

Hello, I would like to re-open the discussion about automatic transparent escaping of all table/column names sent from DBAL to database. It was already discussed in http://www.doctrine-project.org/jira/browse/[DDC-88](http://www.doctrine-project.org/jira/browse/DDC-88) without any satisfactory result.

Why do I have to quote any reserved word used in table or column name? Why Doctrine doesn't do this automatically for all table and column 
names used in generated SQL queries? 

Before you start to explain how complicated it is and what problems you will be faced with, try to look at excellent DIBI database layer -  how it acts in this way - it's behaviour is very cool. Unfortunally at the moment the full documentation is in czech only, but here is a brief automatic google-translation to english - http://dibiphp.com/en/quick-start.

My suggestion to Doctrine 2 ORM/DBAL solution is:
1. Developer should never care about any escaping or avoiding any reserved words - it is not his business, the DBAL shoult solve it transparently and safely.
2. So there should be no need and even no possibility to add any quotation chars in @column or @table annotations as well as in DQL queries. ORM layer has nothing to do with escaping, it is all a business of the DBAL layer. Current possibility for manual escaping the names in mentioned annotations is totally wrong and should be discontinued.
3. DBAL should escape ALL table and column names transparently and automatically. There should be ne option to enable or disable the escaping, there is no reason for disabling it.
4. The escaping should be performed just in the final translation of DBAL queries to native SQL query, not earlier. This is the right place to do that.

So what do you think about that?

msypes:
Not sure if this is the best place to leave this, but a string of searches led me here. Using the Symfony2 console generate entities and schema from a YAML file, I had no problem creating an entity with a "group" property and a matching table with a "group" column. I ran into a syntax error attempting to load data fixtures with Alice. (See https://github.com/hautelook/AliceBundle/issues/179)
While http://doctrine-orm.readthedocs.org/projects/doctrine-orm/en/latest/reference/basic-mapping.html#quoting-reserved-words describes how to deal with quoting these in an annotation, there's nothing there to describe how to do so in a YAML file. If it is possible, I'd appreciate it if the documentation were updated. In the meantime, I'll guess I'll just rename my column/property, which is feeling ugly but unavoidable.

spacetraveller:
Thank you @msypes , you are my hero, finally an answer that gets to the bottom of this. 

[...]

PR #2768, opened 2017-07-05
Add all missing MariaDB keywords to the MySQL platform

roelvanduijnhoven:
Version 10.2.4 introduced `rows`  as a keyword. As doctrine does not escape non-keyword column names; this breaks stuff. And thus this prevents us from upgrading to the lastest MariaDb version.

You connect to MariaDb using the MySql connectors from Doctrine. Thus I think it would be wise to add the new keywords to this file. 

I added all missing keywords from https://mariadb.com/kb/en/mariadb/reserved-words/ and resorted the table in this PR.


Ocramius (approved):
LGTM

@deeky666 can you check this? Do we have a way to test this properly?

leofeyer:
Also see: https://jira.mariadb.org/browse/MDEV-13339

According to the MariaDV devs, `rows` will be a reserved keyword in MySQL 8.0. @Ocramius There is a `MySQLKeywords` and and `MySQL57Keywords` class. Do we need a `MySQL80Keywords` class?

Ocramius:
@leofeyer sadly yes, although we don't have a platform for it yet

roelvanduijnhoven:
@Ocramius Isn't it much simpler to simply escape all column names in an upcoming version? I see there is already some work done on that area: https://github.com/doctrine/dbal/issues/1592. Better investment of time than manually updating those keyword files I'd say.

Ocramius:
We can do it in 3.*, bit quoting by default has issues with case
sensitivity handling on some engines.

roelvanduijnhoven:
What if we simply make the quoting behaviour configurable per engine? We can then already land this in the next minor increment of this repo. For MySQL (and other engines that support this) we can remove the keywords altogether; and simply enable quoting for that engine

Ocramius:
It really is much simpler: we'll move to quoting-only for 3.x, not in 2.x

Ocramius:
:ship: 

Thanks @roelvanduijnhoven!

ausi:
@Ocramius is it possible to merge this into the 2.5 branch too and release a 2.5.14 version?

Currently it is not possible to use MariaDB 10.2.4 together with PHP 5.6 or 7.0 because doctrine/dbal 2.6 requires at least PHP 7.1.

Ocramius:
@ausi no, this is an improvement on the API. If you want to use a newer version of MariaDB, then upgrade also DBAL and consequently PHP.

ausi:
OK, thanks for checking!

roelvanduijnhoven:
I just migrated, and it failed. 

Turns out the `rows` keywords is _not_ escaped if used within an UPDATE statement :thinking:. Not sure why. the `MysqlKeywords` file isn't even used when generating an update query it seems.

`rows` _is_ properly escaped for alter and create statements.

Is this intended behaviour? Not sure if I should open an issue here.

xcy7e:
Manually fixed this in 4.4 LTS and it worked. Why not change the 4.4 globally? Any reason?

fritzmg:
@xcy7e what exactly are you referring to?

PR #2760, opened 2017-06-29
Add documentation regarding transactional exceptions

dsantang:
Fixes #2290

PR #2770, opened 2017-07-05
Fix namespace separator in the exceptions in the transactional handling documentation

Tobion:
Fixes #2760


Ocramius:
Thanks @Tobion!

PR #2772, opened 2017-07-06
Add quoting around index names when altering MySQL schema information

gabrielrincon:
This is the solution for the index name like it have a "-" in the name. The char "`" is necesary for scape the "-" char.

I have the next error:

[Doctrine\DBAL\Exception\SyntaxErrorException]                               
  An exception occurred while executing 'DROP INDEX year-month ON ajustes_man  
  uales':                                                                      
  SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error i  
  n your SQL syntax; check the manual that corresponds to your MySQL server v  
  ersion for the right syntax to use near '-month ON ajustes_manuales' at lin  
  e 1                                                                          
                                                                               

                                                                               
  [Doctrine\DBAL\Driver\PDOException]                                          
  SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error i  
  n your SQL syntax; check the manual that corresponds to your MySQL server v  
  ersion for the right syntax to use near '-month ON ajustes_manuales' at lin  
  e 1                                                                          
                                                                               

                                                                               
  [PDOException]                                                               
  SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error i  
  n your SQL syntax; check the manual that corresponds to your MySQL server v  
  ersion for the right syntax to use near '-month ON ajustes_manuales' at lin  
  e 1

PR #2773, opened 2017-07-08
Bug with index named whit minus symbol "-"

gabrielrincon:
if the name is idx-foo the compile generates error

https://github.com/doctrine/dbal/pull/2772#issuecomment-313867262


Ocramius:
@gabrielrincon this test change is invalid - what is being asked in #2772 is an additional test for indexes with a particular name (such as ones with special characters or non-sql-compliant naming).

Closing here - the change itself can be re-applied in #2772 (just push to the same branch), but additional tests are needed.

PR #2775, opened 2017-07-09
fixed doc param inside Query Builder

AlessandroMinoccheri:
Fixed a parameter documentation from string to mixed because it can be a string or an array

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -409,7 +409,7 @@ class QueryBuilder
      * 'groupBy', 'having' and 'orderBy'.
      *
      * @param string  $sqlPartName
-     * @param string  $sqlPart
+     * @param mixed  $sqlPart
      * @param boolean $append
      *
      * @return $this This QueryBuilder instance.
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -409,7 +409,7 @@ public function getMaxResults()
      * 'groupBy', 'having' and 'orderBy'.
      *
      * @param string  $sqlPartName
-     * @param string  $sqlPart
+     * @param mixed  $sqlPart
```
Ocramius: Should probably be `string`, no?

AlessandroMinoccheri: In some part of the code for example in line 530 is passed as an array not as a string right?

Ocramius: That's because it's upcast (line 423)

AlessandroMinoccheri: Yes I have seen but I thought that you can call this function with a string or with an array so I thought that mixed type is better than string inside documentation.
If It's better `string` I understand but there are two ways for me to call this function for that parameter

Ocramius: @AlessandroMinoccheri use `string|string[]` or whatever matches then :-)

morozov:
Obsolete as of https://github.com/doctrine/dbal/pull/3836.

PR #2777, opened 2017-07-12
failover on dead slave by force connect to master

grzonu:
A little feature to prevent situation when slave is down and application throw connection refuse exception

### DIFF
--- a/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
+++ b/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
@@ -180,7 +180,14 @@ class MasterSlaveConnection extends Connection
                 $this->connections['slave'] = $this->connections['master'];
             }
         } else {
-            $this->connections['slave'] = $this->_conn = $this->connectTo($connectionName);
+            try {
+                $this->connections['slave'] = $this->_conn = $this->connectTo($connectionName);
+            } catch(\Exception $ex) {
+                //failover when slave is dead
+                $this->connections['slave'] = $this->_conn =
+                    (isset($this->connections['master']) && is_object($this->connections['master'])) ?
+                    $this->connections['master'] : $this->connectTo('master');
+            }
         }
 
         if ($this->_eventManager->hasListeners(Events::postConnect)) {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -180,7 +180,14 @@ public function connect($connectionName = null)
                 $this->connections['slave'] = $this->connections['master'];
             }
         } else {
-            $this->connections['slave'] = $this->_conn = $this->connectTo($connectionName);
+            try {
+                $this->connections['slave'] = $this->_conn = $this->connectTo($connectionName);
+            } catch(\Exception $ex) {
+                //failover when slave is dead
+                $this->connections['slave'] = $this->_conn =
```
Ocramius: This scenario still requires a test. Also, I think it's really time for us to have a proper connection pooling abstraction for 3.x :-\

grzonu: I added test for this case, what do you think about that now ?

tests/Doctrine/Tests/DBAL/Functional/MasterSlaveConnectionTest.php
```diff
@@ -139,4 +143,15 @@ public function testMasterSlaveConnectionCloseAndReconnect()
         $conn->connect('master');
         $this->assertTrue($conn->isConnectedToMaster());
     }
+
+    public function testMasterSlaveFailSlave()
+    {
+        $params = $this->prepareConnectionParameters();
+        $invalidParams = $params['master'];
+        $invalidParams['password'] = 'MUUHAAAAHAAAA';
```
Ocramius: I think this test exposes a problem with your implementation (and the test should be adapted too): the only exception to be caught should be a connection failed exception (due to timeout or similar), and not an authentication or setup failure.

Having the entire thing crash on mis-configuration is actually intentional, whereas the functionality you really want here is retrying on failed connection, rather than authentication.

grzonu: You are right, this implementation is not complete, because cover too many cases(like bad auth params) so i think how to make it better ;)
Functionality i want is not retry on failed connection just try other slave or master because if slave is down or network between app server and database server is broken i want to fallback connection to other host

morozov:
Closing as stale.

PR #2782, opened 2017-07-19
Stop using the deprecated `json_array` type internally

lcobucci:


Ocramius (approved):
@lcobucci any user-side change required here? Parch seems good otherwise.

lcobucci:
@Ocramius I think it shouldn't required but TBQH I'm not sure.

Ocramius:
YOLO, :ship:

bendavies:
@Ocramius @lcobucci 
there is a BC break/bug here.

changing the `doctrineTypeMapping` from `json_array` to `json` causes unfixable schema comparison.

i have an old column using `json_array` and this PR results in a schema diff of:
```
ALTER TABLE settings ALTER parameters TYPE JSON
ALTER TABLE settings ALTER parameters DROP DEFAULT
```
Applying the diff does nothing. the same diff persists afterwards.

Only way for the user to fix this is to change the column type from `json_array` to `json`

lcobucci:
@bendavies that's sad news, sorry for causing this issue.

I'll be checking the options to fix the problem caused by 31d01283b0bd9f9b3cafa8c6c06d6a45838ad0dd.

Thanks for letting us know of that

bendavies:
Thanks @lcobucci!

stof:
this requires changing which type requires SQL comment hint. currently, json_array says it does not require a SQL comment hint when native json type is supported in the platform, because platforms would map the native type to it. But as you changed it, you have to change the comment hint requirement here

lcobucci:
@bendavies @stof I think I've managed to fix the reported BC-break on https://github.com/doctrine/dbal/pull/2855, could you please check (it's still WIP, though)?

PR #2786, opened 2017-07-21
Map error code 1429 to `ConnectionException` in the `AbstractMySQLDriver`

SpacePossum:
`1429 Unable to connect to foreign data source`

It might be even better to split up the `ConnectionException`,
making `ConnectionException` an interface and have one implementation for `client-to-server/node` exceptions and one for `server/node-server/node` exception.
However that would be a BC break.
Other way could be to have a new exception for  `server/node-server/node` issue.

Any thoughts?

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
@@ -102,6 +102,7 @@ abstract class AbstractMySQLDriver implements Driver, ExceptionConverterDriver, 
             case '1143':
             case '1227':
             case '1370':
+            case '1429':
             case '2002':
             case '2005':
                 return new Exception\ConnectionException($message, $exception);


Ocramius:
@spacepossum how often would you catch those specific exception types?

SpacePossum:
Communication issues within the cluster (like 1492) typically means someone else has to take a look to see what went wrong, than when for example an `SyntaxErrorException` has been thrown. So I would like to handle those differently.

I can fail back on the error codes of mysql itself, although loose a bit of abstraction from the database.
So if you feel fine graining the (mysql) errors to dedicated exceptions even more is a bit to much, than I can see that point.

Ocramius:
I don't have an overview of all exception types here, but a simplistic approach would be OK as a start (this current patch) and a separate issue should be raised if there is a need for a complete inheritance tree overhaul

Ocramius:
Meanwhile, merging here, thanks!

PR #2753, opened 2017-06-28
problem with GeneratedValue (uuid type) column query mapping

luked3v3l0p3r:

PR #2788, opened 2017-07-25
sqlsrv - force string cast when parameter is of type string

maglnet:
When an object that implements `__toString()` is bound as an parameter of type `PDO::PARAM_STR` it raised the following exception when using `sqlsrv` as driver:

```
1) Doctrine\Tests\DBAL\Functional\Driver\SQLSrv\StatementTest::testObjectsWillBeCastedToString
Doctrine\DBAL\Driver\SQLSrv\SQLSrvException: SQLSTATE [IMSSP, -16]: An invalid PHP type for parameter 1 was specified.
```

The problem was only reproducable with `sqlsrv` driver and came up when using `UUID` objects.

Since we still have projects running PHP 5.6 and MSSQL it would be nice to see this merged and backported to `2.5.x`.

Edit: this handles #2753

### DIFF
--- a/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
+++ b/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
@@ -264,6 +264,9 @@ class SQLSrvStatement implements IteratorAggregate, Statement
                     SQLSRV_PHPTYPE_STREAM(SQLSRV_ENC_BINARY),
                     SQLSRV_SQLTYPE_VARBINARY('max'),
                 ];
+            } elseif (PDO::PARAM_STR === $this->types[$column]) {
+                $variable = (string) $variable;
+                $params[$column - 1] =& $variable ;
             } else {
                 $params[$column - 1] =& $variable;
             }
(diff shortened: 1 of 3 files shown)


lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
```diff
@@ -264,6 +264,9 @@ private function prepare()
                     SQLSRV_PHPTYPE_STREAM(SQLSRV_ENC_BINARY),
                     SQLSRV_SQLTYPE_VARBINARY('max'),
                 ];
+            } elseif (PDO::PARAM_STR === $this->types[$column]) {
+                $variable = (string) $variable;
```
Ocramius: This cast should not be applied. This makes any type mismatch failures silent.

A `StringCastableObject` should be mapped via a custom DBAL type that allows string castable objects instead.

Ocramius (changes requested):
Closing as `invalid`. If a non-string value is expected, then the bug is in the other adapters if they do implicitly apply a cast. Our intent is to handle types as if we were using `declare(strict_types=1)`

PR #2791, opened 2017-07-26
Add support for more mathematical expressions

### LINKED ISSUE #2790: Support more mathematical expressions
The AbstractPlatform already provides a some math expressions, like ACOS, PI, SIN, etc.

However, some expressions are missing that are required to write, for example, a Doctrine DBAL based radial search.

Following expressions are currently unsupported, afaics:
* POWER
* ASIN
* TAN
* ATAN
* RADIANS
* DEGREES

andreaskienast:
This patch adds support for these math expressions:

* POWER
* ASIN
* TAN
* ATAN
* RADIANS
* DEGREES

Fixes #2790

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -737,6 +737,18 @@ abstract class AbstractPlatform
     }
 
     /**
+     * Returns the SQL to get the powered value of a column.
+     *
+     * @param string $value
+     *
+     * @return string
+     */
+    public function getPowerExpression($value)
+    {
+        return 'POWER(' . $value . ')';
+    }
+
+    /**
      * Returns the SQL snippet to round a numeric field to the number of decimals specified.
      *
      * @param string  $column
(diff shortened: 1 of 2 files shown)


morozov:
@andreasfernandez, are implementations of these functions the same for all platforms (part of some SQL standard)? If so, why implement them in the abstraction layer which the DBAL is?

For some domains, both `power()` and `log()` may be used as inverse functions but we only support the former. We support `pi()` but don't support `exp()`. With this approach, we may end up porting everything to the DBAL.

If the team decides the DBAL still needs this, there also should be functional tests in addition to the unit ones to prove that the new code works correctly on all supported platforms.

andreasfernandez:
@morozov, I have checked at least MySQL, MSSQL, Postgresql and Oracle, the functions are all named the same, but IMO that doesn't mean any other supported DMBS names them the same as well.

The fact that `POWER` is only an alias of `POW` in MySQL makes me feel unsure about using it as it may get dropped in future, which renders such a method necessary, IMO. The introduced method `getPowerExpression()` should be overridden in MySqlPlatform to use `POW` explicitly, then.

The original reason I created this patch is that I needed to write a DBAL-compliant radial search. I took [this example](https://developers.google.com/maps/articles/phpsqlsearch_v3#findnearsql) by Google as template for the query and found more mathematical expressions that are missing.

morozov:
> The original reason I created this patch is that I needed to write a DBAL-compliant radial search. I took [this example](https://developers.google.com/maps/articles/phpsqlsearch_v3#findnearsql) by Google as template for the query and found more mathematical expressions that are missing.

The DBAL allows using arbitrary SQL expressions in query parts, and in your example the resulting SQL doesn't depend on the platform. So the query like `$query->select('id', '(3959 * ACOS(…) AS distance')` is DBAL-compliant and portable, and doesn't require any modifications to the DBAL.

PR #2793, opened 2017-07-26
Fix articles of words sounding with a consonant (SQL => ESS-kew-ELL)

afoeder:


morozov:
I think it depends on the reader: most of my colleagues and I pronounce SQL as |ˈsiːkwəl|, so in this case, the existing grammar is correct. However, I'm not claiming to be an expert in English.

Ocramius:
[Both pronunciations OK](https://en.wikipedia.org/wiki/SQL). That said, merging this, as it is still a valuable contribution, and could raise the discussion for other areas in the documentation too.

afoeder:
hah, I wanted to make it "special well" and looked it up in [wiktionary](https://en.wiktionary.org/wiki/SQL#Usage_notes), Ess Kyuu Ell there… However the [usage notes below](https://en.wiktionary.org/wiki/SQL#Usage_notes) also state

> The first form of SQL was called SEQUEL, pronounced like sequel. As a result, many speakers still pronounce SQL like sequel. (For these speakers, SQL is an acronym.)

Thanks for merging :)

PR #2796, opened 2017-07-28
PHP 7.1 typing

tolbon:
Hello, I am working on "typing DBAL" because now minimum version is 7.1 and PHP add this new feature.

WORK IN PROGRESS

this work permit to see some inaccurate docs

### DIFF
--- a/lib/Doctrine/DBAL/Cache/ArrayStatement.php
+++ b/lib/Doctrine/DBAL/Cache/ArrayStatement.php
@@ -58,15 +58,16 @@ class ArrayStatement implements \IteratorAggregate, ResultStatement
     /**
      * {@inheritdoc}
      */
-    public function closeCursor()
+    public function closeCursor(): bool
     {
         unset ($this->data);
+        return true;
     }
 
     /**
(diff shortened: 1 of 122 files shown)


jvasseur:
Adding a type hint to anything that can be extended is a BC break with PHP < 7.2

tolbon:
I am sorry I don't understand why it is BC break ?
composer.json say minimum supported version is PHP 7.1 so I use feature of PHP 7.1

maybe I do it wrong ?

jvasseur:
If have create a class extending one of doctrine's class in your project, adding types in doctrine means your code will break with "incompatible method signature" errors.

The only place we can add types is in method constructors (since they don't have to be compatible) and final classes/methods (since they can't be extended).

With PHP 7.2 we could do it on method arguments (but not method return types): https://wiki.php.net/rfc/parameter-no-type-variance

tolbon:
Oh I understand now. So I close this PR ?

jvasseur:
You can keep the changes for constructors and methods of final classes.

PR #2797, opened 2017-07-28
Typing final class DriverManager

tolbon:


Ocramius:
LGTM, thanks @tolbon!

PR #2799, opened 2017-07-28
Fix Shard Manager docs with more accurate docblocks in `SQLAzureShardManager` and `PoolingShardManager`

tolbon:
Fix Shard Manager Docs mor accurate with code in SQLAzureShardManager and PoolingShardManager


Ocramius:
LGTM! Thanks @tolbon!

PR #2800, opened 2017-07-28
Fix inconsistent/missing return statements

tolbon:


Ocramius:
LGTM, thanks @tolbon! :+1:

PR #2803, opened 2017-07-30
Array Type - Empty String (NOT NULL) support

ruicampos:
Setting a column as Array and NOT NULL for a table with existing data - database (at least MySQL) will set that column as '' (empty string) for existing rows.
And fetching those rows, Doctrine throws a ``Converstion Exception`` with message ``Could not convert database value "" to Doctrine Type array``

This PR fixes that by assuming empty string as empty array.

### DIFF
--- a/lib/Doctrine/DBAL/Types/ArrayType.php
+++ b/lib/Doctrine/DBAL/Types/ArrayType.php
@@ -55,6 +55,11 @@ class ArrayType extends Type
         }
 
         $value = (is_resource($value)) ? stream_get_contents($value) : $value;
+        // Declaring array type on NOT NULL column with existing table data will save '' for array column in existing rows.
+        if( $value === '' )
+        {
+            return [];
+        }
         $val = unserialize($value);
         if ($val === false && $value != 'b:0;') {
             throw ConversionException::conversionFailed($value, $this->getName());
--- a/tests/Doctrine/Tests/DBAL/Types/ArrayTest.php
+++ b/tests/Doctrine/Tests/DBAL/Types/ArrayTest.php
@@ -56,4 +56,10 @@ class ArrayTest extends \Doctrine\Tests\DbalTestCase
     {
         $this->assertFalse($this->_type->convertToPHPValue(serialize(false), $this->_platform));
     }
+
+    public function testEmptyStringConvertsToEmptyArray()
+    {
+        $this->assertEquals([], $this->_type->convertToPHPValue('', $this->_platform));
+    }
+
 }


morozov:
> Setting a column as Array and NOT NULL for a table with existing data - database (at least MySQL) will set that column as '' (empty string) for existing rows.

How does this setting happen? Could you create an isolated test case?

I believe this issue should be fixed during migration, not in run time.

morozov (changes requested):
> Setting a column as Array and NOT NULL for a table with existing data - database (at least MySQL) will set that column as '' (empty string) for existing rows.

How does this setting happen? Could you create an isolated test case?

I believe this issue should be fixed during migration, not in run time.

ruicampos:
The use case is a Symfony 2 application. In that application I already have a ``SentSms`` entity and this table already has several rows.
Then I added the following entity field to my YAML definition:
```
providerReference:
    type: array
    nullable: false
```

Running a ``doctrine:schema:update --dump-sql`` will generate the following change to be applied to the database:

```
ALTER TABLE sent_sms ADD provider_reference LONGTEXT NOT NULL COMMENT '(DC2Type:array)';
```

If I run this ALTER TABLE or if I run ``doctrine:schema:update --force`` it will succeed on applying the schema change to database. So far, so good.
But all existing rows will have ``''`` (empty string) as the provider_reference column is NOT NULL.
Then, trying to find any existing SentSms records within Doctrine will throw a ``ConversionException``

Indeed you can consider this a migration issue rather than a runtime issue that can be solved by:
* Defining the column as NULLABLE in the first place
* After the ``ALTER TABLE`` run a mass update: ``UPDATE sent_sms SET provider_reference='a:0:{}';``

But in my test case the developer experience is that Doctrine generated the update schema SQL, applied it successfully and then it throws errors in runtime. 
So, I believe that with my change there are no other impacts and you make this use case work out of the box in a similar manner as if the entity field was defined as NULLABLE in the first place.

> In our case and to solve the issue we made the column NULLABLE (no problem for our use case) but I decided to submit the PR anyhow.

Ocramius:
> So, I believe that with my change there are no other impacts and you make this use case work out of the box in a similar manner as if the entity field was defined as NULLABLE in the first place.

No, the use-case here is that you now have broken data in your DB, which needs to be fixed by a migration (after a backup, since it's a possibly destructive change). The type needs to be changed indeed, but in such a way that the default column value is either set, or nullability is allowed. This is part of a larger change that could affect any specific type.

This patch as-is is just mitigating a symptom, but not fixing the problem.

ruicampos:
> No, the use-case here is that you now have broken data in your DB, which needs to be fixed by a migration (after a backup, since it's a possibly destructive change)

We're not using migrations on our project. But it may be the solution.

> The type needs to be changed indeed, but in such a way that the default column value is either set, or nullability is allowed. This is part of a larger change that could affect any specific type.

Agreed. I didn't check the other types which may have similar issues with default values on existing rows. 

> This patch as-is is just mitigating a symptom, but not fixing the problem.

My rationale is that with my patch the behaviour is very similar with loading ``NULL`` values from database if column allows ``NULL``.

But I understand it may require a larger change covering all the types and defining the best strategy on how to deal with these issues. Feel free to close the PR.

Ocramius:
Closing as per https://github.com/doctrine/dbal/pull/2803#issuecomment-319111941

Thanks for your understanding, @ruicampos.

PR #2807, opened 2017-08-02
fix for comparator so it can correctly see added indexes.

arminneman:
starting point:
```sql
CREATE TABLE `someTable` (
	`datum` DATE NOT NULL,
	PRIMARY KEY (`datum`),
	INDEX `i_system_date_changed` (`system_date_changed`),
	INDEX `datum` (`datum`)
)
COLLATE='utf8_unicode_ci'
ENGINE=InnoDB
;
```

new situation:
```sql
CREATE TABLE `someTable` (
	`datum` DATE NOT NULL,
	`id` INT(10) UNSIGNED NOT NULL AUTO_INCREMENT,
	PRIMARY KEY (`id`),
	INDEX `i_system_date_changed` (`system_date_changed`),
	INDEX `datum` (`datum`)
)
COLLATE='utf8_unicode_ci'
ENGINE=InnoDB
;
```
The current comparison in comparator will not see a change here because the index name will return primary for both tables. Causing it not to trigger addedIndexes, which in turn will lead to an incorrect statement where the id key and the primary tag get added separately. MySQL will not allow this because AUTOINCREMENT can only be applied on the primary key.

```sql
ALTER TABLE someTable ADD id INT UNSIGNED AUTO_INCREMENT NOT NULL;
ALTER TABLE someTable ADD PRIMARY KEY (id);
```

```
SQLSTATE[42000]: Syntax error or access violation: 1075 Incorrect table definition; there can be only one auto column and it must be defined as a key
```

This, however, is accepted.
```sql
ALTER TABLE someTable ADD id INT UNSIGNED AUTO_INCREMENT NOT NULL, ADD PRIMARY KEY (id); 
```

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -620,8 +620,9 @@ class MySqlPlatform extends AbstractPlatform
             }
             $sql = array_merge(
                 $this->getPreAlterTableIndexForeignKeySQL($diff),
-                $sql,
-                $this->getPostAlterTableIndexForeignKeySQL($diff)
+                $sql
+            //this is causing trouble, adds another drop primary key
+//                $this->getPostAlterTableIndexForeignKeySQL($diff)
             );
         }
 
--- a/lib/Doctrine/DBAL/Schema/Comparator.php
+++ b/lib/Doctrine/DBAL/Schema/Comparator.php
@@ -244,7 +244,10 @@ class Comparator
 
         /* See if all the indexes in table 1 exist in table 2 */
         foreach ($table2Indexes as $indexName => $index) {
-            if (($index->isPrimary() && $table1->hasPrimaryKey()) || $table1->hasIndex($indexName)) {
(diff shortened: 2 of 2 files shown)


Ocramius:
Please target `master` in a new patch.

Also, tests have to be written beforehand.

Closing as `incomplete`

arminneman:
Thanks for the feedback, so I create a pull request on the master branch? Regarding the tests, I have not added any new features, will the current tests not suffice? 

Btw is there a way that allows me to run tests through the CI without polluting the project?

Ocramius:
> Thanks for the feedback, so I create a pull request on the master branch?

Yep!

 > Regarding the tests, I have not added any new features, will the current tests not suffice?

No, every unexpected behavior needs to be tested to ensure that the fix is correct, and that we don't introduce regressions by changing the code later on.

 > Btw is there a way that allows me to run tests through the CI without polluting the project?

You can enable travis-ci for your local repository by going to https://travis-ci.org/ and enabling it there.
I still suggest you run your tests locally. See https://github.com/doctrine/dbal/blob/399ab63ddf3db48a4befb7d4f6bcf699e7796d5f/tests/README.markdown

arminneman:
Okay, I'll look into it, thanks again!

PR #2810, opened 2017-08-03
add @throws declarations to docs in Connection

### LINKED ISSUE #2809: Missing @throws declarations in Connection
In a few cases methods in Connection may throw a DBALException which isn't declared in the docblock. As the Connection class is heavily used by third party code, this code runs the risk of ignoring those exceptions instead of handling them. While it is technically not required to declare exceptions in PHP, it is highly recommended, especially in public APIs.

bestform:
In a few cases methods in Connection may throw a DBALException which
isn't declared in the docblock. As the Connection class is heavily
used by third party code, this code runs the risk of ignoring
those exceptions instead of handling them. While it is
technically not required to declare exceptions in PHP, it is highly
recommended, especially in public APIs.

Fixes #2809

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -413,7 +413,7 @@ class Connection implements DriverConnection
      *
      * @return string|null
      *
-     * @throws \Exception
+     * @throws Exception
      */
     private function getDatabasePlatformVersion()
     {
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -322,6 +322,8 @@ class Connection implements DriverConnection
      * Gets the DatabasePlatform for the connection.
      *
      * @return \Doctrine\DBAL\Platforms\AbstractPlatform
+     *
+     * @throws \Doctrine\DBAL\DBALException
      */
     public function getDatabasePlatform()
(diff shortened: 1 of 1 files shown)


bestform:
To be honest I don't really understand the failure condition this patch is meeting in scrutinizer. Is a changed doc comment a failure condition?

bestform:
Oh, never mind. I see. There are other unrelated errors in doc comments. My patch doesn't fix those - as it shouldn't - which leads to the error in scrutinizer.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -410,6 +412,8 @@ private function detectDatabasePlatform()
      * version without having to query it (performance reasons).
      *
      * @return string|null
+     *
+     * @throws \Exception
```
Ocramius: `Exception` is already imported

Ocramius (changes requested):

lib/Doctrine/DBAL/Connection.php
bestform: True. Fixed.

Ocramius:
@bestform thanks!

Ocramius:
@bestform thanks!

bestform:
anytime. Thanks for the merge.

PR #2813, opened 2017-08-07
fixed return type into OCI8Statement

AlessandroMinoccheri:
Fixed return type from string to array into OCI8Statement


lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -119,7 +119,7 @@ public function __construct($dbh, $statement, OCI8Connection $conn)
      *
      * @param string $statement The SQL statement to convert.
      *
-     * @return string
+     * @return array
```
Ocramius: Can you please add description of the two returned values in the array?

Ocramius (changes requested):

AlessandroMinoccheri:
Updated with a little bit of description

Ocramius (approved):
LGTM :+1:

Issue #2814, opened 2017-08-10
Incorrect documentation example for `QueryBuilder#set()`


Dormilich:
The affected code section (also the `update()` and `where()` examples)
```
     * <code>
     *     $qb = $conn->createQueryBuilder()
     *         ->update('users', 'u')
     *         ->set('u.password', md5('password'))
     *         ->where('u.id = ?');
     * </code>
```
When looking at the source code for `set()`, the `$value` parameter is not escaped (which makes sense) but then any unescaped string value (such as the mentioned MD5 hash) would cause an SQL syntax error due to missing string delimiters. Insofar the code example is misleading as to what can be put in as value.

Maybe a better example would be using one of the SQL functions, e.g.
```
->set('u.last_login', 'NOW()')
```

lcobucci:
@Dormilich would you mind send us a PR with your suggestion?

Dormilich:
not sure if I opened the PR correctly ... it's been ages since I did that last.

Ocramius:
@Dormilich go to https://github.com/doctrine/dbal and you should see a green button that allows you to open a PR, if everything is detected correctly.

Ocramius:
Handled in #2818

PR #2818, opened 2017-08-18
#2814 removing incorrect `QueryBuilder#set()` documentation, which was showing wrong password hashing and value binding practices (combo!)

Dormilich:
Replaced the code examples for `set()` to show that the value parameter is not escaped and thus could pose an SQL injection vulnerability.

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -540,7 +540,7 @@ class QueryBuilder
      * <code>
      *     $qb = $conn->createQueryBuilder()
      *         ->update('users', 'u')
-     *         ->set('u.password', md5('password'))
+     *         ->set('u.last_login', 'NOW()')
      *         ->where('u.id = ?');
      * </code>
      *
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -540,7 +540,7 @@ public function delete($delete = null, $alias = null)
      * <code>
      *     $qb = $conn->createQueryBuilder()
      *         ->update('users', 'u')
-     *         ->set('u.password', md5('password'))
```
Ocramius: OH WOW, we had `md5('password')` in the docs?! Shame on us :-(

Ocramius (approved):

Ocramius:
LGTM :ship: 

Thanks @Dormilich!

Ocramius:
Fixes #2814

lib/Doctrine/DBAL/Query/QueryBuilder.php
Ocramius: Oh, and no parameter binding: combo shame! :-\

PR #2823, opened 2017-08-25
Simplification of a foreach

mikeSimonson:


lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -244,7 +244,8 @@ private function initializeAllDoctrineTypeMappings()
         $this->initializeDoctrineTypeMappings();
 
         foreach (Type::getTypesMap() as $typeName => $className) {
-            foreach (Type::getType($typeName)->getMappedDatabaseTypes($this) as $dbType) {
+            $type = Type::getType($typeName);
```
Ocramius: This actually makes it more complicated by introducing an additional variable.

Ocramius (changes requested):

Issue #2817, opened 2017-08-18
Comparison Problem Default 'NULL'


crash21:
I am using

    PHP                                  v7.1.7
    MariaDB                              v10.2.7
    doctrine/annotations                 v1.2.7    Docblock Annotations Parser
    doctrine/cache                       v1.6.2    Caching library offering an object-oriented API for many cache    backends
    doctrine/collections                 v1.3.0    Collections Abstraction library
    doctrine/common                      v2.6.2    Common Library for Doctrine projects
    doctrine/data-fixtures               v1.1.1    Data Fixtures for all Doctrine Object Managers
    doctrine/dbal                        v2.5.13   Database Abstraction Layer
    doctrine/doctrine-bundle             1.6.8     Symfony DoctrineBundle
    doctrine/doctrine-cache-bundle       1.3.0     Symfony Bundle for Doctrine Cache
    doctrine/doctrine-fixtures-bundle    2.3.0     Symfony DoctrineFixturesBundle
    doctrine/doctrine-migrations-bundle  v1.1.0    Symfony DoctrineMigrationsBundle

When I run command:
    
    php bin/console doctrine:schema:update --force

I get every time same queries run and of course with 

    php bin/console doctrine:schema:update --dump-sql 

I see every time the same  SQL Queries (all of them ALTER).

So I have troubleshooted it a bit and found the following issue in: 

    vendor/doctrine/dbal/lib/Doctrine/DBAL/Schema/Comparator.php:210

    $table1Columns = $table1->getColumns();
    $table2Columns = $table2->getColumns();

More precisely the problem is on :

     vendor/doctrine/dbal/lib/Doctrine/DBAL/Schema/Comparator.php:438

      if ($properties1['default'] != $properties2['default'] ||
            // Null values need to be checked additionally as they tell whether to create or drop a default value.
            // null != 0, null != false, null != '' etc. This affects platform's table alteration SQL generation.
            (null === $properties1['default'] && null !== $properties2['default']) ||
            (null === $properties2['default'] && null !== $properties1['default'])
        ) {
            $changedProperties[] = 'default';
        }

The problem is that default null is different type. In table 1 is string "NULL" while in second table is a null type. Probably here need to cast or also check if property value is "NULL".

![dbal-bug](https://user-images.githubusercontent.com/3415953/29460169-9c99bdce-842f-11e7-9ecb-2fb7fe20118d.png)

[...]

PR #2824, opened 2017-08-26
Mariadb 10.2.7 null values change in information schema, #2817

belgattitude:
Since MariaDB 10.2.7, `information_schema.columns` now stores `'null'` in `column_default` instead of `null`. MySQL 5.7 is not affected. 

This P/R should fix the two doctrine-related issues:

- https://github.com/doctrine/dbal/issues/2817 (invalid detection of column default values)
- https://github.com/doctrine/doctrine2/issues/6565 (schema update will alter all tables with columns_default='null')

Not very used to doctrine, so please check and correct me if I'm wrong somewhere.

*(see https://jira.mariadb.org/browse/MDEV-13132 and https://jira.mariadb.org/browse/MDEV-13341)*


belgattitude:
Sorry messed up with my clone (forgot to create a branch), closing and will reopen a new one in few minutes.

PR #2825, opened 2017-08-26
MariaDB 10.2 initial support

belgattitude:
Initial support for MariaDB 10.2.7+.

> Doctrine currently does not support MariaDB 10.2.7+. Schema creation/update is triggering infinite schema diffs

**Updated**: Title updated from 'MariaDB 10.2 support (will also fix version detection), version detection should be considered as new feature.
  
## Features 

- [x] 1. MariaDb version detection (strip '5.5.5-' prefix)

> To circumvent a limitation in the replication protocol, MariaDB 10+
> use a prefixed version with '5.5.5-' (for instance `5.5.5-Mariadb-10.0.8-xenial`).
> This cause issues with the current detection mechanism with both Mysqli and PDOMysql
> dbal implementations (always detect 5.5.5). While mariadb strips the prefix 
> automatically in their `libmariadb-client`, most (if not all) of php installations
> relies on `mysqlnd`, so the removal of the prefix `5.5.5-` must be made
> in userland (MysqliConnection.php and AbstractMySQLDriver.php).
> See https://jira.mariadb.org/browse/MDEV-4088 for reference.    

- [x] ~2. MariaDb 10.2 support `BLOB/TEXT` defaults values (since 10.2.1).~

**Update:** Removed from this P/R... will be made available in a subsequent P/R (due to architecture decision to be made - current doctrine implementation needs to be reworked)
      
> Note that Mysql (as of 5.7.19) is not yet supporting defaults for BLOB and TEXT.    
> See https://mariadb.com/kb/en/library/blob-and-text-data-types
  
- [x] **UPDATED** ~3. MariaDb 10.2 support `JSON` type.~       

Removed from P/R by @lcobucci, probably to reflect the fact that MariaDB alias JSON to MEDIUMTEXT
      
> ~The JSON type have been introduced in 10.2.7 as an alias to LONGTEXT.
> See https://mariadb.com/kb/en/library/json-data-type/
> To ensure portability between MySQL/Mariadb, the best is to prefer TEXT type over JSON when defining schema (till a *real* JSON type appears in Maria).~  

- [x] 4. MariaDb 10.2 supports `CURRENT_TIME`, `CURRENT_DATE` as default values

> Note that MySQL (as of 5.7.19) does *not* support default values CURRENT_TIME and
> CURRENT_DATE as default values for TIME and DATE columns (CURRENT_TIMESTAMP
> for DATETIME is supported). 
 
- [x] ~5. Default literal values must be escaped.~

[...]

PR #2826, opened 2017-08-26
Remove inconsistencies in table/column escaping for mysql

belgattitude:
Might prevent possible future bugs with reserved keywords.


Ocramius:
> future bugs with reserved keywords.

Anything we can do to assess that? A regression prevention test?

belgattitude:
Was hypothetical ;) 

Anyway quoting table and columns with '`' (which was 50% done) would give some guarantee. But not very important anyway... Feel free to close the P/R, I don't feel urgency to work on this (just added while discovering the code to fix https://github.com/doctrine/dbal/pull/2825)

Ocramius:
@belgattitude the reasoning is quite simple: I need a test that justifies a change in the code. I'd rather have this fail with new MySQL versions (with existing tests) than a change now with no evidence of failure.

belgattitude:
@ocramius thanks I understand. I'll close this one. 

 Keep the good work :)

PR #2832, opened 2017-09-02
Remove useless method

greg0ire:
The name should just be what is provided when adding a type to the type
map. If this turns out to be useful after all, then a check should
probably be added to addType.

### DIFF
--- a/lib/Doctrine/DBAL/Types/Type.php
+++ b/lib/Doctrine/DBAL/Types/Type.php
@@ -159,19 +159,10 @@ abstract class Type
     abstract public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform);
 
     /**
-     * Gets the name of this type.
-     *
-     * @return string
-     *
-     * @todo Needed?
-     */
-    abstract public function getName();
-
-    /**
      * Factory method to create type instances.
      * Type instances are implemented as flyweights.
      *
-     * @param string $name The name of the type (as returned by getName()).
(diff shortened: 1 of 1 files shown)


greg0ire:
This is a BC-break BTW. Should I first mark the method as deprecated on lower branches?

lib/Doctrine/DBAL/Types/Type.php
```diff
@@ -159,19 +159,10 @@ public function getDefaultLength(AbstractPlatform $platform)
     abstract public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform);
 
     /**
-     * Gets the name of this type.
-     *
-     * @return string
-     *
-     * @todo Needed?
-     */
-    abstract public function getName();
```
Ocramius: Please do deprecate it instead.

Also, it is not useless: consumers of `Type` objects do need to get the name eventually, although the class name is indeed sufficient once we replace custom names in mappings.

Until then, this should stay.

greg0ire: Ok, so to be clear on what I'm supposed to do:

1. change the base on 2.5
2. add an `@deprecated` annotation
3. implement the method like this:
```php
<?php
@trigger_error(
    'This method is deprecated, please use the class name directly instead',
    E_USER_DEPRECATED
);
```

Ocramius:
Discussed this on the Symfony Slack directly: not worth doing for now, unless an alternative type mapping system is available.

PR #2834, opened 2017-09-02
WIP: Rework type system

greg0ire:
From @Majkl578 on slack:

just a quick thought of a possible approach:
1) introduce TypeInterface with required methods (or different name)
2) make types implement it
3) introduce some kind of Types enum-like empty class for implictly supported types, just like i.e. GeneratorType in ORM develop (edit: maybe this could stay in Type to lower BC implications and make Type an enum class in 3.0)
4) introduce non-static TypeRegistry with no default types hardcoded in it - this would allow different connections/platforms use different type mappings (as @ocramius suggested in https://symfony-devs.slack.com/archives/C3FQPE6LE/p1504362019000073)
5) make TypeRegistry work with type instances instead of their class names (sounds good, but unsure about consequences)
6) migrate Type::<typenames> to Types enum, keep BC by aliasing them and deprecate in 2,7
7) migrate Type::<static-registry> mechanics to TypeRegistry, keep BC by using private static TypeRegistry internally in 2.7 (some kittens may die during this process)
8) in 3.0 (develop) drop these BC hacks (so no more kittens die)

### DIFF
--- a/lib/Doctrine/DBAL/Types/Type.php
+++ b/lib/Doctrine/DBAL/Types/Type.php
@@ -31,7 +31,7 @@ use Doctrine\DBAL\DBALException;
  * @author Benjamin Eberlei <kontakt@beberlei.de>
  * @since  2.0
  */
-abstract class Type
+abstract class Type implements TypeInterface
 {
     const TARRAY = 'array';
     const SIMPLE_ARRAY = 'simple_array';
@@ -101,13 +101,7 @@ abstract class Type
 
 
     /**
-     * Converts a value from its PHP representation to its database representation
-     * of this type.
-     *
-     * @param mixed                                     $value    The value to convert.
-     * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform The currently used database platform.
-     *
-     * @return mixed The database representation of the value.
+     * {@inheritdoc}
      */
     public function convertToDatabaseValue($value, AbstractPlatform $platform)
     {
@@ -115,13 +109,7 @@ abstract class Type
     }
 
     /**
-     * Converts a value from its database representation to its PHP representation
-     * of this type.
-     *
-     * @param mixed                                     $value    The value to convert.
-     * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform The currently used database platform.
-     *
-     * @return mixed The PHP representation of the value.
+     * {@inheritdoc}
      */
     public function convertToPHPValue($value, AbstractPlatform $platform)
     {
@@ -143,12 +131,7 @@ abstract class Type
     }
 
     /**
-     * Gets the SQL declaration snippet for a field of this type.
-     *
-     * @param array                                     $fieldDeclaration The field declaration.
-     * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform         The currently used database platform.
-     *
-     * @return string
+     * {@inheritdoc}
      */
     abstract public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform);
 
@@ -258,18 +241,7 @@ abstract class Type
     }
 
     /**
-     * Gets the (preferred) binding type for values of this type that
-     * can be used when binding parameters to prepared statements.
-     *
-     * This method should return one of the PDO::PARAM_* constants, that is, one of:
-     *
-     * PDO::PARAM_BOOL
-     * PDO::PARAM_NULL
-     * PDO::PARAM_INT
-     * PDO::PARAM_STR
-     * PDO::PARAM_LOB
-     *
-     * @return integer
+     * {@inheritdoc}
      */
     public function getBindingType()
     {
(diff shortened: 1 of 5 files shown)


lib/Doctrine/DBAL/Types/Type.php
```diff
@@ -201,6 +215,11 @@ public static function getType($name)
      */
     public static function addType($name, $className)
     {
+        self::warnAgainstUsingTheOldAPI();
+        if ($name === $className) {
+            TypeRegistry::addType($className);
```
greg0ire: @Majkl578 that kind of thing can't be done if `TypeRegistry` is not static, unless maybe we add a private static variable to hold the registry?

lib/Doctrine/DBAL/Types/TypeRegistry.php
```diff
@@ -0,0 +1,127 @@
...
+     *
+     * @throws \Doctrine\DBAL\DBALException
+     */
+    public static function getType(string $class): Doctrine\DBAL\Types\Type
+    {
+        if (!self::hasType($class)) {
+            throw DBALException::unknownColumnType($class);
+        }
+        if (!isset(self::$_typeObjects[$class])) {
+            self::$_typeObjects[$class] = new $class();
```
greg0ire: I get an error here because the constructor of the abstract class is private, but I don't understand how it works right now, because the old implementation does something similar.

lib/Doctrine/DBAL/Types/Type.php
Majkl578: Yes, in 2.x it'd have to be done that way.
Then in 3.0, it should probably be configured via Configuration object and use TypeRegistry instance instead of static calls to Type across DBAL/ORM.

lib/Doctrine/DBAL/Types/TypeRegistry.php
Majkl578: Currently, Type is both registry AND the type itself (a sub-class). So in order to create instance of a type, it instantiates its subclass, and since constructor is private in Type, it _just works_. Would require public constructor.

lib/Doctrine/DBAL/Types/Type.php
greg0ire: Ok will change to non-static.

lib/Doctrine/DBAL/Types/TypeRegistry.php
greg0ire: 💡 ok makes sense

lib/Doctrine/DBAL/Types/TypeInterface.php
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * The registry for so-called Doctrine mapping types.
+ *
+ * A Type object is obtained by calling the static {@link getType()} method.
+ *
+ * @author Roman Borschel <roman@code-factory.org>
+ * @author Benjamin Eberlei <kontakt@beberlei.de>
+ * @since  2.0
+ */
+interface TypeInterface
```
greg0ire: Should I find another name? IIRC you don't use the `Interface` suffix, do you?

lib/Doctrine/DBAL/Types/TypeInterface.php
```diff
@@ -0,0 +1,135 @@
...
+     *
+     * PDO::PARAM_BOOL
+     * PDO::PARAM_NULL
+     * PDO::PARAM_INT
+     * PDO::PARAM_STR
+     * PDO::PARAM_LOB
+     *
+     * @return integer
+     */
+    public function getBindingType();
```
greg0ire: Could be moved to another interface

lib/Doctrine/DBAL/Types/TypeInterface.php
```diff
@@ -0,0 +1,135 @@
...
+     * Does working with this column require SQL conversion functions?
+     *
+     * This is a metadata function that is required for example in the ORM.
+     * Usage of {@link convertToDatabaseValueSQL} and
+     * {@link convertToPHPValueSQL} works for any type and mostly
+     * does nothing. This method can additionally be used for optimization purposes.
+     *
+     * @return boolean
+     */
+    public function canRequireSQLConversion();
```
greg0ire: Could be replaced with another interface that would include the 2 following methods

lib/Doctrine/DBAL/Types/TypeInterface.php
```diff
@@ -0,0 +1,135 @@
...
+     * If this Doctrine Type maps to an already mapped database type,
+     * reverse schema engineering can't tell them apart. You need to mark
+     * one of those types as commented, which will have Doctrine use an SQL
+     * comment to typehint the actual Doctrine Type.
+     *
+     * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform
+     *
+     * @return boolean
+     */
+    public function requiresSQLCommentHint(AbstractPlatform $platform);
```
greg0ire: Can be replaced with a marker interface

lib/Doctrine/DBAL/Types/TypeInterface.php
```diff
@@ -0,0 +1,135 @@
...
+    public function convertToPHPValueSQL($sqlExpr, $platform);
+
+    /**
+     * Gets an array of database types that map to this Doctrine type.
+     *
+     * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform
+     *
+     * @return array
+     */
+    public function getMappedDatabaseTypes(AbstractPlatform $platform);
```
greg0ire: Not sure when it is necessary / desirable to implement this one, I found only one implementation : https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/Types/MySqlPointType.php#L20

greg0ire:
This is too big, let's create an issue for tracking + several PRs instead.

PR #2835, opened 2017-09-02
Stop relying on Type::__toString

greg0ire:
It is a somewhat fragile thing to do.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -19,29 +19,29 @@
 
 namespace Doctrine\DBAL\Platforms;
 
-use Doctrine\DBAL\DBALException;
+use Doctrine\Common\EventManager;
 use Doctrine\DBAL\Connection;
-use Doctrine\DBAL\Schema\Identifier;
-use Doctrine\DBAL\Types;
-use Doctrine\DBAL\Schema\Constraint;
-use Doctrine\DBAL\Schema\Sequence;
-use Doctrine\DBAL\Schema\Table;
-use Doctrine\DBAL\Schema\Index;
-use Doctrine\DBAL\Schema\ForeignKeyConstraint;
-use Doctrine\DBAL\Schema\TableDiff;
+use Doctrine\DBAL\DBALException;
+use Doctrine\DBAL\Event\SchemaAlterTableAddColumnEventArgs;
+use Doctrine\DBAL\Event\SchemaAlterTableChangeColumnEventArgs;
+use Doctrine\DBAL\Event\SchemaAlterTableEventArgs;
+use Doctrine\DBAL\Event\SchemaAlterTableRemoveColumnEventArgs;
+use Doctrine\DBAL\Event\SchemaAlterTableRenameColumnEventArgs;
+use Doctrine\DBAL\Event\SchemaCreateTableColumnEventArgs;
+use Doctrine\DBAL\Event\SchemaCreateTableEventArgs;
+use Doctrine\DBAL\Event\SchemaDropTableEventArgs;
+use Doctrine\DBAL\Events;
 use Doctrine\DBAL\Schema\Column;
 use Doctrine\DBAL\Schema\ColumnDiff;
+use Doctrine\DBAL\Schema\Constraint;
+use Doctrine\DBAL\Schema\ForeignKeyConstraint;
+use Doctrine\DBAL\Schema\Identifier;
+use Doctrine\DBAL\Schema\Index;
+use Doctrine\DBAL\Schema\Sequence;
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\Schema\TableDiff;
+use Doctrine\DBAL\Types;
 use Doctrine\DBAL\Types\Type;
-use Doctrine\DBAL\Events;
-use Doctrine\Common\EventManager;
-use Doctrine\DBAL\Event\SchemaCreateTableEventArgs;
-use Doctrine\DBAL\Event\SchemaCreateTableColumnEventArgs;
-use Doctrine\DBAL\Event\SchemaDropTableEventArgs;
-use Doctrine\DBAL\Event\SchemaAlterTableEventArgs;
-use Doctrine\DBAL\Event\SchemaAlterTableAddColumnEventArgs;
-use Doctrine\DBAL\Event\SchemaAlterTableRemoveColumnEventArgs;
-use Doctrine\DBAL\Event\SchemaAlterTableChangeColumnEventArgs;
-use Doctrine\DBAL\Event\SchemaAlterTableRenameColumnEventArgs;
 
 /**
  * Base class for all DatabasePlatforms. The DatabasePlatforms are the central
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -1556,7 +1556,7 @@ abstract class AbstractPlatform
             $columnData['version'] = $column->hasPlatformOption("version") ? $column->getPlatformOption('version') : false;
             $columnData['comment'] = $this->getColumnComment($column);
 
-            if (strtolower($columnData['type']) == "string" && $columnData['length'] === null) {
+            if ($columnData['type'] instanceof Types\StringType && $columnData['length'] === null) {
                 $columnData['length'] = 255;
             }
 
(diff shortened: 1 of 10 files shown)


lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -19,29 +19,29 @@
...
-use Doctrine\DBAL\Events;
-use Doctrine\Common\EventManager;
-use Doctrine\DBAL\Event\SchemaCreateTableEventArgs;
-use Doctrine\DBAL\Event\SchemaCreateTableColumnEventArgs;
-use Doctrine\DBAL\Event\SchemaDropTableEventArgs;
-use Doctrine\DBAL\Event\SchemaAlterTableEventArgs;
-use Doctrine\DBAL\Event\SchemaAlterTableAddColumnEventArgs;
-use Doctrine\DBAL\Event\SchemaAlterTableRemoveColumnEventArgs;
-use Doctrine\DBAL\Event\SchemaAlterTableChangeColumnEventArgs;
-use Doctrine\DBAL\Event\SchemaAlterTableRenameColumnEventArgs;
```
greg0ire: I reordered the use statements. Is it good or did the previous order have some meaning I should keep?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2279,15 +2279,20 @@ public function getDefaultValueDeclarationSQL($field)
         if (isset($field['default'])) {
             $default = " DEFAULT '".$field['default']."'";
             if (isset($field['type'])) {
-                if (in_array((string) $field['type'], ["Integer", "BigInt", "SmallInt"])) {
+                $type = $field['type'];
+                if ($type instanceof Types\IntegerType ||
+                    $type instanceof Types\BigIntType ||
+                    $type instanceof Types\SmallIntType
```
greg0ire: Should I introduce a parent class? This test is done several time ( it done in SQLServerPlatform, for instance)

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2279,15 +2279,20 @@ public function getDefaultValueDeclarationSQL($field)
...
+                ) {
                     $default = " DEFAULT ".$field['default'];
-                } elseif (in_array((string) $field['type'], ['DateTime', 'DateTimeTz']) && $field['default'] == $this->getCurrentTimestampSQL()) {
+                } elseif (($type instanceof Types\DateTimeType || $type instanceof Types\DateTimeTzType)
```
greg0ire: Same here

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
greg0ire: Or maybe a marker interface?

SenseException: Do you mean something like an AbstractNumberType? It wouldn't introduce number-specific functionality and just be an empty class, if you meant such a solution. A marker interface would have no methods, right? I'm not sure about new class or interface files, that are only introduced for checks. Do those have a purpose in userland for custom types or would those be only used internally?

I also considered (static?) methods for internal type checking, but those had to be maintained too and aren't elegant either. It's like we can only introduce more complexity to Doctrine.

greg0ire: > Do you mean something like an AbstractNumberType? It wouldn't introduce number-specific functionality and just be an empty class, if you meant such a solution. 

Correct, but indeed, an interface is better

> A marker interface would have no methods, right?

yup

> I'm not sure about new class or interface files, that are only introduced for checks. Do those have a purpose in userland for custom types or would those be only used internally?

People that introduce numeric types and would like to benefit from the code behind those checks could implement that interface, so yeah, there is a purpose in userland.

> I also considered (static?) methods for internal type checking, but those had to be maintained too and aren't elegant either. It's like we can only introduce more complexity to Doctrine.

Yeah, and I think a marker interface is really not a big thing to make public.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
lcobucci: No worries on reordering it, I usually do it on a separated commit to make it easier for people to review stuff 😄

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
lcobucci: 👍 for the marker interface. I'd suggest to mark it as internal to tell people to simply no use it (`@internal`).

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
lcobucci: Indeed, and we probably should have added the immutable type here as well

lcobucci (changes requested):
Let's move the build fixes out of this PR (I'm working on that) and simplify things using a marker interface.

greg0ire:
The build fix was just to check I didn't actually break things. I will remove them when the PR looks like it's mergeable.

greg0ire:
Also, please see #2838 for a quick fix of the build until you find a better solution if there is one.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
greg0ire: Yeah, I got lazy there, I'll do that.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
greg0ire: @lcobucci the immutable type extends the mutable type, so it's already ok :)

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2279,15 +2279,17 @@ public function getDefaultValueDeclarationSQL($field)
...
+                } elseif ($type instanceof Types\PhpDateTimeMappingType
+                    && $field['default'] == $this->getCurrentTimestampSQL()) {
                     $default = " DEFAULT ".$this->getCurrentTimestampSQL();
-                } elseif ((string) $field['type'] == 'Time' && $field['default'] == $this->getCurrentTimeSQL()) {
+                } elseif ($type instanceof Types\TimeType && $field['default'] == $this->getCurrentTimeSQL()) {
                     $default = " DEFAULT ".$this->getCurrentTimeSQL();
-                } elseif ((string) $field['type'] == 'Date' && $field['default'] == $this->getCurrentDateSQL()) {
-                    $default = " DEFAULT ".$this->getCurrentDateSQL();
-                } elseif ((string) $field['type'] == 'Boolean') {
+                } elseif ($type instanceof Types\DateType && $field['default'] == $this->getCurrentDateSQL()) {
```
greg0ire: Maybe I should name the marker interface differently because of `Types\DateType` does not implement it although it does map to a PHP `\DateTimeInterface` instance? Suggestions?

greg0ire:
@lcobucci please review again. I kept the build-fixing commit to ensure I did not break anything.

greg0ire:
> Let's move the build fixes out of this PR (I'm working on that)

How about merging #2838 right now and do it your way later? That would make things easier for all the outstanding PRs...

Majkl578 (changes requested):
Needs some tests, as you can see, Travis is green even though there's a fatal error. :/

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2279,15 +2279,17 @@ public function getDefaultValueDeclarationSQL($field)
...
+                    $default = " DEFAULT ".$this>getCurrentDateSQL();
```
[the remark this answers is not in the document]
greg0ire: woops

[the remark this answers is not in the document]
SenseException: Really good eyes here. I didn't saw that too. 👍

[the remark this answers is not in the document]
greg0ire: I got very unlucky, that's the only if that was not covered.

greg0ire:
I created #2851 to proove the test work with the previous code. Please review and merge it, and I shall rebase this.

greg0ire:
Oh snap I need to add #2838 to it if I want to prove anything

Majkl578:
Related issue in ORM: https://github.com/doctrine/doctrine2/issues/5760

greg0ire:
Pushed again without the test commit, which was just merged. Please review again @Majkl578 @lcobucci

Ocramius (approved):
LGTM 👍

Ocramius:
:ship:

greg0ire:
Please don't review again then :P

lib/Doctrine/DBAL/Types/PhpIntegerMappingType.php
```diff
@@ -0,0 +1,30 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Types;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
```
SenseException: What was the point for this `use` statement?

lib/Doctrine/DBAL/Types/PhpDateTimeMappingType.php
```diff
@@ -0,0 +1,30 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Types;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
```
SenseException: What was the point for this `use` statement?

lib/Doctrine/DBAL/Types/PhpIntegerMappingType.php
greg0ire: None I'm afraid :sweat:

Issue #2833, opened 2017-09-02
Enable SQLite foreign key on post-connect


Ocramius:
Apparently, SQLite has supported foreign keys for ages! All we need to do is (according to https://sqlite.org/foreignkeys.html) run a statement on post-connect:

```sql
PRAGMA foreign_keys = ON;
```

Not sure where to do it exactly, but this would be a very welcome addition.

Federkun:
> Not sure where to do it exactly

I've an event subscriber to enable that: would you suggest to add an `SqliteSessionInit.php`?

Ocramius:
@Federkun yup :+1:

drhsqlite:
SQLite has a compile-time option to enable foreign_key enforcement by default.  https://www.sqlite.org/compile.html#default_foreign_keys

The default build of SQLite has foreign_key enforcement turned off by default due to backwards-compatibility concerns.

Ocramius:
Compile-time is not in our ballpark: we are a lib sitting on top of
pre-compiled php binaries.

DontShootMe:
Hello,
Just a little up about this functionality.

Have you found any good approach, because foreign keys are missing for ages ?

Many thanks all, for all your work
You are all amazing

alexander-schranz:
I did stumble over https://github.com/doctrine/dbal/pull/5404 and asking myself if FK is now supported via dbal?

derrabus:
Yes. I think, we can close this issue.

Issue #1204, opened 2015-06-26
DBAL-1253: Sqlite: inconsistent (non-)support of foreign keys


doctrinebot:
Jira issue originally created by user adirelle:

Though the SqlitePlatform announces it doesn't support foreign keys, it generates foreign key constraints in "alter table" statements.

This leads to issues when combined with Doctrine ORM and/or Doctrine Migrations : the shcema manager doesn't create foreign keys on new tables, but it generates SQL statement to add them on existing tables.
- Calling the schema update command several times in a row on an existing database causes all tables with relations to be recreated (since Sqlite ALTER TABLE statement is very limited).
- Even if the database is up to date, using the "diff" command of Doctrine Migrations creates a migration the recreate all tables.

I suggest to disable the declaration of foreign keys at all to avoid these discrepancies.

Background:
- Before Sqlite 3.6.19, declarations of foreign keys in CREATE TABLE were accepted (and retrieved) but ignored.
- Starting with 3.6.19, there is a compile option to disable the support of the foreign key syntax in CREATE TABLE statement. There is another compile option to disable the triggers, including foreign key ones. These options can be queried using a PRAGMA statement.
- Starting with 3.6.19, there is a PRAGMA to enforce the respect of foreign keys.
- in any case, foreign keys are not supported in ALTER TABLE statements.

webmozart:
see also #2104

naitsirch:
I found this issue, because DBAL still does not fetch foreign keys from the database in the schema tool. This causes the schema update tool always to generate the whole table creation statements making it useless. The reason is `SqlitePlatform::supportsForeignKeyConstraints()` always returns `false`. What is the reason?

After changing this to `true` (for testing) the diff tool correctly says that there are no differences anymore.

SenseException:
> What is the reason for that?

Simply because Sqlite didn't always supported FK from the start. There are PRs like #821 that were created to introduce them to DBAL's SqlitePlatform.

naitsirch:
Thanks for referencing the PR ;-)

beberlei:
This is the cause for a regression in ORM that tried to fix this on SchemaTool side github.com/doctrine/orm/issue/7841 and github.com/doctrine/orm/pr/7842

morozov:
Fixed by https://github.com/doctrine/dbal/pull/3762.

morozov:
Reopening due to #4255.

morozov:
[...]

PR #2836, opened 2017-09-02
Add sqlite's event subscribe to enable foreign key constraints support

Federkun:
See https://github.com/doctrine/dbal/issues/2833

### DIFF
--- a/lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
+++ b/lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
@@ -10,7 +10,7 @@ use Doctrine\DBAL\Platforms\SqlitePlatform;
 /**
  * Sqlite session init event subscriber enable foreign key constraints.
  */
-class SqliteSessionInit implements EventSubscriber
+final class SqliteSessionInit implements EventSubscriber
 {
     public function postConnect(ConnectionEventArgs $args) : void
     {
--- a/lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
+++ b/lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
@@ -2,19 +2,16 @@
 
 namespace Doctrine\DBAL\Event\Listeners;
 
+use Doctrine\Common\EventSubscriber;
 use Doctrine\DBAL\Event\ConnectionEventArgs;
 use Doctrine\DBAL\Events;
-use Doctrine\Common\EventSubscriber;
 
 /**
  * Sqlite session init event subscriber enable foreign key constraints.
  */
 class SqliteSessionInit implements EventSubscriber
 {
-    /**
-     * @param \Doctrine\DBAL\Event\ConnectionEventArgs $args
-     */
-    public function postConnect(ConnectionEventArgs $args): void
+    public function postConnect(ConnectionEventArgs $args) : void
     {
         $args->getConnection()->exec('PRAGMA foreign_keys = on');
     }
@@ -22,7 +19,7 @@ class SqliteSessionInit implements EventSubscriber
     /**
      * @return string[]
      */
-    public function getSubscribedEvents(): array
+    public function getSubscribedEvents() : array
     {
         return [Events::postConnect];
     }
--- a/lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
+++ b/lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
@@ -1,0 +1,29 @@
+<?php declare(strict_types=1);
+
+namespace Doctrine\DBAL\Event\Listeners;
+
+use Doctrine\DBAL\Event\ConnectionEventArgs;
+use Doctrine\DBAL\Events;
+use Doctrine\Common\EventSubscriber;
+
+/**
+ * Sqlite session init event subscriber enable foreign key constraints.
+ */
+class SqliteSessionInit implements EventSubscriber
+{
+    /**
+     * @param \Doctrine\DBAL\Event\ConnectionEventArgs $args
+     */
+    public function postConnect(ConnectionEventArgs $args): void
+    {
+        $args->getConnection()->exec('PRAGMA foreign_keys = on');
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getSubscribedEvents(): array
+    {
+        return [Events::postConnect];
+    }
+}
--- a/lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
+++ b/lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
@@ -5,6 +5,7 @@ namespace Doctrine\DBAL\Event\Listeners;
 use Doctrine\Common\EventSubscriber;
 use Doctrine\DBAL\Event\ConnectionEventArgs;
 use Doctrine\DBAL\Events;
+use Doctrine\DBAL\Platforms\SqlitePlatform;
 
 /**
  * Sqlite session init event subscriber enable foreign key constraints.
@@ -14,6 +15,12 @@ class SqliteSessionInit implements EventSubscriber
     public function postConnect(ConnectionEventArgs $args) : void
     {
         $args->getConnection()->exec('PRAGMA foreign_keys = on');
+
+        if (! ($args->getDatabasePlatform() instanceof SqlitePlatform)) {
+            return;
+        }
+
+        $args->getDatabasePlatform()->enableForeignKeyConstraintsSupport();
     }
 
     /**
--- a/lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
+++ b/lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
@@ -8,7 +8,7 @@ use Doctrine\DBAL\Events;
 use Doctrine\DBAL\Platforms\SqlitePlatform;
 
 /**
- * Sqlite session init event subscriber enable foreign key constraints.
+ * Sqlite session init event subscriber enables foreign key constraints.
  */
 final class SqliteSessionInit implements EventSubscriber
 {
--- a/tests/Doctrine/Tests/DBAL/Events/SqliteSessionInitTest.php
+++ b/tests/Doctrine/Tests/DBAL/Events/SqliteSessionInitTest.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Events;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Event\ConnectionEventArgs;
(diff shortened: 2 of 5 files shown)


lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
```diff
@@ -0,0 +1,49 @@
...
+{
+    /**
+     * @param \Doctrine\DBAL\Event\ConnectionEventArgs $args
+     *
+     * @return void
+     */
+    public function postConnect(ConnectionEventArgs $args)
+    {
+        $conn = $args->getConnection();
+        $conn->exec('PRAGMA foreign_keys = on');
```
Federkun: "If the command "PRAGMA foreign_keys" returns no data instead of a single row containing "0" or "1", then the version of SQLite you are using does not support foreign keys" - should we throw an exception in this case?

Majkl578:
https://sqlite.org/foreignkeys.html
SQLite supports foreign keys as of 3.6.19 released on 14/10/2009. I think it's safe to assume that users have newer version already.
Not sure about BC implications, this could be a BC break, right?

Federkun:
You are required to explicitly add the event subscribe, so I don't think it's a bc break.
What if this event handler is added but mysql or postgresql is used instead of sqlite? Should be ignored in this case?

SenseException:
That would be a misconfiguration of Doctrine because of the different platforms. Should it throw an exception then?

lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
SenseException: As Doctrine is going forward with versions like PHP, shouldn't old database versions be dropped too? As @Majkl578 showed us in his link, the SQLite 3.6 releases are pretty old.

Majkl578:
> You are required to explicitly add the event subscribe, so I don't think it's a bc break.

Ah yeah, of course, if it's opt-in, it's fine. (Original issue made me think it was supposed to be default behavior.)

---

None of the [existing SessionInit listeners](https://github.com/doctrine/dbal/tree/master/lib/Doctrine/DBAL/Event/Listeners) throws so I'd leave it as is without platform check.

greg0ire:
Regarding the broken build, see #2838 . Closing/reopening should be enough to make this green once it is merged.

lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
Ocramius: Only in major versions, as we have no way to constraint against the DB version via automation. We could do that via `composer`, but not here.

lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
```diff
@@ -0,0 +1,49 @@
...
+ * Sqlite session init event subscriber enable foreign key constraints.
+ */
+class SqliteSessionInit implements EventSubscriber
+{
+    /**
+     * @param \Doctrine\DBAL\Event\ConnectionEventArgs $args
+     *
+     * @return void
+     */
+    public function postConnect(ConnectionEventArgs $args)
```
Majkl578: Please add ` : void` and remove `@return`.

lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        $conn = $args->getConnection();
+        $conn->exec('PRAGMA foreign_keys = on');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSubscribedEvents()
```
Majkl578: Please add ` : array` and add `@return string[]`.

tests/Doctrine/Tests/DBAL/Events/SqliteSessionInitTest.php
```diff
@@ -0,0 +1,30 @@
...
+namespace Doctrine\Tests\DBAL\Events;
+
+use Doctrine\DBAL\Event\ConnectionEventArgs;
+use Doctrine\DBAL\Event\Listeners\SqliteSessionInit;
+use Doctrine\DBAL\Events;
+use Doctrine\Tests\DbalTestCase;
+
+class SqliteSessionInitTest extends DbalTestCase
+{
+    public function testPostConnect()
```
Majkl578: Please add ` : void`.

tests/Doctrine/Tests/DBAL/Events/SqliteSessionInitTest.php
```diff
@@ -0,0 +1,30 @@
...
+                       ->method('exec')
+                       ->with($this->equalTo('PRAGMA foreign_keys = on'));
+
+        $eventArgs = new ConnectionEventArgs($connectionMock);
+
+        $listener = new SqliteSessionInit();
+        $listener->postConnect($eventArgs);
+    }
+
+    public function testGetSubscribedEvents()
```
Majkl578: Please add ` : void`.

lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
```diff
@@ -0,0 +1,49 @@
+<?php
+/*
```
Majkl578: Please remove license header, we're going to drop it everywhere (already done in develop in #2853).

lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
Majkl578: Exceptions should be fine even for minor, as long as this behavior stays opt-in.

tests/Doctrine/Tests/DBAL/Events/SqliteSessionInitTest.php
```diff
@@ -0,0 +1,30 @@
...
+    {
+        $listener = new SqliteSessionInit();
+        $this->assertEquals(array(Events::postConnect), $listener->getSubscribedEvents());
```
Majkl578: Please use `self::` for assertion.

Majkl578 (changes requested):
Just some small CS nitpicking based on recent changes (headers, types).

Federkun:
Hi @Majkl578, does that works for you?

Ocramius (changes requested):
One addition still needed here: an integration test that creates an in-memory sqlite db, adds a FK, then asserts that a FK violation is raised on an invalid insert.

tests/Doctrine/Tests/DBAL/Functional/ExceptionTest.php
```diff
@@ -52,7 +52,13 @@ public function testTableExistsException()
 
     public function testForeignKeyConstraintViolationExceptionOnInsert()
     {
-        if ( ! $this->_conn->getDatabasePlatform()->supportsForeignKeyConstraints()) {
+        if ($this->_conn->getDatabasePlatform()->getName() == 'sqlite') {
```
Federkun: Not sure if this is acceptable as an integration test

Ocramius (changes requested):

tests/Doctrine/Tests/DBAL/Events/SqliteSessionInitTest.php
```diff
@@ -0,0 +1,30 @@
...
+use Doctrine\Tests\DbalTestCase;
+
+class SqliteSessionInitTest extends DbalTestCase
+{
+    public function testPostConnect(): void
+    {
+        $connectionMock = $this->createMock('Doctrine\DBAL\Connection');
+        $connectionMock->expects($this->once())
+                       ->method('exec')
+                       ->with($this->equalTo('PRAGMA foreign_keys = on'));
```
SenseException: You can omit `$this->equalTo()` in `with()`. If you just use `'PRAGMA foreign_keys = on'` in `with()`, PHPUnit will handle it with an `IsEqual` constraint anyway.

Gemorroj:
any updates?

vasilvestre:
What's left to do ?

SenseException:
@Federkun Are you going to work further on your PR?

Federkun:
Yeah. I added the test.
Any idea on how to manage https://github.com/doctrine/dbal/blob/62c2a8b90b58359b6dc0d7be84721f9e83caf3f4/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php#L767
as it's an assertion that can be computed at runtime ( and only after the connection to sqlite)?

Federkun:
> Any idea on how to manage
> 
> https://github.com/doctrine/dbal/blob/62c2a8b90b58359b6dc0d7be84721f9e83caf3f4/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php#L767
> 
> 
> as it's an assertion that can be computed at runtime ( and only after the connection to sqlite)?

The only idea I can think of at the moment is to toggle the support of the platform after the connection has been made: https://github.com/doctrine/dbal/pull/2836/commits/91783c02d4565a9ae0b6c59a388ec3360fd7709a

But it's not really intuitive: platform's support for FK should not change in the middle of a process. Also, it's quite easy to check the flag before making a connection.

Suggestions are welcome.

lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
```diff
@@ -0,0 +1,33 @@
...
+
+use Doctrine\Common\EventSubscriber;
+use Doctrine\DBAL\Event\ConnectionEventArgs;
+use Doctrine\DBAL\Events;
+use Doctrine\DBAL\Platforms\SqlitePlatform;
+
+/**
+ * Sqlite session init event subscriber enable foreign key constraints.
+ */
+class SqliteSessionInit implements EventSubscriber
```
greg0ire:
```suggestion
final class SqliteSessionInit implements EventSubscriber
```

lib/Doctrine/DBAL/Event/Listeners/SqliteSessionInit.php
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Doctrine\DBAL\Event\Listeners;
+
+use Doctrine\Common\EventSubscriber;
+use Doctrine\DBAL\Event\ConnectionEventArgs;
+use Doctrine\DBAL\Events;
+use Doctrine\DBAL\Platforms\SqlitePlatform;
+
+/**
+ * Sqlite session init event subscriber enable foreign key constraints.
```
greg0ire:
```suggestion
 * Sqlite session init event subscriber enables foreign key constraints.
```

Federkun:
I'll need to check how this behave in combination with the ORM and migration, see https://github.com/doctrine/dbal/issues/1204

Gandhi11:
Any news on this PR?

PR #2838, opened 2017-09-02
Create schemas and permissions in more situations

### LINKED ISSUE #2837: Travis failures under Trusty
[Trusty is now default](https://blog.travis-ci.com/2017-08-31-trusty-as-default-status) and some mysql-realted jobs are failing there.
Possibly check #2825, it already contains some sort of fix/hack to make it work.
While at it, we should also migrate to Build Stages.

greg0ire:
On trusty, this is needed for all mysql/mariadb builds.
Fixes #2837

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: php
 sudo: false
+dist: trusty
 
 cache:
     directories:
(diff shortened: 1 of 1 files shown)


Majkl578 (approved):
LGTM, we don't really want non-determinism on CI.
Good enough until we migrate DBAL to Build Stages.

greg0ire:
Restarting the build because I've had a new issue locally even with that change

greg0ire:
Ok still green :)

lcobucci:
@greg0ire @Majkl578 I've handled the this using stages on #2849 (with some other improvements).

greg0ire:
I know, but your PR is big and still red, so not mergeable right now. I just noticed that Travis is not required for the build to pass, which is not great, because people might assume they didn't break the build when they actually did. Right now they can't know for sure unless they check the output of the logs on each commit.

Majkl578:
Handled in #2849, sorry.

greg0ire:
Not a big deal :)

PR #2839, opened 2017-09-03
Introduce WellKnownTextConverter interface

greg0ire:
It removes the need for canRequireSqlConversion.
If you are ok with this, I will write a follow up PR on the ORM that will:

1. trigger a deprecation if any of the `canRequireSQLConversion`, `convertToDatabaseValueSQL`, `convertToPHPValueSQL`. That would probably mean using reflection https://stackoverflow.com/a/17663347/353612
2. ignore said methods if the type implements the new API

### DIFF
--- a/docs/en/reference/types.rst
+++ b/docs/en/reference/types.rst
@@ -929,29 +929,25 @@ Now we implement our ``Doctrine\DBAL\Types\Type`` instance:
 
 The job of Doctrine-DBAL is to transform your type into an SQL
 declaration. You can modify the SQL declaration Doctrine will produce.
-At first, to enable this feature, you must override the
-``canRequireSQLConversion`` method:
+At first, to enable this feature, you must implement  the
+``Doctrine\DBAL\Types\WellKnownTextConverter`` interface:
 
 ::
 
     <?php
-    public function canRequireSQLConversion()
-    {
-        return true;
-    }
+    final class MyType implements WellKnownTextConverter
 
-Then you override the ``convertToPhpValueSQL`` and
-``convertToDatabaseValueSQL`` methods :
+The interface consists of two methods, which we implement like this in our case:
 
 ::
 
     <?php
-    public function convertToPHPValueSQL($sqlExpr, $platform)
+    public function convertToWellKnownTextSQL($sqlExpr, $platform)
     {
         return 'MyMoneyFunction(\''.$sqlExpr.'\') ';
     }
 
-    public function convertToDatabaseValueSQL($sqlExpr, AbstractPlatform $platform)
+    public function convertFromWellKnownTextSQL($sqlExpr, AbstractPlatform $platform)
     {
         return 'MyFunction('.$sqlExpr.')';
     }
--- a/lib/Doctrine/DBAL/Types/Type.php
+++ b/lib/Doctrine/DBAL/Types/Type.php
@@ -292,6 +292,7 @@ abstract class Type
      * does nothing. This method can additionally be used for optimization purposes.
      *
      * @return boolean
+     * @deprecated implement WellKnownTextConverter instead
      */
     public function canRequireSQLConversion()
     {
@@ -305,6 +306,7 @@ abstract class Type
      * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform
      *
      * @return string
+     * @deprecated implement WellKnownTextConverter::convertFromWellKnownTextSQL instead
      */
     public function convertToDatabaseValueSQL($sqlExpr, AbstractPlatform $platform)
     {
@@ -318,6 +320,7 @@ abstract class Type
      * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform
      *
      * @return string
+     * @deprecated implement WellKnownTextConverter::convertToWellKnownTextSQL instead
      */
     public function convertToPHPValueSQL($sqlExpr, $platform)
     {
--- a/lib/Doctrine/DBAL/Types/WellKnownTextConverter.php
+++ b/lib/Doctrine/DBAL/Types/WellKnownTextConverter.php
@@ -1,0 +1,55 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Types;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\DBALException;
+
+/**
+ * Does working with this column require SQL conversion functions?
+ *
+ * This is a metadata function that is required for example in the ORM.
+ * Usage of {@link convertToDatabaseValueSQL} and
+ * {@link convertToPHPValueSQL} works for any type and mostly
+ * does nothing. This method can additionally be used for optimization purposes.
+ *
+ */
+interface WellKnownTextConverter
+{
+    /**
+     * Modifies the SQL expression (identifier, parameter) to convert to a database value.
+     *
+     * @param string                                    $sqlExpr
+     * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform
+     *
+     * @return string
+     */
+    public function convertFromWellKnownTextSQL(string $sqlExpr, AbstractPlatform $platform) : string;
+
+    /**
+     * Modifies the SQL expression (identifier, parameter) to convert to a PHP value.
+     *
+     * @param string                                    $sqlExpr
+     * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform
+     *
+     * @return string
+     */
+    public function convertToWellKnownTextSQL(string $sqlExpr, AbstractPlatform $platform) : string;
+}


lib/Doctrine/DBAL/Types/WellKnownTextConverter.php
```diff
@@ -0,0 +1,58 @@
...
+ * This is a metadata function that is required for example in the ORM.
+ * Usage of {@link convertToDatabaseValueSQL} and
+ * {@link convertToPHPValueSQL} works for any type and mostly
+ * does nothing. This method can additionally be used for optimization purposes.
+ *
+ * @author Roman Borschel <roman@code-factory.org>
+ * @author Benjamin Eberlei <kontakt@beberlei.de>
+ * @since  2.0
+ */
+interface WellKnownTextConverter
```
greg0ire: At first, I named that `RequiresSqlConversion`, tell me what you prefer, or if you'd like something else.

lib/Doctrine/DBAL/Types/WellKnownTextConverter.php
```diff
@@ -0,0 +1,58 @@
...
+{
+    /**
+     * Modifies the SQL expression (identifier, parameter) to convert to a database value.
+     *
+     * @param string                                    $sqlExpr
+     * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform
+     *
+     * @return string
+     */
+    public function convertFromWellKnownTextSQL(string $sqlExpr, AbstractPlatform $platform): string;
```
greg0ire: I deliberately picked other method names, so that I could add type hinting.

lib/Doctrine/DBAL/Types/Type.php
```diff
@@ -305,6 +305,7 @@ public function canRequireSQLConversion()
      * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform
      *
      * @return string
+     * @deprecated implement WellKnownTextConverter::convertFromWellKnownTextSQL instead
```
greg0ire: Should be removed in doctrine 3

greg0ire: Not implementing the new interface though, which is optional and is not needed for any of the native types.

greg0ire:
ping

lib/Doctrine/DBAL/Types/WellKnownTextConverter.php
Majkl578: Actually, since this is targeting 2.7 which will probably require PHP 7.2 (according to current policy of _releasing for latest stable PHP_), you could add parameter typehints without breaking public API. Not return type though, those could only be done later in 3.0.

lib/Doctrine/DBAL/Types/WellKnownTextConverter.php
Majkl578: I don't really like this name but don't have better proposal ATM, let's think about it...

lib/Doctrine/DBAL/Types/WellKnownTextConverter.php
```diff
@@ -0,0 +1,58 @@
...
+ * Does working with this column require SQL conversion functions?
+ *
+ * This is a metadata function that is required for example in the ORM.
+ * Usage of {@link convertToDatabaseValueSQL} and
+ * {@link convertToPHPValueSQL} works for any type and mostly
+ * does nothing. This method can additionally be used for optimization purposes.
+ *
+ * @author Roman Borschel <roman@code-factory.org>
+ * @author Benjamin Eberlei <kontakt@beberlei.de>
+ * @since  2.0
```
Majkl578: `@since` and `@author` should be updated.

lib/Doctrine/DBAL/Types/Type.php
Majkl578: Right, but existing types that do this type of conversion should implement it, if there are any (maybe not).

lib/Doctrine/DBAL/Types/WellKnownTextConverter.php
greg0ire: I see the RC2 is released, so maybe this will be quite soon

Majkl578: ETA @ end of november.

lib/Doctrine/DBAL/Types/WellKnownTextConverter.php
SenseException: Where does "WellKnown" come from? I don't understand that interface name.

lib/Doctrine/DBAL/Types/WellKnownTextConverter.php
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * Modifies the SQL expression (identifier, parameter) to convert to a PHP value.
+     *
+     * @param string                                    $sqlExpr
+     * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform
+     *
+     * @return string
+     */
+    public function convertToWellKnownTextSQL(string $sqlExpr, AbstractPlatform $platform): string;
```
SenseException: Spaces between return types.

lib/Doctrine/DBAL/Types/WellKnownTextConverter.php
SenseException: Spaces between return types. The return types will stay in the interface, right? :-)

lib/Doctrine/DBAL/Types/WellKnownTextConverter.php
morozov: If all types extend the base `Type` class, and it has default implementations for the conversion methods, why do we need an additional interface? The caller doesn't need to know if any conversion is happening within the type implementation, it can just call `convert()` and let the type take care of the conversion.

I'm not an ORM user, so could you explain briefly what problem is being solved?

greg0ire: @SenseException Ctrl+F "well-know" in the link I gave

greg0ire: @morozov this PR is pedantic; it does not solve any issue, it's just about architectural beauty: if you have a boolean method + methods that should be taken into account only if the boolean method returns true, it just begs to be rewritten with an interface. Note that this will make the `Type` class less long with 3 methods that don't do anything removed.

lib/Doctrine/DBAL/Types/WellKnownTextConverter.php
greg0ire: > Not return type though, those could only be done later in 3.0.

@SenseException not completely sure what @Majkl578 meant with that.

lib/Doctrine/DBAL/Types/WellKnownTextConverter.php
morozov: But in any case, the caller should either call a method on the object to see if it implements the WKT representation, or check if it implements the interface? The second is even worse because it has to know about two interfaces (`Type` and `WKT`) instead of one (`Type`).

Is it possible to avoid having the caller call this "marker" method at all and just expect it call the `convert()` method instead?

I don't see beauty in making the Type interface smaller because there's still type-related logic which is implemented on the caller side.

Majkl578: @morozov Please have a look at #2841, the motivation behind this should be clearer.
Basically it's just about killing `Type` god object, splitting it into smaller pieces and also making consuming Type classes more lightweight (99% of types don't even need this typeof conversion at all, so why would they need to have dummy methods).

lib/Doctrine/DBAL/Types/WellKnownTextConverter.php
SenseException: Those methods are new, so my guess is, that type widening isn't relevant here and the interface's API with the return type can stay as it is. Return types aren't covered in type widening and he just mentioned that.

Majkl578: > The return types will stay in the interface

Not in 2.x, adding return types to ancestors is a BC break, only parameter types are acceptable (and even those only as of PHP 7.2).

As I outlined on Slack, it may be easiest to just not add types at all now, just introduce the interface and add types later when master requires PHP 7.2+.

greg0ire:
Rebased

docs/en/reference/types.rst
```diff
@@ -929,29 +929,30 @@ Now we implement our ``Doctrine\DBAL\Types\Type`` instance:
...
 ::
 
     <?php
-    public function canRequireSQLConversion()
-    {
-        return true;
-    }
+    final class MyType implements WellKnownTextConverter
 
+<<<<<<< HEAD
```
Ocramius: @greg0ire rebase introduced issues here

greg0ire: Damn

greg0ire: Fixed

lib/Doctrine/DBAL/Types/WellKnownTextConverter.php
greg0ire: I think we can trust git for this kind of things.

greg0ire:
> I don't really like this name but don't have better proposal ATM, let's think about it...

Does anyone have a better idea?

morozov:
> Does anyone have a better idea?

Whatever the name is, I still don't like the idea of having a separate interface which the caller should check against.

> It removes the need for canRequireSqlConversion.

But introduces the need for checking `implements WellKnownTextConverter` which is exactly the same.

Currently, types only can convert values but cannot build SQL which is needed for some of them. Therefore, types should be responsible for building SQL as well:
```php
interface Type
{
    function buildQuery(Query Builder $builder, $value);
}

class Integer interface Type
{
    function buildQuery(Query Builder $builder, $value)
    {
        return $builder->createPositionalParameter($value, PDO::PARAM_INT);
    }
}

class Point interface Type
{
    function buildQuery(Query Builder $builder, $value)
    {
        return sprintf(
            'Point(%s, %s)', 
            $builder->createPositionalParameter($value->x),
            $builder->createPositionalParameter($value->y)
        );
    }
}
```

With this approach, the consumer of the interface doesn't have to make any assumptions, it just delegates the logic of building type-specific query fragments to the type. Always.

greg0ire:
Conversion is not always needed, it's an optional feature, which is exactly what interfaces are for.
With that marker interface, you can get rid of 3 default methods in your object, and from one in your class if you don't need it. If you do need it, you just read the doc, add the `implements`, and then php will tell you exactly what you need to implement. IDEs will generate a correct skeleton and import use statements for you. So no, it's not exactly the same.

If I understand well, you propose to get rid of the existing check altogether? I don't know the implications, but that sound like a bigger BC-break...

morozov:
> Conversion is not always needed, it's an optional feature, which is exactly what interfaces are for.

O-kay…

> With that marker interface, you can get rid of 3 default methods in your object, and from one in your class if you don't need it.

It's not about whether somebody needs it or not, it's about what features constitute a type. If it turns out that depending on the type, SQL may be different, then it's up to the type to decide what exactly that SQL should look like (even if there's no additional SQL in most cases), not to the caller — whether it needs/wants to call SQL-specific methods on the type.

Following your logic, the `Type::getDefaultLength()` only makes string for the `String` type. Why not introduce a `HasLength` interface and and only cal `$type->getDefaultLength()` if it `implements HasLength`?

> If I understand well, you propose to get rid of the existing check altogether? I don't know the implications, but that sounds like a bigger BC-break...

I propose, as long as we're doing an API cleanup, introduce a cleaner and more consistent API. Otherwise, it's a replacement for one workaround with another. Yes, a marker interface is a bit better than checking if a method exists, but it's still the same from the caller's standpoint — it still needs to perform a check. The BC concern is secondary. If the new API is better for developers, they will be glad to migrate.

greg0ire:
Following your logic, the Type::getDefaultLength() only makes string for the String type. Why not introduce a HasLength interface and and only cal $type->getDefaultLength() if it implements HasLength?

Well, probably with better naming, but yeah, pretty much. At the moment it's not needed because as you said, only the String type has a length, so `Type::getDefaultLength()` could probably be removed, just like this comment seems to suggest:

https://github.com/doctrine/dbal/blob/954ce2e82c5892dfbd51c4bf5bd154c85fc3ecf8/lib/Doctrine/DBAL/Types/Type.php#L144

I find that cleaner than calling X methods that do nothing 90% of the time, and maybe you do too, but if seems you have other ideas in mind, like moving more logic from the caller to the type itself? That does sound like a good plan 🤔 especially if it allows to expose less methods publicly. Closing then.

PR #2846, opened 2017-09-06
Ensure column options map to an existing method

greg0ire:


lib/Doctrine/DBAL/Schema/Column.php
```diff
@@ -118,9 +119,10 @@ public function setOptions(array $options)
     {
         foreach ($options as $name => $value) {
             $method = "set".$name;
-            if (method_exists($this, $method)) {
-                $this->$method($value);
+            if (!method_exists($this, $method)) {
+                throw InvalidArgumentException::fromUnsupportedOption($name);
```
greg0ire: Might be considered a BC-break, tell me if you want me to trigger deprecation errors instead.

greg0ire: I notice that it does not seem to be used anywhere in the code, but maybe you're not against it?

greg0ire:
ping

lib/Doctrine/DBAL/Schema/Column.php
```diff
@@ -118,9 +119,10 @@ public function setOptions(array $options)
     {
         foreach ($options as $name => $value) {
             $method = "set".$name;
-            if (method_exists($this, $method)) {
-                $this->$method($value);
+            if (!method_exists($this, $method)) {
```
Majkl578: There should be a space around `!`.

lib/Doctrine/DBAL/Schema/Column.php
Majkl578: Yes, this could indeed be considered unnecessary BC break in 2.x for no real benefit... And I'm not even sure deprecation notice is worth it. Maybe we should just retarget this to develop/3.0?

Majkl578 (dismissed):

lib/Doctrine/DBAL/Schema/Column.php
greg0ire: How about retargetting only this part, and merging the rest in master, since it's BC?

greg0ire: > for no real benefit...

A smooth upgrade path for people misusing this method is a good benefit IMHO. And the number of misuses I found in `doctrine/dbal` alone suggests this might be useful to many, don't you think.

Majkl578: Agreed, I meant there is _"no real benefit"_ in throwing an exception in 2.x due possible BC concerns.
Anyway, I'm not sure what is the correct approach regarging deprecations. I think the silent variant (what i.e. Symfony does) `@triger_error('blahblah', E_USER_DEPRECATED)` would be just fine. We will eventually need to deprecate multiple things anyway at latest in last 2.x before 3.0 comes (especially in ORM).

greg0ire: Great, I'll go with that then!

lib/Doctrine/DBAL/Exception/InvalidArgumentException.php
```diff
@@ -37,4 +37,9 @@ public static function fromEmptyCriteria()
     {
         return new self('Empty criteria was used, expected non-empty criteria');
     }
+
+    public static function fromUnsupportedOption(string $name): self
```
SenseException: Spaces between return types.

lib/Doctrine/DBAL/Schema/Column.php
SenseException: It's funny to see old options in the tests, that had no use at all. I like the stricter handling with that exception, but I agree on BC and the deprecation trigger.

greg0ire: To test it properly, I will need to make `simple-phpunit` work though :(
Right now I get:

```
PHP Fatal error:  Class 'Doctrine\Tests\DbalTestCase' not found in /home/greg/dev/dbal/tests/Doctrine/Tests/DBAL/Cache/QueryCacheProfileTest.php on line 9

Fatal error: Class 'Doctrine\Tests\DbalTestCase' not found in /home/greg/dev/dbal/tests/Doctrine/Tests/DBAL/Cache/QueryCacheProfileTest.php on line 9

```

.travis.yml
```diff
@@ -26,7 +26,7 @@ before_script:
 install:
   - travis_retry composer -n install
 
-script: ./vendor/bin/phpunit --configuration tests/travis/$DB.travis.xml
+script: SYMFONY_DEPRECATIONS_HELPER=weak_vendors ./vendor/bin/phpunit --configuration tests/travis/$DB.travis.xml
```
greg0ire: Not using the `simple-phpunit` binary because of a crash I don't understand (seems to have to do with autoloading).
The "weak_vendors" will make the tests suite fail if I forget to remove any call to setOption with an unknown option, unless it's in a test marked with the "@legacy" annotation. The test suite won't fail if a vendor suddenly adds a deprecation (that's why it's called weak vendors, it's lenient with deprecations from vendors)

composer.json
```diff
@@ -17,16 +17,20 @@
...
     },
     "require-dev": {
+        "doctrine/coding-standard": "^1.0",
         "phpunit/phpunit": "^6.3",
         "phpunit/phpunit-mock-objects": "!=3.2.4,!=3.2.5",
+        "squizlabs/php_codesniffer": "^3.0",
         "symfony/console": "2.*||^3.0",
-        "doctrine/coding-standard": "^1.0",
-        "squizlabs/php_codesniffer": "^3.0"
+        "symfony/phpunit-bridge": "^3.3"
```
greg0ire: Adding this so that we get a report of deprecations at the end of the test suite.

composer.json
```diff
@@ -17,16 +17,20 @@
...
     },
     "suggest": {
         "symfony/console": "For helpful console commands such as SQL execution and import of files."
     },
     "bin": ["bin/doctrine-dbal"],
+    "config": {
+        "sort-packages": true
+    },
```
greg0ire: Sorted packages in a separate commit, please review commit by commit.

lib/Doctrine/DBAL/Exception/InvalidArgumentException.php
```diff
@@ -37,4 +37,9 @@ public static function fromEmptyCriteria()
     {
         return new self('Empty criteria was used, expected non-empty criteria');
     }
+
+    public static function fromUnsupportedOption(string $name) : self
+    {
+        return new self(sprintf('The "%s" option is not supported', $name));
+    }
```
greg0ire: Unused at the moment, should be used in 3.0.

tests/Doctrine/Tests/DBAL/Schema/ColumnTest.php
```diff
@@ -56,6 +57,17 @@ public function testToArray()
     }
 
     /**
+     * @group legacy
```
greg0ire: I was not able to use the `simple-phpunit` binary, which means I can test the deprecation message as described here: https://symfony.com/doc/current/components/phpunit_bridge.html#write-assertions-about-deprecations

lib/Doctrine/DBAL/Schema/Column.php
```diff
@@ -118,9 +119,17 @@ public function setOptions(array $options)
     {
         foreach ($options as $name => $value) {
             $method = "set".$name;
-            if (method_exists($this, $method)) {
-                $this->$method($value);
+            if ( ! method_exists($this, $method)) {
+                // next major: use InvalidArgumentException::fromUnsupportedOption()
+                @trigger_error(sprintf(
```
greg0ire: Scrutinizer does not like the silencing

lib/Doctrine/DBAL/Exception/InvalidArgumentException.php
greg0ire: Should I remove it or should I strive to make the diff of the 3.0 PR minimal?

composer.json
greg0ire: Is everyone on board with this?

lib/Doctrine/DBAL/Schema/Column.php
SenseException: It seems like Scrutinizer allows silencing in an `if` block, but that's no option for this case. I haven't worked much with the Scrutinizer config, but maybe there's a way to ignore it just for `trigger_error`.

.travis.yml
greg0ire: I pushed something new. Turns out you don't have to use `simple-phpunit` to test deprecations properly.

phpunit.xml.dist
```diff
@@ -54,6 +54,7 @@
 
     <listeners>
         <listener class="Doctrine\Tests\DbalPerformanceTestListener"/>
+        <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener" />
```
greg0ire: This is how you get to test deprecations with annotations

tests/Doctrine/Tests/DBAL/Schema/ColumnTest.php
```diff
@@ -56,6 +57,16 @@ public function testToArray()
...
+     * @expectedDeprecation The "unknown_option" option is not supported, setting it is deprecated and will cause an exception in 3.0
+     * @doesNotPerformAssertions
```
greg0ire: The build passed from red to green after adding this line... not sure why: https://github.com/symfony/symfony/pull/21896

lib/Doctrine/DBAL/Schema/Column.php
greg0ire: Maybe this link would help:  https://scrutinizer-ci.com/docs/reviews/filtering_issues_and_false_positives ?

greg0ire:
I remove the `weak_vendors` mode as asked by @Ocramius , please review again.

lib/Doctrine/DBAL/Schema/Column.php
```diff
@@ -118,9 +119,17 @@ public function setOptions(array $options)
...
+                    'The "%s" option is not supported,'.
+                    ' setting it is deprecated and will cause an exception in 3.0',
+                    $name
+                ), E_USER_DEPRECATED);
+
+                return;
```
Pierstoval: Why not `return $this` like a few lines after?

greg0ire: Good catch, that's a BC-break!

lib/Doctrine/DBAL/Exception/InvalidArgumentException.php
Majkl578: I'd remove it since it's not used anywhere.

lib/Doctrine/DBAL/Schema/Column.php
Majkl578: @lcobucci maybe you can help us here?

composer.json
morozov: What's the purpose of reporting the usage of deprecated code in a test run? As far as I understand, the code is deprecated for the outside usage but the test still has to be tested until removed entirely.

greg0ire: @morozov you are right in that in needs to be tested. Such tests are supposed to be marked with a dedicated annotation `@group legacy`, so that it does not make the build fail. But if somewhere in the lib, we call the deprecated API ourselves, we must be aware of it, and change that. If we don't users, will get a deprecation *and won't be able to do anything about it*. That's what this is about.

morozov: @greg0ire thanks for the explanation. Makes sense.

lib/Doctrine/DBAL/Schema/Column.php
morozov: Could anyone please point me to the article explaining why this silencing operator is used in Symfony and we want it here? As far as I understand, the built-in PHP error handler will not display such errors. A user-defined handler, if it respects `error_reporting()`, it will ignore such errors as well. Does it mean users will have to use some specialized error handler to catch these errors?

greg0ire: @morozov  here is the handler: https://github.com/symfony/symfony/blob/master/src/Symfony/Bridge/PhpUnit/DeprecationErrorHandler.php IIRC most implementations ignore silenced errors, this one does not. I think a PSR is coming regarding all this.

greg0ire: We want it here not only for Symfony users, but for us: we want to know if we use a deprecated path somewhere else in the lib

lib/Doctrine/DBAL/Schema/Column.php
```diff
@@ -118,9 +119,17 @@ public function setOptions(array $options)
     {
         foreach ($options as $name => $value) {
             $method = "set".$name;
-            if (method_exists($this, $method)) {
-                $this->$method($value);
+            if ( ! method_exists($this, $method)) {
+                // next major: throw an exception
+                @trigger_error(sprintf(
```
morozov: The thread started in https://github.com/doctrine/dbal/pull/2846#discussion_r155290648.

> most implementations ignore silenced errors, this one does not

That's my point. Why should this error be ignored by most implementations?

> If we don't users, will get a deprecation *and won't be able to do anything about it*.

If we trigger suppressed errors, the users will not see any errors and won't be able to react unless their code is covered by tests which uses the same listener.

greg0ire: > If we trigger suppressed errors, the users will not see any errors and won't be able to react unless their code is covered by tests which uses the same listener.

Yes, and that's anyone that uses the phpunit-bridge component (so not just sf users, but also any library that relies on it for testing), but anyone who uses that particular component. That's not the only advantage, see https://github.com/doctrine/dbal/pull/2846#discussion_r155290648

Here is the list of the > 1200 libs that use it: https://packagist.org/packages/symfony/phpunit-bridge/dependents

If you are still not convinced, then maybe this should wait for a PSR an even larger portion of the population benefits from that.

morozov: > Yes, and that's anyone that uses the phpunit-bridge component (so not just sf users, but also any library that relies on it for testing)

This is fine. But there are also libraries which don't use the bridge component. Why should we suppress this error for them?

> We want it here not only for Symfony users, but for us: we want to know if we use a deprecated path somewhere else in the lib

I totally get it. I'm asking why we need to suppress the error, not why we need to trigger it.

> Here is a list of the > 1200 libs that use it: https://packagist.org/packages/symfony/phpunit-bridge/dependents

It doesn't explain why the error needs to be suppressed.

> If you are still not convinced, then maybe this should wait for a PSR an even larger portion of the population benefits from that.

You don't have to convince me. I'm just keeping asking the same question which I'd like to get an answer for.

greg0ire: > This is fine. But there are also libraries which don't use the bridge component. Why should we suppress this error for them?

Because otherwise I'm afraid their error handler will turn it into an exception.

greg0ire: So not silencing would not be BC at all :P

morozov: > Because otherwise I'm afraid their error handler will turn it into an exception.

As far as I can tell, most error handlers override the way how the errors are _logged_ (e.g. instead of displaying them, log them to a logging service). If the handler converts errors to exceptions, it should understand that a deprecated error is not an exceptional situation. Logically, its severity is even lower than a notice because the code keeps working as designed.

I see what you're saying but if we want to inform users about them using the deprecated APIs, this message should be available for them without additional means (the bridge). Otherwise, we will not inform the users who don't use the bridge (which is fine) in favor of those who have badly designed error-to-exception converters (which is not).

morozov: > So not silencing would not be BC at all :P

It depends on how BC is defined. If "not being informed about using deprecated APIs" is part of the previous API, then yes. Was the previous API defined like this?

nicolas-grekas:
:+1: that is a proven practice on Symfony since 2.7. we live it.
Ping @ocramius since we already talked about that earlier.

lib/Doctrine/DBAL/Schema/Column.php
greg0ire: > Was the previous API defined like this?

AFAIK it wasn't defined at all, which should probably mean we can't suddenly decide people should be informed through an error, moreover it it might be turned into an exception. 

> I see what you're saying but if we want to inform users about them using the deprecated APIs, this message should be available for them without additional means (the bridge). Otherwise, we will not inform the users who don't use the bridge (which is fine) in favor of those who have badly designed error-to-exception converters (which is not).

I don't know any other dynamic way. There are static ways like the `@deprecated` tag, but that's it.

> in favor of those who have badly designed error-to-exception converters (which is not).

Sorry, I'm a bit slow, I didn't quite understand that last bit. Do you mean error to exception converters who would also ignore the silence operator?

morozov: > [...] we can't suddenly decide people should be informed through an error.

Why not? It's a non-fatal error. If triggering an error is a bad idea why do we want to use `trigger_error()`?

> [...] moreover it it might be turned into an exception

How is it expected to happen?

> Do you mean error to exception converters who would also ignore the silence operator?

No, I mean the converters who convert deprecated errors to exceptions. There's nothing exceptional if deprecated code is used in production.

morozov:
@nicolas-grekas you're talking about the error suppression or about the listener in general? If the former, could you point to an article/discussion thread explaining why Symfony chose this approach?

lib/Doctrine/DBAL/Schema/Column.php
greg0ire: > Why not? It's a non-fatal error. If triggering an error is a bad idea why do we want to use trigger_error()?

I think that would be because how php behaves if you have no error handler at all:

https://3v4l.org/7WsRc

IMO this should probably something in the error_log, but not display anything.

> How is it expected to happen?

I think having both errors and exceptions in the php language is not great, and that flaw started being addressed with php 5.1 with this [ErrorException](http://php.net/manual/en/class.errorexception.php), and since then it's pretty common practice for error handlers to do that, here is a random, non-symfony example to back that up: https://github.com/filp/whoops/blob/2abce9d956589122c6443d6265f01cf7e9388e3c/src/Whoops/Run.php#L349 . So I really think it's a sensible assumptionto make.

> No, I mean the converters who convert deprecated errors to exceptions. There's nothing exceptional if deprecated code is used in production.

I would call that "in favor of", I would call that a big BC-break. So maybe it should first be introduced with the silence operator, and then the silence operator should be removed on next major? But that wouldn't make much sense since the deprecated paths would be removed as well.


Here is my source for saying this could become a PSR: https://medium.com/php-fig/a-month-of-php-fig-1-october-2017-af253682ef5b

morozov: > I think that would be because how php behaves if you have no error handler at all:
> 
> https://3v4l.org/7WsRc

This is exactly how it should work. It will let developers know they are using deprecated APIs without having to use any additional libraries.

> IMO this should probably produce something in the error_log, but not display anything.

It should be controlled by `display_errors` and `error_log` INI settings, not by the DBAL code.

> here is a random, non-symfony example to back that up

Why is a deprecated error an exception? At runtime, there's nothing exceptional about using deprecated APIs.

> I would call that "in favor of", I would call that a big BC-break.

What exactly is being broken?

> Here is my source for saying this could become a PSR

Could you pls quote what could become a PSR and point to a document describing the proposal?

greg0ire: > This is exactly how it should work. It will let developers know they are using deprecated APIs without having to use any additional libraries.

What about test suite of other projects that don't use any error handler? Is that the case for `doctrine/orm` for instance?


> What exactly is being broken?

upgrading the lib produces something on the output. Let me ask the sf community about this, maybe I didn't grasp it entirely.



> Could you pls quote what could become a PSR and point to a document describing the proposal?

That's all the details I have right now: 

> For handling deprecations of classes and methods to trigger errors/notifications

greg0ire: @morozov I think I found the real reason:

> to guarantee opt-in of deprecation. Otherwise, it's only opt-in with a "reasonable error handler", which isn't that opt in. Besides, the default php.ini production value for `error_reporting` only hides PHP deprecations, not user deprecations

— @wouterjnl

So I guess silencing is up to you, and I will remove it if you want me to.

nicolas-grekas:
We started working on adding deprecation notices end of 2014. Back then, we didn't have the `@`.
We worked hard to make that kinda work (just search "silence deprecations" in the PR history.)
The, someone can up with a PR adding `@` everywhere: https://github.com/symfony/symfony/pull/14900
and all the problem we were struggling with vanished.
Since then, we made Symfony 3 and Symfony 4, *because* of this policy, which allowed us to actually do it (remove code.)

So yes, you got it well: the only, simple argument is **to make dealing with deprecations opt-in**. There is no need for any other.

nicolas-grekas:
See also https://symfony.com/doc/3.3/components/phpunit_bridge.html#trigger-deprecation-notices

lib/Doctrine/DBAL/Schema/Column.php
alcaeus: @morozov as mentioned here and elsewhere, the appropriate way to trigger deprecation notices is `@trigger_error(..., E_USER_DEPRECATED);`

As explained by @nicolas-grekas, without the error suppression operator the code would start spitting out deprecation warnings to STDERR which we don't want: deprecation notices are an opt-in feature: users should not be bothered by it unless they specifically decide to care. Thus, the only way to solve this is to suppress the `trigger_error` call and let a dedicated deprecation handler (such as the listener from PhpunitBridge) log the deprecations and display them accordingly.

tests/Doctrine/Tests/DBAL/Schema/ColumnTest.php
```diff
@@ -55,6 +56,15 @@ public function testToArray()
         self::assertEquals($expected, $this->createColumn()->toArray());
     }
 
+    /**
+     * @group legacy
+     * @expectedDeprecation The "unknown_option" option is not supported, setting it is deprecated and will cause an exception in 3.0
```
alcaeus: TIL there's `@expectedDeprecation`. Thanks!

alcaeus (approved):

tests/Doctrine/Tests/DBAL/Schema/ColumnTest.php
greg0ire: You're very welcome!

lib/Doctrine/DBAL/Schema/Column.php
greg0ire: Thanks for providing a clear explanation, I had a hard time doing it!

morozov:
@greg0ire, @nicolas-grekas, @alcaeus, thank you for the details. I understand the reasoning now.

lib/Doctrine/DBAL/Schema/Column.php
morozov: @greg0ire sorry for giving you a hard time :smile:

greg0ire:
@morozov thanks for making me dig deeper, I understand it better too now :)

lib/Doctrine/DBAL/Schema/Column.php
greg0ire: No problem! It was worth it. Very reassuring to see people are doing due diligence on that project!

lib/Doctrine/DBAL/Schema/Column.php
```diff
@@ -19,6 +19,7 @@
 
 namespace Doctrine\DBAL\Schema;
 
+use Doctrine\DBAL\Exception\InvalidArgumentException;
```
Majkl578: This use is unused.

lib/Doctrine/DBAL/Schema/Column.php
```diff
@@ -118,9 +119,17 @@ public function setOptions(array $options)
     {
         foreach ($options as $name => $value) {
             $method = "set".$name;
-            if (method_exists($this, $method)) {
-                $this->$method($value);
+            if ( ! method_exists($this, $method)) {
+                // next major: throw an exception
+                @trigger_error(sprintf(
+                    'The "%s" option is not supported,'.
+                    ' setting it is deprecated and will cause an exception in 3.0',
```
Majkl578: Probably should not mention _exception_ directly (implementation detail), just _error_.

Majkl578 (approved):

Majkl578:
@Ocramius Are you on board with using `@trigger_error(..., E_USER_DEPRECATED)` pattern for deprecations?

Ocramius:
@Majkl578 yes, this is something I discussed in person with @nicolas-grekas, and the approach is solid.

phpunit.xml.dist
Ocramius: Not a fan of this, but hey, it's dev-only, so I can live with it.

Ocramius (approved):

Ocramius:
:ship:

Ocramius:
@greg0ire can I ask you to rename the PR title to something more clear to future readers?.

greg0ire:
@Ocramius sure!

tests/Doctrine/Tests/DBAL/Schema/ColumnTest.php
```diff
@@ -55,6 +56,15 @@ public function testToArray()
         self::assertEquals($expected, $this->createColumn()->toArray());
     }
 
+    /**
+     * @group legacy
+     * @expectedDeprecation The "unknown_option" option is not supported, setting it is deprecated and will cause an error in 3.0
+     */
+    public function testSettingUnknownOptionIsStillSupported() : void
```
morozov: @greg0ire, in some cases, [this test is reported as risky](https://ci.appveyor.com/project/doctrine/dbal/build/1.0.81/job/utt2h7bq2p2b5jj7#L239):

> There was 1 risky test:
> 1) Doctrine\Tests\DBAL\Schema\ColumnTest::testSettingUnknownOptionIsStillSupported
> This test did not perform any assertions
> ...
> Legacy deprecation notices (1)

Is it expected?

wouterj: Older versions of Symfony's phpunit bridge do not count expected deprecation as assertion (and thus mark the test risky).

I would advice to always run all test jobs with the latest version of the bridge

morozov: @wouterj could you suggest the proper `composer.json` constraints? Currently, we have:

https://github.com/doctrine/dbal/blob/46069b6bc84b0f41733a23d3b56e2774d0502c94/composer.json#L24

greg0ire: Dropping the left part of the pipe is completely fine IMO.

morozov: @greg0ire on the build in question, this constraint [was resolved to `v4.0.6`](https://ci.appveyor.com/project/doctrine/dbal/build/1.0.81/job/utt2h7bq2p2b5jj7#L105).

greg0ire: Ah sorry I misread this as 4.0.6 fixes the issue… latest is 4.0.7

Majkl578: Can we somehow isolate the issue and report it to upstream?

greg0ire: I'm trying to reproduce it as we speak but I'm afraid it only ever happens on windows. Should I create a PR with zero tests but the faulty one?

morozov: It's fine if it helps to isolate the issue.

greg0ire: Ok then. See https://github.com/symfony/symfony/issues/26809

lib/Doctrine/DBAL/Schema/Column.php
```diff
@@ -118,9 +118,17 @@ public function setOptions(array $options)
...
-                $this->$method($value);
+            if ( ! method_exists($this, $method)) {
+                // next major: throw an exception
+                @trigger_error(sprintf(
+                    'The "%s" option is not supported,'.
+                    ' setting it is deprecated and will cause an error in 3.0',
+                    $name
+                ), E_USER_DEPRECATED);
+
+                return $this;
```
andytruong: );

greg0ire: What is it @andytruong ?

Majkl578: #3089

PR #2850, opened 2017-09-09
Support for backslashes in default values

PowerKiKi:
It is now possible to use `\`, or other escaped characters, in default
values for a column.

Previously this lead to a confusion when diffing actual and expected
schema leading to perpetual out of sync schema.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -2306,7 +2306,7 @@ abstract class AbstractPlatform
             return " DEFAULT '" . $this->convertBooleans($default) . "'";
         }
 
-        return " DEFAULT " . $this->quoteStringLiteral($default);
+        return " DEFAULT " . $this->quoteDefaultStringLiteral($default);
     }
 
     /**
@@ -3569,5 +3569,17 @@ abstract class AbstractPlatform
     public function getStringLiteralQuoteCharacter()
     {
         return "'";
+    }
+
+    /**
+     * Quote a literal string for usage as DEFAULT value.
+     *
+     * @param string $str
+     *
+     * @return string
+     */
+    protected function quoteDefaultStringLiteral(string $str) : string
+    {
+        return $this->quoteStringLiteral($str);
     }
 }
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -2306,7 +2306,7 @@ abstract class AbstractPlatform
             return " DEFAULT '" . $this->convertBooleans($default) . "'";
         }
 
-        return " DEFAULT '" . $default . "'";
+        return " DEFAULT " . $this->quoteStringLiteral($default);
     }
 
     /**
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -1236,4 +1236,12 @@ class PostgreSqlPlatform extends AbstractPlatform
     {
         return $type instanceof IntegerType || $type instanceof BigIntType;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected function quoteDefaultStringLiteral(string $str) : string
+    {
+        return parent::quoteStringLiteral($str);
+    }
 }
--- a/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
@@ -483,7 +483,6 @@ class PostgreSqlSchemaManager extends AbstractSchemaManager
             return $default;
         }
 
-        $default = str_replace("\\\\", "\\", $default);
         $default = str_replace("''", "'", $default);
 
         return $default;
--- a/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
@@ -477,7 +477,7 @@ class PostgreSqlSchemaManager extends AbstractSchemaManager
      *
      * @return null|string
      */
-    private function unescapeDefaultValue(?string $default): ?string
+    private function unescapeDefaultValue(?string $default) : ?string
     {
         if ($default === null) {
             return $default;
--- a/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
@@ -384,11 +384,12 @@ class PostgreSqlSchemaManager extends AbstractSchemaManager
                 $length = null;
                 break;
             case 'text':
+            case '_varchar':
+            case 'varchar':
+            $tableColumn['default'] = $this->fixDefaultValueQuotes($tableColumn['default']);
                 $fixed = false;
                 break;
-            case 'varchar':
             case 'interval':
-            case '_varchar':
                 $fixed = false;
                 break;
             case 'char':
@@ -467,5 +468,24 @@ class PostgreSqlSchemaManager extends AbstractSchemaManager
         }
 
         return $defaultValue;
+    }
+
+    /**
+     * Un-escape a default value as given by PostgreSQL
+     *
+     * @param null|string $default
+     *
+     * @return null|string
+     */
+    private function fixDefaultValueQuotes(?string $default): ?string
+    {
+        if ($default === null) {
+            return $default;
+        }
+
+        $default = str_replace("\\\\", "\\", $default);
+        $default = str_replace("''", "'", $default);
+
+        return $default;
     }
 }
--- a/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
@@ -386,7 +386,7 @@ class PostgreSqlSchemaManager extends AbstractSchemaManager
             case 'text':
             case '_varchar':
             case 'varchar':
-                $tableColumn['default'] = $this->fixDefaultValueQuotes($tableColumn['default']);
+                $tableColumn['default'] = $this->unescapeDefaultValue($tableColumn['default']);
                 $fixed = false;
                 break;
             case 'interval':
@@ -477,7 +477,7 @@ class PostgreSqlSchemaManager extends AbstractSchemaManager
      *
      * @return null|string
      */
-    private function fixDefaultValueQuotes(?string $default): ?string
+    private function unescapeDefaultValue(?string $default): ?string
     {
         if ($default === null) {
             return $default;
--- a/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
@@ -386,7 +386,7 @@ class PostgreSqlSchemaManager extends AbstractSchemaManager
             case 'text':
             case '_varchar':
             case 'varchar':
-            $tableColumn['default'] = $this->fixDefaultValueQuotes($tableColumn['default']);
+                $tableColumn['default'] = $this->fixDefaultValueQuotes($tableColumn['default']);
                 $fixed = false;
                 break;
             case 'interval':
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
@@ -454,21 +454,7 @@ class MySqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
         $platform = $this->_sm->getDatabasePlatform();
         $this->_conn->query('DROP TABLE IF EXISTS test_column_defaults_with_create');
 
-        $escapeSequences = [
-            "\\0",          // An ASCII NUL (X'00') character
-            "\\'", "''",    // Single quote
-            '\\"', '""',    // Double quote
-            '\\b',          // A backspace character
-            '\\n',          // A new-line character
-            '\\r',          // A carriage return character
-            '\\t',          // A tab character
-            '\\Z',          // ASCII 26 (Control+Z)
-            '\\\\',         // A backslash (\) character
-            '\\%',          // A percent (%) character
-            '\\_',          // An underscore (_) character
-        ];
-
-        $default = implode('+', $escapeSequences);
+        $default = implode('+', $this->getEscapedLiterals());
 
         $sql = "CREATE TABLE test_column_defaults_with_create(
                     col1 VARCHAR(255) NULL DEFAULT {$platform->quoteStringLiteral($default)} 
@@ -476,5 +462,24 @@ class MySqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
         $this->_conn->query($sql);
         $onlineTable = $this->_sm->listTableDetails("test_column_defaults_with_create");
         self::assertSame($default, $onlineTable->getColumn('col1')->getDefault());
+    }
+
+    protected function getEscapedLiterals() : array
+    {
+        return [
+            "\\0", // An ASCII NUL (X'00') character
+            "\\'", // Single quote
+            "''",  // Single quote
+            '\\"', // Double quote
+            '""',  // Double quote
+            '\\b', // A backspace character
+            '\\n', // A new-line character
+            '\\r', // A carriage return character
+            '\\t', // A tab character
+            '\\Z', // ASCII 26 (Control+Z)
+            '\\\\', // A backslash (\) character
+            '\\%', // A percent (%) character
+            '\\_', // An underscore (_) character
+        ];
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
@@ -442,30 +442,4 @@ class MySqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
         $diff = $comparator->diffTable($table, $onlineTable);
         self::assertFalse($diff, "Tables should be identical with column defauts time and date.");
     }
-
-    /**
-     * Returns literals that are platform specific escaped.
-     *
-     * @see https://dev.mysql.com/doc/refman/5.7/en/string-literals.html
-     *
-     * @return array
-     */
-    protected function getEscapedLiterals(): array
-    {
-        return [
-            "\\0", // An ASCII NUL (X'00') character
-            "\\'", // Single quote
-            "''",  // Single quote
-            '\\"', // Double quote
-            '""',  // Double quote
-            '\\b', // A backspace character
-            '\\n', // A new-line character
-            '\\r', // A carriage return character
-            '\\t', // A tab character
-            '\\Z', // ASCII 26 (Control+Z)
-            '\\\\', // A backslash (\) character
-            '\\%', // A percent (%) character
-            '\\_', // An underscore (_) character
-        ];
-    }
 }
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
@@ -444,27 +444,13 @@ class MySqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
     }
 
     /**
-     * Ensure default values (un-)escaping is properly done by mysql platforms.
-     * The test is voluntarily relying on schema introspection due to current
-     * doctrine limitations. Once #2850 is landed, this test can be removed.
+     * Returns literals that are platform specific escaped.
+     *
      * @see https://dev.mysql.com/doc/refman/5.7/en/string-literals.html
+     *
+     * @return array
      */
-    public function testEnsureDefaultsAreUnescapedFromSchemaIntrospection() : void
-    {
-        $platform = $this->_sm->getDatabasePlatform();
-        $this->_conn->query('DROP TABLE IF EXISTS test_column_defaults_with_create');
-
-        $default = implode('+', $this->getEscapedLiterals());
-
-        $sql = "CREATE TABLE test_column_defaults_with_create(
-                    col1 VARCHAR(255) NULL DEFAULT {$platform->quoteStringLiteral($default)} 
-                )";
-        $this->_conn->query($sql);
-        $onlineTable = $this->_sm->listTableDetails("test_column_defaults_with_create");
-        self::assertSame($default, $onlineTable->getColumn('col1')->getDefault());
-    }
-
-    protected function getEscapedLiterals() : array
+    protected function getEscapedLiterals(): array
     {
         return [
             "\\0", // An ASCII NUL (X'00') character
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
@@ -1399,24 +1399,41 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
     }
 
     /**
-     * Returns literals that are platform specific escaped.
+     * Returns potential escaped literals from all platforms combined.
+     *
+     * @see https://dev.mysql.com/doc/refman/5.7/en/string-literals.html
+     * @see http://www.sqlite.org/lang_expr.html
+     * @see https://www.postgresql.org/docs/9.6/static/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-ESCAPE
      *
      * @return array
      */
-    protected function getEscapedLiterals() : array
+    public function getEscapedLiterals(): array
     {
         return [
-            "\\'", // A single quote
-            '\\"', // A double quote
-            '\\n', // A new-line character
-            '\\r', // A carriage return character
+            'An ASCII NUL (X\'00\')' => ["\\0"],
+            'Single quote, C-style' => ["\\'"],
+            'Single quote, doubled-style' => ["''"],
+            'Double quote, C-style' => ['\\"'],
+            'Double quote, double-style' => ['""'],
+            'Backspace' => ['\\b'],
+            'New-line' => ['\\n'],
+            'Carriage return' => ['\\r'],
+            'Tab' => ['\\t'],
+            'ASCII 26 (Control+Z)' => ['\\Z'],
+            'Backslash (\)' => ['\\\\'],
+            'Percent (%)' => ['\\%'],
+            'Underscore (_)' => ['\\_'],
         ];
     }
 
-    public function testEscapedDefaultValueMustBePreserved() : void
+    /**
+     * @dataProvider getEscapedLiterals
+     *
+     * @param string $value
+     */
+    public function testEscapedDefaultValueMustBePreserved(string $value) : void
     {
-        $value = implode('+', $this->getEscapedLiterals());
-
+        $value = 'foo' . $value . 'bar';
         $table = new Table('string_escaped_default_value');
         $table->addColumn('def_string', 'string', array('default' => $value));
         $table->addColumn('def_foo', 'string');
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
@@ -1442,9 +1442,6 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
         $onlineTable = $this->_sm->listTableDetails('string_escaped_default_value');
         self::assertSame($value, $onlineTable->getColumn('def_string')->getDefault(), 'should be able introspect the value of default');
 
-        $comparator = new Comparator();
-        self::assertFalse($comparator->diffTable($table, $onlineTable), 'introspected schema should have no differences');
-
         $this->_conn->insert('string_escaped_default_value', array('def_foo' => 'foo'));
         $row = $this->_conn->fetchAssoc('SELECT def_string FROM string_escaped_default_value');
         self::assertSame($value, $row['def_string'], 'inserted default value should be the configured default value');
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
@@ -1400,7 +1400,7 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
 
     public function testEscapedDefaultValueMustBePreserved()
     {
-        $value = "a\\0a\\'a\"a\na\ra\ta\\Za\\\\a" ;
+        $value = "a\\0b\\'c\"d\te\\Zf\\\\g";
 
         $table = new Table('string_escaped_default_value');
         $table->addColumn('def_string', 'string', array('default' => $value));
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
@@ -1400,7 +1400,7 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
 
     public function testEscapedDefaultValueMustBePreserved()
     {
-        $value = "a\\0b\\'c\"d\te\\Zf\\\\g";
+        $value = "a\\0b\\'c\"d\te\\Zf\\\\g''h";
 
         $table = new Table('string_escaped_default_value');
         $table->addColumn('def_string', 'string', array('default' => $value));
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
@@ -1407,43 +1407,55 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
      *
      * @return array
      */
-    public function getEscapedLiterals(): array
+    private function getEscapedLiterals(): array
     {
         return [
-            'An ASCII NUL (X\'00\')' => ["\\0"],
-            'Single quote, C-style' => ["\\'"],
-            'Single quote, doubled-style' => ["''"],
-            'Double quote, C-style' => ['\\"'],
-            'Double quote, double-style' => ['""'],
-            'Backspace' => ['\\b'],
-            'New-line' => ['\\n'],
-            'Carriage return' => ['\\r'],
-            'Tab' => ['\\t'],
-            'ASCII 26 (Control+Z)' => ['\\Z'],
-            'Backslash (\)' => ['\\\\'],
-            'Percent (%)' => ['\\%'],
-            'Underscore (_)' => ['\\_'],
+            ['An ASCII NUL (X\'00\')', "foo\\0bar"],
+            ['Single quote, C-style', "foo\\'bar"],
+            ['Single quote, doubled-style', "foo''bar"],
+            ['Double quote, C-style', 'foo\\"bar'],
+            ['Double quote, double-style', 'foo""bar'],
+            ['Backspace', 'foo\\bbar'],
+            ['New-line', 'foo\\nbar'],
+            ['Carriage return', 'foo\\rbar'],
+            ['Tab', 'foo\\tbar'],
+            ['ASCII 26 (Control+Z)', 'foo\\Zbar'],
+            ['Backslash (\)', 'foo\\\\bar'],
+            ['Percent (%)', 'foo\\%bar'],
+            ['Underscore (_)', 'foo\\_bar'],
         ];
     }
 
-    /**
-     * @dataProvider getEscapedLiterals
-     *
-     * @param string $value
-     */
-    public function testEscapedDefaultValueMustBePreserved(string $value) : void
+    private function createTableForDefaultValues(): void
     {
-        $value = 'foo' . $value . 'bar';
         $table = new Table('string_escaped_default_value');
-        $table->addColumn('def_string', 'string', array('default' => $value));
+        foreach ($this->getEscapedLiterals() as $i => $literal) {
+            $table->addColumn('field' . $i, 'string', ['default' => $literal[1]]);
+        }
+
         $table->addColumn('def_foo', 'string');
         $this->_sm->dropAndCreateTable($table);
+    }
+
+    public function testEscapedDefaultValueCanBeIntrospected(): void
+    {
+        $this->createTableForDefaultValues();
 
         $onlineTable = $this->_sm->listTableDetails('string_escaped_default_value');
-        self::assertSame($value, $onlineTable->getColumn('def_string')->getDefault(), 'should be able introspect the value of default');
+        foreach ($this->getEscapedLiterals() as $i => $literal) {
+            self::assertSame($literal[1], $onlineTable->getColumn('field' . $i)->getDefault(), 'should be able introspect the value of default for: ' . $literal[0]);
+        }
+    }
 
-        $this->_conn->insert('string_escaped_default_value', array('def_foo' => 'foo'));
-        $row = $this->_conn->fetchAssoc('SELECT def_string FROM string_escaped_default_value');
-        self::assertSame($value, $row['def_string'], 'inserted default value should be the configured default value');
+    public function testEscapedDefaultValueCanBeInserted(): void
+    {
+        $this->createTableForDefaultValues();
+
+        $this->_conn->insert('string_escaped_default_value', ['def_foo' => 'foo']);
+        $row = $this->_conn->fetchAssoc('SELECT * FROM string_escaped_default_value');
+        foreach ($this->getEscapedLiterals() as $i => $literal) {
+            self::assertSame($literal[1], $row['field' . $i], 'inserted default value should be the configured default value for: ' . $literal[0]);
+
+        }
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
@@ -1398,18 +1398,38 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
         self::assertEquals($sequence2InitialValue, $actualSequence2->getInitialValue());
     }
 
-    public function testEscapedDefaultValueMustBePreserved()
+    /**
+     * Returns literals that are platform specific escaped.
+     *
+     * @return array
+     */
+    protected function getEscapedLiterals() : array
     {
-        $value = "a\\0b\\'c\"d\te\\Zf\\\\g''h";
+        return [
+            "\\'", // A single quote
+            '\\"', // A double quote
+            '\\n', // A new-line character
+            '\\r', // A carriage return character
+        ];
+    }
+
+    public function testEscapedDefaultValueMustBePreserved() : void
+    {
+        $value = implode('+', $this->getEscapedLiterals());
 
         $table = new Table('string_escaped_default_value');
         $table->addColumn('def_string', 'string', array('default' => $value));
+        $table->addColumn('def_foo', 'string');
         $this->_sm->dropAndCreateTable($table);
 
         $onlineTable = $this->_sm->listTableDetails('string_escaped_default_value');
-        $this->assertSame($value, $onlineTable->getColumn('def_string')->getDefault());
+        self::assertSame($value, $onlineTable->getColumn('def_string')->getDefault(), 'should be able introspect the value of default');
 
         $comparator = new Comparator();
-        $this->assertFalse($comparator->diffTable($table, $onlineTable));
+        self::assertFalse($comparator->diffTable($table, $onlineTable), 'introspected schema should have no differences');
+
+        $this->_conn->insert('string_escaped_default_value', array('def_foo' => 'foo'));
+        $row = $this->_conn->fetchAssoc('SELECT def_string FROM string_escaped_default_value');
+        self::assertSame($value, $row['def_string'], 'inserted default value should be the configured default value');
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
@@ -1407,7 +1407,7 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
      *
      * @return array
      */
-    private function getEscapedLiterals(): array
+    private function getEscapedLiterals() : array
     {
         return [
             ['An ASCII NUL (X\'00\')', "foo\\0bar"],
@@ -1426,7 +1426,7 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
         ];
     }
 
-    private function createTableForDefaultValues(): void
+    private function createTableForDefaultValues() : void
     {
         $table = new Table('string_escaped_default_value');
         foreach ($this->getEscapedLiterals() as $i => $literal) {
@@ -1437,7 +1437,7 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
         $this->_sm->dropAndCreateTable($table);
     }
 
-    public function testEscapedDefaultValueCanBeIntrospected(): void
+    public function testEscapedDefaultValueCanBeIntrospected() : void
     {
         $this->createTableForDefaultValues();
 
@@ -1447,7 +1447,7 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
         }
     }
 
-    public function testEscapedDefaultValueCanBeInserted(): void
+    public function testEscapedDefaultValueCanBeInserted() : void
     {
         $this->createTableForDefaultValues();
 
@@ -1455,7 +1455,6 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
         $row = $this->_conn->fetchAssoc('SELECT * FROM string_escaped_default_value');
         foreach ($this->getEscapedLiterals() as $i => $literal) {
             self::assertSame($literal[1], $row['field' . $i], 'inserted default value should be the configured default value for: ' . $literal[0]);
-
         }
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
@@ -1397,4 +1397,19 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
         self::assertEquals($sequence2AllocationSize, $actualSequence2->getAllocationSize());
         self::assertEquals($sequence2InitialValue, $actualSequence2->getInitialValue());
     }
+
+    public function testEscapedDefaultValueMustBePreserved()
+    {
+        $value = "a\\0a\\'a\"a\na\ra\ta\\Za\\\\a" ;
+
+        $table = new Table('string_escaped_default_value');
+        $table->addColumn('def_string', 'string', array('default' => $value));
+        $this->_sm->dropAndCreateTable($table);
+
+        $onlineTable = $this->_sm->listTableDetails('string_escaped_default_value');
+        $this->assertSame($value, $onlineTable->getColumn('def_string')->getDefault());
+
+        $comparator = new Comparator();
+        $this->assertFalse($comparator->diffTable($table, $onlineTable));
+    }
 }
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
@@ -473,7 +473,7 @@ abstract class AbstractMySQLPlatformTestCase extends AbstractPlatformTestCase
             "ALTER TABLE mytable ADD PRIMARY KEY (foo)",
         ), $sql);
     }
-    
+
     public function testAlterPrimaryKeyWithNewColumn()
     {
         $table = new Table("yolo");
@@ -483,7 +483,7 @@ abstract class AbstractMySQLPlatformTestCase extends AbstractPlatformTestCase
 
         $comparator = new Comparator();
         $diffTable = clone $table;
-        
+
         $diffTable->addColumn('pkc2', 'integer');
         $diffTable->dropPrimaryKey();
         $diffTable->setPrimaryKey(array('pkc1', 'pkc2'));
@@ -495,7 +495,7 @@ abstract class AbstractMySQLPlatformTestCase extends AbstractPlatformTestCase
                 'ALTER TABLE yolo ADD PRIMARY KEY (pkc1, pkc2)',
             ),
             $this->_platform->getAlterTableSQL($comparator->diffTable($table, $diffTable))
-        );      
+        );
     }
 
     public function testInitializesDoctrineTypeMappings()
@@ -911,11 +911,11 @@ abstract class AbstractMySQLPlatformTestCase extends AbstractPlatformTestCase
     public function testGetDefaultValueDeclarationSQLEscaped()
     {
         // string must be escaped
-        $field = array(
+        $field = [
             'type' => 'string',
             'default' => 'Foo\\Bar'
-        );
+        ];
 
-        self::assertEquals(" DEFAULT 'Foo\\\\Bar'", $this->_platform->getDefaultValueDeclarationSQL($field));
+        self::assertSame(" DEFAULT 'Foo\\\\Bar'", $this->_platform->getDefaultValueDeclarationSQL($field));
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
@@ -907,4 +907,15 @@ abstract class AbstractMySQLPlatformTestCase extends AbstractPlatformTestCase
         self::assertContains('bar', $sql);
         self::assertNotContains('DATABASE()', $sql);
     }
+
+    public function testGetDefaultValueDeclarationSQLEscaped()
+    {
+        // string must be escaped
+        $field = array(
+            'type' => 'string',
+            'default' => 'Foo\\Bar'
+        );
+
+        self::assertEquals(" DEFAULT 'Foo\\\\Bar'", $this->_platform->getDefaultValueDeclarationSQL($field));
+    }
 }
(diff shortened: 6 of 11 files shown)


greg0ire:
Please rebase on the latest master to fix your build

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
```diff
@@ -180,6 +180,21 @@ public function testDoesNotPropagateDefaultValuesForUnsupportedColumnTypes()
         $this->assertFalse($onlineTable->getColumn('def_blob_null')->getNotnull());
     }
 
+    public function testEscapedDefaultValueMustBePreserved()
```
morozov: Since it's a functional test and it doesn't check any platform-specific logic, I think we could move it to a new `SchemaManagerTest` which instead of instantiating the schema manager would take it from `$conn->getSchemaManager()`.

This way, we could make sure this feature is portable and can be safely used on all supported platforms.

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -385,7 +385,7 @@ public function getListTableColumnsSQL($table, $database = null)
         }
 
         return "SELECT COLUMN_NAME AS Field, COLUMN_TYPE AS Type, IS_NULLABLE AS `Null`, ".
-               "COLUMN_KEY AS `Key`, COLUMN_DEFAULT AS `Default`, EXTRA AS Extra, COLUMN_COMMENT AS Comment, " .
+               "COLUMN_KEY AS `Key`, TRIM('\'' FROM QUOTE(COLUMN_DEFAULT)) AS `Default`, EXTRA AS Extra, COLUMN_COMMENT AS Comment, " .
```
morozov: Is the problem really in the introspection query? If I add the following assertion to the newly introduced test,
```php
$this->_conn->executeUpdate('INSERT INTO string_escaped_default_value VALUES()');
$fetched = $this->_conn->executeQuery('SELECT def_string from string_escaped_default_value')
    ->fetchColumn();
$this->assertSame($value, $fetched);
```
it fails as:
```
1) Doctrine\Tests\DBAL\Functional\Schema\MySqlSchemaManagerTest::testEscapedDefaultValueMustBePreserved
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-a\0a\'a"a
+aa'a"a
 a
-a	a\Za\\a
+a	aa\a
```
Which probably means that the actual default value is not what the tests defines and then retrieves.

morozov (changes requested):
@PowerKiKi please see my comments.

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
PowerKiKi: I agree with you, but I don't think I'll be able to test al platforms unfortunately. I might give it a try via Travis though, but that might be cumbersome. I'll see what I can do...

morozov: @PowerKiKi this is the right approach. You don't have to test other platforms yourself. I can run the updated tests on platforms not supported by Travis manually.

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
PowerKiKi: I was under the impression that `getListTableColumnsSQL()` was supposed to return SQL that could be used directly, without any further escaping. And more specifically that they could be compared directly to the value coming from user-defined mapping via `Comparator::diffTable()`. Since that user-defined value must be escaped, the introspected value should also be escaped. That's why I chose to do the escaping in there. 

But maybe we can see the issue from a different point of view and always diff un-escaped values ? so that means that the user-defined value coming from the mapping should be un-escaped ? But since escaping is platform specific where/how should we un-escaped user-defined values ?

morozov: > I was under the impression that `getListTableColumnsSQL()` was supposed to return SQL that could be used directly, without any further escaping.

Designed this way, it would allow stored SQL injections, so no.

> But maybe we can see the issue from a different point of view and always diff un-escaped values?

Yes, the values should be only escaped when building SQL containing them.

> So that means that the user-defined value coming from the mapping should be un-escaped? 

Yes, because escaping is platform-specific.

> But since escaping is platform specific where/how should we un-escaped user-defined values?

Why are user-defined values escaped?

PowerKiKi: > Why are user-defined values escaped?

Because if I want my default value to be `Foo\Bar\Baz`, I will have to escape the `\` and thus writing something like `@ORM\Column(options={"default" = "Foo\\Bar\\Baz"}))`.

I see that for a similar problem with quoted table names, that it was resolved in `AbstractAsset::_setName()` where quotes are detected and removed before comparison.

Should we adopt a similar strategy and modify `Column::setDefault()` to automatically un-escape strings ? If so should we somehow detect which strings need to be un-escaped ? or un-escape all of them and then introducing a breaking change ?

Another approach would be to write the mapping non-escaped, and let Doctrine do the escaping automatically in all cases. But that would directly contradict what is done for table names and would also introduce a breaking change. So I don't think it's the best way to solve that.

The more I think about this, the more I think we won't be able to fix it without introducing a potential breaking change. But maybe there is yet another way... ? Maybe the original way to escape introspected value is less breaking after all ?

morozov: > Because if I want my default value to be `Foo\Bar\Baz`, I will have to escape the `\` and thus writing something like `@ORM\Column(options={"default" = "Foo\\Bar\\Baz"}))`.

It's the escaping required by the PHP string syntax, the value of the default does not contain escaped slashes when used in defining the schema:
```php
echo "Foo\\Bar\\Baz", PHP_EOL;

// Foo\Bar\Baz
```

> I see that for a similar problem with quoted table names, that it was resolved in `AbstractAsset::_setName()` where quotes are detected and removed before comparison.
> 
> Should we adopt a similar strategy and modify `Column::setDefault()` to automatically un-escape strings? If so should we somehow detect which strings need to be un-escaped ? or un-escape all of them and then introducing a breaking change?

I don't know the implementation details of the above, but there shouldn't be any un-escaping in any layer. The un-escaping is only supposed to be done by the parser of the language in whose syntax the escaping is done. Like in my code example, the un-escaping is done by PHP.

> Another approach would be to write the mapping non-escaped, and let Doctrine do the escaping automatically in all cases. But that would directly contradict what is done for table names and would also introduce a breaking change. So I don't think it's the best way to solve that.

This is the only right way how things should work. None values should be escaped anywhere else then before concatenation with SQL/HTML/whatever "program". Furthermore, in the SQL context, if a value is bound to a prepared statement, it must not be escaped because it's not parsed.

PowerKiKi:
> It's the escaping required by the PHP string syntax.

This is incorrect, in my example I use the annotation which are plain text parsed by `doctrine/annotations`, not by PHP parser. So PHP escaping rules do not apply here. As a proof we can compare what happens with a single `\` which are interpreted differently by PHP or MySQL:

```php
echo "Double\\Single\Foo", PHP_EOL; // Single backslash will be preserved by PHP
// Double\Single\Foo
```

```sql
SELECT "Double\\Single\Foo"; -- Single backslash will be removed by MySQL
+------------------+
| Double\SingleFoo |
+------------------+
```

And then running `vendor/bin/doctrine orm:schema-tool:update --dump-sql` we can see that the SQL being executed is exactly as what I wrote. If the mapping is:

```php
/**
 * @ORM\Column(nullable=true, options={"default" = "Double\\Single\Foo"}))
 */
private $name;
```
Then the SQL that will be executed is reported as:

```sql
ALTER TABLE my_table CHANGE name name VARCHAR(255) DEFAULT 'Double\\Single\Foo' NOT NULL;
```

Finally if forcing an update and then checking the actual schema in database with our standard introspection query, we can see that the default is indeed escaped the MySQL way, with the single single dash trimmed:

```sql
SELECT COLUMN_NAME AS Field, COLUMN_DEFAULT AS `Default` FROM information_schema.COLUMNS WHERE TABLE_SCHEMA = database() AND TABLE_NAME = 'my_table';
+-------+------------------+
| Field | Default          |
+-------+------------------+
| id    | NULL             |
| name  | Double\SingleFoo |
+-------+------------------+
```

So we can conclude that **default values are given to MySQL exactly as written in the annotation**, without being affected by PHP escape rules, and without any other form of escaping by doctrine. And we can also clearly confirm that the value in the mapping `Double\\Single\Foo` is not equals to the introspected value `Double\SingleFoo`. So they should not be compared without some kind of escaping/un-escaping beforehand. The question being on what side should we do that extra work ?

One more interesting observation is that `SHOW CREATE TABLE my_table;` will correctly return an escaped value which is ready to be used. That is precisely the behavior I tried to emulate with this PR:

```sql
SHOW CREATE TABLE my_table;
+----------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table    | Create Table                                                                                                                                                                                                             |
+----------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| my_table | CREATE TABLE `my_table` (
  `id` int(11) NOT NULL,
  `name` varchar(255) COLLATE utf8mb4_unicode_ci DEFAULT 'Double\\SingleFoo',
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci |
+----------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

On a sidenote, it is currently possible do SQL injection with an annotation like:

```php
/**
 * @ORM\Column(nullable=true, options={"default" = "'; DROP TABLE my_table; SELECT ''='"}))
 */
private $name;
```

PowerKiKi:
> shouldn't be any un-escaping in any layer

Entirely agree with that. I wouldn't even know how to implement it anyway.

> None values should be escaped anywhere else then before concatenation with SQL/HTML/whatever "program".

Sounds fair to me. So we write annotations non-escaped, and Doctrine will escaped them when building SQL statements. It totally make senses actually.

But it differs from what is done for table names were doctrine never do any escaping/quoting and it is up to end-user to quote the table name if required. Because of that I was not sure if it was the right way to go forward and introduce an inconsistent behavior between table names and default values.

morozov:
> > It's the escaping required by the PHP string syntax.
> 
> This is incorrect, in my example I use the annotation which are plain text parsed by `doctrine/annotations`, not by PHP parser.

@PowerKiKi you're right, thanks for pointing that out.

Currently, the default value is only quoted but not escaped when the column DDL is generated:

https://github.com/doctrine/dbal/blob/80f42dc7afe91781e698e63b7d362b2f51986935/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php#L2280

First, I thought it was done to support default values which are expressions (like `CURRENT_TIMESTAMP()`) but they seem to be taken care of outside of that line. Furthermore, surrounded with quotes, an expression turns into a string literal anyways.

While default values are given to MySQL exactly as written in the annotation, I don't (at least at the moment) see why it's not the bug which should be fixed in the first place. Given that escaping is platform-specific, it's just impossible to create an escaped default value which would be handled correctly by all platforms, so this behavior shouldn't be relied on. And as you pointed out, it allows injecting SQL into annotations.

With that said, could you try adding escaping there and see if it breaks anything? It will definitely break existing metadata which has escaped special characters because the DBAL won't expect them to be escaped anymore.

PowerKiKi:
I just forced push a new patch, which is way cleaner and supports all platform.

I had a look to document the change, but it didn't feel right to add that to http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/schema-representation.html#portable-options. So I suppose it should appear somewhere in UPGRADE.md, but I couldn't figure out where it should go. Any advice about that ?

morozov:
I'd restore your previous test and add the lines I left earlier in comments. The reasons are:

1. Having the default value escaped is not a requirement of the DBAL itself, it's a requirement for integration with a given database platform, so it should be tested in a functional manner.
2. The DBAL is not the source of truth about the way strings should be escaped for a given platform. Thus, the test shouldn't have any assertions against the escaped strings. Otherwise, if the implementation is incorrect, and the test is written after implementation, they will pass even if the implementation is incorrect.
3. What's more important, is that a value specified in the column definition is indeed the value populated by default (covered by https://github.com/doctrine/dbal/pull/2850#discussion_r139862285), and the same value is returned by introspecting the database schema (covered by https://github.com/doctrine/dbal/commit/4dfc5c25a1fdee4002c74e34f4a9b8521a761e98#diff-921db63349be598a376afb1584ccc8b5R192).

>  I had a look to document the change, but it didn't feel right to add that to http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/schema-representation.html#portable-options. So I suppose it should appear somewhere in UPGRADE.md, but I couldn't figure out where it should go. Any advice about that ?

I don't think the documentation needs to be updated as this patch doesn't invalidate the article. As for UPGRADE.md — it will depend on which release this patch gets in. I believe it could be accepted for 2.6 even if it's a minor BC break (the escaped values were not portable anyway). You can draft the notes for 2.6 and we could move them to a different release if needed.

Let's finalize the code changes first to see what exactly the upgrade procedure is.

belgattitude:
@PowerKiKi

Coincidentally, I had to work on that for MariaDB 10.2 support, see #2825. I took the same approach as you did using `quoteStringLiteral()`. I added functional tests and that's where the things became more complex :( 

Note that for MariaDB, I faced a strange issue when backslashing 'single-quotes'... While \' (backslash + single quote) and '' (double single-quotes) should be equivalent (see https://dev.mysql.com/doc/refman/5.7/en/string-literals.html). MariaDB 10.2 silently 'normalize' them when they are stored in information_schema (\' -> '').

So for now #2825 handles backslashes and double-single-quotes and prevent schema diff (important) but keeps limited to MySQL/MariaDB families. 

I will continue to work on this, but we might share some thoughts if you like to. 

PS: See #2824 changed files and search for 'literal'

belgattitude:
@PowerKiKi, 

Thanks for your work, I would love to see automatic quoting of default values ;)

I was just wondering what happen in this scenario:

> If some users have manually quoted their default values:
> ```
> * @ORM\Column(type="string", options={"default" = "O\'Connor"})
> ```
> When doing a schema update, the following change will happen:
>```sql
> ALTER TABLE x CHANGE t t VARCHAR(255) DEFAULT 'O\\''Connor' NOT NULL
> ```
> This is a BC-BREAK, and users needs to update their models like:
> ```
> * @ORM\Column(type="string", options={"default" = "O'Connor"})
> ```

Am I wrong ? or do you have any thought on this ?

belgattitude:
@morozov, no it's not stated in the doc.

But I would be careful with this... I checked some usage of default through github code search:

https://github.com/search?p=1&q=%40ORM%5CColumn%28type%3D%22string%22+options+%22default%22%3A++%5C%5C+language%3APHP+extension%3Aphp&type=Code&utf8=%E2%9C%93

And haven't found any example of people manually backslashing default values yet. But I can imagine some did with things like 'Framework\\Application\\' or even '\'FUNCTION()\''... 

I'm really in favor of escaping default values, but I think it could be nice to add notes in UPGRADE and maybe give some tips:

- A regexp to identify manually quoted defaults
- A note to use 'columnDefinition' if a project needs to support *pre-* and *post-*auto-escaping doctrine versions). 

But for me, all is good. I was just wondering for myself.

belgattitude:
@morozov , please excuse my previous answer. I'll get some rest first ;) and check more deeply later... most of my concerns might have been solved by previous @PowerKiKi answers. Thanks

morozov:
@belgattitude your concerns are perfectly valid. This change will require changes in the depending code if the default values there require escaping and are escaped.

Despite the fact that these value may exist (not only in the form of `@ORM` annotations and not only in public projects), it's unclear to me if a change in non-documented behavior can be considered BC-breaking. Someone more experienced than I am could probably clarify that.

> I'm really in favor of escaping default values, but I think it could be nice to add notes in UPGRADE and maybe give some tips.

It's not really a question of personal preference, it's the only valid solution to this problem. Of course, upgrade notes will help.

> A regexp to identify manually quoted defaults

It may help but it will work only in the case of using DBAL with ORM. Instead, without any additional tools, the user can upgrading and seeing if any unexpected schema changes are generated. It would be a bulletproof solution which doesn't require any knowledge about which characters to look for on our end. Who upgrades dependencies without testing?

> A note to use 'columnDefinition' if a project needs to support pre- and *post-*auto-escaping doctrine versions).

Do you have an example of a project which needs to support multiple DBAL/ORM versions? I have a feeling that most of the packages which depend on DBAL/ORM are applications (not libraries), and thus they lock their dependencies by committing _composer.lock_.

I wouldn't recommend this approach even as a workaround because the value escaped in the definition is not portable.

PowerKiKi:
@belgattitude, I was following your PR for a while and was wondering when our PRs would clash :-)

I believe morozov answered all your questions already. But yes, I think this PR will introduce a "minor"  BC break (as called in UPGRADE.md).

@morozov, I will eventually resume work on this PR, probably in a few days. But right now I need to focus on other priorities.

belgattitude:
@PowerKiKi, clash but very helpful to me

I finally read your description with a clear mind :) and just tested the backslash issue in default. I can confirm the endless schema diff (at least from annotations, don't know for yaml and xml, but let's assume they work the same). 

So to me this change looks is more a bugfix (user perpective: hey doctrine handle my backslashes in defaults without triggering diff, thanks !) and I guess people naturally avoided backslash for this reason. One case still need attention, escape by *repeating* single quotes (i.e. "O''Connor"). Was working perfect but while possible, I doubt there was a lot of use anyway. On the other hand, we don't break any documented feature afaik.

**So more 'bugfix' than 'bc'**. Green for me.

@morozov 

> Do you have an example of a project which needs to support multiple DBAL/ORM versions?

I guess owncloud, nextcloud, sylius... applications/plugins could potentially be affected, I've already partially checked from source if those projects were quoting default values. Nothing yet, but I'll keep you informed next week, I'll install a bunch of them and check information_schema.

belgattitude:
FYI, just tested nextcloud 12 to see if they set default with escape chars... 
(`SELECT * FROM `INFORMATION_SCHEMA`.`COLUMNS` WHERE TABLE_SCHEMA='nextcloud'`) and nothing that will break with your P/R ;)

belgattitude:
@PowerKiKi Belgium is sunny today, I'm going to enjoy it ! I see you're working today, so sorry for you ;) 

Two things that might help you in your weekend quest:

1. Alternative notations (standards)

According to https://dev.mysql.com/doc/refman/5.7/en/string-literals.html#character-escape-sequences and/or https://mariadb.com/kb/en/library/string-literals/. 

> Another way to escape the quoting character is repeating it twice:

Brings the question about other platforms too... Haven't digged into it. But since MariaDB 10.2.7, whatever the notation you use (doubling ') or (\')... The information schema will always store it in the (doubling ') and I fear they might have done it for standardization reasons. This can lead to broken tests with MariaDB > 10.2.7. 

One thing you can do is to add the (doubling ') test, don't care about MariaDB 10.2.7 and when ready I'll normalize on my side (it's already the case: https://github.com/belgattitude/dbal/blob/ad52f401c8d543060eda564f85b772c10dba9cad/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php#L242)


2. SQLServerPlatforms:

 Copy paste (temporarily) the method below the `AbstractPlatformTestCase.php` and run the tests. Haven't digged into, but helped me get a glimpse of what I'll face after covering mysql, postgres and oracle:

```php
    public function testGetDefaultValueDeclarationSQLEscaped()
    {
        $field = [
            'type' => Type::getType('string'),
            'default' => "'O'Connor said: \"Hello\" \ \r'"
        ];

        self::assertSame(sprintf(
                ' DEFAULT %s',
                $this->_platform->quoteStringLiteral("'O'Connor said: \"Hello\" \ \r'")
            ),
            // As demonstration: should be replaced by real expectation behaviour
            // in all platform. Made to illustrate problem with SQLServer
            $this->_platform->getDefaultValueDeclarationSQL($field)
        );
    }

```

Hope it helps and thanks a lot !!!!

belgattitude:
@PowerKiKi  FYI I might be unavailable till tuesday afternoon. But don't hesitate to ask details... I just gave you more problems than solutions ;)

PowerKiKi:
@morozov I re-added the functional test for all platforms, and added support for PostgreSQL and SQLite.

continuousphp is failing, but I don't really understand what is is supposed to do or why it is failing...

Anyway, could you please have a look again at the code ?

belgattitude:
@PowerKiKi Ì could not resist to open my laptop

For the SQLServer issue, you may have luck replacing  `return " DEFAULT '" . $field['default'] . "'";` by `return " DEFAULT ".$this->quoteStringLiteral($field['default']);` in `SQLServerPlatform::getDefaultValueDeclarationSQL($field)` 

Then adding the test in AbstractSQLServerTestCase. 

Never used sql server, but this could help:

1. https://docs.microsoft.com/en-us/sql/integration-services/expressions/numeric-string-and-boolean-literals, look at the string literal section. 
2. https://stackoverflow.com/questions/1586560/how-do-i-escape-a-single-quote-in-sql-server for <doubling '>

Hope it helps.

belgattitude:
@PowerKiKi 

AFAIK, single quote escape by writing two adjacent double quotes is supported on SQL Server, Postgres, Maria, Oracle and Mysql. Could you add tests for it ? (ie. `$default="{ message: ''quoted by adjacent single quotes ''}" `)

PowerKiKi:
@belgattitude I am not willing to install SQLServer to work on that. This will be too much work for me for a platform that I don't really care about. Especially since Travis tests are passing. If SQLServer support is required, then somebody else will have to do it.

I can however include a double single quote in the tested value, to be sure that it works. So, it's done.

belgattitude:
FYI abouot travis failure, it's not related to your P/R, see #2892

belgattitude:
@PowerKiKi ignore my question about functional tests, I should have look deeper. Works great.

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -381,11 +381,12 @@ protected function _getPortableTableColumnDefinition($tableColumn)
                 $length = null;
                 break;
             case 'text':
+            case '_varchar':
+            case 'varchar':
+            $tableColumn['default'] = $this->fixDefaultValueQuotes($tableColumn['default']);
```
belgattitude: Alignment ;)

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
```diff
@@ -907,4 +907,15 @@ public function testListTableForeignKeysSQLEvaluatesDatabase()
         self::assertContains('bar', $sql);
         self::assertNotContains('DATABASE()', $sql);
     }
+
+    public function testGetDefaultValueDeclarationSQLEscaped()
+    {
+        // string must be escaped
+        $field = array(
```
belgattitude: short array syntax. just telling because I got the same ;)

tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
```diff
@@ -919,4 +919,15 @@ public function testQuotesDatabaseNameInCloseActiveDatabaseConnectionsSQL()
             true
         );
     }
+
+    public function testGetDefaultValueDeclarationSQLEscaped()
+    {
+        // string must be escaped
+        $field = array(
```
belgattitude: short array syntax

tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
```diff
@@ -874,4 +874,15 @@ public function testQuotesDatabaseNameInListTableColumnsSQL()
     {
         self::assertContains("'Foo''Bar\\\\'", $this->_platform->getListTableColumnsSQL('foo_table', "Foo'Bar\\"), '', true);
     }
+
+    public function testGetDefaultValueDeclarationSQLEscaped()
+    {
+        // string must be escaped
+        $field = array(
```
belgattitude: short array syntax

morozov:
> I re-added the functional test for all platforms, and added support for PostgreSQL and SQLite.

Please also make sure the `INSERT`/`SELECT` scenario which uses the default value is covered with a functional test (see https://github.com/doctrine/dbal/pull/2850#discussion_r139862285). We want to make sure that the values are not only stored/retrieved from the schema correctly but are also effective as expected.

> Anyway, could you please have a look again at the code ?

As for unescaping on SQLite and PostgreSQL, it looks like instead of the value they return a _SQL expression_ which can be used as a part of another query. Not sure what requirements behind this behavior are but maybe there is a better way of evaluating those values instead of unescaping them in PHP. Some more research/info is needed in this area.

P.S. Do they really return NULL as `null` but strings as escaped literals? With this approach, I'd expect NULL returned as string `NULL` (like MariaDB does) but our code doesn't handle that. Could you please double check and maybe leave some references to the documentation?

belgattitude:
@PowerKiKi ,

> FYI, I finally decided to drop the feature from my PR #2825 in favour of yours.

Some news, PHPunit was updated, so your tests should pass on Travis on your next commit ;) Here's some things that I can share from my attempt:

## Insert/Select

> **From @morozov**: Please also make sure the INSERT/SELECT scenario which uses the default value is covered with a functional test (see #2850 (comment)). We want to make sure that the values are not only stored/retrieved from the schema correctly but are also effective as expected.

I haven't tested and don't know if it helps, but you can add something like:

```php
    public function testInsertEscapedDefaultValueMustBePreserved() {
        $table = new Table("test_insert_escaped_defaults");
        $default = "a\\0b\\'c\"d\te\\Zf\\\\g''h";

        $table->addColumn('test', 'string', ['default' => $default]);
        $this->_sm->dropAndCreateTable($table);

        $this->_conn->executeUpdate("INSERT INTO test_insert_escaped_defaults() VALUES()");
        $row = $this->_conn->fetchAssoc('SELECT test FROM test_insert_escaped_defaults');
        self::assertSame($default, $row['test']);
    }
```

First in `Functional\Schema\MySQLSchemaManagerTest.php` (it should work), then move it into `Functional\Schema\SchemaManagerFunctionalTestCase.php` (I guess there will be problems for some platforms).

## Support for expressions

> From **@morozov**:As for unescaping on SQLite and PostgreSQL, it looks like instead of the value they return a SQL expression which can be used as a part of another query. Not sure what requirements behind this behavior are but maybe there is a better way of evaluating those values instead of unescaping them in PHP. Some more research/info is needed in this area.

@morozov MariaDB 10.2 supports expressions as well (the reason behind the schema changes). While working on that for #2825, my primary reflex was to see if doctrine had an *'Expression'* data type... but I could not find it (I confess I haven't looked thoroughly ;)). Have you heard about some previous work on that ? Is doctrine able to distinguish expressions and defaults values for some platforms yet ? If not, I think escaping backslashes should be safe for now. What do you think ?

morozov:
> While working on that for #2825, my primary reflex was to see if doctrine had an 'Expression' data type... but I could not find it (I confess I haven't looked thoroughly ;)). Have you heard about some previous work on that ? Is doctrine able to distinguish expressions and defaults values for some platforms yet ?

I haven't seen any components you're looking for. I believe this ticket and #2825 have just revealed the need to support expressions returned by some platforms.

> If not, I think escaping backslashes should be safe for now.

You mean escaping backslashes when building DDL? In my understanding, it's not about safety/unsafety but about the semantics of the default attribute of the column in DBAL. As long as the meaning of this "default" attribute is [value](http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/schema-representation.html#portable-options), it must be escaped.

As for fetching the values from the platforms which store expressions in their schema (PostgreSQL, SQLite — according to the changed tests, and MariaDB), we need to:

1. Find some documentation explaining the semantics of the "default" thingy for each platform (whether it's an arbitrary SQL expression or some limited subset of SQL).
2. Based on the above, come up with an abstraction which could be used to handle them in a cross-platform way. For instance, it could be a platform method like `interpretDefault()` which takes whatever value stored in the schema and returns the _value_ which can be compared with what's defined in DBAL.

belgattitude:
@morozov 

> As long as the meaning of this "default" attribute is value, it must be escaped.

Agree.

> @morozov: As for unescaping on SQLite and PostgreSQL, **it looks like instead of the value they return a SQL expression which can be used as a part of another query**. Not sure what requirements behind this behavior are but maybe there is a better way of evaluating those values instead of unescaping them in PHP. Some more research/info is needed in this area.

I could not really find an example of  `which can be used as a part of another query`. Out of curiosity, I've tried to see how it nextval was used in Postgesql:
https://github.com/doctrine/dbal/blob/09e072dee1457bcdf3cae4be60b18854f8829992/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php#L313

Look to me that expressions are not really 'supported', the current code was just grepping some predefined patterns (nextval in this case) and apply a predefined behavior. 
*(As far as I've seen in the current code, only Postgresql seems to use a *trick* to detect an expression (nextval). Other platforms seems to only support 'values'... )*

My personal feeling on this, I would not handle both escaping defaults and supporting expressions in the same P/R, just ensure that code relying on expressions (I could only find postgresql `nextval`) are unescaped (this P/R does not break that)

So to your two questions:

> 1. Find some documentation explaining the semantics of the "default" thingy for each platform (whether it's an arbitrary SQL expression or some limited subset of SQL).
> 2. Based on the above, come up with an abstraction which could be used to handle them in a cross-platform way. For instance, it could be a platform method like interpretDefault() which takes whatever value stored in the schema and returns the value which can be compared with what's defined in DBAL.

Could we imagine handling this in a future P/R ? and ship backslashes escaping before ?

Thanks


Side note: At first sight, the detection of expressions can be done (+/- when text default does not start with a single quote, except null and 'NULL') for MariaDB 10.2 and Postgres). Some expressions are deeply linked to a specific behavior (i.e currently only nextval for postgres), others arbitrary sql (i.e. `concat(first_name, last_name)`, `json_encode(...)`) that will only be processed by the DB. I think we'll need to have an additional type (i.e.: Expression). What I'm not sure is how we'll be able to distinguish expressions and defaults from `options={"default"="cool"},` maybe a new key like 'defaultExpr' ?

morozov:
> I could not really find an example of which can be used as a part of another query.

Yeah. This is in order to distinguish between literals and other expressions (e.g. `CURRENT_TIMESTAMP()`). Not to be used as part of another query.

> At first sight, the detection of expressions can be done (+/- when text default does not start with a single quote, except null and 'NULL') for MariaDB 10.2 and Postgres).

Correct me if I'm wrong, but there's no need to _detect_ expressions. Based on the platform and the use case we will know whether it's an expression or value. For instance, the default column value in MariaDB is always an expression (or `NULL` in some cases).

There has to be a concept of expressions in the DBAL because right now we handle them magically. The starting ideas are:

1. In the DBAL schema, default values may be either values or expressions. Right now default expressions are handled "automagically". Instead, they should be defined explicitly (e.g. `new CurrentTimeStamp()`.
2. Some databases return expressions during schema introspection. Depending on what's defined in the DBAL schema, we should be able to compare expressions to expressions or see if the expression represents a literal and compare it to another literal in the context of a given platform.
3. Schema managers should operate expressions and values as first-class citizens.

belgattitude:
> For instance, the default column value in MariaDB is always an expression (or NULL in some cases).

The default column value in MariaDB 10.2 can be an expression or a value (or null in some cases). MySQL has support for default values and some *magic* `CURRENT_TIMESTAMP...` expressions (they could live with that, time based). When Maria added support for expressions (string, time...), they had to distinguish from default values... They decide to quote values. (like Postgres and I guess Oracle... are doing ?). It looks to me more standard (I'm not expert) and guess they'll have to follow the same path in MySQL. 

> 1. In the DBAL schema, default values may be either values or expressions. Right now default expressions are handled "automagically". Instead, they should be defined explicitly (e.g. new CurrentTimeStamp().

Thanks can be a good start. maybe implementing an `Expression` interface...

> 2. Some databases return expressions during schema introspection. Depending on what's defined in the DBAL schema, we should be able to compare expressions to expressions or see if the expression represents a literal and compare it to another literal in the context of a given platform.

This one I follow less :)

But I think, we go further to the context of this P/R (escaping backslashes)... What do you think if you or me open a separate issue for this and continue the discussion there. If I have directions, I can try a P/R for expressions but I'll need your help :)

Thanks

morozov:
> They decide to quote values.

I think this is a fundamental misconception. I'm sorry if it sounds rude, please feel free to prove me wrong. It's _not_ that they decide to quote values. They move from storing regular values (`(int) 1`, `null`, `(string) fo'o`) and magic values representing expressions (`(string) CURRENT_TIMESTAMP()`) to storing… expressions! Which are all strings `1`, `NULL`, `'fo\'o'`. In a SQL expresssion, you _have to_ quote strings and escape special characters. Is `NULL` a value? Yes. Do they quote it? No. Is `(int) 1` a value? Yes. Do they quote it? No. If it's a `(string) 1`, will it be quoted? Yes.

> This one I follow less :)

I deliberately didn't add any details since it's really a separate project. What I meant is:

1. Assuming, the DBAL schema definition has `(string) foo'bar` as the default value. Assuming, the schema introspection returns `'foo\'bar'` as the default. We know, that this given DB (e.g. MariaDB) returns expressions for default values. In order to see if it's the same value, we do something like:
```php
$actual = new SqlExpression("'foo\'bar'"); // this is what the schema introspection produces
$default = "foo'bar"; // this is what DBAL defines
$expected = (new PortableExpression($default))->toSqlExpression($mariaDb); // we need an expression because it's MariaDB
if (equals($expected, $actual)) { /* the schema is up to date */}
```
What happens internally:
```php
function toPlatformExpression($platform)
{
    return new SqlExpression($platform->quoteString($this->value));
}
```
2. Consider the same setup with MySQL instead of MariaDB:
```php
$actual = "foo'bar"; // this is what the schema introspection produces in MuSQL
$default = "foo'bar"; // this is what DBAL defines — same as previously
if (strcmp($default, $actual)) { /* the schema is up to date */}
```
3. The similar is with "magic" values:
```php
$default = new CurrentTimeStamp(); // this is what DBAL defines 
$actual = new SqlExpression('CURRENT_TIMESTAMP()'); // this is what the schema introspection produces

class CurrentTimeStamp
{
    function toPlatformExpression($platform)
    {
        return $platform->getCurrentTimestampSQL();
    }
}

if (equals($expected, $actual)) { /* the schema is up to date */}
```
With this approach, we won't have to un-escape anything. In the case of expressions, we'll be comparing expressions. In case of values, we'll be comparing values. It will be always apples to apples w/o any magic.

belgattitude:
Good, could you open a separate issue ?

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1192,4 +1192,19 @@ public function testDoesNotListIndexesImplicitlyCreatedByForeignKeys()
         self::assertArrayHasKey('explicit_fk1_idx', $indexes);
         self::assertArrayHasKey('idx_3d6c147fdc58d6c', $indexes);
     }
+
+    public function testEscapedDefaultValueMustBePreserved()
+    {
+        $value = "a\\0b\\'c\"d\te\\Zf\\\\g''h";
```
belgattitude: @PowerKiKi 

If you agree, I propose to test a minimal set of escape sequences in `SchemaManagerFunctionalTestCase`. The current test will fail on SQLServer for example (think about \Z - https://docs.microsoft.com/en-us/sql/integration-services/expressions/numeric-string-and-boolean-literals)

My idea (feel free to implement or not as you think it's best), is to create a method in the base class `SchemaManagerFunctionalTestCase` with those specific escape sequences:

```php
protected function getEscapedLiteralSequence(): string
        return implode('+', [
            "\\'",          // A single quote
            '\\"',          // A double quote
            '\\n',          // A new-line character
            '\\r',          // A carriage return character
            '\\\\',         // A backslash (\) character
        ];
}
```
And change the line above into

```php
   $value = $this->getEscapedLiteralSequence()
```

See my next comment in `AbstractMySQLPlatformTestCase` to test the full MySQL sequences set.

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
```diff
@@ -907,4 +907,15 @@ public function testListTableForeignKeysSQLEvaluatesDatabase()
         self::assertContains('bar', $sql);
         self::assertNotContains('DATABASE()', $sql);
     }
+
+    public function testGetDefaultValueDeclarationSQLEscaped()
```
belgattitude: Following my previous comment in `SchemaManagerFunctionnalTestCase`... it should be now easy to test the full mysql escape sequence:

Override the method:

```
protected function getEscapedLiteralSequence(): string
        return implode('+', [
            "\\0",          // An ASCII NUL (X'00') character
            "\\'", "''",    // Single quote
            '\\"', '""',    // Double quote
            '\\b',          // A backspace character
            '\\n',          // A new-line character
            '\\r',          // A carriage return character
            '\\t',          // A tab character
            '\\Z',          // ASCII 26 (Control+Z)
            '\\\\',         // A backslash (\) character
            '\\%',          // A percent (%) character
            '\\_',          // An underscore (_) character
        ];
}
```

Just an open idea, feel free to do as you like of course. At least this is what works for me in https://github.com/doctrine/dbal/pull/2825/commits/e57dcc0ea65b807366a23603bf1e2c2b1781c988.

PowerKiKi:
I finally had time to rebase, and add functional test via `INSERT` and `SELECT` as requested by @morozov by using some recommendations from @belgattitude. MySQL, PostgreSQL and SQLite should pass.

Could you please have a look again ?

belgattitude:
@PowerKiKi 

Thanks a lot, I've seen you're testing with MariaDB 10.2 too. Nice !

I cannot really give proper feedback for a month or so. Just had a look to the files, at first sight it looks good. I'll add some reviews if I find. Keep up the great work !

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
```diff
@@ -477,4 +463,23 @@ public function testEnsureDefaultsAreUnescapedFromSchemaIntrospection() : void
         $onlineTable = $this->_sm->listTableDetails("test_column_defaults_with_create");
         self::assertSame($default, $onlineTable->getColumn('col1')->getDefault());
     }
+
+    protected function getEscapedLiterals() : array
```
morozov: There should be one test for testing a unified set of special characters on all supported platforms. Testing different sets on different platforms doesn't guarantee that the API is portable.

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
```diff
@@ -477,4 +463,23 @@ public function testEnsureDefaultsAreUnescapedFromSchemaIntrospection() : void
...
+    {
+        return [
+            "\\0", // An ASCII NUL (X'00') character
```
morozov: This looks like a good candidate for a data provider where comments are array keys.

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
```diff
@@ -454,21 +454,7 @@ public function testEnsureDefaultsAreUnescapedFromSchemaIntrospection() : void
         $platform = $this->_sm->getDatabasePlatform();
```
morozov: This entire test should go to `SchemaManagerFunctionalTestCase`.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1360,4 +1360,39 @@ public function commentsProvider() : array
...
+        $table = new Table('string_escaped_default_value');
+        $table->addColumn('def_string', 'string', array('default' => $value));
+        $table->addColumn('def_foo', 'string');
+        $this->_sm->dropAndCreateTable($table);
+
+        $onlineTable = $this->_sm->listTableDetails('string_escaped_default_value');
+        self::assertSame($value, $onlineTable->getColumn('def_string')->getDefault(), 'should be able introspect the value of default');
+
+        $comparator = new Comparator();
+        self::assertFalse($comparator->diffTable($table, $onlineTable), 'introspected schema should have no differences');
```
morozov: Given that this test covers default values and they have been compared in the previous assertion, this assertion seems redundant (as well as the usage of `Comparator` itself). If the default value is introspected correctly, the comparison may fail for whatever irrelevant reason.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1360,4 +1360,39 @@ public function commentsProvider() : array
...
+
+        $this->_conn->insert('string_escaped_default_value', array('def_foo' => 'foo'));
+        $row = $this->_conn->fetchAssoc('SELECT def_string FROM string_escaped_default_value');
+        self::assertSame($value, $row['def_string'], 'inserted default value should be the configured default value');
```
morozov: I'd move this assertion to a different test case. If the introspection assertion fails, it's still worth checking if the value was properly deployed. These requirements are independent.

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
morozov: I'd remove all `testGetDefaultValueDeclarationSQLEscaped()` tests in favor of `testEscapedDefaultValueMustBePreserved()` which will run the same for all platforms. Escaping is part of the SQL syntax whose validity can only be checked against a real DB.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1360,4 +1360,39 @@ public function commentsProvider() : array
...
+        ];
+    }
+
+    public function testEscapedDefaultValueMustBePreserved() : void
+    {
+        $value = implode('+', $this->getEscapedLiterals());
+
+        $table = new Table('string_escaped_default_value');
+        $table->addColumn('def_string', 'string', array('default' => $value));
+        $table->addColumn('def_foo', 'string');
```
morozov: This column doesn't seem needed for the test. Inserting an empty array should work.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2306,7 +2306,7 @@ public function getDefaultValueDeclarationSQL($field)
             return " DEFAULT '" . $this->convertBooleans($default) . "'";
         }
 
-        return " DEFAULT '" . $default . "'";
+        return " DEFAULT " . $this->quoteDefaultStringLiteral($default);
```
morozov: Given that `quoteStringLiteral()` only escapes quotes and backslashes by means of `str_replace()`, it may not work with encodings where the `0x27` byte (the single quote in ASCII) is part of a multi-byte sequence.

This approach to escaping is used everywhere on the platform level, so it'd be good to hear from other maintainers whether this approach is valid in general.

/cc @Ocramius @lcobucci

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -1236,4 +1236,12 @@ private function isNumericType(Type $type) : bool
     {
         return $type instanceof IntegerType || $type instanceof BigIntType;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected function quoteDefaultStringLiteral(string $str) : string
+    {
+        return parent::quoteStringLiteral($str);
```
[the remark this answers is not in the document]
PowerKiKi: Because [PostgreSQL does not support backlash C-style escapes out of the box](https://www.postgresql.org/docs/9.6/static/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS-ESCAPE), the `quoteStringLiteral()` had to be overridden a long time ago. See this quick example:

```
db=# SELECT 'foo\bar', 'foo\\bar', E'foo\\bar';
 ?column? | ?column? | ?column? 
----------+----------+----------
 foo\bar  | foo\\bar | foo\bar
(1 row)
```

This means we cannot use `PostgreSqlPlatform::quoteStringLiteral()` for our use-case. And this is the only reason why we created a new `quoteDefaultStringLiteral()`, to be able to override it to "cancel" the original override.

I wish it was not needed, but I don't say another way around...

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
PowerKiKi: I didn't pay enough attention to it. I now realize that it is a duplicate of `SchemaManagerFunctionalTestCase::testEscapedDefaultValueMustBePreserved()`. I'll remove this one.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
PowerKiKi: Unfortunately inserting an empty array does not work, at the very least on PostgreSQL and possibly on other platforms. While this most likely should be fixed, I think it is out of the scope of this PR.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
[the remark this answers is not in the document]
PowerKiKi: scratch that last comment of mine, I'll think a bit more about that...

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
PowerKiKi: Fixed.

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
PowerKiKi: Fixed.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
PowerKiKi: I would still argue to keep it, because this is the actual end goal of this PR, that a comparator sees the introspected schema and in-memory schema as equal. Keeping this assert ensure that there will never be any regression in whatever may be happening in `diffTable()`.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
PowerKiKi: I feel this would force us to duplicate code for test setup, and double the time needed (table creation is expensive), while those three assert are very closely related together semantically speaking. Are you sure you want to take the performance hit ?

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
PowerKiKi: The classes were `testGetDefaultValueDeclarationSQLEscaped()` appear are already full of tests against quoting/escaping. While I agree that it should ultimately be tested against a real database (which it is), I still see value in those kind of tests, because they are way faster to run, and thus debug, than functional tests. And since it does not seem to go against the trend of the project I'd leave them there.

PowerKiKi:
I think you can give it another round of review.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
morozov: It doesn't have to be a performance hit. On setup, you can create one table with 13 fields representing different defaults and create one record with all default values. The first test would go over fields one by one and introspect them. The second test would select each field and see if the value inserted by default is correct. Each failure will clearly indicate the case where something went wrong.

tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
```diff
@@ -976,4 +976,15 @@ public function testQuotesDatabaseNameInCloseActiveDatabaseConnectionsSQL()
...
+
+    public function testGetDefaultValueDeclarationSQLEscaped() : void
+    {
+        // string must be escaped
+        $field = [
+            'type' => 'string',
+            'default' => 'Foo\\Bar'
+        ];
+
+        self::assertSame(" DEFAULT 'Foo\\Bar'", $this->_platform->getDefaultValueDeclarationSQL($field));
```
morozov: Unlike the functional tests, this one covers one special character out of 13.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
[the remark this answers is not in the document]
morozov: The fact that `quoteStringLiteral()` is not overridden may mean that it's not tested well enough. You could try testing other (non-DDL) cases and see if those tests fail as well.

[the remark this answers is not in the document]
PowerKiKi: I don't understand what you mean by that. `AbstractPlatform::quoteStringLiteral` is indeed overridden by `PostgreSqlPlatform::quoteStringLiteral`. But we intentionally use the non-overridden method in our code. And that is covered by tests.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
PowerKiKi: Test removed.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
PowerKiKi: Is it correct that this going against [what you said earlier](https://github.com/doctrine/dbal/pull/2850#discussion_r155142197) and `getEscapedLiterals()` should not be a dataProvider anymore ? But instead have a private method like `createTableForDefaultValues()` which is called in two different test cases `testIntrospectDefaultValues()` and `testInsertDefaultValues` ?

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
PowerKiKi: Tests removed.

tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
PowerKiKi: Fixed by removing.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
PowerKiKi: Fixed too.

PowerKiKi:
If you can have one more look today... or next year :tada:

Ocramius:
@PowerKiKi there's one remaining conflict in a test

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
Ocramius: This seems to only affect schema definitions, so I'm not too worried about it for now. It is a problem if people rely on it for anything runtime-related.

Ocramius (approved):

PowerKiKi:
@Ocramius, I just rebased everything on master. Would you like me to squash also ?

Ocramius:
@PowerKiKi please don't squash. Besides that, we have a couple CS failures/nits, see https://travis-ci.org/doctrine/dbal/jobs/323796326#L523

Ocramius:
Still one CS issue detected, sorry @PowerKiKi

PowerKiKi:
No worries, I should have checked more carefully.

BTW is there any reason why you don't use `phpcbf` to fix the code style of existing code once and for all ? or PHP-CS-Fixer for that matter ? I feel it would make it easier to check for code style locally (without the need of `git-phpcs`) and thus reduce the time spent (yours and mine) on those silly issues.

PhpSpreadsheet use [composer scripts](https://github.com/PHPOffice/PhpSpreadsheet/blob/develop/composer.json#L25-L34) to easily check and fix code style. I found it very convenient to use.

lcobucci:
@PowerKiKi php supported and phpcbf are properly configured but we have many violations that need to be fixed and we decided to fix them in the develop branch, in order to reduce the amount of conflicts in the existing contributions.

Ocramius:
:ship: thanks for the patience here, @PowerKiKi!

PowerKiKi:
Thanks for the merge, and happy new year :tada:

Ocramius:
Seems like one failure slipped through on oci8: https://continuousphp.com/git-hub/doctrine/dbal/build/46f18bd9-51d3-4d68-a132-fea0b189058f/output

PowerKiKi:
It seems to be related to Oracle, and I am really not familiar with that system. I am afraid I won't be able to help with that.

Ocramius:
That's alright, but we may need to revert the merge until we found the culprit.

Majkl578:
I see this was merged into master, shouldn't it go only to develop since it's BC break?
Also wondering if this could somehow be related to #2926.

Ocramius:
@Majkl578 no, 2.7 is fine, as this can really only cause some issues with the schema tool (and those happen at every upgrade)

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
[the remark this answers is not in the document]
morozov: @PowerKiKi I don't remember what I meant by that :confused:.

morozov:
@PowerKiKi please take a look at the [failures on Oracle](https://app.continuousphp.com/git-hub/doctrine/dbal/build/9d05fb86-47cc-453a-ac6a-4168888350f7):
```xml
<testcase name="testEscapedDefaultValueCanBeIntrospected" class="Doctrine\Tests\DBAL\Functional\Schema\OracleSchemaManagerTest" classname="Doctrine.Tests.DBAL.Functional.Schema.OracleSchemaManagerTest" file="/home/cphp/data/container/9d05fb86-47cc-453a-ac6a-4168888350f7/workspace/tests/Doctrine/Tests/DBAL/Functional/Schema/OracleSchemaManagerTest.php" line="1443" assertions="1" time="0.883744">
    <failure type="PHPUnit\Framework\ExpectationFailedException">
Doctrine\Tests\DBAL\Functional\Schema\OracleSchemaManagerTest::testEscapedDefaultValueCanBeIntrospected should be able introspect the value of default for: An ASCII NUL (X'00') Failed asserting that two strings are identical. --- Expected +++ Actual @@ @@ -'foo\0bar' +'foo\\0bar' /home/cphp/data/container/9d05fb86-47cc-453a-ac6a-4168888350f7/workspace/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php:1449
    </failure>
</testcase>
<testcase name="testEscapedDefaultValueCanBeInserted" class="Doctrine\Tests\DBAL\Functional\Schema\OracleSchemaManagerTest" classname="Doctrine.Tests.DBAL.Functional.Schema.OracleSchemaManagerTest" file="/home/cphp/data/container/9d05fb86-47cc-453a-ac6a-4168888350f7/workspace/tests/Doctrine/Tests/DBAL/Functional/Schema/OracleSchemaManagerTest.php" line="1453" assertions="0" time="0.586455">
    <error type="Exception">
Doctrine\Tests\DBAL\Functional\Schema\OracleSchemaManagerTest::testEscapedDefaultValueCanBeInserted Exception: [PHPUnit\Framework\Error\Notice] Undefined index: field0 With queries: 6. SQL: 'SELECT * FROM sys.user_tables' Params: 5. SQL: 'SELECT * FROM string_escaped_default_value' Params: 4. SQL: 'INSERT INTO string_escaped_default_value (def_foo) VALUES (?)' Params: 'foo' 3. SQL: 'CREATE TABLE string_escaped_default_value (field0 VARCHAR2(255) DEFAULT 'foo\\0bar' NOT NULL, field1 VARCHAR2(255) DEFAULT 'foo\\''bar' NOT NULL, field2 VARCHAR2(255) DEFAULT 'foo''''bar' NOT NULL, field3 VARCHAR2(255) DEFAULT 'foo\\"bar' NOT NULL, field4 VARCHAR2(255) DEFAULT 'foo""bar' NOT NULL, field5 VARCHAR2(255) DEFAULT 'foo\\bbar' NOT NULL, field6 VARCHAR2(255) DEFAULT 'foo\\nbar' NOT NULL, field7 VARCHAR2(255) DEFAULT 'foo\\rbar' NOT NULL, field8 VARCHAR2(255) DEFAULT 'foo\\tbar' NOT NULL, field9 VARCHAR2(255) DEFAULT 'foo\\Zbar' NOT NULL, field10 VARCHAR2(255) DEFAULT 'foo\\\\bar' NOT NULL, field11 VARCHAR2(255) DEFAULT 'foo\\%bar' NOT NULL, field12 VARCHAR2(255) DEFAULT 'foo\\_bar' NOT NULL, def_foo VARCHAR2(255) NOT NULL)' Params: 2. SQL: 'DROP TABLE string_escaped_default_value' Params: 1. SQL: 'DROP TRIGGER STRING_ESCAPED_DEFAULT_VALUE_AI_PK' Params: Trace: /home/cphp/data/container/9d05fb86-47cc-453a-ac6a-4168888350f7/workspace/vendor/symfony/phpunit-bridge/DeprecationErrorHandler.php:105 /home/cphp/data/container/9d05fb86-47cc-453a-ac6a-4168888350f7/workspace/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php:1460 /home/cphp/data/container/9d05fb86-47cc-453a-ac6a-4168888350f7/workspace/vendor/phpunit/phpunit/src/Framework/TestCase.php:1071 /home/cphp/data/container/9d05fb86-47cc-453a-ac6a-4168888350f7/workspace/vendor/phpunit/phpunit/src/Framework/TestCase.php:939 /home/cphp/data/container/9d05fb86-47cc-453a-ac6a-4168888350f7/workspace/vendor/phpunit/phpunit/src/Framework/TestResult.php:698 /home/cphp/data/container/9d05fb86-47cc-453a-ac6a-4168888350f7/workspace/vendor/phpunit/phpunit/src/Framework/TestCase.php:894 /home/cphp/data/container/9d05fb86-47cc-453a-ac6a-4168888350f7/workspace/vendor/phpunit/phpunit/src/Framework/TestSuite.php:755 /home/cphp/data/container/9d05fb86-47cc-453a-ac6a-4168888350f7/workspace/vendor/phpunit/phpunit/src/Framework/TestSuite.php:755 /home/cphp/data/container/9d05fb86-47cc-453a-ac6a-4168888350f7/workspace/vendor/phpunit/phpunit/src/TextUI/TestRunner.php:546 /home/cphp/data/container/9d05fb86-47cc-453a-ac6a-4168888350f7/workspace/vendor/phpunit/phpunit/src/TextUI/Command.php:195 /home/cphp/data/container/9d05fb86-47cc-453a-ac6a-4168888350f7/workspace/vendor/phpunit/phpunit/src/TextUI/Command.php:148 /home/cphp/data/container/9d05fb86-47cc-453a-ac6a-4168888350f7/workspace/vendor/phpunit/phpunit/phpunit:53 /home/cphp/data/container/9d05fb86-47cc-453a-ac6a-4168888350f7/workspace/tests/Doctrine/Tests/DbalFunctionalTestCase.php:88
    </error>
</testcase>
```

morozov:
Looks like it's not just Oracle failures. The changes haven't been tested properly.

1. The test fixtures do not contain lower ASCII range characters as they claim. For instance,

https://github.com/doctrine/dbal/blob/e3ab127672fd72e5e64a17f20b11ab2d12b8bfc5/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php#L1413

creates a `\` followed by a `0` character, not a NULL byte (`\0`). [Fixing](https://github.com/doctrine/dbal/pull/2960) these fixtures introduces [failures](https://travis-ci.org/doctrine/dbal/jobs/324761622#L642) on SQLite where escaping new line characters [needs some more love](https://stackoverflow.com/a/47228906/146187).

2. Fixing failures on Oracle requires implementing more logic in the Oracle's schema manager _and_ reverting [some code](https://github.com/doctrine/dbal/pull/2442/files#diff-a3ae60269159d98094d5946d4d4a5ff0R1168) introduced by #2442. Here, we have a disagreement between tests. The unit ones claim backslashes should be escaped while the functional ones (as well as manual testing) shows they shouldn't. Additional research is needed.

The work in progress is tracked in #2960.

Ocramius:
@morozov should I revert here or just wait?

morozov:
Let’s revert it. Not sure if I’m going to be able to spend enough time on this shortly. And te failures are annoying.

Ocramius:
Reverted in a5d3e33c4cb46e3b81f1224dbdb376328560be6a for now

PR #2854, opened 2017-09-11
Enable strict types

Majkl578:


Ocramius:
Loads of string casts to be added 😅

Majkl578:
The build is green now. :tada:
There may be some discrepancies here and there, we should find them during #3348.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1499,7 +1501,7 @@ private function getBindingInfo($value, $type)
             $value       = $type->convertToDatabaseValue($value, $this->getDatabasePlatform());
             $bindingType = $type->getBindingType();
         } else {
-            $bindingType = $type;
+            $bindingType = $type ?? ParameterType::NULL;
```
morozov: Should we just not specify any type in this case? The less DBAL implies, the better.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
```diff
@@ -48,17 +50,17 @@ class MysqliConnection implements Connection, PingableConnection, ServerInfoAwar
      */
     public function __construct(array $params, $username, $password, array $driverOptions = [])
     {
-        $port = $params['port'] ?? ini_get('mysqli.default_port');
+        $port = (int) ($params['port'] ?? ini_get('mysqli.default_port'));
```
morozov: Let's only cast the value from `ini_get()`. If the explicitly specified port parameter is non-int, let it fail. `??` could be probably replaced with `?:`.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
```diff
@@ -48,17 +50,17 @@ class MysqliConnection implements Connection, PingableConnection, ServerInfoAwar
...
         if (! $port) {
             $port = 3306;
         }
 
         $socket = $params['unix_socket'] ?? ini_get('mysqli.default_socket');
-        $dbname = $params['dbname'] ?? null;
+        $dbname = $params['dbname'] ?? '';
 
-        $flags = $driverOptions[static::OPTION_FLAGS] ?? null;
+        $flags = (int) ($driverOptions[static::OPTION_FLAGS] ?? 0);
```
morozov: The flags are a bitmask of `MYSQLI_*` constants. Shouldn't need any casting.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Exception.php
```diff
@@ -17,6 +19,6 @@ public static function fromErrorInfo($error)
             return new self('Database error occurred but no error information was retrieved from the driver.');
         }
 
-        return new self($error['message'], null, $error['code']);
+        return new self($error['message'] ?? 'An unknown error occurred.', null, $error['code'] ?? 0);
```
morozov: Is it really possible that the error is not `false` but doesn't have a message or code?

lib/Doctrine/DBAL/Schema/AbstractAsset.php
```diff
@@ -166,7 +168,7 @@ public function getName()
             return $this->_namespace . '.' . $this->_name;
         }
 
-        return $this->_name;
+        return $this->_name ?? '';
```
morozov: Would it become much worse if we make the method return a nullable string? Not sure if assets with empty names exist.

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
```diff
@@ -98,7 +100,7 @@ protected function _getPortableTableIndexesList($tableIndexes, $tableName = null
             $keyName = strtolower($tableIndex['name']);
             $buffer  = [];
 
-            if (strtolower($tableIndex['is_primary']) === 'p') {
+            if ($tableIndex['is_primary'] !== null && strtolower($tableIndex['is_primary']) === 'p') {
```
morozov: Most likely, it's just `'P'`.

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
```diff
@@ -137,7 +139,7 @@ protected function _getPortableTableColumnDefinition($tableColumn)
         }
 
         // Default values returned from database sometimes have trailing spaces.
-        $tableColumn['data_default'] = trim($tableColumn['data_default']);
+        $tableColumn['data_default'] = trim((string) $tableColumn['data_default']);
```
morozov: Looks like this is redundant. The value is checked against null and trimmed again a few lines later.

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -323,17 +325,17 @@ protected function _getPortableTableColumnDefinition($tableColumn)
         $matches = [];
 
         $autoincrement = false;
-        if (preg_match("/^nextval\('(.*)'(::.*)?\)$/", $tableColumn['default'], $matches)) {
+        if (preg_match("/^nextval\('(.*)'(::.*)?\)$/", (string) $tableColumn['default'], $matches)) {
```
morozov: If it's not a string, it won't match.

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -323,17 +325,17 @@ protected function _getPortableTableColumnDefinition($tableColumn)
...
             $tableColumn['sequence'] = $matches[1];
             $tableColumn['default']  = null;
             $autoincrement           = true;
         }
 
-        if (preg_match("/^['(](.*)[')]::.*$/", $tableColumn['default'], $matches)) {
+        if (preg_match("/^['(](.*)[')]::.*$/", (string) $tableColumn['default'], $matches)) {
```
morozov: If it's not a string, it won't match.

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -323,17 +325,17 @@ protected function _getPortableTableColumnDefinition($tableColumn)
...
             $tableColumn['default'] = $matches[1];
         }
 
-        if (stripos($tableColumn['default'], 'NULL') === 0) {
+        if (stripos((string) $tableColumn['default'], 'NULL') === 0) {
```
morozov: If it's not a string, it won't match.

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -355,7 +357,7 @@ protected function _getPortableTableColumnDefinition($tableColumn)
         $jsonb     = null;
 
         $dbType = strtolower($tableColumn['type']);
-        if (strlen($tableColumn['domain_type']) && ! $this->_platform->hasDoctrineTypeMappingFor($tableColumn['type'])) {
+        if (strlen((string) $tableColumn['domain_type']) && ! $this->_platform->hasDoctrineTypeMappingFor($tableColumn['type'])) {
```
morozov: A non-string would not pass this check. No need to cast and then check.

tests/Doctrine/Tests/DBAL/Driver/AbstractDriverTest.php
```diff
@@ -269,7 +271,7 @@ private function getExceptionConversions()
 
                     public function __construct($errorCode, $sqlState, $message)
                     {
-                        parent::__construct($message);
+                        parent::__construct($message ?? 'unknown error');
```
morozov: An empty string would be enough since it's the default.

tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
```diff
@@ -400,7 +402,7 @@ public function testJsonbColumn(string $type) : void
         $columns = $this->schemaManager->listTableColumns('test_jsonb');
 
         self::assertSame($type, $columns['foo']->getType()->getName());
-        self::assertTrue(true, $columns['foo']->getPlatformOption('jsonb'));
```
morozov: Nice.

tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php
```diff
@@ -318,14 +319,14 @@ public function testGeneratesPrimaryKeyDeclarationSQL()
         self::assertEquals(
             'CONSTRAINT pk PRIMARY KEY CLUSTERED (a, b)',
             $this->platform->getPrimaryKeyDeclarationSQL(
-                new Index(null, ['a', 'b'], true, true, ['clustered']),
+                new Index('', ['a', 'b'], true, true, ['clustered']),
```
morozov: Another way may be:

https://github.com/doctrine/dbal/blob/92920ede68e1d16246712cd41318dd82bb72c015/lib/Doctrine/DBAL/Schema/ForeignKeyConstraint.php#L67-L69

lib/Doctrine/DBAL/Driver/OCI8/OCI8Exception.php
Majkl578: IIRC the issue was with message being null.

tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
Ocramius: BWAHAHAHAHAHA

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
Majkl578: How about when the value comes i.e. from XML?

tests/Doctrine/Tests/DBAL/Driver/OCI8/OCI8StatementTest.php
```diff
@@ -57,11 +60,16 @@ public function testExecute(array $params)
             ->disableOriginalConstructor()
             ->getMock();
         $conn->expects($this->once())
-            ->method('getExecuteMode');
+            ->method('getExecuteMode')
+            ->willReturn(OCI_NO_AUTO_COMMIT);
```
Ocramius: Constant to be imported?

lib/Doctrine/DBAL/Types/ConversionException.php
```diff
@@ -21,14 +25,14 @@ class ConversionException extends DBALException
...
-     * @param string $value
+     * @param mixed  $value
      * @param string $toType
      *
      * @return \Doctrine\DBAL\Types\ConversionException
      */
     public static function conversionFailed($value, $toType)
     {
-        $value = strlen($value) > 32 ? substr($value, 0, 20) . '...' : $value;
+        $value = is_string($value) && strlen($value) > 32 ? substr($value, 0, 20) . '...' : var_export($value, true);
```
Ocramius: I'm not sure if we can `var_export()` all the time: this can easily go out of memory if somebody passes a complex object to the type conversion.

lib/Doctrine/DBAL/Types/ConversionException.php
```diff
@@ -37,15 +41,15 @@ public static function conversionFailed($value, $toType)
...
+     * @param mixed  $value
      * @param string $toType
      * @param string $expectedFormat
      *
      * @return \Doctrine\DBAL\Types\ConversionException
      */
     public static function conversionFailedFormat($value, $toType, $expectedFormat, ?Throwable $previous = null)
     {
-        $value = strlen($value) > 32 ? substr($value, 0, 20) . '...' : $value;
+        $value = is_string($value) && strlen($value) > 32 ? substr($value, 0, 20) . '...' : var_export($value, true);
```
Ocramius: I'm not sure if we can `var_export()` all the time: this can easily go out of memory if somebody passes a complex object to the type conversion.

lib/Doctrine/DBAL/Schema/AbstractAsset.php
Ocramius: I'd say that an asset needs to have a name or throw an exception.

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -79,10 +82,15 @@ public function bindValue($param, $value, $type = ParameterType::STRING)
      */
     public function bindParam($column, &$variable, $type = ParameterType::STRING, $length = null, $driverOptions = null)
     {
-        $type = $this->convertParamType($type);
+        $type            = $this->convertParamType($type);
+        $extraParameters = array_slice(func_get_args(), 3);
+
+        if (count($extraParameters) !== 0) {
+            $extraParameters[0] = $extraParameters[0] ?? 0;
```
Ocramius: This is very convoluted/hard to read. You are replacing the 4th parameter with `0` if it is set and it is `null`? What happens if `$extraParameters[0]` is always defined by us (instead of doing the `count()` check)?

lib/Doctrine/DBAL/Driver/PDOConnection.php
```diff
@@ -88,7 +90,7 @@ public function query(string $sql) : ResultStatement
      */
     public function quote($input, $type = ParameterType::STRING)
     {
-        return $this->connection->quote($input, $type);
+        return $this->connection->quote((string) $input, $type);
```
Ocramius: Wait, PDO doesn't specify a type for `$input`? :|

Ocramius (changes requested):
Some minor discussion points on code that needed change, but otherwise looking really good!

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
Ocramius: That's up to the caller to fix, not the callee. The definition of `$params['port']` requires an `int`, although that isn't expressed in the docblock (`mixed[]`).

We could improve that with further declarations...

Ocramius:
One thing I didn't mention yet: this should have an `UPGRADE.md` in which we instruct people to run something like Psalm or PHPStan to detect type errors...

lib/Doctrine/DBAL/Schema/AbstractAsset.php
Majkl578: @morozov If we change getName() to return ?string we immediately get 70 PHPStan errors on other places. This is because basically all code assumes this method returns string (and there is e.g. a lot of string concatenation happening).

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
Majkl578: It's trimmed with other characters than whitespace. I'm changing this to be guarded with `!== null` condition instead`.

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
Majkl578: changing to `!== null`

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
Majkl578: changing to `!== null`

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
Majkl578: changing to `!== null`

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
Majkl578: changing to `!== null`

lib/Doctrine/DBAL/Driver/PDOStatement.php
Majkl578: This comes usually as always-defined, defaulting to null, due to:
https://github.com/doctrine/dbal/blob/e39bc0859fbf6e9bf78e30c4dd3329515c6deccc/lib/Doctrine/DBAL/Statement.php#L110-L129

lib/Doctrine/DBAL/Driver/PDOConnection.php
Majkl578: It comes as mixed in here, but quote() expects string: http://php.net/manual/en/pdo.quote.php

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
Majkl578: I have no Oracle instances to test against so I changed this to `!== 'P'` && `!== 'p'`.

lib/Doctrine/DBAL/Schema/AbstractAsset.php
morozov: > I'd say that an asset needs to have a name or throw an exception.

For some types of assets, it's fine to not have a name. E.g. indices are identified by their definition, not by name.

> If we change getName() to return ?string we immediately get 70 PHPStan errors on other places.

Let's leave it as is for now and revisit later :+1:

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
morozov: On [wnameless/oracle-xe-11g](https://hub.docker.com/r/wnameless/oracle-xe-11g), 
```sql
SELECT DISTINCT constraint_type FROM user_constraints

-- +---------------+
-- |CONSTRAINT_TYPE|
-- +---------------+
-- |U              |
-- |R              |
-- |P              |
-- |C              |
-- |O              |
-- +---------------+
```

lib/Doctrine/DBAL/Types/ConversionException.php
morozov: It's also useless from the end user perspective. Maybe just displaying type/FQN of the value is enough.

Ocramius: A test is probably to be added here. Unsure if relevant, but https://3v4l.org/EYdiW could be included to just check that it prints `stdClass`, not much more.

morozov:
> […] something like Psalm or PHPStan to detect type errors...

We could only recommend Psalm if our own codebase passes its test. Otherwise, there will be lots of false-positives. What do you think about adopting Psalm/Phan in our CI? At a glance, it looks like a lot more work :-D

Majkl578:
I am skeptic towards Psalm, last time I saw it it required a lot of Psalm-specific annotations and stuff. :/

Ocramius:
The end users only need to check the contact surface with DBAL, not DBAL itself, so any of these tools would guarantee decent safety. Psalm is way stricter than PHPStan, so that's up for discussion, but it is capable of much more interesting type additions: https://medium.com/vimeo-engineering-blog/uncovering-php-bugs-with-template-a4ca46eb9aeb

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
Majkl578: Removed and changed on caller part.

lib/Doctrine/DBAL/Driver/PDOStatement.php
Ocramius: So is `if (count($extraParameters) !== 0) {` needed at all?

Majkl578: Someone still may call this API directly, I guess.

Ocramius: Yes, I understand that, but the code you provided works with/without `count() !== 0` check, no?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Exception.php
morozov: It's `null` because it's mocked, e.g. in `OCI8StatementTest::testExecute()`. `OCI8Statement::errorInfo()` replaces `false` from the driver with `[]` for DBAL API compliance and therefore cannot be used as input for `OCI8Exception::fromErrorInfo()` which expects `false` as the indicator of no error.

Please try replacing all occurrences of `throw OCI8Exception::fromErrorInfo($this->errorInfo())` in `OCI8Statement` with `throw OCI8Exception::fromErrorInfo(oci_error($this->_sth))`.

I'd like to see `Statement::errorInfo()` and `::errorCode()` gone from the API entirely in the future.

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
```diff
@@ -98,7 +100,7 @@ protected function _getPortableTableIndexesList($tableIndexes, $tableName = null
             $keyName = strtolower($tableIndex['name']);
             $buffer  = [];
 
-            if (strtolower($tableIndex['is_primary']) === 'p') {
+            if ($tableIndex['is_primary'] === 'P' || $tableIndex['is_primary'] === 'p') {
```
morozov: Just `'P'` will be enough. On [wnameless/oracle-xe-11g](https://hub.docker.com/r/wnameless/oracle-xe-11g), 
```sql
SELECT DISTINCT constraint_type FROM user_constraints

-- +---------------+
-- |CONSTRAINT_TYPE|
-- +---------------+
-- |U              |
-- |R              |
-- |P              |
-- |C              |
-- |O              |
-- +---------------+
```

lib/Doctrine/DBAL/Driver/PDOConnection.php
morozov: Thinking out loud: should we restrict it to `string` as well? What's the point in quoting/escaping anything else than a string?

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -913,7 +915,7 @@ static function ($boolean) {
      */
     public function convertFromBoolean($item)
     {
-        if (in_array(strtolower($item), $this->booleanLiterals['false'], true)) {
+        if (in_array(strtolower((string) $item), $this->booleanLiterals['false'], true)) {
```
morozov: Can it really be in upper case? Looks like it [can't](https://www.postgresql.org/docs/9.1/datatype-boolean.html). I'd better remove `(string)` and `strtolower()` altogether. If it's really needed, we'll have an integration test for that case.

lib/Doctrine/DBAL/Driver/PDOConnection.php
Majkl578: I think it makes sense, but not in this PR - in #3348.

lib/Doctrine/DBAL/Driver/PDOStatement.php
Majkl578: No, only when the parameter was really explicitly passed: https://3v4l.org/KORCg
This method is really dirty.

lib/Doctrine/DBAL/Types/ConversionException.php
Majkl578: I have removed var_export(), leaving only is_string() check. The code is still errorneous as it concatenates mixed with string, but let's address that in other PR.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Exception.php
Majkl578: The problem is that oci_error() returns array|false...

lib/Doctrine/DBAL/Connection.php
Majkl578: Probably, but it's out of scope and will require some refactoring (results of this method are used as list() now).

morozov: As far as I know, a common default for a binding type is "string", not "null". Binding as "null" by default may have side effects.

It's a private method, so there shouldn't be much refactoring. In general, while working on the latest refactoring issues, I found most of the methods returning a list really smelling. Why instead of returning some unstructured set of values it doesn't just _do_ the binding?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Exception.php
Majkl578: Another issue is that some calls to `oci_error()` are in context when no _dbh_ is available, so suggested replacement won't work. :|

morozov: > The problem is that oci_error() returns array|false...

It's not a problem, this is exactly what I'm looking for.

> Another issue is that some calls to `oci_error()` are in context when no _dbh_ is available, so suggested replacement won't work. :|

It workes just fine. As long `$_sth` is `protected`, it's accessible from the mock, and `oci_error()` doesn't complain about getting `null` instead of a resource. It happily returns `false`.
```diff
--- a/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
+++ b/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
@@ -298,7 +298,7 @@ class OCI8Statement implements IteratorAggregate, Statement
             $length ?? -1,
             $this->convertParameterType($type)
         )) {
-            throw OCI8Exception::fromErrorInfo($this->errorInfo());
+            throw OCI8Exception::fromErrorInfo(oci_error($this->_sth));
         }
     }
 
@@ -390,7 +390,7 @@ class OCI8Statement implements IteratorAggregate, Statement
 
         $ret = @oci_execute($this->_sth, $this->_conn->getExecuteMode());
         if (! $ret) {
-            throw OCI8Exception::fromErrorInfo($this->errorInfo());
+            throw OCI8Exception::fromErrorInfo(oci_error($this->_sth));
         }
 
         $this->result = true;
```

phpcs.xml.dist
```diff
@@ -65,4 +64,10 @@
     <rule ref="SlevomatCodingStandard.Operators.DisallowEqualOperators.DisallowedNotEqualOperator">
         <exclude-pattern>lib/Doctrine/DBAL/Schema/Comparator.php</exclude-pattern>
     </rule>
+
+    <!-- db2_autocommit() expects bool, but DB2_AUTOCOMMIT_ON/DB2_AUTOCOMMIT_OFF are of int type
```
morozov: It's the opposite: `DB2_AUTOCOMMIT_ON` and `DB2_AUTOCOMMIT_OFF` are of `int` type but `db2_autocommit()` expects `bool` (https://github.com/php/pecl-database-ibm_db2/pull/8#discussion_r258274427).

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
```diff
@@ -273,7 +275,7 @@ public function testGeneratesForeignKeySqlOnlyWhenSupportingForeignKeys()
         }
     }
 
-    protected function getBitAndComparisonExpressionSql($value1, $value2)
```
morozov: I miss the PHP 7.1 strictness.

lib/Doctrine/DBAL/Driver/PDOStatement.php
morozov: I don't really see how we can improve here. Unlike the userland optional arguments defaulting to some value, the internal arguments are counted, so we have to respect the expected number of arguments being passed and their types.

morozov (commented):
Apart from the issue in `OCI8Exception`, I don't see anything else worth reworking.

lib/Doctrine/DBAL/Driver/PDOStatement.php
morozov: Another approach might be to declare `int $length = 0` since on the driver level there's no `null` anyways. But it's fine if we do it later since it's an API change.

phpcs.xml.dist
Majkl578: I think that's exactly what's written in the comment. 🤔

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
Majkl578: I don't understand, I added typehints.

phpcs.xml.dist
morozov: From the original comment, it follows that the constants are defined incorrectly because they are after "but", however, the bug is in the method definition. Or maybe it's just me? Feel free to leave as is if you think it's good enough.

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
morozov: Sorry. The change is correct. I was saying that this issue wouldn't be possible on PHP 7.1 which doesn't allow loosening argument types in subclasses.

phpcs.xml.dist
Majkl578: Reworded. 👍

morozov (approved):

lib/Doctrine/DBAL/Driver/PDOStatement.php
morozov: @Ocramius unless you still have objections and can propose a better approach, I'm going to merge this.

Ocramius (approved):

lib/Doctrine/DBAL/Driver/PDOStatement.php
Ocramius: Good to go 👍

PR #2863, opened 2017-09-15
Fixed unknown database type int4range and int8range requested

mesutaslan:
```
[Doctrine\DBAL\DBALException]
  Unknown database type int4range requested, Doctrine\DBAL\Platforms\PostgreSQL92Platform may not support it.
```

It's fixed.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -1134,6 +1134,7 @@ class PostgreSqlPlatform extends AbstractPlatform
             'year'          => 'date',
             'uuid'          => 'guid',
             'bytea'         => 'blob',
+            'int4range'     => 'integer',
         ];
     }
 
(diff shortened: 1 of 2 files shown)


Ocramius:
@mesutaslan this is missing int8range and similar ranges though, plus integration tests.

mesutaslan:
@Ocramius I'm added int8range.

mesutaslan:
@Ocramius test is completed.

morozov:
@mesutaslan a new set of integration tests is needed where a schema containing fields of the newly introduced types is correctly recognized. Something like a case when the exception in description happened.

mesutaslan:
@morozov and @Ocramius I added test for this fix.

morozov:
@mesutaslan as @Ocramius said it should be an integration test (i.e. creating an actual table in DB and introspecting it).

winkbrace:
I wonder if this solution is technically correct. An int4range is stored as string '[3,10)' which couldn't map to an integer.

If you actually want to work with range types, I believe there is a package for that, but to just get rid of the warning when trying to run `schema update` isn't it better to map to string?

PR #2864, opened 2017-09-16
Removed the "OCI8: SQL Queries with Question Marks" section

morozov:
The [OCI8: SQL Queries with Question Marks](http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/known-vendor-issues.html#oci8-sql-queries-with-question-marks) looks confusing since the shown example doesn't cause a problem. At least on `2.5` and `2.6`, the question mark inside a literal is parsed correctly and is not replaced with a named placeholder.

Currently, there are no known bugs with parsing literals, so I'd better prefer seeing them reported and fixed instead of just documented as a potentially problematic area.

Otherwise, having it documented may prevent people from reporting these bugs because it looks like it's by design.


Ocramius (approved):
LGTM :ship:

PR #2871, opened 2017-09-28
Replace double quotes when the schema is named as keyword

huang-yi:
If some schemas are named as keywords such as `user`, `order`. 
The result of SQL: `select setting from pg_catalog.pg_settings where name = 'search_path'` will be like this.

```
+----------------------------------------------+
| setting                                      |
+----------------------------------------------+
| "$user", public, "user", "order"             |
+----------------------------------------------+
1 row in set
```

All the keywords will be quoted by `""`, so we should replace it.


Ocramius:
This change requires a test to back the change and to prevent regressions.

PR #2877, opened 2017-10-04
Fixed the code style changes

vanbrabantf:
This PR tries to apply the code style applied in the Jenkins code style sniffer.
I'm personally not a big fan of the code style chosen, but that is a different discussion.

This is a combination of the auto-fixer with manual changes for more critical files.

I know this is a very big PR with a ton of changes. And I fully understand you would be very hesitant to merge this. No hard feelings.

PS: I'm very new to open-source. If there are commit guidelines I'm breaking here, please let me know (couldn't find documentation on that)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,11 @@ vendor/
 composer.lock
 /phpunit.xml
 /.phpcs-cache
+
+\.idea/dbal\.iml
+
+\.idea/modules\.xml
+
+\.idea/vcs\.xml
+
+\.idea/workspace\.xml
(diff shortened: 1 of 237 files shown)


vanbrabantf:
The `There must be a single space before a NOT operator` rule still needs to be applied.
The problem there in lies that split up lines look extremely bad.

lcobucci:
@vanbrabantf thanks for sending this PR but our plan is to fix the code standard after we reduce the amount of PRs we have (or maybe only for DBAL v3 - `develop`), therefore I'm going to close this PR as `won't fix`.

vanbrabantf:
No problem. A very sane decision.

PR #2879, opened 2017-10-05
 adding ORDER BY to OVER() on IBMDB2

guidofaecke:
Finished up the missing ORDER BY within OVER() on IBMDB2.  
Might not be the best and most elegant approach, but that's how it works for us.
I am sure many IBM shops would appreciate it.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/DB2Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/DB2Platform.php
@@ -894,10 +894,12 @@ class DB2Platform extends AbstractPlatform
         $orderByArray = [];
 
         //determine if 'ORDER BY' is part of the query
-        $orderByPosition = strrpos($query, 'ORDER BY');
+        $orderByFound = preg_match('/(\bORDER\b)(\s*)(\bBY\b)(?!.*\b\1\b)/i', $query, $orderByPositionArray, PREG_OFFSET_CAPTURE);
 
         // ORDER BY found in query string
-        if (false !== $orderByPosition) {
+        if (0 !== $orderByFound) {
+            $orderByPosition = $orderByPositionArray[0][1];
+
             $queryArray = preg_split('/[, ]/', substr($query, 0, $orderByPosition -1));
             $orderByArray = preg_split('/[, ]/', substr($query, $orderByPosition));
 
@@ -912,8 +914,8 @@ class DB2Platform extends AbstractPlatform
                         $orderByArray[$orderIndex] = ',';
                         break;
                     default:
-                        $orderByArray[$orderIndex] = array_search($orderValue, $queryArray) === false 
-                                                     ? $orderValue 
+                        $orderByArray[$orderIndex] = array_search($orderValue, $queryArray) === false
+                                                     ? $orderValue
                                                      : $queryArray[array_search($orderValue, $queryArray)+2];
                         break;
                 }
--- a/lib/Doctrine/DBAL/Platforms/DB2Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/DB2Platform.php
@@ -900,37 +900,39 @@ class DB2Platform extends AbstractPlatform
         //determine if 'ORDER BY' is part of the query
         $orderByFound = preg_match('/(\bORDER\b)(\s*)(\bBY\b)(?!.*\b\1\b)/i', $query, $orderByPositionArray, PREG_OFFSET_CAPTURE);
 
-        // ORDER BY found in query string
-        if (0 !== $orderByFound) {
-            $orderByPosition = $orderByPositionArray[0][1];
-
-            $queryArray = preg_split('/[, ]/', substr($query, 0, $orderByPosition -1));
-            $orderByArray = explode(',', substr($query, $orderByPosition + strlen('ORDER BY')));
-
-            foreach ($orderByArray as $orderIndex => $orderValue) {
-                $splitOrder = explode(' ', $orderValue);
-
-                foreach ($splitOrder as $splitIndex => $splitValue) {
-                    switch (strtoupper($splitValue)) {
-                        case 'ASC':
-                        case 'DESC':
-                            break;
-                        default:
-                            $splitOrder[$splitIndex] = array_search($splitValue, $queryArray) === false
-                                                       ? $splitValue
-                                                       : $queryArray[array_search($splitValue, $queryArray)+2];
-                            break;
-                    }
-                }
-                $orderByArray[$orderIndex] = array_filter($splitOrder);
-            }
-
-            foreach ($orderByArray as $orderIndex => $orderValue) {
-                $orderByArray[$orderIndex] = implode(' ', $orderValue);
-            }
-
-            $orderByArray[0] = 'ORDER BY ' . $orderByArray[0];
+        // early return if ORDER BY not found in query string
+        if (0 === $orderByFound) {
+            return '';
         }
+        
+        $orderByPosition = $orderByPositionArray[0][1];
+
+        $queryArray = preg_split('/[, ]/', substr($query, 0, $orderByPosition -1));
+        $orderByArray = explode(',', substr($query, $orderByPosition + strlen('ORDER BY')));
+
+        foreach ($orderByArray as $orderIndex => $orderValue) {
+            $splitOrder = explode(' ', $orderValue);
+
+            foreach ($splitOrder as $splitIndex => $splitValue) {
+                switch (strtoupper($splitValue)) {
+                    case 'ASC':
+                    case 'DESC':
+                        break;
+                    default:
+                        $splitOrder[$splitIndex] = array_search($splitValue, $queryArray) === false
+                                                   ? $splitValue
+                                                   : $queryArray[array_search($splitValue, $queryArray)+2];
+                        break;
+                }
+            }
+            $orderByArray[$orderIndex] = array_filter($splitOrder);
+        }
+
+        foreach ($orderByArray as $orderIndex => $orderValue) {
+            $orderByArray[$orderIndex] = implode(' ', $orderValue);
+        }
+
+        $orderByArray[0] = 'ORDER BY ' . $orderByArray[0];
 
         return implode(',', $oderByArray;
     }
--- a/lib/Doctrine/DBAL/Platforms/DB2Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/DB2Platform.php
@@ -934,6 +934,6 @@ class DB2Platform extends AbstractPlatform
 
         $orderByArray[0] = 'ORDER BY ' . $orderByArray[0];
 
-        return implode(',', $oderByArray;
+        return implode(',', $oderByArray);
     }
 }
--- a/lib/Doctrine/DBAL/Platforms/DB2Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/DB2Platform.php
@@ -785,34 +785,8 @@ class DB2Platform extends AbstractPlatform
             return $query;
         }
 
-        //preset empty array
-        $orderByArray = [];
-
-        //determine if 'ORDER BY' is part of the query
-        $orderByPosition = strrpos($query, 'ORDER BY');
-
-        // ORDER BY found in query string
-        if (false !== $orderByPosition) {
-            $queryArray = preg_split('/[, ]/', $query);
-            $orderByArray = preg_split('/[, ]/', substr($query, $orderByPosition));
-
-            foreach ($orderByArray as $orderIndex => $orderValue) {
-                switch (strtoupper($orderValue)) {
-                    case 'ORDER':
-                    case 'BY':
-                    case 'ASC':
-                    case 'DESC':
-                        break;
-                    case '':
-                        $orderByArray[$orderIndex] = ',';
-                        break;
-                    default:
-                        $orderByArray[$orderIndex] = $queryArray[array_search($orderValue, $queryArray)+2];
-                        break;
-                }
-            }
-        }
-
+        $orderByArray = $this->getOrderByForOver($query);
+        
         return sprintf(
             'SELECT db22.* FROM (SELECT db21.*, ROW_NUMBER() OVER(%s) AS DC_ROWNUM FROM (%s) db21) db22 WHERE %s',
             implode(' ', $orderByArray),
@@ -905,5 +879,45 @@ class DB2Platform extends AbstractPlatform
     protected function getReservedKeywordsClass()
     {
         return Keywords\DB2Keywords::class;
+    }
+    
+    /**
+     * Prepare ORDER BY string for OVER() if applicable
+     * 
+     * @param  string $query
+     * 
+     * @return array|string[]|mixed[]
+     */
+    private function getOrderByForOver($query)
+    {
+        //preset empty array
+        $orderByArray = [];
+
+        //determine if 'ORDER BY' is part of the query
+        $orderByPosition = strrpos($query, 'ORDER BY');
+
+        // ORDER BY found in query string
+        if (false !== $orderByPosition) {
+            $queryArray = preg_split('/[, ]/', $query);
+            $orderByArray = preg_split('/[, ]/', substr($query, $orderByPosition));
+
+            foreach ($orderByArray as $orderIndex => $orderValue) {
+                switch (strtoupper($orderValue)) {
+                    case 'ORDER':
+                    case 'BY':
+                    case 'ASC':
+                    case 'DESC':
+                        break;
+                    case '':
+                        $orderByArray[$orderIndex] = ',';
+                        break;
+                    default:
+                        $orderByArray[$orderIndex] = $queryArray[array_search($orderValue, $queryArray)+2];
+                        break;
+                }
+            }
+        }
+
+        return $orderByArray;
     }
 }
--- a/lib/Doctrine/DBAL/Platforms/DB2Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/DB2Platform.php
@@ -785,6 +785,7 @@ class DB2Platform extends AbstractPlatform
             return $query;
         }
 
+        // retrieve ORDER BY string
         $orderBy = $this->getOrderByForOver($query);
         
         return sprintf(
--- a/lib/Doctrine/DBAL/Platforms/DB2Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/DB2Platform.php
@@ -891,9 +891,6 @@ class DB2Platform extends AbstractPlatform
      */
     private function getOrderByForOver(string $query): string
     {
-        //preset empty array
-        $orderByArray = [];
-
         //determine if 'ORDER BY' is part of the query
         $orderByPosition = strripos($query, 'order by');
 
@@ -913,7 +910,11 @@ class DB2Platform extends AbstractPlatform
                                   substr($query, 0, $orderByPosition -1))
                               , function($element) {
                                     // don't return 'AS' and empty elements
-                                    return (strtoupper($element) !== 'AS' && trim($element !== ''));
+                                    return (
+                                        strtoupper($element) !== 'AS'
+                                        && trim($element !== '')
+                                        && $element !== false
+                                    );
                                 }
                           )
         );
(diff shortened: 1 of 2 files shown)


AwakendSleeper2016:
This looks solid enough.  With luck, this will get merged in.  Without it, doctrine can't be used to its fullest in IBM/db2 shops.

Ocramius:
This lacks testing though

guidofaecke:
@Ocramius You caught the missing tests, didn't you 😄

morozov (commented):
@guidofaecke is it possible to create a functional test which describes the value/purpose of this change? (which I'm not questioning). The existing test does describe what the resulting queries should look like but doesn't tell why.

It would be also nice to have a more complex case covered with multiple columns and `ASC`|`DESC` involved. Otherwise, removing corresponding parts of the implementation will not make the new test fail.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -878,4 +880,44 @@ protected function getReservedKeywordsClass()
...
+     * 
+     * @return array|string[]|mixed[]
+     */
+    private function getOrderByForOver($query)
+    {
+        //preset empty array
+        $orderByArray = [];
+
+        //determine if 'ORDER BY' is part of the query
+        $orderByPosition = strrpos($query, 'ORDER BY');
```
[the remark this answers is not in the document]
guidofaecke: Good point... let me get that fixed

tests/Doctrine/Tests/DBAL/Platforms/DB2PlatformTest.php
```diff
@@ -377,6 +377,33 @@ public function testModifiesLimitQuery()
...
+        );
+
+        self::assertEquals(
+            'SELECT db22.* FROM (SELECT db21.*, ROW_NUMBER() OVER(ORDER BY id) AS DC_ROWNUM FROM (SELECT * FROM user) db21) db22 WHERE db22.DC_ROWNUM <= 10',
+            $this->_platform->modifyLimitQuery('SELECT * FROM user ORDER BY id', 10, 0)
+        );
+
+        self::assertEquals(
+            'SELECT db22.* FROM (SELECT db21.*, ROW_NUMBER() OVER(ORDER BY id) AS DC_ROWNUM FROM (SELECT * FROM user) db21) db22 WHERE db22.DC_ROWNUM <= 10',
+            $this->_platform->modifyLimitQuery('SELECT * FROM user', 10)
```
morozov: Doesn't this case cover the same execution path as the previous one? The implementation checks `if ($offset > 0)`, so there's no difference between `0` and `NULL`. And from the SQL standpoint, no offset means "start from zero".

tests/Doctrine/Tests/DBAL/Platforms/DB2PlatformTest.php
```diff
@@ -377,6 +377,33 @@ public function testModifiesLimitQuery()
         );
     }
 
+    public function testModifiesLimitQueryWithOrderBy()
+    {
+        self::assertEquals(
+            'SELECT * FROM user ORDER BY id',
+            $this->_platform->modifyLimitQuery('SELECT * FROM user ORDER BY id', null, null)
+        );
```
morozov: This test could use `@dataProvider` to make assertions independent on each other.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -878,4 +880,46 @@ protected function getReservedKeywordsClass()
...
+
+            foreach ($orderByArray as $orderIndex => $orderValue) {
+                switch (strtoupper($orderValue)) {
+                    case 'ORDER':
+                    case 'BY':
+                    case 'ASC':
+                    case 'DESC':
+                        break;
+                    case '':
+                        $orderByArray[$orderIndex] = ',';
```
morozov: With this approach and the `preg_split()` expression above, if the `ORDER BY` columns have any delimiter than exactly ", ", there's a risk of getting malformed SQL. For example, if there's a leading white space(s) before the comma:
```php
var_dump(preg_split('/[, ]/', 'foo , bar , baz'));

/*
array(7) {
  [0] =>
  string(3) "foo"
  [1] =>
  string(0) ""
  [2] =>
  string(0) ""
  [3] =>
  string(3) "bar"
  [4] =>
  string(0) ""
  [5] =>
  string(0) ""
  [6] =>
  string(3) "baz"
}
*/
```
In this case, the resulting SQL will have more than one comma between the columns.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -878,4 +880,46 @@ protected function getReservedKeywordsClass()
...
+    private function getOrderByForOver($query)
+    {
+        //preset empty array
+        $orderByArray = [];
+
+        //determine if 'ORDER BY' is part of the query
+        $orderByPosition = strrpos($query, 'ORDER BY');
+
+        // ORDER BY found in query string
+        if (false !== $orderByPosition) {
```
morozov: If there's no `ORDER BY`, the method can use early return to avoid the extra indentation below.

tests/Doctrine/Tests/DBAL/Platforms/DB2PlatformTest.php
guidofaecke: I copied form the test case above and added the order by

morozov: I think this test shouldn't focus on testing of different variations of offset and limit since its logic doesn't depend on those, but instead, focus on edge cases of the `ORDER BY` expression.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
guidofaecke: ...and another fix I need to bang my head against

tests/Doctrine/Tests/DBAL/Platforms/DB2PlatformTest.php
guidofaecke: Good point... Will see what I can come with.
Everyone else is more than welcome to chime in 😄

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -878,4 +880,64 @@ protected function getReservedKeywordsClass()
     {
         return Keywords\DB2Keywords::class;
     }
+    
+    /**
+     * Prepare ORDER BY string for OVER() if applicable
+     * 
+     * @param  string $query
+     * 
+     * @return array|string[]|mixed[]
```
morozov: This doesn't look usable. Why doesn't it always return a string?

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -878,4 +880,64 @@ protected function getReservedKeywordsClass()
...
+            $queryArray = preg_split('/[, ]/', substr($query, 0, $orderByPosition -1));
+            $orderByArray = explode(',', substr($query, $orderByPosition));
+
+            foreach ($orderByArray as $orderIndex => $orderValue) {
+                $splitOrder = explode(' ', $orderValue);
+
+                foreach ($splitOrder as $splitIndex => $splitValue) {
+                    switch (strtoupper($splitValue)) {
+                        case 'ORDER':
+                        case 'BY':
```
morozov: Why not take the leading `ORDER BY` out of the parsing? It's already captured by the regex in the beginning, so we only need to parse the actual set of columns.

Also, the `ASC` and `DESC` can be only in the end of the column, so instead of splitting the column in words, can we just `preg_replace()` the trailing `ASC`|`DESC` with an empty string?

I believe the closer the parsing logic to the actual SQL syntax is, the more reliable it is in edge cases.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -878,4 +880,64 @@ protected function getReservedKeywordsClass()
...
+                            $splitOrder[$splitIndex] = '';
+                            break;
+                        case 'ASC':
+                        case 'DESC':
+                            break;
+                        case '':
```
morozov: I think this can be improved. There shouldn't be an empty token. The fact that we capture an empty token means that the delimiter doesn't capture enough characters.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
guidofaecke: Good question... Let me fix that...

tests/Doctrine/Tests/DBAL/Platforms/DB2PlatformTest.php
morozov: Since the regexp matches the last `ORDER BY` expression in the string (as far as I understand, this is to ignore sub-queries), there should be a test covering a query with a sub-query.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -878,4 +881,57 @@ protected function getReservedKeywordsClass()
...
+    private function getOrderByForOver($query)
+    {
+        //preset empty array
+        $orderByArray = [];
+
+        // remove extra white space
+        $query = preg_replace('/\s+/', ' ', $query);
+
+        //determine if 'ORDER BY' is part of the query
+        $orderByFound = preg_match('/(\bORDER\b)(\s*)(\bBY\b)(?!.*\b\1\b)/i', $query, $orderByPositionArray, PREG_OFFSET_CAPTURE);
```
morozov: I believe it's better to use `preg_match_all()` and use the last match. Otherwise, a query like `ORDER BY my_order` will be probably ignored.

morozov: Aren't `\b`'s around `\s` redundant (`\b)(\s*)(\b`)? Could it be just `\s+`?

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -878,4 +881,57 @@ protected function getReservedKeywordsClass()
...
+
+        // ORDER BY found in query string
+        if (0 !== $orderByFound) {
+            $orderByPosition = $orderByPositionArray[0][1];
+
+            $queryArray = preg_split('/[, ]/', substr($query, 0, $orderByPosition -1));
+            $orderByArray = explode(',', substr($query, $orderByPosition + strlen('ORDER BY')));
```
morozov: `strlen('ORDER BY')` will probably corrupt the query if there's more than one space between `ORDER` and `BY` in the original query.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -878,4 +881,57 @@ protected function getReservedKeywordsClass()
...
     }
+    
+    /**
+     * Prepare ORDER BY string for OVER() if applicable
+     * 
+     * @param  string $query
+     * 
+     * @return string
+     */
+    private function getOrderByForOver($query)
```
morozov: Since the DBAL requires PHP ^7.1, we should use scalar type hints and return types where applicable.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -878,4 +881,57 @@ protected function getReservedKeywordsClass()
...
+        $orderByArray = [];
+
+        // remove extra white space
+        $query = preg_replace('/\s+/', ' ', $query);
+
+        //determine if 'ORDER BY' is part of the query
+        $orderByFound = preg_match('/(\bORDER\b)(\s*)(\bBY\b)(?!.*\b\1\b)/i', $query, $orderByPositionArray, PREG_OFFSET_CAPTURE);
+
+        // ORDER BY found in query string
+        if (0 !== $orderByFound) {
```
morozov: Please use an early return.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -878,4 +881,57 @@ protected function getReservedKeywordsClass()
...
+     * 
+     * @return string
+     */
+    private function getOrderByForOver($query)
+    {
+        //preset empty array
+        $orderByArray = [];
+
+        // remove extra white space
+        $query = preg_replace('/\s+/', ' ', $query);
```
morozov: This can (or even should) be replaced with a proper `preg_split()` expression once the `ORDER BY` part is detected. No need to process the entire query.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -878,4 +881,57 @@ protected function getReservedKeywordsClass()
...
+
+        //determine if 'ORDER BY' is part of the query
+        $orderByFound = preg_match('/(\bORDER\b)(\s*)(\bBY\b)(?!.*\b\1\b)/i', $query, $orderByPositionArray, PREG_OFFSET_CAPTURE);
+
+        // ORDER BY found in query string
+        if (0 !== $orderByFound) {
+            $orderByPosition = $orderByPositionArray[0][1];
+
+            $queryArray = preg_split('/[, ]/', substr($query, 0, $orderByPosition -1));
```
morozov: The delimiter should expect any number of any valid whitespaces before and after the comma.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -878,4 +881,57 @@ protected function getReservedKeywordsClass()
...
+
+                foreach ($splitOrder as $splitIndex => $splitValue) {
+                    switch (strtoupper($splitValue)) {
+                        case 'ASC':
+                        case 'DESC':
+                            break;
+                        default:
+                            $splitOrder[$splitIndex] = array_search($splitValue, $queryArray) === false
+                                                       ? $splitValue
+                                                       : $queryArray[array_search($splitValue, $queryArray)+2];
```
morozov: Could you pls add some tests explaining this logic? As I read it, the code tries to find the same column in `SELECT` and if fails, uses the expression as is, otherwise, uses something else (the `+ 2` part). The variable names `$splitOrder` (which is an array of strings) and `$splitValue` look confusing. Is it possible to rename it using some SQL terms?

Not sure if this code properly handles the case when a selected column has an alias (with or without the `AS` keyword) like `SELECT foo foo_alias, bar AS bar_alias`. Not sure what the expected result in this case is. This should be covered by a test too.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -878,4 +881,57 @@ protected function getReservedKeywordsClass()
...
+                            break;
+                    }
+                }
+                $orderByArray[$orderIndex] = array_filter($splitOrder);
```
morozov: What are the expected empty values in this case? Should `'0'` be filtered too? Looks like it can be fixed by a proper replacement of the `explode(' ')` above.

morozov (changes requested):

tests/Doctrine/Tests/DBAL/Platforms/DB2PlatformTest.php
guidofaecke: The test will be there as soon as I cover more edge cases.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
guidofaecke: I think line 898 should take care of the whitespace, which should simplify the whole preg_split and explode madness.

All I tried to do is help with the IBM issue, but you are worth in code review than I am! btw. That's a compliment 😆 

Anyway, maybe with some help we can get than show on the road; Alan, Clark... you are more than welcome in joining me...

lib/Doctrine/DBAL/Platforms/DB2Platform.php
guidofaecke: I was hoping someone would say that!
Based on the current code, without checking composer.json, I wasn't sure if I should apply that.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
guidofaecke: In here I try to eliminate dead whitespace.
However, if it bugs you, I will reconsider.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
guidofaecke: My bad. 
I was coding by talking myself through it.
You are absolutely right.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
guidofaecke: ...bag to the whitespace issue... That's why 898 is there for tight now.. Looks like I should tweak it a little.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
guidofaecke: The logic... DB2 (on IBMi) is kinda strict on certain statements...
The reason for that, and I still not to address the possibility of subqueries here, is -> by default, if I use the entity and slap a limit/offset on it, it always adds a "as xxx" to it. within the "OVER()" you have to sort by the "AS" declaration otherwise DB2 can't handle it.
However!, I have to see how IBM reacts to your statement; without the "AS"...
Dang... I thought that would be an easy ride, like he persistent connection I added to ZF2 for IBMi... LOL

lib/Doctrine/DBAL/Platforms/DB2Platform.php
guidofaecke: not 100% sure yet, the filtered values are primarily "" (empty string).
Might have to investigate more...

guidofaecke:
@morozov Are you in Vegas in October (ZendCon)? We should connect!

lib/Doctrine/DBAL/Platforms/DB2Platform.php
guidofaecke: Just tying to address possible whitespaces between ORDER and BY

tests/Doctrine/Tests/DBAL/Platforms/DB2PlatformTest.php
guidofaecke: Did I mention I don't like "edge cases" 🤣 
Can someone introduce more parts to this class, so we can access certain items before the SQL string is build?

lib/Doctrine/DBAL/Platforms/DB2Platform.php
morozov: You're right, `my_order` will not match `\bORDER\b`.

guidofaecke: This one can stay? Just checking!

tests/Doctrine/Tests/DBAL/Platforms/DB2PlatformTest.php
morozov: Before we proceed with the parser, could you explain what exactly a non-empty `OVER()` is needed in this case? As far as I understand, if it's empty, then the row number assigned in the order specified by `ORDER BY` which is exactly what's needed for implementing `LIMIT, OFFSET`. Doesn't it work in all cases, or it works better with a non-empty `OVER()`.

morozov:
> Are you in Vegas in October (ZendCon)? We should connect!

Not really. Maybe some other conference next year.

tests/Doctrine/Tests/DBAL/Platforms/DB2PlatformTest.php
guidofaecke: On IBMDB2 everything is a little bit more complicated (very unfortunate). The row_number is assigned based on what's in OVER(). If OVER() is empty, the row_number is based on the relative-record-number of the table. The result of your regular order by would not look like what you are expecting if used with row_number.
If you need, as in most cases, the row_number assignment based on your order_by, you need the order in OVER() as well.
Either way, I agree, the current test-cases don't make much sense.

morozov: @guidofaecke thanks for the clarification. With the above said and the fact that the row number in this query is used for emulating `LIMIT`, let's start with a test case in which a selected subset of the rows doesn't match the expectation.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
guidofaecke: Also, it simplifies the processing a little bit along the line.

guidofaecke:
Closed in favor of #3016

PR #2883, opened 2017-10-09
Update OraclePlatform.php

userator:
old formats were incorrect and throw Doctrine\DBAL\Types\ConversionException


Deltachaos:
Thats not correct. Please have a look at https://stackoverflow.com/questions/8168161/how-to-use-doctrine-oraclesessioninit-listener-with-symfony2

Issue #2881, opened 2017-10-06
Doctrine\DBAL\Schema\SqliteSchemaManager: "preg_match(): Compilation failed: missing ) at offset 163"


Bruce17:
Hi,

I'm using Symfony with the [LiipFunctionalTestBundle](https://github.com/liip/LiipFunctionalTestBundle) to mock my database and create a local sqlite database mock. 

Used bundles:
- doctrine/doctrine-bundle: v1.6.3
- doctrine/dbal: v2.5.4

One of my entitie's column name is `FOO#` (legacy database), which causes strange errors in https://github.com/doctrine/dbal/blob/v2.6.2/lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php#L465.

Error:
`preg_match(): Compilation failed: missing ) at offset 163`

Seems the regex is missing a delimiter. Wrapping the regex into `/` fixes my issue.

Fixed code:
```php
private function parseColumnCollationFromSQL($column, $sql)
    {
        if (preg_match(
            '/{(?:'.preg_quote($column).'|'.preg_quote($this->_platform->quoteSingleIdentifier($column)).')
            [^,(]+(?:\([^()]+\)[^,]*)?
            (?:(?:DEFAULT|CHECK)\s*(?:\(.*?\))?[^,]*)*
            COLLATE\s+["\']?([^\s,"\')]+)}isx/', $sql, $match)) {
            return $match[1];
        }

        return false;
    }
```

morozov:
@Bruce17 the original expression has `{` and `}` as delimiters. The version you provided puts the `isx` flags inside the regex which changes its logic. I believe the problem is that the call to `preg_quote()` doesn't have the second parameter passed, and `#` gets escaped (or something like that).

Could you create a pull request first so that the patch could be automatically tested? Also, there should be a functional test which creates a table with the column in question and calls `SqliteSchemaManager::listTableColumns()` on the schema. I believe this is the lowest possible level to test this regex.

Bruce17:
Seems that `#` is also a special character in php regex, but gets not escaped by `preg_quote`. I prepared a pull request including a fix and updated the unit tests.

Majkl578:
> Seems that # is also a special character in php regex, but gets not escaped by preg_quote directly.

I reported this to PHP directly as this is something it should be taken care of by PCRE itself.
https://bugs.php.net/bug.php?id=75355

morozov:
[...]

PR #2884, opened 2017-10-10
Escape "#" in regex in SqliteSchemaManager

Bruce17:
This pull request escapes `#` in sqlite db column names.

See issue https://github.com/doctrine/dbal/issues/2881 for more information.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
@@ -427,7 +427,7 @@ class SqliteSchemaManager extends AbstractSchemaManager
     private function parseColumnCollationFromSQL($column, $sql)
     {
         if (preg_match(
-            '/(?:'.preg_quote($column).'|'.preg_quote($this->_platform->quoteSingleIdentifier($column)).')
+            '/(?:'.$this->pregQuote($column).'|'.$this->pregQuote($this->_platform->quoteSingleIdentifier($column)).')
                 [^,(]+(?:\([^()]+\)[^,]*)?
                 (?:(?:DEFAULT|CHECK)\s*(?:\(.*?\))?[^,]*)*
                 COLLATE\s+["\']?([^\s,"\')]+)/isx', $sql, $match)) {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -457,4 +457,9 @@ private function parseColumnCommentFromSQL($column, $sql)
 
         return false;
     }
+
+    private function escapeRegex($str)
```
Ocramius: I think this is the wrong solution, while the bug spawns from no parameters being passed to `preg_quote()`. We should probably just switch the delimiter to be `/` (as usual) and then it should be fine. Can you give that a spin?

Ocramius (changes requested):

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
Bruce17: Sure, I'll test it..

lcobucci:
@Bruce17 any reason to send this PR to `develop` instead of `master`?

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -457,4 +457,9 @@ private function parseColumnCommentFromSQL($column, $sql)
 
         return false;
     }
+    
+    private function pregQuote($str)
```
lcobucci: Missing type declaration in parameter and return

lcobucci (changes requested):
Please also check the CS violations with `vendor/bin/phpcs lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php` (checking only the changes you've made)

Bruce17:
@lcobucci I was not sure if I should push to master, but I can open a new PR if you like.

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
lcobucci: I was actually talking about `private function pregQuote(string $str): string` not the docblock =)

lcobucci:
> @lcobucci I was not sure if I should push to master, but I can open a new PR if you like.

@Bruce17 I was just checking, `develop` is related to the changes for the ORM v3. It seems that this change is something to part of DBAL v2 😄  so please, rebase your branch and change the destination branch (or simply apply your change on top of `master` and open a new PR).

PR #2865, opened 2017-09-16
Fixes for Sqlite schema regex

GwendolenLynch:
Both column collation & comments were not respecting word boundaries so `foo` and `foobar` would conflict depending on order in the schema.

Capable & happy to make adjustments to tests, commits, etc, as requested.

PR #2885, opened 2017-10-10
Escape "#" in regex in SqliteSchemaManager

Bruce17:
This pull request escapes `#` in sqlite db column names.

See issue https://github.com/doctrine/dbal/issues/2881 for more information.


lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -444,10 +444,10 @@ private function getTableDiffForAlterForeignKey(ForeignKeyConstraint $foreignKey
     private function parseColumnCollationFromSQL($column, $sql)
     {
         if (preg_match(
-            '{(?:'.preg_quote($column).'|'.preg_quote($this->_platform->quoteSingleIdentifier($column)).')
+            '/(?:'.$this->pregQuote($column).'|'.$this->pregQuote($this->_platform->quoteSingleIdentifier($column)).')
```
lcobucci: This is what I told you about CS violations: https://travis-ci.org/doctrine/dbal/jobs/285994030#L2398 can you please fix it?

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -463,9 +463,9 @@ private function parseColumnCollationFromSQL($column, $sql)
     private function parseColumnCommentFromSQL($column, $sql)
     {
         if (preg_match(
-            '{[\s(,](?:'.preg_quote($this->_platform->quoteSingleIdentifier($column)).'|'.preg_quote($column).')
+            '/[\s(,](?:'.$this->pregQuote($this->_platform->quoteSingleIdentifier($column)).'|'.$this->pregQuote($column).')
```
lcobucci: This is what I told you about CS violations: https://travis-ci.org/doctrine/dbal/jobs/285994030#L2412 can you please fix it?

lcobucci (changes requested):

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -474,4 +474,14 @@ private function parseColumnCommentFromSQL($column, $sql)
...
     }
+    
+    /**
+     * @param string $str
+     * 
+     * @return string
+     */
+    private function pregQuote($str)
+    {
+        return preg_quote($str, '#');
```
Majkl578: Why `#`? Patterns are using `/`. This is invalid usage, you are abusing preg_quote for different purpose.

Majkl578 (changes requested):

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -474,4 +474,14 @@ private function parseColumnCommentFromSQL($column, $sql)
 
         return false;
     }
+    
+    /**
+     * @param string $str
+     * 
+     * @return string
+     */
+    private function pregQuote($str)
```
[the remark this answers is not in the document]
morozov: I'd say there shouldn't be a method like this in the first place. Every regex in the class may have its own delimiters, but having a generic method which implies just one may be lead to errors in the future.

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
Bruce17: Please have a look at the linked issue.

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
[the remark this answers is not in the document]
Bruce17: Well I like DRY, but I remove it...

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
Bruce17: The violation was there already in the first place, but I'll fix it ...

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
Majkl578: I see your point, but it is still invalid usage, `#` is not a delimiter. It should be done i.e. with simple `str_replace('#', '\#', $str)` _after_ passing `$str` through `preg_quote()`.

Bruce17: Well I already did that in the old PR, see https://github.com/doctrine/dbal/pull/2884/commits/5bf652ef0ac712116e9f7bbcfc95ba6465c9435c#diff-3ee26e34bba502d9bf80cff65c964b9aR463.

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -444,10 +444,10 @@ private function getTableDiffForAlterForeignKey(ForeignKeyConstraint $foreignKey
     private function parseColumnCollationFromSQL($column, $sql)
     {
         if (preg_match(
-            '{(?:'.preg_quote($column).'|'.preg_quote($this->_platform->quoteSingleIdentifier($column)).')
+            '/(?:' . preg_quote($column, '#') . '|' . preg_quote($this->_platform->quoteSingleIdentifier($column), '#') . ')
```
Majkl578: I still strongly disagree with abusing `preg_quote()` to escape single character. It should be done by `str_replace()` or `addcslashes()`.
Also, changing delimiter to `/` introduces another issue - you have to escape `/` as well, which you can't do if you abuse `preg_quote()`'s 2nd parameter to escape `#`.
So as it currently stands, this would be correct:
```php
addcslashes(preg_quote($column, '/'), '#')
```
https://3v4l.org/8Bnvj

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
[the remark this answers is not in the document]
morozov: Repeating the delimiter character in the calls to `preg_quote()` is not a violation of DRY same as the usage of those delimiters in the expressions. It belongs to the expression as well as the delimiters themselves.

lcobucci:
@Bruce17 it looks like we're missing some tests since the last 2 commits didn't fail during the build. Could you please check that, and if we indeed missed, add them?

lcobucci:
I might be wrong though =)

Bruce17:
@lcobucci I'll check that.

Bruce17:
Can somebody rerun the travis job? There is an `phpenv` error.

lcobucci:
@Bruce17 done 😉

Bruce17:
Any progress on this one?

morozov:
I still have a mixed feeling about using the `/x` modifier in the regex without the reason and then fixing the PHP's bug by using `addcslashes()`. Technically it's correct but it's a workaround. Why can't we just remove the `/x` and leave everything as is including the original `{}` delimiters?

Majkl578:
The fix has been merged into PHP 7.3 tree, can you please update PR and conditionally escape it only for `PHP_VERSION_ID < 70300`? Thanks.

@morozov Yes, well, it sometimes improves readability for large/long/multiline regexps, at least that's where I use it. Probably for further discussion elsewhere (aka send PR?).

Bruce17:
@Majkl578 I updated the code, but it seems the php fix has not landed in php nightly yet.

Majkl578:
Don't worry about nightly, it's probably just some days/weeks old in the Travis build. I'll try to test it against 7.3 locally.

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -437,8 +437,14 @@ private function getTableDiffForAlterForeignKey(ForeignKeyConstraint $foreignKey
 
     private function parseColumnCollationFromSQL(string $column, string $sql) : ?string
     {
-        $pattern = '{(?:\W' . preg_quote($column) . '\W|\W' . preg_quote($this->_platform->quoteSingleIdentifier($column))
-                 . '\W)[^,(]+(?:\([^()]+\)[^,]*)?(?:(?:DEFAULT|CHECK)\s*(?:\(.*?\))?[^,]*)*COLLATE\s+["\']?([^\s,"\')]+)}isx';
+        if (PHP_VERSION_ID < 70300) {
```
Majkl578: Can you please add comment with a link to the bug report explaining this condition? Thanks.
https://bugs.php.net/bug.php?id=75355

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -437,8 +437,14 @@ private function getTableDiffForAlterForeignKey(ForeignKeyConstraint $foreignKey
...
+            $pattern = '{(?:\W' . addcslashes(preg_quote($column, '/'), '#') . '\W|\W' . addcslashes(preg_quote($this->_platform->quoteSingleIdentifier($column), '/'), '#')
```
Majkl578: Since this pattern changed delimiter from `/` to `{...}`, it is now superfluous to specify second parameter to `preg_quote()`.

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -449,8 +455,13 @@ private function parseColumnCollationFromSQL(string $column, string $sql) : ?str
 
     private function parseColumnCommentFromSQL(string $column, string $sql) : ?string
     {
-        $pattern = '{[\s(,](?:\W' . preg_quote($this->_platform->quoteSingleIdentifier($column)) . '\W|\W' . preg_quote($column)
-                 . '\W)(?:\(.*?\)|[^,(])*?,?((?:(?!\n))(?:\s*--[^\n]*\n?)+)}ix';
+        if (PHP_VERSION_ID < 70300) {
+            $pattern = '{[\s(,](?:\W' . addcslashes(preg_quote($this->_platform->quoteSingleIdentifier($column), '/'), '#') . '\W|\W' . addcslashes(preg_quote($column, '/'), '#')
```
Majkl578: Also here, no need for escaping `/`.

Majkl578 (changes requested):
LGTM, please just add a link to PHP bug report so the condition is not confusing in future.

Majkl578:
Btw. tested locally with 7.3.0-dev: Without change of the regexp, tests fail under 7.2 and pass under 7.3, with the code change both pass. ✔️

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
Bruce17: done

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -437,8 +437,16 @@ private function getTableDiffForAlterForeignKey(ForeignKeyConstraint $foreignKey
...
     private function parseColumnCollationFromSQL(string $column, string $sql) : ?string
     {
-        $pattern = '{(?:\W' . preg_quote($column) . '\W|\W' . preg_quote($this->_platform->quoteSingleIdentifier($column))
-                 . '\W)[^,(]+(?:\([^()]+\)[^,]*)?(?:(?:DEFAULT|CHECK)\s*(?:\(.*?\))?[^,]*)*COLLATE\s+["\']?([^\s,"\')]+)}isx';
+        if (PHP_VERSION_ID < 70300) {
+            // PHP < 7.3 does not properly escape '#' in regex.
+            // See bug: https://bugs.php.net/bug.php?id=75355
+            $pattern = '{(?:\W' . addcslashes(preg_quote($column), '#') . '\W|\W' . addcslashes(preg_quote($this->_platform->quoteSingleIdentifier($column)), '#')
+                     . '\W)[^,(]+(?:\([^()]+\)[^,]*)?(?:(?:DEFAULT|CHECK)\s*(?:\(.*?\))?[^,]*)*COLLATE\s+["\']?([^\s,"\')]+)}isx';
+
```
Majkl578: This extra line causes CI Coding Standards failure.

Majkl578 (approved):

Majkl578:
Hi, since #2865 has changed the pattern, the `x` modifier is now meaningless. So I've taken an easier approach of just dropping `x` modifier and taken your tests. See #3030.

Majkl578:
Closing in favor of #3030.

PR #2889, opened 2017-10-13
Add support for PDO_DBLIB driver (SQL Server and potentially Sybase ASE)

Deltachaos:
The current official Version for SQL Server is not available in a stable version for PHP7 and only from the latest version on available for Linux. The FreeTDS PDO_DBLIB driver is there since ages and is also part of PHP7.

It is working without problems for SQL Server, and Sybase ASE. So why should we not support it?

The there are two issues i had to add workarounds for:
 - You can only read the results of one query in parallel.
 - SQL Server and Sybase ASE are very strict in type checking. So we have to pass always the correct type.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractDbLibDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractDbLibDriver.php
@@ -1,0 +1,105 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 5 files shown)


morozov:
> The current official Version for SQL Server is not available in a stable version for PHP7 and only from the latest version on available for Linux.

Could you please clarify what this means? The server itself cannot be available in PHP but there is an [extension](https://github.com/Microsoft/msphpsql/tree/PHP-7.0-Linux) for PHP 7 with a [stable release](https://github.com/Microsoft/msphpsql/releases/tag/v4.3.0) maintained by Microsoft.

Deltachaos:
Year, i meant that driver. Sorry, thought it was still beta for Linux. How ever, DBLib is still a other option to connect.

morozov:
For a new driver, it would be nice if we could set up automated testing from scratch (e.g. using AppVeyor). There's work in progress in #2617. Could you try and see if testing this driver can be done there as well?

Deltachaos:
@morozov will have a look

PR #2890, opened 2017-10-14
MasterSlaveConnection implements MasterSlaveConnectionInterface in or…

ionuttocila:
…der to remove dependency of concrete class implementation

### DIFF
--- a/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
+++ b/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
@@ -80,7 +80,7 @@ use Doctrine\DBAL\Events;
  * @author Lars Strojny <lstrojny@php.net>
  * @author Benjamin Eberlei <kontakt@beberlei.de>
  */
-class MasterSlaveConnection extends Connection
+class MasterSlaveConnection extends Connection implements MasterSlaveConnectionInterface
 {
     /**
      * Master and slave connection (one of the randomly picked slaves).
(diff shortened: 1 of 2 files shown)


lcobucci:
@alteringres why are you depending on that implementation and not on the `Connection` interface? Does your application behaves differently because of that?

That implementation is not an extension point of DBAL so IMO it shouldn't have an interface, maybe @Ocramius and @deeky666 can have a different opinion but I'd say this should be a `won't fix`.

lcobucci:
This class should be `final` too 😜

alteringres:
Because MasterSlaveConnection does not implement an interface which represents the public API of master/slave logic, other packages like doctrine/migrations will do fixes like:
https://github.com/doctrine/migrations/pull/479
so, when this combines with Symfony Doctrine Configuration which allows the usage of custom master/slave wrappers (others then the MasterSlaveConnection provided by DBAL), some functions will be not work any more (in  this example, migrations will not work any more if wrapper is changed into another class)

In my opinion, other packages should not check a concrete class implementation, but an interface which can be implemented by anyone in other ways, keeping the same public API.

I've got this problem, so I created this PR to create the interface, then I'll create another PR to fix doctrine/migrations

lcobucci:
> Because MasterSlaveConnection does not implement an interface which represents the public API of master/slave logic, other packages like doctrine/migrations will do fixes like:
doctrine/migrations#479

@alteringres IMO doctrine/migrations should never force the connection to master, that reveals an encapsulation issue on the `MasterSlaveConnection` (we should rather hear from @Ocramius and @mikeSimonson to have some context on this).

> In my opinion, other packages should not check a concrete class implementation, but an interface which can be implemented by anyone in other ways, keeping the same public API.

You're correct, however that's only applicable when it was designed in that way (as an extension point). In this case I still think that the `MasterSlaveConnection` should be 100% transparent for every single piece of code that relies on DBAL (but I might be wrong, ofc).

PR #2891, opened 2017-10-14
Added guid type support to DB2 platform

tobiasstadler:
DB2 has GENERATE_UNIQUE for "UUID" generation, see https://www.ibm.com/support/knowledgecenter/en/SSEPGG_11.1.0/com.ibm.db2.luw.sql.ref.doc/doc/r0000809.html


morozov:
Is this unique ID a UUID? UUID is a 16-byte string, not 13-char/byte. The article in the description doesn't contain a single mention of UUID.

tobiasstadler:
No it is not a UUID according to the standard, but it is a unique identifier

Ocramius:
Probably better to keep userland UUIDs then (`CHAR(36)`)

Issue #2868, opened 2017-09-22
Doctrine\DBAL\Exception\InvalidFieldNameException when working with DB schema on PostgreSQL 10


skobkin:
Since upgrading to PostgreSQL 10 I have an error in my projects using Doctrine when I trying to generate migrations or just check the schema difference:

```
$ sf doc:sche:up --dump-sql                                                                                          
  [Doctrine\DBAL\Exception\InvalidFieldNameException]                                                        
  An exception occurred while executing 'SELECT min_value, increment_by FROM "admin"."acl_classes_id_seq"':  
  SQLSTATE[42703]: Undefined column: 7 ERROR:  column "min_value" does not exist                             
  LINE 1: SELECT min_value, increment_by FROM "admin"."acl_classes_id_...                                    
                 ^
  [Doctrine\DBAL\Driver\PDOException]                                             
  SQLSTATE[42703]: Undefined column: 7 ERROR:  column "min_value" does not exist  
  LINE 1: SELECT min_value, increment_by FROM "admin"."acl_classes_id_...         
                 ^
  [PDOException]                                                                  
  SQLSTATE[42703]: Undefined column: 7 ERROR:  column "min_value" does not exist  
  LINE 1: SELECT min_value, increment_by FROM "admin"."acl_classes_id_...         
                 ^
doctrine:schema:update [--complete] [--dump-sql] [-f|--force] [--em [EM]]
```

Seems like it's related to [nextcloud/server#5930](https://github.com/nextcloud/server/issues/5930) where @justin-sleep wrote:

> This happens specifically because of changes to how PostgreSQL 10 handles sequence metadata.

wzator:
Suggested solution:
PostgreSqlSchemaManager.php line 292:
>         $data = $this->_conn->fetchAll('SELECT min_value, increment_by FROM ' . $this->_platform->quoteIdentifier($sequenceName));

to :

>         $version = floatval($this->_conn->getWrappedConnection()->getServerVersion());
> 
>         if ($version >= 10) {
>            $data = $this->_conn->fetchAll('SELECT min_value, increment_by FROM pg_sequences WHERE schemaname = \'public\' AND sequencename = '.$this->_conn->quote($sequenceName));
>         }
>         else
>         {
>             $data = $this->_conn->fetchAll('SELECT min_value, increment_by FROM ' . $this->_platform->quoteIdentifier($sequenceName));
>         }

justin-sleep:
[...]

PR #2893, opened 2017-10-19
PostgreSQL 10 support

simPod:
When using Postgres 10, DBAL throws

```
An exception occurred while executing 'SELECT min_value, increment_by FROM "schemaname"."table_id_seq"':

  SQLSTATE[42703]: Undefined column: 7 ERROR:  column "min_value" does not exist
  LINE 1: SELECT min_value, increment_by FROM "schemaname"."table_id_se...
```

This fixes it as PSQL10 reads sequences definitions from `pg_sequences`. https://www.postgresql.org/docs/10/static/view-pg-sequences.html


lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -289,7 +289,13 @@ protected function _getPortableSequenceDefinition($sequence)
             $sequenceName = $sequence['relname'];
         }
 
-        $data = $this->_conn->fetchAll('SELECT min_value, increment_by FROM ' . $this->_platform->quoteIdentifier($sequenceName));
+        $version = (float) $this->_conn->getWrappedConnection()->getServerVersion();
```
lcobucci: This is breaking the test suite, can you please check that?

lcobucci: Also [Postgres 10 yet is not supported in Travis](https://docs.travis-ci.com/user/database-setup/#Using-a-different-PostgreSQL-Version), so there's no way to 100% guarantee the compatibility at this moment.

lcobucci (changes requested):
Can you also add new tests that covers that for PGSQL 10?

Majkl578:
Duplicate of #2868. Waiting for Travis, see https://github.com/travis-ci/travis-ci/issues/8537.

simPod:
@lcobucci It's done. I believe the test was not working properly before (correct me if I'm wrong)

simPod:
There was a case when null was returned for $data and this case was ignored. https://github.com/doctrine/dbal/pull/2893/files#diff-e4bd68e20fcb55598e13e0831ba281abR63 I can add null case it needed.

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -289,7 +289,13 @@ protected function _getPortableSequenceDefinition($sequence)
             $sequenceName = $sequence['relname'];
         }
 
-        $data = $this->_conn->fetchAll('SELECT min_value, increment_by FROM ' . $this->_platform->quoteIdentifier($sequenceName));
+        $version = (float) $this->_conn->getServerVersion();
+
+        if ($version >= 10) {
```
Majkl578: I think this should rather be based on Platform check, there will sure be separate platform for Postgres 10. Anyway I think it should all be done in a single PR, not fixing random pieces randomly... :/

lcobucci: I agree, also exposing `Connection#getServerVersion()` a no-go IMO

simPod:
Tried to refactor. I see this project's internals for the first time tho but IMO it's good start for PSQL10. I can either continue improving this or we can close.

PillowPillow:
Any news about this PR?

iquad:
@PillowPillow Travis still working to implement Postgresql 10. Until that time, we have to wait.

SCIF:
Just ran functional tests on Postgres10 localy and all good. :+1:

lcobucci:
@SCIF it's really good to hear that let's just hope that travis adds support to it soon

Majkl578:
I got tired of infinitely waiting for Travis, so here is externally installed Postgres 10: #2946. Feel free to pick that commit and build a working compat PR on top. 😎

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3570,4 +3570,9 @@ public function getStringLiteralQuoteCharacter()
     {
         return "'";
     }
+
+    public function getSequenceDataSQL($sequenceName, $schemaName)
```
Majkl578: DBAL requires PHP 7.1, please add parameter and return types where applicable.

lib/Doctrine/DBAL/Platforms/Keywords/PostgreSQL100Keywords.php
```diff
@@ -0,0 +1,37 @@
+<?php
+/*
```
Majkl578: No license headers please, we've already removed them in develop.

tests/Doctrine/Tests/DBAL/Driver/AbstractPostgreSQLDriverTest.php
```diff
@@ -67,7 +67,7 @@ protected function getDatabasePlatformsForVersions()
             array('9.4', 'Doctrine\DBAL\Platforms\PostgreSQL94Platform'),
             array('9.4.0', 'Doctrine\DBAL\Platforms\PostgreSQL94Platform'),
             array('9.4.1', 'Doctrine\DBAL\Platforms\PostgreSQL94Platform'),
-            array('10', 'Doctrine\DBAL\Platforms\PostgreSQL94Platform'),
+            array('10', 'Doctrine\DBAL\Platforms\PostgreSQL100Platform'),
```
Majkl578: Please use ::class.

tests/Doctrine/Tests/DBAL/Schema/PostgreSQLSchemaManagerTest.php
```diff
@@ -66,10 +70,18 @@ public function testFiltersSequences()
...
+                new Sequence('schema.bar', 2, 2),
             ],
             $this->schemaManager->listSequences('database')
         );
     }
+
+    public function filtersSequencesDataProvider()
+    {
+        return [
+            ['Doctrine\DBAL\Platforms\PostgreSqlPlatform'],
```
Majkl578: Please use ::class.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
Majkl578: It should probably be renamed to `getSequenceMetadataSQL`, `data` is not explicit.

Majkl578 (changes requested):

Majkl578:
Here's a diff of keywords in [9.6](https://www.postgresql.org/docs/9.6/static/sql-keywords-appendix.html) and [10](https://www.postgresql.org/docs/10/static/sql-keywords-appendix.html): https://www.diffchecker.com/fJ0YkHgw
There are no new reserved keywords, but some of the existing are not reserved anymore. But I'm not a DBAL guy so I can't say whether they are supposed to be removed in PostgreSQL100Keywords or not.

simPod:
They were not reserved before tho.

lib/Doctrine/DBAL/Platforms/PostgreSQL100Platform.php
```diff
@@ -0,0 +1,34 @@
...
+{
+	/**
+	 * {@inheritdoc}
+	 */
+	protected function getReservedKeywordsClass() : string
+	{
+		return PostgreSQL100Keywords::class;
+	}
+
+	public function getSequenceMetadataSQL(string $rawSequenceName, string $schemaName) : string
```
morozov: There shouldn't be a semantic difference between this method and its parents from `AbstractPlatform` and `PostgreSqlPlatform`. Otherwise, the caller which depends on the abstract name will not know whether it should pass raw or non-raw sequence name.

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -289,7 +289,7 @@ protected function _getPortableSequenceDefinition($sequence)
             $sequenceName = $sequence['relname'];
         }
 
-        $data = $this->_conn->fetchAll('SELECT min_value, increment_by FROM ' . $this->_platform->quoteIdentifier($sequenceName));
+        $data = $this->_conn->fetchAll($this->_platform->getSequenceMetadataSQL($sequenceName, $sequence['schemaname']));
```
morozov: Is it really necessary to `fetchAll()` and then use only `$data[0]`? Doing `fetch()` and using the row directly would make more sense to me.

lib/Doctrine/DBAL/Platforms/PostgreSQL100Platform.php
```diff
@@ -0,0 +1,34 @@
...
+	{
+		$sequenceName = $this->cleanSequenceNameFromSchemaName($rawSequenceName, $schemaName);
+
+		return 'SELECT min_value, increment_by FROM pg_sequences WHERE schemaname = '.$this->quoteStringLiteral($schemaName).' AND sequencename = ' . $this->quoteStringLiteral($sequenceName);
```
morozov: Given that in PostgreSQL 10, the sequence metadata is available via a view, would it make sense to rework `getListSequencesSQL()` instead? With this approach, the client will not have to execute the SQL from  `getSequenceMetadataSQL()` for each sequence, and the method itself will not need to be introduced on the abstract platform level.

iquad:
@simPod are you going to fix/apply or need help ? We're waiting to deploy new product to  production for 2 months :P I can help if you wish.

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
simPod: There's no method `fetch()` in connection. In my opinion, this is ok as it fetches one row

lib/Doctrine/DBAL/Platforms/PostgreSQL100Platform.php
simPod: Not sure it's a good idea. `getListSequencesSQL()` lists sequences. It would need to add also `increment_by` and `min_value` columns. That would change the method's functionality. This method is called before in `listSequences()` already.

simPod:
@iquad you can already use this fork so you don't have to wait :)

iquad:
@simPod yeah, but need doctrine to merge them, server tests, 30+ development machine etc. Can't handle them one by one :/

simPod:
@iquad personally I think it's done but waiting for approvals or some other dbal guys reviews

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
Majkl578: There are more fetch*() methods though, fetchArray() is the one you need.

lib/Doctrine/DBAL/Platforms/PostgreSQL100Platform.php
```diff
@@ -0,0 +1,34 @@
...
+    protected function getReservedKeywordsClass() : string
+    {
+        return PostgreSQL100Keywords::class;
+    }
+
+    public function getSequenceMetadataSQL(string $sequenceName, string $schemaName) : string
+    {
+        $sequenceName = $this->cleanSequenceNameFromSchemaName($sequenceName, $schemaName);
+
+        return 'SELECT min_value, increment_by FROM pg_sequences WHERE schemaname = '.$this->quoteStringLiteral($schemaName).' AND sequencename = ' . $this->quoteStringLiteral($sequenceName);
```
Majkl578: Please address CI failure - spaces around dots.

Majkl578 (changes requested):

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
simPod: The data then looks like  
image. 
I personally think the current state is more verbose as I have to reference array values by column names and not by numeric indeces.

lib/Doctrine/DBAL/Platforms/PostgreSQL100Platform.php
morozov: Look at the `OraclePlatform`. Its `getListSequencesSQL()` returns list and metadata. The only reason it doesn't do the same for PostgreSQL is that prior to version 10 there was no view to fetch the metadata from.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3570,4 +3570,9 @@ public function getStringLiteralQuoteCharacter()
     {
         return "'";
     }
+
+    public function getSequenceMetadataSQL(string $sequenceName, string $schemaName) : string
```
morozov: It doesn't belong here. Please implement `getListSequencesSQL()` for the new platform so that it returns sequence names and metadata same as it's done in the `OraclePlatform`. The only platform which needs to be able to provide metadata for one individual sequence is PostgreSQL < 10 because it doesn't have sequence metadata available via a view.

morozov (changes requested):

simPod:
Any idea how to run that failing test? https://travis-ci.org/doctrine/dbal/jobs/320454235 Have not found any way how to reproduce that locally :/

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -283,15 +283,19 @@ protected function getPortableNamespaceDefinition(array $namespace)
...
             $sequenceName = $sequence['schemaname'] . "." . $sequence['relname'];
         } else {
             $sequenceName = $sequence['relname'];
         }
 
-        $data = $this->_conn->fetchAll('SELECT min_value, increment_by FROM ' . $this->_platform->quoteIdentifier($sequenceName));
+        if(!isset($sequence['increment_by'], $sequence['min_value'])){
+            $data = $this->_conn->fetchAll('SELECT min_value, increment_by FROM ' . $this->_platform->quoteIdentifier($sequenceName));
+            $sequence['increment_by'] = $data[0]['min_value'];
+            $sequence['min_value'] = $data[0]['increment_by'];
```
morozov: @simPod is this the reason of the test failures?

simPod: @morozov you're right, thanks! I failed to run those localy but it passes now

lib/Doctrine/DBAL/Platforms/Keywords/PostgreSQL100Keywords.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Doctrine\DBAL\Platforms\Keywords;
+
+/**
+ * PostgreSQL 100 reserved keywords list.
+ *
+ * @link   www.doctrine-project.org
+ * @since  2.6
```
morozov: As of https://github.com/doctrine/coding-standard/pull/9, `@since` will be not allowed. The `@link` above doesn't bring much value as well IMO.

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -283,15 +283,19 @@ protected function getPortableNamespaceDefinition(array $namespace)
...
-        if ($sequence['schemaname'] != 'public') {
+        if ($sequence['schemaname'] !== 'public') {
             $sequenceName = $sequence['schemaname'] . "." . $sequence['relname'];
         } else {
             $sequenceName = $sequence['relname'];
         }
 
-        $data = $this->_conn->fetchAll('SELECT min_value, increment_by FROM ' . $this->_platform->quoteIdentifier($sequenceName));
+        if ( ! isset($sequence['increment_by'], $sequence['min_value'])) {
+            $data                     = $this->_conn->fetchAll('SELECT min_value, increment_by FROM ' . $this->_platform->quoteIdentifier($sequenceName));
```
morozov: You can use `Connection::fetchAssoc()` to fetch one row and then `array_merge($sequence, $data)` to not confuse the keys.

simPod: That would require rework of other tests https://travis-ci.org/doctrine/dbal/jobs/320775111 However, when I launch the testsuite locally, they all pass.

morozov: Which exactly failures do you think are caused by this?

simPod: My bad, found it, fixed it

lib/Doctrine/DBAL/Platforms/PostgreSQL100Platform.php
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Doctrine\DBAL\Platforms;
+
+use Doctrine\DBAL\Platforms\Keywords\PostgreSQL100Keywords;
+
+/**
+ * Provides the behavior, features and SQL dialect of the PostgreSQL 10.0 database platform.
+ *
+ * @link   www.doctrine-project.org
+ * @since  2.6
```
morozov: Please remove annotations.

tests/Doctrine/Tests/DBAL/Schema/PostgreSQLSchemaManagerTest.php
```diff
@@ -32,37 +36,35 @@ protected function setUp()
 
     /**
      * @group DBAL-474
+     * @dataProvider filtersSequencesDataProvider
      */
-    public function testFiltersSequences()
+    public function testFiltersSequences(string $platformClass, array $sequencesData, int $fetchAssocExpectationCount) : void
```
morozov: This test is hard to maintain and not really useful:

1. It used to pass on PostgreSQL 10 before the issue was fixed.
2. It started failing and had to be reworked after the issue was fixed.
3. It enforces implementation details which are not necessarily correct (`fetchAssoc` vs. `fetchAll`).

Let's replace this test with a functional one in `SchemaManagerFunctionalTestCase`. The test should create a sequence using the `AbstractSchemaManager` manager API, and then list the sequences to see if it's created and introspected correctly. This test will not depend on the implementation details of a given platform and will prove that the functionality is portable between the platforms.

morozov (commented):
The code looks good. Now it's time to take care of the test :-)

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1360,4 +1361,26 @@ public function commentsProvider() : array
...
         ];
     }
+
+    public function testCreateAndListSequences() : void
+    {
+        if ( ! $this->_sm->getDatabasePlatform()->supportsSequences()) {
+            $this->markTestSkipped('This test is only supported on platforms that support sequences.');
+        }
+
+        $sequence1Name = 'sequence_1';
```
Majkl578: This seems to error on Oracle now:
```
1) Doctrine\Tests\DBAL\Functional\Schema\OracleSchemaManagerTest::testCreateAndListSequences 
Exception: [PHPUnit\Framework\Error\Notice] Undefined index: sequence_1 
```
https://continuousphp.com/git-hub/doctrine/dbal/build/c3aeacf9-fd40-45c2-8304-e93fc34a5bee/output

simPod: Yup, have to setup oracle here to reproduce.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1360,4 +1361,26 @@ public function commentsProvider() : array
...
+        $sequence2Name = 'sequence_2';
+        $sequence1     = new Sequence($sequence1Name);
```
morozov: @simPod please make sure all sequence attributes are covered by the test, not only the name.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
morozov: Most likely, it’s upper case on Oracle.

simPod: Indeed, it is

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
simPod: Can you review now please? Thanks

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1360,4 +1361,43 @@ public function commentsProvider() : array
...
+
+        $this->_sm->createSequence($sequence1);
+        $this->_sm->createSequence($sequence2);
+        /** @var Sequence[] $actualSequences */
+        $actualSequences = [];
+        foreach ($this->_sm->listSequences() as $sequence) {
+            $actualSequences[$sequence->getName()] = $sequence;
+        }
+
+        $isOracle      = $this->_sm->getDatabasePlatform()->getName() === OraclePlatform::NAME;
```
morozov: Oracle is not the only supported platform which returns upper-cased column names. DB2 is the other one but it looks DBAL doesn't support sequences in DB2. In any event, checking the platform name is not the best approach.

I can think of two solutions:

1. Figure out why the rest of the tests in this class don't fail on Oracle without using the portability layer. For instance, `testListTableColumns()` expects column names to be lower-cased, however, they are upper-cased in Oracle. Must be implemented in:
https://github.com/doctrine/dbal/blob/22844d3c0d52e44c312ad60571b3d9e8da662c9b/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php#L824
2. Use `Portability\Connection` with `PORTABILITY_FIX_CASE` and `fetch_case = CASE_LOWER` when instantiating the schema manager. This is hor a real production code would deal with the case discrepancy.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1360,4 +1361,43 @@ public function commentsProvider() : array
...
+        if ( ! $this->_sm->getDatabasePlatform()->supportsSequences()) {
+            $this->markTestSkipped('This test is only supported on platforms that support sequences.');
+        }
+
+        $sequence1Name           = 'sequence_1';
+        $sequence1AllocationSize = 1;
+        $sequence1InitialValue   = 1;
+        $sequence2Name           = 'sequence_2';
+        $sequence2AllocationSize = 2;
+        $sequence2InitialValue   = 2;
```
morozov: It's better to use different values for different parameters to make sure they are not confused during deployment or introspection.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1360,4 +1361,43 @@ public function commentsProvider() : array
...
+        $actualSequence1 = $actualSequences[$sequence1Name];
+        $actualSequence2 = $actualSequences[$sequence2Name];
+
+        $this->assertSame($sequence1Name, $actualSequence1->getName());
+        $this->assertEquals($sequence1AllocationSize, $actualSequence1->getAllocationSize());
+        $this->assertEquals($sequence1InitialValue, $actualSequence1->getInitialValue());
+
+        $this->assertSame($sequence2Name, $actualSequence2->getName());
+        $this->assertEquals($sequence2AllocationSize, $actualSequence2->getAllocationSize());
+        $this->assertEquals($sequence2InitialValue, $actualSequence2->getInitialValue());
```
morozov: Not sure the assertions are intentionally this verbose, but the same could be checked in a more compact way:
```php
$sequence1 = new Sequence('sequence_1', 1, 2);
$sequence2 = new Sequence('sequence_2', 3, 4);

// create the sequences

self::assertArraySubset([
    $sequence1,
    $sequence2,
], $this->_sm->listSequences());

Comment (simPod) [reply tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php]: I'm fan of verbosity but can remake if you think it better suits code style of this project. Should I proceed then? :)

Comment (morozov) [reply tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php]: Depending on what's the purpose of this verbosity. The approach I suggested is more state-based. It only tests that the collection returned by `listSequences()` contains the objects which were previously created. Isn't it what it's supposed to do?

The current approach also tests the public API of the `Sequence` class. For instance, if there was some logic in `getInitialValue` and `getAllocationSize()`, the test would have to take this logic into account as well which would make it less stable.

There are no strict coding guidelines in this regard, so it's up to you how you want to structure the test.

File: tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php (hunk 19 above)
Comment (simPod) [reply tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php]: `testListTableColumns` also converts table name to lowercase `self::assertEquals('test', strtolower($columns['test']->getname()));`

Can do it this way then. In https://github.com/doctrine/dbal/blob/22844d3c0d52e44c312ad60571b3d9e8da662c9b/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php#L824 it just creates an array with lowercased table names as keys.

Comment (morozov) [reply tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php]: It would be a breaking change. Doing `strtolower()` in a test would mean that the production code should do the same by design which it not true, it's a bug.

Let's skip the test for Oracle for now and fix the schema manager in 3.0.

File: tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php (hunk 21 above)
Comment (simPod) [reply tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php]: I agree. However, assertArraySubset asserts by keys and there are other Sequences created in db as well.

image

image

I reworked it a bit

Comment (simPod) [reply tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php]: It's failing on Psql 9.5. I'd have to either sort it by name or revert to the previous verbose version.

Comment (morozov) [reply tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php]: Eh… I didn't know `assertArraySubset()` is strict about numeric keys. The last approach could fail on any DB because the order of sequences is not enforced.

The problem is not verbosity itself, the problem is also that having the logic like `array_map()` or `array_filter()` outside of assertions which makes the test less readable and less clear about which exactly case it covers.

How about this:
```php
$list = $sm->listSequences();
self::assertContains($sequence1, $list);
self::assertContains($sequence2, $list);
```

simPod: That wouldn't work as those are different objects.

I guess in the end this is the simplier solution:

```php
$this->assertSame($sequence1Name, $actualSequence1->getName());
$this->assertEquals($sequence1AllocationSize, $actualSequence1->getAllocationSize());
$this->assertEquals($sequence1InitialValue, $actualSequence1->getInitialValue());
```

morozov: Okay, let's leave it like this.

simPod: Kk, it's there

morozov (approved):

lib/Doctrine/DBAL/Platforms/Keywords/PostgreSQL100Keywords.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Platforms\Keywords;
+
+/**
+ * PostgreSQL 100 reserved keywords list.
```
Majkl578: typo: `10.0`

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -283,15 +283,18 @@ protected function getPortableNamespaceDefinition(array $namespace)
...
+        if ($sequence['schemaname'] !== 'public') {
             $sequenceName = $sequence['schemaname'] . "." . $sequence['relname'];
         } else {
             $sequenceName = $sequence['relname'];
         }
 
-        $data = $this->_conn->fetchAll('SELECT min_value, increment_by FROM ' . $this->_platform->quoteIdentifier($sequenceName));
+        if ( ! isset($sequence['increment_by'], $sequence['min_value'])) {
+            $data     = $this->_conn->fetchAssoc('SELECT min_value, increment_by FROM ' . $this->_platform->quoteIdentifier($sequenceName));
+            $sequence = array_merge($sequence, $data);
```
Majkl578: Is array_merge needed here? Wouldn't `+` _just_ work since it's _assoc_?

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1360,4 +1361,40 @@ public function commentsProvider() : array
...
+        /** @var Sequence[] $actualSequences */
+        $actualSequences = [];
+        foreach ($this->_sm->listSequences() as $sequence) {
+            $actualSequences[$sequence->getName()] = $sequence;
+        }
+
+        $actualSequence1 = $actualSequences[$sequence1Name];
+        $actualSequence2 = $actualSequences[$sequence2Name];
+
+        $this->assertSame($sequence1Name, $actualSequence1->getName());
```
Majkl578: Please call all asserts statically.

tests/Doctrine/Tests/DBAL/Driver/AbstractPostgreSQLDriverTest.php
```diff
@@ -5,6 +5,7 @@
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Platforms\PostgreSqlPlatform;
 use Doctrine\DBAL\Schema\PostgreSqlSchemaManager;
+use Doctrine\DBAL\Platforms\PostgreSQL100Platform;
```
Majkl578: This one should be above Doctrine\DBAL\Platforms\PostgreSqlPlatform.

Majkl578 (changes requested):
Just last few nitpicks, otherwise LGTM. 👍

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
simPod: Isn't that deprecated?

simPod:
@morozov thanks for help!

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
Majkl578: What do you mean? All assert methods are declared static: https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Assert.php

simPod: Ah, not deprecated but it's there because of something oldhttps://github.com/sebastianbergmann/phpunit/issues/1914#issuecomment-148407321. 

Both can be used tho. I'm used to call it via non-static way. 

I guess it needs to be reworked to static then as the whole project uses static calls. Shall I?

Majkl578: And bad design in PHPUnit apparently, since non-static TestCase extends static Assert class. :/
Since it's declared as static, it was decided to use static calls in tests for assert methods, otherwise it's different context (and personally I think PHP should emit a warning, just like static->instance call).

Majkl578 (approved):

simPod:
@lcobucci is there anything else?

Ocramius (approved):
LGTM :ship:

sergiors:
OMG!

iquad:
thank you so much, all of you!

anonim1133:
Is there any date for 2.7 so I could use postgres 10? Or is there a way to patch 2.6 with this commit?

Ocramius:
No date yet. You can already use 2.6 with PostgreSQL 10, but the schema tools won't be working correctly yet.

nicholasruunu:
Still have this issue with `v2.9.2` and postgres `10.10`

exebece:
I think this issue can easily be solved using proper doctrine connection string specifying the version of postgres like: DATABASE_URL=postgres://user@127.0.0.1:5432/db_name?charset=utf8&serverVersion=10

the serverVersion part helped me solving all issues with Doctrine & Postgres. Try it for free ;)

nicholasruunu:
@exebece Yeah, my bad, updated the doctrine configuration from `9.x` to `10` and it works just fine now. Thanks.

XanderKon:
@exebece Thanks! Spent two hours and your comment save me from another time wasting.

PR #2894, opened 2017-10-23
[WIP] Index can be fulfilled with other index with higher column count

mateuszsip:
Meant to fix #6783.

This is result of my quick investigation.

Comment for conditional is over 7 years old, not sure if this is still our concern.
If so, I could need some guidance/explanation.

Definately needs more test scenarios, I have to do some research about it first.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Index.php
+++ b/lib/Doctrine/DBAL/Schema/Index.php
@@ -228,6 +228,17 @@ class Index extends AbstractAsset implements Constraint
                 return false;
             }
 
+            /**
+             * Index:
+             * unique(foo)
+             * must not be fulfilled with index:
+             * unique(foo, bar)
+             */
+            if (count($other->getColumns()) > count($this->getColumns())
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Schema/Index.php
```diff
@@ -200,9 +200,7 @@ public function spansColumns(array $columnNames)
      */
     public function isFullfilledBy(Index $other)
     {
-        // allow the other index to be equally large only. It being larger is an option
-        // but it creates a problem with scenarios of the kind PRIMARY KEY(foo,bar) UNIQUE(foo)
```
Ocramius: This is the bit that isn't clear.

A test is needed:

 * index `unique(foo)`
 * index `unique(foo, bar)`

The test should verify that the first unique index should not be dropped

kejwmen: Thanks for explanation, another simple test case added.

kejwmen:
Closing in favor of #2895

PR #2895, opened 2017-10-23
Fix functionality for determining whether an index is fulfilled by an…

stevepetcu:
…other one.

- modified Index::isFullfilledBy(Index $index) method to properly identify cases where a composite index is fulfilled by another one, covering Unique constraints;
- added tests;
- added/modified Table tests regarding covering foreign keys.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Index.php
+++ b/lib/Doctrine/DBAL/Schema/Index.php
@@ -200,9 +200,7 @@ class Index extends AbstractAsset implements Constraint
      */
     public function isFullfilledBy(Index $other)
     {
-        // allow the other index to be equally large only. It being larger is an option
-        // but it creates a problem with scenarios of the kind PRIMARY KEY(foo,bar) UNIQUE(foo)
-        if (count($other->getColumns()) != count($this->getColumns())) {
+        if (count($other->getColumns()) < count($this->getColumns())) {
             return false;
         }
 
@@ -222,11 +220,8 @@ class Index extends AbstractAsset implements Constraint
                 return true;
             }
 
-            if ($other->isPrimary() != $this->isPrimary()) {
-                return false;
-            }
-
-            if ($other->isUnique() != $this->isUnique()) {
+            if ( ! $this->isUniqueIndexFulfilledBy($other))
+            {
                 return false;
             }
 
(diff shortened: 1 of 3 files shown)


stevepetcu:
This is related to this issue: https://github.com/doctrine/doctrine2/issues/6783

I _think_ there are some tests that are failing due to missing drivers, not sure if that's normal.

B/c of the large number of skipped tests and some that were failing due to missing drivers, I only ran the tests under `DBAL/Schema` locally, but I will check what's going on more carefully, as soon as I have time. 

By the way, this is my first code PR to Doctrine. Feedback is appreciated, as I couldn't find guidelines about PRs to DBAL. 😅

kejwmen:
I started #2894 in the middle of the night, looks like my reference to this issue was not catched by github.

Wanted to finish it today, but it looks like you already did this.
Thanks, closing #2894

stevepetcu:
@kejwmen Ah, I didn't know that. I'm still a newbie to GitHub, no idea really how to link branches to issues, or anything like that yet. :S

lib/Doctrine/DBAL/Schema/Index.php
```diff
@@ -337,6 +332,23 @@ public function getOptions()
...
     /**
+     * Determines whether a Unique Index is fulfilled by another Index.
+     *
+     * @param Index $other
+     *
+     * @return bool
+     */
+    private function isUniqueIndexFulfilledBy(Index $other)
+    {
+        $isUniqueCovered = $this->isUnique()
```
mateuszsip, line 343: In my opinion readability suffered here.
Maybe grouping connected checks and giving them names would be better?

lib/Doctrine/DBAL/Schema/Index.php
```diff
@@ -222,11 +220,8 @@ public function isFullfilledBy(Index $other)
                 return true;
             }
 
-            if ($other->isPrimary() != $this->isPrimary()) {
-                return false;
-            }
-
-            if ($other->isUnique() != $this->isUnique()) {
+            if ( ! $this->isUniqueIndexFulfilledBy($other))
+            {
```
kejwmen: should be put in previous line

stevepetcu:
@kejwmen Thanks for the review. I'll think about making that function clearer.

Today I spent a couple of hours trying to fix the `SqlitePlatform::testAlterTable()` test that is failing. It's proving a bit more difficult than anticipated.

The problem is due to this method `SqlitePlatform::getPostAlterTableIndexForeignKeySQL(TableDiff $diff)` which, despite its name, has nothing to do with foreign keys. It simply _copies over indexes from the altered table_.

The issue arises when you have a _composite index_ which fulfils a FK, such that the FK does not have its own index. If you remove the composite index, the SqlitePlatform does no checks whatsoever on the FKs, to determine whether it should create new indexes.

I managed to scramble something up to generate new indexes based on FK columns, if they aren't fulfilled already, but a deeper problem is with the naming - I can't generate names for these indexes. This stuff is done in a protected method of the `Table` class - `_generateIdentifierName($columnNames, $prefix='', $maxSize=30)`.

This could, in theory, be circumvented, by making an array of all the indexes required for the altered table. Then _adding those indexes to a temporary table_, in the `SqlitePlatform::getAlterTableSQL(TableDiff $diff)` and _then iterating through it, and creating the actual queries for the indexes_, using the `AbstractPlatform::getCreateIndexSQL(Index $index, $table)`, but that is very inelegant, to say the least.

Plus, the same issue is likely to be found on some other platforms that I don't have the drivers to test installed.

I'm pulling my hair out a little.

PR #2900, opened 2017-10-26
Corrected handling of list parameters with sparse types

### LINKED ISSUE #2897: Incorrect handling of list parameters with sparse types
When a list parameter is bound to a query but the previous parameter types are omitted, the query gets processed incorrectly.

Steps to reproduce:
```php
$query = 'SELECT NULL FROM dummy WHERE ? IN (?)';
$params = ['foo', ['foo', 'bar']];
$types = [1 => Connection::PARAM_STR_ARRAY];

list($query, $params, $types) = SQLParserUtils::expandListParameters($query, $params, $types);
```
Expected result:
```php
$query = 'SELECT NULL FROM dummy WHERE ? IN (?, ?)';
$params = ['foo', 'foo', 'bar'];
$types = [null, PDO::PARAM_STR, PDO::PARAM_STR];
```
Actual result:

> Warning: array_merge(): Argument <span>#</span>2 is not an array in lib/Doctrine/DBAL/SQLParserUtils.php on line 136

```php
$query = 'SELECT NULL FROM dummy WHERE ? IN (?)'; // not modified, invalid
$params = null; // invalid
$types = [PDO::PARAM_STR]; // invalid
```

mateuszsip:
Meant to fix #2897

### DIFF
--- a/lib/Doctrine/DBAL/SQLParserUtils.php
+++ b/lib/Doctrine/DBAL/SQLParserUtils.php
@@ -92,6 +92,7 @@ class SQLParserUtils
 
         if ($isPositional) {
             ksort($params);
+            $types = self::fillMissingTypes($params, $types);
             ksort($types);
         }
 
@@ -239,5 +240,25 @@ class SQLParserUtils
         }
 
         throw SQLParserUtilsException::missingType($paramName);
+    }
+
+    /**
+     * @param array $params
+     * @param array $types
+     * @return array
+     */
+    private static function fillMissingTypes($params, $types)
+    {
+        if (count($types) === 0 || count($params) === count($types)) {
+            return $types;
+        }
+
+        foreach ($params as $paramKey => $param) {
+            if (!isset($types[$paramKey])) {
+                $types[$paramKey] = null;
+            }
+        }
+
+        return $types;
     }
 }
--- a/lib/Doctrine/DBAL/SQLParserUtils.php
+++ b/lib/Doctrine/DBAL/SQLParserUtils.php
@@ -92,8 +92,10 @@ class SQLParserUtils
 
         if ($isPositional) {
             ksort($params);
-            $types = self::fillMissingTypes($params, $types);
+            $types = self::fillMissingPositionalTypes($params, $types);
             ksort($types);
+        } else {
+            $types = self::fillMissingNamedTypes($params, $types);
         }
 
         foreach ($types as $name => $type) {
@@ -110,7 +112,7 @@ class SQLParserUtils
             $arrayPositions[$name] = false;
         }
 
-        if (( ! $arrayPositions && $isPositional)) {
+        if (empty($arrayPositions) && $isPositional) {
             return [$query, $params, $types];
         }
 
@@ -247,7 +249,7 @@ class SQLParserUtils
      * @param array $types
      * @return array
      */
-    private static function fillMissingTypes($params, $types)
+    private static function fillMissingPositionalTypes($params, $types)
     {
         if (count($types) === 0 || count($params) === count($types)) {
             return $types;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/SQLParserUtils.php
```diff
@@ -92,6 +92,7 @@ public static function expandListParameters($query, $params, $types)
 
         if ($isPositional) {
             ksort($params);
+            $types = self::fillMissingTypes($params, $types);
```
morozov: @kejwmen thanks for fixing this. I believe this problem also exists for named parameters but reveals in a slightly different way. Consider the following test:
```php
array(
    "SELECT * FROM Foo WHERE :foo IN (:bar)",
    array('foo' => 1, 'bar' => array(1, 2)),
    array('bar' => Connection::PARAM_INT_ARRAY),
    "SELECT * FROM Foo WHERE ? IN (?, ?)",
    array(1, 1, 2),
    array(null, \PDO::PARAM_INT, \PDO::PARAM_INT)
),
```
On your branch, it produces:
```
1) Doctrine\Tests\DBAL\SQLParserUtilsTest::testExpandListParameters with data set #10 ('SELECT * FROM Foo WHERE :foo IN (:bar)', array(1, array(1, 2)), array(101), 'SELECT * FROM Foo WHERE ? IN (?, ?)', array(1, 1, 2), array(null, 1, 1))
Types dont match
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => null
+    0 => 2
```
Basically, despite the first parameter has an integer value and doesn't have the type specified, it will be bound as a `PDO::PARAM_STR`.

morozov: If you decide to fix that as well, please make sure it works in the case when parameters have leading colons but types don't and vice versa like in these cases:

https://github.com/doctrine/dbal/blob/09e072dee1457bcdf3cae4be60b18854f8829992/tests/Doctrine/Tests/DBAL/SQLParserUtilsTest.php#L324-L339

kejwmen: Thanks for test cases, digging!

kejwmen:
I started focusing on the problem, so `fillMissingNamedTypes` can be not optimal performance-wise. If you see some obvious issues I would love to correct it.

Please tak a look at modified tests, do I understand you correctly?

By the way, it would be nice to make `expandListParameters` readable, but first I need to understand it.

morozov:
> I started focusing on the problem, so `fillMissingNamedTypes` can be not optimal performance-wise. If you see some obvious issues I would love to correct it.

Given that the existing code is performant enough and handles most of the cases, the fix (ideally) should be applied to the existing algorithm instead of being put aside. Otherwise, it fixes a very rare edge case at the cost of overal performance degradation. For instance, type/param normalization seems not required in any of the covered cases most probably it's already implemented somewhere inside.

I cannot give any concrete advice on the approach since it requires detailed understanding of the algorithm.

> Please take a look at modified tests, do I understand you correctly?

Not sure why the tests had to be modified. I think it would be acceptable if instead of a missing type a `NULL` type was returned (as it was done previously), but all explicitly specified types should remain in place.

> By the way, it would be nice to make `expandListParameters` readable, but first I need to understand it.

I think it's enough to have the algorithm documented somewhere in the ticket. Refactoring (if I read "make" correctly) for sake of readability may be too dangerous given how mission critical this piece of code is.

kejwmen:
> Otherwise, it fixes a very rare edge case at the cost of overal performance degradation. For instance, type/param normalization seems not required in any of the covered cases most probably it's already implemented somewhere inside.

I will take a closer look to check what could be done.

> Not sure why the tests had to be modified. I think it would be acceptable if instead of a missing type a NULL type was returned (as it was done previously), but all explicitly specified types should remain in place.

Example:

There were test cases like:

```
array(
    "SELECT * FROM Foo WHERE foo IN (:foo) OR bar = :bar",
    array('foo' => array(1, 2), 'bar' => 'bar'),
    array(':foo' => Connection::PARAM_INT_ARRAY),
    'SELECT * FROM Foo WHERE foo IN (?, ?) OR bar = ?',
    array(1, 2, 'bar'),
    array(\PDO::PARAM_INT, \PDO::PARAM_INT, \PDO::PARAM_STR)
)
```

without explicit type declaration, where `\PDO::PARAM_STR` was expected for not defined type.
So I changed expectations [there](https://github.com/doctrine/dbal/pull/2900/commits/32215904e96f7470696892e0fa9a635cce104537#diff-ce6b24584fe8728ee0f2b14993cd873aL306), because it returns NULL after my changes.

>  I think it would be acceptable if instead of a missing type a NULL type was returned (as it was done previously), but all explicitly specified types should remain in place.

Sounds like what I did, thanks for clariication.

> I think it's enough to have the algorithm documented somewhere in the ticket. 

:ok_hand: 

> Refactoring (if I read "make readable" correctly) for sake of readability may be too dangerous given how mission critical this piece of code is.

So I won't even try, thanks for advice.

PR #2402, opened 2016-05-22
[SQLite Platform] doctrine:database:create --if-not-exists fix

eugene-matvejev:
problem description: https://github.com/doctrine/DoctrineBundle/issues/542

PR #2903, opened 2017-10-31
[SQLite Platform] doctrine:database:create --if-not-exists fix #2402

eugenekurasov:
problem description: doctrine/DoctrineBundle#542
It's continue of Pull request #2402


eugenekurasov:
Hm... Why is database is not created?

PR #2849, opened 2017-09-09
Update build to use stages (adding PHPCS to check for CS violations)

lcobucci:
Since travis is now using Ubuntu Trusty as default (as explained in #2837) we need to create the schema manually in order to execute the tests. I've fixed it (applying basically the same solution that was done on #2835, #2838 and #2825), however I've also changed things to use the build stages (and added PHPCS to ensure that we follow our standards).

It's important to say that I've simplified the build jobs since MySQL 5.6 is the default version on Trusty.

PR #2907, opened 2017-11-07
Don't Set a Default Value on PostgreSQL SERIAL Fields

### LINKED ISSUE #2906: Cannot Generate [BIG]SERIAL Columns Without a Default or NOT NULL
The [postgres docs](https://www.postgresql.org/docs/current/static/ddl-default.html) have an example of a serial column without a `NOT NULL` declaration or default value.

```sql
CREATE TABLE example (
    colname SERIAL,
    -- ...
);
```
This `CREATE TABLE` can't be done in DBAL schema APIs.

Specifying `notnull` to `false` in the column options adds a `DEFAULT NULL` to the column which errors with...

> Syntax error: 7 ERROR:  multiple default values specified for column

Anyway, I'd expect to be able to do something like this:

```php
$table = new \Doctrine\DBAL\Schema\Table('example');
$table->addColumn('colname', 'integer', [
  'autoincrement' => true,
  'notnull' => false,
]);
```

To generate the example SQL above, without a `DEFAULT NULL` in there.

There's a work around for this (setting `default => "nextval('example_colname_seq')"`), but the point of having the serial type is the shorthand and avoiding the sequence stuff. Also ties the schema to postgres rather than being more generic.

chrisguitarguy:
Specifically when `'notnull' => false` is in the column options.

Right now DBAL adds a `DEFAULT NULL` when that happens. `[BIG]SERIAL` fields already have a default value and adding the `DEFAULT NULL` produces an error.

Fixes #2906

Not sure if the functional tests here are necessary, but it seemed like a good idea.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -26,7 +26,9 @@ use Doctrine\DBAL\Schema\Index;
 use Doctrine\DBAL\Schema\Sequence;
 use Doctrine\DBAL\Schema\TableDiff;
 use Doctrine\DBAL\Types\BinaryType;
+use Doctrine\DBAL\Types\BigIntType;
 use Doctrine\DBAL\Types\BlobType;
+use Doctrine\DBAL\Types\IntegerType;
 
 /**
  * PostgreSqlPlatform.
@@ -1185,5 +1187,23 @@ class PostgreSqlPlatform extends AbstractPlatform
         $str = str_replace('\\', '\\\\', $str); // PostgreSQL requires backslashes to be escaped aswell.
 
         return parent::quoteStringLiteral($str);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefaultValueDeclarationSQL($field)
+    {
+        if ($this->isSerialField($field)) {
+            return '';
+        }
+
+        return parent::getDefaultValueDeclarationSQL($field);
+    }
+
+    private function isSerialField(array $field) : bool
+    {
+        return $field['autoincrement'] ?? false === true && isset($field['type'])
+            && ($field['type'] instanceof IntegerType || $field['type'] instanceof BigIntType);
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
@@ -419,6 +419,50 @@ class PostgreSqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
         self::assertEquals(-1.1, $columns['col_decimal']->getDefault());
         self::assertEquals('(-1)', $columns['col_string']->getDefault());
     }
+
+    public static function serialTypes() : array
+    {
+        return [
+            ['integer'],
+            ['bigint'],
+        ];
+    }
+
+    /**
+     * @dataProvider serialTypes
+     * @group 2906
+     */
+    public function testAutoIncrementCreatesSerialDataTypesWithoutADefaultValue(string $type) : void
+    {
+        $tableName = "test_serial_type_$type";
+
+        $table = new Schema\Table($tableName);
+        $table->addColumn('id', $type, ['autoincrement' => true, 'notnull' => false]);
+
+        $this->_sm->dropAndCreateTable($table);
+
+        $columns = $this->_sm->listTableColumns($tableName);
+
+        self::assertNull($columns['id']->getDefault());
+    }
+
+    /**
+     * @dataProvider serialTypes
+     * @group 2906
+     */
+    public function testAutoIncrementCreatesSerialDataTypesWithoutADefaultValueEvenWhenDefaultIsSet(string $type) : void
+    {
+        $tableName = "test_serial_type_with_default_$type";
+
+        $table = new Schema\Table($tableName);
+        $table->addColumn('id', $type, ['autoincrement' => true, 'notnull' => false, 'default' => 1]);
+
+        $this->_sm->dropAndCreateTable($table);
+
+        $columns = $this->_sm->listTableColumns($tableName);
+
+        self::assertNull($columns['id']->getDefault());
+    }
 }
 
 class MoneyType extends Type
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
@@ -144,6 +144,63 @@ abstract class AbstractPostgreSqlPlatformTestCase extends AbstractPlatformTestCa
         self::assertEquals(array('CREATE TABLE autoinc_table (id SERIAL NOT NULL)'), $this->_platform->getCreateTableSQL($table));
     }
 
+    public static function serialTypes() : array
+    {
+        return [
+            ['integer', 'SERIAL'],
+            ['bigint', 'BIGSERIAL'],
+        ];
+    }
+
+    /**
+     * @dataProvider serialTypes
+     * @group 2906
+     */
+    public function testGenerateTableWithAutoincrementDoesNotSetDefault(string $type, string $definition) : void
+    {
+        $table  = new \Doctrine\DBAL\Schema\Table('autoinc_table_notnull');
+        $column = $table->addColumn('id', $type);
+        $column->setAutoIncrement(true);
+        $column->setNotNull(false);
+
+        $sql = $this->_platform->getCreateTableSQL($table);
+
+        self::assertEquals(["CREATE TABLE autoinc_table_notnull (id $definition)"], $sql);
+    }
+
+    /**
+     * @dataProvider serialTypes
+     * @group 2906
+     */
+    public function testCreateTableWithAutoincrementAndNotNullAddsConstraint(string $type, string $definition) : void
+    {
+        $table  = new \Doctrine\DBAL\Schema\Table('autoinc_table_notnull_enabled');
+        $column = $table->addColumn('id', $type);
+        $column->setAutoIncrement(true);
+        $column->setNotNull(true);
+
+        $sql = $this->_platform->getCreateTableSQL($table);
+
+        self::assertEquals(["CREATE TABLE autoinc_table_notnull_enabled (id $definition NOT NULL)"], $sql);
+    }
+
+    /**
+     * @dataProvider serialTypes
+     * @group 2906
+     */
+    public function testGetDefaultValueDeclarationSQLIgnoresTheDefaultKeyWhenTheFieldIsSerial(string $type) : void
+    {
+        $sql = $this->_platform->getDefaultValueDeclarationSQL(
+            [
+                'autoincrement' => true,
+                'type'          => Type::getType($type),
+                'default'       => 1,
+            ]
+        );
+
+        self::assertSame('', $sql);
+    }
+
     public function testGeneratesTypeDeclarationForIntegers()
     {
         self::assertEquals(


chrisguitarguy:
The scrutenizer failure here seems unrelated, just picked up because I changed a file with a pre-existing issue: https://scrutinizer-ci.com/g/doctrine/dbal/inspections/6801d978-162b-436c-a2e2-c1315ea461c5/issues/files/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php?status=new&orderField=path&order=asc&honorSelectedPaths=0

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -1186,4 +1188,23 @@ public function quoteStringLiteral($str)
 
         return parent::quoteStringLiteral($str);
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefaultValueDeclarationSQL($field)
+    {
+        if (self::isSerialField($field) && empty($field['notnull'])) {
```
lcobucci: This condition will result to `true` when `$field['notnull']` doesn't exist, is this the expected behaviour? I'd suggest to be more explicit about that, like `$field['notnull'] ?? false === false`.

lcobucci: I know that we use `empty()` on the parent method, but we didn't have these features back then.

lcobucci: Is there any reason to leave the `empty($field['notnull'])` out of the method?

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -1186,4 +1188,23 @@ public function quoteStringLiteral($str)
...
+            return '';
+        }
+
+        return parent::getDefaultValueDeclarationSQL($field);
+    }
+
+    private static function isSerialField($field)
```
lcobucci: Missing type declarations (parameters and return)

lcobucci: Also, any reason to make this static?

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -1186,4 +1188,23 @@ public function quoteStringLiteral($str)
...
+    {
+        return !empty($field['autoincrement']) && isset($field['type']) && (
```
lcobucci: Same comments about being explicit here (with `empty()`)

tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
```diff
@@ -419,6 +419,34 @@ public function testListNegativeColumnDefaultValue()
...
+    {
+        return [
+            ['integer'],
+            ['bigint'],
+        ];
+    }
+
+    /**
+     * @dataProvider serialTypes
+     * @group https://github.com/doctrine/dbal/issues/2906
```
lcobucci: Please use only `2906` as group

tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
```diff
@@ -144,6 +144,30 @@ public function testGenerateTableWithAutoincrement()
...
+    {
+        return [
+            ['integer', 'SERIAL'],
+            ['bigint', 'BIGSERIAL'],
+        ];
+    }
+
+    /**
+     * @dataProvider serialTypes
+     * @group https://github.com/doctrine/dbal/issues/2906
```
lcobucci: Please use only `2906` as group

tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
```diff
@@ -419,6 +419,34 @@ public function testListNegativeColumnDefaultValue()
         self::assertEquals(-1.1, $columns['col_decimal']->getDefault());
         self::assertEquals('(-1)', $columns['col_string']->getDefault());
     }
+
+    public static function serialTypes()
```
lcobucci: Add ` : array` as return type

tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
```diff
@@ -419,6 +419,34 @@ public function testListNegativeColumnDefaultValue()
...
+            ['integer'],
+            ['bigint'],
+        ];
+    }
+
+    /**
+     * @dataProvider serialTypes
+     * @group https://github.com/doctrine/dbal/issues/2906
+     */
+    public function testAutoIncrementCreatesSerialDataTypesWithoutADefaultValueWheNotNullIsFalse($type)
```
lcobucci: `public function testAutoIncrementCreatesSerialDataTypesWithoutADefaultValueWheNotNullIsFalse(string $type) : void`

tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
```diff
@@ -419,6 +419,34 @@ public function testListNegativeColumnDefaultValue()
...
+        $table->addColumn('id', $type, [
+            'autoincrement' => true,
+            'notnull' => false,
+        ]);
+
+        $this->_sm->dropAndCreateTable($table);
+
+        $columns = $this->_sm->listTableColumns($tableName);
+
+        $this->assertNull($columns['id']->getDefault());
```
lcobucci: Please use `self::assert*` instead

tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
```diff
@@ -144,6 +144,30 @@ public function testGenerateTableWithAutoincrement()
         self::assertEquals(array('CREATE TABLE autoinc_table (id SERIAL NOT NULL)'), $this->_platform->getCreateTableSQL($table));
     }
 
+    public static function serialTypes()
```
lcobucci: Add ` : array` as return type

tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
```diff
@@ -144,6 +144,30 @@ public function testGenerateTableWithAutoincrement()
...
+            ['integer', 'SERIAL'],
+            ['bigint', 'BIGSERIAL'],
+        ];
+    }
+
+    /**
+     * @dataProvider serialTypes
+     * @group https://github.com/doctrine/dbal/issues/2906
+     */
+    public function testGenerateTableWithAutoincrementAndNotNullFalseDoesNotSetADefaultValueOnSerialFields($type, $definition)
```
lcobucci: Please add type declarations here

tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
```diff
@@ -144,6 +144,30 @@ public function testGenerateTableWithAutoincrement()
...
+    {
+        $table = new \Doctrine\DBAL\Schema\Table('autoinc_table_notnull');
+        $column = $table->addColumn('id', $type);
+        $column->setAutoIncrement(true);
+        $column->setNotNull(false);
+
+        $sql = $this->_platform->getCreateTableSQL($table);
+
+        self::assertEquals(array("CREATE TABLE autoinc_table_notnull (id $definition)"), $sql);
```
lcobucci: Please use short-array notation

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
lcobucci: Also do we have unit tests covering all the paths?

lcobucci (changes requested):
Nice job, great commit organisation too.

I got some suggestions, questions, and nitpicking - obviously.

Can you please make sure that the code you're adding follows our standards? `phpcs` is already configured - with `phpcbf` - and can be helpful (I've tried to make it mandatory and simplify the lives of new contributors on #2849 but @Ocramius said NOPE 😂).

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
chrisguitarguy: > Missing type declarations (parameters and return)

Left out the parameter type declaration because the method from which this is called doesn't have one either (`getDefaultValueDeclarationSQL`). `$field` theoretically should be `array`? Happy to add the type declaration for `$field`, but that was the reason I didn't.

Definitely should have the return type!

> Also, any reason to make this static?

Habit of my, sorry. I tend to make "pure" utility functions static -- since nothing from the platform's instance state is required, I made it static. Want me to make it non-static?

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
chrisguitarguy: > Is there any reason to leave the empty($field['notnull']) out of the method?

Yep. `nonnull` doesn't have anything to do with being a `SERIAL` field. Also: this is perfectly okay: `whatever SERIAL NOT NULL`. So if someone wants to make sure the field is `notnull` they are free to do so.

> This condition will result to true when $field['notnull'] doesn't exist, is this the expected behaviour?

`$field['notnull']` missing is the same as `$field['notnull'] = false` yeah? I think you're right though, should be more explicit.

> Also do we have unit tests covering all the paths?

I need to add some for `whatever SERIAL NOT NULL` now that I think about it.

chrisguitarguy: Okay, coming back to this `notnull` thing. I'm not actually it sure it should get checked at all. Any default value for a serial field is invalid -- they already have defaults built in. Maybe `getDefaultValueDeclarationSQL` should just return its empty string for any `SERIAL` field?

chrisguitarguy:
> phpcs is already configured - with phpcbf - and can be helpful

I did the best I could with this. Find some errors in the tests with aligned equals signs. Running either tool on the whole codebase produces a lot of violations, even running it on the files I edited were 40 or 50 fixes per file.

morozov:
Time for a shameless self-promotion:
```
pull-request doctrine dbal 2907 --standard=vendor/doctrine/coding-standard/lib/Doctrine

FILE: tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
--------------------------------------------------------------------------------
FOUND 0 ERRORS AND 2 WARNINGS AFFECTING 2 LINES
--------------------------------------------------------------------------------
 159 | WARNING | Line exceeds 120 characters; contains 147 characters
 175 | WARNING | Line exceeds 120 characters; contains 149 characters
--------------------------------------------------------------------------------
```

@chrisguitarguy, see [morozov/diff-sniffer-pull-request](https://github.com/morozov/diff-sniffer-pull-request) and [morozov/diff-sniffer-pre-commit](https://github.com/morozov/diff-sniffer-pre-commit).

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
lcobucci: I got your point. I don't see much difference in this case, since the caller is already in the scope of the instance. I'd keep it non-static here but was just curious 😄

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -1186,4 +1188,23 @@ public function quoteStringLiteral($str)
...
+    public function getDefaultValueDeclarationSQL($field)
+    {
+        if (self::isSerialField($field)) {
+            return '';
+        }
+
+        return parent::getDefaultValueDeclarationSQL($field);
+    }
+
+    private static function isSerialField($field) : bool
```
lcobucci: Regarding the type, we just don't have the type declaration on `getDefaultValueDeclarationSQL` for historical reasons. Let's add it here too

lcobucci (changes requested):
Great job @chrisguitarguy, just one more thing (apart from what @morozov said).

chrisguitarguy:
Thanks, @morozov, those tools are great!

Updated with the latest requested changes and fixed the PHP CS warnings, @lcobucci.

lcobucci (approved):

lcobucci:
@chrisguitarguy merged and backported to 2.6, thanks for your contribution!

chrisguitarguy:
My pleasure, thanks for the support @lcobucci!

PR #2911, opened 2017-11-08
Add SQL file logger

BenMorel:
I've been using a SQL file logger for years, that proved very helpful during development, are you interested to merge it?

It produces an output similar to:

```
SQL logger starting on Wed, 08 Nov 2017 16:30:05 +0000

Query 1: SELECT ... WHERE t0.id = ?
Parameters: 5
Types: 'integer'
This query: 0.001 seconds; total: 0.001 seconds.

Query 2: SELECT ... WHERE d0_.restaurantId = ? AND d0_.status IN (?)
Parameters: 5, ['active']
Types: 'integer', 102
This query: 0.001 seconds; total: 0.002 seconds.

Query 3: SELECT ... ST_Contains(z0_.boundary, ST_GeomFromWKB(?, 4326)) = 1 AND z0_.status <> ?
Parameters: Brick\Geo\Point(POINT (55.2238832 -21.0401197)), 'deleted'
Types: 'Point', 2
This query: 0.026 seconds; total: 0.028 seconds.
```

*Note: ellipses are mine, the logger outputs the full query.*

It writes a compact representation of parameters, representing objects as their class name along with their string representation if available, or their hash otherwise.

### DIFF
--- a/lib/Doctrine/DBAL/Logging/FileLogger.php
+++ b/lib/Doctrine/DBAL/Logging/FileLogger.php
@@ -1,0 +1,152 @@
+<?php
+
+namespace Doctrine\DBAL\Logging;
+
+/**
+ * A SQL logger that logs to a file.
+ */
+class FileLogger implements SQLLogger
+{
+    /**
+     * Log file pointer.
+     *
+     * @var resource
+     */
+    private $fp;
+
+    /**
+     * Query start time.
+     *
+     * @var float
+     */
+    private $startTime;
+
+    /**
+     * Total time spent on all queries.
+     *
+     * @var float
+     */
+    private $totalTime = 0.0;
+
+    /**
+     * @var int
+     */
+    private $queryNumber = 0;
+
+    /**
+     * @var bool
+     */
+    private $lock;
+
+    /**
+     * Class constructor.
+     *
+     * @param string $filename The log file name.
+     * @param bool   $lock     Whether to use locks when writing to the file.
+     */
+    public function __construct(string $filename, bool $lock = false)
+    {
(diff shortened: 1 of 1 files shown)


lcobucci:
@BenMorel thanks for your contribution. The logger can indeed be useful but I think not everyone might need it, so maybe you could be offered as a separate package. You can create a small lib that depends on DBAL and provides that SQLLogger implementation. What do you think?

morozov:
I agree with @lcobucci. Only the interface belongs to the package itself. All implementations (including the existing ones) don't. Otherwise, we may end up implementing all possible loggers here.

Ocramius:
Probably better to generally use a PSR-3 logger?

Marco Pivetta

http://twitter.com/Ocramius

http://ocramius.github.com/

BenMorel:
Okay... I thought that basic implementations (the ones that don't require extra packages or extensions) did belong to this package.

greg0ire:
Quick implementation:

```php
<?php

namespace Svod\Infrastructure\Persistence\Doctrine;

use Doctrine\DBAL\Logging\SQLLogger;
use Psr\Log\LoggerInterface;

final class PsrSQLLogger implements SQLLogger
{
    /**
     * @var LoggerInterface
     */
    private $logger;

    public function __construct(LoggerInterface $logger)
    {
        $this->logger = $logger;
    }

    /**
     * {@inheritDoc}
     */
    public function startQuery($sql, array $params = null, array $types = null)
    {
        $this->logger->debug(
            $sql,
            compact('params', 'types')
        );
    }

    /**
     * {@inheritDoc}
     */
    public function stopQuery()
    {
    }
}

```

Works for me :tm:

PR #2912, opened 2017-11-09
Allow to pass DateTime to DateTimeImmutableType

ossinkine:
Actually this is the same as #2794 but for immutable types.
It would be nice to be able to set DateTime object to DateTimeImmutableType.

### DIFF
--- a/lib/Doctrine/DBAL/Types/DateImmutableType.php
+++ b/lib/Doctrine/DBAL/Types/DateImmutableType.php
@@ -46,7 +46,7 @@ class DateImmutableType extends DateType
             return $value;
         }
 
-        if ($value instanceof \DateTimeImmutable) {
+        if ($value instanceof \DateTimeInterface) {
             return $value->format($platform->getDateFormatString());
         }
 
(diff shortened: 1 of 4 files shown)


Ocramius:
This is a mistake: make sure your userland code passes right value types to correct endpoints. This patch seems to be fixing a symptom, not the root cause of your problem.

ossinkine:
I set up `datetime_immutable` type to entity field to receive object from DB with immutable value.
But I want to be able to save any DateTimeInetrface object into DB.
Imagine you have a service that generates `DateTimeInterface` object which you going to save into DB. Check that the object really has immutable interface and create a new one if it is not so, it is too boring and senselessly.

Ocramius:
> Imagine you have a service that generates `DateTimeInterface` object which you going to save into DB. 

That requires a `DateTimeInterfaceType` instead

ossinkine:
Value of which class the `DateTimeInterfaceType` will be get from DB to PHP?

ossinkine:
By the same logic `StringType` should validate the value is string but I am still able to pass the object with `__toString` method instead of string.

Ocramius:
Yes, mistakes were made when designing past types.

No more though, please ;-)

lcobucci:
@ossinkine segregating mutable from immutable datetime objects was a design decision,  #2795 was just created to avoid breaking BC. On https://github.com/doctrine/dbal/pull/2795/files#diff-4d2ff4e7be0d7e0ab0eeb9508cb07096R76 you can also see our notes regarding that decision.

lcobucci:
Closing as `wont-fix` as per comments.

PR #2914, opened 2017-11-13
Change visibility of some Connection properties

DavG:
The purpose of this PR is to allow more flexibility in Connection's child, for example, when creating a ConnectionWrapper.

For example, I have a ConnectionWrapper which dynamically change the database name (based on some JWT informations), so with this PR I wouldn't have to duplicate $_params and some other functions.

I think other cases could take advantages of it.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -118,14 +118,14 @@ class Connection implements DriverConnection
      *
      * @var boolean
      */
-    private $_isConnected = false;
+    protected $_isConnected = false;
 
     /**
      * The current auto-commit mode of this connection.
      *
      * @var boolean
      */
-    private $autoCommit = true;
+    protected $autoCommit = true;
 
     /**
      * The transaction nesting level.
@@ -153,7 +153,7 @@ class Connection implements DriverConnection
      *
      * @var array
      */
-    private $_params = [];
+    protected $_params = [];
 
     /**
      * The DatabasePlatform object that provides information about the


Ocramius:
Please avoid extending the Connection class, and implement the `DriverConnection` instead.

Also, `Connection` `private` members are not covered by BC policy, so we really shouldn't make them `protected`, as that would hugely limit the ability of DBAL to be further developed.

DavG:
Hello,

I know about the DriverConnection interface, but honestly it's quite frustrating to re-write it all, just because I need to edit $_params before the connect().
Furthermore, i would copy/paste 99% of the original code, I don't like to do that. Not readable, not (automatically) updatable.

I don't get your point about your BC policy, where is the risk ?
The behavior won't changes at all, except if Connection Child wants.

Ocramius:
> I need to edit $_params before the connect().

Why not add a factory then?

DavG:
I don't think it would work in all my cases. Sometime (in CLI) I need to connect successively to multiple databases, using the same (doctrine) connection (I cannot configure them all in advance, we have > 2500 databases, and it increases everyday)

But maybe I misunderstand you. Would you give me more details ?

Ocramius:
I think you are solving the problem on the wrong end. What you probably
need to look at is the abstract factory pattern, in which you can
paranetrize service instantiation. The connection object is not intended to
be mutable, especially via subclassing.

DavG:
Ok, I'll take a look in this direction.

For the record, we are using a Connection sub-class in production, it works very well for about 2 years.
In fact I opened this PR because i'm doing some refactoring and i was thinking, "it's too bad we need to copy paste some doctrine code."

I'll keep you in touch !

PR #2918, opened 2017-11-16
Type and name sql argument of query function

rey0bs:
Several functions use first argument of `Connection::query` function. This argument is named and typed as string in functions which use it, but is not documented in `query` function.

For example, in `Doctrine\DBAL\Connection`, the logger uses it in `startQuery` function. `$sql` is named and typed in `SQLLogger` interface, but not in `query` function.

This PR names this argument to increase visibility and avoid to use `func_get_args` and `$args[0]` for `$sql` parameter.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1006,25 +1006,26 @@ class Connection implements DriverConnection
     /**
      * Executes an SQL statement, returning a result set as a Statement object.
      *
+     * @param string $sql
+     * @param array $args
+     *
      * @return \Doctrine\DBAL\Driver\Statement
      *
      * @throws \Doctrine\DBAL\DBALException
      */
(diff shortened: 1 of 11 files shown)


lib/Doctrine/DBAL/Connection.php
```diff
@@ -1006,25 +1006,26 @@ public function project($query, array $params, Closure $function)
...
      *
+     * @param string $sql
+     * @param array $args
+     *
      * @return \Doctrine\DBAL\Driver\Statement
      *
      * @throws \Doctrine\DBAL\DBALException
      */
-    public function query()
+    public function query($sql, ...$args)
```
Majkl578: This change is a BC break and couldn't be considered for 2.x.

morozov:
Obsolete as of https://github.com/doctrine/dbal/pull/3803.

PR #2919, opened 2017-11-18
Add tests for column collation to prove it works

Tobion:
Fixes #2551

The main point is that one must use the platform or the custom option for the `collation` and not the standard option for columns. The reasoning I guess is that those options are not cross-RDBMS compatible. There are two things that are inconsistent and causing alot of the confusion:

1. platform vs custom schema option (both result in the same outcome)
    - SchemaManager uses `\Doctrine\DBAL\Schema\Column::setPlatformOption`, see https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php#L203
    - SchemaTool uses `\Doctrine\DBAL\Schema\Column::setCustomSchemaOption`, see https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Tools/SchemaTool.php#L460
2. on the column level the option is called `collation` but on the table level the option is called `collate` (only relevant for mysql, see https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L499 )

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
@@ -474,7 +474,7 @@ abstract class AbstractMySQLPlatformTestCase extends AbstractPlatformTestCase
             "ALTER TABLE mytable ADD PRIMARY KEY (foo)",
         ), $sql);
     }
-    
+
     public function testAlterPrimaryKeyWithNewColumn()
     {
         $table = new Table("yolo");
@@ -484,7 +484,7 @@ abstract class AbstractMySQLPlatformTestCase extends AbstractPlatformTestCase
 
         $comparator = new Comparator();
         $diffTable = clone $table;
-        
+
         $diffTable->addColumn('pkc2', 'integer');
         $diffTable->dropPrimaryKey();
         $diffTable->setPrimaryKey(array('pkc1', 'pkc2'));
@@ -496,7 +496,7 @@ abstract class AbstractMySQLPlatformTestCase extends AbstractPlatformTestCase
                 'ALTER TABLE yolo ADD PRIMARY KEY (pkc1, pkc2)',
             ),
             $this->_platform->getAlterTableSQL($comparator->diffTable($table, $diffTable))
-        );      
+        );
     }
 
     public function testInitializesDoctrineTypeMappings()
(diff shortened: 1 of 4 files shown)


Tobion:
@Ocramius this is ready

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
```diff
@@ -907,4 +907,30 @@ public function testListTableForeignKeysSQLEvaluatesDatabase()
         self::assertContains('bar', $sql);
         self::assertNotContains('DATABASE()', $sql);
     }
+
+    public function testSupportsColumnCollation()
```
Majkl578: Can you please add ` : void` for all new test methods?

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
```diff
@@ -907,4 +907,30 @@ public function testListTableForeignKeysSQLEvaluatesDatabase()
...
+    }
+
+    public function testGetCreateTableSQLWithColumnCollation()
+    {
+        $table = new Table('foo');
+        $table->addColumn('no_collation', 'string');
+        $table->addColumn('column_collation', 'string')->setPlatformOption('collation', 'ascii_general_ci');
+
+        self::assertSame(
+            array('CREATE TABLE foo (no_collation VARCHAR(255) NOT NULL, column_collation VARCHAR(255) NOT NULL COLLATE ascii_general_ci) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB'),
```
Majkl578: Short array syntax should be used.

Majkl578 (commented):
Retriggered the continuousphp build, looks ok now.

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
Tobion: Sure but shouldn't that be changed everywhere? Just doing it here is inconsistent.

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
Tobion: Same as above

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
Majkl578: True, CS is currently not enforced on project level, except checking for some basic rules for new PRs, But soon will be - if we do it like this now, it won't need further fixes in few weeks. 👍

Tobion:
@Majkl578 fixed CS.

Tobion:
@Majkl578 @Ocramius please merge

Tobion:
Rebased. Please merge

Ocramius (approved):

Ocramius:
:+1:

Ocramius:
Had to restart some builds, but otherwise all is green now :) Thanks @Tobion!

Issue #1861, opened 2013-10-27
DBAL-642: Generated IDs are not guaranteed to be unique over the table's lifetime in SQLite


doctrinebot:
Jira issue originally created by user flack:

In

http://docs.doctrine-project.org/en/2.0.x/reference/basic-mapping.html#identifier-generation-strategies

it says that MySQL and SQLite use AUTO_INCREMENT by default. The generated SQL for creating an ID field with GENERATOR_TYPE_AUTO looks like this (abbreviated 
for readability):

```
CREATE*TABLE*TEST (id INTEGER DEFAULT 0 NOT NULL, PRIMARY KEY(id))
```

http://www.sqlite.org/faq.html#q1 states:

If you declare a column of a table to be INTEGER PRIMARY KEY, then whenever you insert a NULL into that column of the table, the NULL is automatically converted into an integer which is one greater than the largest value of that column over all other rows in the table, or 1 if the table is empty. [...] Note that the integer key is one greater than the largest key that was in the table just prior to the insert. The new key will be unique over all keys currently in the table, but it might overlap with keys that have been previously deleted from the table.

So in other words, if you remove an entity and then create a new one, the new one will have the same id as the previously deleted one. If you do both operations on the same entitymanager, id references (in proxies f.x.) will suddenly get confused and point to something else (at least that's my current theory..)

The point is: SQLite doesn't act like MySQL as the documentation implies, and IMHO SQLite's current behaviour makes it useless for more complex scenarios. I've found some reference to this problem here:

https://github.com/doctrine/dbal/pull/66

Unfortunately, it doesn't mention any solution. The problem is that you can't override the columnDefinition, because it would have to be "INTEGER PRIMARY KEY AUTOINCREMENT", but then, you get an exception, because the Platform appends ", PRIMARY KEY(id)", so it's defined twice

flack:
I think this can be closed (implemented by #3141)

PR #2921, opened 2017-11-20
allow creating PRIMARY KEY AUTOINCREMENT fields for sqlite

icewind1991:
If a PRIMARY KEY field is defined in sqlite without autoincrement then
keys might be reused when rows are deleted, explicitly settings autoincrement
will prevent this.

Autoincrement is not enabled by default due to additional overhead introduced
with autoincrement bookkeeping and should only be enabled when the additional
uniqueness is required.

See also https://github.com/doctrine/dbal/issues/1861


Majkl578:
Tests are not passing here...

TimoBakx:
I added unit tests in PR #3141.

TimoBakx:
@Majkl578 The commits in this PR were merged through #3141. This PR may be closed.

Majkl578:
Cool. :+1:

PR #2929, opened 2017-11-30
DBAL-2925: Added functional test checking ability to change PK for po…

SCIF:
…stgres platform

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -2080,6 +2080,7 @@ abstract class AbstractPlatform
         foreach ($diff->changedIndexes as $index) {
             if ($index->isPrimary()) {
                 $sql[] = $this->getDropPrimaryKeySQL($tableName);
+
                 continue;
             }
 
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -2078,6 +2078,11 @@ abstract class AbstractPlatform
             $sql[] = $this->getDropIndexSQL($index, $tableName);
         }
         foreach ($diff->changedIndexes as $index) {
+            if ($index->isPrimary()) {
+                $sql[] = $this->getDropPrimaryKeySQL($tableName);
+                continue;
+            }
+
             $sql[] = $this->getDropIndexSQL($index, $tableName);
         }
 
@@ -3604,5 +3609,15 @@ abstract class AbstractPlatform
     protected function getLikeWildcardCharacters() : string
     {
         return '%_';
+    }
+
+    /**
+     * @param string $table
+     *
+     * @return string
+     */
+    protected function getDropPrimaryKeySQL(string $table): string
+    {
+        return 'ALTER TABLE ' . $table . ' DROP PRIMARY KEY';
     }
 }
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -3611,12 +3611,7 @@ abstract class AbstractPlatform
         return '%_';
     }
 
-    /**
-     * @param string $table
-     *
-     * @return string
-     */
-    protected function getDropPrimaryKeySQL(string $table): string
+    protected function getDropPrimaryKeySQL(string $table) : string
     {
         return 'ALTER TABLE ' . $table . ' DROP PRIMARY KEY';
     }
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -755,6 +755,14 @@ class PostgreSqlPlatform extends AbstractPlatform
     /**
      * {@inheritDoc}
      */
+    protected function getDropPrimaryKeySQL(string $table): string
+    {
+        return $this->getDropConstraintSQL("{$table}_pkey", $table);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
     protected function _getCreateTableSQL($tableName, array $columns, array $options = [])
     {
         $queryFields = $this->getColumnDeclarationListSQL($columns);
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -754,7 +754,7 @@ class PostgreSqlPlatform extends AbstractPlatform
 
     protected function getDropPrimaryKeySQL(string $table) : string
     {
-        return $this->getDropConstraintSQL("{$table}_pkey", $table);
+        return $this->getDropConstraintSQL($table . '_pkey', $table);
     }
 
     /**
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -752,10 +752,7 @@ class PostgreSqlPlatform extends AbstractPlatform
         return $this->getDropConstraintSQL($foreignKey, $table);
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    protected function getDropPrimaryKeySQL(string $table): string
+    protected function getDropPrimaryKeySQL(string $table) : string
     {
         return $this->getDropConstraintSQL("{$table}_pkey", $table);
     }
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -1622,15 +1622,17 @@ class SQLServerPlatform extends AbstractPlatform
 
     protected function getDropPrimaryKeySQL(string $table) : string
     {
-        return "
-DECLARE @table NVARCHAR(512), @sql NVARCHAR(MAX);
+        return <<<EOT
+DECLARE @table NVARCHAR(512), @sql NVARCHAR(MAX), @tableid NVARCHAR(512);
 SELECT @table = N'{$table}';
+SELECT @tableid = OBJECT_ID(@table);
 SELECT @sql = 'ALTER TABLE ' + @table + ' DROP CONSTRAINT '  + name + ';'
     FROM sys.key_constraints
     WHERE [type] = 'PK'
-    AND [parent_object_id] = OBJECT_ID(@table);
+    AND [parent_object_id] = @tableid;
 
-EXEC sp_executeSQL @sql;";
+EXEC sp_executeSQL @sql;
+EOT;
     }
 
 }
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -1619,4 +1619,18 @@ class SQLServerPlatform extends AbstractPlatform
 
         return strtoupper(dechex(crc32($identifier->getName())));
     }
+
+    protected function getDropPrimaryKeySQL(string $table) : string
+    {
+        return "
+DECLARE @table NVARCHAR(512), @sql NVARCHAR(MAX);
+SELECT @table = N'{$table}';
+SELECT @sql = 'ALTER TABLE ' + @table + ' DROP CONSTRAINT '  + name + ';'
+    FROM sys.key_constraints
+    WHERE [type] = 'PK'
+    AND [parent_object_id] = OBJECT_ID(@table);
+
+EXEC sp_executeSQL @sql;";
+    }
+
 }
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
@@ -502,6 +502,34 @@ class PostgreSqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
             'bigint->int' => ['bigint', 'integer', 'INT'],
         ];
     }
+
+    /**
+     * @group 2925
+     */
+    public function testAlterTableChangePrimaryKey() : void
+    {
+        $tableFrom = new Table('primary_key_userid');
+        $column = $tableFrom->addColumn('user_id', 'integer');
+        $column->setAutoincrement(true);
+        $tableFrom->setPrimaryKey(['user_id']);
+        $this->_sm->dropAndCreateTable($tableFrom);
+
+        $tableFrom = $this->_sm->listTableDetails('primary_key_userid');
+        self::assertTrue($tableFrom->getColumn('user_id')->getAutoincrement());
+
+        $tableTo = new Table('primary_key_id');
+        $column = $tableTo->addColumn('id', 'integer');
+        $column->setAutoincrement(true);
+        $tableTo->setPrimaryKey(['id']);
+
+        $c = new Comparator();
+        $diff = $c->diffTable($tableFrom, $tableTo);
+        self::assertInstanceOf(TableDiff::class, $diff, "There should be a difference and not false being returned from the table comparison");
+
+        $this->_sm->alterTable($diff);
+        $tableFinal = $this->_sm->listTableDetails('autoinc_type_modification');
+        self::assertTrue($tableFinal->getColumn('id')->getAutoincrement());
+    }
 }
 
 class MoneyType extends Type
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
@@ -509,7 +509,7 @@ class PostgreSqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
     public function testAlterTableChangePrimaryKey() : void
     {
         $tableFrom = new Table('primary_key_userid');
-        $column = $tableFrom->addColumn('user_id', 'integer');
+        $column    = $tableFrom->addColumn('user_id', 'integer');
         $column->setAutoincrement(true);
         $tableFrom->setPrimaryKey(['user_id']);
         $this->_sm->dropAndCreateTable($tableFrom);
@@ -518,11 +518,11 @@ class PostgreSqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
         self::assertTrue($tableFrom->getColumn('user_id')->getAutoincrement());
 
         $tableTo = new Table('primary_key_id');
-        $column = $tableTo->addColumn('id', 'integer');
+        $column  = $tableTo->addColumn('id', 'integer');
         $column->setAutoincrement(true);
         $tableTo->setPrimaryKey(['id']);
 
-        $c = new Comparator();
+        $c    = new Comparator();
         $diff = $c->diffTable($tableFrom, $tableTo);
         self::assertInstanceOf(TableDiff::class, $diff, "There should be a difference and not false being returned from the table comparison");
 
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
@@ -502,34 +502,6 @@ class PostgreSqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
             'bigint->int' => ['bigint', 'integer', 'INT'],
         ];
     }
-
-    /**
-     * @group 2925
-     */
-    public function testAlterTableChangePrimaryKey() : void
-    {
-        $tableFrom = new Table('primary_key_userid');
-        $column    = $tableFrom->addColumn('user_id', 'integer');
-        $column->setAutoincrement(true);
-        $tableFrom->setPrimaryKey(['user_id']);
-        $this->_sm->dropAndCreateTable($tableFrom);
-
-        $tableFrom = $this->_sm->listTableDetails('primary_key_userid');
-        self::assertTrue($tableFrom->getColumn('user_id')->getAutoincrement());
-
-        $tableTo = new Table('primary_key_id');
-        $column  = $tableTo->addColumn('id', 'integer');
-        $column->setAutoincrement(true);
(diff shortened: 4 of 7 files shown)


lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3570,4 +3575,14 @@ public function getStringLiteralQuoteCharacter()
     {
         return "'";
     }
+
```
simPod: I think the phpdoc is redundant

SCIF: Hi @simPod ,

Yeah, you are correct, it's totally useless and I will drop it soon.

PapsOu:
@simPod Does this PR is ready to be merged ? I've got the same problem for PK under postgresql.

Majkl578:
@PapsOu Unfortunately nobody has simply got to reviewing it yet.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -752,6 +752,11 @@ public function getDropForeignKeySQL($foreignKey, $table)
         return $this->getDropConstraintSQL($foreignKey, $table);
     }
 
+    protected function getDropPrimaryKeySQL(string $table) : string
+    {
+        return $this->getDropConstraintSQL("{$table}_pkey", $table);
```
Majkl578: Please use string concatenation here + `'` instead of variable in string.

tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
```diff
@@ -502,6 +502,34 @@ public function autoIncrementTypeMigrations() : array
...
+        self::assertTrue($tableFrom->getColumn('user_id')->getAutoincrement());
+
+        $tableTo = new Table('primary_key_id');
+        $column  = $tableTo->addColumn('id', 'integer');
+        $column->setAutoincrement(true);
+        $tableTo->setPrimaryKey(['id']);
+
+        $c    = new Comparator();
+        $diff = $c->diffTable($tableFrom, $tableTo);
+        self::assertInstanceOf(TableDiff::class, $diff, "There should be a difference and not false being returned from the table comparison");
```
Majkl578: Constant strings should be enclosed in `'`.

tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
```diff
@@ -502,6 +502,34 @@ public function autoIncrementTypeMigrations() : array
             'bigint->int' => ['bigint', 'integer', 'INT'],
         ];
     }
+
+    /**
+     * @group 2925
+     */
+    public function testAlterTableChangePrimaryKey() : void
```
morozov: Please move this test to `SchemaManagerFunctionalTestCase`. It should be checked and work the same for all platforms, not only for PostgreSQL.

tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
morozov: > There should be a difference and not false

It's a very vague assertion. I'd omit it together with the previous one about the autoincrement and only focus on the end result where the new primary key is `(id)`.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2065,6 +2065,11 @@ protected function getPreAlterTableIndexForeignKeySQL(TableDiff $diff)
             $sql[] = $this->getDropIndexSQL($index, $tableName);
         }
         foreach ($diff->changedIndexes as $index) {
+            if ($index->isPrimary()) {
+                $sql[] = $this->getDropPrimaryKeySQL($tableName);
```
morozov: @Majkl578 is there a way to enforce a new line before flow control operators like `return`, `continue`, etc on the code sniffer level? It would be very nice to have.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
morozov: I found a [couple](http://razorsql.com/features/postgresql_drop_primary_key.html) of [articles](https://gist.github.com/scaryguy/6269293) explaining that this is the way how the PK should be dropped on PostgreSQL but before seeing the explanation, this approach looks like using a "magic" suffix.

It would be nice if a link to documentation describing this approach was provided in the doc-block.

morozov (changes requested):

tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
Majkl578: No sniff I'm aware of, but shouldn't be hard to implement. Just needs to be compatible i.e. with single statements:
```php
if (...) {
    continue;
}
```
or comments:
```php
foo();

// comment
return 123;
```

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
Majkl578: No sniff I'm aware of, but shouldn't be hard to implement. Just needs to be compatible i.e. with single statements:
```php
if (...) {
    continue;
}
```
or comments:
```php
foo();

// comment
return 123;
```

morozov: Yeah. More generically the first case is if the previous line has flow control statement which is a block opener, there shouldn't be an empty line. Like:
```php
function x()
{
    foreach (...) {
        foreach (...) {
        }
    }

    foreach (...) {
    }
}
```

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1394,4 +1394,31 @@ public function testCreateAndListSequences() : void
...
+    public function testAlterTableChangePrimaryKey() : void
+    {
+        $tableFrom = new Table('primary_key_userid');
+        $column    = $tableFrom->addColumn('user_id', 'integer');
+        $column->setAutoincrement(true);
+        $tableFrom->setPrimaryKey(['user_id']);
+        $this->_sm->dropAndCreateTable($tableFrom);
+
+        $tableFrom = $this->_sm->listTableDetails('primary_key_userid');
+        self::assertTrue($tableFrom->getColumn('user_id')->getAutoincrement());
```
SCIF: I can see that this test is failed on Oracle. Could you advise what's the proper way to deal with it (in code check for platform and skip?)?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
SCIF: I've [added new line](https://github.com/doctrine/dbal/pull/2929/files#diff-3ee36db88cb7fcfb3df4415f6952fcd0R2083) before `continue`.

SCIF:
Hi @Majkl578 , @morozov ,

I've fixed MsSQL and MySQL issues discovered after moving test to common Schema testcase.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1619,4 +1619,18 @@ private function generateIdentifierName($identifier)
 
         return strtoupper(dechex(crc32($identifier->getName())));
     }
+
+    protected function getDropPrimaryKeySQL(string $table) : string
+    {
+        return "
```
morozov: Could you use heredoc syntax instead?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1619,4 +1619,18 @@ private function generateIdentifierName($identifier)
...
+DECLARE @table NVARCHAR(512), @sql NVARCHAR(MAX);
+SELECT @table = N'{$table}';
+SELECT @sql = 'ALTER TABLE ' + @table + ' DROP CONSTRAINT '  + name + ';'
+    FROM sys.key_constraints
+    WHERE [type] = 'PK'
+    AND [parent_object_id] = OBJECT_ID(@table);
```
morozov: Would it make sense to move the object ID outside of the `@sql` so that the function is not invoked for every single PK in `sys.key_constraints`?

How will this query behave if the table doesn't have a PK? I'd expect `sp_executeSQL` be invoked with an empty argument and throwing some sort of exception.

morozov (commented):
@SCIF thanks for the cross-platform test. As you can see, it's already bringing value :-)

Please see my comments and address the code style issues.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: @carusogabriel do you want to file this as a requirement for coding standards (I see we have disabled issues for the repository). It'd be nice to have a reference point somewhere. Otherwise, I constantly forget where the discussion started.

/cc @deeky666

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2078,6 +2078,12 @@ protected function getPreAlterTableIndexForeignKeySQL(TableDiff $diff)
             $sql[] = $this->getDropIndexSQL($index, $tableName);
         }
         foreach ($diff->changedIndexes as $index) {
+            if ($index->isPrimary()) {
```
morozov: The [Scrutinizer failure](https://scrutinizer-ci.com/g/doctrine/dbal/inspections/b32d890c-c010-47f9-acf7-7aea17c498cb) reveals this as a decent candidate for refactoring in the future. Instead of the platform knowing how to drop a given index based on its properties (primary, non-primary and so on), the index should tell the platform how to drop itself (sort of a Visitor pattern):
```php
interface Index
{
    function drop(Platform $platform);
}

interface Platform
{
    function dropIndex(string $name);
    function dropPrimaryKey(string $tableName);
}
```
This is outside of scope, just noting down.

/cc @Ocramius @deeky666

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
carusogabriel: @morozov Require new lines before control structions you mean? :)

morozov: Yes. Before and after.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
SCIF: I was thinking about it but that will lead to creation of platform specific indices. Quite heavy refactoring will be required, I worry. I'm afraid I'm not able to do it in reasonable time.

Majkl578: Not really fan of the suggested visitor approach. Index would then need to know how to represent itself on all existing platforms and even those not provided by DBAL out-of-box.

morozov: @SCIF no need to worry about it right now. It's just for the record.

morozov: > Not really fan of the suggested visitor approach. Index would then need to know how to represent itself on all existing platforms and even those not provided by DBAL out-of-box.

Not exactly. They will need to be able to identify themselves as a type while the platform will know how to handle certain types. Same as we do with column types.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
SCIF: Emm. If you will try to drop PK for any other platforms on table not containing it, exception will be thrown as well, so I don't see any inconsistency here.

morozov: Is it the same type of exception?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
SCIF: @morozov , Totally agree with this proposal, but code marked as still required your review :) Mark it properly please.

morozov: @SCIF it's not because of the formatting issue. I want to see if the behavior of dropping a non-existing PK is consistent with other platforms.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
SCIF: Hmm, different one, you are right. Just `syntax error` vs `can't drop unexisted PK`. What's your proposal? I don't have windows pc and that is my very first mssql experience in life :)

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
SCIF: This thread is regarding cs only, the PK ones [here](https://github.com/doctrine/dbal/pull/2929#discussion_r178398463)

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
morozov: You don't need a Windows PC. The SQL Server extension can be [compiled for PHP on Linux](https://github.com/Microsoft/msphpsql/releases/tag/v5.2.0-RC), and the server itself can be running in a Docker container ([microsoft/mssql-server-linux](https://hub.docker.com/r/microsoft/mssql-server-linux/)).

I don't have a proposal currently. My only concern is this code may execute knowingly syntactically invalid SQL. Instead of doing the in the dynamic SQL way, can we just rely on some conventional PK name in SQL Server (like `DROP CONSTRAINT PK_$table;` from [here](https://docs.microsoft.com/en-us/sql/relational-databases/tables/delete-primary-keys#TsqlProcedure))?

morozov: Hold on. `getDropPrimaryKeySQL()` is not even part of the abstract platform API. Why are we using it in the abstract platform code? From what I see, most of the platforms are capable of identifying PK by name including SQL Server. Why should we take special care about the PK on SQL Server if it can be just dropped by name same as other constraints?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3605,4 +3611,9 @@ protected function getLikeWildcardCharacters() : string
     {
         return '%_';
     }
+
+    protected function getDropPrimaryKeySQL(string $table) : string
+    {
+        return 'ALTER TABLE ' . $table . ' DROP PRIMARY KEY';
```
morozov: I see, `getDropPrimaryKeySQL()` is being introduced here for an abstract platform but it's only needed for MySQL and has MySQL-specific syntax. This is a no-go.

I believe, the original issue is filed for MySQL where `ALTER TABLE DROP INDEX PRIMARY` is not supported. Instead of introducing this method, `MySqlPlatform::getDropIndexSQL()` should be updated to produce correct SQL for the PK.

SCIF: The issue originally discovered during work with Postgres. Initially I tried to modify `getDropIndexSQL()` method but it can't be done for postgres at least, so separate method for primary keys only is the best solution.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
SCIF: We can't  rely on any name conventions/best practices. As far as I investigated, the name builds according to schema `PK_{$tableName}_{$someHash}` and can't be predicted.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
SCIF: PK is not just an index, but it's a constraint in some RDBMS. Handling could be done by changing `getDropIndexSQL()` signature but I believe it's not the best option.

SCIF:
Travis fails on getting https://dev.mysql.com/get/mysql-apt-config_0.8.9-1_all.deb which can be downloaded for me.

Majkl578:
Travis issue reported in https://github.com/travis-ci/travis-ci/issues/9421.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: > so separate method for primary keys only is the best solution

It's fine to have it for Postgres and MySQL which seem unable to drop the PK by name but it shouldn't be part of the abstract platform API. Those platforms which can drop PK by its name shouldn't have any changes.

SCIF: I really don't understand what you propose. Do you mean define `getDropPrimaryKeySQL` as:
```
    protected function getDropPrimaryKeySQL(string $table) : string
    {
        $this->getDropIndexSQL('primary', $table);
    }
```
?
I can't get rid of `getDropPrimaryKeySQL` in favour  `getDropIndexSQL()` because `$index` defined as `Index|string`. So, BC break required, if you would like to use only `getDropIndexSQL()`.

morozov: > I really don't understand what you propose. Do you mean define `getDropPrimaryKeySQL` as:
> ```php
> protected function getDropPrimaryKeySQL(string $table) : string
> {
>     $this->getDropIndexSQL('primary', $table);
> }
> ```

No. I propose to get rid of this method entirely. The fact that it's introduced in `AbstractPlatform` with a MySQL-specific implementation and is called from `getPreAlterTableIndexForeignKeySQL()` is itself a BC break. If a platform like SQL Server doesn't support the removal of a PK, it will have to implement a workaround like you did, otherwise it will not work anymore.

Instead, for the platforms which cannot drop a PK by name, I propose an override like this:
```php
function getDropIndexSQL($index)
{
    if ($index->isPrimary()) {
       return $theProperSQLToDropIt;
    }

    return parent::getDropIndexSQL($index);
}

Comment (SCIF) [reply lib/Doctrine/DBAL/Platforms/AbstractPlatform.php]: Hi @morozov ,

I still don't understand why introduction of new `protected` method is worse than changing signature of existing `public` method. Overriding like you proposed, was the very first way I used to fix this issue.

Any other contributors opinions?

Comment (SCIF) [reply lib/Doctrine/DBAL/Platforms/AbstractPlatform.php]: Regarding platform-specific code - I can declare `getDropPrimaryKeySQL()` as `abstract` and implement it in each inherited class.

Comment (morozov) [reply lib/Doctrine/DBAL/Platforms/AbstractPlatform.php]: @SCIF why does the signature of the public method should be changed? If you're referring to my example, it's just pseudo-code.

Declaring the method as abstract is still the same BC break with the same problem: all platforms (including the ones maintained outside of this repo) will have to implement it sometimes at the cost of a workaround.

Instead of dragging all platforms into fixing a platform-specific issue (MySQL and PosrgreSQL, as far as I understand), it should be fixed in those platforms w/o introducing a new API.

Comment (SCIF) [reply lib/Doctrine/DBAL/Platforms/AbstractPlatform.php]: @morozov , here is [short explanation](https://github.com/doctrine/dbal/pull/2929#discussion_r178455710). To be able to introduce platform specific workaround I need to check if index is primary in `getDropIndexSQL()` but method signature allows passing index as a string (just name of index). In several cases, name of index supplied to `getDropIndexSQL()` is not precise i.e. supplied `primary` rather than actual name.

Comment (SCIF) [reply lib/Doctrine/DBAL/Platforms/AbstractPlatform.php]: If my current change will be marked as breaking BC, then I would prefer to change signature of `getDropIndexSQL()` to enforce supplying `Index` only, no string. It will bring much more beauty diff`s. Call to BC guru @Ocramius.

Comment (SCIF) [reply lib/Doctrine/DBAL/Platforms/AbstractPlatform.php]: I've estimated complexity of changing signature of `getDropIndexSQL()`  and it will cause changing signature of `dropIndex()` as well, which is public API.

Comment (morozov) [reply lib/Doctrine/DBAL/Platforms/AbstractPlatform.php]: It doesn't require any signature change in `getDropIndexSQL()`. For Postgres:
```php
public function getDropIndexSQL($index, $table = null)
{
    // after $index contains the index name
    if ($index === 'primary') {
        return $this->getDropConstraintSQL($table . '_pkey', $table);
    }

    return 'DROP INDEX ' . $index;
}
```
Something similar for MySQL.

morozov: Then, in the case when `$index` is a string, you get its object representation and see if it's primary or not.

morozov: You also can fix it only for your use case when the `$index` comes as an object from the schema manager. If there's another case when dropping the PK by name is required, it can be solved separately.

SCIF: @morozov 

> Then, in the case when $index is a string, you get its object representation and see if it's primary or not.

How I can do that? I don't know a way to do it. If you mean to find a workaround for particular `getAlterTableSQL()` usage — yes, I can fix it in another way, but it's bloody stupid idea to find a big issue and solve only one of use-cases rather than fix it in principal. I've spent a lot of time on supporting this PR and would like to solve a root of evil, rather than just superficial solution of the problem.

morozov: > How I can do that? I don't know a way to do it.

Neither do I. That's why I'm saying it's fine if you solve only the case where the passed index is an object.

> I've spent a lot of time on supporting this PR and would like to solve a root of evil, rather than just superficial solution of the problem.

The root can could be only solved by reworking the entire index-related API (I recall there was a discussion or at least mention of that /cc @deeky666). This API would account for all differences between indices, constraints, primary keys, foreign keys, etc. in different platforms and would provide some unified higher level API.

It's impossible to solve it correctly in terms of the current API without introducing more controversial methods (I'm talking about `getDropPrimaryKeySQL`()). You can either solve your case in 2.x or postpone it until 3.x and solve properly.

SCIF: Right now, `Index` class contain primary so, really not necessary to be revised because reflect it's purpose pretty fine. The problem is that index can be supplied by name, which would be my proposal to drop in 3.x.
@morozov , thanks for your input. I will refactor this PR just to fix comparing issue soon.

PR #2933, opened 2017-12-05
Added test case for resource as parameter

holantomas:
Added test case for PHP Resource as parameter in Driver exception.

See [this](https://github.com/doctrine/doctrine2/issues/6864)

Can be fixed by something like this.

`Doctrine\DBAL\DBALException::formatParameters()`
```php
private static function formatParameters(array $params)
    {
        return '[' . implode(', ', array_map(function ($param) {
            if (is_resource($param)) {
                return (string) $param;
            }

            $json = @json_encode($param);

            if (! is_string($json) || $json == 'null' && is_string($param)) {
                // JSON encoding failed, this is not a UTF-8 string.
                return '"\x' . implode('\x', str_split(bin2hex($param), 2)) . '"';
            }

            return $json;
        }, $params)) . ']';
    }
```

### DIFF
--- a/tests/Doctrine/Tests/DBAL/DBALExceptionTest.php
+++ b/tests/Doctrine/Tests/DBAL/DBALExceptionTest.php
@@ -22,7 +22,7 @@ class DBALExceptionTest extends DbalTestCase
     {
         /* @var $driver Driver */
         $driver = $this->createMock(Driver::class);
-        $e = \Doctrine\DBAL\DBALException::driverExceptionDuringQuery($driver, new \Exception, "INSERT INTO file (`content`) VALUES (?)", [1 => tmpfile()]);
+        $e = \Doctrine\DBAL\DBALException::driverExceptionDuringQuery($driver, new \Exception, "INSERT INTO file (`content`) VALUES (?)", [1 => fopen(__FILE__, 'r')]);
         self::assertContains('Resource', $e->getMessage());
     }
 
--- a/tests/Doctrine/Tests/DBAL/DBALExceptionTest.php
+++ b/tests/Doctrine/Tests/DBAL/DBALExceptionTest.php
@@ -17,6 +17,14 @@ class DBALExceptionTest extends DbalTestCase
         $e = DBALException::driverExceptionDuringQuery($driver, new \Exception, '', array('ABC', chr(128)));
         self::assertContains('with params ["ABC", "\x80"]', $e->getMessage());
     }
+    
+    public function testDriverExceptionDuringQueryAcceptsResource()
+    {
+        /* @var $driver Driver */
+        $driver = $this->createMock(Driver::class);
+        $e = \Doctrine\DBAL\DBALException::driverExceptionDuringQuery($driver, new \Exception, "INSERT INTO file (`content`) VALUES (?)", [1 => tmpfile()]);
+        self::assertContains('Resource', $e->getMessage());
+    }
 
     public function testAvoidOverWrappingOnDriverException()
     {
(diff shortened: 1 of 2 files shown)


holantomas:
Any better idea how to fix `Doctrine\DBAL\DBALException::formatParameters()`?
Actually this PR is proof-of-concept which emulate error by tests. I can add fix shown above to pass test a fix this error.

morozov:
> Any better idea how to fix `Doctrine\DBAL\DBALException::formatParameters()`?

@holantomas what exectly are you concerned about? Given the existing implementation, the proposed approach looks good to me. Please add it to the pull request.

As for the method itself, it's really weird that we use JSON encoding for PHP-originated values and eventually end up with two serialization formats chosen at runtime based on the actual value. Something like `var_export()` might fit much better but this is outside of the scope of this issue.

morozov (approved):

tests/Doctrine/Tests/DBAL/DBALExceptionTest.php
```diff
@@ -17,6 +17,14 @@ public function testDriverExceptionDuringQueryAcceptsBinaryData()
         $e = DBALException::driverExceptionDuringQuery($driver, new \Exception, '', array('ABC', chr(128)));
         self::assertContains('with params ["ABC", "\x80"]', $e->getMessage());
     }
+    
+    public function testDriverExceptionDuringQueryAcceptsResource()
+    {
+        /* @var $driver Driver */
+        $driver = $this->createMock(Driver::class);
+        $e = \Doctrine\DBAL\DBALException::driverExceptionDuringQuery($driver, new \Exception, "INSERT INTO file (`content`) VALUES (?)", [1 => tmpfile()]);
```
Majkl578: I'd rather use something that doesn't pollute your environment each run, i.e. `fopen(__FILE__, 'r')`.

Majkl578:
There should be another test case for _closed resource_, especially because its behavior is changing between PHP versions.

holantomas:
@Majkl578 Yes, that's true but I tried some tests and research for handling closed resources. I cannot solve this. [This](https://3v4l.org/JsqtO) is behavioral of closed resources between PHP/HHVM versions.
As you can see `is_resource` returning always `FALSE` for closed resource. Only what can be used is RegExp and search for `"Resource*"` but I don't like this solution and you need to do it only for resources not for standard string/numeric params so we are back in start with this. Any better idea?

**Changed** `tmpfile()` to `fopen(__FILE__, 'r')`.

tests/Doctrine/Tests/DBAL/DBALExceptionTest.php
holantomas: Changed

Majkl578:
We don't support HHVM, only PHP.

holantomas:
@Majkl578 I removed HHVM from my last post. Can we now talk about related things?

Majkl578:
Resources in PHP are mess and pretty unstable land, there have been multiple changes in past (as seen on 3v4l) and there will probably be again in future (converting resources to objects). But for now this PR looks good.

Majkl578 (approved):

holantomas:
@Majkl578 OT: What you mean "converting resources into objects". Is there some PR or FR on PHP to drop resources as we know them to classic PHP Class/Objects? Can you provide more info/links?

Majkl578:
@holantomas There were multiple talks about removing `resource` type and replacing it by normal objects, discussed multiple times in php.internals mailing list. Nothing certain and no RFC though.

holantomas:
@Majkl578 OT:  Ok, I thought I missed some RFC thx. End OT

Majkl578:
@holantomas Thanks!

PR #2934, opened 2017-12-06
Use Null Coalesce Operator

carusogabriel:


lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -121,7 +121,7 @@ public function __construct(array $params, Driver $driver, Configuration $config
             $params['slaves'][$slaveKey]['driver'] = $params['driver'];
         }
 
-        $this->keepSlave = isset($params['keepSlave']) ? (bool) $params['keepSlave'] : false;
+        $this->keepSlave = (bool) $params['keepSlave'] ?? false;
```
morozov: This doesn't look right. The `(bool)` casting seems to has higher [precedence](http://php.net/manual/en/language.operators.precedence.php) than `??`:
```php

$params = [];
$keepSlave = (bool) $params['keepSlave'] ?? false;

// Notice: Undefined index: keepSlave
```

carusogabriel: So `isset` should be keep it here?

stloyd:
```
$keepSlave = (bool) ($params['keepSlave'] ?? false);

Majkl578 (approved):

Ocramius:
🚢  thanks @carusogabriel!

PR #2935, opened 2017-12-08
Improve deprecation wording in `Doctrine\DBAL\Schema\Column#setOptions()`

greg0ire:
People that are starting a new project will immediately know what this
is about, but people upgrading doctrine might get a lot of these and
should have a better clue regarding what this is about.


Ocramius (approved):
LGTM 👍

Issue #2627, opened 2017-02-01
Change QueryBuilder numerical parameters to start from 1


jnvsor:
According to the [querybuilder docs](http://doctrine-orm.readthedocs.io/projects/doctrine-dbal/en/latest/reference/query-builder.html#security-safely-preventing-sql-injection):

> Due to an API design error the numerical parameters in the QueryBuilder API start with the needle 0, not with 1 as in the PDO API. This is very unfortunate, but we have found no BC way to fix this.

The 3.0 milestone seems like a nice place to fix this - could you add this to the 3.0 milestone?

Ocramius:
The BC break is still excessive and hard to spot

jnvsor:
Couldn't you add a property to the querybuilder specifying whether keys should be decremented or not before passing to `executeQuery`/`executeUpdate` and set it to legacy (along with an `E_USER_DEPRECATED`) whenever `setParameter(0, ...)` is called? (Since new code will be 1-indexed)

Ocramius:
No, it's just not worth doing that. Additionally, I'd suggest that following PDO's API is a bad idea in most scenarios, given its exotic/weird design decisions (and also: nobody really wants to fix those in php core, since the extension is quite messy)

jnvsor:
Ok, thanks for the response!

PR #2940, opened 2017-12-15
Clean up note on query builder setParameter documentation

jnvsor:
With regard to the discussion in #2627 it might be better to clean up this note, since it's a bit misleading.


Ocramius (approved):
LGTM 👍

PR #2942, opened 2017-12-19
Combine consecutives unsets

carusogabriel:
`unset` allow us to pass multiple values to check.

[PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) helped me on that.

### DIFF
--- a/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
+++ b/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
@@ -288,8 +288,7 @@ class MasterSlaveConnection extends Connection
      */
     public function close()
     {
-        unset($this->connections['master']);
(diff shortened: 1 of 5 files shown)


Majkl578:
I have mixed feelings about this one, it makes it worse to read... :/

carusogabriel:
I did it to reduce lines, but if readability is more important, why not leave it as it is? :smile:

Majkl578:
I personally think it is, but let @Ocramius decide. :)

Ocramius:
@Majkl578 the readability is quite OK, I just think we aren't used to variadic isset/unset. I myself learned about it just 2 years ago, and I do PHP since aaaaaaages (sadly).

The change is semantically more correct, as this is an atomic operation, so a grouped unset is better than a sequence of unset calls anyway.

PR #2943, opened 2017-12-19
Remove empty verification after isset

carusogabriel:
As `isset` also verifies if the given value isn't null, and by the [`PHP type comparison tables`](http://php.net/manual/en/types.comparisons.php#types.comparisons), couldn't we remove some `empty` verifications?

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
@@ -136,7 +136,7 @@ abstract class AbstractOracleDriver implements Driver, ExceptionConverterDriver
                 $service = 'SERVICE_NAME=' . $serviceName;
             }
 
-            if (isset($params['instancename']) && ! empty($params['instancename'])) {
+            if (isset($params['instancename'])) {
                 $instance = '(INSTANCE_NAME = ' . $params['instancename'] . ')';
             }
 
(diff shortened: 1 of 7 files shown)


Ocramius (changes requested):
Requires test cases verifying empty strings against the introduced changes. So far, patch doesn't seem valid to me.

carusogabriel:
@Ocramius You're right, empty string works differently in `is_null`, `empty` and `=== null`, sorry :sweat_smile: 
Gonna test if it passes.

lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
```diff
@@ -136,7 +136,7 @@ protected function getEasyConnectString(array $params)
                 $service = 'SERVICE_NAME=' . $serviceName;
             }
 
-            if (isset($params['instancename']) && ! empty($params['instancename'])) {
```
Majkl578: You can remove this isset entirely.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Driver.php
```diff
@@ -56,7 +56,7 @@ private function _constructPdoDsn(array $params)
             $dsn .= $params['host'];
         }
 
-        if (isset($params['port']) && !empty($params['port'])) {
```
Majkl578: Same here.

lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
carusogabriel: I did some tests and this `empty`s are valid. Gonna close it :smile:

Majkl578: Yes, but still, these issets are redundant. :)

carusogabriel: Gonna study more and PR this separately :man_technologist:

Issue #1811, opened 2013-09-09
DBAL-599: Add support of insert and insert select statements in DBAL QueryBuilder


doctrinebot:
Jira issue originally created by user koc:

PR #2948, opened 2017-12-21
Feature/add insert select statement in querybuilder

ghost:
related to issue #1811

This PR qill enable the posibility to create INSERT INTO SELECT Statements
examples:
1.
```php
$qb = $conn->createQueryBuilder()`
`->insert('users')`
`->select('u2.name', 'u2.password')`
`->from('users', 'u2')`
`->where('u2.id = ?')
```

2.
```php
$qb = $conn->createQueryBuilder()`
`->insert('users')`
`->values( array( 'name' => null, 'password' => null ) )`
`->select('u2.name', 'u2.password')`
`->from('users', 'u2')`
`->where('u2.id = ?')
```

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -46,7 +46,6 @@ class QueryBuilder
     const DELETE = 1;
     const UPDATE = 2;
     const INSERT = 3;
-    const REPLACE = 3;
 
     /*
      * The builder states.
@@ -583,7 +582,17 @@ class QueryBuilder
      *         );
      * </code>
      *
-     * alternative:
+     * alternative with select
+     *
+     * <code>
+     *     $qb = $conn->createQueryBuilder()
+     *         ->insert('users')
+     *         ->select('u2.name', 'u2.password')
+     *         ->from('users', 'u2')
+     *         ->where('u2.id = ?')
+     * </code>
+     *
+     * with filed definition:
      *
      * <code>
      *     $qb = $conn->createQueryBuilder()
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -592,7 +592,7 @@ class QueryBuilder
      *         ->where('u2.id = ?')
      * </code>
      *
-     * with filed definition:
+     * with field definition:
      *
      * <code>
      *     $qb = $conn->createQueryBuilder()
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -46,6 +46,7 @@ class QueryBuilder
     const DELETE = 1;
     const UPDATE = 2;
     const INSERT = 3;
+    const REPLACE = 3;
 
     /*
      * The builder states.
@@ -64,6 +65,7 @@ class QueryBuilder
      * @var array The array of SQL parts collected.
      */
     private $sqlParts = [
+        'into'    => [],
         'select'  => [],
         'from'    => [],
         'join'    => [],
@@ -220,6 +222,7 @@ class QueryBuilder
      * </code>
      *
      * @return string The SQL query string.
+     * @throws QueryException
      */
     public function getSQL()
     {
@@ -464,7 +467,8 @@ class QueryBuilder
      */
     public function select($select = null)
     {
-        $this->type = self::SELECT;
+        if($this->type !== self::INSERT )
+            $this->type = self::SELECT;
 
         if (empty($select)) {
             return $this;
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
```diff
@@ -476,6 +476,38 @@ public function testInsertValues()
...
+
+        self::assertEquals(QueryBuilder::INSERT, $qb->getType());
+        self::assertEquals('INSERT INTO users (SELECT u2.name, u2.password FROM users u2 WHERE u2.id = ?)', (string) $qb);
+    }
+
+    public function testInsertValuesSelect()
+    {
+        $qb = new QueryBuilder($this->conn);
+        $qb->insert('users')
+            ->values(
```
morozov: This syntax is a bit weird. As far as I understand the resulting SQL, this method defines the columns which the selected columns will be mapped to. Not values.

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
```diff
@@ -476,6 +476,38 @@ public function testInsertValues()
...
+                array(
+                    'name' => null,
+                    'password' => null
```
morozov: This is also weird. Why should there be nulls on the right-hand side?

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1180,12 +1210,22 @@ private function isLimitQuery()
...
-        ' VALUES(' . implode(', ', $this->sqlParts['values']) . ')';
+        $statement = 'INSERT INTO ' . $this->sqlParts['into']['table'];
+        if( !empty($this->sqlParts['values']) && count( array_keys($this->sqlParts['values']) ) > 0 ) {
+            $statement .= ' (' . implode(', ', array_keys($this->sqlParts['values'])) . ')';
+        }
+
+        if( !empty($this->sqlParts['select'] ) ) {
+            $statement .= ' (' . $this->getSQLForSelect() . ')';
+        }
+        else {
```
morozov: What if both `'select'` and `'values'` parts are defined?

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1180,12 +1210,22 @@ private function isLimitQuery()
...
      * @return string
+     * @throws QueryException
      */
     private function getSQLForInsert()
     {
-        return 'INSERT INTO ' . $this->sqlParts['from']['table'] .
-        ' (' . implode(', ', array_keys($this->sqlParts['values'])) . ')' .
-        ' VALUES(' . implode(', ', $this->sqlParts['values']) . ')';
+        $statement = 'INSERT INTO ' . $this->sqlParts['into']['table'];
+        if( !empty($this->sqlParts['values']) && count( array_keys($this->sqlParts['values']) ) > 0 ) {
```
morozov: Isn't it the same as `if (count($this->sqlParts['values']) > 0)`?

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1180,12 +1210,22 @@ private function isLimitQuery()
...
+            $statement .= ' (' . implode(', ', array_keys($this->sqlParts['values'])) . ')';
+        }
+
+        if( !empty($this->sqlParts['select'] ) ) {
+            $statement .= ' (' . $this->getSQLForSelect() . ')';
```
morozov: The parentheses around `SELECT` are not needed. At least in most cases.

viosys: I know, but if it's a slightly more complex statement then I'm on the safe side

lib/Doctrine/DBAL/Query/QueryBuilder.php
viosys: then a 'normal' inster statement is generated. You might even throw an exception if both are specified

lib/Doctrine/DBAL/Query/QueryBuilder.php
viosys: that's right

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
viosys: to only add the columnnames to the insert statement

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
viosys: that's right

lib/Doctrine/DBAL/Query/QueryBuilder.php
Ocramius: Throwing is a better idea here

lib/Doctrine/DBAL/Query/QueryBuilder.php
Ocramius: > slightly more complex statement

If you have such a scenario, please add a test for it

lib/Doctrine/DBAL/Query/QueryBuilder.php
Ocramius: the `! empty()` is sufficient

PR #2853, opened 2017-09-11
Remove copyright headers

Majkl578:
LICENSE file is sufficient.

Targets develop.

(Develop is not up-to-date with master so Travis will fail...)

PR #2949, opened 2017-12-21
Removing license header

mikeSimonson:


Majkl578:
We've already got rid of these here in develop: #2853.

PR #2951, opened 2017-12-26
Allow array of DateTime as param in executeQuery

egircys:
Currently it's not possible to use an array of `\DateTime` objects as parameter, as reported in https://github.com/doctrine/doctrine2/issues/6934

I'm proposing a fix that changes the `PDO::PARAM_STR` to the DBAL `Type::DATETIME` when binding the parameters.

This allows to run the queries with `\DateTime` parameters and the additional types `Connection::PARAM_STR_ARRAY` and `\PDO::PARAM_STR`.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1581,7 +1581,11 @@ class Connection implements DriverConnection
     {
         if (is_string($type)) {
             $type = Type::getType($type);
+        } elseif (PDO::PARAM_STR === $type && $value instanceof \DateTime) {
+            // change the given PDO type to DBAL in case it's a \DateTime object
+            $type = Type::getType(Type::DATETIME);
         }
+
         if ($type instanceof Type) {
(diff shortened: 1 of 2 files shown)


Majkl578:
How do you decided what `DateTime` instance means? You can't safely assume it's `Type::DATETIME`, because it can be `Type::DATETIMETZ` and possibly any custom type. Also doesn't cover immutable variant (which should be preferred imho).

egircys:
True, it could be `Type:DATETIMETZ` or any other. But is there any way to tell what type it should be when the `\PDO::PARAM_STR` is given? I've adopted the same logic now as it's in https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Query/ParameterTypeInferer.php#L57

Ocramius:
The general idea is to have multi-dimensional parameter binding. That
requires complex type definition DSL that would need to replace the current
DBAL types.

egircys:
Closing because it would require bigger refactoring.

PR #2955, opened 2017-12-28
[DBAL-2954] Removed error suppression in IBM DB2 Statement

### LINKED ISSUE #2954: The IBM DB2 Statement driver suppresses errors triggered by db2_execute
`db2_execute()` can trigger different kinds of errors internally:

1. A notice:
   ```php
   $stmt = $conn->executeQuery('SELECT * FROM users WHERE user_name = ?', [[]]);
   ```
   Expected result:
   ```
   Notice: Array to string conversion in lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php on line 175
   ```
   Actual result: nothing. The error is suppressed.
2. A catchable fatal error:
   ```php
   $stmt = $conn->executeQuery('SELECT * FROM users WHERE user_name = ?', [new stdClass()]);
   ```
   Expected result:
   ```
   Catchable fatal error: Object of class stdClass could not be converted to string in lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php on line 175
   ```
   Actual result: nothing. The script silently exists.

Neither of the errors above seems to be needed/possible to handle on the DBAL level. The first one doesn't make `db2_execute()` return `FALSE`, the other cannot be caught using a `try`/`catch` block.

morozov:
Fixes #2954.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
+++ b/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
@@ -172,7 +172,7 @@ class DB2Statement implements \IteratorAggregate, Statement
             }
         }
 
-        $retval = @db2_execute($this->_stmt, $params);
+        $retval = db2_execute($this->_stmt, $params);
 
         if ($retval === false) {
             throw new DB2Exception(db2_stmt_errormsg());
--- a/tests/Doctrine/Tests/DBAL/Functional/Driver/IBMDB2/DB2StatementTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Driver/IBMDB2/DB2StatementTest.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL\Functional\Driver\IBMDB2;
+
+use Doctrine\DBAL\Driver\IBMDB2\DB2Driver;
+use Doctrine\Tests\DbalFunctionalTestCase;
+use PHPUnit\Framework\Error\Notice;
+
+class DB2StatementTest extends DbalFunctionalTestCase
+{
+    protected function setUp()
+    {
+        if ( ! extension_loaded('ibm_db2')) {
+            $this->markTestSkipped('ibm_db2 is not installed.');
+        }
+
+        parent::setUp();
+
+        if ( ! $this->_conn->getDriver() instanceof DB2Driver) {
+            $this->markTestSkipped('ibm_db2 only test.');
+        }
+    }
+
+    public function testExecutionErrorsAreNotSuppressed()
+    {
+        $stmt = $this->_conn->prepare('SELECT * FROM SYSIBM.SYSDUMMY1 WHERE \'foo\' = ?');
+
+        // unwrap the statement to prevent the wrapper from handling the PHPUnit-originated exception
+        $wrappedStmt = $stmt->getWrappedStatement();
+
+        $this->expectException(Notice::class);
+        $wrappedStmt->execute([[]]);
+    }
+}


Majkl578:
Looking into the [source](https://git.php.net/?p=pecl/database/ibm_db2.git;a=blob;f=ibm_db2.c;h=35839a72814adb1fb8017e5182f55ecc3fb92c2c;hb=HEAD], it looks like there is more different errors, not just conversion related but also binding-related errors and execution failure errors. Aren't these the reason for `@`? The next check for `false` suggests it's possibly the reason, to avoid warnings and to properly escalate them as exceptions.

Majkl578:
One solution could be: keep `@`, keep check for `$retval === false` and add another check using `error_get_last()` (cleared with `error_clear_last()` before `db2_execute()`) that escalates to exception as well.

morozov:
> Aren't these the reason for `@`? The next check for `false` suggests it's possibly the reason, to avoid warnings and to properly escalate them as exceptions.

Most probably that's the case.

> One solution could be: keep `@`, keep check for `$retval === false` and add another check using `error_get_last()` (cleared with `error_clear_last()` before `db2_execute()`) that escalates to exception as well. Is it worth it?

IMO, not all errors should be promoted to exceptions. If an error didn't cause the method to return `false`, it shouldn't be converted (e.g. it could be a deprecation notice). Ideally, we should suppress those error which we convert to exceptions but let the others trigger.

In a general case, a single call to `db_execute()` may produce multiple non-fatal and one or more fatal errors at a time. Not sure how it could be implemented.

morozov:
> One solution could be: keep `@` […]

With this solution, an internal fatal error will cause silent script termination. Happy debugging.

Majkl578:
> Happy debugging.

Right, because _catchable fatal_ isn't actually catchable by `try...catch`... :/ (Was renamed to _recoverable fatal_ in 7.1+: https://3v4l.org/Wea6g).
But it is catchable by custom error handler: https://3v4l.org/vJ0Ea without causing script termination.

morozov:
@Majkl578 so you're suggesting something like this?
```php
set_error_handler(function ($errno, $errstr, $errfile, $errline, $errcontext) {
    throw new DB2Exception($errstr);
});

try {
    $retval = db2_execute($this->_stmt, $params);
} finally {
    restore_error_handler();
}
```

It is better than error suppression but it may introduce false positives in the cases which did trigger errors but didn't cause the `false` result. Are we okay with that?

I didn't explore the extension source code but something makes me think that all possible notices are not run-time but logical (revealing bugs in the userland code). Having them also displayed as native PHP notices doesn't look like a big deal to me.

asgrim:
I think this is less a discussion about DB2 itself, and really it's "how should errors from the driver be propagated"... PHP 7 now having `\Throwable` that all errors implement suggests a shift towards preferring exceptions anyway. That is also my personal preference, but it is really up to the Gods of DBAL to decide.

Though, playing devil's advocate, thinking back to when we first started implementing the DB2-based project, I remember the `@` operator being a bit of a pain trying to get things working. I simply removed it whilst initially figuring things out and then re-pulled the deps with Composer when done.

That said, especially if DBAL still supports PHP 5.x, I think @Majkl578's solution of setting an error handler would work universally and nicely (convert everything to exception: if it's a notice that the consumer doesn't care about, that can be handled that in the consumer code).

Ocramius:
@asgrim this is for PHP 7.1+ only though. Unless security issues pop up in `2.5.x`, nothing will be touched for PHP 5 support ;-)

asgrim:
In that case, would it be so bad to remove the `@` operator here, as (almost) everything would be `catch`-able anyway, right?

Majkl578:
> In that case, would it be so bad to remove the @ operator here, as (almost) everything would be catch-able anyway, right?

Not quite, it can still throw a fatal that is not catchable by `try...catch`, that's why I came up with set_error_handler workaround:
```
$stmt = $conn->executeQuery('SELECT * FROM users WHERE user_name = ?', [new stdClass()]);
```

> It is better than error suppression but it may introduce false positives in the cases which did trigger errors but didn't cause the false result. Are we okay with that?

In this case, we'd have to store errors in set_error_handler and throw them conditionally. I'm not a fan of either of these solutions, but I guess it's about what's less evil in the end.
Assuming that it can throw multiple notices/warnings/errors, something like this would be needed:
```php
<?php

$errors = [];
set_error_handler(function ($errno, $errstr, $errfile, $errline, $errcontext) use (&$errors) {
    $errors[] = $errstr;
});

try {
    $retval = db2_execute($this->_stmt, $params);

    if ($retval === false && count($errors) !== 0) {
        throw DB2Exception::fromMultiple($errors);
    }
} finally {
    restore_error_handler();
}
```
Don't like this solution either. :/

morozov:
> Don't like this solution either. :/

And it doesn't solve the original problem. The notice in the example will still be suppressed since it doesn't cause `db2_exec()` to return `false`.

A complete solution would contain the `else` block which would re-trigger all collected errors without throwing the exception. And it would have to convert all collected `E_*` types to `E_USER_*` types. That's… kinda… insane.

Ocramius:
@Netmosfera job for you?

Netmosfera:
@morozov 

> A complete solution would contain the else block which would re-trigger all collected errors without throwing the exception. And it would have to convert all collected E_* types to E_USER_* types. That's… kinda… insane.

if you are not throwing the exception, you are collecting the errors for... what?

Netmosfera:
```php
retrigger_error(string $message, int $type){
    $conversions = [
        E_ERROR      => E_USER_ERROR,
        E_WARNING    => E_USER_WARNING,
        E_NOTICE     => E_USER_NOTICE,
        E_DEPRECATED => E_USER_DEPRECATED,
    ];
    
    foreach($conversions as $from => $to){
        if($type & $from){
            $type = $type & ~$from;
            $type = $type | $to;
        }
    }
    
    trigger_error($message, $type);
}
```

morozov:
@Majkl578 @Netmosfera with the approach of a custom error handler, we will tremendously increase the code complexity in order to overcome the bad design of the `ibm_db2` extension regarding handling errors the userland code.

Are you sure you want to proceed this route?

I still believe just remove the suppression is fine. The worst thing which can happen is in _some_ (which exactly?) cases, the same error may be reported as error and exception. Unless it's a runtime error, I believe, it's perfectly fine.

Netmosfera:
Personally I don't, I would just remove the error suppression, let the notices be, and just check for the return value.

Majkl578:
There is one more issue with _just removing_ the suppression: BC. Because if the function suddenly starts throwing notices/warnings, strict/debug tools will immediately catch them and convert into exception. And since original code depended on DB2Exception but now receives i.e. ErrorException, that could break code.
So at least, it should be in `try...catch` + convert it into DB2Exception + set previous exception to the caught one?

Netmosfera:
why are you converting notices to ErrorException?

Netmosfera:
For correctness' sake and in case you decide to go with that route, I've just been informed that `E_*` are not meant to be triggered combined with each other. So `E_NOTICE | E_STRICT` can never happen.

Majkl578:
> why are you converting notices to ErrorException?

Because even notice means bug. See for example https://github.com/nette/tracy#visualization-of-errors-and-exceptions.

Netmosfera:
programming bugs though

morozov:
> Because if the function suddenly starts throwing notices/warnings, strict/debug tools will immediately catch them and convert into exception.

This is fine because they will indicate bugs in the userland code. Otherwise, we need an example of an error which doesn't produce `false` and is not caused by a bug in the userland code.

> And since original code depended on DB2Exception but now receives i.e. ErrorException, that could break code.

It will receive `ErrorException` from the other error handler, not from DBAL.

> So at least, it should be in `try...catch` + convert it into DB2Exception + set previous exception to the caught one?

You mean we'll be catching `Throwable` from the method which is not supposed to trigger any exceptions just to support custom error handlers? IMO, the code which uses such handlers should be ready to get `ErrorException` from anywhere.

Netmosfera:
I'm lost again :-P

morozov:
So, is there a strong disagreement with the current implementation? I still believe that a simple solution which sometimes will report the same error twice is better than a complex one or the one which will suppress (essentially, lose) some errors.

Majkl578 (approved):

morozov:
@Ocramius do you still have any concerns/reservations?

Ocramius:
Note: this only targets `2.7`, as some users may still be affected due to custom error handlers.

Issue #2953, opened 2017-12-27
Unlike other drivers, DBAL has a hard dependency on PDO


morozov:
Historically, the DBAL public API uses `PDO` constants which creates a hard dependency and makes the abstraction not that abstract. For the end user, it means that even if none of the `PDO` drivers are going to be used, the production environment still has to have `ext-pdo` installed.

To eliminate the dependency, the `PDO` constants in the method signatures should be replaced by the DBAL's own ones:

1. `PDO::PARAM_*` → `Doctrine\DBAL\Driver\Statement::PARAM_*`,
2. `PDO::FETCH_*` → `Doctrine\DBAL\Driver\ResultStatement::FETCH_*`,
3. `PDO::CASE_*` → `Doctrine\DBAL\Driver\ResultStatement::CASE_*`.

It can be done in phases:

1. In `2.x`, introduce the new DBAL constants and assign them corresponding PDO values, e.g. `Statement::PARAM_INT = PDO::PARAM_INT`.
2. Deprecate the usage of `PDO::` constants when calling the DBAL methods.
3. In `3.0`, replace the `PDO` constants with their integer values, e.g. `Statement::PARAM_INT = 1`
4. Remove the dependency.

morozov:
So far, I've stumbled upon unclear issues:

1. [`ResultStatement::fetch()`](https://github.com/doctrine/dbal/blob/fcc00deed202e92c62953eb0d833dd203c42d429/lib/Doctrine/DBAL/Driver/ResultStatement.php#L85) declares support of `PDO::FETCH_ORI_*` constants. In fact, they are only supported by `PDO`. Most likely, it's not used anywhere and therefore can be dropped.
2. [`Statement::bindParam()`](https://github.com/doctrine/dbal/blob/fcc00deed202e92c62953eb0d833dd203c42d429/lib/Doctrine/DBAL/Driver/Statement.php#L73) declares support of `PDO::PARAM_INPUT_OUTPUT`. Given it can only be used with stored procedures which the DBAL doesn't support, there's no way it could be used and therefore its support can be dropped.

Please let me know if you know the reason why the functionality defined by those constants should be still supported.

PR #2958, opened 2018-01-01
Remove hard dependency on PDO

morozov:
The first part of the pull request (prior to the commits labeled `[BC]`) contains the non-breaking changes which should be back-ported to `master`. The remaining part contains the removal of the dependency and breaking API changes.

Fixes: #2953.


lib/Doctrine/DBAL/Driver/Statement.php
```diff
@@ -19,6 +19,8 @@
 
 namespace Doctrine\DBAL\Driver;
 
+use PDO;
```
Ocramius: Probably to be removed

lib/Doctrine/DBAL/Portability/Connection.php
```diff
@@ -20,6 +20,7 @@
 namespace Doctrine\DBAL\Portability;
 
 use Doctrine\DBAL\Cache\QueryCacheProfile;
+use PDO;
```
Ocramius: To be removed, I'd say. Use the FQCN in the docblock

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -27,6 +29,29 @@
  */
 class PDOStatement extends \PDOStatement implements Statement
```
Ocramius: In this case we keep the dependency?

Ocramius (changes requested):
Overall, I understand why this is being attempted, but I don't think users will benefit much from it, as most of our users do in fact use PDO, so the rest shouldn't have a problem with keeping it installed.

What I see though, is using bitmasks to define bound parameter types, which is problematic when we consider things like multi-dimensional parameters. This doesn't affect low-level API, but it is one of the pain-points of our API (binding with `Connection::PARAM_INT_ARRAY`, for example).

I'm not sure if that's relevant to this patch though.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -75,14 +75,14 @@ class Connection implements DriverConnection
      *
      * @var integer
      */
-    const PARAM_INT_ARRAY = 101;
+    const PARAM_INT_ARRAY = Statement::PARAM_INT + self::ARRAY_PARAM_OFFSET;
```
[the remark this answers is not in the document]
Ocramius: Ah, it is an offset... didn't see that.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -75,14 +75,14 @@ class Connection implements DriverConnection
...
 
     /**
      * Represents an array of strings to be expanded by Doctrine SQL parsing.
      *
      * @var integer
      */
-    const PARAM_STR_ARRAY = 102;
+    const PARAM_STR_ARRAY = Statement::PARAM_STR + self::ARRAY_PARAM_OFFSET;
```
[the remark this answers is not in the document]
Ocramius: Ah, it is an offset... didn't see that.

lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
```diff
@@ -147,30 +146,38 @@ public function getIterator()
     /**
      * {@inheritdoc}
      */
-    public function fetch($fetchMode = null, $cursorOrientation = \PDO::FETCH_ORI_NEXT, $cursorOffset = 0)
+    public function fetch($fetchMode = null, $_ = null, $__ = null)
```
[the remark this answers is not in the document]
morozov: We should. This particular commit should only go to `develop`. Initially, I wanted to get rid of them entirely but here's the trick:

1. `DBAL\PDOStatement` extends `PDOStatement` and implements `ResultStatement`, both of which declare `fetch()`.
2. It looks impossible to use a custom PDO statement class which doesn't extend `PDOStatement` since `PDO` instantiates it without calling constructor (this is my current understanding, I may be wrong).
3. Therefore, the declaration of `ResultStatement::fetch()` and all its implementors should be compatible with `PDOStatement::fetch()`.

I'd love to be able to rework `DBAL\PDOStatement extends PDOStatement` into a composition and get rid of those parameters entirely but it's technically challenging. Any ideas?

lib/Doctrine/DBAL/Driver/PDOStatement.php
morozov: Why? This class will be only loaded if a PDO driver is used in runtime. Currently, all tests pass w/o the PDO extension loaded.

morozov:
> @Ocramius added this to the 2.7.0 milestone

This pull request is meant to be split in two parts:

1. All commits prior to the ones marked as `[BC]` go to master and the next minor release.
2. The rest goes to `develop`.

Or the other way around: everything goes to `develop`, then non-breaking changes get ported to `master`. The second approach should be better maintainable IMO.

morozov:
> oci runtime error: exec failed: container_linux.go:265: starting container process caused "could not create session key: disk quota exceeded"

Anyone, have you seen [these errors](https://travis-ci.org/doctrine/dbal/jobs/323725731) before? Looks irrelevant.

Majkl578:
Travis is messed up today, other build failed due to MySQL not starting...

morozov:
> […] but I don't think users will benefit much from it, as most of our users do in fact use PDO, so the rest shouldn't have a problem with keeping it installed.

@Ocramius the hard dependency hardens the adoption of the DBAL by existing projects. Particularly, SugarCRM is migrating its DB layer to DBAL. It supports several DB platforms but doesn't use or support PDO. Currently, it depends on `v2.5` and uses a PDO shim to avoid requiring the extension.

The absence of dependency on PDO makes the upgrade from a non-DBAL-based version to the DBAL-based seamless. Otherwise, the customers will be required to have PDO installed for no reason.

morozov:
> What I see though, is using bitmasks to define bound parameter types, which is problematic when we consider things like multi-dimensional parameters. This doesn't affect low-level API, but it is one of the pain-points of our API (binding with `Connection::PARAM_INT_ARRAY`, for example).

Could you elaborate? I was thinking about replacing `Connection::PARAM_INT_ARRAY` with `Statement::PARAM_INT | Statement::PARAM_ARRAY`. Otherwise, we have related constants in different classes. But this is outside of scope. We can do that in `3.0` before or after getting this pull request merged.

Ocramius:
@morozov ACK on adoption.

As for the `PARAM_INT_ARRAY` question, that was indeed feature creep: let's leave it out.

UPGRADE.md
```diff
@@ -1,3 +1,9 @@
+# Upgrade to NEXT
```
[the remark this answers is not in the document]
morozov: Let's open `develop`, get everything merged there and then back-port the relevant pieces to `master`.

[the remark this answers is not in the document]
Majkl578: Develop already exists here, it just needs rebase.

morozov:
@Ocramius please take a look. Instead of removing additional arguments from the `fetch()` and `fetchAll()` methods I replaced them with `...$args` (see https://github.com/doctrine/dbal/pull/2527#issuecomment-355159152). Therefore, there's no need to deprecate them. Callers will be able to keep passing them as they did but implementors will have to change their signatures accordingly.

Ocramius:
@morozov for `3.0`, wouldn't it be better for us to take ownership of the `fetch` and `fetchAll` signatures?

morozov:
We could. I just don’t have an idea of a better set of signatures. We still need to be able to take variable set of arguments depending on the mode. Do you have any concrete idea?

Ocramius:
Nope, no better ideas so far :-\

.travis.yml
```diff
@@ -306,6 +306,11 @@ jobs:
       if: type = pull_request
       php: 7.2
       script:
+        - |
```
Ocramius: Should be gone after a rebase, no?

morozov: I’ll remove it manually if it doesn’t. I tested the fix here. Gonna rebase develop and this branch tomorrow.

morozov:
Speaking of the ideas of a better API: the additional mode-specific `...$args` are currently applicable to the following supported modes:

1. `FETCH_CLASS` (`$class`, `$ctorArgs`).

    IMO this mode doesn't belong to the DBAL at all. It's the responsibility of the ORM to hydrate objects with the data from DBAL. As far as I understand, it only exists here because it's supported by PDO. But it's not supported by all DBAL drivers (e.g. MySQLi and OCI8).

2. `FETCH_COLUMN` (`$columnIndex`).

   The `$columnIndex` argument seems only supported by PDO natively but is ignored everywhere else. Therefore, it cannot be reliably used. Furthermore, If you're going to fetch one column, why select more? So, to me even if it's technically possible to choose which column to fetch, it doesn't make much sense from the design standpoint.

With all the above said, there's an option of dropping the modes above with their specific arguments in `3.0`. The downsides of dropping them I can see are:

1. Some projects may be using DBAL as a wrapper over PDO with some PDO-only features. Dropping support for `...$args` will break their code for no reason.
2. The DBAL itself will lose the extension point for introducing new fetch modes which require additional arguments. Not that we're going to, but anyways.

Another, more OO approach could be to replace mode constants with objects each of which could take the arguments which make sense for it. Like `$stmt->fetch(new FetchColumn(5))` or `$stmt->fetchAll(new FetchClass('Account', [$listener]))`. I would make the statement API cleaner and better typed but would make it harder to accept by users.

Anyways, it's all outside of the scope since we're trying to get rid of the PDO dependency here and the current solution is about the least BC-breaking. We can break BC later.

morozov:
continuousphp is currently [failing](https://app.continuousphp.com/git-hub/doctrine/dbal/build/fea37953-01cc-4297-b23a-8de4b06d3e06) due to #2963.

morozov:
@Ocramius want to do another round of review?

lib/Doctrine/DBAL/Driver/ResultStatement.php
```diff
@@ -28,66 +81,44 @@ public function columnCount();
...
+     * @param array    $args         Optional mode-specific arguments.
      *
      * @return mixed The return value of this method on success depends on the fetch mode. In all cases, FALSE is
      *               returned on failure.
      *
-     * @see PDO::FETCH_* constants.
+     * @see self::FETCH_* constants.
      */
-    public function fetch($fetchMode = null, $cursorOrientation = \PDO::FETCH_ORI_NEXT, $cursorOffset = 0);
+    public function fetch($fetchMode = null, ...$args);
```
Majkl578: This signature change is BC breaking in a way that is impossible to fix in a code that targets DBAL 2.x and 3.x. These types of BC breaks are really annoying for consumers and should be considered carefully. https://3v4l.org/nFacb

lib/Doctrine/DBAL/Driver/ResultStatement.php
```diff
@@ -9,6 +9,59 @@
...
 {
+    /**
+     * Specifies that the fetch method shall return each row as an array indexed
+     * by column name as returned in the corresponding result set. If the result
+     * set contains multiple columns with the same name, the statement returns
+     * only a single value per column name.
+     *
+     * @see \PDO::FETCH_ASSOC
+     */
+    const FETCH_ASSOC = 2;
```
Majkl578: I'd prefer moving these out to FetchMode enum-like class, pretty much like ORM 3.x will have i.e. GeneratorType. Either as a simple constant holder (ORM case ATM) or proper object encapsulation (encapsulates enum-like validation).
Also this is the opportunity to fix the naming - I'm not a fan of abbreviations tbh.

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -145,9 +200,51 @@ public function fetchAll($fetchMode = null, $fetchArgument = null, $ctorArgs = n
...
+
+        return self::$fetchModeMap[$fetchMode];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getIterator()
+    {
+        yield from $this->stmt;
```
Majkl578: This is interesting, didn't know \PDOStatement is a generator.

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -145,9 +200,51 @@ public function fetchAll($fetchMode = null, $fetchArgument = null, $ctorArgs = n
...
+
+    /**
+     * Converts DBAL fetch mode to PDO fetch mode
+     *
+     * @param int|null $fetchMode Fetch mode
+     * @return int|null
+     */
+    private function convertFetchMode(?int $fetchMode) : ?int
+    {
+        if ($fetchMode === null) {
```
Majkl578: Better move this out of here and drop nullability - null isn't a real fetch mode (or is it?).

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -2,40 +2,71 @@
...
  * Used by all PDO-based drivers.
  *
  * @since 2.0
  */
-class PDOStatement extends \PDOStatement implements Statement
+class PDOStatement implements IteratorAggregate, Statement
 {
     /**
-     * Protected constructor.
+     * @var array<int,int>
```
Majkl578: This format isn't generally well accepted by SA tools, it should be `int[]`.

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -2,40 +2,71 @@
...
      */
-    protected function __construct()
+    private static $paramTypeMap = [
```
Majkl578: Could be const?

lib/Doctrine/DBAL/Driver/Statement.php
```diff
@@ -26,12 +61,11 @@
...
      *                       using question mark placeholders, this will be the 1-indexed position of the parameter.
      * @param mixed   $value The value to bind to the parameter.
-     * @param integer $type  Explicit data type for the parameter using the PDO::PARAM_* constants.
+     * @param integer $type  Explicit data type for the parameter using the self::PARAM_* constants.
      *
      * @return boolean TRUE on success or FALSE on failure.
      */
-    function bindValue($param, $value, $type = null);
-
+    public function bindValue($param, $value, $type = self::PARAM_STR);
```
Majkl578: This is also a BC break, $type was previously nullable, now isn't.

lib/Doctrine/DBAL/Portability/Connection.php
```diff
@@ -13,6 +13,20 @@
  */
 class Connection extends \Doctrine\DBAL\Connection
 {
+    /**
+     * Force column names to upper case.
+     *
+     * @see \PDO::CASE_UPPER
+     */
+    const CASE_UPPER = 1;
```
Majkl578: Same like fetch mode - I'd prefer i.e. ColumnCasing enum-like structure.

Majkl578 (commented):
I like the idea behind this, API based on extending PDO is stinky as soon as non-PDO drivers exist (as they do). But I'm worried about the BC consequences especially around those interface changes.

lib/Doctrine/DBAL/Driver/PDOStatement.php
jvasseur: You can use `yield from` with anything that is `iterable` it doesn't have to be an generator.

Majkl578: Oh neat, didn't realize. 👍

lib/Doctrine/DBAL/Driver/PDOStatement.php
morozov: It means "use the default fetch mode set by `setFetchMode()`".

lib/Doctrine/DBAL/Driver/PDOStatement.php
morozov: I may have misinterpreted https://github.com/doctrine/dbal/pull/2495#discussion_r96125973. And FWIW, PHPStan supports this syntax:

https://github.com/phpstan/phpstan/blob/84b09eba4f2a4f22faa03fc4c94bc34c89bf02a9/tests/PHPStan/Analyser/data/foreach/foreach-with-specified-key-type.php#L9

From the domain standpoint, it's a map of integers to integers, not a list of integers. Do you still believe it should be `int[]`?

lib/Doctrine/DBAL/Driver/ResultStatement.php
morozov: What's the benefit of having an enum-class with constants? Right now, `FETCH_*`, `PARAM_*` and `CASE_*` constants belong to the interfaces which implement their handling which means better cohesion. Do you want to introduce an enum-class for each group of constants?

lib/Doctrine/DBAL/Driver/Statement.php
morozov: The parameter is not typed as `int $type`, so the callers still can pass whatever they want. Or I'm missing something?

lib/Doctrine/DBAL/Driver/PDOStatement.php
Majkl578: I understand, but this syntax does not work in PhpStorm at all (recognized as mixed => mixed) or even our CS. `array<>` syntax has never been part od phpDoc.

lib/Doctrine/DBAL/Driver/Statement.php
Majkl578: You're right. I was just assuming typed code in 3.x which would incorporate either `int` or `?int` typehint, and since `@param` specified `integer` (which would implicitly become nullable with `= null`), I assumed it'd become `int`. With current `mixed` it's irrelevant.

lib/Doctrine/DBAL/Driver/ResultStatement.php
Majkl578: > What's the benefit of having an enum-class with constants?

To me, it seems more consistent, because:
1) they're grouped only by their purpose,
2) they're not mixed with other unrelated constants,
3) their naming is more comprehensive,
3) they could be typed if enum-like structure is used (something like [this](https://github.com/consistence/consistence/blob/master/docs/Enum/enums.md), possibly with better/simpler API)

> Could you draft the non-abbreviated constant names as you see them?

In this case something like this:
```php
class FetchType
{
    public const ASSOCIATIVE = 0b1;
    public const NUMERIC     = 0b01;
    public const COLUMN      = 0b001;
    public const OBJECT      = 0b0001;
    public const MIXED       = self::ASSOCIATIVE | self::NUMERIC;
}
```
```php
class ParameterFormat
{
    public const NULL         = 0b1;
    public const INTEGER      = 0b01;
    public const STRING       = 0b001;
    public const BOOLEAN      = 0b0001;
    public const LARGE_OBJECT = 0b00001;
}
```
(Don't take it as a proposal to change here, rather something to consider for 3.x.)

morozov: > Don't take it as a proposal to change here, rather something to consider for 3.x.

The approach currently taken is to introduce the new constants in `2.x` and remove the assignment of the `PDO` values to them in `3.x`. This way, the callers will have a period of time when either of the APIs can be used.

Not having this changes introduced in `2.x` will make the transition much more painful.

As for the actual values, should they really be bits? It will create an impression that they can be mixed together however they are mutually exclusive. To me, a set of `1`, `2`, `3`… would look more natural. Otherwise, what if somebody calls `$stmt->fetch(FetchType::COLUMN | FetchType::OBJECT)`? We'll have to have a validation layer for those bit masks which could be avoided.

lib/Doctrine/DBAL/Driver/PDOStatement.php
morozov: Agree, will rework.

lib/Doctrine/DBAL/Driver/PDOStatement.php
morozov: Yeah, let's make it a constant.

lib/Doctrine/DBAL/Portability/Connection.php
morozov: I'm fine with enum-like classes and non-abbreviated names. Not sure if we want bit masks and we do need to introduce them in `2.x` to ease the migration.

lib/Doctrine/DBAL/Driver/ResultStatement.php
morozov: `ParameterFormat` should be probably `ParameterType`. Formatting implies a string representation of some value (like `sprintf()`). Here, we're not dealing with any formatting.

morozov: `FetchType` — personally, I'd prefer to keep it as `FetchMode`. The "mode" implies "the way of doing things" which is exactly what those constants define. The "type" is too generic IMO.

morozov: > `public const MIXED = self::ASSOCIATIVE | self::NUMERIC;`

Since it's being brought up, does the DBAL need to support this weird mode, especially defined this way? In my entire experience, I have never needed to fetch the data this way. I think it's the same sort of feature as register globals in PHP 4 which allowed developers to do both `$row[0]` and `$row['id']` without really thinking of what the `$row` actually is.

morozov:
@Majkl578 please see the update. I had to rewrite the history in order to keep the correct order of `2.x` and `3.0` commits.

lib/Doctrine/DBAL/FetchMode.php
```diff
@@ -0,0 +1,69 @@
...
+    public const MIXED = 4;
+
+    /**
+     * Specifies that the fetch method shall return each row as an object with
+     * property names that correspond to the column names returned in the result
+     * set.
+     *
+     * @see \PDO::FETCH_OBJ
+     */
+    public const OBJECT = 5;
```
Ocramius: This may be trouble. `Object` may be a reserved keyword at some point...

lib/Doctrine/DBAL/ParameterType.php
```diff
@@ -0,0 +1,51 @@
...
+ * Contains statement parameter types.
+ */
+class ParameterType
+{
+    /**
+     * Represents the SQL NULL data type.
+     *
+     * @see \PDO::PARAM_NULL
+     */
+    public const NULL = 0;
```
Ocramius: Uhhh... I can see more trouble here. We're using a lot of reserved names here.

Ocramius:
Besides possible issues in the naming of the constants (strongly suggest we prefix them, sorry!), this looks OK to me :+1:

morozov:
> strongly suggest we prefix them

What instead of prefixing them (all, I assume?), we add the `_` suffix to the keyword-like ones like nikic/php-parser does? This way, IDEs will automatically add the suffix but in general `CLASS_`, `OBJECT_` and `NULL_` are extremely rarely used (therefore, the suffix shouldn't be a problem).

Also, not sure if you read https://github.com/doctrine/dbal/pull/2958#issuecomment-355372897. If we drop support for `CLASS` and `OBJECT` modes (besides mimicking the PDO modes, what's the value?), then `NULL_` will remain the only exception.

morozov:
The approach with suffixing has a flaw since `string` is also a keyword which unlike `CLASS`, `OBJECT` and `NULL` can be used very often.

> strongly suggest we prefix them

@Ocramius I understand where the concern comes from but in my understanding, the conflict between constant names and reserved PHP keywords is only possible if PHP introduces a new syntax where a constant can be used without any prefix and therefore will be confused with a keyword. In other cases, `string`, `null` and `object` are already reserved keywords, but it's not a problem. The `CLASS` constant indeed conflicts with the magic `::class` constant and therefore is suffixed.

If you still insist on prefixing, please suggest a prefix and how/where it should be applied.

morozov:
For now, I've added a trailing underscore to the constants which may collide with reserved keywords. The rationale is that `CLASS_` and `OBJECT_` are not entirely supported by DBAL, `NULL_` is relatively rarely used and `STRING_` is the default so can be omitted in many cases. It may look like a lame excuse but I really don't know what prefix to put ahead of the types and I really don't want _all_ constants to have a prefix just because _some_ of them may collide with keywords in the future.

Besides that, while doing that, I found: https://github.com/doctrine/dbal/blob/4bda0867a2b8d1946a42a1d063104e7ceca19ed4/lib/Doctrine/DBAL/Types/Type.php#L36
and:
https://github.com/doctrine/dbal/blob/4bda0867a2b8d1946a42a1d063104e7ceca19ed4/lib/Doctrine/DBAL/Types/Type.php#L38
and:
https://github.com/doctrine/dbal/blob/4bda0867a2b8d1946a42a1d063104e7ceca19ed4/lib/Doctrine/DBAL/Types/Type.php#L42

Could it be used as a reason to leave constants as they are since? Anyways, if/when there's a conflict with new PHP features, we'll have to change the API.

Ocramius:
@morozov let's settle for no suffix. If PHP breaks any BC I shall avenge you with a personal crusade.

morozov:
I've replaced `FetchMode::CLASS_` with `::CUSTOM_OBJECT` and `::OBJECT` with `::STANDARD_OBJECT` as per @guilhermeblanco's suggestion.

> If PHP breaks any BC I shall avenge you with a personal crusade.

`ParameterType::STRING` and `::NULL` are still on me :fire:

composer.json
```diff
@@ -13,7 +13,6 @@
     ],
     "require": {
         "php": "^7.2",
-        "ext-pdo": "*",
```
Ocramius: For unaware readers: this is the big win 👍

lib/Doctrine/DBAL/ParameterType.php
Ocramius: Discussed internally: if this ever breaks, there is a bac break in the language.

Ocramius (approved):

Ocramius:
:ship:

nikic:
FYI class constant names are not subject to reserved keyword restrictions, see https://wiki.php.net/rfc/context_sensitive_lexer. `class` is the only exception, for obvious reasons. No plans to change this ;)

pbowyer:
Thanks everyone, this is excellent work!

morozov:
@nikic I believe if there was no exception for `::class`, we wouldn't have hade those arguments :)

Majkl578:
I'm really happy we decided to stick with the FetchMode and ParameterType classes with unprefixed/unsuffixed constant names, it really makes more sense this way. 🎉

PR #2959, opened 2018-01-03
Use Null Coalesce Operator

carusogabriel:
Following #2934

### DIFF
--- a/lib/Doctrine/DBAL/Cache/ArrayStatement.php
+++ b/lib/Doctrine/DBAL/Cache/ArrayStatement.php
@@ -138,11 +138,8 @@ class ArrayStatement implements \IteratorAggregate, ResultStatement
     public function fetchColumn($columnIndex = 0)
     {
         $row = $this->fetch(PDO::FETCH_NUM);
-        if (!isset($row[$columnIndex])) {
-            // TODO: verify this is correct behavior
-            return false;
-        }
 
-        return $row[$columnIndex];
+        // TODO: verify this is correct behavior
+        return $row[$columnIndex] ?? false;
     }
 }
--- a/lib/Doctrine/DBAL/Cache/ArrayStatement.php
+++ b/lib/Doctrine/DBAL/Cache/ArrayStatement.php
@@ -139,7 +139,7 @@ class ArrayStatement implements \IteratorAggregate, ResultStatement
     {
         $row = $this->fetch(PDO::FETCH_NUM);
 
-        // TODO: verify this is correct behavior
+        // TODO: verify that return false is the correct behavior
         return $row[$columnIndex] ?? false;
     }
 }
--- a/lib/Doctrine/DBAL/DriverManager.php
+++ b/lib/Doctrine/DBAL/DriverManager.php
@@ -153,7 +153,7 @@ final class DriverManager
             self::_checkParams($params);
         }
 
-        $className = $params['driverClass'] ?? $className = self::$_driverMap[$params['driver']];
+        $className = $params['driverClass'] ?? self::$_driverMap[$params['driver']];
 
         $driver = new $className();
 
@@ -393,11 +393,11 @@ final class DriverManager
             // URL schemes must not contain underscores, but dashes are ok
             $driver = str_replace('-', '_', $url['scheme']);
 
-            // The requested driver from the URL scheme takes precedence
-            // over the default driver from the connection parameters (if any).
+            // The requested driver from the URL scheme takes precedence over the
+            // default driver from the connection parameters. If the driver is
+            // an alias (e.g. "postgres"), map it to the actual name ("pdo-pgsql").
+            // Otherwise, let checkParams decide later if the driver exists.
             $params['driver'] = self::$driverSchemeAliases[$driver] ?? $driver;
-            // use alias like "postgres", else we just let checkParams decide later
-            // if the driver exists (for literal "pdo-pgsql" etc)
 
             return $params;
         }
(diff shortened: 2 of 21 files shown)


lib/Doctrine/DBAL/Cache/ArrayStatement.php
```diff
@@ -138,11 +138,8 @@ public function fetchAll($fetchMode = null, $fetchArgument = null, $ctorArgs = n
     public function fetchColumn($columnIndex = 0)
     {
         $row = $this->fetch(PDO::FETCH_NUM);
-        if (!isset($row[$columnIndex])) {
-            // TODO: verify this is correct behavior
-            return false;
-        }
 
-        return $row[$columnIndex];
+        // TODO: verify this is correct behavior
```
morozov: After having two lines smashed together, it's not clear what exactly should be verified. Previously it was clear it was about returning `false`. You may consider updating the comment.

lib/Doctrine/DBAL/DriverManager.php
```diff
@@ -152,11 +152,8 @@ public static function getConnection(
         } else {
             self::_checkParams($params);
         }
-        if (isset($params['driverClass'])) {
-            $className = $params['driverClass'];
-        } else {
-            $className = self::$_driverMap[$params['driver']];
-        }
+
+        $className = $params['driverClass'] ?? $className = self::$_driverMap[$params['driver']];
```
morozov: The second `$className = ` looks redundant.

lib/Doctrine/DBAL/DriverManager.php
```diff
@@ -398,11 +395,9 @@ private static function parseDatabaseUrlScheme(array $url, array $params): array
...
             // The requested driver from the URL scheme takes precedence
             // over the default driver from the connection parameters (if any).
-            $params['driver'] = isset(self::$driverSchemeAliases[$driver])
-                // use alias like "postgres", else we just let checkParams decide later
-                // if the driver exists (for literal "pdo-pgsql" etc)
-                ? self::$driverSchemeAliases[$driver]
-                : $driver;
+            $params['driver'] = self::$driverSchemeAliases[$driver] ?? $driver;
+            // use alias like "postgres", else we just let checkParams decide later
+            // if the driver exists (for literal "pdo-pgsql" etc)
```
morozov: What is this comment now related to?

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -873,10 +873,7 @@ public function testListTableWithBlob()
      */
     protected function createTestTable($name = 'test_table', $data = array())
     {
-        $options = array();
-        if (isset($data['options'])) {
-            $options = $data['options'];
-        }
+        $options = $data['options'] ?? array();
```
morozov: Do you want to use the short array notation for consistency with the previous code style changes?

carusogabriel: I'll use. I did not change it to avoid multiple changes in one PR :sweat_smile:

lib/Doctrine/DBAL/DriverManager.php
carusogabriel: This comment still related to the ternary expression. Should we break the right argument of `??` into the next line or should we remove it now?

morozov: > This comment still related to the ternary expression.

This is not obvious. Usually, comments are added prior to the code they describe but this commit changes the order of the comment and code.

> Should we break the right argument of `??` into the next line or should we remove it now?

I believe it should be fine to keep it above the line with the code. Both `The requested driver from the URL […]` and `use alias like "postgres" […]` describe the same single line of code.

carusogabriel: Can I suggest:
> The requested driver from the URL scheme takes precedence over the default driver from the connection parameters. If any, use alias like "postgres", else we just let checkParams decide later if the driver exists (for literal "pdo-pgsql" etc).

morozov: How about this:

> […] If the driver is an alias (e.g. "postgres"), map it to the actual name ("pdo-pgsql"). Otherwise, let checkParams decide later if the driver exists.

:+1: to combining the comments.

carusogabriel: Amending the previous commit with this comment update :man_technologist:

Ocramius (approved):
LGTM 👍 - inspecting Travis failures.

Ocramius:
Restarted the failed job.

Ocramius:
@carusogabriel that commit message is not truthful/doesn't represent the fix. A future reqder of the `blame` output will not understand why the change was applied.

Ocramius:
Talking of https://github.com/doctrine/dbal/commit/3ffa5b9a95016d201e82768aad0f7017b0ad2106

carusogabriel:
@Ocramius Now it explains everything done :smile:

morozov:
@carusogabriel to avoid having to fix code style issues after pushing your changes, you may want to try using the [pre-commit hook](https://github.com/diff-sniffer/git/releases/download/0.1.0/pre-commit.phar) which does the same checks as the CI build. See the [demo](https://asciinema.org/a/151458) (note that the link in the demo is outdated, however it still will work).

Majkl578 (approved):
Restarted failing job again.
Scrutinizer issues valid, but not caused by this PR.

Ocramius:
:ship:

PR #2960, opened 2018-01-03
Handle default values as values, not SQL expressions

morozov:
This work is based on #2850 and implements it across supported platforms.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
@@ -13,17 +13,36 @@ use function assert;
 use function explode;
 use function is_string;
 use function preg_match;
-use function str_replace;
-use function stripslashes;
 use function strpos;
 use function strtok;
 use function strtolower;
+use function strtr;
 
 /**
  * Schema manager for the MySql RDBMS.
  */
 class MySqlSchemaManager extends AbstractSchemaManager
 {
+    /**
+     * @see https://mariadb.com/kb/en/library/string-literals/#escape-sequences
+     */
+    private const MARIADB_ESCAPE_SEQUENCES = [
+        '\\0' => "\0",
+        "\\'" => "'",
+        '\\"' => '"',
+        '\\b' => "\b",
+        '\\n' => "\n",
+        '\\r' => "\r",
+        '\\t' => "\t",
+        '\\Z' => "\x1a",
+        '\\\\' => '\\',
+        '\\%' => '%',
+        '\\_' => '_',
+
+        // Internally, MariaDB escapes single quotes using the standard syntax
+        "''" => "'",
+    ];
+
     /**
      * {@inheritdoc}
      */
@@ -219,7 +238,7 @@ class MySqlSchemaManager extends AbstractSchemaManager
         }
 
         if (preg_match('/^\'(.*)\'$/', $columnDefault, $matches)) {
-            return stripslashes(str_replace("''", "'", $matches[1]));
+            return strtr($matches[1], self::MARIADB_ESCAPE_SEQUENCES);
         }
 
         switch ($columnDefault) {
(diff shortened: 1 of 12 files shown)


morozov:
Closing this as #2850 has been reverted.

Majkl578:
Is it necessary to do this in 2.x? It's a BC break, even if minor.

morozov:
It's not extremely necessary but it already was accepted for `2.x` last year (https://github.com/doctrine/dbal/pull/2850#event-1405907749). What has changed since then?

I'd agree more to move it to `3.0` if we were actively working on it but it doesn't seem to be the case.

lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
```diff
@@ -219,7 +239,11 @@ private function getMariaDb1027ColumnDefault(MariaDb1027Platform $platform, ?str
         }
 
         if (preg_match('/^\'(.*)\'$/', $columnDefault, $matches)) {
-            return stripslashes(str_replace("''", "'", $matches[1]));
+            return str_replace(
```
Majkl578: Maybe you can use `strtr` with array?

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -329,11 +328,9 @@ protected function _getPortableTableColumnDefinition($tableColumn)
             $autoincrement           = true;
         }
 
-        if (preg_match("/^['(](.*)[')]::.*$/", $tableColumn['default'], $matches)) {
+        if (preg_match("/^['(](.*)[')]::/", $tableColumn['default'], $matches)) {
             $tableColumn['default'] = $matches[1];
-        }
-
-        if (stripos($tableColumn['default'], 'NULL') === 0) {
+        } elseif (preg_match('/^NULL::/', $tableColumn['default'])) {
```
Majkl578: This needs to be case-insensitive (it was `stripos()`).

Majkl578 (changes requested):

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
morozov: Does it? Looks like it's always upper-cased in the introspection output which is covered by a test. The existing implementation may be just too precautious, same as in https://github.com/doctrine/dbal/pull/2854#discussion_r255346946.

Majkl578: I'm not sure (didn't check tests), if you are then :+1:, just from the diff you removed case-insensitive check. If same as https://github.com/doctrine/dbal/pull/2854#discussion_r255346946, then it's fine.

lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
morozov: Actually, `strtr` will give a more correct result:
```php
$str = <<<'STR'
\'\'
STR
;

$tr = [
    "\\'" => "'",
    "''"  => "'",
];

$replaced1 = str_replace(array_keys($tr), array_values($tr), $str);
$replaced2 = strtr($str, $tr);

var_dump($replaced1);
// string(1) "'"

var_dump($replaced2);
// string(2) "''"
```
Going to rework and add to the test.

morozov: The scenario above is actually practically impossible since MariaDB duplicates single quotes instead of prepending a backslash. I added a similar test which somehow passes with both implementations and switched to the proposed implementation anyways.

jwage (approved):
Ship it :+1:

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -395,11 +392,12 @@ protected function _getPortableTableColumnDefinition($tableColumn)
...
             case 'text':
-                $fixed = false;
-                break;
+            case '_varchar':
             case 'varchar':
+                $tableColumn['default'] = $this->parseDefaultExpression($tableColumn['default']);
+                $fixed                  = false;
+                break;
             case 'interval':
-            case '_varchar':
```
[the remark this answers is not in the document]
morozov: No idea. I'll put it on my TODO list.

PR #2965, opened 2018-01-05
Update Doctrine CS requirement

Majkl578:
Disabled checks for strict types and native types to avoid BC issues, should be fine for master.
Must be run on lowest PHP version - some of the sniffs are version-dependent.

Not going to apply CS globally anytime soon, at least until 2.x is closed for features and develop merged. 🔧 

This should be ok since PRs are checked only for issues in modified code, right? 😊


morozov (approved):

Ocramius (approved):
LGTM :+1:

PR #2967, opened 2018-01-05
Drop Doctrine\DBAL\Version in favor of Ocramius\PackageVersions

Majkl578:
cross-linking doctrine/doctrine2#6653


composer.json
```diff
@@ -14,7 +14,8 @@
     "require": {
         "php": "^7.1",
         "ext-pdo": "*",
-        "doctrine/common": "^2.7.1"
+        "doctrine/common": "^2.7.1",
+        "ocramius/package-versions": "^1.2"
```
morozov: I know it's already merged into `doctrine/doctrine2` but given the version is displayed in CLI to humans, would it make sense to use `jean85/pretty-package-versions` for prettier output?

Ocramius: Eh, this is mostly so we can get proper reports when people get something wrong, so I think the human consumer is mostly @doctrine/doctrinecore

Majkl578: Agreed, I would keep it simple and change if requested by users eventually, even now it's 2x more dependencies than before. 😃

Ocramius:
Manually merged via doctrine/dbal@2f085fc00d1a70e2e8fb7298c496bc9a58c68f79

PR #2968, opened 2018-01-06
the preferred way to create unique constraints, UNIQUE INDEX and UNIQ…

kimjuan:
…UE CONSTRAINTS have differences in Postgres

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -1240,6 +1240,33 @@ class PostgreSqlPlatform extends AbstractPlatform
     /**
      * {@inheritDoc}
      */
+    protected function getPreAlterTableIndexForeignKeySQL(TableDiff $diff)
+    {
+        $tableName = $diff->getName($this)->getQuotedName($this);
+
+        $sql = [];
+        if ($this->supportsForeignKeyConstraints()) {
+            foreach ($diff->removedForeignKeys as $foreignKey) {
+                $sql[] = $this->getDropForeignKeySQL($foreignKey, $tableName);
+            }
+            foreach ($diff->changedForeignKeys as $foreignKey) {
+                $sql[] = $this->getDropForeignKeySQL($foreignKey, $tableName);
+            }
+        }
+
+        foreach ($diff->removedIndexes as $index) {
+            $sql = array_merge($sql, $this->getDropIndexSQL($index, $tableName));
+        }
+        foreach ($diff->changedIndexes as $index) {
+            $sql = array_merge($sql, $this->getDropIndexSQL($index, $tableName));
+        }
+
+        return $sql;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
     public function getDropIndexSQL($index, $table = null)
     {
         $isUnique = false;
@@ -1251,9 +1278,12 @@ class PostgreSqlPlatform extends AbstractPlatform
         }
 
         if ($isUnique && $table) {
-            return 'ALTER TABLE ' . $table . ' DROP CONSTRAINT ' . $index;
+            return [
+                sprintf('ALTER TABLE %s DROP CONSTRAINT IF EXISTS %s', $table, $index),
+                'DROP INDEX IF EXISTS ' . $index
+            ];
         } else {
-            return 'DROP INDEX ' . $index;
+            return [ 'DROP INDEX ' . $index ];
         }
     }
 
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -1236,4 +1236,54 @@ class PostgreSqlPlatform extends AbstractPlatform
     {
         return $type instanceof IntegerType || $type instanceof BigIntType;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getDropIndexSQL($index, $table = null)
+    {
+        $isUnique = false;
+        if ($index instanceof Index) {
+            $isUnique = $index->isUnique();
+            $index = $index->getQuotedName($this);
+        } elseif (!is_string($index)) {
+            throw new \InvalidArgumentException('AbstractPlatform::getDropIndexSQL() expects $index parameter to be string or \Doctrine\DBAL\Schema\Index.');
+        }
+
+        if ($isUnique && $table) {
+            return 'ALTER TABLE ' . $table . ' DROP CONSTRAINT ' . $index;
+        } else {
+            return 'DROP INDEX ' . $index;
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getCreateIndexSQL(Index $index, $table)
+    {
+        if ($table instanceof Table) {
+            $table = $table->getQuotedName($this);
+        }
+        $name = $index->getQuotedName($this);
+        $columns = $index->getQuotedColumns($this);
+
+        if (count($columns) == 0) {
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
```diff
@@ -19,7 +19,7 @@ public function getGenerateTableWithMultiColumnUniqueIndexSql()
     {
         return array(
             'CREATE TABLE test (foo VARCHAR(255) DEFAULT NULL, bar VARCHAR(255) DEFAULT NULL)',
-            'CREATE UNIQUE INDEX UNIQ_D87F7E0C8C73652176FF8CAA ON test (foo, bar)'
+            'ALTER TABLE test ADD CONSTRAINT UNIQ_D87F7E0C8C73652176FF8CAA UNIQUE  (foo, bar)'
```
Ocramius: This is actually a breaking change, or will existing schema with unique indexes be correctly handled here?

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -1236,4 +1236,54 @@ private function isNumericType(Type $type) : bool
     {
         return $type instanceof IntegerType || $type instanceof BigIntType;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getDropIndexSQL($index, $table = null)
```
Ocramius: There's a lot of added code in this child class - are we 100% sure that all of it is covered by test automation?

Ocramius (changes requested):

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -1236,4 +1236,54 @@ private function isNumericType(Type $type) : bool
...
+            $isUnique = $index->isUnique();
+            $index = $index->getQuotedName($this);
+        } elseif (!is_string($index)) {
+            throw new \InvalidArgumentException('AbstractPlatform::getDropIndexSQL() expects $index parameter to be string or \Doctrine\DBAL\Schema\Index.');
+        }
+
+        if ($isUnique && $table) {
+            return 'ALTER TABLE ' . $table . ' DROP CONSTRAINT ' . $index;
+        } else {
+            return 'DROP INDEX ' . $index;
```
morozov: If a unique constraint was created using `ALTER TABLE […] ADD CONSTRAINT` and now is being dropped using `DROP INDEX`, will it succeed? If not, I'd say using indices instead of constraints in DBAL is by design since it only operates indices.

kimjuan: _Backward-Remains-Compatible_:

- All operations involve `<indexes>` and `<index>`.
- `<uniqueConstraints>`: Renaming is ok. `<uniqueConstraint name="old_name_constraints" columns="..." />` to `<uniqueConstraint name="new_name_constraints" />`
- Freshly added  `<uniqueConstraints>`

_Things that will BC-Break_ (although won't break production, will likely break their builds):

- Existing schema that uses `<uniqueConstraints>` generated with `CREATE UNIQUE INDEX indexName ON table (...)` will not be able to be dropped. Those can only be dropped using old syntax `DROP INDEX indexName`, but dbal will generate `ALTER TABLE DROP CONSTRAINT indexName`

Admittedly this could be totally prevented with better implementation.

tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
kimjuan: Yes. I didn't thought it out well.

_Backward-Remains-Compatible_:

- All operations involve `<indexes>` and `<index>`.
- `<uniqueConstraints>`: Renaming is ok. `<uniqueConstraint name="old_name_constraints" columns="..." />` to `<uniqueConstraint name="new_name_constraints" />`
- Freshly added  `<uniqueConstraints>`

_Things that will BC-Break_ (although won't break production, will likely break their builds):

- Existing schema that uses `<uniqueConstraints>` generated with `CREATE UNIQUE INDEX indexName ON table (...)` will not be able to be dropped. Those can only be dropped using old syntax `DROP INDEX indexName`, but dbal will generate `ALTER TABLE DROP CONSTRAINT indexName`

This could be totally prevented. Let me work on it.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
kimjuan: context: `drop index` on constraints created using `ALTER TABLE`

`tablename=# DROP INDEX unique_constraint_name;`
`ERROR:  cannot drop index unique_constraint_name because constraint unique_constraint_name on table tablename requires it`
`HINT:  You can drop constraint unique_constraint_name on table tablename instead.`

However: renaming is perfectly fine. In fact, the renaming syntax is not changed and continue to use the original syntax.

kimjuan: context:

Dropping index using `DROP INDEX constraintname` ON constraint created using `ALTER TABLE tablename ADD CONSTRAINT constraintname UNQIUE(...)` will encounter error `constraint indexName not found on table`

Vice versa:

Dropping index using `ALTER TABLE tablename DROP CONSTRAINT constraintname` on constraint created using `CREATE INDEX UNIQUE constraintname...` will encounter error `... table requires it, you can drop constraint constraintname on table tablename instead`

Let me find out if schemaManager provides constraint information on non-foreignKey constraint...

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
kimjuan: Yes

kimjuan:
Schema\Index, schemaManager doesn't have give any signal if an index is a constraint (apart isUnique or not).

The most non intrusive way may be running both using IF EXISTS clause.

`ALTER TABLE tablename DROP CONSTRAINT IF EXISTS constraintname`
`DROP INDEX IF EXISTS constraintname`

This unit of 2 command is also idempotent, if network error happens before/in between/after, rerunning schema updates will work as expected.

simPod:
What is the status of this, can I somehow help to finish?

simPod:
What do you think about adding `DEFERRABLE` to `ADD CONSTRAINT` statement?

morozov:
> What is the status of this, can I somehow help to finish?

@simPod,

1. The build is all read, it needs to be fixed.
2. Is it possible to add another test which would show how the preferred way of creating indices is better than the currently used? I.e. why do we need this change at all if it's just a matter of preference?

simPod:
@morozov the main reason why I was looking for and came accross this PR are [deferred constraints](https://www.postgresql.org/docs/current/static/sql-set-constraints.html). That cannot be done with indices.

simPod:
@morozov

> does it mean that if the code relies on deferred constraints supported by PostgreSQL it will not work on other platforms which don't support them

Yes. When you have _defferable_ + _initially imediate_ (_initially_deffered=no_) constraint, you can deffer them in transaction. I do it like `conn->executeQuery('SET CONSTRAINTS constraint_name DEFERRED')` within `em->transactional()`.

TBH I have not found another way how to overcome cyclic updates.

> What would be the API for deferred constraints and how will it behave on the platforms which don't support them?

That is what I'm indirectly asking about 🤔 Personally, I don't see a difference in having _defferable_ instead of _immediate_ constraint created for postgres platform as it still behaves the same as long as it's _initially_immediate_. It seems to make no difference. But to those people who wish it allows to deffer them. So the API would perhaps stay the same.

morozov:
Closing as stale.

PR #2971, opened 2018-01-09
Use $strict param in functions that have it

carusogabriel:
Some functions, e.g. `in_array`, have a third param to enable strict comparison :shield:

### DIFF
--- a/lib/Doctrine/DBAL/DriverManager.php
+++ b/lib/Doctrine/DBAL/DriverManager.php
@@ -204,7 +204,7 @@ final class DriverManager
             throw DBALException::unknownDriver($params['driver'], array_keys(self::$_driverMap));
         }
 
-        if (isset($params['driverClass']) && ! in_array('Doctrine\DBAL\Driver', class_implements($params['driverClass'], true))) {
+        if (isset($params['driverClass']) && ! in_array('Doctrine\DBAL\Driver', class_implements($params['driverClass'], true), true)) {
             throw DBALException::invalidDriverClass($params['driverClass']);
         }
     }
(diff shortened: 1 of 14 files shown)


lib/Doctrine/DBAL/DriverManager.php
```diff
@@ -204,7 +204,7 @@ private static function _checkParams(array $params): void
             throw DBALException::unknownDriver($params['driver'], array_keys(self::$_driverMap));
         }
 
-        if (isset($params['driverClass']) && ! in_array('Doctrine\DBAL\Driver', class_implements($params['driverClass'], true))) {
+        if (isset($params['driverClass']) && ! in_array('Doctrine\DBAL\Driver', class_implements($params['driverClass'], true), true)) {
```
morozov: I'm not a fan of complicating code for no reason. In this case, both elements of the `$haystack` and `$needle` are strings. What's the point of changing this code?

I agree strict comparison can be used when the types are not clear/enforced, but it's not applicable to many cases in  this pull request.

carusogabriel: @Ocramius So should we wait to add [phpstan/phpstan-strict-rules](https://github.com/phpstan/phpstan-strict-rules)?

Ocramius: Possibly, yes. That would take a lot of work to integrate though.

carusogabriel:
Closing as we will wait to integrate with [phpstan/phpstan-strict-rules](https://github.com/phpstan/phpstan-strict-rules)

PR #2973, opened 2018-01-10
Enhance phpdoc of ResultStatement::fetchAll()

UFOMelkor:
See http://de2.php.net/manual/en/pdostatement.fetchall.php

### DIFF
--- a/lib/Doctrine/DBAL/Driver/ResultStatement.php
+++ b/lib/Doctrine/DBAL/Driver/ResultStatement.php
@@ -87,18 +87,19 @@ interface ResultStatement extends \Traversable
     /**
      * Returns an array containing all of the result set rows.
      *
-     * @param int|null $fetchMode     Controls how the next row will be returned to the caller.
-     *                                The value must be one of the \PDO::FETCH_* constants,
-     *                                defaulting to \PDO::FETCH_BOTH.
-     * @param int|null $fetchArgument This argument has a different meaning depending on the value of the $fetchMode parameter:
-     *                                * \PDO::FETCH_COLUMN: Returns the indicated 0-indexed column.
-     *                                * \PDO::FETCH_CLASS: Returns instances of the specified class, mapping the columns of each
-     *                                  row to named properties in the class.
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Driver/ResultStatement.php
```diff
@@ -87,18 +87,19 @@ public function fetch($fetchMode = null, $cursorOrientation = \PDO::FETCH_ORI_NE
...
+     * @param int|null                 $fetchMode     Controls how the next row will be returned to the caller.
+     *                                                The value must be one of the \PDO::FETCH_* constants,
+     *                                                defaulting to \PDO::FETCH_BOTH.
+     * @param int|string|callable|null $fetchArgument This argument has a different meaning depending on the value of
+     *                                                the $fetchMode parameter:
+     *                                                * \PDO::FETCH_COLUMN: Returns the indicated 0-indexed column.
+     *                                                * \PDO::FETCH_CLASS: Returns instances of the specified class,
+     *                                                  mapping the columns of each row to named properties in the
+     *                                                  class.
+     *                                                * \PDO::FETCH_FUNC: Returns the results of calling the specified
```
morozov: This mode is not supported by any drivers other than PDO. Shouldn’t be part of the generic API documentation.

morozov (changes requested):

lib/Doctrine/DBAL/Driver/ResultStatement.php
UFOMelkor: Should it be moved to PDOStatement or completely removed?

morozov: I'd leave it as is. Technically, `PDO::FETCH_FUNC` and corresponding `callable` type are only supported by PDO and will be deprecated in `2.x` and removed in `3.0` as part of #2958. Adding the type just before deprecating it doesn't make much sense to me. Removing support for a type (even in documentation) is probably a BC break.

pauliusrum:
Would this be merged with out the callable type hint?

morozov:
@pauliusrum no, because it's deprecated in DBAL and will be removed in 3.0.

PR #2980, opened 2018-01-15
Simplify returns

carusogabriel:


carusogabriel:
Should I apply the CS, or there will be a PR just for that?

lib/Doctrine/DBAL/Schema/Comparator.php
```diff
@@ -534,10 +526,6 @@ private function isALegacyJsonComparison(Types\Type $one, Types\Type $other) : b
      */
     public function diffIndex(Index $index1, Index $index2)
     {
-        if ($index1->isFullfilledBy($index2) && $index2->isFullfilledBy($index1)) {
-            return false;
-        }
-
-        return true;
+        return ! ($index1->isFullfilledBy($index2) && $index2->isFullfilledBy($index1));
```
morozov: I'd argue this approach is better than the original one. When debugging code, it may be convenient to set a break point on one of the `return`s and inspect the scope before switching the context. Especially, when the condition is complex. A better rework could look like:
```php
if (!$index1->isFullfilledBy($index2)) {
    return true;
}

if (!$index2->isFullfilledBy($index1)) {
    return true;
}

return false;
```

lib/Doctrine/DBAL/Schema/Index.php
```diff
@@ -251,11 +247,7 @@ public function overrules(Index $other)
             return false;
         }
 
-        if ($this->spansColumns($other->getColumns()) && ($this->isPrimary() || $this->isUnique()) && $this->samePartialIndex($other)) {
-            return true;
-        }
-
-        return false;
+        return $this->spansColumns($other->getColumns()) && ($this->isPrimary() || $this->isUnique()) && $this->samePartialIndex($other);
```
morozov: Same as above, this condition could be split into a series for better readability.

morozov:
> Should I apply the CS, or there will be a PR just for that?

I believe, all tickets marked as "Code Style" should be reflected in the coding standards if technically possible. Otherwise, it will be a never-ending story of writing new code and fixing the style.

Also, having it standardized will help make sure all contributors agree with the code style, not only the ones involved in work on a given pull request.

lib/Doctrine/DBAL/Schema/Comparator.php
carusogabriel: May I do this change in another PR? Keeping one proposal per PR? I agree with you, and there are other places that we can change it :smile:

lcobucci: I find this much more readable and simpler to be honest:

```php
return ! $index1->isFullfilledBy($index2) || ! $index2->isFullfilledBy($index1);
```

Ocramius: @lcobucci that's just De Morgan's law applied (except that @carusogabriel's is the "simplified" version ;-) )

Ocramius:
@morozov Shipping this for now. I think we should get @carusogabriel to propose these in the https://github.com/doctrine/coding-standard once he's joined the party :-)

lib/Doctrine/DBAL/Schema/Comparator.php
carusogabriel: > "De Morgan's law"

Learn something new today :smile:

PR #2986, opened 2018-01-22
Platform names as constants

simPod:
I'd like to reference these from outside for eg. platform mapping to other db libraries.

Is this therefore viable?

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/DB2Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/DB2Platform.php
@@ -12,6 +12,8 @@ use Doctrine\DBAL\Types\Type;
 
 class DB2Platform extends AbstractPlatform
 {
+    public const NAME = 'db2';
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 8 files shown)


lcobucci (changes requested):
@simPod I honestly don't see much benefit on this kind of change, I'd rather remove `AbstractPlatform#getName()` and simply rely on `instanceof` - which you would be able to use the class reference ([even as string](https://3v4l.org/IKBqc)).

@Ocramius what's your take on this?

Ocramius:
`develop` was rebased on `master`, so this patch went a bit off-track. Only https://github.com/doctrine/dbal/pull/2986/commits/bbdae66d294167107199394aa943cad3ee3ad831 should be kept.

Ocramius:
Since this targets DB2, I'd rather prefer using the `PlatformClass::class` as name (consistent with `get_class()` and `instanceof` behavior).

`getName()` should probably be deprecated and removed instead of improved.

simPod:
@Ocramius and is it still worth doing then if it's for DBAL2?

Ocramius:
No, because it would be a BC break

simPod:
Shall we close then?

Ocramius:
Yes, let's close here :-)

If you want, you can send a patch against `develop` that completely removes `getName()`

PR #2989, opened 2018-01-23
Update license's copyright

SenseException:
License file is sufficient (#2853) but outdated. Is an end year even needed for MIT?

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2006-2012 Doctrine Project
+Copyright (c) 2006-2018 Doctrine Project
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of
 this software and associated documentation files (the "Software"), to deal in


Majkl578:
Afaik end year is not required, starting year should be enough. Copyrights [do expire](https://en.wikipedia.org/wiki/List_of_countries'_copyright_lengths) though, but who would use current code in 50 years or so?
@Ocramius may know more about this.

Ocramius:
I just dropped it from my codebases, but not because I know better, just because I stopped caring about it, since the licence of my packages is extremely permissive anyway.

Ocramius:
Please also remember that we DO sign our releases and we DO have a provable git history to claim anything related to copyrights, if ever needed.

Ocramius:
Meanwhile, merging this. Thanks @SenseException!

SenseException:
If that's the case, I would have preferred to remove the end year, so there would be no need to update the `LICENSE` file in the future.

PR #2996, opened 2018-01-25
Forward compatibility with 3.x

### LINKED ISSUE #2953: Unlike other drivers, DBAL has a hard dependency on PDO

morozov:
This is a back-port of non-BC-breaking part of #2958.

Fixes #2953.


docs/en/reference/portability.rst
```diff
@@ -51,12 +51,16 @@ Using the following code block in your initialization will:
...
+
     $params = array(
         // vendor specific configuration
         //...
-        'wrapperClass' => 'Doctrine\DBAL\Portability\Connection',
-        'portability' => \Doctrine\DBAL\Portability\Connection::PORTABILITY_ALL,
-        'fetch_case' => \PDO::CASE_LOWER,
+        'wrapperClass' => PortableConnection::class,
+        'portability'  => PortableConnection::PORTABILITY_ALL,
+        'fetch_case'   => PortableConnection::LOWER,
```
Majkl578: Should be ColumnCase::LOWER.

lib/Doctrine/DBAL/Cache/ArrayStatement.php
```diff
@@ -100,23 +100,30 @@ public function getIterator()
...
-            } elseif ($fetchMode === PDO::FETCH_NUM) {
-                return array_values($row);
-            } elseif ($fetchMode === PDO::FETCH_BOTH) {
-                return array_merge($row, array_values($row));
-            } elseif ($fetchMode === PDO::FETCH_COLUMN) {
-                return reset($row);
-            } else {
-                throw new \InvalidArgumentException("Invalid fetch-style given for fetching result.");
-            }
+        if ( ! isset($this->data[$this->num])) {
```
Majkl578: There shouldn't be space on the left side of `!`. (applies to other occurences too)

lib/Doctrine/DBAL/ColumnCase.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace Doctrine\DBAL;
+
+/**
+ * Contains portable column case conversions.
+ */
+class ColumnCase
```
Majkl578: Can we make these final, maybe?

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
```diff
@@ -62,8 +64,8 @@ class DB2Statement implements \IteratorAggregate, Statement
      * @var array
      */
     static private $_typeMap = [
-        \PDO::PARAM_INT => DB2_LONG,
-        \PDO::PARAM_STR => DB2_CHAR,
+        ParameterType::INTEGER => DB2_LONG,
+        ParameterType::STRING => DB2_CHAR,
```
Majkl578: Do we align array members the same way as assignments? 🤔

lib/Doctrine/DBAL/Cache/ArrayStatement.php
```diff
@@ -100,23 +100,30 @@ public function getIterator()
      */
     public function fetch($fetchMode = null, $cursorOrientation = \PDO::FETCH_ORI_NEXT, $cursorOffset = 0)
```
Majkl578: `\PDO::FETCH_ORI_NEXT` stays as is?

lib/Doctrine/DBAL/Driver/Statement.php
```diff
@@ -43,12 +45,12 @@
...
      * @param mixed   $value The value to bind to the parameter.
-     * @param integer $type  Explicit data type for the parameter using the PDO::PARAM_* constants.
+     * @param integer $type  Explicit data type for the parameter using the {@link \Doctrine\DBAL\ParameterType}
+     *                       constants.
      *
      * @return boolean TRUE on success or FALSE on failure.
      */
-    function bindValue($param, $value, $type = null);
-
+    public function bindValue($param, $value, $type = ParameterType::STRING);
```
Majkl578: Isn't change of default here from null to string BC breaking? (Applies to other signature changes as well.)

Majkl578 (changes requested):
Nice cleanup. 👍 
Travis caught some CS issues on modified lines that are not directly caused by this PR.

docs/en/reference/portability.rst
morozov: Yes, missed it during rework.

lib/Doctrine/DBAL/Cache/ArrayStatement.php
morozov: Yes, in `2.x`.

lib/Doctrine/DBAL/ColumnCase.php
morozov: Yeah. I took it as is from your draft and though `final private function __construct()` is enough. But it only prohibits instantiation, not extension.

lib/Doctrine/DBAL/Driver/Statement.php
morozov: PHP is fine with that judging by https://3v4l.org/95JYL. Semantically, both `null` and `ParameterType::STRING` are the same since the latter is default everywhere.

Ocramius: I'd be cautious with that. Does the test suite at least contain a test with `bindValue()` called only with 2 parameters?

morozov: @Ocramius there is a test like that:

https://github.com/doctrine/dbal/blob/bc23831907cdb0bc462b2460eb31a21d021ac6c2/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php#L180

This change indeed was only required for `develop` to make this code work:
https://github.com/doctrine/dbal/blob/3fd923919fc6bc81661c30c086b97f58289a38f8/lib/Doctrine/DBAL/Driver/PDOStatement.php#L73

I understand your concern but I don't know what to look for. I manually checked all implementations, and the all either default to binding as string explicitly (IBM DB2, MySQLi, PDO) or don't specify the string/null type at all (OCI8, SQL Server) or seem to not even accept `NULL` (SQLAnywhere).

Furthermore, `NULL` appears to not be a valid type for PDO and logically makes little sense as well:
```php
$conn = new PDO('mysql:host=localhost;dbname=mysql');
$stmt = $conn->prepare('SELECT ?');
$stmt->bindValue(1, 'foo', null);
$stmt->execute();
$value = $stmt->fetchColumn();
var_dump($value);

// NULL
```
I can revert changes in all `Statement::bind*()` signatures if needed.

morozov:
@Majkl578 looks like [some of the code style violations](https://travis-ci.org/doctrine/dbal/jobs/333505857#L579) are reported on wrong lines. For instance:
```
FILE: lib/Doctrine/DBAL/Cache/ArrayStatement.php
----------------------------------------------------------------------
FOUND 3 ERRORS AFFECTING 3 LINES
----------------------------------------------------------------------
  45 | ERROR | [x] Expected "int" but found "integer" in @var
     |       |     annotation.
```
The `@var integer` annotation is on line 43 which is not affected by this pull request and therefore shouldn't be reported. It should be fixed on the code standard side. It should report the annotation itself, not the return value type.

Majkl578:
I noticed that too while working on #2998. The errors seem to be reported for the element they're connected to. See [here](https://github.com/slevomat/coding-standard/blob/4.3.1/SlevomatCodingStandard/Sniffs/TypeHints/LongTypeHintsSniff.php#L73) - `$pointer` is the position of matched element (either T_VARIABLE or T_FUNCTION. Not sure if there is any reasoning behind this, normally it's not a problem, but here with the line-based filter it is...

morozov:
@Majkl578 I think we can live with that. I suspect it may be hard to justify the needed changes in a 3rd-party library. On the other hand, we could get rid of the false positives by fixing annotations.

Maybe @carusogabriel could take a look? 😉

carusogabriel:
@morozov How can I help? :nerd_face:

Majkl578:
@morozov Reported here: #250. It should be trivial to fix for LongTypeHInt sniff, but not sure about TypeHintDeclaration sniff which is pretty complex already.

kukulich:
@morozov 

> It should be fixed on the code standard side.

Try `composer update`. It will work now, I hope :)

Majkl578:
@kukulich Awesome, thanks!

@morozov Retriggerred the CS stage: https://travis-ci.org/doctrine/dbal/jobs/334442418#L524, can you check now?

morozov:
@carusogabriel judging by [CI failures](https://travis-ci.org/doctrine/dbal/jobs/333505857), we have some `@param` and `@return` method annotations which define type of value as `integer` or `boolean`. The code style requires them to be represented as `int` and `bool` accordingly.

morozov:
@Majkl578, @kukulich thank you for the update. There's [one more false positive](https://travis-ci.org/doctrine/dbal/jobs/334442418#L601) like the fixed ones:
```
FILE: lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
----------------------------------------------------------------------
----------------------------------------------------------------------
 101 | ERROR | Property
     |       | \Doctrine\DBAL\Driver\SQLSrv\SQLSrvStatement::$defaultFetchMode
     |       | does not have @var annotation.
----------------------------------------------------------------------
```

Unlike those, it doesn't seem possible to fix on the standard side since it reports the absence of annotation. Unless you have an idea.

@carusogabriel, mind fixing issues like this as well?

carusogabriel:
@morozov No problem. I'm just finishing some stuff, I'll PR separately the annotation types and missing `@var`!

morozov:
[One more falses-positive](https://travis-ci.org/doctrine/dbal/jobs/334442418#L785):
```
FILE: tests/Doctrine/Tests/DBAL/Functional/PortabilityTest.php
----------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
----------------------------------------------------------------------
 137 | ERROR | [x] Method
     |       |     \Doctrine\Tests\DBAL\Functional\PortabilityTest::testPortabilityPdoSqlServer()
     |       |     does not need documentation comment.
```
The standard doesn't account for PHPUnit annotations.

Majkl578:
This last one is ours configuration issue - [we do have most of the common annotations whitelisted](https://github.com/doctrine/coding-standard/blob/master/lib/Doctrine/ruleset.xml#L167), but I probably forgot this one.

Ocramius:
Argh, the CS changes introduced recently led to a few merge conflicts here, @morozov. Sorry!

lib/Doctrine/DBAL/Driver/Statement.php
Ocramius: I'm fine if a test already exists and doesn't explicitly pass the third argument to `bindValue()`.

My only issue was that without a test verifying the default parameter we'd have a stability problem :-)

And yes, PDO is (as usual) being weird with default values :-P

tests/Doctrine/Tests/DBAL/DriverManagerTest.php
```diff
@@ -217,7 +240,7 @@ public function databaseUrls()
                 false,
             ),
             'URL without scheme but default PDO driver' => array(
-                array('url' => '//foo:bar@localhost/baz', 'pdo' => $pdoMock),
+                array('url' => '//foo:bar@localhost/baz', 'pdo' => true),
```
Ocramius: This is a test change that I don't understand.

Ocramius (changes requested):
Overall good improvement. Besides a change in a test that isn't really clear to me, this is good to go 👍

tests/Doctrine/Tests/DBAL/DriverManagerTest.php
morozov: @Ocramius look at the changes in `testDatabaseUrl()`. Instead of creating mock, the data provider passes a flag to the test method. It will either create the mock if PDO is loaded or will skip the data set.

Ocramius: Got it 👍

morozov:
I believe we can ignore the code style CI failure for this PR. My entire idea of line-based validation was to avoid the "Fixed code style" commits. But given that this one is a back-port from `develop`, we cannot amend the ported commits because it will introduce further conflicts during rebase of `develop`.

morozov:
@Ocramius anything else you want to get done before merging this?

Ocramius:
Nope, :ship:

Ocramius (approved):

Ocramius:
Handled further CS in #3002

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
```diff
@@ -20,7 +20,8 @@
 namespace Doctrine\DBAL\Driver\SQLSrv;
 
 use Doctrine\DBAL\Driver\StatementIterator;
-use PDO;
```
Majkl578: This is bug - there is still unqualified `PDO::FETCH_ORI_NEXT` reference.

morozov: Looks like the only place like this. Please see https://github.com/doctrine/dbal/pull/3024.

PR #2998, opened 2018-01-27
Extract some constants into TransactionIsolationLevel, TrimMode and DateIntervalUnit

Majkl578:
Similar what we did in #2958 with FetchMode and ParameterType, here comes the same for:
* `Connection::TRANSACTION_*` -> ~`TransactionMode`~ `TransactionIsolationLevel`
* `AbstractPlatform::TRIM_*` -> `TrimMode`
* `AbstractPlatform::DATE_INTERVAL_*` -> `DateIntervalUnit`

This only moves the constants and deprecates the old ones, not actually removing them yet, should be okay for next 2.x.


lib/Doctrine/DBAL/TransactionMode.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL;
+
+final class TransactionMode
```
morozov: Since this class defines transaction isolation levels, would it make sense to name it accordingly?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -765,26 +765,26 @@ public function getModExpression($expression1, $expression2)
     /**
      * Returns the SQL snippet to trim a string.
      *
-     * @param string         $str  The expression to apply the trim to.
-     * @param integer        $pos  The position of the trim (leading/trailing/both).
-     * @param string|boolean $char The char to trim, has to be quoted already. Defaults to space.
+     * @param string      $str  The expression to apply the trim to.
+     * @param int         $pos  The position of the trim (leading/trailing/both).
```
morozov: Should `$pos` be renamed to `$mode`?

lib/Doctrine/DBAL/TransactionMode.php
Majkl578: Makes sense. So `TransactionIsolation` or `TransactionIsolationLevel`?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
Majkl578: Probably, will change that - didn't check variable names to be honest.

lib/Doctrine/DBAL/TransactionMode.php
morozov: FWIW, I'd personally prefer `TransactionIsolationLevel`.

Ocramius (approved):
LGTM 👍

Ocramius:
Manually merged via doctrine/dbal@d9aaf5429ac77e9daf8ef0c06260bb0f3bcc189b

PR #3000, opened 2018-01-29
Use short array syntax

carusogabriel:
As requested by @morozov in #2996


carusogabriel:
This will need some work

morozov:
Let's hold this one off. I wanted to fix the issues which cause false positives in code style validation. This is not one of them.

PR #3003, opened 2018-01-29
Inherit table charset/collation from db charset

tvlooy:
If the database is utf8mb4, tables should be utf8mb4 by default.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
@@ -1058,9 +1058,13 @@ abstract class AbstractSchemaManager
         }
 
         $params = $this->_conn->getParams();
-        if (isset($params['defaultTableOptions'])) {
-            $schemaConfig->setDefaultTableOptions($params['defaultTableOptions']);
+        if (! isset($params['defaultTableOptions'])) {
+            $params['defaultTableOptions'] = [];
         }
+        if (! isset($params['defaultTableOptions']['charset']) && isset($params['charset'])) {
+            $params['defaultTableOptions']['charset'] = $params['charset'];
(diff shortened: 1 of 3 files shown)


tvlooy:
@kimhemsoe I added a test

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -1058,9 +1058,13 @@ public function createSchemaConfig()
         }
 
         $params = $this->_conn->getParams();
-        if (isset($params['defaultTableOptions'])) {
-            $schemaConfig->setDefaultTableOptions($params['defaultTableOptions']);
+        if (! isset($params['defaultTableOptions'])) {
+            $params['defaultTableOptions'] = [];
         }
+        if (! isset($params['defaultTableOptions']['charset']) && isset($params['charset'])) {
```
mikeSimonson: Should we test for emptiness ?

tvlooy: I guess if you set $params['charset'] to null, '', [], an object or ... you are deliberately trying to break stuff. Do I understand your question right?

tvlooy:
Can we merge this?

mikeSimonson:
@tvlooy Thanks

tvlooy:
@mikeSimonson I don't know how doctrine's release process works but will this be in the next 2.6.* too? http://www.doctrine-project.org/contribute.html says that there are release-* branches but I don't see these in dbal. Is this the 2.6 branch?

Majkl578:
New features are not backported to already stable releases, you will need to wait for 2.7.

tvlooy:
but this is a bugfix

Issue #3004, opened 2018-01-30
Rename protected and private class members starting with an underscore


morozov:
According to [PSR-2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md), protected and private property and method names should not be prefixed with an underscore. Currently, we have some of them in the code base:
```
↪ git grep -P '(protected|private) (function |\$)_(?!_)' v2.8.0 | wc -l
284
```

In order to fix the coding standard violations, they should be renamed which would be backward incompatible. A major release could be used to make such a change.

morozov:
Partially fixed in https://github.com/doctrine/dbal/pull/3303 and https://github.com/doctrine/dbal/pull/3306.

Majkl578:
For the rest of protected underscored members, imho the best approach would be to make them private and expose protected constructor instead(https://github.com/doctrine/dbal/pull/3007#issuecomment-362605345). Should this be too annoying, there could be a __get() polyfill of some sort.
@Ocramius @morozov Thoughts?

PR #3007, opened 2018-01-31
Remove underscores from private/protected members names

mateuszsip:
Closes https://github.com/doctrine/dbal/issues/3004


Majkl578:
Restarted the failing build. Although not a blocker, can you please look into the CS issues? We currently have a rule that modified code should follow Doctrine CS (even if rest of the code doesn't yet).
We should also consider a BC layer for master in separate PR.
Otherwise 👍.

morozov:
@Majkl578 I believe we could ignore the code style issues for massive automated changes like this as long as they don't introduce new issues.

What kind of a BC layer do you have in mind? For protected methods, we could have the deprecated old methods which invoke the new ones. But what about properties?

Ocramius:
Benefits here are too marginal to consider the patch, compared to massive BC concerns. Closing as `won't fix`.

Majkl578:
How about reconsidering what really needs to be protected and for those provide final protected getter (and make fields private)?

Ocramius:
@Majkl578 anything that we can make `final` probably should be, but we should do that before considering any style changes. That is a design change :-)

morozov:
Besides the above, we still can fix private methods and properties and protected methods and properties in the non-abstract test classes.

As for private/protected as such, there's also lack of consistency across the codebase. For instance, `MysqliConnection::_conn` if private but `OCI8Connection::dbh` is protected.

PR #3010, opened 2018-02-03
[PoC][WIP] Encapsulate constant crates as enum-like structures

Majkl578:
_This is currently just a proof of concept._

**Advantages**:
 * type safety
   * guarantees in type hints
   * guarantees about value set (no additional validation / _else case_ needed)
   * type analysis
 * working identity comparison

**Disadvantages**:
 * performance (compared to constants - not tested)
 * can't be used as native _default_ value (use null instead?)

**Implementation**:
 * trait
 * final methods, minimalist public API
 * each constant is accessible as a method (either explicit or magic)
 * one unique instance per enum value
 * _secret_ instance registry (no private static field) to prevent anyone messing with reflection
 * no private field for enum value to prevent anyone messing with reflection
 * no direct instantiation/(un)serialization/cloning
 * `()` accessor for real value
 * not really reviewed for optimization
 * values are currently required to be unique (per type)

Tests currently pass for SQLite and Postgres, fail for MySQL/MariaDB - haven't investigated (yet).

Original approach inspired by [the Consistence library](https://github.com/consistence/consistence/blob/master/docs/Enum/enums.md), but completely redesigned.

Naming was chosen to not collide with [PHP `enum` RFC](https://wiki.php.net/rfc/enum) should it ever be accepted into PHP language.
Current implementation is meant as a starting point for discussion/improvement, not as a final piece of code.

Let me know what you think, what to change or improve. 👍


lib/Doctrine/DBAL/ColumnCase.php
```diff
@@ -21,10 +25,13 @@
      */
     public const LOWER = 2;
```
morozov: Are public constants still needed or their values could be moved directly to the methods which use them?

lib/Doctrine/Enumeration/Enumerated.php
```diff
@@ -0,0 +1,139 @@
...
+            throw new NotSupportedException('Invalid name.');
+        }
+    }
+
+    /**
+     * @param mixed $value
+     */
+    final public static function get($value) : self
+    {
+        self::validateValue($value);
```
morozov: Looks like this line does the same as `array_search()` below. Can you instead of calling the same lookup twice just throw an exception if the `$name` is not found?

lib/Doctrine/Enumeration/Enumerated.php
```diff
@@ -0,0 +1,139 @@
...
+     * @param mixed $value
+     */
+    final public static function validateValue($value) : void
+    {
+        if (! in_array($value, self::availableValues(), true)) {
+            throw new NotSupportedException('Invalid value.');
+        }
+    }
+
+    final public static function validateName(string $name) : void
```
morozov: This seems to be not called from anywhere.

lib/Doctrine/Enumeration/Enumerated.php
morozov: Should it be `$name` and `validateName()` instead of `$value` and `validateValue()`? We're getting elements by name, not by value.

lib/Doctrine/Enumeration/Enumerated.php
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * @param mixed $value
+     */
+    final public static function get($value) : self
+    {
+        self::validateValue($value);
+
+        $name      = array_search($value, self::availableValues(), true);
+        $instances = &self::registry(static::class);
```
morozov: Maybe instead of using a method which returns a variable by reference you could use the same static class variable directly to save method calls and avoid the code like `if (!isset($registry[$type]))` in `registry()`:
```php
return self::$instances[static::class][$name] ?? self::$instances[static::class][$name] = new static();
```

lib/Doctrine/Enumeration/Enumerated.php
```diff
@@ -0,0 +1,139 @@
...
+
+        return $instances[$name] ?? $instances[$name] = new static();
+    }
+
+    /**
+     * @return mixed
+     */
+    final private static function value(self $instance)
+    {
+        return self::availableValues()[array_search($instance, self::registry(static::class), true)];
```
morozov: Maybe instead of `array_search()` you could use either `spl_object_hash()` or `SplObjectStorage()`. However, I'm not sure if it'd be more efficient given that usually, the array will contain a handful of values.

lib/Doctrine/DBAL/Cache/ArrayStatement.php
```diff
@@ -81,28 +83,28 @@ public function getIterator()
...
-    public function fetch($fetchMode = null, ...$args)
+    public function fetch(?FetchMode $fetchMode = null, ...$args)
     {
         if (! isset($this->data[$this->num])) {
             return false;
         }
 
         $row       = $this->data[$this->num++];
-        $fetchMode = $fetchMode ?: $this->defaultFetchMode;
+        $fetchMode = $fetchMode ?? $this->defaultFetchMode;
```
Taluu: what if the default is `null` ? Shouldn't it be `MIXED()` as before ?

Taluu: nevermind didn't see the default instantiation in the construct. Carry on.

lib/Doctrine/Enumeration/Enumerated.php
```diff
@@ -0,0 +1,139 @@
...
+    {
+        static $values = [];
+
+        if (isset($values[static::class])) {
+            return $values[static::class];
+        }
+
+        $values[static::class] = [];
+
+        foreach ((new \ReflectionClass(static::class))->getReflectionConstants() as $name => $constantReflection) {
```
morozov: That looks like uber magic. I thought the limited number of existing objects will be enforced by the fact that the class creates its own instances and stores them indexed by name. Not like this.

lib/Doctrine/Enumeration/Enumerated.php
```diff
@@ -0,0 +1,139 @@
...
+     */
+    final public function __invoke()
+    {
+        return self::value($this);
+    }
+
+    /**
+     * @param mixed[] $args
+     */
+    final public static function __callStatic(string $name, array $args) : self
```
Taluu: how about unpacked args (`,,,$args`) ? Would feel more natural imo.

morozov (commented):
I removed my review comment. I sounded like a dick because was too tired.

lib/Doctrine/Enumeration/Enumerated.php
Taluu: Didn't see the assert just below. But I still think it would be okayish for `Foo::BAR('args')` even if it does feel a bit away from enums. So... `¯\_(ツ)_/¯`

Taluu (approved):
Would be awesome to have this as a separate package (e.g `doctrine/enum`)

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -361,63 +362,63 @@ public function getIterator()
...
         }
 
-        if ($fetchMode === FetchMode::STANDARD_OBJECT) {
+        if ($fetchMode === FetchMode::STANDARD_OBJECT()) {
             return oci_fetch_object($this->_sth);
         }
 
-        if (! isset(self::$fetchModeMap[$fetchMode])) {
-            throw new \InvalidArgumentException("Invalid fetch style: " . $fetchMode);
+        if (! isset(self::$fetchModeMap[$fetchMode()])) {
```
Ocramius: I don't think this will work as expected

lib/Doctrine/Enumeration/Enumerated.php
Ocramius: 👍 cached instance should be the happy path

lib/Doctrine/Enumeration/Enumerated.php
```diff
@@ -0,0 +1,139 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\Enumeration;
+
+use Doctrine\Enumeration\Exception\NotSupportedException;
+use function array_search;
+use function constant;
+use function in_array;
+
+trait Enumerated
```
Ocramius: Please mark this as `@internal` - we don't want userland to know/use this

lib/Doctrine/Enumeration/Enumerated.php
Ocramius: This is actually the correct way to do it, although we should probably override `availableValues()` in userland where speed may be a problem (usually not though, since this is only ever executed once.

lib/Doctrine/Enumeration/Enumerated.php
```diff
@@ -0,0 +1,139 @@
...
+        static $values = [];
+
+        if (isset($values[static::class])) {
+            return $values[static::class];
+        }
+
+        $values[static::class] = [];
+
+        foreach ((new \ReflectionClass(static::class))->getReflectionConstants() as $name => $constantReflection) {
+            if (! $constantReflection->isPublic()) {
```
Ocramius: Why only `public` constants?

lib/Doctrine/Enumeration/Enumerated.php
Ocramius: Is this even ever needed/used? Can we kill it? I'd rather not have people call random static methods or skip proper declaration and documentation.

Ocramius (changes requested):
Overall, this looks like a GREAT improvement.

So far, we didn't have any way to properly enforce types of enumerable values, so we relied on `const`, but this is quite a lot better.

The problems here are with migration (lots of replacements to do, although easy) and with performance (unsure how usage of the method affects performance here)

lib/Doctrine/Enumeration/Enumerated.php
```diff
@@ -0,0 +1,139 @@
...
+    {
+        if (! in_array($value, self::availableValues(), true)) {
+            throw new NotSupportedException('Invalid value.');
+        }
+    }
+
+    final public static function validateName(string $name) : void
+    {
+        if (! isset(self::availableValues()[$name])) {
+            throw new NotSupportedException('Invalid name.');
```
greg0ire: A named constructor with `$name` as argument would be nice :P

Ocramius: That's `get()`

lib/Doctrine/Enumeration/Enumerated.php
greg0ire: Looks like we are getting them by value: `return self::get(self::SERIALIZABLE);`. Might be an issue if the values are not unique.

Ocramius: Values are always unique

lib/Doctrine/Enumeration/Exception/NotSupportedException.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Enumeration\Exception;
+
+final class NotSupportedException extends \LogicException implements \Throwable
```
greg0ire: IMO there should be at least another exception if not 2 for invalid values / names, because those will be caught a lot, and should be distinguishable from using a magic method that is not supported.

lib/Doctrine/Enumeration/Enumerated.php
greg0ire: You hope :P

Ocramius: We could also rely on the constant names, rather than their values...

lib/Doctrine/Enumeration/Enumerated.php
morozov: Could we take a simpler approach without reflection and array lookups?
```php
class Enum
{
    private static $instances;

    private function __construct()
    {
    }

    public static function X() : self
    {
        return self::value(__FUNCTION__);
    }

    public static function Y() : self
    {
        return self::value(__FUNCTION__);
    }

    private static function value(string $name) : self
    {
        return self::$instances[$name] ?? self::$instances[$name] = new self();
    }
}

var_dump(Enum::X() === Enum::X());
// true

var_dump(Enum::Y() === Enum::Y());
// true

var_dump(Enum::X() === Enum::Y());
// false
```

lib/Doctrine/Enumeration/Enumerated.php
Majkl578: > We're getting elements by name, not by value.

It's by value. Doing that by name would kill the ability to use constants or compare the instance against constant easily.

lib/Doctrine/Enumeration/Enumerated.php
Majkl578: The only purpose of the `&registry()` method is to hide the instances from reflection. Of course without it it's easier and likely faster, but allows anyone mess with those.

lib/Doctrine/Enumeration/Enumerated.php
Majkl578: Your proposal is no longer by value and changes the semantics. Also it wouldn't be possible to use `$fooBar() === $foo::BAR`.

lib/Doctrine/Enumeration/Enumerated.php
Majkl578: Why would you use any argument at first place?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
Majkl578: I think it will - `()` is an accessor for the value the enum holds. And since PHP arrays don't allow maps from objects to mixed, it's for convenience - yes, it could be SplObjectStorage<SomeEnum, mixed> as well.

lib/Doctrine/Enumeration/Enumerated.php
Majkl578: This method is meant as public API.

lib/Doctrine/DBAL/ColumnCase.php
Majkl578: Explained below.

lib/Doctrine/Enumeration/Enumerated.php
Majkl578: 1) To allow possible private constants i.e. for maps and behavior methods, i.e.:
```php
public const FOO = 1;
public const BAR = 2;
private const DESCRIPTIONS = [
    self::FOO => 'Hello',
    self::BAR => 'World',
];

...

public function getDescription() : string
{
    return self::DESCRIPTIONS[$this()];
}
```

2) To allow `$fooBar() === FOO::BAR` in consumer code (perf?).

lib/Doctrine/Enumeration/Enumerated.php
Majkl578: Yeah it's mostly for convenience/laziness. Not really useful for DBAL due to perf, but okayish for other consumers (if we make this available somewhere as non-internal behavior).

lib/Doctrine/Enumeration/Enumerated.php
morozov: If someone cannot hack something with reflection, they still can use stream filter applied to the class loader. There's no technical way to prevent people from being able to shoot themselves in the foot.

lib/Doctrine/Enumeration/Enumerated.php
morozov: Why expose and even constants? Doesn't it break encapsulation in the way that we want to operate objects, so their associated scalar values are just implementation details.

Why cannot the caller do `$fooBar() === Foo::BAR()` instead?

morozov: If it's because of BC, I think we need to decide first what the proper usage should look like and then build the compatibility layer. Mixing things together looks confusing from the design standpoint (unclear what's by design and what's temporary BC layer).

Majkl578: BC is only one part. It's also because the code with constants is a bit more comprehensible. Without those, everything is anonymous and you need magic like `__FUNCTION__` or so. Also instance can't be used in arrays - that is often useful for maps (i.e. ParameterType mappings in drivers).

Majkl578: I'm not saying it's not possible or that it's a bad idea. But in today's PHP I think using constants gives more flexibility, especially allows easily building maps of enums (their values) to mapped values by simply using the constant and then dereferencing the enum using `()`.

tomasfejfar:
Just out of curiosity: Why reinvent the wheel and not use `consistence`?

Majkl578:
> Just out of curiosity: Why reinvent the wheel and not use consistence?

Personally? Bloated API, PHP 7.0 code, not stateless instances, not a standalone package.

lib/Doctrine/Enumeration/Enumerated.php
morozov: I feel lost. Or, to put it better, I understand the reasoning but I find it confusing that plain constants get replaced with a relatively complex concept to be able to use object and type-hint against their classes, but still, we want to use constants as well.

The fact that the same thing can be referred to by multiple means, and the connection between them is hidden somewhere in the framework, looks similar to the magic which people hate about frameworks like Laravel. This approach may be usable from the consumer's standpoint once you get the concept — like Laravel's facades (which I never used) — but it just doesn't feel straightforward.

Majkl578: Ok, consider we have this currently (in OCI8Statement):
```php
    protected static $fetchModeMap = [
        FetchMode::MIXED       => OCI_BOTH,
        FetchMode::ASSOCIATIVE => OCI_ASSOC,
        FetchMode::NUMERIC     => OCI_NUM,
        FetchMode::COLUMN      => OCI_NUM,
    ];
```

How would you propose to rewrite this with the completely anonymous approach?
This is invalid:
```php
    protected static $fetchModeMap = [
        FetchMode::MIXED()       => OCI_BOTH,
        FetchMode::ASSOCIATIVE() => OCI_ASSOC,
        FetchMode::NUMERIC()     => OCI_NUM,
        FetchMode::COLUMN()      => OCI_NUM,
    ];
```

You would need to move this into a private method, build a SplObjectStorage map in constructor or something like that.

morozov: To me, a replacement of the map with a block of `if`/`else` would be more acceptable.

lib/Doctrine/DBAL/Platforms/DateIntervalUnit.php
```diff
@@ -22,7 +26,43 @@
...
+    {
+        return self::get(self::MONTH);
+    }
+
+    public static function QUARTER() : self
+    {
+        return self::get(self::QUARTER);
+    }
+
+    public static function YEAR() : self
```
Taluu: why define all these accessors, when the `__callStatic` should take over ?

Majkl578: Performance + static analysis.

lib/Doctrine/Enumeration/Enumerated.php
Majkl578: Ok, I will think about it. 👍
@ocramius WDYT?

lib/Doctrine/Enumeration/Enumerated.php
prolic: Maybe have a look at https://github.com/marc-mabe/php-enum.

lib/Doctrine/Enumeration/Enumerated.php
Ocramius: The reflection approach is good enough for me. It is an internal detail that is an `O(1)`, and no I/O is involved, so there is no noticeable overhead. Being able to override the "default magicness" is as easy as importing the method from the trait with a different name

mvrhov:
@Majkl578 Doesn't Elao/PhpEnums do exactly what's proposed here?

Ocramius:
@mvrhov lots of packages possibly do this: we just wanted to go with something internal to explore first, and may depend on something else if it is reliable and stable enough (inheritance is a bad beast)

prolic:
@mvrhov @Ocramius @Majkl578 https://github.com/marc-mabe/php-enum/ is the best library available. Bulletproof and battle-tested.

Ocramius:
@prolic absolutely no-go: way too much inherited `public` API.

Majkl578:
Please stop suggesting existing alternatives, at least at this moment. This is not about _finding the best among the best_, but about a _proof of concept_ solution.

https://github.com/marc-mabe/php-enum/ is not acceptable for me: PHP 5.6, extends (this is probably the worst problem around all existing libraries), bloated public API

lib/Doctrine/Enumeration/Enumerated.php
greg0ire: Say wat?

Ocramius: My mistake: saw the comment on a completely unrelated line.

VasekPurchart:
Hello, author of the `Consistence` lib mentioned as inspiration, just a few notes on the implementation, which I hope might help/clarify:

> traits

I considered using traits in the beginning, because I was afraid of using inheritance. If I remember correctly the final reason why I went with it was that there will always be some "infrastructure" code, which will need to work with all enums. This could be for example a DBAL type serializing the value etc. That means that all of them would have to be of same type, meaning if the actual logic would be in a trait, then you would have to also always not forget to add the implements interface yourself, which was something easily forgotten and when I asked people around was that in the end the inheritance is suitable for this one more and the limitations it brings can be solved by additional traits instead.

> each constant is accessible as a method (either explicit or magic)

Being able to construct enums like this is indeed very comfortable but using magic methods for these (or even offering to do so) is i think bad practice, because you shy away from the best parts of enums - bringing more typed code.

This leaves only explicit methods then, which would have to be either written manually - uncomfortable, error prone (keeping up with the values), or generating them, which needs a lot more infrastructure and has dependencies on global states etc.

That is why this was in the end not part of the library itself, but for example in Slevomat I think they are still using this with generated methods.

PR #3012, opened 2018-02-05
fix if statement when check for an array without elements

AlessandroMinoccheri:
Fix and if statement when check for an array without elements into dbal exception instead of compare It with a boolean

### DIFF
--- a/lib/Doctrine/DBAL/DBALException.php
+++ b/lib/Doctrine/DBAL/DBALException.php
@@ -142,7 +142,7 @@ class DBALException extends \Exception
     public static function driverExceptionDuringQuery(Driver $driver, \Exception $driverEx, $sql, array $params = [])
     {
         $msg = "An exception occurred while executing '".$sql."'";
-        if ($params) {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/DBALException.php
```diff
@@ -142,7 +142,7 @@ public static function unknownDriver($unknownDriverName, array $knownDrivers)
     public static function driverExceptionDuringQuery(Driver $driver, \Exception $driverEx, $sql, array $params = [])
     {
         $msg = "An exception occurred while executing '".$sql."'";
-        if ($params) {
+        if (! empty($params)) {
```
Ocramius: This was already correct, while `empty()` is an extremely risky construct that works with mixed types and can lead to confusion. Overall :-1: to this patch

Ocramius (changes requested):

AlessandroMinoccheri:
Ok I have understood your point @Ocramius 
Thanks for the explanation

PR #3013, opened 2018-02-05
Escape LIKE metacharacters

greg0ire:
There seems to be a lack of tools to fight against [wildcard attacks](https://www.owasp.org/index.php/Testing_for_SQL_Wildcard_Attacks_(OWASP-DS-001) ), or to simply provide a way to let your users search for strings that include metacharacters

# todo

- [x] add a functional test to make sure valid SQL expressions are built by means of the new API.
- [x] allow specifying an escape character in the expression builder `like` and `notLike` methods.
- [x] implement `getWildcardCharacters` for RDBMSs that have more than the default wildcard characters.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -3603,7 +3603,7 @@ abstract class AbstractPlatform
     /**
      * @return string[]
      */
-    protected function getLikeWildcardCharacters() : iterable
+    protected function getLikeWildcardCharacters() : array
     {
         return ['%', '_'];
     }
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -42,6 +42,9 @@ use Doctrine\DBAL\Schema\TableDiff;
 use Doctrine\DBAL\TransactionIsolationLevel;
 use Doctrine\DBAL\Types;
 use Doctrine\DBAL\Types\Type;
+use function sprintf;
+use function strlen;
+use function strtr;
 
 /**
  * Base class for all DatabasePlatforms. The DatabasePlatforms are the central
@@ -3577,5 +3580,28 @@ abstract class AbstractPlatform
     public function getStringLiteralQuoteCharacter()
     {
         return "'";
+    }
+
+    /**
+     * Escapes metacharacters in a string intended to be used with a LIKE
+     * operator.
+     *
+     * @param string $inputString a literal, unquoted string
+     * @param string $escapeChar  should be reused by the caller in the LIKE
+     *                            expression.
+     */
+    final public function escapeStringForLike(string $inputString, string $escapeChar) : string
+    {
+        $replacePairs = [$escapeChar => $escapeChar.$escapeChar];
+        foreach ($this->getLikeWildcardCharacters() as $wildcardChar) {
+            $replacePairs[$wildcardChar] = $escapeChar.$wildcardChar;
+        }
+
+        return strtr($inputString, $replacePairs);
+    }
+
+    protected function getLikeWildcardCharacters() : iterable
+    {
+        return ['%', '_'];
     }
 }
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -43,7 +43,6 @@ use Doctrine\DBAL\TransactionIsolationLevel;
 use Doctrine\DBAL\Types;
 use Doctrine\DBAL\Types\Type;
 use function addcslashes;
-use function implode;
 use function preg_quote;
 use function preg_replace;
 use function sprintf;
@@ -3596,17 +3595,14 @@ abstract class AbstractPlatform
     final public function escapeStringForLike(string $inputString, string $escapeChar) : string
     {
         return preg_replace(
-            '~([' . preg_quote(implode('', $this->getLikeWildcardCharacters()) . $escapeChar, '~') . '])~u',
+            '~([' . preg_quote($this->getLikeWildcardCharacters() . $escapeChar, '~') . '])~u',
             addcslashes($escapeChar, '\\') . '$1',
             $inputString
         );
     }
 
-    /**
-     * @return string[]
-     */
-    protected function getLikeWildcardCharacters() : array
+    protected function getLikeWildcardCharacters() : string
     {
-        return ['%', '_'];
+        return '%_';
     }
 }
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -42,9 +42,12 @@ use Doctrine\DBAL\Schema\TableDiff;
 use Doctrine\DBAL\TransactionIsolationLevel;
 use Doctrine\DBAL\Types;
 use Doctrine\DBAL\Types\Type;
+use function addcslashes;
+use function implode;
+use function preg_quote;
+use function preg_replace;
 use function sprintf;
 use function strlen;
-use function strtr;
 
 /**
  * Base class for all DatabasePlatforms. The DatabasePlatforms are the central
@@ -3592,12 +3595,11 @@ abstract class AbstractPlatform
      */
     final public function escapeStringForLike(string $inputString, string $escapeChar) : string
     {
-        $replacePairs = [$escapeChar => $escapeChar . $escapeChar];
-        foreach ($this->getLikeWildcardCharacters() as $wildcardChar) {
-            $replacePairs[$wildcardChar] = $escapeChar . $wildcardChar;
-        }
-
-        return strtr($inputString, $replacePairs);
+        return preg_replace(
+            '~([' . preg_quote(implode('', $this->getLikeWildcardCharacters()) . $escapeChar, '~') . '])~u',
+            addcslashes($escapeChar, '\\') . '$1',
+            $inputString
+        );
     }
 
     /**
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -3592,14 +3592,17 @@ abstract class AbstractPlatform
      */
     final public function escapeStringForLike(string $inputString, string $escapeChar) : string
     {
-        $replacePairs = [$escapeChar => $escapeChar.$escapeChar];
+        $replacePairs = [$escapeChar => $escapeChar . $escapeChar];
         foreach ($this->getLikeWildcardCharacters() as $wildcardChar) {
-            $replacePairs[$wildcardChar] = $escapeChar.$wildcardChar;
+            $replacePairs[$wildcardChar] = $escapeChar . $wildcardChar;
         }
 
         return strtr($inputString, $replacePairs);
     }
 
+    /**
+     * @return string[]
+     */
     protected function getLikeWildcardCharacters() : iterable
     {
         return ['%', '_'];
--- a/lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php
@@ -120,7 +120,7 @@ class SQLServer2008Platform extends SQLServer2005Platform
     /**
      * {@inheritDoc}
      */
-    protected function getLikeWildcardCharacters() : iterable
+    protected function getLikeWildcardCharacters() : array
     {
         return array_merge(parent::getLikeWildcardCharacters(), ['[', ']', '^']);
     }
--- a/lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php
@@ -117,11 +117,8 @@ class SQLServer2008Platform extends SQLServer2005Platform
         return Keywords\SQLServer2008Keywords::class;
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    protected function getLikeWildcardCharacters() : array
+    protected function getLikeWildcardCharacters() : string
     {
-        return array_merge(parent::getLikeWildcardCharacters(), ['[', ']', '^']);
+        return parent::getLikeWildcardCharacters() . '[]^';
     }
 }
--- a/lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php
@@ -117,6 +117,9 @@ class SQLServer2008Platform extends SQLServer2005Platform
         return Keywords\SQLServer2008Keywords::class;
     }
 
+    /**
+     * {@inheritDoc}
+     */
     protected function getLikeWildcardCharacters() : iterable
     {
         return array_merge(parent::getLikeWildcardCharacters(), ['[', ']', '^']);
--- a/lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php
@@ -116,4 +116,9 @@ class SQLServer2008Platform extends SQLServer2005Platform
     {
         return Keywords\SQLServer2008Keywords::class;
     }
+
+    protected function getLikeWildcardCharacters() : iterable
+    {
+        return array_merge(parent::getLikeWildcardCharacters(), ['[', ']', '^']);
+    }
 }
--- a/lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
+++ b/lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
@@ -257,7 +257,7 @@ class ExpressionBuilder
      *
      * @return string
      */
-    public function like($x, $y)
+    public function like($x, $y/*, ?string $escapeChar = null */)
     {
         return $this->comparison($x, 'LIKE', $y) .
             (func_num_args() >= 3 ? sprintf(" ESCAPE '%s'", func_get_arg(2)) : '');
@@ -271,7 +271,7 @@ class ExpressionBuilder
      *
      * @return string
      */
-    public function notLike($x, $y)
+    public function notLike($x, $y/*, ?string $escapeChar = null */)
     {
         return $this->comparison($x, 'NOT LIKE', $y) .
             (func_num_args() >= 3 ? sprintf(" ESCAPE '%s'", func_get_arg(2)) : '');
--- a/lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
+++ b/lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
@@ -20,7 +20,7 @@
 namespace Doctrine\DBAL\Query\Expression;
 
 use Doctrine\DBAL\Connection;
-use function func_get_args;
+use function func_get_arg;
 use function func_num_args;
 use function sprintf;
 
@@ -260,7 +260,7 @@ class ExpressionBuilder
     public function like($x, $y)
     {
         return $this->comparison($x, 'LIKE', $y) .
-            (func_num_args() >= 3 ? sprintf(" ESCAPE '%s'", func_get_args()[2]) : '');
+            (func_num_args() >= 3 ? sprintf(" ESCAPE '%s'", func_get_arg(2)) : '');
     }
 
     /**
@@ -274,7 +274,7 @@ class ExpressionBuilder
     public function notLike($x, $y)
     {
         return $this->comparison($x, 'NOT LIKE', $y) .
-            (func_num_args() >= 3 ? sprintf(" ESCAPE '%s'", func_get_args()[2]) : '');
+            (func_num_args() >= 3 ? sprintf(" ESCAPE '%s'", func_get_arg(2)) : '');
     }
 
     /**
--- a/lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
+++ b/lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
@@ -20,6 +20,9 @@
 namespace Doctrine\DBAL\Query\Expression;
 
 use Doctrine\DBAL\Connection;
+use function func_get_args;
+use function func_num_args;
+use function sprintf;
 
 /**
  * ExpressionBuilder class is responsible to dynamically create SQL query parts.
@@ -256,7 +259,8 @@ class ExpressionBuilder
      */
     public function like($x, $y)
     {
-        return $this->comparison($x, 'LIKE', $y);
+        return $this->comparison($x, 'LIKE', $y) .
+            (func_num_args() >= 3 ? sprintf(" ESCAPE '%s'", func_get_args()[2]) : '');
     }
 
     /**
@@ -269,7 +273,8 @@ class ExpressionBuilder
      */
     public function notLike($x, $y)
     {
-        return $this->comparison($x, 'NOT LIKE', $y);
+        return $this->comparison($x, 'NOT LIKE', $y) .
+            (func_num_args() >= 3 ? sprintf(" ESCAPE '%s'", func_get_args()[2]) : '');
     }
 
     /**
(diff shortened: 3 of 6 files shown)


lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3578,4 +3578,22 @@ public function getStringLiteralQuoteCharacter()
...
+     */
+    public function escapeStringForLike(string $untrustedString): string
+    {
+        $replacePairs = [];
+        $escapeChar = $this->getLikeEscapeChar();
+        foreach ($this->getLikeMetaCharacters() as $metaChar) {
+            $replacePairs[$metaCharacter] = $escapeChar.$metaChar;
+        }
+
+        return strtr($untrusted, $replacePairs);
```
Taluu: `$untrustedString` ?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3578,4 +3578,22 @@ public function getStringLiteralQuoteCharacter()
...
+    }
+
+    abstract protected function getLikeEscapeChar();
+    abstract protected function getLikeMetaCharacters();
```
Taluu: `:string` and `: iterable` ?

greg0ire:
Please tell me if you're fine with the feature, the naming, and I should add implementations, docs and tests.

morozov:
@greg0ire I have a feeling that in most supported DB platforms, escaping wildcard characters works almost the same.

Instead of introducing new `abstract` methods which is a breaking change, can you introduce default implementations for them?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3578,4 +3578,22 @@ public function getStringLiteralQuoteCharacter()
...
         return "'";
     }
+
+    /**
+     * Escapes metacharacters in a string intended to be used with a LIKE
+     * operator.
+     */
+    public function escapeStringForLike(string $untrustedString): string
+    {
+        $replacePairs = [];
```
morozov: The escape character itself should also be always escaped. Does this approach account for `getLikeMetaCharacters()` always returning `getLikeEscapeChar()` or it `escapeStringForLike()` should handle that itself?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3578,4 +3578,22 @@ public function getStringLiteralQuoteCharacter()
     {
         return "'";
     }
+
+    /**
+     * Escapes metacharacters in a string intended to be used with a LIKE
+     * operator.
+     */
+    public function escapeStringForLike(string $untrustedString): string
```
morozov: This is not about trust. The input is an arbitrary string with no meaning while the output is a LIKE expression.

greg0ire: It could indeed come from a trusted source and still need to be escaped because you genuinely want to search for that 👍

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
greg0ire: I think the best is to let `escapeStringForLike` handle it, and I could rename "meta" to "wildcard", what do you think?

greg0ire:
Funny thing: In MS SQL, the escape character has to be picked by the user and has no default: https://docs.microsoft.com/en-us/sql/t-sql/language-elements/like-transact-sql

morozov:
> Funny thing: In MS SQL, the escape character has to be picked by the user and has no default: https://docs.microsoft.com/en-us/sql/t-sql/language-elements/like-transact-sql

In my personal experience, it defaults to `!` same as in other databases.

> -- Syntax for SQL Server and Azure SQL Database  
> 
> match_expression [ NOT ] LIKE pattern [ ESCAPE escape_character ]  

Why is then the `ESCAPE` part optional?

greg0ire:
> Why is then the ESCAPE part optional?

Probably because you don't always need to escape characters

greg0ire:
For PG, you can specify an escape char too, but it defaults to backslash

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: Agree on both.

morozov:
Also IIRC, in IBM DB2 there's also a different default escape character. Probably, we need to gather escaping examples for all DBs and then see if the `ESCAPE` part can be avoided by using the default escape character for the current platform. Otherwise, it may turn out that just escaping the string may be insufficient and the API will need to look somehow differently.

greg0ire:
MySQL sounds like a lot of fun:

```
 Because MySQL uses C escape syntax in strings (for example, \n to represent a newline character), you must double any \ that you use in LIKE strings. For example, to search for \n, specify it as \\n. To search for \, specify it as \\\\; this is because the backslashes are stripped once by the parser and again when the pattern match is made, leaving a single backslash to be matched against.

Exception: At the end of the pattern string, backslash can be specified as \\. At the end of the string, backslash stands for itself because there is nothing following to escape. Suppose that a table contains the following values: 
```

greg0ire:
> Also IIRC, in IBM DB2 there's also a different default escape character. Probably, we need to gather escaping examples for all DBs and then see if the ESCAPE part can be avoided by using the default escape character for the current platform. Otherwise, it may turn out that just escaping the string may be insufficient and the API will need to look somehow differently.

Well the only which seems problematic to me is MS SQL, because it seems you have to append `ESCAPE somechar` and escape with that char to get it to work. A solution could be to pick some random char and always use `ESCAPE` no matter what but that would imply all systems have that escape char specification capability.

greg0ire:
Ok so let's start the gathering

| Platform   | Escape sequence  |
|------------|------------------|
| PostgreSQL | \                |
| MySQL      | \                |
| MS SQL     | none :trollface: |

greg0ire:
I think an optional `$escapeChar` argument could be added to `escapeStringForLike`, and let the user be responsible for specifying it again when provided whatever they want to the `like()` method of the expression builder.

morozov:
> That's right, I think an optional `$escapeChar` argument could be added to `escapeStringForLike`

It should be either mandatory or not configurable at all and accessible in a read-only mode (e.g. `const ESCAPE_CHAR = '!'`). 

If mandatory, the caller will always manually pass the character of choice to the escaping function and add it to `ESCAPE`. If read-only, the caller will always take its value and also add it to `ESCAPE`.

Otherwise, the caller doesn't know if it's needed for a given string and platform and what to pass to `ESCAPE` if needed and if it's needed.

greg0ire:
I think the mandatory parameter might be best: it offers the most control, and it acts as a better reminder since people not knowing what to specify will read the phpdoc, where I can document that they need to also add `ESCAPE`

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3578,4 +3581,37 @@ public function getStringLiteralQuoteCharacter()
...
+     *
+     * @param string $inputString a literal, unquoted string
+     * @param string $escapeChar  should be exactly one character long and
+     *                            should be reused by the caller in the LIKE
+     *                            expression.
+     */
+    final public function escapeStringForLike(string $inputString, string $escapeChar) : string
+    {
+        /* Some RDBMS' can deal with multibyte characters, but let us assume
+         * that kind of will not actually be needed */
```
greg0ire: Tell me if I should allow them instead, or even remove that check and let the RDBMS deal with that.

morozov: I'd say the DB should decide what's valid and what's not. Even if it's known that it should be one character for all supported platforms, there may exist other platform-specific limitations which we'll not be able to cover. Let's not even try.

morozov:
I'd also add a functional test to make sure valid SQL expressions are built by means of the new API.

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
```diff
@@ -1468,4 +1468,12 @@ public function getGeneratesFloatDeclarationSQL()
             array(array('precision' => 8, 'scale' => 2), 'DOUBLE PRECISION'),
         );
     }
+
+    public function testItEscapesStringsForLikeProperly() : void
+    {
+        self::assertSame(
+            '25\% off \\\\o/',
```
morozov: Could you add the underscore here as well?

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
```diff
@@ -1468,4 +1468,12 @@ public function getGeneratesFloatDeclarationSQL()
             array(array('precision' => 8, 'scale' => 2), 'DOUBLE PRECISION'),
         );
     }
+
+    public function testItEscapesStringsForLikeProperly() : void
```
morozov: The "properly" part seems redundant. It's implied that the test will only pass if the API is implemented properly.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3578,4 +3581,28 @@ public function getStringLiteralQuoteCharacter()
     {
         return "'";
     }
+
+    /**
+     * Escapes metacharacters in a string intended to be used with a LIKE
+     * operator.
+     *
+     * @param string $inputString a literal, unquoted string
+     * @param string $escapeChar  should be exactly one character long and
```
morozov: If it's not enforced by the API, I'd omit the "should be exactly one character long" part. It's still a DB abstraction layer where all DB limitations are implied but don't necessarily have to be called out. The rest is valid.

morozov:
Not sure if it's outside of the scope but maybe the `QueryBuilder` should be able to build the `[NOT] LIKE ? ESCAPE ?` expression to encapsulate the SQL syntax. In that case, the client-side code might look like:
```php
$pattern = $platform->escapeStringForLike($input, '!') . '%';
$builder->like($pattern, '!');
```
The builder, in turn, may or may not append the `ESCAPE` part the expression depending on whether the escape character or wildcard characters are found in the pattern (for sake of compactness).

Looks a bit complicated but I can't see a simpler approach.

greg0ire:
You mean the expression builder, right?

https://github.com/doctrine/dbal/blob/de568088079ad76ccee6c1ac9d1a04c45d4cef9f/lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php#L257-L273

I can't find a `like` method in the query builder.

I'd be 👍 to change the expression builder too, which will mean adding a new argument. Since it isn't final, adding an argument will be a BC-break though, right? Even if it is optional. Shall I rely on `func_get_args()` to achieve this? Or is a BC-break ok on that branch?

morozov:
@greg0ire sorry, I didn't bother to look that those methods already exist in the builder.

You're right. Given it'd be a breaking change, it can only be done in 3.0 or via `func_get_args()`. Let's try the latter and see if the entire effort is accepted by the board. Then we can clean it up in `develop`.

greg0ire:
@morozov what should I do if the user attempts to use a single quote as escape character? Should I throw, or should I attempt to produce the following:

For PG: `SELECT 'test' LIKE 'test' ESCAPE E'\''`
For MySQL: `SELECT 'test' LIKE 'test' ESCAPE '\''`

morozov:
@greg0ire I believe it's not the responsibility of expression builder to escape literals. All arguments it takes are SQL expression, not values. Expressions can be object names (e.g. `user.id`), literals (e.g. `'fo\'o'`) and placeholders (`?`). In all cases, it's the responsibility of the caller to pass valid expressions. Expressions builder just combines them together.

Not sure when the PG syntax with leading `E` means but I believe it should be possible to pass the escape character as a bound parameter to the query same as other parameters. I only checked it on MySQL:
```sql
SELECT name FROM accounts WHERE name LIKE ? ESCAPE ?
```

greg0ire:
The `E` thingy is described here: https://www.postgresql.org/docs/8.3/static/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS

And indeed, the user will be able to do what they want thanks to a bound parameter 👍

lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php
```diff
@@ -116,4 +116,9 @@ protected function getReservedKeywordsClass()
     {
         return Keywords\SQLServer2008Keywords::class;
     }
+
+    protected function getLikeWildcardCharacters() : iterable
+    {
+        return array_merge(parent::getLikeWildcardCharacters(), ['[', ']', '^']);
```
morozov: `array_merge()` will not work with an arbitrary iterable. Either use `array` as the return type or use `yield from` to merge the sets.

tests/Doctrine/Tests/DBAL/Functional/LikeWildcardsEscapingTest.php
```diff
@@ -0,0 +1,22 @@
...
+        $string = '_25% off_ your next purchase \o/';
+        $escapeChar = '!';
+        $stmt = $this->_conn->prepare(sprintf(
+            "SELECT '%s' LIKE '%s' ESCAPE '%s' as it_matches",
+            $string,
+            $this->_conn->getDatabasePlatform()->escapeStringForLike($string, $escapeChar),
+            $escapeChar
+        ));
+        $stmt->execute();
+        $this->assertTrue((bool) $stmt->fetch()['it_matches']);
```
morozov: You can use `$stmt->fetchColumn()` and avoid using the column alias.

lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php
greg0ire: I'll use `array`, can't see a real need for iterable here, sorry @Taluu :P

greg0ire:
Triggering continuousphp

lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
```diff
@@ -256,7 +259,8 @@ public function isNotNull($x)
      */
     public function like($x, $y)
     {
-        return $this->comparison($x, 'LIKE', $y);
+        return $this->comparison($x, 'LIKE', $y) .
+            (func_num_args() >= 3 ? sprintf(" ESCAPE '%s'", func_get_args()[2]) : '');
```
greg0ire: Should I add doc for that new, simulated parameter, or will that confuse CS tools? Should I also add something like  `/* ?$escapeChar = null */` in the signature?

morozov: Not sure what the generally used approach is. Let's try both the phpDoc parameter and commented out method argument, see how PHPCS reacts.

BTW, shouldn't `ESCAPE '%s'` be `ESCAPE %s`? Otherwise, the caller won't be able to bind the escape character a parameter. It's the platform's responsibility to quote literals, not the expression builder's.

morozov: Also, do you want to use `func_get_arg(2)` instead of `func_get_args()[2]`?

tests/Doctrine/Tests/DBAL/Functional/LikeWildcardsEscapingTest.php
```diff
@@ -0,0 +1,25 @@
...
+namespace Doctrine\Tests\DBAL\Functional;
+
+use Doctrine\DBAL\Platforms\OraclePlatform;
+use Doctrine\Tests\DbalFunctionalTestCase;
+
+final class LikeWildcardsEscapingTest extends DbalFunctionalTestCase
+{
+    public function testFetchLikeExpressionResult() : void
+    {
+        $string           = '_25% off_ your next purchase \o/';
```
morozov: Let's make sure all known wildcard characters are covered (see `SQLServer2008Platform`);

tests/Doctrine/Tests/DBAL/Functional/LikeWildcardsEscapingTest.php
```diff
@@ -0,0 +1,25 @@
...
+        $escapeChar       = '!';
+        $databasePlatform = $this->_conn->getDatabasePlatform();
+        $stmt             = $this->_conn->prepare(sprintf(
+            "SELECT (CASE WHEN '%s' LIKE '%s' ESCAPE '%s' THEN 1 ELSE 0 END)" .
+                ($databasePlatform instanceof OraclePlatform ? ' FROM dual' : ''),
```
morozov: Personally, I use `$platform->getDummySelectSQL()` and replace the `1` with the needed expression for better portability.

@Ocramius would it make sense to add an optional parameter to `getDummySelectSQL()` which would accept an arbitrary SQL expression and use it instead of `1`? This might come in handy.

lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
```diff
@@ -249,27 +252,31 @@ public function isNotNull($x)
     /**
      * Creates a LIKE() comparison expression with the given arguments.
      *
-     * @param string $x Field in string format to be inspected by LIKE() comparison.
-     * @param mixed  $y Argument to be used in LIKE() comparison.
+     * @param string $x          Field in string format to be inspected by LIKE() comparison.
+     * @param mixed  $y          Argument to be used in LIKE() comparison.
+     * @param string $escapeChar escape character. Should be used when escaping $y
```
greg0ire: @morozov phpcs does not like this

morozov (approved):

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3578,4 +3581,30 @@ public function getStringLiteralQuoteCharacter()
...
+
+        return strtr($inputString, $replacePairs);
+    }
+
+    /**
+     * @return string[]
+     */
+    protected function getLikeWildcardCharacters() : array
+    {
+        return ['%', '_'];
```
Majkl578: Does this really need to be an array? Why not a string?

greg0ire: Didn't think of it, but that would enforce the characters being characters

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3578,4 +3582,27 @@ public function getStringLiteralQuoteCharacter()
...
+     * operator.
+     *
+     * @param string $inputString a literal, unquoted string
+     * @param string $escapeChar  should be reused by the caller in the LIKE
+     *                            expression.
+     */
+    final public function escapeStringForLike(string $inputString, string $escapeChar) : string
+    {
+        $replacePairs = [$escapeChar => $escapeChar . $escapeChar];
+        foreach (str_split($this->getLikeWildcardCharacters()) as $wildcardChar) {
```
sstok: For RollerworksSearch I am using `addcslashes` is using str_split better for this?

greg0ire: `$escapeChar` is not necessarily a backslash here, so no, `addcslashes` won't do

Majkl578: `str_split` will not work with multibyte (Unicode) character, but would DB work with it?

greg0ire: Not all dbs work with it, pg does, mysql does not

greg0ire: I tried with 💩 the other day, no luck on mysql

sstok: [addcslashes](http://php.net/addcslashes) also accepts a charlist 👍

morozov: So, why do we need to smash all strings together and then decide how to split them back apart? What was wrong with the array?

Majkl578: Personally I would avoid foreach entirely and just used preg_replace. Would work with single and multibyte as well.

morozov: @Majkl578 you can build a regex from an array of inputs without having to split the string (I don't see a problem in using `foreach` or `array_map()` in this case).

It smells if you have to split/parse your own internal values.

greg0ire: > addcslashes also accepts a charlist 👍

@sstok while I'm really glad to read that wonderful piece of news, I think it still does not let you pick what character you want for escaping, does it?

> So, why do we need to smash all strings together and then decide how to split them back apart? What was wrong with the array?

I'm not sure. What's good is that we now are sure we will only get one character.

> Personally I would avoid foreach entirely and just used preg_replace. Would work with single and multibyte as well.

Should I add support for using 💩 even though all RDBMs do not support that?

I'm a bit undecided on the array vs string debate. Regarding `preg_replace`, I'm not sure it makes a lot of sense here, because `str_replace` would do the same job quite well (right?)

greg0ire: > Jokes aside, but why would anyone want to use a multibyte character for escaping.

can't see a reasonable usage

> Pracically, I don't see any difference between them. Unless performance counts which I doubt. Anyways, you'll have to build either the search and replace pieces, or the regex from the raw input of characters which I still prefer to be defined individually.

Yeah performance won't be an issue here IMO. Typical usage of this will be once or twice per http request. Reverting to arrays then

Majkl578: I still think whole froreach + array of chars + strtr (+ single char validation) should be replaced just by something like:
```php
preg_replace('~([' . preg_quote($this->getLikeWildcardCharacters(), '~') . '])~u', addcslashes($escapeChar, '\\') .'$1', $inputString)
```

morozov: @Majkl578 you can `implode()` the array of characters into a string like in your example.

greg0ire:
@Majkl578 @morozov I switched to preg + implode, please review again, and agree on something :P

Majkl578:
I don't really care which way it is, but `array` doesn't give any guarantees about `char` or even `string[]` rather than arbitrary `mixed[]`. I can technically do:
```
protected function getLikeWildcardCharacters() : array
{
    return [$this];
}
```
And there will be lots of dragons. 🐲 

That's why I proposed using simple string instead of array.

morozov:
Agree, let’s revert to the string. It makes more sense with the regexp approach.

greg0ire:
Agreement! Great, let's do this!

tests/Doctrine/Tests/DBAL/Query/Expression/ExpressionBuilderTest.php
```diff
@@ -219,4 +219,33 @@ public function testNotInWithPlaceholder()
...
+        self::assertEquals(
+            "s.last_words NOT LIKE 'this'",
+            $this->expr->notLike('s.last_words', "'this'")
+        );
+    }
+
+    public function testNotLikeWithEscape()
+    {
+        self::assertEquals(
+            "p.description NOT LIKE '20💩%' ESCAPE '💩'",
```
Ocramius: Classy 💩

Ocramius (approved):
LGTM 👍

tests/Doctrine/Tests/DBAL/Query/Expression/ExpressionBuilderTest.php
greg0ire: IKR?

Ocramius:
> Agree, let’s revert to the string. It makes more sense with the regexp approach.

Not needed: `string[]` is a sufficient documentation for subclasses/implementors. In addition to that, there is no guarantee that the wildcard is a single char (although I do really hope that's the case everywhere :-P)

Ocramius:
Oh well, I see it's done anyway. Waiting for CI, then 🚢

greg0ire:
Ready to 🚢

Ocramius:
👍

tests/Doctrine/Tests/DBAL/Functional/LikeWildcardsEscapingTest.php
morozov: Ping, @Ocramius.

PR #3017, opened 2018-02-09
Create date strings compatible to datetime and datetime2

fabiang:
Hello,

in SQLServer the format (dmy or ymd) of datetime values depends on the value of the variable `DATEFORMAT`. When you have a schema you can't manage, you can get wrong results when querying a table with a different `DATEFORMAT` set. It seems that SQLServer support ISO 8601-like values, which seems always to be correct on `datetime` and `datetime2`.

I suggest changing that DateTime objects are always turned into the ISO 8601-like strings, to make it work with both types.

I'm not sure if this is a bc break, but I guess this could introduce some cache-related issue to some end-users.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php
@@ -49,7 +49,7 @@ class SQLServer2008Platform extends SQLServer2005Platform
      */
     public function getDateTimeFormatString()
     {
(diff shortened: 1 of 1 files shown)


Ocramius:
The correct approach would be to tweak [these docs](http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/known-vendor-issues.html#id6) and provide a listener that adapts the format on setup

fabiang:
Closing this for now.

PR #3018, opened 2018-02-11
[CS] Apply Doctrine CS 3.0 in lib

carusogabriel:


Majkl578:
It has already been said that this is no-go as long as develop still exists. :/

carusogabriel:
@Marjkl578 Oh sorry, missed that!

PR #2984, opened 2018-01-21
Allow dynamic intervals in DATE_ADD for SQLite

fogs:
Fixing #2983

Issue #2983, opened 2018-01-21
DATE_ADD does not work with intervals from columns in SQLite


fogs:
Hi there!

We using DATE_ADD() with a dynamic interval parameter from a table column, such as 
`->andWhere($qb->expr()->gt("DATE_ADD(p.beginsOn, p.duration, 'month')", ':timestamp'))`

This works well in MySQL but not in SQLite; the latter produces
`AND DATE(b0_.c_beginsOn,'+b0_.c_duration MONTH') > ?)`

It seems to me like the only static intervals were anticipated; however SQLite can also handle dynamic ones.

morozov:
Fixed by https://github.com/doctrine/dbal/pull/3019.

PR #3019, opened 2018-02-11
Allow dynamic intervals in DATE_ADD & DATE_SUB for SQLite

fogs:
Fixing #2983
Replacing https://github.com/doctrine/dbal/pull/2984

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
@@ -150,6 +150,10 @@ class SqlitePlatform extends AbstractPlatform
                         break;
                 }
 
+                if (!is_numeric($interval)) {
+                    $interval = "' || " . $interval . " || '";
+                }
+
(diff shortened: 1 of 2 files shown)


fogs:
Hi @Ocramius,

This is the replacement for #2984 now including some unit tests. Let me know if that works for you or if you are missing something else.

The Scrutinizer came back with "Tests: passed" but is complaining about "Failure condition met". Is that something I need to fix and if so, how?

Our project is using doctrine/dbal in an older version. Will my patch be backported or can I do something to get it there? Would love to get rid of using my fork of doctrine in our project's composer.json.

Thanks!

Ocramius:
Heya! Please do ignore the Scrutinizer-CI failure here :-)

As for the backport: no, this is an addition, and it will land in DBAL 2.7

morozov:
Someone needs to approve this before it gets into `2.7.0`.

Ocramius (approved):

PR #3020, opened 2018-02-12
Improve handling of schemas in SQL Server >= 2008

stlrnz:
Table listings do not consider schema names for the SQL Server platform.
This fix adds the schema name to SQL Server >= 2008 table listingss using SQL Sever built-in function SCHEMA_NAME() applied to column uid.

Funtion reference: https://docs.microsoft.com/en-us/sql/t-sql/functions/schema-name-transact-sql
Column reference: https://docs.microsoft.com/en-us/sql/relational-databases/system-compatibility-views/sys-sysobjects-transact-sql

Without this fix doctrine will always try to create the whole schema, even if you only triggered an update. This is because the state of the current schema stored in database is not read correctly.

**Note:** This fix might also be necessary for SQL Server platforms older than 2008. Unfortunately, I have no possibility to test it.
If anyone wants to backport this fix, be aware that the built-in function SCHEMA_NAME is only available since SQL Server 2008. Maybe the function USER_NAME will help: https://social.msdn.microsoft.com/Forums/sqlserver/en-US/135d51e5-2c22-4ce9-a308-fbb958f3362e/how-to-get-schema-name-from-uid-column-in-dbosysobjects-in-sql-server-2000?forum=transactsql

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -28,8 +28,10 @@ use Doctrine\DBAL\Schema\Index;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Schema\TableDiff;
 use Doctrine\DBAL\Types;
+use function explode;
 use function implode;
 use function sprintf;
+use function strpos;
 
 /**
  * The SQLServerPlatform provides the behavior, features and SQL dialect of the
@@ -330,13 +332,22 @@ class SQLServerPlatform extends AbstractPlatform
      */
     protected function getCreateColumnCommentSQL($tableName, $columnName, $comment)
     {
+        if (strpos($tableName, '.') !== false) {
+            [$schemaSQL, $tableSQL] = explode('.', $tableName);
+            $schemaSQL              = $this->quoteStringLiteral($schemaSQL);
+            $tableSQL               = $this->quoteStringLiteral($tableSQL);
+        } else {
+            $schemaSQL = "'dbo'";
+            $tableSQL  = $this->quoteStringLiteral($tableName);
+        }
+
         return $this->getAddExtendedPropertySQL(
             'MS_Description',
             $comment,
             'SCHEMA',
-            'dbo',
+            $schemaSQL,
             'TABLE',
-            $tableName,
+            $tableSQL,
             'COLUMN',
             $columnName
         );
@@ -678,13 +689,22 @@ class SQLServerPlatform extends AbstractPlatform
      */
     protected function getAlterColumnCommentSQL($tableName, $columnName, $comment)
     {
+        if (strpos($tableName, '.') !== false) {
+            [$schemaSQL, $tableSQL] = explode('.', $tableName);
+            $schemaSQL              = $this->quoteStringLiteral($schemaSQL);
+            $tableSQL               = $this->quoteStringLiteral($tableSQL);
+        } else {
+            $schemaSQL = "'dbo'";
+            $tableSQL  = $this->quoteStringLiteral($tableName);
+        }
+
         return $this->getUpdateExtendedPropertySQL(
             'MS_Description',
             $comment,
             'SCHEMA',
-            'dbo',
+            $schemaSQL,
             'TABLE',
-            $tableName,
+            $tableSQL,
             'COLUMN',
             $columnName
         );
@@ -708,12 +728,21 @@ class SQLServerPlatform extends AbstractPlatform
      */
     protected function getDropColumnCommentSQL($tableName, $columnName)
     {
+        if (strpos($tableName, '.') !== false) {
+            [$schemaSQL, $tableSQL] = explode('.', $tableName);
+            $schemaSQL              = $this->quoteStringLiteral($schemaSQL);
+            $tableSQL               = $this->quoteStringLiteral($tableSQL);
+        } else {
+            $schemaSQL = "'dbo'";
+            $tableSQL  = $this->quoteStringLiteral($tableName);
+        }
+
         return $this->getDropExtendedPropertySQL(
             'MS_Description',
             'SCHEMA',
-            'dbo',
+            $schemaSQL,
             'TABLE',
-            $tableName,
+            $tableSQL,
             'COLUMN',
             $columnName
         );
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
@@ -3,11 +3,13 @@
 namespace Doctrine\Tests\DBAL\Functional\Schema;
 
 use Doctrine\Common\EventManager;
+use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Events;
 use Doctrine\DBAL\Platforms\OraclePlatform;
 use Doctrine\DBAL\Schema\Column;
 use Doctrine\DBAL\Schema\ColumnDiff;
 use Doctrine\DBAL\Schema\Comparator;
+use Doctrine\DBAL\Schema\SchemaDiff;
 use Doctrine\DBAL\Schema\Sequence;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Schema\TableDiff;
@@ -41,6 +43,23 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
 
         $this->_sm = $this->_conn->getSchemaManager();
     }
+
+
+    protected function tearDown()
+    {
+        parent::tearDown();
+
+        $this->_sm->tryMethod('dropTable', 'testschema.my_table_in_namespace');
+
+        //TODO: SchemaDiff does not drop removed namespaces?
+        try {
+            //sql server versions below 2016 do not support 'IF EXISTS' so we have to catch the exception here
+            $this->_conn->exec('DROP SCHEMA testschema');
+        } catch (DBALException $e) {
+            return;
+        }
+    }
+
 
     /**
      * @group DBAL-1220
@@ -568,6 +587,31 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
             self::assertEquals(array('foreign_key_test'), array_map('strtolower', $foreignKey->getColumns()));
             self::assertEquals(array('id'), array_map('strtolower', $foreignKey->getForeignColumns()));
         }
+    }
+
+
+    public function testTableInNamespace()
+    {
+        if (! $this->_sm->getDatabasePlatform()->supportsSchemas()) {
+            $this->markTestSkipped('Schema definition is not supported by this platform.');
+        }
+
+        //create schema
+        $diff                  = new SchemaDiff();
+        $diff->newNamespaces[] = 'testschema';
+
+        foreach ($diff->toSql($this->_sm->getDatabasePlatform()) as $sql) {
+            $this->_conn->exec($sql);
+        }
+
+        //test if table is create in namespace
+        $this->createTestTable('testschema.my_table_in_namespace');
+        self::assertContains('testschema.my_table_in_namespace', $this->_sm->listTableNames());
+
+        //tables without namespace should be created in default namespace
+        //default namespaces are ignored in table listings
+        $this->createTestTable('my_table_not_in_namespace');
+        self::assertContains('my_table_not_in_namespace', $this->_sm->listTableNames());
     }
 
     public function testCreateAndListViews()
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractSQLServerPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractSQLServerPlatformTestCase.php
@@ -577,6 +577,63 @@ abstract class AbstractSQLServerPlatformTestCase extends AbstractPlatformTestCas
         self::assertEquals('CREATE SCHEMA ' . $schemaName, $sql);
     }
 
+    public function testCreateTableWithSchemaColumnComments()
+    {
+        $table = new Table('testschema.test');
+        $table->addColumn('id', 'integer', ['comment' => 'This is a comment']);
+        $table->setPrimaryKey(['id']);
+
+        $expectedSql = [
+            'CREATE TABLE testschema.test (id INT NOT NULL, PRIMARY KEY (id))',
+            "EXEC sp_addextendedproperty N'MS_Description', N'This is a comment', N'SCHEMA', 'testschema', N'TABLE', 'test', N'COLUMN', id",
+        ];
+
+        self::assertEquals($expectedSql, $this->_platform->getCreateTableSQL($table));
+    }
+
+    public function testAlterTableWithSchemaColumnComments()
+    {
+        $tableDiff                        = new TableDiff('testschema.mytable');
+        $tableDiff->addedColumns['quota'] = new Column('quota', Type::getType('integer'), ['comment' => 'A comment']);
+
+        $expectedSql = [
+            'ALTER TABLE testschema.mytable ADD quota INT NOT NULL',
+            "EXEC sp_addextendedproperty N'MS_Description', N'A comment', N'SCHEMA', 'testschema', N'TABLE', 'mytable', N'COLUMN', quota",
+        ];
+
+        self::assertEquals($expectedSql, $this->_platform->getAlterTableSQL($tableDiff));
+    }
+
+    public function testAlterTableWithSchemaDropColumnComments()
+    {
+        $tableDiff                          = new TableDiff('testschema.mytable');
+        $tableDiff->changedColumns['quota'] = new ColumnDiff(
+            'quota',
+            new Column('quota', Type::getType('integer'), []),
+            ['comment'],
+            new Column('quota', Type::getType('integer'), ['comment' => 'A comment'])
+        );
+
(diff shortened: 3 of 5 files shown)


stlrnz:
Seems like Travis had some trouble yesterday:
https://www.traviscistatus.com/incidents/69gf0dn51g0v
https://www.traviscistatus.com/incidents/f4j8x61l4qgc

@Majkl578 Are you able to rerun the build? Seems like they just killed it.
What do you think about this PR? Does it have a chance to get merged?

morozov:
@stlrnz I restarted the build. Please see the code standards violations.

morozov:
@stlrnz is it possible to test the patch in a functional way? IMO, testing what SQL looks like is the least important thing.

stlrnz:
I've added a test for the modified table listing.

I don't really know which functional test to add, because this PR isn't changing any functionality. It just makes some existing features available for schema-based SQL Server environments (like I use). The only thing I could do is duplicate some existing tests and add a schema name. But does this really makes sense?

stlrnz:
@morozov do you have some more feedback to this PR?
Is there anyone here who could review it?

morozov:
@stlrnz sorry, I missed the update. I was asking for a functional test case which would have failed before the change and would describe the issue you're solving.

> Table listings do not consider schema names for the SQL Server platform.

As far as I understand, this ↑ is the root cause which is fixed in `SQLServer2008Platform::getListTablesSQL()`. 

> Without this fix doctrine will always try to create the whole schema, even if you only triggered an update. This is because the state of the current schema stored in database is not read correctly.

And this ↑ is the actual problem.

Is it possible to create a _functional_ test (the one which interacts with a real DB without mocking it) which initializes the schema, applies some changes to it and makes sure only the required changes are applied, not the entire schema is recreated? Ideally, this test should be applicable not only to SQL server but to an abstract DB platform, since we expect the schema updates to work the same on all platforms.

tests/Doctrine/Tests/DBAL/Schema/SQLServerSchemaManagerTest.php
```diff
@@ -0,0 +1,59 @@
...
+ */
+final class SQLServerSchemaManagerTest extends TestCase
+{
+    public function testListTableNames()
+    {
+        $driverMock = $this->createMock(Driver::class);
+        $platform   = $this->createMock(SQLServerPlatform::class);
+        $connection = $this
+            ->getMockBuilder(Connection::class)
+            ->setMethods(['fetchAll'])
```
morozov: IMO, a test like this doesn't bring much value. Mocking a DB connection means that we know for sure what a DB will return in our case which is often not true, especially given the variety of the supported platforms and their versions.

tests/Doctrine/Tests/DBAL/Platforms/AbstractSQLServerPlatformTestCase.php
```diff
@@ -579,14 +579,83 @@ public function testGetCreateSchemaSQL()
         self::assertEquals('CREATE SCHEMA ' . $schemaName, $sql);
     }
 
+    /**
+     * @group DBAL-42
```
morozov: Where does the 42 come from?

tests/Doctrine/Tests/DBAL/Schema/SQLServerSchemaManagerTest.php
stlrnz: I agree with your opinion in general.
However, this tests shows one special thing: The schema name "dbo" is getting ignored. This is done to increase compatibility to the old behaviour.

tests/Doctrine/Tests/DBAL/Platforms/AbstractSQLServerPlatformTestCase.php
stlrnz: All tests relatet to column comments in AbstractPlatformTestCase.php are annotated with this group. I thought it would be important to keep it grouped. However, if thats not true I will remove it.

stlrnz:
@morozov thank you for your feedback. I got your point and will try to implement a test like this.

stlrnz:
@morozov I pushed another test. I don't think it is 100% what you asked for, but let me explain the initial problem to make clear why it's hard for me to write a test.

I use doctrine2 as ORM component inside a symfony application in combination with a SQL Server database. The database is organized using schemas and created by using the command doctrine:schema:create. Everything works fine until this point.
When I call doctrine:schema:update right after creation, my expectation is that there is nothing to update (since I did not change anything on the entity). Instead, doctrine2 tries to create the whole schema again. This is because **doctrine2** "knows" that my tables should be defined in a schema but **dbal** does not report it back. Consequently, doctrine2 finds no matching table for the entity and tries to create it again.

So IMO the real problem is in the communication between doctrine2 and dbal. I fixed this by modifying the behaviour of dbal. In my (production) scenario this works very well.

I think a good test scenario would consider the following steps:
1. define a model entity (using schema)
2. create the schema
3. calculate the diff of the schema (with unmodified entity)
4. test that nothing has changed

IMO this is impossible to test in dbal because it involves doctrine2 functionality at least in step 3.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -570,6 +572,44 @@ public function testAlterTableScenario()
...
+        self::assertCount(1, $tableDetails->getIndexes());
+
+        $tableDiff                         = new TableDiff('testschema.my_table');
+        $tableDiff->fromTable              = $myTable;
+        $tableDiff->addedColumns['foo']    = new Column('foo', Type::getType('integer'));
+        $tableDiff->removedColumns['test'] = $myTable->getColumn('test');
+
+        $this->_sm->alterTable($tableDiff);
+
+        $tableDetails = $this->_sm->listTableDetails('alter_table');
```
morozov: Should `alter_table` be `testschema.my_table` instead? Otherwise, the last assertion fails since there's no such table at all.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -570,6 +572,44 @@ public function testAlterTableScenario()
...
+            $diff                  = new SchemaDiff();
+            $diff->newNamespaces[] = 'testschema';
+
+            foreach ($diff->toSql($this->_sm->getDatabasePlatform()) as $sql) {
+                $this->_conn->exec($sql);
+            }
+        }
+
+        $myTable = $this->createTestTable('testschema.my_table');
+        self::assertTrue(in_array('testschema.my_table', $this->_sm->listTableNames(), true));
```
morozov: This should be `self::assertContains()`.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -570,6 +572,44 @@ public function testAlterTableScenario()
...
     }
 
+
+    public function testAlterTableInNamespace()
+    {
+        if (! $this->_sm->getDatabasePlatform()->supportsSchemas()) {
+            $this->markTestSkipped('Schema definition is not supported by this platform.');
+        }
+
+        if (! in_array('testschema', $this->_sm->listNamespaceNames(), true)) {
```
morozov: Instead of checking if the schema already exists, could the test create the schema unconditionally and drop it on teardown? Otherwise, if a previous test execution didn't exit cleanly (e.g. the schema was created but the table wasn't), the subsequent executions will always fail.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -570,6 +572,44 @@ public function testAlterTableScenario()
         }
     }
 
+
+    public function testAlterTableInNamespace()
```
morozov: I think the functional testing could be reduced to making sure that if the platform supports schemas, then the table names returned by `listTableNames()` are fully qualified (i.e. are formatted as `schema-name.table.name`). The same seems already true for PostgreSQL.

At the same time, for the SQL Server specifically, if a table belongs to the `dbo` schema, it should be returned without the schema prefix for BC.

Does it fairly describe the fix?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -706,12 +726,21 @@ protected function getAlterColumnCommentSQL($tableName, $columnName, $comment)
      */
     protected function getDropColumnCommentSQL($tableName, $columnName)
     {
+        if (strpos($tableName, '.') !== false) {
+            list($schema, $table) = explode('.', $tableName);
+            $schema               = $this->quoteStringLiteral($schema);
```
morozov: It's better to name the variables as `$tableSQL` and `$schemaSQL` since they are not names but SQL expressions.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -570,6 +572,44 @@ public function testAlterTableScenario()
...
+            $diff->newNamespaces[] = 'testschema';
+
+            foreach ($diff->toSql($this->_sm->getDatabasePlatform()) as $sql) {
+                $this->_conn->exec($sql);
+            }
+        }
+
+        $myTable = $this->createTestTable('testschema.my_table');
+        self::assertTrue(in_array('testschema.my_table', $this->_sm->listTableNames(), true));
+
```
morozov: After this point, the rest of test could be removed as not really relevant to the fix. Covering `dbo` and non-`dbo` tables will be enough to test the changes.

tests/Doctrine/Tests/DBAL/Schema/SQLServerSchemaManagerTest.php
morozov: I see your point. On the other hand, the part about not ignoring the non-`dbo` namespace is already covered by the functional test. If ignoring the `dbo` namespace is implemented in the same functional test, it will have two benefits:

1. The logic of handling `dbo` and non-`dbo` namespaces is tested in a cohesive way which make the test much more informative.
2. The test doesn't need to depend on the DB internal names. It will only test if a table was created in a non-default namespace, it will be returned with the namespace, otherwise, the namespace will be omitted. At this point, `dbo` will remain a platform-specific detail which the test will not care about.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
stlrnz: OMG you are right! This assertion only did not fail in my test because one test before exactly this table was is created and modified in dbo.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
stlrnz: I trusted on the fact that the database is dropped on the beginning of the test. This would remove the schema, too. However, I'll try to implement it more specific.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
stlrnz: Yes, that's correct. I'll implement it in that way.

stlrnz:
@morozov thank you for your patience and the helpful feedback. I've updated the tests according to your comments.

However, there is one nasty thing. The SchemaDiff does not handle dropping of namespaces. So for now, i had to hardcode the SQL.

Majkl578:
@stlrnz Hi, can you please rebase onto current master? We added CI testing for SQL Server recently in #2617 & #3050 so your functional tests will be checked against live SQL Server on CI now.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -42,6 +43,29 @@ protected function setUp()
         $this->_sm = $this->_conn->getSchemaManager();
     }
 
+
+    protected function tearDown()
+    {
+        parent::tearDown();
+
+        if ($this->getName() !== 'testTableInNamespace' || ! $this->_sm->getDatabasePlatform()->supportsSchemas()) {
```
morozov: As far as I understand, this is in order to prevent the exception from `dropTable()` if the table doesn't exist. This is not the best solution since the table may not exist if the test failed and didn't manage to create it.

Please use `$this->_sm->tryMethod()` and get rid of the conditional logic.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -42,6 +43,29 @@ protected function setUp()
...
+
+        //TODO: SchemaDiff does not drop removed Namespaces?
+        $this->_conn->exec('DROP SCHEMA testschema');
+
+        //$diff                      = new SchemaDiff();
+        //$diff->removedNamespaces[] = 'testschema';
+
+        //foreach ($diff->toSql($this->_sm->getDatabasePlatform()) as $sql) {
+        //    $this->_conn->exec($sql);
+        //}
```
morozov: Please remove commented out lines.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
stlrnz: Actually, this is done to reduce the attempts to drop the table / the namespace.
However, I think performance is no requirement on functional tests. So, I will remove it.
But there will be a special handling for the `DROP SCHEMA` left.

stlrnz:
@Majkl578 I merged the changes. However, it seems there are some problems with the AppVeyor tests.

@morozov I've improved the `tearDown()` according to your feedback.

Ocramius:
@stlrnz we don't merge upstream - only rebases are really allowed here.

Also, we now merged 7.2-only on AppVeyor, so maybe the results may be different :)

morozov:
> The SchemaDiff does not handle dropping of namespaces. So for now, I had to hardcode the SQL.

This is okay. We can fix it later if needed.

@Ocramius does schema manager not support dropping schemas by design or it's a missing feature?

morozov (dismissed):

Ocramius:
> @Ocramius does schema manager not support dropping schemas by design or it's a missing feature?

It's something that @deeky666 started handling, but it leads to a lot of tricky scenarios where a DB connection still needs an existing DB (depends on the DB, but take SQLite, for example), so it was likely never improved.

Ocramius (dismissed):
:ship: thanks @stlrnz

Ocramius:
@stlrnz some failures in https://ci.appveyor.com/project/doctrine/dbal/build/1.0.15/job/rjfn6vs2bmmlox3y - can you check if they are related?

Ocramius:
Meanwhile: restarted the build.

stlrnz:
Thank you for approving!
@Ocramius should I try to do a rebase on the current master?

I seems that AppVeyor randomly fails on SQL Server 2008 because of "Cannot open database "doctrine_tests" requested by the login. The login failed.". I cannot see any regularity.
This never happened on my local SQL Server 2008. There was even one good run on AppVeyor: https://ci.appveyor.com/project/doctrine/dbal/build/1.0.14/job/ny1mc1f4nfs1l0c5

So I think its a Problem on AppVeyors side. SQL Server 2012 and 2017 never fails the test.

morozov:
> SQLSTATE [42000, 4060]: [Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Cannot open database "doctrine_tests" requested by the login. The login failed.

I could've sworn I've seen this multiple times but the only other place I can find is [build 1.0.22](https://ci.appveyor.com/project/doctrine/dbal/build/1.0.22/job/g5h80ml62y30h2f0) where the actual error was (see https://github.com/doctrine/dbal/pull/3061#issuecomment-374462620):
> SQLSTATE[HY000]: General error: user-supplied statement class must be derived from PDOStatement

[Build 1.0.36](https://ci.appveyor.com/project/doctrine/dbal/build/1.0.36/job/sfnuc8cr30lvgkbg) doesn't count.

I wouldn't be surprised if this was a manifestation of some other inconsistently reproducible issue (e.g. caused by an unlucky sequence of the tests which is non-deterministic in our case).

Majkl578:
SQL Server issue should be fixed by #3071.

Majkl578:
I did a rebase & squased those 25 commits. Original tree is kept in my fork for now if needed: https://github.com/Majkl578/doctrine-dbal/tree/stlrnz/mssql-schema

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -42,6 +44,23 @@ protected function setUp()
...
     }
 
+
+    protected function tearDown()
+    {
+        parent::tearDown();
+
+        $this->_sm->tryMethod('dropTable', 'testschema.my_table_in_namespace');
+
+        //TODO: SchemaDiff does not drop removed namespaces?
```
Majkl578: Does this need adressing before merge?

morozov: No, it’s OK.

morozov (approved):

Ocramius (approved):

PR #3025, opened 2018-02-17
Added PHPStan, apply changes for level 3

Majkl578:
Mostly annotation changes/fixes to match what the methods/properties really are used for. Added few asserts and instanceof.
Not really sure if you want to consider this as BC issue - I'd hope not, but if so we can retarget to develop instead.

Level 3, not going further as it requires actually changing code itself, that should follow later in develop with strict typing.

There is a PHP notice currently since Travis doesn't have OCI driver:
```
Notice: Use of undefined constant OCI_DEFAULT - assumed 'OCI_DEFAULT'
```

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -19,12 +19,14 @@
 
 namespace Doctrine\DBAL;
 
+use Doctrine\DBAL\Driver\ResultStatement;
 use Doctrine\DBAL\Driver\ServerInfoAwareConnection;
 use Doctrine\DBAL\Exception\InvalidArgumentException;
 use Closure;
 use Exception;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\DBAL\Driver\Connection as DriverConnection;
+use Doctrine\DBAL\Driver\Statement as DriverStatement;
 use Doctrine\Common\EventManager;
 use Doctrine\DBAL\Cache\ResultCacheStatement;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
@@ -32,6 +34,7 @@ use Doctrine\DBAL\Cache\ArrayStatement;
 use Doctrine\DBAL\Cache\CacheException;
 use Doctrine\DBAL\Driver\PingableConnection;
 use Throwable;
+use function assert;
 use function array_key_exists;
 use function array_merge;
 use function func_get_args;
@@ -108,7 +111,7 @@ class Connection implements DriverConnection
     /**
      * The wrapped driver connection.
      *
-     * @var \Doctrine\DBAL\Driver\Connection
+     * @var \Doctrine\DBAL\Driver\Connection|null
      */
     protected $_conn;
 
@@ -409,7 +412,9 @@ class Connection implements DriverConnection
     {
         $version = $this->getDatabasePlatformVersion();
 
-        if (null !== $version) {
+        if ($version !== null) {
+            assert($this->_driver instanceof VersionAwarePlatformDriver);
+
             $this->platform = $this->_driver->createDatabasePlatformForVersion($version);
         } else {
             $this->platform = $this->_driver->getDatabasePlatform();
@@ -875,7 +880,7 @@ class Connection implements DriverConnection
      *
      * @param string $statement The SQL statement to prepare.
      *
-     * @return \Doctrine\DBAL\Statement The prepared statement.
+     * @return DriverStatement The prepared statement.
      *
      * @throws \Doctrine\DBAL\DBALException
      */
@@ -903,7 +908,7 @@ class Connection implements DriverConnection
      * @param array                                       $types  The types the previous parameters are in.
      * @param \Doctrine\DBAL\Cache\QueryCacheProfile|null $qcp    The query cache profile, optional.
      *
-     * @return \Doctrine\DBAL\Driver\Statement The executed statement.
+     * @return ResultStatement The executed statement.
      *
      * @throws \Doctrine\DBAL\DBALException
      */
@@ -955,7 +960,7 @@ class Connection implements DriverConnection
      * @param array                                  $types  The types the previous parameters are in.
      * @param \Doctrine\DBAL\Cache\QueryCacheProfile $qcp    The query cache profile.
      *
-     * @return \Doctrine\DBAL\Driver\ResultStatement
+     * @return ResultStatement
      *
      * @throws \Doctrine\DBAL\Cache\CacheException
      */
@@ -1139,7 +1144,7 @@ class Connection implements DriverConnection
     /**
      * Fetches the SQLSTATE associated with the last database operation.
      *
-     * @return int The last error code.
+     * @return string|null The last error code.
      */
     public function errorCode()
     {
--- a/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
+++ b/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
@@ -21,6 +21,7 @@ namespace Doctrine\DBAL\Connections;
 
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Driver;
+use Doctrine\DBAL\Driver\Connection as DriverConnection;
 use Doctrine\DBAL\Configuration;
 use Doctrine\Common\EventManager;
 use Doctrine\DBAL\Event\ConnectionEventArgs;
@@ -88,7 +89,7 @@ class MasterSlaveConnection extends Connection
     /**
      * Master and slave connection (one of the randomly picked slaves).
      *
-     * @var \Doctrine\DBAL\Driver\Connection[]
+     * @var DriverConnection[]|null[]
      */
     protected $connections = ['master' => null, 'slave' => null];
 
@@ -199,7 +200,7 @@ class MasterSlaveConnection extends Connection
      *
      * @param string $connectionName
      *
-     * @return \Doctrine\DBAL\Driver
+     * @return DriverConnection
      */
     protected function connectTo($connectionName)
     {
--- a/lib/Doctrine/DBAL/Driver/Connection.php
+++ b/lib/Doctrine/DBAL/Driver/Connection.php
@@ -36,7 +36,7 @@ interface Connection
      *
      * @param string $prepareString
      *
-     * @return \Doctrine\DBAL\Driver\Statement
+     * @return Statement
      */
     public function prepare($prepareString);
 
--- a/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
+++ b/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
@@ -22,6 +22,7 @@ namespace Doctrine\DBAL\Driver\IBMDB2;
 use Doctrine\DBAL\Driver\Connection;
 use Doctrine\DBAL\Driver\ServerInfoAwareConnection;
 use Doctrine\DBAL\ParameterType;
+use stdClass;
 use const DB2_AUTOCOMMIT_OFF;
 use const DB2_AUTOCOMMIT_ON;
 use function db2_autocommit;
@@ -74,6 +75,7 @@ class DB2Connection implements Connection, ServerInfoAwareConnection
      */
     public function getServerVersion()
     {
+        /** @var stdClass $serverInfo */
         $serverInfo = db2_server_info($this->_conn);
 
         return $serverInfo->DBMS_VER;
--- a/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
+++ b/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
@@ -252,7 +252,7 @@ class MysqliStatement implements \IteratorAggregate, Statement
     }
 
     /**
-     * @return bool|array
+     * @return mixed[]|false
      */
     private function _fetch()
     {
--- a/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
@@ -934,7 +934,7 @@ class SqlitePlatform extends AbstractPlatform
     /**
      * @param \Doctrine\DBAL\Schema\TableDiff $diff
      *
-     * @return array|bool
+     * @return string[]|false
      */
     private function getSimpleAlterTableSQL(TableDiff $diff)
     {
--- a/lib/Doctrine/DBAL/Schema/Schema.php
+++ b/lib/Doctrine/DBAL/Schema/Schema.php
@@ -76,7 +76,7 @@ class Schema extends AbstractAsset
     protected $_sequences = [];
 
     /**
-     * @var \Doctrine\DBAL\Schema\SchemaConfig
+     * @var SchemaConfig
      */
     protected $_schemaConfig = false;
 
--- a/lib/Doctrine/DBAL/Schema/Table.php
+++ b/lib/Doctrine/DBAL/Schema/Table.php
@@ -77,7 +77,7 @@ class Table extends AbstractAsset
     protected $_options = [];
 
     /**
-     * @var SchemaConfig
+     * @var SchemaConfig|null
      */
     protected $_schemaConfig = null;
 
--- a/lib/Doctrine/DBAL/Sharding/PoolingShardConnection.php
+++ b/lib/Doctrine/DBAL/Sharding/PoolingShardConnection.php
@@ -22,6 +22,7 @@ namespace Doctrine\DBAL\Sharding;
 use Doctrine\Common\EventManager;
 use Doctrine\DBAL\Configuration;
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\Connection as DriverConnection;
 use Doctrine\DBAL\Driver;
 use Doctrine\DBAL\Event\ConnectionEventArgs;
 use Doctrine\DBAL\Events;
@@ -68,19 +69,19 @@ use function is_string;
 class PoolingShardConnection extends Connection
 {
     /**
-     * @var array
+     * @var DriverConnection[]
      */
-    private $activeConnections;
+    private $activeConnections = [];
 
     /**
-     * @var int
+     * @var int|null
      */
     private $activeShardId;
 
     /**
-     * @var array
+     * @var mixed[]
      */
-    private $connections;
+    private $connectionParameters = [];
 
     /**
      * @param array                         $params
@@ -108,7 +109,7 @@ class PoolingShardConnection extends Connection
             throw new \InvalidArgumentException("The 'shardChoser' configuration is not a valid instance of Doctrine\DBAL\Sharding\ShardChoser\ShardChoser");
         }
 
-        $this->connections[0] = array_merge($params, $params['global']);
+        $this->connectionParameters[0] = array_merge($params, $params['global']);
 
         foreach ($params['shards'] as $shard) {
             if ( ! isset($shard['id'])) {
@@ -119,11 +120,11 @@ class PoolingShardConnection extends Connection
                 throw new \InvalidArgumentException("Shard Id has to be a non-negative number.");
             }
 
-            if (isset($this->connections[$shard['id']])) {
+            if (isset($this->connectionParameters[$shard['id']])) {
                 throw new \InvalidArgumentException("Shard " . $shard['id'] . " is duplicated in the configuration.");
             }
 
-            $this->connections[$shard['id']] = array_merge($params, $shard);
+            $this->connectionParameters[$shard['id']] = array_merge($params, $shard);
         }
 
         parent::__construct($params, $driver, $config, $eventManager);
@@ -144,7 +145,7 @@ class PoolingShardConnection extends Connection
      */
     public function getParams()
     {
-        return $this->activeShardId ? $this->connections[$this->activeShardId] : $this->connections[0];
+        return $this->activeShardId ? $this->connectionParameters[$this->activeShardId] : $this->connectionParameters[0];
     }
 
     /**
@@ -241,7 +242,7 @@ class PoolingShardConnection extends Connection
 
         $driverOptions = $params['driverOptions'] ?? [];
 
-        $connectionParams = $this->connections[$shardId];
+        $connectionParams = $this->connectionParameters[$shardId];
 
         $user = $connectionParams['user'] ?? null;
         $password = $connectionParams['password'] ?? null;
@@ -269,7 +270,7 @@ class PoolingShardConnection extends Connection
     public function close()
     {
         $this->_conn             = null;
-        $this->activeConnections = null;
+        $this->activeConnections = [];
         $this->activeShardId     = null;
     }
 }
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,0 +1,46 @@
+parameters:
+    level: 3
+    paths:
+        - %currentWorkingDirectory%/lib
+    autoload_files:
+        - %currentWorkingDirectory%/tests/phpstan-polyfill.php
+    reportUnmatchedIgnoredErrors: false
+    ignoreErrors:
+        # extension not available
+        - '~^(Used )?(Function|Constant) sasql_\S+ not found\.\z~i'
+
+        # removing it would be BC break
+        - '~^Constructor of class Doctrine\\DBAL\\Schema\\Table has an unused parameter \$idGeneratorType\.\z~'
+
+        # changing these would be a BC break, to be done in next major
+        - '~^Method Doctrine\\DBAL\\Driver\\IBMDB2\\DB2Connection::exec\(\) should return int but returns bool\.\z~'
+        - '~^Method Doctrine\\DBAL\\Query\\QueryBuilder::execute\(\) should return Doctrine\\DBAL\\Driver\\Statement\|int but returns Doctrine\\DBAL\\Driver\\ResultStatement\.\z~'
+        - '~^Property Doctrine\\DBAL\\Schema\\Table::\$_primaryKeyName \(string\) does not accept (default value of type )?false\.\z~'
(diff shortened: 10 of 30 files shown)


lib/Doctrine/DBAL/Connection.php
```diff
@@ -402,7 +403,7 @@ private function detectDatabasePlatform()
     {
         $version = $this->getDatabasePlatformVersion();
 
-        if (null !== $version) {
+        if ($version !== null && $this->_driver instanceof VersionAwarePlatformDriver) {
```
morozov: Wouldn't it be enough to check if the driver is not null? Type hint should help static analysis w/o the run time check.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -419,7 +420,7 @@ private function detectDatabasePlatform()
      * or the underlying driver connection cannot determine the platform
      * version without having to query it (performance reasons).
      *
-     * @return string|null
+     * @return string|bool|null
```
morozov: I'd say it's a bug and should be fixed, not documented.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -478,7 +479,7 @@ private function getDatabasePlatformVersion()
     /**
      * Returns the database server version if the underlying driver supports it.
      *
-     * @return string|null
+     * @return string|bool|null
```
morozov: Boolean is a bug.

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -85,7 +86,7 @@ class MasterSlaveConnection extends Connection
     /**
      * Master and slave connection (one of the randomly picked slaves).
      *
-     * @var \Doctrine\DBAL\Driver\Connection[]
+     * @var DriverConnection[]|null[]
```
morozov: Isn't it something like `(DriverConnection|null)[]`? Not all values are necessary of the same type.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2716,7 +2716,7 @@ public function getCurrentTimestampSQL()
      *
      * @param int $level
      *
-     * @return string
+     * @return string|int
```
morozov: Should it be fixed or silenced instead? Doesn't look like it's by design (the SQLite platform impelementation).

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -423,7 +423,7 @@ public function dropView($name)
      *
      * @param string $database The name of the database to create.
      *
-     * @return void
+     * @return void|bool
```
morozov: Looks like a bug. Mixing something with void doesn't make sense.

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -917,7 +917,7 @@ protected function _getPortableTablesList($tables)
     /**
      * @param array $table
      *
-     * @return array
+     * @return mixed[]|string
```
morozov: Can elements of the array be anything else than string? `string[]|string` would make more sense.

lib/Doctrine/DBAL/Schema/Schema.php
```diff
@@ -73,7 +73,7 @@ class Schema extends AbstractAsset
     protected $_sequences = [];
 
     /**
-     * @var \Doctrine\DBAL\Schema\SchemaConfig
+     * @var \Doctrine\DBAL\Schema\SchemaConfig|bool
```
morozov: A nullable type would make more sense here.

lib/Doctrine/DBAL/Sharding/PoolingShardConnection.php
```diff
@@ -65,17 +66,17 @@
 class PoolingShardConnection extends Connection
 {
     /**
-     * @var array
+     * @var DriverConnection[]|null
```
morozov: The variable could be initialized with an empty array instead.

morozov (commented):
I'd say, most if not all places where `false`/`bool` is added as part of the union type are not by design and should be fixed in a BC way (if possible) or silenced instead of being documented.

carusogabriel:
Already prepared changes for `tests/` as well https://github.com/carusogabriel/dbal/tree/phpstan-tests

lib/Doctrine/DBAL/Connection.php
Majkl578: `$this->_platform` is typed as AbstractPlatform, but `createDatabasePlatformForVersion()` is only defined in some drivers implementing VersionAwarePlatformDriver.
I at least replaced this by `assert()` inside the `if` since `Connection::getDatabasePlatformVersion()` returns null for non-version-aware drivers.

lib/Doctrine/DBAL/Connection.php
Majkl578: Hmm this is because `SqlAnywhereConnection::getServerVersion()` calls `ResultStatement::fetchColumn()` which returns `string|bool`. :| Replaced with assert there and removed here.

lib/Doctrine/DBAL/Connection.php
Majkl578: Same as above.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1132,7 +1133,7 @@ public function getTransactionNestingLevel()
     /**
      * Fetches the SQLSTATE associated with the last database operation.
      *
-     * @return int The last error code.
+     * @return string|bool|null The last error code.
```
Majkl578: This on the other hand is not a bug, some implementors explicitly return `false` (`SQLSrvConnection::errorCode()`).

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1163,7 +1164,7 @@ public function errorInfo()
      *
      * @param string|null $seqName Name of the sequence object from which the ID should be returned.
      *
-     * @return string A string representation of the last inserted ID.
+     * @return string|int|bool A string representation of the last inserted ID.
```
Majkl578: Same here (`OCI8Connection::lastInsertId()`).

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
Majkl578: Such notation is more correct, but not supported i.e. by IDEs (PhpStorm will break completely). Pretty much same like intersections vs. unions we discussed earlier somewhere else.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
Majkl578: I can change those ints to strings, but I really have no idea of consequences (either in DBAL or for consumers). SqlAnywherePlatoform returns ints too... Maybe better leave this to 3.x?

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -899,7 +899,7 @@ public function getAlterTableSQL(TableDiff $diff)
     /**
      * @param \Doctrine\DBAL\Schema\TableDiff $diff
      *
-     * @return array|bool
+     * @return string[]|false
```
Majkl578: As much as I hate using `false` as type, there is no better way around here. Usually these cases are something to replace with `null` for 3.x.

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
Majkl578: Yep, removed. But `OracleSchemaManager::createDatabase()` returns true so I removed that one. Should be documented in UPGRADING?

lib/Doctrine/DBAL/Sharding/PoolingShardConnection.php
Majkl578: Probably ok here and for `$connections`, changed (`close()` now sets empty array instead of null).

lib/Doctrine/DBAL/Schema/Schema.php
Majkl578: Can't be done in 2.x I guess...

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
Majkl578: No idea since `$table` is uselessly types as `array`. If you are certain, I can change it.

morozov: I'd look at it from the caller's standpoint: whatever callers can consume by design (`string[]` or `string`) should be valid return, the rest shouldn't.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: I'm all for leaving it for 3.x where we could fix the actual code instead of complicating the API documentation.

carusogabriel:
Any updates on this one? :)

Majkl578:
I'll wait for PHPStan 0.10 (to be released next month) before finishing this.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
```diff
@@ -74,6 +75,7 @@ public function __construct(array $params, $username, $password, $driverOptions
      */
     public function getServerVersion()
     {
+        /** @var stdClass $serverInfo */
         $serverInfo = db2_server_info($this->_conn);
```
carusogabriel: Does this function always return a `stdClass`? Can't we force it in PHPStan stubs?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -300,6 +301,7 @@ public function bindParam($column, &$variable, $type = ParameterType::STRING, $l
         $column = $this->_paramMap[$column] ?? $column;
 
         if ($type === ParameterType::LARGE_OBJECT) {
+            /** @var OCI_Lob $lob */
```
carusogabriel: PHPStan stubs false-positive?

Majkl578: Probably.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
Majkl578: I didn't actually check whether it is stdClass or not, it returns _some object crate_: http://php.net/db2_server_info

Majkl578: oh, and `false`, how unusual ;)

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
morozov: It's actually `OCI-Lob`.

lib/Doctrine/DBAL/Connection.php
morozov: DBAL should eliminate this discrepancy. The clients will not have to know the low level details.

lib/Doctrine/DBAL/Driver/Connection.php
```diff
@@ -99,7 +99,7 @@ public function rollBack();
     /**
      * Returns the error code associated with the last operation on the database handle.
      *
-     * @return string|null The error code, or null if no operation has been run on the database handle.
+     * @return string|bool|null The error code, or null if no operation has been run on the database handle.
```
morozov: `bool|null` should be eliminated, not document.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -287,7 +288,9 @@ public function fetch($fetchMode = null, $cursorOrientation = \PDO::FETCH_ORI_NE
             return $this->fetchColumn();
         }
 
+        /** @var mixed[]|false $values */
```
morozov: Can we change the `_fetch()`'es phpDoc instead?

lib/Doctrine/DBAL/Schema/ForeignKeyConstraint.php
```diff
@@ -333,7 +333,7 @@ public function getOptions()
      * Returns the referential action for UPDATE operations
      * on the referenced table the foreign key constraint is associated with.
      *
-     * @return string|null
+     * @return string|false|null
```
morozov: `false|null` is undesirable.

lib/Doctrine/DBAL/Schema/ForeignKeyConstraint.php
```diff
@@ -344,7 +344,7 @@ public function onUpdate()
      * Returns the referential action for DELETE operations
      * on the referenced table the foreign key constraint is associated with.
      *
-     * @return string|null
+     * @return string|false|null
```
morozov: `false|null` is undesirable.

lib/Doctrine/DBAL/Schema/ForeignKeyConstraint.php
```diff
@@ -357,7 +357,7 @@ public function onDelete()
      *
      * @param string $event Name of the database operation/event to return the referential action for.
      *
-     * @return string|null
+     * @return string|false|null
```
morozov: `false|null` is undesirable.

lib/Doctrine/DBAL/Schema/Schema.php
morozov: Let's ignore it and fix instead of introducing an intermediate incorrect type.

lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -62,7 +62,7 @@ class Table extends AbstractAsset
     protected $_indexes = [];
 
     /**
-     * @var string
+     * @var string|bool
```
morozov: `string|false`?

phpstan.neon.dist
```diff
@@ -0,0 +1,39 @@
...
+        - '~^Call to an undefined method Doctrine\\DBAL\\Driver\\PDOConnection::sqliteCreateFunction\(\)\.\z~'
+
+        # legacy variadic-like signature
+        - '~^Method Doctrine\\DBAL\\Driver\\Connection::query\(\) invoked with \d+ parameters?, 0 required\.\z~'
+
+        # some drivers actually do accept 2nd parameter...
+        - '~^Method Doctrine\\DBAL\\Platforms\\AbstractPlatform::getListTableForeignKeysSQL\(\) invoked with \d+ parameters, 1 required\.\z~'
+
+        # this method is undeclared, relies on PDO ancestor in implementors
+        - '~^Call to an undefined method Doctrine\\DBAL\\Driver\\Statement::nextRowset\(\)\.\z~'
```
morozov: Can this be replaced with `assert($statement instanceof PDOStatement)` in the code?

phpstan.neon.dist
```diff
@@ -0,0 +1,39 @@
+parameters:
+    level: 3
+    paths:
+        - %currentWorkingDirectory%/lib
+    autoload_files:
+        - %currentWorkingDirectory%/tests/phpstan-polyfill.php
+    reportUnmatchedIgnoredErrors: false
+    ignoreErrors:
+        # extension not available
+        - '~^(Used )?(Function|Constant) sasql_\S+ not found\.\z~i'
```
morozov: Could we generate a stub in the longer term? As long as we officially support the driver.

phpstan.neon.dist
```diff
@@ -0,0 +1,39 @@
...
+
+        # legacy remnants from doctrine/common
+        - '~^Class Doctrine\\Common\\(Collections\\Collection|Persistence\\Proxy) not found\.\z~'
+        - '~^.+ on an unknown class Doctrine\\Common\\(Collections\\Collection|Persistence\\Proxy)\.\z~'
+
+        # inheritance variance inference issue
+        - '~^Method Doctrine\\DBAL\\Driver\\PDOConnection::\w+\(\) should return Doctrine\\DBAL\\Driver\\Statement but returns PDOStatement\.\z~'
+
+        # may not exist when pdo_sqlsrv is not loaded
+        - '~^Access to undefined constant PDO::SQLSRV_ENCODING_BINARY\.\z~'
```
morozov: Could we generate a stub in the longer term?

tests/phpstan-polyfill.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+(function () : void {
+    static $map = [
+        'db2' => __DIR__ . '/../vendor/jetbrains/phpstorm-stubs/ibm_db2/ibm_db2.php',
```
morozov: Is it really `'db2'`, not `'ibm_db2'`?
```php
var_dump(extension_loaded('db2'));
// bool(false)

var_dump(extension_loaded('ibm_db2'));
// bool(true)
```

tests/phpstan-polyfill.php
```diff
@@ -0,0 +1,20 @@
...
+        'mysqli' => __DIR__ . '/../vendor/jetbrains/phpstorm-stubs/mysqli/mysqli.php',
+        'oci8' => __DIR__ . '/../vendor/jetbrains/phpstorm-stubs/oci8/oci8.php',
+        'sqlsrv' => __DIR__ . '/../vendor/jetbrains/phpstorm-stubs/sqlsrv/sqlsrv.php',
```
morozov: Given that the stub files are nicely organized, in order to avoid copy/paste and simplify maintenance, can this be reworked as:
```php
$extensions = ['ibm_db2', ...];

foreach ($extensions as $extension) {
    if (extension_loaded($extension)) {
        continue;
    }

    require sprintf(__DIR__ . '/../vendor/jetbrains/phpstorm-stubs/%1$s/%1$s.php', $extension);
}
```

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: Please ignore instead.

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
morozov: Please figure out the correct type from the API standpoint and ignore the deviations.

lib/Doctrine/DBAL/Schema/Schema.php
morozov: Please ignore.

phpstan.neon.dist
Majkl578: I have no idea. PhpStorm stubs don't have sasql and I didn't find any documentation either, nor PECL extension. Working with SAP SQL Anywhere is really hard. :(

lib/Doctrine/DBAL/Schema/Table.php
Majkl578: will ignore

lib/Doctrine/DBAL/Schema/ForeignKeyConstraint.php
Majkl578: adding to ignore, but it's a bug, false is explicitly returned here :/

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
Majkl578:
```
 ------ ------------------------------------------------------------------------------------------------------------------------ 
  Line   lib/Doctrine/DBAL/Schema/OracleSchemaManager.php                                                                        
 ------ ------------------------------------------------------------------------------------------------------------------------ 
  104    Method Doctrine\DBAL\Schema\OracleSchemaManager::_getPortableTableDefinition() should return array but returns string.  
 ------ ------------------------------------------------------------------------------------------------------------------------ 

 ------ ---------------------------------------------------------------------------------------------------------------------------- 
  Line   lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php                                                                        
 ------ ---------------------------------------------------------------------------------------------------------------------------- 
  222    Method Doctrine\DBAL\Schema\PostgreSqlSchemaManager::_getPortableTableDefinition() should return array but returns string.  
 ------ ---------------------------------------------------------------------------------------------------------------------------- 

 ------ --------------------------------------------------------------------------------------------------------------------------- 
  Line   lib/Doctrine/DBAL/Schema/SQLServerSchemaManager.php                                                                        
 ------ --------------------------------------------------------------------------------------------------------------------------- 
  210    Method Doctrine\DBAL\Schema\SQLServerSchemaManager::_getPortableTableDefinition() should return array but returns string.
```
Honestly, no idea. Some managers return `$table` as is and some return a string (schema name + table name).
This definitely looks like a bad design (git-blame indicates addition back in 2008-2011).
Help? :)

Majkl578: putting it into ignores for now, unlikely to be properly fixable in 2.x

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
Majkl578: hmm, that's not a valid class name 🙈 moving to ignore rather than typehinting _hack_

lib/Doctrine/DBAL/Driver/Connection.php
Majkl578: ignored

phpstan.neon.dist
Majkl578: Not sure if possible for a class constant...

morozov: It's missing from the PDO stub as well as many other driver-specific symbols. I'll file a PR for [JetBrains/phpstorm-stubs](https://github.com/JetBrains/phpstorm-stubs) later.

phpstan.neon.dist
morozov: For future reference, I found a few working links at https://archive.sap.com/documents/docs/DOC-40537. I'll try to drop in the Windows DLL into PHP 5 and see if I can dump the signatures. Sometime.

phpstan.neon.dist
```diff
@@ -0,0 +1,49 @@
...
+        - '~^Property Doctrine\\DBAL\\Schema\\Table::\$_primaryKeyName \(string\) does not accept (default value of type )?false\.\z~'
+        - '~^Property Doctrine\\DBAL\\Schema\\Schema::\$_schemaConfig \(Doctrine\\DBAL\\Schema\\SchemaConfig\) does not accept default value of type false\.\z~'
+        - '~^Method Doctrine\\DBAL\\Schema\\ForeignKeyConstraint::onEvent\(\) should return string\|null but returns false\.\z~'
+        - '~^Method Doctrine\\DBAL\\Schema\\(Oracle|PostgreSql|SQLServer)SchemaManager::_getPortableTableDefinition\(\) should return array but returns string\.\z~'
+        - '~^Method Doctrine\\DBAL\\Platforms\\(|SQLAnywhere|Sqlite)Platform::_getTransactionIsolationLevelSQL\(\) should return string but returns int\.\z~'
+        - '~^Method Doctrine\\DBAL\\Driver\\OCI8\\OCI8Connection::lastInsertId\(\) should return string but returns (int|false)\.\z~'
+        - '~^Method Doctrine\\DBAL\\Driver\\SQLSrv\\SQLSrvConnection::errorCode\(\) should return string\|null but returns false\.\z~'
+
+        # actually OCI-Lob - not even a valid class name...
+        - '~^Call to an undefined method object::writeTemporary\(\)\.\z~'
```
morozov: Hmm… all generated stubs have it as `OCI_Lob` (probably to make it valid), shouldn't then PHPStan just use the "fixed" names? The `oci_new_descriptor()` is stubbed as `oci_new_descriptor() : OCI_Lob`, and `OCI_Lob::writeTemporary()` is also stubbed, so I'd expect not seeing this error.

phpstan.neon.dist
morozov: Filed https://github.com/JetBrains/phpstorm-stubs/pull/381.

morozov: > Not sure if possible for a class constant…

Oh, I see what you're saying. The class stub will be not loaded for PDO in the extension itself is loaded, even if it's compiled w/o the missing constants. PDO is so nicely designed.

Majkl578: It's definitely missing in the stubs, but it is also missing in PHP itself unless oci extension is loaded. Who said multiple inheritance in PHP is not possible? :grin:

phpstan.neon.dist
Majkl578: https://github.com/phpstan/phpstan/issues/1123

Majkl578: Not really sure what to think about this. :)

phpstan.neon.dist
Majkl578: The link lead me to: http://dcx.sap.com/index.html#sa160/en/dbprogramming/php-api.html
This looks good enough to build the stubs. Maybe i.e. @carusogabriel may want to help here? :)

But we should probably consider dropping the platform/driver altogether.
Given that DBAL 3 will require PHP 7.2, and SQL Anywhere is barely keeping up (the link only lists PHP 5 modules)...

Majkl578:
CI passed. :green_heart:

phpstan.neon.dist
Majkl578: Nevermind, did this during Slack outage. :stuck_out_tongue: 
https://github.com/JetBrains/phpstorm-stubs/pull/382

morozov (approved):

Majkl578:
@morozov Do you want this in 2.8 or wait for release and then merge into 2.9?

morozov:
@Majkl578 I understand it as a non-breaking low-risk, mostly documentation improvement. Unless it's not true, I'm fine with having it in `2.8.0`.

Majkl578:
Yup, and no tests changed so it should be fine. Let's ship it. 🚢

Ocramius:
:+1:

PR #3027, opened 2018-02-19
Documentation: Warn against using object fields in MySQL and MariaDB

Jasu:
While `types.rst` warns against using `object` in PostgreSQL, it does not warn against using it in MySQL. This PR adds a warning against it, since using object fields will cause backups to silently fail. Based on previous PRs and issues, fixing this would break backward compatibility.


Ocramius (approved):
LGTM :+1:

PR #3031, opened 2018-02-22
Connection parameters are cached hashed

fullbl:
As per https://github.com/doctrine/doctrine2/issues/7086 (I didn't know it was related to dbal and not doctrine2), parameters should be hashed because they can have passwords inside.

### DIFF
--- a/lib/Doctrine/DBAL/Cache/QueryCacheProfile.php
+++ b/lib/Doctrine/DBAL/Cache/QueryCacheProfile.php
@@ -20,18 +20,19 @@
 namespace Doctrine\DBAL\Cache;
 
 use Doctrine\Common\Cache\Cache;
+use function hash;
+use function serialize;
+use function sha1;
 
 /**
  * Query Cache Profile handles the data relevant for query caching.
  *
  * It is a value object, setter methods return NEW instances.
- *
- * @author Benjamin Eberlei <kontakt@beberlei.de>
  */
 class QueryCacheProfile
 {
     /**
-     * @var \Doctrine\Common\Cache\Cache|null
+     * @var Cache|null
      */
     private $resultCacheDriver;
 
@@ -46,19 +47,18 @@ class QueryCacheProfile
     private $cacheKey;
 
     /**
-     * @param int                               $lifetime
-     * @param string|null                       $cacheKey
-     * @param \Doctrine\Common\Cache\Cache|null $resultCache
+     * @param int         $lifetime
+     * @param string|null $cacheKey
      */
-    public function __construct($lifetime = 0, $cacheKey = null, Cache $resultCache = null)
+    public function __construct($lifetime = 0, $cacheKey = null, ?Cache $resultCache = null)
     {
-        $this->lifetime = $lifetime;
-        $this->cacheKey = $cacheKey;
+        $this->lifetime          = $lifetime;
+        $this->cacheKey          = $cacheKey;
         $this->resultCacheDriver = $resultCache;
     }
 
     /**
-     * @return \Doctrine\Common\Cache\Cache|null
+     * @return Cache|null
      */
     public function getResultCacheDriver()
     {
@@ -76,7 +76,7 @@ class QueryCacheProfile
     /**
      * @return string
      *
-     * @throws \Doctrine\DBAL\Cache\CacheException
+     * @throws CacheException
      */
     public function getCacheKey()
     {
@@ -102,7 +102,7 @@ class QueryCacheProfile
         $realCacheKey = 'query=' . $query .
             '&params=' . serialize($params) .
             '&types=' . serialize($types) .
-            '&connectionParams=' . serialize($connectionParams);
+            '&connectionParams=' . hash('sha256', serialize($connectionParams));
 
         // should the key be automatically generated using the inputs or is the cache key set?
         if ($this->cacheKey === null) {
@@ -111,11 +111,10 @@ class QueryCacheProfile
             $cacheKey = $this->cacheKey;
         }
 
-        return array($cacheKey, $realCacheKey);
+        return [$cacheKey, $realCacheKey];
     }
 
     /**
-     * @param \Doctrine\Common\Cache\Cache $cache
      *
      * @return \Doctrine\DBAL\Cache\QueryCacheProfile
      */
--- a/tests/Doctrine/Tests/DBAL/Cache/QueryCacheProfileTest.php
+++ b/tests/Doctrine/Tests/DBAL/Cache/QueryCacheProfileTest.php
@@ -5,14 +5,43 @@ namespace Doctrine\Tests\DBAL\Cache;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
 use Doctrine\DBAL\ParameterType;
 use Doctrine\Tests\DbalTestCase;
+use function parse_str;
 
 class QueryCacheProfileTest extends DbalTestCase
 {
-    const LIFETIME = 3600;
-    const CACHE_KEY = 'user_specified_cache_key';
+    private const LIFETIME  = 3600;
+    private const CACHE_KEY = 'user_specified_cache_key';
 
-    /** @var QueryCacheProfile */
+    /**
+     * @var QueryCacheProfile
+     */
     private $queryCacheProfile;
+
+    /**
+     * @var string
+     */
+    private $query = 'SELECT * FROM foo WHERE bar = ?';
+
+    /**
+     * @var int[]
+     */
+    private $params = [666];
+
+    /**
+     * @var string[]
+     */
+    private $types = [ParameterType::INTEGER];
+
+    /**
+     * @var string[]
+     */
+    private $connectionParams = [
+            'dbname'   => 'database_name',
+            'user'     => 'database_user',
+            'password' => 'database_password',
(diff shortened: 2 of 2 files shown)


lib/Doctrine/DBAL/Cache/QueryCacheProfile.php
```diff
@@ -99,10 +101,14 @@ public function getCacheKey()
      */
     public function generateCacheKeys($query, $params, $types, array $connectionParams = [])
     {
+        $hashedParams = array_map(function ($param) {
+            return crypt($param, md5($param));
```
morozov: This approach assumes that all `$param`s are scalar which may be true for now but not necessarily true in the future. Why not just hash the serialized representation once instead of hashing parameters individually?

morozov: I'm not a security expert but `md5($param)` is not random (the salt should be random). Should something like `random_bytes()` be used instead, or can we just get away with something like `md5()` or `sha1()` for this purpose?

fullbl: First one seems legit, since I don't know where is it really called I didn't want to change too much stuff. If it's ok I'll change it.
Salt is not random because with same strings we want same outputs!

Ocramius (changes requested):

lib/Doctrine/DBAL/Cache/QueryCacheProfile.php
```diff
@@ -20,18 +20,19 @@
 namespace Doctrine\DBAL\Cache;
 
 use Doctrine\Common\Cache\Cache;
+use function serialize;
+use function sha1;
```
Ocramius: Unused

lib/Doctrine/DBAL/Cache/QueryCacheProfile.php
```diff
@@ -102,7 +102,7 @@ public function generateCacheKeys($query, $params, $types, array $connectionPara
         $realCacheKey = 'query=' . $query .
             '&params=' . serialize($params) .
             '&types=' . serialize($types) .
-            '&connectionParams=' . serialize($connectionParams);
+            '&connectionParams=' . hash('sha256', serialize($connectionParams));
```
Ocramius: :+1:

tests/Doctrine/Tests/DBAL/Cache/QueryCacheProfileTest.php
```diff
@@ -5,65 +5,61 @@
...
-        $types  = [ParameterType::INTEGER];
+    /** @var array<int> */
+    private $params = [666];
 
-        $connectionParams = array(
+    /** @var array<string> */
+    private $types = [ParameterType::INTEGER];
+
+    /** @var array<string> */
+    private $connectionParams = [
```
Ocramius: Alignment went a bit bonkers here. Also, the type is `string[]`, as most tooling doesn't work with `array<foo>`

tests/Doctrine/Tests/DBAL/Cache/QueryCacheProfileTest.php
```diff
@@ -5,65 +5,61 @@
...
-    public function testShouldUseTheGivenCacheKeyIfPresent()
-    {
-        $query  = 'SELECT * FROM foo WHERE bar = ?';
-        $params = [666];
-        $types  = [ParameterType::INTEGER];
+    /** @var array<int> */
+    private $params = [666];
 
-        $connectionParams = array(
+    /** @var array<string> */
```
Ocramius: `string[]`

tests/Doctrine/Tests/DBAL/Cache/QueryCacheProfileTest.php
```diff
@@ -5,65 +5,61 @@
...
-    }
+    /** @var string */
+    private $query = 'SELECT * FROM foo WHERE bar = ?';
 
-    public function testShouldUseTheGivenCacheKeyIfPresent()
-    {
-        $query  = 'SELECT * FROM foo WHERE bar = ?';
-        $params = [666];
-        $types  = [ParameterType::INTEGER];
+    /** @var array<int> */
```
Ocramius: `int[]`

tests/Doctrine/Tests/DBAL/Cache/QueryCacheProfileTest.php
```diff
@@ -5,65 +5,61 @@
...
 use Doctrine\DBAL\ParameterType;
 use Doctrine\Tests\DbalTestCase;
+use function parse_str;
+use function strpos;
 
 class QueryCacheProfileTest extends DbalTestCase
 {
-    const LIFETIME = 3600;
-    const CACHE_KEY = 'user_specified_cache_key';
+    public const LIFETIME  = 3600;
```
Ocramius: Can be `private`

tests/Doctrine/Tests/DBAL/Cache/QueryCacheProfileTest.php
```diff
@@ -5,65 +5,61 @@
...
+    public const CACHE_KEY = 'user_specified_cache_key';
```
Ocramius: `private`

tests/Doctrine/Tests/DBAL/Cache/QueryCacheProfileTest.php
```diff
@@ -77,67 +73,64 @@ public function testShouldGenerateAnAutomaticKeyIfNoKeyHasBeenGiven()
 
     public function testShouldGenerateDifferentKeysForSameQueryAndParamsAndDifferentConnections()
     {
-        $query  = 'SELECT * FROM foo WHERE bar = ?';
```
Ocramius: Some in the diffs in the test seem completely wrong/unrelated. Also, as I mentioned before, a new method verifying that the connection parameters cannot be found in the key is needed.

lib/Doctrine/DBAL/Cache/QueryCacheProfile.php
fullbl: lines 103, 104 and 109!

Ocramius: Aha, I just saw the diff, thanks :+1:

tests/Doctrine/Tests/DBAL/Cache/QueryCacheProfileTest.php
```diff
@@ -5,65 +5,61 @@
 use Doctrine\DBAL\Cache\QueryCacheProfile;
 use Doctrine\DBAL\ParameterType;
 use Doctrine\Tests\DbalTestCase;
+use function parse_str;
+use function unserialize;
 
 class QueryCacheProfileTest extends DbalTestCase
```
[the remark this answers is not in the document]
fullbl: testConnectionParamsShouldBeHashed

tests/Doctrine/Tests/DBAL/Cache/QueryCacheProfileTest.php
```diff
@@ -77,67 +83,64 @@ public function testShouldGenerateAnAutomaticKeyIfNoKeyHasBeenGiven()
...
+            $this->connectionParams
         );
 
+        $params = [];
+        parse_str($queryString, $params);
+
+        self::assertArrayHasKey('connectionParams', $params);
+
+        foreach ($this->connectionParams as $param) {
+            self::assertEquals(false, strpos($param, $params['connectionParams']));
```
Ocramius: Is there something like `assertStringNotContains` that you can use here?

tests/Doctrine/Tests/DBAL/Cache/QueryCacheProfileTest.php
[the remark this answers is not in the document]
Ocramius: Thanks! Didn't see it due to diff noise :)

.gitignore
```diff
@@ -8,3 +8,4 @@ vendor/
 composer.lock
 /phpunit.xml
 /.phpcs-cache
+.idea
```
Ocramius: This should be in your [global `.gitignore`](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c), and not in the project

tests/Doctrine/Tests/DBAL/Cache/QueryCacheProfileTest.php
fullbl: Yes, I tried to make it more dry!

Ocramius (dismissed):

Ocramius:
Some CS issues remain, but I'll not hassle you with them :)

lib/Doctrine/DBAL/Cache/QueryCacheProfile.php
```diff
@@ -20,18 +20,19 @@
 namespace Doctrine\DBAL\Cache;
 
 use Doctrine\Common\Cache\Cache;
+use function hash;
+use function serialize;
+use function sha1;
```
Majkl578: This one is now redundant.

Majkl578 (approved):

jwage (approved):

Ocramius:
:ship:

Ocramius:
Thanks @fullbl!

PR #3033, opened 2018-02-28
Added support for available DSN parameters for the PDOSqlsrv driver

aashmelev:
This is done to transfer parameters to the DSN.
The available parameters are described by link http://php.net/manual/en/ref.pdo-sqlsrv.connection.php


aashmelev:
At the moment, there is no way to pass parameters to the driver.

Please review this pull-request.

morozov:
@aashmelev thank you for the patch. From what I see, the non-PDO SQL Server driver supports the same DSN parameters, so it might make sense to move this logic to `AbstractSQLServerDriver`.

Besides that, it would be great if we could avoid mixing DBAL-specific parameters like `dbname` and platform-specific ones in the same namespace (`$params`). Doing so requires listing all known platform-specific parameters to distinct between them. E.g. `sqlsrv_connect()` takes `$connectionOptions` as a seaprate parameter. So could try doing we.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Driver.php
```diff
@@ -61,16 +64,65 @@ private function _constructPdoDsn(array $params)
...
+
+    /**
+     * Returns available DSN params for driver
+     *
+     * @return string[] Available DSN params
+     */
+    private function getAvailableDsnParam()
+    {
+        return [
+            'APP',
```
morozov: https://pecl.php.net/package-changelog.php?package=sqlsrv&release=5.2.0 introduces the `ColumnEncryption` parameter which indicates that it's better if we don't maintain this list ourselves at all.

aashmelev:
@morozov thanks for the answer! I changed the code.

The problem is that the `pdo_sqlsrv` driver receives the connection parameters only via the `dsn` and it does not receive the parameters via the `driverOptions` parameter like the `sqlsrv` driver.

It's not working for the `pdo_sqlsrv` driver (https://docs.microsoft.com/en-us/sql/connect/php/pdo-construct?view=sql-server-2017):
```
DriverManager::getConnection([
    'url' => 'mssql://user:password@server/dbname,
    'driverOptions' => [
        'APP' => 'APP_NAME',
    ]
]);
```

But it's working for the `sqlsrv` driver (https://docs.microsoft.com/en-us/sql/connect/php/sqlsrv-connect?view=sql-server-2017):
```
DriverManager::getConnection([
    'url' => 'sqlsrv://user:password@server/dbname,
    'driverOptions' => [
        'APP' => 'APP_NAME',
    ]
]);
```
I suggest adding the new parameter `connectionOptions` to pass the connection options (https://docs.microsoft.com/en-us/sql/connect/php/connection-options?view=sql-server-2017), and then passing them through the `dsn` into the `PDO` instance:
```
DriverManager::getConnection([
    'url' => 'mssql://user:password@server/dbname,
    'connectionOptions' => [
        'APP' => 'APP_NAME',
    ]
]);
```

morozov:
Why not use the `driverOptions` key which is supported by the wrapper connection generically and is part of the `Driver::connect()` inteface? I mean, instead of introducing a new parameter, we start properly supporting the existing one.

aashmelev:
I thought about it, but then we'll have to keep a list of connection options. And this is not very good, as you wrote about it above. Or, to distinguish them, is it enough that PDO constants have a numeric key?

morozov:
To summarize the state we're in: there's already an inconsistency in how we structure connection and driver options in existing drivers:

1. In `MysqliConnection`, the `$driverOptions` argument accepts the options which `mysqli::set_options()` accepts and the `flags` key which may contain the flags which will be passed to `mysqli::real_connect`.
2. In all PDO-based drivers, `$driverOptions` are the options which will be passed to the PDO constructor.
3. In the SQL Server driver, `$driverOptions` are connection options like "APP", "MultipleActiveResultSets", etc.

In the PDO SQL Server driver, we need to combine both 2 and 3.

> Or, to distinguish them, is it enough that PDO constants have a numeric key?

I think it's better than introducing a new option. This way, the PDO SQL Server configuration will look the same as the non-PDO SQL Server (will contain dsn-specific parameters) and will look like other PDO configurations (will contain `PDO::` constants).

I think, in the future, every driver will need to have more than one `$driverOptions` parameter specific to their nature. E.g. PDO-based drivers will have explicit `$pdoOptions`, etc.

aashmelev:
I made changes to the code.
It is important to note that the `MultipleActiveResultSets` parameter is no longer accepted via the DSN, but it is only accepted through the driver options. This breaks backward compatibility.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Driver.php
```diff
@@ -33,22 +35,25 @@ class Driver extends AbstractSQLServerDriver
      */
     public function connect(array $params, $username = null, $password = null, array $driverOptions = [])
     {
+        $splitedOptions = $this->splitOptions($driverOptions);
```
morozov: It should be `$splitOptions` (irregular verb).

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Driver.php
```diff
@@ -61,16 +66,58 @@ private function _constructPdoDsn(array $params)
         }
 
         if (isset($params['dbname'])) {
-            $dsn .= ';Database=' .  $params['dbname'];
+            $dsn .= ';Database=' . $params['dbname'];
         }
 
-        if (isset($params['MultipleActiveResultSets'])) {
-            $dsn .= '; MultipleActiveResultSets=' . ($params['MultipleActiveResultSets'] ? 'true' : 'false');
+        if (! empty($connectionOptions)) {
```
morozov: This condition seems redundant. If the method is invoked with an empty set, it will return an empty string which is acceptable.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Driver.php
```diff
@@ -61,16 +66,58 @@ private function _constructPdoDsn(array $params)
...
+        foreach ($options as $optionKey => $optionValue) {
+            if (is_int($optionKey)) {
+                $driverOptions[$optionKey] = $optionValue;
+            } else {
+                $connectionOptions[$optionKey] = $optionValue;
+            }
+        }
+
+        return [
+            'driverOptions' => $driverOptions,
```
morozov: Given it's a private API, instead of introducing named keys, would it be cleaner to just use a tuple? Like:
```php
function splitOptions()
{
    return [$connectionOptions, $driverOptions];
}

[$connectionOptions, $driverOptions] = splitOptions();
```

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Driver.php
```diff
@@ -61,16 +66,58 @@ private function _constructPdoDsn(array $params)
...
+        ];
+    }
+
+    /**
+     * Converts a connection options array to the DSN
+     *
+     * @param string[] $connectionOptions
+     * @return string
+     */
+    private function getConnectionOptionsDsn(array $connectionOptions)
```
morozov: Please add a return type instead of the annotation.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Driver.php
```diff
@@ -61,16 +66,58 @@ private function _constructPdoDsn(array $params)
...
         return $dsn;
     }
 
+    /**
+     * Separates a connection options from a driver options
+     *
+     * @param array $options
+     * @return array
+     */
+    private function splitOptions(array $options)
```
morozov: Please use a return type.

morozov (commented):
Please see a few more comments. Besides those, could you think of a way of testing the changes? Preferably, it should be a functional test showing that both PDO driver options and SQL Server connection options are effective and therefore the patch works as expected.

E.g., setting `WSID` to some value and then executing `HOST_NAME()` to see if it reached the server (see the [article](https://docs.microsoft.com/en-us/sql/relational-databases/native-client/applications/using-connection-string-keywords-with-sql-server-native-client?view=sql-server-2017)).

aashmelev:
Okay, I'll do the functional tests as you wrote about it above.

aashmelev:
l added tests

tests/Doctrine/Tests/DBAL/Driver/PDOSqlsrv/DriverTest.php
```diff
@@ -16,4 +17,48 @@ protected function createDriver()
...
+            $driverOptions
+        );
+    }
+
+    /**
+     * @throws \PHPUnit_Framework_SkippedTestError
+     */
+    private function skipWhenNotUsingPdoSqlsrv()
+    {
+        if (isset($GLOBALS['db_type']) && $GLOBALS['db_type'] === 'pdo_sqlsrv') {
```
morozov: Please check connection platform or driver depending on what's being tested, not the global variables.

morozov (commented):
Please see some more comments and revert the irrelevant changes in `UPGRADE.md`.

UPGRADE.md
```diff
@@ -12,15 +12,6 @@
 
 ``Doctrine\DBAL\Connection::TRANSACTION_*`` were moved into ``Doctrine\DBAL\TransactionIsolationLevel`` class without the ``TRANSACTION_`` prefix. 
 
-## DEPRECATION: direct usage of the PDO APIs in the DBAL API
```
morozov: This shouldn't have been removed. Also, the deprecation note on using `MultipleActiveResultSets` seems mistakenly gone.

tests/Doctrine/Tests/DBAL/Driver/AbstractSQLServerDriverTest.php
```diff
@@ -54,4 +56,40 @@ protected function getDatabasePlatformsForVersions()
             array('12', 'Doctrine\DBAL\Platforms\SQLServer2012Platform'),
         );
     }
+
+    protected function getDriver() : AbstractSQLServerDriver
+    {
+        return static::createDriver();
+    }
+
+    protected function checkForSkippingTest(AbstractSQLServerDriver $driver) : void
```
morozov: Please remove this method. The abstract class shouldn't use any knowledge of its descendants.

tests/Doctrine/Tests/DBAL/Driver/AbstractSQLServerDriverTest.php
```diff
@@ -54,4 +56,40 @@ protected function getDatabasePlatformsForVersions()
             array('12', 'Doctrine\DBAL\Platforms\SQLServer2012Platform'),
         );
     }
+
+    protected function getDriver() : AbstractSQLServerDriver
```
morozov: Please remove this method. You can use `$this->_conn->getDriver()` instead.

tests/Doctrine/Tests/DBAL/Driver/PDOSqlsrv/DriverTest.php
```diff
@@ -16,4 +19,22 @@ protected function createDriver()
     {
         return new Driver();
     }
+
+    protected function checkForSkippingTest(AbstractSQLServerDriver $driver) : void
+    {
+        if (extension_loaded('pdo_sqlsrv') && $driver instanceof Driver) {
```
morozov: As long as this test uses a real connection, it should be part of the functional suite, where the availability of the extension will be checked in `setUp()`.

tests/Doctrine/Tests/DBAL/Driver/AbstractSQLServerDriverTest.php
```diff
@@ -54,4 +56,40 @@ protected function getDatabasePlatformsForVersions()
...
+            $driverOptions
+        );
+    }
+
+    public function testConnectionOptions() : void
+    {
+        $driver = $this->getDriver();
+        $this->checkForSkippingTest($driver);
+        $connection = $this->getConnection($driver, ['APP' => 'APP_NAME']);
+        $result     = $connection->query('select APP_NAME() as app')->fetch();
```
morozov: You can use `fetchColumn()` instead of `fetch()` and remove the "app" alias for simplicity.

morozov: Please use upper case for SQL keywords for better readability (`'SELECT'`).

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Driver.php
```diff
@@ -61,16 +66,56 @@ private function _constructPdoDsn(array $params)
...
+        $dsn .= $this->getConnectionOptionsDsn($connectionOptions);
+
         return $dsn;
     }
 
+    /**
+     * Separates a connection options from a driver options
+     *
+     * @param mixed[] $options
+     * @return mixed[][]
```
morozov: I know @Majkl578 won't like these `mixed` annotation. Taking into account that most existing tools will not recognize annotations like `(int|string)[]`, would it be fair to replace them with `int[]|string[]` and `int[][]|string[][]` respectively?

Majkl578: Probably, better than mixed. :+1:

aashmelev:
Thanks for the comments!
I renamed `tests/Doctrine/Tests/DBAL/Functional/Driver/PDOSqlsrv/Driver.php` because it should be called DriverTest.php

aashmelev:
Perhaps it would be better if I make a separate PR to rename the file?

morozov:
@aashmelev, no need to make another PR for the rename. Let's keep everything in one place. If you want me to help you rename the file, rebase the PR or squash the commits, let me know.

aashmelev:
@morozov, I squashed the commits.

morozov (approved):
Thanks, @aashmelev!

PR #3034, opened 2018-02-28
Missing type hints in $query, $params, $types sets

simPod:


Ocramius (approved):
Looking good from here, although unsure if this will be a BC break (too little :coffee: and toi early in the morning)

Majkl578:
Wouldn't pursue similar changes until develop lands in master, too early in the DBAL 3.0 cycle.
Otherwise needs rebase and fixing CS.

Majkl578 (changes requested):
I still think we shouldn't do this until develop is merged into master. (Same as fixing CS globally.)

ostrolucky:
superseded with https://github.com/doctrine/dbal/pull/3348

PR #634, opened 2014-07-21
Introduce a Transaction object

BenMorel:
This is a proposal for a Transaction object, as suggested by @guilhermeblanco and @beberlei in doctrine/doctrine2#949; this is a foundation for the proposed ORM feature. This supersedes #571.

This PR makes the following changes to `Connection`:
- **Adds** the following method:
  - `getTransactionManager()` : returns the `TransactionManager`
- **Deprecates** the following methods, in favour of their counterparts on `Transaction`:
  - `commit()`
  - `rollBack()`
  - `setRollbackOnly()`
  - `isRollbackOnly()`
- **Deprecates** the following methods, in favour of their counterparts on `TransactionManager`:
  - `beginTransaction()`
  - `isTransactionActive()`
  - `getTransactionNestingLevel()`
  - `setNestTransactionsWithSavepoints()`
  - `getNestTransactionsWithSavepoints()`

The new way of dealing with transactions is now:

```
$transactionManager = $connection->getTransactionManager();
$transaction = $transactionManager->beginTransaction();
// ...
$transaction->commit();
```

A fluent API is available to configure a transaction:

```
$transaction = $transactionManager->createTransaction()
    ->withIsolationLevel(Connection::TRANSACTION_SERIALIZABLE)
    ->begin();
```

Calls to `commit()` and `rollback()` are automatically propagated to nested transactions:

```
$transaction1 = $transactionManager->beginTransaction();
$transaction2 = $transactionManager->beginTransaction();
$transaction1->commit(); // will commit $transaction2 then $transaction1
```

Overall, it's not a complicated change, does not introduce any BC break, and passes all existing tests.

PR #3041, opened 2018-03-08
Introduce a null SQL logger

BenMorel:
As suggested by @morozov in [#634](https://github.com/doctrine/dbal/pull/634#discussion_r172039417), this adds a null SQL logger and avoids wrapping every `start|stopQuery()` call in a `if ($logger)` condition.

I extracted this in a separate PR to avoid polluting too much the Transaction object PR.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -9,6 +9,10 @@ The Doctrine\DBAL\Version class is no longer available: please refrain from chec
 1. The support of `PDO::PARAM_*`, `PDO::FETCH_*`, `PDO::CASE_*` and `PDO::PARAM_INPUT_OUTPUT` constants in the DBAL API is removed.
 2. `\Doctrine\DBAL\Driver\PDOStatement` does not extend `\PDOStatement` anymore.
 
+## BC BREAK: the SQLLogger interface has changed
+
+The methods are the same but use scalar type hints, return types, and non-nullable arrays.
+
 Before:
 
     use Doctrine\DBAL\Portability\Connection;
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -12,6 +12,7 @@ The Doctrine\DBAL\Version class is no longer available: please refrain from chec
 ## BC BREAK: the SQLLogger interface has changed
 
 The methods are the same but use scalar type hints, return types, and non-nullable arrays.
+SQLLogger implementations are now final.
 
 Before:
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -9,11 +9,6 @@ The Doctrine\DBAL\Version class is no longer available: please refrain from chec
 1. The support of `PDO::PARAM_*`, `PDO::FETCH_*`, `PDO::CASE_*` and `PDO::PARAM_INPUT_OUTPUT` constants in the DBAL API is removed.
 2. `\Doctrine\DBAL\Driver\PDOStatement` does not extend `\PDOStatement` anymore.
 
-## BC BREAK: the SQLLogger interface has changed
-
-The methods are the same but use scalar type hints, return types, and non-nullable arrays.
-SQLLogger implementations are now final.
-
 Before:
 
     use Doctrine\DBAL\Portability\Connection;
@@ -44,6 +39,12 @@ After:
 ## BC BREAK: Removed Drizzle support
 
 The Drizzle project is abandoned and is therefore not supported by Doctrine DBAL anymore.
+
+## BC BREAK: SQLLogger changes
+
+- The SQLLogger interface has changed; the methods are the same but use scalar type hints, return types, and non-nullable arrays.
+- SQLLogger implementations: `DebugStack`, `EchoSQLLogger`, `LoggerChain` are now final.
+- `Configuration::getSQLLogger()` does not return `null` anymore, but a `NullLogger` implementation.
 
 # Upgrade to 2.7
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -42,9 +42,10 @@ The Drizzle project is abandoned and is therefore not supported by Doctrine DBAL
 
 ## BC BREAK: SQLLogger changes
 
-- The SQLLogger interface has changed; the methods are the same but use scalar type hints, return types, and non-nullable arrays.
-- SQLLogger implementations: `DebugStack`, `EchoSQLLogger`, `LoggerChain` are now final.
+- The `SQLLogger` interface has changed; the methods are the same but use scalar type hints, return types, and non-nullable arrays.
+- `SQLLogger` implementations: `DebugStack`, `EchoSQLLogger`, `LoggerChain` are now final.
 - `Configuration::getSQLLogger()` does not return `null` anymore, but a `NullLogger` implementation.
+- `Configuration::setSQLLogger()` does not allow `null` anymore.
 
 # Upgrade to 2.7
 
--- a/lib/Doctrine/DBAL/Configuration.php
+++ b/lib/Doctrine/DBAL/Configuration.php
@@ -28,8 +28,6 @@ class Configuration
 
     /**
      * Sets the SQL logger to use.
-     *
-     * @param SQLLogger $logger
      */
     public function setSQLLogger(SQLLogger $logger) : void
     {
--- a/lib/Doctrine/DBAL/Configuration.php
+++ b/lib/Doctrine/DBAL/Configuration.php
@@ -27,13 +27,11 @@ class Configuration
     protected $_attributes = [];
 
     /**
-     * Sets the SQL logger to use. Defaults to NULL which means SQL logging is disabled.
+     * Sets the SQL logger to use.
      *
-     * @param \Doctrine\DBAL\Logging\SQLLogger|null $logger
-     *
-     * @return void
+     * @param SQLLogger $logger
      */
-    public function setSQLLogger(SQLLogger $logger = null)
+    public function setSQLLogger(SQLLogger $logger) : void
     {
         $this->_attributes['sqlLogger'] = $logger;
     }
--- a/lib/Doctrine/DBAL/Configuration.php
+++ b/lib/Doctrine/DBAL/Configuration.php
@@ -45,11 +45,7 @@ class Configuration
      */
     public function getSQLLogger() : SQLLogger
     {
-        if (! isset($this->_attributes['sqlLogger'])) {
-            $this->_attributes['sqlLogger'] = new NullLogger();
-        }
-
-        return $this->_attributes['sqlLogger'];
+        return $this->_attributes['sqlLogger'] ?? $this->_attributes['sqlLogger'] = new NullLogger();
     }
 
     /**
--- a/lib/Doctrine/DBAL/Configuration.php
+++ b/lib/Doctrine/DBAL/Configuration.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\DBAL;
 
+use Doctrine\DBAL\Logging\NullLogger;
 use Doctrine\DBAL\Logging\SQLLogger;
 use Doctrine\Common\Cache\Cache;
 
@@ -40,11 +41,15 @@ class Configuration
     /**
      * Gets the SQL logger that is used.
      *
-     * @return \Doctrine\DBAL\Logging\SQLLogger|null
+     * @return \Doctrine\DBAL\Logging\SQLLogger
      */
-    public function getSQLLogger()
+    public function getSQLLogger() : SQLLogger
     {
-        return $this->_attributes['sqlLogger'] ?? null;
+        if (! isset($this->_attributes['sqlLogger'])) {
+            $this->_attributes['sqlLogger'] = new NullLogger();
+        }
+
+        return $this->_attributes['sqlLogger'];
     }
 
     /**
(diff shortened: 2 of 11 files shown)


lib/Doctrine/DBAL/Configuration.php
```diff
@@ -40,11 +41,15 @@ public function setSQLLogger(SQLLogger $logger = null)
...
      */
-    public function getSQLLogger()
+    public function getSQLLogger() : SQLLogger
     {
-        return $this->_attributes['sqlLogger'] ?? null;
+        if (! isset($this->_attributes['sqlLogger'])) {
+            $this->_attributes['sqlLogger'] = new NullLogger();
+        }
+
+        return $this->_attributes['sqlLogger'];
```
Majkl578:
```php
$this->_attributes['sqlLogger'] ?? $this->_attributes['sqlLogger'] = new NullLogger();
```
And you can drop the condition above.

lib/Doctrine/DBAL/Logging/NullLogger.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace Doctrine\DBAL\Logging;
+
+/**
+ * A SQL logger that does nothing.
+ */
+class NullLogger implements SQLLogger
```
Majkl578: Could be final.

lib/Doctrine/DBAL/Configuration.php
```diff
@@ -40,11 +41,15 @@ public function setSQLLogger(SQLLogger $logger = null)
     /**
      * Gets the SQL logger that is used.
      *
-     * @return \Doctrine\DBAL\Logging\SQLLogger|null
+     * @return \Doctrine\DBAL\Logging\SQLLogger
```
Majkl578: This should be dropped.

lib/Doctrine/DBAL/Logging/NullLogger.php
```diff
@@ -0,0 +1,23 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function startQuery($sql, array $params = null, array $types = null)
```
Majkl578: `?array $params`
`?array $types`

Majkl578 (dismissed):
:+1:

lib/Doctrine/DBAL/Logging/NullLogger.php
BenMorel: This requires changing the interface. And quite a lot of code ([1](https://github.com/doctrine/dbal/blob/develop/lib/Doctrine/DBAL/Connection.php#L1007), [2](https://github.com/doctrine/dbal/blob/develop/lib/Doctrine/DBAL/Connection.php#L1089), [3](https://github.com/doctrine/dbal/blob/develop/lib/Doctrine/DBAL/Connection.php#L1247), ...) only provides the first parameter, this would force them to call `startQuery($query, null, null)` which IMHO is not a very good idea.

Majkl578: No it doesn't (I am not suggesting to drop `= null`, only add `?` - also see CS violation on Travis). Nullable is implied by the interface specifying `null` as default value.
https://3v4l.org/mKITu

lib/Doctrine/DBAL/Configuration.php
BenMorel: Why remove the `@return` statement? Is there a coding standard that Doctrine follows? I tried to propose one a few years ago, but this lead nowhere at the time.

lib/Doctrine/DBAL/Logging/NullLogger.php
BenMorel: OK!

lib/Doctrine/DBAL/Logging/NullLogger.php
morozov: IMO there's no need for params and types to be nullable. We can default them to empty arrays and avoid dealing with nulls: `array $params = [], array $types = []`.

Otherwise, the nullable signature will allow calls like `startQuery('SQL', null, null)` which doesn't make any sense.

Majkl578: That is better, but a BC break.

lib/Doctrine/DBAL/Configuration.php
BenMorel: Why not, done.

lib/Doctrine/DBAL/Logging/NullLogger.php
morozov: True. Still worth doing separately for develop.

morozov: True. Still worth doing separately for develop.

BenMorel: Agreed, let's change the interface then.

lib/Doctrine/DBAL/Configuration.php
Majkl578: Yes, Doctrine projects use [Doctrine Coding Standard](https://github.com/doctrine/coding-standard), it has been improved a lot in recent months.
This is redundant and useless information (there is already native type), thus should be removed.

lib/Doctrine/DBAL/Logging/NullLogger.php
BenMorel: Done, and because we're changing the interface anyway, I also added `string` type hint and `void` return types.

lib/Doctrine/DBAL/Configuration.php
BenMorel: Thanks for the pointer, I missed it! Removed `@return` then.

lib/Doctrine/DBAL/Logging/NullLogger.php
BenMorel: Should I make the other logger classes `final`, by the way?

lib/Doctrine/DBAL/Logging/NullLogger.php
Majkl578: > because we're changing the interface anyway

Well, adding type for parameter is not a BC break (we require PHP 7.2), but adding void is. :) But should be okay for develop. Please document it in UPGRADE.md.

BenMorel: Now is the time for BC breaks! I added a note to `UPGRADE.md`.

morozov:
Not sure if we care about test coverage much, but `NullLogger` should be excluded from coverage (`@codeCoverageIgnore`). Otherwise, we need a test :-)

backbone87:
Imho we should make the SQLLogger dependant on a PSR Logger. The SQL logger "just" adds 2 utility methods that can internally use a normal PSR Logger. Then we dont need to reimplement something like a NullLogger because we can just use the NullLogger of PSR and decorate it with the SQLLogger.

morozov:
@backbone87 yes, it was already suggested in https://github.com/doctrine/dbal/pull/2911#issuecomment-343755819. Looks out of scope to me.

Majkl578:
Imho SQL logger and PSR logger are two entirely different concepts. It is perfectly fine to have PsrSqlLogger that logss queries into PSR-compatible logger though.

morozov:
@Majkl578 on that note, a feature I'd love to see in the SQL logger is logging queries with inlined parameters (like MySQL's query log does). This is very useful for being able to copy the query from the log and then play with it in a SQL client. I use some helper functions for inlining on a daily basis which saves me a lot of time.

backbone87:
> Imho SQL logger and PSR logger are two entirely different concepts. It is perfectly fine to have PsrSqlLogger that logs queries into PSR-compatible logger though.

I would keep the SQLLogger interface, but would only offer 2 included impls: PsrSqlLogger (side note: whats with the Code style here? PSRSQLLogger?) and the DebugStack logger

morozov:
Naming-wise, I'd say the "SQL" part could be entirely omitted as the DBAL is not expected to log anything else than SQL.

BenMorel:
Not sure how to test `NullLogger` as there are no assertions to make. I marked it as `@codeCoverageIgnore`.

Should I rename `SQLLogger` to `Logger` then?
And should I make all logger classes final?

morozov:
@BenMorel testing `NullLogger` was a joke, sorry. `@codeCoverageIgnore` is fine.

I'm all for `final` classes. Not sure about renaming. Just not introducing "SQL" in new class names should be fine.

BenMorel:
@morozov Maybe we should rename `EchoSQLLogger` to `EchoLogger` at least?

morozov:
Once the PSR logger is introduced, I'd remove `EchoSQLLogger` entirely. The same functionality can be achieved by using the PSR one.

BenMorel:
OK then! Is anybody working on the PSR logger?

morozov:
@BenMorel there's no issue or pull request to my knowledge. Please feel free to file one so that we can track it there.

UPGRADE.md
```diff
@@ -9,6 +9,11 @@ The Doctrine\DBAL\Version class is no longer available: please refrain from chec
 1. The support of `PDO::PARAM_*`, `PDO::FETCH_*`, `PDO::CASE_*` and `PDO::PARAM_INPUT_OUTPUT` constants in the DBAL API is removed.
 2. `\Doctrine\DBAL\Driver\PDOStatement` does not extend `\PDOStatement` anymore.
 
+## BC BREAK: the SQLLogger interface has changed
```
Majkl578: Finalizing implementors needs to be documented too.

lib/Doctrine/DBAL/Configuration.php
```diff
@@ -39,12 +40,10 @@ public function setSQLLogger(SQLLogger $logger = null)
...
     /**
      * Gets the SQL logger that is used.
-     *
-     * @return \Doctrine\DBAL\Logging\SQLLogger|null
      */
-    public function getSQLLogger()
+    public function getSQLLogger() : SQLLogger
     {
-        return $this->_attributes['sqlLogger'] ?? null;
+        return $this->_attributes['sqlLogger'] ?? $this->_attributes['sqlLogger'] = new NullLogger();
```
Majkl578: This needs to be documented too, apart from dropping nullability.

Majkl578: Also setLogger should no longer be nullable or default to null.

UPGRADE.md
BenMorel: It's been done already in my [previous commit](https://github.com/doctrine/dbal/pull/3041/commits/6c5aacc91ab10571ce761f8895ff2fb07fa470cb).

lib/Doctrine/DBAL/Configuration.php
BenMorel: Actually calling `setSQLLogger(null)` really is a shortcut for `setSQLLogger(new NullLogger)` now, a quick way to remove the logger. I don't mind not allowing null, just wondering if it can be useful to anyone?

Majkl578: Yes I understand, but it doesn't make much sense. The logger is no longer nullable, you can't _remove it_.

Majkl578: If anything, defaulting to null is nonsense, calling `$configuration->setSQLLogger()` is just bad design.

BenMorel: Sure, at least it would need to be `?SQLLogger`. But I'm OK to disallow `null` altogether, that's done. I also updated `UPGRADE.md`, and added `void` return type to `setSQLLogger()`.

Majkl578:
(Note: Once merged, this will likely break ORM tests.)

BenMorel:
I can fix the ORM tests after the merge, if needed.

Ocramius (dismissed):
LGTM :+1:

Ocramius:
@Majkl578 PSR-3 based SQL logger is something I'd drag in later

Ocramius:
@BenMorel few minors remain:

```
FILE: lib/Doctrine/DBAL/Configuration.php
----------------------------------------------------------------------
----------------------------------------------------------------------
 32 | ERROR | [x] Method \Doctrine\DBAL\Configuration::setSQLLogger()
    |       |     has useless @param annotation for parameter
    |       |     $logger.
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
FILE: lib/Doctrine/DBAL/Logging/SQLLogger.php
----------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
----------------------------------------------------------------------
 21 | ERROR | @param annotation of method
    |       | \Doctrine\DBAL\Logging\SQLLogger::startQuery() does not
    |       | specify type hint for items of its traversable
    |       | parameter $params.
 22 | ERROR | @param annotation of method
    |       | \Doctrine\DBAL\Logging\SQLLogger::startQuery() does not
    |       | specify type hint for items of its traversable
    |       | parameter $types.
----------------------------------------------------------------------
```

BenMorel:
For the traversable type hints, it's hard to say as `$params` can have any type (int, string, even object when using an appropriate Type). Does not it make sense to document `mixed[]` as opposed to `array`? 

For `$types`, is it always strings?

By the way, does the CS expect `type[]` or `array<type>`?

Ocramius:
@BenMorel `mixed[]` would do. `array<mixed>` is not recognized by most tooling

Ocramius:
> For `$types`, is it always strings?

Yep: `string[]`

BenMorel:
OK, done!

Ocramius (approved):
Thanks, @BenMorel!

PR #3042, opened 2018-03-09
Don't catch `Exception` anymore, as `Throwable` is already handled

BenMorel:
This was only needed for PHP 5.x.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1151,7 +1151,6 @@ class Connection implements DriverConnection
      *
      * @return mixed The value returned by $func
      *
-     * @throws Exception
      * @throws Throwable
      */
     public function transactional(Closure $func)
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Connection.php
```diff
@@ -1151,7 +1151,6 @@ public function lastInsertId($seqName = null)
      *
      * @return mixed The value returned by $func
      *
-     * @throws Exception
      * @throws Throwable
      */
     public function transactional(Closure $func)
```
morozov: In the light of dropping PHP 5 related code, should this be changed to `callable` instead of `Closure`?

BenMorel: Doesn't hurt, but `callable` was available since PHP 5.4!

Ocramius: Let's move that to a separate PR if needed :+1:

Ocramius (approved):
:+1:

morozov:
Superseded by https://github.com/doctrine/dbal/pull/4253 in `2.11.0`.

PR #3044, opened 2018-03-13
Functional test for allowing dynamic intervals in date sub/add

AshleyDawson:
Use catenation strategy to include either a literal interval or a dynamic interval from the database its self using a column reference

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
@@ -5,7 +5,9 @@ namespace Doctrine\Tests\DBAL\Functional;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\FetchMode;
 use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Platforms\SqlitePlatform;
 use Doctrine\DBAL\Platforms\TrimMode;
+use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Types\Type;
 use const CASE_LOWER;
 use const PHP_EOL;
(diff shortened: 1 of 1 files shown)


AshleyDawson:
This is actually how the DBAL MySQL platform works - just making SQLite consistent as I use SQLite for functional tests and they're failing because of this inconsistency.

morozov:
@AshleyDawson could you include a failing test which is fixed by your changes in the pull request?

AshleyDawson:
Have written format tests for all unit/operator permutations when performing date add/subtract arithmetic for the SQLite platform

morozov:
@AshleyDawson I was talking about functional tests which failed for you and helped discover the issue. The ones which are executed against a read DB, not just make assertions against SQL.

AshleyDawson:
@morozov Sorry, should have explained in a previous comment. A query containing "DATE_SUB", etc. with a column reference instead of a literal fails silently and doesn't produce any results - with no exception raised. Only by inspecting the generated SQL can you see where the issue is. Please could you let me know where the SQLite functional tests for "DATE_SUB", etc. are and I'll modify them to take into account dynamic intervals.

To explain the issue with examples:

**Current implementation, with literal (WORKS):**
```sql
... WHERE ((DATE(r0_.next_renewal_at,'-30 DAY') ...
```

**Current implementation, with column reference (FAILS):**
```sql
... WHERE ((DATE(r0_.next_renewal_at,'-r1_.renewal_risk_threshold_in_days DAY') ...
```
(Notice the embedding of the column reference within the interval string)

**New implementation, with literal (WORKS):**
```sql
... WHERE ((DATE(r0_.next_renewal_at,'-' || 30 || ' DAY') ...
```

**New implementation, with column reference (WORKS):**
```sql
... WHERE ((DATE(r0_.next_renewal_at,'-' || r1_.renewal_risk_threshold_in_days || ' DAY') ...
```

morozov:
@AshleyDawson the logic you describe is the test I'm asking for. `Doctrine\Tests\DBAL\Functional\DataAccessTest` seems a good place to put it in. The logic can be the following:

1. Create two records with different dates/intervals.
2. Create a query with the `WHERE` clause which one of them matches but the other doesn't.
3. Fetch the records.
4. Make sure one and only one of them is selected. This way, we'll validate the correctness of the SQL expression by means of a real DB and make sure this implementation survives any future rework without a regression. Furthermore, We'll make sure it works the same on other platforms which support that.

AshleyDawson:
@morozov I've now added a functional test as instructed to `tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php`

morozov (dismissed):
:+1:

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
```diff
@@ -561,6 +563,34 @@ public function testDateArithmetics()
         self::assertEquals('2004-01-01', date('Y-m-d', strtotime($row['sub_years'])), "Subtracting years should end up on 2004-01-01");
     }
 
+    public function testSqliteDateArithmeticWithDynamicInterval()
+    {
+        if (! $this->_conn->getDriver() instanceof PDOSqliteDriver) {
```
morozov: One last thing from me: please check platform instead of the driver since it's a platform-specific feature, not driver-specific.

AshleyDawson: Good spot, I've made that change now.

morozov:
@AshleyDawson looks good. Could you please rebase and squash all commits? The patch is quite straightforward and doesn't require all the commits to understand.

morozov (dismissed):

Ocramius:
@morozov to be ported back to `2.6`?

morozov:
> This branch is out-of-date with the base branch

@Ocramius do we really need this, especially combined auto-dismissal of approvals after rebase? We don't usually have many concurrent conflicting changes.

Porting to `2.6` is fine with me.

Ocramius:
No, I think we're fine without that check. I will remove it when home.

morozov:
@AshleyDawson sorry for the delay. Could you please rebase?

morozov:
Looks like the issue is already resolved by #3019. I'll merge the functional test.

morozov (approved):

PR #3045, opened 2018-03-13
When rendering SQL, only render the alias if it's different from the table name

morozov:
When DBAL is used as part of a higher level abstraction where queries are built dynamically, the table name and the alias in `SELECT`, `UPDATE` and `DELETE` queries may be defined independently and may end up being the same. In this case, the resulting SQL contains a redundant alias:
```php
$builder = new QueryBuilder($this->conn);
$builder->select('id')
    ->from('users', 'users');

// SELECT id FROM users users
```
…which is technically valid but not needed. In order to avoid that, the caller needs to compare `$table` and `$alias` before calling `$builder->from()` (in the case above) and pass `null` as the alias in the case if they are equal.

The caller logic could be simplified without an impact on generated SQL if the DBAL itself only rendered the alias if it's different from the table name.

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -1120,7 +1120,7 @@ class QueryBuilder
 
         // Loop through all FROM clauses
         foreach ($this->sqlParts['from'] as $from) {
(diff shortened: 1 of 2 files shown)


Ocramius (approved):
:+1:

PR #3048, opened 2018-03-14
Adds generation expression to MySQL57 getListTableColumns method

octavianparalescu:
The generation expression is very important for the table columns, it is actually the definition of virtual columns.

https://www.percona.com/blog/2016/03/04/virtual-columns-in-mysql-and-mariadb/

### DIFF
--- a/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
@@ -22,6 +22,9 @@ namespace Doctrine\DBAL\Schema;
 use Doctrine\DBAL\Platforms\MariaDb1027Platform;
 use Doctrine\DBAL\Platforms\MySqlPlatform;
 use Doctrine\DBAL\Types\Type;
+use function strpos;
+use function strstr;
+use function strtoupper;
 
 /**
  * Schema manager for the MySql RDBMS.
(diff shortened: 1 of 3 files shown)


Ocramius:
Requires integration test additions

OctavianParalescu:
@Ocramius Unfortunately I cannot seem to be able to add virtual columns using \Doctrine\DBAL\Schema\Column 's option array, it seems we are completely lacking virtual/stored columns support from MySQL 5.7.

Ocramius:
I'd run the DDL in a raw query, then verify what the introspection does.

lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
```diff
@@ -204,6 +204,18 @@ protected function _getPortableTableColumnDefinition($tableColumn)
             $column->setPlatformOption('collation', $tableColumn['collation']);
         }
 
+        // MySQL 5.7 specific https://dev.mysql.com/doc/refman/5.7/en/create-table-generated-columns.html
+        if (strpos($tableColumn['extra'], 'GENERATED') !== false) {
+            // extra's syntax in the case of a generated columns:
+            // <UPPERCASE_GENERATION_TYPE> GENERATED
+            $pieces = explode(' ', $tableColumn['extra']);
```
Ocramius: You only use `$pieces[0]` here. Maybe you'd rather use `[$generationType] = explode(' ', $tableColumn['extra']);`? Also, is there a limit in the `explode()` expression here? Do you care only about the first space? If so, add `2` as a parameter to `explode()`

Ocramius (dismissed):
This looks good to me, but I'm unsure about MySQL 5.7-specific details in the schema manager rather than in the platform. @morozov where do you think this would fit best?

lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
morozov: Or `$generationType = strstr($tableColumn['extra'], ' ', true);`?

morozov:
> I'm unsure about MySQL 5.7-specific details in the schema manager rather than in the platform. @morozov where do you think this would fit best?

It depends on what the separation of concerns between the platform and schema manager is. I think we can merge it as is and think it through later. E.g., having a MariaDB-specific code in MySQL schema manager is a bigger concern to me.

OctavianParalescu:
We also need to modify the docs at http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/schema-representation.html (10.1.1.3. Vendor specific options')

PR #3056, opened 2018-03-16
Set Appveyor to use PHP 7.2

photodude:
This makes it so the SQLSVR DLL packages install the correct version to work with PHP 7.2

In #2617 it appeared like there was a limitation preventing testing with PHP 7.2, That was based on an oversite in the appveyor set up vs the PHP packages. This addresses that oversite and allows testing on PHP 7.2

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,9 +9,9 @@ clone_depth: 1
 cache:
   - C:\ProgramData\chocolatey\bin -> .appveyor.yml
   - C:\ProgramData\chocolatey\lib -> .appveyor.yml
-  - c:\tools\php -> .appveyor.yml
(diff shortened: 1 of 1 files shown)


Majkl578 (dismissed):
:+1: 
Maybe we can leave one/two of them to run on 7.1?

photodude:
Running on a mixed php 7.1/7.2 setup would require a new environment variable to handle the VC versions.
One or two I would choose the lowest versions for PHP 7.1.

Majkl578:
Ok, let's just keep 7.2. We test different versions on Travis anyway.

Ocramius (approved):
Thanks @photodude! \o/

PR #3064, opened 2018-03-20
Allow some larger indexes to fulfil another index

hinrik:
This allows a larger index to fulfil another index when it doesn't
compromise the uniqueness constraint. This prevents Doctrine from
creating superfluous indexes on foreign keys in some cases. E.g.:

```sql
CREATE TABLE `location_picture` (
  `location_picture_id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `location_id` int(10) unsigned NOT NULL,
  `order_id` tinyint(3) unsigned NOT NULL,
  `type` enum('normal','wide') NOT NULL,
  `filename` varchar(255) NOT NULL,
  PRIMARY KEY (`location_picture_id`),
  KEY `location_id_order_id` (`location_id`, `order_id`)
);
```

For the above table, Doctrine would unnecessarily try to create an
index on `location_id` even though it's covered by the existing
`location_id_order_id` index.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Index.php
+++ b/lib/Doctrine/DBAL/Schema/Index.php
@@ -20,6 +20,7 @@
 namespace Doctrine\DBAL\Schema;
 
 use Doctrine\DBAL\Platforms\AbstractPlatform;
+use function count;
 
 class Index extends AbstractAsset implements Constraint
 {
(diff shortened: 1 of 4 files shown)


Ocramius:
@hinrik please provide test cases to support the change

hinrik:
Will do.

hinrik:
All good? The continuousphp check doesn't seem to want to report its status.

morozov:
@hinrik if you squash your commits and force-push, the build will restart.

hinrik:
There we go.

hinrik:
Anyone care to review this? :)

Issue #2369, opened 2016-04-20
SQLSrvStatement doesn't respect PDO::PARAM_STR when binding parameters


morozov:
``` php
$conn = Doctrine\DBAL\DriverManager::getConnection(array(
    'driver' => 'sqlsrv',
    'host' => 'localhost',
    'user' => 'user',
    'password' => 'passw0rd',
    'dbname' => 'test',
));

$conn->executeStatement('CREATE TABLE test(
    id varchar(36) NOT NULL,
    name nvarchar(60) NULL
)');
$conn->executeStatement('INSERT INTO test (id, name) VALUES(\'xxx\', \'Test\')');
$query = 'SELECT name FROM test WHERE id = ?';
$stmt = $conn->prepare($query);
$stmt->bindValue(1, 1, \Doctrine\DBAL\ParameterType::STRING);

try {
    var_dump($stmt->execute()->fetchOne());
} catch (Exception $e) {
    echo $e->getMessage(), PHP_EOL;
}

$conn->executeStatement('DROP TABLE test');

// Outputs: An exception occurred while executing 'SELECT name FROM test WHERE id = ?' with params [1]:
//
// SQLSTATE [22018, 245]: [Microsoft][ODBC Driver 11 for SQL Server][SQL Server]Conversion failed when converting the varchar value 'xxx' to data type int.
```

According to the [implementation of `SQLSrvStatement::bindParam()`](https://github.com/doctrine/dbal/blob/v2.5.4/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php#L141-L145), it only considers `PDO::PARAM_LOB` if it is specified as `$type` but ignores other types.

The expected behavior is to have SQL Server informed that the parameter is of a string type, not int.

slucas-nwx:
I'm wondering if the problem i'm encountering is the same. I'm trying to send a comma delimited string to my stored procedure but it fails. If i execute the procedure from Mssql directly it works, but if i execute it through doctrine, I get no results back.

`        $statement = $this->em
            ->getConnection()
            ->getWrappedConnection()
            ->prepare('EXEC [dbo].[GetCatalogItems]
                @CultureId = ?,
                @OrderTypeCountryId = ?,
                @TargetDateTime = ?,
                @strSKU = ?'
            );


        $statement->bindValue(1, $cultureId, ParameterType::INTEGER);
        $statement->bindValue(2, $orderTypeCountryId, ParameterType::INTEGER);
[...]

PR #3065, opened 2018-03-22
Respect ParameterType for string in preparing statement for SQLServer…

lkeijmel:
…, otherwise you can get an "Conversion failed when converting the varchar value 'xxx' to data type int.". This was reported in #2369


Ocramius:
Requires test case additions with the actual simulation of the failure.

lkeijmel:
@Ocramius all right, I have to find some time to add it. What is the best case? Expand a current testcase or add a new one? And what's the best place to add one?

Majkl578:
It should be a new test in tests/Doctrine/Tests/DBAL/Functional/Ticket. Have a look at other tests there for inspiration.

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
```diff
@@ -267,6 +267,13 @@ private function prepare()
                     SQLSRV_PHPTYPE_STREAM(SQLSRV_ENC_BINARY),
                     SQLSRV_SQLTYPE_VARBINARY('max'),
                 ];
+            } elseif ($this->types[$column] === ParameterType::STRING) {
+                $params[$column - 1] = [
+                    &$variable,
+                    SQLSRV_PARAM_IN,
+                    SQLSRV_PHPTYPE_STRING(SQLSRV_ENC_CHAR),
+                    SQLSRV_SQLTYPE_VARCHAR('max'),
```
morozov: The trick is, depending on the column type and/or value, it could be `SQLSRV_SQLTYPE_NVARCHAR`. We need a test which shows that this code works with non-ASCII strings.

lkeijmel:
@Majkl578 I've finally got some time again to work on this issue. Where can I find or request a official ticket number for this? Or is this the issue ID from github (2369)?

morozov:
@lkeijmel the official ticket number for this is #2369.

Ocramius:
Please rebase instead of merging master :+1:

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
```diff
@@ -49,6 +50,7 @@
 use function sqlsrv_rows_affected;
 use function SQLSRV_SQLTYPE_VARBINARY;
 use function stripos;
+use function SQLSRV_SQLTYPE_VARCHAR;
```
Majkl578: These are really declared as upper-case? Oh gosh, this extension is a mess (class names with `-` are worse though). :see_no_evil:

Ocramius: That's the best you can say about SQLSRV? :P

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
```diff
@@ -294,6 +296,15 @@ private function prepare()
                     ];
                     break;
 
+                case ParameterType::STRING:
+                    $params[$column - 1] = [
+                        &$variable,
+                        SQLSRV_PARAM_IN,
+                        SQLSRV_PHPTYPE_STRING(SQLSRV_ENC_CHAR),
+                        SQLSRV_SQLTYPE_VARCHAR('max'),
```
morozov: It could be either `SQLSRV_SQLTYPE_VARCHAR` or `SQLSRV_SQLTYPE_NVARCHAR` — that's the challenge. I don't know the exact consequences of providing an incorrect type, but in my inderstanding:

1. If the column is `NVARCHAR` but the parameter is bound as `VARCHAR`, then a unicode value stored in the variable may be corrupted (requires research).
2. If the column is `VARCHAR` but the parameter is bound as `NVARCHAR`, then the DB may not be able to use the index by this field because it will have to cast the stored value to the binding type.

Before this change is accepted, we need to see what the potential issues are and that we're not introducing additional issues.

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL2369Test.php
```diff
@@ -0,0 +1,115 @@
...
+class DBAL2369Test extends DbalFunctionalTestCase
+{
+    /**
+     * @throws DBALException
+     */
+    protected function setUp() : void
+    {
+        parent::setUp();
+
+        if ($this->_conn->getDatabasePlatform()->getName() !== 'mssql') {
```
morozov: It shouldn't be required. All positive scenarios should pass on all platforms.

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL2369Test.php
```diff
@@ -0,0 +1,115 @@
...
+            $this->markTestSkipped('Related to SQLSRV only');
+        }
+
+        /* @var $sm \Doctrine\DBAL\Schema\AbstractSchemaManager */
+        $sm = $this->_conn->getSchemaManager();
+        if ($sm->tablesExist(['dbal2369'])) {
```
morozov: Please do `$sm->dropAndCreateTable()` instead. The fact that the table exists doesn't mean it has the expected schema.

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL2369Test.php
```diff
@@ -0,0 +1,115 @@
...
+        $stmt  = $this->_conn->prepare($query);
+        $stmt->bindValue(1, 1, ParameterType::INTEGER);
+        $stmt->execute();
+
+        $ret = $stmt->fetch();
+
+        self::assertArrayHasKey('id', $ret);
+        self::assertEquals($ret['id'], 1);
+        self::assertArrayHasKey('textfield', $ret);
+        self::assertEquals($ret['textfield'], 'test');
```
morozov: You can use the following for brevity:
```php
self::assertArraySubset([
    'id' => 1,
    'textfield' => 'test',
], $ret);
```

morozov:
See https://github.com/doctrine/dbal/issues/2369#issuecomment-882196985,

Issue #3066, opened 2018-03-22
Missing parameter fetchArgument for fetch and fetchAll methods


akrz:
If I want to fetch a record and I want to use PDO::FETCH_CLASS, I can't do

```php
$example = $qb->execute()->fetch(\PDO::FETCH_CLASS, Example::class);
```

because the method signature only has one parameter:
```php
# Doctrine\DBAL\Statement
public function fetch($fetchMode = null) {...}
```

The fetchAll method actually has a second parameter in the signature to assign an argument
```php
# Doctrine\DBAL\Statement
public function fetchAll($fetchMode = null, $fetchArgument = 0) {...}
```
But the Interface only has one parameter:
```php
# Doctrine\DBAL\Driver\ResultStatement
public function fetchAll($fetchMode = null);
```
So it throws an annoying warning when I want to use
```php
$example = $qb->execute()->fetchAll(\PDO::FETCH_CLASS, Example::class);
```

The workaround is setting the fetch mode and then do a fetch/fetchAll but it should be possible to pass an argument directly to those methods.
```php
$stmt = $qb->execute();
$stmt->setFetchMode(\PDO::FETCH_CLASS, Example::class);
$example = $stmt->fetch();
```

Ocramius:
@morozov @deeky666 we should really really really disallow the `\PDO::FETCH_CLASS` abomination.

akrz:
@Ocramius Can I ask why?

Ocramius:
@akrz it sets property fields with something inconsistent with reflection behavior, and creates instances by skipping ctor calls as well. If you really want to go in that direction, then you want to use a standard hydrator and avoid the bugs/quirks of PDO, by using something built in userland: even if slower, it is not worth the pain to go with the PDO internals here.

morozov:
Please see https://github.com/doctrine/dbal/pull/2958#issuecomment-355372897 for more details.

morozov:
I'd remove it together with `$columnIndex` for `FETCH_COLUMN` and drop `...$args` entirely.

deeky666:
Not to mention all of those non-PDO drivers not/rarely supporting Class/Object hydration natively. We are currently doing basic "userland" hydration for those drivers just to keep compatibility with our driver API. This is something that should be done outside of DBAL. I wish we could get rid of fetch modes completely and simply provide associative arrays as results and leave everything else for other libraries/userland...

mihas2:
[...]

PR #3070, opened 2018-03-26
Dropped support of fetching objects and non-zero column

morozov:
See https://github.com/doctrine/dbal/issues/3066#issuecomment-375338776 and https://github.com/doctrine/dbal/pull/2958#issuecomment-355372897 for the source.

Summary:
1. `FetchMode::STANDARD_OBJECT` and `FetchMode::CUSTOM_OBJECT` are no longer supported.
2. `FetchMode::COLUMN` with a non-zero index is no longer supported.
3. Incompatible change in the `Connection::fetchColumn()` signature where the 3rd argument is now `$types`, not `$columnIndex`.

Reasoning:
1. The object fetching modes exist only for parity with PDO. Fetching an object from DBAL is a poor man solution to object hydration. It works only if the DB schema and the object's properties are tightly coupled together and makes decoupling harder (see [example](https://github.com/morozov/dbal/blob/78546ce9be08752d9ea2e621b9df816098969e3b/phpcs.xml.dist#L74-L76)).
2. It's not implemented in the MySQL driver and in the caching layer without demand which means that it doesn't bother consumers.
3. Fetching of a non-zero column is a sign of bad practice of selecting more data than needs to be fetched. Not implemented in `fetch()` and `fetchAll()` in any of the non-PDO drivers and in the caching layer.
4. Getting rid of the above lets us get rid of the variadic `$args` in fetch statement methods and therefore simplify the interface.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,13 @@
 # Upgrade to 3.0
 
+## BC BREAK: Dropped support for `FetchMode::CUSTOM_OBJECT` and `::STANDARD_OBJECT`
+
+Instead of fetching an object, fetch an array and map it to an object of the desired class.
+
+## BC BREAK: Dropped support for the `$columnIndex` argument in `ResultStatement::fetchColumn()`, other `ResultStatement::fetch*()` methods invoked with `FetchMode::COLUMN` and `Connection::fetchColumn()`.
+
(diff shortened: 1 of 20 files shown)


morozov:
Reported the AppVeyor build failure as https://github.com/appveyor/ci/issues/2204.

greg0ire (approved):

Issue #3005, opened 2018-01-30
Statement objects should throw exception in exceptional cases


morozov:
As it was pointed out in https://github.com/doctrine/dbal/pull/2546#discussion_r96259485, the current API contract doesn't allow some methods to throw exceptions even in exceptional situations.

For instance:

1. `ResultStatement::fetch()`, `::fetchColumn()` will return `false` (or empty array in case of `fetchAll()`) if there's no more data to fetch or the statement hasn't been executed yet. In the latter, it should throw an exception instead.
2. `ResultStatement::closeCursor()` should not return anything. Some underlying implementations (e.g. `mysqli`) have a `void` return type, for others (e.g. `ibm_db2` and `oci8`) it's just a formality (they both return false if the `$stmt` is not passed).
3. `Statement::bindParam()`, `::bindValue()` and `::execute()` return a boolean or throw exceptions depending on the driver (see `DB2Statement` or `MysqliStatement`) for example. These methods shouldn't return any value and should throw an exception in case of failure.
4. In `ResultCacheStatement` and `ArrayStatement`, calling `::fetchColumn()` with an invalid column index should throw an exception.
5. `ResultStatement::setFetchMode()` shouldn't return any value.

morozov:
@Ocramius is fixing this one worth breaking BC in your opinion?

Majkl578:
> is fixing this one worth breaking BC in your opinion?

I'd say it is. Returning false on error is PHP relic.
It can be implemented in backward- & forward-compatible way, i.e. by a temporary switch (2.x only) to enable exceptions in 2.x. Or even as Configuration option which will won't be possible to disable in 3.0.

Ocramius:
Yes, returning `false` on error is just problematic. If the exception type is specific enough, adding `try-catch` around it is trivial.

Majkl578:
What's the resolution here @morozov?

morozov:
It's partially resolved by the rework of the types. None of the methods return `false` to indicate the error but the exceptions are only thrown when the underlying indicates an error.

PR #3074, opened 2018-03-31
Fix last insert ID consistency across drivers

morozov:
Reopened #2765 from the personal repo (replaces #2648).

## TODO
- [x] Why should the ID be a `string`, not an `int`? Because on a 32-bit system, a numeric ID will be represented as a string by the driver (e.g. `mysqli`). In a general case, an extremely large ID will not fit the int range anyways.
- [x] Simplify the test to have a better score on Scrutinizer.
- [x] Why should we return a zero when the last inserted ID is unavailable, e.g. why not `NULL`? Because some drivers return `0` when there's no ID. Besides that, for the caller, there's no difference between `0` and `NULL`, they both mean "none".
- [x] Why do we need to suppress errors? Let's suppress them for BC. Error reporting will be improved in #3005.
- [ ] Address code review concerns.


morozov:
Looks like we have a connection leak in general and especially with PDO driver. Here's the number of MySQL connections obtained from `SHOW PROCESSLIST` on each test teardown:

|MySQLi|PDO MySQL|
|-----:|--------:|
|3|4|
|4|6|
|5|8|
|6|10|
|7|12|
|8|14|
|9|16|
|10|18|
|11|20|
|12|22|
|13|24|
|14|26|
|15|28|
|16|30|
|17|32|
|18|34|
|19|36|
|20|38|
|21|40|
|22|42|
|23|44|
|24|46|
|25|48|
|26|50|
|27|52|
|28|54|
|29|56|
|30|58|
|31|60|
|32|62|
|33|64|
|34|66|
|35|68|
|37|73|
|38|74|
|39|75|
|40|76|
|41|77|

morozov:
Almost there. Down to one failure on SQL Server.

tests/Doctrine/Tests/DBAL/Functional/LastInsertIdTest.php
```diff
@@ -29,7 +29,7 @@ protected function tearDown()
             $this->_conn->getSchemaManager()->dropTable('last_insert_id_table');
         }
 
-        return parent::tearDown();
+        parent::tearDown();
```
Ocramius: `setUp` and `tearDown` should be marked as `void` in here, and we should add type hints too (doing a sweep for low hanging issues with the patch)

morozov:
It's green now, @Ocramius, @Majkl578.

lib/Doctrine/DBAL/Driver/LastInsertId.php
```diff
@@ -0,0 +1,28 @@
...
+{
+    /** @var string */
+    private $value = '0';
+
+    public function get() : string
+    {
+        return $this->value;
+    }
+
+    public function register(string $value) : void
```
Majkl578: I'm not a fan of having this crate mutable.

lib/Doctrine/DBAL/Driver/PDOConnection.php
```diff
@@ -32,6 +31,9 @@
  */
 class PDOConnection extends PDO implements Connection, ServerInfoAwareConnection
 {
+    /** @var string */
+    private $lastInsertId = '0';
```
Majkl578: This default value + doc block don't seem valid since you assign LastInsertId below.

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
```diff
@@ -154,11 +158,7 @@ public function __construct($conn, $sql, LastInsertId $lastInsertId = null)
     {
         $this->conn = $conn;
         $this->sql = $sql;
-
-        if (stripos($sql, 'INSERT INTO ') === 0) {
-            $this->sql .= self::LAST_INSERT_ID_SQL;
-            $this->lastInsertId = $lastInsertId;
-        }
+        $this->lastInsertId = $lastInsertId;
```
Majkl578: Shouldn't this use Driver\LastInsertId too instead of Driver\SQLSrv\LastInsertId? For BC, Driver\SQLSrv\LastInsertId can exend Driver\LastInsertId and be deprecated.
[same for SQLSrvConnection]

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
```diff
@@ -413,4 +409,23 @@ public function rowCount()
     {
         return sqlsrv_rows_affected($this->stmt);
     }
+
+    private function trackLastInsertId() : void
+    {
+        if (! $this->lastInsertId) {
```
Majkl578: `=== null`

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php
```diff
@@ -155,7 +170,7 @@ public function lastInsertId($name = null)
             $stmt = $this->query('SELECT @@IDENTITY');
         }
 
-        return $stmt->fetchColumn();
+        return $stmt->fetchColumn() ?: '0';
```
Majkl578: I don't like`?:` here, it hides possible _error values_ (would that be `false`, `'0'` or `0` here?) - is it possible to use `!== false` for example?

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php
```diff
@@ -140,7 +143,19 @@ public function exec($statement)
...
 
-        return sqlsrv_rows_affected($stmt);
+        $affectedRowCount = sqlsrv_rows_affected($stmt);
+
+        $stmt = sqlsrv_query($this->conn, 'SELECT @@IDENTITY');
+
+        if ($statement !== false) {
+            sqlsrv_fetch($stmt);
+
+            $lastInsertId = sqlsrv_get_field($stmt, 0) ?: '0';
```
Majkl578: I don't like`?:` here, it hides possible _error values_ (would that be `false`, `'0'` or `0` here?) - is it possible to use `!== false` for example?

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php
```diff
@@ -73,6 +75,7 @@ public function __construct($serverName, $connectionOptions)
             throw SQLSrvException::fromSqlSrvErrors();
         }
         $this->lastInsertId = new LastInsertId();
```
Majkl578: Shouldn't this use Driver\LastInsertId too instead of Driver\SQLSrv\LastInsertId? For BC, Driver\SQLSrv\LastInsertId can exend Driver\LastInsertId and be deprecated.

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereConnection.php
```diff
@@ -155,13 +155,13 @@ public function getServerVersion()
     /**
      * {@inheritdoc}
      */
-    public function lastInsertId($name = null)
+    public function lastInsertId($name = null) : string
     {
         if (null === $name) {
-            return sasql_insert_id($this->connection);
+            return (string) sasql_insert_id($this->connection);
```
Majkl578: Looks BC break-ish, what returned `false` before would now return `''`.

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereConnection.php
```diff
@@ -155,13 +155,13 @@ public function getServerVersion()
...
         }
 
-        return $this->query('SELECT ' . $name . '.CURRVAL')->fetchColumn();
+        return (string) $this->query('SELECT ' . $name . '.CURRVAL')->fetchColumn();
```
Majkl578: Looks BC break-ish, what returned `false` before would now return `''`.

lib/Doctrine/DBAL/Driver/PDOConnection.php
```diff
@@ -133,4 +138,55 @@ public function requiresQueryForServerVersion()
...
+     * Tracks the last insert ID at the current state.
+     *
+     * If this PDO driver is not able to fetch the last insert ID for identity columns
+     * without influencing connection state or transaction state, this is a noop method.
+     *
+     * @internal this method is only supposed to be used in DBAL internals
+     *
+     * @throws \PDOException
+     */
+    public function trackLastInsertId() : void
```
Majkl578: The whole design decision behind this method is a bit shady - making subclasses override/disable this behavior looks a bit weird. Would love to see this extracted into some InsertIdStrategyTracking with multiple providers (no-op/identity/...).

lib/Doctrine/DBAL/Driver/OCI8/OCI8Connection.php
```diff
@@ -163,21 +163,22 @@ public function exec($statement)
     /**
      * {@inheritdoc}
      */
-    public function lastInsertId($name = null)
+    public function lastInsertId($name = null) : string
     {
         if ($name === null) {
-            return false;
+            return '0';
```
Majkl578: Looks BC break-ish, what returned `false` before would now return `'0'`.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Connection.php
```diff
@@ -163,21 +163,22 @@ public function exec($statement)
...
+            $stmt = $this->prepare('SELECT ' . $name . '.CURRVAL FROM DUAL');
+            $stmt->execute();
+        } catch (OCI8Exception $exception) {
+            // In case no sequence value is available yet, we get the error:
+            // ORA-08002: sequence $name.CURRVAL is not yet defined in this session
+            return '0';
         }
 
-        return (int) $result;
+        return (string) $stmt->fetchColumn();
```
Majkl578: Looks BC break-ish, what returned `0` before would now return `'0'`.

Majkl578 (commented):
I am not sure about the possible BC impact especially around those type changes (making everything string). :|

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php
morozov: In the original proposal, `lastInsertId()` suppresses errors everywhere by design (see the OCI implementation specifically). Not sure I understand and agree with it. We definitely could improve here.

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php
morozov: Same as above. Agree on improving error handling.

lib/Doctrine/DBAL/Driver/LastInsertId.php
morozov: We can make it immutable. Would it make more sense to rename this to something like `with() : self` in this case? What about `get()`? Should it be something like `toValue()` instead?

morozov:
Given the amount of remaining work which is entirely about the API and BC-related concerns, I propose to move it out of `v2.7.0`. The pressure of this ticket being a release blocker is demotivating and provocative to taking shortcuts.

lib/Doctrine/DBAL/Driver/Connection.php
```diff
@@ -69,6 +69,14 @@ public function exec($statement);
     /**
      * Returns the ID of the last inserted row or sequence value.
      *
+     * If no sequence name is given, "0" is returned if there currently is no last insert ID available
+     * in the current session or in case of an error or if the driver does not support this operation.
```
morozov: @Majkl578 this is where the error suppression is explicitly defined. It looks dangerous to me that the return value may be different from what's expected because of a runtime issue.

Ocramius:
@morozov agreed, pushing this back into `3.0`.

lib/Doctrine/DBAL/Driver/LastInsertId.php
morozov: @Majkl578, sadly, the immutable approach will not just work since `MysqliConnection` and `MysqliStatement` share the same instance of `LastInsertId`. It's probably done like this to mitigate:

> MySQL drivers reset the last insert ID to `0` after each non-insert query

It could be worked around by keeping the reference to the connection from the statement. But this approach stinks too (also implemented in PDO Statement/Connection pair).

morozov:
I'm closing this as "won't fix" and not a bug. Below is the explanation of what exactly was attempted to get fixed and why it shouldn't.

### Inconsistency in _when_ the last inserted ID is available across platforms

This is what the largest part of the code change is about including the introduction of the `LastInsertID` class: MySQL "forgets" the ID right after the next query following the `INSERT` one, PostgreSQL "forgets" it after dropping the table, etc.

This issue is being fixed at the cost of creating either unnecessary coupling between the `Statement` and `Connection` objects (see MySQL and PDO for example), or code duplication (see SQL Server). At the same time, it allows the caller to be careless about capturing the last inserted ID any time and counting on it always being available. At the same time, if the caller is careless enough, there may be another `INSERT` happened between the original one and the moment of the capture, so this approach is error-prone by definition.

Instead of fixing this issue, we can explicitly recommend fetching the last insert ID immediately after `INSERT`. Ironically, Oracle which doesn't support identity columns, provides even better way of fetching the ID using the `INSERT ... RETURNING id INTO :id` syntax.

### Inconsistency in _what_ the last inserted ID looks like

Yes, there is an inconsistency between what an empty ID looks like (all kinds of empty values) and the non-empty ones (integer or a string). But it's natural for a weakly-typed language like PHP.

1. If the ID is empty, it can be reliably checked as `if (!$lastInsertedID)` without any knowledge of its type. Furthermore, without the proper error reporting (https://github.com/doctrine/dbal/issues/3005) the puristic approach of `if (!$lastInsertedID === '0')` is dangerous in the way that it doesn't account for potentially existing bug in the DBAL or underlying drivers (e.g. `$lastInsertedID = false` as a result of an error). Technically, any empty value means "none" regardless of its type.
2. If it's not empty, the most common use case is to use it in subsequent `INSERT`/`UPDATE` statements. At this point, the type of the ID doesn't matter as long as it can be bound to the statement.

Depite the fact that IDs are mostly integers in the DB internals, we cannot represent them as such due to potential overflow issues. At the same time, declaring them strings on the API level is misleading in the way that they are not just strings (e.g. `"xyz"`), they are strings representing integers, and PHP is fine with that.

Technically, we can re-iterate on the type consistency later, after https://github.com/doctrine/dbal/issues/3005 is resolved but besides being theoretically correct, I don't see a practical value in it.

PR #3077, opened 2018-04-01
CS bump to 4.0, imports for global functions & constants

Majkl578:
CS is a simple bump (no change required since we use incremental diffs here).

Had to hack some sniffs to:
* only add imports for functions/constants (and not classes as well)
* to re-order only functions/constants (and not classes as well)

Would love to see this in 2.7, but not a dealbreaker.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -84,6 +84,7 @@ install:
           }
         }
     # install composer dependencies
+    - rm composer.lock
     - appveyor-retry php composer.phar self-update
     - appveyor-retry php composer.phar install --no-progress --profile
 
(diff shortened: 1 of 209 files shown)


Ocramius (approved):
This is an easy and big win. It won't make a huge difference in this test suite alome, but things like the DBAL types are used a gazillion times, and we may have some improvements there 👍

Ocramius:
Checked scrutinizer issues: mostly raised by non-existence of exotic extensions (db2, sql anywhere, sqlsrv)

Bilge:
But why was `composer.lock` readded?

Ocramius:
`composer.lock` is needed when testing with tools like phpcs, because they keep inadvertedly breaking BC.

We use `composer install` (with `composer.lock`) in some CI pipeline steps.

PR #3080, opened 2018-04-02
[DBAL-3079] Reworked the usage of PDO in PDOConnection from inheritance to composition

### LINKED ISSUE #3079: Connection::query() should accept string $sql as first parameter
Currently, `Connection::query()` doesn't accept arguments and makes implementors use `func_get_args()` internally. It causes [false positives on Scrutinizer](https://scrutinizer-ci.com/g/doctrine/dbal/inspections/3c50cd7c-82e7-4e06-99d1-60a8e4c525aa/code-structure/class/Doctrine%5CTests%5CDBAL%5CFunctional%5CLastInsertIdTest):

> The call to `Doctrine\DBAL\Driver\Connection::query()` has too many arguments starting with `'INSERT INTO last_insert...table (foo) VALUES (1)'`.

Can be fixed by not extending `PDO` by `PDOConnection` similarly to what we previously did with `PDOStatement`.

morozov:
Fixes #3079.

1. `PDOConnection` no longer extends `PDO`.
2. The `PDO` object can be obtained via `PDOStatement::getWrappedConnection()`. It's required to be able to call PDO-specific methods like `setAttribute()`.
3. As long as the signature of `PDOStatement::query()` doesn't have to be compatible with `PDO::query()`, it's updated to `PDOStatement::query(string $sql) : ResultStatement`. This is the primary goal of this pull request.
4. Other query-related methods of `Connection` have been updated with typed arguments for consistency: `prepare()`, `executeQuery()`, `executeUpdate()`, `exec()`,
5. `Driver\Statement::rowCount()` was updated with `: int` since it's invoked from some methods above.
6. The tests wich invoked `PDO` methods on `PDOStatement` have been reworked to use `PDOStatement::getWrappedConnection()`.
7. `DriverConnectionMock` has been removed since it required an update but wasn't use anywhere.


lib/Doctrine/DBAL/Driver/Connection.php
```diff
@@ -19,16 +20,18 @@
...
     /**
      * Executes an SQL statement, returning a result set as a Statement object.
      *
-     * @return \Doctrine\DBAL\Driver\Statement
+     * @return Statement
+     *
+     * @throws DBALException
      */
-    public function query();
+    public function query(string $sql);
```
morozov: I didn't add the return type since the wrapper connection will have to declare that it returns a driver connection. This will likely change in the future.

lib/Doctrine/DBAL/Driver/PDOConnection.php
```diff
@@ -48,17 +52,17 @@ public function exec($statement)
...
     {
-        return PDO::getAttribute(PDO::ATTR_SERVER_VERSION);
+        return $this->conn->getAttribute(PDO::ATTR_SERVER_VERSION);
     }
 
     /**
      * {@inheritdoc}
      */
-    public function prepare($prepareString, $driverOptions = [])
+    public function prepare($prepareString)
```
morozov: No longer needed to follow the `PDO` signature. Will add to the changelog later.

lib/Doctrine/DBAL/Driver/PDOConnection.php
```diff
@@ -115,4 +117,44 @@ protected function createStatement(\PDOStatement $stmt) : PDOStatement
     {
         return new PDOStatement($stmt);
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function beginTransaction()
+    {
+        return $this->conn->beginTransaction();
```
morozov: Not wrapping this into a driver exception. Could be resolved in a different ticket.

lib/Doctrine/DBAL/Driver/Connection.php
Ocramius: @morozov I didn't understand this comment: if we change this in future, wouldn't it be better to let this become a loud BC break via a change of the signature?

lib/Doctrine/DBAL/Driver/PDOConnection.php
```diff
@@ -7,12 +7,16 @@
...
- * @since 2.0
+ * Used by all PDO-based drivers.
  */
-class PDOConnection extends PDO implements Connection, ServerInfoAwareConnection
+class PDOConnection implements Connection, ServerInfoAwareConnection
 {
+    /**
+     * @var PDO
+     */
+    private $conn;
```
Ocramius: Suggest renaming this to `connection` or even `pdoConnection` to avoid abbreviation and confusion overall

Ocramius (changes requested):

morozov:
@Ocramius want to take a look again?

lib/Doctrine/DBAL/Connection.php
```diff
@@ -881,16 +880,14 @@ public function prepare($statement)
...
      *
-     * @param string                                      $query  The SQL query to execute.
-     * @param array                                       $params The parameters to bind to the query, if any.
-     * @param array                                       $types  The types the previous parameters are in.
-     * @param \Doctrine\DBAL\Cache\QueryCacheProfile|null $qcp    The query cache profile, optional.
-     *
-     * @return \Doctrine\DBAL\Driver\Statement The executed statement.
+     * @param string                 $query  The SQL query to execute.
+     * @param mixed[]                $params The parameters to bind to the query, if any.
+     * @param mixed[]                $types  The types the previous parameters are in.
```
Ocramius: Types should be `(int|string)[]`

lib/Doctrine/DBAL/Connection.php
```diff
@@ -929,16 +926,14 @@ public function executeQuery($query, array $params = [], $types = [], QueryCache
     /**
      * Executes a caching query.
      *
-     * @param string                                 $query  The SQL query to execute.
-     * @param array                                  $params The parameters to bind to the query, if any.
-     * @param array                                  $types  The types the previous parameters are in.
-     * @param \Doctrine\DBAL\Cache\QueryCacheProfile $qcp    The query cache profile.
+     * @param string            $query  The SQL query to execute.
+     * @param mixed[]           $params The parameters to bind to the query, if any.
+     * @param mixed[]           $types  The types the previous parameters are in.
```
Ocramius: Types should be `(int|string)[]`

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1479,16 +1460,11 @@ public function convertToPHPValue($value, $type)
...
-     * @param array                           $params The map/list of named/positional parameters.
-     * @param array                           $types  The parameter types (PDO binding types or DBAL mapping types).
-     *
-     * @return void
-     *
-     * @internal Duck-typing used on the $stmt parameter to support driver statements as well as
-     *           raw PDOStatement instances.
+     * @param DriverStatement $stmt   The statement to bind the values to.
+     * @param mixed[]         $params The map/list of named/positional parameters.
+     * @param mixed[]         $types  The parameter types.
```
Ocramius: Types should be `(int|string)[]`

lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
```diff
@@ -22,7 +22,7 @@ class Driver extends AbstractPostgreSQLDriver
     public function connect(array $params, $username = null, $password = null, array $driverOptions = [])
     {
         try {
-            $pdo = new PDOConnection(
+            $conn = new PDOConnection(
```
Ocramius: Can we rename `$conn` here? Let's avoid abbreviations

lib/Doctrine/DBAL/Driver/PDOSqlite/Driver.php
```diff
@@ -36,7 +36,7 @@ public function connect(array $params, $username = null, $password = null, array
         }
 
         try {
-            $pdo = new PDOConnection(
+            $conn = new PDOConnection(
```
Ocramius: Same: can we avoid the abbreviation?

lib/Doctrine/DBAL/Portability/Connection.php
```diff
@@ -4,9 +4,11 @@
 
 use Doctrine\DBAL\Cache\QueryCacheProfile;
 use Doctrine\DBAL\ColumnCase;
+use Doctrine\DBAL\Driver\PDOConnection;
```
Ocramius: Unrelated to this patch, but do we still need the portability connection in 3.x?

Ocramius (changes requested):
Besides nits, this looks really good :+1:

lib/Doctrine/DBAL/Connection.php
morozov: When iterating over `(int|string)[]`, PhpStorm will recognize elements as `mixed`, but if it's `int[]|string[]` (not exactly correct), it will recognize them as `int|string`. This was previously brought up in https://github.com/doctrine/dbal/pull/3025#discussion_r168909829.

I'm still in favor of the proper annotation instead of making PhpStorm happy, especially if it's about scalar types (@Majkl578).

lib/Doctrine/DBAL/Portability/Connection.php
morozov: What's wrong with it? The column case conversion is very helpful when dealing with Oracle and IBM DB2.

morozov: On the other hand, I agree that `PORTABILITY_RTRIM` is just a crutch for badly designed code which happens to work on MySQL. And as for `PORTABILITY_EMPTY_TO_NULL` — no idea what it could be used for.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -929,16 +926,14 @@ public function executeQuery($query, array $params = [], $types = [], QueryCache
...
-     * @param \Doctrine\DBAL\Cache\QueryCacheProfile $qcp    The query cache profile.
-     *
-     * @return \Doctrine\DBAL\Driver\ResultStatement
+     * @param string              $query  The SQL query to execute.
+     * @param mixed[]             $params The parameters to bind to the query, if any.
+     * @param (int|string)|Type[] $types  The types the previous parameters are in.
+     * @param QueryCacheProfile   $qcp    The query cache profile.
      *
-     * @throws \Doctrine\DBAL\Cache\CacheException
+     * @throws DBALException
```
morozov: `CacheException` is a subclass of `DBALException` which can happen underneath. Therefore, replaced.

Ocramius (approved):
:shipit:

Ocramius:
Thanks @morozov!

lib/Doctrine/DBAL/Connection.php
Majkl578: Please let's not introduce types like `(int|string)[]`. Although unfortunate, it doesn't work in PhpStorm, our CS and most of other tools either.

Majkl578: Hmm I see you actually did the change... :/
I am strongly against and it should be reverted given the current state of PHP ecosystem.

morozov: We need to achieve some agreement before reverting it. The state of the PHP ecosystem is PHP 7.1 = ~37%. Yet, we don't support the older versions.

Majkl578: The state of PHP ecosystem is: PHPStan supports this, any other software doesn't.

stof:
Could we have a way for DBAL's PDOStatement to expose the underlying PDOStatement ? PDO has a `getColumnMeta` method, which is useful to get info about the type of returned columns (I have a specific use case for that, where I have a controller running SQL queries stored in dedicated files and returning the data in CSV, and so the controller cannot assume the names of column headers as it changes depending on the query).

With DBAL 2.x, I can access this info thanks to the inheritance. But I don't see any way to achieve this in current state of 3.x now that composition is used. The PDOConnection allows accessing the wrapped connection, but the PDOStatement does not have the equivalent API.

morozov:
@stof please file a separate ticket. For a use case like yours, it'd make sense to expose the underlying statement not only for PDO drivers.

stof:
@morozov here it is: https://github.com/doctrine/dbal/issues/3534

PR #3085, opened 2018-04-04
Tests: remove implicit verbose flag

Majkl578:
Having this by default has little value and only clutters the output. PHPUnit will still show skipped/incomplete tests as S/I and in summary.

Opting in is easy by just running PHPUnit with `-v`.
Does not affect error/fail output.


morozov:
Isn't it cute?
screenshot from 2018-04-04 11-45-59

morozov (approved):

Majkl578:
Looks like `VisualHostKey=yes` for SSH. :sunglasses:

Majkl578:
@Ocramius I'll merge this and if you go yolo about it or if it makes some things harder to spot/debug, we can revert later. :blush:

PR #3086, opened 2018-04-04
Postgres, Oracle and SQL Server Sequence increment_by & min_value as int

simPod:
When comparing schema, it causes problem at https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/Comparator.php#L196 where `$initialValue` is int in first case and string in second case and therefore `schema:validate` command never passes as `--dump-sql` always generates statements like:

```sql
ALTER SEQUENCE asequence_id_seq INCREMENT BY 1;
```

### DIFF
--- a/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
@@ -311,7 +311,7 @@ class PostgreSqlSchemaManager extends AbstractSchemaManager
             $sequence += $data;
         }
 
-        return new Sequence($sequenceName, $sequence['increment_by'], $sequence['min_value']);
+        return new Sequence($sequenceName, (int) $sequence['increment_by'], (int) $sequence['min_value']);
     }
 
     /**
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
@@ -123,7 +123,7 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
 
         $name = 'dropcreate_sequences_test_seq';
 
-        $this->_sm->dropAndCreateSequence(new \Doctrine\DBAL\Schema\Sequence($name, 20, 10));
+        $this->_sm->dropAndCreateSequence(new Sequence($name, 20, 10));
 
         self::assertTrue($this->hasElementWithName($this->_sm->listSequences(), $name));
     }
@@ -142,11 +142,11 @@ class SchemaManagerFunctionalTestCase extends \Doctrine\Tests\DbalFunctionalTest
 
     public function testListSequences()
     {
-        if(!$this->_conn->getDatabasePlatform()->supportsSequences()) {
-            $this->markTestSkipped($this->_conn->getDriver()->getName().' does not support sequences.');
+        if (! $this->_conn->getDatabasePlatform()->supportsSequences()) {
+            $this->markTestSkipped($this->_conn->getDriver()->getName() . ' does not support sequences.');
         }
 
-        $sequence = new \Doctrine\DBAL\Schema\Sequence('list_sequences_test_seq', 20, 10);
+        $sequence = new Sequence('list_sequences_test_seq', 20, 10);
         $this->_sm->createSequence($sequence);
 
         $sequences = $this->_sm->listSequences();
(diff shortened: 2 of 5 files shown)


Majkl578 (changes requested):
Requires a test case.

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -294,7 +294,7 @@ protected function _getPortableSequenceDefinition($sequence)
             $sequence += $data;
         }
 
-        return new Sequence($sequenceName, $sequence['increment_by'], $sequence['min_value']);
+        return new Sequence($sequenceName, (int) $sequence['increment_by'], (int) $sequence['min_value']);
```
morozov: Should we update the signature of `Sequence::__construct()` instead and declare them integers? With the absence of strict scalar types on the calling end, the effect will be the same.

simPod: Didn't notice the absence of strict declaration. Will do.

simPod: Shouldn't I leave there `return new Sequence($sequenceName, (int) $sequence['increment_by'], (int) $sequence['min_value']);` as well? It will have to be added later anyway I suppose.

morozov: Probably. I'd start with a test.

simPod: I think `(int)` conversion should stay and adding type hints would be nice. Done

morozov:
@simPod do the test changes cover the changes in the code? I'm almost certain, if the code changes are reverted, the test will still pass.

simPod:
@morozov I tried and there was the failure that `'20'` is not `20` so I think `assertSame` covers it

morozov:
@simPod oh, it's a functional test where the schema gets actually created in the DB and then introspected. Makes sense.

morozov:
@simPod I've restarted the build. The fix looks good for me. The only thing I'm unsure about is if we're allowed to introduce a type hint in a minor release (however, I believe it's the right thing to do).

morozov (dismissed):

simPod:
@morozov any idea who can say we're allowed to do that? :D I noticed @Ocramius is working on some Roave's tool from another dimension that might be able to do that.

@Majkl578 tag `Missing Tests` is not applicable now I believe

morozov:
@simPod, @Ocramius can. Most likely, his tool will flag it as a BC break :(

Ocramius:
Adding a type hint here doesn't seem like a BC break, because:

 1. constructors are "special" and escape the rules of signature variance
 2. the constructor was already documented with these exact types

Yes, my tool will flag this up, but that's because it's a naive approximation, so I didn't even think yet of this scenario :-)

morozov:
@Ocramius my concern was not about static API compatibility but about the contract. If a calling method declared strict scalar types and passed a numeric string, it would previously work but now will fail.

But I'm not trying to be a purist. If it's fine with you, it's fine with me.

Ocramius:
> If a calling method declared strict scalar types and passed a numeric string, it would previously work but now will fail.

True, but we guarantee the BC boundary according to the docblock. Let's face it: PHP is a shitty language, and we need to at least trust the declared types (even if just as comments) or else DOOM.

Majkl578 (changes requested):
I am still not sure about those type hints for Sequence.
Since `strict_types` is declared on caller's side, this will break code using non-integer values when `strict_types=1`. I'd rather not do that in 2.x.

simPod:
Removing typehints discovered that other managers needed the tweak too 🤔

Majkl578:
I added a functional test to this PR that is failing on master with Postgres 10.
TL;DR: This happens because Postges 10 platform automatically detects definition values during _list sequences_ and provides them to schema manager as is (as strings). In Postges <10, these values were queried by schema manager itself.

morozov:
@Majkl578 is the Oracle test supposed to fail? The code changes themselves look right.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1452,4 +1456,37 @@ public function testCreateAndListSequences() : void
...
+        $sequenceInitialValue   = 10;
+        $sequence               = new Sequence($sequenceName, $sequenceAllocationSize, $sequenceInitialValue);
+
+        $this->_sm->dropAndCreateSequence($sequence);
+
+        $createdSequence = array_values(
+            array_filter(
+                $this->_sm->listSequences(),
+                function (Sequence $sequence) use ($sequenceName) : bool {
+                    return strcasecmp($sequence->getName(), $sequenceName) === 0;
```
Majkl578: Things in Oracle are uppercase... Changed to case insensitive comparison. @morozov any pointers how to do this better, without array_filter?

morozov: No idea given the current API. We could improve the API to make it return sequences indexed by name.

morozov: BTW, I found it literally a couple of days ago that DB2 supports [`DB2_CASE_LOWER`](http://php.net/manual/en/ibm-db2.constants.php#constant.db2-case-lower), but we use driver level case conversion only with PDO. If the same thing existed for Oracle, we could throw away case conversion from the portability layer.

Majkl578 (approved):

morozov (approved):

morozov:
Back-ported to `2.7` via bd5e190a8281c59a539bd1f7a394fda541df16cd.

PR #3092, opened 2018-04-06
CI: Aggregate coverage from multiple jobs

Majkl578:
_not ready yet_

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -4,13 +4,15 @@ platform:
 #matrix:
 #  fast_finish: true # kills the build at the first failure
 clone_folder: C:\projects\dbal
-clone_depth: 1
+clone_depth: 2
 
 cache:
   - C:\ProgramData\chocolatey\bin -> .appveyor.yml
   - C:\ProgramData\chocolatey\lib -> .appveyor.yml
   - C:\tools\php -> .appveyor.yml
   - composer.phar
+  - ocular.phar
+  - cacert.pem
   - '%LOCALAPPDATA%\Composer\files'
   #- vendor
 
@@ -20,19 +22,23 @@ environment:
   - db: mssql
     driver: sqlsrv
     db_version: sql2008r2sp2
+    coverage: yes
     php: 7.2
   - db: mssql
     driver: sqlsrv
     db_version: sql2012sp1
     php: 7.2
+    coverage: yes
   - db: mssql
     driver: sqlsrv
     db_version: sql2017
+    coverage: no
     php: 7.2
   - db: mssql
     driver: pdo_sqlsrv
     db_version: sql2017
     php: 7.2
+    coverage: yes
 
 init:
   - SET PATH=C:\Program Files\OpenSSL;c:\tools\php;%PATH%
@@ -59,6 +65,7 @@ install:
           Add-Content php.ini "`n extension=php_fileinfo.dll"
           Add-Content php.ini "`n extension=php_pdo_sqlite.dll"
           Add-Content php.ini "`n extension=php_sqlite3.dll"
+          Add-Content php.ini "`n extension=php_curl.dll"
 
           # Get and install the MSSQL DLL's
           $DLLVersion = "5.2.0rc1"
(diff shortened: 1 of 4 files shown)


Majkl578:
Hmm, looks like it won't be _that_ easy...
> Code coverage data is not yet available.
> Scrutinizer was notified that the tests failed.
>
> If you are sending coverage for multiple environments, make sure that all environments
> support generating code coverage, otherwise the first build that finishes
> defines the final status.

.scrutinizer.yml
```diff
@@ -21,6 +21,7 @@ before_commands:
 tools:
     external_code_coverage:
         timeout: 3600
+        runs: 17
```
morozov: If the syntax allows, could you leave a comment about how the expected number of runs is determined? It'll have to be synchronized with the changes in `.travis.yml` and ContinuousPHP configuration. AppVeyor probably doesn't count since it doesn't cover any additional DB platforms.

.travis.yml
```diff
@@ -18,26 +18,64 @@ env:
...
 
 before_install:
-  - mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{,.disabled} || echo "xdebug not available"
+  - mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{,.disabled} || true
 
 before_script:
   - if [[ "$DB" == "mysql" || "$DB" == "mysqli" || "$DB" == *"mariadb"* ]]; then mysql < tests/travis/create-mysql-schema.sql; fi;
+  - |
+    if [ -v $COVERAGE ]; then
+      mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{.disabled,} || true
```
morozov: So we unconditionally disable Xdebug `before_install` and then conditionally enable it here? Wouldn't it be more straightforward to conditionally disable it `before_install` and then conditionally check if the extension is loaded here?

.scrutinizer.yml
Majkl578: YAML comment is used below so it should be ok. Actually forgot to count ContinuousPHP in. :) We can add AppVeyor too for the older (2008, 2012) platforms.

.travis.yml
morozov: But between `before_install` and `before_script`, no PHP scripts are run. So the net effect will be the same.

Majkl578: There is _install_ between those two (although it's not ordered accordingy in our config).

* before_install
* install
* before_script
* script

https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle

morozov: But Composer disables Xdebug internally, so it shouldn't affect it.

Majkl578: Hmm, is that even possible? Afaik if PHP is started with XDebug registered as zend_extension, this itself has impact on the runtime since XDebug overrides default execution handlers...

morozov: @Majkl578 see `Composer\XdebugHandler`. It restarts PHP with explicit INI configuration which excludes Xdebug (something like that).

Majkl578: I see, it was added in Composer 1.3.0, 15 months ago. Probably safe to assume this behavior now then. :)

morozov:
@Majkl578 do you want to add https://github.com/doctrine/dbal/compare/master...morozov:phpunit-configuration on top? Just found that the whitelist is not configured for the default config and other params are inconsistent. Didn't want it to conflict with your changes as a separate PR.

Majkl578:
Feel free to land that separately, this is still incomplete and shouldn't conflict difectly with your changes. :)

morozov:
Then I'll add whitelist configuration to all files for consistency, and you'll have to remove yours after my PR gets merged.

Majkl578:
This is ready for Travis + AppVeyor. ContinuousPHP will need some further work.

.appveyor.yml
```diff
@@ -4,13 +4,15 @@ platform:
 #matrix:
 #  fast_finish: true # kills the build at the first failure
 clone_folder: C:\projects\dbal
-clone_depth: 1
+clone_depth: 2
```
morozov: Is it for `ocular` to be able to determine something? Will it be always enough?

Majkl578: Without that the command used to list commit's parents fails: `git log --pretty="%P" -n1 HEAD`
With depth 2 it works, seems to be fine for both single and merge commits (for merge commits it fetches both parents).

```
$ git clone git@github.com:doctrine/dbal --branch=master --depth=2 dbal-shallow -q
$ cd dbal-shallow/
$ git log --oneline 
6ee7d7d (HEAD -> master, origin/master, origin/HEAD) Merge pull request #3117 from morozov/develop-badges
9b7a04a (grafted) Develop badges
5ba43cf (grafted) Merge pull request #3112 from morozov/run-all-cleanup
```

Majkl578:
I removed the expected coverage from Continuous PHP so it's 17x Travis + 3x AppVeyor now.

morozov (dismissed):
:+1: Looking forward to seeing the updated coverage!

Majkl578:
Hmm, strange AppVeyor failure. 🤔

morozov (approved):

PR #3097, opened 2018-04-07
Fix compatibility for pre-2.7 DateIntervalType format

### LINKED ISSUE #3093: When updating to version v2.7 type DateInterval throws errors
When using doctrine/dbal with Symfony, the following error is thrown after updating from 2.6.3 to 2.7.0:

```
Could not convert database value "P00Y00M07DT00H00M00S" to Doctrine Type dateinterval. Expected format: %RP%YY%MM%DDT%HH%IM%SS
```

Using `"doctrine/dbal": "2.6.*"` fixes this. Is this a usage issue or a bug?

Majkl578:
After #2579 DateIntervalType always expects a sign, which is not there for records created by pre-2.7 DBAL.

Fixes #3093.

### DIFF
--- a/lib/Doctrine/DBAL/Types/DateIntervalType.php
+++ b/lib/Doctrine/DBAL/Types/DateIntervalType.php
@@ -56,10 +56,17 @@ class DateIntervalType extends Type
             return $value;
         }
 
-        try {
-            $interval = new \DateInterval(substr($value, 1));
+        $negative = false;
 
-            if (substr($value, 0, 1) === '-') {
+        if (isset($value[0]) && ($value[0] === '+' || $value[0] === '-')) {
+            $negative = $value[0] === '-';
+            $value    = substr($value, 1);
+        }
+
+        try {
+            $interval = new \DateInterval($value);
+
+            if ($negative) {
                 $interval->invert = 1;
             }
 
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Types/DateIntervalType.php
```diff
@@ -57,9 +57,16 @@ public function convertToPHPValue($value, AbstractPlatform $platform)
         }
 
         try {
-            $interval = new \DateInterval(substr($value, 1));
+            $negative = false;
 
-            if (substr($value, 0, 1) === '-') {
+            if ($value[0] === '+' || $value[0] === '-') {
```
morozov: The value may be an empty string, so before it gets converted to an interval, we'll have a notice. Probably, the previous approach with `substr()` should be used instead.

morozov (changes requested):

lib/Doctrine/DBAL/Types/DateIntervalType.php
Majkl578: Used isset.
Note that using `substr('', 1)` produces FALSE, not empty string, but no warning. #php

morozov (approved):

morozov:
Back-ported to `2.7` via f62711ae809e3f8a45497f53813127f40396f0d0.

PR #3103, opened 2018-04-10
Added keywords OVER and RETURNING for MariaDb102

mdriessen:
I ran into a problem with MariaDB 10.2.14 when creating a field named 'over'. I noticed this keyword is missing from the MariaDb102Keywords file. While checking the entire list I also saw RETURNING was missing so I added that as well.

Other things I'm not sure about:
1) WINDOW is disallowed for table aliases, so should that be added as a keyword?
2) VIRTUAL is in the keyword file, but not listed on the MariaDB site. Any reason for that?
3) EXCEPT and INTERSECT are new keywords for MariaDB 10.3. Should we add them to this file or create a new Platform for that?

For reference:
https://mariadb.com/kb/en/library/reserved-words/


mdriessen:
@belgattitude Could you take a look at this PR?

mdriessen:
Thanks for the feedback, I've updated the pull request with the new keywords.

belgattitude (dismissed):

morozov:
> If possible I would like @morozov advice on this last question. If ok, let's proceed.

IMO, the entire concept of handling keywords in DBAL should go away (not sooner than in 3.0). See https://github.com/doctrine/dbal/issues/3152.

For now, the change looks good to me.

morozov (dismissed):
:+1:

mdriessen:
@morozov Done and thanks!

morozov (approved):

morozov:
@Ocramius do we backport patches like this? Technically, it's not a bug fix but it feels like it belongs to 2.7 where we introduced support for MariaDB.

Ocramius:
@morozov yeah, this is most likely worth backporting

PR #2720, opened 2017-05-09
`SQLParserUtils::getPlaceholderPositions()` fails if there are quoted strings containing only backslashes

mondrake:
Passing the following statement to ```SQLParserUtils::getPlaceholderPositions``` results in failure to identify the second placeholder ```condition_1```.

```
SELECT data.age AS age, data.id AS id, data.name AS name, data.id AS id
FROM test_data data 
WHERE 
  (data.description LIKE :condition_0 ESCAPE '\\') AND 
  (data.description LIKE :condition_1 ESCAPE '\\')
ORDER BY id ASC
```
Apparently, ```::getUnquotedStatementFragments``` gets confused by escapes.

Added a test to prove the bug.

PR #3104, opened 2018-04-10
Fix like escape for sqlite

nickvergessen:
Follow up to #2720 from @mondrake approved by @deeky666 (["I simply trust you guys on this one" 😿](https://github.com/doctrine/dbal/pull/2720#pullrequestreview-38390044))

Not sure if that is also the case for other DBs, but at least for SQLite `ESCAPE "…"` must only contain one char. So that is `ESCAPE "\\"` in php. The unit test accidently used `ESCAPE "\\\\"` and therefor didn't match the actual syntax and broke for SQLite.

In case `ESCAPE "\\\\"` is needed for another DB, we should just add both unit tests.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/NamedParametersTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/NamedParametersTest.php
@@ -5,6 +5,7 @@ namespace Doctrine\Tests\DBAL\Functional\Ticket;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\FetchMode;
 use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Platforms\PostgreSqlPlatform;
 use Doctrine\DBAL\Platforms\SqlitePlatform;
 use Doctrine\DBAL\Schema\Table;
 use const CASE_LOWER;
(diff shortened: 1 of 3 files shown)


MorrisJobke:
Maybe @Ocramius can have a look at this as well 🤔

tests/Doctrine/Tests/DBAL/SQLParserUtilsTest.php
```diff
@@ -68,11 +68,11 @@ public function dataGetPlaceholderPositions()
...
+            array('SELECT data.age AS age, data.id AS id, data.name AS name, data.id AS id FROM test_data data WHERE (data.description LIKE :condition_0 ESCAPE "\\") AND (data.description LIKE :condition_1 ESCAPE "\\") ORDER BY id ASC', false, array(121 => 'condition_0', 173 => 'condition_1')),
+            array('SELECT data.age AS age, data.id AS id, data.name AS name, data.id AS id FROM test_data data WHERE (data.description LIKE :condition_0 ESCAPE "\\") AND (data.description LIKE :condition_1 ESCAPE \'\\\') ORDER BY id ASC', false, array(121 => 'condition_0', 173 => 'condition_1')),
+            array('SELECT data.age AS age, data.id AS id, data.name AS name, data.id AS id FROM test_data data WHERE (data.description LIKE :condition_0 ESCAPE `\\`) AND (data.description LIKE :condition_1 ESCAPE `\\`) ORDER BY id ASC', false, array(121 => 'condition_0', 173 => 'condition_1')),
+            array('SELECT data.age AS age, data.id AS id, data.name AS name, data.id AS id FROM test_data data WHERE (data.description LIKE :condition_0 ESCAPE \'\\\') AND (data.description LIKE :condition_1 ESCAPE `\\`) ORDER BY id ASC', false, array(121 => 'condition_0', 173 => 'condition_1')),
```
morozov: > Not sure if that is also the case for other DBs […]

Is it possible to convert this (at least the relevant parts) to a functional test to make sure it does work on a real DB. Otherwise, we may end up changing tests after implementation w/o any evidence that the query is actually valid.

nickvergessen: Can you tell me a file where i should put them?

morozov: Please look at https://github.com/doctrine/dbal/tree/master/tests/Doctrine/Tests/DBAL/Functional.

nickvergessen: Added to tests/Doctrine/Tests/DBAL/Functional/NamedParametersTest.php

tests/Doctrine/Tests/DBAL/Functional/NamedParametersTest.php
```diff
@@ -212,6 +212,67 @@ public function testTicket($query,$params,$types,$expected)
...
         }
 
-        self::assertEquals($result, $expected);
+        self::assertEquals($expected, $result);
+    }
+
+    public function ticketGitHub3104Provider()
+    {
+        $escapes = [
+            "ESCAPE '\\'",
```
nickvergessen: So SQLite needs `\\` and all others need `\\\\`…
How to best adjust the test for this?

morozov: The question is, how to write a portable query which will work on SQLite and others using the DBAL? If there's no proper solution now, then there's nothing to be fixed or tested: no matter how it's implemented, it's not portable.

nickvergessen: Okay found a way, please have a look again.

And well, the databases require a different query, that is exactly what this PR is fixing. SQLite is broken at the moment, all others work. With this PR SQLite and all others work.

morozov: @nickvergessen now Postgres is broken. Please take a look at the build.

tests/Doctrine/Tests/DBAL/Functional/NamedParametersTest.php
```diff
@@ -212,6 +214,63 @@ public function testTicket($query,$params,$types,$expected)
...
+            ['SELECT 1 FROM ddc1372_foobar WHERE (:param_0 LIKE :find ESCAPE "{{escape}}") OR (:param_1 LIKE :find ESCAPE "{{escape}}") LIMIT 1', true],
+        ];
+    }
+
+    /**
+     * @dataProvider namedParametersAfterEscapeProvider
+     * @param string $query
+     * @param bool $ignorePostgres
+     */
+    public function testNamedParametersAfterEscapeFirstMatches($query, $ignorePostgres)
```
morozov: The functional test should only only cover the literals quoted according to the platform rules: `AbstractPlatform::quoteStringLiteral()`. The fact that the parser supports other quote characters than the current platform supports will be covered in the unit test for the parser.

tests/Doctrine/Tests/DBAL/Functional/NamedParametersTest.php
```diff
@@ -212,6 +214,35 @@ public function testTicket($query,$params,$types,$expected)
...
+        ];
+    }
+
+    /**
+     * @dataProvider namedParametersAfterEscapeProvider
+     * @param array $params
+     */
+    public function testNamedParametersAfterEscape(array $params)
+    {
+        $escapeChar = $this->_conn->getDatabasePlatform()->quoteStringLiteral('\\');
```
nickvergessen: @morozov the problem is, PostgreSqlPlatform escapes `\`, but the string in `ESCAPE '…'` must not be escaped for PostgreSql

morozov: @nickvergessen please see if you can use the API introduced in https://github.com/doctrine/dbal/pull/3013.

Issue #2787, opened 2017-07-24
Filtering by binary field doesn't work on Oracle


morozov:
Setup:
```php
$table = new \Doctrine\DBAL\Schema\Table('test');
$table->addColumn('id', 'integer');
$table->addColumn('value', 'binary', ['length' => 16]);

$sm = $conn->getSchemaManager();
$sm->createTable($table);

$conn->insert('test', [
    'id' => 1,
    'value' => 'test',
], [
    PDO::PARAM_INT,
    PDO::PARAM_LOB,
]);
```
Binding param as LOB (same as during insertion):
```php
$stmt = $conn->executeQuery('SELECT id FROM test WHERE value = ?', ['test'], [PDO::PARAM_LOB]);
var_dump($stmt->fetch());

// ORA-00932: inconsistent datatypes: expected - got BLOB
```

Binding param as a string:
```php
$stmt = $conn->executeQuery('SELECT id FROM test WHERE value = ?', ['test']);
var_dump($stmt->fetch());

// bool(false) – not found
```

Representing the value as a hexadecimal string:
```php
$stmt = $conn->executeQuery('SELECT id FROM test WHERE value = ?', [bin2hex('test')]);
var_dump($stmt->fetch());

// array(1) {
//  'ID' =>
//  string(1) "1"
// }
```

Binding param as a binary string:
```php
$stmt = oci_parse($conn, 'SELECT id FROM test WHERE value = :param');
$param = 'test';
oci_bind_by_name($stmt, 'param', $param, -1, OCI_B_BIN);
oci_execute($stmt);
var_dump(oci_fetch_array($stmt, OCI_ASSOC));

// array(1) {
//  'ID' =>
//  string(1) "1"
// }
```

Ocramius:
We can't really detect the server-side parameter type on our own, but I suppose that we could add `bin2hex()` when `PDO::PARAM_LOB` is used.

What happens for insert operations?

morozov:
> What happens for insert operations?

I think the fact that binding as LOB for INSERT/UPDATE happens to work is an unintended behavior on the server/driver side. Probably, during a write operation, Oracle doesn't really care if it's a string or a stream — it just gets written, but during the comparison, it cannot compare anything with a LOB by design.

[...]

PR #3107, opened 2018-04-13
Fix "Unknown database type binary requested" for SQLite

spyric:
Hello!

Currently there is no type mapping for binary that is cause of error: Unknown database type binary requested, Doctrine\DBAL\Platforms\SqlitePlatform

This PR fix it

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
@@ -642,6 +642,7 @@ class SqlitePlatform extends AbstractPlatform
             'decimal'          => 'decimal',
             'numeric'          => 'decimal',
             'blob'             => 'blob',
+            'binary'           => 'blob',
         ];
     }
 


lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -642,6 +642,7 @@ protected function initializeDoctrineTypeMappings()
             'decimal'          => 'decimal',
             'numeric'          => 'decimal',
             'blob'             => 'blob',
+            'binary'           => 'blob',
```
morozov: There may be a confusion between binary and blob types in the DBAL. The "binary" is supposed to represent a binary string of a known (fixed or variable) length like `BINARY(N)` or `VARBINARY(N)` in [MySQL](https://dev.mysql.com/doc/refman/5.7/en/binary-varbinary.html). Why "blob" is a large chunk of binary data.

The binary fields should be possible to use for comparison and as a primary key (e.g. a UUID PK could be stored in a `BINARY(16)`) field (see https://github.com/doctrine/dbal/issues/2787 for a reference).

Please make sure the way how "binary" is represented in SQLite meets these criteria.

morozov: @spyric sorry for the confusion. According to the `BinaryTest` added as part of #3133, `BLOB` indeed can be used as implementation of the `BINARY` column type on SQLite.

Please provide a test case.

morozov:
According to the [documentation](https://www.sqlite.org/datatype3.html#affinity_name_examples), SQLite doesn't support `BINARY` columns, therefore `SQLitePlatform::getDoctrineTypeMapping()` cannot expect `'binary'` as input.

PR #3109, opened 2018-04-14
Allow to specify arbitrary SQL expression in AbstractPlatform::getDummySelectSQL()

morozov:
For testing purposes and not only, it may be needed to evaluate an arbitrary SQL expression (see https://github.com/doctrine/dbal/pull/3013, https://github.com/doctrine/dbal/pull/3108). Instead of doing `str_replace()` on an expected string, one could specify the expression explicitly.


Majkl578 (changes requested):
Makes sense, but can't go into master due to the added parameter.

Majkl578 (approved):

PR #3112, opened 2018-04-19
Removed hard-coded configuration filenames from the test runner

morozov:
Currently, `run-all.sh` contains hard-coded configuration file names some of which are obsolete (`pdo-ibm`) and some are missing (`pdo-pgsql`).

### DIFF
--- a/run-all.sh
+++ b/run-all.sh
@@ -1,22 +1,9 @@
 #!/bin/bash
 
 # This script is a small convenience wrapper for running the doctrine testsuite against a large bunch of databases.
-# Just create the phpunit.xmls as described in the array below and configure the specific files <php /> section
-# to connect to that database. Just omit a file if you dont have that database and the tests will be skipped.
+# Create *.phpunit.xml files and specify database connection parameters in the <php /> section.
 
-configs[1]="mysql.phpunit.xml"
-configs[2]='postgres.phpunit.xml'
-configs[3]='sqlite.phpunit.xml'
-configs[4]='oracle.phpunit.xml'
-configs[5]='db2.phpunit.xml'
-configs[6]='pdo-ibm.phpunit.xml'
-configs[7]='sqlsrv.phpunit.xml'
-configs[8]='sqlanywhere.phpunit.xml'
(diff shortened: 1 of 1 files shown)


run-all.sh
```diff
@@ -1,22 +1,9 @@
...
-configs[7]='sqlsrv.phpunit.xml'
-configs[8]='sqlanywhere.phpunit.xml'
-
-for i in "${configs[@]}"; do
-    if [ -f "$i" ];
-    then
-        echo "RUNNING TESTS WITH CONFIG $i"
-        phpunit -c "$i" "$@"
-    fi;
+for i in phpunit.xml.dist *.phpunit.xml; do
```
Majkl578: Not really sure about the dist file here, probably shouldn't be included (also doesn't match description above).

run-all.sh
```diff
@@ -1,22 +1,9 @@
...
+    echo "RUNNING TESTS WITH CONFIG $i"
+    phpunit -c "$i" "$@"
```
Majkl578: Slightly unrelated, but this should use vendor/bin/phpunit, otherwise it looks up in $PATH.

Majkl578 (commented):
We may also rename the file to run-all-tests.sh to be more descriptive.

run-all.sh
morozov: Which is fine. My `$PATH` contains `vendor/bin`. Or for other users, it could be their `phpunit` alias.

run-all.sh
morozov: What's wrong with running the dist file first? Otherwise, in order to test `pdo-sqlite` one will need to create an extra file however the extension doesn't require configuration. The description doesn't say the script will run only those files.

morozov:
I'd leave the name as is to make sure nobody is required to update their scripts.

run-all.sh
Majkl578: Besides it wasn't there previously, dist file is meant to be a template for script like this.

Majkl578 (approved):

PR #3113, opened 2018-04-20
Revert complex type specification in Connection

Majkl578:
This reverts non-standard type specification using parenthesis.
Such format is not supported by any tools or IDEs, PHPStan likely being the sole exception in the PHP ecosystem. It breaks type inspection in IDEs etc.
Additionally it has no standard to support this (the syntax is currently just a PHPStan made-up).

Note that this is not even part of PSR-5 proposal.

Follow-up from https://github.com/doctrine/dbal/pull/3080#discussion_r181420821 and https://github.com/doctrine/dbal/pull/3025#discussion_r168909829.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -882,7 +882,7 @@ class Connection implements DriverConnection
      *
      * @param string                 $query  The SQL query to execute.
      * @param mixed[]                $params The parameters to bind to the query, if any.
-     * @param (int|string|Type)[]    $types  The types the previous parameters are in.
+     * @param int[]|string[]|Type[]  $types  The types the previous parameters are in.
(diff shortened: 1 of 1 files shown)


jvasseur:
This syntax is not something PHPStan made-up but something that comes from the PSR-5 proposal:

https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#arrays
> specified as containing multiple types, the Type definition informs the reader of the type of each array value. Each value can be of any of the given types. Example: `@return (int|string)[]`

Majkl578:
Here is how it breaks PhpStorm:
PhpStorm broken

And Sami:
SAMI broken

morozov (approved):
I have to agree. Especially because the more precise annotation doesn't bring any noticeable benefits from the static analysis standpoint.

Majkl578:
rebased

Majkl578:
rebased again :grimacing:

Issue #3114, opened 2018-04-20
Portability Statement still fetches all data in iterator mode


morozov:
As of https://github.com/doctrine/dbal/pull/2718, in iterator mode, statements are supposed to fetch entries one by one for more efficient memory usage.

Apparently, the portability statement wrapper still fetches all entries in memory:
https://github.com/doctrine/dbal/blob/11037b4352c008373561dc6fc836834eed80c3b5/lib/Doctrine/DBAL/Portability/Statement.php#L140-L145

It renders the entire optimization useless if the portability wrapper is used.

Issue #2705, opened 2017-04-26
SQLSrv (and other non PDO drivers) iterator inefficient


jenkoian:
https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php#L296-L301

When iterating (using foreach or something) over results using the sqlsrv driver it can easily run out of memory when dealing with large data sets. Looking at the code above, it seems it's because it's just doing a fetch all and populating an array iterator. The problem doesn't occur if you iterate over the results by doing `while ($result = $results->fetch())`. I notice we do this with [some of the other non pdo drivers also](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php#L407-L412). 

I'm wondering if we could improve this by building a bespoke iterator at this point rather than calling fetch all and populating an array iterator. I'm happy to take a look into this if you think it's possible, but I'm not 100% sure it is just yet, and possibly is a limitation of using the non pdo drivers.

One idea I was thinking of was building an iterator (using an anonymous class or something, fuck knows about pre php7 though) which uses the core functions such as [sqlsrv_next_result()](http://php.net/manual/en/function.sqlsrv-next-result.php) etc. Again, not sure if this is possible but was something that looked at least somewhat feasible? 

Is this something worth looking into? Is it even possible? Do people even care? Do people other than me even use this driver?!

 Like mentioned, be happy to give it a go, but don't want to waste my time if a non starter.

Ocramius:
Just a note: please use tagged versions or hashes when linking code, as code changes, and the links may not point at what you wanted to highlight in 1 year or so :-)

Ocramius:
> fuck knows about pre php7 though

Nobody cares - we bumped the requirement to PHP 7.0 anyway, so if you want to go ahead and implement something that requires 7.x, it can be done now.

jenkoian:
Started work on this here: https://github.com/doctrine/dbal/compare/master...jenkoian:efficient-iterating?expand=1 have only tested the SQLSrv one as that's the one I'm dealing with but it fixes the problem for me (didn't actually need to do anything too clever in the end). 

Still need to add some tests, work out ways of dealing with different fetch modes for different drivers (e.g. some drivers seem to do some custom mapping for certain fetch modes) and ultimately test the other drivers (other than SQLSrv). 

If anyone is working with these other drivers and wishes to help out that would be great!

PR #3115, opened 2018-04-22
Replaced ArrayIterator with StatementIterator in Portability\Connection

morozov:
Fixes https://github.com/doctrine/dbal/issues/3114

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
+++ b/tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
@@ -2,39 +2,101 @@
 
 namespace Doctrine\Tests\DBAL\Driver;
 
+use Doctrine\DBAL\Driver\IBMDB2\DB2Statement;
+use Doctrine\DBAL\Driver\Mysqli\MysqliStatement;
+use Doctrine\DBAL\Driver\OCI8\OCI8Statement;
+use Doctrine\DBAL\Driver\SQLAnywhere\SQLAnywhereStatement;
+use Doctrine\DBAL\Driver\SQLSrv\SQLSrvStatement;
 use Doctrine\DBAL\Driver\Statement;
 use Doctrine\DBAL\Driver\StatementIterator;
+use Doctrine\DBAL\Portability\Statement as PortabilityStatement;
+use IteratorAggregate;
+use PHPUnit\Framework\MockObject\MockObject;
+use Traversable;
+use function extension_loaded;
 
 class StatementIteratorTest extends \Doctrine\Tests\DbalTestCase
 {
-    public function testGettingIteratorDoesNotCallFetch()
+    /**
+     * @dataProvider statementProvider()
+     */
+    public function testGettingIteratorDoesNotCallFetch(string $class) : void
     {
-        $stmt = $this->createMock(Statement::class);
+        /** @var IteratorAggregate|MockObject $stmt */
+        $stmt = $this->createPartialMock($class, ['fetch', 'fetchAll', 'fetchColumn']);
         $stmt->expects($this->never())->method('fetch');
         $stmt->expects($this->never())->method('fetchAll');
         $stmt->expects($this->never())->method('fetchColumn');
 
-        $stmtIterator = new StatementIterator($stmt);
-        $stmtIterator->getIterator();
+        $stmt->getIterator();
     }
 
-    public function testIterationCallsFetchOncePerStep()
+    public function testIteratorIterationCallsFetchOncePerStep() : void
+    {
+        $stmt = $this->createMock(Statement::class);
+
+        $calls = 0;
+        $this->configureStatement($stmt, $calls);
+
+        $stmtIterator = new StatementIterator($stmt);
+
+        $this->assertIterationCallsFetchOncePerStep($stmtIterator, $calls);
+    }
+
+    /**
+     * @dataProvider statementProvider()
+     */
+    public function testStatementIterationCallsFetchOncePerStep(string $class) : void
+    {
+        $stmt = $this->createPartialMock($class, ['fetch']);
+
+        $calls = 0;
+        $this->configureStatement($stmt, $calls);
+        $this->assertIterationCallsFetchOncePerStep($stmt, $calls);
+    }
+
+    private function configureStatement(MockObject $stmt, int &$calls) : void
     {
         $values = ['foo', '', 'bar', '0', 'baz', 0, 'qux', null, 'quz', false, 'impossible'];
         $calls = 0;
 
-        $stmt = $this->createMock(Statement::class);
         $stmt->expects($this->exactly(10))
             ->method('fetch')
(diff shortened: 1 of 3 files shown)


tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
```diff
@@ -2,39 +2,91 @@
...
+    private function assertIterationCallsFetchOncePerStep($iterator, &$calls) : void
+    {
+        foreach ($iterator as $i => $_) {
             $this->assertEquals($i + 1, $calls);
         }
     }
+
+    public static function statementProvider()
+    {
+        if (extension_loaded('ibm_db2')) {
```
morozov: Unlike `MysqliStatement` and `SQLAnywhereStatement`, `DB2Statement` and other driver statement class declarations depend on corresponding extension symbols, so they only can be loaded if extensions are loaded.

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
```diff
@@ -2,39 +2,94 @@
...
+use IteratorAggregate;
+use PHPUnit\Framework\MockObject\MockObject;
+use Traversable;
+use function extension_loaded;
 
 class StatementIteratorTest extends \Doctrine\Tests\DbalTestCase
 {
-    public function testGettingIteratorDoesNotCallFetch()
+    /**
+     * @dataProvider statementProvider()
```
Majkl578: First time I see data providers with `()`, is this reliable?

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
```diff
@@ -2,39 +2,94 @@
...
+        /** @var IteratorAggregate|MockObject $stmt */
+        $stmt = $this->createPartialMock($class, ['fetch', 'fetchAll', 'fetchColumn']);
         $stmt->expects($this->never())->method('fetch');
         $stmt->expects($this->never())->method('fetchAll');
         $stmt->expects($this->never())->method('fetchColumn');
 
+        $stmt->getIterator();
+    }
+
+    public function testIteratorIterationCallsFetchOncePerStep()
```
Majkl578: can haz void?

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
```diff
@@ -2,39 +2,94 @@
...
-        $stmtIterator->getIterator();
+
+        $this->assertIterationCallsFetchOncePerStep($stmtIterator, $calls);
     }
 
-    public function testIterationCallsFetchOncePerStep()
+    /**
+     * @dataProvider statementProvider()
+     */
+    public function testStatementIterationCallsFetchOncePerStep(string $class)
```
Majkl578: can haz void?

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
```diff
@@ -2,39 +2,94 @@
...
+    {
+        $stmt = $this->createPartialMock($class, ['fetch']);
+        $this->configureStatement($stmt, $calls);
+        $this->assertIterationCallsFetchOncePerStep($stmt, $calls);
+    }
+
+    private function configureStatement(MockObject $stmt, &$calls) : void
```
Majkl578: can haz int typehint?
I am pretty sure @Ocramius will hate the reference here. :laughing:

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
```diff
@@ -2,39 +2,94 @@
...
                 $value = $values[$calls];
                 $calls++;
+
                 return $value;
             });
+    }
 
-        $stmtIterator = new StatementIterator($stmt);
-        foreach ($stmtIterator as $i => $_) {
+    private function assertIterationCallsFetchOncePerStep(Traversable $iterator, &$calls) : void
```
Majkl578: can haz int typehint?

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
```diff
@@ -2,39 +2,94 @@
...
+    {
+        foreach ($iterator as $i => $_) {
             $this->assertEquals($i + 1, $calls);
         }
     }
+
+    public static function statementProvider()
```
Majkl578: can haz `iterable` return type?

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
morozov: Yes, I've been always using them. A function name without parentheses looks awkward.

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
morozov: We've been through that a couple of times. Sometimes they just make sense.

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
morozov: Was lazy to initialize the variable :)

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
morozov: Yeah, again, was lazy to add the element type hint which the CS forces.

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
morozov: Would it make sense to require a non-empty return type on test methods? It's `void` in 99% of cases, but if a test returns a dependency, it's in most cases of a known type.

Majkl578:
This also fixes #2705 imho.

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
Majkl578: I don't understand the "non-empty return type" part. You mean making all tests return something specific, instead of void?

Majkl578 (approved):

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
morozov: Not sure it could be a requirement project-wide. For instance, `fetch()` could return either a row (`array`) or one field (`string|null`) or `false` (end of set). You'll have to have a very strict API to enable this, can't see it happening in the near future.

morozov:
Back-ported to `2.7` via 5003847465dc447ff3c2886f39c91caf9ce04e48.

PR #3119, opened 2018-04-28
Added "string" mapping in SqlitePlatform 

DeyV:
- fix 'Unknown database type string requested, Doctrine\DBAL\Platforms\SqlitePlatform may not support it.' error


Ocramius (approved):
LGTM! :+1:

PR #3120, opened 2018-04-29
Excluded tests from Scrutinizer CI analysis

morozov:
1. Judging by the [inspection results](https://scrutinizer-ci.com/g/doctrine/dbal/inspections/e291a0d7-dc12-4a37-b9b2-5fd74c4a5edd/issues/files/tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php?status=failed&conditionIds%5B0%5D=271126), Scrutinizer doesn't understand PHPUnit mocks which unify the type being tested and the mocking API. Adding tests with mocks impacts the code quality metric of the entire project.
2. The issues found in tests should not be considered when the project's code quality is assessed since tests are not used in production.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -26,6 +26,7 @@ tools:
 filter:
     excluded_paths:
         - docs
+        - tests
 
 build_failure_conditions:
     - 'elements.rating(<= C).new.exists'                        # No new classes/methods with a rating of C or worse allowed


lcobucci:
> Judging by the inspection results, Scrutinizer doesn't understand PHPUnit mocks which unify the type being tested and the mocking API. Adding tests with mocks impacts the code quality metric of the entire project.

It indeed doesn't but we can easily filter those org-wise (it's a bit annoying, I know), marking them as false-positives (which will then not count them as bugs anymore). It should be smatter though, PHPStan is way better in this sense by offering extensions.

> The issues found in tests should not be considered when the project's code quality is assessed since tests are not used in production.

I think that not filtering tests helps us to have good code in tests, which is quite important IMO. I'd keep them there so that we can have a complete view of the project we maintain, keeping an eye for the hotspots, and trying to improve things as much as possible.

I see a rating of 10 just like 100% of code coverage: it should just be a consequence.

Majkl578:
Generally I'd like to move most of these checks (we already did it for CS) to Travis and PHPCS + PHPStan/Psalm [+ <insert interesting tool here>]*.
Scrutinizer is good for coverage, maybe some code quality checks, but it's really lacking in static analysis area these days. :/

morozov:
> It should be smatter though, PHPStan is way better in this sense by offering extensions.

and

> Scrutinizer is […] lacking in static analysis area these days. :/

This is my concern together with almost complete absence of support and development of Scrutinizer.

morozov:
@lcobucci I see you marked the issues as ignored but the build itself is still flagged as failed. Is there a way to make Scrutinizer re-report the updated status to GitHub?

lcobucci:
@morozov @Majkl578 I completely get your points, what I like about scrutinizer is also how they show info, like: https://scrutinizer-ci.com/g/doctrine/dbal/code-structure/master/hot-spots

I'm fine with removing if we think that we won't neglect the quality of our tests.

> @lcobucci I see you marked the issues as ignored but the build itself is still flagged as failed. Is there a way to make Scrutinizer re-report the updated status to GitHub?

It picks it up on the next inspection after the configuration changes

PR #3122, opened 2018-04-30
Introduced a smoke testing phase on Travis to run SQLite tests and CS checks first

morozov:
Sometimes, the CS verification is the only failing stage of the build (especially, if CS is not verified locally before submitting the patch). In this case, after fixing the CS issue, it'd be very convenient to get immediate feedback instead of waiting 15-20 minutes until all test builds finish first.

A solution could be to introduce a "smoke testing" stage which would precede the testing one and include the following:

1. Testing against SQLite to find non-platform-specific issues since it's the fastest platform.
2. Code style verification.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,21 @@ jobs:
     - php: nightly
 
   include:
+
+    - stage: Smoke Testing
+      php: 7.2
+      env: DB=sqlite COVERAGE=yes
+    - stage: Smoke Testing
+      php: 7.2
+      env: PHPStan
+      install: travis_retry composer install --prefer-dist
+      script: vendor/bin/phpstan analyse
+    - stage: Smoke Testing
+      php: 7.2
+      env: PHP_CodeSniffer
+      install: travis_retry composer install --prefer-dist
+      script: vendor/bin/phpcs
+
     - stage: Test
       php: 7.2
       env: DB=mysql COVERAGE=yes
@@ -170,9 +185,6 @@ jobs:
         - docker
       before_script:
         - bash ./tests/travis/install-postgres-11.sh
-    - stage: Test
-      php: 7.2
-      env: DB=sqlite COVERAGE=yes
     - stage: Test
       php: 7.2
       env: DB=sqlsrv COVERAGE=yes
@@ -319,14 +331,3 @@ jobs:
       install:
         - composer config minimum-stability dev
         - travis_retry composer update --prefer-dist
-
-    - stage: Code Quality
-      php: 7.2
-      env: PHPStan
-      install: travis_retry composer install --prefer-dist
-      script: vendor/bin/phpstan analyse
-    - stage: Code Quality
-      php: 7.2
-      env: PHP_CodeSniffer
-      install: travis_retry composer install --prefer-dist
-      script: vendor/bin/phpcs


.travis.yml
```diff
@@ -64,231 +69,268 @@ jobs:
...
-    - stage: Test
+    - stage: smoke testing
       php: 7.1
       env: DB=sqlite COVERAGE=yes
 
-    - stage: Test
+    - stage: smoke testing
+      if: type = pull_request
+      php: 7.1
+      env: DB= PHP_CodeSniffer=yes
```
morozov: This is a purely cosmetic addition. Neither of the variable influence the build but are displayed in the job list and help to understand what the job is about.

.travis.yml
```diff
@@ -50,10 +50,15 @@ after_script:
...
+stages:
+  - smoke testing
+  - test
+  - coding standard
+
 jobs:
   allow_failures:
     - php: nightly
-    - stage: Coding standard
+    - stage: coding standard
```
morozov: The stages are lower-cased since otherwise, Travis fails to sort them properly.

morozov:
@Majkl578 depending on how important and fast they are — either at the beginning of the build or at the end. If they are fast and required, they will go into the smoke testing since there's no point in checking platform compatibility if the build is not going to be accepted due to quality or BC concerns.

.travis.yml
morozov: Looks like my bad. I've reverted the changes, and the order still seems correct.

Majkl578:
One more thing: I'm afraid this will cause issues with Scrutinizer. When the initial stage fails, further coverage runs will not be executed, thus Scrutinizer will be waiting for them infinitely (well, until timeout = 3600 seconds). :/ Not sure how to handle this.

morozov:
> Not sure how to handle this.

Do we need to handle this? If the build fails, the coverage is irrelevant. Isn't it the same as a test run fails and doesn't send any coverage results to Scrutinizer?

Majkl578:
It will block all other builds for an hour.

> Isn't it the same as a test run fails and doesn't send any coverage results to Scrutinizer?

Yes, something to still take care of, see the _Tip_ here: https://scrutinizer-ci.com/docs/tools/external-code-coverage/#changes-in-third-party-configuration

morozov:
So, it looks like at the end of the code style check job, it if fails, we need to call `ocular` which will cancel the build on Scrutinizer. Gitter seems down.

Majkl578:
Yes. We are currently doing that (in after_script): https://travis-ci.org/doctrine/dbal/jobs/373140364#L1014
I am just not sure whether one call like that is sufficient when 17 is expected. I guess we will have to try or we will find out. :)

> Gitter seems down.

Yes. 😢 Slack is up though. \o/

morozov:
We still can solve this problem if we collect coverage only nightly. Coverage is not taken into account to validate the pull request and is not displayed anywhere on GitHub.

morozov:
> I am just not sure whether one call like that is sufficient when 17 is expected. I guess we will have to try or we will find out. :)

@Majkl578 the Scrutinizer support said that once the platform receives an `ocular` call without a file, it recognizes it as a test failure and will proceed without code coverage.

Ocramius (approved):
:+1:

PR #3123, opened 2018-05-01
Opt in deprecations docs

greg0ire:


docs/en/reference/upgrading.rst
```diff
@@ -0,0 +1,53 @@
...
+    );
+
+Since this error is silenced, it will not produce any effect unless you
+opt-in by setting up an error handler designed to ignore the silence
+operator in that case. Such an error handler could look like this:
+
+.. code-block:: php
+
+    <?php
+    set_error_handler(function (
```
Majkl578: Is it actually a good idea to overwrite handler for all messages like that? Can we maybe suggest this instead only for deprecations?
```php
set_error_handler(
    function (int $code, string $message, string $file, int $line) : void {
        echo sprintf('Deprecation: "%s" in %s:%d', $message, $file, $line);
    },
    E_USER_DEPRECATED
);
```
https://3v4l.org/AFfPk

docs/en/reference/upgrading.rst
```diff
@@ -0,0 +1,53 @@
...
+            return;
+        }
+
+        // normal error handling here
+    });
+
+This is of course overly simplified, and if you are looking for such an
+error handler, consider the ``symfony/debug``, error handler that will
+log deprecations. You may also be interested by the
+``symfony/phpunit-bridge`` error handler that will display deprecations
```
Majkl578: It'd be nice to expand on the `symfony/phpunit-bridge` part. Imho that's the most convenient opt-in method.

docs/en/reference/upgrading.rst
greg0ire: Fixed

jwage (approved):

Majkl578:
Uh, we don't usually merge to 2.x branches, rather cherry-pick from master. This needs to be picked up to master now. :|

greg0ire:
> Uh, we don't usually merge to 2.x branches, rather cherry-pick from master. This needs to be picked up to master now. :|

Ah right, sorry for this

Issue #3118, opened 2018-04-28
MasterSlaveConnection::ping() does't work, it throws PDOException


fredeastside:
Hi,

When I use ```MasterSlaveConnection::ping()``` it doesn't work correctly. The problem's in custom ```MasterSlaveConnection::query()``` method. Now there don't exist block try-catch and line ```$statement->setFetchMode($this->defaultFetchMode);```. A script falls down with PDOException when execute ```SELECT 1``` query. Actually, this method has one different line ```$this->connect('master');``` and it could be work with these lines of code. But it seems like copy-paste

Doctrine\DBAL\Connection:
```php
/**
 * Executes an SQL statement, returning a result set as a Statement object.
 *
 * @return \Doctrine\DBAL\Driver\Statement
 *
 * @throws \Doctrine\DBAL\DBALException
 */
public function query()
{
    $this->connect();

    $args = func_get_args();

    $logger = $this->_config->getSQLLogger();
    if ($logger) {
        $logger->startQuery($args[0]);
    }

    try {
        $statement = $this->_conn->query(...$args);
    } catch (Exception $ex) {
        throw DBALException::driverExceptionDuringQuery($this->_driver, $ex, $args[0]);
    }

    $statement->setFetchMode($this->defaultFetchMode);

    if ($logger) {
        $logger->stopQuery();
    }

    return $statement;
}
```

Doctrine\DBAL\Connections\MasterSlaveConnection:
```php
/**
 * {@inheritDoc}
 */
public function query()
{
    $this->connect('master');

    $args = func_get_args();

    $logger = $this->getConfiguration()->getSQLLogger();
    if ($logger) {
        $logger->startQuery($args[0]);
    }

    $statement = $this->_conn->query(...$args);

    if ($logger) {
        $logger->stopQuery();
    }

    return $statement;
}
```

morozov:
@fredeastside please feel free to provide a pull request with a fix and a test for it.

greg0ire:
> But it seems like copy-paste

That's a bit weird indeed… it's like that since e7d6fc2cf6b1393b1a8095465f5909c3ccaf3a7e , but I cannot find an explanation with that commit.

greg0ire:
Also, duplicates #2769 ?

morozov:
Irrelevant as of https://github.com/doctrine/dbal/pull/4128.

PR #3127, opened 2018-05-02
fix for #3118

fredeastside:


lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -374,7 +376,13 @@ public function query()
             $logger->startQuery($args[0]);
         }
 
-        $statement = $this->_conn->query(...$args);
+        try {
+            $statement = $this->_conn->query(...$args);
+        } catch (\Throwable $ex) {
```
morozov: We probably need to catch a more specific type of exception here. Whatever issues unrelated to the connection failure should not be converted to the driver exception (e.g. all types of `Error`).

tests/Doctrine/Tests/DBAL/Connections/MasterSlaveConnectionTest.php
```diff
@@ -0,0 +1,64 @@
...
+            ->will($this->returnValue($this->createMock(Statement::class)))
+        ;
+
+        $driverMock->expects($this->any())
+            ->method('connect')
+            ->will($this->returnValue($driverConnectionMock));
+
+        $connection = new MasterSlaveConnection([
+            'platform' => new MockPlatform(),
+            'driver' => 'pdo_mysql',
```
morozov: Please remove irrelevant parameters. Looks like this is the required minimum:
```php
'platform' => new MockPlatform(),
'driver' => null,
'master' => [],
'slaves' => [
    [
    ],
],
```

tests/Doctrine/Tests/DBAL/Connections/MasterSlaveConnectionTest.php
```diff
@@ -0,0 +1,64 @@
...
+                ],
+            ],
+        ], $driverMock);
+
+        $this->assertTrue($connection->ping());
+
+        $driverConnectionMock
+            ->expects($this->any())
+            ->method('query')
+            ->willThrowException(new Exception('Query error.'))
```
morozov: The connection cannot throw an arbitrary exception. If it does, it's against the API and should be fixed.

morozov (changes requested):

morozov:
Closing as stale.

PR #3128, opened 2018-05-03
Add MySQL 8 reserved keywords

mlocati:
MySQL 8 added a few reserved keywords.
Comparing https://dev.mysql.com/doc/refman/8.0/en/keywords.html with [MySQL57Keywords.php](https://github.com/doctrine/dbal/blob/e4af109a206352e27c1ddab0d438c8e5f3a80d1c/lib/Doctrine/DBAL/Platforms/Keywords/MySQL57Keywords.php) here's the list of the new keywords:

- `ADMIN`
- `CUBE`
- `CUME_DIST`
- `DENSE_RANK`
- `EMPTY`
- `EXCEPT`
- `FIRST_VALUE`
- `FUNCTION`
- `GROUPING`
- `GROUPS`
- `JSON_TABLE`
- `LAG`
- `LAST_VALUE`
- `LEAD`
- `NTH_VALUE`
- `NTILE`
- `OF`
- `OVER`
- `PERCENT_RANK`
- `PERSIST`
- `PERSIST_ONLY`
- `RANK`
- `RECURSIVE`
- `ROW`
- `ROWS`
- `ROW_NUMBER`
- `SYSTEM`
- `WINDOW`

What should I write in the `@since` phodoc?

PS: it would be nice to add this to the 2.5.x series, for projects using PHP older than 7.1

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/Keywords/MySQL80Keywords.php
+++ b/lib/Doctrine/DBAL/Platforms/Keywords/MySQL80Keywords.php
@@ -1,0 +1,81 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Platforms\Keywords;
+
+use function array_merge;
+
+/**
+ * MySQL 8.0 reserved keywords list.
+ *
+ * @link   www.doctrine-project.org
+ */
+class MySQL80Keywords extends MySQL57Keywords
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        return 'MySQL80';
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @link https://dev.mysql.com/doc/refman/8.0/en/keywords.html
+     */
+    protected function getKeywords()
+    {
+        $keywords = parent::getKeywords();
+
+        $keywords = array_merge($keywords, [
+            'ADMIN',
+            'CUBE',
+            'CUME_DIST',
+            'DENSE_RANK',
+            'EMPTY',
+            'EXCEPT',
+            'FIRST_VALUE',
+            'FUNCTION',
+            'GROUPING',
+            'GROUPS',
+            'JSON_TABLE',
+            'LAG',
+            'LAST_VALUE',
+            'LEAD',
+            'NTH_VALUE',
+            'NTILE',
+            'OF',
+            'OVER',
+            'PERCENT_RANK',
+            'PERSIST',
+            'PERSIST_ONLY',
+            'RANK',
+            'RECURSIVE',
+            'ROW',
+            'ROWS',
+            'ROW_NUMBER',
+            'SYSTEM',
+            'WINDOW',
+        ]);
+
+        return $keywords;
+    }
+}
--- a/docs/en/reference/platforms.rst
+++ b/docs/en/reference/platforms.rst
@@ -35,6 +35,7 @@ MySQL
 
 -  ``MySqlPlatform`` for version 5.0 and above.
 -  ``MySQL57Platform`` for version 5.7 (5.7.9 GA) and above.
+-  ``MySQL80Platform`` for version 8.0 (8.0 GA) and above.
 
 Oracle
 ^^^^^^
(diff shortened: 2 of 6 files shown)


lib/Doctrine/DBAL/Platforms/Keywords/MySQL8Keywords.php
```diff
@@ -0,0 +1,311 @@
...
+
+namespace Doctrine\DBAL\Platforms\Keywords;
+
+/**
+ * MySQL 8 reserved keywords list.
+ *
+ * @author Michele Locati <michele@locati.it>
+ * @link   www.doctrine-project.org
+ */
+class MySQL8Keywords extends MySQL57Keywords
```
Majkl578: To follow similar naming pattern, this class should be named MySQL80Keywords.
Also if possible, should also only add new keywords (or remove removed ones), like here: https://github.com/doctrine/dbal/blob/9f3188ab3d56db6f6fc76c2ddf6b0b4aae06192f/lib/Doctrine/DBAL/Platforms/Keywords/PostgreSQL94Keywords.php#L49-L55

lib/Doctrine/DBAL/Platforms/MySQL8Platform.php
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Doctrine\DBAL\Platforms;
+
+/**
+ * Provides the behavior, features and SQL dialect of the MySQL 8.0 (8.0 GA) database platform.
+ *
+ * @author Michele Locati <michele@locati.it>
+ * @link   www.doctrine-project.org
+ */
+class MySQL8Platform extends MySqlPlatform
```
Majkl578: To follow similar naming pattern, this class should be named MySQL80Platform. Also should extend MySQL57Platform and only expand it.

lib/Doctrine/DBAL/Platforms/Keywords/MySQL8Keywords.php
mlocati: > To follow similar naming pattern, this class should be named MySQL80Keywords.

:+1:

> Also if possible, should also only add new keywords (or remove removed ones), like here:

What about the current `MySQL57Keywords` class? Should it be updated to have the same behavior vs its underlying `MySQLKeywords` class

lib/Doctrine/DBAL/Platforms/MySQL8Platform.php
mlocati: 👍

lib/Doctrine/DBAL/Platforms/Keywords/MySQL8Keywords.php
mlocati: Furthermore: if a keywords will be added to `MySQL57Keywords`, and it's already present in `MySQL80Keywords`, with the [approach you suggest](https://github.com/doctrine/dbal/blob/9f3188ab3d56db6f6fc76c2ddf6b0b4aae06192f/lib/Doctrine/DBAL/Platforms/Keywords/PostgreSQL94Keywords.php#L49-L55) we'll have duplicated kewords: should be call `array_unique` (and `array_values` to keep the array indexes sane)?

lib/Doctrine/DBAL/Platforms/Keywords/MySQL80Keywords.php
```diff
@@ -0,0 +1,84 @@
...
+            'ROWS',
+            'ROW_NUMBER',
+            'SYSTEM',
+            'WINDOW',
+        ]);
+
+        // Avoid having duplicated keywords in the list of the underlying class is updated
+        $keywords = array_unique($keywords);
+        // Sort the final list and reset the array indexes
+        sort($keywords);
```
morozov: It doesn't seem to be required. Unless it is, please remove.

lib/Doctrine/DBAL/Platforms/Keywords/MySQL80Keywords.php
```diff
@@ -0,0 +1,84 @@
...
+            'RECURSIVE',
+            'ROW',
+            'ROWS',
+            'ROW_NUMBER',
+            'SYSTEM',
+            'WINDOW',
+        ]);
+
+        // Avoid having duplicated keywords in the list of the underlying class is updated
+        $keywords = array_unique($keywords);
```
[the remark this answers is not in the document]
mlocati: For instance, if `Groups` becomes a reserved word for MySQL 5.7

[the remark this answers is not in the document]
morozov: It shouldn't happen if they follow semver. If it does, both MySQL 5.7 and MySQL 8.0 will have to be updated.

lib/Doctrine/DBAL/Platforms/Keywords/MySQL80Keywords.php
mlocati: Ok, let's assume that Oracle follows semver...

mlocati:
> @mlocati please fix the coding style violations.

Done.

mlocati:
> Please squash them since the commit history is not really relevant in this case.

Done.

MisatoTremor (approved):

morozov (approved):

morozov:
Thanks @mlocati!

mlocati:
Will this code be added to the 2.5.x series too? For projects using PHP older than 7.1...

Ocramius:
No, all improvements are only forward-ported. Backporting is only for bug-fixes that are really worth backporting.

iLevye:
From the [mysql docs](https://dev.mysql.com/doc/refman/8.0/en/keywords.html):

> ADMIN  became nonreserved in 8.0.12

mlocati:
> No, all improvements are only forward-ported. Backporting is only for bug-fixes that are really worth backporting.

@Ocramius Consider the case of concrete5 (an open source CMS with tons of users), which uses a table called `Groups`: without this fix it won't run. So for me it'd be worth backporting it...

morozov:
@iLevye I think it won't hurt if `ADMIN` stays marked as a reserved keyword. The DBAL doesn't prevent them from being used. On the contrary, it adds additional quotes around them to make the query work. The same quotation syntax can be used for any object name, not only for reserved keywords.

Ocramius:
> @Ocramius Consider the case of concrete5 (an open source CMS with tons of users), which uses a table called `Groups`: without this fix it won't run. So for me it'd be worth backporting it...

Upgrade the DBAL version of concrete5

morozov:
@mlocati In my own experience, the DBAL is extensible enough to let you use custom `Connection`, `Statement` and `Platform` implementations. In the project I'm working on, we use this approach to introduce the needed changes in the DBAL which have not been yet released. Then we upgrade and remove custom implementations.

KorvinSzanto:
> Upgrade the DBAL version of concrete5

This is absolutely something we'll do in v9 of concrete5 as long as it ends up with a minimum version higher than PHP 7.1.x, but we can't do that in any v8.x versions since it'd be a backwards incompatible change.

With that in mind, I think adding a layer on our end would be a reasonable solution.

Ocramius:
> since it'd be a backwards incompatible change.

https://www.doctrine-project.org/2017/07/25/php-7.1-requirement-and-composer.html

This depends on whether you are pushing a library or a deployable:

 * in a deployable, any change in `composer.lock` is a BC break (everything, really)
 * in a library, you can bump dependencies in every minor release

KorvinSzanto:
You're right that composer based installs will just automatically get the latest version already. The problem is that concrete5 is largely distributed as a prebuilt zip file set at minimum requirements.

mlocati:
> @mlocati In my own experience, the DBAL is extensible enough to let you use custom Connection, Statement and Platform implementations.

@morozov Thanks for pointimg me to the right direction: you are correct, extending DBAL is very easy (see https://github.com/concrete5/concrete5/pull/6753 )

morozov:
@mlocati, awesome! Problem solved.

Issue #3124, opened 2018-05-01
Split DBALException into smaller specific exceptions, consider dropping "Exception" suffixes


Majkl578:
Much like ORM has split ORMException into smaller specific exceptions and made ORMException a marker interface (doctrine/doctrine2#6743), DBAL should follow this principle as well.

TBD

morozov:
Resolved by #3525.

PR #3131, opened 2018-05-04
[3.0] Reorganize exceptions (part 1 - specialized exceptions)

Majkl578:
Extracting all exception factory methods to specific sub-classes, keeping the type hiearchy intact. These methods are now removed (thus BC break).
No test changes except mocks or specific factory tests.

In the next PR(s) these (now empty) exception classes will be converted to interfaces and moved to their appropriate namespaces.

First part to complete #3124.
UPGRADE update will come with the 2nd part.


Majkl578:
@morozov Please take a look at this, currently one extracted/removed method per commit. Once reviewed, I'll squash them.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -211,7 +218,7 @@ public function __construct(array $params, Driver $driver, Configuration $config
 
         if (isset($params["platform"])) {
             if ( ! $params['platform'] instanceof Platforms\AbstractPlatform) {
-                throw DBALException::invalidPlatformType($params['platform']);
+                throw InvalidPlatformType::new($params['platform']);
```
morozov: Would it make sense to change the constructor signature and get rid of the exception?

lib/Doctrine/DBAL/Connection.php
```diff
@@ -656,7 +663,7 @@ private function gatherConditions(array $identifiers)
     public function delete($tableExpression, array $identifier, array $types = [])
     {
         if (empty($identifier)) {
-            throw InvalidArgumentException::fromEmptyCriteria();
+            throw EmptyCriteriaNotAllowed::new();
```
morozov: Personally, I'd better not have this check at all. `DELETE FROM table` is a valid SQL expression. Sometimes you just need to remove all records from a cache/projection table. Why should we not allow it? Probably a topic for another discussion.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -937,7 +944,7 @@ public function executeCacheQuery($query, $params, $types, QueryCacheProfile $qc
     {
         $resultCache = $qcp->getResultCacheDriver() ?: $this->_config->getResultCacheImpl();
         if ( ! $resultCache) {
```
morozov: It would be simpler if enforced by the interfaces of `$qcp` and the config. E.g they both are required to have an instance of the cache driver which defaults to a null-implementation which throws exceptions from any method.

morozov (commented):
Structure-wise, the changes look fine but therу are cases (I just commented on a few) where exceptions could be eliminated by introducing proper APIs.

Not sure how we want to proceed with that but I feel that without doing so we'll just introduce a bunch of useless classes.

Let me know what you think.

Majkl578:
> there are cases (I just commented on a few) where exceptions could be eliminated by introducing proper APIs.

Agreed.

> Not sure how we want to proceed with that

At this moment I think it's easier to keep it as is and refactor later, either as part of introducing strict types or as separate improvements. Also when an exception becomes useless during that process, it can simply be removed, it's not going to a stable release anytime soon. (And we may eventually mark them as `@internal` before 3.0 anyway.)
Do you want me to open issues to track those possible improvements?

morozov:
A few issues with the current structure (not necessarily affected by this patch):

exceptions

1. All transaction-related exceptions extend `ConnectionException` which probably could renamed to `TransactionException`.
2. `NonUniqueFieldNameException` directly extending `ServerException` is weird.
3. `TableExistsException` extending `ServerException` is weird too.
4. `ServerException` extending `DriverException`
5. More like these.

@Majkl578 please file one or two tiсkets for 3.0 abour cleaning up and restructuring exceptions. Besides that, it looks good!

morozov (approved):

Majkl578:
@morozov At this moment I only extracted the methods to specific classes.
The plan for next PR is to make them interfaces and move them around to their appropriate namespaces, also split some of them or slightly change their hiearchy as we discussed earlier. Looking forward to discussing this then. :+1:

morozov:
@Majkl578 what do you say about the Scrutinizer issues like the one in [`CommitFailedRollbackOnly`](https://scrutinizer-ci.com/g/doctrine/dbal/inspections/7fe58d2c-2fd6-43a9-84b3-09f16446b2df/issues/files/lib/Doctrine/DBAL/Exception/CommitFailedRollbackOnly.php?status=new&orderField=path&order=asc&honorSelectedPaths=0)? There are indeed two classes with the same name within the same hierarchy:
```
↪ find -name ConnectionException.php
./lib/Doctrine/DBAL/Exception/ConnectionException.php
./lib/Doctrine/DBAL/ConnectionException.php
```
Also, please squash.

Majkl578:
> There are indeed two classes with the same name within the same hierarchy:

Correct, that is also true for 2.x. We should fix that in the 2nd part following this PR.

lib/Doctrine/DBAL/Connection.php
Majkl578: moved to #3163

lib/Doctrine/DBAL/Connection.php
Majkl578: #3164

lib/Doctrine/DBAL/Connection.php
Majkl578: moved to #3165

Majkl578:
@Ocramius WDYT?

Ocramius (changes requested):
Overall, patch looks good, but it needs testing of the new exception methods where decisions are being taken. Specifically:
 1. where we have logical switches
 2. where the method allows union typed parameters

Majkl578:
@Ocramius Are you asking me to actually add these missing tests here? Because all I did was basically moving these methods from a _god exceptions_ to their specific exception classes + named constructors... :roll_eyes: :no_mouth:

PR #3133, opened 2018-05-06
Travis DB2

Majkl578:
_very broken, much cool_

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/DB2Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/DB2Platform.php
@@ -42,6 +42,14 @@ class DB2Platform extends AbstractPlatform
     /**
      * {@inheritdoc}
      */
+    public function getCharMaxLength() : int
+    {
+        return 254;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function getBinaryMaxLength()
     {
         return 32704;
@@ -53,6 +61,19 @@ class DB2Platform extends AbstractPlatform
     public function getBinaryDefaultLength()
     {
         return 1;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getVarcharTypeDeclarationSQL(array $field)
+    {
+        // for IBM DB2, the CHAR max length is less than VARCHAR default length
+        if (! isset($field['length']) && ! empty($field['fixed'])) {
+            $field['length'] = $this->getCharMaxLength();
+        }
+
+        return parent::getVarcharTypeDeclarationSQL($field);
     }
 
     /**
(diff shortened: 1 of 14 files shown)


Majkl578:
Rebased onto current master.

tests/travis/install-db2.sh
```diff
@@ -0,0 +1,22 @@
...
+sudo docker run \
+    -d \
+    -p 50000:50000 \
+    -e DB2INST1_PASSWORD=Doctrine2018 \
+    -e LICENSE=accept \
+    --name db2 \
+    ibmcom/db2express-c:10.5.0.5-3.10.0 \
+    db2start
+
+sleep 60
```
morozov: This may be too much but `sleep 5` became not enough when I added creation of the tablespace. We could try looking for a more elegant solution.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -107,18 +128,10 @@ public function isCommentedDoctrineType(Type $doctrineType)
...
+        return $fixed ? ($length ? 'CHAR(' . $length . ')' : 'CHAR(254)')
                 : ($length ? 'VARCHAR(' . $length . ')' : 'VARCHAR(255)');
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    protected function getBinaryTypeDeclarationSQLSnippet($length, $fixed)
-    {
-        return $fixed ? 'BINARY(' . ($length ?: 255) . ')' : 'VARBINARY(' . ($length ?: 255) . ')';
```
morozov: Most likely, binary data can be safely stored in "char" fields in DB2. For now, just disabling the type since it's broken anyways.

lib/Doctrine/DBAL/Types/DateIntervalType.php
```diff
@@ -25,7 +26,7 @@ public function getName()
      */
     public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform)
     {
-        $fieldDeclaration['length'] = 255;
+        $fieldDeclaration['length'] = min($platform->getCharMaxLength(), 255);
         $fieldDeclaration['fixed']  = true;
```
morozov: According to the format, the interval is definitely not a 255-char fixed length string.

@Majkl578 would `CHAR(255)` → `VARCHAR(32)` be a breaking change given that the current type is far from truth? `VARCHAR(128)` still should be safe and portable according to https://github.com/doctrine/dbal/pull/2316/files#r53146155.

Majkl578:
Added PHP 7.2 stage, [should be supported now](https://pecl.php.net/package/ibm_db2/2.0.3).

morozov (approved):

PR #3134, opened 2018-05-08
Fix numeric comparison for sequence initial value

maxperrimond:
Small fix to avoid a diff on sequences cause by a numeric string/integer comparison.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Comparator.php
+++ b/lib/Doctrine/DBAL/Schema/Comparator.php
@@ -189,11 +189,8 @@ class Comparator
      */
     public function diffSequence(Sequence $sequence1, Sequence $sequence2)
     {
-        if ($sequence1->getAllocationSize() != $sequence2->getAllocationSize()) {
-            return true;
-        }
-
-        return $sequence1->getInitialValue() != $sequence2->getInitialValue();
+        return intval($sequence1->getAllocationSize()) !== intval($sequence2->getAllocationSize())
+            || intval($sequence1->getInitialValue()) !== intval($sequence2->getInitialValue());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
```diff
@@ -435,8 +435,8 @@ public function testCompareSequences()
         $seq1 = new Sequence('foo', 1, 1);
         $seq2 = new Sequence('foo', 1, 2);
         $seq3 = new Sequence('foo', 2, 1);
-        $seq4 = new Sequence('foo', "1", "1");
-        $seq5 = new Sequence('foo', "10", "1");
+        $seq4 = new Sequence('foo', '1', '1');
+        $seq5 = new Sequence('foo', '10', '1');
```
Ocramius: Can you please add the scenario where the comparator was failing?

Ocramius (changes requested):

Majkl578:
Wasn't this fixed by  #3086?

lib/Doctrine/DBAL/Schema/Comparator.php
```diff
@@ -189,11 +189,8 @@ private function isAutoIncrementSequenceInSchema($schema, $sequence)
      */
     public function diffSequence(Sequence $sequence1, Sequence $sequence2)
     {
-        if ($sequence1->getAllocationSize() != $sequence2->getAllocationSize()) {
-            return true;
-        }
-
-        return $sequence1->getInitialValue() !== $sequence2->getInitialValue();
+        return intval($sequence1->getAllocationSize()) !== intval($sequence2->getAllocationSize())
+            || intval($sequence1->getInitialValue()) !== intval($sequence2->getInitialValue());
```
morozov: The `Sequence` is a DBAL-level type which should enforce proper types of its internal values. If it doesn't, it should be fiexed in the `Sequence`, not in `Comparator`. Otherwise, we'll end up having type casting across the entire codebase.

maxperrimond:
As mentioned by @Majkl578, it was already fixed by a previous PR. So no need this change anyway.

PR #3137, opened 2018-05-10
Workaround for the inability to use a post-PHPUnit script on ContinuousPHP

morozov:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | none

#### Summary

According to the ContinuousPHP support, currently, it's impossible to specify a post-PHPUnit script in the build configuration. A workaround is implemented to do that as part of the PHPUnit run.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -21,7 +21,7 @@ before_commands:
 tools:
     external_code_coverage:
         timeout: 3600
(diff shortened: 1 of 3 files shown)


Majkl578 (approved):
What a nice dirty hack, I like it!

Issue #2426, opened 2016-06-25
SQLITE Autoincrement


s0me0ther:
doctrine: 2.5.4
sqlite: 3

The [docs](http://docs.doctrine-project.org/en/latest/reference/basic-mapping.html#identifier-generation-strategies) are talking about, that this annotation should result in an SQLITE Autoincrement ID field:

```
    /**
     * @var int
     * @ORM\Id
     * @ORM\Column(type="integer")
     * @ORM\GeneratedValue(strategy="AUTO")
     */
```

The id is created, but the autoincrement flag is not checked in the sqlitedb..

if i insert an new entity on a fresh db it will have id 1
if i delte this entity and create a new entity afterwards, the new entity will also have the id 1

do i miss somethis or is this a bug in doctrine?

sebka:
i just searched thru the code and found that:
```
    /**
     * @var int
     * @ORM\Id
     * @ORM\Column(type="integer", options={"autoincrement":true})
     * @ORM\GeneratedValue(strategy="AUTO")
     */
```
will do the job.

deeky666:
can you please provide the DDL that was used to create the table?

sebka:
This Query will be executed:
`CREATE TABLE jobs (id INTEGER NOT NULL, created DATETIME NOT NULL, keyword VARCHAR(255) NOT NULL, weight INTEGER DEFAULT 1000 NOT NULL, status VARCHAR(255) NOT NULL, master BOOLEAN DEFAULT '0' NOT NULL, PRIMARY KEY(id));`

From this Entity:
```<?php

namespace Entity;

use Doctrine\ORM\Mapping as ORM;

/**
 *
 * @ORM\Entity
 * @ORM\Table(name="jobs")
 * @ORM\HasLifecycleCallbacks
 */
class Job
{
    /**
     * @var int
     * @ORM\Id
     * @ORM\Column(type="integer", options={"autoincrement":true})
     * @ORM\GeneratedValue(strategy="AUTO")
     */
    protected $id;

    /**
     * @var \DateTime
     * @ORM\Column(type="datetime", nullable=false, unique=false)
     */
    protected $created;
    
    /**
     * @var string
     * @ORM\Column(type="string", nullable=false, unique=false)
     */
    protected $keyword;

    /**
     * @var int
     * @ORM\Column(type="integer", nullable=false, unique=false, options={"default":1000})
     */
    protected $weight;

    /**
     * @var string
[...]

PR #3141, opened 2018-05-11
allow creating PRIMARY KEY AUTOINCREMENT fields for sqlite (unit tests)

TimoBakx:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes?
| Fixed issues | #1861, #2426 

#### Summary

These are the updated unit tests for PR #2921.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
@@ -307,9 +307,9 @@ class SqlitePlatform extends AbstractPlatform
      */
     protected function _getCommonIntegerTypeDeclarationSQL(array $columnDef)
     {
-        // sqlite autoincrement is implicit for integer PKs, but not when the field is unsigned
+        // sqlite autoincrement is only possible for the primary key
         if ( ! empty($columnDef['autoincrement'])) {
-            return '';
+            return ' PRIMARY KEY AUTOINCREMENT';
         }
 
         return ! empty($columnDef['unsigned']) ? ' UNSIGNED' : '';
@@ -345,7 +345,10 @@ class SqlitePlatform extends AbstractPlatform
 
         if (isset($options['primary']) && ! empty($options['primary'])) {
             $keyColumns = array_unique(array_values($options['primary']));
-            $queryFields.= ', PRIMARY KEY('.implode(', ', $keyColumns).')';
+            // the primary key is already defined in the column definition for auto increment fields
+            if (count($keyColumns) > 1 || !$columns[$keyColumns[0]]['autoincrement']) {
+                $queryFields .= ', PRIMARY KEY(' . implode(', ', $keyColumns) . ')';
+            }
         }
 
         if (isset($options['foreignKeys'])) {
--- a/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
@@ -343,13 +343,7 @@ class SqlitePlatform extends AbstractPlatform
             }
         }
 
-        if (isset($options['primary']) && ! empty($options['primary'])) {
-            $keyColumns = array_unique(array_values($options['primary']));
-            // the primary key is already defined in the column definition for auto increment fields
-            if (count($keyColumns) > 1 || !$columns[$keyColumns[0]]['autoincrement']) {
-                $queryFields .= ', PRIMARY KEY(' . implode(', ', $keyColumns) . ')';
-            }
-        }
+        $queryFields .= $this->getNonAutoincrementPrimaryKeyDefinition($columns, $options);
 
         if (isset($options['foreignKeys'])) {
             foreach ($options['foreignKeys'] as $foreignKey) {
@@ -376,6 +370,29 @@ class SqlitePlatform extends AbstractPlatform
         }
 
         return $query;
+    }
+
+    /**
+     * Generate a PRIMARY KEY definition if no autoincrement value is used
+     *
+     * @param string[] $columns
+     * @param mixed[]  $options
+     */
+    private function getNonAutoincrementPrimaryKeyDefinition(array $columns, array $options) : string
+    {
+        if (empty($options['primary'])) {
+            return '';
+        }
+
+        $keyColumns = array_unique(array_values($options['primary']));
+
+        foreach ($keyColumns as $keyColumn) {
+            if (isset($columns[$keyColumn]['autoincrement']) && ! empty($columns[$keyColumn]['autoincrement'])) {
+                return '';
+            }
+        }
+
+        return ', PRIMARY KEY(' . implode(', ', $keyColumns) . ')';
     }
 
     /**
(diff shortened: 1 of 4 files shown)


tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
```diff
@@ -232,4 +232,63 @@ public function getDiffListIntegerAutoincrementTableColumnsData()
             array('bigint', true, true),
         );
     }
+
+    /**
+     * @group DBAL-2921
+     */
+    public function testPrimaryKeyAutoIncrement()
```
morozov: @TimoBakx thank you for the patch.

There doesn't seem to be anything SQLite-specific in this test which is good. Could you try moving it to a platform-agnostic test case (e.g. `SchemaManagerFunctionalTestCase`) to make sure it works the same on all platforms which support autoincrement fields?

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
```diff
@@ -232,4 +232,63 @@ public function getDiffListIntegerAutoincrementTableColumnsData()
...
+        $table->addColumn('id', 'integer', [
+            'autoincrement' => true,
+        ]);
+        $table->addColumn('text', 'text');
+        $table->setPrimaryKey(['id']);
+        $this->_sm->dropAndCreateTable($table);
+
+        $this->_conn->insert('test_pk_auto_increment', ['text' => '1']);
+        $this->_conn->insert('test_pk_auto_increment', ['text' => '2']);
+        $this->_conn->insert('test_pk_auto_increment', ['text' => '3']);
```
morozov: If one record is enough to test that autoincrement works in fact, please leave one out of three.

TimoBakx: Good point. I removed two lines in this test and in `testPrimaryKeyNoAutoIncrement`.

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
TimoBakx: Good idea. It took a bit of tweaking, but it now runs (and passes) on all tests.

TimoBakx:
@morozov Thank you for your review.

I made some changes. Could you please review again?
Thanks in advance.

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -345,7 +346,10 @@ protected function _getCreateTableSQL($name, array $columns, array $options = []
 
         if (isset($options['primary']) && ! empty($options['primary'])) {
             $keyColumns = array_unique(array_values($options['primary']));
-            $queryFields.= ', PRIMARY KEY('.implode(', ', $keyColumns).')';
+            // the primary key is already defined in the column definition for auto increment fields
+            if (count($keyColumns) > 1 || ! $columns[$keyColumns[0]]['autoincrement']) {
+                $queryFields .= ', PRIMARY KEY(' . implode(', ', $keyColumns) . ')';
```
morozov: Do we still need to use two syntaxes? The `PRIMARY KEY(x, y, z)` looks more generic since it can describe multiple fields. Given that we take this case into account, should we use just the generic syntax.

This `$columns[$keyColumns[0]]` doesn't look good. What if the key consists of two columns where the second one has autoincrement?

morozov: If that's the case, can we leave the `AUTOINCREMENT` keyword with the field and define the PK later? Otherwise, it's unclear if it's possible and how it will be handled if one of the columns is autoincremented but the PK consists of two columns.

TimoBakx: Unfortunately, it only works when the field definition is `INTEGER PRIMARY KEY AUTOINCREMENT` as a whole. In that specific order.

morozov: Still, we need to figure out how this is going to be handled if at all possible:
> one of the columns is autoincremented but the PK consists of two columns.

The `0` in `count($keyColumns) > 1 || ! $columns[$keyColumns[0]]['autoincrement']` looks like a magic constant.

TimoBakx: I agree. I'll run some tests to see if I can improve that.

TimoBakx: I've made some improvements. This should cover it.

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -375,6 +373,31 @@ protected function _getCreateTableSQL($name, array $columns, array $options = []
...
+    {
+        if (! isset($options['primary']) || empty($options['primary'])) {
+            return '';
+        }
+
+        $keyColumns = array_unique(array_values($options['primary']));
+
+        foreach ($keyColumns as $keyColumn) {
+            if (isset($columns[$keyColumn]['autoincrement']) && ! empty($columns[$keyColumn]['autoincrement'])) {
+                return '';
```
morozov: Could you check if it's technically possible to have a PK which consists of multiple columns one or more of which has `AUTOINCREMENT`? If yes, then it would be a breaking change since in this case, the PK will consist only declared for the autoincremented column.

TimoBakx: That's currently not possible. It will result in a `MappingException` while generating the scheme:
"Single id is not allowed on composite primary key in entity...."

morozov: Then we're not breaking anything. Sounds good.

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -375,6 +372,31 @@ protected function _getCreateTableSQL($name, array $columns, array $options = []
...
 
+    /**
+     * Generate a PRIMARY KEY definition if no autoincrement value is used
+     *
+     * @param string[] $columns
+     * @param mixed[] $options
+     *
+     * @return string
+     */
+    private function getNonAutoincrementPrimaryKeyDefinition(array $columns, array $options)
```
morozov: Please return type explicitly instead of an annotation.

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -375,6 +372,31 @@ protected function _getCreateTableSQL($name, array $columns, array $options = []
...
+    {
+        if (! isset($options['primary']) || empty($options['primary'])) {
```
morozov: The combination of `!isset() || empty()` is redundant. It's covered by `empty()`.

morozov (approved):

TimoBakx:
@morozov Thank you for your reviews, insight and patience.

morozov:
Glad it helped. Thank you for your contribution!

Issue #3145, opened 2018-05-13
Index should be fulfilled if its columns match the leftmost part of a longer compound index


fabriciojs:
DBAL does not consider `KEY(A, B)` to fulfill `KEY(A)`, assuming that it should, since the leftmost part matches the whole later index.

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.6.x

# Context

I am working in a project where I stumbled upon a DBAL (2.6) behavior that I couldn't understand why it was built that way (the situation happens at least since 2.5 where duplicated indexes started being identified up to newer versions and `master`).

In this case we use Mysql and I was trying to remove redundant indexes (indexes with columns that match the leftmost part of other indexes).

I noticed that we had a lot of implicit indexes automatically added for foreign keys by Doctrine ORM that became redundant by custom indexes that started with the same column as the foreign key field. So I tried removing those indexes, which Mysql accepts (since we have another index that can be used for the foreign key lookups just the same).

After doing so I noticed that `doctrine:schema:validate` no longer validated my database as in sync with my mapping. Doctrine auto auto-schema mapping is trying to add again the indexes for the foreign keys, even though there were compound indexes that work just the same (all mapped as well).

Since we wanted to help RDMS performance by removing redundant indexes, now the auto schema tools (`doctrine:migration:diff`, `doctrine:schema:update`, `doctrine:schema:validate`) cannot be properly used since they try every time to add dozens of indexes we know we don't want or need.

# How to Reproduce

I am gonna provide and link here a PR with a failing test.

PR: https://github.com/doctrine/dbal/pull/3146

# Investigation

So I narrowed the issue to this check (added back in 2010):

https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/Index.php#L209-L213

# Background

That appears to be there since the beginning of duplicate/overruling indexes: https://github.com/doctrine/dbal/commit/bc9c8c2b139ebee2e94e94ca5b572ee921b336fa

Work has been done since then on best identifying duplicated/redundant indexes (custom or implict), but never touched the compound indexes issue I noticed now.
- https://github.com/doctrine/dbal/pull/764
- https://github.com/doctrine/dbal/pull/769
- https://github.com/doctrine/dbal/commit/0d817986c293893d45a4ec4768558dc7a86b4643

Specifically for FK naming (not our case here, but touches the indexing as well): https://github.com/doctrine/doctrine2/issues/3753

[...]

PR #3147, opened 2018-05-13
Compounded indexes fulfilling

fabriciojs:
This is just a work in progress that aims to be an approach to fix the redundant indexes issue presented at #3145.

|      Q       |   A
|------------- | -----------
| Type         | TBD
| BC Break     | probably not
| Issues | #3145

#### Summary

- Removed the limitation of indexes only fulfilling same size indexes.
- Added tests and special case treatment for old concern that looks like prevented this behavior initially.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,16 +1,3 @@
-# Upgrade to 2.6.2
-
-## MINOR BC BREAK:
-1 ``Doctrine\DBALDBALException::invalidPlatformSpecified()`` now has one argument which takes the invalid platform option.
-
-Before:
-
-    Doctrine\DBALDBALException::invalidPlatformSpecified();
-
-After:
-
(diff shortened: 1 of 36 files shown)


fabriciojs:
Hello @morozov ! I guess the change to this PR on last Nov 20th  - making new commits to appear in the diff as well as changed the target branch - was some automatic housekeeping that went down?

Just wanted to check if this is still worthy pursuing - updating to modern branches - or should we close this after all these years?

morozov:
> I guess the change to this PR on last Nov 20th - making new commits to appear in the diff as well as changed the target branch - was some automatic housekeeping that went down?

I believe so. I removed all unmaintained branches around that time.

> Just wanted to check if this is still worthy pursuing - updating to modern branches - or should we close this after all these years?

If the issue is still relevant, I believe it's worth pursuing.

PR #3149, opened 2018-05-13
Introduced binary binding type to support binary parameters on Oracle

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | #2787

#### Summary

When binding parameters to a statement, `oci8` requires a special binding type for binary parameters (`OCI_B_BIN`).

### DIFF
--- a/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
+++ b/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
@@ -25,6 +25,7 @@ use Doctrine\DBAL\FetchMode;
 use Doctrine\DBAL\ParameterType;
 use IteratorAggregate;
 use const OCI_ASSOC;
+use const OCI_B_BIN;
 use const OCI_B_BLOB;
 use const OCI_BOTH;
 use const OCI_D_LOB;
@@ -35,6 +36,7 @@ use const OCI_RETURN_LOBS;
 use const OCI_RETURN_NULLS;
 use const OCI_TEMP_BLOB;
 use const PREG_OFFSET_CAPTURE;
+use const SQLT_CHR;
 use function array_key_exists;
 use function count;
 use function implode;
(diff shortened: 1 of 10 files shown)


lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -301,18 +303,35 @@ public function bindParam($column, &$variable, $type = ParameterType::STRING, $l
...
+            $variable,
+            $length ?? -1,
+            $this->convertParamType($type)
+        );
+    }
+
+    /**
+     * Converts DBAL parameter type to oci8 parameter type
+     */
+    private function convertParamType(int $type) : int
```
Majkl578: `convertParameterType()`

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -301,18 +303,35 @@ public function bindParam($column, &$variable, $type = ParameterType::STRING, $l
...
+    {
+        switch ($type) {
```
Majkl578: This could also be written as:
```php
return [
    ParameterType::BINARY => OCI_B_BIN,
    ParameterType::LARGE_OBJECT => OCI_B_BLOB,
][$type] ?? SQLT_CHR;
```
Up to you which variant you prefer.

Majkl578 (commented):
Looks good, I'd prefer some comment why references are required, it may be a bit confusing.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
morozov: I'll keep it implemented using `switch`. In PHP 7, the lookup is the same efficient as using an array, it's more maintainable (you can easily add logic per key) and it doesn't initialize a temporary array upon each call.

morozov:
> I'd prefer some comment why references are required, it may be a bit confusing.

Where exactly? Referencing variables in `OCI8Statement::$boundValues` is indeed non-obvious, but there's already a commend in the property declaration. Using references to the variables bound as parameters is by design and consistent across the platforms.

Majkl578 (approved):

Majkl578:
> Where exactly?

Whole OCI8Statement looks quite magical due to those references, if that is by design, fine.

PR #3150, opened 2018-05-14
Fix sp_rename in dbs with case sensitive collation

lucassch:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

In MSSQLs DBs that use case sensitive collation, the command `sp_RENAME` will not work, so i'm changing SQLServerPlatform.php it to use  `sp_rename` instead.


morozov:
Given that all existing tests pass, we need a new one which would cover the code change.

morozov:
This issue is already addressed by https://github.com/doctrine/dbal/pull/4238.

PR #3151, opened 2018-05-14
MSSQL DLL 5.2.0 has been released.

photodude:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | was using the RC version of the MSSQL DLL

#### Summary

uses the release version rather than the outdated RC1 version

Release DLL info https://pecl.php.net/package-changelog.php?package=sqlsrv&release=5.2.0

PR #3153, opened 2018-05-15
Don't use the RC versions of sqlsrv/pdo_sqlsrv in AppVeyor

mlocati:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

AppVeyor is currently configured to test the 5.2.0rc1 versions of the `sqlsrv`/`pdo_sqlsrv` PHP extensions.
BTW the very latest version is now [5.2.0](https://github.com/Microsoft/msphpsql/releases): let's use it.
This is done by using the [PhpManager](https://github.com/mlocati/powershell-phpmanager) PowerShell module, that's able to automatically install the most recent stable version of PHP extensions, without the need of manually specifying their download URL and version.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -83,9 +83,6 @@ install:
           Install-PhpExtension -Extension xdebug -DontEnable
           Write-Host "xdebug version: $((Get-PhpExtension | Where-Object { $_.Handle -eq 'xdebug' }).Version)"
 
-          # install sqlite
-          appveyor-retry cinst -y sqlite
-
           # download Composer
           if (!(Test-Path C:\tools\composer)) {
             New-Item -Path c:\tools -Name composer -ItemType Directory | Out-Null
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,6 +9,7 @@ clone_depth: 2
 cache:
   - C:\ProgramData\chocolatey\bin -> .appveyor.yml
   - C:\ProgramData\chocolatey\lib -> .appveyor.yml
+  - '%ProgramFiles%\WindowsPowerShell\Modules\PhpManager -> .appveyor.yml'
   - C:\tools\php -> .appveyor.yml
   - C:\tools\composer -> .appveyor.yml
   - C:\tools\ocular -> .appveyor.yml
@@ -46,16 +47,16 @@ init:
 ## Install PHP and composer, and run the appropriate composer command
 install:
     - ps: |
-        # Check if installation is cached
-        if (!(Test-Path c:\tools\php)) {
-
-          # install PhpManager
+        # install PhpManager
+        if (-Not(Get-Module -Name PhpManager -ListAvailable)) {
           if (-Not(Get-PackageProvider -Name NuGet -ErrorAction SilentlyContinue | Where-Object { $_.Version -ge '2.8.5.201'})) {
             Install-PackageProvider -Name NuGet -MinimumVersion 2.8.5.201 -Force | Out-Null
           }
-          if (-Not(Get-Module -ListAvailable | Where-Object { $_.Name -eq 'PhpManager' })) {
-            Install-Module -Name PhpManager -Force -SkipPublisherCheck
-          }
+          Install-Module -Name PhpManager -Force -SkipPublisherCheck
+        }
+
+        # Check if installation is cached
+        if (!(Test-Path c:\tools\php)) {
 
           # install PHP
           Install-Php -Version $env:php -Architecture $env:platform -ThreadSafe $false -Path c:\tools\php -TimeZone UTC -AddToPath System -InitialPhpIni Production
@@ -79,7 +80,7 @@ install:
           Write-Host "sqlsrv version: $((Get-PhpExtension | Where-Object { $_.Handle -eq 'sqlsrv' }).Version)"
           Install-PhpExtension -Extension pdo_sqlsrv
           Write-Host "pdo_sqlsrv version: $((Get-PhpExtension | Where-Object { $_.Handle -eq 'pdo_sqlsrv' }).Version)"
-          Install-PhpExtension -Extension xdebug
+          Install-PhpExtension -Extension xdebug -DontEnable
           Write-Host "xdebug version: $((Get-PhpExtension | Where-Object { $_.Handle -eq 'xdebug' }).Version)"
 
           # install sqlite
@@ -127,7 +128,9 @@ test_script:
       }
   - ps: >-
       if ($env:coverage -eq "yes") {
+        Enable-PhpExtension -Extension xdebug
         vendor\bin\phpunit -c $($env:phpunit_config) --coverage-clover clover.xml
+        Disable-PhpExtension -Extension xdebug
         appveyor-retry ocular code-coverage:upload --format=php-clover clover.xml
       } else {
         vendor\bin\phpunit -c $($env:phpunit_config)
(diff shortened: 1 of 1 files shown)


mlocati:
To speed up AppVeyor we could also install xdebug only if `$Env:coverage` is `yes` (otherwise it's not needed).

.appveyor.yml
```diff
@@ -101,20 +96,12 @@ install:
...
           }
-
-          # download CA bundle
-          if (!(Test-Path C:\tools\cacert)) {
-            New-Item -path c:\tools\ -name cacert -itemtype directory
-          }
-          if (!(Test-Path c:\tools\cacert\bundle.pem)) {
-            appveyor-retry appveyor DownloadFile https://curl.haxx.se/ca/cacert.pem -Filename C:\tools\cacert\bundle.pem
-          }
         }
```
Majkl578: Just noticed now, PHP block should end before installing Composer/Ocular - could you please move it above `download Composer`? Thanks.

Majkl578:
> AppVeyor is currently configured to test the 5.2.0rc1

That is because we started testing against PHP 7.2 at the time when no stable PECL version was compatible.

Otherwise I can't review the rest since I am not a Windows guy.

.appveyor.yml
mlocati: What about `appveyor-retry cinst -y sqlite`?

mlocati:
>> AppVeyor is currently configured to test the 5.2.0rc1
> That is because we started testing against PHP 7.2 at the time when no stable PECL version was compatible.

Yes, the 5.2.0 (not RC) is quite recent.

.appveyor.yml
Majkl578: That should stay as part of the PHP block.
The question is whether we even need it, we don't test against SQLite (although we originally wanted to). Imho we can drop SQLite completely.

morozov:
The pull request header looks misleading in the way that it's not about the RC versions but is about introducing PhpManager. BTW, the extension versions are updated in https://github.com/doctrine/dbal/pull/3151 which looks like a much simpler change.

To the question of using PhpManager, given that our installation process is scripted anyways, what's the real use of having another CI dependency? Instead of just using the PHPs own way of configuring extensions, we'll use a 3rd party tool.

mlocati:
The only reason to use PhpManager is that's really easy to use. BTW obviously it's not mandatory.
Should I close this PR?

mlocati:
That's why there are libraries maintained by people that knows a specific language better...

morozov:
From what I can see, it uses the same PowerShell language but instead of operating on the level of PECL packages and php.ini configuration which everyone knows about it has its own API. This API should be convenient for a daily use as a developer or for scripting different environments but in our case, we have just one environment which is static.

mlocati:
Yes, they are both PowerShell commands, but PhpManager abstracts pear/pecl, and make it very easy to write and maintain PHP installations.

For example, here's how the current DBAL script is translated to PhpManager:

- From  
```powershell
appveyor-retry cinst --params '""/InstallDir:C:\tools\php""' --ignore-checksums -y php --version ((choco search php --exact --all-versions -r | select-string -pattern $env:php | sort { [version]($_ -split '\|' | select -last 1) } -Descending | Select-Object -first 1) -replace '[php|]','')
copy php.ini-production php.ini
Add-Content php.ini "`n date.timezone=UTC"
Add-Content php.ini "`n extension_dir=ext"
```
- To  
```powershell
Install-Php -Version $env:php -Architecture $env:platform -ThreadSafe $false -Path c:\tools\php -TimeZone UTC -AddToPath System -InitialPhpIni Production
```
- From
```powershell
if (!(Test-Path C:\tools\cacert)) {
  New-Item -path c:\tools\ -name cacert -itemtype directory
}
if (!(Test-Path c:\tools\cacert\bundle.pem)) {
  appveyor-retry appveyor DownloadFile https://curl.haxx.se/ca/cacert.pem -Filename C:\tools\cacert\bundle.pem
}
Add-Content php.ini "`n curl.cainfo=C:\tools\cacert\bundle.pem"
```
- To  
```powershell
Update-PhpCAInfo
```
- From
```powershell
Add-Content php.ini "`n memory_limit=1G"
```
- To  
```powershell
Set-PhpIniKey -Key memory_limit -Value 1G
```
- From
```powershell
Add-Content php.ini "`n extension=php_openssl.dll"
Add-Content php.ini "`n extension=php_mbstring.dll"
Add-Content php.ini "`n extension=php_fileinfo.dll"
Add-Content php.ini "`n extension=php_pdo_sqlite.dll"
Add-Content php.ini "`n extension=php_sqlite3.dll"
Add-Content php.ini "`n extension=php_curl.dll"
```
- To  
```powershell
Enable-PhpExtension -Extension openssl
Enable-PhpExtension -Extension mbstring
Enable-PhpExtension -Extension fileinfo
Enable-PhpExtension -Extension pdo_sqlite
Enable-PhpExtension -Extension sqlite3
Enable-PhpExtension -Extension curl
```
- From
```powershell
$DLLVersion = "5.2.0rc1"
cd c:\tools\php\ext
$source = "https://windows.php.net/downloads/pecl/releases/sqlsrv/$($DLLVersion)/php_sqlsrv-$($DLLVersion)-$($env:php)-nts-vc15-x64.zip"
$destination = "c:\tools\php\ext\php_sqlsrv-$($DLLVersion)-$($env:php)-nts-vc15-x64.zip"
Invoke-WebRequest $source -OutFile $destination
7z x -y php_sqlsrv-$($DLLVersion)-$($env:php)-nts-vc15-x64.zip > $null
$source = "https://windows.php.net/downloads/pecl/releases/pdo_sqlsrv/$($DLLVersion)/php_pdo_sqlsrv-$($DLLVersion)-$($env:php)-nts-vc15-x64.zip"
$destination = "c:\tools\php\ext\php_pdo_sqlsrv-$($DLLVersion)-$($env:php)-nts-vc15-x64.zip"
Invoke-WebRequest $source -OutFile $destination
7z x -y php_pdo_sqlsrv-$($DLLVersion)-$($env:php)-nts-vc15-x64.zip > $null
$DLLVersion = "2.6.0"
$source = "https://xdebug.org/files/php_xdebug-$($DLLVersion)-$($env:php)-vc15-nts-x86_64.dll"
$destination = "c:\tools\php\ext\php_xdebug.dll"
Invoke-WebRequest $source -OutFile $destination
Remove-Item c:\tools\php\* -include .zip
Add-Content php.ini "`nextension=php_sqlsrv.dll"
Add-Content php.ini "`nextension=php_pdo_sqlsrv.dll"
Add-Content php.ini "`nzend_extension=php_xdebug.dll"
```
- To  
```powershell
Install-PhpExtension -Extension sqlsrv
Install-PhpExtension -Extension pdo_sqlsrv
Install-PhpExtension -Extension xdebug
```

Furthermore, disabling a PHP extension using `Get-Content`/`Set-Content` would be quite complex, whereas with PhpManager you only have to call `Disable-PhpExtension`.

About the problem of learning a new tool, well, it's not that hard with PowerShell (the first commands you learn when developing PS scripts are `Get-Command -Module <module>` to list the commands offered by a module, and `Get-Help <command>` to get extensive help about that command).

TL;DR: I thought that using PhpManager would have made maintainers' life easier, but it seems it's not the case: so please feel absolutely free to close this PR :wink:

morozov:
Sorry, I'm going to close it.

PR #3157, opened 2018-05-21
When building a limit query, zero offset without a limit should be ignored

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

Semantically, the absence of the offset in a `LIMIT`ed query and the zero offset mean the same: the rows should be fetched starting the first one. At the same time, some platforms handle the zero offset in the same way as the non-zero ones by adding unnecessary keywords and/or sub-queries.

It could be avoided by proper handling of the zero value of the offset.


Ocramius (approved):
👌 Thanks @morozov!

PR #3158, opened 2018-05-21
postgres: correctly produce alter table sql comment to update column definition

bendavies:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues |

#### Summary

The schema diff for postgres wasn't correctly updating the column comment for an altered column, for example updating a column type from `datetime` to `datetime_immutable`

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -580,11 +580,8 @@ SQL
                 }
             }
 
-            $oldComment = null;
             $newComment = $this->getColumnComment($column);
-            if (null !== $columnDiff->fromColumn) {
-                $oldComment = $this->getColumnComment($columnDiff->fromColumn);
-            }
+            $oldComment = $this->getOldColumnComment($columnDiff);
 
             if ($columnDiff->hasChanged('comment') || ($columnDiff->fromColumn !== null && $oldComment !== $newComment)) {
                 $commentsSQL[] = $this->getCommentOnColumnSQL(
@@ -1253,5 +1250,10 @@ SQL
     private function isNumericType(Type $type) : bool
     {
         return $type instanceof IntegerType || $type instanceof BigIntType;
+    }
+
+    private function getOldColumnComment(ColumnDiff $columnDiff) : ?string
+    {
+        return $columnDiff->fromColumn ? $this->getColumnComment($columnDiff->fromColumn) : null;
     }
 }
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -580,11 +580,17 @@ SQL
                 }
             }
 
-            if ($columnDiff->hasChanged('comment')) {
+            $oldComment = null;
+            $newComment = $this->getColumnComment($column);
+            if (null !== $columnDiff->fromColumn) {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -591,11 +591,17 @@ public function getAlterTableSQL(TableDiff $diff)
                 }
             }
 
-            if ($columnDiff->hasChanged('comment')) {
+            $oldComment = null;
```
Ocramius: Can the entire comment block be moved to a private method with correct nullable hints?

tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
```diff
@@ -819,6 +819,29 @@ public function testAltersTableColumnCommentWithExplicitlyQuotedIdentifiers()
...
+        $comparator = new Comparator();
+
+        $tableDiff = $comparator->diffTable($table1, $table2);
+
+        $this->assertInstanceOf('Doctrine\DBAL\Schema\TableDiff', $tableDiff);
+        $this->assertSame(
+            [
+                'ALTER TABLE "foo" ALTER "bar" TYPE TIMESTAMP(0) WITHOUT TIME ZONE',
+                'ALTER TABLE "foo" ALTER "bar" DROP DEFAULT',
+                'COMMENT ON COLUMN "foo"."bar" IS \'(DC2Type:datetime_immutable)\'',
```
Ocramius: Does the patch only affect types that require comments, or comments in general? If the latter is the case, can it be generalised to use explicit comments rather than going through DBAL types?

Ocramius (changes requested):

tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
bendavies: only when the type requires a comment via `requiresSQLCommentHint` as far as i can see. General comments (at least those added via `options={"comment":"..."}` to an orm column are correctly picked up.

bendavies: secondly, there is a further 'bug' here. the first two sql statements are not required as they are the same between the types being changed. I can fix this in a separate PR rather than pollute this one. they won't cause a problem.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
bendavies: do you mean extract the whole comment adding section to a method? i.e 594-606?

bendavies: ping @Ocramius

Ocramius: @bendavies a `private function` that retrieves `$oldComment`

bendavies: done

bendavies:
@Ocramius should be good, and rebased.

Ocramius (approved):
:+1:

Ocramius:
@morozov can you merge/backport this one, when you got a sec?

bendavies:
thanks!

PR #3178, opened 2018-06-06
Fix incorrect exception thrown from SQLAnywhere16Platform

Majkl578:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | maybe?

#### Summary

SQLAnywhere16Platform was throwing exception from doctrine/common's Proxy.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,9 @@
+# Upgrade to 2.8
+
+## Corrected exception thrown by ``Doctrine\DBAL\Platforms\SQLAnywhere16Platform::getAdvancedIndexOptionsSQL()``
+
+This method now throws SPL ``UnexpectedValueException`` instead of accidentally throwing ``Doctrine\Common\Proxy\Exception\UnexpectedValueException``.
+
 # Upgrade to 2.7
 
 ## Doctrine\DBAL\Platforms\AbstractPlatform::DATE_INTERVAL_UNIT_* constants deprecated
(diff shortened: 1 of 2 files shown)


UPGRADE.md
```diff
@@ -1,3 +1,9 @@
+# Upgrade to 2.8
+
+## Corrected exception thrown by ``Doctrine\DBAL\Platforms\SQLAnywhere16Platform::getAdvancedIndexOptionsSQL()``
+
+This method now throws SPL ``UnexpectedValueException`` instead of accidentally throwing ``Doctrine\Common\Proxy\Exception\UnexpectedValueException``.
```
morozov: I see that a couple of other drivers already throw SPL `UnexpectedValueException` however it's not reflected anywhere in their API documentation. Meaning that the caller doesn't know it can expect this kind of exception to be thrown.

I get the idea behind the change but maybe instead of just getting rid of an incorrect exception, we should introduce some sort of DBAL exception which the caller could catch together with other exceptions which a driver can throw?

Or just throw a generic `DBALException`?

Majkl578: Is it worth doing given that we're reworking exceptions in #3131 + even further?
We can leave it as is but we won't be able to get rid of doctrine/common in 2.8 then.

morozov: I'm all in favor of getting rid of doctrine/common, I'm just saying we need to replace it with a DBAL exception rather than SPL one.

Majkl578: There is currently other 35 occurences where platforms throw SPL exception directly.
I am still not sure whether it's worth changing it to DBALException solely for 2.8 (thus also breaking UnexpectedValueException inheritance assumption). I would rather leave this for 3.0 and fix it globally there. :/

morozov: If it’ 35 of them un total, I agree. And aso agree with the inheritance argument.

morozov (approved):

PR #3179, opened 2018-06-06
Removed the logic of using BLOB columns for BINARY type of fields

morozov:
In DBAL 2.x, if a `BINARY` column has a length greater than the max size on a given platform, it's represented as a `BLOB`. This idea may be inspired by the type systems of PostgreSQL and SQLite which don't make a difference between `BINARY` columns and `BLOB`s.

On the DBAL level, however, it looks like a bad idea for the following reasons:

1. `BLOB` columns cannot be used in `DISTINCT`, `WHERE`, `ORDER BY` and `GROUP BY` clauses on some platforms (e.g. Oracle, IBM DB2) for a reason. A `BLOB` is just a _large_ piece of binary data which is not supposed to be compared. It should be just exported or imported.
2. The max binary length is inconsistent between platforms: 2K, 4K, 8K, 32K, 64K. It means that depending on the length of a particular field, you'll get it represented as a `BLOB` on some platforms and as a `BINARY` on others.
3. To make `BLOB` and `BINARY` types look alike, DBAL represents `BINARY` columns in PHP as streams. It's pointless because unlike the `BLOB` type which could be used to store images or other media files, `BINARY` columns are used to store values like UUID or hashes like SHA1 and can be used as key columns. You don't want to have your key value as a stream.
4. DBAL doesn't silently convert `(VAR)?CHAR` columns to `TEXT`. The same shouldn't be done for binary columns.

Depending on the purpose of the column, the developer should decide whether it should be a `BINARY` column or a `BLOB` explicitly.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
@@ -516,19 +516,24 @@ abstract class AbstractMySQLPlatformTestCase extends AbstractPlatformTestCase
 
     public function testReturnsBinaryTypeDeclarationSQL()
     {
-        self::assertSame('VARBINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(array()));
-        self::assertSame('VARBINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 0)));
-        self::assertSame('VARBINARY(65535)', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 65535)));
-        self::assertSame('MEDIUMBLOB', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 65536)));
-        self::assertSame('MEDIUMBLOB', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 16777215)));
-        self::assertSame('LONGBLOB', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 16777216)));
+        self::assertSame('VARBINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL([]));
+        self::assertSame('VARBINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(['length' => 0]));
+        self::assertSame('VARBINARY(65535)', $this->_platform->getBinaryTypeDeclarationSQL(['length' => 65535]));
+        self::assertSame('VARBINARY(65536)', $this->_platform->getBinaryTypeDeclarationSQL(['length' => 65536]));
 
-        self::assertSame('BINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true)));
-        self::assertSame('BINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 0)));
-        self::assertSame('BINARY(65535)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 65535)));
-        self::assertSame('MEDIUMBLOB', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 65536)));
-        self::assertSame('MEDIUMBLOB', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 16777215)));
-        self::assertSame('LONGBLOB', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 16777216)));
+        self::assertSame('BINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(['fixed' => true]));
+        self::assertSame('BINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL([
+            'fixed' => true,
+            'length' => 0,
+        ]));
+        self::assertSame('BINARY(65535)', $this->_platform->getBinaryTypeDeclarationSQL([
+            'fixed' => true,
+            'length' => 65535,
+        ]));
+        self::assertSame('BINARY(65536)', $this->_platform->getBinaryTypeDeclarationSQL([
+            'fixed' => true,
+            'length' => 65536,
+        ]));
     }
 
     public function testDoesNotPropagateForeignKeyCreationForNonSupportingEngines()
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractSQLServerPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractSQLServerPlatformTestCase.php
@@ -978,15 +978,24 @@ abstract class AbstractSQLServerPlatformTestCase extends AbstractPlatformTestCas
 
     public function testReturnsBinaryTypeDeclarationSQL()
     {
-        self::assertSame('VARBINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(array()));
-        self::assertSame('VARBINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 0)));
-        self::assertSame('VARBINARY(8000)', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 8000)));
-        self::assertSame('VARBINARY(MAX)', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 8001)));
+        self::assertSame('VARBINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL([]));
+        self::assertSame('VARBINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(['length' => 0]));
+        self::assertSame('VARBINARY(8000)', $this->_platform->getBinaryTypeDeclarationSQL(['length' => 8000]));
+        self::assertSame('VARBINARY(8001)', $this->_platform->getBinaryTypeDeclarationSQL(['length' => 8001]));
 
-        self::assertSame('BINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true)));
-        self::assertSame('BINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 0)));
-        self::assertSame('BINARY(8000)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 8000)));
-        self::assertSame('VARBINARY(MAX)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 8001)));
+        self::assertSame('BINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(['fixed' => true]));
+        self::assertSame('BINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL([
+            'fixed' => true,
+            'length' => 0,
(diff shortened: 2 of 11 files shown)


tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
```diff
@@ -519,16 +519,10 @@ public function testReturnsBinaryTypeDeclarationSQL()
         self::assertSame('VARBINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(array()));
         self::assertSame('VARBINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 0)));
         self::assertSame('VARBINARY(65535)', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 65535)));
-        self::assertSame('MEDIUMBLOB', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 65536)));
```
Ocramius: I suggest keeping the assertion and checking that `VARBINARY(65536)` is generated.

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
```diff
@@ -519,16 +519,10 @@ public function testReturnsBinaryTypeDeclarationSQL()
...
-        self::assertSame('MEDIUMBLOB', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 16777215)));
-        self::assertSame('LONGBLOB', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 16777216)));
 
         self::assertSame('BINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true)));
         self::assertSame('BINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 0)));
         self::assertSame('BINARY(65535)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 65535)));
-        self::assertSame('MEDIUMBLOB', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 65536)));
```
Ocramius: Same as above: asserted value should change, assertion should be kept

tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
```diff
@@ -446,12 +446,10 @@ public function testReturnsBinaryTypeDeclarationSQL()
         self::assertSame('RAW(255)', $this->_platform->getBinaryTypeDeclarationSQL(array()));
         self::assertSame('RAW(2000)', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 0)));
         self::assertSame('RAW(2000)', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 2000)));
-        self::assertSame('BLOB', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 2001)));
```
Ocramius: Same as above: let's keep the assertion with the correct comparison instead

tests/Doctrine/Tests/DBAL/Platforms/AbstractSQLServerPlatformTestCase.php
```diff
@@ -981,12 +981,10 @@ public function testReturnsBinaryTypeDeclarationSQL()
         self::assertSame('VARBINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(array()));
         self::assertSame('VARBINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 0)));
         self::assertSame('VARBINARY(8000)', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 8000)));
-        self::assertSame('VARBINARY(MAX)', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 8001)));
```
Ocramius: Same as above: let's keep the assertion with the correct comparison instead

tests/Doctrine/Tests/DBAL/Platforms/AbstractSQLServerPlatformTestCase.php
```diff
@@ -981,12 +981,10 @@ public function testReturnsBinaryTypeDeclarationSQL()
...
 
         self::assertSame('BINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true)));
         self::assertSame('BINARY(255)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 0)));
         self::assertSame('BINARY(8000)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 8000)));
-        self::assertSame('VARBINARY(MAX)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 8001)));
```
Ocramius: Same as above: let's keep the assertion with the correct comparison instead

tests/Doctrine/Tests/DBAL/Platforms/DB2PlatformTest.php
```diff
@@ -422,11 +422,9 @@ public function testReturnsBinaryTypeDeclarationSQL()
         self::assertSame('VARCHAR(1) FOR BIT DATA', $this->_platform->getBinaryTypeDeclarationSQL([]));
         self::assertSame('VARCHAR(255) FOR BIT DATA', $this->_platform->getBinaryTypeDeclarationSQL(['length' => 0]));
         self::assertSame('VARCHAR(32704) FOR BIT DATA', $this->_platform->getBinaryTypeDeclarationSQL(['length' => 32704]));
-        self::assertSame('BLOB(1M)', $this->_platform->getBinaryTypeDeclarationSQL(['length' => 32705]));
```
Ocramius: Same as above: let's keep the assertion with the correct comparison instead

tests/Doctrine/Tests/DBAL/Platforms/DB2PlatformTest.php
```diff
@@ -422,11 +422,9 @@ public function testReturnsBinaryTypeDeclarationSQL()
...
 
         self::assertSame('CHAR(1) FOR BIT DATA', $this->_platform->getBinaryTypeDeclarationSQL(['fixed' => true]));
         self::assertSame('CHAR(254) FOR BIT DATA', $this->_platform->getBinaryTypeDeclarationSQL(['fixed' => true, 'length' => 0]));
-        self::assertSame('BLOB(1M)', $this->_platform->getBinaryTypeDeclarationSQL(['fixed' => true, 'length' => 32705]));
```
Ocramius: Same as above: let's keep the assertion with the correct comparison instead

tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
```diff
@@ -446,12 +446,10 @@ public function testReturnsBinaryTypeDeclarationSQL()
...
 
         self::assertSame('RAW(255)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true)));
         self::assertSame('RAW(2000)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 0)));
         self::assertSame('RAW(2000)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 2000)));
-        self::assertSame('BLOB', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 2001)));
```
Ocramius: Same as above: let's keep the assertion with the correct comparison instead

tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php
```diff
@@ -938,12 +938,10 @@ public function testReturnsBinaryTypeDeclarationSQL()
         self::assertSame('VARBINARY(1)', $this->_platform->getBinaryTypeDeclarationSQL(array()));
         self::assertSame('VARBINARY(1)', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 0)));
         self::assertSame('VARBINARY(32767)', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 32767)));
-        self::assertSame('LONG BINARY', $this->_platform->getBinaryTypeDeclarationSQL(array('length' => 32768)));
```
Ocramius: Same as above: let's keep the assertion with the correct comparison instead

tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php
```diff
@@ -938,12 +938,10 @@ public function testReturnsBinaryTypeDeclarationSQL()
...
 
         self::assertSame('BINARY(1)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true)));
         self::assertSame('BINARY(1)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 0)));
         self::assertSame('BINARY(32767)', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 32767)));
-        self::assertSame('LONG BINARY', $this->_platform->getBinaryTypeDeclarationSQL(array('fixed' => true, 'length' => 32768)));
```
Ocramius: Same as above: let's keep the assertion with the correct comparison instead

tests/Doctrine/Tests/DBAL/Types/BinaryTest.php
```diff
@@ -52,19 +52,34 @@ public function testBinaryNullConvertsToPHPValue()
...
-        self::assertSame($databaseValue, $phpValue);
+        self::assertSame('binary string', $phpValue);
+    }
+
+    /**
+     * Creates a binary string containing all possible byte values.
+     *
+     * @return string
+     */
+    private function getBinaryString()
```
Ocramius: ` : string`

tests/Doctrine/Tests/DBAL/Types/BinaryTest.php
```diff
@@ -52,19 +52,34 @@ public function testBinaryNullConvertsToPHPValue()
...
         $phpValue      = $this->type->convertToPHPValue($databaseValue, $this->platform);
 
-        self::assertSame($databaseValue, $phpValue);
+        self::assertSame('binary string', $phpValue);
+    }
+
+    /**
+     * Creates a binary string containing all possible byte values.
+     *
+     * @return string
```
Ocramius: Omit

tests/Doctrine/Tests/DBAL/Types/BinaryTest.php
```diff
@@ -52,19 +52,34 @@ public function testBinaryNullConvertsToPHPValue()
...
+     */
+    private function getBinaryString()
+    {
+        $string = '';
+
+        for ($i = 0; $i < 256; $i++) {
```
Ocramius: `return implode(array_map('chr', range(1, 256)))`? Or am I pushing it too far? :D

Ocramius (changes requested):

Ocramius:
Note: I agree with the overall patch idea - the comments are just about details

morozov:
@Ocramius mind taking another look?

Ocramius (approved):
LGTM 👍

Ocramius:
Thanks @morozov!

PR #3181, opened 2018-06-07
Remove dependency on doctrine/common

### LINKED ISSUE #3176: Eliminate dependency on doctrine/common
There's currently hard dependency on `doctrine/common` which pulls unnecessary packages like Lexer or Annotations.
What is actually used:
`Doctrine\Common\Cache` -> to be replaced by `doctrine/cache`
`Doctrine\Common\EventManager` -> needs to be extracted into `doctrine/event-manager`
`Doctrine\Common\Util\Debug` -> move into `doctrine/dbal` / extract to `doctrine/debug` / replace by something else?
`Doctrine\Common\Proxy\Exception\UnexpectedValueException` thrown from `SQLAnywhere16Platform` - this looks like a bug (wrong import)

Blocker for https://github.com/doctrine/migrations/issues/722

Majkl578:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues |  #3176

#### Summary

Replace doctrine/common with dependency on doctrine/cache + doctrine/event-manager.

This also drops indirect dependency on:
* doctrine/lexer
* doctrine/collections
* doctrine/annotations
* doctrine/inflector

closes  #3176

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
     "require": {
         "php": "^7.1",
         "ext-pdo": "*",
-        "doctrine/common": "^2.7.1"
+        "doctrine/cache": "^1.0",
+        "doctrine/event-manager": "^1.0"
     },
     "require-dev": {
         "doctrine/coding-standard": "^4.0",
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,12 @@
 # Upgrade to 2.8
 
+## Removed dependency on doctrine/common
+
+The dependency on doctrine/common package has been removed.
+DBAL now depends on doctrine/cache and doctrine/event-manager instead.
+If you are using any other component from doctrine/common package,
+you will have to add an explicit dependency to your composer.json.
+
 ## Corrected exception thrown by ``Doctrine\DBAL\Platforms\SQLAnywhere16Platform::getAdvancedIndexOptionsSQL()``
 
 This method now throws SPL ``UnexpectedValueException`` instead of accidentally throwing ``Doctrine\Common\Proxy\Exception\UnexpectedValueException``.


composer.json
```diff
@@ -14,7 +14,8 @@
     "require": {
         "php": "^7.1",
         "ext-pdo": "*",
-        "doctrine/common": "^2.7.1"
+        "doctrine/cache": "^1.0",
+        "doctrine/event-manager": "^1.0@dev"
```
morozov: Should we release this first?

Majkl578: Depends, leaving the decision up to you. 😛 We should definitely have 1.0.0 before DBAL 2.8 either way.

Majkl578: Bumped to stable. :+1:

jwage:
If this is not considered a BC break, I think we should release this in a Doctrine DBAL 1.7.x release. @Ocramius thoughts?

morozov:
I have a gut feeling it should be `2.8.0` rather than `2.7.x`. If it was a BC break, it would have to be `3.0.0`. If it was just a patch or a bug fix, it would be `2.7.x`. But it's more of an improvement and code reorganization.

morozov (approved):

morozov:
Waiting for the consensus before merging (the upgrade file specifically say the change will have happened in `2.8.0`).

Ocramius:
> If this is not considered a BC break, I think we should release this in a Doctrine DBAL 1.7.x release. 
 > @Ocramius thoughts?

Patch releases MUST leave dependencies un-touched, because every older system MUST be able to run a `composer update` at any time in order to receive eventual security patches, so the constraints are very important there.

We can make this land in `2.8`, but not `2.7`.

Ocramius (approved):

Ocramius:
:ship:

Issue #2566, opened 2016-11-30
Column altering in migration from TEXT to LONGTEXT may not apply.


janokary:
Based on this
https://github.com/laravel/framework/issues/12363
and my experience here
https://laracasts.com/discuss/channels/laravel/alter-table-change-field-to-longtext

I think there is a problem when migrating a table column from TEXT to LONG TEXT

Ocramius:
@janokary what RDBMs does this apply to? Maybe this is missing schema introspection?

janokary:
It is allways mysql

> version: 5.5.52-0+deb8u1

or

> version: 5.7.11-log

or
>version: 5.7.12-0ubuntu1.1

deeky666:
@janokary I need more information. What is the Doctrine DBAL type before and what should it be after? Also `length` property is is significant.

ariews:
Hi all, I'm facing same issue, using Laravel 5.4, doctrine/dbal v2.5.12, php 5.6.30, mysql  Ver 14.14 Distrib 5.6.33.

before: `$table->text('body')` alter using: `$table->longText('body')->change()`. 

The column type should be `longtext`, but the column type is unchanged.

Ocramius:
@ariews can this be reproduced with DBAL-only?

frandieguez:
Any news on getting this integrated?

Ocramius:
@frandiguez this needs a test case first.

Ocramius:
Can't find a PR for that: can you open or dig up one?

cbotsikas:
@Ocramius here's your PR.

cbotsikas:
I came across this issue trying to change a column in a Laravel application from text to longtext. I traced the migration and the issue was that the change was ignored since [$changedProperties was empty](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/Comparator.php#L235) although the _length was different.
The creation migration in Laravel was like this:
```
Schema::create('myTable', function (Blueprint $table) {
    $table->increments('id');
    $table->text('myColumn')->nullable();
});
```
and the change migration like this:
```
Schema::table('myTable', function (Blueprint $table) {
    $table->longText('myColumn')->change();
});
```

cbotsikas:
I have added some comments [on my PR](https://github.com/doctrine/dbal/pull/3182). There's an issue with a failing test which i feel is not well defined. 
Please have a look and let me know if i'm missing something..

uphlewis:
[...]

PR #3182, opened 2018-06-09
Fix text column length change

cbotsikas:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2566 

#### Summary

Schema Comparator's diffColumn did not check if the length had changed when the column type was Text. In MySQL it wasn't possible to change between TINYTEXT, TEXT, MEDIUMTEXT and LONGTEXT.
I've added a new check for TextType where i only check if the length property is the same in both columns. I did not append the case with StringType and BinaryType because 'fixed' property doesn't apply to TextType and the default 'length' of 255 feels wrong. 

I've included a test case.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Comparator.php
+++ b/lib/Doctrine/DBAL/Schema/Comparator.php
@@ -469,7 +469,7 @@ class Comparator
             }
         } elseif ($properties1['type'] instanceof Types\TextType) {
             // check if values of length match. There is no default value for all adapters.
-            if ($properties1['length'] !== $properties2['length']) {
+            if ($properties1['length'] != $properties2['length']) {
                 $changedProperties[] = 'length';
             }
         } elseif ($properties1['type'] instanceof Types\DecimalType) {
--- a/lib/Doctrine/DBAL/Schema/Comparator.php
+++ b/lib/Doctrine/DBAL/Schema/Comparator.php
@@ -467,6 +467,11 @@ class Comparator
             if ($properties1['fixed'] != $properties2['fixed']) {
                 $changedProperties[] = 'fixed';
             }
+        } elseif ($properties1['type'] instanceof Types\TextType) {
+            // check if values of length match. There is no default value for all adapters.
+            if ($properties1['length'] !== $properties2['length']) {
+                $changedProperties[] = 'length';
+            }
         } elseif ($properties1['type'] instanceof Types\DecimalType) {
             if (($properties1['precision'] ?: 10) != ($properties2['precision'] ?: 10)) {
                 $changedProperties[] = 'precision';
--- a/tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
+++ b/tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
@@ -1052,6 +1052,28 @@ class ComparatorTest extends \PHPUnit\Framework\TestCase
         self::assertEquals($expected, Comparator::compareSchemas($oldSchema, $newSchema));
     }
 
+    public function testCompareChangedTextColumn()
+    {
+        $oldSchema = new Schema();
+
+        $tableFoo = $oldSchema->createTable('foo');
+        $tableFoo->addColumn('id', 'text');
+
+        $newSchema = new Schema();
+        $table = $newSchema->createTable('foo');
+        $table->addColumn('id', 'text', array('length' => 16777215 + 1));
+
+        $expected = new SchemaDiff();
+        $expected->fromSchema = $oldSchema;
+        $tableDiff = $expected->changedTables['foo'] = new TableDiff('foo');
+        $tableDiff->fromTable = $tableFoo;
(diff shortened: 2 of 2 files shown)


cbotsikas:
I need to check this later today and also use the non strict comparison (!=) instead. I suppose this is why Travis failed and I guess that's why it's not strict for the StringType case.

On Jun 9, 2018 1:20 PM, "Jonathan H. Wage" <notifications@github.com> wrote:

After this change, the correct DDL statement is generated for you?

cbotsikas:
@jwage I've successfully tested it using Laravel with MySQL and all migrations i tried switching between  text, mediumtext and longtext where executed correctly. The DDL was correct. Laravel doesn't support tinytext, so i didn't do that.

Regarding the travis [failing test](https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php#L393-L407), I think it's out of my scope and not sure how to proceed. Either the failing test is wrong or there's a deeper issue with the field's length default handling in dbal. Let me explain the issue:
The test is [creating a text field](https://github.com/doctrine/dbal/blob/7021bdf74c7a678b6eaa0bd1303cfc77201dd2cf/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php#L238) without specifying the length. This means that the object (and the [$offlineTable](https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php#L399)) has [null length for the column](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/Column.php#L165). This column is created as longtext in MySQL for some reason. When the test reads the table in $onlineTable, [the length is 0](https://travis-ci.org/doctrine/dbal/jobs/390012492#L604) since there is no case for longtext [as there exists for the rest](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php#L159-L167) (tinytext, text and mediumtext). My [latest commit with non-strict comparison](https://github.com/doctrine/dbal/pull/3182/commits/a184e1c15a4435c7d14c8fa67082fdc7a3a7d145) is checking 0 over null, which is making the test for MySQL to pass. For SqlServer, i'm not sure what's the actual type but the [length is -1](https://travis-ci.org/doctrine/dbal/jobs/390210423#L1038), which is the MAX length, if i'm not mistaken. Comparing -1 over null is not the same and shouldn't. Without my fix, all tests are passing but clearly it shouldn't since the length values are not equal.

What do you think about this issue and the failing test? 

In my opinion, it should use the strict comparison (but could also stay with the non-strict one) and fix the failing test with explicit length or by checking manually if the diff is the text column's length and ignore it.

I don't have any more capacity atm to dive deeper in dbal.

cbotsikas:
With MySQL, changing the test to explicitly stating the text length has the following results:

- no length, **Fail** (existing test):
$table->addColumn('foo', 'text', array('notnull' => true));
$offlineTable's column length: null
$onlineTable's column length: 0
- tinytext. Pass:
$table->addColumn('foo', 'text', array('length' => 255, 'notnull' => true));
$offlineTable's column length: 255
$onlineTable's column length: 255
- text, Pass:
$table->addColumn('foo', 'text', array('length' => 65535, 'notnull' => true));
$offlineTable's column length: 65535
$onlineTable's column length: 65535
- mediumtext, Pass:
$table->addColumn('foo', 'text', array('length' => 16777215, 'notnull' => true));
$offlineTable's column length: 16777215
$onlineTable's column length: 16777215
- longtext, **Fail**:
$table->addColumn('foo', 'text', array('length' => 4294967295, 'notnull' => true));
$offlineTable's column length: 4294967295
$onlineTable's column length: 0
- random, Fail (expected):
$table->addColumn('foo', 'text', array('length' => 50, 'notnull' => true));
$offlineTable's column length: 50
$onlineTable's column length: 255
- -1, Fail (expected):
$table->addColumn('foo', 'text', array('length' => -1, 'notnull' => true));
$offlineTable's column length: -1
$onlineTable's column length: 255

The longtext shouldn't be failing but, as i said, the case is missing when the table is read.
I feel that the SqlServer would fail in any case except maybe the last, but i didn't check how is the SqlServer implementation done and what would it mean to set the length for text.

PVince81:
Any chance to get the tests fixed to have this merged ?

Is there another workaround for the time being ?

aschempp:
Any chance of this getting fixed?

@cbotsikas a suggestion to fix the failing test. Try this:

```php
if (($properties1['length'] > 0 || $properties2['length'] > 0) && (int) $properties1['length'] !== (int) $properties2['length']) {
    $changedProperties[] = 'length';
}
```

SenseException:
It seems the changes break sqlsrv and ibm_db2 tests. Any chances that you can make your changes work with them too?

morozov:

PR #3183, opened 2018-06-14
CI: Test against MySQL 8.0 on Travis

Majkl578:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Testing MySQL 8.0 on Travis.

### DIFF
--- a/tests/travis/install-mysql-8.0.sh
+++ b/tests/travis/install-mysql-8.0.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -ex
+
+echo "Starting MySQL 8.0..."
+
+echo -e "[mysqld]\ndefault_authentication_plugin=mysql_native_password" > /tmp/mysql-auth.cnf
+
+sudo docker pull mysql:8.0
+sudo docker run \
+    -d \
+    -e MYSQL_ALLOW_EMPTY_PASSWORD=yes \
+    -e MYSQL_DATABASE=doctrine_tests \
+    -v /tmp/mysql-auth.cnf:/etc/mysql/conf.d/auth.cnf:ro \
+    -p 33306:3306 \
+    --name mysql80 \
+    mysql:8.0
+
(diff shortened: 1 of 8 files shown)


morozov:
There's one more failure I inconsistently see on MySQL 8:
```
1) Doctrine\Tests\DBAL\Functional\Schema\MySqlSchemaManagerTest::testAlterTableScenario
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => 'foo'
-    1 => 'foreign_key_test'
+    0 => 'foreign_key_test'
+    1 => 'foo'
 )
```
There must be the same reason as for the previously fixed failure.

tests/travis/install-mysql-8.0.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -ex
+
+echo "Starting MySQL 8.0..."
+
+echo -e "[mysqld]\ndefault_authentication_plugin=mysql_native_password" > /tmp/mysql-auth.cnf
```
morozov: As far as I understand, this file cannot be directly created in the container because by the moment the container has started, the configuration will have been read, and the newly created file will be ignored. Right?

Majkl578: Yes, this is done according to [documentation](https://hub.docker.com/_/mysql/) and is a standard way to provide custom configs to dockerized services:

> Using a custom MySQL configuration file
> ...
> This will start a new container some-mysql where the MySQL instance uses the combined startup settings from /etc/mysql/my.cnf and /etc/mysql/conf.d/config-file.cnf, with settings from the latter taking precedence.

Majkl578: Maybe we can try altering the user instead if you don't like linking a config: https://dev.mysql.com/doc/refman/8.0/en/upgrading-from-previous-series.html#upgrade-caching-sha2-password

morozov: No, it's fine leaving it as is. I didn't realize it's the right way to do things and started asking, then realized but decided to post the question to make sure.

PR #3188, opened 2018-06-19
Deprecated usage of binary fields whose length exceeds the platform maximum

### LINKED ISSUE #3187: Deprecate usage of binary fields whose length exceeds maximum
As of #3179 (`3.0.0`), `BLOB`s are no longer used to represent binary fields whose length exceeds maximum on a given platform. This logic should be documented as deprecated in the next minor release. If technically possible, deprecation notices should be triggered.

morozov:
Closes #3187.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -42,6 +42,7 @@ use Doctrine\DBAL\Schema\TableDiff;
 use Doctrine\DBAL\TransactionIsolationLevel;
 use Doctrine\DBAL\Types;
 use Doctrine\DBAL\Types\Type;
+use const E_USER_DEPRECATED;
 use function addcslashes;
 use function array_map;
 use function array_merge;
@@ -68,6 +69,7 @@ use function strlen;
 use function strpos;
 use function strtolower;
 use function strtoupper;
+use function trigger_error;
 
 /**
  * Base class for all DatabasePlatforms. The DatabasePlatforms are the central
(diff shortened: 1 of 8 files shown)


lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -325,7 +327,17 @@ public function getBinaryTypeDeclarationSQL(array $field)
 
         $fixed = $field['fixed'] ?? false;
 
-        if ($field['length'] > $this->getBinaryMaxLength()) {
+        $maxLength = $this->getBinaryMaxLength();
+
+        if ($field['length'] > $maxLength) {
+            if ($maxLength > 0) {
+                @trigger_error(sprintf(
+                    'Binary field length %d is greater than supported by the platform (%d)',
```
Majkl578: Can we also add a hint here? _"use getBlobTypeDeclarationSQL() instead"_

Majkl578 (dismissed):

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: It may be confusing. Besides using a blob, a user may want to reduce the length of the field. Also, most likely, a user is not calling this method directly, it's called based on their entity definitions.

What about: reduce the field size or use a BLOB field instead?

Majkl578: Sounds good. 👍

Ocramius (dismissed):
LGTM 👍

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
Majkl578: Hah @Ocramius was too fast. 😂

morozov:
@Majkl578 @Ocramius did you guys just see a race condition here? I force-pushed the updated branch a moment before the pull request got merged, it invalidated the review but still let the merged. Both the merge and invalidation happened on the same second: 3:48 PDT.

Ocramius:
Yeah, saw the dismissed review.

Is the current `master` state diffing anything with your local dev branch?

Majkl578:
Yup, noticed, original commit got merged, the new one interestingly only fired ContinuousPHP build which immediately failed.
Nice!


@morozov Can you send the reworded patch in new PR? 👍

morozov:
> Is the current `master` state diffing anything with your local dev branch?

Actually yes (#3192). What I thought, is the changes I force-pushed got merged even if they weren't approved. So, false alarm.

PR #514, opened 2014-01-16
[DBAL-789] Fix default values for TEXT/BLOB column type on MySQL

deeky666:
MySQL does not support default values for `TEXT` and `BLOB` type columns. This causes `CREATE TABLE` statements to fail if a default value is supplied for those column types:

``` sql
Doctrine\DBAL\DBALException: An exception occurred while executing 'CREATE TABLE text_blob_default_value (def_text LONGTEXT DEFAULT 'def' NOT NULL, def_text_null LONGTEXT DEFAULT 'def', def_blob LONGBLOB DEFAULT 'def' NOT NULL, def_blob_null LONGBLOB DEFAULT 'def') DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB':

SQLSTATE[42000]: Syntax error or access violation: 1101 BLOB/TEXT column 'def_text' can't have a default value
```

Additionally useless `ALTER TABLE` statements are created for those mappings with the schema tool because online and offline schema differ for those column types concerning default values:

``` sql
ALTER TABLE creative CHANGE filename filename LONGTEXT DEFAULT '' NOT NULL;
```

PR #3189, opened 2018-06-19
In Schema Diff, ignore default value alteration for MySQL TEXT and BLOB types

JarJak:
MySQL (all versions) and MariaDB (<10.2.1) does not support default value for TEXT and BLOB types.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #514 

#### Summary

If you look at #514 Doctrine already ignores those, but if you run schema diff you can still see that default value for these columns is altered.

See: https://github.com/bolt/bolt/pull/7477

I am not sure if this is best place to fix it. Any better proposals are more than welcomed :)

### DIFF
--- a/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
@@ -168,7 +168,12 @@ class MySqlSchemaManager extends AbstractSchemaManager
         if ($this->_platform instanceof MariaDb1027Platform) {
             $columnDefault = $this->getMariaDb1027ColumnDefault($this->_platform, $tableColumn['default']);
         } else {
-            $columnDefault = $tableColumn['default'];
+            if ($dbType === 'text' || $dbType === 'blob') {
+                // MySQL (all versions) and MariaDB (<10.2.1) does not support default value for TEXT and BLOB types
+                $columnDefault = null;
+            } else {
+                $columnDefault = $tableColumn['default'];
+            }
         }
 
         $options = [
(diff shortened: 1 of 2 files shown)


morozov:
> MySQL (all versions) and MariaDB (<10.2.1) does not support default value for TEXT and BLOB types.

@JarJak, as far as I understand the patch, it disables default value even for for all MySQL-like platforms even if they support the default values. What about the non-MySQL-like platforms ones?

Regardless of what the final decision is, we need to have a functional test which would pass on all supported platforms and therefore explain what the intended behavior is.

lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
```diff
@@ -188,7 +188,12 @@ protected function _getPortableTableColumnDefinition($tableColumn)
         if ($this->_platform instanceof MariaDb1027Platform) {
             $columnDefault = $this->getMariaDb1027ColumnDefault($this->_platform, $tableColumn['default']);
         } else {
-            $columnDefault = $tableColumn['default'];
+            if ($dbType === 'text' || $dbType === 'blob') {
+                // MySQL (all versions) and MariaDB (<10.2.1) does not support default value for TEXT and BLOB types
```
morozov: This comment is a good candidate for a platform feature.

morozov (changes requested):

JarJak:
@morozov this patch does not disable default value because it is already disabled in #514. It only stops Schema Siff from showing this as a change in schema, which is false positive.

morozov:
> Regardless of what the final decision is, we need to have a functional test which would pass on all supported platforms and therefore explain what the intended behavior is.

Let's start with this.

morozov:

PR #3197, opened 2018-06-25
fix for pg10 support

lorenzo-pezzali:
In case the db in use is a pg10 the function "_getPortableSequenceDefinition" for retrieving the sequence information won't work.
I've updated the class so it can get the pg version in the constructor and then use in the sequence definition lookup changing the query executed.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | no issue yet

#### Summary

I've created a constructor that gets the version number from prostgresql and then it compares the version in the "_getPortableSequenceDefinition" method.
In case the version is equal or grater to 10 it uses the new view pg_sequences to get the sequence information.


lpwebit:
Hi, DBAL 2.7 requires php 7.1 if i can't update to that php version how can i handle this?
thanks in advance!

Ocramius:
Hi @lpwebit,

We highly endorse a PHP upgrade. Support for newer platforms in older DBAL releases will not be provided.

PR #3201, opened 2018-06-29
Fix support for URL to account for master-slave and pooling-shard connections

stof:
|      Q       |   A
|------------- | -----------
| Type         | bug/feature
| BC Break     | no
| Fixed issues | 

#### Summary

When adding support for url-based configuration, it was implemented at the DriverManager level rather than at the Connection level (which indeed has a bunch of advantages). Due to that, MasterSlaveConnection and PoolingShardConnection were missing support for the URL because the params are not at the top-level, but in sub-keys.
Moving the parsing of the URL to the connection itself when dealing with params would probably require much more work than making DriverManager aware of the config structure for these 2 special connections. 

I would consider it as a bugfix for the URL support (and so I hope it gets cherry-picked at least in 2.7), but that's your choice.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/DriverManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/DriverManagerTest.php
@@ -2,12 +2,15 @@
 
 namespace Doctrine\Tests\DBAL;
 
+use Doctrine\DBAL\Connections\MasterSlaveConnection;
 use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Driver\DrizzlePDOMySql\Driver as DrizzlePDOMySqlDriver;
 use Doctrine\DBAL\Driver\PDOMySql\Driver as PDOMySQLDriver;
 use Doctrine\DBAL\Driver\PDOSqlite\Driver as PDOSqliteDriver;
 use Doctrine\DBAL\Driver\SQLSrv\Driver as SQLSrvDriver;
 use Doctrine\DBAL\DriverManager;
+use Doctrine\DBAL\Sharding\PoolingShardConnection;
+use Doctrine\DBAL\Sharding\ShardChoser\MultiTenantShardChoser;
 use Doctrine\Tests\DBAL\Mocks\MockPlatform;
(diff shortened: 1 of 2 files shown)


stof:
I would even appreciate a backport to the 2.6 branch btw, as I'm currently unable to upgrade DBAL (due to being unable to upgrade the ORM because of a regression affecting me)

yguedidi (approved):

deguif:
Any chance to get this merged?

tests/Doctrine/Tests/DBAL/DriverManagerTest.php
```diff
@@ -133,6 +136,74 @@ public function testValidDriverClass()
...
+        self::assertInstanceOf(PDOMySQLDriver::class, $conn->getDriver());
+
+        $expected = [
+            'user'     => 'foo',
+            'password' => 'bar',
+            'host'     => 'localhost',
+            'port'     => 11211,
+        ];
+
+        foreach ($expected as $key => $value) {
```
morozov: Would it make sense to use `self::assertArraySubset()` instead?

morozov (approved):

morozov:
Thank you @stof!

stof:
@morozov will this get backported or no ?

morozov:
@stof, no. It doesn't look that severe to make an exception and revive an unsupported branch. You should be able to pass structured connection parameters instead of a URL and use master/slave connections and sharding on older DBAL versions.

deguif:
Would it be possible to get a release published with this fix?

Majkl578:
@deguif Right now this is targeting 2.9.0, whether this can be considered a bugfix and backported to 2.8.x is up to @morozov.

deguif:
Any ETA about the release of 2.9.0?

Ocramius:
None yet

morozov:
@stof right now, `master` is in an unstable state due to https://github.com/doctrine/dbal/issues/3336. We'll need some more time to double check if there are other regressions caused by the conversion of weak comparisons to strict.

PR #3207, opened 2018-07-05
Drop legacy submodules, empty README and ORM leftovers

Majkl578:
8 years :wave:


morozov (approved):
I don't know what it is but I think we can leave without it :+1:

Majkl578:
Some relics, probably from the time when ORM and DBAL were not separate repos.
The git module is IIRC useless with new website.

PR #3208, opened 2018-07-06
Add throws phpdoc

asuran:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Add throws tag to a couple of methods


lib/Doctrine/DBAL/Types/Type.php
```diff
@@ -119,6 +119,8 @@
      * @param \Doctrine\DBAL\Platforms\AbstractPlatform $platform The currently used database platform.
      *
      * @return mixed The database representation of the value.
+     *
+     * @throws ConversionException
```
morozov: @carusogabriel what's our take on fixing issues like this? Looks like it's far from being the only place where the thrown exception is not documented. I'd like these improvements to be done in a more systems approach rather than random fixes here and there.

carusogabriel: @morozov I don't know any solution besides https://github.com/kalessil/phpinspectionsea to ensure that. Maybe @Majkl578 knows? :thinking:

Majkl578: Take a look at https://github.com/pepakriz/phpstan-exception-rules.
I'm not saying we should blindly start using it (it's an experimental extension ATM), but we/you can use it to analyse the code locally.

scaytrase: @morozov so would the system approach be done by core maintainers or you suggest to replace this PR with CI improvements for throws signature detection and fix them all at once?

morozov: @scaytrase my suggestion is to find a way to identify all or most missing annotations and fix them at once. It doesn't have to be done by the maintainers. It doesn't have to be enforced on CI but you can give it a try.

Please keep in mind that some of the "missing" `@throws` annotations are by design. For instance, the exception thrown in:

https://github.com/doctrine/dbal/blob/d475ade698bf00c84d20946350f24f0cbd07f164/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php#L132

…is an API violation and shouldn't be documented since it'd be a BC break.

scaytrase: if it is API violation - should it be wrapped with some kind of API-expected exceptions or this would violate API too, since end-users expect broken API for now?

I think we can try to engage mentioned phpstan extension here

morozov: For this particular method, the API says:

https://github.com/doctrine/dbal/blob/d475ade698bf00c84d20946350f24f0cbd07f164/lib/Doctrine/DBAL/Driver/Statement.php#L51

Therefore, throwing any exception is agains the API. In `2.x`, we can ignore such cases but in `3.0` (e.g. https://github.com/doctrine/dbal/issues/3005) we should change the API by documenting exceptions and getting rid of ambiguous return values.

Majkl578: Maybe not worth the work until we finish #3131 + follow-up PR.

morozov: I think research-wise it makes sense to do it even in `2.x`. We're not going to introduce any new types, it's only about adding annotations.

Given we know the mapping between the `2.x` exceptions and the `3.x` ones, it shouldn't be hard to translate them afterwards (if this gets finished prior to #3131).

Issue #3167, opened 2018-05-27
The semantics of the GUID expression is inconsistent across platforms


morozov:
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.7.1

#### Summary

As a developer, I expect the GUID expression generated by `AbstractPlatform::getGuidExpression()` to have the same semantic (namely, [UUIDv1](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_1_(date-time_and_MAC_address))) across supported platforms.

#### Current behavior

1. On MySQL platform, it's implemented as `UUID()` which generates UUIDv1 and is therefore correct.
2. On PostgreSQL platform, it's implemented it as `UUID_GENERATE_V4()` which generates UUIDv4. For some reason, `UUID_GENERATE_V1()` is [unavailable](https://travis-ci.org/doctrine/dbal/jobs/384203956#L663) on Travis CI:
   > SQLSTATE[42883]: Undefined function: 7 ERROR:  function uuid_generate_v1() does not exist
3. On Oracle platform, it's implemented as `SYS_GUID()` which returns `RAW(16)` which is not a UUID [according to Stackoverflow](https://stackoverflow.com/a/3037511/146187).
4. On SQL Server and SQLAnywhere, it's implemented as `NEWID()`. From the [documentation](https://docs.microsoft.com/en-us/sql/t-sql/functions/newid-transact-sql?view=sql-server-2017), it's not clear which UUID version it generates, but according to the [test output](https://travis-ci.org/doctrine/dbal/jobs/384203968#L1035), it's not UUIDv1:
   > Failed asserting that 'DBB4D428-2338-4D4F-823F-4AA2ECEC985A' is greater than 'F4959EE5-02EC-4136-A21F-211813F67B63'
5. On SQLite, it's implemented using purely `RANDOMBLOB()` which makes it not compliant with the the UUIDv1 definition.

There's also a [summary](https://en.wikipedia.org/wiki/Universally_unique_identifier#As_database_keys) of the above on Wikipedia.

#### How to reproduce

Below is the pseudo-code. See a real example in [#3162](https://github.com/doctrine/dbal/pull/3162/files#diff-fe3cbd6e714521d88bd0e81b269c7863R26).

```php
$previous = generateGuid();

for ($i = 0; $i < 99; $i++) {
    $value = generateGuid();
    self::assertGreaterThan($previous, $value);
    $previous = $value;
}
```

#### Expected behavior

[...]

PR #3212, opened 2018-07-11
Deprecated usage of DB-generated UUIDs

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #3167

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,13 @@
 # Upgrade to 2.8
 
+## Deprecated usage of DB-generated UUIDs
+
+The format of DB-generated UUIDs is inconsistent across supported platforms and therefore is not portable. Some of the platforms produce UUIDv1, some produce UUIDv4, some produce the values which are not even UUID.
+
+Unless UUIDs are used in stored procedures which DBAL doesn't support, there's no real benefit of DB-generated UUIDs comparing to the application-generated ones.
+
+Use a PHP library (e.g. [ramsey/uuid](https://packagist.org/packages/ramsey/uuid)) to generate UUIDs on the application side.
+
 ## Deprecated usage of binary fields whose length exceeds the platform maximum
 
 - The usage of binary fields whose length exceeds the maximum field size on a given platform is deprecated.
--- a/docs/en/reference/sharding.rst
+++ b/docs/en/reference/sharding.rst
@@ -100,11 +100,15 @@ platforms:
 
     <?php
     use Doctrine\DBAL\DriverManager;
+    use Ramsey\Uuid\Uuid;
 
     $conn = DriverManager::getConnection(/**..**/);
-    $guid = $conn->fetchColumn('SELECT ' . $conn->getDatabasePlatform()->getGuidExpression());
+    $guid = Uuid::uuid1();
 
-    $conn->insert("my_table", array("id" => $guid, "foo" => "bar"));
+    $conn->insert('my_table', [
+        'id'  => $guid->toString(),
+        'foo' => 'bar',
+    ]);
 
 In your application you should hide this details in Id-Generation services:
 
@@ -113,15 +117,13 @@ In your application you should hide this details in Id-Generation services:
     <?php
     namespace MyApplication;
 
+    use Ramsey\Uuid\Uuid;
+
     class IdGenerationService
     {
-        private $conn;
-
-        public function generateCustomerId()
+        public function generateCustomerId() : Uuid
         {
-            return $this->conn->fetchColumn('SELECT ' .
-                $this->conn->getDatabasePlatform()->getGuidExpression()
-            );
+            return Uuid::uuid1();
         }
     }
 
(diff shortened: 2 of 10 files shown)


Majkl578:
I like it. But it has an impact on ORM which has a UuidGenerator, therefore not sure about 2.8.0 milestone. :/

morozov:
> I like it. But it has an impact on ORM which has a UuidGenerator, therefore not sure about 2.8.0 milestone. :/

@Majkl578, we're going to do a minor ORM release prior to releasing DBAL 2.8.0, right? As far as I understand, the changes needed to be made to the UuidGenerator are just a replacement of the DBAL API call with `Ramsey\Uuid`. Is there a problem with the schedule or with the introduction of a new last minute dependency or nobody just has time for it?

Majkl578:
> the changes needed to be made to the UuidGenerator are just a replacement of the DBAL API call with `Ramsey\Uuid`

Unfortunately not _just_ that. To quote your deprecation note:
> Some of the platforms produce UUIDv1, some produce UUIDv4, some produce the values which are not even UUID.

It's not consistent and it's platform dependent. So it can't simply be universally replaced by a Uuid1/Uuid4 version.

> Is there a problem with the schedule or with the introduction of a new last minute dependency

This is something that can't happen in a patch release, see https://github.com/doctrine/dbal/pull/3181#issuecomment-396370664.
We can only provide deprecation in ORM 2.7 and removal in 3.0.

morozov:
> We can only provide deprecation in ORM 2.7 and removal in 3.0.

You mean ORM 2.8 is not expected? We still can introduce new dependencies in minor releases. Besides that, is it a huge problem if ORM keeps using a deprecated feature of the DBAL? Deprecation is just a message, it's not a breaking change.

Majkl578:
> You mean ORM 2.8 is not expected?

Exactly. ORM 2.7 will be the last, deprecation-only release, but it's not scheduled to be released anytime soon (not within weeks).

> Besides that, is it a huge problem if ORM keeps using a deprecated feature of the DBAL?

Personally, I wouldn't bother, but I am pretty sure someone will rush in saying _it's unacceptable_.

morozov:
@Majkl578 I removed the `trigger_error()` pieces and only left documentation changes.

morozov:
> Personally, I wouldn't bother, but I am pretty sure someone will rush in saying _it's unacceptable_.

I acknowledge that but this _is unhealthy_. You cannot demand the library to not tell you that you're using a deprecated API. People take those messages and errors however they are just messages.

Majkl578:
> You cannot demand the library to not tell you that you're using a deprecated API.

I agree. But in case of ORM here, the consumers are using UuidGenerator, the fact it uses DBAL platform to generate the SQL is just an implementation detail. So they would start getting deprecations from a code they don't directly use, and at the same time ORM can't provide a fixed version without using deprecated API.

Marking it as `@deprecated` now and leaving `trigger_error()` for later 2.x release is imho more consumer-friendly at this stage. This is similar to `doctrine/common`, we [deprecated most of the classes this way](https://github.com/doctrine/common/commit/6e2d6228b05a0664ee2a027ffe17b0c18f3a2e0a), while ORM is still using them in 2.6.y (but won't be in 2.7/3.x).

Majkl578 (approved):

UPGRADE.md
```diff
@@ -1,5 +1,13 @@
 # Upgrade to 2.8
 
+## Deprecated usage of DB-generated UUIDs
+
+The format of DB-generated UUIDs is inconsistent across supported platforms and therefore is not portable. Some of the platforms produce UUIDv1, some produce UUIDv4, some produce the values which are not even UUID.
+
+Unless UUIDs are used in stored procedures which DBAL doesn't support, there's no real benefit of DB-generated UUIDs comparing to the application-generated ones.
+
+Use a PHP library (e.g. [ramsey/uuid](https://packagist.org/packages/ramsey/uuid)) to generate UUIDs on the application side.
```
lyrixx: I would suggest the [UUID PECL](https://pecl.php.net/package/uuid) instead.
It less invasive and the code is much more simple.

morozov: What do you mean it's less invasive?

lcobucci: I got intrigued with the "less invasive" too :smile: but DBAL would simply handle a string so it doesn't really matter which lib/extension is being used.

I'd still recommend @ramsey's lib, though. It's better documented (IMO) and supports all UUID versions (even the namespaced ones)...

Majkl578: I also don't understand the _less invasive_ part.
The referenced PECL extension requires setup by server admin, has no documentation on php.net, is unix-only and has last release in 2015.

lyrixx: It has not been updated because it does not need to be. It just works. 

About less invasive, i may be unclear. Sorry. I don't like the PHP lib because there is so many code for a really simple thing.

Ocramius: Applies also to core libs: 90%+ of the code is just to emulate userland behaviour. If you can avoid an extension for doing something that can be done in userland, then do avoid it.

PR #3217, opened 2018-07-13
Fix that MysqliStatement cannot handle streams

mpdude:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

The `binary` and `blob` types map `BLOB`/`BINARY`/`VARBINARY` columns to PHP streams. Internally, they use the `ParameterType::LARGE_OBJECT` (i. e. `\PDO::PARAM_LOB`) binding type, which suggests that efficient handling of PHP stream resources was intended.

However, at least when using the `mysqli` driver, stream resources passed into `insert()` or `update()` are simply cast to strings. As a result, a literal string like `"Resource id #126"` will end up in the database.

This PR fixes the issue by correctly processing streams in the `MysqliStatement` when they are passed with the `ParameterType::LARGE_OBJECT` binding type. It uses the `mysqli::send_long_data()` method to pass stream data in chunks to the MySQL server, thus keeping the memory footprint low. This method *does not* (despite claims to the contrary) allow to bypass the `max_allowed_package` size.

*Update:* The `pdo_mysql` driver was already capable of handling streams this way. Now this is covered by tests.

#### Helpful documentation

* http://php.net/manual/en/mysqli-stmt.send-long-data.php
* http://php.net/manual/en/mysqli-stmt.bind-param.php - see first "Note"
* http://php.net/manual/en/pdo.lobs.php
* https://blogs.oracle.com/oswald/phps-mysqli-extension:-storing-and-retrieving-blobs

### DIFF
--- a/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
+++ b/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
@@ -21,11 +21,16 @@ namespace Doctrine\DBAL\Driver\Mysqli;
 
 use Doctrine\DBAL\Driver\Statement;
 use Doctrine\DBAL\Driver\StatementIterator;
+use Doctrine\DBAL\Exception\InvalidArgumentException;
 use Doctrine\DBAL\FetchMode;
 use Doctrine\DBAL\ParameterType;
 use function array_combine;
 use function array_fill;
 use function count;
+use function feof;
+use function fread;
+use function get_resource_type;
+use function is_resource;
 use function str_repeat;
 
 /**
@@ -42,7 +47,7 @@ class MysqliStatement implements \IteratorAggregate, Statement
         ParameterType::BOOLEAN      => 'i',
         ParameterType::NULL         => 's',
         ParameterType::INTEGER      => 'i',
-        ParameterType::LARGE_OBJECT => 's',
+        ParameterType::LARGE_OBJECT => 'b',
     ];
 
     /**
@@ -169,9 +174,11 @@ class MysqliStatement implements \IteratorAggregate, Statement
                     throw new MysqliException($this->_stmt->error, $this->_stmt->errno);
                 }
             } else {
-                if (! $this->_stmt->bind_param($this->types, ...$this->_bindedValues)) {
+                list($types, $values, $streams) = $this->separateBoundValues();
+                if (! $this->_stmt->bind_param($types, ...$values)) {
                     throw new MysqliException($this->_stmt->error, $this->_stmt->sqlstate, $this->_stmt->errno);
                 }
+                $this->sendLongData($streams);
             }
         }
 
@@ -226,6 +233,63 @@ class MysqliStatement implements \IteratorAggregate, Statement
         $this->result = true;
 
         return true;
+    }
+
+    /**
+     * Split $this->_bindedValues into those values that need to be sent using mysqli::send_long_data()
+     * and those that can be bound the usual way.
+     *
+     * @return array<int, array<int|string, mixed>|string>
+     */
+    private function separateBoundValues()
+    {
+        $streams = $values = [];
+        $types   = $this->types;
+
+        foreach ($this->_bindedValues as $parameter => $value) {
+            if (! isset($types[$parameter - 1])) {
+                $types[$parameter - 1] = static::$_paramTypeMap[ParameterType::STRING];
+            }
+
+            if ($types[$parameter - 1] === static::$_paramTypeMap[ParameterType::LARGE_OBJECT]) {
+                if (is_resource($value)) {
+                    if (get_resource_type($value) !== 'stream') {
+                        throw new InvalidArgumentException('Resources passed with the LARGE_OBJECT parameter type must be stream resources.');
+                    }
+                    $streams[$parameter] = $value;
+                    $values[$parameter]  = null;
+                    continue;
+                } else {
+                    $types[$parameter - 1] = static::$_paramTypeMap[ParameterType::STRING];
+                }
+            }
+
+            $values[$parameter] = $value;
+        }
+
+        return [$types, $values, $streams];
+    }
+
+    /**
+     * Handle $this->_longData after regular query parameters have been bound
+     *
+     * @throws MysqliException
+     */
+    private function sendLongData($streams)
+    {
+        foreach ($streams as $paramNr => $stream) {
+            while (! feof($stream)) {
+                $chunk = fread($stream, 8192);
+
+                if ($chunk === false) {
+                    throw new MysqliException("Failed reading the stream resource for parameter offset ${paramNr}.");
+                }
+
+                if (! $this->_stmt->send_long_data($paramNr - 1, $chunk)) {
+                    throw new MysqliException($this->_stmt->error, $this->_stmt->sqlstate, $this->_stmt->errno);
+                }
+            }
+        }
     }
 
     /**
--- a/tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
@@ -3,11 +3,13 @@
 namespace Doctrine\Tests\DBAL\Functional;
 
 use Doctrine\DBAL\Driver\PDOSqlsrv\Driver as PDOSQLSrvDriver;
+use Doctrine\DBAL\FetchMode;
 use Doctrine\DBAL\ParameterType;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Types\Type;
-use const CASE_LOWER;
-use function array_change_key_case;
+use function fopen;
+use function in_array;
+use function str_repeat;
 use function stream_get_contents;
 
 /**
@@ -49,6 +51,28 @@ class BlobTest extends \Doctrine\Tests\DbalFunctionalTestCase
         self::assertEquals(1, $ret);
     }
 
+    public function testInsertProcessesStream()
+    {
+        if (in_array($this->_conn->getDatabasePlatform()->getName(), ['oracle', 'db2'], true)) {
+            // https://github.com/doctrine/dbal/issues/3288 for DB2
+            // https://github.com/doctrine/dbal/issues/3290 for Oracle
+            $this->markTestIncomplete('Platform does not support stream resources as parameters');
+        }
+
+        $longBlob = str_repeat('x', 4 * 8192); // send 4 chunks
+        $this->_conn->insert('blob_table', [
+            'id'        => 1,
+            'clobfield' => 'ignored',
+            'blobfield' => fopen('data://text/plain,' . $longBlob, 'r'),
+        ], [
+            ParameterType::INTEGER,
+            ParameterType::STRING,
+            ParameterType::LARGE_OBJECT,
+        ]);
+
+        $this->assertBlobContains($longBlob);
+    }
+
     public function testSelect()
     {
         $this->_conn->insert('blob_table', [
@@ -86,14 +110,63 @@ class BlobTest extends \Doctrine\Tests\DbalFunctionalTestCase
         $this->assertBlobContains('test2');
     }
 
+    public function testUpdateProcessesStream()
+    {
+        if (in_array($this->_conn->getDatabasePlatform()->getName(), ['oracle', 'db2'], true)) {
+            // https://github.com/doctrine/dbal/issues/3288 for DB2
+            // https://github.com/doctrine/dbal/issues/3290 for Oracle
+            $this->markTestIncomplete('Platform does not support stream resources as parameters');
+        }
+
+        $this->_conn->insert('blob_table', [
+            'id'          => 1,
+            'clobfield'   => 'ignored',
+            'blobfield'   => 'test',
+        ], [
+            ParameterType::INTEGER,
+            ParameterType::STRING,
+            ParameterType::LARGE_OBJECT,
+        ]);
+
+        $this->_conn->update('blob_table', [
+            'id'          => 1,
+            'blobfield'   => fopen('data://text/plain,test2', 'r'),
+        ], ['id' => 1], [
+            ParameterType::INTEGER,
+            ParameterType::LARGE_OBJECT,
+        ]);
+
+        $this->assertBlobContains('test2');
+    }
+
+    public function testBindParamProcessesStream()
+    {
+        if (in_array($this->_conn->getDatabasePlatform()->getName(), ['oracle', 'db2'], true)) {
+            // https://github.com/doctrine/dbal/issues/3288 for DB2
+            // https://github.com/doctrine/dbal/issues/3290 for Oracle
+            $this->markTestIncomplete('Platform does not support stream resources as parameters');
+        }
+
+        $stmt = $this->_conn->prepare("INSERT INTO blob_table(id, clobfield, blobfield) VALUES (1, 'ignored', ?)");
+
+        $stream = null;
+        $stmt->bindParam(1, $stream, ParameterType::LARGE_OBJECT);
+
+        // Bind param does late binding (bind by reference), so create the stream only now:
+        $stream = fopen('data://text/plain,test', 'r');
+
+        $stmt->execute();
+
+        $this->assertBlobContains('test');
+    }
+
     private function assertBlobContains($text)
     {
-        $rows = $this->_conn->fetchAll('SELECT * FROM blob_table');
+        $rows = $this->_conn->query('SELECT blobfield FROM blob_table')->fetchAll(FetchMode::COLUMN);
 
         self::assertCount(1, $rows);
-        $row = array_change_key_case($rows[0], CASE_LOWER);
 
-        $blobValue = Type::getType('blob')->convertToPHPValue($row['blobfield'], $this->_conn->getDatabasePlatform());
+        $blobValue = Type::getType('blob')->convertToPHPValue($rows[0], $this->_conn->getDatabasePlatform());
 
         self::assertInternalType('resource', $blobValue);
         self::assertEquals($text, stream_get_contents($blobValue));


mpdude:
Build failures seem to be unrelated/unreliable CI

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -230,6 +242,27 @@ public function execute($params = null)
...
 
+    /**
+     * Send BLOB type streams using mysqli::send_long_data
+     */
+    private function _sendLongData()
+    {
+        foreach ($this->_longData as $paramNr => $resource) {
+            if (is_resource($resource) && get_resource_type($resource) == 'stream') {
+                while (!feof($resource)) {
+                    $longData = fread($resource, 8192);
```
mpdude: 8192 was just an arbitrary choice. Don't know if another chunk size would make more sense.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -230,6 +246,27 @@ public function execute($params = null)
...
 
+    /**
+     * Send BLOB type streams using mysqli::send_long_data
+     */
+    private function _sendLongData()
+    {
+        foreach ($this->_longData as $paramNr => $resource) {
+            if (is_resource($resource) && get_resource_type($resource) === 'stream') {
+                while (! feof($resource)) {
+                    $longData = fread($resource, 8192);
```
mpdude: 8192 was just an arbitrary choice. Don't know if another chunk size would make more sense.

mpdude:
@morozov Does this make sense to you?

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -230,6 +246,27 @@ public function execute($params = null)
...
+                    if (! $this->_stmt->send_long_data($paramNr, $longData)) {
+                        throw new MysqliException($this->_stmt->error, $this->_stmt->sqlstate, $this->_stmt->errno);
+                    }
+                }
+            } else {
+                if (! $this->_stmt->send_long_data($paramNr, $resource)) {
```
morozov: Why use `send_long_data()` for non-streams? We should bind them as usual.

tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
```diff
@@ -55,6 +57,43 @@ public function testInsert()
         self::assertEquals(1, $ret);
     }
 
+    public function testInsertProcessesStream()
+    {
+        $this->_conn->insert('blob_table', [
+            'id'          => 1,
+            'clobfield'   => 'ignored',
+            'blobfield'   => fopen('data://text/plain,test', 'r'),
+            'binaryfield' => fopen('data://text/plain,test', 'r'),
```
morozov: > The `binary` and `blob` types map `BLOB`/`BINARY`/`VARBINARY` columns to PHP streams.

It's not exactly true (see https://github.com/doctrine/dbal/commit/5465c6c90fb41bd9f040716f902f015abb710453). Only BLOB and CLOB (TEXT) fields should be populated from streams.

tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
```diff
@@ -55,6 +57,43 @@ public function testInsert()
         self::assertEquals(1, $ret);
     }
 
+    public function testInsertProcessesStream()
+    {
+        $this->_conn->insert('blob_table', [
+            'id'          => 1,
+            'clobfield'   => 'ignored',
```
morozov: Please test this as well.

tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
```diff
@@ -55,6 +57,43 @@ public function testInsert()
...
+            ParameterType::STRING,
+            ParameterType::LARGE_OBJECT,
+            ParameterType::LARGE_OBJECT,
+        ]);
+
+        $this->assertBlobContains('test');
+        $this->assertBinaryContains('test');
+    }
+
+    public function testInsertCanHandleStreamLongerThanChunkSize()
```
morozov: Wouldn't one test be enough to test a large-ish value? If a large value can be stored, a small one can too.

morozov (changes requested):
@mpdude thank you, the improvement makes sense. Please see my comments.

Besides that, would it make sense to test the case when the data size is actually longer than `max_allowed_packet`? Otherwise, the test scenarios would work even without the fix.

mpdude:
Thanks for the review!

My initial problem (which is why I created this PR) was that when passing in a stream, a literal value like `"Resource id #126"` was written into the DB.

I did not intend to work around the `max_allowed_packet` setting, and quick tests show that this is not possible: When I configure MySQL to use a low `max_allowed_packet` value, an error will be triggered stating `Parameter of prepared statement which is set through mysql_send_long_data() is longer than 'max_allowed_packet' bytes`. 

Documentation for the underlying `mysql_stmt_send_long_data()` C API function ([here](https://dev.mysql.com/doc/refman/8.0/en/mysql-stmt-send-long-data.html), the first sentence [here](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_max_allowed_packet) and also [this bug](https://bugs.mysql.com/bug.php?id=83958)) suggests that `max_allowed_packet` is always a hard limit. 

What `mysqli::send_long_data()` seems to allow is that every data chunk of data passed to it is immediately sent out to the network. I have confirmed this using `tcpdump`, and so the advantage might be that we can keep the memory footprint low on the PHP side while processing streams.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
mpdude: I could change that like so: https://gist.github.com/mpdude/1ad1c03556d9fd4191396159b7929124

Basically, if you bind as `ParameterType::LARGE_OBJECT` but do *not* pass in a stream, just treat it like `ParameterType::STRING`. Might lose the advantage of sending the parameter in a network package by itself, though.

tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
mpdude: So you want to process streams also for `ParameterType::STRING`?

Of course we could do, but then I don't really get the purpose of the parameter types... On `SELECT`, only the `LARGE_OBJECT` type will return a stream, whereas `STRING` is a... string.

morozov: No, you should be able to bind a bind a character stream to a CLOB field. `LARGE_OBJECT` doesn't imply *B*LOB, it implies _any_ LOB.

tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
mpdude: I think I need some more explications here 🤔

Please have a look at the [current test on 2.7](https://github.com/doctrine/dbal/blob/9efbf7f3e260442ceefe4aec163c18b193c8e0c2/tests/Doctrine/Tests/DBAL/Functional/BlobTest.php):

* `clobfield` is always used with `ParameterType::STRING`. It does not return a stream on `SELECT`, so I also did not expect streams to be bound here (could change that).
* `blobfield` is a `ParameterType::LARGE_OBJECT`, should process streams when binding data and return streams from `SELECT` (as-is right now).
* `binaryfield` is always treated like `blobfield`, including the [assertions that streams are returned](https://github.com/doctrine/dbal/blob/9efbf7f3e260442ceefe4aec163c18b193c8e0c2/tests/Doctrine/Tests/DBAL/Functional/BlobTest.php#L102). Only difference is that [its database type is `binary`](https://github.com/doctrine/dbal/blob/9efbf7f3e260442ceefe4aec163c18b193c8e0c2/tests/Doctrine/Tests/DBAL/Functional/BlobTest.php#L30).

Your remarks would make sense to me if `clobfield` were treated as `ParameterType::LARGE_OBJECT` and `binaryfield` were `ParameterType::STRING`, because then stream processing (on read as well as write) would only happen for the `LARGE_OBJECT` type.

tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
morozov: > On `SELECT`, only the `LARGE_OBJECT` type will return a stream, whereas `STRING` is a... string.

I think this is an issue we should fix in 3.0. Regardless of whether it's TEXT or BLOB, it should be returned as a stream.

mpdude:
@morozov I think I need some more feedback, please see my comments above.

morozov:
@mpdude, all above is correct. What I meant by "the test scenarios would work even without the fix", is that currently, it is already possible to store strings like "test" or 40K times "x" using the currently used MySQL API but it's not possible to store strings longer than 16M. If we don't cover this case with a test, then instead of using `send_long_data()`, we can use `stream_get_contents()` and all tests still would pass. If we test a really long string, then changing the implementation will fail the test.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -42,9 +46,7 @@ class MysqliStatement implements \IteratorAggregate, Statement
         ParameterType::BOOLEAN      => 'i',
         ParameterType::NULL         => 's',
         ParameterType::INTEGER      => 'i',
-
-        // TODO Support LOB bigger then max package size
```
morozov: @mpdude the fact that this is done is not covered by tests.

mpdude:
> What I meant by "the test scenarios would work even without the fix", is that currently, it is already possible to store strings like "test" or 40K times "x" using the currently used MySQL API but it's not possible to store strings longer than 16M.

I think there is no way we can exceed the `max_allowed_packet` size in MySQL, and this PR does not try to attempt this. 

What this PR actually *does* fix is that passing streams together with the `ParameterType::LARGE_OBJECT` indeed processes the stream instead of casting it to a string (which results in a "Resource id #..." string). And this is covered by the test, they don't pass against the 2.7 implementation of `MysqliStatement`.

> If we don't cover this case with a test, then instead of using send_long_data(), we can use stream_get_contents() and all tests still would pass. If we test a really long string, then changing the implementation will fail the test.

As I said above, we cannot process a value larger than `max_allowed_package` and so a test for it makes no sense.

But you are right that instead of using `mysqli::send_long_data()`, someone might change the implementation of stream processing to use `stream_get_contents()`. This would increase the memory footprint in PHP (loads the entire stream into memory at once), but otherwise would still work.

Any idea what a test might look like that guards against this?

mpdude:
@morozov Please have a look at https://github.com/mpdude/dbal/pull/1. Does `BlobTest` make more sense that way?

It would only deal with `text` and `blob` type columns and make sure that both can be updated with strings and streams when the `ParameterType::LARGE_OBJECT` binding is used. Additionally, it checks for `SELECT`s that the `text` type returns a string and `blob` returns a stream.

tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
mpdude: Your remark regarding changed behaviour is true for 3.x, but this is against 2.7.

tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
mpdude: I think the test covers it now?

mpdude:
@morozov better like this?

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
morozov: Yeah, valid point. Let's send strings via `send_long_data()` too. Please rename `$resource` to `$value` then, since it's not always a resource and it's by design.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
morozov: @mpdude isn't using `send_long_data()` the way to store the data longer than the max package size?

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -84,6 +86,14 @@ class MysqliStatement implements \IteratorAggregate, Statement
      */
     protected $_values = [];
 
+    /**
+     * Contains values from bindValue() that need to be sent
+     * using send_long_data *after* bind_param has been called.
+     *
+     * @var mixed[]
```
morozov: It's probably `string[]|resource[]`.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -230,6 +246,27 @@ public function execute($params = null)
         return true;
     }
 
+    /**
+     * Send BLOB type streams using mysqli::send_long_data
+     */
+    private function _sendLongData()
+    {
+        foreach ($this->_longData as $paramNr => $resource) {
+            if (is_resource($resource) && get_resource_type($resource) === 'stream') {
```
morozov: Should we throw an exception if it's a non-stream resource? No point in trying to send the resource itself in the `else` block.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
morozov: `fread()` can return `false` in case of a failure. It should be handled before trying to send the data.

tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
```diff
@@ -81,20 +124,54 @@ public function testUpdate()
...
+        $this->assertBinaryContains('test2');
+    }
+
+    public function testUpdateProcessesStream()
+    {
+        $this->_conn->insert('blob_table', [
+            'id'          => 1,
+            'clobfield'   => 'test',
+            'blobfield'   => 'test',
+            'binaryfield' => 'test',
```
morozov: Should be ignored.

tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
```diff
@@ -81,20 +124,54 @@ public function testUpdate()
...
+            ParameterType::LARGE_OBJECT,
+            ParameterType::LARGE_OBJECT,
+            ParameterType::LARGE_OBJECT,
+        ]);
+
+        $this->_conn->update('blob_table', [
+            'id'          => 1,
+            'clobfield'   => fopen('data://text/plain,test2', 'r'),
+            'blobfield'   => fopen('data://text/plain,test2', 'r'),
+            'binaryfield' => fopen('data://text/plain,test2', 'r'),
```
morozov: Should be ignored.

tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
```diff
@@ -47,14 +49,53 @@ public function testInsert()
...
         self::assertEquals(1, $ret);
     }
 
+    public function testInsertProcessesStream()
+    {
+        $this->_conn->insert('blob_table', [
+            'id'          => 1,
+            'clobfield'   => fopen('data://text/plain,test', 'r'),
+            'blobfield'   => fopen('data://text/plain,test', 'r'),
+            'binaryfield' => fopen('data://text/plain,test', 'r'),
```
morozov: Should be ignored.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -230,6 +246,27 @@ public function execute($params = null)
...
+                while (! feof($resource)) {
+                    $longData = fread($resource, 8192);
+                    if (! $this->_stmt->send_long_data($paramNr, $longData)) {
```
morozov: The `send()`/`throw` block could be moved into a separate private method. Otherwise, there's code duplication and the nesting is getting too deep.

morozov:
> I think there is no way we can exceed the `max_allowed_packet` size in MySQL, and this PR does not attempt this.

Why not?
```php
$conn->insert('blob_table', [
    'id'          => 1,
    'blobfield'   => str_repeat('x', 0x1000001),
]);
```
Results in:
> Warning: Doctrine\DBAL\Connection::executeUpdate(): Error occurred while closing statement in lib/Doctrine/DBAL/Connection.php on line 1055
> 
> Fatal error: Uncaught Doctrine\DBAL\Driver\Mysqli\MysqliException: Got a packet bigger than 'max_allowed_packet' bytes in lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php on line 109

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
mpdude: No, I tried to explain that in the comments and discussion. You *cannot* overcome the `max_allowed_package` limitation. 

`send_long_data()` will – for each invocation – send out a chunk of data immediately to the MySQL server. So in the first place, it saves you from having to buffer one big query on the client side (memory allocation) before sending that out. We can read streams in small steps and send data to the server, never keeping everything at once in memory.

The MySQL server will still enforce the `max_allowed_packet` limitation and reject queries exceeding it. That happens even if you have sent more data than allowed in small chunks. 

It's not obvious, but the documentation and also the bug report I linked above shortly mention it.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
mpdude: It can be either a resource or a string. I'll rename the variable.

mpdude: :+1:

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
mpdude: :+1:

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
mpdude: :+1:

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
mpdude: :+1:

mpdude:
@morozov I think I've got all your comments covered and fixed.

Regarding the `max_allowed_packet`, please see https://github.com/doctrine/dbal/pull/3217#discussion_r216558493.

tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
```diff
@@ -124,4 +175,17 @@ private function assertBlobContains($text)
         self::assertInternalType('resource', $blobValue);
         self::assertEquals($text, stream_get_contents($blobValue));
     }
+
+    private function assertClobContains($text)
+    {
+        $rows = $this->_conn->fetchAll('SELECT * FROM blob_table');
+
+        self::assertCount(1, $rows);
+        $row = array_change_key_case($rows[0], CASE_LOWER);
```
morozov: You can use `$conn->query()->fetchAll(FetchMode::COLUMN)` and avoid the case conversion (it's not part of the SUT).

tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
```diff
@@ -124,4 +175,17 @@ private function assertBlobContains($text)
...
+
+        $blobValue = Type::getType('text')->convertToPHPValue($row['clobfield'], $this->_conn->getDatabasePlatform());
```
morozov: As long as the assertion is about the CLOB field, please use `$clob*` names.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -230,6 +246,44 @@ public function execute($params = null)
...
+
+    /**
+     * Bind parameters using send_long_data
+     *
+     * @param int    $paramNr  Parameter offset
+     * @param string $longData A chunk of data to send
+     *
+     * @throws MysqliException
+     */
+    private function _sendLongData($paramNr, $longData): void
```
morozov: There's definitely something wrong with the CI, there [should be a space](https://github.com/doctrine/coding-standard/blob/1d35fd5/lib/Doctrine/ruleset.xml#L269) before the colon.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -230,6 +246,44 @@ public function execute($params = null)
         return true;
     }
 
+    /**
+     * Handle $this->_longData after regular query parameters have been bound
+     *
+     * @throws MysqliException
+     */
+    private function _processLongData()
```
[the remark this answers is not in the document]
morozov: @carusogabriel do we enforce this in the standard? This is what I'm getting locally for this PR:
```
$ git-phpcs upstream/2.7...c71d64c

FILE: lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
------------------------------------------------------------------------------------------------
------------------------------------------------------------------------------------------------
 280 | ERROR | [x] There must be exactly 1 whitespace between closing parenthesis and return type
     |       |     colon.
     |       |     (SlevomatCodingStandard.TypeHints.ReturnTypeHintSpacing.IncorrectWhitespaceBeforeColon)
------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
------------------------------------------------------------------------------------------------
```

morozov:
@mpdude please retarget this against `master`. `2.7` is now security-only, I'll backport it to `2.8` myself.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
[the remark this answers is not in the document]
carusogabriel: Yeap, we do. Here’s how our return types should be

    public function foo() : string

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -230,6 +246,44 @@ public function execute($params = null)
...
+    {
+        foreach ($this->_longData as $paramNr => $value) {
+            if (is_resource($value) && get_resource_type($value) === 'stream') {
+                $stream = $value;
+                while (! feof($stream)) {
+                    $chunk = fread($stream, 64 * 1024);
```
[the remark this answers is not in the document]
mpdude: From the commit msg:

> Set chunk size to 64 KB. Still, this is an arbitrary choice. But as send_long_data() is designed to send out data to the network immediately, it is probably a good thing performance-wise to have a little bit bigger chunks (think TCP windows or offloading). And, on the other hand, 64 KB of memory in PHP should not hurt that much.

Does that make sense? WDYT?

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -230,6 +246,44 @@ public function execute($params = null)
...
 
+    /**
+     * Handle $this->_longData after regular query parameters have been bound
+     *
+     * @throws MysqliException
+     */
+    private function _processLongData()
+    {
+        foreach ($this->_longData as $paramNr => $value) {
+            if (is_resource($value) && get_resource_type($value) === 'stream') {
```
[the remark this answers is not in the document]
mpdude: I see. Will fix that.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -230,6 +246,44 @@ public function execute($params = null)
...
+                $stream = $value;
```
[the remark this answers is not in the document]
mpdude: This was meant to improve clarity in the code and has no additional cost in PHP.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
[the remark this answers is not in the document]
mpdude: I've fixed the return type hint (missing space).

Regarding non-public members with a leading underscore, I just followed the practice in the rest of the fields. 🤷‍♂️

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -230,6 +246,44 @@ public function execute($params = null)
...
+                while (! feof($stream)) {
+                    $chunk = fread($stream, 64 * 1024);
+                    if ($chunk === false) {
+                        throw new MysqliException("Failed processing the stream resource for parameter offset ${paramNr}.");
```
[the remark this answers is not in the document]
mpdude: fixed

mpdude:
@morozov This is now based on master and I hope I did not miss any of your comments.

One open question remains regarding the use of underscore-prefixed names for non-public methods and fields. I just followed the way the class was written before, so...?

mpdude:
Is there anything I can do here to make the builds pass?

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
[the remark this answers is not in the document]
morozov: @carusogabriel I meant the leading underscore in the `_processLongData()` method name. We shouldn't allow it (PSR-2) doesn't.

@mpdude the codebase is full of coding standard violations since fixing some of them (particularly, renaming methods) would lead to BC breaks. We took an incremental approach and only check that the newly added or edited code is compliant with the standard.

morozov:
> Is there anything I can do here to make the builds pass?

You can follow https://github.com/doctrine/dbal/pull/3217#discussion_r216890919. DB2 and Oracle capitalize column names. Given that you only fetch one column, you don't need to refer to it by name, just use its value.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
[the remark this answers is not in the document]
morozov: I don't think that performance optimization without numbers at hand makes much sense. 8K is a commonly used default and looks clear. Anything else implies that we did some research and decided that size is better than the default and therefore may be confusing.

Let's stick to the size from the [PHP manual](http://www.php.net/manual/en/mysqli-stmt.send-long-data.php#example-1935):
```php
$stmt = $mysqli->prepare("INSERT INTO messages (message) VALUES (?)");
$null = NULL;
$stmt->bind_param("b", $null);
$fp = fopen("messages.txt", "r");
while (!feof($fp)) {
    $stmt->send_long_data(0, fread($fp, 8192));
}
fclose($fp);
$stmt->execute();
```

morozov:
> One open question remains regarding the use of underscore-prefixed names for non-public methods and fields. I just followed the way the class was written before, so...?

Please remove them. The Doctrine coding standard is [based on PSR-2](https://github.com/doctrine/coding-standard/blob/1d35fd5aca57d2a216778904320dcf39f6d90a6a/lib/Doctrine/ruleset.xml#L10) which in turn [doesn't allow using underscores](https://www.php-fig.org/psr/psr-2/#43-methods).

> The other is about the chunk size of 8 vs 64 KB, see [#3217 (comment)](https://github.com/doctrine/dbal/pull/3217#discussion_r216889410).

Answered above.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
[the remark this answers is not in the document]
mpdude: Will do

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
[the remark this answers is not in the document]
mpdude: Ok, but I'm just going to fix the new ones introduced in this PR.

tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
mpdude: D'oh! I thought you confused `COLUMN` and `FETCH_ASSOC`, but now it makes sense.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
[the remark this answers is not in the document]
carusogabriel: @morozov Gotcha. Gonna write some tests for our CS and open a PR there if we don't have this rule :)

mpdude:
Regarding the failing builds I meant that appveyor and continuousphp seem to fail almost immediately and I don't understand what's wrong.

mpdude:
Also rebased against master since #3285 has been merged.

mpdude:
DB2 was like :trollface:

morozov:
> Regarding the failing builds I meant that appveyor and continuousphp seem to fail almost immediately and I don't understand what's wrong.

Right now, AppVeyor passes, and ContinuousPHP has failures in `BlobTest`. Let me know if you get stuck fixing them.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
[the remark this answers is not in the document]
morozov: > Ok, but I'm just going to fix the new ones introduced in this PR.

Sure. This is how our incremental checks work.

> Gotcha. Gonna write some tests for our CS and open a PR there if we don't have this rule :)

@carusogabriel it turns out to be a configuration issue (https://github.com/doctrine/dbal/pull/3287).

mpdude:
Build failures: In DB2, inserts and updates from streams cause the same issue that this PR addresses for MySQLi, namely that the stream is cast into a string and `Resource id #...` is written to the DB. 

Not sure if this is a new issue or just new tests documenting it.

morozov:
@mpdude it's a known issue documented by your test. Please file the issue and skip DB2 in your test.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
morozov: > Might lose the advantage of sending the parameter in a network package by itself, though.

This did make sense since I thought one could send a string which exceeds `max_allowed_packet` using `send_long_data()`. Given it's not the case, what's the advantage of sending the parameter in a network package by itself? Isn't it better to just bind the string as is and let the driver decide how to send it?

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -228,6 +247,50 @@ public function execute($params = null)
...
 
+    /**
+     * Handle $this->_longData after regular query parameters have been bound
+     *
+     * @throws MysqliException
+     */
+    private function processLongData()
+    {
+        foreach ($this->longData as $paramNr => $value) {
+            if (!is_resource($value)) {
```
morozov: From https://github.com/doctrine/dbal/pull/3217#discussion_r216139584:

> Might lose the advantage of sending the parameter in a network package by itself, though.

This did make sense since I thought one could send a string which exceeds `max_allowed_packet` using `send_long_data()`. Given it's not the case, what's the advantage of sending the parameter in a network package by itself? Isn't it better to just bind the string as is and let the driver decide how to send it?

mpdude: Seems to make a difference regarding memory consumption, but that must happen down inside the PHP core/mysqli driver.

Please confirm by trying the following:

Patch the `BlobTest` like so:

```
diff --git a/tests/Doctrine/Tests/DBAL/Functional/BlobTest.php b/tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
index 2c41ed89e..bc90c679d 100644
--- a/tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
@@ -37,17 +37,20 @@ class BlobTest extends \Doctrine\Tests\DbalFunctionalTestCase

     public function testInsert()
     {
+        $lob = str_repeat('x', 8 * 1024 * 1024);
         $ret = $this->_conn->insert('blob_table', [
             'id'          => 1,
-            'clobfield'   => 'test',
-            'blobfield'   => 'test',
+            'clobfield'   => $lob,
+            'blobfield'   => $lob,
         ], [
             ParameterType::INTEGER,
-            ParameterType::LARGE_OBJECT,
-            ParameterType::LARGE_OBJECT,
+            ParameterType::STRING,
+            ParameterType::STRING,
         ]);

         self::assertEquals(1, $ret);
+
+        print memory_get_peak_usage() / (1024*1024);
     }

     public function testInsertProcessesStream()
```

This creates a 8 MB BLOB, inserts it as a string and effectively short-circuits the entire `send_long_data()` logic introduced in this PR. At the end, it prints the peak memory usage, which for me was around 44 MB.

Then, change back to `ParameterType::LARGE_OBJECT`. For me, memory usage drops to 28 MB.

As the difference is two times the size of the LOB, my assumption is that somewhere down inside PHP, the MySQL driver must copy together all the data to be sent to MySQL into some buffer. Maybe `send_long_data()` can prevent this by sending out the data immediately.

Bottom line: I'd like to keep it this way as it leaves it to the user to decide whether to try any advanced optimizations, or just use `ParameterType::STRING` and pass in large strings. YMMV.

mpdude:
@morozov Now it's Oracle failing at https://app.continuousphp.com/git-hub/doctrine/dbal/build/ee788937-49f1-4be2-a652-d8dc7fe72712/logs/a9bb62f2-3809-4d79-93b4-f24dc1719040.

I don't have a clue about BLOB/CLOB types on Oracle, do you see what's wrong?

morozov:
@mpdude `testInsert()` fails because Oracle is the only supported platform which handles BLOB and CLOB streams differently (it needs a different resource for them but we cannot guess which since we use `ParameterType::LARGE_OBJECT` for both). `testUpdate()` fails because Oracle has its own API for streams (`OCI-Lob`) which we currently don't support.

Please mark both tests incomplete on Oracle.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -82,6 +87,14 @@ class MysqliStatement implements \IteratorAggregate, Statement
      */
     protected $_values = [];
 
+    /**
+     * Contains values from bindValue() that need to be sent
+     * using send_long_data *after* bind_param has been called.
+     *
+     * @var string[]|resource[]
+     */
+    protected $longData = [];
```
morozov: Please declare it `private` instead. This data is not supposed to be reused via inheritance.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
morozov: The real problem is storing an 8 MB string in memory, not the way we send it to the DB.

I don't think we should solve this problem at the cost of complicating the API. In fact, every DB driver will have to be able to bind a string and a resource as `ParameterType::LARGE_OBJECT` which adds unnecessary complexity.

If memory usage critical is important but the data is already in memory, it can be dumped into a temporary file which then can be `fopen()`ed and bound to the statement.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -228,6 +247,50 @@ public function execute($params = null)
...
 
+    /**
+     * Handle $this->_longData after regular query parameters have been bound
+     *
+     * @throws MysqliException
+     */
+    private function processLongData()
+    {
+        foreach ($this->longData as $paramNr => $value) {
+            if (! is_resource($value)) {
```
morozov: Unfinished discussion: https://github.com/doctrine/dbal/pull/3217#discussion_r217223893.

mpdude:
Regarding Oracle test failures:

One failure is in `Doctrine\Tests\DBAL\Functional\Schema\OracleSchemaManagerTest::testDropsDatabaseWithActiveConnections`, where I don't see why this PR is relevant at all.

Apart from that, it fails all tests in `BlobTest`, even those marked with 

```php
if (in_array($this->_conn->getDatabasePlatform()->getName(), ['oracle', 'db2'], true)) {
    $this->markTestSkipped('Platform does not support stream resources as parameters');
}
```

I would have expected those to pass? Does the platform have another name.

Could you please explain why there are two columns in `BlobTest` at all, one for `text` and one for `blob`? If Oracle is the only platform that makes a difference between those types and we'd need to skip the `text` half for it, can't we just drop it?

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
mpdude: done

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
mpdude: See above.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
mpdude: I don't get your point 😕. But re-reading my previous comment, I found it rather confusing and misunderstandable. So, let me try to restate it and I hope we can get along.

If you have your data – for whatever reason –  as a string in memory and pass it as the `STRING` type, there's not much we can do about it.

If you use the `LARGE_OBJECT` type to indicate your data is something "heavy", DBAL should do its best to handle it the most efficient way. 

If you pass in a stream, `mysqli` can process it in chunks. To make this work is what this PR is about. (Other platforms still lack support as the skipped tests show. Maybe the particular drivers can find efficient ways of handling streams, maybe they need to fall back to reading the entire stream and treat it as `STRING`. But all that is out of scope here.)

If you pass in a string, `mysqli` has a better memory footprint when using `send_long_data` instead of `bind_param`. This optimization is a side-effect of this PR, but it is also an internal detail of the DBAL MySQLi driver the user need not worry about.

So, is your suggestion that `LARGE_OBJECT` should no longer accept strings but only streams?

`BlobTest` suggests that passing strings as `LARGE_OBJECT` is supported for all platforms right now. It's stream support where we only have partial coverage (not DB2, not Oracle?)

mpdude:
Strange, the last ContinuousPhp build at https://app.continuousphp.com/git-hub/doctrine/dbal/build/bd400d80-e773-4709-8ab1-a2c7c2915256/logs/715b6472-4a32-42df-804d-53182192e5e8 had the stream-based tests skipped (as I'd expect) and does not exhibit the problem with `Doctrine\Tests\DBAL\Functional\Schema\OracleSchemaManagerTest::testDropsDatabaseWithActiveConnections`.

Still the question remains why the CLOB/`text` field makes sense in `BlobTest` if Oracle cannot handle it as `LARGE_OBJECT` and all other DBMSes don't make a difference. Maybe I don't get the real purpose of the `BlobTest`.

morozov:
> Still the question remains why the CLOB/`text` field makes sense in `BlobTest` if Oracle cannot handle it as `LARGE_OBJECT` and all other DBMSes don't make a difference. Maybe I don't get the real purpose of the `BlobTest`.

The purpose if the test is to test the `Blob` type but historically it covers other types as well. It covers `Binary` because it's the same as `Blob` on PostgeSQL and SQLite. It covers `Text` because it's almost the same as `Blob` (streams, etc).

Given that this ticket originally wasn't about CLOBs, I think it makes sense to revert non-BLOB related changes from the test and let it pass on Oracle. Later, we can clean the test up and fix handling BLOBs and CLOBs on Oracle.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
morozov: Please link to the comment.

mpdude: https://github.com/doctrine/dbal/pull/3217#discussion_r217895850 (I wanted to keep the discussion together in one thread.)

mpdude:
Reverted test changes covering CLOBs.

Remaining discussion: Handle string inputs for `LARGE_OBJECT` parameters? At https://github.com/doctrine/dbal/pull/3217#discussion_r217895850.

mpdude:
Tests ✅🏝

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
morozov: > So, is your suggestion that `LARGE_OBJECT` should no longer accept strings but only streams?

It should remain as is in order to not introduce breaking changes — a string bound as `LARGE_OBJECT` should be handled as a string, without using the `send_long_data()`. In the future, I'd like to deprecate the ability to bind strings as LOB. A string parameter should expect a string value, a LOB should expect a stream — just as simple as that.

> If you use the `LARGE_OBJECT` type to indicate your data is something "heavy", DBAL should do its best to handle it the most efficient way.

I don't agree here. DBAL doesn't have to optimize internal defects of a driver at the cost of increasing its own complexity. As a driver user, I'd indeed expect a large string bound to a statement to not increase the memory footprint significantly. If it's a bug, it should be fixed in the driver so that not only DBAL users could benefit from that.

> `BlobTest` suggests that passing strings as `LARGE_OBJECT` is supported for all platforms right now. It's stream support where we only have partial coverage (not DB2, not Oracle?)

This is just because the stream support is not yet implemented in other platforms, and it was relatively easy to implement handling strings. From the DB application standpoint, there's really no reason to represent LOBs as strings. They are just files stored in the DB.

morozov: > But re-reading my previous comment, I found it rather confusing and misunderstandable. So, let me try to restate it and I hope we can get along.

I do completely get your point, I just don't agree.

tests/Doctrine/Tests/DBAL/Functional/BlobTest.php
```diff
@@ -49,6 +51,27 @@ public function testInsert()
         self::assertEquals(1, $ret);
     }
 
+    public function testInsertProcessesStream()
+    {
+        if (in_array($this->_conn->getDatabasePlatform()->getName(), ['oracle', 'db2'], true)) {
+            // https://github.com/doctrine/dbal/issues/3288 for DB2
+            $this->markTestSkipped('Platform does not support stream resources as parameters');
```
morozov: Nitpicking: both platforms do support streams, their support is just not implemented in DBAL. Therefore, the test should be marked as incomplete (therefore, requiring a fix), not skipped (would be correct if the platforms really didn't support the feature).

mpdude: fixed.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
mpdude: Changed. 

Once this is merged I can open another issue – maybe it helps someone else in the future or we might want to revisit this sometime.

morozov: That sounds good to me. You may also want to file a ticket with `ext-mysqli` to get it potentially fixed in a proper place.

morozov:
@mpdude looks like we're approaching the finish. Once you get the code style fixed, please squash the commits.

mpdude:
🤞

morozov (dismissed):

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -141,6 +154,11 @@ public function bindParam($column, &$variable, $type = ParameterType::STRING, $l
      */
     public function bindValue($param, $value, $type = ParameterType::STRING)
```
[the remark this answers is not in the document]
morozov: That's a valid concern.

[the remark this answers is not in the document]
mpdude: Too bad we did not notice this earlier 🤔

Seems `bindParam` is completely missed at least in `BlobTest`.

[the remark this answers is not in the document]
morozov: IIRC, in most drivers, `bindValue()` just proxies to `bindParam()`, so the latter is not tested individually. I once tried to do the same for `mysqli` but it didn't work and it wasn't my primary goal at the time. Maybe we could try to generalize the implementation again.

mpdude:
Note to self: Tests hitting `bindParam()` at all:

- `DataAccessTest`
- `StatementTest`

mpdude:
@kimhemsoe I've added a test for `bindParam()` and hope I've got it all covered. Maybe you could have a look?

@morozov `bindParam()` has to take a variable reference and cannot check if it has a stream resource or a string until `execute` is called. Sorting things out inside `execute()` will probably get very messy... But if we re-consider using `send_long_data` also for strings passed as `ParameterType::LARGE_OBJECT`, we can get away with minimal changes (see last commit).

WDYT?

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -228,6 +253,50 @@ public function execute($params = null)
...
+    /**
+     * Handle $this->_longData after regular query parameters have been bound
+     *
+     * @throws MysqliException
+     */
+    private function processLongData()
+    {
+        foreach ($this->longData as $paramNr => $value) {
+            if (! is_resource($value)) {
+                $this->sendLongData($paramNr, $value);
```
morozov: Please bind strings as strings. Even if a LOB parameter is bound by reference, its value is already known by the moment `execute()` is called.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -82,6 +87,14 @@ class MysqliStatement implements \IteratorAggregate, Statement
      */
     protected $_values = [];
 
+    /**
+     * Contains values from bindValue() that need to be sent
+     * using send_long_data *after* bind_param has been called.
```
morozov: After https://github.com/doctrine/dbal/pull/3217/files#r219689244 is addressed, this will no longer be true. It will contain the values which we bound as LOB, not necessarily will be sent via `send_long_data()`.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -228,6 +245,63 @@ public function execute($params = null)
...
+     * Bind parameters using send_long_data
+     *
+     * @param int    $paramNr  Parameter offset
+     * @param string $longData A chunk of data to send
+     *
+     * @throws MysqliException
+     */
+    private function sendLongData($paramNr, $longData) : void
+    {
+        if (! $this->_stmt->send_long_data($paramNr, $longData)) {
```
morozov: Does this need to exist as a separate method? It's only called from `processLongData()`. It'd make sense to leave only one named `sendLongData()` since the "process" term is too abstract,

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -228,6 +245,63 @@ public function execute($params = null)
         return true;
     }
 
+    /**
+     * Move all stream resources that are bound with the LARGE_OBJECT type
+     * from the _bindesValues array over to longData so that we can
```
morozov: Typo in "_bindesValues".

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -228,6 +245,63 @@ public function execute($params = null)
...
+        foreach ($this->_bindedValues as $parameter => $value) {
+            if ($this->types[$parameter - 1] !== static::$_paramTypeMap[ParameterType::LARGE_OBJECT]) {
+                continue;
+            }
+
+            if (is_resource($value)) {
+                if (get_resource_type($value) !== 'stream') {
+                    throw new InvalidArgumentException('Resources passed with the LARGE_OBJECT parameter type must be stream resources.');
+                }
+                $this->_bindedValues[$parameter] = null;
```
morozov: I don't think it's a good idea to change the object state at the statement execution time. If somebody decides to reuse an existing stream and an existing statement, it won't work.

Instead of changing the object state, this method could return an array like `[$boundValues, $longData]`, each of which would be processed accordingly. This way, the state of the object will remain intact, and the `$longData` property will be no longer needed.

mpdude:
@morozov Updated.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -228,6 +235,59 @@ public function execute($params = null)
...
 
+    /**
+     * Split $this->_bindedValues into those values that need to be sent using mysqli::send_long_data()
+     * and those that can be bound the usual way.
+     *
+     * @return array
+     */
+    private function separateLongData()
+    {
+        $streams = $otherData = [];
```
morozov: Would it make sense to call `$otherData` `$values` same as in the calling method?

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -228,6 +235,59 @@ public function execute($params = null)
         return true;
     }
 
+    /**
+     * Split $this->_bindedValues into those values that need to be sent using mysqli::send_long_data()
+     * and those that can be bound the usual way.
+     *
+     * @return array
+     */
+    private function separateLongData()
```
morozov: Would it make sense to call it `separateBoundValues()` instead? We want to take all bound values and separate them in two groups, one of which is the streams which need to be sent using `mysqli::send_long_data()`.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -228,6 +235,59 @@ public function execute($params = null)
...
+            }
+
+            if (is_resource($value)) {
+                if (get_resource_type($value) !== 'stream') {
+                    throw new InvalidArgumentException('Resources passed with the LARGE_OBJECT parameter type must be stream resources.');
+                }
+                $streams[$parameter]   = $value;
+                $otherData[$parameter] = null;
+            } else {
+                $this->types[$parameter - 1] = static::$_paramTypeMap[ParameterType::STRING];
```
morozov: Can we return `$types` as part of the return value instead of modifying the object state? Look at `SQLParserUtils::expandListParameters()` for example.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -228,6 +235,59 @@ public function execute($params = null)
...
+     * and those that can be bound the usual way.
+     *
+     * @return array
+     */
+    private function separateLongData()
+    {
+        $streams = $otherData = [];
+
+        foreach ($this->_bindedValues as $parameter => $value) {
+            if ($this->types[$parameter - 1] !== static::$_paramTypeMap[ParameterType::LARGE_OBJECT]) {
```
morozov: AFAIK, types may be omitted and default to `STRING`. We probably should check if the value is set.

mpdude:
@morozov what would be the right `@return` type hunt for the `separateBoundValues` method?

morozov:
It should be `mixed[][]` since it's an array of arrays where the inner arrays can contain almost anything.

mpdude:
Would that also pass static code analysis? 

I cannot really make sense of https://travis-ci.org/doctrine/dbal/jobs/433673047#L543.

morozov:
It should. You can run `vendor/bin/phpstan A` locally and see before pushing the changes.

.gitignore
```diff
@@ -8,3 +8,4 @@ vendor/
 /phpunit.xml
 /.phpcs-cache
 /phpstan.neon
+/.idea
```
morozov: Please move this to [your own `.gitignore`](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore).

mpdude: Sorry

mpdude:
Squashed and 🤞🏻also for CS builds.

We should get this finalized as our (comments/LOC changed) ratio approaches 1.

morozov (approved):
Looks good to me :+1: Unless @kimhemsoe has objections, I'll merge it later today.

kimhemsoe:
Looks good to me, great work.

About the spelling mistakes in variables.. those are mine :D Hope to fix them in dbal 3 ..

Ocramius (approved):
:+1:

morozov:
Awesome work @mpdude!

mpdude:
Please don’t forget to merge this back into the lowest possible 2.x release.

Ocramius:
@mpdude this is targeting 2.9, as it is an improvement

mpdude:
Ok, not going to argue about your definitions ;-)

Just wanted to mention it because the PR is against master and not a 2.x branch.

morozov:
We expect all pull requests to target `master` which represents the next minor release (currently, `2.9`). Bug and security fixes get manually backported to `2.x` branches.

Issue #3140, opened 2018-05-11
Lost parameters: wrong ESCAPED_BRACKET_QUOTED_TEXT


riccardonar:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.5.7

#### Summary

This commit https://github.com/doctrine/dbal/commit/3a351369582dade60c750e2cef540eb7b568e6b3 change preg_match_all and not consider the shortest match so can lost some parameters

#### Current behavior

`Doctrine\DBAL\SQLParserUtils::getUnquotedStatementFragments` for this statement

``` php
$statement = "SELECT table.field1, ARRAY['3']::integer[]
FROM schema.table table
WHERE table.f1 = :some AND ARRAY['3']::integer[]";
```

return this array

``` php
Array
(
    [0] => Array
        (
            [0] => SELECT table.f1, ARRAY
            [1] => 0
        )

    [1] => Array
        (
            [0] => 3
            [1] => 24
        )

    [2] => Array
        (
            [0] => 3
            [1] => 97
        )

    [3] => Array
        (
            [0] => ]::integer
            [1] => 99
        )
)
```

without sql part with `table.f1 = :some` so parameter `some` is lost

#### How to reproduce

``` php
$statement = "SELECT table.field1, ARRAY['3']::integer[]
FROM schema.table table
WHERE table.f1 = :some AND ARRAY['3']::integer[]";

$fragments = \Doctrine\DBAL\SQLParserUtils::getUnquotedStatementFragments($statement);
print_r($fragments);
```

#### Expected behavior

from previous version, something like this:

``` php
Array
(
    [0] => Array
        (
            [0] => SELECT table.f1, ARRAY
            [1] => 0
        )

    [1] => Array
        (
            [0] => ::integer
            [1] => 27
        )

    [2] => Array
        (
            [0] => 
FROM schema.table table
WHERE table.f1 = :some AND ARRAY
            [1] => 38
        )

    [3] => Array
        (
            [0] => ::integer
            [1] => 100
        )
)
```

ohvitorino:
I'm facing the same problem.

I'm upgrading an application that used doctrine/dbal version 2.3.5 to 2.5.14.

I managed to figure out that the problem might lie on the method: `\Doctrine\DBAL\SQLParserUtils::getUnquotedStatementFragments`. The regular expression used to split the query in fragments has change since 2.3.5. This causes, as far as I can see, the query fragments to be incorrectly determined.

[...]

PR #3226, opened 2018-07-23
Ignore case when matching ARRAY

ohvitorino:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

Named parameters get lost because the regular expression used to determine the query fragments only match the keyword `ARRAY` (capital letters). This should be made case insensitive since this is also correct syntax.


Ocramius:
Is this fixing #3140?

Ocramius:
@ohvitorino please target DBAL `master`: we will arrange backporting ourselves after having evaluated/merged the patch.

ohvitorino:
@Ocramius I changed the target to master. And unfortunately this doesn't seem to fix #3140.

Ocramius (dismissed):
LGTM :+1:

morozov (dismissed):

morozov:
@ohvitorino please fix the code style issues before I can merge it.

ohvitorino:
@morozov the php code sniffer configuration being used expects, for example, short array syntax. But the whole file uses the `array()` syntax, should I then change the whole file or only the lines I changed in my PR? Changing only a couple of statements to comply with the expected code style seems a bit inconsistent.

ohvitorino:
@morozov Code styles are fixed 🙂

ohvitorino:
@morozov done and building

morozov (approved):

morozov:
Thank you, @ohvitorino! Merged.

morozov:
Backported to `2.8` via 7d9f7d0ed7ef9efed64ce2e3f7f7697fc30860d0.

PR #3231, opened 2018-07-26
fix type : simpleArrayType, empty set will make mistake

zYeee:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

If mysql type was **SET**, and mapping to the doctrine entity will be **simpleArray**, and if the set was empty, the value passed to the **convertToPHPValue** was '', the origin code would return [''], the correct result should be [].

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Types/SimpleArrayTest.php
+++ b/tests/Doctrine/Tests/DBAL/Types/SimpleArrayTest.php
@@ -15,26 +15,21 @@ class SimpleArrayTest extends \Doctrine\Tests\DbalTestCase
     /**
      * @var AbstractPlatform
      */
-    protected $_platform;
+    private $platform;
 
     /**
      * @var Type
      */
-    protected $_type;
+    private $type;
 
     protected function setUp()
     {
-        $this->_platform = new MockPlatform();
-        $this->_type = Type::getType('simple_array');
-    }
-
-    protected function tearDown()
-    {
-        error_reporting(-1); // reactive all error levels
+        $this->platform = new MockPlatform();
+        $this->type = Type::getType('simple_array');
     }
 
     public function testNullConversion()
     {
-        self::assertEquals([], $this->_type->convertToPHPValue("", $this->_platform));
+        self::assertEquals([], $this->type->convertToPHPValue("", $this->platform));
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Types/SimpleArrayTest.php
+++ b/tests/Doctrine/Tests/DBAL/Types/SimpleArrayTest.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Types;
+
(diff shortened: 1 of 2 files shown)


Ocramius:
Integration and unit test are very much required here

zYeee:
added the test file, is it right?

tests/Doctrine/Tests/DBAL/Types/SimpleArrayTest.php
```diff
@@ -0,0 +1,40 @@
...
+
+    protected function setUp()
+    {
+        $this->_platform = new MockPlatform();
+        $this->_type = Type::getType('simple_array');
+    }
+
+    protected function tearDown()
+    {
+        error_reporting(-1); // reactive all error levels
```
Ocramius: Shouldn't be used

tests/Doctrine/Tests/DBAL/Types/SimpleArrayTest.php
```diff
@@ -0,0 +1,40 @@
...
+use const E_STRICT;
+use function error_reporting;
+use function serialize;
+
+class SimpleArrayTest extends \Doctrine\Tests\DbalTestCase
+{
+    /**
+     * @var AbstractPlatform
+     */
+    protected $_platform;
```
Ocramius: `private` by default please

Ocramius: The underscore naming should also be used (our CS checks will likely catch this)

tests/Doctrine/Tests/DBAL/Types/SimpleArrayTest.php
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * @var Type
+     */
+    protected $_type;
+
+    protected function setUp()
+    {
+        $this->_platform = new MockPlatform();
+        $this->_type = Type::getType('simple_array');
```
Ocramius: Should probably just instantiate the simple array type

Ocramius (changes requested):
Test looks good (besides comments), but we also need a scenario in which a `simple_array` manages to generate a `''` value in MySQL and converts it back.

Ocramius:
To be clear, the added scenario would be a larger integration test with a connected DB

zYeee:
Ok, Thanks for your reply, when a column with type "set" in mysql, and this column was empty,  **simply array** would generate a **""** value, I can make a integration test with a connected DB via my local env, but I don't know how to add this scenario here :(

Ocramius:
See https://github.com/doctrine/dbal/tree/54054275689c9b9759bd607f6d0cf63c77d52375/tests/Doctrine/Tests/DBAL/Functional/Ticket for examples

Also, I think the `simple_array` type was thought to be used with `VARCHAR`, not with `SET`, so this patch may be invalid if that's the case.

PR #3232, opened 2018-07-28
Removed NOW() from QueryBuilder usage examples

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

The `QueryBuilder` API documentation contains usages of the `NOW()` function. It's not the best idea for the following reasons:
1. This function is non-portable. Therefore, users who are not enough familiar with the supported platforms may be tricked into using this function in production code as recommended.
2. It promotes sourcing dates from the DB platform which may lead to data inconsistency between app-generated and DB-generated dates.


Ocramius (approved):
LGTM 👍

PR #3235, opened 2018-08-02
Use PSR-4 autoloader

Majkl578:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Use PSR-4 Composer autoloader (**only**). This does NOT change the directory structure in any way. We use PSR-4 autoloader in most Doctrine libraries already.

Related: https://github.com/doctrine/doctrine2/pull/1451 + #621

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,10 +34,10 @@
         "sort-packages": true
     },
     "autoload": {
-        "psr-0": { "Doctrine\\DBAL\\": "lib/" }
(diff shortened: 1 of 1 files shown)


Ocramius:
Woah, didn't even realise that we are still on PSR-0 here: good catch!

Ocramius (approved):

Issue #3237, opened 2018-08-03
Regression in ImportCommand


morozov:
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.0.0-dev

On `develop`, the following [issue](https://travis-ci.org/doctrine/dbal/jobs/411436518#L511) is reported by PHPStan:
```
------ ----------------------------------------------------------- 
 Line   lib/Doctrine/DBAL/Tools/Console/Command/ImportCommand.php  
------ ----------------------------------------------------------- 
 95     Call to an undefined method                                
        Doctrine\DBAL\Driver\PDOStatement::nextRowset().           
------ ----------------------------------------------------------- 
```
The issue looks legit. In order to fix the PHPStan job on Travis, I'm going to white-list this issue but it will have to be fixed properly before the release.

morozov:
Anyone knows how `doctrine-dbal dbal:import` is supposed to work? According to the help message, it will "Import SQL file(s) directly to Database", however judging by the implementation, it cannot handle 
files which have more than one statements. Here's what I'm trying to do:

1. ```bash
   mysql << 'EOF'
   CREATE TABLE `test` (
     `id` int NOT NULL,
     PRIMARY KEY (`id`)
   )
   EOF
   ```
2. ```bash
   cat > data.sql << 'EOF'
   INSERT INTO test (`id`) VALUES (1);
   INSERT INTO test (`id`) VALUES (2);
   EOF
   ```
3. ```bash
   bin/doctrine-dbal dbal:import data.sql
   ```

The above results into:

> An exception occurred while executing 'INSERT INTO test (`id`) VALUES (1);                   
>  INSERT INTO test (`id`) VALUES (2);                                                          
>  ':                                                                                           
>
>  You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'INSERT INTO test (`id`) VALUES (2)' at line 2

Other import implementations I saw in PHP would parse the file into multiple statements and execute them one by one but this just passes the file contents to the driver as is.

/cc @Ocramius @Majkl578

Ocramius:
Never used it, also unaware of its purpose 😰

morozov:
Linking existing related issue https://github.com/doctrine/dbal/issues/2700.

morozov:
I spent some more time researching and figured out two more things:

[...]

Issue #3236, opened 2018-08-03
Regression introduced by changes in Schema\Table


morozov:
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.0.0-dev

On `develop`, the following [issue](https://travis-ci.org/doctrine/dbal/jobs/411436518#L504) is reported by PHPStan:
```
------ ----------------------------------------------------------- 
 Line   lib/Doctrine/DBAL/Schema/Table.php                         
------ ----------------------------------------------------------- 
 909    Call to an undefined method                                
        Doctrine\DBAL\Schema\UniqueConstraint::getLocalColumns().  
------ ----------------------------------------------------------- 
```
The issue looks legit. In order to fix the PHPStan job on Travis, I'm going to white-list this issue but it will have to be fixed properly before the release.

It looks like it was caused by this commit https://github.com/doctrine/dbal/commit/2f4809aa1557ef8361d1ea7eed9fc99c970b66db

morozov:
Fixed by #3519.

PR #3238, opened 2018-08-03
Fixed and suppressed issues found by PHPStan on develop

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

As of #3025, the `develop` builds [fail](https://travis-ci.org/doctrine/dbal/jobs/411436518) due to the issues which didn't exist and weren't fixed in `master`.

#3236 and #3237 have been separately filed as release blockers for `3.0` since fixing them requires a deeper understanding of the underlying issues than I currently have.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -24,7 +24,7 @@ parameters:
         - '~^Method Doctrine\\DBAL\\Driver\\SQLSrv\\SQLSrvConnection::errorCode\(\) should return string\|null but returns false\.\z~'
 
         # http://php.net/manual/en/pdo.sqlitecreatefunction.php
-        - '~^Call to an undefined method Doctrine\\DBAL\\Driver\\PDOConnection::sqliteCreateFunction\(\)\.\z~'
+        - '~^Call to an undefined method PDO::sqliteCreateFunction\(\)\.\z~'
 
         # legacy variadic-like signature
         - '~^Method Doctrine\\DBAL\\Driver\\Connection::query\(\) invoked with \d+ parameters?, 0 required\.\z~'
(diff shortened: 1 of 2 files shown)


phpstan.neon.dist
```diff
@@ -24,7 +24,7 @@ parameters:
         - '~^Method Doctrine\\DBAL\\Driver\\SQLSrv\\SQLSrvConnection::errorCode\(\) should return string\|null but returns false\.\z~'
 
         # http://php.net/manual/en/pdo.sqlitecreatefunction.php
-        - '~^Call to an undefined method Doctrine\\DBAL\\Driver\\PDOConnection::sqliteCreateFunction\(\)\.\z~'
+        - '~^Call to an undefined method PDO::sqliteCreateFunction\(\)\.\z~'
```
morozov: @Majkl578 while I understand the PHPStan's inability to use stubbed `PDO::SQLSRV_*` constants, what's the reason we should white-list this method? As long as `pdo_sqlite` is loaded on Travis, shouldn't PHPStan be able to see it via the extension reflection?

Majkl578: > As long as pdo_sqlite is loaded

Exactly, as long as it is loaded. And pdo_sqlite is not direct dependency in composer.json, so you may not have it enabled locally. And polyfills can't add methods/constants to existing classes.
Maybe we should have two phpstan configs with specific ignores for Travis.

Majkl578 (approved):

phpstan.neon.dist
Majkl578: This is als why I disabled `reportUnmatchedIgnoredErrors`, since different people/environments have different extensions.

Ocramius:
:+1:

phpstan.neon.dist
morozov: Gotcha. I thought this issue was reproducible despite whether `pdo_sqlite` was loaded or not.

The necessity to use `reportUnmatchedIgnoredErrors()` across the board and the fact that PHP extensions _can_ add members to existing classes (which is impossible in PHP) looks like a ground for PHPStan to implement this "extensibility" feature for stubs as well. I.e. it should initialize from the reflection extension and merge the stubbed symbols in. Does it sound like a reasonable feature request?

Issue #3152, opened 2018-05-14
When building queries, DBAL doesn't consider reserved keywords


morozov:
### Bug Report

When building queries, DBAL doesn't consider reserved keywords

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.7.1

#### Summary

Assuming `'FILE'` is a reserved keyword on the Oracle platform, I can create a table with the `'file'` column using the schema manager but I can't execute an `INSERT` or `UPDATE` statement using the `Connection` API. I also cannot `SELECT` from the table using `QueryBuilder`.

#### How to reproduce

```php
$conn = \Doctrine\DBAL\DriverManager::getConnection([
    'driver' => 'oci8',
    'host' => 'localhost',
    'user' => 'system',
    'password' => 'oracle',
    'dbname' => 'XE',
]);

$table = new Doctrine\DBAL\Schema\Table('test_table');
$table->addColumn('file', 'string');

$conn->getSchemaManager()
    ->createTable($table);

$conn->insert('test_table', ['file' => 'foo']);

// ORA-00928: missing SELECT keyword

$conn->createQueryBuilder()
    ->select('file')
    ->from('test_table')
    ->execute()
    ->fetchAll();

// ORA-00936: missing expression
```

#### Expected behavior

The keywords are handled consistently by the Schema Manager and Query Builder.

morozov:
The same is reproducible with `'select'` as the column name on other platforms.

morozov:
@Ocramius we can use this issue as a starting point for discussion of whether/how identifiers should be quoted in the DBAL.

leoqbc:
Hi guys I'm facing the same problem in MYSQL, I have a field called 'desc':

```php

$conn->insert('my_table', [
    'desc' => 'Hello',
    ...
]);
```

The "desc" word throws a SyntaxException.

My solution was to put mysql quote as array parameter like this:

```php

$conn->insert('my_table', [
    '`desc`' => 'Hello', // `desc`
    ...
]);
```

May it can be a parcial solution, can we discuss a fix??

Best.

leoqbc:
Make sense, as I imagined, not easy to patch...

Well ```Connection::quoteIdentifier()``` worked fine for me.

Maybe it can be refactored in future, since It breaks SOLID's SRP IMHO, but I dont know the complete API.

Thanks @morozov and good luck on this.

morozov:
[...]

PR #3241, opened 2018-08-07
Removed dbal:import CLI command

morozov:
|      Q       |   A
|------------- | -----------
| BC Break     | yes
| Fixed issues | #3237

The details are in https://github.com/doctrine/dbal/issues/3237#issuecomment-410923341.


UPGRADE.md
```diff
@@ -1,5 +1,10 @@
 # Upgrade to 3.0
 
+## BC BREAK: Removed dbal:import CLI command
+
+The `dbal:import` CLI command has been removed since it only worked with PDO-based drivers by relying on a non-documented behavior of the extension, and it was impossible to make it work with other drivers.
+Please use other database client applications for import, e.g. `mysql < data.sql`.
```
Majkl578: I'm going to be annoying and say that _including only MySQL looks biased_. :P Can we add alternatives that are also obvious?
```md

Please use other database client applications for import,
e.g. `mysql < data.sql`, `psql < data.sql`, `cat data.sql | sqlite3 my.db` or equivalent.
```

morozov: I'll have to figure out the ways to import SQL dumps on all supported platforms, but in general, the proposal is legit. If you can add some more examples from the top of your head, please do.

BTW, isn't `cat data.sql | sqlite3 my.db` same as `sqlite3 my.db < data.sql` but with an extra process for `cat`?

Majkl578: I don't think we need _all_ supported platforms, MySQL, MariaDB, PostgreSQL and SQLite is enough imho.

> isn't `cat data.sql | sqlite3 my.db` same as `sqlite3 my.db < data.sql`

Likely yes. :)
```
$ sqlite3 <<<'select 123'
123
```

morozov: > I don't think we need _all_ supported platforms, MySQL, MariaDB, PostgreSQL and SQLite is enough imho.

Makes sense. I'll update the PR later.

> ```
> $ sqlite3 <<<'select 123'
> 123
> ```

Speaking of which, isn't it a proper replacement for `dbal:run-sql`? Yes, it should works across the platforms but in the entire `Tools` namespace we have only two commands remaining: `dbal:import`, `dbal:run-sql` and `dbal:reserved-words`. The latter, based on some recently discovered issues (e.g. https://github.com/doctrine/dbal/issues/3152) may become obsolete in 3.0 since on the one hand, keywords cannot be fully supported in DBAL, on the other ORM should just quote everything. The notion of which names are reserved shouldn't be relevant for an app developer.

Majkl578 (approved):

PR #3243, opened 2018-08-08
Fix getTruncateTableSQL for Sqlite platform

raziel057:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes

#### Summary

Reset sequence when calling getTruncateTableSQL as if it was a real truncate.

Currently if you have a table with auto-increment primary key, the sequence is not reset when using the ORMPurger with ``PURGE_MODE_TRUNCATE`` for sqlite databases.

The behaviour was different in ``doctrine/dbal: 2.7.2`` because the keys where reset when calling only DELETE (as the field was not created with autoincrement). This have been changed by the following commit (which is BC BREAK) https://github.com/doctrine/dbal/commit/51358603bba352e46d56310125d68430342ed26a#diff-3fa09ba089913d40a1fbeb10b8cb8536

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
@@ -558,7 +558,7 @@ class SqlitePlatform extends AbstractPlatform
         $tableIdentifier = new Identifier($tableName);
         $tableName = str_replace('.', '__', $tableIdentifier->getQuotedName($this));
 
-        return 'DELETE FROM ' . $tableName;
+        return 'DELETE FROM ' . $tableName . '; UPDATE SQLITE_SEQUENCE SET SEQ=0 WHERE NAME="' . $tableName . '";';
     }
 
     /**


morozov:
@raziel057 please add a functional test which validates the fix.

raziel057:
@morozov In fact this PR produce an error when all tables in the database doesn't contains an autoincrement PK.

In this case we have:

``SQLSTATE[HY000]: General error: 1 no such table: SQLITE_SEQUENCE``

So I will close this PR and open a new one with the current failing test because I don't know how to fix it as I can't check the existance of this ``SQLITE_SEQUENCE`` in ``SqlitePlatform`` (we don't have the entity manager).

Maybe the best solution would be to revert the following 5135860 because autoincrement is not recommended (https://www.sqlite.org/autoinc.html) on SQLITE and by default the rowid automatically increments if not specified. Currently when you run a test suite (generally with sqlite) it fails because rowid is not reset between each unit tests when using the ORMPurger. So it's a big issue.

PR #3244, opened 2018-08-09
Deprecated dbal:import CLI command

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no
| Fixed issues | #3237

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,14 @@
+# Upgrade to 2.9
+
+## Deprecated dbal:import CLI command
+
+The `dbal:import` CLI command has been deprecated since it only works with PDO-based drivers by relying on a non-documented behavior of the extension, and it's impossible to make it work with other drivers.
+Please use other database client applications for import, e.g.:
+
+ * For MySQL and MariaDB: `mysql [dbname] < data.sql`.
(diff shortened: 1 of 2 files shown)


Ocramius (approved):
:+1:

lcobucci:
@Ocramius @morozov in some places we're using `@deprecated` + `@trigger_error()` in others just `@deprecated` to handle deprecations, which way is the one to go?

Ocramius:
To be honest, I don't know. The entire situation with requiring an additional component to run tests with deprecated bits of code is extremely confusing, IMO, so the silent deprecation approach isn't any better eithe.

lcobucci:
@Ocramius I agree, I really dislike this whole phpunit-bridge stuff and I think that https://github.com/phpstan/phpstan-deprecation-rules is the appropriated tool to check for deprecations. 

This means that I believe that we could make a call and standardise things in the repos.... shall we move this discussion to gitter?

Ocramius:
I'd be on board with using @phpstan's approach: simpler and checks only the sources, which is exactly what we want.

Yes, that means that many users won't see anything until we remove a feature, but the silenced operation is just a mess.

morozov:
@lcobucci, @Ocramius, I'm with you.

PR #3245, opened 2018-08-09
Add Failing test using getTruncateTableSQL

raziel057:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

The test ``testTruncatePrimaryKeyAutoIncrement`` is currently failing with SQLITE because TRUNCATE doesn't exists with this platform. Using DELETE in replacement of TRUNCATE doesn't reset the PK with autoincrement. This issue happens on 2.8.x because of the merge of the following PR https://github.com/doctrine/dbal/commit/51358603bba352e46d56310125d68430342ed26a#diff-3fa09ba089913d40a1fbeb10b8cb8536

Before, all PK were created without autoincrement (by default rowid are automatically incremented in SQLITE if not specified explicitely when inserting data). On delete the rowid was reset so it was OK for this use case.

I'm not sure how to fix this issue because the best would be to reset the linked sequence, a bit like in this PR https://github.com/doctrine/dbal/pull/3243 but currently there is another side effect.

Maybe the best solution would be to revert the following 5135860 because autoincrement is not recommended (https://www.sqlite.org/autoinc.html) on SQLITE and by default the rowid automatically increments if not specified. Currently when you run a test suite (generally with sqlite) it fails because rowid is not reset between each unit tests when using the ORMPurger. So it's a big issue.

### DIFF
--- a/.github/ISSUE_TEMPLATE/BC_Break.md
+++ b/.github/ISSUE_TEMPLATE/BC_Break.md
@@ -1,6 +1,6 @@
 ---
 name: 💥 BC Break
-about: Have you encountered an issue during upgrade? 💣
+about: Have you encountered an issue during an upgrade? 💣
 ---
 
 <!--
@@ -18,15 +18,15 @@ Before reporting a BC break, please consult the upgrading document to make sure 
 
 #### Summary
 
-<!-- Provide a summary desciribing the problem you are experiencing. -->
+<!-- Provide a summary describing the problem you are experiencing. -->
 
-#### Previous behavior
+#### Previous behaviour
 
-<!-- What was the previous (working) behavior? -->
+<!-- What was the previous (working) behaviour? -->
 
 #### Current behavior
 
-<!-- What is the current (broken) behavior? -->
+<!-- What is the current (broken) behaviour? -->
 
 #### How to reproduce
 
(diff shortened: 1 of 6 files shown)


morozov:
@raziel057 thank you for the update. Please retarget the pull request against `master`. We'll take care of the backport once the issue is resolved.

raziel057:
@morozov Done.

morozov:
@raziel057, thank you. Please make sure there are no conflicting changes in the pull request (`git rebase HEAD~2 --onto upstream/master` should do the job).

morozov:
@TimoBakx any idea on hot this can get fixed?

TimoBakx:
@morozov Would it be possible to execute this for sqlite when truncating?

`delete from [table];    
delete from sqlite_sequence where name='[table]';`

TimoBakx:
A check for the existence of the `sqlite_sequence` table could be done by executing this:
`SELECT name FROM sqlite_master WHERE type='table' AND name='sqlite_sequence';`

TimoBakx:
Ah, but we can't execute any queries at that point. And there's no way to put those two in 1 optional query. And I can't find any way to see if the given tableName would have an auto increment column or not.

I'm afraid that that leaves no solution that works in both cases (have a proper truncate AND no re-used auto_increment ID when deleting the last row in a table).

TimoBakx:
Sorry for the problems my previous change has caused. I did not realise that it would have these consequences.

PR #3246, opened 2018-08-09
Fix url resolution for master/slave connections

deguif:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

When using url option for configuring master/slave connection, the url is never parsed and the connection parameters are not filled as they should.


deguif:
Just see that it was already proposed by @stof in #3201

Majkl578:
Duplicate of #3201

Majkl578:
Closing as per @deguif's comment above.

PR #3247, opened 2018-08-10
ResultStatement::fetchColumn() can return null if the result set is empty

ondrejmirtes:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

For PHPStan to stop complaining when the result is compared against `null`, I added `null` to the return typehint of this method. It can return null when the result set is empty.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/ResultStatement.php
+++ b/lib/Doctrine/DBAL/Driver/ResultStatement.php
@@ -109,7 +109,7 @@ interface ResultStatement extends \Traversable
      *                         If no value is supplied, PDOStatement->fetchColumn()
      *                         fetches the first column.
      *
-     * @return string|boolean A single column in the next row of a result set, or FALSE if there are no more rows.
+     * @return string|boolean|null A single column in the next row of a result set, or FALSE if there are no more rows.
      */
     public function fetchColumn($columnIndex = 0);
 }


lib/Doctrine/DBAL/Driver/ResultStatement.php
```diff
@@ -109,7 +109,7 @@ public function fetchAll($fetchMode = null, $fetchArgument = null, $ctorArgs = n
      *                         If no value is supplied, PDOStatement->fetchColumn()
      *                         fetches the first column.
      *
-     * @return string|boolean A single column in the next row of a result set, or FALSE if there are no more rows.
+     * @return string|boolean|null A single column in the next row of a result set, or FALSE if there are no more rows.
```
Ocramius: When can `null` be returned? Can we have a scenario where this is tested?

Ocramius (changes requested):

lib/Doctrine/DBAL/Driver/ResultStatement.php
ondrejmirtes: When you have an aggregate query (like `SELECT AVG`) and it does not match any rows. It might be MySQL-specific.

I'm not familiar with DBAL tests so not sure how to add it.

Ocramius: Probably somewhere like in https://github.com/doctrine/dbal/blob/66daecd0084118ec04c54288915a8eb9bac7511d/tests/Doctrine/Tests/DBAL/Functional/PDOStatementTest.php (or the `Functional` dir - didn't pin-point it to a specific test)

ondrejmirtes: OK, I'll look into it 😊

morozov: Here's a simple example:
```php
$value = $conn->executeQuery(
    $conn->getDatabasePlatform()
        ->getDummySelectSQL('NULL')
)->fetchColumn();

var_dump($value);
// NULL
```

morozov:
@Majkl578 why is it a "Missing Tests"?

Ocramius:
yes, I'd need the `null` being returned scenario to be added, so we don't introduce a regression when/if we get rid of that scenario implicitly.

morozov:
This looks outdated: https://github.com/doctrine/dbal/blob/3b3c0004f42a145098114da43c8f76a866cd3a56/src/Driver/Result.php#L30-L37

PR #3248, opened 2018-08-11
Made the OFFSET in LIMIT queries non-nullable integer defaulting to 0

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | none

Semantically, the absence of the offset in a `LIMIT`ed query (`$offset = null`) means "start from the beginning", i.e. with the zero offset. Therefore, `$offset = null` and `$offset = 0` are semantically equivalent.

From the practical standpoint, allowing `$offset = null` means that we have to cast it to an `(int)` or consider its nullability during strict comparison.

This patch makes the offset non-nullable defaulting to `0`. The usage of the `NULL` offset in DBAL 2.x will be deprecated in favor of using `0`.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
@@ -692,10 +692,10 @@ class SqlitePlatform extends AbstractPlatform
     /**
      * {@inheritDoc}
      */
-    protected function doModifyLimitQuery($query, $limit, $offset)
+    protected function doModifyLimitQuery(string $query, ?int $limit, int $offset) : string
     {
         if ($limit === null && $offset > 0) {
-            return $query . ' LIMIT -1 OFFSET ' . $offset;
+            $limit = -1;
         }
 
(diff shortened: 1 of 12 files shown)


lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -692,10 +692,10 @@ protected function getPostAlterTableIndexForeignKeySQL(TableDiff $diff)
     /**
      * {@inheritDoc}
      */
-    protected function doModifyLimitQuery($query, $limit, $offset)
+    protected function doModifyLimitQuery(string $query, ?int $limit, int $offset) : string
     {
         if ($limit === null && $offset > 0) {
-            return $query . ' LIMIT -1 OFFSET ' . $offset;
+            $limit = -1;
```
morozov: Not strictly related to the issue, just a code duplication cleanup.

Majkl578 (approved):
LGTM, makes sense.

PR #3252, opened 2018-08-16
Replaced call_user_func_array() of a fixed method with the usage of variadic arguments

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

The regular `$object->method(...$arguments)` syntax is cleaner and more friendly to SA.


Ocramius (approved):
LGTM

PR #3253, opened 2018-08-16
Deprecated usage of the NULL offset in LIMITed queries

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no
| Fixed issues | none

See #3248 for more details.


Ocramius (approved):
:+1:

PR #3254, opened 2018-08-17
Throw ConversionException when unserialization fail for array and object types

seferov:
|      Q       |   A
|------------- | -----------
| BC Break     | no

#### Summary

**Problem:** Converting invalid data with ArrayType/ObjectType throws PHP notice error.

**Solution:** inside convert method unserialize notice error can be suppressed since a proper exception (ConversionException) will be thrown right after that. No need to deal with error reporting.

### DIFF
--- a/lib/Doctrine/DBAL/Types/ConversionException.php
+++ b/lib/Doctrine/DBAL/Types/ConversionException.php
@@ -120,4 +120,13 @@ class ConversionException extends \Doctrine\DBAL\DBALException
             $error
         ));
     }
+
+    public static function conversionFailedUnserialization(string $format, string $error) : self
+    {
+        return new self(sprintf(
+            "Could not convert database value to '%s' as an error was triggered by the unserialization: '%s'",
+            $format,
+            $error
+        ));
+    }
 }
--- a/lib/Doctrine/DBAL/Types/ObjectType.php
+++ b/lib/Doctrine/DBAL/Types/ObjectType.php
@@ -21,7 +21,9 @@ namespace Doctrine\DBAL\Types;
 
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use function is_resource;
+use function restore_error_handler;
 use function serialize;
+use function set_error_handler;
 use function stream_get_contents;
 use function unserialize;
 
@@ -58,12 +60,16 @@ class ObjectType extends Type
         }
 
         $value = (is_resource($value)) ? stream_get_contents($value) : $value;
-        $val = unserialize($value);
-        if ($val === false && $value !== 'b:0;') {
-            throw ConversionException::conversionFailed($value, $this->getName());
-        }
 
-        return $val;
+        set_error_handler(function (int $code, string $message) : void {
+            throw ConversionException::conversionFailedUnserialization($this->getName(), $message);
+        });
+
+        try {
+            return unserialize($value);
+        } finally {
+            restore_error_handler();
+        }
     }
 
     /**
--- a/tests/Doctrine/Tests/DBAL/Types/ArrayTest.php
+++ b/tests/Doctrine/Tests/DBAL/Types/ArrayTest.php
@@ -5,9 +5,6 @@ namespace Doctrine\Tests\DBAL\Types;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\Tests\DBAL\Mocks\MockPlatform;
-use const E_ALL;
-use const E_STRICT;
-use function error_reporting;
 use function serialize;
 
 class ArrayTest extends \Doctrine\Tests\DbalTestCase
@@ -28,12 +25,6 @@ class ArrayTest extends \Doctrine\Tests\DbalTestCase
         $this->_type = Type::getType('array');
     }
 
-    protected function tearDown()
-    {
-        error_reporting(-1); // reactive all error levels
-    }
-
-
     public function testArrayConvertsToDatabaseValue()
     {
         self::assertInternalType(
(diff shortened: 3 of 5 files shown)


tests/Doctrine/Tests/DBAL/Types/ArrayTest.php
```diff
@@ -52,7 +46,6 @@ public function testArrayConvertsToPHPValue()
 
     public function testConversionFailure()
     {
-        error_reporting( (E_ALL | E_STRICT) - \E_NOTICE );
```
guilliamxavier: Then you should probably also delete the now-obsolete imports at the top of the file:

```php
use const E_ALL;
use const E_STRICT;
use function error_reporting;
```

Also, there's the same logic in file `tests/Doctrine/Tests/DBAL/Types/ObjectTest.php` (from the same original commit 2cc1cdc56d7dc09445ca935ce59b5a2755699302)

seferov: @guilliamxavier you are right. I will make the changes you suggested. However I am not sure what the maintainers think about this PR.

morozov (changes requested):
@seferov does the error message contain the reason of the conversion failure or just states the fact? It looks like the latter.

I don't like the idea of error suppression since it leads to poor developer experience. Instead, we could try using a temporary error handler to capture the error and pass it to the exception message. Similarly to `ConversionException::conversionFailedSerialization()`, we could create `ConversionException::conversionFailedUnserialization()`.

lib/Doctrine/DBAL/Types/ObjectType.php
```diff
@@ -58,10 +58,13 @@ public function convertToPHPValue($value, AbstractPlatform $platform)
...
-        $val   = @unserialize($value);
-        if ($val === false && $value !== 'b:0;') {
-            throw ConversionException::conversionFailed($value, $this->getName());
-        }
+
+        set_error_handler(function ($errno, $errstr) use ($value) {
+            throw ConversionException::conversionFailedUnserialization($value, $this->getName(), $errstr);
+        });
+
+        $val = unserialize($value);
```
morozov: In the case of an exception, the error handler should be restored as well. Please use `try/finally`.

lib/Doctrine/DBAL/Types/ConversionException.php
```diff
@@ -120,4 +120,16 @@ public static function conversionFailedSerialization($value, $format, $error)
             $error
         ));
     }
+
+    public static function conversionFailedUnserialization($value, $format, $error)
+    {
+        $actualType = is_object($value) ? get_class($value) : gettype($value);
+
+        return new self(sprintf(
+            "Could not convert database value %s to %s, as an '%s' error was triggered by the unserialization",
```
morozov: Please use quotes around `%s` for consistency with existing error messages.

lib/Doctrine/DBAL/Types/ObjectType.php
```diff
@@ -58,9 +60,15 @@ public function convertToPHPValue($value, AbstractPlatform $platform)
         }
 
         $value = (is_resource($value)) ? stream_get_contents($value) : $value;
-        $val = unserialize($value);
-        if ($val === false && $value !== 'b:0;') {
-            throw ConversionException::conversionFailed($value, $this->getName());
+
+        set_error_handler(function (int $errno, string $errstr) use ($value) {
```
morozov: The last nitpick, please add `: void` to the fn signature and squash all commits. The test looks fine.

lib/Doctrine/DBAL/Types/ConversionException.php
```diff
@@ -120,4 +120,16 @@ public static function conversionFailedSerialization($value, $format, $error)
             $error
         ));
     }
+
+    public static function conversionFailedUnserialization($value, string $format, string $error) : self
+    {
+        $actualType = is_object($value) ? get_class($value) : gettype($value);
+
+        return new self(sprintf(
+            "Could not convert database value '%s' to '%s', as an '%s' error was triggered by the unserialization",
```
Majkl578: Since `$value` will very likely be always `string` (unlike the _serialize_ case), I think it doesn't add any value to the error message.

I'd suggest something like this:
```php

"Could not convert database value to '%s' as an error was triggered by the unserialization:  '%s'"
```

lib/Doctrine/DBAL/Types/ObjectType.php
Majkl578: Also `$code` instead of `$errno` and `$message` instead of `$errstr` would be more readable. :yum:

lib/Doctrine/DBAL/Types/ObjectType.php
```diff
@@ -58,9 +60,15 @@ public function convertToPHPValue($value, AbstractPlatform $platform)
...
+            throw ConversionException::conversionFailedUnserialization($value, $this->getName(), $errstr);
+        });
+
+        try {
+            $val = unserialize($value);
```
Majkl578: No need for `$val` variable, you should use `return unserialize(...)` directly here.

tests/Doctrine/Tests/DBAL/Types/ArrayTest.php
```diff
@@ -52,8 +43,8 @@ public function testArrayConvertsToPHPValue()
...
         $this->expectException('Doctrine\DBAL\Types\ConversionException');
+        $this->expectExceptionMessage('Could not convert database value \'string\' to \'array\', as an \'unserialize(): Error at offset 0 of 7 bytes\' error was triggered by the unserialization');
```
Majkl578: You can avoid cryptic escaping by using double qoutes for the message.

tests/Doctrine/Tests/DBAL/Types/ObjectTest.php
```diff
@@ -44,8 +36,8 @@ public function testObjectConvertsToPHPValue()
 
     public function testConversionFailure()
     {
-        error_reporting( (E_ALL | E_STRICT) - \E_NOTICE );
         $this->expectException('Doctrine\DBAL\Types\ConversionException');
+        $this->expectExceptionMessage('Could not convert database value \'string\' to \'object\', as an \'unserialize(): Error at offset 0 of 7 bytes\' error was triggered by the unserialization');
```
Majkl578: You can avoid cryptic escaping by using double qoutes for the message.

Majkl578:
Please also [squash changes](https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit) into one commit. (Or we can do it before merge.) Thanks. :100:

Majkl578 (approved):

morozov (approved):
Thanks @seferov! :shipit:

Issue #2841, opened 2017-09-03
Refactor type system


greg0ire:
From @Majkl578 on slack:

just a quick thought of a possible approach:

    introduce TypeInterface with required methods (or different name)
    make types implement it
    introduce some kind of Types enum-like empty class for implictly supported types, just like i.e. GeneratorType in ORM develop (edit: maybe this could stay in Type to lower BC implications and make Type an enum class in 3.0)
    introduce non-static TypeRegistry with no default types hardcoded in it - this would allow different connections/platforms use different type mappings (as @Ocramius suggested in https://symfony-devs.slack.com/archives/C3FQPE6LE/p1504362019000073)
    make TypeRegistry work with type instances instead of their class names (sounds good, but unsure about consequences)
    migrate Type:: to Types enum, keep BC by aliasing them and deprecate in 2,7
    migrate Type:: mechanics to TypeRegistry, keep BC by using private static TypeRegistry internally in 2.7 (some kittens may die during this process)
    in 3.0 (develop) drop these BC hacks (so no more kittens die)

greg0ire:
> use FQCNs instead of strings to identify types

In light of the need to have so-called parameterized types, this seems wrong. If we are going that way, the type registry should look more like a PSR-11 container (which it does already). FQCNs could still be used to identify types _by default_, but one should be able to have a type called `Doctrine\DBAL\Types\Enum.Country`.

Ocramius:
TBH, just register `My\Enum\Type\Country`, or `My\Enum\Type<My\Domain\Thing\Country>`.

EDIT: to be clear, being able to have a `class-string<DBALType>` would already help some validation a lot, especially in attributes.

greg0ire:
Oh yes, that second one sounds like a great idea, it would make it easier for static analysis tools to get make sense of the output of the type registry! So `Doctrine\DBAL\Types\Enum<MyProject\Country>`

Ocramius:
Right, so:

```php
/** @template-extends SomeEnum<Country> */
class MyCountryEnum extends SomeEnum
{
    // ...
}
```

therefore you would register it as `MyCountryEnum::class` :D

greg0ire:
Ah sorry, I should have made it a bit more obvious that I was referring to PHP 8 enums

It would look more like this:

```php
<?php
enum MyProject\Country
{
    // …
}

/**
 * @template E of enum-string
 */
class Doctrine\DBAL\Types\Enum
{
    /** @param E $enum */
[...]

PR #3255, opened 2018-08-17
Remove Doctrine\DBAL\Types\Type::getDefaultLength()

Majkl578:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

#### Summary

Not used by DBAL itself, only overridden by StringType. Some other types (DateIntervalType) hardcode limits in `getSQLDeclaration()` directly.

1st step towards #2841 (_Refactoring type system_).

### DIFF
--- a/lib/Doctrine/DBAL/Types/StringType.php
+++ b/lib/Doctrine/DBAL/Types/StringType.php
@@ -22,14 +22,6 @@ class StringType extends Type
     /**
      * {@inheritdoc}
      */
-    public function getDefaultLength(AbstractPlatform $platform)
-    {
-        return $platform->getVarcharDefaultLength();
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function getName()
     {
         return Type::STRING;
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade to 3.0
 
+## BC BREAK: `Doctrine\DBAL\Types\Type::getDefaultLength()` removed
+
+The `Doctrine\DBAL\Types\Type::getDefaultLength()` method has been removed as it served no purpose.
+
 ## BC BREAK: `Doctrine\DBAL\Types\Type::__toString()` removed
 
 Relying on string representation was discouraged and has been removed.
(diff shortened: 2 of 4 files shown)


morozov (approved):
Makes sense.

lib/Doctrine/DBAL/Types/StringType.php
```diff
@@ -19,14 +19,6 @@ public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $pla
         return $platform->getVarcharTypeDeclarationSQL($fieldDeclaration);
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function getDefaultLength(AbstractPlatform $platform)
-    {
-        return $platform->getVarcharDefaultLength();
```
morozov: Speaking of which, why do we need to know the default type length on the platform level? E.g.:

https://github.com/doctrine/dbal/blob/caf55ee7df850234e21c257fdc63d9be00e02a98/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php#L298-L300

We're taking our hard-coded value and passing it explicitly to the field declaraion. Would it make sense to leave the length unspecified if the developer doesn't care? Otherwise, it leads to the pointless expctations like https://github.com/doctrine/dbal/pull/3133/commits/e2e0de250b078119dd19c3926bb8eaedc9501a4d#diff-21fd51266275832f478e50a5b1f22748L219.

Ocramius: Completely unspecified length is probably better: should be explicitly specified by developers anyway.

Ocramius (approved):

lib/Doctrine/DBAL/Types/StringType.php
Majkl578: How portalbe is that? From what I remember MySQL requires maximum length when declaring VARCHAR columns while PostgreSQL doesn't.

morozov: There are two points here:

1. If a platform doesn't require max length (PostgreSQL), DBAL shouldn't require max length either but it shouldn't provide any hard-coded length in DDL. It should omit it.
2. If a platform does require MySQL, DBAL should require it in the column definition but it shouldn't provide any hard-coded length in DDL. It should fail.

This way, if you use only one platform (PostgreSQL) and don't care about portability, you can omit the length and use DBAL w/o any artificial restrictions. Otherwise, you have to explicitly specify the length. It's not the DBAL's job to specify it for you.

PR #3257, opened 2018-08-17
Remove Doctrine\DBAL\Types\Type::__toString()

Majkl578:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

#### Summary

Relying on __toString() internally was already removed in #2835, it's not obsolete and untested API.

1st step towards #2841 (_Refactoring type system_).


morozov (approved):
Not sure if relying on FQCN is any better than relying on a made-up name but `__toString()` shouldn't bepart of the API in any event.

PR #3259, opened 2018-08-17
Update export ignores

Majkl578:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Export-ignore more files, use absolute paths.

Also dropped archive.exclude section from composer.json since [Composer respects .gitattributes](https://github.com/composer/composer/blob/158e1c95da02cc0b932de74f9a09a1c7b6cf654f/src/Composer/Package/Archiver/GitExcludeFilter.php) [automatically](https://github.com/composer/composer/blob/122e422682d961233cc5db8b2102cd98b049d4f9/src/Composer/Package/Archiver/ArchivableFilesFinder.php#L53).

Result:


.gitattributes
```diff
@@ -1,12 +1,15 @@
...
-build.xml export-ignore
-phpunit.xml.dist export-ignore
-run-all.sh export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.appveyor.yml export-ignore
+/.travis.yml export-ignore
+/.scrutinizer.yml export-ignore
+/build.properties export-ignore
+/build.xml export-ignore
```
morozov: Are these Ant build files still relevant?

.gitattributes
```diff
@@ -1,12 +1,15 @@
 /tests export-ignore
```
morozov: Let's sort everything alphabetically. It's confusing to see `phpunit.xml` ignored but not `tests/` because tests go first.

.gitattributes
Majkl578: Same question crossed my mind. I guess not, we should remove them in another PR.

.gitattributes
Majkl578: :+1:

morozov (approved):
:ship:

Issue #2323, opened 2016-02-18
Problem with TEXT type in MSSQL due to VARCHAR(MAX) instead of NVARCHAR(MAX)


cfales:
The below issue was set for milestone 2.5.2 but was marked as Closed - Incomplete.
DBAL-1233: TEXT type in MSSQL should be NVARCHAR(MAX) not VARCHAR(MAX)

To my knowledge the issue remains (as there are other similar problems reported eg. DBAL-1823: Doctrine2 + pdo_sqlsrv the createQueryBuilder() Order By with varchar field does not work)

This is an issue that is affected our deployment.

Could someone have a look and see if this is a problem that can be addressed in an update?

mwey:
+1

Ocramius:
Ref: https://github.com/doctrine/dbal/issues/1182

Ocramius:
Ref: https://github.com/doctrine/dbal/issues/1351

PatrickLouys:
https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/SQLServerSchemaManager.php#L75

Introduced here https://github.com/doctrine/dbal/commit/56c0806d08d605d46cbb7cacb6b2254cb48c061f by @deeky666 

Could this be fixed by just switching varchar to nvarchar? Or is there a reason why varchar was used?

rwitchell:
+1

hockdudu:
Is there an update on this issue? The pull request #3262 was closed without being merged, but I don't quite understand why.

Ocramius:
@hockdudu looks like the original author deleted their fork, hence why the PR was closed.

Send one **WITH** a test, and then it shall be considered.

michalbundyra:
I've created PR #6421 solve this issue. I hope this time it will be considered as a bugfix and released. Thanks 🙌

Issue #1351, opened 2015-12-03
DBAL-1823: Doctrine2 + pdo_sqlsrv the createQueryBuilder() Order By with varchar field does not work


doctrinebot:
Jira issue originally created by user leoFelipe:

I am creating a QueryBuilder receiving values to createQueryBuilder() coming by parameter. I created testing in my local database with pdo_mysql driver and it worked perfectly, both ordered by a field of type int (id) or by a field of type varchar (titulo).

However, after testing in other database in external server with pdo_sqlsrv driver (and also sqlsrv driver) works when the "order by" is by a field of type int (id), but when ordination by a field of type varchar (titulo) gives timeout in the connection.

The following code prints the return of the mySQL database and the SQLServer database:

``` java
public function getBy($fields, $orderBy, $where = null, $params = null)
{
    if (empty($where))
        $where = '1 = 1';

    $query = $this->getEntityManager()->createQueryBuilder()
        ->select($fields)
        ->from($this->entity, $this->alias)
        ->where($where)
        ->add('orderBy', $orderBy)
        ->setFirstResult(URIHelper::$arrayParams['offset'])
        ->setMaxResults(URIHelper::$arrayParams['limit'])
        ->getQuery();

    if (!empty($params))
        $query->setParameters($params);

    print_r($query->getSQL());
    //print_r($query->getParameters());
    //print_r($query->getArrayResult());
    //die();
    return $query->getArrayResult();
}
```

---

**Return OK when I use MySQL database ordination titulo(varchar):**
!img01.png|thumbnail!

---

**Return OK when you use SQLServer database ordination id (int):**
!img02.png|thumbnail!

---

**Return ERROR when I use SQLServer database ordination titulo(varchar):**
!img03.png|thumbnail!

Issue #1182, opened 2015-05-19
DBAL-1233: TEXT type in MSSQL should be NVARCHAR(MAX) not VARCHAR(MAX)


doctrinebot:
Jira issue originally created by user Javad Rahimi:

PR #3262, opened 2018-08-22
SQLSRV / MSSQL / TSQL - Fix default CLOB type from varchar(max) to nchar(max) & nvarchar(max)

rwitchell:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | #2323, #1182, #1351

#### Summary

Modifies SQL Server (MSSQL, TSQL) VARCHAR column type to be NVARCHAR, to better handle unicode.

Initial breaking change: #451.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SQLServer2005Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServer2005Platform.php
@@ -48,7 +48,13 @@ class SQLServer2005Platform extends SQLServerPlatform
      */
     public function getClobTypeDeclarationSQL(array $field)
     {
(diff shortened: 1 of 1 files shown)


morozov:
@rwitchell please add a test case which would fail without the fix.

morozov (changes requested):
Test is missing.

rwitchell:
modifying this PR slightly as the library has lost the ability to accept a table as VARCHAR and allow you to keep the table as VARCHAR. Investigating whether an option 'is_unicode' or a unicodeStringType will solve this.

Issue #2880, opened 2017-10-06
Symfony 3 - Postgresql JSON type always appear in diff


PapsOu:
With Symfony 3.3.9, doctrine/common v2.8.1, doctrine/dbal v2.6.2 and doctrine/orm v2.5.11,  `bin/console doctrine:schema:update --dump-sql` always display changes related to JSON types even if applying them manually:
```sql
ALTER TABLE sylius_product_attribute_value ALTER json_value TYPE JSON;
ALTER TABLE sylius_product_attribute_value ALTER json_value DROP DEFAULT;
ALTER TABLE sylius_payment ALTER details TYPE JSON;
ALTER TABLE sylius_payment ALTER details DROP DEFAULT;
ALTER TABLE sylius_gateway_config ALTER config TYPE JSON;
ALTER TABLE sylius_gateway_config ALTER config DROP DEFAULT;
```

Here's gist for more details about versions https://gist.github.com/PapsOu/12805469b21bdeac6fe757174b8743f8

lcobucci:
Being fixed on https://github.com/doctrine/dbal/pull/2855

PapsOu:
Do you have any news about this PR ?

lcobucci:
@PapsOu no, quite busy with paid work that I haven't been able to finish stuff 😢

bordeux:
also that bug is for JSONB colum :) I

PapsOu:
Hi @lcobucci, do you have any news about the PR ?

lcobucci:
@PapsOu I'm waiting for reviews now. The fix is failing for Oracle but I didn't have time to look into that yet.

bordeux:
@lcobucci  what is your branch/repo name ? Where can i make some fixes ?

PapsOu:
@bordeux Everything is here... https://github.com/doctrine/dbal/pull/2855

jbdelhommeau:
Do you say if this bug is resolved ? I still have the problem with jsonb.

PapsOu:
@lcobucci (or any maintainer) : I'm experiencing again this bug.

A `bin/console do:sc:up --dump-sql` displays those lines :

```sql
ALTER TABLE pia_processing_data_type ALTER data TYPE TEXT;
ALTER TABLE pia_processing_data_type ALTER data DROP DEFAULT;
```

Those lines are always displayed after a `bin/console do:sc:up --force && bin/console do:sc:up --dump-sql`, even after executing those raw SQL manually.

[...]

PR #3266, opened 2018-08-22
Add test case for issue #2880

PapsOu:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2880

#### Summary

Adds a test case to cover specific situation : adding a `json_array` column to a table, executing changes, changing the `json_array` to a `string` still show a schema diff.


Ocramius:
Is this supposed to be closed?

PapsOu:
Yes because it is stale and I can reproduce the initial bug.

PR #3270, opened 2018-08-24
Extend SQLServerPlatform Tests to inherit previous version's test functions

rwitchell:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

- changed name from "Fix sql server tests".

#### Summary

- Changed SQL Server Platform tests to extend previous version of platform. 

Previously, the SQL Server Platform classes were not correctly inheriting their previous version.
This change is made on the assumption that each new version of a platform should inherit it's previous versions' tests too, so those tests are also run and pass in the newer version. If those tests are no longer relevant, than the test can be marked as skipped.

Due to this extension, two asserts in one test (\Doctrine\Tests\DBAL\Platforms\SQLServerPlatformTest::testScrubInnerOrderBy) are failing as the doModifyLimitQuery() function has been modified in 2012 compared to the base SQLServerPlatform.

We can either improve the doModifyLimitQuery() to handle SQL queries with 'dctrn_minrownum', otherwise I have marked the function incomplete in 2012.

This change will also allow a SQL Server 2016 platform to be added, and have all previous 2008, 2012 tests run against the platform.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Platforms/SQLServer2008PlatformTest.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/SQLServer2008PlatformTest.php
@@ -4,7 +4,7 @@ namespace Doctrine\Tests\DBAL\Platforms;
 
 use Doctrine\DBAL\Platforms\SQLServer2008Platform;
 
-class SQLServer2008PlatformTest extends AbstractSQLServerPlatformTestCase
+class SQLServer2008PlatformTest extends SQLServerPlatformTest
 {
     public function createPlatform()
     {
--- a/tests/Doctrine/Tests/DBAL/Platforms/SQLServer2012PlatformTest.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/SQLServer2012PlatformTest.php
@@ -6,7 +6,7 @@ use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Platforms\SQLServer2012Platform;
 use Doctrine\DBAL\Schema\Sequence;
 
-class SQLServer2012PlatformTest extends AbstractSQLServerPlatformTestCase
+class SQLServer2012PlatformTest extends SQLServer2008PlatformTest
 {
     public function createPlatform()
     {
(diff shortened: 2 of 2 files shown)


tests/Doctrine/Tests/DBAL/Schema/ColumnTest.php
```diff
@@ -62,7 +62,8 @@ public function testToArray()
      */
     public function testSettingUnknownOptionIsStillSupported() : void
     {
-        new Column('foo', $this->createMock(Type::class), ['unknown_option' => 'bar']);
+        $col1 = new Column('foo', $this->createMock(Type::class), ['unknown_option' => 'bar']);
+        self::assertInstanceOf(Column::class, $col1);
```
morozov: This assertion is redundant. The `new` operator cannot produce any else than an instance of the requested class.

morozov:
> Due to this extension, two asserts in one test (`\Doctrine\Tests\DBAL\Platforms\SQLServerPlatformTest::testScrubInnerOrderBy`) are failing as the `doModifyLimitQuery()` function has been modified in 2012 compared to the base `SQLServerPlatform`.
> 
> Did you want to improve the `doModifyLimitQuery()` to handle SQL queries with 'dctrn_minrownum'?

@rwitchell the issue description says it's a bug. What bug are you trying to fix?

tests/Doctrine/Tests/DBAL/Schema/ColumnTest.php
rwitchell: based on the test function, i was guessing that this was testing that the object constructor would not throw an exception if you pass 'unknown_option'. The pass is that the $col1 is correctly a Column. Otherwise, my local php unit was giving an alert that the test has no assertion, so no test was actually being performed.

morozov: It has something to do with the Symfony bridge for PHPUnit (https://github.com/doctrine/dbal/pull/2846). It should take care of this case. Please omit this change as long as it's not part of the fix you're working on.

morozov:
> […] each SQL server version should run the previous version tests, for better coverage - do you agree?

@rwitchell I don't know the history behind the platform classes and their tests and why they are organized the way they are. If there were breaking changes between two SQL Server versions, then the tests defined in the older version's test class won't pass in the newer one's. So in general it's not necessarily true.

> Changes i'm making in another branch rely on all previous SQL server tests to pass in previous versions […]

Would it make sense to combine the changes together so that we could see the full picture?

> […] however some tests are modified to meet SQL Server 2012/2014/azure requirements.

All tests currently pass on the servers provided by AppVeyor and SQL Server 2017 for Linux. Are you saying some fragments of the tests or code are needed only for Azure?

I'm sorry, I'm still missing the point of why exactly those changes are being made.

tests/Doctrine/Tests/DBAL/Schema/ColumnTest.php
rwitchell: reverted.

rwitchell: reverted.

PR #3275, opened 2018-08-31
Add .doctrine-project.json to root of the project.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Add .doctrine-project.json to root of the project for the website build process.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,8 +1,17 @@
 {
     "name": "doctrine/dbal",
     "type": "library",
-    "description": "Database Abstraction Layer",
-    "keywords": ["dbal", "database", "persistence", "queryobject"],
+    "description": "Powerful PHP database abstraction layer (DBAL) with many features for database schema introspection and management.",
+    "keywords": [
+        "php",
+        "mysql",
+        "pgsql",
+        "dbal",
+        "database",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -1,8 +1,17 @@
 {
     "name": "doctrine/dbal",
     "type": "library",
-    "description": "Database Abstraction Layer",
-    "keywords": ["dbal", "database", "persistence", "queryobject"],
+    "description": "Powerful PHP database abstraction layer (DBAL) with many features for database schema introspection, schema management and PDO abstraction.",
```
morozov: Despite the fact it's inspired by PDO, it's not a PDO abstraction and may diverge from PDO in 3.0 even more. Would it make sense to omit the PDO reference?

jwage: I can update it to whatever you want it to be. Want to write something?

morozov: No, I'm fine with the rest of the statement, just the PDO part may be misleading. Therefore:

> "Powerful PHP database abstraction layer (DBAL) with many features for database schema introspection and management."

morozov (approved):
:+1:

PR #3279, opened 2018-09-07
[Docs] wrong charset in mysql example

pscheit:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     |  no

#### Summary

I think the name of the UTF-8 Charset in mysql is utf8 not UTF-8, right?


morozov:
`utf8` in MySQL is `utf8mb3` which doesn't entirely correspond to the specification.`utf8mb4` is the recommended alternative.

morozov:
Yes, it should be `utf8mb4`.

Issue #3202, opened 2018-06-29
MariaDB 10.3 Native JSON Support


kiler129:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary
https://github.com/doctrine/dbal/commit/52b4692ebefd5a25420238aa53176ef8d4dc2e83 disabled JSON support for MariaDB 10.2, since the engine is based on 5.6.

MariaDB 10.3 doesn't have its own platform file, but I think it needs one now to enable native JSON storage.

WDYT?

morozov:
I think it makes sense. If you're going to file a pull request, please make sure there's a functional test which creates and uses a JSON column. This way, we'll make sure that all platforms which declare JSON as supported actually support it.

hal869:
with
```
doctrine/dbal v2.7.1
```
on
```
Ver 10.3.8-MariaDB-log for Linux on x86_64 (Source distribution)
```
I'm seeing similar issues to those previously reported (& solved) for MDB 10.2x here

	https://github.com/doctrine/dbal/pull/2825

Particularly, the 'infinite schema updates' and failing
```
doctrine schema:validate
```
after a (reported) 'successful'
```
doctrine schema:update
```
Other than the common issue of 'MDB 10.3 support', not clear that this is the right place for this.

If so, and it's useful, I can add the details here on request.

Atm, I've no specific functional test, other than a new DB init on a symfony4 project install.

hal869:
@morozov 

On a new SF4 + phpcr project, after package installs,

starting with new/clean DBs

<details>
  <summary>mysqlshow ...</summary>

```
	mysqlshow testdb
		Database: testdb
			+--------+
			| Tables |
			+--------+
			+--------+

	mysqlshow testdbphpcr
		Database: testdbphpcr
			+--------+
			| Tables |
			+--------+
			+--------+
```

</details>

a phpcr DB init reports successful completion
```
	bin/console doctrine:phpcr:init:dbal --force
		Jackalope Doctrine DBAL tables have been initialized successfully.
```

<details>
  <summary>mysqlshow ...</summary>

```
	mysqlshow testdbphpcr
		Database: testdbphpcr
		+----------------------------+
		|           Tables           |
		+----------------------------+
		| phpcr_binarydata           |
		| phpcr_internal_index_types |
		| phpcr_namespaces           |
		| phpcr_nodes                |
		| phpcr_nodes_references     |
[...]

PR #3283, opened 2018-09-10
MariaDB improvements, support 10.3

sidz:
Support for MariaDB 10.3+

### **Supported Reserved Keywords List Improvements:** (https://mariadb.com/kb/en/library/reserved-words/)

- [x] ~~Remove unsupported reserved keywords by MariaDB 10.2.~~
- [x] ~~Support for MariaDB 10.3 reserved keywords.~~
- [x] Added MariaDB 10.3 in travis (both pdo_mysql and mysqli)

### Native JSON Support (https://github.com/doctrine/dbal/issues/3202) ?

MariaDB still doesn't have native support for the JSON type. I mean that it still an alias for the LONGTEXT type: https://mariadb.com/kb/en/library/json-data-type/

So In fact it still doesn't support a native JSON. Do we need to mark it as native and replace LONGTEXT via JSON ?

### Work In Progress
Please fill free to let me know what I missed or may miss.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -180,6 +180,18 @@ jobs:
         mariadb: 10.2
 
     - stage: Test
+      php: 7.2
+      env: DB=mariadb MARIADB_VERSION=10.3
+      addons:
+          mariadb: 10.3
+
+    - stage: Test
+      php: 7.2
+      env: DB=mariadb.mysqli MARIADB_VERSION=10.3
+      addons:
+          mariadb: 10.3
+
+    - stage: Test
       php: 7.1
       env: DB=pgsql POSTGRESQL_VERSION=9.2 COVERAGE=yes
       services:
--- a/docs/en/reference/platforms.rst
+++ b/docs/en/reference/platforms.rst
@@ -37,6 +37,11 @@ MySQL
 -  ``MySQL57Platform`` for version 5.7 (5.7.9 GA) and above.
 -  ``MySQL80Platform`` for version 8.0 (8.0 GA) and above.
 
+MariaDB
+^^^^^
+
+-  ``MariaDb1027Platform`` for version 10.2 (10.2.7 GA) and above.
+
 Oracle
 ^^^^^^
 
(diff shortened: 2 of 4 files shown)


lib/Doctrine/DBAL/Platforms/MariaDb102Platform.php
```diff
@@ -17,19 +17,21 @@
...
  * Provides the behavior, features and SQL dialect of the MariaDB 10.2 (10.2.7 GA) database platform.
  *
- * Note: Should not be used with versions prior ro 10.2.7.
+ * Note: Should not be used with versions prior to 10.2.7.
  *
  * @author Vanvelthem Sébastien
  * @link   www.doctrine-project.org
  */
-final class MariaDb1027Platform extends MySqlPlatform
+class MariaDb102Platform extends MySqlPlatform
```
Ocramius: The rename is a BC break, and should be reverted: if you need an additional abstract inheritance level, adding it will be more effective than renaming herr 👍

Ocramius (dismissed):

sidz:
So need an advice. Looks like that is what I really may contribute for MariaDB 10.3 support and I don't see a lot of sense in this PR.

based on @morozov comment here https://github.com/doctrine/dbal/pull/3103#issuecomment-393025529 it was done on purpose. 

Should I separate platforms or may close this PR for now?

morozov:
> So need an advice. Looks like that is what I really may contribute for MariaDB 10.3 support and I don't see a lot of sense in this PR.

Adding new builds to the matrix makes perfect sense to me.

> based on @morozov comment here https://github.com/doctrine/dbal/pull/3103#issuecomment-393025529 it was done on purpose.

What exactly do you mean by "it"?

> Should I separate platforms or may close this PR for now?

You mean, given that support for MariaDB 10.3 doesn't require any code changes besides the keywords, do we need a separate platform at all? I don't know. If there were changes in reserved keywords between the versions (a link would be appreciated) or the original list was inaccurate, having separate platforms would make sense.

morozov:
@sidz I was okay with not having dedicated platform/keywords classes for MariaDB 10.3 at that moment and I still am. But if you think it's important to have them separated — this is fine too. I'm just curious if there are any practically visible consequences of having quotes around an identifier which doesn't have to have the quotes?

morozov:
@sidz please squash. Looks good to me :+1:

sidz:
@morozov done

morozov (approved):

morozov:
Thank you, @sidz! Merged.

javiereguiluz:
@sidz thanks for working on this and thanks @morozov for reviewing and merging!

Issue #3296, opened 2018-09-22
doctrine:database:create use Oracle database get Notice.


lcp0578:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.5.13

#### Current behaviour

```
php7 bin/console doctrine:database:create --connection=oracle -v

In AbstractOracleDriver.php line 121:

  [Symfony\Component\Debug\Exception\ContextErrorException]
  Notice: Undefined index: dbname


Exception trace:
 Doctrine\DBAL\Driver\AbstractOracleDriver->getEasyConnectString() at /Users/lcp0578/htdocs/sxjicheng/water_datacenter/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/OCI8/Driver.php:61
 Doctrine\DBAL\Driver\OCI8\Driver->_constructDsn() at /Users/lcp0578/htdocs/sxjicheng/water_datacenter/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/OCI8/Driver.php:42
 Doctrine\DBAL\Driver\OCI8\Driver->connect() at /Users/lcp0578/htdocs/sxjicheng/water_datacenter/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php:360
 Doctrine\DBAL\Connection->connect() at /Users/lcp0578/htdocs/sxjicheng/water_datacenter/vendor/doctrine/doctrine-bundle/Command/CreateDatabaseDoctrineCommand.php:84
 Doctrine\Bundle\DoctrineBundle\Command\CreateDatabaseDoctrineCommand->execute() at /Users/lcp0578/htdocs/sxjicheng/water_datacenter/vendor/symfony/symfony/src/Symfony/Component/Console/Command/Command.php:252
 Symfony\Component\Console\Command\Command->run() at /Users/lcp0578/htdocs/sxjicheng/water_datacenter/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:964
 Symfony\Component\Console\Application->doRunCommand() at /Users/lcp0578/htdocs/sxjicheng/water_datacenter/vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Console/Application.php:86
 Symfony\Bundle\FrameworkBundle\Console\Application->doRunCommand() at /Users/lcp0578/htdocs/sxjicheng/water_datacenter/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:248
[...]

PR #3297, opened 2018-09-23
Do not generate SID or SERVICE_NAME when dbname or service name is not specified

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3296

Additionally, refactored the logic of connection string generation to use array representation of the connection parameters before rendering the string.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractOracleDriver.php
@@ -20,6 +20,7 @@
 namespace Doctrine\DBAL\Driver;
 
 use Doctrine\DBAL\Driver;
+use Doctrine\DBAL\Driver\AbstractOracleDriver\EasyConnectString;
 use Doctrine\DBAL\Exception;
 use Doctrine\DBAL\Platforms\OraclePlatform;
 use Doctrine\DBAL\Schema\OracleSchemaManager;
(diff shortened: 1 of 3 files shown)


tests/Doctrine/Tests/DBAL/Driver/AbstractOracleDriverTest.php
```diff
@@ -44,6 +46,61 @@ public function testReturnsDatabaseNameWithConnectDescriptor()
...
+    /**
+     * @param mixed[] $params
+     * @dataProvider easyConnectStringProvider
+     */
+    public function testEasyConnectString(array $params, string $expected) : void
+    {
+        $driver = $this->createMock(AbstractOracleDriver::class);
+
+        $reflection = new ReflectionMethod($driver, 'getEasyConnectString');
+        $reflection->setAccessible(true);
```
Ocramius: This is extremely ugly: would it make sense to move this to a value object instead?

Ocramius (changes requested):

tests/Doctrine/Tests/DBAL/Driver/AbstractOracleDriverTest.php
morozov: Having a `ConnectionString` class which implements rendering of `['DESCRIPTION' => ...]` to `(DESCRIPTION=...)` would make sense but it's not what being tested here. I'm testing the logic of mapping of DBAL configuration parameters to the connection string ones which doesn't beling to the `ConnectionString` itself.

Do we need another class which would do the mapping? It will only be used by `AbstractOracleDriver` and will only have one implementation. Besides being able to test it in isolation, what's its purpose?

morozov: Or… instead of adding the class for mapping, we can create a named constructor like `fromConfigurationParameters()` in the value object and get it tested.

morozov:
@Ocramius does this look better?

Ocramius (approved):
:+1: :ship:

PR #3301, opened 2018-09-25
First example of proposed handleTypeConversion function for Type

TomHAnderson:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | maybe/no

#### Summary

This PR is the result of the conversation at https://github.com/doctrine/dbal/pull/3291

Instead of aiming towards fixing the existing Type classes which do not allow the same input type as the output type this creates a new function on the base Type class.  For this commit the base Type class is modified and one existing Type class which may benefit from less code.

### DIFF
--- a/lib/Doctrine/DBAL/Types/DateTimeType.php
+++ b/lib/Doctrine/DBAL/Types/DateTimeType.php
@@ -66,7 +66,7 @@ class DateTimeType extends Type implements PhpDateTimeMappingType
      */
     public function convertToPHPValue($value, AbstractPlatform $platform)
     {
-        if ($value === null || $value instanceof \DateTimeInterface) {
+        if ($value === null) {
             return $value;
         }
 
@@ -82,4 +82,16 @@ class DateTimeType extends Type implements PhpDateTimeMappingType
 
         return $val;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+     public function handleTypeConversion($value, AbstractPlatform $platform)
+     {
+        if ($value instanceof \DateTimeInterface) {
+            return $value;
+        }
+
+        return $this->convertToPHPValue($value, $platform);
+     }
 }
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Types/DateTimeType.php
```diff
@@ -66,7 +66,7 @@ public function convertToDatabaseValue($value, AbstractPlatform $platform)
      */
     public function convertToPHPValue($value, AbstractPlatform $platform)
     {
-        if ($value === null || $value instanceof \DateTimeInterface) {
```
beberlei: I think this is still needed in case of `DateTime`, because as far as I remember PDO Sqlite or sqlsrv immediately return DateTime objects for date/time/datetime typed fields.

beberlei (commented):
Not putting the "self" checks into each `convertToPHPValue` function makes a lot of sense, because as Marco said on the previous discussion, the DBAL Types exist in the DBAL context and not in hydration.

That said, `handleTypeConversion` is a bit far away naming wise from the underlying action `convertToPHPValue` and it doesn't hint at which conversion is handled the db=>php or the php=>db one. Can I suggest `normalizeToPHPValue` as an option for another name?

lib/Doctrine/DBAL/Types/DateTimeType.php
```diff
@@ -82,4 +82,16 @@ public function convertToPHPValue($value, AbstractPlatform $platform)
 
         return $val;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+     public function handleTypeConversion($value, AbstractPlatform $platform)
```
morozov: When I suggested adding a new method, I was thinking of something like `isValueOfType($value)`. So that the calling code could check it first and then call the conversion function if needed.

If in 3.0 we get rid of type inheritance, every `Type` interface implementor would have to have the same copy/pasted implementation of this method.

lib/Doctrine/DBAL/Types/DateTimeType.php
morozov: If that's was the case, we'd most likely have failures in `TypeConversionTest` with those drivers. However, double checking won't hurt.

Issue #2610, opened 2017-01-19
Weird createQueryBuilder behavior on UPDATE on SQL Server 2014


Norcoen:
I had this code running for some years with a mysql database.
Recently we had to switch to SQL Server 2014, now I'm getting weird errors because somehow it does not like the UPDATE-Syntax doctrine is using.

It seems like sqlsrv does not like the table alias, while in the [documentation](http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/query-builder.html#set-clause) this seems to be fine
 ```php
$afrmol = $db->createQueryBuilder()
	->update('tbl_priv_user', 'u')
	//->set('u.lastlogin', 'CURRENT_TIMESTAMP()')
	->set('u.logincount', 'u.logincount + 1')
	//->where('u.userid = :usermail')
	//->setParameter('usermail', $_SESSION['user'])
	->execute()
;
echo "affected rows:" . $afrmol;
exit;
```

```
Internal Server Error

An exception occurred while executing 'UPDATE tbl_priv_user u SET u.logincount = u.logincount + 1': SQLSTATE[42000]: [Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Incorrect syntax near 'u'.

[D:/wwwroot_test/includes/doctrine/doctrine-2.4.5/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php:864] Doctrine\DBAL\DBALException::driverExceptionDuringQuery(PDOException::__set_state(array('errorInfo'=>array('42000',102,'[Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Incorrect syntax near 'u'.'))),'UPDATE tbl_priv_user u SET u.logincount = u.logincount + 1',array())
[D:/wwwroot_test/includes/doctrine/doctrine-2.4.5/vendor/doctrine/dbal/lib/Doctrine/DBAL/Query/QueryBuilder.php:206] Doctrine\DBAL\Connection->executeUpdate('UPDATE tbl_priv_user u SET u.logincount = u.logincount + 1',array(),array())
[D:/wwwroot_test/configuration/legacy_configuration.php:78] Doctrine\DBAL\Query\QueryBuilder->execute()
[D:/wwwroot_test/configuration/extended_configuration.php:37] require_once('D:\wwwroot_test\configuration\legacy_configuration.php')
[D:/wwwroot_test/index.php:36] require('D:\wwwroot_test\configuration\extended_configuration.php')
```

[...]

PR #3304, opened 2018-09-27
Fix executing insert() with empty data

simPod:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

Before this change it run query with invalid syntax. Now `insert()` with empty data returns 0 rows.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -785,7 +785,7 @@ class Connection implements DriverConnection
     public function insert($tableExpression, array $data, array $types = [])
     {
         if (empty($data)) {
-            return $this->executeUpdate('INSERT INTO ' . $tableExpression . ' ()' . ' VALUES ()');
+            return 0;
         }
 
         $columns = [];
--- a/tests/Doctrine/Tests/DBAL/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/ConnectionTest.php
@@ -324,15 +324,11 @@ class ConnectionTest extends \Doctrine\Tests\DbalTestCase
         self::assertFalse($conn->isTransactionActive());
     }
 
-    public function testEmptyInsert()
+    public function testEmptyInsert() : void
     {
         $conn = $this->getExecuteUpdateMockConnection();
 
-        $conn->expects($this->once())
-            ->method('executeUpdate')
-            ->with('INSERT INTO footable () VALUES ()');
-
-        $conn->insert('footable', array());
+        self::assertSame(0, $conn->insert('footable', []));
     }
 
     /**


morozov:
> Before this change it run query with invalid syntax. Now `insert()` with empty data returns 0 rows.

It's not invalid syntax:
```
mysql>  CREATE TABLE test (id INT NOT NULL AUTO_INCREMENT PRIMARY KEY);
Query OK, 0 rows affected (0.12 sec)

mysql> INSERT INTO test() VALUES();
Query OK, 1 row affected (0.02 sec)

mysql> INSERT INTO test() VALUES();
Query OK, 1 row affected (0.01 sec)

mysql> INSERT INTO test() VALUES();
Query OK, 1 row affected (0.01 sec)

mysql> select * from test;
+----+
| id |
+----+
|  1 |
|  2 |
|  3 |
+----+
3 rows in set (0.00 sec)
```

simPod:
It is in postgres for example.

```psqls
test=# CREATE TABLE test (id uuid);
CREATE TABLE
test=# insert into test() values();
ERROR:  syntax error at or near ")"
LINE 1: insert into test() values();
                         ^
```

There's no need to call that query in mysql as well.

Ocramius:
The test case needs to be adapted rather than removed then :+1:

simPod:
Will probably not work in oracle as well as they're very similar with postgres in terms of sql dialect. Not sure about others.

morozov:
> There's no need to call that query in mysql as well.

What do you mean? That query inserts a row with all default values and may affect a non-zero number of rows.

simPod:
> That query inserts a row with all default values

Ah didn't notice. Hm and what to do about this behaviour then? It isn't unified in all supported platforms. IMO it shouldn't be allowed to pass invalid query in non-mysql platforms.

morozov:
We can only document it, same as https://github.com/doctrine/dbal/issues/2610#issuecomment-273823889. It's a feature which is not supported by all platforms.

> IMO it shouldn't be allowed to pass invalid query in non-mysql platforms.

How do we not allow using it?

Majkl578:
>  That query inserts a row with all default values

In Postgres you can do this to achieve the same:
```sql
insert into sometable default values;
```

Majkl578:
Maybe the way to go would be `AbstractPlatform::getEmptyInsertStatement()` or something like that?

morozov:
What would that method do for the platforms which don't support this syntax? I see that some platforms like SQL Server allow using `DEFAULT` to explicitly specify the default value. But first, it'd be good to have some summary of what supported platforms are capable of.

Majkl578:
It can throw `NotSupported` for example. Right now this doesn't work with some platforms anyway. I don't know about other platforms - quickly googling:
SQLite also supports `DEFAULT VALUES`: http://www.sqlitetutorial.net/sqlite-insert/
Oracle and DB2 support `DEFAULT` keyword for values, not sure about empty column list: https://docs.oracle.com/javadb/10.6.2.1/ref/rrefsqlj40774.html + https://www.ibm.com/support/knowledgecenter/en/ssw_ibm_i_72/sqlp/rbafyinsert.htm

morozov:
I'm fine if we want to fix this behavior everywhere we can (most likely, it can be done for all platforms). The original proposal was to break it even where it worked :-)

morozov:
Closing as stale. The proposal in https://github.com/doctrine/dbal/pull/3304#issuecomment-425132507 looks like the right API to be used by the consumers. As for `Connection::insert()` we can consider it non-portable in this edge case.

PR #3305, opened 2018-10-01
Fix Fetch mode query in MasterSlaveConnection

BradCrumb:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

The `Doctrine\DBAL\Connections\MasterSlaveConnection` class doesn't set the default fetchmode when using die `query()` method.

The issue came up when we were migrating a normal connection to master/slave. First we got a FETCH_ASSOC and after the config change we had FETCH_BOTH.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/MasterSlaveConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/MasterSlaveConnectionTest.php
@@ -3,6 +3,7 @@
 namespace Doctrine\Tests\DBAL\Functional;
 
 use Doctrine\DBAL\Connections\MasterSlaveConnection;
+use Doctrine\DBAL\Driver\Statement;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Schema\AbstractSchemaManager;
 use Doctrine\DBAL\Schema\Table;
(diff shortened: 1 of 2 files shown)


Ocramius (changes requested):
Requires test case

BradCrumb:
@Ocramius Just checked the current existing test in `/test/Doctrine/Tests/DBAL/Functional/MasterSlaveConnectionTest.php`. But a lot of methods of the MasterSlaveConnection class (also the query method where i added the setFetchMode) are uncovered in the test class, so I think it is better to create a new issue for completing the coverage of this connection class.

BradCrumb:
@Ocramius Fine I add the test for the query method in the MasterSlaveConnectionTest because I changed the code in that method. I will add an issue for all other methods.

tests/Doctrine/Tests/DBAL/Functional/MasterSlaveConnectionTest.php
```diff
@@ -189,4 +190,33 @@ public function testMasterSlaveConnectionCloseAndReconnect()
...
+
+    public function testQueryOnMaster()
+    {
+        $conn = $this->createMasterSlaveConnection();
+
+        $query = 'SELECT count(*) as num FROM master_slave_table';
+
+        $statement = $conn->query($query);
+
+        self::assertTrue($statement instanceof Statement);
```
morozov: This assertion seems redundant. The `query()` method always returns a statement.

tests/Doctrine/Tests/DBAL/Functional/MasterSlaveConnectionTest.php
```diff
@@ -189,4 +190,33 @@ public function testMasterSlaveConnectionCloseAndReconnect()
...
+    public function testQueryOnSlave()
+    {
+        $conn = $this->createMasterSlaveConnection();
+        $conn->connect('slave');
+
+        $query = 'SELECT count(*) as num FROM master_slave_table';
+
+        $statement = $conn->query($query);
+
+        self::assertTrue($statement instanceof Statement);
```
morozov: This assertion seems redundant. The `query()` method always returns a statement.

morozov (commented):
@BradCrumb could you clarify how the tests are related to the code being changed? The issue and the fix are about the fetch mode, but nothing is fetched in the test.

Ocramius (dismissed):
@BradCrumb this looks good, just needs a squash before being considered for merge :+1:

morozov (approved):

morozov:
Thank you, @BradCrumb.

PR #3310, opened 2018-10-04
Ensuring correct ALTER TABLE statement for creation of an AUTO INCREMENT column as new PRIMARY KEY

arnegroskurth:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2807 

#### Summary

Ensures that the primary key is created within the same "alter table" statement that an auto-increment column is added to the table as part of the new primary key.

Before the fix for this problem this resulted in a database error:
> SQLSTATE[42000]: Syntax error or access violation: 1075 Incorrect table definition; there can be only one auto column and it must be defined as a key


Ocramius:
@arnegroskurth please target `master`

PR #3311, opened 2018-10-04
Fix: Ensuring correct ALTER TABLE statement for creation of an AUTO INCREMENT column as new PRIMARY KEY

arnegroskurth:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2807

#### Summary

Ensures that the primary key is created within the same "alter table" statement that an auto-increment column is added to the table as part of the new primary key.

Before the fix for this problem this resulted in a database error:

> SQLSTATE[42000]: Syntax error or access violation: 1075 Incorrect table definition; there can be only one auto column and it must be defined as a key

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -599,6 +599,17 @@ class MySqlPlatform extends AbstractPlatform
             $keyColumns   = array_unique(array_values($diff->addedIndexes['primary']->getColumns()));
             $queryParts[] = 'ADD PRIMARY KEY (' . implode(', ', $keyColumns) . ')';
             unset($diff->addedIndexes['primary']);
+        } elseif (isset($diff->changedIndexes['primary'])) {
+            // Necessary in case the new primary key includes a new auto_increment column
+            foreach ($diff->changedIndexes['primary']->getColumns() as $columnName) {
+                if (isset($diff->addedColumns[$columnName]) && $diff->addedColumns[$columnName]->getAutoincrement()) {
+                    $keyColumns   = array_unique(array_values($diff->changedIndexes['primary']->getColumns()));
+                    $queryParts[] = 'DROP PRIMARY KEY';
+                    $queryParts[] = 'ADD PRIMARY KEY (' . implode(', ', $keyColumns) . ')';
+                    unset($diff->changedIndexes['primary']);
+                    break;
+                }
+            }
         }
 
         $sql      = [];
--- a/tests/Doctrine/Tests/DBAL/Functional/Platform/NewPrimaryKeyWithNewAutoIncrementColumnTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Platform/NewPrimaryKeyWithNewAutoIncrementColumnTest.php
@@ -1,0 +1,69 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Functional\Platform;
+
+use Doctrine\DBAL\Schema\Comparator;
+use Doctrine\Tests\DbalFunctionalTestCase;
+use function in_array;
+
+final class NewPrimaryKeyWithNewAutoIncrementColumnTest extends DbalFunctionalTestCase
+{
+    /**
+     * {@inheritDoc}
+     */
+    protected function setUp()
+    {
+        parent::setUp();
+
+        if (in_array($this->getPlatform()->getName(), ['mysql'])) {
+            return;
+        }
+
+        $this->markTestSkipped('Restricted to MySQL.');
+    }
+
+    /**
+     * Ensures that the primary key is created within the same "alter table" statement that an auto-increment column
+     * is added to the table as part of the new primary key.
+     *
+     * Before the fix for this problem this resulted in a database error: (at least on mysql)
+     * SQLSTATE[42000]: Syntax error or access violation: 1075 Incorrect table definition; there can be only one auto
+     * column and it must be defined as a key
+     */
+    public function testAlterPrimaryKeyToAutoIncrementColumn()
+    {
+        $schemaManager = $this->connection->getSchemaManager();
+        $schema        = $schemaManager->createSchema();
+
+        $table = $schema->createTable('dbal2807');
+        $table->addColumn('initial_id', 'integer');
+        $table->setPrimaryKey(['initial_id']);
+
+        $schemaManager->dropAndCreateTable($table);
+
+        $newSchema = clone $schema;
+        $newTable  = $newSchema->getTable($table->getName());
+        $newTable->addColumn('new_id', 'integer', ['autoincrement' => true]);
+        $newTable->dropPrimaryKey();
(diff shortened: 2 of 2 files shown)


arnegroskurth:
Will probably also resolve issue [3245](https://github.com/doctrine/doctrine2/issues/3245) of doctrine/doctrine2

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL2807Test.php
```diff
@@ -0,0 +1,56 @@
...
+
+        $newTable = clone $table;
+        $newTable->addColumn('new_id', 'integer', ['autoincrement' => true]);
+        $newTable->dropPrimaryKey();
+        $newTable->setPrimaryKey(['new_id']);
+
+        $diff = (new Comparator())->diffTable($table, $newTable);
+
+        $this->assertSame(
+            ['ALTER TABLE my_table ADD new_id INT AUTO_INCREMENT NOT NULL, DROP PRIMARY KEY, ADD PRIMARY KEY (new_id)',],
```
morozov: Instead of making an assertion against SQL, please apply the diff and introspect the resulting table. This way, we could make sure that the SQL is correct and we can update it if needed w/o reworking the test.

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL2807Test.php
```diff
@@ -0,0 +1,56 @@
...
+final class DBAL2807Test extends DbalFunctionalTestCase
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function setUp()
+    {
+        parent::setUp();
+
+        if (!in_array($this->getPlatform()->getName(), ['mysql']))
```
morozov: While the current test implementation checks MySQL-specific SQL, would it make sense to enable the test for all relevant platforms? Dropping and setting back the PK should work on most if not all of them.

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL2807Test.php
```diff
@@ -0,0 +1,56 @@
...
+        {
+            $this->markTestSkipped('Restricted to MySQL.');
+
+            return;
```
morozov: This is unnecessary.

morozov (commented):
@arnegroskurth, thank you for the patch. Please see a few comments.

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL2807Test.php
arnegroskurth: I agree that this -should- work on every platform but I cannot test this locally. I will just remove the restriction and see how the CI tests go.

arnegroskurth:
@morozov The test now asserts the correct resulting schema and is no longer restricted to mysql

morozov:
@arnegroskurth please look at the failing tests. Before getting the fix merged, the tests should be fixed for all supported platforms or marked as skipped. In the latter case, please also file a new issue or issues describing the reason of the failure.

arnegroskurth:
The test is now restricted to mysql again. I think this made sense in the first place as I fixed the problem only within Doctrine\DBAL\Platforms\MysqlPlatform anyway.

arnegroskurth:
@morozov Is there anything else you need to merge this?

morozov:
@arnegroskurth please resolve https://github.com/doctrine/dbal/pull/3311#discussion_r222887715 and move the new test under the `DBAL\Functional\Platform` namespace. Naming tests after issue numbers is not informative. Also, please fix the build (should be fixed by a rebase).

arnegroskurth:
@morozov The test is now renamed, moved and your code style has been applied. As already discussed the skip for non-mysql platforms IS necessary as the fix only covers the mysql platform. The follow-up ticket for other platforms already exists.

morozov:
@arnegroskurth I was referring to the `return` statement. It's not necessary since `markTestSkipped()` throws an exception.

tests/Doctrine/Tests/DBAL/Functional/Platform/NewPrimaryKeyWithNewAutoIncrementColumn.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Functional\Platform;
+
+use Doctrine\DBAL\Schema\Comparator;
+use Doctrine\Tests\DbalFunctionalTestCase;
+use function in_array;
+
+final class NewPrimaryKeyWithNewAutoIncrementColumn extends DbalFunctionalTestCase
```
morozov: Should be `NewPrimaryKeyWithNewAutoIncrementColumnTest`.

tests/Doctrine/Tests/DBAL/Functional/Platform/NewPrimaryKeyWithNewAutoIncrementColumn.php
```diff
@@ -0,0 +1,68 @@
...
+{
+    /**
+     * {@inheritDoc}
+     */
+    public function setUp()
```
morozov: `setUp()` is protected.

arnegroskurth:
@morozov The tests now failing are not related to my changes

morozov:
@arnegroskurth I've restarted the errored job. Such hiccups sometimes happen on Travis, and we cannot do anything about them.

arnegroskurth:
@morozov I had to re-add the return-statement to meet the codestyle requirements

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL2807Test.php
morozov: @arnegroskurth I see. @Majkl578 what if we disable this rule? It's often so annoying and forces write cumbersome code. In this case it could probably be:
```php
if (in_array(/* ... */)) {
    return;
}

$this->markTestSkipped();
```

arnegroskurth: @morozov Please just change the code in the way you want it to be yourself. I opened up the MR two months (!!) ago and seven out of eleven commits are dedicated to code-styling and such. I know that I can use phpcs locally and I did. But if you still have issues with the code after your own style-definition said it would be OK then I really cannot help it :-1:

morozov: > Please just change the code in the way you want it to be yourself.

I just did.

> But if you still have issues with the code after your own style-definition said it would be OK then I really cannot help it -1

Our coding standard has issues as well as any software does.

morozov:
@arnegroskurth as long as you're okay with the changes I pushed, please squash all commits, and I'll merge it.

morozov (approved):

morozov:
:ship:

Issue #2769, opened 2017-07-05
Connection ping method can trigger uncaught PDOExceptions


gregl83:
Reading the ping method phpdoc it's not immediately clear to me if the method should allow driver exceptions to bubble up.

My expectation was that this method would always return true/false based on the success of connecting to the database and performing a dummy query.

When having a connection timeout or close it's possible to encounter PDOException instead.

Ocramius:
@gregl83 this should be fixed rather than documented as a possible crash: a `ping` operation failing due to invalid connection information should simply return `false`

gregl83:
@Ocramius my sentiment as well.

morozov:
Closing per https://github.com/doctrine/dbal/pull/4128.

PR #3313, opened 2018-10-04
MasterSlaveConnection::ping() throws PDOException

sidz:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | https://github.com/doctrine/dbal/issues/3118 https://github.com/doctrine/dbal/issues/2769 

#### Summary

Fix existing issue with uncaught exception in the `MasterSlaveConnection::query`.
Also looks like `ping` method in the MasterSlaveConnection class will send pings to the slave connection:
https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php#L124-L133

I've marked it as BC Break as `ping` method will change it behavior in the `MasterSlaveConnection` and will ping master instead of slave

p.s. this PR covers 2 issues: https://github.com/doctrine/dbal/issues/3118 and https://github.com/doctrine/dbal/issues/2769

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -9,6 +9,7 @@ use Doctrine\DBAL\Cache\CacheException;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
 use Doctrine\DBAL\Cache\ResultCacheStatement;
 use Doctrine\DBAL\Driver\Connection as DriverConnection;
+use Doctrine\DBAL\Driver\PDOException;
 use Doctrine\DBAL\Driver\PingableConnection;
 use Doctrine\DBAL\Driver\ResultStatement;
 use Doctrine\DBAL\Driver\ServerInfoAwareConnection;
@@ -1671,7 +1672,7 @@ class Connection implements DriverConnection
             $this->query($this->getDatabasePlatform()->getDummySelectSQL());
 
             return true;
-        } catch (DBALException $e) {
+        } catch (PDOException | DBALException $e) {
             return false;
         }
     }
--- a/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
+++ b/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
@@ -5,11 +5,13 @@ namespace Doctrine\DBAL\Connections;
 use Doctrine\Common\EventManager;
 use Doctrine\DBAL\Configuration;
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Driver;
 use Doctrine\DBAL\Driver\Connection as DriverConnection;
 use Doctrine\DBAL\Event\ConnectionEventArgs;
 use Doctrine\DBAL\Events;
 use InvalidArgumentException;
+use Throwable;
 use function array_rand;
 use function count;
 use function func_get_args;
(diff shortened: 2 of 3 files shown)


lib/Doctrine/DBAL/Connection.php
```diff
@@ -1673,7 +1674,7 @@ public function ping()
             $this->query($this->getDatabasePlatform()->getDummySelectSQL());
 
             return true;
-        } catch (DBALException $e) {
+        } catch (PDOException | DBALException $e) {
```
Ocramius: When does a `PDOException` occur? I believe we upcast those, so this is fixing a symptom

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -370,4 +378,24 @@ public function prepare($statement)
 
         return parent::prepare($statement);
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function ping()
+    {
+        $this->connect('master');
```
Ocramius: Seems incorrect to me: it should simply ping whichever connection is already selected, as otherwise we may attempt to select the slave and then not get a valid response

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -352,7 +356,11 @@ public function query()
             $logger->startQuery($args[0]);
         }
 
-        $statement = $this->_conn->query(...$args);
+        try {
+            $statement = $this->_conn->query(...$args);
+        } catch (Throwable $e) {
```
Ocramius: Same as above: we upcast these, so if we need to further wrap it there's something else going on

Ocramius (changes requested):

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
sidz: I follow the same idea from the regular Connection class https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Connection.php#L1011-L1015

lib/Doctrine/DBAL/Connection.php
sidz: we may stay with `DBALException` only for case when we will catch `Throwable` and convert them to the DBALException in the `query` method.

https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Connection.php#L1011-L1015

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
sidz: will revert

sidz:
is the latest build random? oO

morozov:
@sidz no, it just errored. I restarted the needed jobs.

sidz:
@morozov doesn't help :(

morozov:
@sidz should be related to https://www.traviscistatus.com/incidents/k35bqcmm5274.

morozov:
@sidz please try rebasing on the latest master.

sidz:
@morozov done

morozov:
@sidz the Travis CI build is green but AppVeyor is all red. I've restarted it to see if it's something accidental.

sidz:
@Ocramius ping for re-review

morozov:
Irrelevant as of https://github.com/doctrine/dbal/pull/4128.

PR #3315, opened 2018-10-06
Got rid of func_num_args() in AbstractPlatform::getDummySelectSQL()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

While working on https://github.com/doctrine/dbal/pull/3109, we agreed that the method signature couldn't be changed due to BC concerns. Now the method signature is explicit.


Ocramius (approved):
:+1:

PR #3317, opened 2018-10-06
Implemented proper escaping of string literals in platforms and schema managers

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

Some SQL queries used internally in DBAL wrap string literals in quotes w/o doing proper escaping. While it doubtfully can be considered a security issue on its own, it still makes sense to fix them at least in order to set a proper example of writing SQL queries.


Ocramius (approved):
LGTM :+1:

Majkl578:
WIP? 😁

morozov:
@Majkl578 It was actually done. Thanks, @Ocramius.

PR #3331, opened 2018-10-25
Fetch all should use the driver statement's fetchAll method

michaelcullum:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

At the moment it just loops over calling fetch instead of calling fetchAll on the ResultStatement which may mean losing performance gains implemented in fetchAll.


Ocramius (approved):
LGTM :+1:

Ocramius:
Thanks @michaelcullum!

Issue #3332, opened 2018-10-29
MsSQL Windows Authentication 


odinsey:
In dbal/lib/Doctrine/DBAL/Driver/SQLSrv/Driver.php:34

For sqlsrv_connect, if any UID nor PWD are provided in $driverOptions array, it will use Windows Authentication
BUT there are always those keys even if $username and $password are null.

Please add those keys to $driverOptions only if values are not null.

odinsey:
is doctrine dead ?

AlessandroMinoccheri:
Hi @odinsey, it could be amazing if you create a pull request about it because you know the problem and the solution. After discuss it with the community!
Try to create the pull request please

odinsey:
Ok it's done ;) please be gentle, it's my first public pull request.

Majkl578:
#3333 has been merged.

PR #3333, opened 2018-10-31
Allow windows (userless/passwordless) authentication for SQL Server

odinsey:
Adding Windows Authentication compatibility according to http://php.net/manual/en/function.sqlsrv-connect.php
"By default, the connection is attempted using Windows Authentication. To connect using SQL Server Authentication, include "UID" and "PWD" in the connection options array."

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3332 

#### Summary

Simply don't provide useless drivers options.
because [] != ['UID'=>null, 'PWD'=>null]

Thx

### DIFF
--- a/lib/Doctrine/DBAL/Driver/SQLSrv/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/SQLSrv/Driver.php
@@ -31,8 +31,13 @@ class Driver extends AbstractSQLServerDriver
             $driverOptions['CharacterSet'] = $params['charset'];
         }
 
-        $driverOptions['UID'] = $username;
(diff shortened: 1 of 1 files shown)


Ocramius (changes requested):
Is there a way to verify this behavior, somehow? Comparing two `SQLSrvConnection` instances produced by the `Driver`?

odinsey:
To verify this behavior you need at least Active Directory and Kerberos

Ocramius (approved):
LGTM :+1: 

@morozov do you have any idea if this can be verified somehow?

Issue #3334, opened 2018-10-31
DBAL 3.0 : improvements to the Driver\Connection interface


BenMorel:
I can see on the `develop` branch that the `PDOConnection` class now wraps `PDO` instead of inheriting from it, which is a very nice improvement, as the `Driver\Connection` class can now evolve away from the PDO interface! 👍 

Building on this, I would like to suggest a few changes to this interface:

## Stop returning false on error

Pretty much all PDO methods (`query()`, `beginTransaction()`, ...) can return `false` on error. This is actually only true, as far as I know, when using the `PDO::ERRMODE_SILENT` or `PDO::ERRMODE_WARNING` modes.

Because Doctrine forces `PDO::ERRMODE_EXCEPTION`, it makes no sense IMO to return false on error in any method; we'd rather enforce always throwing exceptions when something goes wrong, which would make the interface consistent with the methods that already guarantee to throw on error, such as `prepare()` and `query()`.

So I would make the following methods return `void` and throw an exception on failure:

- `beginTransaction()`
- `commit()`
- `rollBack()`

As it stands, these methods in the `PDOConnection` can never return `false`, they always return `true` or throw an exception, so the `bool` return type makes no sense in this driver.

Furthermore, I can see that many implementations actually don't respect the current contract, and don't return any value in these methods ([DBAL\Connection](https://github.com/doctrine/dbal/blob/develop/lib/Doctrine/DBAL/Connection.php#L1195-L1212), [MasterSlaveConnection](https://github.com/doctrine/dbal/blob/develop/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php#L232-L237), [DB2Connection](https://github.com/doctrine/dbal/blob/develop/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php#L137-L140), ...), so this change makes even more sense.

## Remove the errorCode() and errorInfo() methods

If we're always going to throw exceptions on error, then the error code and error info should be accessible from the exception. Therefore, I think it's redundant to offer this information in the Connection interface.

## Strongly type wherever possible

- `public function quote(string $input, int $type = ParameterType::STRING) : ?string`
  <s>As far as I can see, the input is always a string</s> not sure about this, it looks like we're accepting other types here. Is this really needed? Can't we just use `quote(string $input) : string`?
[...]

PR #3154, opened 2018-05-17
Statement objects should throw exception in exceptional cases

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | #3005

#### Summary

TBD

PR #3335, opened 2018-10-31
DBAL 3.0: suggestions to improve the Driver\Connection interface

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | #3334 

#### Summary

This PR proposes a first batch of changes to the `Driver\Connection` interface, suggested in #3334:

## Return `void` in `beginTransaction()`, `commit()`, `rollBack()`

Takeaway from the issue:

> Because Doctrine forces PDO::ERRMODE_EXCEPTION, it makes no sense IMO to return false on error in any method; we'd rather enforce always throwing exceptions when something goes wrong, which would make the interface consistent with the methods that already guarantee to throw on error, such as prepare() and query().

> As it stands, these methods in the PDOConnection can never return false, they always return true or throw an exception, so the bool return type makes no sense in this driver.

> Furthermore, I can see that many implementations actually don't respect the current contract, and don't return any value in these methods (DBAL\Connection, MasterSlaveConnection, DB2Connection, ...), so this change makes even more sense.

# Type-hint parameters and return types

- `beginTransaction()`, `commit()`, `rollBack()` as described above
- `lastInsertId()` that always receives a `string|null` and always returns `string`
- `quote()` that should always return `string` or throw an exception (?)

**These changes do not break any tests.**


morozov:
@BenMorel the patch looks reasonable code-wise. Are you planning on fixing [failures on Oracle](https://app.continuousphp.com/git-hub/doctrine/dbal/build/d468de9a-3417-492c-b0cd-4e5375c11278/logs/3f1b118b-70c2-478e-8e05-07f30e6b3ab9#line-76)? I'd like to get this in before I finish my project-wide type changes for `3.0`.

BenMorel:
@morozov Oracle fixed in 76fd95f8e81645f984edb26d1d41963602504827.

The first failure was that the test expected `lastInsertId()` to return an `int`, where we now always return `string`:

```php
self::assertSame(1, $this->driverConnection->lastInsertId($sequence));
```

The second failure, however, expected `false` when we also return string:

```php
self::assertFalse($this->connection->lastInsertId(null));
```

If fixed it with `assertSame('', ...`, but should we document the fact that is returns an empty string when no sequence name is given? Or should we throw an exception in this case?

morozov:
> The first failure was that the test expected `lastInsertId()` to return an `int`, where we now always return `string`:
> 
> ```php
> self::assertSame(1, $this->driverConnection->lastInsertId($sequence));
> ```

Makes sense.

> The second failure, however, expected `false` when we also return string:
> 
> ```php
> self::assertFalse($this->connection->lastInsertId(null));
> ```
> 
> If fixed it with `assertSame('', ...`, but should we document the fact that is returns an empty string when no sequence name is given? Or should we throw an exception in this case?

It should be an exception. In fact, previously `false` indicated an exceptional situation (same as in PDO in a non-exception error mode). Please make sure that the new code doesn't rely on strict types disabled. For instance, currently `OCI8Connection::lastInsertId()` declared as returning a `string` but returns an `int`:
https://github.com/doctrine/dbal/blob/5da6be5a8b6f415595ae917b5f8f2b4930da773a/lib/Doctrine/DBAL/Driver/OCI8/OCI8Connection.php#L142-L156

Please rebase on the latest `develop` to fix the failure on IBM DB2 and see if there are any PHPStan and PHP_CodeSniffer failures.

BenMorel:
> It should be an exception. In fact, previously false indicated an exceptional situation (same as in PDO in a non-exception error mode).

OK, which kind of exception should it throw? A `DriverException`, or a `BadMethodCallException` because as I understand it, an exception can only be thrown when calling the method without a parameter, on platforms that require it?

> Please make sure that the new code doesn't rely on strict types disabled.

Fully agree. Should we actually plan to use `strict_types` everywhere in 3.0? I'm sure there are plently of other places where implicit conversion happens without us knowing. I'd be happy to file another PR for this.

> For instance, currently OCI8Connection::lastInsertId() declared as returning a string but returns an int

✔️ This one is fixed.

> Please rebase on the latest develop to fix the failure on IBM DB2 and see if there are any PHPStan and PHP_CodeSniffer failures.

- ✔️ rebased on the latest develop
- ✔️ fixed PHPStan failures

About PHPCS failures, unfortunately it does not seem to recognize the existing `{@inheritdoc}` docblocks, and is full of errors:

> Parameter $name has null default value, but is not marked as nullable.

Is there a way to make it understand `{@inheritdoc}`? Or do you want me to copy/paste the actual docblock from the parent class/interface instead?

morozov:
> OK, which kind of exception should it throw?

`OCI8Exception` should be fine.

> Should we actually plan to use `strict_types` everywhere in 3.0? I'm sure there are plently of other places where implicit conversion happens without us knowing. I'd be happy to file another PR for this.

I meant a due-dilligence visual check. to prevent new inconsistencies. We'll introduce `strict_types` across the codebase after https://github.com/doctrine/dbal/pull/3348.

>Is there a way to make it understand `{@inheritdoc}`? Or do you want me to copy/paste the actual docblock from the parent class/interface instead?

`{@inheritdoc}` should work just fine. Not sure how exactly it's handled but it worked for me a lot of times. Please don't copy/paste doc-blocks.

morozov:
> Parameter $name has null default value, but is not marked as nullable.

It means that the parameter should be declared as `?string $name = null`, not `string $name = null`. You can run `vendor/bin/phpcbf` and fix all issues automatically.

BenMorel:
> It means that the parameter should be declared as ?string $name = null, not string $name = null. You can run vendor/bin/phpcbf and fix all issues automatically.

Ah, gotcha. I got somehow convinced that this error was related to improper documentation. Fixed 👍 

In `Driver\Connection`, I prefered to add explicit descriptions of parameters and return types, rather than removing the existing `@param` and `@return` annotations. I like very thoroughly documented interfaces, even if it feels redundant sometimes.

Oh, and I explicitly added `@throws DriverException` on `Driver\Connection::lastInsertId()`, now that OCI8 throws such an exception when using a null sequence name.

This is a first step towards [my suggestion regarding exception handling in #3334](https://github.com/doctrine/dbal/issues/3334#issuecomment-434680396).

Anyway, I'm done on this PR now, do you see anything else?

morozov:
> In `Driver\Connection`, I prefered to add explicit descriptions of parameters and return types, rather than removing the existing `@param` and `@return` annotations. I like very thoroughly documented interfaces, even if it feels redundant sometimes.

You're misinterpreting the standard. The standard forbids _useless_ annotations which only contain the variable type which is already specified in the method signature and not containing a description. Meaningful descriptions are more than welcome.

BenMorel:
> You're misinterpreting the standard. The standard forbids useless annotations which only contain the variable type which is already specified in the method signature and not containing a description. Meaningful descriptions are more than welcome.

I got that, and I had 2 choices: removing the annotation, or adding a description to it. I chose the second option!

morozov (changes requested):

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
```diff
@@ -25,13 +25,13 @@ public function lastInsertId($name = null)
         $stmt = $this->prepare('SELECT CONVERT(VARCHAR(MAX), current_value) FROM sys.sequences WHERE name = ?');
         $stmt->execute([$name]);
 
-        return $stmt->fetchColumn();
+        return (string) $stmt->fetchColumn();
```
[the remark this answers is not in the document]
BenMorel: Indeed!

I have a problem regarding exceptions, though: I have documented `Driver\Connection::lastInsertId()` to throw a `Driver\DriverException` on error, but the only current implementations of this interface are specific to other drivers: `MysqliException`, `OCI8Exception` etc. I cannot use `Driver\PDOException`, which even though it implements the interface, expects a native `\PDOException` in its constructor, and I can only pass an error message string here.

One way to make things right, is to create a `PdoSqlSrvException` class implementing `Driver\DriverException`, but I'm wondering: why is `DriverException` an interface, and why are there several driver-specific implementations?

I cannot think of a use case where someone would want to catch an `OCI8Exception` specifically, for example. You're using a `Driver\Connection` **interface**, that tells you that methods throw a `Driver\DriverException`. Why would you want to catch a subclass of that?

I guess the primary reason why `Driver\DriverException` was an interface, is that `Driver\PDOException` needed at some point to extend `PDOException`, and at the same time be an instanceof `Driver\PDOException`. This is not the case anymore: `PDOConnection` does not extend `\PDO` anymore, so I see no reason to keep `Driver\PDOException` extending `\PDOException`.

As such, I would suggest to:

- drop all driver-specific exception classes, including `Driver\PDOException`
- make `Driver\DriverException` an instantiable class instead
- make all drivers throw `DriverException`s directly

What do you think?

I you're happy with the concept but think this is too much for this PR, I can create a `PDOSqlSrvException` class for now, and after this PR is merged, create another PR to deal with exceptions.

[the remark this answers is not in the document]
morozov: > I cannot use `Driver\PDOException`, which even though it implements the interface, expects a native `\PDOException` in its constructor, and I can only pass an error message string here.

I'm experiencing the same problem in https://github.com/doctrine/dbal/pull/3154.

> One way to make things right, is to create a `PdoSqlSrvException` class implementing `Driver\DriverException`, but I'm wondering: why is `DriverException` an interface, and why are there several driver-specific implementations?
> 
> I cannot think of a use case where someone would want to catch an `OCI8Exception` specifically, for example. You're using a `Driver\Connection` **interface**, that tells you that methods throw a `Driver\DriverException`. Why would you want to catch a subclass of that?

You wouldn't want to catch a specific implementation of an interface but as long as it's an interface, implementations can be independent on each other. E.g. have methods with the same name but different signatures like `fromError(array $error)` and `fromError(PDOException $e)`.

> As such, I would suggest to:
> 
> * drop all driver-specific exception classes, including `Driver\PDOException`
> * make `Driver\DriverException` an instantiable class instead
> * make all drivers throw `DriverException`s directly
> 
> What do you think?

I think we need to leave the interface for now. There is an ongoing work https://github.com/doctrine/dbal/pull/3131 around reorganizing exceptions, so I'd like us to refrain from introducing any major changes.

If it's only the case for PDO (I assume all other driver-specific classes can be used), can we use something like:
```php
throw new class ($message) extends \Doctrine\DBAL\Driver\AbstractDriverException {};
```
It should satisfy the interface and let us move forward. It will be cleaned up later once have a better understanding of the approach.

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereConnection.php
```diff
@@ -137,13 +133,13 @@ public function getServerVersion()
     /**
      * {@inheritdoc}
      */
-    public function lastInsertId($name = null)
+    public function lastInsertId(?string $name = null) : string
     {
         if ($name === null) {
-            return sasql_insert_id($this->connection);
+            return (string) sasql_insert_id($this->connection);
```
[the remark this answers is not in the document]
BenMorel: Could you please clarify what the method should do when no sequence name is given, and the last statement was not an INSERT, or did not target a table with an autoincrement column?

`sasql_insert_id()` returns `0` in this case, `mysqli_insert_id()` as well, I guess PDO does the same, while SQL Server's `@@IDENTITY` returns `NULL`

Should we thrown an exception?
Or return an empty string `''`?
Or return `'0'`?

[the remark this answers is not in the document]
morozov: We should throw an exception. Otherwise, the client will have to explicitly check for an empty value.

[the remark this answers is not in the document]
BenMorel: Makes sense. Done!

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereConnection.php
```diff
@@ -137,13 +133,13 @@ public function getServerVersion()
...
         }
 
-        return $this->query('SELECT ' . $name . '.CURRVAL')->fetchColumn();
+        return (string) $this->query('SELECT ' . $name . '.CURRVAL')->fetchColumn();
```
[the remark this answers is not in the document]
BenMorel: Done!

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php
```diff
@@ -129,13 +129,13 @@ public function lastInsertId($name = null)
             $stmt = $this->query('SELECT @@IDENTITY');
         }
 
-        return $stmt->fetchColumn();
+        return (string) $stmt->fetchColumn();
```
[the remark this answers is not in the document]
BenMorel: Done!

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
```diff
@@ -189,7 +190,10 @@ public function testLastInsertIdNoSequenceGiven()
             $this->markTestSkipped("Test only works consistently on platforms that support sequences and don't support identity columns.");
         }
 
-        self::assertFalse($this->connection->lastInsertId(null));
+        $this->expectException(OCI8Exception::class);
```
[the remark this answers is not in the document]
BenMorel: Don't know what I had in mind. Fixed.

lib/Doctrine/DBAL/Driver/Connection.php
```diff
@@ -15,62 +15,72 @@
...
      * @throws DBALException
      */
     public function query(string $sql) : ResultStatement;
 
     /**
      * Quotes a string for use in a query.
      *
-     * @param mixed $input
-     * @param int   $type
+     * If the driver does not support quoting, an exception is thrown.
```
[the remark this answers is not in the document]
BenMorel: We don't have any. I added this sentence because `PDO::quote()` [can return false](http://php.net/manual/en/pdo.quote.php) *"if the driver does not support quoting in this way"*. 
I have no idea what this means. Please check #3334 where I was discussing `Connection::quote()`, and was suggesting that we accept `string` only, I guess this would solve the problem as every driver should indeed be able to quote strings, and I cannot see a value in quoting other types. Then we would have no reason to say that it can throw an exception.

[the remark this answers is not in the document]
morozov: The confusion comes from the method name. I think in the broadest sense, the purpose of `quote()` is to take a value and convert it into a SQL expression representing that value. Hence, the return type of `string` makes the most sense. Given the purpose, any value is acceptable (e.g. `(int) 0 => '0'`, `null => 'NULL'`). If `quote()` doesn't do that, then the client side code would have to handle everything else than strings which is tedious.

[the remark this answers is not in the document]
morozov: Regarding the "if the driver does not support quoting", I'd remove it.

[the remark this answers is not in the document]
BenMorel: I just tested `PDO::quote()`'s implementation on both sqlite and mysql:

```php
echo $pdo->quote(null); // ''
echo $pdo->quote(null, PDO::PARAM_NULL); // ''
```

They all return an empty single-quoted string. For other types (int, float), this is just basically adding single quotes around the value type-casted to string.

Basically, what the PDO implementation seems to be doing is type-casting the input to a string, then applying driver-specific quoting to that string (actually, `PDO::quote()` [explicitly says](http://php.net/manual/en/pdo.quote.php) that the input type is `string`). This is also what DBAL implementations of `Driver\Connection::quote()` currently do.

So even though what you said above makes sense, it is in all likelihood not how PDO works, and it is not how the current DBAL implementation of `quote()` works. So it looks like nobody is currently relying on this behaviour.

As such, I would still strongly suggest that the API becomes:

```php
public function quote(string $string) : string;
```

This would be much simpler and remove all confusion around this topic.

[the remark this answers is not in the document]
morozov: Indeed. Out of the supported PDO drivers, only [`mysql`](https://github.com/php/php-src/blob/0f7f1498be549f5988bf9d8150f35bedf70741c5/ext/pdo_pgsql/pgsql_driver.c#L334) and [`pgsql`](https://github.com/php/php-src/blob/8d3f8ca12a0b00f2a74a27424790222536235502/ext/pdo_mysql/mysql_driver.c#L306-L311) implement _some_ type-specific logic but it's now what we support in DBAL.

Let's update the API to the suggested about.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -814,18 +814,15 @@ public function quoteIdentifier($str)
...
      */
-    public function quote($input, $type = ParameterType::STRING) : string
+    public function quote(string $input) : string
     {
         $this->connect();
 
-        [$value, $bindingType] = $this->getBindingInfo($input, $type);
-
-        return $this->_conn->quote($value, $bindingType);
+        return $this->_conn->quote($input);
```
BenMorel: `DBAL\Connection` was using `getBindingInfo()` in `quote()`, like it was doing in `resolveParams()`, `executeQuery()` and `executeUpdate()`. But it looks like no code was relying on this "feature", as [no tests are broken](https://travis-ci.org/doctrine/dbal/builds/456856337). `getBindingInfo()` is, AFAICS, only actually used for binding parameters to prepared statements, so this shouldn't be a problem.

lib/Doctrine/DBAL/Schema/DB2SchemaManager.php
```diff
@@ -25,7 +25,7 @@ class DB2SchemaManager extends AbstractSchemaManager
     public function listTableNames()
     {
         $sql  = $this->_platform->getListTablesSQL();
-        $sql .= ' AND CREATOR = UPPER(' . $this->_conn->quote($this->_conn->getUsername()) . ')';
+        $sql .= ' AND CREATOR = UPPER(' . $this->_conn->quote((string) $this->_conn->getUsername()) . ')';
```
BenMorel: Without this explicit typecast, this was failing with a *"TypeError: Argument must be of the type string, null given"* when the connection username is null. I chose to just typecast to `string` and not do something like `AND CREATOR IS NULL` in this case, for 2 reasons:

- first of all because I don't know DB2 so I have no idea if this is a nullable column;
- and perhaps more importantly, because that was what `quote()` was doing so far: type-casting `null` to an empty string, so at least we're not changing the current behaviour.

It could be interesting to get someone with some DB2 knowledge to check that this behaviour is the proper one, though.

lib/Doctrine/DBAL/Driver/Connection.php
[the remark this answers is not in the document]
BenMorel: Cool, done in e6dcadb. I've inline-commented on 2 gotchas, though.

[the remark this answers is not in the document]
BenMorel: 2 other gotchas, that I can't inline-comment on: [static analysis](https://travis-ci.org/doctrine/dbal/jobs/456856374) showed 2 `quote()` calls with the second parameter in use:

- [ExpressionBuilder::literal()](https://github.com/doctrine/dbal/blob/e6dcadb6f4d099d297527d81617d132e28f1713a/lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php#L296)
- [SQLAzureShardManager::splitFederation()](https://github.com/doctrine/dbal/blob/e6dcadb6f4d099d297527d81617d132e28f1713a/lib/Doctrine/DBAL/Sharding/SQLAzure/SQLAzureShardManager.php#L211)

These methods are not unit-tested, and not used anywhere in the code.

Because of the current `quote()` behaviour discussed above, I think we can safely rewrite them to not use the second parameter. What do you think?

Also, it would be nice to check if these functions are actually useful, or could be removed from the codebase altogether.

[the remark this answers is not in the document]
morozov: > I think we can safely rewrite them to not use the second parameter. What do you think?

I agree. Especially in the second case, the distribution type is always a string.

> Also, it would be nice to check if these functions are actually useful, or could be removed from the codebase altogether.

Let's leave them for now. I haven't used `ExpressionBuilder::literal()` myself but it seems to make sense. Just remove the second argument from the call.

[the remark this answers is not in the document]
BenMorel: Done! Notes:

- I changed the parameter type to `string` in both functions, as any other type does not make sense anymore
- I removed the `$type` parameter from `ExpressionBuilder::literal()` as it's unused now

lib/Doctrine/DBAL/Schema/DB2SchemaManager.php
morozov: Regardless of whether it's a nullable column, a condition like `WHERE CREATOR = UPPER(NULL)` will always yield `NULL`. I can check what's going on later, but this `(string)` casting doesn't look good.

lib/Doctrine/DBAL/Driver/Connection.php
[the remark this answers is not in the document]
morozov: :+1:

lib/Doctrine/DBAL/Connection.php
morozov: Looks legit to me. I have an impression that earlier, DBAL unnecessarily combined two ideas: quoting based on the type and mixing binding types with data types. And this is the result.

lib/Doctrine/DBAL/Schema/DB2SchemaManager.php
morozov: Still, it's the same as suppressing an error with the `@` operator. Most likely, the `WHERE CREATOR = ''` condition won't yield any rows. Therefore, what's the point in running this query at all?

BenMorel: I honestly don't know, that's why I brought this to your attention in the first place. I just wanted to emphasize that my patch doesn't change the current behaviour.

BenMorel: Note that the [comment right above](https://github.com/doctrine/dbal/blob/e6dcadb6f4d099d297527d81617d132e28f1713a/lib/Doctrine/DBAL/Schema/DB2SchemaManager.php#L22) says:

> Apparently creator is the schema not the user who created it:

followed by a broken link to ibm.com.

So it looks like this code does not do what was intended in the first place anyway?

morozov: > I just wanted to emphasize that my patch doesn't change the current behaviour.

It doesn't but it violates the purpose of adding explicit types: instead of addressing the issue revealed by the type system, it hides it.

According to what I see by running `SELECT NAME, CREATOR FROM SYSIBM.SYSTABLES WHERE TYPE = 'T'`, `CREATOR` is never empty defaulting to `SYSIBM.SYSTABLES` but the username may be empty if some different authentication schema is used.

Similarly to what we do with MySQL,
https://github.com/doctrine/dbal/blob/a7cffda89e6154cf75502f250d0120523e73fac0/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L392
we can default it to the [`CURRENT USER`](https://www.ibm.com/support/knowledgecenter/en/SSEPGG_9.7.0/com.ibm.db2.luw.sql.ref.doc/doc/r0011138.html) expression. Something like:
```php
if ($user !== null) {
    $sql = 'UPPER(' . $this->_conn->quote((string) $this->_conn->getUsername() . ')';
} else {
    $sql = 'CURRENT USER';
}
```

BenMorel: I totally trust you on this one! Fixed in e401695.

BenMorel:
@morozov Do you see any other issue with this PR?

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereConnection.php
```diff
@@ -137,13 +130,29 @@ public function getServerVersion()
...
      */
-    public function lastInsertId($name = null)
+    public function lastInsertId(?string $name = null) : string
     {
         if ($name === null) {
-            return sasql_insert_id($this->connection);
+            $result = sasql_insert_id($this->connection);
+
+            if ($result === false) {
+                throw new SQLAnywhereException('The connection is not valid.');
```
morozov: The connection is established in the constructor. If there's a run-time error, it shouldn't be hidden but instead reflected in the exception.

lib/Doctrine/DBAL/Driver/Connection.php
```diff
@@ -15,62 +14,80 @@
...
+     * that was inserted into the database.
+     *
+     * If a sequence name was specified, lastInsertId() returns a string representing the last value retrieved from the
+     * specified sequence object.
      *
-     * @return string
+     * This method throws a DriverException if a value cannot be returned, in particular when:
+     *
+     * - no sequence name was provided, but the platform requires one, or the last statement did not return an identity;
+     * - a sequence name was provided, but the platform does not support sequences, or no such sequence exists.
```
morozov: I'd make sense to split the parts separated by "or" into different items.

lib/Doctrine/DBAL/Driver/Connection.php
```diff
@@ -15,62 +14,80 @@
...
+     * If a sequence name was not specified, lastInsertId() returns a string representing the row ID of the last row
+     * that was inserted into the database.
+     *
+     * If a sequence name was specified, lastInsertId() returns a string representing the last value retrieved from the
+     * specified sequence object.
      *
-     * @return string
+     * This method throws a DriverException if a value cannot be returned, in particular when:
+     *
+     * - no sequence name was provided, but the platform requires one, or the last statement did not return an identity;
```
morozov: > but the platform requires one

It's a driver-specific API, so "driver" seems more suitable here than "platform".

lib/Doctrine/DBAL/Driver/Connection.php
morozov: Some drivers (PDO OCI) don't support this API at all. In this case, we'll throw an exception as well.

lib/Doctrine/DBAL/Driver/Connection.php
morozov: > the last statement did not return an identity

Do we really want an exception here? In https://github.com/doctrine/dbal/pull/3335#discussion_r234417100, I answered "exception" to all cases without having noticed that the question contained multiple scenarios. The current implementation of this case doesn't seem to be consistent across drivers (e.g. compare SQL Anywhere and IBM DB2). Which means it's not tested and not documented.

I feel like the last one is not an exceptional situation, but a candidate for `null`. What do you think?

morozov (commented):
The more I deal with `Connection::lastInsertId()` (see https://github.com/doctrine/dbal/pull/3074), the less I like it. I think a significant part of the confusion comes from the fact that the calls with and without the `$name` are targeted at the same method but have totally different semantics (named sequences and identity columns). This confusing API was originally borrowed from PDO but we don't have to follow it strictrly.

What if you drop `Connection::lastInsertId()` from this pull request so that we can focus on what's already clear?

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereConnection.php
BenMorel: From the [docs](http://infocenter.sybase.com/help/index.jsp?topic=/com.sybase.infocenter.dc01776.1600/doc/html/san1357754986165.html), it looks like this method can only return an error (`false`) *"if the $conn is not valid*". So IMO the error message should just reflect that (for example if the connection was closed in the meantime).

lib/Doctrine/DBAL/Driver/Connection.php
BenMorel: I further clarified the docblocks, with 5 bullet points.

BenMorel:
You're right about `lastInsertId()`, it could be split into 2 methods (`lastInsertId()` and `getSequenceNumber()`?), this would clarify the intent a lot!

I can definitely remove `lastInsertId()` from this PR, but:
- there are a few things that would need to be ported to another PR: the clarified docs on the method itself, and the modifications to drivers (`OCI8`, `PDOSqlsrv`, ...)
- I quickly checked your other PR, and don't get the point of the `LastInsertId` object?

From what I could see by quickly browsing the (quite numerous) changes, it looks like you're attempting to retain the last insert ID, even if there were other statements executed in between. It's a feature that could create more problems than it would solve, IMHO:

As I said in a comment above, whenever I used `lastInsertId()`, I know for a fact that the previous statement generated such an ID. If I executed another statement (a `SELECT`?) in the meantime, I consider this a defect in my code and I want to know about it.

What if today I have an `INSERT` followed by a `SELECT`? Your approach will work fine, and allow me to get the last insert ID after both statements.
Then some time later, I introduce another `INSERT` somewhere after this `SELECT`. At first glance it's safe: the previous statement is a `SELECT`; and suddenly the previous code fails silently (getting the wrong ID).

Again, the more I think about it, the more I think this is an exceptional situation: either you get an ID **from the very last statement executed**, or you get an exception if it didn't yield an ID.

What do you think?

morozov:
> I quickly checked your other PR, and don't get the point of the `LastInsertId` object?

It's not the point. I mentioned that PR to show that this `Connection::lastInsertId()` API is a can of worms. We already have a history of burning days of work without producing any changes. And I don't want this API to derail the originally planned work: return `void` in `beginTransaction()`, `commit()`, `rollBack()`

> Again, the more I think about it, the more I think this is an exceptional situation: either you get an ID **from the very last statement executed**, or you get an exception if it didn't yield an ID.

I agree. I thought about it some more and also came to the conclusion that you shouldn't want just to look if there's a last inserted ID, you expect it to be there in a certain situation and be able to use it.

> What do you think?

You'll need to add a test next to `\Doctrine\Tests\DBAL\Functional\WriteTest::testLastInsertId*()` for the case when the ID is empty, and an exception is thrown.

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereConnection.php
morozov: An invalid connection may have many reasons behind it. If it was valid at the construction time but isn't anymore, I want to know the exact reason (should be available via `SQLAnywhereException::fromSQLAnywhereError($this->connection)`).

lib/Doctrine/DBAL/Driver/Connection.php
morozov: > Do you have a use case where this would actually be useful?

No.

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereConnection.php
BenMorel: I didn't know about this method, fixed!

BenMorel:
I added a test for `lastInsertId()` after a `SELECT` (not just on a fresh connection). My problem is, only mysql/mariadb consistently resets the last insert ID after a non-`INSERT` statement.

AFAICS, on other platforms there is no clean way to check whether a non-zero value comes from the last statement, or from a statement before (we'd have to parse every query to check if it's an INSERT, which is a huge can of worms).

I'm afraid we might have to relax the `lastInsertId()` description to:

> If the last statement was not an `INSERT` to an autoincrement column, this method *may* return an ID from a previous statement. Do not rely on this behaviour which is driver-dependent: always use `lastInsertId()` right after executing an `INSERT` statement.

Do you have a better idea, @morozov?

morozov:
> Do not rely on this behaviour which is driver-dependent: always use `lastInsertId()` right after executing an `INSERT` statement.

This is exactly how it should be used, which is the reason why #3074 and its predecessors were fixed as "Won't fix".

lib/Doctrine/DBAL/Driver/Connection.php
```diff
@@ -15,62 +14,87 @@
...
     public function exec(string $statement) : int;
 
     /**
      * Returns the ID of the last inserted row or sequence value.
      *
-     * @param string|null $name
+     * If a sequence name was not specified, lastInsertId() returns a string representing the row ID of the last row
+     * that was inserted into the database.
+     *
+     * If a sequence name was specified, lastInsertId() returns a string representing the last value retrieved from the
```
morozov: > the last value

It's the current value of the sequence.

lib/Doctrine/DBAL/Driver/Connection.php
```diff
@@ -15,62 +14,87 @@
...
+     *
      * @throws DBALException
      */
     public function exec(string $statement) : int;
 
     /**
      * Returns the ID of the last inserted row or sequence value.
      *
-     * @param string|null $name
+     * If a sequence name was not specified, lastInsertId() returns a string representing the row ID of the last row
```
morozov: "Row ID" may be confused with the `ROWID` pseudocolumn some platforms like Oracle implement. It's not necessarily the "id" column, and AFAIK not every insert affects the value returned by this method.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
```diff
@@ -126,23 +121,35 @@ public function exec(string $statement) : int
...
      * {@inheritdoc}
      */
-    public function lastInsertId($name = null)
+    public function lastInsertId(?string $name = null) : string
     {
-        return db2_last_insert_id($this->conn);
+        if ($name !== null) {
+            // DB2 does not support sequences. However, PDO::lastInsertId() ignores the name parameter, and returns
+            // the last insert ID even if a sequence name is given. We expect an exception in that case.
+            throw new DB2Exception('DB2 does not support sequences.');
```
morozov: This may be misleading since IBM DB2 LUW [supports sequences](https://www.ibm.com/support/knowledgecenter/en/SSEPGG_9.7.0/com.ibm.db2.luw.sql.ref.doc/doc/r0004201.html):
> The `CREATE SEQUENCE` statement defines a sequence at the application server.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
```diff
@@ -166,35 +165,37 @@ public function exec(string $statement) : int
     /**
      * {@inheritdoc}
      */
-    public function lastInsertId($name = null)
+    public function lastInsertId(?string $name = null) : string
     {
-        return $this->conn->insert_id;
+        if ($this->conn->insert_id === 0) {
```
morozov: Let's assign the value of `$this->conn->insert_id` to a variable and use it in all cases for better atomicity.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Connection.php
```diff
@@ -153,7 +147,7 @@ public function lastInsertId($name = null)
             throw new OCI8Exception('lastInsertId failed: Query was executed but no result was returned.');
         }
 
-        return (int) $result;
+        return (string) $result;
```
morozov: Is casting to string needed? Most likely, the previously existing casting was needed since the driver returns numbers as strings which is exactly what we need.

lib/Doctrine/DBAL/Driver/PDOConnection.php
```diff
@@ -84,17 +83,30 @@ public function query(string $sql) : ResultStatement
...
+    public function lastInsertId(?string $name = null) : string
     {
-        return $this->connection->lastInsertId($name);
+        try {
+            $lastInsertId = $this->connection->lastInsertId($name);
+        } catch (\PDOException $e) {
+            throw new PDOException($e);
+        }
+
+        if ($lastInsertId === '0' || $lastInsertId === '') {
```
morozov: Is this needed since different drivers return different empty values? If so, please add a comment to the code.

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereConnection.php
```diff
@@ -137,13 +130,29 @@ public function getServerVersion()
...
         if ($name === null) {
-            return sasql_insert_id($this->connection);
+            $result = sasql_insert_id($this->connection);
+
+            if ($result === false) {
+                throw SQLAnywhereException::fromSQLAnywhereError($this->connection);
+            }
+
+            if ($result === 0) {
+                throw new SQLAnywhereException('The last insert did not affect an AUTOINCREMENT column.');
```
morozov: This message should be consistent across drivers. We can use a factory method like:

https://github.com/doctrine/dbal/blob/16ef12680c39f73cabffcf8814694ec2846ef54e/lib/Doctrine/DBAL/Schema/SchemaException.php#L24-L27

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereConnection.php
```diff
@@ -137,13 +130,29 @@ public function getServerVersion()
...
+            }
+
+            return (string) $result;
         }
 
-        return $this->query('SELECT ' . $name . '.CURRVAL')->fetchColumn();
+        $result = $this->query('SELECT ' . $name . '.CURRVAL')->fetchColumn();
+
+        if ($result === false) {
+            throw new SQLAnywhereException('No sequence with name "' . $name . '" found.');
```
morozov: Please use a factory method for consistency.

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereConnection.php
```diff
@@ -137,13 +130,29 @@ public function getServerVersion()
...
+        }
+
+        return (string) $result;
```
morozov: Not sure whether casting is needed here. No idea how to verify that. Most likely, the SQLAnywhere driver in DBAL is broken for a long time and we'll have to get rid of it.

morozov (commented):
Looks good in general. Just needs some polishing :+1:

BenMorel:
I just finished fixing tests for all drivers, polishing now! 👍

lib/Doctrine/DBAL/Driver/Connection.php
BenMorel: That's mostly a copy/paste from `PDO::lastInsertId()`. What phrasing do you suggest? Something along these lines?

> If a sequence name was not specified, `lastInsertId()` returns a string representing the value of the auto-increment field from the last row inserted into the database, if any.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
BenMorel: Damn. I checked `DB2Platform::supportsSequences()` and it returns `false`. Should this be fixed?
Anyway, I updated the comment and exception message to remove any confusion.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
```diff
@@ -126,23 +121,35 @@ public function exec(string $statement) : int
     /**
      * {@inheritdoc}
      */
-    public function lastInsertId($name = null)
+    public function lastInsertId(?string $name = null) : string
     {
-        return db2_last_insert_id($this->conn);
+        if ($name !== null) {
+            // db2_last_insert_id() ignores the name parameter, and returns the last insert ID even if a sequence name
```
morozov: > db2_last_insert_id() ignores the name parameter

I don't think it's correct. Besides the poorly designed PDO API, no other drivers support the sequence name in a call to the "last insert id" method. It should be implemented in a different way like `SELECT $name.currval FROM dual` (not part of this ticket).

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
```diff
@@ -126,23 +121,35 @@ public function exec(string $statement) : int
...
+            // is given. We expect an exception in that case.
+            throw new DB2Exception('This DB2 driver does not support sequences.');
```
morozov: It's still confusing. Drivers usually don't need support for sequences, they are accessed via SQL. "Sequences on IBM DB2 are not currently supported" would be more accurate.

lib/Doctrine/DBAL/Driver/Connection.php
BenMorel: Confusing wording indeed, courtesy PDO. Fixed.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
BenMorel: Fixed!

lib/Doctrine/DBAL/Driver/OCI8/OCI8Connection.php
BenMorel: That's hard to know without activating `strict_types` and hoping that the test suite is comprehensive enough! I would be surprised if the database itself didn't return a sequence value as an integer, however I don't know how OCI8 transports the data. Hopefully it's not like PDO which casts as `string` or not depending on a seemingly unrelated setting (namely ATTR_EMULATE_PREPARES). In doubt, does the `(string)` cast hurt?

morozov: > That's hard to know without activating `strict_types`

Not really:
```php
$stmt = oci_parse($conn, 'SELECT 1 FROM DUAL');
oci_execute($stmt);
$row = oci_fetch_assoc($stmt);

var_dump($row);

/*
array(1) {
  [1] =>
  string(1) "1"
}
*/
```
It doesn't hurt but it's not needed and looks confusing. Most DB drivers return numbers (integer and float) as strings. The reason is that sometimes, the precision a DB platforms provides (e.g. `BIGINT` on MySQL) is higher than the PHP can handle on the current OS platform (e.g. x86). That's why drivers return number as strings and let the application do the conversion if needed.

lib/Doctrine/DBAL/Driver/PDOConnection.php
BenMorel: It is indeed. Added a comment.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Connection.php
BenMorel: Fair enough. I'm never this confident as, as I said above, at least PDO can return MySQL's integers as `int` depending on a configuration setting (probably only on x64, and/or only for ints < 2^31, I haven't tested). I hope there is no such thing in OCI8; anyway, once strict_types will be enabled everywhere, and if our tests are extensive enough, we should spot any such issue later on.

Anyway, I've removed the type cast.

lib/Doctrine/DBAL/Driver/Connection.php
BenMorel: I committed the sentence above. Review welcome.

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereConnection.php
BenMorel: Done.

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereConnection.php
BenMorel: Are there no CI tests for SQLAnywhere? Should we leave the typecast, in doubt?

lib/Doctrine/DBAL/Driver/AbstractDriverException.php
```diff
@@ -36,6 +36,42 @@ public function __construct($message, $sqlState = null, $errorCode = null)
...
+     * @return AbstractDriverException
+     */
+    public static function noInsertId() : self
+    {
+        $message = 'No identity value was generated by the last statement.';
+
+        if (static::class === self::class) {
+            // WIP regarding exceptions, see:
+            // https://github.com/doctrine/dbal/pull/3335#discussion_r234381175
+            return new class($message) extends AbstractDriverException {
```
BenMorel: I had to resort to this nasty code, as I can't use a factory method on an anonymous class.
Exceptions refactoring is really needed here!

BenMorel:
> What if you drop Connection::lastInsertId() from this pull request so that we can focus on what's already clear?

`lastInsertId()` is now becoming clear, are you OK to merge it together with this PR, and once this is done, I'll open another PR to split it into 2 methods?

morozov:
> lastInsertId() is now becoming clear, are you OK to merge it together with this PR […]

Yes. I only suggested it to not hold the rest of the changes.

lib/Doctrine/DBAL/Driver/AbstractDriverException.php
BenMorel: Note that Scrutinizer wrongly reports that `static::class === self::class` is always true, I'll file a bug with them.

BenMorel:
So apart from [the type-cast on SQLAnywhere](https://github.com/doctrine/dbal/pull/3335#discussion_r236861426), that we don't know how to test (it doesn't hurt to leave it, as for sure it cannot create a bug, whereas removing it *could* create one), I think I'm done on this PR, @morozov!

Please let me know if you see anything else; I'd like to move forward with splitting `lastInsertId()` into 2 methods, and other stuff too. Thanks!

lib/Doctrine/DBAL/Driver/AbstractDriverException.php
```diff
@@ -36,6 +36,42 @@ public function __construct($message, $sqlState = null, $errorCode = null)
...
     }
 
+    /**
+     * @return AbstractDriverException
+     */
+    public static function noInsertId() : self
+    {
+        $message = 'No identity value was generated by the last statement.';
+
+        if (static::class === self::class) {
```
morozov: Is this method supposed to be called on the abstract class? I'd rather remove this block since it's a temporary solution than complicate it that much.

Another thought I had in mind is, do we really need driver-specific exceptions which are not originated in the underlying drivers? The primary reason for having `OCI8Exception`, `MysqliException`, etc. separated out is that different drivers provide different information about errors and different APIs for obtaining this information.

In the case of exceptions like this which originate in the DBAL, it's not applicable. The information is always the same.

Unlike most other use cases, where this separation is useful when a client is interested in catching only a specific type of exceptions, DBAL clients are not interested in driver-specific exceptions because the driver is abstracted out.

Another point is, an exception originated in DBAL, cannot properly implement `Doctrine\DBAL\Driver\DriverException` because it doesn't have error code and SQLSTATE by definition.

Can we move these new methods to `Doctrine\DBAL\DBALException` (even if it's an API change) and find a better place later?

BenMorel: > Is this method supposed to be called on the abstract class? I'd rather remove this block since it's a temporary solution than complicate it that much.

Yes, this is to allow calling `AbstractDriverException::noInsertId()`. There is no other solution to factorize the `noInsertId()` method as you requested it, without either making `AbstractDriverException` not abstract, or creating more driver-specific exception classes.

> Another thought I had in mind is, do we really need driver-specific exceptions which are not originated in the underlying drivers? (...) Unlike most other use cases, where this separation is useful when a client is interested in catching only a specific type of exceptions, DBAL clients are not interested in driver-specific exceptions because the driver is abstracted out.

Exactly. That's why I [suggested earlier](https://github.com/doctrine/dbal/pull/3335#discussion_r234362640) that `DriverException` might not be an interface, but a concrete class.

I get your point that subclasses allow for driver-specific factory methods such as `fromSqlSrvErrors()`, `fromPDOException()` and so on, but they provide nothing that could not be implemented in the `Driver\Connection` classes.

For example, in `SqlSrvConnection`: `throw SQLSrvException::fromSqlSrvErrors()` could be replaced with `throw $this->createExceptionFromSqlSrvErrors()`, that would create a generic `DriverException`. This removes the need from an `SQLSrvException` class altogether.

> Another point is, an exception originated in DBAL, cannot properly implement Doctrine\DBAL\Driver\DriverException because it doesn't have error code and SQLSTATE by definition.

Error code and SQLSTATE are currently optional in `DriverException`, but anyway I'm not sure why you would throw a DriverException from another place of the DBAL?

As I see it, the driver is the lowest layer of DBAL, and the rest of the DBAL is another layer built on top of it: it uses drivers and handles DriverExceptions, but never throws them.

> Can we move these new methods to Doctrine\DBAL\DBALException (even if it's an API change) and find a better place later?

`DBALException` does not implement `DriverException`, so this would not respect the contract!

What I would suggest, if you see no other issue here, is that we merge this PR as is: it provides well encapsulated changes, and has just a handful of well-documented exotic ways to throw exceptions; as soon as this is merged, I can file another PR to propose to change `interface DriverException` to `class DriverException`, and get rid of the driver-specific exceptions. This should be a simple change that will clean up the codebase, and will fix all the temporary hacks introduced here. And we can start a fresh discussion there if issues arise.

What do you think?

morozov: > For example, in `SqlSrvConnection`: `throw SQLSrvException::fromSqlSrvErrors()` could be replaced with `throw $this->createExceptionFromSqlSrvErrors()`, that would create a generic `DriverException`. This removes the need from an `SQLSrvException` class altogether.

This way, you could throw these exceptions only from connections. They should be also available to statements. That's why they are separate classes.

> `DBALException` does not implement `DriverException`, so this would not respect the contract!

What contract? `Connection::lastInsertId()` doesn't declare any thrown exception. Throwing an exception (which can happen even in master) is already vilation of the contract.

BenMorel: > This way, you could throw these exceptions only from connections. They should be also available to statements. That's why they are separate classes.

No problem if implemented as a static method on the Connection, that can be used by the Statement as well.

> What contract? Connection::lastInsertId() doesn't declare any thrown exception. 

[It does](https://github.com/doctrine/dbal/pull/3335/files#diff-7dbc9d70e253495c562b8b2d74f27957R79) now, that's the whole point of what we've been discussing so far, and the contract that the newly introduced tests check.

BenMorel: I've pushed a commit in another branch, that I will move to a PR once/if this is merged:
https://github.com/BenMorel/dbal/commit/f9e0c5a55d10a499a57444aeffaab6bd207861cc

It does exactly what I suggested:

- `AbstractDriverException` is now a concrete `DriverException`, replacing the interface
- All sub-exceptions are now gone
- All exotic `new class ... extends AbstractDriverException` are now replaced with `new DriverException`
- Factory methods from sub-exceptions are moved to a `public static` method in the Connection, that can be used from within the Connection itself and from the Statement
- **All relevant Connection methods are now explicitly documented as throwing `DriverException`**

This last point is very important: so far, most of the methods had no documented exceptions, while `query()` and `exec()` were documented as throwing `DBALException`, **a contract that was never respected**:

- `PDOConnection::query()` and `exec()` throw `Driver\PDOException`, which is a `DriverException`, not a `DBALException`
- `DB2Exception::query()` and `exec()` throw `DB2Exception`, which is a `DriverException`
- `SQLSrvException::query()` and `exec()` throw `SQLSrvException` (via `SQLSrvStatement::execute()`), which is a `DriverException`
- ... and so on.

This future PR will make these things right.

BenMorel:
@morozov If you don't see any other issue here, I'll be happy get this merged and move forward to #3367 and #3368, that are ready for review!

morozov:
The PR cannot be merged in its current state. If it requires extra rework of exceptions, it should be done first, not the other way around.

BenMorel:
I've updated #3367 to be a standalone PR, not based on this one.
It's ready for review, @morozov.

Some things are intrinsically related, though: for now methods such as `commit()` or `rollBack()` are marked both as `@return false` and `@throws DriverException`.

This may feel weird, but reflects the current reality: some drivers such as DB2 already throw exceptions here, instead of respecting the current contract.

As described in the PR description, that PR is aimed to "make things right" regarding DriverExceptions; once/if merged, the present PR will finish the job by removing the `@return false`-style methods and make other drivers respect the new contract.

If you think that changes in both PRs are too intrisically related, I can also merge the 2 PRs, but this will be a bigger change, harder to review.

morozov:
Closing as superseded by #3480 and #3568.

PR #3337, opened 2018-11-04
Fix of links in documentation

SenseException:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

During my check if link-targets in the documentation became deprecated or removed, I only found a few rst-formats for the links, that need to be fixed. Some of the linked pages are old, but are still available with their supposed content.


ostrolucky (approved):

ostrolucky:
Should we target older branch?

PR #3340, opened 2018-11-07
Distinct in the QueryBuilder

michaelcullum:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no

#### Summary

Add functionality to generate distinct queries using the DBAL query builder

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -420,10 +420,8 @@ class QueryBuilder
 
     /**
      * Returns whether or not the query object is set to return only distinct results
-     *
-     * @return bool
      */
-    public function isDistinct(): bool
+    public function isDistinct() : bool
     {
         return $this->isDistinct;
     }
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -126,6 +126,13 @@ class QueryBuilder
     private $boundCounter = 0;
 
     /**
+     * Whether or not to include a distinct flag in the SELECT clause
+     *
+     * @var bool
+     */
+    private $isDistinct = false;
+
+    /**
      * Initializes a new <tt>QueryBuilder</tt>.
      *
      * @param Connection $connection The DBAL Connection.
@@ -394,6 +401,29 @@ class QueryBuilder
     public function getMaxResults()
     {
         return $this->maxResults;
+    }
+
+    /**
+     * Sets the flag to only retrieve distinct results
+     *
+     * @return $this This QueryBuilder instance.
+     */
+    public function distinct()
+    {
+        $this->state = self::STATE_DIRTY;
+        $this->isDistinct = true;
+
+        return $this;
+    }
+
+    /**
+     * Returns whether or not the query object is set to return only distinct results
+     *
+     * @return bool
+     */
+    public function getDistinct()
+    {
+        return $this->isDistinct;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -396,6 +403,29 @@ public function getMaxResults()
         return $this->maxResults;
     }
 
+    /**
+     * Sets the flag to only retrieve distinct results
+     *
+     * @return $this This QueryBuilder instance.
+     */
+    public function distinct()
```
IonBazan: You should allow users reset the flag back to `false`. Please add `$flag = true` argument to achieve this.

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
```diff
@@ -576,6 +587,15 @@ public function testSetFirstResult()
         self::assertEquals(10, $qb->getFirstResult());
     }
 
+    public function testSetDistinct()
+    {
+        $qb = new QueryBuilder($this->conn);
+        $qb->distinct();
+
+        self::assertEquals(QueryBuilder::STATE_DIRTY, $qb->getState());
+        self::assertEquals(10, $qb->getDistinct());
```
IonBazan: Why `10`?

IonBazan (changes requested):

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
michaelcullum: Good spot, a silly typo.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -396,6 +403,31 @@ public function getMaxResults()
...
     }
 
+    /**
+     * Sets the flag to only retrieve distinct results
+     *
+     * @param bool $isDistinct Set true to add the distinct flag, set to false to remove it
+     *
+     * @return $this This QueryBuilder instance.
+     */
+    public function distinct($isDistinct = true)
```
morozov: The API would be cleaner if instead of `distinct(false)`, a developer could call `nonDistinct()`. Otherwise, it looks sort of asymmetric: `distinct()` ↔ `distinct(false)`.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -396,6 +403,31 @@ public function getMaxResults()
...
+        $this->state      = self::STATE_DIRTY;
+        $this->isDistinct = $isDistinct;
+
+        return $this;
+    }
+
+    /**
+     * Returns whether or not the query object is set to return only distinct results
+     *
+     * @return bool
```
morozov: Please provide a return type hint instead.

morozov (changes requested):
Besides the comments below, looks good to me.

lib/Doctrine/DBAL/Query/QueryBuilder.php
IonBazan: I think he is trying to make it consistent with Doctrine ORM's QueryBuilder: https://github.com/doctrine/doctrine2/blob/6dc46e78ccf4fe75c80884e7abe41b811f2204b6/lib/Doctrine/ORM/QueryBuilder.php#L765

morozov: We can make it better and improve the ORM’s builder later.

michaelcullum: I'm inclined to disagree, the main reason was for consistency with the ORM but I think the ORM makes sense.

The method is essentially a boolean setter and the most common thing it's being set to is true so the fact it defaults to true is just to make it nicer to use (`$qb->distinct()` instead of a regularly superfluous true). Assuming you don't rely on the default added for DX it is symmetric: `distinct(true)` ↔ `distinct(false)`

It also means you could pass a statement straight into the argument e.g. `$qb->distinct($this->myBooleanFlag)` should you so wish instead of `if ($this->myBooleanFlag) { $qb->distinct(); } else { $qb->nonDistinct(); }`.

Finally, DX again, a method called `nonDistinct()` is much less likely to actually be found by a developer compared to `distinct(bool $flag)`, particularly inflated by the fact that is the ORM's behaviour.

morozov: It's not a setter. It's a set of builder commands which you will call or not call depending on the flag and the previous state. If this justification was valid, then instead of `assertEquals()` and `assertNotEquals()`, we'd see something like `assert(bool $equals)` in xUnit frameworks.

Issue #2926, opened 2017-11-27
Type names with backslashes cause failing upgrades


greg0ire:
I'm planning ahead and starting to use FQCNs as custom doctrine type getName implementation, as well as registering them under that name with Type::addType. Sadly, it seems SQL comments with backslashes are somehow not matched with the desired value, and the upgrade command fails with :

```
[Doctrine\DBAL\DBALException]                                                                                                                                                               
  Unknown column type "EMA\\MD\\Doctrine\\Type\\CountryType" requested. Any Doctrine type that you use has to be registered with \Doctrine\DBAL\Types\Type::addType(). You can get a list of  
   all the known types with \Doctrine\DBAL\Types\Type::getTypesMap(). If this error occurs during database introspection then you might have forgotten to register all database types for a   
  Doctrine Type. Use AbstractPlatform#registerDoctrineTypeMapping() or have your custom types implement Type#getMappedDatabaseTypes(). If the type name is empty you might have a problem wi  
  th the cache or forgot some mapping information. 
```

This is slightly different than what I described in doctrine/doctrine2#6752 .

Here is the column comment I have:

```
| country_of_origin | character varying(255) |  not null   | extended  |         <null> | (DC2Type:EMA\\MD\\Doctrine\\Type\\CountryType)          |

```

Majkl578:
Observed this as well, possibly platform-dependent (double-)quoting issue, needs investigation.

ostrolucky:
Fixed by https://github.com/doctrine/dbal/pull/3330

PR #3342, opened 2018-11-11
#2926 Fix backslashes double escaping

ostrolucky:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | #2926 

Seems during #2442 these backslash strategies were added to several DB engines blindly, without verifying if they require such backslashes. What several people experience at least with Postgres then is that DBAL generates SQLs with duplicated backslashes and upon retrieving it from database can't match it with definition (where it's single backslashed).

I can't write backlash specific test case for every single DB operation, but I verified several operations manually (schema, foreign keys, closeActiveDatabaseConnectionsSQL...) and from that I see postgresql does not treat backslash as escape character anywhere.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -28,7 +28,6 @@ use function is_bool;
 use function is_numeric;
 use function is_string;
 use function sprintf;
-use function str_replace;
 use function strpos;
 use function strtolower;
 use function trim;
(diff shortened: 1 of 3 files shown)


morozov:
Looks like a duplicate of https://github.com/doctrine/dbal/pull/3330.

ostrolucky:
Indeed :cry: What's odd though is that this testcase passes for oracle

ostrolucky:
I don't see a good, platform agnostic way to do that, do you? SchemaManager removes doctrine type from comments, so it's empty.

morozov:
Does it have to be tested this indirectly (custom type → column comment → escaping) if we're only modifying the way how escaping works?

> I can't write backlash specific test case for every single DB operation

And you don't have to unless there's an escaping-specific case apart from just escaping tested in #3330.

ostrolucky:
Concern of this testcase isn't really about internals of escaping, but about ensuring whatever DBAL writes can read too. If it screws up the write part, at least it should also screw up read part same way, so it matches.

> And you don't have to unless there's an escaping-specific case apart from just escaping tested in #3330.

In theory yes, but it turns out indirect test passes for oracle, so there might be other things wrong.

ostrolucky:
So this is because of shared connection between tests. Comments are init only once :roll_eyes:

PR #3344, opened 2018-11-13
serialize(null) = "N;" which is not a null db value

vinise:
- not null column capability lost
- is null capability lost


Majkl578 (changes requested):
This kind of change requires a test case.

Issue #2283, opened 2016-01-07
Removed Unsigned Integer Support in SQL Server (and unsupported platforms)


scalp:
Hi,

In my opinion [PR-207](https://github.com/doctrine/dbal/pull/207) didn't remove full support in SQL Server of **UNSIGNED integer**.

Like [ISSUE-1695](https://github.com/doctrine/dbal/issues/1695) for SQLite, _Schema\Comparator_ return differences for each column with "unsigned => true" attribute.

As i see, **UNSIGNED integer** is not available from neither SQL Server, PostgreSQL, SQLite, DB2, ...

I tried to see in which way it would be possible to fix it :
1. i think the best way would be to remove "unsigned" from defaults attributes and switch it to _Schema\Column\PlatformOptions_ or a newer property like **PlatformAttributes**. The problem is that it is a lot of work and will cause BC breaks.
2. another way, would be to add a method **supportsUnsignedInteger** in _Platforms\AbstractPlatform_ and test it in _Schema\SchemaDiff_toSql_ method to don't add SQL lines if unsupported. But i've an unresolved problem with this option, if (int) type change (small int to int for example) or nullable property.
3. last way was to inject **Platform** in _Schema\Comparator_ and test support (like previous point with **$platform->supportsUnsignedInteger**) in _Schema\Comparator\diffColumn_ method.

What do you think about theses solutions ? I can make a PR for solution 2 or 3. First way is too complicated for me and maybe a little bit unreasonable.

deeky666:
@scalp I'm not sure if I get the issue here. Where is SQL Server platform even using `unsigned`? Please provide us some information on how to reproduce the actual issue. Thanks!

scalp:
Hi @deeky666, it's the same issue as [ISSUE-1695](https://github.com/doctrine/dbal/issues/1695) but for SQLServer.

Requests are finely generated without ``UNSIGNED`` attribute by DBAL but *Schema\Comparator* method always detect differences on syncing methods.

[...]

PR #3345, opened 2018-11-13
Do not add query part if only unsigned has changed

npoa-cm:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2283

#### Summary

Don't create query if only (unsupported) unsigned has changed.


morozov (changes requested):
Please add an integration test. Also, please retarget against `master`.

PR #3323, opened 2018-10-13
CI: Test against PHP 7.3 on Travis

Majkl578:
|      Q       |   A
|------------- | -----------
| Type         | improvement

Issue #1479, opened 2012-07-05
DBAL-298: [GH-166] Fixes a bug caused by unquoted reserved table name being referenced during schema creation


doctrinebot:
Jira issue originally created by user @beberlei:

This issue is created automatically through a Github pull request on behalf of klaussilveira:

  Url: https://github.com/doctrine/dbal/pull/166

Message:

When creating a schema with tables with reserved names and relationships between them, the schema creation tool failed to write valid SQL in MySQL due to unquoted table names. For example:

```
CREATE TABLE `Group` (id INT AUTO_INCREMENT NOT NULL, name VARCHAR(255) NOT NULL DEFAULT NULL, PRIMARY KEY(id)) ENGINE = InnoDB;
CREATE TABLE User (id INT AUTO_INCREMENT NOT NULL, username VARCHAR(255) NOT NULL PRIMARY KEY(id)) ENGINE = InnoDB;
CREATE TABLE UserHasGroup (user*id INT NOT NULL, group_id INT NOT NULL, INDEX IDX_617A865CA76ED395 (user_id), INDEX IDX_617A865CFE54D947 (group_id), PRIMARY KEY(user_id, group*id)) ENGINE = InnoDB;
ALTER TABLE UserHasGroup ADD CONSTRAINT FK*617A865CA76ED395 FOREIGN KEY (user*id) REFERENCES User (id);
ALTER TABLE UserHasGroup ADD CONSTRAINT FK*617A865CFE54D947 FOREIGN KEY (group*id) REFERENCES Group (id);
```

This fix creates a small function for creating quoted foreign key table names.

PR #3347, opened 2018-11-14
Travis CI tests on PHP 7.3

BenMorel:
Travis CI [nightly is now PHP 7.4-dev](https://github.com/travis-ci/travis-ci/issues/9717#issuecomment-414259993). PHP 7.3 is [now available on Travis](https://github.com/travis-ci/travis-ci/issues/9717#issuecomment-429564626).

This PR adds Travis tests on PHP 7.3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ cache:
 php:
   - 7.1
   - 7.2
+  - 7.3
   - nightly
 
 env:
@@ -62,28 +63,52 @@ jobs:
       env: DB=mysql
     - php: 7.1
       env: DB=mysqli
+    - php: 7.2
+      env: DB=sqlite
+    - php: 7.2
+      env: DB=mysql
+    - php: 7.2
+      env: DB=mysqli
+    - php: 7.3
+      env: DB=mysql
+    - php: 7.3
+      env: DB=mysqli
+    - php: nightly
+      env: DB=mysql
+    - php: nightly
+      env: DB=mysqli
 
   include:
     - stage: Test
       php: 7.1
+      env: DB=sqlite
+
+    - stage: Test
+      php: 7.2
       env: DB=sqlite COVERAGE=yes
 
     - stage: Test
-      php: 7.1
+      php: 7.2
       env: DB=mysql COVERAGE=yes
 
     - stage: Test
-      php: 7.1
+      php: 7.2
       env: DB=mysqli COVERAGE=yes
 
     - stage: Test
       php: 7.1
+      env: DB=mysql MYSQL_VERSION=5.7
+      sudo: required
+      before_script:
+        - bash ./tests/travis/install-mysql-5.7.sh
+    - stage: Test
+      php: 7.2
       env: DB=mysql MYSQL_VERSION=5.7 COVERAGE=yes
       sudo: required
       before_script:
         - bash ./tests/travis/install-mysql-5.7.sh
     - stage: Test
-      php: 7.2
+      php: 7.3
       env: DB=mysql MYSQL_VERSION=5.7
       sudo: required
       before_script:
@@ -97,12 +122,18 @@ jobs:
 
     - stage: Test
       php: 7.1
+      env: DB=mysqli MYSQL_VERSION=5.7
+      sudo: required
+      before_script:
+        - bash ./tests/travis/install-mysql-5.7.sh
+    - stage: Test
+      php: 7.2
       env: DB=mysqli MYSQL_VERSION=5.7 COVERAGE=yes
       sudo: required
       before_script:
         - bash ./tests/travis/install-mysql-5.7.sh
     - stage: Test
-      php: 7.2
+      php: 7.3
       env: DB=mysqli MYSQL_VERSION=5.7
       sudo: required
       before_script:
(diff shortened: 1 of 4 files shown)


Majkl578:
Duplicate of #3323

BenMorel:
@Majkl578 Ah, sorry for the dup. Closing this one.

Majkl578:
If you want, take my commit and you can finish what @morozov suggested, I won't have time for that for next 5 days or so.

BenMorel:
Actually your commit is based on master, while mine is based on develop (no PHP 7.1 support).

Otherwise they're pretty much equivalent. @morozov are you OK if I apply your suggestions on the commit here?

morozov:
@BenMorel as long as these are non-breaking changes, we can accept them in `master`. Sure, you can implement the suggestions from https://github.com/doctrine/dbal/pull/3323#discussion_r224969133 here and retarget the pull request.

BenMorel:
@morozov The problem is that master targets PHP 7.1+, while develop targets 7.2+.

Should I create 2 PRs, one targeting master and one targeting develop?

morozov:
@BenMorel what is the problem? Regardless of the minimal supported version, we'll support PHP 7.3 in all actively supported branches. No need for two separate pull requests. `develop` is periodically rebased on top of the latest `master`.

BenMorel:
Ah, true, rebasing develop on master should do the job, if there are no conflicts.

OK, I'll pick up @Majkl578's commit and implement your suggestion.

BenMorel:
OK, so here's what I've done:

- Retargeted the PR on `master` and picked @Majkl578's commit
- Default versions of `sqlite`, `mysql` and `mysqli` now run only on PHP 7.3 (they were previously running only on PHP 7.2, explicitly excluding 7.1, so I guess it makes sense to increment the version here)
- Kept PHP 7.3 tests only on the latest version of each platform:
  - DB=mysql MYSQL_VERSION=5.7
  - DB=mysqli MYSQL_VERSION=5.7
  - DB=mariadb MARIADB_VERSION=10.3
  - DB=mariadb.mysqli MARIADB_VERSION=10.3
  - DB=pgsql POSTGRESQL_VERSION=11.0
- Removed `sqlsrv`, `pdo_sqlsrv` and `ibm_db2` tests on PHP 7.3: the [sqlsrv](https://pecl.php.net/package-changelog.php?package=sqlsrv) and [ibm_db2](https://pecl.php.net/package-changelog.php?package=ibm_db2) PECL extensions do not yet support PHP 7.3 and do not compile

@morozov @Majkl578 What do you think?

morozov:
Looks good so far. A couple of questions:

1. While adding new builds, could you also reduce the number of nightly builds to not increase (or maybe even reduce) the number of builds in total? The pattern I want to achieve is:
   1. For the the latest stable PHP version (which is currently 7.2), test all drivers with and platform versions.
   2. All other supported PHP versions (7.1, 7.3, nightly), test each driver with the latest version of the platform.
1. Instead of removing `sqlsrv` and `pdo_sqlsrv`, could you try testing their [v5.4.0-preview](https://github.com/Microsoft/msphpsql/releases/tag/v5.4.0-preview) if it's not too difficult to build? It should work.

If the algorighm above makes sense, maybe we could have a script which takes all versions as input and generates the matrix? Just for the sake of discussion.

BenMorel:
1. Sounds like a good strategy. I'm not sure whether I feel like writing a script to generate the matrix, but I can surely give it a try.
  Note that PHP 7.3 should be GA around December 6, so maybe we can be proactive and consider this one stable right now?
2. You're right, `5.4.0preview` works for both [sqlsrv](https://travis-ci.org/doctrine/dbal/jobs/455772606) and [pdo_sqlsrv](https://travis-ci.org/doctrine/dbal/jobs/455772609)! Fixed in b85ff57 👍

morozov:
> I'm not sure whether I feel like writing a script to generate the matrix, but I can surely give it a try.

You don't have to. I can give it a try myself. It's just easier to reason about an algorithm and play with the switches when it's implemented as code.

> Note that PHP 7.3 should be GA around December 6, so maybe we can be proactive and consider this one stable right now?

We can do that but I was hoping we'd have all extensions in place by GA (at least `sqlsrv` and `pdo_sqlsrv`). Otherwise, we won't have full coverage where we're supposed to have it.

tests/travis/install-mssql-pdo_sqlsrv.sh
```diff
@@ -3,4 +3,9 @@
 set -ex
 
 echo "Installing extension"
-pecl install pdo_sqlsrv
+
+if [ "$TRAVIS_PHP_VERSION" == "7.3" ]; then
+  pecl install pdo_sqlsrv-devel
```
morozov: Can we pin it to `5.4.0-preview`? If `devel` gets broken after `5.4.0` is released but before we update this config, I don't want our builds to fail for no reason.

BenMorel:
OK, [I actually gave the build script a shot](https://github.com/doctrine/dbal/blob/051911fc0d99e1ef14b714a92e5949bbb533be87/travis-matrix-builder.php), and it works fine! The script itself is not a work of art, but it's simple enough, and it's easy to modify its variables.

The good news is: we're down from [60](https://travis-ci.org/doctrine/dbal/builds/455697439) builds to [53](https://travis-ci.org/doctrine/dbal/builds/456067665?utm_source=github_status&utm_medium=notification), including PHP 7.3 (44 without) 🎉 

---

Notes: I actually removed the base `php` and `env` variables from the .travis.yml file, as well as the `jobs`.`exclude` entries, as we're now explicitly `include`ing every single job.

The only 4 jobs that are not part of the script, and [left as is in the .travis.yml file](https://github.com/doctrine/dbal/blob/051911fc0d99e1ef14b714a92e5949bbb533be87/.travis.yml#L375-L399), are:

- the tests for composer's `--prefer-lower` and `--prefer-dist`
- the Code Quality and Coding standard jobs

As they're not like the other jobs in the build.

What do you think, @morozov?

morozov:
> The good news is: we're down from [60](https://travis-ci.org/doctrine/dbal/builds/455697439) builds to [53](https://travis-ci.org/doctrine/dbal/builds/456067665?utm_source=github_status&utm_medium=notification), including PHP 7.3 (44 without) 🎉 

This is what I was aiming for. Thank you.

Could you please remove the script itself from the pull request, put it somewhere as a gist and reference in the pull request? As I said, I wanted this script to be a reference implementation of the algorithm we agreed on, not part of the codebase (at least for now). Given the pace of the changes in the landscape and potential number of temporary exceptions, it will be hard for us to maintain that script in a working state.

Instead, please apply as little modifications to `.travis.yml` as it's reasonably possibe based on the script output.

.travis.yml
```diff
@@ -54,47 +44,38 @@ jobs:
...
-
-  exclude:
-    - php: 7.1
-      env: DB=sqlite
-    - php: 7.1
-      env: DB=mysql
-    - php: 7.1
-      env: DB=mysqli
+    - php: 7.3
+      env: DB=ibm_db2
```
morozov: It should be just excluded from the matrix.

.travis.yml
```diff
@@ -3,19 +3,9 @@ sudo: false
 dist: trusty
 
 cache:
-    directories:
-        - vendor
-        - $HOME/.composer/cache
```
morozov: Is it possible to avoid these changes?

.travis.yml
```diff
@@ -3,19 +3,9 @@ sudo: false
...
-
-php:
-  - 7.1
-  - 7.2
-  - nightly
```
morozov: Please avoid changing this if possible.

.travis.yml
```diff
@@ -54,47 +44,38 @@ jobs:
...
-      env: DB=mysql COVERAGE=yes
-
-    - stage: Test
-      php: 7.1
-      env: DB=mysqli COVERAGE=yes
-
+# BEGIN TEST BUILD MATRIX - DO NOT EDIT BELOW ▼▼▼ USE THE TRAVIS MATRIX BUILDER SCRIPT
     - stage: Test
       php: 7.1
       env: DB=mysql MYSQL_VERSION=5.7 COVERAGE=yes
```
morozov: Would it make sense to collect coverage on the latest stable PHP version where we expect all drivers to be tested against all platform versions?

BenMorel:
The problem is that it will be quite hard to rewrite the output of the script to match the (somewhat messy) current .travis.yml file. In particular, the fact that a matrix is built from the root `php` and `env` variables, then some of it is `exclude`d from the `jobs`, then new ones are `include`d again later.

At least the script keeps everything clean and predictable: no base matrix, only jobs built by the script or added by hand outside of the DO NOT EDIT section.

> Given the pace of the changes in the landscape and potential number of temporary exceptions, it will be hard for us to maintain that script in a working state.

You can still `include` and `exclude` jobs outside of the DO NOT EDIT section, so you can add temporary exceptions without touching the script at all?

.travis.yml
BenMorel: This is just an indenting change for consistency. I can revert it if you wish.

.travis.yml
BenMorel: It may start working at any time, though, once a newer `ibm_db2` PECL package is available; shouldn't we keep it as an allowed failure in the meantime?

.travis.yml
BenMorel: It does indeed! Changed to 7.2.

.travis.yml
BenMorel: I need to check if I can re-introduce these, without the matrix magically creating new jobs. At first glance this may work. What's the point, though?

morozov:
> The problem is that it will be quite hard to rewrite the output of the script to match the (somewhat messy) current .travis.yml file.

Please see https://github.com/doctrine/dbal/pull/3347#issuecomment-439573971. I didn't mean replacing the existing configuration with autogenerated. I meant generating a new matrix as a reference implementation of the algorithm we were discussing (https://github.com/doctrine/dbal/pull/3347#issuecomment-439231396).

BenMorel:
I got that, but my worries still hold. I'll give it a try, though.

BenMorel:
OK, after a lot of diffing, I finally managed to align the original .travis.yml file matrix to the matrix generated by the script. The jobs are now the same:

https://travis-ci.org/doctrine/dbal/builds/456618855

Here is how the final diff looks like, keeping as much of the original .travis.yml file as possible:

https://github.com/doctrine/dbal/pull/3347/files

---

If you want my opinion, it's a total mess, though: jobs are created by the base matrix, for example:

```
php: 7.2
env: DB=mysql
```

Then some of these jobs have to be manually excluded in the `jobs` section:

```
  exclude:
    - php: 7.2
      env: DB=mysql
```

To be re-included later on, with code coverage enabled:

```
  include:
    - stage: Test
      php: 7.2
      env: DB=mysql COVERAGE=yes
```

In summary, the base matrix:

```
php:
  - 7.1
  - 7.2
  - 7.3
  - nightly

env:
  - DB=sqlite
  - DB=mysql
  - DB=mysqli
```

Is totally useless, and the requirements are such that we should only rely on the `jobs` section, which is what the script does.

Anyway, for posterity, here is a copy of the matrix builder script:

https://gist.github.com/BenMorel/67e5d94c864af9b9ae1d3a1db8010b98

tests/travis/install-mssql-pdo_sqlsrv.sh
```diff
@@ -3,4 +3,9 @@
 set -ex
 
 echo "Installing extension"
-pecl install pdo_sqlsrv
+
+if [ "$TRAVIS_PHP_VERSION" == "7.3" ]; then
```
morozov: Should it be the same for "nightly"?

tests/travis/install-mssql-sqlsrv.sh
```diff
@@ -3,4 +3,9 @@
 set -ex
 
 echo "Installing extension"
-pecl install sqlsrv
+
+if [ "$TRAVIS_PHP_VERSION" == "7.3" ]; then
```
morozov: Should it be the same for "nightly"?

.travis.yml
```diff
@@ -403,13 +398,22 @@ jobs:
...
+      env: DB=ibm_db2 COVERAGE=yes
       sudo: required
       services:
         - docker
       before_script:
         - bash ./tests/travis/install-db2.sh
         - bash ./tests/travis/install-db2-$DB.sh
+    - stage: Test
+      php: nightly
+      env: DB=ibm_db2
```
morozov: It's not currently supported. Please remove.

morozov:
@BenMorel thank you for your hard work and patience.

> If you want my opinion, it's a total mess, though […]

I like the idea of cleaning things up but usually refactoring and changing the logic at the same time is a bad idea. I also liked the way how in the autogenerated matrix, builds were ordered by PHP version, driver and platform. This way, if something fails, one could visually see a pattern of the failure.

If you don't mind, once this is merged, please create another pull request with the refactoring.

morozov:
> Do you mean with the script-based approach above, including the script in the codebase?

No, only the changes in the YAML file.

morozov:
Judging by the [Scrutinizer failure](https://scrutinizer-ci.com/g/doctrine/dbal/inspections/ef98a244-21ec-4e2f-b56d-9347eb5db9c1), we may be not collecting all new coverage reports. In [`master`](https://travis-ci.org/doctrine/dbal/builds/457167572), we have 18 of them, while [here](https://travis-ci.org/doctrine/dbal/builds/456849295), we have 23.

Please try updating `.scrutinizer.yml` and seeing if it fixes the failure:

https://github.com/doctrine/dbal/blob/ae0365bd95782d2b94eb0b16f5d1ba5aa4f5722e/.scrutinizer.yml#L24

BenMorel:
> > > If you don't mind, once this is merged, please create another pull request with the refactoring.

> > Do you mean with the script-based approach above, including the script in the codebase?

> No, only the changes in the YAML file.

I didn't get what you mean then. Do you mean a PR with the build matrix as generated by the script (without the root `php:` and `env:` entries), but without committing the script itself?

BenMorel:
Changing `runs` in `.scrutinizer.yml` seems to have fixed the code coverage: back to [74%](https://scrutinizer-ci.com/g/doctrine/dbal/inspections/cb8b3cfc-552c-424a-8aea-4f2e02aee523) from [69%](https://scrutinizer-ci.com/g/doctrine/dbal/inspections/ef98a244-21ec-4e2f-b56d-9347eb5db9c1) before, but for some reason it's still failing:

https://scrutinizer-ci.com/g/doctrine/dbal/inspections/cb8b3cfc-552c-424a-8aea-4f2e02aee523

With no link to code coverage this time, and a "configuration has changed" warning.
Any idea what's wrong?

morozov:
My Scrutinizer extension for Chrome shows:

>  Inspection found no new issues, test coverage decreased by 0.03% to 74.42%.

Which is probably the reason for the failure.

> The configuration has changed since Scrutinizer has last run. This page might contain changes that are not related to code modifications.

It probably means that Scrutinizer cannot compare two builds in proper details because they have different configuration. We probably can ignore it since it's expected to be a one off.

Without digging too much, could you double check how it is possible that the coverage went down after the number of runs went up? Maybe we removed some well-covered code or something? Some details can be found [here](https://scrutinizer-ci.com/g/doctrine/dbal/inspections/cb8b3cfc-552c-424a-8aea-4f2e02aee523/code-structure/test-coverage).

BenMorel:
> Without digging too much, could you double check how it is possible that the coverage went down after the number of runs went up? Maybe we removed some well-covered code or something?

I checked 3 different files, and the differences are always in `switch`es:

image
becomes ↓
image

or:

image
becomes ↓
image

Always the same pattern.

As I understand it, it's a difference (bug?) in covered lines reported by a different version of the software. Probably caused by the change to running coverage on PHP 7.2. All these lines are still covered.

morozov:
@BenMorel thank you for the thorough explanation. It's most likely not a bug but internal changes in how PHP code is compiled to bytecodes. The change in the coverage is a side effect of that (e.g. like the [Xdebug issue #1479](https://bugs.xdebug.org/view.php?id=1479)).

morozov:
Please squash the commits and let's get it merged.

BenMorel:
Can't you squash the commits with 1 button press?

image

Majkl578:
We don't squash-merge using GitHub UI because GitHub doesn't create a merge commit (which is what we want) - please squash the commits manually. 😊

Majkl578:
Also since you did quite some work, you may keep your work as separate commit, or add yourself into [`Co-Authored-By`](https://help.github.com/articles/creating-a-commit-with-multiple-authors/).

morozov:
If everyone is comfortable with the way the work is attributed, I'll merge it in a couple of hours:
```
Author: Michael Moravec <me@majkl.me>
Date:   Sun Aug 26 22:04:27 2018 +0200

    CI: Test against PHP 7.3 on Travis
    
    Co-authored-by: Michael Moravec <me@majkl.me>
    Co-authored-by: Benjamin Morel <benjamin.morel@gmail.com>
```
If it's a single commit, based on the amount of work done, I think it'd be fair if it's attributed like:
```
Author: Benjamin Morel <benjamin.morel@gmail.com>
Date:   Sun Aug 26 22:04:27 2018 +0200

    CI: Test against PHP 7.3 on Travis
    
    Co-authored-by: Michael Moravec <me@majkl.me>
```

What do you say?

morozov (approved):

BenMorel:
I don't mind either way!

Ocramius (approved):

Majkl578:
> For my personal knowledge, could you please explain very briefly what's the difference between a merge commit and the squash-and-merge functionality of GitHub?

The difference is:
a) merge commit exists so it's clear where the commit comes from,
b) multiple commits are allowed,
c) GPG signatures are retained.

PR #3260, opened 2018-08-18
Enable PHPStan in tests

Majkl578:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Enabling PHPStan in tests/ at level 3, just like for lib/. :tada: 
Rewritten some Prophecy-based unit tests into PHPUnit mock system to avoid having another (unofficial) extension for Prophecy.

Replaces incomplete #3203 with no changes like (_Class_ -> _mixed_).

Currently depends on my fork of phpstan-phpunit due to https://github.com/phpstan/phpstan-phpunit/pull/28.

PR #2384, opened 2016-05-10
Fluent methods in QueryBuilder are now documented as returning `$this`

Deltachaos:
According to the discussion in #2349 i would rather suggest to document the return type for the fluent interfaces methods with `$this`.
[phpDocumentor](https://www.phpdoc.org/docs/latest/guides/types.html#keywords) recommends `$this` to be used for refer "This exact object instance, usually used to denote a fluent interface." instead of `self` "An object of the class where this type was used, if inherited it will still represent the class where it was originally defined.".

PR #3348, opened 2018-11-15
Enforce parameter and return value types in the codebase

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | none

Enabled `SlevomatCodingStandard.TypeHints.TypeHintDeclaration.Missing{Parameter,Return}TypeHint` sniffs across the codebase.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,11 @@
 # Upgrade to 3.0
 
+## BC BREAK: Changes in the `Doctrine\DBAL\Schema` API
+
+- Column precision no longer defaults to 10. The default value is NULL.
+- Asset names are no longer nullable. An empty asset name should be represented as an empty string.
+- `Doctrine\DBAL\Schema\AbstractSchemaManager::_getPortableTriggersList()` and `::_getPortableTriggerDefinition()` have been removed.
+
 ## BC BREAK: Changes in the `Doctrine\DBAL\Event` API
 
 - `SchemaAlterTableAddColumnEventArgs::addSql()` and the same method in other `SchemaEventArgs`-based classes no longer accept an array of SQL statements. They accept a variadic string.
--- a/lib/Doctrine/DBAL/Schema/AbstractAsset.php
+++ b/lib/Doctrine/DBAL/Schema/AbstractAsset.php
@@ -25,7 +25,7 @@ use function substr;
 abstract class AbstractAsset
 {
     /** @var string */
-    protected $_name;
+    protected $_name = '';
 
     /**
      * Namespace of the asset. If none isset the default namespace is assumed.
@@ -138,7 +138,7 @@ abstract class AbstractAsset
             return $this->_namespace . '.' . $this->_name;
         }
 
-        return $this->_name ?? '';
+        return $this->_name;
     }
 
     /**
@@ -167,7 +167,7 @@ abstract class AbstractAsset
      */
     protected function _generateIdentifierName(array $columnNames, string $prefix = '', int $maxSize = 30) : string
     {
-        $hash = implode('', array_map(static function ($column) {
+        $hash = implode('', array_map(static function ($column) : string {
             return dechex(crc32($column));
         }, $columnNames));
 
--- a/lib/Doctrine/DBAL/Schema/Column.php
+++ b/lib/Doctrine/DBAL/Schema/Column.php
@@ -23,7 +23,7 @@ class Column extends AbstractAsset
     protected $_length;
 
     /** @var int|null */
-    protected $_precision = 10;
+    protected $_precision;
 
     /** @var int */
     protected $_scale = 0;
@@ -106,10 +106,6 @@ class Column extends AbstractAsset
 
     public function setPrecision(?int $precision) : self
     {
-        if ($precision === null) {
-            $precision = 10; // defaults to 10 when no precision is given.
-        }
-
         $this->_precision = $precision;
 
         return $this;
@@ -195,7 +191,7 @@ class Column extends AbstractAsset
         return $this->_precision;
     }
 
-    public function getScale() : ?int
+    public function getScale() : int
     {
         return $this->_scale;
     }
--- a/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
@@ -92,7 +92,7 @@ class PostgreSqlSchemaManager extends AbstractSchemaManager
         $names = $this->getSchemaNames();
         $paths = $this->getSchemaSearchPaths();
 
-        $this->existingSchemaPaths = array_filter($paths, static function ($v) use ($names) {
+        $this->existingSchemaPaths = array_filter($paths, static function ($v) use ($names) : bool {
             return in_array($v, $names);
         });
     }
@@ -159,14 +159,6 @@ class PostgreSqlSchemaManager extends AbstractSchemaManager
             $tableForeignKey['conname'],
             ['onUpdate' => $onUpdate, 'onDelete' => $onDelete]
         );
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    protected function _getPortableTriggerDefinition(array $trigger) : string
-    {
-        return $trigger['trigger_name'];
     }
 
     /**
--- a/lib/Doctrine/DBAL/Schema/Schema.php
+++ b/lib/Doctrine/DBAL/Schema/Schema.php
@@ -408,8 +408,6 @@ class Schema extends AbstractAsset
 
     /**
      * Cloning a Schema triggers a deep clone of all related assets.
-     *
-     * @return void
      */
     public function __clone()
     {
--- a/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
@@ -223,7 +223,7 @@ abstract class AbstractSchemaManager
      *
      * @return array<int, mixed>
      */
-    protected function filterAssetNames(array $assetNames)
+    protected function filterAssetNames(array $assetNames) : array
     {
         $filter = $this->_conn->getConfiguration()->getSchemaAssetsFilter();
         if (! $filter) {
@@ -603,35 +603,6 @@ abstract class AbstractSchemaManager
     protected function getPortableNamespaceDefinition(array $namespace) : string
     {
         return array_shift($namespace);
-    }
-
-    /**
-     * @param array<int, array<int, mixed>> $triggers
-     *
-     * @return array<int, string>
-     */
-    protected function _getPortableTriggersList(array $triggers)
-    {
-        $list = [];
-        foreach ($triggers as $value) {
-            $value = $this->_getPortableTriggerDefinition($value);
-
-            if (! $value) {
-                continue;
-            }
-
-            $list[] = $value;
-        }
-
-        return $list;
-    }
-
-    /**
-     * @param array<string|int, mixed> $trigger
-     */
-    protected function _getPortableTriggerDefinition(array $trigger) : string
-    {
-        return array_shift($trigger);
     }
 
     /**
--- a/lib/Doctrine/DBAL/Schema/ForeignKeyConstraint.php
+++ b/lib/Doctrine/DBAL/Schema/ForeignKeyConstraint.php
@@ -59,14 +59,12 @@ class ForeignKeyConstraint extends AbstractAsset implements Constraint
      * @param array<int, string>   $localColumnNames   Names of the referencing table columns.
      * @param Table|string         $foreignTableName   Referenced table.
      * @param array<int, string>   $foreignColumnNames Names of the referenced table columns.
-     * @param string|null          $name               Name of the foreign key constraint.
+     * @param string               $name               Name of the foreign key constraint.
      * @param array<string, mixed> $options            Options associated with the foreign key constraint.
      */
-    public function __construct(array $localColumnNames, $foreignTableName, array $foreignColumnNames, $name = null, array $options = [])
+    public function __construct(array $localColumnNames, $foreignTableName, array $foreignColumnNames, string $name = '', array $options = [])
     {
-        if ($name !== null) {
-            $this->_setName($name);
-        }
+        $this->_setName($name);
 
         $this->_localColumnNames = $this->createIdentifierMap($localColumnNames);
 
--- a/lib/Doctrine/DBAL/Schema/SQLAnywhereSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/SQLAnywhereSchemaManager.php
@@ -45,13 +45,13 @@ class SQLAnywhereSchemaManager extends AbstractSchemaManager
         parent::dropDatabase($database);
     }
 
-    public function startDatabase(string $database)
+    public function startDatabase(string $database) : void
     {
         assert($this->_platform instanceof SQLAnywherePlatform);
         $this->_execSql($this->_platform->getStartDatabaseSQL($database));
     }
 
-    public function stopDatabase(string $database)
+    public function stopDatabase(string $database) : void
     {
         assert($this->_platform instanceof SQLAnywherePlatform);
         $this->_execSql($this->_platform->getStopDatabaseSQL($database));
--- a/lib/Doctrine/DBAL/Schema/SchemaDiff.php
+++ b/lib/Doctrine/DBAL/Schema/SchemaDiff.php
@@ -120,7 +120,7 @@ class SchemaDiff
             }
         }
 
-        if ($platform->supportsSequences() === true) {
+        if ($platform->supportsSequences()) {
             foreach ($this->changedSequences as $sequence) {
                 $sql[] = $platform->getAlterSequenceSQL($sequence);
             }
--- a/lib/Doctrine/DBAL/Schema/Table.php
+++ b/lib/Doctrine/DBAL/Schema/Table.php
@@ -460,7 +460,7 @@ class Table extends AbstractAsset
 
         $colNames = array_unique(array_merge($pkCols, $fkCols, array_keys($columns)));
 
-        uksort($columns, static function ($a, $b) use ($colNames) {
+        uksort($columns, static function ($a, $b) use ($colNames) : bool {
             return array_search($a, $colNames) >= array_search($b, $colNames);
         });
 
@@ -578,7 +578,7 @@ class Table extends AbstractAsset
      *
      * @return array<string, ForeignKeyConstraint>
      */
-    public function getForeignKeys()
+    public function getForeignKeys() : array
     {
         return $this->_fkConstraints;
     }
@@ -623,8 +623,6 @@ class Table extends AbstractAsset
 
     /**
      * Clone of a Table triggers a deep clone of all affected assets.
-     *
-     * @return void
      */
     public function __clone()
     {
--- a/lib/Doctrine/DBAL/Schema/Visitor/SchemaDiffVisitor.php
+++ b/lib/Doctrine/DBAL/Schema/Visitor/SchemaDiffVisitor.php
@@ -17,25 +17,25 @@ interface SchemaDiffVisitor
     /**
      * Visit an orphaned foreign key whose table was deleted.
      */
-    public function visitOrphanedForeignKey(ForeignKeyConstraint $foreignKey);
+    public function visitOrphanedForeignKey(ForeignKeyConstraint $foreignKey) : void;
 
     /**
      * Visit a sequence that has changed.
      */
-    public function visitChangedSequence(Sequence $sequence);
+    public function visitChangedSequence(Sequence $sequence) : void;
 
     /**
      * Visit a sequence that has been removed.
      */
-    public function visitRemovedSequence(Sequence $sequence);
+    public function visitRemovedSequence(Sequence $sequence) : void;
 
-    public function visitNewSequence(Sequence $sequence);
+    public function visitNewSequence(Sequence $sequence) : void;
 
-    public function visitNewTable(Table $table);
+    public function visitNewTable(Table $table) : void;
 
-    public function visitNewTableForeignKey(Table $table, ForeignKeyConstraint $foreignKey);
+    public function visitNewTableForeignKey(Table $table, ForeignKeyConstraint $foreignKey) : void;
 
-    public function visitRemovedTable(Table $table);
+    public function visitRemovedTable(Table $table) : void;
 
-    public function visitChangedTable(TableDiff $tableDiff);
+    public function visitChangedTable(TableDiff $tableDiff) : void;
 }
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -20,14 +20,6 @@
         <exclude name="SlevomatCodingStandard.Classes.DisallowLateStaticBindingForConstants.DisallowedLateStaticBindingForConstant"/>
     </rule>
 
-    <rule ref="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingParameterTypeHint">
-        <exclude-pattern>*/lib/*</exclude-pattern>
-    </rule>
-
-    <rule ref="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingReturnTypeHint">
-        <exclude-pattern>*/lib/*</exclude-pattern>
-    </rule>
-
     <rule ref="PSR1.Classes.ClassDeclaration.MultipleClasses">
         <exclude-pattern>*/tests/*</exclude-pattern>
     </rule>
--- a/tests/Doctrine/Tests/DBAL/Driver/AbstractOracleDriver/EasyConnectStringTest.php
+++ b/tests/Doctrine/Tests/DBAL/Driver/AbstractOracleDriver/EasyConnectStringTest.php
@@ -22,7 +22,7 @@ class EasyConnectStringTest extends TestCase
     }
 
     /**
-     * @return mixed[]
+     * @return iterable<string, array<int, mixed>>
      */
     public static function connectionParametersProvider() : iterable
     {
--- a/tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
@@ -251,17 +251,6 @@ class ConnectionTest extends DbalFunctionalTestCase
         self::assertEquals(42, $res);
     }
 
-    /**
-     * Tests that the quote function accepts DBAL and PDO types.
-     */
-    public function testQuote() : void
-    {
-        self::assertEquals(
-            $this->connection->quote('foo'),
-            $this->connection->quote('foo')
-        );
-    }
-
     public function testPingDoesTriggersConnect() : void
     {
         $this->connection->close();
--- a/tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
@@ -512,11 +512,9 @@ class DataAccessTest extends DbalFunctionalTestCase
     }
 
     /**
-     * @param string|false $char
-     *
      * @dataProvider getTrimExpressionData
      */
-    public function testTrimExpression(string $value, int $position, $char, string $expectedResult) : void
+    public function testTrimExpression(string $value, int $position, ?string $char, string $expectedResult) : void
     {
         $sql = 'SELECT ' .
             $this->connection->getDatabasePlatform()->getTrimExpression($value, $position, $char) . ' AS trimmed ' .
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
@@ -293,54 +293,54 @@ abstract class SchemaManagerFunctionalTestCase extends DbalFunctionalTestCase
 
         self::assertArrayHasKey('test', $columns);
         self::assertEquals(1, array_search('test', $columnsKeys));
-        self::assertEquals('test', strtolower($columns['test']->getname()));
-        self::assertInstanceOf(StringType::class, $columns['test']->gettype());
-        self::assertEquals(255, $columns['test']->getlength());
-        self::assertEquals(false, $columns['test']->getfixed());
-        self::assertEquals(false, $columns['test']->getnotnull());
-        self::assertEquals('expected default', $columns['test']->getdefault());
+        self::assertEquals('test', strtolower($columns['test']->getName()));
+        self::assertInstanceOf(StringType::class, $columns['test']->getType());
+        self::assertEquals(255, $columns['test']->getLength());
+        self::assertEquals(false, $columns['test']->getFixed());
+        self::assertEquals(false, $columns['test']->getNotnull());
+        self::assertEquals('expected default', $columns['test']->getDefault());
         self::assertIsArray($columns['test']->getPlatformOptions());
 
-        self::assertEquals('foo', strtolower($columns['foo']->getname()));
+        self::assertEquals('foo', strtolower($columns['foo']->getName()));
         self::assertEquals(2, array_search('foo', $columnsKeys));
-        self::assertInstanceOf(TextType::class, $columns['foo']->gettype());
-        self::assertEquals(false, $columns['foo']->getunsigned());
-        self::assertEquals(false, $columns['foo']->getfixed());
-        self::assertEquals(true, $columns['foo']->getnotnull());
-        self::assertEquals(null, $columns['foo']->getdefault());
+        self::assertInstanceOf(TextType::class, $columns['foo']->getType());
+        self::assertEquals(false, $columns['foo']->getUnsigned());
+        self::assertEquals(false, $columns['foo']->getFixed());
+        self::assertEquals(true, $columns['foo']->getNotnull());
+        self::assertEquals(null, $columns['foo']->getDefault());
         self::assertIsArray($columns['foo']->getPlatformOptions());
 
-        self::assertEquals('bar', strtolower($columns['bar']->getname()));
+        self::assertEquals('bar', strtolower($columns['bar']->getName()));
         self::assertEquals(3, array_search('bar', $columnsKeys));
-        self::assertInstanceOf(DecimalType::class, $columns['bar']->gettype());
-        self::assertEquals(null, $columns['bar']->getlength());
-        self::assertEquals(10, $columns['bar']->getprecision());
-        self::assertEquals(4, $columns['bar']->getscale());
-        self::assertEquals(false, $columns['bar']->getunsigned());
-        self::assertEquals(false, $columns['bar']->getfixed());
-        self::assertEquals(false, $columns['bar']->getnotnull());
-        self::assertEquals(null, $columns['bar']->getdefault());
+        self::assertInstanceOf(DecimalType::class, $columns['bar']->getType());
+        self::assertEquals(null, $columns['bar']->getLength());
+        self::assertEquals(10, $columns['bar']->getPrecision());
+        self::assertEquals(4, $columns['bar']->getScale());
+        self::assertEquals(false, $columns['bar']->getUnsigned());
+        self::assertEquals(false, $columns['bar']->getFixed());
+        self::assertEquals(false, $columns['bar']->getNotnull());
+        self::assertEquals(null, $columns['bar']->getDefault());
         self::assertIsArray($columns['bar']->getPlatformOptions());
 
-        self::assertEquals('baz1', strtolower($columns['baz1']->getname()));
+        self::assertEquals('baz1', strtolower($columns['baz1']->getName()));
         self::assertEquals(4, array_search('baz1', $columnsKeys));
-        self::assertInstanceOf(DateTimeType::class, $columns['baz1']->gettype());
-        self::assertEquals(true, $columns['baz1']->getnotnull());
-        self::assertEquals(null, $columns['baz1']->getdefault());
+        self::assertInstanceOf(DateTimeType::class, $columns['baz1']->getType());
+        self::assertEquals(true, $columns['baz1']->getNotnull());
+        self::assertEquals(null, $columns['baz1']->getDefault());
         self::assertIsArray($columns['baz1']->getPlatformOptions());
 
-        self::assertEquals('baz2', strtolower($columns['baz2']->getname()));
+        self::assertEquals('baz2', strtolower($columns['baz2']->getName()));
         self::assertEquals(5, array_search('baz2', $columnsKeys));
-        self::assertContains($columns['baz2']->gettype()->getName(), ['time', 'date', 'datetime']);
-        self::assertEquals(true, $columns['baz2']->getnotnull());
-        self::assertEquals(null, $columns['baz2']->getdefault());
+        self::assertContains($columns['baz2']->getType()->getName(), ['time', 'date', 'datetime']);
+        self::assertEquals(true, $columns['baz2']->getNotnull());
+        self::assertEquals(null, $columns['baz2']->getDefault());
         self::assertIsArray($columns['baz2']->getPlatformOptions());
 
-        self::assertEquals('baz3', strtolower($columns['baz3']->getname()));
+        self::assertEquals('baz3', strtolower($columns['baz3']->getName()));
         self::assertEquals(6, array_search('baz3', $columnsKeys));
-        self::assertContains($columns['baz3']->gettype()->getName(), ['time', 'date', 'datetime']);
-        self::assertEquals(true, $columns['baz3']->getnotnull());
-        self::assertEquals(null, $columns['baz3']->getdefault());
+        self::assertContains($columns['baz3']->getType()->getName(), ['time', 'date', 'datetime']);
+        self::assertEquals(true, $columns['baz3']->getNotnull());
+        self::assertEquals(null, $columns['baz3']->getDefault());
         self::assertIsArray($columns['baz3']->getPlatformOptions());
     }
 
--- a/tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
@@ -154,7 +154,6 @@ class WriteTest extends DbalFunctionalTestCase
         self::assertEquals(1, $this->connection->insert('write_table', ['test_int' => 2, 'test_string' => 'bar']));
         $num = $this->lastInsertId();
 
-        self::assertNotNull($num, 'LastInsertId() should not be null.');
         self::assertGreaterThan(0, $num, 'LastInsertId() should be non-negative number.');
     }
 
@@ -343,11 +342,9 @@ class WriteTest extends DbalFunctionalTestCase
      * Returns the ID of the last inserted row or skips the test if the currently used driver
      * doesn't support this feature
      *
-     * @return string|false
-     *
      * @throws DriverException
      */
-    private function lastInsertId(?string $name = null)
+    private function lastInsertId(?string $name = null) : string
     {
         try {
             return $this->connection->lastInsertId($name);
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
@@ -426,16 +426,14 @@ abstract class AbstractPostgreSqlPlatformTestCase extends AbstractPlatformTestCa
     }
 
     /**
-     * @param string|bool $databaseValue
-     *
      * @group DBAL-457
      * @dataProvider pgBooleanProvider
      */
     public function testConvertBooleanAsLiteralStrings(
-        $databaseValue,
+        string $databaseValue,
         string $preparedStatementValue,
-        ?int $integerValue,
-        ?bool $booleanValue
+        int $integerValue,
+        bool $booleanValue
     ) : void {
         $platform = $this->createPlatform();
 
@@ -458,16 +456,14 @@ abstract class AbstractPostgreSqlPlatformTestCase extends AbstractPlatformTestCa
     }
 
     /**
-     * @param string|bool $databaseValue
-     *
      * @group DBAL-630
      * @dataProvider pgBooleanProvider
      */
     public function testConvertBooleanAsDatabaseValueStrings(
-        $databaseValue,
+        string $databaseValue,
         string $preparedStatementValue,
-        ?int $integerValue,
-        ?bool $booleanValue
+        int $integerValue,
+        bool $booleanValue
     ) : void {
         $platform = $this->createPlatform();
 
@@ -487,11 +483,9 @@ abstract class AbstractPostgreSqlPlatformTestCase extends AbstractPlatformTestCa
     }
 
     /**
-     * @param string|bool $databaseValue
-     *
      * @dataProvider pgBooleanProvider
      */
-    public function testConvertFromBoolean($databaseValue, string $prepareStatementValue, ?int $integerValue, ?bool $booleanValue) : void
+    public function testConvertFromBoolean(string $databaseValue, string $prepareStatementValue, int $integerValue, bool $booleanValue) : void
     {
         $platform = $this->createPlatform();
 
@@ -730,7 +724,6 @@ abstract class AbstractPostgreSqlPlatformTestCase extends AbstractPlatformTestCa
     {
         return [
             // Database value, prepared statement value, boolean integer value, boolean value.
-            [true, 'true', 1, true],
             ['t', 'true', 1, true],
             ['true', 'true', 1, true],
             ['y', 'true', 1, true],
@@ -738,15 +731,12 @@ abstract class AbstractPostgreSqlPlatformTestCase extends AbstractPlatformTestCa
             ['on', 'true', 1, true],
             ['1', 'true', 1, true],
 
-            [false, 'false', 0, false],
             ['f', 'false', 0, false],
             ['false', 'false', 0, false],
             [ 'n', 'false', 0, false],
             ['no', 'false', 0, false],
             ['off', 'false', 0, false],
             ['0', 'false', 0, false],
-
-            [null, 'NULL', null, null],
         ];
     }
 
@@ -809,14 +799,6 @@ abstract class AbstractPostgreSqlPlatformTestCase extends AbstractPlatformTestCa
     protected function getQuotesDropForeignKeySQL() : string
     {
         return 'ALTER TABLE "table" DROP CONSTRAINT "select"';
-    }
-
-    public function testGetNullCommentOnColumnSQL() : void
-    {
-        self::assertEquals(
-            'COMMENT ON COLUMN mytable.id IS NULL',
-            $this->platform->getCommentOnColumnSQL('mytable', 'id', null)
-        );
     }
 
     /**
--- a/tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
@@ -246,7 +246,7 @@ class OraclePlatformTest extends AbstractPlatformTestCase
      */
     public function testGeneratesAdvancedForeignKeyOptionsSQL(array $options, string $expectedSql) : void
     {
-        $foreignKey = new ForeignKeyConstraint(['foo'], 'foreign_table', ['bar'], null, $options);
+        $foreignKey = new ForeignKeyConstraint(['foo'], 'foreign_table', ['bar'], '', $options);
 
         self::assertSame($expectedSql, $this->platform->getAdvancedForeignKeyOptionsSQL($foreignKey));
     }
--- a/tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php
@@ -234,11 +234,9 @@ class SQLAnywherePlatformTest extends AbstractPlatformTestCase
     }
 
     /**
-     * @param int|bool|null $lockMode
-     *
      * @dataProvider getLockHints
      */
-    public function testAppendsLockHint($lockMode, string $lockHint) : void
+    public function testAppendsLockHint(?int $lockMode, string $lockHint) : void
     {
         $fromClause     = 'FROM users';
         $expectedResult = $fromClause . $lockHint;
--- a/tests/Doctrine/Tests/DBAL/Platforms/SQLServerPlatformTest.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/SQLServerPlatformTest.php
@@ -16,12 +16,10 @@ class SQLServerPlatformTest extends AbstractSQLServerPlatformTestCase
     }
 
     /**
-     * @param int|bool|null $lockMode
-     *
      * @group DDC-2310
      * @dataProvider getLockHints
      */
-    public function testAppendsLockHint($lockMode, string $lockHint) : void
+    public function testAppendsLockHint(?int $lockMode, string $lockHint) : void
     {
         $fromClause     = 'FROM users';
         $expectedResult = $fromClause . $lockHint;
@@ -33,7 +31,7 @@ class SQLServerPlatformTest extends AbstractSQLServerPlatformTestCase
      * @group DBAL-2408
      * @dataProvider getModifyLimitQueries
      */
-    public function testScrubInnerOrderBy(string $query, int $limit, ?int $offset, string $expectedResult) : void
+    public function testScrubInnerOrderBy(string $query, int $limit, int $offset, string $expectedResult) : void
     {
         self::assertSame($expectedResult, $this->platform->modifyLimitQuery($query, $limit, $offset));
     }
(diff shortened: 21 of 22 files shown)


Majkl578:
This looks impossible to review and properly document. Who don't we do it in smaller parts?

Also eliminating type casts/checks is probably task for separate PR (PHPStan should catch them).

morozov:
>> Clean up type casting and checks like (bool) $flag or is_string($string) which are no longer needed.

> Also eliminating type casts/checks is probably task for separate PR (PHPStan should catch them).

I just though about the same. Will double check if PHPStan can do that and will file a ticket otherwilse and attach it here.

Majkl578:
It definitely does, but probably on higher level than we have.
IMO we should've first worked on level 7 and then on enabling scalar types, this way we may end up with fatal errors in uncovered code (100% coverage is a dream).

morozov:
> This looks impossible to review and properly document. Why don't we do it in smaller parts?

I'd better try to come up with a systems approach of reviewing what's already done then splitting it apart and then putting back together. It took me a week to make this change including almost one day full time, and I may not have that much time for additional coding.

morozov:
> Hmm, where? Maybe I just forgot.

I meant https://github.com/doctrine/dbal/pull/3260 and https://github.com/doctrine/dbal/pull/2854.

> Enabling strict types is separate task, not blocking scalar types

I see. I misread this:

> IMO we should've first worked on level 7 and then on enabling scalar types

.phpunit.result.cache
```diff
@@ -0,0 +1 @@
...h data set #2";d:0;s:58:"testReturnsForeignKeyReferentialActionSQL with data set #3";d:0;s:58:"testReturnsForeignKeyReferentialActionSQL with data set #4";d:0;s:58:"testReturnsForeignKeyReferentialActionSQL with data set #5";d:0;s:25:"testQuoteSingleIdentifier";d:0;s:19:"testQuoteIdentifier";d:0;s:33:"testColumnCollationDeclarationSQL";d:0;s:61:"testDoesNotPropagateForeignKeyCreationForNonSupportingEngines";d:0.001;s:27:"testSupportsColumnCollation";d:0;s:44:"testListTableForeignKeysSQLEvaluatesDatabase";d:0;s:43:"testQuotesDatabaseNameInListTableColumnsSQL";d:0;s:40:"testQuotesTableNameInListTableColumnsSQL";d:0;s:47:"testQuotesDatabaseNameInListTableForeignKeysSQL";d:0;s:44:"testQuotesTableNameInListTableForeignKeysSQL";d:0;s:36:"testQuotesDatabaseNameInListViewsSQL";d:0;s:43:"testQuotesDatabaseNameInListTableIndexesSQL";d:0;s:40:"testQuotesTableNameInListTableIndexesSQL";d:0;s:33:"testReturnsGuidTypeDeclarationSQL";d:0;s:58:"testDoesNotPropagateDefaultValuesForUnsupportedColumnTypes";d:0.001;s:63:"testDoesNotPropagateForeignKeyAlterationForNonSupportingEngines";d:0.001;s:35:"testZeroOffsetWithoutLimitIsIgnored";d:0;s:45:"testIdempotentDataConversion with data set #0";d:0.179;}}}
```
greg0ire: Should this really be under version control?

morozov: No, I accidentally committed it when I tried using new PHPUnit flags like `--cache-result` combined with `--order-by`. Does anyone have experience using them?

greg0ire: I read about them the other day, but did not try yet… but it sounds cool!

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1005,254 +879,169 @@ public function getCosExpression($value)
...
      *
-     * @param string $date
-     * @param int    $quarters
-     *
-     * @return string
-     *
      * @throws DBALException If not supported on this platform.
      */
-    public function getDateAddQuartersExpression($date, $quarters)
+    public function getDateAddQuartersExpression(string $date, string $quarters) : string
```
Majkl578: `$quarters` should be `int` according to phpDoc so this is a BC break.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1005,254 +879,169 @@ public function getCosExpression($value)
...
      *
-     * @param string $date
-     * @param int    $quarters
-     *
-     * @return string
-     *
      * @throws DBALException If not supported on this platform.
      */
-    public function getDateSubQuartersExpression($date, $quarters)
+    public function getDateSubQuartersExpression(string $date, string $quarters) : string
```
Majkl578: `$quarters` should be `int` according to phpDoc so this is a BC break.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1005,254 +879,169 @@ public function getCosExpression($value)
...
      *
-     * @param string $date
-     * @param int    $years
-     *
-     * @return string
-     *
      * @throws DBALException If not supported on this platform.
      */
-    public function getDateAddYearsExpression($date, $years)
+    public function getDateAddYearsExpression(string $date, string $years) : string
```
Majkl578: `$years` should be `int` according to phpDoc so this is a BC break.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1005,254 +879,169 @@ public function getCosExpression($value)
...
      *
-     * @param string $date
-     * @param int    $years
-     *
-     * @return string
-     *
      * @throws DBALException If not supported on this platform.
      */
-    public function getDateSubYearsExpression($date, $years)
+    public function getDateSubYearsExpression(string $date, string $years) : string
```
Majkl578: `$years` should be `int` according to phpDoc so this is a BC break.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1262,51 +1051,55 @@ public function getDateSubYearsExpression($date, $years)
...
      * @param string $unit     The unit of the interval that shall be calculated into the date.
-     *                         One of the DATE_INTERVAL_UNIT_* constants.
-     *
-     * @return string
+     *                                 One of the DATE_INTERVAL_UNIT_* constants.
      *
      * @throws DBALException If not supported on this platform.
      */
-    protected function getDateArithmeticIntervalExpression($date, $operator, $interval, $unit)
+    protected function getDateArithmeticIntervalExpression(string $date, string $operator, string $interval, string $unit) : string
```
Majkl578: `$interval` should be `int` according to phpDoc so this is a BC break.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1262,51 +1051,55 @@ public function getDateSubYearsExpression($date, $years)
...
-     * @param string $value2
+     * @param string $interval   SQL expression describing the value of the interval
+     * @param string $unit       Interval unit
+     * @param int    $multiplier Interval multiplier
      *
-     * @return string
+     * @throws DBALException
      */
-    public function getBitAndComparisonExpression($value1, $value2)
+    protected function convertNonNativeInterval(string $interval, string $unit, int $multiplier) : string
```
Majkl578: `$multiplier` should be `int` according to phpDoc so this is a BC break. Also visibility change is a BC break.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1367,11 +1152,9 @@ public function getDropDatabaseSQL($database)
      *
      * @param Table|string $table
```
Majkl578: Not directly related, but we should take care of these unions in a major version.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1480,19 +1255,13 @@ public function getDropForeignKeySQL($foreignKey, $table)
      * Returns the SQL statement(s) to create a table with the specified name, columns and constraints
      * on this platform.
      *
-     * @param int $createFlags
-     *
      * @return string[] The sequence of SQL statements.
      *
      * @throws DBALException
      * @throws InvalidArgumentException
```
Majkl578: No longer thrown.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2625,10 +2328,8 @@ public function convertBooleans($item)
      * The default conversion tries to convert value into bool "(bool)$item"
      *
      * @param mixed $item
-     *
-     * @return bool|null
      */
-    public function convertFromBoolean($item)
+    public function convertFromBoolean($item) : ?bool
     {
         return $item === null ? null: (bool) $item;
```
Majkl578: space in front of `:` (looks like a CS issue)

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3095,80 +2703,64 @@ public function supportsColumnLengthIndexes() : bool
...
 
     /**
      * Whether this platform supports onUpdate in foreign key constraints.
-     *
-     * @return bool
      */
-    public function supportsForeignKeyOnUpdate()
+    public function supportsForeignKeyOnUpdate() : bool
     {
-        return $this->supportsForeignKeyConstraints() && true;
```
Majkl578: What was this for? A boolean cast?

lib/Doctrine/DBAL/Platforms/MySQL57Platform.php
```diff
@@ -14,45 +14,47 @@ class MySQL57Platform extends MySqlPlatform
...
-    protected function getRenameIndexSQL($oldIndexName, Index $index, $tableName)
+    protected function getRenameIndexSQL(string $oldIndexName, Index $index, string $tableName) : array
     {
         return ['ALTER TABLE ' . $tableName . ' RENAME INDEX ' . $oldIndexName . ' TO ' . $index->getQuotedName($this)];
     }
 
     /**
      * {@inheritdoc}
+     *
+     * @return string
```
Majkl578: Should be native type.

lib/Doctrine/DBAL/Platforms/MySQL80Platform.php
```diff
@@ -9,6 +9,8 @@ class MySQL80Platform extends MySQL57Platform
 {
     /**
      * {@inheritdoc}
+     *
+     * @return string
```
Majkl578: Should be native type.

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -194,23 +193,23 @@ public function getListTableForeignKeysSQL($table, $database = null)
...
+    public function getDropViewSQL(string $name) : string
     {
         return 'DROP VIEW ' . $name;
     }
 
     /**
      * {@inheritDoc}
      */
-    protected function getVarcharTypeDeclarationSQLSnippet($length, $fixed)
+    protected function getVarcharTypeDeclarationSQLSnippet(?int $length, bool $fixed) : string
```
Majkl578: Wouldn't `0` be better default?

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -219,7 +218,7 @@ protected function getVarcharTypeDeclarationSQLSnippet($length, $fixed)
     /**
      * {@inheritdoc}
      */
-    protected function getBinaryTypeDeclarationSQLSnippet($length, $fixed)
+    protected function getBinaryTypeDeclarationSQLSnippet(?int $length, bool $fixed) : string
```
Majkl578: Wouldn't `0` be better default?

lib/Doctrine/DBAL/Platforms/PostgreSQL100Platform.php
```diff
@@ -19,7 +19,7 @@ protected function getReservedKeywordsClass() : string
         return PostgreSQL100Keywords::class;
     }
 
-    public function getListSequencesSQL($database) : string
+    public function getListSequencesSQL(?string $database) : ?string
```
Majkl578: This doesn't seem to ever return `null`.

lib/Doctrine/DBAL/Platforms/PostgreSQL94Platform.php
```diff
@@ -23,6 +25,8 @@ public function getJsonTypeDeclarationSQL(array $field)
 
     /**
      * {@inheritdoc}
+     *
+     * @return string
```
Majkl578: Should be native type.

lib/Doctrine/DBAL/Platforms/PostgreSQL94Platform.php
```diff
@@ -31,8 +35,10 @@ protected function getReservedKeywordsClass()
 
     /**
      * {@inheritdoc}
+     *
+     * @return void
```
Majkl578: Should be removed.

lib/Doctrine/DBAL/Schema/Column.php
```diff
@@ -92,106 +88,60 @@ public function setOptions(array $options)
...
-    public function setPrecision($precision)
+    public function setPrecision(?int $precision) : Column
     {
-        if (! is_numeric($precision)) {
-            $precision = 10; // defaults to 10 when no valid precision is given.
-        }
-
-        $this->_precision = (int) $precision;
+        // defaults to 10 when no valid precision is given.
+        $this->_precision = $precision ?? 10;
```
Majkl578: Maybe should use `int $precision = 10` directly?

lib/Doctrine/DBAL/Schema/Column.php
```diff
@@ -92,106 +88,60 @@ public function setOptions(array $options)
...
-     */
-    public function setScale($scale)
+    public function setScale(?int $scale) : Column
     {
-        if (! is_numeric($scale)) {
-            $scale = 0;
-        }
-
-        $this->_scale = (int) $scale;
+        $this->_scale = $scale ?? 0;
```
Majkl578: Maybe should use `int $scale = 0` directly?

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
```diff
@@ -312,12 +312,8 @@ protected function _getPortableDatabaseDefinition($database)
     /**
      * {@inheritdoc}
      */
-    public function createDatabase($database = null)
+    public function createDatabase(string $database) : void
     {
-        if ($database === null) {
```
Majkl578: This is a BC break.

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -442,14 +440,17 @@ protected function _getPortableTableColumnDefinition($tableColumn)
...
             'scale'         => $scale,
-            'fixed'         => $fixed,
             'unsigned'      => false,
             'autoincrement' => $autoincrement,
             'comment'       => isset($tableColumn['comment']) && $tableColumn['comment'] !== ''
                 ? $tableColumn['comment']
                 : null,
         ];
 
+        if ($fixed !== null) {
```
Majkl578: What's the background of this change?

lib/Doctrine/DBAL/Schema/Schema.php
```diff
@@ -283,11 +248,11 @@ public function getSequences()
      *
      * @param string $namespaceName The name of the namespace to create.
      *
-     * @return \Doctrine\DBAL\Schema\Schema This schema instance.
+     * @return $this
```
Majkl578: Please avoid `$this`, use `static` or nothing instead.

lib/Doctrine/DBAL/Schema/Schema.php
```diff
@@ -322,12 +283,9 @@ public function createTable($tableName)
     /**
      * Renames a table.
      *
-     * @param string $oldTableName
-     * @param string $newTableName
-     *
-     * @return \Doctrine\DBAL\Schema\Schema
+     * @return $this
```
Majkl578: Please avoid `$this`, use `static` or nothing instead.

lib/Doctrine/DBAL/Schema/Schema.php
```diff
@@ -341,11 +299,9 @@ public function renameTable($oldTableName, $newTableName)
     /**
      * Drops a table from the schema.
      *
-     * @param string $tableName
-     *
-     * @return \Doctrine\DBAL\Schema\Schema
+     * @return $this
```
Majkl578: Please avoid `$this`, use `static` or nothing instead.

lib/Doctrine/DBAL/Schema/Schema.php
```diff
@@ -372,11 +322,9 @@ public function createSequence($sequenceName, $allocationSize = 1, $initialValue
     }
 
     /**
-     * @param string $sequenceName
-     *
-     * @return \Doctrine\DBAL\Schema\Schema
+     * @return $this
```
Majkl578: Please avoid `$this`, use `static` or nothing instead.

lib/Doctrine/DBAL/Schema/Visitor/AbstractVisitor.php
```diff
@@ -14,34 +14,36 @@
...
 {
-    public function acceptSchema(Schema $schema)
+    public function acceptSchema(Schema $schema) : void
     {
     }
 
     /**
      * {@inheritdoc}
+     *
+     * @return void
```
Majkl578: To be dropped.

lib/Doctrine/DBAL/Schema/Visitor/CreateSchemaSqlCollector.php
```diff
@@ -32,8 +32,10 @@ public function __construct(AbstractPlatform $platform)
 
     /**
      * {@inheritdoc}
+     *
+     * @return void
```
Majkl578: To be dropped.

lib/Doctrine/DBAL/Sharding/ShardChoser/ShardChoser.php
```diff
@@ -12,10 +12,6 @@
 {
     /**
      * Picks a shard for the given distribution value.
-     *
-     * @param string|int $distributionValue
-     *
-     * @return string|int
      */
-    public function pickShard($distributionValue, PoolingShardConnection $conn);
+    public function pickShard(string $distributionValue, PoolingShardConnection $conn) : string;
```
Majkl578: Dropping `int` is intentional BC break?

lib/Doctrine/DBAL/Tools/Console/Command/RunSqlCommand.php
```diff
@@ -21,6 +21,8 @@ class RunSqlCommand extends Command
 {
     /**
      * {@inheritdoc}
+     *
+     * @return void
```
Majkl578: Should become native.

lib/Doctrine/DBAL/Types/DateTimeImmutableType.php
```diff
@@ -13,14 +13,18 @@ class DateTimeImmutableType extends DateTimeType
 {
     /**
      * {@inheritdoc}
+     *
+     * @return string
```
Majkl578: To be dropped.

lib/Doctrine/DBAL/Types/DateImmutableType.php
```diff
@@ -40,6 +44,8 @@ public function convertToDatabaseValue($value, AbstractPlatform $platform)
 
     /**
      * {@inheritdoc}
+     *
+     * @return DateTimeImmutable|null
```
Majkl578: Can be expressed natively.

lib/Doctrine/DBAL/Types/DateTimeImmutableType.php
```diff
@@ -41,6 +45,8 @@ public function convertToDatabaseValue($value, AbstractPlatform $platform)
 
     /**
      * {@inheritdoc}
+     *
+     * @return DateTimeImmutable|null
```
Majkl578: Can be expressed natively.

lib/Doctrine/DBAL/Types/DateTimeTzImmutableType.php
```diff
@@ -40,6 +44,8 @@ public function convertToDatabaseValue($value, AbstractPlatform $platform)
 
     /**
      * {@inheritdoc}
+     *
+     * @return DateTimeImmutable|null
```
Majkl578: Can be expressed natively.

lib/Doctrine/DBAL/Types/DateTimeTzImmutableType.php
```diff
@@ -12,14 +12,18 @@ class DateTimeTzImmutableType extends DateTimeTzType
...
-    public function getName()
+    public function getName() : string
     {
         return Type::DATETIMETZ_IMMUTABLE;
     }
 
     /**
      * {@inheritdoc}
+     *
+     * @return string|false|null
```
Majkl578: `false` should be eliminated.

lib/Doctrine/DBAL/Types/DateTimeImmutableType.php
```diff
@@ -13,14 +13,18 @@ class DateTimeImmutableType extends DateTimeType
...
-    public function getName()
+    public function getName() : string
     {
         return Type::DATETIME_IMMUTABLE;
     }
 
     /**
      * {@inheritdoc}
+     *
+     * @return string|false|null
```
Majkl578: `false` should be eliminated.

lib/Doctrine/DBAL/Types/DateImmutableType.php
```diff
@@ -12,14 +12,18 @@ class DateImmutableType extends DateType
...
-    public function getName()
+    public function getName() : string
     {
         return Type::DATE_IMMUTABLE;
     }
 
     /**
      * {@inheritdoc}
+     *
+     * @return string|false|null
```
Majkl578: `false` should be eliminated.

lib/Doctrine/DBAL/Types/JsonArrayType.php
```diff
@@ -30,16 +30,20 @@ public function convertToPHPValue($value, AbstractPlatform $platform)
...
-    public function getName()
+    public function getName() : string
     {
         return Type::JSON_ARRAY;
     }
 
     /**
      * {@inheritdoc}
+     *
+     * @return true
```
Majkl578: Should be dropped.

lib/Doctrine/DBAL/Types/JsonArrayType.php
```diff
@@ -30,16 +30,20 @@ public function convertToPHPValue($value, AbstractPlatform $platform)
 
     /**
      * {@inheritdoc}
+     *
+     * @return string
```
Majkl578: To be dropped.

lib/Doctrine/DBAL/Types/GuidType.php
```diff
@@ -11,24 +11,28 @@ class GuidType extends StringType
...
-    public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform)
+    public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform) : string
     {
         return $platform->getGuidTypeDeclarationSQL($fieldDeclaration);
     }
 
     /**
      * {@inheritdoc}
+     *
+     * @return string
```
Majkl578: To be dropped.

lib/Doctrine/DBAL/Types/DateTimeTzImmutableType.php
```diff
@@ -12,14 +12,18 @@ class DateTimeTzImmutableType extends DateTimeTzType
 {
     /**
      * {@inheritdoc}
+     *
+     * @return string
```
Majkl578: To be dropped.

lib/Doctrine/DBAL/Types/DateImmutableType.php
```diff
@@ -12,14 +12,18 @@ class DateImmutableType extends DateType
 {
     /**
      * {@inheritdoc}
+     *
+     * @return string
```
Majkl578: To be dropped.

lib/Doctrine/DBAL/Types/GuidType.php
```diff
@@ -11,24 +11,28 @@ class GuidType extends StringType
 {
     /**
      * {@inheritdoc}
+     *
+     * @return string
```
Majkl578: To be dropped.

lib/Doctrine/DBAL/Types/TimeImmutableType.php
```diff
@@ -12,14 +12,18 @@ class TimeImmutableType extends TimeType
 {
     /**
      * {@inheritdoc}
+     *
+     * @return string
```
Majkl578: To be dropped.

lib/Doctrine/DBAL/Types/TimeImmutableType.php
```diff
@@ -40,6 +44,8 @@ public function convertToDatabaseValue($value, AbstractPlatform $platform)
 
     /**
      * {@inheritdoc}
+     *
+     * @return DateTimeImmutable|null
```
Majkl578: Can be expressed natively.

lib/Doctrine/DBAL/Types/TimeImmutableType.php
```diff
@@ -12,14 +12,18 @@ class TimeImmutableType extends TimeType
...
-    public function getName()
+    public function getName() : string
     {
         return Type::TIME_IMMUTABLE;
     }
 
     /**
      * {@inheritdoc}
+     *
+     * @return string|false|null
```
Majkl578: `false` should be eliminated.

lib/Doctrine/DBAL/Types/VarDateTimeImmutableType.php
```diff
@@ -13,14 +13,18 @@ class VarDateTimeImmutableType extends VarDateTimeType
...
-    public function getName()
+    public function getName() : string
     {
         return Type::DATETIME_IMMUTABLE;
     }
 
     /**
      * {@inheritdoc}
+     *
+     * @return string|false|null
```
Majkl578: `false` should be eliminated.

lib/Doctrine/DBAL/Types/VarDateTimeImmutableType.php
```diff
@@ -13,14 +13,18 @@ class VarDateTimeImmutableType extends VarDateTimeType
 {
     /**
      * {@inheritdoc}
+     *
+     * @return string
```
Majkl578: To be dropped.

lib/Doctrine/DBAL/Types/VarDateTimeImmutableType.php
```diff
@@ -41,6 +45,8 @@ public function convertToDatabaseValue($value, AbstractPlatform $platform)
 
     /**
      * {@inheritdoc}
+     *
+     * @return DateTimeImmutable|null
```
Majkl578: Can be expressed natively.

lib/Doctrine/DBAL/Types/VarDateTimeType.php
```diff
@@ -17,6 +17,8 @@ class VarDateTimeType extends DateTimeType
 {
     /**
      * {@inheritdoc}
+     *
+     * @return DateTime|null
```
Majkl578: Can be expressed natively.

Majkl578 (changes requested):

Ocramius:
Possibly a good idea to have https://github.com/doctrine/dbal/pull/2854 in first - would catch a lot of runtime errors due to type mismatch, if there are any left

morozov:
Yea, we discussed that internally.

lib/Doctrine/DBAL/Configuration.php
```diff
@@ -87,20 +79,18 @@ public function setFilterSchemaAssetsExpression($filterExpression)
...
 
-    /**
-     * @param string $filterExpression
-     */
-    private function buildSchemaAssetsFilterFromExpression($filterExpression) : callable
+    private function buildSchemaAssetsFilterFromExpression(string $filterExpression) : callable
     {
-        return static function ($assetName) use ($filterExpression) {
+        return /**
+                * @return int|false
```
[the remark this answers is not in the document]
morozov: Hm… this probably just was autogenerated based on the signature of `preg_match()`. The filter function will return `bool`.

UPGRADE.md
```diff
@@ -1,5 +1,9 @@
 # Upgrade to 3.0
 
+## BC BREAK `ResultStatement::setFetchMode()` and `::closeCursor()` return `void`.
+
+Unlike `PDOStatement`, `::setFetchMode()` and `::closeCursor()` of `ResultStatement` won't longer return a boolean value. In the future, logical errors during invocation of these methods may result in exceptions.
```
```suggestion
Unlike `PDOStatement`, `::setFetchMode()` and `::closeCursor()` of `ResultStatement` will no longer return a boolean value. In the future, logical errors during invocation of these methods may result in exceptions.
```

lib/Doctrine/DBAL/Connection.php
```diff
@@ -342,7 +324,7 @@ public function getExpressionBuilder()
      * @return bool TRUE if the connection was successfully established, FALSE if
      *              the connection is already open.
      */
-    public function connect()
+    public function connect() : bool
```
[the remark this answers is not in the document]
morozov: Done.

lib/Doctrine/DBAL/Driver/Statement.php
```diff
@@ -45,17 +45,17 @@ public function bindValue($param, $value, $type = ParameterType::STRING);
...
-     *                           constants.
+     *                               constants.
      * @param int|null $length   You must specify maxlength when using an OUT bind
-     *                           so that PHP allocates enough memory to hold the returned value.
+     *                             so that PHP allocates enough memory to hold the returned value.
      *
      * @return bool TRUE on success or FALSE on failure.
      */
-    public function bindParam($column, &$variable, $type = ParameterType::STRING, $length = null);
+    public function bindParam($column, &$variable, $type = ParameterType::STRING, $length = null) : bool;
```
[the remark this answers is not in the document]
morozov: Yes, but I'd defer it to https://github.com/doctrine/dbal/pull/3154. It will also require restructuring the exception hierarchy first.

lib/Doctrine/DBAL/Driver/Statement.php
```diff
@@ -24,11 +24,11 @@
      *                     using question mark placeholders, this will be the 1-indexed position of the parameter.
      * @param mixed $value The value to bind to the parameter.
      * @param int   $type  Explicit data type for the parameter using the {@link \Doctrine\DBAL\ParameterType}
-     *                     constants.
+     *                        constants.
      *
      * @return bool TRUE on success or FALSE on failure.
      */
-    public function bindValue($param, $value, $type = ParameterType::STRING);
+    public function bindValue($param, $value, $type = ParameterType::STRING) : bool;
```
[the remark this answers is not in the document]
morozov: Will be resolved in https://github.com/doctrine/dbal/pull/3154.

lib/Doctrine/DBAL/Driver/Statement.php
```diff
@@ -87,7 +87,7 @@ public function errorInfo();
      *
      * @return bool TRUE on success or FALSE on failure.
      */
-    public function execute($params = null);
+    public function execute(?array $params = null) : bool;
```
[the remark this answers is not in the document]
morozov: Will be resolved in https://github.com/doctrine/dbal/pull/3154.

lib/Doctrine/DBAL/Event/SchemaAlterTableEventArgs.php
```diff
@@ -27,28 +27,22 @@ public function __construct(TableDiff $tableDiff, AbstractPlatform $platform)
...
+    public function getPlatform() : AbstractPlatform
     {
         return $this->platform;
     }
 
     /**
      * @param string|string[] $sql
      *
-     * @return \Doctrine\DBAL\Event\SchemaAlterTableEventArgs
+     * @return $this
```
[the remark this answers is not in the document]
morozov: `$this` is a valid phpDoc return type. It's more specific than `static` or `self` meaning that it will return the same object.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -747,11 +657,9 @@ public function getModExpression($expression1, $expression2)
      *
      * @param string      $str  The expression to apply the trim to.
      * @param int         $mode The position of the trim (leading/trailing/both).
-     * @param string|bool $char The char to trim, has to be quoted already. Defaults to space.
-     *
-     * @return string
+     * @param string|null $char The char to trim, has to be quoted already. Defaults to space.
      */
-    public function getTrimExpression($str, $mode = TrimMode::UNSPECIFIED, $char = false)
+    public function getTrimExpression(string $str, int $mode = TrimMode::UNSPECIFIED, ?string $char = null) : string
```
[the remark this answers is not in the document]
morozov: It is. I'll add it to the upgrade notes.

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -111,18 +111,16 @@ public function __construct(array $params, Driver $driver, ?Configuration $confi
...
+    public function isConnectedToMaster() : bool
     {
         return $this->_conn !== null && $this->_conn === $this->connections['master'];
     }
 
     /**
      * {@inheritDoc}
      */
-    public function connect($connectionName = null)
+    public function connect($connectionName = null) : bool
```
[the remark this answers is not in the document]
morozov: This looks creepy. Would it make sense to split it to `connect()` and `connectToSlave(string $connectionName)`?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -152,17 +151,15 @@ public function __construct()
     /**
      * Sets the EventManager used by the Platform.
      */
-    public function setEventManager(EventManager $eventManager)
+    public function setEventManager(?EventManager $eventManager) : void
```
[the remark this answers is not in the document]
morozov: Not sure why it got nullable, will revert.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -152,17 +151,15 @@ public function __construct()
...
         $this->_eventManager = $eventManager;
     }
 
     /**
      * Gets the EventManager used by the Platform.
-     *
-     * @return EventManager
      */
-    public function getEventManager()
+    public function getEventManager() : ?EventManager
```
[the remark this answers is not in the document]
morozov: Before it gets set, it's null. Why doesn't it correspond?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1005,254 +879,169 @@ public function getCosExpression($value)
...
      *
-     * @param string $date
-     * @param int    $seconds
-     *
-     * @return string
-     *
      * @throws DBALException If not supported on this platform.
      */
-    public function getDateAddSecondsExpression($date, $seconds)
+    public function getDateAddSecondsExpression(string $date, string $seconds) : string
```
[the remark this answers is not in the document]
morozov: It is a BC break but it should be a `string` by design. When it comes to expressions, every component is an expression as well (e.g. can be a reference to a different field rather than a numeric value). There's no point in restricting it to an `int`, unless some platform-specific corner cases.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: There's no visibility change. It's a different method.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: I'd like to but AFAIK it may complicate some cases. Should be resolved on a per-case basis.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: This is the only purpose I can think of. As long as the return type is now enforced, it's no longer needed.

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
morozov: There's no default. It's a nullable length for platforms like SQLite which don't support length or PostgreSQL which don't require one.

I have some PoC changes in https://github.com/morozov/dbal/commits/remove-default-max-length.

lib/Doctrine/DBAL/Schema/Column.php
morozov: I'd better get rid of it entirely. What's the point in calling a method with knowingly wrong value and rely on some implementation details?

lib/Doctrine/DBAL/Schema/Column.php
morozov: Same as above. It'd rather try to remove this default entirely and make it non-nullable.

lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
morozov: `null` is not a valid value of a boolean parameter but it's initialized with a `null` and then assigned a boolean value based on the column metadata. We pass the value down to the column definition only if it was explicitly assigned.

lib/Doctrine/DBAL/Schema/Schema.php
morozov: I'd rather leave it as `$this`. Please see http://docs.phpdoc.org/guides/types.html#keywords.

lib/Doctrine/DBAL/Types/TimeImmutableType.php
morozov: It was probably inferred from `DateTimeImmutable::format()`.

lib/Doctrine/DBAL/Sharding/ShardChoser/ShardChoser.php
morozov: No, it should remain `string|int`.

lib/Doctrine/DBAL/Types/DateTimeImmutableType.php
morozov: Shouldn't PHPCS detect this as redundant?

lib/Doctrine/DBAL/Exception/DriverException.php
```diff
@@ -21,7 +21,7 @@ class DriverException extends DBALException
      * @param string                                $message         The exception message.
      * @param \Doctrine\DBAL\Driver\DriverException $driverException The DBAL driver exception to chain.
      */
-    public function __construct($message, \Doctrine\DBAL\Driver\DriverException $driverException)
+    public function __construct(string $message, \Doctrine\DBAL\Driver\DriverException $driverException)
```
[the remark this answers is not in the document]
morozov: Shouldn't PHPCS detect this?

lib/Doctrine/DBAL/Sharding/ShardChoser/ShardChoser.php
morozov: Well, if `$distributionValue` is the partition key what determines the shard, it can be anything including a list of different values. In light of this, the implementation of `MultiTenantShardChoser` looks questionable. Should we just declare it `mixed`?

lib/Doctrine/DBAL/Event/SchemaAlterTableEventArgs.php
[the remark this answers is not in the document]
morozov: FWTW, `self` may lead to false-positive static analysis issues like https://github.com/doctrine/dbal/pull/2384#issuecomment-308863577. I'd still stick with `$this`.

lib/Doctrine/DBAL/Driver/PDOConnection.php
```diff
@@ -17,17 +16,17 @@ class PDOConnection implements Connection, ServerInfoAwareConnection
...
-     * @param mixed[]|null $options
+     * @param mixed[] $options
      *
      * @throws PDOException In case of an error.
      */
-    public function __construct($dsn, $user = null, $password = null, ?array $options = null)
+    public function __construct(string $dsn, ?string $username = null, ?string $password = null, array $options = [])
     {
+        $username = (string) $username;
+        $password = (string) $password;
```
morozov: These are not needed.

lib/Doctrine/DBAL/Driver/Statement.php
```diff
@@ -24,11 +24,11 @@ interface Statement extends ResultStatement
      *                     using question mark placeholders, this will be the 1-indexed position of the parameter.
      * @param mixed $value The value to bind to the parameter.
      * @param int   $type  Explicit data type for the parameter using the {@link \Doctrine\DBAL\ParameterType}
-     *                     constants.
+     *                        constants.
```
morozov: This should be reverted.

lib/Doctrine/DBAL/Event/SchemaAlterTableRemoveColumnEventArgs.php
```diff
@@ -32,36 +32,25 @@ public function __construct(Column $column, TableDiff $tableDiff, AbstractPlatfo
...
         return $this->platform;
     }
 
     /**
      * @param string|string[] $sql
-     *
-     * @return \Doctrine\DBAL\Event\SchemaAlterTableRemoveColumnEventArgs
      */
-    public function addSql($sql)
+    public function addSql($sql) : self
```
morozov: Should the type of `$sql` be restricted to `string`? Or `...$string`?

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1257,7 +1234,7 @@ public function __toString()
      *
      * @return string the placeholder name used.
      */
-    public function createNamedParameter($value, $type = ParameterType::STRING, $placeHolder = null)
+    public function createNamedParameter($value, $type = ParameterType::STRING, $placeHolder = null) : string
```
morozov: Should be `?string $placeHolder`?

lib/Doctrine/DBAL/Schema/Comparator.php
```diff
@@ -273,7 +260,10 @@ public function diffTable(Table $table1, Table $table2)
                 if ($this->diffForeignKey($constraint1, $constraint2) === false) {
                     unset($fromFkeys[$key1], $toFkeys[$key2]);
                 } else {
-                    if (strtolower($constraint1->getName()) === strtolower($constraint2->getName())) {
+                    $name1 = $constraint1->getName();
+                    $name2 = $constraint2->getName();
+
+                    if ($name1 !== null && $name2 !== null && strcasecmp($name1, $name2) === 0) {
```
morozov: This looks incorrect. Should be something like: https://github.com/doctrine/dbal/blob/34c3ad11496d8065d164fc095ac5569ebbc3d74f/lib/Doctrine/DBAL/Schema/Comparator.php#L436

lib/Doctrine/DBAL/Schema/AbstractAsset.php
```diff
@@ -172,13 +152,14 @@ public function getName()
...
      * Gets the quoted representation of this asset but only if it was defined with one. Otherwise
      * return the plain unquoted value as inserted.
-     *
-     * @return string
      */
-    public function getQuotedName(AbstractPlatform $platform)
+    public function getQuotedName(AbstractPlatform $platform) : string
     {
+        $name = $this->getName();
+        assert(is_string($name));
```
morozov: The assertion will fail if the method is called in an anonymous asset. Should be an exception instead.

lib/Doctrine/DBAL/Schema/DB2SchemaManager.php
```diff
@@ -173,8 +169,10 @@ protected function _getPortableTableForeignKeysList($tableForeignKeys)
 
     /**
      * {@inheritdoc}
+     *
+     * @return string|null
```
morozov: This is redundant.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -476,7 +451,7 @@ private function getServerVersion()
      *
      * @return bool True if auto-commit mode is currently enabled for this connection, false otherwise.
      */
-    public function isAutoCommit()
+    public function isAutoCommit() : bool
     {
         return $this->autoCommit === true;
```
morozov: The comparison is no longer needed.

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereConnection.php
```diff
@@ -203,7 +184,7 @@ public function rollBack()
      *
      * @throws SQLAnywhereException
      */
-    private function endTransaction()
+    private function endTransaction() : bool
```
morozov: This should be `void`.

lib/Doctrine/DBAL/Driver/PingableConnection.php
```diff
@@ -10,8 +10,6 @@
     /**
      * Pings the database server to determine if the connection is still
      * available. Return true/false based on if that was successful or not.
-     *
-     * @return bool
      */
-    public function ping();
+    public function ping() : bool;
```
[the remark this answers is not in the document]
morozov: This is being worked on in #3485.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
```diff
@@ -142,32 +142,23 @@ private function bind($parameter, &$variable, int $parameterType, int $dataType)
...
-    public function closeCursor()
+    public function closeCursor() : void
     {
         if (! $this->stmt) {
-            return false;
+            return;
         }
 
         $this->bindParam = [];
+        $this->result    = false;
```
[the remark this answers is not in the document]
morozov: Reverted.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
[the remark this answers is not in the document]
morozov: Addressed in #3491.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -783,75 +691,61 @@ public function getTrimExpression($str, $mode = TrimMode::UNSPECIFIED, $char = f
...
-     *
-     * @return string
-     *
-     * @throws DBALException If not supported on this platform.
+     * @param string $str      Literal string.
+     * @param string $substr   Literal string to find.
+     * @param int    $startPos Position to start at, beginning of string by default.
      */
-    public function getLocateExpression($str, $substr, $startPos = false)
+    public function getLocateExpression(string $str, string $substr, int $startPos = 1) : string
```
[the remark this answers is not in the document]
morozov: Addressed in #3494.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -881,13 +773,11 @@ public function getSubstringExpression($value, $from, $length = null)
     /**
      * Returns a SQL snippet to concatenate the given expressions.
      *
-     * Accepts an arbitrary number of string parameters. Each parameter must contain an expression.
-     *
-     * @return string
+     * @param string[] ...$expressions
      */
-    public function getConcatExpression()
+    public function getConcatExpression(string ...$expressions) : string
```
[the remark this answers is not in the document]
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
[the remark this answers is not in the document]
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1005,254 +879,169 @@ public function getCosExpression($value)
...
      *
-     * @param string $date
-     * @param int    $minutes
-     *
-     * @return string
-     *
      * @throws DBALException If not supported on this platform.
      */
-    public function getDateAddMinutesExpression($date, $minutes)
+    public function getDateAddMinutesExpression(string $date, string $minutes) : string
```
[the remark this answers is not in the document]
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1005,254 +879,169 @@ public function getCosExpression($value)
...
      *
-     * @param string $date
-     * @param int    $hours
-     *
-     * @return string
-     *
      * @throws DBALException If not supported on this platform.
      */
-    public function getDateAddHourExpression($date, $hours)
+    public function getDateAddHourExpression(string $date, string $hours) : string
```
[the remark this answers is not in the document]
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1005,254 +879,169 @@ public function getCosExpression($value)
...
      *
-     * @param string $date
-     * @param int    $days
-     *
-     * @return string
-     *
      * @throws DBALException If not supported on this platform.
      */
-    public function getDateAddDaysExpression($date, $days)
+    public function getDateAddDaysExpression(string $date, string $days) : string
```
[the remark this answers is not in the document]
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1005,254 +879,169 @@ public function getCosExpression($value)
...
      *
-     * @param string $date
-     * @param int    $hours
-     *
-     * @return string
-     *
      * @throws DBALException If not supported on this platform.
      */
-    public function getDateSubHourExpression($date, $hours)
+    public function getDateSubHourExpression(string $date, string $hours) : string
```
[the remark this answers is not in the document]
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1005,254 +879,169 @@ public function getCosExpression($value)
...
      *
-     * @param string $date
-     * @param int    $days
-     *
-     * @return string
-     *
      * @throws DBALException If not supported on this platform.
      */
-    public function getDateSubDaysExpression($date, $days)
+    public function getDateSubDaysExpression(string $date, string $days) : string
```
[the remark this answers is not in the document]
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1005,254 +879,169 @@ public function getCosExpression($value)
...
      *
-     * @param string $date
-     * @param int    $weeks
-     *
-     * @return string
-     *
      * @throws DBALException If not supported on this platform.
      */
-    public function getDateAddWeeksExpression($date, $weeks)
+    public function getDateAddWeeksExpression(string $date, string $weeks) : string
```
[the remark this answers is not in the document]
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1005,254 +879,169 @@ public function getCosExpression($value)
...
      *
-     * @param string $date
-     * @param int    $weeks
-     *
-     * @return string
-     *
      * @throws DBALException If not supported on this platform.
      */
-    public function getDateSubWeeksExpression($date, $weeks)
+    public function getDateSubWeeksExpression(string $date, string $weeks) : string
```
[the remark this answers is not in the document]
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1005,254 +879,169 @@ public function getCosExpression($value)
...
      *
-     * @param string $date
-     * @param int    $seconds
-     *
-     * @return string
-     *
      * @throws DBALException If not supported on this platform.
      */
-    public function getDateSubSecondsExpression($date, $seconds)
+    public function getDateSubSecondsExpression(string $date, string $seconds) : string
```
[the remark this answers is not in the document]
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1005,254 +879,169 @@ public function getCosExpression($value)
...
      *
-     * @param string $date
-     * @param int    $minutes
-     *
-     * @return string
-     *
      * @throws DBALException If not supported on this platform.
      */
-    public function getDateSubMinutesExpression($date, $minutes)
+    public function getDateSubMinutesExpression(string $date, string $minutes) : string
```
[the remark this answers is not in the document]
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1005,254 +879,169 @@ public function getCosExpression($value)
...
      *
-     * @param string $date
-     * @param int    $months
-     *
-     * @return string
-     *
      * @throws DBALException If not supported on this platform.
      */
-    public function getDateAddMonthExpression($date, $months)
+    public function getDateAddMonthExpression(string $date, string $months) : string
```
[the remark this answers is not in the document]
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1005,254 +879,169 @@ public function getCosExpression($value)
...
      *
-     * @param string $date
-     * @param int    $months
-     *
-     * @return string
-     *
      * @throws DBALException If not supported on this platform.
      */
-    public function getDateSubMonthExpression($date, $months)
+    public function getDateSubMonthExpression(string $date, string $months) : string
```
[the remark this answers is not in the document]
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: Addressed in #3498.

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
morozov: Actually, it shouldn't be nullable. For a field with the non-variable length, there's no point in having any default. I have a branch with some rework in this area.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
```diff
@@ -142,32 +142,23 @@ private function bind($parameter, &$variable, int $parameterType, int $dataType)
...
+        db2_free_result($this->stmt);
     }
 
     /**
      * {@inheritdoc}
      */
-    public function columnCount()
+    public function columnCount() : int
     {
-        if (! $this->stmt) {
```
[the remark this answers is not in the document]
morozov: This has been reverted. I don't remember why this change was made.

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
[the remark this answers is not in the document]
morozov: Marking as resolved for now (addressed in #3486). Making the API nicer is not a requirement for this specific issue.

lib/Doctrine/DBAL/Driver/DriverException.php
```diff
@@ -22,19 +22,10 @@
      */
     public function getErrorCode();
 
-    /**
-     * Returns the driver error message.
-     *
-     * @return string
-     */
-    public function getMessage();
```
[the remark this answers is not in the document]
morozov: Why? To me, it looks like a minor cosmetic improvement which may confuse the reader of the code.

After the changes in `phpcs.xml.dist` are merged, all implementors will be required to declare the return type which will be enforced by the standard.

jwage (approved):

PR #3353, opened 2018-11-22
Allow setting previous exception in AbstractDriverException

elprawn:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

It is useful to be able to set the previous exception in AbstractDriverException, for logging and debugging purposes. In the case where the doctrine driver is an adapter for the driver of the underlying technology, that driver may have its own exceptions, which are currently lost if they are converted to AbstractDriverException. This change allows the exception from the underlying context to be retrieved.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractDriverException.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractDriverException.php
@@ -3,6 +3,7 @@
 namespace Doctrine\DBAL\Driver;
 
 use Exception;
+use Throwable;
 
 /**
  * Abstract base implementation of the {@link DriverException} interface.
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Driver/AbstractDriverException.php
```diff
@@ -27,10 +28,15 @@
...
+     * @param Throwable|null  $previous  The previous exception.
      */
-    public function __construct($message, $sqlState = null, $errorCode = null)
-    {
-        parent::__construct($message);
+    public function __construct(
+        $message,
+        $sqlState = null,
+        $errorCode = null,
+        ?Throwable $previous = null
```
Ocramius: Disagree with this change: `AbstractDriverException` is an internal exception type used by DBAL, and shouldn't be used/extended by consumers. This is most likely to be fixed by adding an `@internal`

Ocramius (changes requested):

Ocramius:
Closing as "Won't fix" - please use a custom implementation of a `DriverException` instead.

PR #3354, opened 2018-11-24
Extract type factory and registry from Type into TypeRegistry

Majkl578:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Related issues | #2841

#### Summary

This PR splits Type factory, registry and abstract ancestor for all implementors.
Factory and registry stuff is moved to new TypeRegistry which holds no default types on its own.

##### TODO
* [x] Make `addType()` and `overrideType()` accept instances or factories instead of class name strings? This would remove internal API dependency on `Type`'s constructor and allow users use it.

##### Further scope
* Add support type aliases on registry level (internally one level of indirection, not leaked outside registry).
* Make TypeRegistry per-connection, so different platforms may declare different types.
* Eliminate static access to `Type::*()` registry methods.
* Deprecate and remove `Type::getName()` and leverage ~`TypeRegistry::getName()`~`TypeRegistry::lookupName()` instead.
* Un`@internal`ize TypeRegistry.

cc @greg0ire

### DIFF
--- a/lib/Doctrine/DBAL/DBALException.php
+++ b/lib/Doctrine/DBAL/DBALException.php
@@ -6,6 +6,7 @@ use Doctrine\DBAL\Driver\DriverException as DriverExceptionInterface;
 use Doctrine\DBAL\Driver\ExceptionConverterDriver;
 use Doctrine\DBAL\Exception\DriverException;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\Type;
 use Exception;
 use Throwable;
 use function array_map;
@@ -18,6 +19,7 @@ use function is_resource;
 use function is_string;
 use function json_encode;
 use function preg_replace;
+use function spl_object_hash;
 use function sprintf;
 
 class DBALException extends Exception
@@ -276,5 +278,17 @@ class DBALException extends Exception
     public static function typeNotFound($name)
     {
         return new self('Type to be overwritten ' . $name . ' does not exist.');
+    }
+
+    public static function typeNotRegistered(Type $type) : self
+    {
+        return new self(sprintf('Type of the class %s@%s is not registered.', get_class($type), spl_object_hash($type)));
+    }
+
+    public static function typeAlreadyRegistered(Type $type) : self
+    {
+        return new self(
+            sprintf('Type of the class %s@%s is already registered.', get_class($type), spl_object_hash($type))
+        );
     }
 }
--- a/lib/Doctrine/DBAL/Types/Type.php
+++ b/lib/Doctrine/DBAL/Types/Type.php
@@ -5,6 +5,8 @@ namespace Doctrine\DBAL\Types;
 use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\ParameterType;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
+use function array_map;
+use function get_class;
 use function str_replace;
 use function strrpos;
 use function substr;
@@ -16,76 +18,70 @@ use function substr;
  */
 abstract class Type
 {
-    public const TARRAY               = 'array';
-    public const SIMPLE_ARRAY         = 'simple_array';
-    public const JSON_ARRAY           = 'json_array';
-    public const JSON                 = 'json';
     public const BIGINT               = 'bigint';
+    public const BINARY               = 'binary';
+    public const BLOB                 = 'blob';
     public const BOOLEAN              = 'boolean';
+    public const DATE                 = 'date';
+    public const DATE_IMMUTABLE       = 'date_immutable';
+    public const DATEINTERVAL         = 'dateinterval';
     public const DATETIME             = 'datetime';
     public const DATETIME_IMMUTABLE   = 'datetime_immutable';
     public const DATETIMETZ           = 'datetimetz';
     public const DATETIMETZ_IMMUTABLE = 'datetimetz_immutable';
-    public const DATE                 = 'date';
-    public const DATE_IMMUTABLE       = 'date_immutable';
-    public const TIME                 = 'time';
-    public const TIME_IMMUTABLE       = 'time_immutable';
     public const DECIMAL              = 'decimal';
-    public const INTEGER              = 'integer';
-    public const OBJECT               = 'object';
-    public const SMALLINT             = 'smallint';
-    public const STRING               = 'string';
-    public const TEXT                 = 'text';
-    public const BINARY               = 'binary';
-    public const BLOB                 = 'blob';
     public const FLOAT                = 'float';
     public const GUID                 = 'guid';
-    public const DATEINTERVAL         = 'dateinterval';
-
-    /**
-     * Map of already instantiated type objects. One instance per type (flyweight).
-     *
-     * @var self[]
-     */
-    private static $_typeObjects = [];
+    public const INTEGER              = 'integer';
+    public const JSON                 = 'json';
+    public const JSON_ARRAY           = 'json_array';
+    public const OBJECT               = 'object';
+    public const SIMPLE_ARRAY         = 'simple_array';
+    public const SMALLINT             = 'smallint';
+    public const STRING               = 'string';
+    public const TARRAY               = 'array';
+    public const TEXT                 = 'text';
+    public const TIME                 = 'time';
+    public const TIME_IMMUTABLE       = 'time_immutable';
 
     /**
      * The map of supported doctrine mapping types.
-     *
-     * @var string[]
      */
-    private static $_typesMap = [
-        self::TARRAY => ArrayType::class,
-        self::SIMPLE_ARRAY => SimpleArrayType::class,
-        self::JSON_ARRAY => JsonArrayType::class,
-        self::JSON => JsonType::class,
-        self::OBJECT => ObjectType::class,
-        self::BOOLEAN => BooleanType::class,
-        self::INTEGER => IntegerType::class,
-        self::SMALLINT => SmallIntType::class,
-        self::BIGINT => BigIntType::class,
-        self::STRING => StringType::class,
-        self::TEXT => TextType::class,
-        self::DATETIME => DateTimeType::class,
-        self::DATETIME_IMMUTABLE => DateTimeImmutableType::class,
-        self::DATETIMETZ => DateTimeTzType::class,
+    private const BUILTIN_TYPES_MAP = [
+        self::BIGINT               => BigIntType::class,
+        self::BINARY               => BinaryType::class,
+        self::BLOB                 => BlobType::class,
+        self::BOOLEAN              => BooleanType::class,
+        self::DATE                 => DateType::class,
+        self::DATE_IMMUTABLE       => DateImmutableType::class,
+        self::DATEINTERVAL         => DateIntervalType::class,
+        self::DATETIME             => DateTimeType::class,
+        self::DATETIME_IMMUTABLE   => DateTimeImmutableType::class,
+        self::DATETIMETZ           => DateTimeTzType::class,
         self::DATETIMETZ_IMMUTABLE => DateTimeTzImmutableType::class,
-        self::DATE => DateType::class,
-        self::DATE_IMMUTABLE => DateImmutableType::class,
-        self::TIME => TimeType::class,
-        self::TIME_IMMUTABLE => TimeImmutableType::class,
-        self::DECIMAL => DecimalType::class,
-        self::FLOAT => FloatType::class,
-        self::BINARY => BinaryType::class,
-        self::BLOB => BlobType::class,
-        self::GUID => GuidType::class,
-        self::DATEINTERVAL => DateIntervalType::class,
+        self::DECIMAL              => DecimalType::class,
+        self::FLOAT                => FloatType::class,
+        self::GUID                 => GuidType::class,
+        self::INTEGER              => IntegerType::class,
+        self::JSON                 => JsonType::class,
+        self::JSON_ARRAY           => JsonArrayType::class,
+        self::OBJECT               => ObjectType::class,
+        self::SIMPLE_ARRAY         => SimpleArrayType::class,
+        self::SMALLINT             => SmallIntType::class,
+        self::STRING               => StringType::class,
+        self::TARRAY               => ArrayType::class,
+        self::TEXT                 => TextType::class,
+        self::TIME                 => TimeType::class,
+        self::TIME_IMMUTABLE       => TimeImmutableType::class,
     ];
 
+    /** @var TypeRegistry|null */
+    private static $typeRegistry;
+
     /**
-     * Prevents instantiation and forces use of the factory method.
+     * @internal Do not instantiate directly - use {@see Type::addType()} method instead.
      */
-    final private function __construct()
+    final public function __construct()
     {
     }
 
@@ -149,6 +145,29 @@ abstract class Type
     abstract public function getName();
 
     /**
+     * @internal This method is only to be used within DBAL for forward compatibility purposes. Do not use directly.
+     */
+    final public static function getTypeRegistry() : TypeRegistry
+    {
+        if (self::$typeRegistry === null) {
+            self::$typeRegistry = self::createTypeRegistry();
+        }
+
+        return self::$typeRegistry;
+    }
+
+    private static function createTypeRegistry() : TypeRegistry
+    {
+        $registry = new TypeRegistry();
+
+        foreach (self::BUILTIN_TYPES_MAP as $name => $class) {
+            $registry->register($name, new $class());
+        }
+
+        return $registry;
+    }
+
+    /**
      * Factory method to create type instances.
      * Type instances are implemented as flyweights.
      *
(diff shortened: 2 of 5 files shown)


lib/Doctrine/DBAL/Types/Type.php
```diff
@@ -42,50 +42,44 @@
...
 
+    /** @var TypeRegistry|null */
+    private static $typeRegistry;
+
     /**
-     * Prevents instantiation and forces use of the factory method.
+     * @internal Do not instantiate directly - use the factory method instead.
      */
-    final private function __construct()
+    final public function __construct()
```
morozov: Is this method kept only for the sake of being marked `@internal`? As far as I understand, eventually `Type` should be converted to an interface and not have a constructor at all. Shoudn't it?

lib/Doctrine/DBAL/Types/Type.php
```diff
@@ -42,50 +42,44 @@
...
-        self::INTEGER => IntegerType::class,
-        self::SMALLINT => SmallIntType::class,
-        self::BIGINT => BigIntType::class,
-        self::STRING => StringType::class,
-        self::TEXT => TextType::class,
-        self::DATETIME => DateTimeType::class,
-        self::DATETIME_IMMUTABLE => DateTimeImmutableType::class,
-        self::DATETIMETZ => DateTimeTzType::class,
+    private const BUILTIN_TYPES_MAP = [
+        self::TARRAY               => ArrayType::class,
```
morozov: For easier lookup, could this be sorted alphabetically by key?

lib/Doctrine/DBAL/Types/Type.php
```diff
@@ -160,14 +174,7 @@ public function getDefaultLength(AbstractPlatform $platform)
      */
     public static function getType($name)
```
morozov: Is this and other extracted methods supposed to be marked `@deprecated`?

lib/Doctrine/DBAL/Types/TypeRegistry.php
```diff
@@ -0,0 +1,118 @@
...
+     * @var Type[]
+     */
+    private $typeInstances = [];
+
+    /**
+     * Map of defined types and their classes
+     *
+     * @var string[]
+     */
+    private $typesMap = [];
```
morozov: `$typeClasses` might be better.

lib/Doctrine/DBAL/Types/TypeRegistry.php
```diff
@@ -0,0 +1,118 @@
...
+ * @internal
+ */
+final class TypeRegistry
+{
+    /**
+     * Map of already instantiated type objects. One instance per type (flyweight).
+     *
+     * @var Type[]
+     */
+    private $typeInstances = [];
```
morozov: This data structure doesn't prevent multiple keys pointing to the same instance, neither does the implementation of `addType()` (multiple names → same class). Would it make sense to use `SplObjectStorage` where instances were keys?

lib/Doctrine/DBAL/Types/TypeRegistry.php
```diff
@@ -0,0 +1,118 @@
...
+     *
+     * @throws DBALException
+     */
+    public function overrideType(string $name, string $className) : void
+    {
+        if (! isset($this->typesMap[$name])) {
+            throw DBALException::typeNotFound($name);
+        }
+
+        if (isset($this->typeInstances[$name])) {
```
morozov: Wouldn't `unset()` alone be enough?

lib/Doctrine/DBAL/Types/TypeRegistry.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * Adds a custom type to the type map.
+     *
+     * @param string $name      The name of the type. This should correspond to what getName() returns.
+     * @param string $className The class name of the custom type.
+     *
+     * @throws DBALException
+     */
+    public function addType(string $name, string $className) : void
```
morozov: Could `$class` be used for naming consistency and consistency with the code in `Type`? The `$className` seems to be coming from JavaScript where one cannot use reserved keywords as variable names.

lib/Doctrine/DBAL/Types/TypeRegistry.php
```diff
@@ -0,0 +1,118 @@
...
+        $this->typesMap[$name] = $className;
+    }
+
+    /**
+     * Gets the types array map which holds all registered types and the corresponding
+     * type class
+     *
+     * @return string[]
+     */
+    public function getTypesMap() : array
```
morozov: Besides BC, what could this method be used for? Unless it's by design, we should mark it `@internal`.

Given that we have two maps internally, and `getType()` returns an instance, would it make sense to name it as `getTypeClasses()`?

morozov:
> Make `addType()` and `overrideType()` accept instances or factories instead of class name strings? This would remove internal API dependency on `Type`'s constructor and allow users use it.

Would it make sense to accept type instances instead? Otherwise, the factory would still depend on the constructor signatures which would have to be the same. Given that objects are lightweight, there shouldn't be much overhead in just registering them all at once.

lib/Doctrine/DBAL/Types/Type.php
```diff
@@ -148,6 +142,26 @@ public function getDefaultLength(AbstractPlatform $platform)
...
+    private static function createTypeRegistry() : TypeRegistry
+    {
+        $registry = new TypeRegistry();
+
+        foreach (self::BUILTIN_TYPES_MAP as $name => $class) {
+            $registry->addType($name, $class);
+        }
+
+        return $registry;
+    }
```
greg0ire: Why make this a separate method? It's only used once I think.

lib/Doctrine/DBAL/Types/Type.php
Majkl578: It's to force empty constructor, but since Types are no longer instantiated from within themselves, it must be public.

lib/Doctrine/DBAL/Types/Type.php
Majkl578: Eventually yes, but not at this moment - there is no way to use the registry directly, it's only used internally. We may design different approach.

lib/Doctrine/DBAL/Types/TypeRegistry.php
Majkl578: This structure doesn't, but current API does. If we decide to change API, then we may need to change this too.

lib/Doctrine/DBAL/Types/TypeRegistry.php
Majkl578: Whole TypeRegistry is `@internal` now. But yes, this method is useless and only for compatibility with `Type`.

lib/Doctrine/DBAL/Types/TypeRegistry.php
Majkl578: Yes, dunno, maybe it served as some optimization? `isset` is O(1) while `unset` is O(n).

lib/Doctrine/DBAL/Types/Type.php
Majkl578: Readability.

lib/Doctrine/DBAL/Types/TypeRegistry.php
Majkl578: `SplObjectStorage` may not be the best fit for this use case since the primary use case is _finding objects by keys_. IMO `array` is fine as long as we maintain consistency inside TypeRegistry and don't leak it outside.

lib/Doctrine/DBAL/Types/Type.php
morozov: So this is only needed because the registry instantiates types. It's fine as a temporary solution, but it shouldn't be needed if instead of identifying types by classes, we identify them by instances implementing `Type`.

lib/Doctrine/DBAL/Types/TypeRegistry.php
morozov: Agree with the above. I still don't see how `register()` prevents subsequent calls like `register('type1', T::class)`, `register('type2', T::class)`.

Majkl578: Please see the new version (just pushed) where register() accepts instances directly.

lib/Doctrine/DBAL/Types/Type.php
Majkl578: That's my thought, once we only rely on Type interface (not the abstract class), we can remove restriction on the constructor. Right now we can't since we instantiate the types without parameters.

Majkl578:
Pushed new version where TypeRegistry accepts instances directly.

lib/Doctrine/DBAL/Types/Type.php
morozov: Agreed.

lib/Doctrine/DBAL/Types/TypeRegistry.php
```diff
@@ -0,0 +1,114 @@
...
+use Doctrine\DBAL\DBALException;
+use function array_search;
+use function in_array;
+
+/**
+ * @internal
+ */
+final class TypeRegistry
+{
+    /** @var Type[] Map of already instantiated type objects. */
```
morozov: The "already instantiated" seems no longer relevant.

lib/Doctrine/DBAL/Types/Type.php
```diff
@@ -148,6 +144,26 @@ public function getDefaultLength(AbstractPlatform $platform)
      */
     abstract public function getName();
 
+    private static function getTypeRegistry() : TypeRegistry
+    {
+        if (self::$typeRegistry === null) {
+            return self::$typeRegistry = self::createTypeRegistry();
```
tarlepp: what about removing `return` on this, single return point on method should be cleaner

lib/Doctrine/DBAL/Types/TypeRegistry.php
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * @throws DBALException
+     */
+    public function override(string $name, Type $type) : void
+    {
+        if (! isset($this->instances[$name])) {
+            throw DBALException::typeNotFound($name);
+        }
+
+        if (! in_array($this->findTypeName($type), [$name, null], true)) {
```
Taluu: why don't you throw if the type is already registered to the same name ? Unless you consider that doing it again is harmless (it is, but for consistency check maybe it'd be better to throw imo)

Taluu (dismissed):
👍

lib/Doctrine/DBAL/Types/TypeRegistry.php
Majkl578: > Unless you consider that doing it again is harmless

Exactly for this reason, and also because the method is named _override_, so overriding by _itself_ doesn't seem harmful (unlike setting the same in _register_).

lib/Doctrine/DBAL/Types/TypeRegistry.php
```diff
@@ -0,0 +1,114 @@
...
+    {
+        if (! isset($this->instances[$name])) {
+            throw DBALException::unknownColumnType($name);
+        }
+
+        return $this->instances[$name];
+    }
+
+    /**
+     * Find a name for the given type.
```
greg0ire: On line 51, you use the 3rd person, and here the imperative, maybe be stick to either one?

lib/Doctrine/DBAL/Types/Type.php
```diff
@@ -182,11 +191,7 @@ public static function getType($name)
      */
     public static function addType($name, $className)
```
tarlepp: Hmm, should these `$name` and `$className` also have types defined, because later on eg. that `register` method is expecting certain types.

Majkl578: That will not be possible in master due to BC - phpDoc types are sufficient here.

tests/Doctrine/Tests/DBAL/Types/TypeTest.php
```diff
@@ -0,0 +1,34 @@
...
+namespace Doctrine\Tests\DBAL\Types;
+
+use Doctrine\DBAL\Types\Type;
+use PHPUnit\Framework\TestCase;
+use ReflectionClass;
+
+class TypeTest extends TestCase
+{
+    /**
+     * @dataProvider defaultTypesProvider()
```
enumag: Drop the parenthesis, they are not used in any other tests except `StatementIteratorTest`. Drop them there too.

Majkl578: @morozov WDYT? Didn't we agree to prefer them?

morozov: Personally, I like it more with parentheses since this is how functions and methods are usually notated in PHP documentation. It'd be nice to have it clarified in the standard.

Majkl578:
Rebased onto current master and addressed remaining open discussion.

lib/Doctrine/DBAL/DBALException.php
```diff
@@ -277,4 +279,16 @@ public static function typeNotFound($name)
...
+    {
+        return new self(sprintf('Type of the class %s@%s is not registered.', get_class($type), spl_object_hash($type)));
+    }
+
+    public static function typeAlreadyRegistered(Type $type) : self
+    {
+        return new self(
+            sprintf('Type of the class %s@%s is already registered.', get_class($type), spl_object_hash($type))
+        );
+    }
```
greg0ire: On one hand, not a big fan of making a big exception class that grows with time, on the other hand, if these exceptions are not meant to be caught but to help the user figure out their issue, maybe this is fine?

lib/Doctrine/DBAL/Types/Type.php
```diff
@@ -245,7 +245,12 @@ public function getBindingType()
      */
     public static function getTypesMap()
     {
-        return self::$_typesMap;
+        return array_map(
+            static function (Type $type) : string {
+                return get_class($type);
+            },
```
greg0ire: why not directly `array_map('get_class', self::getTypeRegistry()->getMap())`? Can't SA tools figure this out? Or is it because `get_class` might return `false`?

lib/Doctrine/DBAL/Types/TypeRegistry.php
```diff
@@ -0,0 +1,118 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Types;
+
+use Doctrine\DBAL\DBALException;
+use function array_search;
+use function in_array;
+
+/**
+ * Type registry is responsible for holding holding a map of known DBAL types.
```
greg0ire:
```suggestion
 * Type registry is responsible for holding a map of known DBAL types.
```

lib/Doctrine/DBAL/Types/TypeRegistry.php
```diff
@@ -0,0 +1,118 @@
...
+ * Type instances are implemented as flyweight pattern.
```
greg0ire:
```suggestion
 * Type instances implement the flyweight pattern.
```

lib/Doctrine/DBAL/Types/TypeRegistry.php
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * @internal TypeRegistry class exists for forward compatibility, its API should not be considered stable.
```
greg0ire:
```suggestion
 * @internal the TypeRegistry class exists for forward compatibility, its API should not be considered stable.
```

greg0ire (changes requested):

lib/Doctrine/DBAL/Types/Type.php
Majkl578: I personally always avoid string callbacks for these functions. It can't return false in this context.

lib/Doctrine/DBAL/Types/TypeRegistry.php
Majkl578: This is wrong, the types don't implement the pattern, they are used as such. Maybe using different word than ambiguous "implement" is a good idea.

lib/Doctrine/DBAL/Types/TypeRegistry.php
Majkl578: IMO not, TypeRegistry is a name.

lib/Doctrine/DBAL/Types/Type.php
greg0ire: ok then

lib/Doctrine/DBAL/DBALException.php
Majkl578: It's only for 2.x.

lib/Doctrine/DBAL/Types/TypeRegistry.php
greg0ire: "Type instances are used as implementations of the flyweight pattern" then?

lib/Doctrine/DBAL/Types/TypeRegistry.php
greg0ire: "TypeRegistry exists" would be correct because of this, but "TypeRegistry class exists" isn't I think.

greg0ire (approved):

morozov (approved):

Ocramius (approved):

Issue #3336, opened 2018-10-31
Comparator incorrectly diffs default values (regression in master)


Majkl578:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.9.0-dev

#### Summary

After applying Doctrine CS 5.0, weak comparisons became strict comparisons.
One of the affected places is Comparator which compares default values _as is_ (always `string|null` from the DB while it's `scalar|null` when declared in ORM):
https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/Comparator.php#L435

#### Previous behaviour
The following test works with snapshot/develop/2018-04-17
https://github.com/doctrine/doctrine2/blob/2726636e55fdbdf6fb1ed466f22d34de116fb8f9/tests/Doctrine/Tests/ORM/Functional/SchemaTool/DBAL483Test.php#L24-L41

#### Current behavior

Comparator now produces diff SQL when `default` is `(int) 0`.

#### How to reproduce

Run ORM 2.6 test suite with DBAL 2.9.0-dev.
https://travis-ci.org/doctrine/doctrine2/jobs/448657137#L608

PR #3355, opened 2018-11-24
Implemented comparison of default values as strings regardless of their PHP types

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3336

Compare default values as strings (as they appear in SQL and introspected schema) regardless of their PHP types.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Comparator.php
+++ b/lib/Doctrine/DBAL/Schema/Comparator.php
@@ -432,12 +432,10 @@ class Comparator
             $changedProperties[] = 'comment';
         }
 
-        if ($properties1['default'] !== $properties2['default'] ||
-            // Null values need to be checked additionally as they tell whether to create or drop a default value.
-            // null != 0, null != false, null != '' etc. This affects platform's table alteration SQL generation.
-            ($properties1['default'] === null && $properties2['default'] !== null) ||
-            ($properties2['default'] === null && $properties1['default'] !== null)
-        ) {
+        // Null values need to be checked additionally as they tell whether to create or drop a default value.
+        // null != 0, null != false, null != '' etc. This affects platform's table alteration SQL generation.
+        if (($properties1['default'] === null) !== ($properties2['default'] === null)
+            || (string) $properties1['default'] !== (string) $properties2['default']) {
             $changedProperties[] = 'default';
         }
 
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
@@ -23,6 +23,8 @@ class MySqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
             return;
         }
 
+        $this->resetSharedConn();
+
         Type::addType('point', MySqlPointType::class);
     }
 
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/ComparatorTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/ComparatorTest.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL\Functional\Schema;
+
+use Doctrine\DBAL\Schema\AbstractSchemaManager;
+use Doctrine\DBAL\Schema\Comparator;
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\Tests\DbalFunctionalTestCase;
+
+class ComparatorTest extends DbalFunctionalTestCase
+{
+    /** @var AbstractSchemaManager */
+    private $schemaManager;
+
+    /** @var Comparator */
+    private $comparator;
+
+    protected function setUp()
+    {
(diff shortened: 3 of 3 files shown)


tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
```diff
@@ -23,6 +23,8 @@ protected function setUp()
             return;
         }
 
+        $this->resetSharedConn();
```
morozov: This is needed since a new functional test is added prior to this one. In this case, the platform instance associated with the shared connection has already initialized the type mappings and won't reinitialize them with the newly added type.

lib/Doctrine/DBAL/Schema/Comparator.php
```diff
@@ -432,12 +433,12 @@ public function diffColumn(Column $column1, Column $column2)
...
-            // null != 0, null != false, null != '' etc. This affects platform's table alteration SQL generation.
-            ($properties1['default'] === null && $properties2['default'] !== null) ||
-            ($properties2['default'] === null && $properties1['default'] !== null)
-        ) {
+        $default1IsNull = $properties1['default'] === null;
+        $default2IsNull = $properties2['default'] === null;
+
+        // Null values need to be checked additionally as they tell whether to create or drop a default value.
+        // null != 0, null != false, null != '' etc. This affects platform's table alteration SQL generation.
+        if ($default1IsNull !== $default2IsNull || strcmp($properties1['default'], $properties2['default']) !== 0) {
```
Majkl578: From my understanding, the original problem was that `['default']` could be i.e. integer (when defined in ORM for example), so we need an explicit cast here otherwise we will run into problems when we enable `strict_types`.

Majkl578 (changes requested):

lib/Doctrine/DBAL/Schema/Comparator.php
morozov: > From my understanding, the original problem was that `['default']` could be i.e. integer (when defined in ORM for example)

Yes.

> so we need an explicit cast here otherwise we will run into problems when we enable `strict_types`.

Does it apply to builtin functions? Anyways, my original implementation was done under the assumption that `strcmp(mixed $value1, mixed $value2)` is the same as `(string) $value1 === (string) $value2` which is not true. `strcmp()` expects strings.

Majkl578: Yep it does: https://3v4l.org/BVrSK

Majkl578 (approved):
:+1:, fixes the ORM issue.

lib/Doctrine/DBAL/Schema/Comparator.php
```diff
@@ -432,12 +432,10 @@ public function diffColumn(Column $column1, Column $column2)
...
 
-        if ($properties1['default'] !== $properties2['default'] ||
-            // Null values need to be checked additionally as they tell whether to create or drop a default value.
-            // null != 0, null != false, null != '' etc. This affects platform's table alteration SQL generation.
-            ($properties1['default'] === null && $properties2['default'] !== null) ||
-            ($properties2['default'] === null && $properties1['default'] !== null)
-        ) {
+        // Null values need to be checked additionally as they tell whether to create or drop a default value.
+        // null != 0, null != false, null != '' etc. This affects platform's table alteration SQL generation.
+        if (($properties1['default'] === null) !== ($properties2['default'] === null)
```
svycka: @morozov not sure that's not my problem but with annotation like this:
```
@ORM\Column(type="boolean", nullable=false, options={"default" = false})
```
it always generates migration becouse of this fix. compared with old version https://3v4l.org/HBLG7 :
````php
<?php
$changedProperties = [];
$properties1['default'] = false;
$properties2['default'] = 0;

if ($properties1['default'] != $properties2['default'] ||
    (null === $properties1['default'] && null !== $properties2['default']) ||
    (null === $properties2['default'] && null !== $properties1['default'])
) {
    $changedProperties[] = 'default';
}

var_dump($changedProperties);

if (($properties1['default'] === null) !== ($properties2['default'] === null)
    || (string) $properties1['default'] !== (string) $properties2['default']) {
    $changedProperties[] = 'default';
}

var_dump($changedProperties);
````
result:
```
array(0) {
}
array(1) {
  [0]=>
  string(7) "default"
}
```

svycka: If we do not support `true/false` values for boolean type maybe do not allow them?

Majkl578: Because `('0' == false) === true` but `('0' === (string) false) === false`.

svycka: @Majkl578 yes yes... but should I update my default value to `0` or this will be fixed?

morozov: I think we'll loosen `===` back to `==` for the patch release, but later, the missing component here seems to be `Type::convertToDatabaseValue()`. We're loosely comparing a PHP value with a DB value which may have its own not yet discovered edge cases.

PR #3356, opened 2018-11-24
Extract constants for built-in types from Type to Types

Majkl578:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Deprecations | yes
| Related issues | #2841

#### Summary

Move constants for built-in types to a separate constant-only class.

Some constants were renamed for clarity:
* `TARRAY`-> `ARRAY` (this was [due to a PHP 5 limitation](https://3v4l.org/Qaj1T))
* `DATE` -> `DATE_MUTABLE`
* `DATETIME` -> `DATETIME_MUTABLE`
* `DATETIMETZ` -> `DATETIMETZ_MUTABLE`
* `TIME` -> `TIME_MUTABLE`

Old constants are now deprecated.

##### Naming

I originally used `BuiltinTypes`, but it seemed strange, so I decided for `DefaultTypes`. Thoughts?

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/DB2Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/DB2Platform.php
@@ -8,6 +8,7 @@ use Doctrine\DBAL\Schema\Identifier;
 use Doctrine\DBAL\Schema\Index;
 use Doctrine\DBAL\Schema\TableDiff;
 use Doctrine\DBAL\Types\Type;
+use Doctrine\DBAL\Types\Types;
 use function array_merge;
 use function count;
 use function current;
@@ -96,7 +97,7 @@ class DB2Platform extends AbstractPlatform
      */
     public function isCommentedDoctrineType(Type $doctrineType)
     {
-        if ($doctrineType->getName() === Type::BOOLEAN) {
+        if ($doctrineType->getName() === Types::BOOLEAN) {
             // We require a commented boolean type in order to distinguish between boolean and smallint
             // as both (have to) map to the same native type.
             return true;
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,16 @@
+# Upgrade to 2.10
+
+## Deprecated `Type::*` constants
+
+The constants for built-in types have been moved from `Doctrine\DBAL\Types\Type` to a separate class `Doctrine\DBAL\Types\Types`.
+
+Some of the constants were renamed in the process:
+* `TARRAY`-> `ARRAY`
+* `DATE` -> `DATE_MUTABLE`
+* `DATETIME` -> `DATETIME_MUTABLE`
+* `DATETIMETZ` -> `DATETIMETZ_MUTABLE`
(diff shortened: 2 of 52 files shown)


lib/Doctrine/DBAL/Types/DefaultTypes.php
```diff
@@ -0,0 +1,43 @@
...
+    public const JSON                 = 'json';
+    public const OBJECT               = 'object';
+    public const SIMPLE_ARRAY         = 'simple_array';
+    public const SMALLINT             = 'smallint';
+    public const STRING               = 'string';
+    public const TEXT                 = 'text';
+    public const TIME_MUTABLE         = 'time';
+    public const TIME_IMMUTABLE       = 'time_immutable';
+
+    /** @deprecated json_array type is deprecated, use {@see BuiltinTypes::JSON} instead. */
```
kejwmen: Comments still refer to `BuiltinTypes`

Majkl578: Thanks, PhpStorm refactoring didn't catch these.

Taluu (dismissed):
`Types\BuiltIn` ? But `DefaultTypes` is good too.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -97,7 +98,7 @@ public function initializeDoctrineTypeMappings()
      */
     public function isCommentedDoctrineType(Type $doctrineType)
     {
-        if ($doctrineType->getName() === Type::BOOLEAN) {
+        if ($doctrineType->getName() === DefaultTypes::BOOLEAN) {
```
morozov: The name `DefaultTypes` (as well as `BuiltinTypes`) feels weird in this context. It creates an impression that the code works only with default types but not with custom. Would it make sense to rename it to just `Types`? It's irrelevant whether they are default, built-in or custom in most contexts.

It may be hard to come up with a decent name because this class doesn't seem to have any purpose. What exactly is the reason for extraction?

enumag (dismissed):

lib/Doctrine/DBAL/Platforms/DB2Platform.php
Majkl578: I'm fine with `Types`, but it may cause confusion that there are only these types and no other custom ones.

> What exactly is the reason for extraction?

- We extracted constants elsewhere into constant-only classes, but not here.
- In order to make Type an interface, we need to do a cleanup of the interface and these names don't belong there.

Majkl578: Renamed to `Types`.

> What exactly is the reason for extraction?

Did I sufficiently answer your question in my previous comment?

morozov: Yes, thank you.

morozov (approved):

greg0ire (approved):

PR #3359, opened 2018-11-27
Removed DataAccessTest::testFetchColumnNonExistingIndex() since it covers a bug in PDO

### LINKED ISSUE #3358: DataAccessTest::testFetchColumnNonExistingIndex() covers a PDO bug
According to the discussion in https://github.com/php/php-src/pull/3655, the fact that `PDOStatement::fetchColumn()` returns `null` for an invalid column index, is a bug. Starting 7.2.14 and other newer PHP versions, `DataAccessTest::testFetchColumnNonExistingIndex()` will fail:
```
↪ php --version
PHP 7.2.14-dev (cli) (built: Nov 26 2018 15:14:02) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies

↪ phpunit --filter testFetchColumnNonExistingIndex tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
PHPUnit 7.4.0 by Sebastian Bergmann and contributors.

Testing Doctrine\Tests\DBAL\Functional\DataAccessTest
E                                                                                   1 / 1 (100%)

Time: 39 ms, Memory: 6.00MB

There was 1 error:

1) Doctrine\Tests\DBAL\Functional\DataAccessTest::testFetchColumnNonExistingIndex
Exception: [Doctrine\DBAL\Driver\PDOException] SQLSTATE[HY000]: General error: Invalid column index

ERRORS!
Tests: 1, Assertions: 0, Errors: 1.
```
This test was introduced as part of https://github.com/doctrine/dbal/pull/709 but it doesn't cover the issue which was fixed.

Instead of supporting this behavior by means of a workaround, I think we should just remove the test and document the changes. Invalid column index usually is a sign of a logical error in the client code and shouldn't be silently suppressed by the DBAL.

morozov:
Fixes #3358. I'll add the documentation once we agree on the approach.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade to 2.9
 
+## Deprecated `Statement::fetchColumn()` with an invalid index
+
+Calls to `Statement::fetchColumn()` with an invalid column index currently return `NULL`. In the future, such calls will result in a exception.
+
 ## Deprecated `Configuration::getFilterSchemaAssetsExpression()`, `::setFilterSchemaAssetsExpression()` and `AbstractSchemaManager::getFilterSchemaAssetsExpression()`.
 
 Regular expression-based filters are hard to extend by combining together. Instead, you may use callback-based filers via `::getSchemaAssetsFilter()` and `::getSchemaAssetsFilter()`. Callbacks can use regular expressions internally.
--- a/tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
@@ -891,22 +891,6 @@ class DataAccessTest extends DbalFunctionalTestCase
     /**
      * @group DBAL-1028
      */
-    public function testFetchColumnNonExistingIndex()
-    {
-        if ($this->connection->getDriver()->getName() === 'pdo_sqlsrv') {
-            $this->markTestSkipped(
-                'Test does not work for pdo_sqlsrv driver as it throws a fatal error for a non-existing column index.'
-            );
-        }
-
-        self::assertNull(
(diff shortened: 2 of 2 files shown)


deeky666 (dismissed):
:+1:

deeky666:
@morozov looking at #709 I think we should throw an exception instead of returning `null`, just like PDO now triggers an error natively. What do you think? In the end it is a client code error which should not be silenced...

UPGRADE.md
```diff
@@ -1,5 +1,9 @@
 # Upgrade to 2.9
 
+## Deprecated `Statement::fetchColumn()` with an invalid index
+
+Calls to `Statement::fetchColumn()` with an invalid column index currently return `NULL`. In the future, such calls will result in a exception.
```
deeky666: Well this is not entirely true considering that errors will be triggered by PHP itself and we are propagating them already. I would even question if this is a deprecation at all as it is undocumented behaviour. Our API does not tell anything about invalid index and `null`. Also why would someone expect such behaviour when it is clearly a bug in the client code? Even if someone would use `null` to tell whether the index in question is out of range, you couldn't even safely do that because `null` is not distinguishable from a regular index `null`.
IMO we could simply implement exceptions and maybe rather document it as "minor BC break" (even though it still is nothing we ever intended to support).

Majkl578: IMO better to not tease consumers with sneaky exceptions, unless it's a security issue or so.

Majkl578 (dismissed):

Majkl578:
I guess "_Removed_" is missing in the title / commit message?

Majkl578 (approved):

PR #3363, opened 2018-11-29
Remove redundant implements

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

`Driver\PDOSqlsrv\Connection` extends `Driver\PDOConnection`, which already implements `Driver\Connection`, therefore the `implements` is redundant and can be safely removed.


morozov (commented):
It looks like a non-breakinig change. Would it make sense to target it to master?

BenMorel:
Rebased on and retargeted to `master`.

morozov (approved):

morozov:
Thank you, @BenMorel.

PR #3367, opened 2018-11-29
[3.0] Rework driver exceptions

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | N/A

#### Summary

Consumers of the `Driver\Connection` API are not interested in catching individual exceptions implementing `Driver\DriverException`, and will always `catch (DriverException)`.

As such, the multiple implementations of DriverException provide little to no value, clutter the codebase, and force introducing a new exception class or [using `new class`](https://github.com/doctrine/dbal/blob/1c680e191af590039976cf18cc0d7a8e63fa3695/tests/Doctrine/Tests/DBAL/Driver/AbstractDriverTest.php#L86-L108) whenever we need to throw a DriverException from a new place.

This design was probably chosen for 2 reasons:

- a historical reason: `Driver\PDOConnection` originally extended `\PDO`, and as such `Driver\PDOException` extended `\PDOException`. To implement an interface common to all drivers, it was chosen to create an `interface DriverException`, that `Driver\PDOException` would implement, while still extending `\PDOException`. **This requirement is gone**, as the DBAL is evolving away from extending PDO, towards encapsulation.
- a practical reason: custom exceptions allow for [custom factory methods](https://github.com/doctrine/dbal/blob/1c680e191af590039976cf18cc0d7a8e63fa3695/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvException.php#L17); **these can easily be replaced** with [static methods on the Connection classes](https://github.com/doctrine/dbal/blob/46f6dfa99075967fbe8b35d8de83b41bab9817f7/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php#L194-L218).

#### Proposed changes

- `AbstractDriverException` is now a concrete `DriverException`, replacing the interface
- All sub-exceptions are now gone
- All exotic `new class ... extends AbstractDriverException` are now replaced with `new DriverException`
- Factory methods from sub-exceptions are moved to a `public static` method in the Connection, that can be used from within the Connection itself and from the Statement
- **All relevant Connection methods are now explicitly documented as throwing `DriverException`**

This last point is very important: so far, most of the methods had no documented exceptions, while `query()` and `exec()` were documented as throwing `DBALException`, **a contract that was never respected**:

- `PDOConnection::query()` and `exec()` throw `Driver\PDOException`, which is a `DriverException`, not a `DBALException`
- `DB2Exception::query()` and `exec()` throw `DB2Exception`, which is a `DriverException`
- `SQLSrvException::query()` and `exec()` throw `SQLSrvException` (via `SQLSrvStatement::execute()`), which is a `DriverException`
- ... and so on.


Majkl578:
FWIW please don't change exceptions hiearchy, it will be a nightmare to rebase #3131 even without this proposal.

> As such, the multiple implementations of DriverException provide little to no value, clutter the codebase, and force introducing a new exception class

This is decision made for ORM 3.0 and also planned for DBAL 3.0 - base exceptions will be interfaces, specific exceptions will be _named exceptions_.

BenMorel:
> This is decision made for ORM 3.0 and also planned for DBAL 3.0 - base exceptions will be interfaces, specific exceptions will be named exceptions.

Makes sense, but I wouldn't consider that `MysqliException` or `SqlsrvException` qualify as *specific exceptions*: `DriverException` is the only exception that should be part of the Connection contract, and there is no use case you would want to catch `MysqliException` specifically. These exceptions are only there for historical reasons IMO.

Majkl578:
IMO for driver exceptions, the tree should look like:
```
Throwable
 \
  DbalException
   \
    DriverException
     \
      MysqliException
       \
        UnknownType

```

BenMorel:
Again, I would stop at DriverException and make it a concrete class. There is no need for driver-specific subclasses.

Having an interface instead of a common, concrete class leads to [this kind of atrocities](https://github.com/doctrine/dbal/blob/1c680e191af590039976cf18cc0d7a8e63fa3695/tests/Doctrine/Tests/DBAL/Driver/AbstractDriverTest.php#L86-L108).

Subclassing to use-case-specific classes such as `DeadlockException` makes sense. Subclassing to driver-specific `MysqliException` does not.

Majkl578:
> There is no need for driver-specific subclasses.
> [...]
> Subclassing to use-case-specific classes such as DeadlockException makes sense. Subclassing to driver-specific MysqliException does not

Disagree here: different drivers lead to different problems, which means driver-specific exceptions and driver-specific handling (we have full tests of driver-specific skippers).

BenMorel:
Driver-specific exception handling is performed in [ExceptionConverterDriver::convertException()](https://github.com/doctrine/dbal/blob/2179996d6aa7bbbf6c83c98cab16b625db8ac17b/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php#L30-L107), which use `DriverException::getSQLState()` and `DriverException::getErrorCode()`, that are enough to discriminate exceptions. No existing driver need anything else.

Majkl578:
If you actually check my PR, you will find out that factories will be replaced by named exceptions. Additionally DriverException will be an interface, so anything that throws DriverException now is invalid and will be a specific exception instead.

Majkl578:
> Your PR is hard to read I'm afraid

It's just not rebased.

morozov:
> @morozov What should we do here, to get #3335 merged? You preferred to have exceptions handled before, but this conflicts with @Majkl578's PR.
> 
> Where should we head?

The easiest would be to drop the part with exceptions and merge the non-conflicting parts. If you want to do the redising of exceptions, we all need to get into an agreement of the approach and then whoever has time does the implementation.

BenMorel:
> The easiest would be to drop the part with exceptions and merge the non-conflicting parts.

There is absolutely no conflict in #3335 with exceptions: it's just using the `new class` where a concrete DriverException class is not available. Would you please explain what's wrong with merging it then?

morozov:
> There is absolutely no conflict in #3335 with exceptions: it's just using the `new class` where a concrete DriverException class is not available.

BenMorel:
@morozov #3335 does not make any changes to exceptions, on the contrary it tries to accommodate them as best as it can.

@Majkl578 Could you please review #3335, to see if it poses any problem to you?

morozov:
@BenMorel I made some progress recently cleaning up driver exceptions in https://github.com/doctrine/dbal/pull/4072, https://github.com/doctrine/dbal/pull/4085, https://github.com/doctrine/dbal/pull/4125, https://github.com/doctrine/dbal/pull/4130, and other PRs (see the history of the `3.0.x` branch).

Currently, some of the problems you were trying to solve are solved but not everything is changed the way you planned to. Do you want to try and rebase your PR onto `3.0.x` and see if any changes are still relevant?

BenMorel:
@morozov I’m on vacation right now, I’ll try to look into this as soon as I get back!

BenMorel:
Looking at `3.0.x` and `master`, exceptions look in good shape and the project now looks much closer to what I had in mind.

I'm therefore closing this PR, and would rather open a new one if I were to suggest other changes in the future.

Thank you.

morozov:
Sounds good. Thanks for following up, @BenMorel.

Issue #3366, opened 2018-11-29
OCI8Statement::execute() ignores binding errors


morozov:
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.8.0

`Statement::bind*()` methods return a boolean value to indicate whether the binding was successful:
https://github.com/doctrine/dbal/blob/df4343721d9981e58afd15eeaf18c885a2399766/lib/Doctrine/DBAL/Driver/Statement.php#L51-L53

This value is ignored in other DBAL components. For instance:
https://github.com/doctrine/dbal/blob/2ec1743e8876c20ba371e80fea17a0cd8e1ac919/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php#L391-L394

It leads to the issues like:
```php

use Doctrine\DBAL\DriverManager;

$conn = DriverManager::getConnection([
    'driver' => 'oci8',
    'host' => 'localhost',
    'user' => 'system',
    'password' => 'oracle',
    'dbname' => 'XE',
]);

$data = $conn->fetchAll('SELECT 1 FROM DUAL', [123]);

var_dump($data);

/*
Warning: oci_bind_by_name(): ORA-01036: illegal variable name/number in lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php on line 316

array(1) {
  [0] =>
  array(1) {
    [1] =>
    string(1) "1"
  }
}
*/
```
`Statement::execute()` should throw an exception in the case of a binding error instead of trying to execute an incorrectly constructed query.

Ocramius:
Handled in #3369

PR #3369, opened 2018-12-01
Handle binding errors in OCI8Statement::execute() and MySQLiStatement::execute()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #3366

1. `MysqliStatement`: regardless of whether the underlying statement has bound variables, try binding the parameters passed to the DBAL statement to get them validated by the driver.
2. `OCI8Statement`: do not ignore the return value of `oci_bind_by_name()`.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade to 3.0
 
+## BC BREAK `Statement::execute()` with redundant parameters.
+
+Similarly to the drivers based on `pdo_pgsql` and `pdo_sqlsrv`, `OCI8Statement::execute()` and `MySQLiStatement::execute()` do not longer ignore redundant parameters.
+
 ## BC BREAK: `Doctrine\DBAL\Types\Type::getDefaultLength()` removed
 
 The `Doctrine\DBAL\Types\Type::getDefaultLength()` method has been removed as it served no purpose.
--- a/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\Tests\DBAL\Functional;
 
+use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Driver\Statement;
 use Doctrine\DBAL\FetchMode;
 use Doctrine\DBAL\ParameterType;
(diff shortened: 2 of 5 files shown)


UPGRADE.md
```diff
@@ -1,5 +1,9 @@
 # Upgrade to 2.9
 
+## MINOR BC BREAK `Statement::execute()`.
+
+Similarly to the drivers based on `pdo_pgsql` and `pdo_sqlsrv`, `OCI8Statement::execute()` and `MySQLiStatement::execute()` do not longer ignore redundant parameters.
```
Ocramius: This is a major BC break, since apps built on these will really blow up spectacularly

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
```diff
@@ -292,4 +294,37 @@ public function testFetchInColumnMode() : void
...
+
+        switch ($driver) {
+            case 'pdo_mysql':
+            case 'pdo_oracle':
+            case 'pdo_sqlsrv':
+                self::markTestSkipped(sprintf(
+                    'PDOStatement::execute() implemented in the "%s" driver does not report redundant parameters',
+                    $driver
+                ));
+                break;
```
Ocramius: The `break` should rather be `return`

Ocramius (changes requested):

UPGRADE.md
morozov: I marked it as minor because it will only occur in the case of a logic error in application code. This was exactly my case. The same code already doesn't work with other drivers. Do you suggest to change the wording or move it to `develop`?

In a similar issue https://github.com/doctrine/dbal/pull/3359, @deeky666 [suggested](https://github.com/doctrine/dbal/pull/3359#discussion_r237781659) to implement exceptions and document it as a minor BC break.

I'm confused.

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
morozov: True.

UPGRADE.md
Ocramius: Yeah, while it is indeed an application error, consumers may have built code that just stuffs literally everything in query parameters, so we are actively breaking their stuff.

I think this should be moved to major 😣

morozov:
@Ocramius, retargeted.

Ocramius (approved):

Ocramius:
:ship:

PR #3372, opened 2018-12-01
Travis CI tests on MySQL 8.0

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

MySQL 8.0 has been out for some time, yet Doctrine DBAL is not tested against this version right now.

Following the rules edicted in #3347 and #3365, PHP `7.1`, <s>`7.3`</s>① and `nightly` are now only tested against the latest version, MySQL 8.0.

① **Edit:** MySQL 8.0 requires Xenial, and Travis CI does not support PHP 7.3 on Xenial yet. PHP 7.3 tests are therefore kept on MySQL 5.7 for the time being.

### DIFF
--- a/tests/travis/install-mysql-8.0.sh
+++ b/tests/travis/install-mysql-8.0.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -ex
+
+echo "Installing MySQL 8.0..."
+
+echo mysql-apt-config mysql-apt-config/select-server select mysql-8.0 | sudo debconf-set-selections
+wget https://dev.mysql.com/get/mysql-apt-config_0.8.10-1_all.deb
+sudo dpkg --install mysql-apt-config_0.8.10-1_all.deb
+sudo apt-get update -q
+sudo apt-get install -q -y --force-yes -o Dpkg::Options::=--force-confnew mysql-server
+echo -e "[mysqld]\ndefault_authentication_plugin=mysql_native_password" | sudo tee --append /etc/mysql/my.cnf
+sudo /etc/init.d/mysql start
+sudo mysql_upgrade
+
+mysql --version
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -21,7 +21,7 @@ before_commands:
 tools:
     external_code_coverage:
         timeout: 3600
-        runs: 28 # 23x Travis (jobs with COVERAGE=yes) + 3x AppVeyor (jobs with coverage=yes) + 2x ContinuousPHP
+        runs: 30 # 25x Travis (jobs with COVERAGE=yes) + 3x AppVeyor (jobs with coverage=yes) + 2x ContinuousPHP
 
 filter:
     excluded_paths:
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,16 +47,18 @@ jobs:
   include:
     - stage: Test
       php: 7.1
-      env: DB=mysql MYSQL_VERSION=5.7
+      env: DB=mysql MYSQL_VERSION=8.0
+      dist: xenial
       sudo: required
       before_script:
-        - bash ./tests/travis/install-mysql-5.7.sh
+        - bash ./tests/travis/install-mysql-8.0.sh
     - stage: Test
       php: 7.1
-      env: DB=mysqli MYSQL_VERSION=5.7
+      env: DB=mysqli MYSQL_VERSION=8.0
+      dist: xenial
       sudo: required
       before_script:
-        - bash ./tests/travis/install-mysql-5.7.sh
+        - bash ./tests/travis/install-mysql-8.0.sh
     - stage: Test
       php: 7.1
       env: DB=mariadb MARIADB_VERSION=10.3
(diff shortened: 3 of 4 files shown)


morozov:
Referencing https://github.com/doctrine/dbal/pull/3183.

tests/travis/install-mysql-8.0.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -ex
+
+echo "Installing MySQL 8.0..."
+
+echo mysql-apt-config mysql-apt-config/select-server select mysql-8.0 | sudo debconf-set-selections
+wget https://dev.mysql.com/get/mysql-apt-config_0.8.10-1_all.deb
```
Ocramius: Is there no package manager setup for this?

Ocramius (changes requested):

Ocramius:
Seems to fail to upgrade MySQL. What are your thoughts about running a tagged docker container instead?

BenMorel:
I'm debugging it right now, it should work soon.

BenMorel:
[It runs MySQL 8.0 alright](https://travis-ci.org/doctrine/dbal/jobs/462306731), there's just a `ConnectionException` not thrown because of a [MySQL error 2054](https://dev.mysql.com/doc/refman/8.0/en/client-error-reference.html) (CR_NOT_IMPLEMENTED), which is due to this error:

> The server requested authentication method unknown to the client

If I add code `2054` to `AbstractMySQLDriver::convertException()`, it works, but because this looks like a generic error code for any non-implemented feature, I don't think it's a good idea.

Instead I'm adding `default-authentication-plugin=mysql_native_password` to my.cnf to fix this error.

BenMorel:
All good now 👍 

`MysqlPlatform::getListTableIndexesSQL()` and `getListTableColumnsSQL()` are fixed for MySQL 8.0.

Ocramius:
Give it a squash, please

BenMorel:
Squashed!

Ocramius (dismissed):

morozov (changes requested):
@BenMorel please make sure Scrutinizer code coverage configuration reflects the changes.

tests/travis/install-mysql-8.0.sh
```diff
@@ -0,0 +1,16 @@
...
```
Majkl578: Please use Docker container instead of arbitrary 3rd-party debfile. As done in #3183.

Majkl578 (dismissed):

Majkl578:
How is this PR better than #3183? Why not finish that one instead of duplicating changes here?

tests/travis/install-mysql-8.0.sh
BenMorel: I'm not familiar with Docker, however for the record this isn't an arbitrary debfile, this is MySQL's official APT repository.

BenMorel:
@morozov Updated Scrutinizer config and rebased on latest `master`.

@Majkl578 Sorry, I didn't mean to interfer with your PR, I didn't find it when I opened this one. I did check it afterwards, and saw that it was left in a failing state with debug information. Because I don't have write access, I can't commit to your PR, so I chose to finish mine.

Feel free to merge either PR, I really don't mind!

tests/travis/install-mysql-8.0.sh
Majkl578: Given the nature of APT repositories, it's not a good idea to depend on a debfile as debfiles may randomly disappear when they get superseded by newer version.
Also Docker is easier to setup, configure environmen´t, maintain and in the end is more reliable.

BenMorel: Note that MySQL 5.7 [is already installed using MySQL's APT repo](https://github.com/doctrine/dbal/blob/master/tests/travis/install-mysql-5.7.sh). If you move to Docker, then this one should be updated as well for consistency.

Majkl578: Sure.

morozov (approved):

morozov:
@Majkl578 I think it will be fair to merge this pull request. It gets the job done, and we didn't have any problems with deb-packaged MySQL in the past. We can replace deb packages with Docker images later if needed.

morozov:
> I'm not familiar with Docker […]

@BenMorel you can try using the bits from #3183. Let us know if you need help. Maybe @Majkl578 or I could help you finish it.

Alternatively, if we rebase #3183 and tests no longer fail there, we can merge it instead. @Majkl578 let me know if you need help rebasing it.

No pressure, but this is the last open ticket for [2.9.0](https://github.com/doctrine/dbal/milestone/68) :smile:

morozov:
> If you're in a hurry for 2.9.0, then maybe the first option could be a good compromise?

Right now it's not that much of a blocker, so I'd like to get input from @Majkl578 first. Both options look legit to me.

Ocramius:
I'd say let's merge here and then we rebase the other patch, mostly due to this being ready to be shipped.

morozov:
Thank you @BenMorel. Time to put together some release notes.

Majkl578:
> mostly due to this being ready to be shipped

I disagree, Xenial is not acceptable - it's unstable, undocumented and doesn't support PHP 7.3.

Ocramius:
@Majkl578 let's fix it up, then - for now we reached the goal of getting 8.0 tested

PR #3375, opened 2018-12-03
[SQL Anywhere] Fix bound parameter references in PHP 7

deeky666:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | n/a

#### Summary

The SQL Anywhere driver suffers the same problem that `oci8` did with PHP 7 (See: ae314b137211be8c167250254b06ef388b36c0cc)

Binding multiple parameters in a prepared statement does not work as PHP does not manage the references itself. We have to manage the references to bound parameters ourself, otherwise only the first bound parameter is used and overwritten subsequently.

The tests are revealing the issue already.


Ocramius:
@deeky666 any way to test this, or do integration tests already fail?

deeky666:
@Ocramius integration tests already fail big time. The other failures are unrelated and need to be fixed separately.

Before: https://pastebin.com/CS0dGa3i
After: https://pastebin.com/yMdUXW3Q

Ocramius (approved):
:+1:

PR #3378, opened 2018-12-04
[SQL Anywhere] Fix query limit values "0" and "null"

deeky666:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | n/a

#### Summary

Integration tests on SQL Anywhere revealed that the platform doesn't handle `$limit = null` when rewriting `SELECT` statements for LIMIT/OFFSET.
Also handling of `$limit = 0` was wrong as it was selecting ALL results instead of NONE.

Reference: http://dcx.sybase.com/1200/en/dbreference/select-statement.html

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
@@ -1308,25 +1308,20 @@ SQL
      */
     protected function doModifyLimitQuery($query, $limit, $offset)
     {
-        $limitOffsetClause = '';
+        $limitOffsetClause = $this->getTopClauseSQL($limit, $offset);
 
-        if ($limit > 0) {
-            $limitOffsetClause = 'TOP ' . $limit . ' ';
-        }
+        return $limitOffsetClause === ''
+            ? $query
+            : preg_replace('/^\s*(SELECT\s+(DISTINCT\s+)?)/i', '\1' . $limitOffsetClause . ' ', $query);
+    }
 
+    private function getTopClauseSQL(?int $limit, ?int $offset) : string
+    {
         if ($offset > 0) {
-            if ($limit === 0) {
-                $limitOffsetClause = 'TOP ALL ';
-            }
-
-            $limitOffsetClause .= 'START AT ' . ($offset + 1) . ' ';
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
```diff
@@ -1310,16 +1310,14 @@ protected function doModifyLimitQuery($query, $limit, $offset)
...
-        if ($limit > 0) {
+        if ($limit !== null) {
             $limitOffsetClause = 'TOP ' . $limit . ' ';
-        }
 
-        if ($offset > 0) {
-            if ($limit === 0) {
-                $limitOffsetClause = 'TOP ALL ';
+            if ($offset > 0) {
+                $limitOffsetClause .= 'START AT ' . ($offset + 1) . ' ';
```
Ocramius: Can you move the code that generates `$limitOffsetClause` to a `private` method? All of these if/else become early returns that way.

Ocramius (changes requested):

deeky666:
@Ocramius done

lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
```diff
@@ -1308,25 +1308,24 @@ protected function _getTransactionIsolationLevelSQL($level)
...
+        return $limitOffsetClause === ''
+            ? $query
+            : preg_replace('/^\s*(SELECT\s+(DISTINCT\s+)?)/i', '\1' . $limitOffsetClause . ' ', $query);
+    }
 
+    /**
+     * @param int|null $limit
+     * @param int|null $offset
+     */
+    private function getTopClauseSQL($limit, $offset) : string
```
Ocramius: These can be `?int` now

Ocramius (changes requested):

lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
deeky666: @Ocramius fixed

Ocramius (approved):
LGTM :+1: 

@morozov assigning to you to do the backporting dance, if you have time

Issue #3377, opened 2018-12-04
Schema comparison behavior changed for boolean default values


Majkl578:
```
@ORM\Column(type="boolean", nullable=false, options={"default" = false})
```
it always generates migration becouse of #3355. compared with old version https://3v4l.org/HBLG7 :
````php
<?php
$changedProperties = [];
$properties1['default'] = false;
$properties2['default'] = 0;

if ($properties1['default'] != $properties2['default'] ||
    (null === $properties1['default'] && null !== $properties2['default']) ||
    (null === $properties2['default'] && null !== $properties1['default'])
) {
    $changedProperties[] = 'default';
}

var_dump($changedProperties);

if (($properties1['default'] === null) !== ($properties2['default'] === null)
    || (string) $properties1['default'] !== (string) $properties2['default']) {
    $changedProperties[] = 'default';
}

var_dump($changedProperties);
````
result:
```
array(0) {
}
array(1) {
  [0]=>
  string(7) "default"
}
```

_Originally posted by @svycka in https://github.com/doctrine/dbal/pull/3355_

Ocramius:
Handled in #3382

PR #3382, opened 2018-12-04
Reverted strict comparison back to loose because of a new regression

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3377

Reverted strict comparison back to loose.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
@@ -128,7 +128,7 @@ class OracleSchemaManager extends AbstractSchemaManager
             }
         }
 
-        $unsigned = $fixed = null;
+        $unsigned = $fixed = $precision = $scale = $length = null;
 
         if (! isset($tableColumn['column_name'])) {
             $tableColumn['column_name'] = '';
@@ -146,8 +146,13 @@ class OracleSchemaManager extends AbstractSchemaManager
             $tableColumn['data_default'] = trim($tableColumn['data_default'], "'");
         }
 
-        $precision = null;
-        $scale     = null;
+        if ($tableColumn['data_precision'] !== null) {
+            $precision = (int) $tableColumn['data_precision'];
+        }
+
+        if ($tableColumn['data_scale'] !== null) {
+            $scale = (int) $tableColumn['data_scale'];
+        }
 
         $type                    = $this->_platform->getDoctrineTypeMapping($dbType);
         $type                    = $this->extractDoctrineTypeFromComment($tableColumn['comments'], $type);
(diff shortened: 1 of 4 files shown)


Ocramius (dismissed):

tests/Doctrine/Tests/DBAL/Functional/Schema/ComparatorTest.php
```diff
@@ -25,15 +25,31 @@ protected function setUp()
         $this->comparator    = new Comparator();
     }
 
-    public function testDefaultValueComparison()
+    /**
+     * @param mixed $value
+     *
+     * @dataProvider defaultValueProvider
+     */
+    public function testDefaultValueComparison(string $type, $value)
```
lcobucci: ` : void` could be added

tests/Doctrine/Tests/DBAL/Functional/Schema/ComparatorTest.php
```diff
@@ -25,15 +25,31 @@ protected function setUp()
         $this->comparator    = new Comparator();
     }
 
-    public function testDefaultValueComparison()
+    /**
+     * @param mixed $value
```
lcobucci: Does it make sense to be a bit more explicit here (as in `int|bool`)?

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
```diff
@@ -146,32 +146,30 @@ protected function _getPortableTableColumnDefinition($tableColumn)
...
-                    $precision = 1;
-                    $scale     = 0;
-                    $type      = 'boolean';
+                if ($precision === 20 && $scale === 0) {
+                    $type = 'bigint';
+                } elseif ($precision === 5 && $scale === 0) {
+                    $type = 'smallint';
+                } elseif ($precision === 1 && $scale === 0) {
+                    $type = 'boolean';
                 } elseif ($tableColumn['data_scale'] > 0) {
```
lcobucci: Nice clean up :+1: 

This could be `$scale > 0` (not that it would make a big difference :stuck_out_tongue:)

lcobucci (dismissed):
LGTM just some thoughts

tests/Doctrine/Tests/DBAL/Functional/Schema/ComparatorTest.php
morozov: It could be used with other types later, so I'd leave it as is.

tests/Doctrine/Tests/DBAL/Functional/Schema/ComparatorTest.php
morozov: @Majkl578 we may have already discussed it, but is it possible for the coding standard to require/add `: void` if a function doesn't `return` or `yield` anything?

Majkl578: It does (SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingReturnTypeHint), but it's disabled in master.

morozov: It should be safe to enable it for tests, right? Maybe except for mocks. I hate those mocks.

lcobucci: > It should be safe to enable it for tests, right?

It should.

lcobucci (approved):

lcobucci:
@morozov :ship:

PR #3390, opened 2018-12-06
Add optional detection and logging of query source in debug stack

carnage:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | N/A

#### Adds optional detection and logging of the query source in the debug stack logger

Adds a new flag, which when enabled, adds an additional field to the logged data "querySource" this is populated by a best attempt to detect where (file + line number) in your code base the query originated from.

This uses debug_backtrace and is a performance hit if enabled, so it is disabled by default.

### DIFF
--- a/lib/Doctrine/DBAL/Logging/DebugStack.php
+++ b/lib/Doctrine/DBAL/Logging/DebugStack.php
@@ -9,6 +9,9 @@ use function microtime;
  */
 class DebugStack implements SQLLogger
 {
+    const SOURCE_DISABLED = 0;
+    const SOURCE_ENABLED = 1;
+
     /**
      * Executed SQL queries.
      *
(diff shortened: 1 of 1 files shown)


Ocramius:
@carnage can you isolate this design into a completely separate logger? Mixing it with the existing one is not a good idea, as the unit now does multiple things.

carnage:
Was thinking about this over the weekend, but I'm not sure I agree. The issue came up as I was trying to decide what the new class should be named, a lot of the options didn't really define what the class was intended for (eg QuerySourceLogger misses out the fact it's intended for debugging, VerboseDebugLogger was somewhat imprecise) 

The unit still only does a single thing - logs information for debugging purposes; the additional feature is to add more information to it. The /only/ reason I didn't enable this by default is that debug_backtrace is an expensive call. 

This does highlight that my implementation is probably wrong; using a flag for the function is deceptive as it implies that there are two features here; instead there should just be a verbosity level - this means if anything gets added in future an additional verbosity level can be included. A verbosity level would also better match with user expectations - more output at a potential extra performance cost by increasing the level.

It would also open up the possibility of lower levels of verbosity - excluding parameter information from the log, this would allow applications to log queries without risking logging sensitive information.

Let me know your thoughts and I'll amend the PR; needs a test regardless of the route we go.

PR #3392, opened 2018-12-06
fixed annotations of parameter $columns in Table

### LINKED ISSUE #3380: 2.9 BC Break on Doctrine\DBAL\Schema\Table method signatures
|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.9.0

#### Summary

Sorry if this is intended, but I couldn't find anywhere mentioning these changes on these `Doctrine\DBAL\Schema\Table` method signatures :
- setPrimaryKey(int, string) -> setPrimaryKey(array)
- addUniqueIndex(int, string) -> addUniqueIndex(array)
- addIndex(int, string) -> addIndex(array)

I can drop a PR either reverting the method signature's back or update the docs if needed.

LukasVitek:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

Fixes #3380.

#### Summary

Fixed annotations of parameter `$columns` in Table. Parameter `$columns` is array of mixed, nowhere in the code does not treated with him as with an array of array. 

Before the fix the code, phpStan reported an error with this code:
```php
$table = new Table('tableName');
$table->addUniqueIndex(['col1', 'col2']);
```
`Parameter #1 $columnNames of method Doctrine\DBAL\Schema\Table::addUniqueIndex() expects array<array>, array<int, string> given. `

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Table.php
+++ b/lib/Doctrine/DBAL/Schema/Table.php
@@ -9,8 +9,6 @@ use const ARRAY_FILTER_USE_KEY;
 use function array_filter;
 use function array_merge;
 use function in_array;
-use function is_numeric;
-use function is_string;
 use function preg_match;
 use function strlen;
 use function strtolower;
@@ -100,16 +98,16 @@ class Table extends AbstractAsset
     /**
      * Sets the Primary Key.
      *
-     * @param mixed[][]   $columns
+     * @param string[]    $columnNames
      * @param string|bool $indexName
      *
      * @return self
      */
-    public function setPrimaryKey(array $columns, $indexName = false)
+    public function setPrimaryKey(array $columnNames, $indexName = false)
     {
-        $this->_addIndex($this->_createIndex($columns, $indexName ?: 'primary', true, true));
+        $this->_addIndex($this->_createIndex($columnNames, $indexName ?: 'primary', true, true));
 
-        foreach ($columns as $columnName) {
+        foreach ($columnNames as $columnName) {
             $column = $this->getColumn($columnName);
             $column->setNotnull(true);
         }
@@ -118,7 +116,7 @@ class Table extends AbstractAsset
     }
 
     /**
-     * @param mixed[][]   $columnNames
+     * @param string[]    $columnNames
      * @param string|null $indexName
      * @param string[]    $flags
      * @param mixed[]     $options
@@ -168,7 +166,7 @@ class Table extends AbstractAsset
     }
 
     /**
-     * @param mixed[][]   $columnNames
+     * @param string[]    $columnNames
      * @param string|null $indexName
      * @param mixed[]     $options
      *
@@ -236,15 +234,15 @@ class Table extends AbstractAsset
     /**
      * Checks if an index begins in the order of the given columns.
      *
-     * @param mixed[][] $columnsNames
+     * @param string[] $columnNames
      *
      * @return bool
      */
-    public function columnsAreIndexed(array $columnsNames)
+    public function columnsAreIndexed(array $columnNames)
     {
         foreach ($this->getIndexes() as $index) {
             /** @var $index Index */
-            if ($index->spansColumns($columnsNames)) {
+            if ($index->spansColumns($columnNames)) {
                 return true;
             }
         }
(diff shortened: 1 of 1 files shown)


morozov:
@LukasVitek is this related to https://github.com/doctrine/dbal/issues/3380? Could you double check if those `mixed[][] $columns/$columnNames` are actually `string[] $columnNames`, not `mixed[]`? See the details in the related ticket.

LukasVitek:
Actually, yes, issue https://github.com/doctrine/dbal/issues/3380 is very related. 
And yes, it is always assumed that `$columns/$columnNames` are type of `string[]`, 
except for one case here: https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/Table.php#L274 where it is verified if the variable in the array is type of `string`.
Should I therefore type of annotations to remake to `string[]`?

lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -253,12 +253,12 @@ public function columnsAreIndexed(array $columnsNames)
     }
 
     /**
-     * @param mixed[][] $columnNames
-     * @param string    $indexName
-     * @param bool      $isUnique
-     * @param bool      $isPrimary
-     * @param string[]  $flags
-     * @param mixed[]   $options
+     * @param string[] $columnNames
```
morozov: > […] except for one case here: https://github.com/doctrine/dbal/blob/4b4d44f90ae6efde2a494b12167348bbc0551e09/lib/Doctrine/DBAL/Schema/Table.php#L274
where it is verified if the variable in the array is type of `string`.

Here, it should be something like `string[]|mixed[][]`. Judging by the code, it allows the following syntax:
```php
_createIndex([
    'column_1',
    'column_2' => [
        'option' => 'value'
    ],
]);
```

koftikes (dismissed):

lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -253,12 +253,12 @@ public function columnsAreIndexed(array $columnsNames)
     }
 
     /**
-     * @param mixed[][] $columnNames
-     * @param string    $indexName
-     * @param bool      $isUnique
-     * @param bool      $isPrimary
-     * @param string[]  $flags
-     * @param mixed[]   $options
+     * @param string[]|mixed[][] $columnNames
```
Ocramius: This is probably not `mixed[][]`. To inspect that, I'd suggest digging into the `Index` constructor.

Ocramius (dismissed):

lib/Doctrine/DBAL/Schema/Table.php
morozov: If it's not `mixed[][]`, then the code handling `$indexColOptions` should be removed. Given that this is a private API, I wanted to make the annotations consistent with the code and then clean up the code in a minor release.

morozov: Indeed, it's only `string[]` Judging by the four calls from the same class. Please revert back to `string[] $columnNames` and rework the `foreach` loop to:
```php
foreach ($columnNames as $columnName) {
}
```

morozov:
@LukasVitek please squash and I’ll merge it.

thewilkybarkid:
`columnsAreIndexed()` needs updating too.

morozov:
@LukasVitek please update `columnsAreIndexed()` as well.

morozov (approved):

morozov:
Thank you @LukasVitek and @thewilkybarkid.

leofeyer:
I'm afraid this change breaks indexes with a fixed length. We have an index named `path(768)`, which now throws a `There is no column with name 'path(768)'` exception, because the following code has been removed:

```php
    foreach ($columnNames as $columnName => $indexColOptions) {
        if (is_numeric($columnName) && is_string($indexColOptions)) {
            $columnName = $indexColOptions;
        }
```

Ocramius:
@leofeyer can you please report a new issue? Got a test case for it?

leofeyer:
@Ocramius Sure, here you go: #3409

Issue #3387, opened 2018-12-05
Generates invalid up() migration


svycka:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 1.8.1

#### Summary

generates incorrect up migration but down is ok

#### Current behavior
generated migration just throw exception becous it should first remove constrant and then remove index
````php
<?php declare(strict_types=1);

namespace DatabaseMigrations;

use Doctrine\DBAL\Schema\Schema;
use Doctrine\Migrations\AbstractMigration;

/**
 * Auto-generated Migration: Please modify to your needs!
 */
final class Version20181205094238 extends AbstractMigration
{
    public function up(Schema $schema) : void
    {
        // this up() migration is auto-generated, please modify it to your needs
        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');

        $this->addSql('DROP INDEX idx_7ee5e3886dc044c5 ON fields');
        $this->addSql('CREATE INDEX IDX_7EE5E388D3CB4A96 ON fields (`group`)');
    }

    public function down(Schema $schema) : void
    {
        // this down() migration is auto-generated, please modify it to your needs
        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');

        $this->addSql('ALTER TABLE fields DROP FOREIGN KEY FK_7EE5E388D3CB4A96');
        $this->addSql('DROP INDEX idx_7ee5e388d3cb4a96 ON fields');
        $this->addSql('CREATE INDEX IDX_7EE5E3886DC044C5 ON fields (`group`)');
        $this->addSql('ALTER TABLE fields ADD CONSTRAINT FK_7EE5E388D3CB4A96 FOREIGN KEY (`group`) REFERENCES field_groups (`id`)');
    }
}
````

#### How to reproduce
````
     * @ORM\ManyToOne(targetEntity="FarmField\Entity\FieldGroup")
     * @ORM\JoinColumn(name="group", referencedColumnName="id")
changed to:
     * @ORM\ManyToOne(targetEntity="FarmField\Entity\FieldGroup")
     * @ORM\JoinColumn(name="`group`", referencedColumnName="id")
````
#### Expected behavior

generated valid migration maybe like this:
````php
    public function up(Schema $schema) : void
    {
[...]

PR #3395, opened 2018-12-07
fixes incorrect database migration for quoted columns

svycka:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3387

#### Summary

I think this should accept as the same column if one is quoted and other is not


lib/Doctrine/DBAL/Schema/ForeignKeyConstraint.php
```diff
@@ -364,9 +364,9 @@ private function onEvent($event)
      */
     public function intersectsIndexColumns(Index $index)
     {
-        foreach ($index->getColumns() as $indexColumn) {
+        foreach ($index->getUnquotedColumns() as $indexColumn) {
             foreach ($this->_localColumnNames as $localColumn) {
-                if (strtolower($indexColumn) === strtolower($localColumn->getName())) {
+                if (strtolower($indexColumn) === strtolower($localColumn->trimQuotes($localColumn->getName()))) {
```
svycka: I am not sure that I have to quote `$localColumn->getName()` this should always be unquoted?

tests/Doctrine/Tests/DBAL/Schema/ForeignKeyConstraintTest.php
```diff
@@ -16,14 +16,9 @@ class ForeignKeyConstraintTest extends TestCase
      */
     public function testIntersectsIndexColumns(array $indexColumns, $expectedResult)
     {
-        $foreignKey = new ForeignKeyConstraint(['foo', 'bar'], 'foreign_table', ['fk_foo', 'fk_bar']);
+        $foreignKey = new ForeignKeyConstraint(['foo', '`bar`'], 'foreign_table', ['fk_foo', 'fk_bar']);
```
Ocramius: Please design a new test: don't modify existing ones.

If an existing test is being modified out of necessity, we will likely hit a BC break scenario.

lib/Doctrine/DBAL/Schema/ForeignKeyConstraint.php
Ocramius: Can't really help with that due to my unfamiliarity with this code area. @deeky666 @morozov can you check this?

Ocramius (changes requested):

lib/Doctrine/DBAL/Schema/ForeignKeyConstraint.php
morozov: I'm unfamiliar with this either but it feels bad. An unquoted column name is the value itself, a quoted column name is a SQL expression representing this value. Why would we want to mix them and compare by trimming quotes?

svycka: @morozov maybe I did not understand you. I need them to ignore escaping because of #3387 while debugging I noticed that for up and down migrations it compares escaped or unescaped values and generates or not gerenates migration. This place checks if the column name is included in the foreign constraint and it should not matter if it is escaped or not should say it is in both cases. I just not sure about this `$localColumn->trimQuotes($localColumn->getName())`.

Also I am not sure this will not break other parts but tests looks green so maybe ok?

svycka:
@deeky666 would be nice if you  could review as this is your code :) implemented  with https://github.com/doctrine/dbal/pull/756

lib/Doctrine/DBAL/Schema/ForeignKeyConstraint.php
morozov: I don't think it's a valid case for DBAL. Right now there's no difference between quoted and unquoted identifiers in the DBAL type system but it doesn't mean it should handle them transparently. Most likely, there's a lot of other cases where mixing quoted and unquoted identifiers will cause unexpected behavior.

Instead of changing the comparison logic in DBAL, I'd try to fix the inconsistency on the calling side.

morozov (changes requested):

lib/Doctrine/DBAL/Schema/ForeignKeyConstraint.php
SenseException: Isn't this a valid case where a database allows reserved words when they are quoted?

svycka:
@Ocramius I have created a new test, is it ok now?

tests/Doctrine/Tests/DBAL/Schema/ForeignKeyConstraintTest.php
```diff
@@ -8,6 +8,20 @@
 
 class ForeignKeyConstraintTest extends TestCase
 {
+    /**
+     * @param string[] $indexColumns
+     *
+     * @dataProvider getIntersectsIndexColumnsData
+     */
+    public function testIntersectsIndexEscapedColumns(array $indexColumns, $expectedResult)
```
Ocramius: What's `$expectedResult`? Needs type documentation here

Ocramius (changes requested):

tests/Doctrine/Tests/DBAL/Schema/ForeignKeyConstraintTest.php
```diff
@@ -18,12 +32,7 @@ public function testIntersectsIndexColumns(array $indexColumns, $expectedResult)
     {
         $foreignKey = new ForeignKeyConstraint(['foo', 'bar'], 'foreign_table', ['fk_foo', 'fk_bar']);
 
-        $index = $this->getMockBuilder(Index::class)
-            ->disableOriginalConstructor()
-            ->getMock();
-        $index->expects($this->once())
-            ->method('getColumns')
-            ->will($this->returnValue($indexColumns));
+        $index = new Index('INDEX_NAME', $indexColumns);
```
Ocramius: I'm vary of test changes here: what was broken here?

svycka: Nothing, just thought maybe better to use an actual object instead of mock no?

tests/Doctrine/Tests/DBAL/Schema/ForeignKeyConstraintTest.php
svycka: I guess I could add bool type hint here. Is that would be ok? or I have to update doc block?

Ocramius: Adding a type declaration would be best (instead of docblock)

tests/Doctrine/Tests/DBAL/Schema/ForeignKeyConstraintTest.php
Ocramius: If the previous code works without changes, let's avoid changing it at all. Let's put test refactoring in separate patches, to avoid hiding potential BC breaks

svycka: @Ocramius as you can see tests fail and now I remember why. Because I use `getUnquotedColumns` instead of `getColumns` and I added some columns with quotes into data provider. And if I mock here can't test quoted names. I mean this:
```
$index->expects($this->once())
            ->method('getUnquotedColumns')
            ->will($this->returnValue($indexColumns));
``` 
should always return unquoted names and I can't add quoted names in data provider but maybe I don't have to?
 I still think actual implementation would be better here and anyway I have to change mocked method anyway.

lib/Doctrine/DBAL/Schema/ForeignKeyConstraint.php
```diff
@@ -366,7 +366,7 @@ public function intersectsIndexColumns(Index $index)
     {
         foreach ($index->getUnquotedColumns() as $indexColumn) {
             foreach ($this->_localColumnNames as $localColumn) {
-                if (strtolower($indexColumn) === strtolower($localColumn->trimQuotes($localColumn->getName()))) {
```
svycka: it seems this was not required `testIntersectsIndexEscapedColumns` should cover this

tests/Doctrine/Tests/DBAL/Schema/ForeignKeyConstraintTest.php
svycka: @Ocramius if you want I can still leave this mock but will have to change mocked method and will also have to duplicate this test and test data to actually test quoted names.

morozov:
Given that this is a cross-platform change, we need a new functional test that reproduces the problem. Just a unit test that covers some implementation details doesn't tell anyone anything.

svycka:
> Given that this is a cross-platform change, we need a new functional test that reproduces the problem. Just a unit test that covers some implementation details doesn't tell anyone anything.

maybe you are right. But I think this is just a bug I don't know database with could say they are different columns just because it is quoted for example `column1` and ``` `column1` ``` do you know databases with allows to have those booths? So I think if we check if it exists we should check both at least in this case when we actually mean to check if the column exists.

And about a test, I am not sure how to test this it does not treat it as the same column and you say it is not a bug as I understand. You can see an example here #3387 but I don't know how to write a test for this. Maybe someone could help or write a test for this. Or at least some kind of similar example I could write a test if this is a bug.

morozov:
> > Given that this is a cross-platform change, we need a new functional test that reproduces the problem. Just a unit test that covers some implementation details doesn't tell anyone anything.
> 
> maybe you are right.

I usually am.

> But I think this is just a bug I don't know database with could say they are different columns just because it is quoted for example `column1` and `` `column1` `` do you know databases with allows to have those booths?

No.

> And about a test, I am not sure how to test this it does not treat it as the same column and you say it is not a bug as I understand.

I’m not saying it's not a bug. I’m saying that the unit test doesn't represent the bug.

> You can see an example here #3387 but I don't know how to write a test for this. Maybe someone could help or write a test for this. Or at least some kind of similar example I could write a test if this is a bug.

As for the high-level “what the expected test should look like”, please look at [`DateExpressionTest.php`](https://github.com/doctrine/dbal/blob/4c22e80517d5458ed4025a35bdcbcd0c27447d55/tests/Doctrine/Tests/DBAL/Functional/Platform/DateExpressionTest.php) and [`DefaultExpressionTest.php`](https://github.com/doctrine/dbal/blob/4c22e80517d5458ed4025a35bdcbcd0c27447d55/tests/Doctrine/Tests/DBAL/Functional/Platform/DefaultExpressionTest.php). The test should deploy the initial schema to a real DB, make the needed changes to it and assert that the changes have been applied correctly.

Issue #3389, opened 2018-12-06
Column option "autoincrement" => "auto" causes schema diffs


sandermarechal:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.9.0

#### Summary

Schema diff generates a diff for columns that use the (undocumented) value of `"auto"` for the "autoincrement" option. This would normally be trivial, but the rather popular symfony/security-acl package defines all it's tables with an `"id"` column with the option `"autoincrement" => "auto"`.

#### Previous behaviour

A schema diff (update) when using Symfony ACL only generates expected updates.

#### Current behavior

A schema diff (update) when using Symfony ACL also generates updates for the `"id"` column of all ACL tables.

#### How to reproduce

Setup a standard Symfony, add the security ACL, initialize the ACL tables, run "bin/console doctrine:schema:update --dump-sql".

### Cause

I have identified commit dca1465bdc0836c71ed97e46820490127ccc89ab as the cause, specifically in `Doctrine\DBAL\Schema\Comparator` [line 415](https://github.com/doctrine/dbal/commit/dca1465bdc0836c71ed97e46820490127ccc89ab#diff-6a252b2e905db4e03fc6a121a41e6466L415) which changes the column property comparison from != to ===. In the old version "auto" and true would compare equal. In the new version they do not compare equal, causing the schemas to differ.

sandermarechal:
See also symfony/security-acl#43

Ocramius:
Would it be possible to isolate this into a test case? Patch seems easy to
do, once we prevent regressions via a test.

sandermarechal:
That should be easy. Just use a Schema with a column that has `"autoincrement" => "auto"` as an option as the "from" schema, and one with `"autoincrement" => true` as an option in the "to" schema. Diff 'em and the diff should be empty.

sandermarechal:
I have added a failing test case. I am not sure what the best solution would be though. Do a sloppy comparison? Or just add a special-case for the string 'auto'?

morozov:
@sandermarechal looks like most schema managers cast the value of `autoincrement` to a boolean but the one for MySQL doesn't. If so, the manager should be fixed, not the comparator.

sandermarechal:
Symfony ACL is fixing it on their end, but that doesn't account for already released versions. I would propose adding a special case for the string "auto" only, and documenting that it is for BC with symfony/security-acl.

Ocramius:
[...]

PR #3399, opened 2018-12-10
Add failing test case for DBAL #3389

sandermarechal:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

Add a failing test case for #3389

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
+++ b/tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
@@ -1302,4 +1302,32 @@ class ComparatorTest extends TestCase
         self::assertCount(1, $actual->changedTables['table2']->addedForeignKeys, 'FK to table3 should be added.');
         self::assertEquals('table3', $actual->changedTables['table2']->addedForeignKeys[0]->getForeignTableName());
     }
+
+    public function testAutoincrementColumnWithStringConfiguration()
+    {
+        $schema1 = new Schema([
+            'bugdb' => new Table(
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
```diff
@@ -1302,4 +1302,32 @@ public function testForeignKeyRemovalWithRenamedLocalColumn()
...
+                    ]),
+                ]
+            ),
+        ]);
+        $schema2 = new Schema([
+            'bugdb' => new Table(
+                'bugdb',
+                [
+                    'integerfield1' => new Column('integerfield1', Type::getType('integer'), [
+                        'autoincrement' => 'auto',
```
morozov: It should be an integration test. The comparator itself shouldn't account for all potential inconsistencies in platform and schema manager implementations. Does `"auto"` have any different meaning than `true` for MySQL? If not, it should be solved in the schema manager level.

morozov (changes requested):

Ocramius:
Closing as per decision in https://github.com/doctrine/dbal/issues/3389#issuecomment-446226991

@sandermarechal thanks for reporting and speccing this further out: I think we shouldn't consider `'auto'` a special case, and therefore this becomes `invalid` and to be fixed in the downstream consumer instead.

Another alternative would be to reject `'autoincrement' => 'auto'` by throwing a hard exception, forcing downstream consumers to change, but that would probably just worsen the situation even further.

Issue #3410, opened 2018-12-14
MySqlSchemaManager::parseCreateOptions() must be of the type string, null given


jarnovanleeuwen:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.9.1

#### Summary

After upgrading from 2.9.0 to 2.9.1 we started seeing the following error when migrating databases:

```
Argument 1 passed to Doctrine\DBAL\Schema\MySqlSchemaManager::parseCreateOptions() must be of the type string, null given
```

#### Previous behaviour

In 2.9.0, there was a check if `CREATE_OPTIONS` was empty. See the difference: https://github.com/doctrine/dbal/commit/b18f63ccbfb3bd99a59fb67a900991f0c478c2f3#diff-f8e1d5c54371540dc546d17c426c36e6L309

#### Current behavior

Now, `parseCreateOptions` is called even when `$tableOptions['CREATE_OPTIONS']` is null. As the type hint shows, a `string` needs to be passed.

#### How to reproduce

Not sure, but I think the code clearly shows that we expect a null value here while the typehint disallows null values.

jarnovanleeuwen:
Linking https://github.com/doctrine/dbal/pull/3393

morozov:
Indeed, `CREATE_OPTIONS` in `INFORMATION_SCHEMA.TABLES` is nullable:
```sql
SHOW CREATE TABLE INFORMATION_SCHEMA.TABLES;

CREATE TEMPORARY TABLE `TABLES` (
  `TABLE_CATALOG` varchar(512) NOT NULL DEFAULT '',
  `TABLE_SCHEMA` varchar(64) NOT NULL DEFAULT '',
  `TABLE_NAME` varchar(64) NOT NULL DEFAULT '',
  `TABLE_TYPE` varchar(64) NOT NULL DEFAULT '',
  `ENGINE` varchar(64) DEFAULT NULL,
  `VERSION` bigint(21) unsigned DEFAULT NULL,
  `ROW_FORMAT` varchar(10) DEFAULT NULL,
  `TABLE_ROWS` bigint(21) unsigned DEFAULT NULL,
  `AVG_ROW_LENGTH` bigint(21) unsigned DEFAULT NULL,
  `DATA_LENGTH` bigint(21) unsigned DEFAULT NULL,
  `MAX_DATA_LENGTH` bigint(21) unsigned DEFAULT NULL,
[...]

PR #3412, opened 2018-12-17
`CREATE_OPTIONS` is nullable in MySQL

morozov:
I haven't found any documentation on when exactly empty `CREATE_OPTIONS` is `NULL` but it looks like it's true for all views and only views.

Fixes https://github.com/doctrine/dbal/issues/3410.


Ocramius (approved):
LGTM :+1:

Issue #3414, opened 2018-12-18
Schema migration doesn't detect index length change


morozov:
Originally reported by @leofeyer in https://github.com/doctrine/dbal/issues/3409#issuecomment-448134762.

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.9.1

```php
public function testUpdatesIndexLength()
{
    $this->connection->exec('CREATE TABLE index_length (path text NULL)');
    $this->connection->exec('CREATE INDEX path ON index_length (path(333))');

    $table = new Table('index_length');
    $table->addColumn('path', 'text', ['notnull' => false]);
    $table->addIndex(['path'], 'path', [], ['lengths' => [768]]);

    $fromSchema = $this->connection->getSchemaManager()->createSchema();
    $toSchema = new Schema([$table]);

    $diff = $fromSchema->getMigrateToSql($toSchema, $this->connection->getDatabasePlatform());

    self::assertContains('ALTER TABLE index_length DROP INDEX path, ADD INDEX path (path(768))', $diff);
}
```

yvoyer:
This fix seems to introduce another regression, see #3458

PR #3415, opened 2018-12-19
Consider column lengths when comparing indices

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug

When comparing indices, compare column lengths as well. Fixes https://github.com/doctrine/dbal/issues/3414.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Index.php
+++ b/lib/Doctrine/DBAL/Schema/Index.php
@@ -4,6 +4,7 @@ namespace Doctrine\DBAL\Schema;
 
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use InvalidArgumentException;
+use function array_filter;
 use function array_keys;
 use function array_map;
 use function array_search;
@@ -211,6 +212,10 @@ class Index extends AbstractAsset implements Constraint
                 return false;
             }
 
+            if (! $this->hasSameColumnLengths($other)) {
+                return false;
+            }
+
             if (! $this->isUnique() && ! $this->isPrimary()) {
                 // this is a special case: If the current key is neither primary or unique, any unique or
                 // primary key will always have the same effect for the index and there cannot be any constraint
@@ -335,5 +340,18 @@ class Index extends AbstractAsset implements Constraint
         }
 
         return ! $this->hasOption('where') && ! $other->hasOption('where');
+    }
+
+    /**
+     * Returns whether the index has the same column lengths as the other
+     */
+    private function hasSameColumnLengths(self $other) : bool
+    {
+        $filter = static function (?int $length) : bool {
+            return $length !== null;
+        };
+
+        return array_filter($this->options['lengths'] ?? [], $filter)
+            === array_filter($other->options['lengths'] ?? [], $filter);
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
+++ b/tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
@@ -109,6 +109,36 @@ class IndexTest extends TestCase
     }
 
     /**
+     * @param string[]     $columns
+     * @param int[]|null[] $lengths1
+     * @param int[]|null[] $lengths2
+     *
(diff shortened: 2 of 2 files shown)


morozov:
Not sure how this is supposed to work when the target platform doesn't support index lengths but they are specified in the metadata. Something should remove lengths from the target definitions in order to avoid the diff which is impossible to eliminate.

lib/Doctrine/DBAL/Schema/Index.php
```diff
@@ -336,4 +341,17 @@ private function samePartialIndex(Index $other)
...
+    /**
+     * Returns whether the index has the same column lengths as the other
+     */
+    private function hasSameColumnLengths(self $other) : bool
+    {
+        $filter = static function (?int $length) : bool {
+            return $length !== null;
+        };
+
+        return array_filter($this->options['lengths'] ?? [], $filter)
```
Ocramius: Does the order inside `lengths` affect the results? Also, is it always guaranteed to be an `array`?

tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
```diff
@@ -108,6 +108,36 @@ public function testOverrulesWithPartial()
...
+     * @return mixed[][]
+     */
+    public static function indexLengthProvider() : iterable
+    {
+        return [
+            'empty' => [['column'], [], [], true],
+            'same' => [['column'], [64], [64], true],
+            'different' => [['column'], [32], [64], false],
+            'sparse-different-positions' => [['column1', 'column2'], [0 => 32], [1 => 32], false],
+            'sparse-same-positions' => [['column1', 'column2'], [null, 32], [1 => 32], true],
```
Ocramius: Looking at this data, are multiple `$lengths` ever possible?

Ocramius (changes requested):

lib/Doctrine/DBAL/Schema/Index.php
morozov: The order of lengths is expected to be the same as the order of columns, lengths should be an array:

https://github.com/doctrine/dbal/blob/e37fc54d91a9bad1a7579520f2ee63c90e9099e3/lib/Doctrine/DBAL/Schema/Index.php#L101-L107

tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
morozov: Nothing in the [documentation](https://dev.mysql.com/doc/refman/8.0/en/create-index.html) says it's impossible. If the length is omitted, the entire field length will be used. I haven't tried it myself though. Why do you think it's impossible?

Ocramius: I'm wondering about this syntax: `[1 => 32, 2 => 33, 3 => 34]`

Is this something possible? I'm mostly wondering about why the type is an `array` in first place.

lib/Doctrine/DBAL/Schema/Index.php
Ocramius: Aha, that clears it up :+1:

tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
Ocramius: Handled by https://github.com/doctrine/dbal/pull/3415#discussion_r243029973

Ocramius (approved):
:+1:

leofeyer:
Thank you @morozov. 👍

rvanlaak:
We started seeing the following error after updating from `2.9.0` to `2.9.2`

>  [PDOException (42000)]                                                       
>  SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key was t  
>  oo long; max key length is 767 bytes

The schema tool did try to drop and create the index, but was not able to make the change. We were able to solve it by manually removing the index length, after which the schema tool did not detect a change for the index anymore.

PR #3417, opened 2018-12-22
Statement::fetchColumn() will throw an exception in the case of invalid index

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | #3358

This is a BC-breaking implementation of the changes originally implemented in #3359.

### DIFF
--- a/lib/Doctrine/DBAL/Cache/ArrayStatement.php
+++ b/lib/Doctrine/DBAL/Cache/ArrayStatement.php
@@ -3,10 +3,12 @@
 namespace Doctrine\DBAL\Cache;
 
 use ArrayIterator;
+use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Driver\ResultStatement;
 use Doctrine\DBAL\FetchMode;
 use InvalidArgumentException;
 use IteratorAggregate;
+use function array_key_exists;
 use function array_merge;
 use function array_values;
 use function count;
@@ -130,7 +132,14 @@ class ArrayStatement implements IteratorAggregate, ResultStatement
     {
         $row = $this->fetch(FetchMode::NUMERIC);
 
-        // TODO: verify that return false is the correct behavior
-        return $row[$columnIndex] ?? false;
+        if ($row === false) {
+            return false;
+        }
+
+        if (! array_key_exists($columnIndex, $row)) {
+            throw DBALException::invalidColumnIndex($columnIndex, count($row));
+        }
+
+        return $row[$columnIndex];
     }
 }
--- a/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
+++ b/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
@@ -4,14 +4,17 @@ namespace Doctrine\DBAL\Cache;
 
 use ArrayIterator;
 use Doctrine\Common\Cache\Cache;
+use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Driver\ResultStatement;
 use Doctrine\DBAL\Driver\Statement;
 use Doctrine\DBAL\FetchMode;
 use InvalidArgumentException;
 use IteratorAggregate;
+use function array_key_exists;
 use function array_merge;
 use function array_values;
 use function assert;
+use function count;
 use function reset;
 
 /**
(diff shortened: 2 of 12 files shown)


lib/Doctrine/DBAL/Cache/ArrayStatement.php
```diff
@@ -130,7 +131,14 @@ public function fetchColumn($columnIndex = 0)
     {
         $row = $this->fetch(FetchMode::NUMERIC);
 
-        // TODO: verify that return false is the correct behavior
-        return $row[$columnIndex] ?? false;
+        if ($row === false) {
+            return false;
+        }
+
+        if ($columnIndex >= count($row)) {
```
Ocramius: Would an `array_key_exists()` be better here?

lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
```diff
@@ -175,8 +177,15 @@ public function fetchColumn($columnIndex = 0)
     {
         $row = $this->fetch(FetchMode::NUMERIC);
 
-        // TODO: verify that return false is the correct behavior
-        return $row[$columnIndex] ?? false;
+        if ($row === false) {
+            return false;
+        }
+
+        if ($columnIndex >= count($row)) {
```
Ocramius: Same as above: `array_key_exists()`?

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
```diff
@@ -350,7 +351,11 @@ public function fetchColumn($columnIndex = 0)
             return false;
         }
 
-        return $row[$columnIndex] ?? null;
+        if ($columnIndex >= count($row)) {
```
Ocramius: Same as above: `array_key_exists()`?

lib/Doctrine/DBAL/DBALException.php
```diff
@@ -282,4 +282,14 @@ public static function typeNotFound($name)
     {
         return new self('Type to be overwritten ' . $name . ' does not exist.');
     }
+
+    public static function invalidColumnIndex(int $index, int $count) : self
```
Ocramius: Would it make sense to provide the `$row`, and add the array keys to the message?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -500,7 +501,11 @@ public function fetchColumn($columnIndex = 0)
             return false;
         }
 
-        return $row[$columnIndex] ?? null;
+        if ($columnIndex >= count($row)) {
```
Ocramius: Same as above: `array_key_exists()`?

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereStatement.php
```diff
@@ -282,7 +283,9 @@ public function fetchColumn($columnIndex = 0)
             return false;
         }
 
-        return $row[$columnIndex] ?? null;
+        if ($columnIndex >= count($row)) {
```
Ocramius: Same as above: `array_key_exists()`?

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
```diff
@@ -401,7 +402,11 @@ public function fetchColumn($columnIndex = 0)
             return false;
         }
 
-        return $row[$columnIndex] ?? null;
+        if ($columnIndex >= count($row)) {
```
Ocramius: Same as above: `array_key_exists()`?

Ocramius (changes requested):
The comments suggesting `array_key_exists()` usage are just suggestions: I think they are semantically more correct than using `count()`.

lib/Doctrine/DBAL/Cache/ArrayStatement.php
morozov: Actually, yes. Otherwise, the case when `$columnIndex < 0` won't be handled.

lib/Doctrine/DBAL/DBALException.php
morozov: `$rows` is always a zero-based array since we're fetching a column by index, so the count is enough.

Ocramius: :+1:

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
```diff
@@ -349,4 +350,34 @@ public function testExecWithRedundantParameters() : void
...
+
+    /**
+     * @throws DBALException
+     *
+     * @dataProvider nonExistingIndexProvider
+     */
+    public function testFetchColumnNonExistingIndex(int $index) : void
+    {
+        if ($this->connection->getWrappedConnection() instanceof PDOConnection) {
+            $this->markTestSkipped('PDO supports this behavior natively but throws a different exception');
```
morozov: I wouldn't bother handling the PDO exception since the only way to identify it is by the message. It's a logical exception, so it's not supposed to be handled at runtime anyways.

Majkl578 (approved):

PR #3420, opened 2019-01-02
Index length can be a `string`: ensure that it is an integer when read by the `MySqlSchemaManager`

leofeyer:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

Fixes #3415.

#### Summary

When using MySQL, `AbstractSchemaManager::listTableIndexes()` will return the index length as string, which makes the strict comparison in `Index::hasSameColumnLengths()` fail. `$this->options['length']` will have the index length as string and `$other->options['length']` as integer.

I am not 100% sure if my fix is in the right place, as we could also cast to int in the [_getPortableTableIndexesList()](https://github.com/doctrine/dbal/blob/2.9/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php#L865) method. So feel free to adjust the PR as necessary.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
@@ -107,6 +107,19 @@ class MySqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
         self::assertTrue($indexes['s_index']->hasFlag('spatial'));
     }
 
+    public function testIndexWithLength() : void
+    {
+        $table = new Table('index_length');
+        $table->addColumn('text', 'string', ['length' => 255]);
+        $table->addIndex(['text'], 'text_index', [], ['lengths' => [128]]);
+
+        $this->schemaManager->dropAndCreateTable($table);
+
+        $indexes = $this->schemaManager->listTableIndexes('index_length');
+        self::assertArrayHasKey('text_index', $indexes);
+        self::assertSame([128], $indexes['text_index']->getOption('lengths'));
+    }
+
     /**
      * @group DBAL-400
      */
--- a/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
@@ -68,7 +68,7 @@ class MySqlSchemaManager extends AbstractSchemaManager
             } elseif (strpos($v['index_type'], 'SPATIAL') !== false) {
                 $v['flags'] = ['SPATIAL'];
             }
-            $v['length'] = $v['sub_part'] ?? null;
+            $v['length'] = isset($v['sub_part']) ? (int) $v['sub_part'] : null;
 
             $tableIndexes[$k] = $v;
         }


lib/Doctrine/DBAL/Schema/Index.php
```diff
@@ -351,7 +351,7 @@ private function hasSameColumnLengths(self $other) : bool
             return $length !== null;
         };
 
-        return array_filter($this->options['lengths'] ?? [], $filter)
-            === array_filter($other->options['lengths'] ?? [], $filter);
+        return array_map('intval', array_filter($this->options['lengths'] ?? [], $filter))
```
Ocramius: I'd say `lengths` should always be `int[]`. Any way to enforce that at introspection/validation time?

leofeyer: Yes, we could cast to int in the [`_getPortableTableIndexesList()`](https://github.com/doctrine/dbal/blob/2.9/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php#L865) method:

```php
$result[$keyName]['options']['lengths'][] = isset($tableIndex['length']) ? (int) $tableIndex['length'] : null;
```

Ocramius: Probably better to `assert()` that it is integer-ish, rather than blindly applying `(int)`

greg0ire: Asserting great, but you don't actually mean `assert()` should be used for this, right?

Ocramius: I expect a crash, so yes, this should be a conditional with an exception.

morozov: > Yes, we could cast to int in the [`_getPortableTableIndexesList()`](https://github.com/doctrine/dbal/blob/2.9/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php#L865) method:

Please do. It should be cast to the proper type as soon as it crosses the architecture boundary DB driver → DBAL. The rest of the code should only expect an `int`.

leofeyer: I have adjusted the PR accordingly, however, I am not sure where to put the test. Can you point me in the right direction please?

Ocramius: Possibly https://github.com/doctrine/dbal/blob/f28f3a83e29baca5d9b4ee6bebc11465053055aa/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php ?

leofeyer: I have added a test in 3d3973ead11bd3457898731aaebd03fb70b9defd.

leofeyer:
@morozov Can we be certain that this issue is not limited to MySQL? Otherwise I would prefer to apply the changes in the `MySqlSchemaManager` class only.

lib/Doctrine/DBAL/Schema/Index.php
greg0ire: @morozov what do you think about https://github.com/doctrine/dbal/pull/3420#discussion_r244757515 ?

morozov: I don't think this case is different from all others where we just cast it, especially given that the bug is not about the value not being an integer. It is an integer represented as a string which is true for most cases.

Unless @Ocramius can suggest a way to check the value, I'm fine with not having it. Should it be `preg_match()` or `filter_var()` or `ctype_digit()`? Note that the latter two will introduce implicit dependencies on a not yet used PHP extension.

morozov:
@leofeyer if I'm not mistaken, prefix indices are only supported by MySQL, so it should be fine to fix/test it only for MySQL. Once the changes are complete, please squash the commits.

Ocramius (dismissed):
:+1:

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
```diff
@@ -107,6 +107,19 @@ public function testSpatialIndex()
         self::assertTrue($indexes['s_index']->hasFlag('spatial'));
     }
 
+    public function testIndexWithLength()
```
Ocramius: `: void`

leofeyer: Addded in d82b245a25c8dc5d4b216769946e85bf3a1685a4.

I wasn't sure about adding it, because only 7 of the 20 test methods have it. What's the rule here? Keeping the existing methods as they are but adding return types to all new ones?

Ocramius: I don't think that we've added `doctrine/coding-standards` to this codebase, but we try to follow it for any new diffs anyway.

In `doctrine/orm`, any diff like this would have led to an automated CS check failure

morozov:
@leofeyer please rebase on master and fix the code style violation.

leofeyer:
Done.

morozov (approved):

Issue #950, opened 2014-10-22
DBAL-1020: Postgres and using Schema tool throws cardinality errors


doctrinebot:
Jira issue originally created by user intellix:

Postgres: 9.3.5.0 (Postgres App for OSX) w/ PostGIS extensions
doctrine/common: 2.4.x-dev ae92d076442e27b6910dd86a1292a8867cf5cfe4
doctrine/dbal: dev-master 1c9c24a7e2295b71249ae2a719ce38861fccd551
creof/doctrine2-spatial: https://github.com/intellix/doctrine2-spatial 4023ca8fbe703043012c31d6df26b9bc7b0a972d
## 

It seems every now and again when I come to use the schema-tool I'm getting exceptions which can only be fixed by dropping the database and recreating from scratch.

The following SQL looks to be generated here: \Doctrine\DBAL\Platforms\AbstractPlatform::getListTableForeignKeysSQL

`SELECT quote*ident(r.conname) as conname, pg_catalog.pg_get*constraintdef(r.oid, true) as condef
                  FROM pg*catalog.pg*constraint r
                  WHERE r.conrelid =
                  (
                      SELECT c.oid
                      FROM pg*catalog.pg_class c, pg_catalog.pg*namespace n
                      WHERE n.nspname NOT IN ('pg*catalog', 'information_schema', 'pg_toast') AND c.relname = 'state' AND n.nspname = ANY(string_to_array((select replace(replace(setting,'"$user"',user),' ','') from pg_catalog.pg_settings where name = 'search*path'),',')) AND n.oid = c.relnamespace
                  )
                  AND r.contype = 'f'`

The full stack trace is as follows:

```
 ---<sub>---</sub>---<sub>---</sub>---<sub>---</sub>---<sub>---</sub>---<sub>---</sub>---<sub>---</sub>---<sub>---</sub>---<sub>---</sub>---<sub>--
  Dropping database schema...
      ./bin/doctrine-module orm:schema-tool:drop --force --verbose
---</sub>---<sub>---</sub>---<sub>---</sub>---<sub>---</sub>---<sub>---</sub>---<sub>---</sub>---<sub>---</sub>---<sub>---</sub>---<sub>---</sub>--
Dropping database schema...



  [Doctrine\DBAL\Exception\DriverException]                                                                                                                                                            
  An exception occurred while executing 'SELECT quote*ident(r.conname) as conname, pg_catalog.pg_get*constraintdef(r.oid, true) as condef                                                              
[...]

PR #3422, opened 2019-01-02
Modified subquery in PostgreSqlPlatform->getListTableForeignKeysSQL()

jwilson-cee:
Fixed cardinality issue where subquery could return more than than one row in getListTableForeignKeysSQL function

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #950 

#### Summary

The cardinality error would happen when the schema wasn't prefixed in the $table parameter. As the subquery in the `WHERE =` clause would return more than one result if there was the same table name in two different schemas. (This is specific to postgres only.)

This adds:

- `LIMIT 1` to the subquery of the `WHERE = (SELECT ...)` clause
- A clearer parameter name `$tableWithSchema` and doc blocks to convey what's expected for this function
- A deprecation message for when the schema is not included

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -16,6 +16,7 @@ use Doctrine\DBAL\Types\BlobType;
 use Doctrine\DBAL\Types\IntegerType;
 use Doctrine\DBAL\Types\Type;
 use UnexpectedValueException;
+use const E_USER_DEPRECATED;
 use function array_diff;
 use function array_merge;
 use function array_unique;
@@ -31,6 +32,7 @@ use function is_string;
 use function sprintf;
 use function strpos;
 use function strtolower;
+use function trigger_error;
 use function trim;
 
 /**
@@ -272,9 +274,9 @@ class PostgreSqlPlatform extends AbstractPlatform
     }
 
     /**
-     * {@inheritDoc}
+     * @param string $tableWithSchema The name of the table prefixed with the schema (i.e. 'some_schema.some_table').
      */
-    public function getListTableForeignKeysSQL(string $table, ?string $database = null) : string
+    public function getListTableForeignKeysSQL(string $tableWithSchema, ?string $database = null) : string
     {
         return 'SELECT quote_ident(r.conname) as conname, pg_catalog.pg_get_constraintdef(r.oid, true) as condef
                   FROM pg_catalog.pg_constraint r
@@ -282,7 +284,8 @@ class PostgreSqlPlatform extends AbstractPlatform
                   (
                       SELECT c.oid
                       FROM pg_catalog.pg_class c, pg_catalog.pg_namespace n
-                      WHERE ' . $this->getTableWhereClause($table) . " AND n.oid = c.relnamespace
+                      WHERE ' . $this->getTableWhereClause($tableWithSchema) . " AND n.oid = c.relnamespace
+                      LIMIT 1
                   )
                   AND r.contype = 'f'";
     }
@@ -331,11 +334,9 @@ SQL
     }
 
     /**
-     * {@inheritDoc}
-     *
-     * @link http://ezcomponents.org/docs/api/trunk/DatabaseSchema/ezcDbSchemaPgsqlReader.html
+     * @param string $tableWithSchema The name of the table prefixed with the schema (i.e. 'some_schema.some_table').
      */
-    public function getListTableIndexesSQL(string $table, ?string $currentDatabase = null) : string
+    public function getListTableIndexesSQL(string $tableWithSchema, ?string $currentDatabase = null) : string
     {
         return 'SELECT quote_ident(relname) as relname, pg_index.indisunique, pg_index.indisprimary,
                        pg_index.indkey, pg_index.indrelid,
@@ -344,25 +345,33 @@ SQL
                  WHERE oid IN (
                     SELECT indexrelid
                     FROM pg_index si, pg_class sc, pg_namespace sn
-                    WHERE ' . $this->getTableWhereClause($table, 'sc', 'sn') . ' AND sc.oid=si.indrelid AND sc.relnamespace = sn.oid
+                    WHERE ' . $this->getTableWhereClause($tableWithSchema, 'sc', 'sn') . ' AND sc.oid=si.indrelid AND sc.relnamespace = sn.oid
                  ) AND pg_index.indexrelid = oid';
     }
 
-    private function getTableWhereClause(string $table, string $classAlias = 'c', string $namespaceAlias = 'n') : string
+    /**
+     * @param string $tableWithSchema The name of the table prefixed with the schema (i.e. 'some_schema.some_table').
+     */
+    private function getTableWhereClause(string $tableWithSchema, string $classAlias = 'c', string $namespaceAlias = 'n') : string
     {
-        $whereClause = $namespaceAlias . ".nspname NOT IN ('pg_catalog', 'information_schema', 'pg_toast') AND ";
-        if (strpos($table, '.') !== false) {
-            [$schema, $table] = explode('.', $table);
+        if (strpos($tableWithSchema, '.') !== false) {
+            [$schema, $table] = explode('.', $tableWithSchema);
             $schema           = $this->quoteStringLiteral($schema);
         } else {
+            $table  = $tableWithSchema;
             $schema = "ANY(string_to_array((select replace(replace(setting,'\"\$user\"',user),' ','') from pg_catalog.pg_settings where name = 'search_path'),','))";
+            @trigger_error(sprintf(
+                'Providing a table name without a schema prefix, i.e. \'some_schema.some_table\', (%s given) is deprecated and will cause an error in Doctrine 3.0',
+                $tableWithSchema
+            ), E_USER_DEPRECATED);
         }
 
         $table = new Identifier($table);
         $table = $this->quoteStringLiteral($table->getName());
 
-        return $whereClause . sprintf(
-            '%s.relname = %s AND %s.nspname = %s',
+        return sprintf(
+            "%s.nspname NOT IN ('pg_catalog', 'information_schema', 'pg_toast') AND %s.relname = %s AND %s.nspname = %s",
+            $namespaceAlias,
             $classAlias,
             $table,
             $namespaceAlias,
@@ -371,9 +380,9 @@ SQL
     }
 
     /**
-     * {@inheritDoc}
+     * @param string $tableWithSchema The name of the table prefixed with the schema (i.e. 'some_schema.some_table').
      */
-    public function getListTableColumnsSQL(string $table, ?string $database = null) : string
+    public function getListTableColumnsSQL(string $tableWithSchema, ?string $database = null) : string
     {
         return "SELECT
                     a.attnum,
@@ -400,7 +409,7 @@ SQL
                         FROM pg_description WHERE pg_description.objoid = c.oid AND a.attnum = pg_description.objsubid
                     ) AS comment
                     FROM pg_attribute a, pg_class c, pg_type t, pg_namespace n
-                    WHERE " . $this->getTableWhereClause($table, 'c', 'n') . '
+                    WHERE " . $this->getTableWhereClause($tableWithSchema, 'c', 'n') . '
                         AND a.attnum > 0
                         AND a.attrelid = c.oid
                         AND a.atttypid = t.oid
--- a/tests/Doctrine/Tests/DBAL/Schema/PostgreSqlListTableFunctionsTest.php
+++ b/tests/Doctrine/Tests/DBAL/Schema/PostgreSqlListTableFunctionsTest.php
@@ -1,0 +1,73 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL\Schema;
+
+use Doctrine\DBAL\DriverManager;
+use Doctrine\DBAL\Platforms\PostgreSqlPlatform;
+use Doctrine\Tests\DbalFunctionalTestCase;
+use function extension_loaded;
+use function in_array;
+
+class PostgreSqlListTableFunctionsTest extends DbalFunctionalTestCase
+{
+    protected function setUp() : void
+    {
+        if (! extension_loaded('pdo_pgsql')) {
+            $this->markTestSkipped('pdo_pgsql is not loaded.');
+
+            return;
+        }
+
+        parent::setUp();
+
+        if (! in_array('pdo_pgsql', DriverManager::getAvailableDrivers())) {
+            $this->markTestSkipped('PostgreSQL driver not available');
+
+            return;
+        }
+
+        if (! $this->connection->getDatabasePlatform() instanceof PostgreSqlPlatform) {
+            $this->markTestSkipped('PostgreSQL Only test.');
+
+            return;
+        }
+
+        $this->connection->executeQuery('CREATE SCHEMA test_schema1;');
+        $this->connection->executeQuery('CREATE SCHEMA test_schema2;');
+        $this->connection->executeQuery('set search_path = "test_schema1", "test_schema2";');
+        $this->connection->executeQuery('CREATE TABLE test_schema1.test_foreign1 ( id integer PRIMARY KEY );');
+        $this->connection->executeQuery('CREATE TABLE test_schema1.test_table ( test_column1 varchar(5), test_foreign1_id integer constraint fk_test_foreign1 references test_schema1.test_foreign1 (id) );');
+        $this->connection->executeQuery('CREATE INDEX idx_test_column1 ON test_schema1.test_table (test_column1);');
+        $this->connection->executeQuery('CREATE TABLE test_schema2.test_foreign2 ( id integer PRIMARY KEY );');
+        $this->connection->executeQuery('CREATE TABLE test_schema2.test_table ( test_column2 varchar(5), test_foreign2_id integer constraint fk_test_foreign2 references test_schema2.test_foreign2 (id) );');
+        $this->connection->executeQuery('CREATE INDEX idx_test_column2 ON test_schema2.test_table (test_column2);');
+    }
+
+    public function testListTableFunctions() : void
+    {
+        $foreignKeys = $this->connection->getSchemaManager()->listTableForeignKeys('test_table');
+        $this->assertNotEmpty($foreignKeys);
+        $foreignKeys = $this->connection->getSchemaManager()->listTableForeignKeys('test_schema1.test_table');
+        $columns     = isset($foreignKeys[0]) ? $foreignKeys[0]->getColumns() : [];
+        $this->assertContains('test_foreign1_id', $columns);
+        $foreignKeys = $this->connection->getSchemaManager()->listTableForeignKeys('test_schema2.test_table');
+        $columns     = isset($foreignKeys[0]) ? $foreignKeys[0]->getColumns() : [];
+        $this->assertContains('test_foreign2_id', $columns);
+
+        $columns = $this->connection->getSchemaManager()->listTableColumns('test_table');
+        $this->assertNotEmpty($columns);
+        $columns = $this->connection->getSchemaManager()->listTableColumns('test_schema1.test_table');
+        $this->assertArrayHasKey('test_column1', $columns);
+        $columns = $this->connection->getSchemaManager()->listTableColumns('test_schema2.test_table');
+        $this->assertArrayHasKey('test_column2', $columns);
+
+        $indexes = $this->connection->getSchemaManager()->listTableIndexes('test_table');
+        $this->assertNotEmpty($indexes);
+        $indexes = $this->connection->getSchemaManager()->listTableIndexes('test_schema1.test_table');
+        $this->assertArrayHasKey('idx_test_column1', $indexes);
+        $indexes = $this->connection->getSchemaManager()->listTableIndexes('test_schema2.test_table');
+        $this->assertArrayHasKey('idx_test_column2', $indexes);
+    }
+}
--- a/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
@@ -98,6 +98,18 @@ class PostgreSqlSchemaManager extends AbstractSchemaManager
     }
 
     /**
+     * Lists the foreign keys for the given table in a given schema.
+     *
+     * @param string $tableWithSchema The name of the table prefixed with the schema (i.e. 'some_schema.some_table').
+     *
+     * @return array<int|string, ForeignKeyConstraint>
+     */
+    public function listTableForeignKeys(string $tableWithSchema, ?string $database = null) : array
+    {
+        return parent::listTableForeignKeys($tableWithSchema, $database);
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function dropDatabase(string $database) : void
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractPostgreSqlPlatformTestCase.php
@@ -961,7 +961,7 @@ abstract class AbstractPostgreSqlPlatformTestCase extends AbstractPlatformTestCa
     {
         self::assertStringContainsStringIgnoringCase(
             "'Foo''Bar\\'",
-            $this->platform->getListTableForeignKeysSQL("Foo'Bar\\")
+            $this->platform->getListTableForeignKeysSQL("public.Foo'Bar\\")
         );
     }
 
@@ -994,7 +994,7 @@ abstract class AbstractPostgreSqlPlatformTestCase extends AbstractPlatformTestCa
     {
         self::assertStringContainsStringIgnoringCase(
             "'Foo''Bar\\'",
-            $this->platform->getListTableIndexesSQL("Foo'Bar\\")
+            $this->platform->getListTableIndexesSQL("public.Foo'Bar\\")
         );
     }
 
@@ -1016,7 +1016,7 @@ abstract class AbstractPostgreSqlPlatformTestCase extends AbstractPlatformTestCa
     {
         self::assertStringContainsStringIgnoringCase(
             "'Foo''Bar\\'",
-            $this->platform->getListTableColumnsSQL("Foo'Bar\\")
+            $this->platform->getListTableColumnsSQL("public.Foo'Bar\\")
         );
     }
 


greg0ire:
Please format your commit message according to our rules : 

- The commit message subject must be less than 50 characters and tell us *what* you did.
- The commit message body should tell us *why* you did it. It is optional but highly recommended.

Bad example :

```
Fixed bug #989 by removing call to defraculate()
```

Good example:

```
Remove call to defraculate()

Calling this function caused a bug because it interferes 
with calls to getSchmeckles().
Fixes #989
```

More details in CONTRIBUTING.md

greg0ire (dismissed):

Ocramius:
Change looks good, but I'm wondering if there is a way to test this patch?

jwilson-cee:
> Change looks good, but I'm wondering if there is a way to test this patch?

In order to reproduce this error you need to have a postgres database that has two schemas that have a table with the same name:
```
TestDB
    ->SchemaName1
        ->TestTable
    ->SchemaName2
        ->TestTable
````
Then to trigger the error, fetch a column type from the schema:
`Schema::getColumnType('TestTable', 'id')`

**Comment by Ocramius** [comment]
Yes, and that needs to be in our automation test suite with the other tests

**Comment by jwilson-cee** [comment]
> Yes, and that needs to be in our automation test suite with the other tests

@Ocramius, you asked a question and I gave answer. What's the purpose of this comment? Do you want me to write a test for this?

**Comment by Ocramius** [comment]
Yes, in @doctrine/doctrinecore we don't merge fixes without tests, unless they fix an already failing test

File: lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php [hunk 1]
```diff
@@ -283,7 +283,7 @@ public function getListTableForeignKeysSQL($table, $database = null)
     {
         return 'SELECT quote_ident(r.conname) as conname, pg_catalog.pg_get_constraintdef(r.oid, true) as condef
                   FROM pg_catalog.pg_constraint r
-                  WHERE r.conrelid =
+                  WHERE r.conrelid IN
```
morozov: Isn't it because it ignores the `$database` parameter? Based on the scenario described in https://github.com/doctrine/dbal/pull/3422#issuecomment-463331428, I'd expect the query to select the rows corresponding to the table in its schema, not other tables with the same name in other schemas.

morozov (dismissed):

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
jwilson-cee: The `$database` parameter would not be useful for the query in this function as it would just be a string of the database name. We'd need the name of the schema to be able to make this query be more specific, which this function isn't equipped to receive. 

IMO this is the best solution to this issue. Otherwise you'll have to go down the rabbit trail of how to pass a schema name down to this function. And I don't believe that would be worth the effort (my effort anyway) as this is such a rare situation where there would be two schemas with the same table name AND have foreign key constraints.

jwilson-cee:
Ok, I'm giving up on this. You guys can take it or leave it.

I can't get it to pass some of your code checks. 

The code sniffer in travis is complaining about type hinting array items, I can't find any examples of type hinting array items. And it's complaining about the in_array() function, of which cannot find any documentation or reference to this error `Function in_array() should not be referenced via a fallback global name, but via a use statement. (SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly.ReferenceViaFallbackGlobalName`.

Also getting an error from appveyor about it not being able to find a driver. Not sure what to do about that, as I can get the phpunit test to pass locally.

Ocramius:
> `Function in_array() should not be referenced via a fallback global name, but via a use statement.`

This means that at the top of the file, after the other `use` statements, there should be a `use function in_array;`

 > I can't find any examples of type hinting array items.

The  syntax is something like `@var Foo[]` or `@param Foo[] $paramName` or `@return Foo[]`

andrew-demb:
Seems like travis failures is not related to PR.
Can we retry checks, or merge it?

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
morozov: > The `$database` parameter would not be useful for the query in this function as it would just be a string of the database name. We'd need the name of the schema to be able to make this query be more specific […]

This is what I'm talking about. If I'm not mistaken, `$database` in the DBAL terminology corresponds to the PostgreSQL schema.

jwilson-cee: > If I'm not mistaken, $database in the DBAL terminology corresponds to the PostgreSQL schema.

This is incorrect. `$database` contains the name of the database, not the name of the schema. In PostgreSQL, a database can have multiple schemas: `database -> schemas -> tables` . It's the schema name that is needed for added specificity, not the database name.

So, the `$database` parameter is actually completely useless for this function. I'm not sure why it's even there. My guess is it's just carry-over (copy/paste) from other drivers that override this abstract function (see MySqlPlatform->getListTableForeignKeysSQL()).

jwilson-cee:
> Seems like travis failures is not related to PR.
> Can we retry checks, or merge it? - @andrew-demb 

I believe this PR passes all the checks. I think it is being blocked by @morozov 's Change Request comment? 
https://github.com/doctrine/dbal/pull/3422/files/0a55ce69a634009624f30f11971770ffbeb42099
(you may need to click on "Show Conversation" to see the discussion)

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
morozov: You're right. DBAL doesn't use `$database` to indicate a schema for PostgreSQL. However, it still doesn't look right to return foreign keys for tables with the same name from all schemas as a workaround. Please see if you can use `schema.table` as the table name since it seems to be supported in:

https://github.com/doctrine/dbal/blob/0a55ce69a634009624f30f11971770ffbeb42099/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php#L366-L371

jwilson-cee: You've pretty much just shown that the solution you are looking for is already in the `getListTableForeignKeysSQL` function.

`getListTableForeignKeysSQL` utilizes the `getTableWhereClause` function to filter for schema. 
https://github.com/doctrine/dbal/blob/0a55ce69a634009624f30f11971770ffbeb42099/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php#L290

But the `$table` parameter must have the schema in it (i.e. `$table = 'some_schema.some_table'`), otherwise the `getTableWhereClause` will extrapolate a list of schemas from the users search_path. 

So, passing specificity to a single schema is already possible in this function. The developer can included the schema with the table name in the `$table` parameter ('some_schema.some_table'). But if they don't, it'll fall back on a list of schemas in the DB users search_path.

Therefore, the very simple fix this PR (literally just one key stroke on one line code) is still necessary. And there is no need to figure out how to pass a specific schema to this function - as you have found that it is already available in the code.

With all due respect, I implore you to stop beating around the bush with this question of schema specificity, and please let this PR be merged.

Thank you.

greg0ire: I think the right fix would be to trigger a deprecation if `$table` doesn't contain a dot, because that's just an incorrect usage. I think @morozov is right, listing all the foreign keys of all the tables named like `$table` makes no sense. Right now, this doesn't crash when there is only one table, right?

greg0ire:

How to do that?

1. `git rebase -i origin/master`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

tests/Doctrine/Tests/DBAL/Schema/PostgreSqlListTableForeignKeysTest.php
```diff
@@ -0,0 +1,68 @@
...
+                $connection->executeQuery('CREATE TABLE test_schema1.test_foreign ( id integer PRIMARY KEY );');
+                $connection->executeQuery('CREATE TABLE test_schema1.test ( test_column varchar(5), test_foreign_id integer constraint fk_test_foreign references test_schema1.test_foreign (id) );');
+                $connection->executeQuery('CREATE TABLE test_schema2.test_foreign ( id integer PRIMARY KEY );');
+                $connection->executeQuery('CREATE TABLE test_schema2.test ( test_column varchar(5), test_foreign_id integer constraint fk_test_foreign references test_schema2.test_foreign (id) );');
+            } else {
+                $this->markTestSkipped('PostgreSQL Only test');
+            }
+        } else {
+            $this->markTestSkipped('PostgreSQL driver not available');
+        }
```
greg0ire: Please use early return statements in this method rather than `else` statements.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
jwilson-cee: The main issue here is that this **is** a bug that needs to be fixed. A WHERE clause that is comparing against a sub SELECT statement must either be an IN comparison or if it's an equals (=) comparison then the SELECT statement **must** return only one row (i.e. LIMIT 1). The current code does not comply with those SQL rules, and that is what this PR is intended to address.

The theology of what the `getListTableForeignKeysSQL()` should return is a different discussion. Yes, it is related to how this fix could be implemented. And I'm willing to have that discussion. But I'm not just going arbitrarily pick a direction.

Here is my opinion - it is **not** incorrect usage to pass just the table name without the schema, as listing all foreign keys for the table that is passed is what the function is built to do. The `getListTableForeignKeysSQL()` function is used in the `AbstractSchemaManager->listTableForeignKeys()` and this is the description in the DocBlock for the purpose of this function: `Lists the foreign keys for the given table.` It might _"make no sense"_ to you, but as odd as it may be, someone might find a use case for getting a list of all foreign keys for a table name across multiple schemas. So, IMHO, this is fine as it is.

But if you feel strongly about limiting this function to require a schema in the table name (in the context of postresql), then `AbstractSchemaManager->listTableForeignKeys()` should be overridden in `PostgreSqlSchemaManager` with the check for the dot in the `$table` in there (not in the `getListTableForeignKeysSQL()` function), and the DocBlock description changed to 
`Lists the foreign keys for the given table in a given schema.`
`@param string      $table    The name of the table including schema ('some_schema.some_table').`

greg0ire: > someone might find a use case for getting a list of all foreign keys for a table name across multiple schemas.

But that person cannot be using the feature at the moment, since it will inevitably crash, correct? 

That means that we could:

1. catch the current exception when it happens, and wrap it in an `\InvalidArgumentException` with a meaningful message if and only if the database name is not provided by the user. If somebody can think of a use case for this, they will see the meaningful message, and we will use `IN`, and add a test that document the use case.
2. when the exception does not happen, trigger a deprecation informing the user that they should include the database name.
3. once that is merged, we will remove the deprecation and throw an exception as soon as the database name is not provided in the next major version of this library. 

I agree with you about the override of `listTableForeignKeys` but it should only trigger a deprecation for now. The docblock could be changed as you suggested since that is not a BC-break.

WDYT?

tests/Doctrine/Tests/DBAL/Schema/PostgreSqlListTableForeignKeysTest.php
```diff
@@ -0,0 +1,68 @@
...
+        'driver'   => 'pdo_pgsql',
+        'user'     => 'postgres',
+        'password' => '',
+        'host'     => 'localhost',
+        'port'     => '5432',
+    ];
+
+    protected function setUp() : void
+    {
+        if (! extension_loaded('pdo_pgsql')) {
```
greg0ire: Running the test suite with the extension loaded but no database started will cause a crash… Please look into existing tests to avoid this and hardcoding database parameters above. Maybe you should proceed as in https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOPgsqlConnectionTest.php

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
jwilson-cee: Just to provide clarity - you were using the term "database name" in your last comment, when I believe you meant "schema name"

After digging into this further, the `getTableWhereClause()` function is where queries using that function could end up returning multiple results from same table names across multiple schemas, via this line of code:
`$schema = "ANY(string_to_array((select replace(replace(setting,'\"\$user\"',user),' ','') from pg_catalog.pg_settings where name = 'search_path'),','))";` _(which means: any schema in the db user search path)_

And the `getTableWhereClause()` function is used in 3 functions in the `PostgreSqlPlatform` class:
`getListTableIndexesSQL()` which uses the IN comparison in the WHERE clause,
`getListTableColumnsSQL()`,
and `getListTableForeignKeysSQL()` 

So, if you're wanting to test for a dot in the `$table` parameter to trigger a depreciation message, then it should be done in the `getTableWhereClause()` function. Then in the next major version remove the fall back `$schema = 'any schema in db user search path'` and throw an exception if the `$table` doesn't include the schema.

I still say that the malformed SQL in the `getListTableForeignKeysSQL()` needs to be fixed. The `getListTableIndexesSQL()` uses the IN comparison exactly as this PR proposes. So there is precedent, and this leads me to believe that the original developer may have made a mistake in using the = comparison in the `getListTableForeignKeysSQL()` function.

greg0ire: If we use `IN`, then we will start "technically correct" results. That sounds like malicious compliance to me. The fact that there's a precedent would lead me to question the code in `getListTableIndexesSQL`. You can change my mind by showing a plausible use case for getting objects for all the tables with the same name in different schemas, maybe I'm just lacking imagination.

jwilson-cee: > You can change my mind by showing a plausible use case for getting objects for all the tables with the same name in different schemas

For me personally, I don't see a "plausible" use case for this (but someone might). If you are wanting to require the schema be given in the `$table` parameter - I believe that is that is reasonable considering that a "plausible" use case would be rare. I just don't think it's the right approach to use the error resulting from malformed SQL to check for the schema requirement - as you suggested. That's essentially using a bug to validate requirements.

> If we use IN, then we will start "technically correct" results. That sounds like malicious compliance to me. The fact that there's a precedent would lead me to question the code in getListTableIndexesSQL.

You don't have to use `IN` to have the SQL be "compliant" with the rules. You can use `LIMIT 1` in the sub select in the `WHERE = (SELECT ...)` clause. 

So, this is what I suggest, and I'll make these changes if you agree:

- Change the queries in `getListTableForeignKeysSQL` and `getListTableIndexesSQL` to use `LIMIT 1`
- Have the `getTableWhereClause` validate that `$table` is prefixed with a schema, and if it doesn't, trigger a depreciation.
- In the next major version, remove the `$schema` fall back query: `$schema = "ANY(string_to_array((select replace(replace(setting,'\"\$user\"',user),' ','') from pg_catalog.pg_settings where name = 'search_path'),','))"` in the `getTableWhereClause`

greg0ire: Sounds almost perfect to me. I would remove the `LIMIT 1` too in the next major version, but maybe that was implied.

jwilson-cee: Ok, I'll work on implementing this. Why would you want to remove `LIMIT 1` from in the next version?

greg0ire: Because a crash would be an indication that something is wrong, just like it made you detect that something was wrong right now (not limiting the query to a particular schema).

greg0ire (approved):

greg0ire (changes requested):
There are some conflicts

greg0ire:
>  @jwilson-cee jwilson-cee requested a review from greg0ire 4 hours ago 

Github still complains about conflicts… can you please rebase?

greg0ire:
The build is failing because of coding style issues. You can try running `vendor/bin/phpcbf` to fix them.

jwilson-cee:
> The build is failing because of coding style issues. You can try running `vendor/bin/phpcbf` to fix them.

I keep getting this error when trying to run phpcbf:
`ERROR: Referenced sniff "Doctrine" does not exist`

Ocramius:
@jwilson-cee run a `composer install` again: I think this depends on whether post-install scripts were run or not

jwilson-cee:
> @jwilson-cee run a `composer install` again: I think this depends on whether post-install scripts were run or not

Thanks, but getting the same error after re-running composer install. I'll just manually fix the code style.

tests/Doctrine/Tests/DBAL/Schema/PostgreSqlListTableFunctionsTest.php
```diff
@@ -0,0 +1,77 @@
...
+        $this->connection->executeQuery('CREATE TABLE test_schema1.test_table ( test_column1 varchar(5), test_foreign1_id integer constraint fk_test_foreign1 references test_schema1.test_foreign1 (id) );');
+        $this->connection->executeQuery('CREATE INDEX idx_test_column1 ON test_schema1.test_table (test_column1);');
+        $this->connection->executeQuery('CREATE TABLE test_schema2.test_foreign2 ( id integer PRIMARY KEY );');
+        $this->connection->executeQuery('CREATE TABLE test_schema2.test_table ( test_column2 varchar(5), test_foreign2_id integer constraint fk_test_foreign2 references test_schema2.test_foreign2 (id) );');
+        $this->connection->executeQuery('CREATE INDEX idx_test_column2 ON test_schema2.test_table (test_column2);');
+    }
+
+    /**
+     * @group legacy
+     * @expectedDeprecation Providing a table name without a schema prefix, i.e. 'some_schema.some_table', (%s given) is deprecated and will cause an error in Doctrine 3.0
```
greg0ire: Does this actually work? I would expect https://github.com/doctrine/dbal/pull/3643 to disable this. If yes, we forgot to remove other occurrences of `@expectedDeprecation` in that PR.

greg0ire (commented):
This is starting to look really good!

How to do that?

1. `git rebase -i origin/master`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

When doing so, please make sure to follow [the contributing guide](https://www.doctrine-project.org/contribute/index.html#working-on-topic-branches) regarding the commit message. This PR is non-trivial, and that means the commit message should be high quality.

tests/Doctrine/Tests/DBAL/Schema/PostgreSqlListTableFunctionsTest.php
jwilson-cee: Before I rebased to `doctrine:master` I had to have that there for the test to pass. It appears that this test will now pass without having it there. Do you want me to take this out?

greg0ire: Yeah because it's a bit misleading now: we are saying that something is asserted, and it actually isn't:
- false sense of security
- wtf effect if the actual message is changed in a subsequent PR for some reason and that message isn't, and the tests still pass

To be clear, we're talking about line 50, line 49 has to stay.

jwilson-cee: Ok, I've removed it. 
BTW there are three other tests that still have the `@expectedDeprecation` notation.
`PDOStatementTest.php testPDOSpecificModeIsAccepted (line 33)`
`ColumnTest.php testSettingUnknownOptionIsStillSupported (line 67)`
`ColumnTest.php testOptionsShouldNotBeIgnored (line 78)`

greg0ire: I'm going to handle these in a separate PR, thanks.

greg0ire: See #3719

greg0ire (approved):
LGTM

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -344,25 +345,33 @@ public function getListTableIndexesSQL(string $table, ?string $currentDatabase =
...
-        if (strpos($table, '.') !== false) {
-            [$schema, $table] = explode('.', $table);
+        if (strpos($tableWithSchema, '.') !== false) {
+            [$schema, $table] = explode('.', $tableWithSchema);
             $schema           = $this->quoteStringLiteral($schema);
         } else {
+            $table  = $tableWithSchema;
             $schema = "ANY(string_to_array((select replace(replace(setting,'\"\$user\"',user),' ','') from pg_catalog.pg_settings where name = 'search_path'),','))";
+            @trigger_error(sprintf(
+                'Providing a table name without a schema prefix, i.e. \'some_schema.some_table\', (%s given) is deprecated and will cause an error in Doctrine 3.0',
```
morozov: FWIW, `master` already corresponds to `3.0.0-dev`.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -272,17 +274,18 @@ public function getListViewsSQL(string $database) : string
...
     {
         return 'SELECT quote_ident(r.conname) as conname, pg_catalog.pg_get_constraintdef(r.oid, true) as condef
                   FROM pg_catalog.pg_constraint r
                   WHERE r.conrelid =
                   (
                       SELECT c.oid
                       FROM pg_catalog.pg_class c, pg_catalog.pg_namespace n
-                      WHERE ' . $this->getTableWhereClause($table) . " AND n.oid = c.relnamespace
+                      WHERE ' . $this->getTableWhereClause($tableWithSchema) . " AND n.oid = c.relnamespace
+                      LIMIT 1
```
morozov: Instead of failing the query and letting the user know that they are using the API incorrectly, we're hiding the issue and returning one random value instead. I do not agree with this.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
greg0ire: Oh, so this PR should target another branch.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
jwilson-cee: @greg0ire You approved this PR, so why are you giving this a thumbs up?

We are letting the user know that they are using the API incorrectly thru the deprecation message. That's what we agreed on in a previous conversation.

IMO, this is improper use of SQL. When this crashed for me, the error had no useful meaning as to why. I had to dig deep and debug this code myself to find out the problem was in this SQL - not because I was using the API incorrectly. The API was allowing me to pass the table name without a schema. This fix will provide a useful warning, and in the next version it won't allow it.

If you have a specific way you want this fixed, then either fix it your self, or be clear about exactly how you want it fixed. Stating your opinion isn't helpful if you aren't going to provide a solution. Especially since @greg0ire and I have already had this conversation.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
jwilson-cee: Okay? So what needs to be done?? And by who?

morozov: This PR should target `master` and not contain any deprecations. Once this is accepted and merged, it should be rebased and filed again, this time against `2.10`. However, I still fail to see how this issue can be solved in a non-breaking way. I'd rather fix it only in `3.0` and document the proper usage for `2.10`.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
morozov: @jwilson-cee please do not resolve the conversations that you didn't start.

greg0ire: > @greg0ire You approved this PR, so why are you giving this a thumbs up?

Because I mistakenly approved the PR, and I realized @morozov was right, `LIMIT 1` will have the effect of returning something random if you have several schemas (malicious compliance, remember), and shouldn't be needed when using the function properly, right?

> When this crashed for me, the error had no useful meaning as to why.

This is why I proposed this:

> catch the current exception when it happens, and wrap it in an \InvalidArgumentException with a meaningful message if and only if the database schema name is not provided by the user. If somebody can think of a use case for this, they will see the meaningful message, and we will use IN, and add a test that document the use case.

That wouldn't be a BC-break if there already is a crash IMO. Not sure why you didn't address this in your answer (https://github.com/doctrine/dbal/pull/3422#discussion_r340834148). Maybe I wasn't clear enough about this applying to the stable branch.

greg0ire (dismissed):
- retarget needed
- remove `LIMIT 1`
- throw a meaningful exception if and only if there is a crash

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
greg0ire: @morozov are we not using a merge up policy? IMO a non-breaking way to fix this would be to trigger a deprecation when misusing the methods, and to wrap any confusing exception in a higher level, more meaningful exception.

greg0ire:
We're debating this internally, and will do the necessarily retargets for you when we agree. We're in the middle of changing workflows for Doctrine, hence the confusion, sorry for this.

jwilson-cee:
I'm going to close this PR, unless you prefer otherwise.

My reasoning is that I've dug into all the `getListTable...SQL()` functions and I've discovered that all of them return SQL that will generate a list of all objects for a table name that is in multiple schemas (the `getListTableForeignKeysSQL` is the only one that crashes because of the `=` operator). I'll provide all the SQL for this that you can run to see for yourself.

Because of this discovery, I believe might be better to move this discussion to the #950 issue thread than continue in here, as this may warrant further discussion on the direction you want to take for this. And, because of this, it will probably be cleaner to start with a new branch, and let this one die.

If you'd rather keep it in this branch/PR, I'll keep this open and put my findings (the SQL test code) in this comment thread.

Thanks

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
jwilson-cee: > please do not resolve the conversations that you didn't start.

@morozov Sorry if that annoys you, that was not my intent.

When I've addressed a conversation and I feel that I don't need anything more from it, I'll mark it as resolved. As is encouraged in githubs docs https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/commenting-on-a-pull-request#resolving-conversations

This makes it easier for me, the requestor, to see what conversations still need my attention.

Thanks.

greg0ire:
You can start afresh if you think it's best. I'm AFK this weekend though, so don't expect much more from me until Tuesday.

jwilson-cee:
Closing this PR as the fix for this needs further discussion (see issue #950)

Issue #3423, opened 2019-01-04
Rollback called when outside of a transaction when using deferred constraints


grongor:
### Bug Report

In PostgreSQL there are deferrable constraints - constraints that are **checked at the time of commit**, not immediately after issuing an SQL statement.

And when the commit fails because of such constraint (duplicate key in unique constraint, for example) then the transaction is rolled back and an exception is thrown inside PDO.

The problem is that both DBAL and ORM don't expect that behaviour and they try to rollback the transaction after catching that exception. That triggers another exception - trying to rollback a transaction while there is no transaction active.

So this bug creates several issues:
- we try to rollback a transaction when there is no transaction
- we lose information about what error happened in the first place
- the Connection object is unusable after this happens as the internal "transaction level counter" remains at 1 while it should be 0

@simPod already discovered the issue and mentioned it here https://github.com/doctrine/doctrine2/issues/7545 . The problem with his issue is that he didn't find the root cause/only mentioned "previous exception being discarded", which is partly correct but doesn't reflect the severity of the issue. It's not like we only need to know "what happened before that last exception" - the second exception shouldn't have been thrown in the first place.

Code is better then 1000 words, so here you go:
- Failing build; added tests "explaining" the issue: https://travis-ci.org/grongor/dbal/builds/475298852
- Green build; added fix for the failing tests: https://travis-ci.org/grongor/dbal/builds/475299477
- Pull request with both the tests and the fix: https://github.com/doctrine/dbal/pull/3424

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | all up to this day, afaik

#### Current behaviour

You can't use PostgreSQLs deferred constraints as they will kill DBAL/ORM when the constraint is violated.

#### How to reproduce

Create a table with deferrable unique constraint. Try to insert some data to the table when in transaction with deferrable constraint set to defer. Both DBAL and ORM will throw made up errors and end up in unexpected state.

#### Expected behaviour

I would expect to get an unique constraint violation exception. I would also expect to be able to use the connection after the fact.

rela589n:
Has the fix been released yet?

Currently using 3.9.3, and it looks like it doesn't yet have it

simPod:
No, there has been no release since.

rela589n:
@greg0ire , could you please tell if there's any particular date when this is planned for release?

greg0ire:
Yes: there is no particular date planned for any release we do (except major releases)

rela589n:
[...]

PR #3424, opened 2019-01-04
Fix incorrect handling of transactions using deferred constraints

grongor:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3423

#### Summary

Rollback called when outside of a transaction when using deferred constraints (PostgreSQL).

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -1230,14 +1230,15 @@ class Connection
         $this->beginTransaction();
         try {
             $res = $func($this);
-            $this->commit();
-
-            return $res;
         } catch (Throwable $e) {
             $this->rollBack();
 
             throw $e;
         }
+
+        $this->commit();
+
+        return $res;
     }
 
     /**
@@ -1346,7 +1347,13 @@ class Connection
                 $logger->startQuery('"COMMIT"');
             }
 
-            $result = $connection->commit();
+            try {
+                $result = $connection->commit();
+            } catch (Driver\Exception $e) {
+                $this->updateTransactionStateAfterCommit();
+
+                throw $this->convertExceptionDuringQuery($e, 'COMMIT');
+            }
 
             if ($logger !== null) {
                 $logger->stopQuery();
@@ -1362,15 +1369,20 @@ class Connection
             }
         }
 
+        $this->updateTransactionStateAfterCommit();
+
+        return $result;
+    }
+
+    private function updateTransactionStateAfterCommit(): void
+    {
         --$this->transactionNestingLevel;
 
         if ($this->autoCommit !== false || $this->transactionNestingLevel !== 0) {
-            return $result;
+            return;
         }
 
         $this->beginTransaction();
-
-        return $result;
     }
 
     /**
--- a/src/Driver/PDO/Connection.php
+++ b/src/Driver/PDO/Connection.php
@@ -133,7 +133,11 @@ final class Connection implements ServerInfoAwareConnection
      */
     public function commit()
     {
-        return $this->connection->commit();
+        try {
+            return $this->connection->commit();
+        } catch (PDOException $exception) {
+            throw Exception::new($exception);
+        }
     }
 
     /**
--- a/tests/Functional/Ticket/GH3423Test.php
+++ b/tests/Functional/Ticket/GH3423Test.php
@@ -1,0 +1,124 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Tests\Functional\Ticket;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Exception;
+use Doctrine\DBAL\Tests\FunctionalTestCase;
+
+use function sprintf;
+
+/**
+ * @see https://github.com/doctrine/dbal/issues/3423
+ */
+class GH3423Test extends FunctionalTestCase
+{
+    /** @var bool */
+    private static $tableCreated = false;
+
+    /** @var string */
+    private $constraintName = '';
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $platformName = $this->connection->getDatabasePlatform()->getName();
+
+        if ($platformName === 'oracle') {
+            $constraintName = 'GH3423_UNIQUE';
+        } elseif ($platformName === 'postgresql') {
+            $constraintName = 'gh3423_unique';
+        } else {
+            self::markTestSkipped('Only databases supporting deferrable constraints are eligible for this test.');
+        }
+
+        $this->constraintName = $constraintName;
+
+        if (self::$tableCreated) {
+            return;
+        }
+
+        $this->connection->executeStatement(
+            <<<SQL
+            CREATE TABLE gh3423 (
+                unique_field INTEGER NOT NULL
+                    CONSTRAINT $constraintName
+                        UNIQUE
+                        DEFERRABLE INITIALLY DEFERRED
+            )
+            SQL
+        );
+        $this->connection->executeStatement('INSERT INTO gh3423 VALUES (1)');
+
+        self::$tableCreated = true;
+    }
+
+    /**
+     * @group GH3423
+     */
+    public function testTransactionalWithDeferredConstraint(): void
+    {
+        $this->connection->transactional(function (Connection $connection): void {
+            $connection->executeStatement(sprintf('SET CONSTRAINTS "%s" DEFERRED', $this->constraintName));
+            $connection->executeStatement('INSERT INTO gh3423 VALUES (1)');
+
+            $this->expectException(Exception::class);
+            $this->expectExceptionMessage(sprintf('violates unique constraint "%s"', $this->constraintName));
+        });
+    }
+
+    /**
+     * @group GH3423
+     */
+    public function testTransactionalWithDeferredConstraintAndTransactionNesting(): void
+    {
+        $this->connection->setNestTransactionsWithSavepoints(true);
+
+        $this->connection->transactional(function (Connection $connection): void {
+            $connection->executeStatement(sprintf('SET CONSTRAINTS "%s" DEFERRED', $this->constraintName));
+            $connection->beginTransaction();
+            $connection->executeStatement('INSERT INTO gh3423 VALUES (1)');
+            $connection->commit();
+
+            $this->expectException(Exception::class);
+            $this->expectExceptionMessage(sprintf('violates unique constraint "%s"', $this->constraintName));
+        });
+    }
+
+    /**
+     * @group GH3423
+     */
+    public function testCommitWithDeferredConstraint(): void
+    {
+        $this->connection->beginTransaction();
+        $this->connection->executeStatement(sprintf('SET CONSTRAINTS "%s" DEFERRED', $this->constraintName));
+        $this->connection->executeStatement('INSERT INTO gh3423 VALUES (1)');
+
+        $this->expectException(Exception::class);
+        $this->expectExceptionMessage(sprintf('violates unique constraint "%s"', $this->constraintName));
+
+        $this->connection->commit();
+    }
+
+    /**
+     * @group GH3423
+     */
+    public function testCommitWithDeferredConstraintAndTransactionNesting(): void
+    {
+        $this->connection->setNestTransactionsWithSavepoints(true);
+
+        $this->connection->beginTransaction();
+        $this->connection->executeStatement(sprintf('SET CONSTRAINTS "%s" DEFERRED', $this->constraintName));
+        $this->connection->beginTransaction();
+        $this->connection->executeStatement('INSERT INTO gh3423 VALUES (1)');
+        $this->connection->commit();
+
+        $this->expectException(Exception::class);
+        $this->expectExceptionMessage(sprintf('violates unique constraint "%s"', $this->constraintName));
+
+        $this->connection->commit();
+    }
+}


grongor:
Not sure why the continuousphp build failed - I think it's not related to my changes.

Travis's MySQL died for some reason, here is the build of this branch against my fork: https://travis-ci.org/grongor/dbal/builds/475348248

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1172,15 +1172,17 @@ public function transactional(Closure $func)
...
-            return $res;
         } catch (Exception $e) {
             $this->rollBack();
             throw $e;
         } catch (Throwable $e) {
             $this->rollBack();
             throw $e;
         }
+
+        $this->commit();
```
Ocramius: Shouldn't a crash on `commit()` lead to a rollback as well?

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1297,6 +1305,11 @@ public function commit()
             }
         }
 
+        $this->commitDone();
+    }
+
+    private function commitDone()
```
Ocramius: ` : void`

Also: naming suggests that committing finished, yet the code below begins a new transaction now?

Ocramius: Documentation block also needed - or better/more explicative naming.

Ocramius (changes requested):

lib/Doctrine/DBAL/Connection.php
grongor: That is precisely the issue here - when you call `commit()` and it fails, then the transaction is no more :) There is nothing to rollback ... so this PR is all about not calling the rollback if the failure occurs in the `commit()`

Ocramius: > then the transaction is no more

Ok, this is my misconception then. Are we sure that this holds consistently for all platforms that we support?

lib/Doctrine/DBAL/Connection.php
grongor: Oh yeah, I wasn't sure about that part at all :D Do you have any suggestions for the name? It basically just resets the internal state of the object, after the commit.

lib/Doctrine/DBAL/Connection.php
grongor: AFAIK all other platforms throw the exceptions before `commit()` (they do not support constraint deferring), so we will always call rollback in those cases. But that might not be true - I'm not familiar enough with all of them to be perfectly honest. 

Anyway, shouldn't the existing tests cover that?

Ocramius: > shouldn't the existing tests cover that?

If they run on all platforms, yes.

tests/Doctrine/Tests/DBAL/Functional/Ticket/GH3423Test.php
```diff
@@ -0,0 +1,116 @@
...
+class GH3423Test extends DbalFunctionalTestCase
+{
+    /** @var bool */
+    private static $tableCreated = false;
+
+    protected function setUp()
+    {
+        parent::setUp();
+
+        if ($this->connection->getDatabasePlatform()->getName() !== 'postgresql') {
```
Ocramius: Are all tests of this patch related to postgresql only? Which bits could be made DB-agnostic?

grongor: Well as they all target the deferrable constraint then they are only usefull for databases which supports them. From a quick search it looks like Oracle could be targeted as well, if you support it. SQLite, MySQL and SQLServer don't support it them.

Ocramius: OracleDB is supported here, but I was wondering about whether the test could've been made generic. Anyway, no need for it if none of the tests is fully generic: resolving here.

lib/Doctrine/DBAL/Connection.php
grongor: Purpose of this method is to set the `Connection` to the correct state. Until now the code didn't expect the `commit()` to ever throw any exception. Now that it is possible we need to manage the state so that if someone catches the exception, then they can still use the connection afterwards. Without doing that the connection would be stuck with `transactionNestingLevel` set to 1 even though the transaction is long gone. The `->beginTransaction()` is there because of the `autoCommit` thingie - without it the `Connection` object would again be in an unexpected state.

lib/Doctrine/DBAL/Connection.php
grongor: I looked at the functional tests of `Connection` class and it looks like there are no tests covering constraints violations, only basic nesting level checks and whatnot :/ I'll try to write some on Monday.

grongor: @Ocramius I improved the ConnectionTest functional test https://github.com/doctrine/dbal/pull/3425 . Could that be enough? If so, I can rebase this branch onto it...

Ocramius: I'll have to defer that decision to @deeky666 and @morozov

lib/Doctrine/DBAL/Connection.php
Ocramius: Right, and that isn't clear by reading the method: can we somehow improve that?

Consider comments to be the last resort: having a clear method name would be a good improvement at first.

lib/Doctrine/DBAL/Connection.php
SenseException: What about a case where the Exception happens during an open transaction? Shouldn't this improvement be marked as a BC break, when there are no rollbacks anymore?

SenseException (changes requested):

morozov (changes requested):
We need a test which covers the change on all platforms, not only on those which support deferred constraints.

lib/Doctrine/DBAL/Connection.php
grongor: @SenseException I'm sorry, I don't understand. I don't think that I changed the behavior for "common" transactions in any way; please provide an example code if you think so - thanks.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1172,15 +1172,17 @@ public function transactional(Closure $func)
         $this->beginTransaction();
         try {
             $res = $func($this);
-            $this->commit();
-            return $res;
         } catch (Exception $e) {
```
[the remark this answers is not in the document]
grongor: I didn't want to do more changes than was really necessary in this PR, but yeah, I think catching `Throwable` only is a sensible and possible. Should I add the patch to this PR?

tests/Doctrine/Tests/DBAL/Functional/Ticket/GH3423Test.php
```diff
@@ -0,0 +1,117 @@
...
+        $this->connection->exec('INSERT INTO gh3423 VALUES (true)');
+        $this->connection->commit();
+    }
+
+    /**
+     * @group GH3423
+     */
+    public function testCommitWithDeferredConstraintAndTransactionNesting() : void
+    {
+        $this->expectException(DBALException::class);
```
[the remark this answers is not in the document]
grongor: Yeah I usually do it like that but all tests in DBAL seemed to start with it, so I just went with it ... I'll update the code shortly

tests/Doctrine/Tests/DBAL/Functional/Ticket/GH3423Test.php
```diff
@@ -0,0 +1,117 @@
...
+class GH3423Test extends DbalFunctionalTestCase
+{
+    /** @var bool */
+    private static $tableCreated = false;
+
+    protected function setUp()
+    {
+        parent::setUp();
+
+        if (! in_array($this->connection->getDatabasePlatform()->getName(), ['postgresql', 'oracle'], true)) {
```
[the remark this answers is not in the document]
grongor: @morozov Yeah, I understand the fear, trust me :D I don't see another way to do it - how would I test something that the other platforms don't support? I'm not aware of a way to do that ... I think it must be sufficient that these changes don't break the existing tests as those tests should cover all the core/common functionality. And just to be extra sure I adjusted the `ConnectionTest` to let the database throw the errors, you can check it out here https://github.com/doctrine/dbal/pull/3425 .

grongor:
@morozov I think that we do have a test that covers it on all platforms: the functional `ConnectionTest`. I don't see a better way to test this. The new test `GH3423Test` just covers that extra portion that only postgres and oracle support.

lib/Doctrine/DBAL/Connection.php
grongor: @Ocramius added the patch. Does `updateTransactionStateAfterCommit()` make more sense?

tests/Doctrine/Tests/DBAL/Functional/Ticket/GH3423Test.php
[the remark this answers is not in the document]
grongor: Here is a build on top of that improved `ConnectionTest` https://travis-ci.org/grongor/dbal/builds/479358748 - passing

lib/Doctrine/DBAL/Connection.php
SenseException: You fixed your issue "Rollback called when outside of a transaction" in this method by moving the `commit()` call out of the try-catch. Before your change an exception thrown in `commit()` would call a rollback in the catch, but now this isn't the case anymore. What if `commit()` throws an exception inside a transaction and a rollback is needed?

grongor: Thanks, yeah, I understand now.

> What if commit() throws an exception inside a transaction and a rollback is needed?

Then that case wouldn't be handled. The thing is that I was not able to come up with an example where `COMMIT` would throw an exception and also where would need to call `ROLLBACK` afterwards. I asked all devs in my company and no one was able to think of a situation like that. I also asked one guy who knows basically everything around MySQL and he was not able to think of such example (even tried poking master/slave and cluster setups).

But I guess unless someone checks all source codes/documentations of all supported databases we can't be 100 % sure. All I can do is to deploy those changes to our production systems (Postgres and MySQL) and report back if we experienced anything unusual...

Or can you think of a better solution to this problem?

danydev:
@morozov @grongor we recently hit this bug as well. Since it looks like quite a bit of work has been done, do you think the fix is gonna land in doctrine soon? Do you need help in some specific task to get it done?.

tests/Doctrine/Tests/DBAL/Functional/Ticket/GH3423Test.php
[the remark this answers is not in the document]
morozov: > how would I test something that the other platforms don't support? I'm not aware of a way to do that ...

Except for not being able to create a deferred constraint, how is the new test expected to behave on a platform that doesn't support deferred constraint? I.e. instead of skipping the test entirely, can we just create the deferred constraint if the platform supports it, and otherwise create a regular one?

> I think it must be sufficient that these changes don't break the existing tests as those tests should cover all the core/common functionality.

Could you identify these tests?

tests/Doctrine/Tests/DBAL/Functional/Ticket/GH3423Test.php
```diff
@@ -0,0 +1,119 @@
...
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\DBALException;
+use Doctrine\Tests\DbalFunctionalTestCase;
+use function in_array;
+
+/**
+ * @see https://github.com/doctrine/dbal/issues/3423
+ */
+class GH3423Test extends DbalFunctionalTestCase
```
morozov, line 15: Please avoid using issue numbers in tests. It should be named after the feature/behavior being tested.

morozov:
> […] we recently hit this bug as well. Since it looks like quite a bit of work has been done, do you think the fix is gonna land in doctrine soon?

@danydev, if fixing this bug doesn't require breaking changes, this PR may be retargeted against `2.10.x`. This way, it could be released soon after it's resolved.

At the moment, there's no enough confidence that changing the code at the abstraction level and testing it only with certain platforms is sufficient.

> Do you need help in some specific task to get it done?

You can help by answering my questions above.

grongor:
So ... I tried rebasing it today (4.x), and just running the tests against the current implementation. Here is the build: https://github.com/grongor/dbal/actions/runs/646044144

Postgres is broken: the constraint violation is completely throw away and replaced with "There is no active transaction".
Oracle passes the test, as it includes the information about the constraint violation in the exception throw when rolling back the transaction. You can see it here https://github.com/grongor/dbal/runs/2094504708?check_suite_focus=true#step:6:70 (where I yet didn't handle the OCI warning). I personally think that the rollback here isn't correct either, but I'm not familiar with Oracle and I don't intend to be :D 

So ... do you want to continue this somehow, or should we close? I checked the tests around the Connection class and it seems they are a bit too synthetic. We could add a test that would check the constraint violation (without deferring), and if the new implementation won't break it, I think it's safe to say it will fix this issue without causing another. Would that be enough for this to get merged? If so, I'll update the PR.

ps: Sorry for the "late response", it cost me a lot of time and I got tired at some point. We also stopped using the deferred constraints so it wasn't a priority for me anymore...

morozov:
> I personally think that the rollback here isn't correct either, but I'm not familiar with Oracle and I don't intend to be :D

That's the problem. The code being modified works for all platforms.

> So ... do you want to continue this somehow, or should we close?

What do _you_ want? If you want to close it, feel free to. If you want to proceed, you'll need to gather more information. Likely, not all platforms will behave the same, and the code change will become more complex. E.g. [here](https://stackoverflow.com/a/24321060/146187) (just as a reference) it's stated that a rollback is needed. And [here](https://developer.jboss.org/thread/261507) is an example of using `tx.getStatus()` to observe the transaction status to make the call. What could help:
1. Research the documentation on all supported platforms and see if a rollback is needed after a failed commit.
2. Find a way to observe the status of a transaction programmatically and write tests.
3. Research other projects like JDBC and see how it's done there.

> I checked the tests around the Connection class and it seems they are a bit too synthetic.

What do you mean?

grongor:
> That's the problem. The code being modified works for all platforms.

No, it does not. See the link I provided. It doesn't work for the deferred constraints.

> E.g. here (just as a reference) it's stated that a rollback is needed

Nope. To quote exactly: `either explicitly with a COMMIT or ROLLBACK statement`. And my proposed change does exactly that. Your current implementation does both, which is what causes the unexpected errors (again, please check the links to the build I provided in the comment above).

> What do you mean?

There isn't a single test that would check how the transaction behaves when there is an error, like for example violation of a constraint. By an error, I mean something that is triggered by the database, not something we throw explicitly in the test.

> What do _you_ want?

Nothing really. I just didn't want to be that guy that just closes the issue without giving it at least one last try. I've given this issue (and the one in ORM package) a lot of time already, and I only met with resistance, which to be honest I understand and think is warranted. I just expected someone to meet me half-way. I'm not gonna spend another 20 hours just to learn that "we aren't quite confident yet" - I hope you can understand that.

---------

Maybe there is another way to approach this. I wanted to remove the `rollback` call because in the case of deferred statements it is _probably_ redundant (and causes exception), but maybe we could try to put another `try-catch` around the `rollback` statement. If there is an exception, we can ignore it and re-throw the "original" exception that caused the `rollback` in the first place. That way we wouldn't have to change how the `commit`/`rollback` is handled, and we would also solve the issue with deferred constraints, where the actual error is hidden away by the "there is no transaction" error caused by the `rollback`.

danydev:
It's kind of hard to guarantee that a rollback is not needed after a failed commit in all vendors. For example sqlite may return a SQLITE_BUSY exception, and after that you still need to rollback.

I think your second proposal may be the way to sort this out. Maybe It's not the cleanest, but it's the safer.

grongor:
Yeah I agree. To be honest I am quite surprised I didn't come up with it sooner lol, now it seems like an obvious way to do it.

simPod:
@morozov what branch should this target?

morozov:
I've retargeted to `3.1.x`.

PR #3426, opened 2019-01-08
Set wait_timeout for MySQL in travis tests to prevent timeouts

grongor:
Also remove redundant bootstrap option in mariadb tests config

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | none

#### Summary

This patch should fix issues like this one https://travis-ci.org/doctrine/dbal/jobs/476488849 . After a quick research this seems like a solution to these kinds of errors on Travis CI.


Ocramius (approved):

SenseException (approved):

Majkl578:
This should be already fixed by #3473 using different approach, if not, we can reopen.

PR #3428, opened 2019-01-10
Update OCI8Statement.php

ToGo101:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

Converting ParameterType::Integer to SQLT_INT

Requires for "insert into xxx returning id into :id" statements
otherwise a numeric id will be casted to string.

Example: Saved entity will genertate an ID from a sequence, lets say 1234567890

Before this change this will return a '1' 
After this chang this will return 1234567890

See: https://stackoverflow.com/questions/1710617/oci-bind-by-name-returning-into-truncates-value


Ocramius:
Requires a test case

PR #3432, opened 2019-01-20
Updated PHPStan to v0.11

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1227,9 +1227,7 @@ class Connection implements DriverConnection
     }
 
     /**
-     * Starts a transaction by suspending auto-commit mode.
-     *
-     * @return void
+     * {@inheritDoc}
      */
     public function beginTransaction()
     {
@@ -1256,12 +1254,12 @@ class Connection implements DriverConnection
                 $logger->stopQuery();
             }
         }
+
+        return true;
     }
 
     /**
-     * Commits the current transaction.
-     *
-     * @return void
+     * {@inheritDoc}
      *
      * @throws ConnectionException If the commit failed due to no active transaction or
      *                                            because the transaction was marked for rollback only.
@@ -1300,10 +1298,12 @@ class Connection implements DriverConnection
         --$this->transactionNestingLevel;
 
         if ($this->autoCommit !== false || $this->transactionNestingLevel !== 0) {
-            return;
+            return true;
         }
 
         $this->beginTransaction();
+
+        return true;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


lcobucci (approved):

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1256,12 +1254,12 @@ public function beginTransaction()
...
         }
+
+        return true;
     }
 
     /**
-     * Commits the current transaction.
-     *
-     * @return void
+     * {@inheritDoc}
```
Majkl578: We should get rid of this in future - inheritance is implicit.

morozov: What do you mean? It explicitly implements the driver connection interface and therefore inherits its API. Are you talking about not implementing the driver connection interface in the wrapper connection in the future?

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -368,7 +368,7 @@ private function getNonAutoincrementPrimaryKeyDefinition(array $columns, array $
         $keyColumns = array_unique(array_values($options['primary']));
 
         foreach ($keyColumns as $keyColumn) {
-            if (isset($columns[$keyColumn]['autoincrement']) && ! empty($columns[$keyColumn]['autoincrement'])) {
+            if (! empty($columns[$keyColumn]['autoincrement'])) {
```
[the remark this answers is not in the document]
morozov: We could use a `Schema\Column` instance here and call `getAutoincrement()`. Not sure if we can get of `empty()` while having `mixed[][]` everywhere.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -327,28 +328,23 @@ public function fetch($fetchMode = null, $cursorOrientation = PDO::FETCH_ORI_NEX
...
         }
 
-        switch ($fetchMode) {
-            case FetchMode::NUMERIC:
-                return $values;
+        if ($fetchMode === FetchMode::NUMERIC) {
+            return $values;
+        }
 
+        assert(is_array($this->_columnNames));
```
morozov: This is a good example (one of many) which shows that currently the `Statement` and `Result` concerns are intermixed in the same API just because of many underlying drivers do so. Column names belong to the result, which should be explicitly returned (not stored as a state) by the statement.

lib/Doctrine/DBAL/Connection.php
Majkl578: I mean that `{@inheritDoc}` is redundant - inheritance is implicit and this shouldn't be needed. But we need to first address this on CS level, I also discussed this with @kukulich somewhere.

morozov: Yeah, this is true. It shouldn't be needed according to [phpDocumentor](http://docs.phpdoc.org/guides/inheritance.html#methods). Looks like PhpStorm doesn't agree with that ([WI-38168](https://youtrack.jetbrains.com/issue/WI-38168)).

PR #3436, opened 2019-01-23
PHPStan Level 5

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

On level 5, things are getting a bit awkward. Fixing some issues the right way would cause a BC break.

1. The `rowCount()` method logically belongs to `ResultStatement`, however, it's declared in the `Statement` class. This issue has manifested in `ResultCacheStatement`.
2. The same should be applicable to `errorCode()` and `errorInfo()`. The error is a _result_ of statement execution, not an intent to execute it. See `Portability\Statement`.
3. Additionally, some methods of `Portability\Statement` requires a `Statement` while the others require a `ResultStatement`. Probably it could be split into two classes and represent two interfaces independently.
4. Type-wise, `SQLParserUtils::getPlaceholderPositions()` should be split into two methods for the statements with positional and named parameters.
5. PHPStan doesn't seem to understand property annotations like `SplObjectStorage|T[]` or `SplObjectStorage<T>`. That's why the annotations are added inline. See https://github.com/phpstan/phpstan/issues/1838.
6. `sha1(mt_rand())` has been removed from the `Graphviz` visitor since the object name is optional and it cannot be referenced from outside.

### DIFF
--- a/lib/Doctrine/DBAL/Event/SchemaCreateTableEventArgs.php
+++ b/lib/Doctrine/DBAL/Event/SchemaCreateTableEventArgs.php
@@ -3,7 +3,6 @@
 namespace Doctrine\DBAL\Event;
 
 use Doctrine\DBAL\Platforms\AbstractPlatform;
-use Doctrine\DBAL\Schema\Column;
 use Doctrine\DBAL\Schema\Table;
 use function array_merge;
 use function is_array;
@@ -16,7 +15,7 @@ class SchemaCreateTableEventArgs extends SchemaEventArgs
     /** @var Table */
     private $table;
 
-    /** @var Column[] */
+    /** @var mixed[][] */
     private $columns;
 
     /** @var mixed[] */
@@ -29,8 +28,8 @@ class SchemaCreateTableEventArgs extends SchemaEventArgs
     private $sql = [];
 
     /**
-     * @param Column[] $columns
-     * @param mixed[]  $options
+     * @param mixed[][] $columns
+     * @param mixed[]   $options
      */
     public function __construct(Table $table, array $columns, array $options, AbstractPlatform $platform)
     {
@@ -49,7 +48,7 @@ class SchemaCreateTableEventArgs extends SchemaEventArgs
     }
 
     /**
-     * @return Column[]
+     * @return mixed[][]
      */
     public function getColumns()
     {
--- a/lib/Doctrine/DBAL/Schema/Visitor/DropSchemaSqlCollector.php
+++ b/lib/Doctrine/DBAL/Schema/Visitor/DropSchemaSqlCollector.php
@@ -78,15 +78,18 @@ class DropSchemaSqlCollector extends AbstractVisitor
     {
         $sql = [];
 
+        /** @var ForeignKeyConstraint $fkConstraint */
         foreach ($this->constraints as $fkConstraint) {
             $localTable = $this->constraints[$fkConstraint];
             $sql[]      = $this->platform->getDropForeignKeySQL($fkConstraint, $localTable);
         }
 
+        /** @var Sequence $sequence */
         foreach ($this->sequences as $sequence) {
             $sql[] = $this->platform->getDropSequenceSQL($sequence);
         }
 
+        /** @var Table $table */
         foreach ($this->tables as $table) {
             $sql[] = $this->platform->getDropTableSQL($table);
         }
--- a/lib/Doctrine/DBAL/Schema/Visitor/Graphviz.php
+++ b/lib/Doctrine/DBAL/Schema/Visitor/Graphviz.php
@@ -8,8 +8,6 @@ use Doctrine\DBAL\Schema\Table;
 use function current;
 use function file_put_contents;
 use function in_array;
-use function mt_rand;
-use function sha1;
 use function strtolower;
 
 /**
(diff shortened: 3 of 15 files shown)


lib/Doctrine/DBAL/Event/SchemaCreateTableEventArgs.php
```diff
@@ -49,7 +48,7 @@ public function getTable()
     }
 
     /**
-     * @return Column[]
+     * @return mixed[][]
```
Ocramius: This change kinda hurts... Do we have a way to upcast the `$columns` to `Column[]`?

lib/Doctrine/DBAL/Schema/Visitor/DropSchemaSqlCollector.php
```diff
@@ -78,15 +78,18 @@ public function getQueries()
     {
         $sql = [];
 
+        /** @var ForeignKeyConstraint $fkConstraint */
```
Ocramius: Would it make sense to document these on the iterators above?

lib/Doctrine/DBAL/Schema/Visitor/DropSchemaSqlCollector.php
```diff
@@ -78,15 +78,18 @@ public function getQueries()
...
         foreach ($this->constraints as $fkConstraint) {
             $localTable = $this->constraints[$fkConstraint];
             $sql[]      = $this->platform->getDropForeignKeySQL($fkConstraint, $localTable);
         }
 
+        /** @var Sequence $sequence */
```
Ocramius: Would it make sense to document these on the iterators above?

lib/Doctrine/DBAL/Schema/Visitor/DropSchemaSqlCollector.php
```diff
@@ -78,15 +78,18 @@ public function getQueries()
...
         foreach ($this->sequences as $sequence) {
             $sql[] = $this->platform->getDropSequenceSQL($sequence);
         }
 
+        /** @var Table $table */
```
Ocramius: Would it make sense to document these on the iterators above?

lib/Doctrine/DBAL/Schema/Visitor/Graphviz.php
```diff
@@ -41,7 +39,7 @@ public function acceptForeignKey(Table $localTable, ForeignKeyConstraint $fkCons
      */
     public function acceptSchema(Schema $schema)
     {
-        $this->output  = 'digraph "' . sha1(mt_rand()) . '" {' . "\n";
```
Ocramius: Should we use the schema FQN for this instead?

Ocramius (changes requested):

lib/Doctrine/DBAL/Event/SchemaCreateTableEventArgs.php
morozov: It never returned `Column[]`, there shouldn't be existing clients which would rely on that.

lib/Doctrine/DBAL/Schema/Visitor/DropSchemaSqlCollector.php
morozov: See the description.

lib/Doctrine/DBAL/Schema/Visitor/Graphviz.php
morozov: Not sure what the code would look like. Getting the FQN requires some default namespace.

Besides that, I was unable to generate the chart from the output of the visitor:
```php
$schema = new Schema();
$table  = $schema->createTable('test');
$table->addColumn('foo', 'integer');
$table->addColumn('bar', 'string');

$visitor = new Graphviz();
$visitor->acceptSchema($schema);
$visitor->write('php://stdout');
```
produces the output which doesn't contain any relevant information:
```
digraph {
splines = true;
overlap = false;
outputorder=edgesfirst;
mindist = 0.6;
sep = .2;
}
```

morozov: Could you provide a snippet for getting the FQN for the chart?

Ocramius: Assuming DB "foo" and schema "bar"

`diagraph "foo.bar" {`

or just schema "foo":

`diagraph "foo" {`

lib/Doctrine/DBAL/Schema/Visitor/DropSchemaSqlCollector.php
morozov: Not as of now. See the description.

lib/Doctrine/DBAL/Schema/Visitor/DropSchemaSqlCollector.php
morozov: Not as of now. See the description.

lib/Doctrine/DBAL/Schema/Visitor/Graphviz.php
Ocramius: Let's go with an anonymous `diagraph` block :+1:

Ocramius (approved):

PR #3435, opened 2019-01-22
PHPStan Level 4

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

PR #3439, opened 2019-01-24
Fixed QueryBuilder type hints

petrnemec:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | N/A

#### Summary

Fixed type hints for `$firstResult` and `$maxResults` properties

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -107,14 +107,14 @@ class QueryBuilder
     /**
      * The index of the first result to retrieve.
      *
-     * @var int
+     * @var int|null
      */
     private $firstResult = null;
 
     /**
      * The maximum number of results to retrieve.
      *
-     * @var int
+     * @var int|null
      */
     private $maxResults = null;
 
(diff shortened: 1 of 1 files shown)


morozov:
@petrnemec thank you for the patch. I'm currently working on making the codebase compliant with PHPStan Level 6 and 7 requirements (see the previous changes in #3435 and #3436). Most likely, this will be taken care of by the upcoming changes.

I'm not going to accept these changes right now since they are not enforced by raising PHPStan level requirement.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -107,14 +107,14 @@ class QueryBuilder
     /**
      * The index of the first result to retrieve.
      *
-     * @var int
+     * @var int|null
```
morozov: It's not supposed to be nullable. The "non specified" first result is the same as `0`.

petrnemec: Default value of property this is `null`, init value is not set in constructor, so it should be marked as nullable.

morozov: The default value should be `0`, not `null` unless it breaks some existing API contract.

PR #3440, opened 2019-01-24
Add test for ignoring default value for MySQL TEXT and BLOB types

JarJak:
See #3189


SenseException:
Hi @JarJak and thank you for this contribution. Is this supposed to be the test for #3189?

JarJak:
@SenseException Yes. I've added it as separate PR just to run CI and check if it fails.

PR #3442, opened 2019-01-27
PHPStan Level 6

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

### DIFF
--- a/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
+++ b/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
@@ -19,6 +19,7 @@ use function feof;
 use function fread;
 use function get_resource_type;
 use function is_array;
+use function is_int;
 use function is_resource;
 use function sprintf;
 use function str_repeat;
@@ -41,7 +42,7 @@ class MysqliStatement implements IteratorAggregate, Statement
     /** @var mysqli_stmt */
     protected $_stmt;
 
-    /** @var string[]|bool|null */
+    /** @var string[]|false|null */
     protected $_columnNames;
 
     /** @var mixed[]|null */
@@ -78,10 +79,14 @@ class MysqliStatement implements IteratorAggregate, Statement
     public function __construct(mysqli $conn, $prepareString)
     {
         $this->_conn = $conn;
-        $this->_stmt = $conn->prepare($prepareString);
-        if ($this->_stmt === false) {
+
+        $stmt = $conn->prepare($prepareString);
+
+        if ($stmt === false) {
             throw new MysqliException($this->_conn->error, $this->_conn->sqlstate, $this->_conn->errno);
         }
+
+        $this->_stmt = $stmt;
 
         $paramCount = $this->_stmt->param_count;
         if (0 >= $paramCount) {
@@ -97,14 +102,14 @@ class MysqliStatement implements IteratorAggregate, Statement
      */
     public function bindParam($column, &$variable, $type = ParameterType::STRING, $length = null)
     {
+        assert(is_int($column));
+
         if (! isset(self::$_paramTypeMap[$type])) {
             throw new MysqliException(sprintf("Unknown type: '%s'", $type));
         }
 
-        $type = self::$_paramTypeMap[$type];
-
         $this->_bindedValues[$column] =& $variable;
-        $this->types[$column - 1]     = $type;
+        $this->types[$column - 1]     = self::$_paramTypeMap[$type];
 
         return true;
     }
@@ -114,15 +119,15 @@ class MysqliStatement implements IteratorAggregate, Statement
      */
     public function bindValue($param, $value, $type = ParameterType::STRING)
     {
+        assert(is_int($param));
+
         if (! isset(self::$_paramTypeMap[$type])) {
             throw new MysqliException(sprintf("Unknown type: '%s'", $type));
         }
 
-        $type = self::$_paramTypeMap[$type];
-
         $this->_values[$param]       = $value;
         $this->_bindedValues[$param] =& $this->_values[$param];
-        $this->types[$param - 1]     = $type;
+        $this->types[$param - 1]     = self::$_paramTypeMap[$type];
 
         return true;
     }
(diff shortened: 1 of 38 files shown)


lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -41,7 +42,7 @@ class MysqliStatement implements IteratorAggregate, Statement
     /** @var mysqli_stmt */
     protected $_stmt;
 
-    /** @var string[]|bool|null */
+    /** @var string[]|false|null */
```
Ocramius: `false` is not really a type in PHP - we should probably stick to `bool` here

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -97,14 +102,14 @@ public function __construct(mysqli $conn, $prepareString)
      */
     public function bindParam($column, &$variable, $type = ParameterType::STRING, $length = null)
     {
+        assert(is_int($column));
```
Ocramius: Wondering if we should run CI with `dzend_assertions` enabled and throwing

lib/Doctrine/DBAL/Driver/OCI8/OCI8Connection.php
```diff
@@ -71,17 +73,23 @@ public function __construct($username, $password, $db, $charset = null, $session
      */
     public function getServerVersion()
     {
-        if (! preg_match('/\s+(\d+\.\d+\.\d+\.\d+\.\d+)\s+/', oci_server_version($this->dbh), $version)) {
+        $version = @oci_server_version($this->dbh);
```
Ocramius: We should most likely avoid introducing `@` where not 100000% sure if required. Yes, the DB2 extension is horrible, but what can fail here?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -360,7 +376,7 @@ public function execute($params = null)
             $hasZeroIndex = array_key_exists(0, $params);
             foreach ($params as $key => $val) {
                 if ($hasZeroIndex && is_numeric($key)) {
-                    $this->bindValue($key + 1, $val);
+                    $this->bindValue((int) $key + 1, $val);
```
Ocramius: Maybe we should document `$params` as `array<int, mixed>` instead?

Ocramius: This should apply to any cast, by the way: casting leads to silent failures in way too many scenarios

lib/Doctrine/DBAL/Driver/PDOConnection.php
```diff
@@ -69,23 +69,13 @@ public function prepare($prepareString, $driverOptions = [])
      */
     public function query()
```
Ocramius: What happens if we declare a variadic argument here, btw? Would that break inheritance in any way?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -315,6 +315,10 @@ protected function _getCreateTableSQL($tableName, array $columns, array $options
      */
     public function getCreatePrimaryKeySQL(Index $index, $table)
     {
+        if ($table instanceof Table) {
+            $table = $table->getQuotedName($this);
```
Ocramius: Should probably use a different variable name, so that we don't end up with reasoning about a union type at all

Ocramius (changes requested):

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
morozov: > This should apply to any cast, by the way: casting leads to silent failures in way too many scenarios

Hm… in this particular case, we're casting to an integer something which `is_numeric()`, and we're using it only because there's no `is_integerish()` in PHP. Pracically, I can't see any problem here.

lib/Doctrine/DBAL/Driver/PDOConnection.php
morozov: Yes:
> Fatal error: Declaration of Doctrine\DBAL\Driver\MyConnection::query() must be compatible with Doctrine\DBAL\Driver\PDOConnection::query(...$args)

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
Ocramius: I'd surely get rid of the cast though - even asserting would be better.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
morozov: According to the [documentation](http://php.net/manual/en/function.assert.php#function.assert.expectations), `zend.assertions` are enabled by default (development mode) and according to the internet, Travis CI doesn't disable them. As for `assert.exception` regardless of the configuration, PHPUnit will handle a warning or a `Throwable` and turn into an error. So unless we want to be paranoid, the existing configuration is good enough.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
morozov: Actually, I'm even sure which assertion would help here. In fact, it looks like a false-positive on the PHPStan end (see the [example](https://phpstan.org/r/2edaabe7-40cb-4a94-a312-d6024d9a55b0)). Adding a number to a numeric string won't result in an error.

But I think we can replace `is_numeric()` with `is_int()`: since the value comes from an array key, the only numeric but non-integer value would be something like `'1.5'` which is not a valid positional or named parameter anyways.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
Ocramius: :+1:

Ocramius (approved):

PR #3445, opened 2019-01-31
Change type from TEXT to BYTEA for storing objects in PostgreSQL

NeoFusion:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | doctrine/orm#4029

#### Summary
Change type from `TEXT` to `BYTEA` on `ObjectType`, because PostgreSQL itself does not allow null byte (`\0` adds by `serialize`) on `TEXT` fields.


morozov:
@NeoFusion this change requires a test case. If it ends up being BC-breaking, it will have to be retargeted against `develop` instead of `master`.

NeoFusion:
I added functional test, but there was an error with DB2 (`db2_bind_param(): Describe Param Failed`)

tests/Doctrine/Tests/DBAL/Functional/Types/ObjectTest.php
```diff
@@ -0,0 +1,57 @@
...
+
+        $result = $this->connection->insert('object_table', [
+            'id'    => $id,
+            'value' => $value,
+        ]);
+
+        $this->assertSame(1, $result);
+    }
+
+    private function select(int $id)
```
Ocramius: Please add return types where applicable

tests/Doctrine/Tests/DBAL/Functional/Types/ObjectTest.php
```diff
@@ -0,0 +1,57 @@
...
+        $object1->setData('Test');
+
+        $this->insert(1, $object1);
+        $resultObject1 = $this->select(1);
+
+        $this->assertInstanceOf(SimpleClass::class, $resultObject1);
+        $this->assertSame('Test', $resultObject1->getData());
+    }
+
+    private function insert(int $id, object $object)
```
Ocramius: Please add `: void`

Ocramius (changes requested):
Requires adjustments as per code style (see build failures)

tests/Doctrine/Tests/DBAL/Functional/Types/ObjectTest.php
NeoFusion: mixed

tests/Doctrine/Tests/DBAL/Functional/Types/ObjectTest.php
NeoFusion: done

lib/Doctrine/DBAL/Types/ObjectType.php
```diff
@@ -20,15 +20,15 @@ class ObjectType extends Type
      */
     public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $platform)
     {
-        return $platform->getClobTypeDeclarationSQL($fieldDeclaration);
+        return $platform->getObjectTypeDeclarationSQL($fieldDeclaration);
```
morozov: Instead of introducing new Platform methods, could the object type use BLOB column instead? Otherwise, even if they me binary-safe on some platforms, CLOB fields are subject to character set conversion and therefore are unsafe to store binary data.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -1273,4 +1282,12 @@ private function getOldColumnComment(ColumnDiff $columnDiff) : ?string
     {
         return $columnDiff->fromColumn ? $this->getColumnComment($columnDiff->fromColumn) : null;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function escapeStringForObject($data)
+    {
+        return pg_escape_bytea($data);
```
morozov: As far as I can tell, this function should be used to represent the data as a string literal in SQL. It's not needed when used as a parameter for a prepared statement.

morozov (changes requested):
Please map the `Object` type to `BLOB` instead. It should be safe as per [existing](https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/Functional/StatementTest.php#L106) [tests](https://github.com/doctrine/dbal/blob/master/tests/Doctrine/Tests/DBAL/Functional/Types/BinaryTest.php#L41).

morozov:
> @morozov this is probably a BC break for most downstream consumers anyway, as they'd get an `ALTER TABLE ...` for each field (in ORM) mapped as `object`. Still, it is a reasonable migration, and the schema changes can be ignored.

This way, I wouldn't consider it a BC-breaking. The mapping of PHP types to DB columns is an internal implementation detail of the DBAL which consumers should not rely on.

morozov:
Before proceeding, please rebase in order to make sure:

1. The test is compatible with PHPUnit 8.
2. PostgreSQL 11 jobs are no longer failing.

NeoFusion:
There is new PR based on this duscussion: doctrine/dbal/pull/3453

Issue #3449, opened 2019-02-06
Run test suite with executionOrder="random"


Majkl578:
Triggered by https://twitter.com/phpdudeaz/status/1093176110047584257.

My main concern with this approach is determinism. I know it's the sole point of this idea, but debugging these failures locally (or getting them randomly in PRs) is terrifying.
Therefore it may be a good idea to run tests randomly only when `if: NOT type = pr`.

I'm not sure whether PHPUnit provides _seed_ used for the execution that could later be used for reproducible failure. If not, adding such feature into PHPUnit may be a pre-requisite.

Edit: PHPUnit does provide a way to set the seed, as well as printing it as part of output, but only in verbose mode (https://github.com/sebastianbergmann/phpunit/issues/3515).

morozov:
In general, I agree that the random order should be only used on release branches, not in PRs. A contributor of a given change doesn't have to solve unrelated issues caused by hidden dependencies in existing tests.

On the other hand, a new PR should be prevented from introducing those dependencies. Therefore, once the tests get stable enough on feature branches, we can try enabling random order on PRs as well.

morozov:
I took an initial stab at it in https://github.com/doctrine/dbal/pull/4731. Since there doesn't seem to be a way to make failures optional on GitHub, running tests in random order on CI doesn't seem feasible.

PR #3450, opened 2019-02-06
[WIP] Run test suite with executionOrder="random"

Majkl578:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #3449

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "ddba260b7dfb84d2591b0a71a885166a",
+    "content-hash": "2d55f7a8d33401da4b8b6415a96ae022",
     "packages": [
         {
             "name": "doctrine/cache",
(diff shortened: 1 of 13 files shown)


Ocramius (changes requested):
The fact that all builds went completely haywire is quite telling: I'd love to have this merged though.

Is there any way to make sure that the random seed is displayed in the build? That's a real blocker.

Majkl578:
@Ocramius Yes, see https://github.com/sebastianbergmann/phpunit/issues/3515 + https://github.com/sebastianbergmann/phpunit/pull/3524 - the seed will be displayed by default in the next PHPUnit release. :+1:

composer.lock
```diff
@@ -650,7 +652,7 @@
                     "homepage": "https://nette.org/contributors"
                 }
             ],
-            "description": "💎 Nette Dependency Injection Container: Flexible, compiled and full-featured DIC with perfectly usable autowiring and support for all new PHP 7.1 features.",
+            "description": "? Nette Dependency Injection Container: Flexible, compiled and full-featured DIC with perfectly usable autowiring and support for all new PHP 7.1 features.",
```
Majkl578: This looks like a composer bug on Windows.

morozov: Did you run the update on Windows? 😱

Majkl578: Yes. 😹 Composer should be portable, right? 😅

morozov:
While we are fixing the failures, let's try to document what the end result should like from the CI perspective. IIRC, we considered running PR tests with a fixed seed in order to avoid irrelevant failures. The question is, what if a PR introduces an implicit dependency? Should cron-scheduled builds and the push ones have the same order settings? Or we just go wild and enable the random order everywhere?

composer.lock
morozov: Apparently, it's not only a Windows issue. I'm getting the same consistently on one of my two Linux machines (see #3604). Did you have a chance to look deeper or maybe report an issue? I haven't found anything related.

morozov:
Closing in favor of https://github.com/doctrine/dbal/pull/4731. There are no more known failures in failures in random mode but there doesn't seem to be a way to allow build failures on GitHub Actions for a certain configuration.

PR #3454, opened 2019-02-09
CI: Test against PHP 7.4snapshot instead of nightly (8.0)

Majkl578:
|      Q       |   A
|------------- | -----------
| Type         | improvement

#### Summary

Since PHP 7.4 was branched off recently, `nightly` is now PHP 8.0-dev and PHP 7.4 is not tested against.
This way we get rid of unreachable PHP 8.0 target and get the code tested against PHP 7.4-dev instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ after_script:
 
 jobs:
   allow_failures:
-    - php: nightly
+    - php: 7.4snapshot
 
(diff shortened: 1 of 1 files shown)


Ocramius (approved):
:+1: if/when the build passes :D

morozov (approved):

PR #3455, opened 2019-02-11
Use variadic instead of func_get_args

danizord:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

So that it can be proxied with ProxyManager.
See https://github.com/Ocramius/ProxyManager/issues/269

### DIFF
--- a/lib/Doctrine/DBAL/Driver/PDOConnection.php
+++ b/lib/Doctrine/DBAL/Driver/PDOConnection.php
@@ -5,7 +5,6 @@ namespace Doctrine\DBAL\Driver;
 use Doctrine\DBAL\ParameterType;
 use PDO;
 use function assert;
-use function func_get_args;
 
 /**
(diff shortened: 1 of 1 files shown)


danizord:
nvm, it would require a BC breaking change in `Connection` interface :/

danizord:
Actually, it seems like PHP allows to add variadic arguments to the implementation even if the interface does not require any argument. So I'm reopening :)

Majkl578:
This is still a BC break since PDOConnection is not final. However it may be fine for 3.x if you rebase & retarget to develop.

morozov:
Closing this since it's not acceptable for `master` and is already solved in `develop` (#3080).

gggeek:
It seems that this is still an issue when using ProxyManager 2.8.1 with DBAL 2.11.1, even though in that release Connection is not extending PDO any more...

Issue #3427, opened 2019-01-09
Regression in 2.9 constantly outputting table differences


cs278:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.9.*

#### Summary

I've tried to upgrade a project to DBAL 2.9.2 from 2.8.1, however the Doctrine bundle schema diff command constantly outputs schema changes. This is triggered as we've overloaded the default `datetime` type with `datetime_immutable`.

#### Current behaviour

Schema diff constantly generates the following query:

```sql
ALTER TABLE test CHANGE dt dt DATETIME NOT NULL COMMENT '(DC2Type:datetime_immutable)'
```

#### How to reproduce

Given schema:

```sql
CREATE TABLE test (dt DATETIME NOT NULL COMMENT '(DC2Type:datetime_immutable)') DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB;
```

Running the following, will exit 0 on 2.8.1 and error on 2.9.2.

```php
<?php

use Doctrine\DBAL\Schema\Comparator;
use Doctrine\DBAL\Schema\Schema;
use Doctrine\DBAL\Types\Type;
use Doctrine\DBAL\DriverManager;

require 'vendor/autoload.php';

Type::overrideType('datetime', \Doctrine\DBAL\Types\DateTimeImmutableType::class);

$conn = DriverManager::getConnection([
    'url' => 'mysql://root@127.0.0.1:3306/main',
]);

$conn->connect();

$platform = $conn->getDatabasePlatform();
$schemaManager = $conn->getSchemaManager();

$schema1 = $schemaManager->createSchema();

$schema2 = new Schema();
$table2 = $schema2->createTable('test');
$table2->addColumn('dt', 'datetime');
$diff = Comparator::compareSchemas($schema1, $schema2)->toSql($platform);

var_dump($diff);

exit($diff === [] ? 0 : 1);
```

#### Expected behaviour

No update queries should be generated.

#### Patch

Cause: https://github.com/doctrine/dbal/commit/dca1465bdc0836c71ed97e46820490127ccc89ab#diff-6a252b2e905db4e03fc6a121a41e6466L415

This fixes the problem, not entirely sure if this is the correct resolution though.

```diff
diff --git a/lib/Doctrine/DBAL/Schema/Comparator.php b/lib/Doctrine/DBAL/Schema/Comparator.php
index 8297e05..33ee482 100644
[...]

PR #3456, opened 2019-02-12
Compare type class when comparing columns.

garret-gunter:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3427 

#### Summary
Comparing columns that have different instances of the same type no longer detects a difference in type.

I'm a little hesitant to override a type in a test, but I don't know of a better way to get two instances of a type.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Comparator.php
+++ b/lib/Doctrine/DBAL/Schema/Comparator.php
@@ -12,6 +12,7 @@ use function array_shift;
 use function array_unique;
 use function assert;
 use function count;
+use function get_class;
 use function strtolower;
 
 /**
@@ -421,7 +422,11 @@ class Comparator
 
         $changedProperties = [];
 
-        foreach (['type', 'notnull', 'unsigned', 'autoincrement'] as $property) {
+        if (get_class($properties1['type']) !== get_class($properties2['type'])) {
+            $changedProperties[] = 'type';
+        }
+
+        foreach (['notnull', 'unsigned', 'autoincrement'] as $property) {
             if ($properties1[$property] === $properties2[$property]) {
                 continue;
             }
--- a/tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
+++ b/tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
@@ -16,6 +16,7 @@ use Doctrine\DBAL\Schema\TableDiff;
 use Doctrine\DBAL\Types\Type;
 use PHPUnit\Framework\TestCase;
 use function array_keys;
+use function get_class;
 
 class ComparatorTest extends TestCase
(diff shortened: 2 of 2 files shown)


garret-gunter:
Fixed conflicts.

lib/Doctrine/DBAL/Schema/Comparator.php
```diff
@@ -421,7 +422,11 @@ public function diffColumn(Column $column1, Column $column2)
 
         $changedProperties = [];
 
-        foreach (['type', 'notnull', 'unsigned', 'autoincrement'] as $property) {
+        if (get_class($properties1['type']) !== get_class($properties2['type'])) {
```
morozov: How does it work with non-object types? Isn't it supposed to throw some warning?
```sh
$ php -r "var_dump(get_class('string'));"

Warning: get_class() expects parameter 1 to be object, string given in Command line code on line 1

Call Stack:
    0.0001     466744   1. {main}() Command line code:0

Command line code:1:
bool(false)
```
Also, will it work correctly with two different non-object types?
```sh
$ php -r "var_dump(get_class('string') === get_class('integer'));"

Warning: get_class() expects parameter 1 to be object, string given in Command line code on line 1

Call Stack:
    0.0001     466880   1. {main}() Command line code:0


Warning: get_class() expects parameter 1 to be object, string given in Command line code on line 1

Call Stack:
    0.0001     466880   1. {main}() Command line code:0

Command line code:1:
bool(true)
```

morozov: I see. A type is always an object but it's a nullable object.

tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
```diff
@@ -193,6 +194,36 @@ public function testCompareChangedColumnsChangeType() : void
...
+        $integerType2 = Type::getType('integer');
+
+        $column1 = new Column('integerfield1', $integerType1);
+        $column2 = new Column('integerfield1', $integerType2);
+
+        $c = new Comparator();
+        self::assertEquals([], $c->diffColumn($column1, $column2));
+    }
+
+    public function testCompareColumnsOverridenType() : void
```
morozov: Should this be "overridden"?

garret-gunter: Yep, thanks.  It's fixed now.

morozov:
Looks good. Please squash since the history doesn't seem relevant.

garret-gunter:
I squashed commits.

morozov (approved):

morozov:
Thank you, @garret-gunter! Merged.

PR #3458, opened 2019-02-14
[WIP] Index are recreated during a diff even when not changed

yvoyer:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #3780

#### Summary

This PR adds the steps to reproduce for #3780
When generating a diff (using DoctrineMigration package) for a table containing an index, the index is always dropped and then recreated.

I wrote a test in the `Comparator` to see if it could be the source, but he seems to work properly. So I just added a test on the `MysqlPlatform` (the place where I got the error). 

I will try to work on a patch when I get the time, unless someone fixes it before.

**Note:** This seems to be a **regression** introduced in `2.9.2`, since it worked in `<=2.9.1`. 

**TODO**

- [X] Step to reproduce 
- [ ] Provide a fix

image

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Platforms/MySqlPlatformTest.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/MySqlPlatformTest.php
@@ -3,6 +3,8 @@
 namespace Doctrine\Tests\DBAL\Platforms;
 
 use Doctrine\DBAL\Platforms\MySqlPlatform;
+use Doctrine\DBAL\Schema\Index;
+use Doctrine\DBAL\Schema\TableDiff;
 use Doctrine\DBAL\TransactionIsolationLevel;
 
 class MySqlPlatformTest extends AbstractMySQLPlatformTestCase
@@ -18,5 +20,12 @@ class MySqlPlatformTest extends AbstractMySQLPlatformTestCase
             TransactionIsolationLevel::REPEATABLE_READ,
             $this->platform->getDefaultTransactionIsolationLevel()
         );
+    }
+
+    public function testPreserveIndexOnAlterTable()
+    {
+        $index = new Index('index_name', ['column']);
+        $diff = new TableDiff('table_name', [], [], [], [], [$index]);
+        $this->assertEquals([], $this->createPlatform()->getAlterTableSQL($diff));
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
+++ b/tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
@@ -398,6 +398,40 @@ class ComparatorTest extends TestCase
         self::assertEquals($expected, Comparator::compareSchemas($schema1, $schema2));
     }
 
+    public function testPreserveIndexesWhenNoChange()
+    {
+        $schema1 = new Schema(
+            [
+                'table' => new Table(
+                    'table',
+                    [
+                        'field' => new Column('field', Type::getType('integer')),
+                    ],
+                    [
+                        'primary' => new Index('primary', ['field']),
+                    ]
+                ),
+            ]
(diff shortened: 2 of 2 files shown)


tests/Doctrine/Tests/DBAL/Platforms/MySqlPlatformTest.php
```diff
@@ -19,4 +21,11 @@ public function testHasCorrectDefaultTransactionIsolationLevel()
             $this->platform->getDefaultTransactionIsolationLevel()
         );
     }
+
+    public function testPreserveIndexOnAlterTable()
+    {
+        $index = new Index('index_name', ['column']);
+        $diff = new TableDiff('table_name', [], [], [], [], [$index]);
+        $this->assertEquals([], $this->createPlatform()->getAlterTableSQL($diff));
```
SenseException: Can you please use `self::assertSame()` here?

tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
```diff
@@ -398,6 +398,40 @@ public function testCompareChangedIndex()
...
+                    [
+                        'primary' => new Index('primary', ['field']),
+                    ]
+                ),
+            ]
+        );
+
+        $expected = new SchemaDiff();
+        $expected->fromSchema = $schema1;
+        self::assertEquals($expected, Comparator::compareSchemas($schema1, $schema2));
```
SenseException: Can you please use `self::assertSame()` here?

SenseException (changes requested):
`testPreserveIndexOnAlterTable()` fails on my dbal 2.9.1 installation showing that index_name is dropped and added afterwards like described. `testPreserveIndexesWhenNoChange()` passes.

PowerKiKi:
Still seeing this in v2.10.0

PowerKiKi:
I finally dug deeper in this issue. Across five of my projects it was always not because a bug in dbal, but because I did not specify a length on a varchar index. That meant that MariaDB created an index with a default length ([which may vary](https://mariadb.com/kb/en/innodb-system-variables/#innodb_large_prefix)). Then DBAL correctly got confused because he saw a length in DB, but it was not specified in my code.

Specifying the length, like below, fixes the problem:

```diff
 /**
  * @ORM\Table(indexes={
- *     @ORM\Index(name="message", columns={"message"})
+ *     @ORM\Index(name="message", columns={"message"}, options={"lengths" = {191}})
  * })
  * @ORM\Entity(repositoryClass="Application\Repository\LogRepository")
  */
```

I think this issue should be closed.

PowerKiKi (changes requested):
I think this PR should be rejected as the introduced tests are not valid.

It is likely that the original behavior observed that lead to this PR is actually a misuse of DBAL. Possible solutions for the end-users are:

- [Adding parentheses](https://github.com/doctrine/dbal/issues/3780#issuecomment-290080207)
- [Using lowercase keywords](https://github.com/doctrine/dbal/issues/3780#issuecomment-381138100)
- [Specifying lengths](https://github.com/doctrine/dbal/pull/3458#issuecomment-569087641)
- [Quoting lengths](https://github.com/doctrine/dbal/issues/3780#issuecomment-565859839)

yvoyer:
@PowerKiKi, indeed it could be closed. but I would have liked to reproduce in a better way. I could rework it to assert that the varchar index has a length, so that this misuse never happens. Would it be better to move the test to doctrine/orm, with a mapping test? and maybe make a pr here to assert that a varchar index has a length?

PowerKiKi:
Requiring a length for varchar index sounds like a breaking to me. I'm not quite sure if it's the best way to fix this. We might need the insight of someone more involved in Doctrine to decide on that.

PR #3465, opened 2019-02-20
Replaced MySQL 5.7 installed from a PPA with an official Docker image

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | [job #495144274](https://travis-ci.org/doctrine/dbal/jobs/495144274)

### DIFF
--- a/tests/travis/install-mysql-5.7.sh
+++ b/tests/travis/install-mysql-5.7.sh
@@ -2,21 +2,14 @@
 
 set -ex
 
-echo "Installing MySQL 5.7..."
+echo "Starting MySQL 5.7..."
 
-sudo service mysql stop
-sudo apt-get remove "^mysql.*"
-sudo apt-get autoremove
-sudo apt-get autoclean
-echo mysql-apt-config mysql-apt-config/select-server select mysql-5.7 | sudo debconf-set-selections
-wget http://dev.mysql.com/get/mysql-apt-config_0.8.9-1_all.deb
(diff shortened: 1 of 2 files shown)


tests/travis/install-mysql-5.7.sh
```diff
@@ -2,21 +2,14 @@
...
-
-echo "Restart mysql..."
-sudo mysql -e "use mysql; update user set authentication_string=PASSWORD('') where User='root'; update user set plugin='mysql_native_password';FLUSH PRIVILEGES;"
+sudo docker run \
+    -d \
+    -e MYSQL_ALLOW_EMPTY_PASSWORD=yes \
+    -e MYSQL_DATABASE=doctrine_tests \
+    -p 33306:3306 \
+    --name mysql57 \
+    mysql:5.7
```
SenseException: Should a fixed 5.7 version/tag be used here or does the build always have to use the newest 5.7 version? I don't expect problems though.

morozov: We have the same for all other images except for DB2 (not sure if it was tagged deliberately). Unless stability issues start showing up, I think testing against the latest is fine.

lcobucci (approved):
Looks good, I'd suggest to add a tmpfs mount for the data folder, which makes things slightly faster but it can be done in another PR.

More info: https://vladmihalcea.com/how-to-run-integration-tests-at-warp-speed-with-docker-and-tmpfs/

Ocramius (approved):
:+1:

PR #3469, opened 2019-02-23
Replaced Xdebug with PCOV for code coverage

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

On the SQLite build with code coverage enabled, the test run with PCOV takes [~20 seconds](https://travis-ci.org/doctrine/dbal/jobs/497330533#L731) comparing to more than [1½ minutes](https://travis-ci.org/doctrine/dbal/jobs/497267081#L575) with Xdebug.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,18 +8,14 @@ cache:
         - $HOME/.composer/cache
 
 before_install:
+  - phpenv config-rm xdebug.ini || true
   - |
-    if [ "x$COVERAGE" != "xyes" ]; then
-      mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{,.disabled} || true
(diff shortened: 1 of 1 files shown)


morozov:
The Scrutinizer failure seems irrelevant since the code coverage was and remains 75%.

Majkl578:
How reliable is this in long-term? PCOV exists since last month...

morozov:
> How reliable is this in long-term? PCOV exists since last month...

PCOV was developed specifically to address performance issues of Xdebug when it comes to code coverage, so as long as it does its job well (which seems to be the case) I wouldn't expect it to disappear in one more month. If something goes wrong, nothing prevents us from switching back to Xdebug.

greg0ire (approved):

PR #3470, opened 2019-02-24
Replaced MockPlatform with the ones generated by PHPUnit

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

This is something I've wanted to do for quite some time. While I was working on #3348, it appeared that the introduction of return types made the existing mock implementation incompatible with the changes according to Scrutinizer but didn't break any tests. We can take it as a chance to get rid of it instead of maintaining.

More generally, the downsides of using non-autogenerated mocks are:

1. During refactoring, they require extra maintenance (e.g. changing method signatures or stubbing newly introduced methods). Unlike those, if the method signature declares a return type, PHPUnit will respect it and can even generate an object mock to be returned.
2. In the mocked platform class, a significant part of the methods was not used by the tests but still had to be maintained.
3. Shared mocks may create implicit dependencies between tests.
4. Hard-coded mocks cannot track method invocations, so it's not clear if a test using it just stubs a method or it tests the interaction between the SUT and the mock,


tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -153,9 +158,8 @@ public function testConnectDispatchEvent()
         $driverMock = $this->createMock(Driver::class);
         $driverMock->expects($this->at(0))
                    ->method('connect');
-        $platform = new Mocks\MockPlatform();
 
-        $conn = new Connection(['platform' => $platform], $driverMock, new Configuration(), $eventManager);
+        $conn = new Connection([], $driverMock, new Configuration(), $eventManager);
```
Majkl578: Was `platform` removed before it was unused or for some other reason?

morozov: It wasn't used in the test.

Majkl578 (approved):

PR #3473, opened 2019-02-27
Avoid database connection from PHPUnit data providers

### LINKED ISSUE #3471: DriverTest::testDatabaseParameters() is unstable
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.0-dev

In the last couple of weeks, I've seen this test failing a few times for no reason.

When running against PostgreSQL, it may [fail](https://travis-ci.org/doctrine/dbal/jobs/498266025) like this:
```
1) Warning
The data provider specified for Doctrine\Tests\DBAL\Functional\Driver\PDOPgSql\DriverTest::testDatabaseParameters is invalid.
An exception occurred in driver: SQLSTATE[08006] [7] server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

When running against MySQL (yes), it may [fail](https://travis-ci.org/doctrine/dbal/jobs/496130765) like this:
```
1) Warning
The data provider specified for Doctrine\Tests\DBAL\Functional\Driver\PDOPgSql\DriverTest::testDatabaseParameters is invalid.
An exception occurred in driver: SQLSTATE[HY000] [2006] MySQL server has gone away
```

In its data provider, the test tries to detect the database name specified in the connection parameters and by that triggers the database connection while the database seems to be not yet ready.

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

Fixes #3471.

The data provider `PDOPgSql\DriverTest::getDatabaseParameter()` triggers a database connection in order to obtain the configured database name. Under certain circumstances, this connection may fail and invalidate the entire test suite. What are those circumstances?

1. PostgreSQL running in a Docker container:
   ```bash
   docker run -d -p 5432:5432 --name pgsql postgres:11.1
   docker exec -i pgsql bash <<< 'until pg_isready -U postgres > /dev/null 2>&1 ; do sleep 1; done'
   php -r 'try {new PDO("pgsql:host=127.0.0.1;port=5432", "postgres", "");} catch (PDOException $e) {echo $e->getMessage(), PHP_EOL;}'

   # SQLSTATE[08006] [7] server closed the connection unexpectedly
   #     This probably means the server terminated abnormally
   #     before or while processing the request.
   ```
   This failure [happens](https://travis-ci.org/doctrine/dbal/jobs/498266025) inconsistently, one of a few times.
2. MySQL running in a Docker container without an explicitly created database. Even if the test is going to be skipped, the data provider is still executed:
   ```bash
   docker run -d -e MYSQL_ALLOW_EMPTY_PASSWORD=yes -p 33306:3306 --name mysql mysql:5.7
   docker exec -i mysql bash <<< 'until mysql -e "show databases"; do sleep 1; done'
   php -r 'try {new PDO("mysql:host=127.0.0.1;port=33306", "root", "");} catch (PDOException $e) {echo $e->getMessage(), PHP_EOL;}'

   # SQLSTATE[HY000] [2006] MySQL server has gone away
   ```
   This failure [happens](https://travis-ci.org/doctrine/dbal/jobs/496130765) in 100% of the cases.

The solution is to expose `TestUtil::getConnectionParams()` to the test suite in order to let it get the parameters from configuration instead of connection. The logic from `TestUtil::getSpecifiedConnectionParams()` has been worked into `TestUtil::initializeDatabase()` in order to avoid side effects in `TestUtil::getConnectionParams()`.

This way, until we find a better solution, at least the MySQL jobs will not be impacted by an issue in PostgreSQL tests.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/Driver/AbstractDriverTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Driver/AbstractDriverTest.php
@@ -55,7 +55,7 @@ abstract class AbstractDriverTest extends DbalFunctionalTestCase
         );
 
         self::assertSame(
-            $this->getDatabaseNameForConnectionWithoutDatabaseNameParameter(),
+            static::getDatabaseNameForConnectionWithoutDatabaseNameParameter(),
             $this->driver->getDatabase($connection)
         );
     }
@@ -65,10 +65,7 @@ abstract class AbstractDriverTest extends DbalFunctionalTestCase
      */
     abstract protected function createDriver();
 
-    /**
-     * @return string|null
-     */
-    protected function getDatabaseNameForConnectionWithoutDatabaseNameParameter()
+    protected static function getDatabaseNameForConnectionWithoutDatabaseNameParameter() : ?string
     {
         return null;
     }
--- a/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOPgSql/DriverTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOPgSql/DriverTest.php
@@ -50,9 +50,12 @@ class DriverTest extends AbstractDriverTest
         );
     }
 
-    public function getDatabaseParameter()
+    /**
+     * @return mixed[][]
+     */
+    public static function getDatabaseParameter() : iterable
     {
-        $params            = TestUtil::getConnection()->getParams();
+        $params            = TestUtil::getConnectionParams();
         $realDatabaseName  = $params['dbname'] ?? '';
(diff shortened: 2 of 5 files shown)


morozov:
@Majkl578 apart form moving the connection out of the data provider, can you think of a way to improve the readiness probe for these containers? I tried poking the database from a second container instead of running the probe inside the running one in order to account for potential connectivity issues but it didn't help.

Ocramius (approved):

Majkl578:
@morozov It depends on the database, at least postgres can use pg_isready with remote servers too. It may be more tricky with other DBs (i.e. sqlsrv). Can you create new issue for that?

PR #3474, opened 2019-02-27
Remove more hard-coded mock classes

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

The same reasons as in #3470 apply.


tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -165,10 +164,22 @@ public function testConnectDispatchEvent()
...
     {
-        $driverMock = new DriverMock();
-        $connection = new Connection($this->params, $driverMock);
-        self::assertInstanceOf(EventManager::class, $connection->getDatabasePlatform()->getEventManager());
-        self::assertSame($connection->getEventManager(), $connection->getDatabasePlatform()->getEventManager());
+        $eventManager = new EventManager();
+
+        /** @var AbstractPlatform|MockObject $driver */
+        $platform = $this->createMock(AbstractPlatform::class);
+        $platform->expects($this->once())
```
Ocramius: Not important here, but `self::any()` should be the correct access type

Ocramius (approved):

PR #3475, opened 2019-02-27
Make PHP 7.3 the primary PHP version on Travis CI

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

Since PHP 7.3 is currently the latest stable version, according to https://github.com/doctrine/dbal/pull/3347#issuecomment-439231396, PHP 7.2 and PHP 7.3 jobs have been swapped. The only exception is that the IBM DB2 driver is still tested on PHP 7.2 since PHP 7.3 is not yet supported.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,30 +43,85 @@ jobs:
   include:
 
     - stage: Smoke Testing
-      php: 7.2
+      php: 7.3
       env: DB=sqlite COVERAGE=yes
     - stage: Smoke Testing
-      php: 7.2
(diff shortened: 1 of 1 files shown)


SenseException (approved):
Looks okay to me. Scrutinizer seems to have that test_coverage issue again.

morozov:
@SenseException thank you. Most likely, the coverage changes are caused by the difference in Opcache implementation details on PHP 7.2 and 7.3.

Issue #2901, opened 2017-10-27
SQLite in-memory option useless without shared cache option


bitwombat:
DBAL supports the `'memory' => true` option for SQLite database, but the way SQLite works, each new connection opening gets a new database.  This makes the feature useless because of how Doctrine opens up a new connection for every operation.

SQLite has an option that can be passed in the URI:  `sqlite::memory:?cache=shared`  This makes subsequent connection open operations share the same 'cache' - for some reason this means they share the in-memory database.

Unfortunately, PDO doesn't support this URI scheme, nor a setting via another mechanism (the docs for SQLite make it seem like the URI is the only way to get this behaviour, though there is a SQLite function called sqlite3_enable_shared_cache().  PHP's sqlite3 module supports this, but I don't think that's much help to Doctrine/DBAL).

I spent quite a few hours figuring this out, so I'd like to help document it somewhere to save the next poor sap.

Comments?

References:  
#2898 - This same issue, but I closed it because I was confused and the comments muddy everthing.
https://github.com/php/php-src/pull/2734/files (a PR that added support for an SQLite attribute - does a  similar attribute exist for the shared cache option?)
http://www.sqlite.org/sharedcache.html
http://www.sqlite.org/inmemorydb.html
https://stackoverflow.com/questions/9150319/enable-shared-pager-cache-in-sqlite-using-php-pdo
https://stackoverflow.com/questions/16487679/shared-cache-on-pdo-sqlite

mikeryan776:
Added a pull request to support this - just a simple matter of adding "?cache=shared" to the DSN. Before attempting tests, a couple of questions:

1. I think shared cache mode is supported beyond :memory:, so this perhaps should go just before the return?
2. Are there other similar settings driver settings that could be supported?

Thanks.

mikeryan776:
The current idea in the PR is to add 'dsn' as a possible connection param for the pdo_sqlite driver. For my own purposes, for now I'm using this solution as follows:

In my composer.json, referencing patch file [sqlite-shared-cache.patch.txt](https://github.com/doctrine/dbal/files/2912534/sqlite-shared-cache.patch.txt) (minus the .txt):

```json

    "require-dev": {
        "phpunit/phpunit" : "^8.0",
        "squizlabs/php_codesniffer": "^2.3",
        "cweagans/composer-patches": "~1.0"
    },

    "extra": {
        "patches" : {
            "doctrine/dbal": {
[...]

PR #3476, opened 2019-02-27
#2901: Pass shared cache setting through DSN.

mikeryan776:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | 

#### Summary

With this change, 'sqlite:///:memory:?cache=shared' will allow multiple Sqlite connections within a process to share a memory database.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/PDOSqlite/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/PDOSqlite/Driver.php
@@ -61,14 +61,14 @@ class Driver extends AbstractSQLiteDriver
      */
     protected function _constructPdoDsn(array $params)
     {
+        if (isset($params['dsn'])) {
+            return $params['dsn'];
+        }
         $dsn = 'sqlite:';
         if (isset($params['path'])) {
             $dsn .= $params['path'];
         } elseif (isset($params['memory'])) {
             $dsn .= ':memory:';
-            if (isset($params['cache'])) {
-                $dsn .= '?cache=' . $params['cache'];
-            }
         }
 
         return $dsn;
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Driver/PDOSqlite/Driver.php
```diff
@@ -66,6 +66,9 @@ protected function _constructPdoDsn(array $params)
             $dsn .= $params['path'];
         } elseif (isset($params['memory'])) {
             $dsn .= ':memory:';
+            if (isset($params['cache'])) {
```
Ocramius: Possibly a better idea to pass in a pre-built PDO instance, when using advanced driver settings?

morozov: Or maybe accept a pre-built DSN? I don't like the idea of passing a PDO instance since it breaks encapsulation (therefore, unexpected side effects are possible) and is not supported by non-PDO drivers.

mikeryan776: Yeah, I think it's better to not have application code touching PDO directly - passing the dsn seems cleaner to me.

```php
            'connection' => [
//                'url' => 'sqlite:///:memory:?cache=shared',
                'driver' => 'pdo_sqlite',
                'dsn' => 'sqlite::memory:?cache=shared',
            ],
```

mikeryan776:
Per https://github.com/doctrine/dbal/issues/2901#issuecomment-468075680, this doesn't actually work. I'm going back to the drawing board - if I can't figure out a way to do this via the DSN, maybe I need to take up the PDO instance per @Ocramius?

PR #3481, opened 2019-03-14
Updated PHPStan to v0.11.3

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

### DIFF
--- a/lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereException.php
+++ b/lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereException.php
@@ -4,7 +4,6 @@ namespace Doctrine\DBAL\Driver\SQLAnywhere;
 
 use Doctrine\DBAL\Driver\AbstractDriverException;
 use InvalidArgumentException;
-use function is_resource;
 use function sasql_error;
 use function sasql_errorcode;
 use function sasql_sqlstate;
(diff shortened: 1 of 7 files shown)


lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereException.php
```diff
@@ -28,14 +28,6 @@ class SQLAnywhereException extends AbstractDriverException
      */
     public static function fromSQLAnywhereError($conn = null, $stmt = null)
     {
-        if ($conn !== null && ! is_resource($conn)) {
```
morozov: Judging by the number of newly reported issues, we enforce union types only in a couple of places. Let's revert them for consistency and rely on static analysis rather than on runtime checks.

lib/Doctrine/DBAL/Platforms/DrizzlePlatform.php
```diff
@@ -576,7 +576,7 @@ public function convertBooleans($item)
     {
         if (is_array($item)) {
             foreach ($item as $key => $value) {
-                if (! is_bool($value) && ! is_numeric($item)) {
+                if (! is_bool($value) && ! is_numeric($value)) {
```
morozov: This most likely was a copy-pasta bug.

Ocramius (changes requested):

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -928,7 +928,7 @@ protected function _getPortableTableDefinition($table)
     /**
      * @param mixed[][] $users
      *
-     * @return string[][]
+     * @return mixed[][]
```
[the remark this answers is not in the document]
morozov: Yeah… I left this unchanged and restricted the input instead.

Ocramius (approved):

PR #3483, opened 2019-03-15
Moved rowCount() from Statement to ResultStatement

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

The issue was discovered using PHPStan as part of [#3436](https://github.com/doctrine/dbal/pull/3436/files#diff-f4ea3306103e9a98e0c10981a495d980R200).

### DIFF
--- a/lib/Doctrine/DBAL/Cache/ArrayStatement.php
+++ b/lib/Doctrine/DBAL/Cache/ArrayStatement.php
@@ -60,6 +60,18 @@ class ArrayStatement implements IteratorAggregate, ResultStatement
     /**
      * {@inheritdoc}
      */
+    public function rowCount() : int
+    {
+        if ($this->data === null) {
+            return 0;
+        }
+
+        return count($this->data);
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


lib/Doctrine/DBAL/Cache/ArrayStatement.php
```diff
@@ -57,6 +57,18 @@ public function columnCount()
         return $this->columnCount;
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function rowCount() : int
+    {
+        if (! isset($this->data)) {
```
Ocramius: Since the property is never unset, I'd use `=== null` here

Ocramius (approved):

lib/Doctrine/DBAL/Cache/ArrayStatement.php
Ocramius: @morozov feel free to merge once this is handled

PR #3485, opened 2019-03-15
Connection::ping() will return void

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

As suggested in https://github.com/doctrine/dbal/pull/3348#discussion_r236016744.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade to 3.0
 
+## BC BREAK `Connection::ping()` returns `void`.
+
+`Connection::ping()` and `PingableConnection::ping()` no longer return a boolean value. They will throw an exception in case of failure.
+
 ## BC BREAK User-provided `PDO` instance is no longer supported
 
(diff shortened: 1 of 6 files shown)


Ocramius:
The unhappy path could be tested if we were able to cut the connection to the DB. For PDO, we could mock the PDO instance itself. For drivers relying on global functions, it would require us to bring down the real DB connection.

I'd say that if the PDO testing is possible, we should at least try that.

morozov:
Can't inject `PDO` on `develop` due to #3487.

morozov:
Testing the happy path with PDO is no longer possible as of #3548. The fact that it takes us that long to come up with a testing approach makes me question the need for this API even more:

> Given that only MySQL implements it natively, what's the value of the specialized API?

@Ocramius, in the interest of making progress with enforcing types, I think we should just merge it as is.

Ocramius (approved):

Ocramius:
@morozov :ship:

PR #3486, opened 2019-03-15
Converted Connection and Statement methods which returned false in case of a failure into void

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

### DIFF
--- a/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
+++ b/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
@@ -123,7 +123,7 @@ class MasterSlaveConnection extends Connection
     /**
      * {@inheritDoc}
      */
-    public function connect($connectionName = null)
+    public function connect($connectionName = null) : void
     {
         $requestedConnectionChange = ($connectionName !== null);
         $connectionName            = $connectionName ?: 'slave';
@@ -136,7 +136,7 @@ class MasterSlaveConnection extends Connection
         // change request, then abort right here, because we are already done.
         // This prevents writes to the slave in case of "keepSlave" option enabled.
         if ($this->_conn !== null && ! $requestedConnectionChange) {
-            return false;
+            return;
         }
 
         $forceMasterAsSlave = false;
@@ -153,7 +153,7 @@ class MasterSlaveConnection extends Connection
                 $this->connections['slave'] = $this->_conn;
             }
 
-            return false;
+            return;
         }
 
         if ($connectionName === 'master') {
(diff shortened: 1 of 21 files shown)


lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -288,13 +288,15 @@ public function bindParam($column, &$variable, $type = ParameterType::STRING, $l
 
         $this->boundValues[$column] =& $variable;
 
-        return oci_bind_by_name(
+        if (! oci_bind_by_name(
```
Majkl578: This might be better as separate variable assignment than multiline if.

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -92,15 +92,9 @@ public function bindParam($column, &$variable, $type = ParameterType::STRING, $l
     /**
      * {@inheritdoc}
      */
-    public function closeCursor()
+    public function closeCursor() : void
     {
-        try {
-            return $this->stmt->closeCursor();
-        } catch (\PDOException $exception) {
```
Majkl578: Does this mean it can now throw generic PDOException here? Should we wrap it?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
morozov: I wouldn’t like to introduce a variable just because the function call spans multiple lines.

Let’s leave it like this unless you’re really dislike it.

morozov: I wouldn’t like to introduce a variable just because the function call spans multiple lines.

Let’s leave it like this unless you’re really dislike it.

lib/Doctrine/DBAL/Driver/PDOStatement.php
morozov: I thought I'd put it into the description but I hadn't. `PDOStatement::closeCursor()` doesn't throw exceptions in any of the scenarios we have in our tests including the newly added ones. That's why the interface annotation doesn't declare any `@throws`. I cannot imagine a situation when the result cannot be freed on a valid statement and how this could be handled.

AFAIK, procedural APIs like `oci_cancel($stmt)` and the similar from other drivers only return `false` when the statement handle is invalid which is impossible in our case.

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -167,12 +167,12 @@ public function connect($connectionName = null)
...
 
-        if ($this->_eventManager->hasListeners(Events::postConnect)) {
-            $eventArgs = new ConnectionEventArgs($this);
-            $this->_eventManager->dispatchEvent(Events::postConnect, $eventArgs);
+        if (! $this->_eventManager->hasListeners(Events::postConnect)) {
+            return;
         }
 
-        return true;
+        $eventArgs = new ConnectionEventArgs($this);
```
Ocramius: Variable redundant

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
Ocramius: Looks acceptable: an added single-use variable is more confusing to the reader, IMO

Ocramius (approved):

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
morozov: It was like this prior to the change and only got changed because of `EarlyExitNotUsed` which often makes more harm than good if enforced (e.g. https://github.com/slevomat/coding-standard/issues/371).

PR #3488, opened 2019-03-16
Connection::quote() can only quote strings

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

The signature of `Connection::quote()` allows `mixed $input` and `int $type`. On the one hand, not all drivers support all existing types, on the other hand, from the SQL perspective, it doesn't make any sense to quote anything else than a string.

As part of the `mixed` type, the method accepts `null` but implicitly converts it to a `string` which is unacceptable in the SQL context. Enforcing the accepted type to `string` helped reveal a few bugs (to be reported later, see the test failures).

The wrapper connection also supports DBAL types which in the absence of the sense in driver-level types in the escaping context boils down to just converting the value to a PHP value which may not even be a string. Besides being redundant, this functionality doesn't have any test coverage.

It's proposed to keep `Connection::quote()` responsible only for quoting strings.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade to 3.0
 
+## BC BREAK `Statement::quote()` only accepts strings.
+
+`Statement::quote()` and `ExpressionBuilder::literal()` no longer accept arguments of an arbitrary type and and don't implement type-specific handling. Only strings can be quoted.
+
 ## BC BREAK `Statement` and `Connection` methods return `void`.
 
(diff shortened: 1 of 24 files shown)


Ocramius (approved):

Ocramius:
Patch looks good, but CI is still unhappy with it

morozov:
> Patch looks good, but CI is still unhappy with it

@Ocramius I thought it was due to some implementations of `AbstractPlatform::getListViewsSQL()`, `::getListSequencesSQL()` and `::getListTableForeignKeysSQL()` not properly handling `$database = null` and was planning to report/address them separately as a dependency for this issue. However, it turned out the failures weren't because of that.

Those issues are still to be reported and address later.

Ocramius (approved):
:+1: :ship:

PR #3489, opened 2019-03-17
Update doctrine coding standard from 5.0 to 6.0

ahmedsbytes:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Upgrade doctrine coding standard to 6.0

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
@@ -36,28 +36,34 @@ abstract class AbstractMySQLDriver implements Driver, ExceptionConverterDriver, 
                 return new Exception\LockWaitTimeoutException($message, $exception);
             case '1050':
                 return new Exception\TableExistsException($message, $exception);
+
             case '1051':
             case '1146':
                 return new Exception\TableNotFoundException($message, $exception);
+
             case '1216':
             case '1217':
             case '1451':
             case '1452':
             case '1701':
                 return new Exception\ForeignKeyConstraintViolationException($message, $exception);
+
             case '1062':
             case '1557':
             case '1569':
             case '1586':
                 return new Exception\UniqueConstraintViolationException($message, $exception);
+
             case '1054':
             case '1166':
             case '1611':
                 return new Exception\InvalidFieldNameException($message, $exception);
+
             case '1052':
             case '1060':
             case '1110':
                 return new Exception\NonUniqueFieldNameException($message, $exception);
+
             case '1064':
             case '1149':
             case '1287':
@@ -71,6 +77,7 @@ abstract class AbstractMySQLDriver implements Driver, ExceptionConverterDriver, 
             case '1554':
             case '1626':
                 return new Exception\SyntaxErrorException($message, $exception);
+
             case '1044':
             case '1045':
             case '1046':
@@ -84,6 +91,7 @@ abstract class AbstractMySQLDriver implements Driver, ExceptionConverterDriver, 
             case '2002':
             case '2005':
                 return new Exception\ConnectionException($message, $exception);
+
             case '1048':
             case '1121':
             case '1138':
--- a/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
@@ -36,34 +36,28 @@ abstract class AbstractMySQLDriver implements Driver, ExceptionConverterDriver, 
                 return new Exception\LockWaitTimeoutException($message, $exception);
             case '1050':
                 return new Exception\TableExistsException($message, $exception);
-
             case '1051':
             case '1146':
                 return new Exception\TableNotFoundException($message, $exception);
-
             case '1216':
             case '1217':
             case '1451':
             case '1452':
             case '1701':
                 return new Exception\ForeignKeyConstraintViolationException($message, $exception);
-
             case '1062':
             case '1557':
             case '1569':
             case '1586':
                 return new Exception\UniqueConstraintViolationException($message, $exception);
-
             case '1054':
             case '1166':
             case '1611':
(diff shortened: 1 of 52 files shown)


lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
```diff
@@ -239,7 +239,7 @@ private function setDriverOptions(array $driverOptions = [])
         $exceptionMsg = "%s option '%s' with value '%s'";
 
         foreach ($driverOptions as $option => $value) {
-            if ($option === static::OPTION_FLAGS) {
+            if ($option === self::OPTION_FLAGS) {
```
Ocramius: Wondering if this is BC compliant: does any subclass override this?

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -247,15 +247,15 @@ public function getClobTypeDeclarationSQL(array $field)
         if (! empty($field['length']) && is_numeric($field['length'])) {
             $length = $field['length'];
 
-            if ($length <= static::LENGTH_LIMIT_TINYTEXT) {
```
Ocramius: Also here: BC compliant? Are there overrides in subclasses?

Ocramius (changes requested):

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
morozov: Even if we don't some dependencies in theory may. I'd disable the rule for `master` and add it in `develop`. In order to avoid further confusion, all affected constants (not sure what specifically the rule is above) will need to be marked `private`~/`final`~ …we don't have `final const` in PHP?!

lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
```diff
@@ -36,34 +36,28 @@ public function convertException($message, DriverException $exception)
                 return new Exception\LockWaitTimeoutException($message, $exception);
             case '1050':
                 return new Exception\TableExistsException($message, $exception);
-
```
morozov: I'd like these new lines to be kept. They visually separate `case`s from each other.

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php
```diff
@@ -100,7 +100,9 @@ public function quote($value, $type = ParameterType::STRING)
     {
         if (is_int($value)) {
             return $value;
-        } elseif (is_float($value)) {
+        }
```
morozov: Nice one.

lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php
```diff
@@ -100,7 +100,9 @@ public function quote($value, $type = ParameterType::STRING)
     {
         if (is_int($value)) {
             return $value;
-        } elseif (is_float($value)) {
```
morozov: Nice one.

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
amaabdou: Nope, non of them overridden in any subclass

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
amaabdou: I did not see any subclass for MysqliConnection,                 
I tried to find `final const` but I did not find in any, https://3v4l.org/NU0ic#output         

The rule of prohibiting late static binding is only for constants, feels logical, late static binding is always annoying         
SlevomatCodingStandard.Classes.DisallowLateStaticBindingForConstants.DisallowedLateStaticBindingForConstant

but I'm sure you have seen much more cases than me :D , so what should be done ?

lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
amaabdou: If I kept them I will be breaking 

SlevomatCodingStandard.ControlStructures.ControlStructureSpacing.IncorrectLinesCountAfterLastControlStructure

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
Ocramius: @morozov's plan makes sense. This change is a potential BC break for downstream consumers, so we can only apply it for `develop` (which is for the next major release)

amaabdou: Should I re-base this PR against the develop ?

morozov: @amaabdou please revert all `static` → `self` changes from this PR and keep it targeting master. Since the coding standard will report the violations, please exclude the corresponding rules for those files.

Once this is merged, I'll rebase `develop` and we can apply the breaking changes.

lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
morozov: That's a bummer. I'd say it's an oversight on the standard side. Would you mind filing an issue? If `return` is part of a `case` followed by another `case`, the new line should be allowed.

For the time being, please revert these changes and disable the rule.

Ocramius (approved):

Ocramius:
Thanks @amaabdou!

morozov:
`develop` is up to date with `master`. We can enable `DisallowedLateStaticBindingForConstant` there.

PR #3491, opened 2019-03-17
Changed the type of $char in AbstractPlatform::getTrimExpression()from string|false to ?string

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

This breaking change is needed to enforce the type of `$char` natively to `?string`. The existing approach of using `false` to specify the absence of a value is incorrect since `false` is the opposite of `true` which doesn't make sense as the value of `$char`, while `null` is exactly what should be used in this case.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
@@ -24,6 +24,7 @@ use function explode;
 use function func_get_args;
 use function get_class;
 use function implode;
+use function in_array;
 use function is_string;
 use function preg_match;
 use function sprintf;
@@ -1121,10 +1122,16 @@ SQL
     /**
      * {@inheritdoc}
      */
-    public function getTrimExpression($str, $pos = TrimMode::UNSPECIFIED, $char = false)
+    public function getTrimExpression(string $str, int $mode = TrimMode::UNSPECIFIED, ?string $char = null) : string
     {
-        if (! $char) {
-            switch ($pos) {
+        if (! in_array($mode, [TrimMode::UNSPECIFIED, TrimMode::LEADING, TrimMode::TRAILING, TrimMode::BOTH], true)) {
+            throw new InvalidArgumentException(
+                sprintf('The value of $mode is expected to be one of the TrimMode constants, %d given', $mode)
+            );
+        }
+
+        if ($char === null) {
+            switch ($mode) {
                 case TrimMode::LEADING:
                     return $this->getLtrimExpression($str);
                 case TrimMode::TRAILING:
@@ -1136,7 +1143,7 @@ SQL
 
         $pattern = "'%[^' + " . $char . " + ']%'";
 
-        switch ($pos) {
+        switch ($mode) {
             case TrimMode::LEADING:
                 return 'SUBSTR(' . $str . ', PATINDEX(' . $pattern . ', ' . $str . '))';
             case TrimMode::TRAILING:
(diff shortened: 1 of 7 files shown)


tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php
```diff
@@ -665,11 +665,6 @@ public function testGeneratesSQLSnippets()
             "REVERSE(SUBSTR(REVERSE(column), PATINDEX('%[^' + c + ']%', REVERSE(column))))",
             $this->platform->getTrimExpression('column', TrimMode::TRAILING, 'c')
         );
-        self::assertEquals(
-            "REVERSE(SUBSTR(REVERSE(SUBSTR(column, PATINDEX('%[^' + c + ']%', column))), PATINDEX('%[^' + c + ']%', " .
-            "REVERSE(SUBSTR(column, PATINDEX('%[^' + c + ']%', column))))))",
-            $this->platform->getTrimExpression('column', null, 'c')
-        );
```
morozov: This cases abuses the API and covers the same scenario as the following one.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1031,23 +1031,36 @@ public function getModExpression($expression1, $expression2)
     /**
      * {@inheritDoc}
      */
-    public function getTrimExpression($str, $pos = TrimMode::UNSPECIFIED, $char = false)
+    public function getTrimExpression(string $str, int $mode = TrimMode::UNSPECIFIED, ?string $char = null) : string
     {
-        if (! $char) {
-            switch ($pos) {
+        switch ($mode) {
```
Ocramius: I'd change this to `if (! in_array(...)) {`

lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
```diff
@@ -1121,10 +1121,26 @@ public function getTimeTypeDeclarationSQL(array $fieldDeclaration)
     /**
      * {@inheritdoc}
      */
-    public function getTrimExpression($str, $pos = TrimMode::UNSPECIFIED, $char = false)
+    public function getTrimExpression(string $str, int $mode = TrimMode::UNSPECIFIED, ?string $char = null) : string
     {
-        if (! $char) {
-            switch ($pos) {
+        switch ($mode) {
```
Ocramius: I'd change this to `if (! in_array(...)) {`

Ocramius: BTW: thoughts on including an assertion library for this stuff, @doctrine/doctrinecore?

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
```diff
@@ -547,6 +548,12 @@ public function getTrimExpressionData()
         ];
     }
 
+    public function testTrimExpressionInvalidMode()
```
Ocramius: ` : void`

Ocramius (changes requested):

lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
morozov: In this specific case, it'd reduce code duplication but wouldn't reduce knowledge duplication. Feels like sweeping mess under the rug.

Ocramius (approved):

PR #3494, opened 2019-03-19
Modified AbstractPlatform::getLocateExpression() and ::getSubstringExpression() signatures

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

This breaking change is needed to enforce the type of `$start` natively. Additionally, all function arguments that accept an SQL expression will be type-hinted as `string` or `?string`, regardless of which SQL type they represent. Function arguments have been rename closer to their SQL analogs.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -850,17 +850,16 @@ abstract class AbstractPlatform
     }
 
     /**
-     * Returns the SQL snippet to get the position of the first occurrence of substring $substr in string $str.
+     * Returns the SQL snippet to get the position of the first occurrence of the substring in the string.
      *
-     * @param string    $str      Literal string.
-     * @param string    $substr   Literal string to find.
-     * @param int|false $startPos Position to start at, beginning of string by default.
-     *
-     * @return string
+     * @param string      $string    SQL expression producing the string to locate the substring in.
+     * @param string      $substring SQL expression producing the substring to locate.
+     * @param string|null $start     SQL expression producing the position to start at.
+     *                               Defaults to the beginning of the string.
      *
      * @throws DBALException If not supported on this platform.
      */
-    public function getLocateExpression($str, $substr, $startPos = false)
+    public function getLocateExpression(string $string, string $substring, ?string $start = null) : string
     {
         throw DBALException::notSupported(__METHOD__);
     }
@@ -876,25 +875,22 @@ abstract class AbstractPlatform
     }
 
     /**
-     * Returns a SQL snippet to get a substring inside an SQL statement.
+     * Returns an SQL snippet to get a substring inside the string.
      *
      * Note: Not SQL92, but common functionality.
      *
-     * SQLite only supports the 2 parameter variant of this function.
-     *
-     * @param string   $value  An sql string literal or column name/alias.
-     * @param int      $from   Where to start the substring portion.
-     * @param int|null $length The substring portion length.
-     *
-     * @return string
+     * @param string      $string SQL expression producing the string from which a substring should be extracted.
+     * @param string      $start  SQL expression producing the position to start at,
+     * @param string|null $length SQL expression producing the length of the substring portion to be returned.
+     *                            By default, the entire substring is returned.
      */
-    public function getSubstringExpression($value, $from, $length = null)
+    public function getSubstringExpression(string $string, string $start, ?string $length = null) : string
     {
         if ($length === null) {
-            return 'SUBSTRING(' . $value . ' FROM ' . $from . ')';
+            return sprintf('SUBSTRING(%s FROM %s)', $string, $start);
         }
 
-        return 'SUBSTRING(' . $value . ' FROM ' . $from . ' FOR ' . $length . ')';
+        return sprintf('SUBSTRING(%s FROM %s FOR %s)', $string, $start, $length);
     }
 
     /**
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -78,18 +78,6 @@ class PostgreSqlPlatform extends AbstractPlatform
     /**
      * {@inheritDoc}
      */
-    public function getSubstringExpression($value, $from, $length = null)
-    {
-        if ($length === null) {
-            return 'SUBSTRING(' . $value . ' FROM ' . $from . ')';
-        }
-
-        return 'SUBSTRING(' . $value . ' FROM ' . $from . ' FOR ' . $length . ')';
-    }
-
-    /**
-     * {@inheritDoc}
-     */
     public function getNowExpression()
     {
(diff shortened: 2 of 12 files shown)


lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -75,18 +75,6 @@ public function setUseBooleanTrueFalseStrings($flag)
         $this->useBooleanTrueFalseStrings = (bool) $flag;
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    public function getSubstringExpression($value, $from, $length = null)
```
morozov: This is the same as the default implementation.

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -96,28 +96,26 @@ public function getTrimExpression(string $str, int $mode = TrimMode::UNSPECIFIED
 
     /**
      * {@inheritDoc}
-     *
-     * SQLite only supports the 2 parameter variant of this function
```
morozov: This is obsolete.

tests/Doctrine/Tests/DBAL/Platforms/DB2PlatformTest.php
```diff
@@ -332,7 +332,6 @@ public function testGeneratesSQLSnippets()
         self::assertEquals("'1987/05/02' - 10 YEAR", $this->platform->getDateSubYearsExpression("'1987/05/02'", 10));
         self::assertEquals(' WITH RR USE AND KEEP UPDATE LOCKS', $this->platform->getForUpdateSQL());
         self::assertEquals('LOCATE(substring_column, string_column)', $this->platform->getLocateExpression('string_column', 'substring_column'));
-        self::assertEquals('LOCATE(substring_column, string_column)', $this->platform->getLocateExpression('string_column', 'substring_column'));
```
morozov: This is a duplicate of the above.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -850,17 +850,15 @@ public function getLowerExpression($str)
...
     /**
-     * Returns the SQL snippet to get the position of the first occurrence of substring $substr in string $str.
+     * Returns the SQL snippet to get the position of the first occurrence of the substring in the string.
      *
-     * @param string    $str      Literal string.
-     * @param string    $substr   Literal string to find.
-     * @param int|false $startPos Position to start at, beginning of string by default.
-     *
-     * @return string
+     * @param string      $string    String to locate the substring in, SQL expression.
```
Ocramius: This comment is unclear to me - can you rephrase it, please?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -850,17 +850,15 @@ public function getLowerExpression($str)
...
+     * @param string      $substring Substring to locate, SQL expression.
```
Ocramius: I think a better naming could be "needle"

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -850,17 +850,15 @@ public function getLowerExpression($str)
...
+     * @param string|null $start     Position to start at, SQL expression. Defaults to the beginning of the string.
```
Ocramius: Alao confusing - why is this a `string`? Could we maybe add example usages?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -876,25 +874,22 @@ public function getNowExpression()
...
      *
-     * SQLite only supports the 2 parameter variant of this function.
-     *
-     * @param string   $value  An sql string literal or column name/alias.
-     * @param int      $from   Where to start the substring portion.
-     * @param int|null $length The substring portion length.
-     *
-     * @return string
+     * @param string      $string String to get a substring in, SQL expression.
+     * @param string      $start  Position to start at, SQL expression.
```
Ocramius: Similar to the above, `start` being a `string` is confusing to me - we'll need usage examples, IMO

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
Ocramius: Do we know since when? We should probably require SQLite to be at least newer than X

Ocramius (changes requested):

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: This is the best I could come up with. Otherwise, it was "The SQL expression which represents the string in which the substring should be located". And this "the SQL expression" will be in a lot of places. Is that better?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: I'd keep it closer to the SQL naming, not to PHP. Otherwise, it'd have to be `$haystack` and `$needle`.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: Because it's an SQL expression, not a value (see the test). It could a reference to another column in the query or a function call or whatever else. There will be a lot of changes like this in the API. Not sure if we want to put an example in every occurrence. What would you suggest?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
Ocramius: Aha, the wording "sql expression" put after the comma did confuse me

morozov:
> Do we know since when? We should probably require SQLite to be at least newer than X

No, but [at least as of 2008](https://web.archive.org/web/20081219142839/https://www.sqlite.org/lang_corefunc.html).

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: So what do you recommend?

Ocramius: "SQL expression producing the haystack" or something like that?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: There are more than 40 methods in this class which accept SQL expressions. We do need some explanation but it doesn't have to be a phpDoc example for every one of them. Maybe it could be some footnote or a link to the online documentation?

I'd like to leave the doc block in a good enough shape and proceed with the changes. I hope, someone could join later and improve the documentation as well.

morozov:
@Ocramius do you still have any concerns or we can proceed?

Ocramius (approved):
Let's proceed 👍

Issue #3472, opened 2019-02-26
getUnquotedStatementFragments no longer works for very long query since 2.8.1


fletanoux:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.8.1

#### Summary

Function getUnquotedStatementFragments no longer works properly if the sql (postgresql) request is too long/complex. As a result params passed to executeQuery/executeUpdate are no longer used and we are getting this error :

`"SQLSTATE[08P01]: <<Unknown error>>: 7 ERROR:  bind message supplies 0 parameters, but prepared statement "" requires 2"`

#### Previous behaviour

You could have an extremy long and complex query with several json_build_object imbricated into each other, and params would get replaced as intended.

#### Current behavior

WIth the exact same request, params are no longer replaced, and we get an error instead.

#### How to reproduce

Just make a very long query, with a lot of fields (it seems to bug above 40 fields) and some parameters to replace. Pass that query to a executeUpdate/Query method, and wait for the error to come. In my case the sql query is mostly composed of `json_build_object` imbricated in each other.

I've traced back the error to the method  getUnquotedStatementFragments in the file SQLParsertUtils.php which seemed have been modified in the following commit :

https://github.com/doctrine/dbal/commit/0a7d5f0c264e9cd2122a9efb6cd20ee0fe8ba1b9#diff-e7dd540cfb86c74ce7361e2bd00053cf

I've run the same test on the current version of DBAL so `2.9.2` but with the old code from 2.8.0 for the method getUnquotedStatementFragments and it works, so the problem is 100% there.

I've also found out that by increasing the PCRE's backtracking limit (         ini_set('pcre.backtrack_limit', 10000000);
) it also works with current code. (

SenseException:
The bugfix of #3277 probably changed recursion/backtrack. What's the size of your statement, that is injected into the mentioned method?

fletanoux:
The statement i used to do my test with is just over 15k characters long.

duvall:
Same problem here with a query of 17748 characters, I set up ini_set('pcre.backtrack_limit', 10000000); and it works just fine

SenseException:
A query this long is probably an edge case and it can be averted by the change of the pcre config value.

[...]

PR #3277, opened 2018-09-02
Lost parameters: wrong ESCAPED_BRACKET_QUOTED_TEXT

ohvitorino:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | #3140 

#### Summary

The first capturing group of the parser regex was stopping from the moment it would find a `[` character. I expanded the original capturing group to include `ARRAY[.*]` as a capturable expression.

PR #3497, opened 2019-03-27
Fix quotes in single-line comments breaking parameters

TPHRyan:
An odd number of quotes in comments would cause getUnquotedStatementFragments to start returning the inside of strings, rather than the outside.
Ignore comment lines to solve this problem.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Could not find existing issue

#### Summary

This PR ensures that comments are ignored when breaking a query into fragments. Without this PR, comments with quotes **(including apostrophes in words)** can arbitrarily break parameter substitution.
While ideal production code does not include comments, test code can, and will confuse people if something breaks because they're present.

When trying to test a query locally, I got an SQLSTATE error indicating there was a mismatch between positional parameters and the parameters provided.
Inspecting the query just before execution, I noted that not all of the parameters had been substituted, it seemed that the ones that were later in the query had not.
After stepping through, I found the culprit to be getUnquotedSegmentFragments, later on in the query it was returning short fragments such as "variable_foo" etc. (I recognised these as names of json fields that I was accessing)
**I realised that the pattern was such that it broke my comments into fragments, and then encountered the final quote in a comment and continued searching for its pair, then treating every following _opening_ quote as a _closing_ one!**

This is fixable by adding syntax for comments to the beginning of the RegEx, so they are skipped if found.
I also needed to add the "-" character to the list of banned characters preceding a quote, so that it would be interrupted by a comment. This doesn't seem to prevent important edge cases, but I'm happy if one is found to prove me wrong.

### DIFF
--- a/lib/Doctrine/DBAL/SQLParserUtils.php
+++ b/lib/Doctrine/DBAL/SQLParserUtils.php
@@ -254,7 +254,7 @@ class SQLParserUtils
         $expression = sprintf('/(?:[\s]*--.*?\n)|((.+(?i:ARRAY)\\[.+\\])|([^-\'"`\\[]+))(?:%s)?/s', $literal);
         preg_match_all($expression, $statement, $fragments, PREG_OFFSET_CAPTURE);
 
-        return array_filter($fragments[1], static function ($match) {
+        return array_filter($fragments[1], static function (string $match) : bool {
             return $match !== '';
         });
     }
--- a/lib/Doctrine/DBAL/SQLParserUtils.php
+++ b/lib/Doctrine/DBAL/SQLParserUtils.php
@@ -4,6 +4,7 @@ namespace Doctrine\DBAL;
 
 use const PREG_OFFSET_CAPTURE;
 use function array_fill;
+use function array_filter;
 use function array_key_exists;
 use function array_merge;
 use function array_slice;
@@ -250,11 +251,12 @@ class SQLParserUtils
             self::ESCAPED_DOUBLE_QUOTED_TEXT . '|' .
             self::ESCAPED_BACKTICK_QUOTED_TEXT . '|' .
             self::ESCAPED_BRACKET_QUOTED_TEXT;
-        $expression = sprintf('/((.+(?i:ARRAY)\\[.+\\])|([^\'"`\\[]+))(?:%s)?/s', $literal);
-
+        $expression = sprintf('/(?:[\s]*--.*?\n)|((.+(?i:ARRAY)\\[.+\\])|([^-\'"`\\[]+))(?:%s)?/s', $literal);
         preg_match_all($expression, $statement, $fragments, PREG_OFFSET_CAPTURE);
 
-        return $fragments[1];
+        return array_filter($fragments[1], static function ($match) {
+            return $match !== '';
+        });
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Ocramius (changes requested):
This kind of change requires test additions to back it

Ocramius:
Please also check https://travis-ci.org/doctrine/dbal/jobs/511822586

TPHRyan:
Apologies, I could not find contributer guidelines from this repo.

I have added a regression test for this bugfix and fixed formatting issues.

Ocramius:
> Apologies, I could not find contributer guidelines from this repo.

No worries, that's what CI is for on CS :-)

Ocramius (dismissed):

tests/Doctrine/Tests/DBAL/SQLParserUtilsTest.php
```diff
@@ -86,6 +86,25 @@ public function dataGetPlaceholderPositions()
...
+OR bar=':not_a_param2'
+SQLDATA
+                ,
+                false,
+                [
+                    29 => 'a_param1',
+                    145 => 'a_param2',
+                    162 => 'a_param3',
+                ],
+            ],
```
morozov: @TPHRyan could you add an extra data entry where a multi-line string literal would contain a line starting with `--`? E.g.:
```sql
INSERT INTO notes(text) values('
-- This is a comment
And this is the note itself
')
```

TPHRyan: Good point. It doesn't break with the current regex but I can see how it might.
Have pushed a further test for this case.

lib/Doctrine/DBAL/SQLParserUtils.php
```diff
@@ -250,11 +251,12 @@ private static function getUnquotedStatementFragments($statement)
             self::ESCAPED_DOUBLE_QUOTED_TEXT . '|' .
             self::ESCAPED_BACKTICK_QUOTED_TEXT . '|' .
             self::ESCAPED_BRACKET_QUOTED_TEXT;
-        $expression = sprintf('/((.+(?i:ARRAY)\\[.+\\])|([^\'"`\\[]+))(?:%s)?/s', $literal);
-
+        $expression = sprintf('/(?:[\s]*--.*?\n)|((.+(?i:ARRAY)\\[.+\\])|([^-\'"`\\[]+))(?:%s)?/s', $literal);
         preg_match_all($expression, $statement, $fragments, PREG_OFFSET_CAPTURE);
 
-        return $fragments[1];
+        return array_filter($fragments[1], static function ($match) {
```
morozov: Please use argument and return value types where possible (`function (string $match) : bool`).

tests/Doctrine/Tests/DBAL/SQLParserUtilsTest.php
```diff
@@ -86,6 +86,42 @@ public function dataGetPlaceholderPositions()
...
+            // Check comments in strings don't reverse it either
+            [
+                <<<'SQLDATA'
+INSERT INTO foo VALUES
+('
+-- Not a comment
```
morozov: What if this line ends the literal with a single quote? I.e.
```sql
INSERT INTO foo VALUES
('
-- Not a comment', :a_param1)
```

I'm trying to think of a use case similar to being fixed but inside of a string literal.

TPHRyan:
All requested changes have now been made.

morozov:
@TPHRyan the build is unhappy. Please fix the failures.

lib/Doctrine/DBAL/SQLParserUtils.php
```diff
@@ -250,11 +251,12 @@ private static function getUnquotedStatementFragments($statement)
             self::ESCAPED_DOUBLE_QUOTED_TEXT . '|' .
             self::ESCAPED_BACKTICK_QUOTED_TEXT . '|' .
             self::ESCAPED_BRACKET_QUOTED_TEXT;
-        $expression = sprintf('/((.+(?i:ARRAY)\\[.+\\])|([^\'"`\\[]+))(?:%s)?/s', $literal);
-
+        $expression = sprintf('/(?:[\s]*--.*?\n)|((.+(?i:ARRAY)\\[.+\\])|([^-\'"`\\[]+))(?:%s)?/s', $literal);
```
morozov: There's an issue (https://github.com/doctrine/dbal/issues/3472) caused by making this regular expression more complex (https://github.com/doctrine/dbal/pull/3277). I don't know of a proper solution but it seems like making it even more complex may make things even worse.

morozov:
As far as the parser is concerned, this is fixed in https://github.com/doctrine/dbal/pull/4397.

PR #3500, opened 2019-03-27
Replace self with static in fabric methods for ConversionException

nitso:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

 This change enables compatibility with inherited classes for fabric methods in base exception class.

I like to use `\Doctrine\DBAL\Types\ConversionException` factory methods when working with types. 
But for custom types I prefer to extend base exception class to get more specific behavior.

### DIFF
--- a/lib/Doctrine/DBAL/Types/ConversionException.php
+++ b/lib/Doctrine/DBAL/Types/ConversionException.php
@@ -30,7 +30,7 @@ class ConversionException extends DBALException
     {
         $value = strlen($value) > 32 ? substr($value, 0, 20) . '...' : $value;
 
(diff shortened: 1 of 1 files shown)


Ocramius:
Sorry, but this is an incorrect approach, since there is no guarantee that your child class constructor behaves like the parent one: please use your own implementation of these classes and static methods.

nitso:
Thanks for quick reply, I see the problem. 
Do you plan to refactor type system (\Doctrine\DBAL\Types\Type)  in future releases?

Ocramius:
Yes, but right now the BC break would be too massive. Please check `develop` for current `3.x` work

Majkl578:
Also see #2841.

PR #3501, opened 2019-03-29
Reworked AbstractSchemaManager::extractDoctrineTypeFromComment(), removed ::removeDoctrineTypeFromComment()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

Currently, `AbstractSchemaManager::extractDoctrineTypeFromComment()` and `::removeDoctrineTypeFromComment()` have to be called separately, one after the other, and duplicate implementation, however they are parts of a single whole: extracting a Doctrine type from the column comment.

The fact that the former requires passing `$currentType` makes it hard to use when the type is not yet known (see `SqliteSchemaManager`) and hard to guess if the type was actually extracted (it takes passing a knowingly wrong type and compare it with the result).

The proposed implementation makes the usage more straightforward. I would like to keep `testExtractDoctrineTypeFromComment` in place but I don't think the method is worth extraction into a separate class. Therefore, reflection is used in the test.

~It's still unclear to me whether column comments should be nullable or not in general (I'd love them to not be), but making a decision requires a larger effort which will be made later.~ It's actually not being currently changed.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
@@ -23,7 +23,6 @@ use function func_get_args;
 use function is_array;
 use function is_callable;
 use function preg_match;
-use function str_replace;
 use function strtolower;
 
 /**
@@ -1099,35 +1098,19 @@ abstract class AbstractSchemaManager
     }
 
     /**
-     * Given a table comment this method tries to extract a typehint for Doctrine Type, or returns
-     * the type given as default.
+     * Given a table comment this method tries to extract a type hint for Doctrine Type. If the type hint is found,
+     * it's removed from the comment.
      *
-     * @param string|null $comment
-     * @param string      $currentType
-     *
-     * @return string
+     * @return string|null The extracted Doctrine type or NULL of the type hint was not found.
      */
-    public function extractDoctrineTypeFromComment($comment, $currentType)
+    final protected function extractDoctrineTypeFromComment(?string &$comment) : ?string
     {
-        if ($comment !== null && preg_match('(\(DC2Type:(((?!\)).)+)\))', $comment, $match)) {
-            return $match[1];
-        }
-
-        return $currentType;
-    }
-
-    /**
-     * @param string|null $comment
-     * @param string|null $type
-     *
-     * @return string|null
-     */
-    public function removeDoctrineTypeFromComment($comment, $type)
-    {
-        if ($comment === null) {
+        if ($comment === null || ! preg_match('/(.*)\(DC2Type:(((?!\)).)+)\)(.*)/', $comment, $match)) {
             return null;
         }
 
-        return str_replace('(DC2Type:' . $type . ')', '', $comment);
+        $comment = $match[1] . $match[4];
+
+        return $match[2];
     }
 }
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade to 3.0
 
+## BC BREAK `AbstractSchemaManager::extractDoctrineTypeFromComment()` changed, `::removeDoctrineTypeFromComment()` removed
+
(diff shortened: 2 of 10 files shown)


lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -1097,35 +1096,19 @@ public function getSchemaSearchPaths()
...
-
-    /**
-     * @param string|null $comment
-     * @param string|null $type
-     *
-     * @return string|null
+     * @return string|null The extracted Doctrine type or NULL of the type hint was not found.
      */
-    public function removeDoctrineTypeFromComment($comment, $type)
+    protected function extractDoctrineTypeFromComment(?string &$comment) : ?string
```
morozov: For those who don't like arguments passed by reference, in this case, it's not much different than `array_shift()`.

Ocramius: I'd still rather not have them. What's the alternative that warrants such an API?

morozov: I don't understand the question. Could you rephrase?

morozov: The need for the API change was discovered during the introduction of type hints. An alternative might be to return a tuple of `($comment, $type)` which is bad from the static analysis standpoint or to have two methods which is weird because you'd do one thing in two steps.

There's nothing wrong about the comment being passed by reference and mutated. That's why I made it protected. A caller doesn't need to keep the unmodified version of the comment for future use since the Doctrine type hint contained in the original value is internal to DBAL and just should be removed from it.

I do understand the implications of passing arguments by reference but IMO it's justified in this case.

greg0ire: Should this method be `final`?

morozov: It’d make sense.

Majkl578: I don't like the side-effect of this method - the name is "extract" but it also removes the type part from the comment. This may or may not be expected by the caller.
Maybe at least using separate parameter, like `&$filteredComment` would be a good idea?

morozov: > I don't like the side-effect of this method - the name is "extract" but it also removes the type part from the comment. This may or may not be expected by the caller.

In my understanding, "extract" means exactly what the method does (think of a tooth or an intelligence asset extraction). You have the object of extraction as the result, and it's no longer where it was.

> Maybe at least using separate _out_ parameter, like `&$filteredComment = null` would be a good idea?

There's no need for that. See above:

> A caller doesn't need to keep the unmodified version of the comment for future use since the Doctrine type hint contained in the original value is internal to DBAL and just should be removed from it.

Majkl578: > You have the object of extraction as the result, and it's no longer where it was.

Hmm, it makes sense. Maybe it's just me but when a method is called _extract_, I'd expect the portion being returned and the input untouched.

morozov: > I'd expect the portion being returned and the input untouched.

I'd expect exactly the opposite:

image

It literally means "remove".

morozov: > Maybe at least using separate _out_ parameter, like `&$filteredComment = null` would be a good idea?

It makes the API just harder to use since the caller will eventually have to replace their original `$comment` with `$filteredComment` which the proposed API already does for them. Look at the calling code.

morozov:
@Ocramius, @Majkl578 while I understand your arguments, I don't find them sufficient for blocking the change or changing the proposed API. Let's move on, please.

Majkl578:
Hmm, I'm not a native speaker so my perception of this may be wrong. I don't really have anything else to say except the naming.

Majkl578 (approved):

PR #3505, opened 2019-04-05
Reworked driver exceptions

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

Besides the breaking changes necessitated by enforcing strict scalar types and adding return type hints, named constructors were added to `DB2Exception` and `MysqliException`.


lib/Doctrine/DBAL/Driver/DriverException.php
```diff
@@ -14,29 +14,10 @@
...
-
     /**
      * Returns the SQLSTATE the driver was in at the time the error occurred.
      *
      * Returns null if the driver does not provide a SQLSTATE for the error occurred.
-     *
-     * @return string|null
      */
-    public function getSQLState();
+    public function getSQLState() : ?string;
```
greg0ire: Shouldn't this BC-break be listed too?

morozov: Why is it a BC break?

Ocramius: Indeed, not a BC break: https://3v4l.org/uvKF4

Ocramius (approved):
:+1:

lib/Doctrine/DBAL/Driver/DriverException.php
Majkl578: It's a BC break for those who implemented that without typehint, i.e. in mocks.

UPGRADE.md
```diff
@@ -1,5 +1,13 @@
 # Upgrade to 3.0
 
+## BC BREAK Changes in driver exceptions
+
+1. The `Doctrine\DBAL\Driver\DriverException::getErrorCode()` method is removed. In order to obtain the driver error code, please use `::getCode()`.
+2. `Doctrine\DBAL\Driver\PDOException` no longer extends `PDOException`.
+3. The value returned by `Doctrine\DBAL\Driver\PDOException::getSQLState()` no longer falls back to the driver error code.
+
+The method was used internally and is no longer needed.
```
Majkl578: This shouldn't be here, looks like a copy&paste error.

lib/Doctrine/DBAL/Driver/DriverException.php
Ocramius: My mistake indeed: my snippet was missing something: https://3v4l.org/4933a

UPGRADE.md
morozov: Yes, incorrect conflict resolution.

lib/Doctrine/DBAL/Driver/DriverException.php
morozov: While it's a technically a BC break, it's not something worth noting in upgrade notes. Eventually, we'll have return type hints added across the entire codebase, so if documented explicitly, this list would become unreadably huge.

greg0ire: That's right, a global comment on release will do

PR #3507, opened 2019-04-06
Removed errorCode() and errorInfo() methods from Connection and Statement APIs

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

The reasons or removal are:

1. It's an archaism inherited from PDO and relevant only for the procedural type of API (`do_stuff()`, `get_last_error()`) which was only relevant for the `PDO::ERRMODE_WARNING` mode.
2. The current implementation is not portable: for DB2, the meaning of the code is SQLState, for MySQL it's the driver error code. Unlike others drivers where the error info is a tuple of error code and message, `sqlsrv` returns a list of such tuples.
3. This API is not covered with integration tests which proves #‌2.
4. Tuples are not supported natively by PHP which makes it harder to analyze these methods statically.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade to 3.0
 
+## BC BREAK `::errorCode()` and `::errorInfo()` removed from `Connection` and `Statement` APIs
+
+The error information is available in `DriverException` trown in case of an error.
+
 ## BC BREAK Changes in driver exceptions
 
 1. The `Doctrine\DBAL\Driver\DriverException::getErrorCode()` method is removed. In order to obtain the driver error code, please use `::getCode()`.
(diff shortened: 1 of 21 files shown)


Ocramius (approved):

UPGRADE.md
```diff
@@ -1,5 +1,9 @@
 # Upgrade to 3.0
 
+## BC BREAK `::errorCode()` and `::errorInfo()` removed from `Connection` and `Statement` APIs
+
+The error information is available in `DriverException` trown in case of an error.
```
Majkl578: "thrown" 😛

morozov: We need a spellchecker on CI?

PR #3511, opened 2019-04-07
Added type hints to the Column class

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Column.php
+++ b/lib/Doctrine/DBAL/Schema/Column.php
@@ -7,7 +7,6 @@ namespace Doctrine\DBAL\Schema;
 use Doctrine\DBAL\Types\Type;
 use const E_USER_DEPRECATED;
 use function array_merge;
-use function is_numeric;
 use function method_exists;
 use function sprintf;
 use function trigger_error;
@@ -21,9 +20,9 @@ class Column extends AbstractAsset
     protected $_type;
 
     /** @var int|null */
-    protected $_length = null;
+    protected $_length;
 
-    /** @var int */
+    /** @var int|null */
     protected $_precision = 10;
 
     /** @var int */
@@ -38,8 +37,8 @@ class Column extends AbstractAsset
     /** @var bool */
     protected $_notnull = true;
 
-    /** @var string|null */
-    protected $_default = null;
+    /** @var mixed */
+    protected $_default;
 
     /** @var bool */
     protected $_autoincrement = false;
@@ -48,10 +47,10 @@ class Column extends AbstractAsset
     protected $_platformOptions = [];
 
     /** @var string|null */
-    protected $_columnDefinition = null;
+    protected $_columnDefinition;
 
     /** @var string|null */
-    protected $_comment = null;
+    protected $_comment;
 
     /** @var mixed[] */
     protected $_customSchemaOptions = [];
@@ -59,22 +58,19 @@ class Column extends AbstractAsset
     /**
      * Creates a new Column.
      *
-     * @param string  $columnName
      * @param mixed[] $options
      */
-    public function __construct($columnName, Type $type, array $options = [])
+    public function __construct(string $name, Type $type, array $options = [])
     {
-        $this->_setName($columnName);
+        $this->_setName($name);
         $this->setType($type);
         $this->setOptions($options);
     }
 
     /**
      * @param mixed[] $options
-     *
-     * @return Column
      */
-    public function setOptions(array $options)
+    public function setOptions(array $options) : self
     {
         foreach ($options as $name => $value) {
             $method = 'set' . $name;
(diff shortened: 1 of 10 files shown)


lib/Doctrine/DBAL/Schema/Column.php
```diff
@@ -21,13 +20,13 @@ class Column extends AbstractAsset
     protected $_type;
 
     /** @var int|null */
-    protected $_length = null;
+    protected $_length;
 
     /** @var int */
-    protected $_precision = 10;
```
Ocramius: Not sure if this should be changed

lib/Doctrine/DBAL/Schema/Column.php
```diff
@@ -21,13 +20,13 @@ class Column extends AbstractAsset
...
+    protected $_precision;
 
     /** @var int */
-    protected $_scale = 0;
```
Ocramius: Same here: default value to be preserved IMO

Ocramius (dismissed):

lib/Doctrine/DBAL/Schema/Column.php
Majkl578: As precision is only applicable for numerical columns, I think it does make sense to drop the default.

Majkl578: Another solution could be moving this information to Type.

lib/Doctrine/DBAL/Schema/Column.php
```diff
@@ -94,106 +90,59 @@ public function setOptions(array $options)
...
-     */
-    public function setPrecision($precision)
+    public function setPrecision(?int $precision) : self
     {
-        if (! is_numeric($precision)) {
-            $precision = 10; // defaults to 10 when no valid precision is given.
-        }
-
-        $this->_precision = (int) $precision;
+        $this->_precision = $precision;
```
Majkl578: Should we check for `$precision >= 0`?

lib/Doctrine/DBAL/Schema/Column.php
```diff
@@ -94,106 +90,59 @@ public function setOptions(array $options)
...
-     */
-    public function setScale($scale)
+    public function setScale(?int $scale) : self
     {
-        if (! is_numeric($scale)) {
-            $scale = 0;
-        }
-
-        $this->_scale = (int) $scale;
+        $this->_scale = $scale;
```
Majkl578: Should we check for `$scale >= 0`?

lib/Doctrine/DBAL/Schema/Column.php
```diff
@@ -94,106 +90,59 @@ public function setOptions(array $options)
...
-     */
-    public function setLength($length)
+    public function setLength(?int $length) : self
     {
-        if ($length !== null) {
-            $this->_length = (int) $length;
-        } else {
-            $this->_length = null;
-        }
+        $this->_length = $length;
```
Majkl578: Should we check for `$length > 0`?

lib/Doctrine/DBAL/Schema/Column.php
morozov: This is a DTO which just carries what's given to it. It shouldn't imply any defaults. Whatever owns the definition (DB schema or the definition in the application) should define it explicitly.

I'll add it to the upgrade notes.

lib/Doctrine/DBAL/Schema/Column.php
morozov: I don't think so:

1. It's out of scope.
2. The class is just a DTO, not a domain object. DBAL doesn't do any handling of the precision AFAIK, and the DB will validate it itself.
3. In the domain, the value is not only bounded from at bottom (is it?), it also may be bounded at the top which we know nothing about at the abstraction layer.
4. The max value may be dependent on other values (e.g. on Oracle, the sum of precision and scale can't be greater than 38).

jwage:
Can we (should we?) separate the BC breaks due to adding type hints and BC breaks from us changing default behavior? IMO, while changing the defaults may be a good idea long term, I am not sure if making the change along side adding type hints is good. May be better to have 1 PR to add the type hints while maintaining behavior and then another PR to change the defaults. This way the defaults changing can be discussed in isolation from all the other changes and the type hints can probably be merged faster.

morozov:
@jwage I get the intent but if we want to take this route, the changes have to be implemented in the reverse order:

1. Drop the defaults (the real BC break)
2. Enforce strict argument types (a break only for API implementors, not consumers).

The existing defaults only make sense with non-enforced argument types (hence the `is_numeric()` checks). If we introduce the strict types first, it will implicitly break the behavior covered by `is_numeric()`.

I like the idea of the split in general. Especially, given that there are still some leftovers of the default value of `10` in `Comparator`. I'll open a separate PR.

lib/Doctrine/DBAL/Schema/Column.php
morozov: Reverted from this patch.

lib/Doctrine/DBAL/Schema/Column.php
morozov: Reverted from this patch.

jwage (approved):

PR #535, opened 2014-02-28
added backtrace to query logging

dmecke:
I've added a stacktrace to each logged query to be able to display it later on. This can help to find out why a query has been executed.
This one is needed for pull request https://github.com/doctrine/DoctrineBundle/pull/275 in doctrine/doctrineBundle.

PR #3513, opened 2019-04-13
Add query backtrace logger

ottaviano:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | ~

#### Summary
inspired by [doctrine-stats](https://github.com/steevanb/doctrine-stats) I think it'll be nice to log backtrace of each query.

use case with doctrine-bundle:
image

[doctrine-bundle pull request](https://github.com/doctrine/DoctrineBundle/pull/954)

### DIFF
--- a/lib/Doctrine/DBAL/Logging/BacktraceLogger.php
+++ b/lib/Doctrine/DBAL/Logging/BacktraceLogger.php
@@ -11,14 +11,6 @@ class BacktraceLogger extends DebugStack
     {
         parent::startQuery($sql, $params, $types);
 
-        $this->queries[$this->currentQuery]['backtrace'] = array_reverse(debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS));
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function stopQuery()
-    {
-        parent::stopQuery();
+        $this->queries[$this->currentQuery]['backtrace'] = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
     }
 }
--- a/lib/Doctrine/DBAL/Logging/BacktraceLogger.php
+++ b/lib/Doctrine/DBAL/Logging/BacktraceLogger.php
@@ -2,6 +2,9 @@
 
 namespace Doctrine\DBAL\Logging;
 
+use const DEBUG_BACKTRACE_IGNORE_ARGS;
+use function debug_backtrace;
+
 class BacktraceLogger extends DebugStack
 {
     /**
--- a/lib/Doctrine/DBAL/Logging/BacktraceLogger.php
+++ b/lib/Doctrine/DBAL/Logging/BacktraceLogger.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace Doctrine\DBAL\Logging;
+
+class BacktraceLogger extends DebugStack
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function startQuery($sql, ?array $params = null, ?array $types = null)
+    {
+        parent::startQuery($sql, $params, $types);
+
+        $this->queries[$this->currentQuery]['backtrace'] = array_reverse(debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function stopQuery()
+    {
+        parent::stopQuery();
+    }
+}
--- a/lib/Doctrine/DBAL/Logging/BacktraceLogger.php
+++ b/lib/Doctrine/DBAL/Logging/BacktraceLogger.php
@@ -3,6 +3,7 @@
 namespace Doctrine\DBAL\Logging;
 
 use const DEBUG_BACKTRACE_IGNORE_ARGS;
+use function array_shift;
 use function debug_backtrace;
 
 class BacktraceLogger extends DebugStack
@@ -14,6 +15,11 @@ class BacktraceLogger extends DebugStack
     {
         parent::startQuery($sql, $params, $types);
 
-        $this->queries[$this->currentQuery]['backtrace'] = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
+        $backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
+
+        // skip first since it's always the current method
+        array_shift($backtrace);
+
+        $this->queries[$this->currentQuery]['backtrace'] = $backtrace;
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Logging/BacktraceLoggerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Logging/BacktraceLoggerTest.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Logging;
+
+use Doctrine\DBAL\Logging\BacktraceLogger;
+use Doctrine\Tests\DbalTestCase;
+
+class BacktraceLoggerTest extends DbalTestCase
+{
+    public function testBacktraceLogged()
+    {
+        $logger = new BacktraceLogger();
+
+        $logger->startQuery('SELECT column FROM table');
+
+        $currentQuery = current($logger->queries);
+
+        self::assertSame('SELECT column FROM table', $currentQuery['sql']);
+        self::assertNull($currentQuery['params']);
+        self::assertNull($currentQuery['types']);
+        self::assertIsArray($currentQuery['backtrace']);
+    }
+}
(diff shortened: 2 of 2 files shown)


lib/Doctrine/DBAL/Logging/BacktraceLogger.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Doctrine\DBAL\Logging;
+
+class BacktraceLogger extends DebugStack
```
Ocramius: I'd avoid coupling the `BacktraceLogger` to the existing `DebugStack` one, if possible

greg0ire:
Links to previous attempts: #3390 , #535 , #719 cc @pierredup @carnage @dmecke . People want that feature badly, it would be great to have a list of requirements for its implementation somewhere.

ottaviano:
Thank you @greg0ire for this summary 👍 

1. `debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS)` ignores all arguments and doesn't keep any object in memory. Am I wrong, @stof?

2. honesty I don't really see how to do it :) with DebugStack public properties 

3. cf point n1, as the backtrace contains only scalar values, it's completely serializable

4. I'm not sure I understand this point

morozov:
Regarding #‌4 and https://github.com/doctrine/dbal/pull/719#issuecomment-345302849, after trying to implement a PSR-3 adapter (https://github.com/doctrine/dbal/pull/3156) and later reading an [article on Domain-Oriented Observability](https://martinfowler.com/articles/domain-oriented-observability.html), I realized that our logger is not really a logger. It's a probe. It allows plugging in loggers, metrics and their combinations.

I'm still not convinced that we should be adding any new implementations to the stack. As long as the interface allows for extensibility, the needed functionality can be part of the application which needs it or a separate package.

stof:
@morozov well, if we use that feature in DoctrineBundle itself, it means that the implementation would be maintained by Doctrine anyway. Then, having it in DBAL itself means that other framework integrations can also use it directly instead of having to copy it from DoctrineBundle.

lib/Doctrine/DBAL/Logging/BacktraceLogger.php
stof: well, the whole point here is to add the backtrace into the data stored for each query. Not coupling it would defeat the point.

lib/Doctrine/DBAL/Logging/BacktraceLogger.php
```diff
@@ -0,0 +1,24 @@
...
+
+        $this->queries[$this->currentQuery]['backtrace'] = array_reverse(debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function stopQuery()
+    {
+        parent::stopQuery();
```
stof: this is useless code. you could remove the method override entirely.

lib/Doctrine/DBAL/Logging/BacktraceLogger.php
```diff
@@ -0,0 +1,24 @@
...
+class BacktraceLogger extends DebugStack
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function startQuery($sql, ?array $params = null, ?array $types = null)
+    {
+        parent::startQuery($sql, $params, $types);
+
+        $this->queries[$this->currentQuery]['backtrace'] = array_reverse(debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS));
```
stof: should we actually reverse the backtrace ? This could be confusing for users, as it would display it in the reverse order compared to what they are used to.

alcaeus: I agree. Backtraces always show the most recent call first and we should keep it that way.

ottaviano: agree, removed 👍

lib/Doctrine/DBAL/Logging/BacktraceLogger.php
ottaviano: removed

greg0ire:
> honesty I don't really see how to do it :) with DebugStack public properties

@Ocramius can you maybe help with this? I don't understand what you are suggesting either (composition, copy/paste, something else entirely, no suggestion?)

Ocramius:
I'm suggesting c&p and avoiding the inheritance

ostrolucky (approved):
LGTM. inheritance > c&p

Ocramius:
Disagree massively: adding inheritance reduces our ability to introduce international changes. If inheritance is not also introducing added "meaning", it is absolutely **NEVER** to be used as a code reuse mechanism.

greg0ire:
If this goes towards copy and paste, would you say public properties should be kept, or should encapsulation be used this time?

Ocramius:
Considering how widespread the library is, I'd at least use a safer API around here, because properties are invariants, therefore they expose a BC surface both for reads and writes.

In other words: yes, we'd need some sort of accessors/reset methods.

Ocramius:
Thinking more about API and state, @morozov got to the right thought at https://github.com/doctrine/dbal/pull/3513#issuecomment-483062136

The logger being the primary reason why doctrine has memleaks in symfony-related setuos (think profiler), maybe it is a good idea if we didn't even have a logger ourselves at all, and instead only provided the hook mechanism?

greg0ire:
Finally got around to reading the article linked by @morozov , that was really a great read! Just to be clear, are you saying that the `DebugStack` class should be deprecated, and that the code in this PR should be moved to the doctrine bundle, making that part of the code a bit more Symfony-owned?

morozov:
> Just to be clear, are you saying that the `DebugStack` class should be deprecated, and that the code in this PR should be moved to the doctrine bundle, making that part of the code a bit more Symfony-owned (although it would still be in the Doctrine org)?

I didn't think of it in these details but sounds right. I believe everyone would benefit from having code closer to the owner in order to avoid unproductive discussions and wasted effort.

I do believe that neither `DebugStack` nor backtrace logger belong to the DBAL because they solve the application developer experience problems, not the DBAL ones. In a longer run, a good DX may involve adding some more details to the bracktrace which would again require changes in DBAL which would trigger the same round of discussions.

On the other hand, in my own experience of developing DBAL, having `DebugStack` enabled in the test suite doesn't add _any_ value and just creates noise. Unlike an application, where failure is likely to be caused by a sequence of SQL queries, in the DBAL test suite, a test will fail because a specific query failed and the list of previously executed SQL queries is irrelevant.

greg0ire:
Looks like we have reached an agreement :) @ottaviano, I think you can close this PR and make your bundle PR independent from it by adding a new SqlLogger implementation to it.

Ocramius:
Closing here as per https://github.com/doctrine/dbal/pull/3513#issuecomment-487394513

alcaeus:
For the record, while I agree with the semantics surrounding `DebugStack`, its very unfortunate public API (public properties, no clear structure for logged queries, etc.) I don't agree that this shouldn't be in DBAL. If we move this into DoctrineBundle, we'll also have to have it in the corresponding module for Zend Framework or any other framework integrations. People not using a framework integration will have to duplicate and maintain the code for this logic themselves, which is even more unfortunate than having it here.

I'm all for cleaning up the API and dropping the current `SQLLogger` interface in 3.0, but I don't think not adding new functionality because the current API is not as good as we'd like to have it is unfortunate.

Regarding decoupling from `DebugStack`, this is not possible unfortunately. Since the `SQLLogger` interface does not specify how something is logged/can be retrieved from the logger, any packages using such instrumentation have to type hint against a concrete implementation. In the case of the Symfony Doctrine bridge, the data collectors that make use of SQL logging are coded against `DebugStack`, meaning that any additional loggers would always have to extend the `DebugStack` class.

morozov:
> If we move this into DoctrineBundle, we'll also have to have it in the corresponding module for Zend Framework or any other framework integrations.

Which is natural. As long as observability logic can be framework-specific, it's fine to have separate implementations in each bundle/module.

> People not using a framework integration will have to duplicate and maintain the code for this logic themselves, which is even more unfortunate than having it here.

Which is also fine. People may have different needs than frameworks. Besides that, the current implementation of `DebugStack` primitive but not extensible. It won't be a maintenance burden for the end developer but is for us.

> Regarding decoupling from `DebugStack`, this is not possible unfortunately. Since the `SQLLogger` interface does not specify how something is logged/can be retrieved from the logger, any packages using such instrumentation have to type hint against a concrete implementation. In the case of the Symfony Doctrine bridge, the data collectors that make use of SQL logging are coded against `DebugStack`, meaning that any additional loggers would always have to extend the `DebugStack` class.

Frameworks may define their own interfaces which extend `SQLLogger` and define the missing APIs.

Issue #2315, opened 2016-02-15
pdo_mysql with persistent connection


jeff1985:
While trying to use doctrine with persistent pdo_mysql connections, it seemed that this is not supported. 
To be clear, when you pass `PDO::ATTR_PERSISTENT => true` as config option for connection, an exception will be raised stating that `General error: PDO::ATTR_STATEMENT_CLASS cannot be used with persistent PDO instances`.

After some research, i found [this answer on stackoverflow](http://stackoverflow.com/questions/16217426/is-it-possible-to-use-doctrine-with-persistent-pdo-connections) suggesting to pass a pre-created PDO instance as configuration parameter like this:

```
$conn = array('pdo' => $dbh);`
$this->entityManager = EntityManager::create($conn, $config);
```

It does work and my application is able use doctrine as usual. BUT it feels strange. My understanding is that this way we fool doctrine to use pdo without the statement class (which is by the way `Doctrine\DBAL\Driver\PDOStatement`). Can someone please point out, if there are any risks of using doctrine like that?

I also suggest to add the corresponding info regarding persistent connection to the docs.

Ocramius:
> this way we fool doctrine

Passing in a PDO instance is the correct way to handle edge cases in the configuration DSL of the [`DriverManager`](https://github.com/doctrine/dbal/blob/625dd63568e386e11c01022c22e50adc2a8a7af6/lib/Doctrine/DBAL/DriverManager.php#L132), actually.

jeff1985:
Yes, but do you need a pdo instance configured to work with `Doctrine\DBAL\Driver\PDOStatement`? If you don't, then what is the purpose of this class?

Ocramius:
@jeff1985 doctrine wraps around a number of drivers, and many of them are not PDO-based, so we built internal abstractions to get rid of that impedance.

jeff1985:
@Ocramius 
OK, i understand, why you've built a wrapper class.  

You write, that `Doctrine\DBAL\Driver\PDOStatement` would still be used. How do you mean that? The whole point of providing an own pdo instance is not to be using this class. Because custom statement class is not supported by PDO for persistent connections. 

So the actual question is will using the plain `\PDOStatement` class break any of doctrine code? So far i can see the only change that this class does is to wrap  \PDOException with `Doctrine\DBAL\Driver\PDOException`. As far as i can see, the code still uses  \PDOException everywhere, so the change actually would not break anything.

Ocramius:
[...]

PR #3515, opened 2019-04-14
Implemented support for persistent connections in PDO and mysqli drivers

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #2315

As of #3080, DBAL doesn't use a custom PDO statement class which enables the use of persistent connections with PDO drivers. Additionally, the support for persistent connections has been implemented for the `mysqli` driver.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
+++ b/lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
@@ -56,6 +56,11 @@ class MysqliConnection implements Connection, PingableConnection, ServerInfoAwar
 
         $socket = $params['unix_socket'] ?? ini_get('mysqli.default_socket');
         $dbname = $params['dbname'] ?? '';
+        $host   = $params['host'];
+
+        if (! empty($params['persistent'])) {
+            $host = 'p:' . $host;
+        }
 
         $flags = $driverOptions[static::OPTION_FLAGS] ?? 0;
 
@@ -67,7 +72,7 @@ class MysqliConnection implements Connection, PingableConnection, ServerInfoAwar
         set_error_handler(static function () {
         });
         try {
-            if (! $this->conn->real_connect($params['host'], $username, $password, $dbname, $port, $socket, $flags)) {
+            if (! $this->conn->real_connect($host, $username, $password, $dbname, $port, $socket, $flags)) {
                 throw MysqliException::fromConnectionError($this->conn);
             }
         } finally {
--- a/tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
@@ -7,11 +7,16 @@ namespace Doctrine\Tests\DBAL\Functional;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\ConnectionException;
 use Doctrine\DBAL\Driver\Connection as DriverConnection;
+use Doctrine\DBAL\Driver\PDOConnection;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Platforms\SqlitePlatform;
+use Doctrine\DBAL\Platforms\SQLServerPlatform;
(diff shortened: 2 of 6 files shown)


morozov:
At the moment, I don't see a reliable way to test the persistence of a connection across supported platforms/drivers.

In pseudo-code, a test may look like following:
```php
$conn = connect($persistent = false);
$id1 = $conn->id;

$conn = connect($persistent = true);
$id2 = $conn->id;

assertNotEquals($id1, $id2);

$conn = connect($persistent = true);
$id3 = $conn->id;

assertEquals($id2, $id3);
```

The results are:
1. The approach above works for MySQL (`SELECT CONNECTION_ID()`) and PostgreSQL (`SELECT pg_backend_pid()`).
2. SQL Server and Oracle return the same ID even for non-persistent connections (`SELECT @@SPID` and `SELECT SYS_CONTEXT('USERENV', 'SID') FROM DUAL` respectively).
3. I didn't find a way to obtain the connection ID for IBM DB2 and SQL Anywhere.

Majkl578:
Can we at least test that persistent connection flag is active when `persistent` parameter is enabled?
For PDO, it can be done using `PDO::getAttribute()`. For mysqli, it can be done using `mysqli::get_connection_stats()`.

morozov:
> Can we at least test that persistent connection flag is active when `persistent` parameter is enabled?
For PDO, it can be done using `PDO::getAttribute()`.

Good idea. Done.

> For mysqli, it can be done using `mysqli::get_connection_stats()`.

If we use the `active_persistent_connections` key from there, the test may pass if there is a persistent connection in the server, not necessarily the current one. Let's keep the test skipped for non-PDO drivers.

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
```diff
@@ -313,4 +318,28 @@ public function testDeterminesDatabasePlatformWhenConnectingToNonExistentDatabas
...
+        $connection       = DriverManager::getConnection($params);
+        $driverConnection = $connection->getWrappedConnection();
+
+        if (! $driverConnection instanceof PDOConnection) {
+            self::markTestSkipped('Unable to test if the connection is persistent');
+        }
+
+        $pdo = $driverConnection->getWrappedConnection();
+
+        $this->assertTrue($pdo->getAttribute(PDO::ATTR_PERSISTENT));
```
Majkl578: `self::`

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
```diff
@@ -56,6 +56,11 @@ public function __construct(array $params, $username, $password, array $driverOp
 
         $socket = $params['unix_socket'] ?? ini_get('mysqli.default_socket');
         $dbname = $params['dbname'] ?? '';
+        $host   = $params['host'];
+
+        if (! empty($params['persistent'])) {
```
Majkl578: I'd prefer `array_key_exists()` over more dangerous `! empty()`.

morozov: In this case it's irrelevant. `array_key_exists()` over a mistyped key will have the same effect.

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
morozov: I wish it was part of the standard.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
morozov: I'm thinking of introducing a `ConnectionParameters` object which will be just a typed structure which will also contain username and password if they are specified. It could avoid optional `$username` and `$pasword` in driver/connection methods. Any suggestions/ideas?

jwage (approved):
Nice! quick and easy win :+1:

dasmfm:
Hi, @morozov, can you please backport this fix to upcoming 2.9.3 or 2.10.0?
It looks like small, but very useful feature.

morozov:
No, it depends on #3080 which is a breaking change.

Issue #3194, opened 2018-06-20
Doctrine\DBAL\Connection issue with "platform" key of $params array


Perf:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.7.1

#### Summary

I have a Symfony 4.1 project with installed Doctrine packages.
I found an issue, when trying to use my custom database Platform class, which I created to workaround another issue with default MySql Platform class - when you create database it's not adding ` CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci` to create table statement.

I created class with just one method:
```php
namespace App\Doctrine\DBAL\Platforms;

class MySQL57Platform extends \Doctrine\DBAL\Platforms\MySQL57Platform
{
    public function getCreateDatabaseSQL($name)
    {
        return parent::getCreateDatabaseSQL($name).' CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci';
    }
}
```
Then added it to doctrine configuration:
```yml
doctrine:
    dbal:
        ...
        platform_service: App\Doctrine\DBAL\Platforms\MySQL57Platform
        ...
```
When trying to execute `doctrine:database:create` command it creates database with default character set and collation, but should with `utf8mb4`.
I performed a debug session with script `\Doctrine\Bundle\DoctrineBundle\Command\CreateDatabaseDoctrineCommand` and found that:
- in `execute()` method at line 47 `$connection = $this->getDoctrineConnection($connectionName);` return correct connection object with correct `platform` property set.
- at line 51 `$params = $connection->getParams();` and `$params` doesn't contain `platform` key
- at line 83 `$tmpConnection = DriverManager::getConnection($params);` returns connection object without `platform` property set.

Tbh, I don't know why it is necessary to get another `$tmpConnection` instead of using already set `$connection`, but new connection object is created without proper platform, as `$params` doesn't have key `platform`.

#### Current behavior

Command `doctrine:database:create` doesn't use configured `platform_service` object.

#### How to reproduce

Create custom platform class (see above) and configure doctrine dbal to use it via parameter `platform_service`.
Execute command `doctrine:database:create` and check created database. It will be created with default character set and collation, instead of enforced `utf8mb4` in custom Platform object.

#### Expected behavior

Any creation of a connection object should respect configured value of `platform_service`.
As a possible fix in `\Doctrine\DBAL\Connection::__construct` at line 234:
- remove code `unset($this->_params["platform"]);`

[...]

PR #3521, opened 2019-04-17
Maintain platform parameter in connection params

jwage:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3194

#### Summary

Maintain platform parameter in connection params but remove it before using params in query cache key generation.

We still have the same issue with the `pdo` parameter being unset in `Connection::__construct()` but we're not going to deal with that in this PR.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -200,7 +200,6 @@ class Connection implements DriverConnection
             }
 
             $this->platform = $params['platform'];
-            unset($this->params['platform']);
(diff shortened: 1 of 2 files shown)


morozov (approved):

morozov:
In the longer term, we probably should get rid of the "platform" as a connection parameter. A connection is a transport level entity while the platform represents the server on the other end of that connection. The platform can be detected from the connection but doesn't define the connection itself. Mixing these concerns together creates the issue.

Issue #3516, opened 2019-04-15
Invalid last inserted id when the table has a trigger with insert statements in SQLSrvConnection


sarven:
### Bug Report
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.9.0

#### Summary

I discovered a problem when the table in MSSQL has a trigger with insert statements. Id returned by method lastInsertId is invalid. It isn't id from the table in where we were inserting data, but from the table that trigger was inserting data.

#### Current behaviour

Like I wrote in previous paragraph, currently when we have the table with a trigger inserting some data in another table, returned last Inserted Id is related with another table, not the main one that we expect.

#### Expected behaviour

Returned last inserted ID should be related to the main insert statement, not that executed in a trigger.

Ocramius:
@sarven is this even achievable with the current MSSQL clients? What extension are you using to connect to the DB?

sarven:
@Ocramius Yes, I have current MSSQL clients and PHP 7.3. I added draft PR with a simple test for this issue, I will try to make proper implementation to fix it.

arnegroskurth:
I ran into the same problem and as long as it is not fixed in DBAL I helped myself with a dirty trick:

Create a dummy table and extend your trigger to write the original id into this table at the end of the trigger execution:

```sql
CREATE TABLE [fake_last_inserted_id] ([id] INT IDENTITY(1, 1) PRIMARY KEY);

CREATE TRIGGER [my_trigger] ON [some_table] FOR INSERT AS BEGIN
    
   ...

   -- fake last inserted id
    SET IDENTITY_INSERT fake_last_inserted_id ON;
    DELETE FROM fake_last_inserted_id; -- prevent collision
    INSERT INTO fake_last_inserted_id ([id]) VALUES ([inserted].[id]); -- adjust id column name
    SET IDENTITY_INSERT fake_last_inserted_id OFF;
END;
```

sarven:
@arnegroskurth I think that I created the proper solution for this problem, but it is still waiting for review. In my situation extending and modifying triggers/tables are not good practice, because I work with the database that is using by a foreign application. So currently after insert I just fetch IDENT_CURRENT. It isn't good, but works.

mherderich:
stumbled upon this issue. i guess following change broke it: cadd79c
prior to this SCOPE_IDENTITY was used as following:

`;SELECT SCOPE_IDENTITY() AS LastInsertId;`

[...]

PR #3484, opened 2019-03-15
Use latest stable versions of sqlsrv and pdo_sqlsrv on CI

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

1. [`sqlsrv` and `pdo_sqlsrv` v5.6.0](https://github.com/Microsoft/msphpsql/releases/tag/v5.6.0) are compatible with all supported PHP versions.
2. [`xdebug` v2.7.0](https://github.com/xdebug/xdebug/releases/tag/2.7.0) is available.
3. `sqlsrv` on AppVeyor is temporarily rolled back to `5.5.0preview` due to (https://github.com/Microsoft/msphpsql/issues/955).

PR #3523, opened 2019-04-18
Fix returning the proper last inserted id in MSSQL

sarven:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | #3516 

#### Summary
I discovered a problem when the table in MSSQL has a trigger with insert statements. Id returned by method lastInsertId is invalid. It isn't id from the table in where we were inserting data, but from the table that trigger was inserting data.

The problem concerns sqlsrv and pdo_sqlsrv extensions. The fix for sqlsrv was quite simple, I only added a few lines of code. 

But the fix for pdo_sqlsrv required to rebuild the implementation of Connection and Statement. I made it similar to the sqlsrv implementation.


tests/Doctrine/Tests/DBAL/Functional/Driver/PDOSqlsrv/DriverTest.php
```diff
@@ -64,11 +63,4 @@ public function testConnectionOptions() : void
 
         self::assertSame('APP_NAME', $result);
     }
-
-    public function testDriverOptions() : void
```
SenseException: Is this test breaking with your changes?

sarven: Yes, because previously Connection extended PDOConnection, now it only implements Connection interface, where is no method like getAttribute.

SenseException: master-branch is currently for 2.10 while develop is for 3.0. It looks like the changes are more fitting for the next major release.

sarven: So should I make a branch based on develop then apply these changes and create a new pull request or just override this?

morozov:
@sarven could you rebase this please?

sarven:
@morozov Done.

.appveyor.yml
```diff
@@ -68,7 +68,7 @@ install:
           Add-Content php.ini "`n extension=php_curl.dll"
           Add-Content php.ini "`n curl.cainfo=C:\tools\cacert\bundle.pem"
 
-          $DLLVersion = "5.5.0preview"
+          $DLLVersion = "5.6.1"
```
morozov: Please revert unless this is required. If possible, we should roll back to using the latest stable version as it was prior to #3484 but that would be a different PR.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Statement.php
```diff
@@ -1,37 +0,0 @@
...
-namespace Doctrine\DBAL\Driver\PDOSqlsrv;
-
-use Doctrine\DBAL\Driver\PDOStatement;
-use Doctrine\DBAL\ParameterType;
-use PDO;
-
-/**
- * PDO SQL Server Statement
- */
-class Statement extends PDOStatement
```
morozov: Why does this need to be renamed?

morozov (commented):
@sarven thank you. Could you please describe the problem in the pull request description?

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL3516Test.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+class DBAL3516Test extends DbalFunctionalTestCase
+{
+    protected function setUp() : void
+    {
+        parent::setUp();
+
+        $platform = $this->connection->getDatabasePlatform()->getName();
+
+        if ($platform !== 'mssql') {
```
morozov: Since this is not the only driver/platform which implements `lastInsertId()`, the test should be implemented for all supported platforms and skipped if there's a known issue with a specific platform.

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL3516Test.php
```diff
@@ -0,0 +1,46 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL\Functional\Ticket;
+
+use Doctrine\Tests\DbalFunctionalTestCase;
+
+/**
+ * @group DBAL-3516
+ */
+class DBAL3516Test extends DbalFunctionalTestCase
```
morozov: Please move it under `Doctrine\Tests\DBAL\Functional\ConnectionTest`.

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
```diff
@@ -373,7 +374,9 @@ public function testExecWithRedundantParameters() : void
      */
     public function testFetchColumnNonExistingIndex(int $index) : void
     {
-        if ($this->connection->getWrappedConnection() instanceof PDOConnection) {
+        $wrappedConnection = $this->connection->getWrappedConnection();
+
+        if ($wrappedConnection instanceof PDOConnection || $wrappedConnection instanceof PDOSqlsrvConnection) {
```
morozov: Why isn't `PDOSqlsrvConnection` sub-type of `PDOConnection`?

.appveyor.yml
sarven: @morozov I reverted, but probably some tests won't pass.

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL3516Test.php
sarven: I thought about make this tests for all platforms, but there is a problem with create trigger for all platforms. So I think that creating it only for MSSQL is ok, because this problem probably doesn't concern others platforms.

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL3516Test.php
sarven: Ok

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Statement.php
sarven: I just wanted to name it similarly to SQLSrvStatement. It is not subtype of PDOStatement, so I wanted to mark it clearly.

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
sarven: I did some changes and now Connection and Statement in PDOSqlsrv are subtypes of specific doctrine PDO classes. Previously I wanted it to be similarly to the Sqlsrv implementation, but it is also possible to just extends PDO classes and do some overrides. Now it seems to be simpler.

.appveyor.yml
sarven: @morozv Now in appveyor is the error:
`
PHP Fatal error:  Meta data vector out of sync in
`
Should I create a new pull request with a change of version?

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
```diff
@@ -14,19 +17,31 @@
  */
 class Connection extends PDOConnection
 {
+    /** @var LastInsertId|null */
+    protected $lastInsertId;
+
+    /**
+     * Append to any INSERT query to retrieve the last insert id.
+     */
+    protected const LAST_INSERT_ID_SQL = ';SELECT SCOPE_IDENTITY() AS LastInsertId;';
```
Ocramius: `private` unless needed elsewhere

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
```diff
@@ -14,19 +17,31 @@
  */
 class Connection extends PDOConnection
 {
+    /** @var LastInsertId|null */
+    protected $lastInsertId;
```
Ocramius: `private` unless needed elsewhere

lib/Doctrine/DBAL/Driver/PDOSqlsrv/LastInsertId.php
```diff
@@ -0,0 +1,30 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Driver\PDOSqlsrv;
+
+/**
+ * Last Id Data Container.
+ */
+class LastInsertId
+{
+    /** @var string */
```
Ocramius: According to (missing) constructor semantics, this is `string|null`

lib/Doctrine/DBAL/Driver/PDOSqlsrv/LastInsertId.php
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * @param string $id
+     */
+    public function setId($id)
+    {
+        $this->id = $id;
+    }
+
+    /**
+     * @return string
```
Ocramius: According to (missing) constructor semantics, this is `string|null`

lib/Doctrine/DBAL/Driver/PDOSqlsrv/LastInsertId.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+class LastInsertId
+{
+    /** @var string */
+    private $id;
+
+    /**
+     * @param string $id
+     */
+    public function setId($id)
```
Ocramius: Type declarations over docblocks: this is `string` and `: void`. Same for the rest of the patch overall.

Ocramius (changes requested):
I'd need @deeky666 to check this patch, since he initially authored https://github.com/doctrine/dbal/pull/2648 (which was closed due to its high complexity/risk for little gain).

I'd say that also here the same applies: if a prepared statement is causing side-effects through stored procedures and triggers, I'd rather call the schema "foobar'd" and "needs to be fixed there" rather than trying to fix issues on this end.

warslett:
hey @sarven @Ocramius any update on this PR? looks like the branches might need clearing up (presuming this PR doesn't include 141 commits).

PR #3525, opened 2019-04-19
Extract exception factory methods into specific exceptions

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Continuing work started in #3131


jwage:
@morozov I got this rebased and cleaned up. I didn't do too much else from where the PR already was when work on it was stopped. Can we merge get this merged?

morozov:
@jwage sure, as soon as the builds are green. BTW would it make sense if you created PRs from your personal fork in order to not have Travis run the same jobs twice?

jwage:
The failures seem not related to any of the changes in this PR. Travis fails because of sqlsrv and php 7.4

morozov:
The SQL Server failures will probably be fixed by the changes in https://github.com/doctrine/dbal/pull/3528. Could you try rebasing?

> I wish there was another way to make it not run twice without having to work in a fork.

What's the problem with working in a fork?

jwage:
> What's the problem with working in a fork?

Nothing, it is just different and another "hop" in the workflow that I am not used to. If this is how everyone else is doing it, then I will adapt.

morozov:
> If this is how everyone else is doing it […]

I believe so unless it's someone's personal project.

jwage:
This passes now after rebasing against develop.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -912,10 +919,10 @@ public function executeQuery(string $query, array $params = [], $types = [], ?Qu
      */
     public function executeCacheQuery($query, $params, $types, QueryCacheProfile $qcp) : ResultStatement
     {
-        $resultCache = $qcp->getResultCacheDriver() ?? $this->_config->getResultCacheImpl();
+        $resultCache = $qcp->getResultCacheDriver() ?: $this->_config->getResultCacheImpl();
```
morozov: @jwage is this intentional? Most likely, it comes from an incorrect conflict resolution.

morozov (commented):
Looks like there's one unintentional change. Otherwise, good to go.

lib/Doctrine/DBAL/Connection.php
jwage: Yes, bad conflict resolution. I will fix, one second.

morozov:
Thank you @Majkl578 and @jwage!

Issue #1205, opened 2015-06-27
DBAL-1254: FIX TINYINT(1) - Make BOOL use BIT


doctrinebot:
Jira issue originally created by user jonny827:
1. MySQL doesnt have boolean type but it supports BIT(1) which in my more novice opinion would be more effective if you wanted to use something to map. OR just not made a boolean type.  Ideally, a programmer would go I am using the MySQL database engine, no boolean, let me use a TINY INT.  So instead of making the programmer think you dumbed down a field that can store 255unsigned valyes into one that stores TRUE OR FALSE. 
2. Now I need to use 2 bytes of data to store the integer 2-255unsigned in smallint.  So think about all of the fields that could use TINYINT to store configurations or settings where there is more than a true false. Account types, account status, day of week, day of month, current AGE, date at death, age required to do XYZ,  You do not offer enum and you made a strong tiny integer into a weak bipolar newborn.
3. By requiring two bites of data to these fields you will be double the data storage in probably 1 out of 5 of my columns. If you had a million records with this issue that is an extra 1MB of unneeded database size.  This may slow performance of select and other MySQL operation fractionally.
4. In summary, you spoon fed developers a weak FAT bipolar newborn.  Give us back out functionality, allow us to trim down and NORMALIZE properly, stop cross-mapping data types. You take away all enum options.  This is ONE major major major major benefit of Propel is that they at least have TINYINT. Also: built in behaviors versus complaining that its too hard to test and maintain behaviors.  I respect Doctrine and chose it for its stability and strength. Everyone makes mistakes but please fix this one. Sorry in advance for being rude but I preferred being honest versus being a suck-up.

Love your application, thank you very much for being SRP friendly(data mapper) stable 2.x, part of symphony install by default/massive tutorial/knowledge base online, more contributors, active contributors, ROADMAP(they refuse to use one in PROPEL), easier to unit test.  I spent weeks reading every word about both.  Thank you for listening.

I decided to post this as an issue and left the comment on (http://www.doctrine-project.org/jira/browse/[DBAL-781](http://www.doctrine-project.org/jira/browse/DBAL-781)) so others searching for this issue would realize other people have this frustration and it is not just them.  Also everyone was searching online for a list of comparable to Propel benefits so I added that in for a little SEO love.

PR #3526, opened 2019-04-21
MySQL: Use BOOLEAN with comment hint to store boolean (issue #1205)

homersimpsons:
|      Q       |   A
|------------- | -----------
| Type         | feature/improvement
| BC Break     | yes
| Fixed issues | #1205 
#### Summary

Originally this PR should make use of `BIT(1)` to store boolean (instead of `TINYINT(1)`).

But as the real problem was the restriction about using TINYINT datatype I switched to a comment hint instead.

Reasons for this PR are explained in the issue #1205 (sum up here):
1. Programmer can't use TINYINT that can store 256 values, that dumbed down the field  
2. Programmer have to use SMALLINT which is twice larger, while TINYINT is much more adapted for a lot of case
3. Requiring SMALLINT doubles the size required and the input lag on selects

Furthermore this can cause issue on relation with TINYINT id column

**Note**: In MySQL `BOOLEAN` type internally map to `TINYINT(1)` and then the column is of type `TINYINT(1)`. An issue is already open in MySQL https://bugs.mysql.com/bug.php?id=8485


lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -1049,7 +1053,8 @@ public function getReadLockSQL()
     protected function initializeDoctrineTypeMappings()
     {
         $this->doctrineTypeMapping = [
-            'tinyint'       => 'boolean',
+            'bit'           => 'boolean',
+            'tinyint'       => 'tinyint',
```
morozov: Please provide some info on how it's expected to work with an already deployed schema.

lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
```diff
@@ -159,7 +162,10 @@ protected function _getPortableTableColumnDefinition($tableColumn)
                 break;
         }
 
-        if ($this->_platform instanceof MariaDb1027Platform) {
+        if ($type === Types::BOOLEAN) {
+            // As BIT (0b'0' or 0b'1') is used to store boolean values we must cast this default to a php boolean
+            $columnDefault = (bool) bindec(substr($tableColumn['default'], 2, -1));
```
morozov: A `BIT` column can be nullable. Looks like it's not taken into account.

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
homersimpsons: I updated the PR to keep BOOLEAN (so TINYINT) and to use comment hint instead. On an already deployed schema, reverse engineering will break strict comparison (`=== true`).

lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
homersimpsons: I updated the test to test for `null`. I also removed this part as it's no more relevant.

homersimpsons:
I Updated this PR to use BOOLEAN datatype and Comment Hint the column. I also updated the description.

PR #3528, opened 2019-04-23
Reworked SQL Server installer on Travis CI

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

The existing implementation has a couple of issues:

1. The installer scripts are executed in the following order:
   https://github.com/doctrine/dbal/blob/74a2714de32d7de161790d0f54561c5b2b3e4e4b/.travis.yml#L103-L104

   Which means that a PHP extension is installed first, and then its dependencies are installed. Not sure how it works with `dist: trusty` but it [won't work](https://travis-ci.org/doctrine/dbal/jobs/523284700#L462) with `dist: xenial`. Also, the script is missing the `unixodbc-dev` package which is needed to be able to compile the extension and is probably included into the VM image.
2. The scripts install the `mssql-tools` package only in order to be able to connect to the DB from outside the container, however, it can be done from within the container which we do with [MySQL](https://github.com/doctrine/dbal/blob/9b75cd5febbca8591075157046e739b8ab655f6b/tests/travis/install-mysql-8.0.sh#L19) and [PostgreSQL](https://github.com/doctrine/dbal/blob/master/tests/travis/install-postgres-11.sh#L10).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,6 +100,7 @@ jobs:
       services:
         - docker
       before_script:
+        - bash ./tests/travis/install-sqlsrv-dependencies.sh
         - bash ./tests/travis/install-mssql-sqlsrv.sh
         - bash ./tests/travis/install-mssql.sh
     - stage: Test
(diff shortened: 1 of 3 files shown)


jwage (approved):
Ship it :+1:

Issue #3531, opened 2019-04-23
Improve exception messages


jwage:
Before 3.0, we want to audit all exception messages and improve/standardize the language and formatting of the messages.

The following can be improved:

- Inside a message, names should be quoted using double quotes, not single quotes.
- All sentences should be terminated with a period.
- `sprintf()` should be preferred for embedding data in messages over concatenation.

PR #3532, opened 2019-04-24
Extra info when fetching column definition fails

bashilbers:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary
Adds extra information when fetching column definition fails.


SenseException:
Why only at this call of `_getPortableTableColumnDefinition()` and not the others? Is this an edge case or convenience for development?

jwage:
I don't think this is the appropriate fix. We have an issue for 3.0 to improve all the exception messages https://github.com/doctrine/dbal/issues/3531

PR #3533, opened 2019-04-25
Updated mariadb 10.4 support

nesk:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

Since [version 10.4.3](https://mariadb.com/kb/en/library/mariadb-1043-release-notes/), MariaDB uses the `JSON_VALID()` function as a `CHECK` constraint for the JSON data type alias in order to ensure that a valid json document is inserted.

Currently, Doctrine considers MariaDB has no native support for the type JSON type because its an alias of LONGTEXT. However, since the 10.4.3 release adds some constraints to the JSON type alias, I think it's important to use the latter instead of the LONGTEXT type which doesn't add a `CHECK` constraint.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,16 +74,6 @@ jobs:
         - bash ./tests/travis/install-mysql-8.0.sh
     - stage: Test
       php: 7.2
-      env: DB=mariadb MARIADB_VERSION=10.3
-      addons:
-        mariadb: 10.3
-    - stage: Test
-      php: 7.2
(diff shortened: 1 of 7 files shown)


SenseException:
@nesk Did you find the time to look for the failing tests in travis-ci?

morozov:
@nesk please retarget against `master`. Since it doesn't look like a breaking change, it doesn't have to wait until `3.0`.

nesk:
Closing since I didn't take time to fix this, sorry…

PR #3535, opened 2019-04-29
Add IfExists and IfNotExists convenience methods to Schema and Table

### LINKED ISSUE #3529: remove/drop/add and if exists check within the same method call.
### Feature Request

Combine add/remove/drop operations with if exists for simpler migrations.

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes/no
| BC Break    | yes/no

#### Summary

Would others find it of benefit to have the following methods available?

[_remove|drop|add_][_ForeignKey|Index|Column|Table_]IfExists()

For example

```
public function removeForeignKeyIfExists($constraintName)
{
    if ($this->hasForeignKey($constraintName)) {
        $this->removeForeignKey($constraintName);
    }
}
```

I personally believe these methods would be helpful in keeping migrations clean. If others think so to I am more than happy to put in the work and submit a PR.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | fixes #3529 

#### Summary

Add IfExists and IfNotExists convenience methods to Schema and Table public API:

- Schema::createNamespaceIfNotExists
- Schema::createTableIfNotExists
- Schema::dropTableIfExists
- Schema::createSequenceIfNotExists
- Schema::dropSequenceIfExists
- Table::addIndexIfNotExists
- Table::addUniqueIndexIfNotExists
- Table::dropIndexIfExists
- Table::addColumnIfNotExists
- Table::dropColumnIfExists
- Table::addForeignKeyConstraintIfNotExists
- Table::removeForeignKeyIfExists

#### Notes

- `addForeignKeyConstraint` and `removeForeignKey` method names are not consistent. Should we make this consistent in 3.0? Either both with `Constraint` suffix or not.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Table.php
+++ b/lib/Doctrine/DBAL/Schema/Table.php
@@ -123,11 +123,29 @@ class Table extends AbstractAsset
     public function addIndex(array $columnNames, $indexName = null, array $flags = [], array $options = [])
     {
         if ($indexName === null) {
-            $indexName = $this->_generateIdentifierName(
-                array_merge([$this->getName()], $columnNames),
-                'idx',
-                $this->_getMaxIdentifierLength()
-            );
+            $indexName = $this->generateNormalIndexName($columnNames);
+        }
+
+        return $this->_addIndex($this->_createIndex($columnNames, $indexName, false, false, $flags, $options));
+    }
+
+    /**
+     * @param array<int, string>   $columnNames
+     * @param array<int, string>   $flags
+     * @param array<string, mixed> $options
+     */
+    public function addIndexIfNotExists(
+        array $columnNames,
+        ?string $indexName = null,
+        array $flags = [],
+        array $options = []
+    ) : self {
+        if ($indexName === null) {
+            $indexName = $this->generateNormalIndexName($columnNames);
+        }
+
+        if ($this->hasIndex($indexName)) {
+            return $this;
         }
 
         return $this->_addIndex($this->_createIndex($columnNames, $indexName, false, false, $flags, $options));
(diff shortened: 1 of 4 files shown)


morozov:
I'm not sure that the "add*IfNotExists" methods are a good idea. The existing object is likely to have a different definition than the one being added. Therefore, the end result of the operation is non-deterministic.

In general, it's the responsibility of the schema manager and comparator to compare schemas and produce the diff which will guarantee that the end result matches the expectation.

I'm skeptical about introducing this API.

jwage:
I don't have too much of an opinion. I've written my own `if` checks in my own scripts a few times to avoid errors so having the convenience methods are just a nice to have. @NeilMasters do have an opinion?

NeilMasters:
@jwage your implementation is exactly what I previously considered, due to recent time  constraints I have had to put this on the back burner so 👍 for picking it up.

IMO this is a helpful addition to anyone who uses migrations extensively in particular those who are passionate about protecting migrations from failure. Personally I always protect my migrations to prevent repeated query executions and this change cuts down on boiler plate code by extending already implemented functionality.

lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -123,11 +123,29 @@ public function setPrimaryKey(array $columnNames, $indexName = false)
...
-            );
+            $indexName = $this->generateNormalIndexName($columnNames);
+        }
+
+        return $this->_addIndex($this->_createIndex($columnNames, $indexName, false, false, $flags, $options));
+    }
+
+    /**
+     * @param array<int, string>   $columnNames
+     * @param array<int, string>   $flags
```
AlessandroMinoccheri: could it be that the index name is missing?

pvsaintpe (approved):

pvsaintpe:
@jwage Please fix conflicts.

lib/Doctrine/DBAL/Schema/Table.php
SenseException: The coding style wants to add phpdocs if the type isn't expressive enough. An array can contain any possible type as element and a key can be one of some generics. A nullable string as a type is sufficient for the Doctrine coding style.

morozov:
Closing as stale.

PR #3518, opened 2019-04-15
Remove @​deprecated stuff

Majkl578:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

#### Summary

Removes so far @​deprecated stuff for DBAL 3.0. Fixes #3508.

PR #3536, opened 2019-04-30
Improve consistency of exception message formatting.

### LINKED ISSUE #3531: Improve exception messages

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | fixes #3531

#### Summary

Improve consistency of exception message formatting. See #3531 for details.

#### TODO

- [ ] Add tests for exceptions that weren't tested.


lib/Doctrine/DBAL/DBALException.php
```diff
@@ -28,21 +28,22 @@ class DBALException extends Exception
      */
     public static function driverExceptionDuringQuery(Driver $driver, Throwable $driverEx, $sql, array $params = [])
     {
-        $msg = "An exception occurred while executing '" . $sql . "'";
-        if ($params) {
-            $msg .= ' with params ' . self::formatParameters($params);
-        }
-        $msg .= ":\n\n" . $driverEx->getMessage();
+        $message = sprintf(
+            "An exception occurred while executing \"%s\"%s:\n\n%s.",
```
greg0ire: How about
```
<<<MESSAGE
An exception occurred while executing "%s"%s:\n\n%s.
MESSAGE
```
or even

```
<<<'MESSAGE'
An exception occurred while executing "%s"%s:

%s.
MESSAGE
```
?

lib/Doctrine/DBAL/DBALException.php
```diff
@@ -28,21 +28,22 @@ class DBALException extends Exception
...
+        return static::wrapException($driver, $driverEx, $message);
     }
 
     /**
      * @return self
      */
     public static function driverException(Driver $driver, Throwable $driverEx)
     {
-        return static::wrapException($driver, $driverEx, 'An exception occurred in driver: ' . $driverEx->getMessage());
+        return static::wrapException($driver, $driverEx, sprintf('An exception occurred in driver "%s".', $driverEx->getMessage()));
```
morozov: I'm not sure we need to enclose sentences in quotes. When a message contains another message, the syntax with the colon makes more sense to me. The same syntax is used in Golang for error chaining.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -344,7 +344,7 @@ public function fetch($fetchMode = null, ...$args)
                 return (object) $assoc;
 
             default:
-                throw new MysqliException(sprintf("Unknown fetch type '%s'", $fetchMode));
+                throw new MysqliException(sprintf('Unknown fetch type "%s".', $fetchMode));
```
morozov: Let's replace "type" with "mode" and `"%s"` with `%d` (it's an integer).

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -156,7 +156,7 @@ public static function convertPositionalToNamedPlaceholders($statement)
 
         if ($currentLiteralDelimiter) {
             throw new OCI8Exception(sprintf(
-                'The statement contains non-terminated string literal starting at offset %d',
+                'The statement contains non-terminated string literal starting at offset "%d".',
```
morozov: I don't think numbers need to be enclosed in quotes. Syntactically, quotes indicate a string which is not the case.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -407,7 +407,7 @@ public function fetch($fetchMode = null, ...$args)
         }
 
         if (! isset(self::$fetchModeMap[$fetchMode])) {
-            throw new InvalidArgumentException('Invalid fetch style: ' . $fetchMode);
+            throw new InvalidArgumentException(sprintf('Invalid fetch style "%s".', $fetchMode));
```
morozov: Let's replace "style" with "mode" for consistency.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -434,7 +434,7 @@ public function fetchAll($fetchMode = null, ...$args)
...
```
morozov: Let's replace "style" with "mode" for consistency.

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereStatement.php
```diff
@@ -73,7 +73,7 @@ class SQLAnywhereStatement implements IteratorAggregate, Statement
     public function __construct($conn, $sql)
     {
         if (! is_resource($conn)) {
-            throw new SQLAnywhereException('Invalid SQL Anywhere connection resource: ' . $conn);
+            throw new SQLAnywhereException(sprintf('Invalid SQL Anywhere connection resource "%s".', $conn));
```
morozov: This won't be informative if the value is not a scalar (and it's not expected to be a scalar). Likely, it won't happen often since we enforce types statically, but otherwise, we need a simple formatting function which can indicate (not dump) an arbitrary value (e.g. class if it's an object, the value itself if it's a string or number, etc).

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereStatement.php
```diff
@@ -108,7 +108,7 @@ public function bindParam($column, &$variable, $type = ParameterType::STRING, $l
                 break;
 
             default:
-                throw new SQLAnywhereException('Unknown type: ' . $type);
+                throw new SQLAnywhereException(sprintf('Unknown type "%s".', $type));
```
morozov: Type is an integer, not a string.

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereStatement.php
```diff
@@ -215,7 +215,7 @@ public function fetch($fetchMode = null, ...$args)
                 return sasql_fetch_object($this->result);
 
             default:
-                throw new SQLAnywhereException('Fetch mode is not supported: ' . $fetchMode);
+                throw new SQLAnywhereException(sprintf('Fetch mode is not supported "%s".', $fetchMode));
```
morozov: Mode is an integer.

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereStatement.php
```diff
@@ -314,7 +314,7 @@ private function castObject(stdClass $sourceObject, $destinationClass, array $ct
         if (! is_string($destinationClass)) {
             if (! is_object($destinationClass)) {
                 throw new SQLAnywhereException(sprintf(
-                    'Destination class has to be of type string or object, %s given.',
+                    'Destination class has to be of type string or object, "%s" given.',
```
morozov: Value printer is needed.

lib/Doctrine/DBAL/Exception/MissingSQLParam.php
```diff
@@ -12,7 +12,7 @@ final class MissingSQLParam extends SQLParserUtilsException
     public static function new(string $paramName) : self
     {
         return new self(
-            sprintf('Value for :%1$s not found in params array. Params array key should be "%1$s"', $paramName)
+            sprintf('Value for :%1$s not found in params array. Params array key should be "%1$s"', $paramName) // what is this syntax?
```
morozov: It allows reusing one parameter multiple times by its position. See [`sprintf()`](https://php.net/sprintf#example-6099).

lib/Doctrine/DBAL/Id/TableGenerator.php
```diff
@@ -139,7 +140,7 @@ public function nextValue($sequenceName)
             $this->conn->commit();
         } catch (Throwable $e) {
             $this->conn->rollBack();
-            throw new DBALException('Error occurred while generating ID with TableGenerator, aborted generation: ' . $e->getMessage(), 0, $e);
+            throw new DBALException(sprintf('Error occurred while generating ID with TableGenerator, aborted generation with error "%s".', $e->getMessage()), 0, $e);
```
morozov: Let's find some consensus on message chaining. If the chained exception also contains another exception, the nested quotes will look ugly:

> Authentication failed with error "Database connection failed with error: "Invalid user name"".

With colons, it will look better:

> Authentication failed: Database connection failed: Invalid user name

When a message chains another message, it probably shouldn't add its own period since the period will be contained in the innermost message.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -828,7 +828,7 @@ private function convertSingleBooleanValue($value, $callback)
             return $callback(true);
         }
 
-        throw new UnexpectedValueException("Unrecognized boolean literal '${value}'");
+        throw new UnexpectedValueException(sprintf('Unrecognized boolean literal "%s".', $value));
```
morozov: Variable printer needed.

lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
```diff
@@ -551,7 +551,7 @@ public function getDropIndexSQL($index, $table = null)
 
         if (! is_string($index)) {
             throw new InvalidArgumentException(
-                'SQLAnywherePlatform::getDropIndexSQL() expects $index parameter to be string or ' . Index::class . '.'
+                sprintf('SQLAnywherePlatform::getDropIndexSQL() expects $index parameter to be "string" or "%s".', Index::class)
```
morozov: Should it be "…or an instance of %s", Index::class? Classes probably shouldn't be enclosed in quotes since they are symbols, not strings.

lib/Doctrine/DBAL/Query/Exception/NonUniqueAlias.php
```diff
@@ -17,8 +17,8 @@ public static function new(string $alias, array $registeredAliases) : self
     {
         return new self(
             sprintf(
-                "The given alias '%s' is not unique in FROM and JOIN clause table. "
-                    . 'The currently registered aliases are: %s.',
+                'The given alias "%s" is not unique in FROM and JOIN clause table. '
+                    . 'The currently registered aliases are "%s".',
```
morozov: It's weird to see all list elements enclosed in one pair of quotes.

morozov:
@jwage please see the comments above. I haven't reviewed the entire patch yet.

lib/Doctrine/DBAL/Id/TableGenerator.php
jwage: Agree. Colons work better. I will make the changes :+1:

lib/Doctrine/DBAL/DBALException.php
```diff
@@ -28,21 +35,57 @@ class DBALException extends Exception
...
     /**
      * @return self
      */
     public static function driverException(Driver $driver, Throwable $driverEx)
     {
-        return static::wrapException($driver, $driverEx, 'An exception occurred in driver: ' . $driverEx->getMessage());
+        return static::wrapException($driver, $driverEx, sprintf('An exception occurred in driver with message: %s', $driverEx->getMessage()));
+    }
+
+    public static function formatValue($value) : string
```
jwage: Is there a better place for this? or do you want me to create a new class for it? New class probably makes the most sense.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -109,7 +109,7 @@ public function bindParam($column, &$variable, $type = ParameterType::STRING, $l
         assert(is_int($column));
 
         if (! isset(self::$_paramTypeMap[$type])) {
-            throw new MysqliException(sprintf("Unknown type: '%s'", $type));
+            throw new MysqliException(sprintf('Unknown type "%s".', $type));
```
jwage: Should this be `Unknown type, "%s" given.` to be consistent with other exceptions?

lib/Doctrine/DBAL/DBALException.php
```diff
@@ -28,21 +35,57 @@ class DBALException extends Exception
      */
     public static function driverExceptionDuringQuery(Driver $driver, Throwable $driverEx, $sql, array $params = [])
     {
-        $msg = "An exception occurred while executing '" . $sql . "'";
-        if ($params) {
-            $msg .= ' with params ' . self::formatParameters($params);
-        }
-        $msg .= ":\n\n" . $driverEx->getMessage();
+        $messageFormat = <<<MESSAGE
```
greg0ire: Nowdoc is more appropriate if you pick the version w/o `\n`
```suggestion
        $messageFormat = <<<'MESSAGE'
```

lib/Doctrine/DBAL/DBALException.php
greg0ire: `Doctrine\DBAL\Exception\VarFormatter::__invoke()`?

jwage: If we use `__invoke()`, would it be named like an action? e.g. `FormatVariable` instead of `VarFormatter`

greg0ire: Yeah that might make more sense 👍

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
greg0ire: I think that format should be used when you oppose the first part to the second part ("$type should be a known type, XXX given). I think it is fine as is.

jwage: Isn't that the case here? I think the above should be something like:

```
Unknown type, "%s" given.
```

or as you said:

```
$type should be a known type, "%s" given.
```

lib/Doctrine/DBAL/DBALException.php
```diff
@@ -28,21 +22,28 @@ class DBALException extends Exception
...
     {
-        $msg = "An exception occurred while executing '" . $sql . "'";
-        if ($params) {
-            $msg .= ' with params ' . self::formatParameters($params);
-        }
-        $msg .= ":\n\n" . $driverEx->getMessage();
+        $messageFormat = <<<'MESSAGE'
+An exception occurred while executing "%s"%s:
+
+%s.
```
morozov: The driver exception will likely have a period in the end. In a worst-case scenario, it's better to not have a period at all than have two.

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -206,7 +206,7 @@ private function convertParamType(int $type) : int
         if (! isset(self::PARAM_TYPE_MAP[$type])) {
             // TODO: next major: throw an exception
             @trigger_error(sprintf(
-                'Using a PDO parameter type (%d given) is deprecated and will cause an error in Doctrine 3.0',
+                'Using a PDO parameter type ("%d" given) is deprecated and will cause an error in Doctrine 3.0.',
```
morozov: A number doesn't need quotes.

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -226,8 +226,8 @@ private function convertFetchMode(int $fetchMode) : int
         if (! isset(self::FETCH_MODE_MAP[$fetchMode])) {
             // TODO: next major: throw an exception
             @trigger_error(sprintf(
-                'Using a PDO fetch mode or their combination (%d given)' .
-                ' is deprecated and will cause an error in Doctrine 3.0',
+                'Using a PDO fetch mode or their combination ("%d" given)' .
```
morozov: A number doesn't need quotes.

tests/Doctrine/Tests/DBAL/Driver/OCI8/OCI8StatementTest.php
```diff
@@ -104,15 +104,15 @@ public static function nonTerminatedLiteralProvider()
         return [
             'no-matching-quote' => [
                 "SELECT 'literal FROM DUAL",
-                '/offset 7/',
+                '/offset "7"./',
```
morozov: This shouldn't be changed.

lib/Doctrine/DBAL/Exception/FormatArray.php
```diff
@@ -0,0 +1,41 @@
...
+{
+    /**
+     * Returns a human-readable representation of an array of parameters.
+     * This properly handles binary data by returning a hex representation.
+     *
+     * @param array<mixed, mixed> $array
+     */
+    public function __invoke(array $array) : string
+    {
+        return '[' . implode(', ', array_map(static function ($param) {
```
morozov: I may have misspoken about printing the entire thing. It may be an overkill and cause issues (e.g. exposing passwords or memory leaks). What we need is a printable representation of a variable in the context of the exception, not a dump.

In most cases where a function accepts `mixed $param`, the type (including class name or resource type) is enough. Printing the actual value may not be needed since the exception will most likely be about the type, not the value.

lib/Doctrine/DBAL/Exception/InvalidColumnIndex.php
```diff
@@ -12,7 +12,7 @@ final class InvalidColumnIndex extends DBALException
     public static function new(int $index, int $count) : self
     {
         return new self(sprintf(
-            'Invalid column index %d. The statement result contains %d column%s.',
+            'Invalid column index "%d". The statement result contains "%d" column%s.',
```
morozov: `"%d"` → `%d`

lib/Doctrine/DBAL/Exception/MissingSQLType.php
```diff
@@ -12,7 +12,7 @@ final class MissingSQLType extends SQLParserUtilsException
     public static function new(string $typeName) : self
     {
         return new self(
-            sprintf('Value for :%1$s not found in types array. Types array key should be "%1$s"', $typeName)
+            sprintf('Value for :%1$s not found in types array. Types array key should be "%1$s"', $typeName) // what is this syntax?
```
morozov: Please remove the comment. The second sentence looks grammatically incorrect. Not even sure what it means.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -762,7 +766,7 @@ public function getTrimExpression(string $str, int $mode = TrimMode::UNSPECIFIED
             default:
                 throw new InvalidArgumentException(
                     sprintf(
-                        'The value of $mode is expected to be one of the TrimMode constants, %d given',
+                        'The value of $mode is expected to be one of the TrimMode constants, "%d" given.',
```
morozov: `"%d"` → `%d`

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3286,14 +3290,14 @@ public function getTimeFormatString()
     {
         if ($offset < 0) {
             throw new DBALException(sprintf(
-                'Offset must be a positive integer or zero, %d given',
+                'Offset must be a positive integer or zero, "%d" given.',
```
morozov: `"%d"` → `%d`

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1037,7 +1037,7 @@ public function getTrimExpression(string $str, int $mode = TrimMode::UNSPECIFIED
     {
         if (! in_array($mode, [TrimMode::UNSPECIFIED, TrimMode::LEADING, TrimMode::TRAILING, TrimMode::BOTH], true)) {
             throw new InvalidArgumentException(
-                sprintf('The value of $mode is expected to be one of the TrimMode constants, %d given', $mode)
+                sprintf('The value of $mode is expected to be one of the TrimMode constants, "%d" given', $mode)
```
morozov: `"%d"` → `%d`

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -80,7 +80,7 @@ public function getTrimExpression(string $str, int $mode = TrimMode::UNSPECIFIED
             default:
                 throw new InvalidArgumentException(
                     sprintf(
-                        'The value of $mode is expected to be one of the TrimMode constants, %d given',
+                        'The value of $mode is expected to be one of the TrimMode constants, "%d" given.',
```
morozov: `"%d"` → `%d`

lib/Doctrine/DBAL/Schema/Exception/IndexNameInvalid.php
```diff
@@ -12,7 +12,7 @@ final class IndexNameInvalid extends SchemaException
     public static function new(string $indexName) : self
     {
         return new self(
-            sprintf('Invalid index-name %s given, has to be [a-zA-Z0-9_]', $indexName),
+            sprintf('Invalid index-name "%s" given, has to be [a-zA-Z0-9_]', $indexName),
```
morozov: Why is it index-name with a dash?

lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -649,7 +650,7 @@ public function getPrimaryKeyColumns()
         $primaryKey = $this->getPrimaryKey();
 
         if ($primaryKey === null) {
-            throw new DBALException('Table ' . $this->getName() . ' has no primary key.');
+            throw new DBALException(sprintf('Table "%s", has no primary key.', $this->getName()));
```
morozov: Looks like the comma is redundant.

lib/Doctrine/DBAL/Schema/UniqueConstraint.php
```diff
@@ -178,7 +178,7 @@ public function getOptions()
     protected function _addColumn($column)
     {
         if (! is_string($column)) {
-            throw new InvalidArgumentException('Expecting a string as Index Column');
+            throw new InvalidArgumentException('Expecting a string as Index Column.');
```
morozov: This should be converted to a type hint if possible.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -191,7 +191,7 @@ public function testEventManagerPassedToPlatform()
     public function testDriverExceptionIsWrapped($method)
     {
         $this->expectException(DBALException::class);
-        $this->expectExceptionMessage("An exception occurred while executing 'MUUHAAAAHAAAA':\n\nSQLSTATE[HY000]: General error: 1 near \"MUUHAAAAHAAAA\"");
+        $this->expectExceptionMessage("An exception occurred while executing \"MUUHAAAAHAAAA\":\n\nSQLSTATE[HY000]: General error: 1 near \"MUUHAAAAHAAAA\"");
```
morozov: `phpcbf` should have converted the outer quotes to `'`. Not sure why it didn't.

tests/Doctrine/Tests/DBAL/Exception/InvalidColumnIndexTest.php
```diff
@@ -15,14 +15,14 @@ public function testNew() : void
         $exception = InvalidColumnIndex::new(5, 1);
 
         self::assertInstanceOf(DBALException::class, $exception);
-        self::assertSame('Invalid column index 5. The statement result contains 1 column.', $exception->getMessage());
+        self::assertSame('Invalid column index "5". The statement result contains "1" column.', $exception->getMessage());
```
morozov: Please revert.

tests/Doctrine/Tests/DBAL/Functional/PDOStatementTest.php
```diff
@@ -32,7 +32,7 @@ protected function setUp() : void
 
     /**
      * @group legacy
-     * @expectedDeprecation Using a PDO fetch mode or their combination (%d given) is deprecated and will cause an error in Doctrine 3.0
+     * @expectedDeprecation Using a PDO fetch mode or their combination ("%d" given) is deprecated and will cause an error in Doctrine 3.0.
```
morozov: Please revert.

lib/Doctrine/DBAL/Exception/FormatArray.php
jwage: I will remove this from the `FormatVariable` class:

```
if (is_array($value)) {
    return (new FormatArray())->__invoke($value);
}
```

lib/Doctrine/DBAL/Exception/MissingSQLType.php
jwage: I will look at where this exception is being thrown and take a stab at writing a better message.

lib/Doctrine/DBAL/Schema/Exception/IndexNameInvalid.php
jwage: No good reason as far as I can tell. I will change to `index name`.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
jwage: Maybe because it has `\n\n` inside?

lib/Doctrine/DBAL/Exception/MissingSQLType.php
jwage: Take a look at it now and let me know what you think.

lib/Doctrine/DBAL/Exception/MissingSQLType.php
```diff
@@ -11,8 +11,6 @@ final class MissingSQLType extends SQLParserUtilsException
 {
     public static function new(string $typeName) : self
     {
-        return new self(
-            sprintf('Value for :%1$s not found in types array. Types array key should be "%1$s"', $typeName)
-        );
+        return new self(sprintf('Type named "%s" not found in types array.', $typeName));
```
morozov: `SQLParserUtils::extractParam()` is really overloaded/generalized. What it's trying to say is that it's unable to find the type of the array parameter with the given name (`$paramName`, not `$typeName`). Something like this (add the missing articles):

> Type of array parameter %s is missing (not specified).

morozov: It should be probably renamed to `MissingArrayParameterType` together with `MissingSQLParam` → `MissingArrayParameter`.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
morozov: Indeed.

lib/Doctrine/DBAL/Exception/MissingSQLType.php
jwage: Ok, I made another change. I think we're getting closer :)

lib/Doctrine/DBAL/Exception/MissingArrayParameterType.php
```diff
@@ -7,12 +7,10 @@
 use Doctrine\DBAL\SQLParserUtilsException;
 use function sprintf;
 
-final class MissingSQLType extends SQLParserUtilsException
+final class MissingArrayParameterType extends SQLParserUtilsException
 {
     public static function new(string $typeName) : self
```
morozov: `$typeName` should be renamed to `$paramName`.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -109,7 +109,7 @@ public function bindParam($column, &$variable, $type = ParameterType::STRING, $l
         assert(is_int($column));
 
         if (! isset(self::$_paramTypeMap[$type])) {
-            throw new MysqliException(sprintf("Unknown type: '%s'", $type));
+            throw new MysqliException(sprintf('Unknown type, "%s" given.', $type));
```
morozov: Unlike the wrapper statement, wrapper statements only accept `ParameterType::*` constants for `$type` which are integers. Even if the type of `$type` is not enforced to `int`, we shouldn't expect anything else here. Therefore, `"%s"` → `%d`.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -124,7 +124,7 @@ public function bindValue($param, $value, $type = ParameterType::STRING) : void
         assert(is_int($param));
 
         if (! isset(self::$_paramTypeMap[$type])) {
-            throw new MysqliException(sprintf("Unknown type: '%s'", $type));
+            throw new MysqliException(sprintf('Unknown type, "%s" given.', $type));
```
morozov: Same as above, `"%s"` → `%d`.

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -206,7 +206,7 @@ private function convertParamType(int $type) : int
         if (! isset(self::PARAM_TYPE_MAP[$type])) {
             // TODO: next major: throw an exception
             @trigger_error(sprintf(
-                'Using a PDO parameter type (%d given) is deprecated and will cause an error in Doctrine 3.0',
+                'Using a PDO parameter type (%d given) is deprecated and will cause an error in Doctrine 3.0.',
```
morozov: For the reference, this should be removed in #3518.

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -227,7 +227,7 @@ private function convertFetchMode(int $fetchMode) : int
             // TODO: next major: throw an exception
             @trigger_error(sprintf(
                 'Using a PDO fetch mode or their combination (%d given)' .
-                ' is deprecated and will cause an error in Doctrine 3.0',
+                ' is deprecated and will cause an error in Doctrine 3.0.',
```
morozov: For the reference, this should be removed in #3518.

lib/Doctrine/DBAL/Exception/InvalidDriverClass.php
```diff
@@ -14,7 +14,7 @@ public static function new(string $driverClass) : self
     {
         return new self(
             sprintf(
-                "The given 'driverClass' %s has to implement the %s interface.",
+                'The given "driverClass" "%s" has to implement the "%s" interface.',
```
morozov: Since the class and interface are symbols, the quotes are not needed.

lib/Doctrine/DBAL/Exception/InvalidPlatformType.php
```diff
@@ -21,7 +21,7 @@ public static function new($invalidPlatform) : self
         if (is_object($invalidPlatform)) {
             return new self(
                 sprintf(
-                    "Option 'platform' must be a subtype of '%s', instance of '%s' given",
+                    'Option "platform" must be a subtype of "%s", instance of "%s" given.',
```
morozov: Since the class and interface are symbols, the quotes are not needed.

lib/Doctrine/DBAL/Exception/InvalidPlatformType.php
```diff
@@ -30,7 +30,7 @@ public static function new($invalidPlatform) : self
 
         return new self(
             sprintf(
-                "Option 'platform' must be an object and subtype of '%s'. Got '%s'",
+                'Option "platform" must be an object and subtype of "%s". Got "%s".',
```
morozov: Since the class is a symbol, the quotes are not needed. Got "%s" should use the variable formatter. Otherwise, it will print Got "int" for an integer which is misleading.

lib/Doctrine/DBAL/Exception/InvalidWrapperClass.php
```diff
@@ -14,7 +14,7 @@ public static function new(string $wrapperClass) : self
     {
         return new self(
             sprintf(
-                "The given 'wrapperClass' %s has to be a subtype of %s.",
+                'The given "wrapperClass" "%s" has to be a subtype of "%s".',
```
morozov: Both parameters indicate symbols.

lib/Doctrine/DBAL/Id/TableGenerator.php
```diff
@@ -126,7 +127,7 @@ public function nextValue($sequenceName)
                 $rows = $this->conn->executeUpdate($sql, [$sequenceName, $row['sequence_value']]);
 
                 if ($rows !== 1) {
-                    throw new DBALException('Race-condition detected while updating sequence. Aborting generation');
+                    throw new DBALException('Race-condition detected while updating sequence. Aborting generation.');
```
morozov: "Race condition" doesn't need a dash.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -420,7 +420,11 @@ public function getDoctrineTypeMapping($dbType)
         $dbType = strtolower($dbType);
 
         if (! isset($this->doctrineTypeMapping[$dbType])) {
-            throw new DBALException('Unknown database type ' . $dbType . ' requested, ' . static::class . ' may not support it.');
+            throw new DBALException(sprintf(
+                'Unknown database type "%s" requested, "%s" may not support it.',
```
morozov: The second parameter is a symbol. Quotes are not needed.

lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
```diff
@@ -551,7 +551,7 @@ public function getDropIndexSQL($index, $table = null)
 
         if (! is_string($index)) {
             throw new InvalidArgumentException(
-                'SQLAnywherePlatform::getDropIndexSQL() expects $index parameter to be string or ' . Index::class . '.'
+                sprintf('SQLAnywherePlatform::getDropIndexSQL() expects $index parameter to be "string" or an instance of %s.', Index::class)
```
morozov: Should it be "…a string or an instance…"?

lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
```diff
@@ -565,7 +565,7 @@ public function getDropIndexSQL($index, $table = null)
 
         if (! is_string($table)) {
             throw new InvalidArgumentException(
-                'SQLAnywherePlatform::getDropIndexSQL() expects $table parameter to be string or ' . Index::class . '.'
+                sprintf('SQLAnywherePlatform::getDropIndexSQL() expects $table parameter to be "string" or an instance of %s.', Index::class)
```
morozov: Should it be "…a string or an instance…"?

lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
```diff
@@ -1446,7 +1446,7 @@ protected function getTableConstraintDeclarationSQL(Constraint $constraint, $nam
         }
 
         if (! $constraint instanceof Index) {
-            throw new InvalidArgumentException('Unsupported constraint type: ' . get_class($constraint));
+            throw new InvalidArgumentException(sprintf('Unsupported constraint type "%s".', get_class($constraint)));
```
morozov: The quotes are not needed (symbol).

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -215,7 +215,7 @@ public function getDropIndexSQL($index, $table = null)
         if ($index instanceof Index) {
             $index = $index->getQuotedName($this);
         } elseif (! is_string($index)) {
-            throw new InvalidArgumentException('AbstractPlatform::getDropIndexSQL() expects $index parameter to be string or \Doctrine\DBAL\Schema\Index.');
+            throw new InvalidArgumentException('AbstractPlatform::getDropIndexSQL() expects $index parameter to be "string" or "\Doctrine\DBAL\Schema\Index".');
```
morozov: Can we make it "…a string or an instance of %s", Index::class?

lib/Doctrine/DBAL/Types/Exception/InvalidType.php
```diff
@@ -31,7 +31,7 @@ public static function new($value, string $toType, array $possibleTypes) : self
         if (is_scalar($value)) {
             return new self(
                 sprintf(
-                    "Could not convert PHP value '%s' of type '%s' to type '%s'. Expected one of the following types: %s",
+                    'Could not convert PHP value "%s" of type "%s" to type "%s". Expected one of the following types "%s".',
```
morozov: The last parameter is a comma-separated list.

lib/Doctrine/DBAL/Types/Exception/InvalidType.php
```diff
@@ -42,7 +42,7 @@ public static function new($value, string $toType, array $possibleTypes) : self
 
         return new self(
             sprintf(
-                "Could not convert PHP value of type '%s' to type '%s'. Expected one of the following types: %s",
+                'Could not convert PHP value of type "%s" to type "%s". Expected one of the following types "%s".',
```
morozov: The last parameter is a comma-separated list.

lib/Doctrine/DBAL/Types/Exception/ValueNotConvertible.php
```diff
@@ -20,7 +20,7 @@ public static function new($value, $toType, ?string $message = null) : self
         if ($message !== null) {
             return new self(
                 sprintf(
-                    "Could not convert database value to '%s' as an error was triggered by the unserialization: '%s'",
+                    'Could not convert database value to "%s" as an error was triggered by the unserialization: "%s".',
```
morozov: Quotes are not needed for the last parameter (chained error message). The period is probably okay since the error message will come from PHP, not from us.

lib/Doctrine/DBAL/Types/Exception/SerializationFailed.php
```diff
@@ -21,7 +21,7 @@ public static function new($value, string $format, string $error) : self
 
         return new self(
             sprintf(
-                "Could not convert PHP type '%s' to '%s', as an '%s' error was triggered by the serialization",
+                'Could not convert PHP type "%s" to "%s", as an "%s" error was triggered by the serialization',
```
morozov: Can we move the chained error message the end, similarly to `ValueNotConvertible`?

lib/Doctrine/DBAL/Exception/FormatVariable.php
```diff
@@ -0,0 +1,51 @@
...
+        if (is_bool($value)) {
+            return $value === true ? 'true' : 'false';
+        }
+
+        if (is_array($value)) {
+            return 'array';
+        }
+
+        if (is_string($value)) {
+            return $value;
```
morozov: This probably should be just `'string'` since otherwise, it may display a long or security-sensitive string. With that, the class should be named as `GetVariableType`.

lib/Doctrine/DBAL/Exception/FormatVariable.php
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        throw new InvalidArgumentException(sprintf('Could not format value of type "%s".', gettype($value)));
```
morozov: It's not an exceptional situation. If implemented as `return gettype($value)`, it will cover `array`, `string`, `int` and `float` as well.

lib/Doctrine/DBAL/DBALException.php
```diff
@@ -28,21 +22,28 @@ class DBALException extends Exception
...
+An exception occurred while executing "%s"%s:
+
+%s
+MESSAGE;
 
-        return static::wrapException($driver, $driverEx, $msg);
+        $message = sprintf(
+            $messageFormat,
+            $sql,
+            $params !== [] ? sprintf(' with params %s', (new FormatArray())->__invoke($params)) : '',
```
morozov: Looks like it's the only use case for `FormatArray`. Can we leave the implementation here and not introduce the class? Embedding query parameters into an exception is a bad idea in general from the security perspective, so let's not make any improvements in this area.

lib/Doctrine/DBAL/Exception/MissingSQLType.php
morozov: Definitely. But we may need a couple more passes :)

lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
jwage: So no quotes around string?

lib/Doctrine/DBAL/Cache/ArrayStatement.php
```diff
@@ -48,7 +49,7 @@ public function __construct(array $data)
      */
     public function closeCursor() : void
     {
-        unset($this->data);
+        $this->data = null;
```
jwage: I was adding tests for ArrayStatement and this seemed wrong since it can result in PHP errors if you call other public methods after `closeCursor()`

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereStatement.php
```diff
@@ -73,7 +73,10 @@ class SQLAnywhereStatement implements IteratorAggregate, Statement
     public function __construct($conn, $sql)
     {
         if (! is_resource($conn)) {
-            throw new SQLAnywhereException('Invalid SQL Anywhere connection resource: ' . $conn);
+            throw new SQLAnywhereException(sprintf(
+                'Invalid SQL Anywhere connection resource, "%s" given.',
```
jwage: Sometimes `GetVariableType` will return a token (class name) and it will be wrapped in double quotes here. Should we move the quoting up in to `GetVariableType` or is that acceptable in this case?

lib/Doctrine/DBAL/Sharding/PoolingShardConnection.php
```diff
@@ -69,34 +70,34 @@ class PoolingShardConnection extends Connection
...
+                throw new InvalidArgumentException('Missing "id" for one configured shard. Please specify a unique shard-id.');
             }
 
             if (! is_numeric($shard['id']) || $shard['id'] < 1) {
                 throw new InvalidArgumentException('Shard Id has to be a non-negative number.');
             }
 
             if (isset($this->connectionParameters[$shard['id']])) {
-                throw new InvalidArgumentException('Shard ' . $shard['id'] . ' is duplicated in the configuration.');
+                throw new InvalidArgumentException(sprintf('Shard "%s" is duplicated in the configuration.', $shard['id']));
```
jwage: Is `$shard['id']` a string?

morozov: It's annotated as `int|string` in most places but looks like it's a numeric string (or an integer):

https://github.com/doctrine/dbal/blob/82f5fe317bedf9f328c06aeaab5a2f586810e94a/lib/Doctrine/DBAL/Sharding/PoolingShardConnection.php#L94-L96

Also, the `id` comes from `sys.federation_member_distributions.member_id`:

https://github.com/doctrine/dbal/blob/68e48e83021c22f19354558a5fdb46e53e9e6c1a/lib/Doctrine/DBAL/Sharding/SQLAzure/SQLAzureShardManager.php#L158-L164

which according to this [article](https://convective.wordpress.com/2012/03/05/introduction-to-sql-azure-federations/) is `INT`.

These details are mostly for myself, I'll need this info later to finish the work on strict types.

lib/Doctrine/DBAL/Cache/ArrayStatement.php
morozov: Makes perfect sense.

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereStatement.php
morozov: We should omit quotes here. The way I see it, we should quote string _values_. Something which would make sense as a value of type string in the code (e.g. 'Configuration parameter "username" is missing' or 'Database "doctrine" does not exist').

In this case, it's not a string `"\Foo\Bar"` or string `"array"`. It's just a word in the sentence.

lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
morozov: No quotes. We're not quoting a value "string", we're referring to a type.

morozov:
@jwage I rebased `develop` yesterday to accommodate recent conflicting changes from `master`. Please rebase.

jwage:
@morozov Done!

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereStatement.php
morozov: It's no longer a value, it's a type, so quotes are not needed.

lib/Doctrine/DBAL/Exception/InvalidPlatformType.php
```diff
@@ -30,9 +29,9 @@ public static function new($invalidPlatform) : self
 
         return new self(
             sprintf(
-                "Option 'platform' must be an object and subtype of '%s'. Got '%s'",
+                'Option "platform" must be an object and subtype of %s. Got "%s".',
```
morozov: No quotes around type needed.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -828,7 +829,10 @@ private function convertSingleBooleanValue($value, $callback)
             return $callback(true);
         }
 
-        throw new UnexpectedValueException("Unrecognized boolean literal '${value}'");
+        throw new UnexpectedValueException(sprintf(
+            'Unrecognized boolean literal, "%s" given.',
```
morozov: No quotes around type needed.

morozov:
@jwage thanks. Could you look at the build failure and the new comments? Looks like we're really close to finishing it.

jwage:
Fixed latest feedback and fixed the tests.

morozov (approved):

morozov:
Thank you :ship:

jwage:
Cool! Thanks!

Issue #3487, opened 2019-03-16
Using a user-provided PDO connection breaks assertion


morozov:
|    Q        |   A
|------------ | ------
| Regression | yes
| Version     | 2.10.0-dev

The following code is expected to work:

```php
var_dump(
    DriverManager::getConnection([
        'pdo' => new PDO('sqlite::memory:'),
    ])->ping()
);

// bool(true)
```
On `master`, it currently produces:
> Warning: assert(): assert($this->_conn instanceof DriverConnection) failed in lib/Doctrine/DBAL/Connection.php on line 1429
> 
> bool(true)

On `develop`:
> Warning: assert(): assert($this->_conn instanceof DriverConnection) failed in /home/morozov/Projects/dbal/lib/Doctrine/DBAL/Connection.php on line 1374
> 
> TypeError: Return value of Doctrine\DBAL\Connection::query() must be an instance of Doctrine\DBAL\Driver\ResultStatement, instance of PDOStatement returned in lib/Doctrine/DBAL/Connection.php on line 995

The existing implementation relies on weak/duck typing since instead of injecting the `PDO` instance into `PDOConnection`, the latter gets _replaced_. Besides the poor design, this "feature" is completely untested.

We can loosen or remove the assertion from `master` but before fixing it in `develop` I'd like to understand the use cases this feature was implemented for.

lcobucci:
> We can loosen or remove the assertion from master but before fixing it in develop I'd like to understand the use cases this feature was implemented for.

I think that loosing doesn't address the root issue. Passing a PDO instance is kind of a can of worms that we barely support.

There're a couple use cases that I can think of:

* having a PDO instance with arbitrary attributes: **imagine** that in a long long time ago (pre v2.x) the `driverOptions` argument was not a thing
* legacy software: reusing a connection opened outside of DBAL/ORM control (not opening multiple transactions and all)

I think that the way we implemented is definitely not ideal, IMHO we should rather have an adapter to our interface (that wraps the instance).

Now, looking at the future (v3 or v4), maybe we should drop it?

jwage:
I've seen this used quite a bit in apps when migrating to Doctrine. They'll have an existing PDO connection open already or some other ORM with a PDO connection. If we don't support it, then they would be forced to have two connections open. I talked to @morozov and I am going to take a look at properly refactoring this for 3.0.

morozov:
[...]

PR #3543, opened 2019-05-09
Fix wrong type assertion and annotation

nicolas-grekas:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

This issue introduced in #3443 breaks Symfony's test suite.

### DIFF
--- a/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
+++ b/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
@@ -10,9 +10,11 @@ use Doctrine\DBAL\FetchMode;
 use InvalidArgumentException;
 use IteratorAggregate;
 use PDO;
+use PDOStatement;
 use function array_merge;
 use function array_values;
 use function assert;
+use function is_array;
 use function reset;
 
 /**
@@ -42,7 +44,7 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement
     /** @var int */
     private $lifetime;
 
-    /** @var ResultStatement */
+    /** @var ResultStatement|PDOStatement */
     private $statement;
 
     /**
@@ -59,11 +61,12 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement
     private $defaultFetchMode = FetchMode::MIXED;
 
     /**
-     * @param string $cacheKey
-     * @param string $realKey
-     * @param int    $lifetime
+     * @param ResultStatement|PDOStatement $stmt
+     * @param                              string                       $cacheKey
+     * @param                              string                       $realKey
+     * @param                              int                          $lifetime
      */
-    public function __construct(ResultStatement $stmt, Cache $resultCache, $cacheKey, $realKey, $lifetime)
+    public function __construct($stmt, Cache $resultCache, $cacheKey, $realKey, $lifetime)
     {
         $this->statement   = $stmt;
         $this->resultCache = $resultCache;
@@ -167,7 +170,10 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement
      */
     public function fetchAll($fetchMode = null, $fetchArgument = null, $ctorArgs = null)
     {
-        $this->data    = $this->statement->fetchAll($fetchMode, $fetchArgument, $ctorArgs);
+        $data = $this->statement->fetchAll($fetchMode, $fetchArgument, $ctorArgs);
+        assert(is_array($data));
+
+        $this->data    = $data;
         $this->emptied = true;
 
         return $this->data;
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -20,6 +20,7 @@ use Doctrine\DBAL\Query\QueryBuilder;
 use Doctrine\DBAL\Schema\AbstractSchemaManager;
 use Doctrine\DBAL\Types\Type;
 use Exception;
+use PDO;
 use Throwable;
 use function array_key_exists;
 use function assert;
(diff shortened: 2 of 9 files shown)


jwage:
@morozov and I were just talking about this issue in #3487 and how we plan to improve this in 3.0. This is good for master.

jwage (approved):

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1424,12 +1424,12 @@ public function rollbackSavepoint($savepoint)
     /**
      * Gets the wrapped driver connection.
      *
-     * @return DriverConnection
+     * @return DriverConnection|\PDO
```
jwage: Can you use PDO?

jwage (changes requested):

lib/Doctrine/DBAL/Connection.php
nicolas-grekas: done

jwage:
Hmm, does phpstan not like that?

lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
```diff
@@ -59,11 +61,12 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement
     private $defaultFetchMode = FetchMode::MIXED;
 
     /**
-     * @param string $cacheKey
-     * @param string $realKey
-     * @param int    $lifetime
+     * @param ResultStatement|PDOStatement $stmt
```
alcaeus: I wanted to add an `assert` statement to the constructor to check for this, but apparently PHPStan thinks it's redundant. The code in question would be

```php
assert($stmt instanceof ResultStatement || $stmt instanceof PDOStatement);
```

Do we want to add it and silence the error or leave it as is?

alcaeus:
I've fixed the PHPStan failures that occurred due to allowing PDO classes in `Connection`. Unfortunately that means adding a bunch of `assert` calls as for example `fetchAll` will happily return `false` depending on the error handling mechanism. I don't think it's feasible to check for `false` and throw exceptions; at least not in the context of this PR.

lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
```diff
@@ -59,11 +61,12 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement
...
+     * @param                              string                       $cacheKey
+     * @param                              string                       $realKey
+     * @param                              int                          $lifetime
      */
-    public function __construct(ResultStatement $stmt, Cache $resultCache, $cacheKey, $realKey, $lifetime)
+    public function __construct($stmt, Cache $resultCache, $cacheKey, $realKey, $lifetime)
```
lcobucci: Even when passing the PDO instance we modify the statement class, so I'd say we shouldn't remove this type hint here.

alcaeus: PHPStan was complaining about it, that's why I had to modify it.

jwage (approved):

lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
nicolas-grekas: This should be removed and the type hint added back, isn't it?

lcobucci: I agree here, it looks like we're missing an assert on the caller and not here.

tests/Doctrine/Tests/DBAL/DriverManagerTest.php
```diff
@@ -38,11 +38,11 @@ public function testInvalidPdoInstance()
      */
     public function testValidPdoInstance()
     {
-        $conn = DriverManager::getConnection([
-            'pdo' => new PDO('sqlite::memory:'),
-        ]);
+        $pdo  = new PDO('sqlite::memory:');
```
lcobucci: Perhaps we need more tests covering the basic DBAL features for when an instance is used?

lcobucci (commented):
I feel like we're doing unnecessary things here and that we should try an alternative approach (wrapping). I'm working on it atm and we can compare the two PRs...

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -207,11 +210,12 @@ protected function _getPortableTableIndexesList($tableIndexes, $tableName = null
             $idx['primary']    = false;
             $idx['non_unique'] = ! $tableIndex['unique'];
 
-                $stmt       = $this->_conn->executeQuery(sprintf(
-                    'PRAGMA INDEX_INFO (%s)',
-                    $this->_conn->quote($keyName)
-                ));
-                $indexArray = $stmt->fetchAll(FetchMode::ASSOCIATIVE);
+            $stmt       = $this->_conn->executeQuery(sprintf(
```
SenseException:
```suggestion
            $stmt = $this->_conn->executeQuery(sprintf(
```

Because of the newline above this, I assume that the spaces aren't needed here.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -986,7 +994,7 @@ public function project($query, array $params, Closure $function)
     /**
      * Executes an SQL statement, returning a result set as a Statement object.
      *
-     * @return \Doctrine\DBAL\Driver\Statement
+     * @return DriverStatement|PDOStatement
```
SenseException: The Interface containing `query()` only expects a return type of `Doctrine\DBAL\Driver\Statement`. Should `PDOStatement` be added to the interface too?

SenseException (changes requested):

morozov:
Closing as no longer relevant.

PR #3544, opened 2019-05-10
Wrap pdo instances

lcobucci:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no (based on our doc blocks)
| Related issues | #3487 

#### Summary

This is an alternative of #3543 created by @alcaeus and @nicolas-grekas.

I'm not really happy naming-wise and am open for suggestions.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -13,6 +13,7 @@ use Doctrine\DBAL\Driver\PingableConnection;
 use Doctrine\DBAL\Driver\ResultStatement;
 use Doctrine\DBAL\Driver\ServerInfoAwareConnection;
 use Doctrine\DBAL\Driver\Statement as DriverStatement;
+use Doctrine\DBAL\Driver\WrappedPDOConnection;
 use Doctrine\DBAL\Exception\InvalidArgumentException;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Query\Expression\ExpressionBuilder;
@@ -189,7 +190,7 @@ class Connection implements DriverConnection
         $this->params  = $params;
 
         if (isset($params['pdo'])) {
-            $this->_conn       = $params['pdo'];
+            $this->_conn       = WrappedPDOConnection::fromInstance($params['pdo']);
             $this->isConnected = true;
             unset($this->params['pdo']);
         }
--- a/lib/Doctrine/DBAL/Driver/WrappedPDOConnection.php
+++ b/lib/Doctrine/DBAL/Driver/WrappedPDOConnection.php
@@ -1,0 +1,161 @@
+<?php
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Driver;
+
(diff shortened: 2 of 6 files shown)


lib/Doctrine/DBAL/Portability/Connection.php
```diff
@@ -66,7 +67,7 @@ public function connect()
             }
 
             if (isset($params['fetch_case']) && $this->portability & self::PORTABILITY_FIX_CASE) {
-                if ($this->_conn instanceof PDOConnection) {
+                if ($this->_conn instanceof PDOConnection || $this->_conn instanceof WrappedPDOConnection) {
```
SenseException: Is there a chance that `$this->_conn` can be of one type only?

lcobucci: Considering that modifying `PDOConnection` to use wrapped instances would be a BC-breaking change, I'd say no =/

I just modified things here too to be 100% safe that things work in a similar way to when we create the instance...

lib/Doctrine/DBAL/Connection.php
```diff
@@ -189,7 +190,7 @@ public function __construct(
         $this->params  = $params;
 
         if (isset($params['pdo'])) {
-            $this->_conn       = $params['pdo'];
+            $this->_conn       = WrappedPDOConnection::fromInstance($params['pdo']);
```
morozov: It still would be a breaking change since it's no longer a PDO instance (won't match a type hint, can't call PDO-specific methods, etc).

lib/Doctrine/DBAL/Driver/WrappedPDOConnection.php
```diff
@@ -0,0 +1,161 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Driver;
+
+use Doctrine\DBAL\ParameterType;
+use PDO;
+use function assert;
+use function func_get_args;
+
+final class WrappedPDOConnection implements Connection, ServerInfoAwareConnection
```
greg0ire: > I'm not really happy naming-wise and am open for suggestions.

`PDOConnectionAdapter`?

lib/Doctrine/DBAL/Connection.php
lcobucci: Not according to our docblocks...

lcobucci: We never expect to handle raw PDO connection in the internals

morozov: True. I didn't notice it was about the internal API. But still, even if we can afford introducing this class, I don't think we should. Having two PDO connections which share most of the code is not a proper solution. I'd like to propose #3549 instead.

lcobucci: @morozov I'm fine with not adding this adapter :smile: I'd suggest to port the tests to your PR, just to ensure that the basic stuff works fine.

lcobucci:
Closed as per https://github.com/doctrine/dbal/pull/3544#discussion_r286442106

PR #3547, opened 2019-05-20
Default column expressions do not work on SQL Server

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

The combination of #2960 and #3348 helped identify that the SQL Server platform doesn't properly generate DDL for the date/time columns with the default value described as an expression (see the change in `AbstractSQLServerPlatformTestCase`). Instead of using the expression as such, it converts it to a string literal.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/Platform/DefaultExpressionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Platform/DefaultExpressionTest.php
@@ -1,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL\Functional\Platform;
+
+use Doctrine\DBAL\FetchMode;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Platforms\MySqlPlatform;
+use Doctrine\DBAL\Platforms\OraclePlatform;
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\Types\Types;
+use Doctrine\Tests\DbalFunctionalTestCase;
+use function sprintf;
+
+class DefaultExpressionTest extends DbalFunctionalTestCase
+{
+    public function testCurrentDate() : void
+    {
+        $platform = $this->connection->getDatabasePlatform();
+
+        if ($platform instanceof MySqlPlatform) {
+            self::markTestSkipped('Not supported on MySQL');
+        }
+
+        $this->assertDefaultExpression(Types::DATE_MUTABLE, static function (AbstractPlatform $platform) : string {
+            return $platform->getCurrentDateSQL();
+        });
+    }
+
+    public function testCurrentTime() : void
+    {
+        $platform = $this->connection->getDatabasePlatform();
+
+        if ($platform instanceof MySqlPlatform) {
+            self::markTestSkipped('Not supported on MySQL');
+        }
+
+        if ($platform instanceof OraclePlatform) {
+            self::markTestSkipped('Not supported on Oracle');
+        }
+
+        $this->assertDefaultExpression(Types::TIME_MUTABLE, static function (AbstractPlatform $platform) : string {
+            return $platform->getCurrentTimeSQL();
+        });
+    }
+
+    public function testCurrentTimestamp() : void
+    {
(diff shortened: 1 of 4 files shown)


tests/Doctrine/Tests/DBAL/Functional/Schema/SQLServerSchemaManagerTest.php
```diff
@@ -75,12 +73,9 @@ public function testDefaultConstraints()
         self::assertEquals('Doctrine rocks!!!', $columns['df_string_2']->getDefault());
         self::assertEquals('another default value', $columns['df_string_3']->getDefault());
         self::assertEquals(1, $columns['df_boolean']->getDefault());
-        self::assertSame($platform->getCurrentDateSQL(), $columns['df_current_date']->getDefault());
-        self::assertSame($platform->getCurrentTimeSQL(), $columns['df_current_time']->getDefault());
```
morozov: During introspection, the string representation of the actual expression on the column doesn't necessarily match the expression in the schema definition. They only used to match because of the bug.

While the expression being deployed looks like `CONVERT(date, GETDATE())`, the introspected expression could be `CONVERT([date],getdate())` (type is quoted, no whitespace after the colon, lowerr-cased function name) or even [`CONVERT([date],getdate(),0)`](https://ci.appveyor.com/project/doctrine/dbal/build/job/75o41rjwc4b3u291#L274) depending on the platform version.

tests/Doctrine/Tests/DBAL/Platforms/AbstractSQLServerPlatformTestCase.php
```diff
@@ -1496,7 +1496,7 @@ public function testGetDefaultValueDeclarationSQLForDateType() : void
             ];
 
             self::assertSame(
-                " DEFAULT '" . $currentDateSql . "'",
```
morozov: 🤔

tests/Doctrine/Tests/DBAL/Functional/Platform/DefaultExpressionTest.php
```diff
@@ -0,0 +1,82 @@
...
+
+        yield 'current-time' => [
+            Types::TIME_MUTABLE,
+            static function (AbstractPlatform $platform) : string {
+                if ($platform instanceof MySqlPlatform) {
+                    self::markTestSkipped('Not supported on MySQL');
+                }
+
+                if ($platform instanceof OraclePlatform) {
+                    self::markTestSkipped('Not supported on Oracle');
```
Ocramius: Uhh, I don't think we should skip tests from code within the data-provider: confusing approach IMO

Ocramius (dismissed):
Besides issues with the test implementation, I don't see anything wrong with this change

tests/Doctrine/Tests/DBAL/Functional/Platform/DefaultExpressionTest.php
morozov: We can move this test under `AbstractPlatformTest` and split it into three, one for each expression. Then, in each concrete class, skip the corresponding method if the platform doesn't support it.

jwage (dismissed):
💯

SenseException (changes requested):
There's isn't much I can add here.

tests/Doctrine/Tests/DBAL/Functional/Platform/DefaultExpressionTest.php
```diff
@@ -0,0 +1,82 @@
...
+        $this->connection->getSchemaManager()->dropAndCreateTable($table);
+
+        $query = sprintf(
+            'INSERT INTO default_expr_test (actual_value) VALUES (%s)',
+            $defaultSql
+        );
+
+        $this->connection->exec($query);
+
+        $query                        = sprintf('SELECT default_value, actual_value FROM default_expr_test');
```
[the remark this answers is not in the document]
morozov: It's a copy-paste of the previous query. Will fix.

tests/Doctrine/Tests/DBAL/Functional/Schema/SQLServerSchemaManagerTest.php
morozov: This is already tested in `DefaultExpressionTest`.

SenseException (approved):

SenseException:
When the build goes green, then this PR is :+1:

PR #3558, opened 2019-05-24
Deprecated Driver::getName()

morozov:
The method is not used for anything else than skipping tests for specific drivers. Cross-driver portability should be established by drivers, not outside of them based on their name. The method has been removed in #3553.

|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOConnectionTest.php
@@ -4,9 +4,13 @@ namespace Doctrine\Tests\DBAL\Functional\Driver;
 
 use Doctrine\DBAL\Driver\PDOConnection;
 use Doctrine\DBAL\Driver\PDOException;
+use Doctrine\DBAL\Driver\PDOOracle\Driver as PDOOracleDriver;
+use Doctrine\DBAL\Driver\PDOPgSql\Driver as PDOPgSQLDriver;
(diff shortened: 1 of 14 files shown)


jwage (dismissed):
💯

tests/Doctrine/Tests/DBAL/Functional/Driver/PDOConnectionTest.php
```diff
@@ -66,29 +70,31 @@ public function testThrowsWrappedExceptionOnExec()
...
         }
 
+        // Some PDO adapters do not check the query server-side
+        // even though emulated prepared statements are disabled,
+        // so an exception is thrown only eventually.
+        if ($driver instanceof PDOOracleDriver
+            || $driver instanceof PDOPgSQLDriver
+        ) {
+            self::markTestSkipped(sprintf(
+                'The underlying implementation of the "%s" driver does not check the query to be prepared server-side.',
```
jwage: The class name shouldn't be wrapped in double quotes right since it is a symbol?

morozov: No, it shouldn't 😬

jwage: hehe, nice. Makes me feel better about how many times I got it wrong in my previous PR where you had to point it out to me.

jwage (approved):

PR #3560, opened 2019-05-25
Enforced parameter and return value types in Driver classes

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

The rationale behind the breaking changes is the following:

1. Most of the underlying drivers do not allow NULL for `$username` or `$password`, therefore if we allow NULL, we'll have to replace it with an empty string in most of the implementation. At the same time, fro the drivers which support NULL (e.g. PDO) there's no difference between NULL and empty string.
2. Some drivers (e.g. `mysqli`, `pdo_mysql`, `pdo_sqlite`) allow establishing a connection without selecting a database, therefore it should be possible to indicate the absence of a value in `getDatabase()`. This is important because consumers (mostly, schema managers) should handle the situation of "no database selected" (which they currently don't) and raise an error instead of returning an empty list of assets.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,10 @@
 # Upgrade to 3.0
 
+## BC BREAK: Changes in the `Doctrine\DBAL\Driver` API
+
+1. The `$username` and `$password` arguments of `::connect()` are no longer nullable. Use an empty string to indicate empty username or password.
+2. The return value of `::getDatabase()` has been documented as nullable since some of the drivers allow establishing a connection without selecting a database.
+
 ## BC BREAK: `Doctrine\DBAL\Driver::getName()` removed
 
 The `Doctrine\DBAL\Driver::getName()` has been removed.
--- a/lib/Doctrine/DBAL/Driver.php
+++ b/lib/Doctrine/DBAL/Driver.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\DBAL;
 
+use Doctrine\DBAL\Driver\Connection as DriverConnection;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Schema\AbstractSchemaManager;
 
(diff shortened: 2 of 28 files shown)


UPGRADE.md
```diff
@@ -1,5 +1,10 @@
 # Upgrade to 3.0
 
+## BC BREAK: Changes in the `Doctrine\DBAL\Driver` API
+
+1. The `$username` and `$password` arguments of `::connect()` are no longer nullable. Use empty string to indicate empty username or password.
```
greg0ire: Either
```suggestion
1. The `$username` and `$password` arguments of `::connect()` are no longer nullable. Use an empty string to indicate empty username or password.
```
or 
```suggestion
1. The `$username` and `$password` arguments of `::connect()` are no longer nullable. Use empty strings to indicate empty username or password.
```

UPGRADE.md
```diff
@@ -1,5 +1,10 @@
...
+2. The return value of `::getDatabase()` has been documented as nullable since some of the drivers allow establishing connection without selecting a database.
```
```suggestion
2. The return value of `::getDatabase()` has been documented as nullable since some of the drivers allow establishing a connection without selecting a database.
```

lib/Doctrine/DBAL/Driver.php
```diff
@@ -16,35 +17,38 @@ interface Driver
...
-     * @return AbstractSchemaManager
      */
-    public function getSchemaManager(Connection $conn);
+    public function getSchemaManager(Connection $conn) : AbstractSchemaManager;
 
     /**
      * Gets the name of the database connected to for this driver.
      *
-     * @return string The name of the database.
+     * @return string The name of the database or NULL if no database currently selected.
```
greg0ire:
```suggestion
     * @return string The name of the database or NULL if no database is currently selected.
```

morozov:
Thank you @greg0ire. I wish I could use Grammarly with PhpStorm.

jwage (approved):
Hooray!

PR #3562, opened 2019-05-28
Enforced parameter and return value types in Platform classes

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

**Notes:**

1. The nullability of the `$database` argument in `AbstractPlatform` methods is not final. It just reflects the current type. The type can be restricted to a non-nullable `string` in a subsequent PR.
2. Some mocks have been reworked from using the PHPUnit mock API instead of Prophecy because the latter doesn't respect the return types of the mocked methods.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractSQLiteDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractSQLiteDriver.php
@@ -83,7 +83,7 @@ abstract class AbstractSQLiteDriver implements Driver, ExceptionConverterDriver
     {
         $params = $conn->getParams();
 
-        return $params['path'] ?? null;
+        return $params['path'] ?? '';
     }
 
     /**
--- a/lib/Doctrine/DBAL/Platforms/DB2Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/DB2Platform.php
@@ -33,7 +33,7 @@ class DB2Platform extends AbstractPlatform
     /**
      * {@inheritdoc}
      */
-    public function getBinaryMaxLength()
+    public function getBinaryMaxLength() : int
     {
         return 32704;
     }
@@ -41,7 +41,7 @@ class DB2Platform extends AbstractPlatform
     /**
      * {@inheritdoc}
      */
-    public function getBinaryDefaultLength()
+    public function getBinaryDefaultLength() : int
     {
         return 1;
     }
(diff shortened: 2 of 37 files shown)


lib/Doctrine/DBAL/Driver/AbstractSQLiteDriver.php
```diff
@@ -83,7 +83,7 @@ public function getDatabase(Connection $conn) : ?string
     {
         $params = $conn->getParams();
 
-        return $params['path'] ?? null;
+        return $params['path'] ?? '';
```
morozov: The returned NULL implies that there's no database currently selected, however, SQLite doesn't support databases and one is always selected. Some of the platform methods do not allow NULL for `$database` which previously was only documented but not enforced.

lib/Doctrine/DBAL/Platforms/DB2Platform.php
```diff
@@ -269,13 +269,8 @@ public function getTruncateTableSQL($tableName, $cascade = false)
...
     /**
      * This code fragment is originally from the Zend_Db_Adapter_Db2 class, but has been edited.
-     *
-     * @param string $table
-     * @param string $database
-     *
-     * @return string
      */
-    public function getListTableColumnsSQL($table, $database = null)
+    public function getListTableColumnsSQL(string $table, ?string $database = null) : string
```
morozov: The nullability of the `$database` argument in `AbstractPlatform` methods is not final. It just reflects the current type. The type can be restricted to a non-nullable `string` in a subsequent PR.

tests/Doctrine/Tests/DBAL/Functional/Driver/PDOSqlite/DriverTest.php
```diff
@@ -25,6 +25,14 @@ protected function setUp() : void
         $this->markTestSkipped('pdo_sqlite only test.');
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function testReturnsDatabaseNameWithoutDatabaseNameParameter()
+    {
+        $this->markTestSkipped('SQLite does not support the concept of a database.');
```
morozov: Previously, this test would only pass with an in-memory SQLite database for which the `Driver::getDatabase()` would return NULL. The test is invalid with SQLite by design.

tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php
```diff
@@ -884,10 +875,6 @@ public function testGeneratesSequenceSqlCommands()
             'SELECT myseq.NEXTVAL',
             $this->platform->getSequenceNextValSQL('myseq')
         );
-        self::assertEquals(
-            'SELECT sequence_name, increment_by, start_with, min_value FROM SYS.SYSSEQUENCE',
-            $this->platform->getListSequencesSQL(null)
```
morozov: This is no longer valid.

jwage (approved):
💯

PR #3565, opened 2019-05-28
Add types to Schema namespace

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

#### Summary

Improving types in `Doctrine\DBAL\Schema`.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,17 @@
 # Upgrade to 3.0
 
+## BC BREAK: Changes in the `Doctrine\DBAL\Schema` API
+
+- Method `Doctrine\DBAL\Schema\AbstractSchemaManager::_getPortableViewDefinition()` no longer optionally returns false. It will always return a `Doctrine\DBAL\Schema\View` instance.
+- Method `Doctrine\DBAL\Schema\Comparator::diffTable()` now optionally returns null instead of false.
+- Property `Doctrine\DBAL\Schema\TableDiff::$newName` is now optionally null instead of false.
+- Method `Doctrine\DBAL\Schema\AbstractSchemaManager::tablesExist()` no longer accepts a string. Use `Doctrine\DBAL\Schema\AbstractSchemaManager::tableExists()` instead.
+- Method `Doctrine\DBAL\Schema\OracleSchemaManager::createDatabase()` no longer accepts `null` for `$database` argument.
+- Removed unused method `Doctrine\DBAL\Schema\AbstractSchemaManager::_getPortableFunctionsList()`
+- Removed unused method `Doctrine\DBAL\Schema\AbstractSchemaManager::_getPortableFunctionDefinition()`
+- Removed unused method `Doctrine\DBAL\Schema\OracleSchemaManager::_getPortableFunctionDefinition()`
+- Removed unused method `Doctrine\DBAL\Schema\SqliteSchemaManager::_getPortableTableIndexDefinition()`
+
 ## BC BREAK: Changes in the `Doctrine\DBAL\Driver` API
 
 1. The `$username` and `$password` arguments of `::connect()` are no longer nullable. Use an empty string to indicate empty username or password.
--- a/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
@@ -671,7 +671,7 @@ class SqlitePlatform extends AbstractPlatform
         $sql       = [];
         $tableName = $diff->getNewName();
 
-        if ($tableName === false) {
+        if ($tableName === null) {
             $tableName = $diff->getName($this);
         }
 
@@ -898,7 +898,7 @@ class SqlitePlatform extends AbstractPlatform
 
             $newName = $diff->getNewName();
 
-            if ($newName !== false) {
+            if ($newName !== null) {
                 $sql[] = sprintf(
                     'ALTER TABLE %s RENAME TO %s',
                     $newTable->getQuotedName($this),
@@ -980,7 +980,7 @@ class SqlitePlatform extends AbstractPlatform
         }
 
         if (! $this->onSchemaAlterTable($diff, $tableSql)) {
-            if ($diff->newName !== false) {
+            if ($diff->newName !== null) {
                 $newTable = new Identifier($diff->newName);
                 $sql[]    = 'ALTER TABLE ' . $table->getQuotedName($this) . ' RENAME TO ' . $newTable->getQuotedName($this);
             }
--- a/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
@@ -10,12 +10,14 @@ use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Event\SchemaColumnDefinitionEventArgs;
 use Doctrine\DBAL\Event\SchemaIndexDefinitionEventArgs;
 use Doctrine\DBAL\Events;
+use Doctrine\DBAL\Exception\DatabaseRequired;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Platforms\Exception\NotSupported;
 use Throwable;
 use function array_filter;
 use function array_intersect;
 use function array_map;
+use function array_shift;
 use function array_values;
 use function assert;
 use function call_user_func_array;
@@ -57,10 +59,8 @@ abstract class AbstractSchemaManager
 
     /**
      * Returns the associated platform.
-     *
-     * @return AbstractPlatform
      */
-    public function getDatabasePlatform()
+    public function getDatabasePlatform() : AbstractPlatform
     {
         return $this->_platform;
     }
@@ -97,9 +97,9 @@ abstract class AbstractSchemaManager
     /**
      * Lists the available databases for this connection.
      *
-     * @return string[]
+     * @return array<int, string>
      */
-    public function listDatabases()
+    public function listDatabases() : array
     {
         $sql = $this->_platform->getListDatabasesSQL();
 
@@ -111,9 +111,9 @@ abstract class AbstractSchemaManager
     /**
      * Returns a list of all namespaces in the current database.
      *
-     * @return string[]
+     * @return array<int, string>
      */
-    public function listNamespaceNames()
+    public function listNamespaceNames() : array
     {
         $sql = $this->_platform->getListNamespacesSQL();
 
@@ -125,15 +125,15 @@ abstract class AbstractSchemaManager
     /**
      * Lists the available sequences for this connection.
      *
-     * @param string|null $database
-     *
-     * @return Sequence[]
+     * @return array<int, Sequence>
      */
-    public function listSequences($database = null)
+    public function listSequences(?string $database = null) : array
     {
-        if ($database === null) {
-            $database = $this->_conn->getDatabase();
-        }
+        $database = $this->ensureDatabase(
+            $database ?? $this->_conn->getDatabase(),
+            __METHOD__
+        );
+
         $sql = $this->_platform->getListSequencesSQL($database);
 
         $sequences = $this->_conn->fetchAll($sql);
@@ -151,16 +151,14 @@ abstract class AbstractSchemaManager
      * of a table. We're a RDBMS specifies more details these are held
      * in the platformDetails array.
      *
-     * @param string      $table    The name of the table.
-     * @param string|null $database
-     *
-     * @return Column[]
+     * @return array<string, Column>
      */
-    public function listTableColumns($table, $database = null)
+    public function listTableColumns(string $table, ?string $database = null) : array
     {
-        if (! $database) {
-            $database = $this->_conn->getDatabase();
-        }
+        $database = $this->ensureDatabase(
+            $database ?? $this->_conn->getDatabase(),
+            __METHOD__
+        );
 
         $sql = $this->_platform->getListTableColumnsSQL($table, $database);
 
@@ -174,11 +172,9 @@ abstract class AbstractSchemaManager
      *
      * Keys of the portable indexes list are all lower-cased.
      *
-     * @param string $table The name of the table.
-     *
-     * @return Index[]
+     * @return array<string, Index>
      */
-    public function listTableIndexes($table)
+    public function listTableIndexes(string $table) : array
     {
         $sql = $this->_platform->getListTableIndexesSQL($table, $this->_conn->getDatabase());
 
@@ -190,23 +186,26 @@ abstract class AbstractSchemaManager
     /**
      * Returns true if all the given tables exist.
      *
-     * @param string|string[] $tableNames
-     *
-     * @return bool
+     * @param array<int, string> $tableNames
      */
-    public function tablesExist($tableNames)
+    public function tablesExist(array $tableNames) : bool
     {
-        $tableNames = array_map('strtolower', (array) $tableNames);
+        $tableNames = array_map('strtolower', $tableNames);
 
         return count($tableNames) === count(array_intersect($tableNames, array_map('strtolower', $this->listTableNames())));
+    }
+
+    public function tableExists(string $tableName) : bool
+    {
+        return $this->tablesExist([$tableName]);
     }
 
     /**
      * Returns a list of all tables in the current database.
      *
-     * @return string[]
+     * @return array<int, string>
      */
-    public function listTableNames()
+    public function listTableNames() : array
     {
         $sql = $this->_platform->getListTablesSQL();
 
@@ -220,11 +219,11 @@ abstract class AbstractSchemaManager
      * Filters asset names if they are configured to return only a subset of all
      * the found elements.
      *
-     * @param mixed[] $assetNames
+     * @param array<int, mixed> $assetNames
      *
-     * @return mixed[]
+     * @return array<int, mixed>
      */
-    protected function filterAssetNames($assetNames)
+    protected function filterAssetNames(array $assetNames)
     {
         $filter = $this->_conn->getConfiguration()->getSchemaAssetsFilter();
         if (! $filter) {
@@ -237,9 +236,9 @@ abstract class AbstractSchemaManager
     /**
      * Lists the tables for this connection.
      *
-     * @return Table[]
+     * @return array<int, Table>
      */
-    public function listTables()
+    public function listTables() : array
     {
         $tableNames = $this->listTableNames();
 
@@ -251,12 +250,7 @@ abstract class AbstractSchemaManager
         return $tables;
     }
 
-    /**
-     * @param string $tableName
-     *
-     * @return Table
-     */
-    public function listTableDetails($tableName)
+    public function listTableDetails(string $tableName) : Table
     {
         $columns     = $this->listTableColumns($tableName);
         $foreignKeys = [];
@@ -273,13 +267,17 @@ abstract class AbstractSchemaManager
     /**
      * Lists the views this connection has.
      *
-     * @return View[]
+     * @return array<string, View>
      */
-    public function listViews()
+    public function listViews() : array
     {
-        $database = $this->_conn->getDatabase();
-        $sql      = $this->_platform->getListViewsSQL($database);
-        $views    = $this->_conn->fetchAll($sql);
+        $database = $this->ensureDatabase(
+            $this->_conn->getDatabase(),
+            __METHOD__
+        );
+
+        $sql   = $this->_platform->getListViewsSQL($database);
+        $views = $this->_conn->fetchAll($sql);
 
         return $this->_getPortableViewsList($views);
     }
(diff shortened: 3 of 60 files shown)


lib/Doctrine/DBAL/Event/SchemaColumnDefinitionEventArgs.php
```diff
@@ -33,9 +33,9 @@ class SchemaColumnDefinitionEventArgs extends SchemaEventArgs
     private $connection;
 
     /**
-     * @param mixed[] $tableColumn
-     * @param string  $table
-     * @param string  $database
+     * @param mixed[]     $tableColumn
+     * @param string      $table
+     * @param string|null $database
```
morozov: Is it possible to not declare it nullable? A column event outside of a database doesn't make much sense.

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -897,7 +897,7 @@ public function getAlterTableSQL(TableDiff $diff)
         if (! $this->onSchemaAlterTable($diff, $tableSql)) {
             $dataTable = new Table('__temp__' . $table->getName());
 
-            $newTable = new Table($table->getQuotedName($this), $columns, $this->getPrimaryIndexInAlteredTable($diff), [], $this->getForeignKeysInAlteredTable($diff), $table->getOptions());
+            $newTable = new Table($table->getQuotedName($this), array_values($columns), $this->getPrimaryIndexInAlteredTable($diff), [], $this->getForeignKeysInAlteredTable($diff), $table->getOptions());
```
morozov: After switching from `array<int, Column>` to `array<int|string, Column>`, this is probably no longer needed.

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -97,9 +96,9 @@ public function tryMethod()
     /**
      * Lists the available databases for this connection.
      *
-     * @return string[]
+     * @return array<int, string>
```
morozov: This is more of a leading question than a suggestion: do we really need to enforce the type of the key if it's not used? If enforced, it may force unnecessary changes to already correct code (I'm referring to the newly introduced `array_values()`).

lib/Doctrine/DBAL/Schema/SQLServerSchemaManager.php
```diff
@@ -56,15 +56,15 @@ public function dropDatabase($database)
     /**
      * {@inheritdoc}
      */
-    protected function _getPortableSequenceDefinition($sequence)
+    protected function _getPortableSequenceDefinition($sequence) : Sequence
```
morozov: The argument type doesn't match the parent definition. PhpStorm in the PHP 7.1 mode reports such violations but it doesn't in PHP 7.2 mode.

tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
```diff
@@ -453,7 +453,7 @@ public function testCompareSequences()
         $seq1 = new Sequence('foo', 1, 1);
         $seq2 = new Sequence('foo', 1, 2);
         $seq3 = new Sequence('foo', 2, 1);
-        $seq4 = new Sequence('foo', '1', '1');
+        $seq4 = new Sequence('foo', 1, 1);
```
morozov: Now, this is a duplicate of `$seq1`.

lib/Doctrine/DBAL/Schema/Visitor/Graphviz.php
```diff
@@ -116,13 +110,9 @@ private function createNode($name, $options)
...
     /**
-     * @param string   $node1
-     * @param string   $node2
-     * @param string[] $options
-     *
-     * @return string
+     * @param array<string, string> $options
      */
-    private function createNodeRelation($node1, $node2, $options)
+    private function createNodeRelation(string $node1, string $node2, $options) : string
```
morozov: Could the type of `$options` be enforced as well?

lib/Doctrine/DBAL/Schema/TableDiff.php
```diff
@@ -130,25 +129,16 @@ public function __construct(
...
-     */
-    public function getNewName()
+    public function getNewName() : ?Identifier
     {
-        if ($this->newName === false) {
-            return false;
-        }
-
-        return new Identifier($this->newName);
+        return $this->newName !== null ? new Identifier($this->newName) : null;
```
morozov: I'd rather leave the `if` instead of the ternary for better readability and more accurate code coverage.

lib/Doctrine/DBAL/Schema/Sequence.php
```diff
@@ -22,74 +22,44 @@ class Sequence extends AbstractAsset
...
-    /**
-     * @param int $initialValue
-     *
-     * @return \Doctrine\DBAL\Schema\Sequence
-     */
-    public function setInitialValue($initialValue)
+    public function setInitialValue(int $initialValue) : self
     {
-        $this->initialValue = (int) $initialValue ?: 1;
+        $this->initialValue = $initialValue ?: 1;
```
morozov: Is the default value still relevant? In the past, it might cover some NULL value which is no longer relevant.

lib/Doctrine/DBAL/Schema/Sequence.php
```diff
@@ -22,74 +22,44 @@ class Sequence extends AbstractAsset
...
-    /**
-     * @param int $allocationSize
-     *
-     * @return \Doctrine\DBAL\Schema\Sequence
-     */
-    public function setAllocationSize($allocationSize)
+    public function setAllocationSize(int $allocationSize) : self
     {
-        $this->allocationSize = (int) $allocationSize ?: 1;
+        $this->allocationSize = $allocationSize ?: 1;
```
morozov: Is the default value still relevant? In the past, it might cover some NULL value which is no longer relevant.

morozov (commented):
Besides the need to introduce `array_values()`, looks solid. I'll see if I can run some more consistency checks on the changes later.

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
jwage: Hmm, good question. I think yes, because without it, if you introduce new code that uses that key, phpstan won't validate that you are using the type of the key correctly.

lib/Doctrine/DBAL/Event/SchemaColumnDefinitionEventArgs.php
jwage: I will give it a try. I didn't change any signatures so far, just documented what it is today as best as I could tell.

jwage: It is because of this and this https://github.com/doctrine/dbal/blob/develop/lib/Doctrine/DBAL/Driver.php#L53 and https://github.com/doctrine/dbal/blob/develop/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php#L159

morozov: Right. Please use something like 933b850a88b1151430046fe56908389cc57f18c1. If there's no DB specified in the connection parameters and not DB is currently selected, the schema manager won't be able to fetch any assets anyways. It should be an error.

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
morozov: But you don't know the type of the key before you introduce it. And once you do, if the signature says a method accepts `array<int|string, T>`, it will have to be able to handle both `int` and `string` keys correctly.

morozov: Another reason to not introduce key types just everywhere is that `array<K, T>` implies that the method accepts (or returns) a map (correspondence) between keys and values and the order of the key-value pairs is irrelevant (or is at least secondary). While `T[]` means that it's an ordered list of `T` and the order matters while the keys don't.

lib/Doctrine/DBAL/Event/SchemaColumnDefinitionEventArgs.php
jwage: Can we make that behavior change in another PR?

morozov: I'd rather make it now in order to not widen the type constraint. I wouldn't object if it already was `?string`. If it's an error reported by PHPStan, it's fine to temporarily suppress the error.

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
jwage: Hmm, I am not sure. I started doing this because @Ocramius suggested it in another pull request. Without the key type, the user can pass in anything they want. Adding the type made sense because it makes the usage explicit.

morozov: Which is fine as long as the key is not used (`T[]`). Having to put `array_values()` everywhere IMO indicates a problem.

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -973,4 +978,13 @@ public function getSchemaSearchPaths() : array
 
         return $match[2];
     }
+
+    private function ensureDatabase(?string $database) : string
+    {
+        if ($database === null) {
+            throw new InvalidArgumentException('Needs a database.');
```
morozov: Later, it will need a special exception.

tests/Doctrine/Tests/DBAL/Functional/Driver/AbstractDriverTest.php
```diff
@@ -69,6 +69,6 @@ public function testReturnsDatabaseNameWithoutDatabaseNameParameter()
 
     protected static function getDatabaseNameForConnectionWithoutDatabaseNameParameter() : ?string
     {
-        return null;
+        return '';
```
morozov: The test makes sense for the abstract driver because this is how a driver indicates that it's not connected to a DB. The test needs to be skilled in SQLite driver because it doesn't support databases. If not skipped, it'd fail with a non-in-memory database (the driver will return the database path, not null).

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
jwage: I will go ahead and add this now.

UPGRADE.md
```diff
@@ -1,5 +1,12 @@
 # Upgrade to 3.0
 
+## BC BREAK: Changes in the `Doctrine\DBAL\Schema` API
+
+- Method `Doctrine\DBAL\Schema\AbstractSchemaManager::_getPortableViewDefinition()` now optionally returns null instead of false.
```
morozov: Is there a valid case where it would return `false`? In my version, the default behavior was `throw DBALException::notSupported('Views');`.

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
```diff
@@ -288,7 +278,7 @@ protected function _getPortableDatabaseDefinition($database)
     /**
      * {@inheritdoc}
      */
-    public function createDatabase($database = null)
+    public function createDatabase(?string $database = null) : void
```
morozov: `$database` should be non-nullable same as in the parent method.

lib/Doctrine/DBAL/Schema/Schema.php
```diff
@@ -282,23 +245,19 @@ public function getSequence($sequenceName)
...
     {
         return $this->_sequences;
     }
 
     /**
      * Creates a new namespace.
      *
-     * @param string $namespaceName The name of the namespace to create.
-     *
-     * @return \Doctrine\DBAL\Schema\Schema This schema instance.
```
morozov: It'd make sense to keep it annotated as `@return $this`. See #2384 for the reference.

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -941,27 +828,27 @@ protected function _getPortableUsersList($users)
...
+     * @return array<string, View>
      */
-    protected function _getPortableViewsList($views)
+    protected function _getPortableViewsList(array $views) : array
     {
         $list = [];
         foreach ($views as $value) {
             $view = $this->_getPortableViewDefinition($value);
 
-            if (! $view) {
```
morozov: This doesn't look like a valid case. All platforms will create a view from the definition if they support views.

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
jwage: Do you know why it was allowed to be null for Oracle? I assumed there was a reason and I thought I remembered reading something somewhere in the code about it.

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
jwage: I will see if I can get rid of this and remove the nullable on the method too.

UPGRADE.md
jwage: I will see if I can work that out.

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
Ocramius: If we can guarantee `array<int, T>`, then it should be like that, as it allows downstream to use it in variadic calls. Effectively, the `int` or `string` key is something that will be a BC break downstream regardless, if we change it, so if we can guarantee it by declaring it in first place, that is a win for us.

Ocramius: To make the above comment more explicit: having `array<int|string, T>` and changing implementation in the method to return `array<int, T>` to `array<string, T>` is technically not a BC break, but in practice it is one for downstream. If we can restrict types under our control, we should do so to allow these BC breaks to become explicit (when the docblock changes)

UPGRADE.md
jwage: Good call. Fixed here https://github.com/doctrine/dbal/pull/3565/commits/02d4639428a6714ad8fa2a57869916440cffa19b

lib/Doctrine/DBAL/Schema/Schema.php
jwage: Fixed in https://github.com/doctrine/dbal/pull/3565/commits/05eca9944e37463a6ff0636c8c1c89a081321402

lib/Doctrine/DBAL/Schema/OracleSchemaManager.php
jwage: Fixed in https://github.com/doctrine/dbal/pull/3565/commits/733e1b6f792adcaf3894cba2a55c25b35686e14b

lib/Doctrine/DBAL/Schema/Visitor/CreateSchemaSqlCollector.php
```diff
@@ -47,15 +48,15 @@ public function acceptNamespace($namespaceName)
     /**
      * {@inheritdoc}
      */
-    public function acceptTable(Table $table)
+    public function acceptTable(Table $table) : void
     {
-        $this->createTableQueries = array_merge($this->createTableQueries, $this->platform->getCreateTableSQL($table));
+        $this->createTableQueries = array_merge($this->createTableQueries, array_values($this->platform->getCreateTableSQL($table)));
```
morozov: @Ocramius my major concern about adding `array<int, T>` is the need for such `array_values()` calls here and there. Specifically, in this case, it seems to massage the `string[]` returned by `AbstractPlatform::getCreateTableSQL()` to `array<int, string>` (the type of `$this->createTableQueries`) for no real reason.

lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -57,17 +57,16 @@ class Table extends AbstractAsset
     protected $_schemaConfig = null;
 
     /**
-     * @param string                 $tableName
-     * @param Column[]               $columns
-     * @param Index[]                $indexes
-     * @param UniqueConstraint[]     $uniqueConstraints
-     * @param ForeignKeyConstraint[] $fkConstraints
-     * @param mixed[]                $options
+     * @param array<int|string, Column>               $columns
```
morozov: @Ocramius as for the definitions like this, it's not clear for the caller what an `int` or `string` keys mean. How does the caller know which keys to use in order to pass the columns to the method, and what is the meaning of those keys? The method just iterates over the array and doesn't use the key at all.

As a rule of thumb, for a union type, you should be able to explain the difference between the types and when to use what. E.g. `function bindParam(int|string $param, mixed $value)` means that the parameter is either bound by its position (`int`) or name (`string`). In the case of `array<int|string, Column>`, `int` and `string` keys mean "nothing" and "nothing" respectively.

jwage: I think these cases should probably be fixed so it is only `array<int|string, Column>`. I only documented the way it is today and I think it is that way today just by chance and inconsistency due to there not being any types before.

morozov: I don't agree with `array<int, Column>` in this case either. If a caller currently passes `array<string, Column>`, this constraint will force them to use `array_values()` for no real reason (the original code works just fine) — same as it's being added in this pull request.

The bottom line is, I wouldn't like us to enforce the type of the variables (keys) which we don't use..

jwage: I can see both sides and I could go either way. I would rather get this merged so I will change it back.

morozov: I can accept the `array<>` approach as long as the patch doesn't contain unnecessary calls to `array_values()`. They do indicate the problem and I'd rather have it fixed than masked.

jwage: Take a look at this commit https://github.com/doctrine/dbal/pull/3565/commits/125265bd436523fee013e38c7e57d29b3c275995. I backed out the ones where we don't use the key and the type causes the 
 need to use `array_values`.

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -100,8 +106,15 @@ public function dropAndCreateForeignKey(ForeignKeyConstraint $foreignKey, $table
      */
     public function dropForeignKey($foreignKey, $table) : void
     {
-        $tableDiff                       = $this->getTableDiffForAlterForeignKey($table);
-        $tableDiff->removedForeignKeys[] = $foreignKey;
+        $table = $this->ensureTable($table);
```
jwage: These changes were made to fix this type https://github.com/doctrine/dbal/pull/3565/commits/125265bd436523fee013e38c7e57d29b3c275995#diff-6b78caddb762256c1425fe6c316e07dcL93

lib/Doctrine/DBAL/Schema/SchemaDiff.php
```diff
@@ -165,6 +164,6 @@ protected function _toSql(AbstractPlatform $platform, $saveMode = false)
             $sql = array_merge($sql, $platform->getAlterTableSQL($tableDiff));
         }
 
-        return $sql;
+        return array_values($sql);
```
morozov: Is it still needed? I'd rather mark the required `AbstractPlatform` methods as returning `array<int, string>` (even though that's a different namespace) than call `array_values()` at runtime. I want to avoid the introduction of unnecessary `array_values()` because later, after all types are settled, it will be hard to figure out which of them are no longer needed.

UPGRADE.md
```diff
@@ -1,5 +1,13 @@
 # Upgrade to 3.0
 
+## BC BREAK: Changes in the `Doctrine\DBAL\Schema` API
+
+- Method `Doctrine\DBAL\Schema\AbstractSchemaManager::_getPortableViewDefinition()` no longer optionally returns false. It will always return a `Doctrine\DBAL\Schema\View` instance.
```
morozov: Nitpicking: although it's a change, it's not a breaking change.

morozov (commented):
The recent changes look fine. We're almost there.

lib/Doctrine/DBAL/Schema/SchemaDiff.php
jwage: Should be able to kill it. I'll make another commit.

UPGRADE.md
morozov: Also, should the removed methods from the description be listed here?

jwage: Doesn't hurt. I will add.

lib/Doctrine/DBAL/Schema/SchemaDiff.php
jwage: Fixed with https://github.com/doctrine/dbal/pull/3565/commits/d382e342d37227b734b0b18183a3d3fcd8923ebc

morozov (approved):
Awesome :+1: Size-wise, this is about 25% of the remaining type-related changes. In one shot!

jwage:
Platforms has a bullseye on it now!

PR #3568, opened 2019-05-30
Enforced parameter and return value types in Connection classes

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

**Notes:**
1. Alongside with the removed `Connection` methods, I'd like to be able to remove `::getParams()` as well in order to make connection parameters internal to the object and be able to modify them without introducing further BC breaks.
2. I didn't document the new `DriverException` thrown by `OCI8Connection::lastInsertId()` because this method needs to be split apart: one for sequences, another for identity columns.


tests/Doctrine/Tests/DBAL/Sharding/PoolingShardConnectionTest.php
```diff
@@ -243,80 +243,4 @@ public function testGetParamsOverride()
             'host' => 'foo',
         ], $conn->getParams());
     }
-
-    public function testGetHostOverride()
```
morozov: The functionality of switching shards is still covered by the test above.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -199,99 +199,49 @@ public function __construct(
      *
      * @return mixed[]
```
jwage: I probably would have changed this to `array<string, mixed>`. Whether that is right or not is another question :)

morozov: It definitely would be a more proper annotation.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -501,58 +443,58 @@ public function setFetchMode($fetchMode) : void
      * Prepares and executes an SQL query and returns the first row of the result
      * as an associative array.
      *
-     * @param string         $statement The SQL query.
-     * @param mixed[]        $params    The query parameters.
-     * @param int[]|string[] $types     The query parameter types.
+     * @param string                                           $query  The SQL query.
+     * @param array<int, mixed>|array<string, mixed>           $params The prepared statement params.
+     * @param array<int, int|string>|array<string, int|string> $types  The query parameter types.
```
jwage: Could this be written as `array<int|string, int|string>`?

jwage (approved):

lib/Doctrine/DBAL/Connection.php
morozov: Not really. It's either an array with integer keys corresponding to parameter positions or an array with string keys corresponding to their names. Not a mix.

PR #3570, opened 2019-05-30
Add proper types to Doctrine\DBAL\Tools namespace.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Add proper types to `Doctrine\DBAL\Tools` namespace.


greg0ire (approved):

morozov (approved):

jwage:
I don't think this one had any BC breaks in it, did it?

morozov:
Formally, it is: `public function setKeywordListClass(string $name, string $class)` can no longer be called with as `setKeywordListClass(1, 2)`, etc. There are no breaking changes worth mentioning in `UPGRADE.md` though.

PR #3571, opened 2019-05-30
Add proper types to the Doctrine\DBAL\Sharding namespace.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Add proper types to the `Doctrine\DBAL\Sharding` namespace.

### DIFF
--- a/lib/Doctrine/DBAL/Sharding/PoolingShardManager.php
+++ b/lib/Doctrine/DBAL/Sharding/PoolingShardManager.php
@@ -18,7 +18,7 @@ class PoolingShardManager implements ShardManager
     /** @var ShardChoser */
     private $choser;
 
-    /** @var string|null */
+    /** @var mixed */
     private $currentDistributionValue;
 
     public function __construct(PoolingShardConnection $conn)
@@ -31,7 +31,7 @@ class PoolingShardManager implements ShardManager
     /**
      * {@inheritDoc}
      */
-    public function selectGlobal()
+    public function selectGlobal() : void
     {
         $this->conn->connect(0);
         $this->currentDistributionValue = null;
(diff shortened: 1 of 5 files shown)


jwage:
Is the shard id always an integer? I see we allow string|int|null but I see it casted to int here https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Sharding/PoolingShardConnection.php#L187

Can we change it to be int everywhere?

morozov:
I believe the `(int)` cast here is wrong. The method didn't allow `string` prior to https://github.com/doctrine/dbal/commit/7bafa27cf7467031f35bfefc8eea8b0805f479bb#diff-9b68bdf5dcd3705b351878bd1ee20492R56 and it caused a type mismatch:
```
$ git diff
@@ -131,7 +131,7 @@ class PoolingShardConnection extends Connection
     /**
      * Connects to a given shard.
      *
-     * @param string|int|null $shardId
+     * @param int|null $shardId

$ phpstan a
 ------ --------------------------------------------------------------------------------------- 
  Line   Doctrine/DBAL/Sharding/PoolingShardManager.php                                         
 ------ --------------------------------------------------------------------------------------- 
  46     Parameter #1 $shardId of method                                                        
         Doctrine\DBAL\Sharding\PoolingShardConnection::connect() expects int|null, int|string  
         given.                                                                                 
 ------ --------------------------------------------------------------------------------------- 
```

The major problem is that I personally don't know which sharding implementations we support and we don't test anything on CI. I'd assume that depending on the implementation, shards IDs may be integers or strings.

Enforcing the ID to `int` only might be the right approach, but it will require some more rework of the calling code.

morozov:
Honestly, I'm not even sure that the sharding support is practically usable. The only implementation of `ShardChoser` is `MultiTenantShardChoser`:
https://github.com/doctrine/dbal/blob/efd0922d36c10876131030725d360ee7e6948c5d/lib/Doctrine/DBAL/Sharding/ShardChoser/MultiTenantShardChoser.php#L18-L21

The implementation means that for every single distribution value, there's a separate shard which does not make any sense. Is there a way to see this sharding in action?

jwage:
To be honest, I am not sure. We might need to create a new issue for 3.0 to look at this and figure out how to properly test it and confirm it is working.

morozov (approved):

PR #3572, opened 2019-05-30
Make LoggerChain use constructor to add loggers instead of adder method

ostrolucky:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

This will help us implement configurable list of loggers to inject into LoggerChain on DI level in DoctrineBundle much more easily (then user can just use `!tagged`). Will help us resolve issues like https://github.com/doctrine/DoctrineBundle/pull/692 and https://github.com/doctrine/DoctrineBundle/issues/225

Deprecation was a result of conversation I had with @morozov on Slack.

No test cases because there are none for LoggerChain and this patch seems simple enough for not needing it.

After merging, I will create new PR for `develop` with removed method.

Before:
```php
$chain->addLogger($logger1);
$chain->addLogger($logger2);
```
After:
```php
$chain = new ChainLogger([$logger1, $logger2]);
```

### DIFF
--- a/lib/Doctrine/DBAL/Logging/LoggerChain.php
+++ b/lib/Doctrine/DBAL/Logging/LoggerChain.php
@@ -11,7 +11,17 @@ class LoggerChain implements SQLLogger
     private $loggers = [];
 
     /**
+     * @param SQLLogger[] $loggers
+     */
+    public function __construct(array $loggers = [])
+    {
+        $this->loggers = $loggers;
+    }
+
+    /**
      * Adds a logger in the chain.
+     *
+     * @deprecated Inject list of loggers via constructor instead
      *
      * @return void
      */
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -42,6 +42,10 @@ The  `Doctrine\DBAL\Driver\SQLSrv\SQLSrvStatement::LAST_INSERT_ID_SQL` constant 
 
 The constants in `Doctrine\DBAL\SQLParserUtils` have been deprecated and will be made private in 3.0.
 
+## Deprecated `LoggerChain::addLogger` method
+
+The `Doctrine\DBAL\Logging\LoggerChain::addLogger` method has been deprecated. Inject list of loggers via constructor instead.
+
 # Upgrade to 2.9
 
 ## Deprecated `Statement::fetchColumn()` with an invalid index


lib/Doctrine/DBAL/Logging/LoggerChain.php
```diff
@@ -7,12 +7,22 @@
...
 class LoggerChain implements SQLLogger
 {
-    /** @var SQLLogger[] */
+    /** @var iterable<SQLLogger> */
     private $loggers = [];
 
+    /**
+     * @param SQLLogger[] $loggers
+     */
+    public function __construct(iterable $loggers = [])
```
morozov: It cannot be any iterable because it has to support `$this->loggers[] = $logger`. Please use `array` in `master`.

lib/Doctrine/DBAL/Logging/LoggerChain.php
```diff
@@ -7,12 +7,22 @@
  */
 class LoggerChain implements SQLLogger
 {
-    /** @var SQLLogger[] */
+    /** @var iterable<SQLLogger> */
     private $loggers = [];
 
+    /**
+     * @param SQLLogger[] $loggers
```
morozov: Please use the same type as on the property. Preferrable, `array<int, SQLLogger>`.

lib/Doctrine/DBAL/Logging/LoggerChain.php
ostrolucky: can't do, `!tagged` does not use `array`. This is a requirement for lazyness

greg0ire: Maybe throw an exception in the adder if `$this->loggers` is not an array then?

ostrolucky: sure if it seems worth it to you, @morozov would you like that?

lib/Doctrine/DBAL/Logging/LoggerChain.php
ostrolucky: we don't care about keys so type shouldn't specify them

lib/Doctrine/DBAL/Logging/LoggerChain.php
ostrolucky: alternatively, `iterator_to_array` can be triggered from within `addLogger`

morozov: What is `!tagged`?

ostrolucky: Symfony DI feature https://symfony.com/blog/new-in-symfony-3-4-simpler-injection-of-tagged-services

morozov: Is this limitation of Symfony DI by design or just a shortcoming? It should be possible to inject an array of proxies which would instantiate the actual objects at the first call.

Why should loggers be lazily injected? They should be fairly lightweight by definition.

I don't like the hackery with `iterator_to_array` either. Does it mean adding a new logger will instantiate all existing one as a side effect?

Can we make it `array` in `2.x` and loosen to `iterable` in `3.0`?

ostrolucky: > It should be possible to inject an array of proxies which would instantiate the actual objects at the first call.

Proxies require classes to be non-final

> Why should loggers be lazily injected? They should be fairly lightweight by definition.

Feature doesn't distinguish it's dbal logger, it's universal for all services, so performance boost applies everywhere. And that's good, it doesn't hurt to apply it for lightweight services too.

> Does it mean adding a new logger will instantiate all existing ones as a side effect?

Yes. But remember, it's same case currently. Benefit of this approach would be that they are instantiated only when you use both methods - constructor and adder, which shouldn't really be done.

> Can we make it array in 2.x and loosen to iterable in 3.0?

What about deprecation path? With regards upgrade path, that doesn't seem correct to me. And I wanted to utilize this in doctrinebundle before dbal 2.0.

morozov: > Proxies require classes to be non-final

DBAL doesn't depend on a specific final class. It depends on an interface which definitely can be proxied.

> Feature doesn't distinguish it's dbal logger, it's universal for all services, so performance boost applies everywhere. And that's good, it doesn't hurt to apply it for lightweight services too.

Proxying has its cost as well which may be even higher than no proxying for lightweight services which loggers usually are. So if we're talking about performance, let's look at the numbers.

> Yes. But remember, it's same case currently. Benefit of this approach would be that they are instantiated only when you use both methods - constructor and adder, which shouldn't really be done.

IMO it's too much complication for the reason I don't currently see. Instead of throwing an exception, one could use `AppendIterator` but I still don't think it's worth it.

> I wanted to utilize this in doctrinebundle before dbal 3.0.

If it's that important, you can add a temporary implementation with the logic you think fits and remove it once 3.0 is out.

ostrolucky: As you wish. I did the change you asked me to do. When is dbal 3.0 release scheduled?

morozov (approved):

morozov:
Thanks @ostrolucky.

PR #3574, opened 2019-05-30
Add proper types to Doctrine\DBAL\Query namespace.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

#### Summary

Add proper types to `Doctrine\DBAL\Query` namespace.

### DIFF
--- a/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
+++ b/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
@@ -41,7 +41,7 @@ class CompositeExpression implements Countable
      * @param string          $type  Instance type of composite expression.
      * @param self[]|string[] $parts Composition of expressions to be joined on composite expression.
      */
-    public function __construct($type, array $parts = [])
+    public function __construct(string $type, array $parts = [])
     {
         $this->type = $type;
 
@@ -51,11 +51,11 @@ class CompositeExpression implements Countable
     /**
      * Adds multiple parts to composite expression.
      *
-     * @param self[]|string[] $parts
+     * @param array<int, self|string> $parts
      *
-     * @return \Doctrine\DBAL\Query\Expression\CompositeExpression
+     * @return $this
      */
-    public function addMultiple(array $parts = [])
+    public function addMultiple(array $parts = []) : self
     {
         foreach ($parts as $part) {
             $this->add($part);
@@ -67,11 +67,11 @@ class CompositeExpression implements Countable
     /**
      * Adds an expression to composite expression.
      *
-     * @param mixed $part
+     * @param self|string $part
      *
-     * @return \Doctrine\DBAL\Query\Expression\CompositeExpression
+     * @return $this
      */
-    public function add($part)
+    public function add($part) : self
     {
         if (empty($part)) {
             return $this;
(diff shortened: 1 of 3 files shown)


lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
```diff
@@ -51,11 +51,11 @@ public function __construct($type, array $parts = [])
     /**
      * Adds multiple parts to composite expression.
      *
-     * @param self[]|string[] $parts
+     * @param array<int, self>|array<int, string> $parts
```
morozov: Should it be `array<int, self|string>`? Unlike parameters in query builder, every expression part is independent and can be either a string or expression.

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
```diff
@@ -69,9 +69,9 @@ public function addMultiple(array $parts = [])
      *
      * @param mixed $part
```
morozov: This is probably `self|string`, the same as the element of `addMultiple($parts)`.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -83,14 +83,14 @@ class QueryBuilder
     /**
      * The query parameters.
      *
-     * @var mixed[]
+     * @var array<string|int, mixed>
```
morozov: I don't think that parameter keys can be mixed. It's more of `array<int, mixed>|array<string, mixed>`.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -83,14 +83,14 @@ class QueryBuilder
...
      */
     private $params = [];
 
     /**
      * The parameter type map of this query.
      *
-     * @var int[]|string[]
+     * @var array<string|int, mixed>
```
morozov: The same as above.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -113,12 +113,12 @@ class QueryBuilder
...
-    private $firstResult = null;
+    private $firstResult = 0;
 
     /**
      * The maximum number of results to retrieve.
      *
-     * @var int
+     * @var int|null
      */
     private $maxResults = null;
```
morozov: Can we get rid of ` = null` since it's redundant?

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -291,12 +285,12 @@ public function setParameter($key, $value, $type = null)
      *         ));
      * </code>
      *
-     * @param mixed[]        $params The query parameters to set.
-     * @param int[]|string[] $types  The query parameters types to set.
+     * @param array<string|int, mixed> $params The query parameters to set.
+     * @param array<string|int, mixed> $types  The query parameters types to set.
```
morozov: The types of `$params` and `$types` keys cannot be mixed (should be `array<…>|array<…>`).

morozov (approved):

PR #3575, opened 2019-05-30
Enforced parameter and return value types in test classes

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

This patch enables the "MissingParameterTypeHint" and "MissingReturnTypeHint" hints in tests. The patch does not change method bodies or data providers leaving the test logic unchanged.


tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -198,7 +198,10 @@ public function testDriverExceptionIsWrapped($method)
         $connection->$method('MUUHAAAAHAAAA');
     }
 
-    public function getQueryMethods()
+    /**
+     * @return array<int, array<int, mixed>>
```
jwage: Should it be `array<int, array<int, string>>`? I'd usually try to get it as close as possible to what the provider is returning but also understand it can be overkill sometimes so I am ok with it being `array<int, array<int, mixed>>` for all providers.

morozov:
This is a backport of the existing work originally done on top of the `develop` branch. I made a mistake of first making a backport and then starting to rework the newly added `T[]` annotations to `array<T>` ones. 

After having done a few, I realized that we don't have PHPStan enabled for tests which means that those modifications would be blind and given the scale of the change inevitably have errors. The proposal is to leave them as is and improve once PHPStan is enabled for tests (#3260).

jwage (commented):
Just a question.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
morozov: The inner type (`array<int, mixed>`) represents a function arguments. Even if we pass them as an array, they are not. They are a tuple where each element has its own independent type. Even if they all happen to be of a certain type at a certain place (e.g. all strings), it doesn't mean they have to be. If tomorrow I add one more argument to the test method and reflect it in the data provider, it shouldn't require changing the types.

jwage (approved):
🥇

PR #3583, opened 2019-06-01
Getting rid of the column name index

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

Right now, column names are contained both in array keys and the actual definitions which creates an ambiguity which is a potential source of issues. The proposal is to not rely on the keys and always use the name from the column definition as the single source of truth.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
@@ -367,17 +367,16 @@ class OraclePlatform extends AbstractPlatform
         $options['indexes'] = [];
         $sql                = parent::_getCreateTableSQL($tableName, $columns, $options);
 
-        foreach ($columns as $name => $column) {
+        foreach ($columns as $column) {
             if (isset($column['sequence'])) {
                 $sql[] = $this->getCreateSequenceSQL($column['sequence']);
             }
 
-            if (! isset($column['autoincrement']) || ! $column['autoincrement'] &&
-               (! isset($column['autoinc']) || ! $column['autoinc'])) {
+            if (empty($column['autoincrement'])) {
                 continue;
             }
 
-            $sql = array_merge($sql, $this->getCreateAutoincrementSql($name, $tableName));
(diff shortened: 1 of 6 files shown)


lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -367,17 +367,16 @@ protected function _getCreateTableSQL(string $tableName, array $columns, array $
...
         $sql                = parent::_getCreateTableSQL($tableName, $columns, $options);
 
-        foreach ($columns as $name => $column) {
+        foreach ($columns as $column) {
             if (isset($column['sequence'])) {
                 $sql[] = $this->getCreateSequenceSQL($column['sequence']);
             }
 
-            if (! isset($column['autoincrement']) || ! $column['autoincrement'] &&
-               (! isset($column['autoinc']) || ! $column['autoinc'])) {
```
morozov: Not sure if it's worth mentioning the removal of `'autoinc'` as a breaking change. I don't see it mentioned in the documentation and this is the only platform which has it. The rest of the usages were removed prior to DBAL 2.0.

jwage:
Should this be done in `Doctrine\DBAL\Schema\Table` too?

morozov:
This is a very good question. Internally, `Table` uses string keys in order to enforce column name uniqueness. But at the same time, they don't need to be exposed outside. Let me see if the same is applicable to the rest of the codebase. For now, I'll put this on hold.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
jwage: I would remove it in another PR and add it to the upgrade doc. That isn't really related to getting rid of the column name index.

jwage:
@morozov We can just merge this and create another issue for improving `Doctrine\DBAL\Schema\Table`. They can be done separate?

morozov:
Yes, let's get this one merged then.

jwage (approved):

PR #3480, opened 2019-03-13
Transaction-related Statement methods return void

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

As was suggested in https://github.com/doctrine/dbal/pull/3348#issuecomment-439734809, I'm beginning to split the pull request into smaller parts starting with breaking changes.

Issue #3175, opened 2018-06-05
Commits silently fail after node quorum rejects changes (Percona cluster + Galera plugin)


tonykiselis:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | ^2.0

#### Summary

Commit from code side passes as if commit was successful when no data was written into the database. Issue reveals itself when Doctrine is being used with Percona multi-master cluster with galera plugin.

Commits silently fail. You end up with "lost data" as rest of the code executes as if everything went fine.

Establish Percona database cluster with galera plugin. Run multiple queries which would use Pessimistic lock to different nodes at the same time. Durining the quorum the requested will be rejected. However, the transanction from code side will pass as if commit was successful despite of it.

To fix: Doctrine\DBAL\Connection::commit() consistently returns false when commit didn't happen. However, this information is tightly wrapped and inaccessible from the Doctrine's API.

Best in my opinion: thow exception on Doctrine\ORM\Connection::commit() when it calls Doctrine\DBAL\Connection::commit() and it returns false. This will uphold the consistency instead of dealing with bool values.

To have exception being thrown after failure to commit in this instance.

morozov:
Looks like all implementors of the `Connection::commit()` violate the signature and return nothing and/or throw an exception in the case of failure. As part of `2.x`, we can return the boolean value consistently across platforms, in `3.x` (#3005) we should start throwing exceptions.

@tonykiselis what about rollbacks? Besides having been called in a wrong state, can it fail and therefore have to be handled by the caller?

tonykiselis:
I found that weird too. Although, it's also kinda weird that in most cases it should throw exception so you can wrap commit in a try/catch, but on deeper level it returns bool? 

Well, there is no rollback. I mean, you can try/catch commit which fails in this case and it will pass as data would have been commited. Although none is present. I don't feel qualified to comment on inner workings of Percona / galera cluster, but I recall that changes are 'discarded' if nodes fail to find consensus during the quorum. 

This happens, as I've mentioned,  when multiple nodes try to write into same record. Nodes communicate and ask other members of the cluster if it's okey for them to commit. If some other node currently has that row locked, transanction doesn't proceed and is discarded. However, from code side in doctrine, commit does not throw exception on this part. However, as I've mentioned DBAL\Connection::commit() returns bool with value false if it failed and true if it succeeded.

morozov:
[...]

PR #3588, opened 2019-06-04
Add commit result bool

otazniksk:
Add commit result bool from connection commit

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1266,6 +1266,8 @@ class Connection implements DriverConnection
             throw ConnectionException::commitFailedRollbackOnly();
         }
 
+        $result = true;
+
         $connection = $this->getWrappedConnection();
 
         $logger = $this->_config->getSQLLogger();
@@ -1275,7 +1277,7 @@ class Connection implements DriverConnection
                 $logger->startQuery('"COMMIT"');
             }
 
-            $connection->commit();
+            $result = $connection->commit();
 
             if ($logger) {
                 $logger->stopQuery();
@@ -1293,12 +1295,12 @@ class Connection implements DriverConnection
         --$this->transactionNestingLevel;
 
         if ($this->autoCommit !== false || $this->transactionNestingLevel !== 0) {
-            return true;
+            return $result;
         }
 
         $this->beginTransaction();
 
-        return true;
+        return $result;
     }
 
     /**
--- a/tests/Doctrine/Tests/DBAL/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/ConnectionTest.php
@@ -298,6 +298,36 @@ class ConnectionTest extends DbalTestCase
     }
 
     /**
+     * @dataProvider resultProvider
+     */
+    public function testCommitReturn(bool $expectedResult) : void
+    {
+        $driverConnection = $this->createMock(DriverConnection::class);
+        $driverConnection->expects($this->once())
+            ->method('commit')->willReturn($expectedResult);
+
+        $driverMock = $this->createMock(Driver::class);
+        $driverMock->expects($this->any())
+            ->method('connect')
+            ->will($this->returnValue($driverConnection));
+
+        $conn = new Connection([], $driverMock);
+
+        $conn->connect();
+        $conn->beginTransaction();
+
+        self::assertSame($expectedResult, $conn->commit());
+    }
+
+    /**
+     * @return bool[][]
+     */
+    public function resultProvider() : array
+    {
+        return [[true], [false]];
+    }
+
+    /**
      * @group DBAL-81
      */
     public function testRollBackStartsTransactionInNoAutoCommitMode() : void
--- a/tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
@@ -88,7 +88,7 @@ class ConnectionTest extends DbalFunctionalTestCase
                 self::assertEquals(2, $this->connection->getTransactionNestingLevel());
                 $this->connection->beginTransaction();
                 self::assertEquals(3, $this->connection->getTransactionNestingLevel());
-                $this->connection->commit();
+                self::assertTrue($this->connection->commit());
                 self::assertEquals(2, $this->connection->getTransactionNestingLevel());
                 throw new Exception();
                 $this->connection->commit(); // never reached
--- a/tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Functional;
+
+use Doctrine\DBAL\Platforms\MySqlPlatform;
+use Doctrine\Tests\DbalFunctionalTestCase;
+use function sleep;
+
+class TransactionTest extends DbalFunctionalTestCase
+{
+    protected function setUp() : void
+    {
+        parent::setUp();
+
+        if ($this->connection->getDatabasePlatform() instanceof MySqlPlatform) {
+            return;
+        }
+
+        $this->markTestSkipped('Restricted to MySQL.');
+    }
+
+    protected function tearDown() : void
+    {
+        $this->resetSharedConn();
+
+        parent::tearDown();
+    }
+
+    public function testCommitFalse() : void
+    {
+        $this->connection->query('SET SESSION wait_timeout=1');
+
+        $this->assertTrue($this->connection->beginTransaction());
+
+        sleep(2); // during the sleep mysql will close the connection
+
+        $this->assertFalse(@$this->connection->commit()); // we will ignore `MySQL server has gone away` warnings
+    }
+}


jwage:
@morozov Is this ok to be done in master or should we target develop (3.0)?

morozov:
I think `master` is fine. However, it must have an integration test to pass. Right now, I believe this area is not well tested and there are no existing expectations of the proper behavior.

morozov:
Please also see a previous discussion on the topic in https://github.com/doctrine/dbal/issues/3175.

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
```diff
@@ -104,7 +104,7 @@ public function testTransactionNestingBehaviorWithSavepoints() : void
             } catch (ConnectionException $e) {
                 self::assertTrue($this->connection->getNestTransactionsWithSavepoints());
             }
-            self::assertTrue($this->connection->commit()); // should not throw exception
+            $this->connection->commit(); // should not throw exception
```
AlessandroMinoccheri: is necessary this comment on this line?

otazniksk:
@morozov @jwage Added tests and return value bool to rollback

morozov:
@otazniksk what is the point in returning a boolean if the boolean can be only `true`?

pulzarraider:
[lib/Doctrine/DBAL/Connection.php](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Connection.php) implements the interface [/lib/Doctrine/DBAL/Driver/Connection.php](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Driver/Connection.php#L69).

For methods `commit()` and `rollback()` it says: `@return bool TRUE on success or FALSE on failure.`

This PR tries to solve the problem, that Doctrine DBAL is not implementing it's own interface correctly and it's hiding the real result of these methods from PDO.

All this is already discussed in #3175. We are experienced the same issues with Percona cluster as it's described in the related issue.

Together with @otazniksk we would like to create also another PR which will throws exception if the `false` occurs in `commit()` or `rollback()`. But that PR will be more controversial because the exception IMHO should be thrown inside PDO if the ERRMODE_EXCEPTION is set. So we will try to fix the php/PDO bug https://bugs.php.net/bug.php?id=66528 inside dbal.

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
pulzarraider: That's the original code. I think there is no reason to drop the comment in this PR.

AlessandroMinoccheri: ok 👍

morozov:
> This PR tries to solve the problem, that Doctrine DBAL is not implementing its own interface correctly and it's hiding the real result of these methods from PDO.

Please show me the code which didn't work before these changes and now will work. Ideally, in the form of a test.

pulzarraider:
@morozov I have [added test for `commit()`](https://github.com/doctrine/dbal/blob/7205a51adb65311996d33ada9eb5154b5654d1dc/tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php) returning `false`. Seems that the `rollback()` method is returning `true` even the real command cannot be send to mysql so I can't add real functional test.

Is it enough? Do you prefer replacing the `false` return value with exception in this PR?

tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php
```diff
@@ -0,0 +1,53 @@
...
+        $this->connection->query('SET SESSION wait_timeout=1');
+
+        $table = new Table('foo');
+        $table->addColumn('id', 'integer');
+        $table->setPrimaryKey(['id']);
+
+        $this->connection->getSchemaManager()->createTable($table);
+
+        self::assertEquals('foo', $table->getName());
+        try {
```
morozov: Please avoid code control structures in tests. Each code path should be tested individually and explicitly. Not conditionally.

tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php
```diff
@@ -0,0 +1,53 @@
...
+            $this->assertTrue($this->connection->beginTransaction());
+            $this->connection->executeUpdate('INSERT INTO foo (id) VALUES(1)');
+
+            sleep(2); // during the sleep mysql will close the connection
+
+            $this->assertFalse(@$this->connection->commit()); // we will ignore `MySQL server has gone away` warnings
```
morozov: This is the test I was asking for!

morozov: Please disable handling warnings explicitly: `PHPUnit\Framework\Error\Warning::$enabled = false`. Then enable back on teardown.

tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php
```diff
@@ -0,0 +1,53 @@
...
+        $table = new Table('foo');
+        $table->addColumn('id', 'integer');
+        $table->setPrimaryKey(['id']);
+
+        $this->connection->getSchemaManager()->createTable($table);
+
+        self::assertEquals('foo', $table->getName());
+        try {
+            $this->assertTrue($this->connection->beginTransaction());
+            $this->connection->executeUpdate('INSERT INTO foo (id) VALUES(1)');
```
morozov: The query and the table seem unnecessary. The transaction will fail to commit even without a query.

tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        $this->connection->query('SET SESSION wait_timeout=1');
+
+        $table = new Table('foo');
+        $table->addColumn('id', 'integer');
+        $table->setPrimaryKey(['id']);
+
+        $this->connection->getSchemaManager()->createTable($table);
+
+        self::assertEquals('foo', $table->getName());
```
morozov: This assertion is irrelevant to the test.

tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php
```diff
@@ -0,0 +1,53 @@
...
+        try {
+            $this->assertTrue($this->connection->beginTransaction());
+            $this->connection->executeUpdate('INSERT INTO foo (id) VALUES(1)');
+
+            sleep(2); // during the sleep mysql will close the connection
+
+            $this->assertFalse(@$this->connection->commit()); // we will ignore `MySQL server has gone away` warnings
+        } catch (Throwable $exception) {
+            $this->fail('Is the PHP/PDO bug https://bugs.php.net/bug.php?id=66528 fixed? Normally no exception is thrown.');
```
morozov: What is the purpose of this instruction? Why do we want our test to fail because of a bug in PHP?

tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php
```diff
@@ -0,0 +1,53 @@
...
+use function in_array;
+use function sleep;
+
+class TransactionTest extends DbalFunctionalTestCase
+{
+    protected function setUp() : void
+    {
+        parent::setUp();
+
+        if (in_array($this->connection->getDatabasePlatform()->getName(), ['mysql'])) {
```
morozov: Please use `instanceof` instead of the name for the better class discoverability. The `in_array()` seems redundant.

morozov (commented):
The new test is exactly what I wanted to see :+1: Some cleanup is still needed.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1254,6 +1254,8 @@ public function beginTransaction()
     /**
      * {@inheritDoc}
      *
+     * @return bool
```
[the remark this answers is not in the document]
pulzarraider: Fixed.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1322,6 +1326,8 @@ private function commitAll()
     /**
      * Cancels any database changes done during the current transaction.
      *
+     * @return bool
```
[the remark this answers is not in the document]
pulzarraider: Fixed.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -297,6 +298,98 @@ public function testCommitStartsTransactionInNoAutoCommitMode() : void
         self::assertTrue($conn->isTransactionActive());
     }
 
+    /**
+     * @group DBAL-81
```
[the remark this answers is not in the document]
pulzarraider: Removed. Copy-paste bug.

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
```diff
@@ -63,7 +63,7 @@ public function testTransactionNestingBehavior() : void
             }
             self::assertTrue($this->connection->isRollbackOnly());
 
-            $this->connection->commit(); // should throw exception
+            self::assertTrue($this->connection->commit()); // should throw exception
```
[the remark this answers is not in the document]
pulzarraider: Assertion removed.

tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php
pulzarraider: Fixed. try/catch was removed.

tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php
pulzarraider: Table and query removed. Thanks.

tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php
pulzarraider: Fixed. Assertion removed.

tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php
pulzarraider: I wanted to identify when will be the bug fixed in PHP/Pdo. Your're  right it's wrong idea. Removed.

tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php
pulzarraider: Fixed.

tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php
pulzarraider: I have used `Warning::$enabled = false` and removed `@`. It [does not throw exceptions, but the warnings mark the tests as failed](https://travis-ci.org/doctrine/dbal/builds/550168914).
Did I miss something?

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -297,6 +298,98 @@ public function testCommitStartsTransactionInNoAutoCommitMode() : void
...
+        $driverMock->expects($this->any())
+            ->method('connect')
+            ->will($this->returnValue(
+                $this->createMock(DriverConnection::class)
+            ));
+        $pdo = $this->createMock(PDO::class);
+        $pdo->expects($this->once())
+            ->method('commit')->willReturn(true);
+
+        $conn = new Connection(['pdo' => $pdo], $driverMock);
```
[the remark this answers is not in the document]
pulzarraider: Fixed.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -297,6 +298,98 @@ public function testCommitStartsTransactionInNoAutoCommitMode() : void
...
+        $conn->connect();
+        $conn->beginTransaction();
+
+        self::assertTrue($conn->commit());
+    }
+
+    /**
+     * @group DBAL-81
+     */
+    public function testCommitReturnFalse() : void
```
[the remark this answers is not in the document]
pulzarraider: Done.

pulzarraider:
@morozov thank you for your review. I fixed all issues [expect one](https://github.com/doctrine/dbal/pull/3588#discussion_r297095722).

tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php
morozov: This is weird. It works for me in my local environment on c08baa3e5095260049f6e39bbe6a75ae25d39cef without any code modifications (on both `mysqli` and `pdo_sqlsrv`).

It seems to mark the test risky because it produces unexpected output as a side effect of reporting the warning. If you can't reproduce this failure locally and fix it appropriately, please feel free to revert to `@`.

morozov:
> […] Seems that the `rollback()` method is returning `true` even the real command cannot be send to mysql so I can't add real functional test for this method. […]

Then we need to revert this part. Otherwise, what's the point?

> […] Do you prefer replacing the `false` return value with exception in this PR?

Not in `2.x`. We already throw exceptions in `3.0-dev` (#3480).

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -297,6 +298,58 @@ public function testCommitStartsTransactionInNoAutoCommitMode() : void
...
+    public function testCommitReturn(bool $expectedResult) : void
+    {
+        $pdo = $this->createMock(PDO::class);
+        $pdo->expects($this->once())
+            ->method('commit')->willReturn($expectedResult);
+
+        $driverMock = $this->createMock(Driver::class);
+        $driverMock->expects($this->any())
+            ->method('connect')
+            ->will($this->returnValue($pdo));
```
morozov: Why `PDO`? https://github.com/doctrine/dbal/blob/0152fbf4d72a7889996bfd7dc7ccc4bc678f4a15/lib/Doctrine/DBAL/Driver.php#L22-L24

tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php
pulzarraider: Unfortunately I can't solve this without `@`.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -297,6 +298,98 @@ public function testCommitStartsTransactionInNoAutoCommitMode() : void
...
+        $conn->connect();
+        $conn->beginTransaction();
+
+        self::assertTrue($conn->rollback());
+    }
+
+    /**
+     * @group DBAL-81
+     */
+    public function testRollackReturnFalse() : void
```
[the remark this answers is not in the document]
pulzarraider: Done.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
pulzarraider: Fixed. Thank you.

pulzarraider:
> > […] Seems that the `rollback()` method is returning `true` even the real command cannot be send to mysql so I can't add real functional test for this method. […]
> 
> Then we need to revert this part. Otherwise, what's the point?

@morozov rollback() changes reverted.

Commits squshed. Tests pass. 
Ready to be reviewed and merged if everzthing is ok.

morozov (approved):

morozov:
Thank you @otazniksk and @pulzarraider!

pulzarraider:
Thank you @morozov.

beberlei:
@morozov The commit testing for return false is restricted to MySQL only. As such it doesn't actually verify that this works cross platform and not some other platform now returns false from `Connection::commit()`.

morozov:
@beberlei I must have overseen this or there was another reason for that. I usually require that functional tests are run across all supported platforms/drivers.

Issue #3592, opened 2019-06-06
Inconsistent and unspecified behavior of Connection::getDatabase()


morozov:
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.9.3

The `Connection::getDatabase()` method description looks like following:

https://github.com/doctrine/dbal/blob/682003690879a625d97db3602e61786390992112/lib/Doctrine/DBAL/Connection.php#L232-L233

This is a bit unclear and doesn't always work in the same way.

### 1. Depending on the connection parameters

Consider the example:
```php
$conn = DriverManager::getConnection([
    'driver' => 'mysqli',
    'host' => 'localhost',
    'user' => 'root',
]);

var_dump($conn->getDatabase());
// NULL
$conn->exec('USE doctrine');
var_dump($conn->getDatabase());
// string(8) "doctrine"
$conn->exec('USE mysql');
var_dump($conn->getDatabase());
// string(5) "mysql"
```
The above looks correct since it returns the _currently_ selected database.

Here's another example:
```php
$conn = DriverManager::getConnection([
    'driver' => 'mysqli',
    'host' => 'localhost',
    'user' => 'root',
    'dbname' => 'doctrine',
]);

var_dump($conn->getDatabase());
// string(8) "doctrine"
$conn->exec('USE mysql');
var_dump($conn->getDatabase());
// string(8) "doctrine"
```
Here, it always returns the _originally_ selected database.

This behavior is consistent across drivers. They all try to use the value from the configuration and fall back to the value reported by the connection which causes the inconsistency.

### 2. Depending on the authentication method

Most of the drivers support external authentication where most if not all connection parameters including username and database can be omitted and will be picked up by the driver from another place (e.g. php.ini, OS configuration, etc).

For instance, in order to connect to an Oracle database via `oci8` using external authentication, one needs to specify `/` as the user name. If `Connection::getDatabase()` is called on such a connection, it will return `/`, not the current database. The same example can be made for most of the supported drivers.

morozov:
Fixed by #3606.

PR #3606, opened 2019-06-12
Removed Driver::getDatabase() in favor of Connection::getDatabase()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

Fixes: #3592.

1. The changes from the end user perspective are stated in `UPGRADE.md`.
2. Some unit tests have been removed as no longer relevant.
3. The code is still covered by functional tests, specifically the ones which perform schema introspection.
4. I didn't add tests for changing the database on the fly since it's not currently supported by the abstraction. The main goal is to make DBAL not hinder valid scenarios which it doesn't explicitly support.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
@@ -152,6 +152,19 @@ class SqlitePlatform extends AbstractPlatform
 
     /**
      * {@inheritDoc}
+     *
+     * The SQLite platform doesn't support the concept of a database, therefore, it always returns an empty string
+     * as an indicator of an implicitly selected database.
+     *
+     * @see \Doctrine\DBAL\Connection::getDatabase()
+     */
+    public function getCurrentDatabaseExpression() : string
+    {
+        return "''";
+    }
+
+    /**
+     * {@inheritDoc}
      */
     protected function _getTransactionIsolationLevelSQL(int $level) : string
     {
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,11 @@
 # Upgrade to 3.0
 
+## BC BREAK: Changes in obtaining the currently selected database name
+
+- The `Doctrine\DBAL\Driver::getDatabase()` method has been removed. Please use `Doctrine\DBAL\Connection::getDatabase()` instead.
+- `Doctrine\DBAL\Connection::getDatabase()` will always return the name of the database currently connected to, regardless of the configuration parameters and will initialize a database connection if it's not yet established.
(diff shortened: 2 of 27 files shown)


lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -150,6 +150,19 @@ public function getDateDiffExpression(string $date1, string $date2) : string
         return sprintf("JULIANDAY(%s, 'start of day') - JULIANDAY(%s, 'start of day')", $date1, $date2);
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * The SQLite platform doesn't support the concept of a database, therefore, it always returns an empty string
```
Ocramius: Is there no way to get the path to the DB?

Ocramius (changes requested):

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
morozov: What would be the practical application of this return value? On the platforms which support databases, the current database can be changed after establishing a connection and may be explicitly used for schema introspection. That's why we need to be able to get it out of the connection. In the case of SQLite, the path is an implementation detail which may not even exist (in-memory database).

Ocramius: Well, I don't think `getDatabase()` should be used at all, but if used, it is likely used for display/debug reasons.

Having this return the path (yes, even for in-memory) would allow making that clear, IMO

morozov: I don't think you can return the path from SQL. I agree it's nice to have but as long as the underlying platform doesn't expose the path as a database, we cannot return it. Or does it?

> Well, I don't think `getDatabase()` should be used at all […]

You mean we shouldn't consider the fact that a different database can be selected after establishing a connection? Yes, that would simplify things but IIRC some platforms (e.g. `information_schema` in MySQL) require the schema/database to be explicitly specified for introspection. On the other hand, some platforms ignore the `$database` argument during introspection (e.g. `getListTableColumnsSQL()` in the SQL Server and IBM DB2 platforms).

With that said, I believe it is reasonable to keep `Connection::getDatabase()` around and use it appropriately. At the same time, the SQLite file path doesn't look like proper usage.

Ocramius: Fair enough :+1:

Ocramius (approved):

jwage (approved):
🤟

PR #3607, opened 2019-06-15
Added .phpstorm.meta.php for PHPUnit mocks

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

With this file in the root of the project, PhpStorm will be able to correctly determine the type of mock objects, therefore resulting in fewer false positive type errors and better code completion.

Screenshot from 2019-06-14 12-05-34

### DIFF
--- a/.phpstorm.meta.php
+++ b/.phpstorm.meta.php
@@ -1,0 +1,31 @@
+<?php
+
+/**
+ * @see https://confluence.jetbrains.com/display/PhpStorm/PhpStorm+Advanced+Metadata
+ */
+namespace PHPSTORM_META
+{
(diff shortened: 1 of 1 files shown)


Ocramius:
This is already part of PHPUnit 8.2.0 as per https://github.com/sebastianbergmann/phpunit/pull/3708

Ocramius (changes requested):
This is already part of PHPUnit 8.2.0 as per sebastianbergmann/phpunit#3708

morozov:
Cool. I may have forgotten to update my local dependencies since it only worked with the added file.

PR #3609, opened 2019-06-15
Reworked the mocks generated by Prophecy using PHPUnit

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | build [#545999926](https://travis-ci.org/doctrine/dbal/jobs/545999926#L609)

As of PHP 7.4.0alpha1 `ReflectionType::__toString()` seems to have started triggering deprecation notices however it's deprecated since PHP 7.1. We already reworked the corresponding mocks in #3582 (`develop`), so this is a backport of the relevant changes.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Types/DateImmutableTypeTest.php
+++ b/tests/Doctrine/Tests/DBAL/Types/DateImmutableTypeTest.php
@@ -9,13 +9,13 @@ use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\ConversionException;
 use Doctrine\DBAL\Types\DateImmutableType;
 use Doctrine\DBAL\Types\Type;
+use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
-use Prophecy\Prophecy\ObjectProphecy;
 use function get_class;
 
 class DateImmutableTypeTest extends TestCase
 {
-    /** @var AbstractPlatform|ObjectProphecy */
+    /** @var AbstractPlatform|MockObject */
     private $platform;
 
     /** @var DateImmutableType */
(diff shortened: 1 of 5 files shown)


tests/Doctrine/Tests/DBAL/Types/DateImmutableTypeTest.php
```diff
@@ -9,13 +9,13 @@
...
 use Doctrine\DBAL\Types\Type;
+use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
-use Prophecy\Prophecy\ObjectProphecy;
 use function get_class;
 
 class DateImmutableTypeTest extends TestCase
 {
-    /** @var AbstractPlatform|ObjectProphecy */
+    /** @var AbstractPlatform|MockObject */
```
Ocramius: Not important, but we should start using `&MockObject`

tests/Doctrine/Tests/DBAL/Types/DateImmutableTypeTest.php
```diff
@@ -24,7 +24,7 @@ class DateImmutableTypeTest extends TestCase
     protected function setUp() : void
     {
         $this->type     = Type::getType('date_immutable');
-        $this->platform = $this->prophesize(AbstractPlatform::class);
+        $this->platform = $this->createMock(AbstractPlatform::class);
```
Ocramius: Thank you! Much better off using regular PHPUnit mocks :+1:

Ocramius (approved):

tests/Doctrine/Tests/DBAL/Types/DateImmutableTypeTest.php
morozov: Is it the time for PHPStan to drop the conversion of `|` to `&` in tests? At least if they had made it possible to disable this behavior, we could disable it and make sure all our types are declared correctly.

Ocramius: Hmm, not sure how the PHPStan PHPUnit plugin works here: `&MockObject` is the correct type here, but let's leave it for another patch/discussion.

PR #3614, opened 2019-06-26
Typo: behavior -> behaviour

noplanman:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | little typo

#### Summary

Typo fixed: `behavior` -> `behaviour`


lcobucci (commented):
I think we're using the American spelling for our docs, right @morozov?

morozov:
I don't think we should be accepting such changes w/o having a publicly available standard and we don't have one AFAIK. Personally, I'd be in favor (without the _u_) of the American spelling.

noplanman:
Ok, no problem. Was just adapting to what was being used in other places already.

SenseException:
Those are a lot of commits for fixing a typo. Is there a base branch mismatch?

PR #3615, opened 2019-06-27
Allow getCustomSchemaOption 'using' to be used

### LINKED ISSUE #2376: Explicitly cast old type's value from new type's value when altering a column's type with PostgreSQL
When you change a column's type with PostgreSQL, for example from boolean to integer, using the annotation and then run a schema update, you end up with the following error:

> ERROR: column "type" cannot be cast automatically to type integer
> SQL state: 42804
> Hint: You might need to specify "USING type::integer".

I guess the queries generated in `PostgreSqlPlatform::getAlterTableSQL()` have to cast the value with the `USING` statement?

I'm willing to do a PR if it's the right way to proceed but I need some guidelines.

### LINKED ISSUE #28963: Error when converting boolean column to decimal in PostgreSQL
- Laravel Version: 5.8.26
- PHP Version: 7.3.6
- Database Driver & Version: PostgreSQL 11.4

### Description:
Converting boolean to decimal/numeric in migrations results in:

```
Datatype mismatch: 7 ERROR:  column "${column_name}" cannot be cast automatically to type numeric
```

### Steps To Reproduce:
1. Create migration with a boolean column. Example:

```php
$table->boolean('special_price')->nullable();
```
2. Create another migration that will alter the column to make it decimal. Example:

```php
$table->decimal('special_price', 12, 4)->nullable()->change();
```
3. Run `php artisan migrate` and it will show an error like above. More complete error message:
```
   Illuminate\Database\QueryException  : SQLSTATE[42804]: Datatype mismatch: 7 ERROR:  column "special_price" cannot be cast automatically to type numeric
HINT:  You might need to specify "USING special_price::numeric(12,4)". (SQL: ALTER TABLE product_flat ALTER special_price TYPE NUMERIC(12, 4))
```

Thanks for your help.

shemgp:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | Might fix #2376 

#### Summary

Allow `customSchemaOption` 'using' to be used as an additional SQL to an `ALTER COLUMN`  type change.  This is to allow a way of adding custom SQL that can be used to convert from one type to another.  Mainly used to fix: https://github.com/laravel/framework/issues/28963 with `syntax ->change()->using('special_price::int::numeric(12,4)')`.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -548,7 +548,7 @@ SQL
                 // here was a server version check before, but DBAL API does not support this anymore.
                 $query = 'ALTER ' . $oldColumnName . ' TYPE ' . $type->getSQLDeclaration($columnDefinition, $this);
                 if ($columnDiff->hasChanged('using')) {
-                    $query = $query . ' USING ' . $column->getCustomSchemaOption('using');
+                    $query .= ' USING ' . $column->getCustomSchemaOption('using');
                 }
                 $sql[] = 'ALTER TABLE ' . $diff->getName($this)->getQuotedName($this) . ' ' . $query;
             }
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -546,7 +546,11 @@ SQL
                 $columnDefinition['autoincrement'] = false;
 
                 // here was a server version check before, but DBAL API does not support this anymore.
-                $query = 'ALTER ' . $oldColumnName . ' TYPE ' . $type->getSQLDeclaration($columnDefinition, $this);
+                $customUsing = '';
+                if ($columnDiff->hasChanged('using')) {
+                    $customUsing = ' USING '.$column->getCustomSchemaOption('using');
+                }
+                $query = 'ALTER ' . $oldColumnName . ' TYPE ' . $type->getSQLDeclaration($columnDefinition, $this) . $customUsing;
                 $sql[] = 'ALTER TABLE ' . $diff->getName($this)->getQuotedName($this) . ' ' . $query;
             }
 
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -548,7 +548,7 @@ SQL
                 // here was a server version check before, but DBAL API does not support this anymore.
                 $customUsing = '';
                 if ($columnDiff->hasChanged('using')) {
-                    $customUsing = ' USING '.$column->getCustomSchemaOption('using');
+                    $customUsing = ' USING ' . $column->getCustomSchemaOption('using');
                 }
                 $query = 'ALTER ' . $oldColumnName . ' TYPE ' . $type->getSQLDeclaration($columnDefinition, $this) . $customUsing;
                 $sql[] = 'ALTER TABLE ' . $diff->getName($this)->getQuotedName($this) . ' ' . $query;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -546,7 +546,11 @@ public function getAlterTableSQL(TableDiff $diff)
                 $columnDefinition['autoincrement'] = false;
 
                 // here was a server version check before, but DBAL API does not support this anymore.
-                $query = 'ALTER ' . $oldColumnName . ' TYPE ' . $type->getSQLDeclaration($columnDefinition, $this);
+                $customUsing = '';
+                if ($columnDiff->hasChanged('using')) {
```
SenseException:
``` php
$query = 'ALTER ' . $oldColumnName . ' TYPE ' . $type->getSQLDeclaration($columnDefinition, $this);
if ($columnDiff->hasChanged('using')) {
    $query = $query . ' USING ' . $column->getCustomSchemaOption('using');
}
```

This way, `$customUsing` shouldn't be needed.

SenseException (changes requested):
Thank you for your contribution :tada:. Please add tests to cover the functionality that you have introduced.

@doctrine/doctrinecore I also would like to have more reviewers on this one. Should `USING` be a PostgreSQL only, or could this be usable somehow for other supported databases too? Should `USING` be introduced this way?

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
morozov: The way it reads right now is weird: the `using` property of the column has changed. I'd rather introduce a new pair of methods for setting and retrieving this expression of a very specific type `string` than put it into an `array<string, mixed>`.

We also need a usage example. Not sure how it's supposed to be used outside of DBAL.

shemgp:
Added test and made requested changes.

For a use case, I need it for fixing: https://github.com/laravel/framework/issues/28963
eg. in Laravel language:
```php
$table->decimal('special_price', 12, 4)->nullable()->change()->using('special_price::int::numeric(12,4)');
```

This is not needed in MySQL as it can convert directly (or ignores it), but is needed in PostgreSQL, since it doesn't know how to convert a boolean to numeric.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
shemgp: I've done the code changes above.

pvsaintpe (approved):

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
SenseException: @morozov @shemgp Sure, new methods for `using` is also possible. If it makes sense, I would like to see this for other RDBMS too.

pvsaintpe: @morozov, @SenseException,

1. Mysql/MariaDB doesn't have USING construction:
https://mariadb.com/kb/en/library/alter-table/

2. Oracle has method USING:
https://docs.oracle.com/cd/B28359_01/server.111/b28286/statements_3001.htm#i2103924

[ USING 'encrypt_algorithm' ]
[ IDENTIFIED BY password ]
[ [ NO ] SALT ]

But I'have doesn't work with Oracle.

Sorry.


3. SQL Server doesn't have USING method:
https://docs.microsoft.com/en-us/sql/t-sql/statements/alter-table-transact-sql?view=sql-server-2017

I think this is a rather specific thing and in this context is not required.

This Merge Request is also needed for me in solving my Postgres SQL tasks, I have never worked with Oracle and I don’t know the functionality of the USING analogue and, all the more, I cannot write tests.

I think it will be right to give you @todo and implement it for other languages later so that the author of this commit and I can continue to developing now.

rhzs: Hi any update on this? I think this one will game changer to support across multiple databases. Just a thought, it'd be good if we can just implement the specific db migration using `trait` inside the migration class to indicate specific database migration / even a version.  (e.g. trait `use PostgresOnly`  etc.)

Issue #3618, opened 2019-07-05
Travic CI tests are failing because ibmcom/db2express-c docker is gone


shemgp:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | master

#### Summary

Travis-CI tests are failing due to docker ibmcom/db2express-c being removed.

#### Current behaviour

```
sudo docker pull ibmcom/db2express-c:10.5.0.5-3.10.0
Error response from daemon: pull access denied for ibmcom/db2express-c, repository does not exist or may require 'docker login'
The command "bash ./tests/travis/install-db2.sh" failed and exited with 1 during .
```

#### How to reproduce

Just commit and that commit or PR will have a Travis CI failure.

#### Expected behaviour

Tests for db2 should probably use https://hub.docker.com/r/ibmcom/db2 instead.

morozov:
@shemgp thanks for jumping on this. I see that the DB2 build still cannot start. Let me know if get stuck with something. I tried switching to this image a week ago and I couldn't make it work.

shemgp:
I'm stuck actually. Still figuring out why it's crashing with 137.

PR #3620, opened 2019-07-05
Make db2 start

shemgp:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3618

#### Summary

Used https://hub.docker.com/r/ibmcom/db2 instead of `ibmcom/db2express-c` which doesn't exist any more.

### DIFF
--- a/tests/travis/install-db2.sh
+++ b/tests/travis/install-db2.sh
@@ -6,20 +6,21 @@ echo Setting up IBM DB2
 
 sudo docker pull ibmcom/db2
 docker run \
-    -itd \
+    -d \
     --name db2 \
     --privileged=true \
     -p 50000:50000 \
     -e LICENSE=accept \
     -e DB2INST1_PASSWORD=Doctrine2018 \
     -e DBNAME=doctrine \
-    ibmcom/db2
+    ibmcom/db2 \
+    "sh -c 'tail -f /dev/null'"
 
-while ! sudo docker exec db2 su - db2inst1 -c 'which db2 2> /dev/null'; do sleep 1; done
+while ! sudo docker exec db2 su - db2inst1 -c 'which db2 2> /dev/null'; do sleep 7; done
 
-while ! sudo docker exec db2 su - db2inst1 -c 'db2start'; do sleep 1; done
+while ! docker logs db2 | grep "All databases are now active."; do sleep 7; done
 
 sudo docker exec db2 su - db2inst1 -c \
-    'db2 CREATE DB doctrine && db2 CONNECT TO doctrine && db2 CREATE USER TEMPORARY TABLESPACE doctrine_tbsp PAGESIZE 4 K'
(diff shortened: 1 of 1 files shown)


tests/travis/install-db2.sh
```diff
@@ -4,17 +4,20 @@ set -ex
...
+    -e DBNAME=doctrine \
+    ibmcom/db2
+
+while ! sudo docker exec db2 su - db2inst1 -c 'which db2 2> /dev/null'; do sleep 1; done
 
-sleep 15
+while ! sudo docker exec db2 su - db2inst1 -c 'db2start'; do sleep 1; done
 
 sudo docker exec db2 su - db2inst1 -c \
     'db2 CREATE DB doctrine && db2 CONNECT TO doctrine && db2 CREATE USER TEMPORARY TABLESPACE doctrine_tbsp PAGESIZE 4 K'
```
morozov: `db2 CREATE DB doctrine` is no longer needed since the DB is created via `-e DBNAME`. Additionally, according to the [documentation](https://hub.docker.com/r/ibmcom/db2#build-your-own),
> Any script copied into the `/var/custom` will be automatically executed after main Db2 setup has completed.

We can try putting this in a file instead of keeping here.

tests/travis/install-db2.sh
```diff
@@ -4,17 +4,20 @@ set -ex
...
-    ibmcom/db2express-c:10.5.0.5-3.10.0 \
-    db2start
+    -e DB2INST1_PASSWORD=Doctrine2018 \
+    -e DBNAME=doctrine \
+    ibmcom/db2
+
+while ! sudo docker exec db2 su - db2inst1 -c 'which db2 2> /dev/null'; do sleep 1; done
 
-sleep 15
+while ! sudo docker exec db2 su - db2inst1 -c 'db2start'; do sleep 1; done
```
morozov: Won't the container start automatically?

> To confirm Db2 container is ready, in the logs we will see the message **Setup has completed**.

Maybe we could try the above?

tests/travis/install-db2.sh
```diff
@@ -4,17 +4,20 @@ set -ex
 
 echo Setting up IBM DB2
 
-sudo docker pull ibmcom/db2express-c:10.5.0.5-3.10.0
-sudo docker run \
-    -d \
+sudo docker pull ibmcom/db2
+docker run \
+    -itd \
```
morozov: Are the `-it` flags really needed?

tests/travis/install-db2.sh
shemgp: Yea, removed it.

tests/travis/install-db2.sh
shemgp: Yes, it will. Removed it and changed it to wait for db2 to start.

tests/travis/install-db2.sh
shemgp: Not necessary at all. Removed.

morozov:
Fixed by #3622.

PR #3622, opened 2019-07-07
Switched from ibmcom/db2express-c to ibmcom/db2

### LINKED ISSUE #3618: Travic CI tests are failing because ibmcom/db2express-c docker is gone

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

Fixes #3618.

### DIFF
--- a/tests/travis/install-db2.sh
+++ b/tests/travis/install-db2.sh
@@ -4,19 +4,20 @@ set -ex
 
 echo Setting up IBM DB2
 
-sudo docker pull ibmcom/db2express-c:10.5.0.5-3.10.0
+echo "su - db2inst1 -c 'db2 CONNECT TO doctrine && db2 CREATE USER TEMPORARY TABLESPACE doctrine_tbsp PAGESIZE 4 K'" > /tmp/doctrine-init.sh
+chmod +x /tmp/doctrine-init.sh
+
 sudo docker run \
     -d \
     -p 50000:50000 \
     -e DB2INST1_PASSWORD=Doctrine2018 \
     -e LICENSE=accept \
+    -e DBNAME=doctrine \
+    -v /tmp/doctrine-init.sh:/var/custom/doctrine-init.sh:ro \
     --name db2 \
-    ibmcom/db2express-c:10.5.0.5-3.10.0 \
-    db2start
+    --privileged=true \
+    ibmcom/db2:11.5.0.0
 
-sleep 15
-
-sudo docker exec db2 su - db2inst1 -c \
-    'db2 CREATE DB doctrine && db2 CONNECT TO doctrine && db2 CREATE USER TEMPORARY TABLESPACE doctrine_tbsp PAGESIZE 4 K'
+sudo docker logs -f db2 | sed '/(*) Setup has completed./ q'
 
 echo DB2 started


tests/travis/install-db2.sh
```diff
@@ -4,19 +4,20 @@ set -ex
...
     -e DB2INST1_PASSWORD=Doctrine2018 \
     -e LICENSE=accept \
+    -e DBNAME=doctrine \
+    -v /tmp/doctrine-init.sh:/var/custom/doctrine-init.sh:ro \
     --name db2 \
-    ibmcom/db2express-c:10.5.0.5-3.10.0 \
-    db2start
-
-sleep 15
+    --privileged=true \
```
lcobucci: Is it really necessary to have this flag enabled? Not a big deal but it's usually recommended to give the minimum permissions to the container.

tests/travis/install-db2.sh
```diff
@@ -4,19 +4,20 @@ set -ex
...
+    ibmcom/db2
```
lcobucci: Can we use a tag here, other than "latest" (sure)? Otherwise we might rely on different images if they change things.

tests/travis/install-db2.sh
```diff
@@ -4,19 +4,20 @@ set -ex
...
 
-sudo docker exec db2 su - db2inst1 -c \
-    'db2 CREATE DB doctrine && db2 CONNECT TO doctrine && db2 CREATE USER TEMPORARY TABLESPACE doctrine_tbsp PAGESIZE 4 K'
+sudo docker logs -f db2 | sed '/(*) Setup has completed./ q'
```
lcobucci: Is it possible to use `dockerize` instead?

tests/travis/install-db2.sh
morozov: Yes, it's necessary.

tests/travis/install-db2.sh
morozov: Yeah, let's use a tag.

tests/travis/install-db2.sh
morozov: I haven't used it before. What would be the benefit and what would the command look like?

morozov: The [documentation](https://hub.docker.com/r/ibmcom/db2#quick-start) says:

> To confirm Db2 container is ready, in the logs we will see the message **Setup has completed**.

While we can ping the container until it's ready, but our experience with MySQL shows that often it's not reliable.

lcobucci (approved):

PR #3626, opened 2019-07-09
Adding support use expression's in default value

pvsaintpe:
I have difficulty generating SQL code in PostgreSQL for ALTER TABLE when using the following cases:

1. SET DEFAULT nextval ('some_table_id_seq' :: regclass)
2. SET DEFAULT '' :: character varying

This Merge Request adds Expression support to implement this feature. I did not add dependencies and my own implementations of the simpler Expression class, so I used the one that is in Doctrine (CompositeExpression) - it basically solves the problem because it has the toString () method and returns the part unchanged - what I need.

I beg you to hold this request in the main branch - (if possible).

|      Q       |   A
|------------- | -----------
| Type         | bug/feature
| BC Break     | no


pvsaintpe:
Now my migrations give incorrect code, in particular, they wrap something that should not be wrapped in single quotes.

For example, in Postgres, the default value may be a procedure, etc.

pvsaintpe:
[Bug with set default value](https://github.com/umbrellio/laravel-pg-extensions/issues/8)

Now there are some problems, since AbstractPlatform generates incorrect SQL for ALTER TABLE, ex.:

```php
Schema::table('some_table', function (Blueprint $table) {
   $table->integer('code')
        ->default(DB::raw("''::character varying"))
       ->change();
});
```

Results:
```SQL
ALTER TABLE some_table ALTER COLUMN code SET DEFAULT '\'\'::character varying'::character varying
```

Expected:

```SQL
ALTER TABLE some_table ALTER COLUMN code SET DEFAULT ''::character varying
```

and etc..

Ocramius:
I'd say this is to be performed with hand-crafted DDL instead

Zlob (dismissed):

pvsaintpe:
> I'd say this is to be performed with hand-crafted DDL instead

I do not agree with you. 

I find it much more convenient to use the builder than raw SQL 

I have already done everything for you added a compatible code that does not break the standard behavior and wrote tests.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2291,6 +2292,10 @@ public function getDefaultValueDeclarationSQL($field)
             return " DEFAULT '" . $default . "'";
         }
 
+        if ($default instanceof CompositeExpression) {
```
morozov: It cannot be a `Doctrine\DBAL\Query\Expression\CompositeExpression`. It could be a new class like `Doctrine\DBAL\Schema\Expression`.

morozov (commented):
> I'd say this is to be performed with hand-crafted DDL instead

We could take this as an opportunity of getting rid of the magic in handling default date values like `CURRENT_TIMESTAMP` and enable the same in a generic way for other platforms. While it kind of contradicts the concept of ORM where the data should be completely controlled by the application, it's fine from the DBAL perspective.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
pvsaintpe: Do you mind if I add this Expression class with such a namespace?  If this is the only problem I will make the necessary changes.

morozov: Adding the new expression class would be the first step. After that, we need to make sure that these new expressions are correctly introspected from a deployed schema and compared by the comparator (i.e. adding new tests or identifying the existing ones which already cover this case).

pvsaintpe: ok, let's start with this!

pvsaintpe:
@morozov I have been added Expression class in namespace (Doctrine\DBAL\Schema\Expressions).
Check it, please.

pvsaintpe:
Why "AppVeyor build failed"?
This is a bug?

morozov:
Looks like something's wrong on AppVeyor. The same build failed after I restarted it. Let's ignore the failure for now.

pvsaintpe:
> Looks like something's wrong on AppVeyor. The same build failed after I restarted it. Let's ignore the failure for now.

I understand that yes, there is a global problem in AppVeor, and not one that is related to my changes.

For my part, I made the necessary edits.

What are our further actions?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -17,6 +17,7 @@
 use Doctrine\DBAL\Schema\Column;
 use Doctrine\DBAL\Schema\ColumnDiff;
 use Doctrine\DBAL\Schema\Constraint;
+use Doctrine\DBAL\Schema\Expressions\Expression;
```
morozov: The `Expressions` piece seems redundant.

lib/Doctrine/DBAL/Schema/Expressions/Expression.php
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * @return mixed
+     */
+    public function getValue()
+    {
+        return $this->value;
+    }
+
+    public function __toString() : string
```
morozov: Let's avoid magic and declare it as `toString()` which will need to be called explicitly.

lib/Doctrine/DBAL/Schema/Expressions/Expression.php
```diff
@@ -0,0 +1,30 @@
...
+     */
+    public function __construct($value)
+    {
+        $this->value = $value;
+    }
+
+    /**
+     * @return mixed
+     */
+    public function getValue()
```
morozov: Not needed.

lib/Doctrine/DBAL/Schema/Expressions/Expression.php
```diff
@@ -0,0 +1,30 @@
...
+
+class Expression
+{
+    /** @var mixed $value */
+    protected $value;
+
+    /**
+     * @param mixed $value
+     */
+    public function __construct($value)
```
morozov: Why would we want to accept `mixed` and cast it to a `string`? We should accept `string`.

lib/Doctrine/DBAL/Schema/Expressions/Expression.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace Doctrine\DBAL\Schema\Expressions;
+
+class Expression
+{
+    /** @var mixed $value */
+    protected $value;
```
morozov: `private`

lib/Doctrine/DBAL/Schema/Expressions/Expression.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace Doctrine\DBAL\Schema\Expressions;
+
+class Expression
```
morozov: `final`

morozov (commented):
> What are our further actions?

See https://github.com/doctrine/dbal/pull/3626#discussion_r302050498.

lib/Doctrine/DBAL/Schema/Expressions/Expression.php
pvsaintpe: fixed

lib/Doctrine/DBAL/Schema/Expressions/Expression.php
pvsaintpe: fixed

lib/Doctrine/DBAL/Schema/Expressions/Expression.php
pvsaintpe: fixed

lib/Doctrine/DBAL/Schema/Expressions/Expression.php
pvsaintpe: fixed

lib/Doctrine/DBAL/Schema/Expressions/Expression.php
pvsaintpe: fixed

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
pvsaintpe: fixed

pvsaintpe:
@morozov,

I have completed all your confirmation requests.
When can I expect these changes to appear in the next release?

morozov:
No, you haven't. The next release is not yet planned.

pvsaintpe:
@morozov 

> No, you haven't. The next release is not yet planned.

Yes, but I think that this is a flaw and users suffer from it.  This merge request does not have backwards incompatible changes and does not require any changes from other developers.  What is the technological or political difficulty to make version v2.9.3 with this fix?

morozov:
This is not how we work here. Please, either do what is suggested or close the PR.

pvsaintpe:
@morozov 

> This is not how we work here. Please, either do what is suggested or close the PR.

For all your comments, I made all the necessary changes. What else is required from me to push this MP into the master or at least create a release 2.9.3? thank you in advance

morozov:
@pvsaintpe what exactly are requesting to review? The suggestion in https://github.com/doctrine/dbal/pull/3626#discussion_r302050498 is still not addressed and I already pointed it out in https://github.com/doctrine/dbal/pull/3626#pullrequestreview-261533819.

PR #3630, opened 2019-07-17
Process NOT NULL in gatherConditions

githabas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary
Adds possibility to specify WHERE IS NOT NULL in conditions.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -615,14 +615,14 @@ class Connection implements DriverConnection
         $platform = $this->getDatabasePlatform();
 
         foreach ($identifier as $columnName => $value) {
-            if (!is_array($value)) {
-                if ($value === null) {
-                    $conditions[] = $platform->getIsNullExpression($columnName);
-                    continue;
-                } elseif (strtoupper($value) === 'NOT NULL') {
+            if ($value === null) {
+                if (substr(strtoupper($columnName), -6) === 'IS NOT') {
+                    $columnName = substr($columnName, 0, -7);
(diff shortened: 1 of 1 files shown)


SenseException (changes requested):
Thank your for your contribution. These changes are breaking the tests with the error message 
`strtoupper() expects parameter 1 to be string, object given`. There are also a few coding style fixes needed.

If you have questions about your PR or the build feel free to ask. Can you please describe your use case for your changes?

lib/Doctrine/DBAL/Connection.php
```diff
@@ -615,9 +615,14 @@ private function addIdentifierCondition(
...
 
         foreach ($identifier as $columnName => $value) {
-            if ($value === null) {
-                $conditions[] = $platform->getIsNullExpression($columnName);
-                continue;
+            if (!is_array($value)) {
+                if ($value === null) {
+                    $conditions[] = $platform->getIsNullExpression($columnName);
+                    continue;
+                } elseif (strtoupper($value) === 'NOT NULL') {
```
morozov: How is it supposed to work with a string value "NOT NULL"?

githabas: $this->connection->update('connections', ['timezone' => 'my zone'], ['timezone' => 'NOT NULL']);

morozov: What if I want to check that the value is not equal to the string "NOT NULL"?

githabas: Ok it's bad idea. I think we should rather look at columnName

Issue #3631, opened 2019-07-23
Doctrine DBAL MysqliStatement broken for Float to Double conversion in de_DE locale (or locales that use ',' for floats)


mmucklo:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.9.2

#### Summary

When using the mysqli driver, the MysqliStatement does not correctly parse the type of a float during bind params, setting it as type 's' (string), when it should be type 'd' (double) - is what I think is going on.

This does not seem to happen with the PDO mysql driver.

An error occurs to the user such as the following:
```
  An exception occurred while executing 'INSERT INTO grumpy_puppy (broken) VALUES (?)' with params [0.12345678901]:

  Data truncated for column 'broken' at row 1
```

This was discovered here:
https://github.com/mmucklo/DtcQueueBundle/issues/98 

And for now the following work around seems to work: https://github.com/mmucklo/DtcQueueBundle/commit/33358df208ef6ef9dab1a254c739d538b568111a

I think this is because https://www.php.net/manual/en/mysqli-stmt.bind-param.php specifies that 'd' should be used for type double, but MysqliStatement does not even have a 'd' in the list of possible types here: https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php#L30-L37

So consequently neither does the conversion code ever check for php floats and specify that they are of type 'd', however I have not tested this hypothesis yet. It took me about a week part time just to reproduce the problem consistently and package it up so that someone else could run it (see below):

#### Current behaviour

Floats can't be inserted into Mysql double columns using the Mysqli driver when in a locale that uses a ',' instead of a '.'

This happens because the float is turned into a string and the SQL standard necessitates floats instead to use '.' regardless of locale.

#### How to reproduce

First, install docker and docker-compose for your platform.

```
git clone https://github.com/mmucklo/docker-symfony.git
cd docker-symfony
git checkout doctrine-bug
docker-compose build
docker-compose up
```

Wait for mysql, then PHP to start up.
This will take up to 30-45 seconds (as it sleeps for mysql warmup).
You should then see a Migration that goes by in php:

```
php_1    |   ++ migrating 20190719053443
php_1    |
[...]

PR #3639, opened 2019-07-31
Issue #3631 - possible fix for Mysqli

mmucklo:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3631 

#### Summary

Create a new DOUBLE type to satisfy the 'd' type necessary for full Mysqli compatibility as referenced here: [mysqli_stmt::bind_param](https://www.php.net/manual/en/mysqli-stmt.bind-param.php).

### DIFF
--- a/lib/Doctrine/DBAL/ParameterType.php
+++ b/lib/Doctrine/DBAL/ParameterType.php
@@ -50,6 +50,11 @@ final class ParameterType
     public const BINARY = 16;
 
     /**
+     * Represents a double data type.
+     */
+    public const DOUBLE = 32;
+
+    /**
      * This class cannot be instantiated.
      */
     private function __construct()
--- a/lib/Doctrine/DBAL/ParameterType.php
+++ b/lib/Doctrine/DBAL/ParameterType.php
@@ -52,7 +52,7 @@ final class ParameterType
     /**
      * Represents a double data type.
      */
-    public const DOUBLE = 32;
+    public const DOUBLE = 17;
 
     /**
      * This class cannot be instantiated.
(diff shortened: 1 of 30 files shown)


mmucklo:
Mysql8.0 build failing due to authentication issues, all other tests pass.

Ocramius:
> for full Mysqli compatibility

Would most likely need an integration test. Something that runs `SELECT 1 WHERE 1 > ?` would suffice, for example.

lib/Doctrine/DBAL/ParameterType.php
```diff
@@ -49,6 +49,11 @@ final class ParameterType
      */
     public const BINARY = 16;
 
+    /**
+     * Represents a double data type.
+     */
+    public const DOUBLE = 32;
```
morozov: It doesn't have to be a power of two since parameter types do aren't used as bits in a bitmask. Please use `17`.

lib/Doctrine/DBAL/Types/FloatType.php
```diff
@@ -29,4 +30,12 @@ public function convertToPHPValue($value, AbstractPlatform $platform)
     {
         return $value === null ? null : (float) $value;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getBindingType()
+    {
+        return ParameterType::INTEGER;
```
morozov: Why do we want to bind floats as integers?

morozov:
> Mysql8.0 build failing due to authentication issues, all other tests pass.

Please rebase on top of the latest `master`. We fixed a few related issues yesterday.

morozov:
> Would most likely need an integration test. Something that runs SELECT 1 WHERE 1 > ? would suffice, for example.

Please make sure you reproduce the scenario described in #3631 which warrants the code change. Just using a float in a query already works and doesn't require changes.

PR #3642, opened 2019-08-03
Fixed test failures on PHP 7.4

morozov:
The failures are caused by the changes in PHP according to [PHP RFC: E_WARNING for invalid container read array-access](https://wiki.php.net/rfc/notice-for-non-valid-array-container).

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | [build #567162538](https://travis-ci.org/doctrine/dbal/builds/567162538)

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Table.php
+++ b/lib/Doctrine/DBAL/Schema/Table.php
@@ -34,7 +34,9 @@ class Table extends AbstractAsset
     protected $_fkConstraints = [];
 
     /** @var mixed[] */
-    protected $_options = [];
+    protected $_options = [
+        'create_options' => [],
+    ];
 
     /** @var SchemaConfig|null */
     protected $_schemaConfig = null;
@@ -69,7 +71,7 @@ class Table extends AbstractAsset
             $this->_addForeignKeyConstraint($constraint);
         }
 
-        $this->_options = $options;
+        $this->_options = array_merge($this->_options, $options);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


tests/Doctrine/Tests/DBAL/Sharding/PoolingShardManagerTest.php
```diff
@@ -47,6 +47,10 @@ public function testSelectGlobal() : void
     {
         $conn = $this->createConnectionMock();
         $conn->expects($this->once())->method('connect')->with($this->equalTo(0));
+        $conn->method('getParams')
+            ->willReturn([
+                'shardChoser' => $this->createMock(ShardChoser::class),
```
morozov: This illustrates one more advantage of using classes for data structures instead of associative arrays. Unlike the array element, a method return value would be automatically mocked.

morozov:
The existing errors seem to be caused by the changes in PHPUnit 8.3 (https://github.com/sebastianbergmann/phpunit/commit/db1dce395c1843f095319a4d9652c2853ea785df) and the fact that `symfony/phpunit-bridge` relies on the old API marked as `@internal`.

/cc @greg0ire

lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -34,7 +34,9 @@ class Table extends AbstractAsset
     protected $_fkConstraints = [];
 
     /** @var mixed[] */
-    protected $_options = [];
+    protected $_options = [
+        'create_options' => [],
```
morozov: As I already said, associative arrays of associative arrays are barely manageable.

morozov:
The rest of the failures are caused by some unknown issue with authentication against MySQL 8 (irrelevant to this PR).

Ocramius (approved):

PR #3654, opened 2019-08-17
fix Appveyor builds

mmucklo:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3657

#### Summary

Fix Appveyor builds which just started crashing a week or two ago due to the disappearance of PHP 7.2 from chocolatey.

Upgrades to PHP 7.3 as it's the only version that's available.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -22,21 +22,21 @@ environment:
     driver: sqlsrv
     db_version: sql2008r2sp2
     coverage: yes
-    php: 7.2
+    php: 7.3
   - db: mssql
     driver: sqlsrv
     db_version: sql2012sp1
-    php: 7.2
+    php: 7.3
     coverage: yes
   - db: mssql
     driver: sqlsrv
     db_version: sql2017
     coverage: no
-    php: 7.2
+    php: 7.3
   - db: mssql
     driver: pdo_sqlsrv
     db_version: sql2017
-    php: 7.2
+    php: 7.3
     coverage: yes
 
 init:


morozov:
@mmucklo could you please point to some changelog which documents the removal of PHP 7.2 from chocolatey?

mmucklo:
@morozov 

Chocolatey is a package repository (akin to packagist.org), I'm not aware that it has a CHANGELOG per se, like a traditional development repository (maybe it does and I don't know where to find it).

Anyway search for yourself with the link below and see if you find PHP 7.2 anymore (also install Chocolatey on a Windows machine and try to install PHP 7.2 if you don't believe me):

[chocolatey.org/search?q=php&prerelease=true&sortOrder=relevance](https://chocolatey.org/search?q=php&prerelease=true&sortOrder=relevance)


As an aside, as we all should probably know from [NPM](https://news.ycombinator.com/item?id=16087024), packages can and do disappear from such places seemingly without notice, although packagist.org is one of the better ones allowing for an abandoned state, but not all repositories follow its example.

morozov:
> […] if you don't believe me […]

It's not about whether I believe you or not. It's about documenting the decisions and referencing them to the changes in the outside world if that's the case.

In this specific case, it's fine just to bump to PHP 7.3 even if it didn't fail due to the rules we agreed upon earlier in https://github.com/doctrine/dbal/pull/3347#issuecomment-439231396. 

> […] as it's the only version that's available.

FWIW, [PHP 7.2.18](https://chocolatey.org/packages/php/7.2.18) is still available on Chocolatey.

morozov (approved):

morozov:
Thank you @mmucklo. Merged.

PR #3660, opened 2019-08-18
WIP - Issue #3631 - fix for Double data type when using MySQLi with languages that do not use '.' for the decimal point

mmucklo:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3631 

#### Summary

   * Add a special ParameterType for Double
   * Detect floats in MysqliStatement and pass the 'd' as the parameter type (see: [mysqli_stmt::bind-param](https://www.php.net/manual/en/mysqli-stmt.bind-param.php#types))
   * Create two tests cases, one for the standard '.' version, and one that extends the same test case, but switches the language to de_DE and tests the ',' case.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
+++ b/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
@@ -34,6 +34,7 @@ class MysqliStatement implements IteratorAggregate, Statement
         ParameterType::NULL         => 's',
         ParameterType::INTEGER      => 'i',
         ParameterType::LARGE_OBJECT => 'b',
+        ParameterType::DOUBLE       => 'd',
     ];
 
     /** @var mysqli */
--- a/lib/Doctrine/DBAL/Types/FloatType.php
+++ b/lib/Doctrine/DBAL/Types/FloatType.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\DBAL\Types;
 
+use Doctrine\DBAL\ParameterType;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 
 class FloatType extends Type
@@ -28,5 +29,13 @@ class FloatType extends Type
     public function convertToPHPValue($value, AbstractPlatform $platform)
     {
         return $value === null ? null : (float) $value;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getBindingType()
+    {
+        return ParameterType::DOUBLE;
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Functional/Types/DoubleTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Types/DoubleTest.php
@@ -1,0 +1,123 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL\Functional\Types;
+
+use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\Tests\DbalFunctionalTestCase;
+use function abs;
+use function floatval;
+use function is_int;
+use function is_string;
+use function microtime;
+use function sprintf;
+
+class DoubleTest extends DbalFunctionalTestCase
+{
+    protected function setUp() : void
+    {
+        parent::setUp();
+
+        $table = new Table('double_table');
+        $table->addColumn('id', 'integer');
+        $table->addColumn('val', 'float');
+        $table->setPrimaryKey(['id']);
+
+        $sm = $this->connection->getSchemaManager();
+        $sm->dropAndCreateTable($table);
+    }
+
+    public function testInsertAndSelect() : void
+    {
+        $value1 = 1.1;
+        $value2 = 77.99999999999;
+        $value3 = microtime(true);
+
+        $this->insert(1, $value1);
+        $this->insert(2, $value2);
+        $this->insert(3, $value3);
+
+        $result1 = $this->select(1);
+        $result2 = $this->select(2);
+        $result3 = $this->select(3);
+
+        if (is_string($result1)) {
+            $result1 = floatval($result1);
+            $result2 = floatval($result2);
+            $result3 = floatval($result3);
+        }
+
+        if ($result1 === false) {
+            $this->fail('Expected $result1 to not be false');
+
+            return;
+        }
+        if ($result2 === false) {
+            $this->fail('Expected $result2 to not be false');
+
+            return;
+        }
+        if ($result3 === false) {
+            $this->fail('Expected $result3 to not be false');
+
+            return;
+        }
+
+        $diff1 = abs($result1 - $value1);
+        $diff2 = abs($result2 - $value2);
+        $diff3 = abs($result3 - $value3);
+
+        $this->assertLessThanOrEqual(0.0001, $diff1, sprintf('%f, %f, %f', $diff1, $result1, $value1));
+        $this->assertLessThanOrEqual(0.0001, $diff2, sprintf('%f, %f, %f', $diff2, $result2, $value2));
+        $this->assertLessThanOrEqual(0.0001, $diff3, sprintf('%f, %f, %f', $diff3, $result3, $value3));
+
+        $result1 = $this->selectDouble($value1);
+        $result2 = $this->selectDouble($value2);
+        $result3 = $this->selectDouble($value3);
+
+        $this->assertSame(is_int($result1) ? 1 : '1', $result1);
+        $this->assertSame(is_int($result2) ? 2 : '2', $result2);
+        $this->assertSame(is_int($result3) ? 3 : '3', $result3);
+    }
+
+    private function insert(int $id, float $value) : void
+    {
+        $result = $this->connection->insert('double_table', [
+            'id'  => $id,
+            'val' => $value,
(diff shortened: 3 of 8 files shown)


lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -272,10 +274,17 @@ private function sendLongData($streams)
     private function bindUntypedValues(array $values)
     {
         $params = [];
-        $types  = str_repeat('s', count($values));
-
+        $types  = '';
         foreach ($values as &$v) {
             $params[] =& $v;
+            // fix for issue #3631 - detect parameter types as they have to be bound differently
```
morozov: Why would we want to autodetect the type of untyped parameters? They should be bound as strings by design.

lib/Doctrine/DBAL/Types/FloatType.php
```diff
@@ -29,4 +30,12 @@ public function convertToPHPValue($value, AbstractPlatform $platform)
     {
         return $value === null ? null : (float) $value;
     }
+
+    /**
+     * {@inheritdoc}
```
morozov: Is this necessary?

tests/Doctrine/Tests/DBAL/Functional/Types/DoubleTest.php
```diff
@@ -0,0 +1,123 @@
...
+        $table->addColumn('val', 'float');
+        $table->setPrimaryKey(['id']);
+
+        $sm = $this->connection->getSchemaManager();
+        $sm->dropAndCreateTable($table);
+    }
+
+    public function testInsertAndSelect() : void
+    {
+        $value1 = 1.1;
```
morozov: If these values are independent, they should be injected using a data provider.

tests/Doctrine/Tests/DBAL/Functional/Types/DoubleTest.php
```diff
@@ -0,0 +1,123 @@
...
+        $value2 = 77.99999999999;
+        $value3 = microtime(true);
```
morozov: How are these three values different? Do they test different code paths?

tests/Doctrine/Tests/DBAL/Functional/Types/DoubleTest.php
```diff
@@ -0,0 +1,123 @@
...
+            $this->fail('Expected $result3 to not be false');
+
+            return;
+        }
+
+        $diff1 = abs($result1 - $value1);
+        $diff2 = abs($result2 - $value2);
+        $diff3 = abs($result3 - $value3);
+
+        $this->assertLessThanOrEqual(0.0001, $diff1, sprintf('%f, %f, %f', $diff1, $result1, $value1));
```
morozov: Where does `0.001` come from?

tests/Doctrine/Tests/DBAL/Functional/Types/DoubleTest.php
```diff
@@ -0,0 +1,123 @@
...
+        $this->assertLessThanOrEqual(0.0001, $diff2, sprintf('%f, %f, %f', $diff2, $result2, $value2));
+        $this->assertLessThanOrEqual(0.0001, $diff3, sprintf('%f, %f, %f', $diff3, $result3, $value3));
+
+        $result1 = $this->selectDouble($value1);
+        $result2 = $this->selectDouble($value2);
+        $result3 = $this->selectDouble($value3);
+
+        $this->assertSame(is_int($result1) ? 1 : '1', $result1);
```
morozov: There shouldn't be any conditionals in a test.

tests/Doctrine/Tests/DBAL/Functional/Types/DoubleTest.php
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * @return mixed
+     */
+    private function select(int $id)
+    {
+        return $this->connection->fetchColumn(
+            'SELECT val FROM double_table WHERE id = ?',
+            [$id],
+            0,
+            [ParameterType::INTEGER]
```
morozov: How is this relevant to the test? We're testing binding double values, not integers.

tests/Doctrine/Tests/DBAL/Functional/Types/LanguageDoubleTest.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL\Functional\Types;
+
+use const LC_ALL;
+use function setlocale;
+
+class LanguageDoubleTest extends DoubleTest
```
morozov: Please avoid extending tests if possible. Specifically in this case, if you want to test the same code paths using different locales, use a data provider.

tests/Doctrine/Tests/DBAL/Functional/Types/LanguageDoubleTest.php
```diff
@@ -0,0 +1,17 @@
...
+{
+    protected function setUp() : void
+    {
+        setlocale(LC_ALL, 'de_DE.UTF-8', 'de_DE', 'de', 'ge');
```
morozov: What if neither of the locales is available at the runtime? Should there be any teardown logic in the test?

morozov (changes requested):

tests/travis/install-language-de.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -ex
+
+echo "Installing language pack de..."
+
+sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 6B05F25D762E3157
```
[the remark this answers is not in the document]
mmucklo: @morozov 

apt update (the next command) was failing before without adding that key, appears to be fixed.

Removed work around.

tests/Doctrine/Tests/DBAL/Functional/Types/DoubleTest.php
```diff
@@ -0,0 +1,123 @@
...
+
+        $this->insert(1, $value1);
+        $this->insert(2, $value2);
+        $this->insert(3, $value3);
+
+        $result1 = $this->select(1);
+        $result2 = $this->select(2);
+        $result3 = $this->select(3);
+
+        if (is_string($result1)) {
```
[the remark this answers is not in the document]
mmucklo: @morozov 

Not all SQL drivers operate the same as I discovered while testing this.

PDO in most if not all cases seems to return floats as strings, while mysqli appears to return them as actual floats.

I discovered this while testing.

As I wanted the test to work on all drivers, I need to convert the ones that are strings into floats so that my comparisons properly work.

This could probably be done differently, I'm open to suggestions.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
mmucklo: @morozov,

Removed this for now (was missing the case for double / d anyway, as I was still testing things out) - this is for a separate "part II" of the bug where another test fails when under de_DE when dealing with floats.

I may add it back later, or do a follow on PR once this is accepted.

lib/Doctrine/DBAL/Types/FloatType.php
mmucklo: @morozov 

Just following convention set in IntegerType - do you want me to remove the inheritdoc?:

[https://github.com/doctrine/dbal/blob/9ff47e738575ab208f3ab4031412c58eb676ed1e/lib/Doctrine/DBAL/Types/IntegerType.php#L38](https://github.com/doctrine/dbal/blob/9ff47e738575ab208f3ab4031412c58eb676ed1e/lib/Doctrine/DBAL/Types/IntegerType.php#L38)

tests/Doctrine/Tests/DBAL/Functional/Types/DoubleTest.php
mmucklo: @morozov 

Similar to my other comment, some drivers seem to return integers as strings in which case I have to either cast everything, or just detect.

I'm open to alternatives?

tests/Doctrine/Tests/DBAL/Functional/Types/DoubleTest.php
mmucklo: @morozov 

As this is a test with Doubles, it's trying to test both insertion and result sets of the doubles to make sure what was inserted matches what's queried.

I can just test insertions / updates if you want things more concise. Although without querying the database, I may not know for sure if my insertion actually worked.

Please advise?

tests/Doctrine/Tests/DBAL/Functional/Types/LanguageDoubleTest.php
mmucklo: @morozov 

Let me look into it.

tests/Doctrine/Tests/DBAL/Functional/Types/LanguageDoubleTest.php
mmucklo: @morozov 

I can make a tear down.

For travis, every run has the de_DE language installed, but I suppose I could try to add detection logic here if it works.

morozov:
Closing as stale.

PR #3662, opened 2019-08-23
Marked connection exception test incomplete on MySQL 8

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | [job #575597509](https://travis-ci.org/doctrine/dbal/jobs/575597509)

It looks like a MySQL bug or miscommunication between PHP and MySQL 8. I can reproduce it as follows:

1. Start a MySQL 8 container:
   ```bash
   docker run \
       --rm \
       -e MYSQL_ALLOW_EMPTY_PASSWORD=yes \
       -p 33306:3306 \
       mysql:8.0 \
       --default-authentication-plugin=mysql_native_password
   ```
2. When the server is ready, run the following script (PHP 7.3.8):
   ```php
   $conn = mysqli_connect('127.0.0.1:33306', 'root', '', 'mysql');
   
   if (!$conn) {
       exit(1);
   }
   
   $query = 'SELECT VERSION()';
   
   $result = mysqli_query($conn, $query);
   
   if (!$result) {
       echo mysqli_error($conn), PHP_EOL;
       exit(1);
   }
   
   echo mysqli_fetch_row($result)[0], PHP_EOL;
   ```
   The expected output is something like `8.0.17`
3. Change the username in the connection parameters to something invalid. Re-run the script. The expected output is:
   >Warning: mysqli_connect(): (HY000/1045): Access denied for user 'anonymous'@'172.17.0.1' (using password: YES)

   The actual output is:
   > Warning: mysqli_connect(): The server requested authentication method unknown to the client [caching_sha2_password]
   > 
   > Warning: mysqli_connect(): (HY000/2054): The server requested authentication method unknown to the client

   At the same time, the server reports its configuration as configured:
   ```
   mysql> show variables like 'default_authentication_plugin';
   +-------------------------------+-----------------------+
   | Variable_name                 | Value                 |
   +-------------------------------+-----------------------+
   | default_authentication_plugin | mysql_native_password |
   +-------------------------------+-----------------------+
   1 row in set (0.01 sec)
   ```

My speculation is that when a connection cannot be established by means of the default `mysql_native_password`, MySQL falls back to `caching_sha2_password` which is not yet supported by PHP. Therefore, instead of error 1045 we're getting 2054.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/ExceptionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/ExceptionTest.php
@@ -3,14 +3,20 @@
 namespace Doctrine\Tests\DBAL\Functional;
 
 use Doctrine\DBAL\Driver\ExceptionConverterDriver;
+use Doctrine\DBAL\Driver\ServerInfoAwareConnection;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Exception;
+use Doctrine\DBAL\Platforms\DrizzlePlatform;
+use Doctrine\DBAL\Platforms\MySqlPlatform;
+use Doctrine\DBAL\Platforms\PostgreSqlPlatform;
+use Doctrine\DBAL\Platforms\SqlitePlatform;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\Tests\DbalFunctionalTestCase;
 use Throwable;
 use const PHP_OS;
 use function array_merge;
+use function assert;
 use function chmod;
 use function exec;
 use function file_exists;
@@ -20,6 +26,7 @@ use function sprintf;
 use function sys_get_temp_dir;
 use function touch;
 use function unlink;
+use function version_compare;
 
 class ExceptionTest extends DbalFunctionalTestCase
 {
@@ -289,7 +296,7 @@ class ExceptionTest extends DbalFunctionalTestCase
 
     public function testConnectionExceptionSqLite() : void
     {
-        if ($this->connection->getDatabasePlatform()->getName() !== 'sqlite') {
+        if ($this->connection instanceof SqlitePlatform) {
             $this->markTestSkipped('Only fails this way on sqlite');
         }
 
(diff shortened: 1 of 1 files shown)


Ocramius (approved):

Ocramius:
> Therefore, instead of error 1045 we're getting 2054.

Seems OK to skip for now, merged :+1:

morozov:
For the reference, here's the wireshark dump of a failed login.:

```
No.     Time           Source                Destination           Protocol Length Info
   1198 66.931375953   172.17.0.1            172.17.0.2            MySQL    165    Login Request user=incognito db=mysql

Frame 1198: 165 bytes on wire (1320 bits), 165 bytes captured (1320 bits) on interface 0
Linux cooked capture
Internet Protocol Version 4, Src: 172.17.0.1, Dst: 172.17.0.2
Transmission Control Protocol, Src Port: 41856, Dst Port: 3306, Seq: 1, Ack: 79, Len: 97
MySQL Protocol
    Packet Length: 93
    Packet Number: 1
    Login Request
        Client Capabilities: 0xa28d
        Extended Client Capabilities: 0x000a
        MAX Packet: 3221225472
        Charset: Unknown (255)
        Username: incognito
        Schema: mysql
        Client Auth Plugin: mysql_native_password
    Payload: 150c5f636c69656e745f6e616d65076d7973716c6e64

No.     Time           Source                Destination           Protocol Length Info
   1202 66.931443447   172.17.0.2            172.17.0.1            MySQL    116    Auth Switch Request

Frame 1202: 116 bytes on wire (928 bits), 116 bytes captured (928 bits) on interface 0
Linux cooked capture
Internet Protocol Version 4, Src: 172.17.0.2, Dst: 172.17.0.1
Transmission Control Protocol, Src Port: 3306, Dst Port: 41856, Seq: 79, Ack: 98, Len: 48
MySQL Protocol
    Packet Length: 44
    Packet Number: 2
    Status: 0xfe
    Auth Method Name: caching_sha2_password
    Auth Method Data: 7c23785c060c0311181751252564064a4b12016600
```

The server indeed may [request to switch](https://dev.mysql.com/doc/internals/en/connection-phase-packets.html#packet-Protocol::AuthSwitchRequest) to a different authentication method without checking/knowing if the client supports it. This method [cannot be disabled](https://dev.mysql.com/doc/refman/5.6/en/sha256-pluggable-authentication.html#sha256-pluggable-authentication-installation) on the server, hence the failure is by design.

PR #3664, opened 2019-08-26
Refactor usage of MockBuilder's deprecated setMethods()

baspeeters:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

- `setMethods()` has been deprecated ([ref](https://github.com/sebastianbergmann/phpunit/blob/b61d4d78a2657f5c65a296ffc86944b473211c88/src/Framework/MockObject/MockBuilder.php#L195)) in `PHPUnit\Framework\MockObject\MockBuilder` (https://github.com/sebastianbergmann/phpunit/commit/d618fa3fda437421264dcfa1413a474f306f79c4)

  - Mocking nonexistent classes requires
    ```
    $this->getMockBuilder('Foo')
    ```
    to be refactored into
    ```
    $this->getMockBuilder($this->getMockClass('Foo'))
    ```
    and then set the mock methods with [`addMethods()`](https://github.com/sebastianbergmann/phpunit/blob/b61d4d78a2657f5c65a296ffc86944b473211c88/src/Framework/MockObject/MockBuilder.php#L259-L266)

  - Mocking existing classes requires existing methods to be set with [`onlyMethods()`](https://github.com/sebastianbergmann/phpunit/blob/b61d4d78a2657f5c65a296ffc86944b473211c88/src/Framework/MockObject/MockBuilder.php#L206-L213)
- In 2 occurrances a class was calling its own dynamic method in a static way. Changed this to use `$this` instead of `self::`

### DIFF
--- a/tests/Doctrine/Tests/DBAL/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/ConnectionTest.php
@@ -67,7 +67,7 @@ class ConnectionTest extends DbalTestCase
         $platform = $this->getMockForAbstractClass(AbstractPlatform::class);
 
         return $this->getMockBuilder(Connection::class)
-            ->setMethods(['executeUpdate'])
+            ->onlyMethods(['executeUpdate'])
             ->setConstructorArgs([['platform' => $platform], $driverMock])
             ->getMock();
     }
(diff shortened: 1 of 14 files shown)


Ocramius (approved):

morozov (approved):

Ocramius:
Thanks @baspeeters!

PR #3668, opened 2019-08-31
Quote collation on MySQL

staudenmeir:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | laravel/framework#29804

#### Summary

`binary` doesn't work as a collation for tables or columns. It's a keyword and needs to be quoted.

This is already happening on PostgreSQL: 

https://github.com/doctrine/dbal/blob/d9b377cda7bb9f97cc03fe77449cfac16f0152c2/lib/Doctrine/DBAL/Platforms/PostgreSQL91Platform.php#L31-L34

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -503,7 +503,7 @@ SQL
             $options['collate'] = $options['charset'] . '_unicode_ci';
         }
 
-        $tableOptions[] = sprintf('COLLATE %s', $options['collate']);
+        $tableOptions[] = $this->getColumnCollationDeclarationSQL($options['collate']);
 
         // Engine
         if (! isset($options['engine'])) {
@@ -963,6 +963,14 @@ SQL
     public function getColumnCharsetDeclarationSQL($charset)
     {
         return 'CHARACTER SET ' . $charset;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getColumnCollationDeclarationSQL($collation)
+    {
+        return 'COLLATE ' . $this->quoteSingleIdentifier($collation);
     }
 
     /**
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
@@ -8,6 +8,7 @@ use Doctrine\DBAL\Platforms\MySqlPlatform;
 use Doctrine\DBAL\Schema\Comparator;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\Types\BlobType;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\DBAL\Types\Types;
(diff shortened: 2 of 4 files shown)


lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -500,7 +500,7 @@ private function buildTableOptions(array $options)
 
         // Collate
         if (! isset($options['collate'])) {
-            $options['collate'] = $options['charset'] . '_unicode_ci';
+            $options['collate'] = $this->quoteSingleIdentifier($options['charset'] . '_unicode_ci');
```
morozov: Please quote when adding to SQL (3 lines below), not here.

morozov (commented):
Please add a functional test which would have failed without the fix.

staudenmeir:
What's the best place for that?

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -503,7 +503,7 @@ private function buildTableOptions(array $options)
             $options['collate'] = $options['charset'] . '_unicode_ci';
         }
 
-        $tableOptions[] = sprintf('COLLATE %s', $options['collate']);
+        $tableOptions[] = sprintf('COLLATE %s', $this->quoteSingleIdentifier($options['collate']));
```
morozov: Can/should we call `$this->getColumnCollationDeclarationSQL()` here to reuse code?

morozov:
Functional tests are located in `tests/Doctrine/Tests/DBAL/Functional`.

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
```diff
@@ -283,6 +285,7 @@ public function testColumnCollation() : void
         self::assertEquals('latin1_swedish_ci', $columns['text']->getPlatformOption('collation'));
         self::assertEquals('latin1_swedish_ci', $columns['foo']->getPlatformOption('collation'));
         self::assertEquals('utf8_general_ci', $columns['bar']->getPlatformOption('collation'));
+        self::assertInstanceOf(BlobType::class, $columns['baz']->getType());
```
morozov: So as far as I understand, if MySQL is told to create a `(VAR)CHAR`/`TEXT` column with the `BINARY` collation, it converts it to `(VAR)BINARY`/`BLOB` which doesn't expose its collation via the `INFORMATION_SCHEMA.COLUMNS`. Therefore, the assertion is made based on the column type.

While the approach seems legit, I'm curious what's the practical use case where you'd need to create a `BINARY` column as `CHAR COLLATION BINARY`?

staudenmeir: Yes. I can't say anything about the practical use case. We fixed/enabled it in Laravel and this is the port to Doctrine (which Laravel uses to alter columns).

You can argue about the use case, but I think quoting the collation is not unreasonable.

morozov: It is reasonable.

morozov (approved):

morozov:
Thank you @staudenmeir.

Issue #3111, opened 2018-04-18
PDO::FETCH_KEY_PAIR Missing from FetchMode Constants


chrisguitarguy:
There's no constant in `Doctrine\DBAL\FetchMode` for `PDO::FETCH_KEY_PAIR`.

Was this intentional? Could it be added? Happy to send a PR for this, but wasn't sure if the omission was intentional.

See also #3082 where FETCH_KEY_PAIR was also mentioned.

morozov:
@chrisguitarguy it is intentional because this mode is only supported in PDO but not other underlying drivers. We may want to implement it across platforms in `3.0` but it requires some refactoring of the driver layer first.

Majkl578:
You can still use the constant in 2.x though and it should work as long as you use PDO-based driver. Note that it'll emit deprecation warning now in 2.x since it's not explicitly supported a fetch mode across drivers.
See #2996 + #2958.

Marking as feature request for 3.0 where it may eventually become a supported FetchMode.

chrisguitarguy:
> it is intentional because this mode is only supported in PDO but not other underlying drivers.

Ah, I gotcha. Didn't think about non-PDO drivers.

For what it's worth, I "fixed" my own deprecation warning by just doing `FetchMode::ASSOCIATIVE` and using `array_column` to pull out what I needed. That may be a solution for other drivers, but DBAL probably doesn't have enough info about the query to do that without parsing SQL I'd guess.

morozov:
@chrisguitarguy you probably can use `FetchMode::NUMERIC` and the first two columns instead without depending on the column names.

chrisguitarguy:
Looked into this a bit more and I think I pretty much understand how it would get implemented. PDO stuff just works, but things like the mysqli, oci, and sql server drivers would need a supported added in their `{DriverName}Statement` classes.

Is that on the right track?

I've only every fixed stuff in DBAL for postgres and mysql, are there docs on how to run the OCI or other tests anywhere?

morozov:
@chrisguitarguy, please hold off on implementing it. While it's technically possible, currently, there's no proper place to put the implementation. If it's implemented on the `Driver\Statement` level, it will have to be duplicated for every non-PDO driver (same as currently `FetchMode::STANDARD_OBJECT` and `FetchMode::CUSTOM_OBJECT` are).

As for testing on OCI, your options are:

[...]

PR #3669, opened 2019-09-03
Change fetchAll typehint

realFlowControl:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | no issue

#### Summary

i hope i didn't get this wrong, but i think the type hint for the `$fetchArgument` argument in `fetchAll` method is wrong, at least in case you are using PDO. The argument `$fetchArgument` may also be a string when `$fetchMode` is set to `\Doctrine\DBAL\FetchMode::CUSTOM_OBJECT`


Ocramius:
I don't think we support `CUSTOM_OBJECT` at all...

flow-control:
In `Doctrine\DBAL\Driver\PDOStatement` it is supported. Nothing special done with it, just passed on to `\PDOStatement`

Ocramius:
It is still *intentionally* not supported: https://github.com/doctrine/dbal/issues/3111#issuecomment-382450385

flow-control:
Thanks for the link, that makes complete sense :+1:

PR #3670, opened 2019-09-05
Add support for multi table truncate when possible (Postgres as example)

goetas:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | - 

#### Summary

Hi,
this PR allows to truncate multiple tables with a single query when supported (postgres as example).

It allowed us to have a test suite running in ~20s instead of 6m. I'm open for suggestions on how to improve it.

This PR is a follow up on https://github.com/doctrine/data-fixtures/pull/314

Is this solution acceptable?

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -3103,6 +3103,16 @@ abstract class AbstractPlatform
     }
 
     /**
+     * Whether the platform supports multi table truncate.
+     *
+     * @return bool
+     */
+    public function supportsTruncateMultiTable()
+    {
+        return false;
+    }
+
+    /**
      * Whether the platform supports partial indexes.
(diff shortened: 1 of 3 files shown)


morozov:
If it's only supported by one platform, why does it need to be part of the abstraction? In https://github.com/doctrine/data-fixtures/pull/314, I see there's branching depending on whether it's PostgreSQL. With this feature implemented, there still will be a branching based on whether the multi-table truncation is supported. No real benefit.

What might help is if we introduced emulation of multi-table truncation for the platforms which don't natively support it. This way, it would indeed simplify the client-side logic.

goetas:
> What might help is if we introduced emulation of multi-table truncation for the platforms which don't natively support it. This way, it would indeed simplify the client-side logic.

This could be done, but for other platforms will return an array of queries (multiple items probably), while for postgres will return only one item.

PR #3677, opened 2019-09-16
Relax statement type declaration

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

`fetch()` is part of the `ResultStatement` interface. Using `Statement` as a
type declaration needlessly constrains people to use this class only
with implementations of Statement.

See https://github.com/doctrine/dbal/blob/b0726e7aab01080fc385599a88491a0ccdbd9417/lib/Doctrine/DBAL/Driver/ResultStatement.php#L65

### DIFF
--- a/lib/Doctrine/DBAL/Driver/StatementIterator.php
+++ b/lib/Doctrine/DBAL/Driver/StatementIterator.php
@@ -6,10 +6,10 @@ use IteratorAggregate;
 
 class StatementIterator implements IteratorAggregate
 {
-    /** @var Statement */
+    /** @var ResultStatement */
     private $statement;
 
-    public function __construct(Statement $statement)
+    public function __construct(ResultStatement $statement)
     {
         $this->statement = $statement;
     }


greg0ire:
Should I also rename the class (in a BC way, like this: https://dev.to/greg0ire/how-to-deprecate-a-type-in-php-48cf)?

greg0ire:
I tried completing the tests with `ArrayStatement`, but they fail because they assert that `fetchAll` is not called when calling `getIterator()` which is not true for `ArrayStatement`.

morozov:
> Should I also rename the class to `ResultStatementIterator` […]?

I don't think it's necessary. The entire hierarchy `Statement extends ResultStatement` looks odd. There's no reason for one to extend the other. There should be a statement and a result. If/once it's done, the iterator can be renamed accordingly.

morozov (approved):

morozov:
Thank you @greg0ire!

PR #3679, opened 2019-09-20
fix begin trasaction after reconnect

kalinin-k-a:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

This is a small fix if we lose the connection and work with transactions after reconnect.
The code examples below demonstrate a problem:

```php
/** Doctrine\DBAL\Connection $co */
$co->executeQuery('CREATE TABLE if not exists test11(test int not null)');
$co->executeQuery('truncate table test11');
$co->executeQuery('insert into test11 values (1)');
$co->beginTransaction();
$co->executeQuery('insert into test11 values (2)');
$v = $co->fetchAll('select * from test11');
$co->close(); // here we lost a connection or closed it for some reason
$co->beginTransaction(); /* p1 */
$co->executeQuery('insert into test11 values (33)');
$co->rollback(); /* p2 */
$v = $co->fetchAll('select * from test11'); // I expect the value 33 will not be selected from the table because it was rollbacked. But the transaction was not started on the line "p1" because of nestingLevel had not beed reset after open new connection.
```

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
@@ -70,6 +70,40 @@ class ConnectionTest extends DbalFunctionalTestCase
             $this->connection->rollBack();
             self::assertEquals(0, $this->connection->getTransactionNestingLevel());
         }
+
+        $this->connection->beginTransaction();
+        $this->connection->close();
+        $this->connection->beginTransaction();
+        self::assertEquals(1, $this->connection->getTransactionNestingLevel());
+    }
+
+    public function testTransactionNestingLevelIsResetOnReconnect() : void
+    {
+        if ($this->connection->getDatabasePlatform()->getName() === 'sqlite') {
+            $params           = $this->connection->getParams();
+            $params['memory'] = false;
+            $params['path']   = '/tmp/test_nesting.sqlite';
+
+            $connection = DriverManager::getConnection(
+                $params,
+                $this->connection->getConfiguration(),
+                $this->connection->getEventManager()
+            );
+        } else {
+            $connection = $this->connection;
+        }
+
+        $connection->executeQuery('CREATE TABLE test_nesting(test int not null)');
+
+        $this->connection->beginTransaction();
+        $this->connection->beginTransaction();
+        $connection->close(); // connection closed in runtime (for example if lost or another application logic)
+
+        $connection->beginTransaction();
+        $connection->executeQuery('insert into test_nesting values (33)');
+        $connection->rollback();
+
+        self::assertEquals(0, $connection->fetchColumn('select count(*) from test_nesting'));
     }
 
     public function testTransactionNestingBehaviorWithSavepoints() : void
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -354,6 +354,8 @@ class Connection implements DriverConnection
         $this->_conn       = $this->_driver->connect($this->params, $user, $password, $driverOptions);
         $this->isConnected = true;
 
+        $this->transactionNestingLevel = 0;
+
         if ($this->autoCommit === false) {
             $this->beginTransaction();
         }


kalinin-k-a:
This is a small fix if we lose the connection and work with transactions after reconnect.
The code examples below demonstrate a problem:
/** Doctrine\DBAL\Connection $co */
$co->executeQuery('CREATE TABLE if not exists test11(test int not null)');
         $co->executeQuery('truncate table test11');
         $co->executeQuery('insert into test11 values (1)');
         $co->beginTransaction();
         $co->executeQuery('insert into test11 values (2)');
         $v = $co->fetchAll('select * from test11');
         $co->close(); // here we lost a connection or closed it for some reason
         $co->beginTransaction(); /* p1 */
         $co->executeQuery('insert into test11 values (33)');
         $co->commit(); /* p2 */
        $v = $co->fetchAll('select * from test11'); // I expect to see here value of 33 in the table, but it was not commited because transaction nesting level had not been reset after reconnect

morozov:
@kalinin-k-a please turn the scenario on the description into a test.

kalinin-k-a:
> 
> 
> @kalinin-k-a please turn the scenario in the description into a test.

done

greg0ire:
Why are there 2 PRs for this? #3679 looks the same, but on another branch and without tests, which one is the right one?

greg0ire:
Please squash your commits

> fix begin trasaction ofter reconnect

Please explain in your commit message what was right, and why your changes fix that situation. Maybe

```
Reset transaction nesting level on connection loss

When the connection is lost, subsequent transactions will no longer be nested because they are executed in a brand new session. Our internal representation of the nesting should take this into account.
```

kalinin-k-a:
> 
> 
> Why are there 2 PRs for this? #3679 looks the same, but on another branch and without tests, which one is the right one?

Another is into v2.5. I'll also write tests

greg0ire (dismissed):

greg0ire:
Please associate your email address with your Github account, or change the email in your commits to an address already associated with it.

kalinin-k-a:
commits are squashed, email is changed. Thank you for reminding !

greg0ire (dismissed):

kalinin-k-a:
pushed again because mysql server had gone away during functional tests previously

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
```diff
@@ -72,6 +72,18 @@ public function testTransactionNestingBehavior() : void
...
+    public function testTransactionNestingLevelIsResetOnReconnect() : void
+    {
+        $this->connection->beginTransaction();
+        $this->connection->beginTransaction();
+        self::assertEquals(2, $this->connection->getTransactionNestingLevel());
+        $this->connection->close(); // connection is lost
+        $this->connection->beginTransaction(); // should connect, reset nesting level and increase it once
+        self::assertEquals(1, $this->connection->getTransactionNestingLevel());
+        $this->connection->commit();
+        self::assertEquals(0, $this->connection->getTransactionNestingLevel());
```
morozov: Please use the scenario from the description. The test as it's written now does not reflect the actual issue.

kalinin-k-a: I’m not sure that scenario from description is good for test. It just shows one case that may be affected by nesting level issue. Also the result from the description may happen if for example commit works wrong.

I fixed nesting level reset, so i tested the same.

Do you insist on change test?

morozov: Without that case, it's not clear why we want to reset the transaction nesting level after establishing a connection. The level is an internal implementation detail of the DBAL, it doesn't have to be tested in a functional way but I'd rather have a functional case which relies on it. Even if it's just one case.

kalinin-k-a: Ok.
Strange, but Insert statement fails the test now with "Table not found". Seems like table exist only before closing connection in testing environment. I'll look again a little later.

kalinin-k-a: Fixed. It was sqlite connection with memory=true. Please review once more

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
```diff
@@ -70,6 +70,40 @@ public function testTransactionNestingBehavior() : void
             $this->connection->rollBack();
             self::assertEquals(0, $this->connection->getTransactionNestingLevel());
         }
+
+        $this->connection->beginTransaction();
+        $this->connection->close();
+        $this->connection->beginTransaction();
+        self::assertEquals(1, $this->connection->getTransactionNestingLevel());
```
morozov: Does this test need to be updated or it's better to put this in a separate test?

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
```diff
@@ -70,6 +70,40 @@ public function testTransactionNestingBehavior() : void
...
+            $connection = $this->connection;
+        }
+
+        $connection->executeQuery('CREATE TABLE test_nesting(test int not null)');
+
+        $this->connection->beginTransaction();
+        $this->connection->beginTransaction();
+        $connection->close(); // connection closed in runtime (for example if lost or another application logic)
+
+        $connection->beginTransaction(); // should connect, reset nesting level and increase it once
```
morozov: The comment doesn't seem relevant to what the test is about.

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
kalinin-k-a: The name of this test is very common  - testTransactionNestingBehavior(), and my fix affects transaction nesting behavior, so I uppdate this)

I can move this lines into a separate test and rename this to something

morozov: Please do.

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
kalinin-k-a: "Should realy start the transaction, and the rollback() call below realy leads to rollback it "

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
```diff
@@ -70,6 +70,40 @@ public function testTransactionNestingBehavior() : void
...
+        $connection->executeQuery('insert into test_nesting values (33)');
+        $connection->rollback();
+
+        self::assertEquals(0, $connection->fetchColumn('select count(*) from test_nesting'));
```
SenseException: Should the transaction nesting level be checked here too?

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
SenseException: The transaction nesting level could be checked here too.

SenseException (dismissed):

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
morozov: > "Should realy start the transaction, and the rollback() call below realy leads to rollback it " - better?

Just remove the comment. The previous one says enough.

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
morozov: Not really. The fact that the transaction was successfully rolled back is enough. The nesting level is an implementation detail in this case.

kalinin-k-a: > 
> 
> Should the transaction nesting level be checked here too?

The exact values of nestingLevel is completely tested in testTransactionNestingBehavior()

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
kalinin-k-a: Ok,. done

morozov (dismissed):

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
morozov: Hm… it's still there.

kalinin-k-a: Hm, it really was. Forgot to commit, sorry. Now it's realy done)

morozov (approved):

morozov:
Thank you @kalinin-k-a.

PR #3680, opened 2019-09-20
fix_begin_transaction_after_reconnect

kalinin-k-a:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary
This is a small fix if we lose the connection and work with transactions after reconnect.
The code examples below demonstrate a problem:
/** Doctrine\DBAL\Connection $co */
This is a small fix if we lose the connection and work with transactions after reconnect.
The code examples below demonstrate a problem:
/** Doctrine\DBAL\Connection $co */
$co->executeQuery('CREATE TABLE if not exists test11(test int not null)');
         $co->executeQuery('truncate table test11');
         $co->executeQuery('insert into test11 values (1)');
         $co->beginTransaction();
         $co->executeQuery('insert into test11 values (2)');
         $v = $co->fetchAll('select * from test11');
         $co->close(); // here we lost a connection or closed it for some reason
         $co->beginTransaction(); /* p1 */
         $co->executeQuery('insert into test11 values (33)');
         $co->rollback(); /* p2 */
        $v = $co->fetchAll('select * from test11'); // I expect the value 33 will not be selected from the table because it was rollbacked. But the transaction was not started on the line "p1" because of nestingLevel had not beed reset after open new connection.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -358,6 +358,8 @@ class Connection implements DriverConnection
             $this->_params['password'] : null;
 
         $this->_conn = $this->_driver->connect($this->_params, $user, $password, $driverOptions);
+        
+        $this->_transactionNestingLevel = 0;
         $this->_isConnected = true;
 
         if (null === $this->platform) {
--- a/tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
@@ -62,6 +62,18 @@ class ConnectionTest extends \Doctrine\Tests\DbalFunctionalTestCase
         }
     }
 
+    public function testTransactionNestingLevelIsResetOnReconnect()
+    {
+        $this->connection->beginTransaction();
+        $this->connection->beginTransaction();
+        self::assertEquals(2, $this->connection->getTransactionNestingLevel());
+        $this->connection->close(); // connection is lost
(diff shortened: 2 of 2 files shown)


SenseException:
I'm not sure if I understand the problem for this solution. What is responsible for closing the connection? It looks like this fix is for a human created problem. Can you please elaborate the use case more?

Also: DBAL 2.5 isn't supported anymore. Please target the master branch for your PR.

morozov:
> What is responsible for closing the connection?

@SenseException this is a good question. As far as I understand, the connection may be lost due to external factors and then _something_ reestablishes it. However, I'm not sure that DBAL can reconnect automatically. E.g. this is why one could do `Connection::ping()` manually.

kalinin-k-a:
> 
> Also: DBAL 2.5 isn't supported anymore. Please target the master branch for your PR.

Ok, I close it. There is another request into master.


> 
> I'm not sure if I understand the problem for this solution. What is responsible for closing the connection? It looks like this fix is for a human created problem. Can you please elaborate the use case more?
> 

More detailed usecase can be described as code below
```
$attempts = 5;
while ($attempts--) {
  try {
    return $connection->fetcAll($query); 
  catch (DBALException $e) {
    if ($connection->errorCode() === 2006) { // gone away
        $connection->close();
    }
  }
}
```
and if nested level was not 0, than Connection object still think that it is still in transaction even after I call close().

SenseException:
I'll close this PR in favor of your new PR #3679. Thank you.

PR #3691, opened 2019-10-08
Fixed empty defaults (In MySqlSchemaManager.php line 223: Notice: Uni…

SerheyDolgushev:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | `null`

#### Summary

Fixes following error which happens when the default is an empty string:
```
In MySqlSchemaManager.php line 223:                                    
  Notice: Uninitialized string offset: 0  
```

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL3691Test.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL3691Test.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Functional\Ticket;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Platforms\MariaDb1027Platform;
+use Doctrine\DBAL\Schema\MySqlSchemaManager;
+use PHPUnit\Framework\TestCase;
+use ReflectionMethod;
+
+/**
+ * @group DBAL-3691
+ */
+class DBAL3691Test extends TestCase
+{
+    public function testIssue()
+    {
+        $conn     = $this->createMock(Connection::class);
+        $platform = new MariaDb1027Platform();
+
+        $schemaManager = new MySqlSchemaManager($conn, $platform);
+
+        $reflectionMethod = new ReflectionMethod($schemaManager, '_getPortableTableColumnDefinition');
+        $reflectionMethod->setAccessible(true);
+
+        $column = $reflectionMethod->invoke($schemaManager, [
+            'field' => 'string_empty_by_default',
+            'type' => 'varchar',
+            'length' => 255,
+            'default' => '',
+            'notnull' => true,
+            'extra' => false,
+            'null' => false,
+        ]);
+
+        $this->assertEquals(null, $column->getDefault());
+    }
+}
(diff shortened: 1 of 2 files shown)


Ocramius (changes requested):
This kind of change requires test case additions

SenseException:
@SerheyDolgushev If you have any questions about the required test cases feel free to ask them :-)

SenseException (changes requested):
@SerheyDolgushev The projects of Doctrine are using automated tests to verify that the existing code is working as well as when they were introduced the first time. Changes can introduce regressions that cause BC breaks or new bugs, that can be covered by those written test cases.

The testing framework we use is [PHPUnit](https://phpunit.readthedocs.io/en/8.4/) and you can find other tests for `MySqlSchemaManager` in the test directory of the project in [`MySqlSchemaManagerTest`](https://github.com/doctrine/dbal/blob/97b99f5302c098fae40962f10479f15e9784369e/tests/Doctrine/Tests/DBAL/Schema/MySqlSchemaManagerTest.php). To be able to prevent future regressions for your new introduced fix, we need you to add a new test that covers the functionality of your code.

From my point of view an empty string shouldn't result in an error. I'm not sure if an empty string has to become a null value, because it is a contradicted behaviour for a user who would expect an empty string. On the other side is a default empty string in a database column a sign, that there should be a value and no `null` in there. If this column can be kept empty, I would expect the application to be nullable to see a difference between "no one put a value in it" and an empty string explicitly set by a user/application who want it to be an empty string.

SerheyDolgushev:
@SenseException thanks for explanations. I just pushed a test for this PR (used PR id as ticket id). Please let me know if any other changes are required.

SenseException (approved):
Thank you for adding the test case.

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL3691Test.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        $conn     = $this->createMock(Connection::class);
+        $platform = new MariaDb1027Platform();
+
+        $schemaManager = new MySqlSchemaManager($conn, $platform);
+
+        $reflectionMethod = new ReflectionMethod($schemaManager, '_getPortableTableColumnDefinition');
+        $reflectionMethod->setAccessible(true);
+
+        $column = $reflectionMethod->invoke($schemaManager, [
```
morozov: This type of a test doesn't really tell why this change is needed. It just proves that internally the code works as expected. Could you please provide a functional test that would cover all supported platforms and potentially find the same issue in others?

morozov (changes requested):

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL3691Test.php
SerheyDolgushev: Sorry, but I guess this issue is outdated and was already fixed. Is it OK to close this PR?

morozov: Sure, feel free to close it.

PR #3692, opened 2019-10-10
Possibility to test the entire database

icetee:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | https://github.com/doctrine/migrations/issues/793

#### Summary

I would like to fix a problem in the doctrine/migrations library and need this commit. Provides the ability to scan the entire database.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
@@ -200,18 +200,41 @@ abstract class AbstractSchemaManager
     }
 
     /**
+    * Returns true if all the given tables exist. This method ignores filters.
+    *
+    * @param string[] $tableNames
+    *
+    * @return bool
+    */
+    public function allTablesExistInSchema(array $tableNames) : bool
+    {
+       $tableNames = array_map('strtolower', $tableNames);
+
+       return count($tableNames) === count(array_intersect($tableNames, array_map('strtolower', $this->listAllTableNames())));
+    }
+
+    /**
      * Returns a list of all tables in the current database.
      *
      * @return string[]
      */
     public function listTableNames()
     {
+        return $this->filterAssetNames($this->listAllTableNames());
+    }
+
+    /**
+     * Returns a list of all tables in the current database.
+     *
+     * @return string[]
+     */
+    final public function listAllTableNames(): iterable
+    {
         $sql = $this->_platform->getListTablesSQL();
 
-        $tables     = $this->_conn->fetchAll($sql);
-        $tableNames = $this->_getPortableTablesList($tables);
+        $tables = $this->_conn->fetchAll($sql);
 
-        return $this->filterAssetNames($tableNames);
+        return $this->_getPortableTablesList($tables);
     }
 
     /**
--- a/tests/Doctrine/Tests/DBAL/Schema/DB2SchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Schema/DB2SchemaManagerTest.php
@@ -10,6 +10,7 @@ use Doctrine\DBAL\Platforms\DB2Platform;
 use Doctrine\DBAL\Schema\DB2SchemaManager;
 use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
+
 use function in_array;
 
 /**
@@ -58,6 +59,64 @@ final class DB2SchemaManagerTest extends TestCase
             ],
             $this->manager->listTableNames()
         );
+    }
+
+    /**
+     * @return void
+     *
+     * @group DBAL-3692
+     */
+    public function testListAllTableNamesFiltersAssetNamesCorrectly()  : void
+    {
+        $this->conn->getConfiguration()->setFilterSchemaAssetsExpression('/^(?!T_)/');
+        $this->conn->expects($this->once())->method('fetchAll')->will($this->returnValue([
+            ['name' => 'FOO'],
+            ['name' => 'T_FOO'],
+            ['name' => 'BAR'],
+            ['name' => 'T_BAR'],
+        ]));
+
+        self::assertSame(
+            [
+                'FOO',
+                'T_FOO',
+                'BAR',
+                'T_BAR',
+            ],
+            $this->manager->listAllTableNames()
+        );
+    }
+
+    /**
+     * @return void
+     *
+     * @group DBAL-3692
+     */
+    public function testTablesExistWithoutFilteredParameter() : void
+    {
+        $this->conn->getConfiguration()->setFilterSchemaAssetsExpression('/^(?!T_)/');
+        $this->conn->expects($this->once())->method('fetchAll')->will($this->returnValue([
+            ['name' => 'FOO'],
+            ['name' => 'T_FOO'],
+        ]));
+
+        self::assertEquals(false, $this->manager->tablesExist('T_FOO'));
+    }
+
+    /**
+     * @return void
+     *
+     * @group DBAL-3692
+     */
+    public function testTablesExistWhithFilteredFalseParameter() : void
+    {
+        $this->conn->getConfiguration()->setFilterSchemaAssetsExpression('/^(?!T_)/');
+        $this->conn->expects($this->once())->method('fetchAll')->will($this->returnValue([
+            ['name' => 'FOO'],
+            ['name' => 'T_FOO'],
+        ]));
+
+        self::assertEquals(true, $this->manager->allTablesExistInSchema(['T_FOO']));
     }
 
     /**


lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -209,6 +212,21 @@ public function listTableNames()
...
+     * Returns a list of all tables in the current database.
+     *
+     * @return string[]
+     */
+    public function listAllTableNames()
+    {
+        $sql = $this->_platform->getListTablesSQL();
+
+        $tables     = $this->_conn->fetchAll($sql);
+        $tableNames = $this->_getPortableTablesList($tables);
```
stof: no need to assign this `$tableNames` variable.

icetee:
@stof Can you look at him? :)

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -183,19 +183,22 @@ public function listTableIndexes($table)
...
      * Returns true if all the given tables exist.
      *
-     * @param string[] $tableNames
+     * @param string|string[] $tableNames
+     * @param bool            $filtered
      *
      * @return bool
      */
-    public function tablesExist($tableNames)
+    public function tablesExist($tableNames, $filtered = true)
```
greg0ire: This looks like a BC-break for extending classes that override this method. I think you should use `func_get_args()` to retrieve that argument when passed.

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -209,6 +212,20 @@ public function listTableNames()
         return $this->filterAssetNames($tableNames);
     }
 
+    /**
+     * Returns a list of all tables in the current database.
+     *
+     * @return string[]
+     */
+    public function listAllTableNames()
```
greg0ire: Let's make this method signature more changeable (`final`) and modern (`iterable`):
```suggestion
    final public function listAllTableNames(): iterable
```

greg0ire:
Can you please write tests for this?

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
icetee: I covered it with tests, the original operation is preserved. I think it's unnecessary to call `func_get_args`. It only increases complexity.

greg0ire: Doesn't make it less of a BC-break. Let me illustrate my point: https://3v4l.org/g4Er6
:boom: warning!

greg0ire (dismissed):

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -183,19 +183,26 @@ public function listTableIndexes($table)
     /**
      * Returns true if all the given tables exist.
      *
-     * @param string[] $tableNames
+     * To avoid BC-break not all parameters are included.
+     *
+     * @param string|string[] $tableNames
+     * @param bool            $filtered   Optional parameter, if false ignores filters.
```
Ocramius: @greg0ire I followed the discussion about this, and the BC break is indeed fixed, but the API is really really horrible.

Isn't it possible to declare a new method, instead of using dynamic args?

Also, patch should probably target `master`, since it's an addition (for `2.10`)

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -183,19 +183,26 @@ public function listTableIndexes($table)
...
      *
      * @return bool
      */
     public function tablesExist($tableNames)
     {
         $tableNames = array_map('strtolower', (array) $tableNames);
 
-        return count($tableNames) === count(array_intersect($tableNames, array_map('strtolower', $this->listTableNames())));
+        $args = func_get_args();
```
Ocramius: As mentioned above, using separate methods for checking existence in current DB or all DBs would be best

tests/Doctrine/Tests/DBAL/Schema/DB2SchemaManagerTest.php
```diff
@@ -62,6 +63,80 @@ public function testListTableNamesFiltersAssetNamesCorrectly()
...
+     */
+    public function testTablesExistWhithFilteredTrueParameter()
+    {
+        $this->conn->getConfiguration()->setFilterSchemaAssetsExpression('/^(?!T_)/');
+        $this->conn->expects($this->once())->method('fetchAll')->will($this->returnValue([
+            ['name' => 'FOO'],
+            ['name' => 'T_FOO'],
+        ]));
+
+        self::assertEquals(false, $this->manager->tablesExist('T_FOO', true));
```
Ocramius: Is any internal DBAL API using the second parameter at this point in time?

tests/Doctrine/Tests/DBAL/Schema/DB2SchemaManagerTest.php
```diff
@@ -62,6 +63,80 @@ public function testListTableNamesFiltersAssetNamesCorrectly()
...
+    }
+
+    /**
+     * @return void
```
Ocramius: Anything `@return void` can be `: void`, since we require PHP 7.1 anyway

Ocramius (changes requested):

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
greg0ire: Uh yeah sure I should have spotted this, even phpmd says booleans in signatures are a smell…

icetee: What would be the appropriate method name? `tablesExistWithAllTable`?

greg0ire: `tablesExistInUnfilteredSchema`?

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -209,6 +216,20 @@ public function listTableNames()
...
+     * @return string[]
+     */
+    final public function listAllTableNames(): iterable
+    {
+        $sql = $this->_platform->getListTablesSQL();
+
+        $tables = $this->_conn->fetchAll($sql);
+
+        return $this->_getPortableTablesList($tables);
+    }
```
greg0ire: I would make `listTableNames` call `listAllTableNames`

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -195,18 +195,41 @@ public function tablesExist($tableNames)
...
 
     /**
-     * Returns a list of all tables in the current database.
+     * Returns true if all the given tables exist. This method ignores filters.
+     *
+     * @param string|string[] $tableNames
+     *
+     * @return bool
+     */
+    public function tablesExistInUnfilteredSchema($tableNames) : bool
```
Ocramius: From the API name, it is unclear if all tables must exist, or only one suffices.

`allTablesExistInSchema` would solve that

Ocramius (changes requested):

Ocramius:
Please mind the target branch (should be `master` ;-) )

icetee:
I had to rebuild the branch in order to get into the "master". So clean, no more other diff commit.

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -199,19 +199,42 @@ public function tablesExist($tableNames)
         return count($tableNames) === count(array_intersect($tableNames, array_map('strtolower', $this->listTableNames())));
     }
 
+    /**
+    * Returns true if all the given tables exist. This method ignores filters.
+    *
+    * @param string[] $tableNames
+    *
+    * @return bool
```
greg0ire: That line is unnecessary

lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```diff
@@ -199,19 +199,42 @@ public function tablesExist($tableNames)
...
+    */
+    public function allTablesExistInSchema(array $tableNames) : bool
+    {
+       $tableNames = array_map('strtolower', $tableNames);
+
+       return count($tableNames) === count(array_intersect($tableNames, array_map('strtolower', $this->listAllTableNames())));
```
greg0ire: Why are we using `strtolower` here?

greg0ire: Why are we using `strtolower here`?

icetee: Primarily because it was an earlier code, so it was in the tablesExist method. The table name can be both uppercase and lowercase.

But I do not see what the consequences would be if we changed this now.

greg0ire: Oh right. Sounds like you could refactor this in a private method, right? `isTableListSubsetOf($leftSet, $rightSet): bool`, something like that…

icetee: You mean that?

```php
     /**
     * Returns true if all the given tables exist.
     *
     * @param string|string[] $tableNames
     *
     * @return bool
     */
    public function tablesExist($tableNames)
    {
        return $this->isTableListSubsetOf(
          array_map('strtolower', (array) $tableNames),
          array_map('strtolower', $this->listTableNames())
        );
    }

    /**
    * Returns true if all the given tables exist. This method ignores filters.
    *
    * @param string[] $tableNames
    *
    * @return bool
    */
    public function allTablesExistInSchema(array $tableNames) : bool
    {
        return $this->isTableListSubsetOf(
          array_map('strtolower', $tableNames),
          array_map('strtolower', $this->listAllTableNames())
        );
    }

    private function isTableListSubsetOf(array $leftSet, array $rightSet) : bool
    {
        return count($leftSet) === count(array_intersect($leftSet, $rightSet));
    }
```

greg0ire: I mean having the `array_map` calls in there too…

morozov:
Closing as stale. Additionally, I don't think that the introduction of a version of a method that ignores the filter is the right approach. This way, we may end up introducing a similar version for more methods and end up having two responsibilities in the one API. The consumer of the API can temporarily disable the filter instead.

PR #3693, opened 2019-10-11
fix MySQL bit type does not support exceptions.

Yi-pixel:
The following error occurred when I was generating phpdocs.
```
Exception: Unknown database type bit requested, Doctrine\DBAL\Platforms\MySqlPlatform may not support it.
Could not analyze class App\Models\UserModel.
```
By checking the source code, I found out that it was a type mapping problem, so I added this line of code.

|      Q       |   A
|------------- | -----------
| Type            | improvement
| BC Break     | yes
| Fixed issues | fix MySQL bit type does not support exceptions.

#### Summary

Type mapping with bit added

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -1059,6 +1059,7 @@ SQL
             'smallint'      => 'smallint',
             'mediumint'     => 'integer',
             'int'           => 'integer',
+            'bit'           => 'integer',
             'integer'       => 'integer',
             'bigint'        => 'bigint',
             'tinytext'      => 'text',


lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -1085,6 +1085,7 @@ protected function initializeDoctrineTypeMappings()
             'binary'        => 'binary',
             'varbinary'     => 'binary',
             'set'           => 'simple_array',
+            'bit'           => 'integer',
```
greg0ire: Looks like entries are grouped by values. Could you move this line to line 1062?

WonlyCms: Ok, I have already done that.

greg0ire:
Please squash your 2 commits together, we don't need to know it was not on the right line from the start, and it's hard to understand what you are fixing with a commit message that just says "fix"

greg0ire (dismissed):

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
WonlyCms: > Please squash your 2 commits together, we don't need to know it was not on the right line from the start, and it's hard to understand what you are fixing with a commit message that just says "fix"

Ok, please check the latest commit

WonlyCms:
Sorry, there is a problem with the branch.

PR #3695, opened 2019-10-11
Add support for DISTINCT statement

bingo-soft:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

The same functionality as in Doctrine/ORM. New method `distinct($flag = true)` in `QueryBuilder` class and new test `testSimpleSelectWithDistinct` in `QueryBuilderTest` are provided.

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -482,7 +482,7 @@ class QueryBuilder
      */
     public function distinct(bool $flag = true)
     {
-        $this->sqlParts['distinct'] = (bool) $flag;
+        $this->sqlParts['distinct'] = $flag;
 
         return $this;
     }
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -480,13 +480,12 @@ class QueryBuilder
      *
      * @return $this This QueryBuilder instance.
      */
-    public function distinct($flag = true)
+    public function distinct(bool $flag = true)
     {
         $this->sqlParts['distinct'] = (bool) $flag;
 
         return $this;
     }
-
 
     /**
      * Adds an item that is to be returned in the query result.
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -476,7 +476,7 @@ class QueryBuilder
      *         ->from('users', 'u');
      * </code>
      *
-     * @param bool $flag
+     * @param bool $flag If true, then DISTINCT is added to the query
      *
      * @return $this This QueryBuilder instance.
      */
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -62,6 +62,7 @@ class QueryBuilder
      * @var array<string, mixed>
      */
     private $sqlParts = [
+        'distinct' => false,
         'select'  => [],
         'from'    => [],
         'join'    => [],
@@ -464,6 +465,28 @@ class QueryBuilder
 
         return $this->add('select', $selects);
     }
+
+    /**
+     * Adds a DISTINCT flag to this query.
+     *
+     * <code>
+     *     $qb = $conn->createQueryBuilder()
+     *         ->select('u.id')
+     *         ->distinct()
+     *         ->from('users', 'u');
+     * </code>
+     *
+     * @param bool $flag
+     *
+     * @return $this This QueryBuilder instance.
+     */
+    public function distinct($flag = true)
+    {
+        $this->sqlParts['distinct'] = (bool) $flag;
+
+        return $this;
+    }
+
 
     /**
      * Adds an item that is to be returned in the query result.
@@ -1088,7 +1111,8 @@ class QueryBuilder
      */
     private function getSQLForSelect() : string
     {
-        $query = 'SELECT ' . implode(', ', $this->sqlParts['select']);
+        $query = 'SELECT ' . ($this->sqlParts['distinct'] === true ? 'DISTINCT ' : '') .
+            implode(', ', $this->sqlParts['select']);
 
         $query .= ($this->sqlParts['from'] ? ' FROM ' . implode(', ', $this->getFromClauses()) : '')
             . ($this->sqlParts['where'] !== null ? ' WHERE ' . ((string) $this->sqlParts['where']) : '')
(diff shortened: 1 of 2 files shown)


greg0ire:
That functionality is already available by using `select('DISTINCT u.id')` correct? Is the point to be able to make it conditional easily? Can you provide a code sample that shows in what situation this will make your life easier?

bingo-soft:
@greg0ire. There is a number of arguments for providing this functionality.

First of all, in SQL `DISTINCT` is a separate operation, similar to `GROUP BY`, but with a specific focus on eliminating duplicates. So, if we have `groupBy` in API, why not to provide a more granular `distinct` method?

Secondly, `DISTINCT` is a row operation, not a column function, so if you write higher level query language on top of Doctrine DBAL, which allows users to specify returned fields on the fly, it would be easier to apply `distinct=true` once, then to inject this logic into the part of the algorithm responsible for handling fields. So, two lines of code in Doctrine DBAL would save numbers of lines of client code (looping through all fields, injecting DISTINCT to the first fields; if client does not provide the list of fields, than adding DISTINCT to `*`). So, what is expected to be rather a row operation, becomes a column operation in DBAL client code.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -465,6 +466,27 @@ public function select($select = null) : self
...
+     *         ->from('users', 'u');
+     * </code>
+     *
+     * @param bool $flag If true, then DISTINCT is added to the query
+     *
+     * @return $this This QueryBuilder instance.
+     */
+    public function distinct(bool $flag = true)
+    {
+        $this->sqlParts['distinct'] = (bool) $flag;
```
greg0ire: I don't think this cast is needed.
```suggestion
        $this->sqlParts['distinct'] = $flag;
```

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -465,6 +466,27 @@ public function select($select = null) : self
...
+     *         ->select('u.id')
+     *         ->distinct()
+     *         ->from('users', 'u');
+     * </code>
+     *
+     * @param bool $flag If true, then DISTINCT is added to the query
+     *
+     * @return $this This QueryBuilder instance.
+     */
+    public function distinct(bool $flag = true)
```
greg0ire:
```suggestion
    final public function distinct(bool $flag = true) : self
```

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1088,7 +1110,8 @@ public function resetQueryPart(string $queryPartName) : self
      */
     private function getSQLForSelect() : string
     {
-        $query = 'SELECT ' . implode(', ', $this->sqlParts['select']);
+        $query = 'SELECT ' . ($this->sqlParts['distinct'] === true ? 'DISTINCT ' : '') .
```
greg0ire: This is already supposed to be a boolean, and if happens not to be, the bug might still go undetected. If you feel unsafe about it being a boolean, then check that and throw if it isn't, or introduce a private accessor to do the type check. If you feel safe, then simply do this:

```suggestion
        $query = 'SELECT ' . ($this->sqlParts['distinct'] ? 'DISTINCT ' : '') .
```

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -62,6 +62,7 @@ class QueryBuilder
      * @var array<string, mixed>
      */
     private $sqlParts = [
+        'distinct' => false,
```
greg0ire: I'd move it one line below, it looks like there this array is ordered with the same order in which you would find these keywords in an SQL query.

greg0ire (changes requested):
> Also, is develop the correct target branch when there are no BC breaks?

Please address this

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -465,6 +466,27 @@ public function select($select = null) : self
...
+     * Adds a DISTINCT flag to this query.
+     *
+     * <code>
+     *     $qb = $conn->createQueryBuilder()
+     *         ->select('u.id')
+     *         ->distinct()
+     *         ->from('users', 'u');
+     * </code>
+     *
+     * @param bool $flag If true, then DISTINCT is added to the query
```
greg0ire: In what scenario is that parameter going to be useful?

lib/Doctrine/DBAL/Query/QueryBuilder.php
bingo-soft: Looks like that. You are right

lib/Doctrine/DBAL/Query/QueryBuilder.php
bingo-soft: Agree with `self` part, but not sure of `final` specificator. It would be the only one `final` method in QueryBuilder.

lib/Doctrine/DBAL/Query/QueryBuilder.php
bingo-soft: Do you mean, that there is no sense to mention this parameter in the docstring, since it is self evident? And by the way, as for `develop` branch, should I make a new merge request to the master branch to fix this?

lib/Doctrine/DBAL/Query/QueryBuilder.php
greg0ire: Yeah but it is one of the most recent ones, right? We can wait for the advice of another maintainer if you want, I will not merge this without the approval of some other maintainers anyway.

lib/Doctrine/DBAL/Query/QueryBuilder.php
greg0ire: No I mean, maybe we should remove the parameter entirely?

> And by the way, as for develop branch, should I make a new merge request to the master branch to fix this?

You can edit your PR to target master, and rebase your branch on it: `git rebase --onto master origin/develop develop` (you should probably not have named your local branch `develop` BTW, it makes this a bit confusing :P)

greg0ire:
You should probably add something to https://github.com/doctrine/dbal/blob/master/docs/en/reference/query-builder.rst

lib/Doctrine/DBAL/Query/QueryBuilder.php
bingo-soft: Ups, I've never done that magic yet and made another fork for another PR. So, I think I will close this one and start another PR, if you don't mind. And I will take into account all you suggestions, probably except `final` key word

PR #3697, opened 2019-10-15
Add command to ping the database

mcfedr:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no

#### Summary

Adds a useful command to ping the database, it can do this in a loop to wait for it to be available. 

Particularly useful when you want to run commands linked to docker containers, and you need to wait for the database to be up, and you don't have `mysql` client installed, plus its more useful than that because it uses your existing doctrine config and can connect to any database.

Also great for tests when you need to provision and spin up a db server on the fly.

### DIFF
--- a/lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
+++ b/lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
@@ -1,0 +1,82 @@
+<?php
+
+namespace Doctrine\DBAL\Tools\Console\Command;
+
+use Doctrine\DBAL\Connection;
+use RuntimeException;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Throwable;
+use function is_numeric;
+use function sleep;
+use function sprintf;
+
+class PingCommand extends Command
+{
+    protected function configure()
+    {
+        $this
+            ->setName('dbal:ping')
+            ->setDescription('Check db is available')
+            ->addOption('limit', null, InputOption::VALUE_REQUIRED, 'Max number of pings to try', '1')
+            ->addOption('sleep', null, InputOption::VALUE_REQUIRED, 'Length of time (seconds) to sleep between pings', '1')
+            ->setHelp(<<<EOT
+Connects to the database to check if it is accessible.
+
+The exit code will be non-zero when the connection fails.
+EOT
+            );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output) : int
+    {
+        $limit = $input->getOption('limit');
+        if (! is_numeric($limit) || $limit < 0) {
+            throw new RuntimeException('Option "limit" must contain a positive integer value');
+        }
+        $sleep = $input->getOption('sleep');
+        if (! is_numeric($sleep) || $sleep < 0) {
+            throw new RuntimeException('Option "sleep" must contain a positive integer value');
+        }
+
+        return $this->waitForPing($this->getHelper('db')->getConnection(), (int) $limit, (int) $sleep, $output);
+    }
+
+    /**
+     * @return int > 0 for error
+     */
+    private function waitForPing(Connection $conn, int $limit, int $sleep, OutputInterface $output) : int
+    {
+        while (true) {
+            $last = $this->ping($conn, $output);
+            if ($last === 0 || --$limit <= 0) {
+                break;
+            }
+            sleep($sleep);
+        }
+
+        return $last;
+    }
+
+    /**
+     * @return int > 0 for error
+     */
+    private function ping(Connection $conn, OutputInterface $output) : int
+    {
+        try {
+            if ($conn->ping()) {
+                return 0;
+            }
+
+            $output->writeln('Ping failed');
+
+            return 1;
+        } catch (Throwable $e) {
+            $output->writeln(sprintf('Ping failed: <error>%s</error>', $e->getMessage()));
+
+            return 2;
+        }
+    }
+}
--- a/lib/Doctrine/DBAL/Tools/Console/ConsoleRunner.php
+++ b/lib/Doctrine/DBAL/Tools/Console/ConsoleRunner.php
@@ -4,6 +4,7 @@ namespace Doctrine\DBAL\Tools\Console;
 
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Tools\Console\Command\ImportCommand;
+use Doctrine\DBAL\Tools\Console\Command\PingCommand;
 use Doctrine\DBAL\Tools\Console\Command\ReservedWordsCommand;
 use Doctrine\DBAL\Tools\Console\Command\RunSqlCommand;
 use Doctrine\DBAL\Tools\Console\Helper\ConnectionHelper;
@@ -58,6 +59,7 @@ class ConsoleRunner
             new RunSqlCommand(),
             new ImportCommand(),
             new ReservedWordsCommand(),
+            new PingCommand(),
         ]);
     }
 
--- a/tests/Doctrine/Tests/DBAL/Tools/Console/Command/PingCommandTest.php
+++ b/tests/Doctrine/Tests/DBAL/Tools/Console/Command/PingCommandTest.php
@@ -1,0 +1,147 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Tools\Console\Command;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Tools\Console\Command\PingCommand;
+use Doctrine\DBAL\Tools\Console\ConsoleRunner;
+use PDOException;
+use PHPUnit\Framework\TestCase;
+use RuntimeException;
+use Symfony\Component\Console\Application;
+use Symfony\Component\Console\Tester\CommandTester;
+
+class PingCommandTest extends TestCase
+{
+    /** @var CommandTester */
+    private $commandTester;
+    /** @var PingCommand */
+    private $command;
+
+    /** @var Connection */
+    private $connectionMock;
+
+    protected function setUp() : void
+    {
+        $application = new Application();
+        $application->add(new PingCommand());
+
+        $this->command       = $application->find('dbal:ping');
+        $this->commandTester = new CommandTester($this->command);
+
+        $this->connectionMock = $this->createMock(Connection::class);
+
+        $helperSet = ConsoleRunner::createHelperSet($this->connectionMock);
+        $this->command->setHelperSet($helperSet);
+    }
+
+    public function testConnectionWorking() : void
+    {
+        $this->connectionMock
+            ->expects($this->once())
+            ->method('ping')
+            ->willReturn(true);
+
+        $this->commandTester->execute([]);
+
+        self::assertSame(0, $this->commandTester->getStatusCode());
+    }
+
+    public function testConnectionNotWorking() : void
+    {
+        $this->connectionMock
+            ->expects($this->once())
+            ->method('ping')
+            ->willReturn(false);
+
+        $this->commandTester->execute([]);
+
+        self::assertSame(1, $this->commandTester->getStatusCode());
+        self::assertSame("Ping failed\n", $this->commandTester->getDisplay(true));
+    }
+
+    public function testConnectionErrors() : void
+    {
+        $this->connectionMock
+            ->expects($this->once())
+            ->method('ping')
+            ->willThrowException(new PDOException('Connection failed'));
+
+        $this->commandTester->execute([]);
+
+        self::assertSame(2, $this->commandTester->getStatusCode());
+        self::assertSame("Ping failed: Connection failed\n", $this->commandTester->getDisplay(true));
+    }
+
+    public function testConnectionNotWorkingLoop() : void
+    {
+        $this->connectionMock
+            ->expects($this->exactly(3))
+            ->method('ping')
+            ->willReturn(false);
+
+        $this->commandTester->execute([
+            '--limit' => '3',
+            '--sleep' => '0',
+        ]);
+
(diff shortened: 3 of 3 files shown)


lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
```diff
@@ -0,0 +1,80 @@
...
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Throwable;
+use function is_numeric;
+use function sleep;
+use function sprintf;
+
+class PingCommand extends Command
```
greg0ire: Should we mark this as `final`?

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
```diff
@@ -0,0 +1,80 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected static $defaultName = 'dbal:ping';
+
+    protected function configure()
+    {
+        $this
+            ->setDescription('Check db is available')
```
greg0ire:
```suggestion
            ->setDescription('Check whether a database is available')
```

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
```diff
@@ -0,0 +1,80 @@
...
+        $limit = $input->getOption('limit');
+        if (! is_numeric($limit)) {
+            throw new LogicException("Option 'limit' must contain an integer value");
+        }
+        $sleep = $input->getOption('sleep');
+        if (! is_numeric($sleep)) {
+            throw new LogicException("Option 'sleep' must contain an integer value");
+        }
+
+        return $this->waitForPing($this->getHelper('db')->getConnection(), (int) $limit, (int) $sleep, $output);
```
greg0ire: :thinking: doesn't the dbal allow for several connections?

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
```diff
@@ -0,0 +1,80 @@
...
+            ->setDescription('Check db is available')
+            ->addOption('limit', null, InputOption::VALUE_REQUIRED, 'Max number of pings to try', '1')
+            ->addOption('sleep', null, InputOption::VALUE_REQUIRED, 'Length of time (seconds) to sleep between pings', '1');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output) : int
+    {
+        $limit = $input->getOption('limit');
+        if (! is_numeric($limit)) {
+            throw new LogicException("Option 'limit' must contain an integer value");
```
greg0ire: I'd use `UnexpectedValueException` here, because there is no issue with the code, so it must be a `RuntimeException`

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
```diff
@@ -0,0 +1,80 @@
...
+     * {@inheritdoc}
+     */
+    protected static $defaultName = 'dbal:ping';
+
+    protected function configure()
+    {
+        $this
+            ->setDescription('Check db is available')
+            ->addOption('limit', null, InputOption::VALUE_REQUIRED, 'Max number of pings to try', '1')
+            ->addOption('sleep', null, InputOption::VALUE_REQUIRED, 'Length of time (seconds) to sleep between pings', '1');
```
greg0ire:
```suggestion
            ->addOption('sleep', null, InputOption::VALUE_REQUIRED, 'Amount of time (in seconds) to sleep between pings', '1');
```

greg0ire (changes requested):

mcfedr:
Commands shouldn't be final as they are wrapped in [DoctrineBundle](https://github.com/doctrine/DoctrineBundle/blob/1.12.x/Command/Proxy/RunSqlDoctrineCommand.php) -  this bundle also provides the arguments to allow connection switching.

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
mcfedr: I followed the style of https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Tools/Console/Command/RunSqlCommand.php#L61 - I can change it if you want but seems more consistent not to.

greg0ire: I think that line might be wrong too, sadly.

https://www.php.net/manual/en/class.runtimeexception.php
https://www.php.net/manual/en/class.logicexception.php

If you read the description for the 2 exception parent classes above what do you conclude?

greg0ire:
> @greg0ire Commands shouldn't be final as they are wrapped in DoctrineBundle - this bundle also provides the arguments to allow connection switching.
Would it be possible to do the wrapping using decoration? This could avoid issues with type declarations.

mcfedr:
> Would it be possible to do the wrapping using decoration?

I'm sure it would, but it seems outside of the scope of this PR, rather some refactoring to be done, first in DoctrineBundle, and then here with all the commands.

> This could avoid issues with type declarations.

What issues?

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
mcfedr: You are probabably right, although its a bit subjective, for example the command can be trigger like it is in testing, with code

```
$this->commandTester->execute([
    '--limit' => '5',
    '--sleep' => '0',
]);
```

I would be inclined to follow the given style, and fix it in a future PR across all the commands.

greg0ire: > I would be inclined to follow the given style, and fix it in a future PR across all the commands.
I'm ok with it if you commit to fixing it in a future PR. If not, please don't add more lines to fix.

mcfedr: Ok, I've switched this out for `RuntimeException`

mcfedr:
@greg0ire I think all the issues have been addressed.

greg0ire:
> This could avoid issues with type declarations.


Changes in type declarations in parent classes can be incompatible with the parent, and *vice versa*, and disappear if you use decoration.

For instance, phpunit has been adding the `: void` return type hint to `TestCase::setUp` lately, causing a lot of trouble in a big number of libraries (example: https://github.com/symfony/symfony/issues/30071)

If you choose decoration, I would not require you to refactor the doctrine bundle and the dbal. Maybe documenting extension as a deprecated way of doing things in the project or as a github issue would be sufficient.

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
greg0ire: Let's mark that one as unresolved until we get feedback from other maintainers on the decoration vs extension debate.

lcobucci: We should, makes things clearer IMHO

mcfedr:
I'm just looking at what it would take to decorate this command in bundle, it feels like a much more brittle solution in this case - for example I'd have to get all the options, and maybe the (currently) empty arguments and recreate them - the only non-private parts of the command is the interface inherited from `Command`

greg0ire:
Ok, if it's not straightforward let's stay with extension I guess. I think the command in the bundle will not be just about renaming this command and using a different description is it? Could you please open the PR on the bundle to show us how it will look like?

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
```diff
@@ -0,0 +1,80 @@
...
+                return 0;
+            }
+
+            $output->writeln('Ping failed');
+
+            return 1;
+        } catch (Throwable $e) {
+            $output->writeln(sprintf('Connection error: <error>%s</error>', $e->getMessage()));
+
+            return 2;
```
morozov: What difference does it make for the caller? The ping either succeeded or didn't. Note that in DBAL v3, `ping` will throw an exception on failure, so it's not clear how this would have to be reimplemented w/o breaking the existing behavior.

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * @return int > 0 for error
+     */
+    private function waitForPing(Connection $conn, int $limit, int $sleep, OutputInterface $output) : int
+    {
+        while (true) {
+            $last = $this->ping($conn, $output);
+            if ($last === 0 || --$limit === 0) {
+                return $last;
```
morozov: This could be replaced with `break` to avoid the duplication of the `return $last` logic and allow for further changes.

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
```diff
@@ -0,0 +1,80 @@
...
+        $this
+            ->setDescription('Check db is available')
+            ->addOption('limit', null, InputOption::VALUE_REQUIRED, 'Max number of pings to try', '1')
+            ->addOption('sleep', null, InputOption::VALUE_REQUIRED, 'Length of time (seconds) to sleep between pings', '1');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output) : int
+    {
+        $limit = $input->getOption('limit');
+        if (! is_numeric($limit)) {
```
morozov: Should it be checked to be an integer `>= 1`? Otherwise, combined with `--$limit === 0`, a negative number will result in an infinite loop.

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
```diff
@@ -0,0 +1,80 @@
...
+            throw new RuntimeException("Option 'limit' must contain an integer value");
+        }
+        $sleep = $input->getOption('sleep');
+        if (! is_numeric($sleep)) {
```
morozov: Should it be checked to be an integer `>= 0`?

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
```diff
@@ -0,0 +1,80 @@
...
+            ->setDescription('Check db is available')
+            ->addOption('limit', null, InputOption::VALUE_REQUIRED, 'Max number of pings to try', '1')
+            ->addOption('sleep', null, InputOption::VALUE_REQUIRED, 'Length of time (seconds) to sleep between pings', '1');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output) : int
+    {
+        $limit = $input->getOption('limit');
+        if (! is_numeric($limit)) {
+            throw new RuntimeException("Option 'limit' must contain an integer value");
```
morozov: Please use double quotes for parameter names.

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
```diff
@@ -0,0 +1,80 @@
...
+        }
+        $sleep = $input->getOption('sleep');
+        if (! is_numeric($sleep)) {
+            throw new RuntimeException("Option 'sleep' must contain an integer value");
```
morozov: Please use double quotes for parameter names.

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /**
+     * @return int > 0 for error
+     */
+    private function waitForPing(Connection $conn, int $limit, int $sleep, OutputInterface $output) : int
+    {
+        while (true) {
+            $last = $this->ping($conn, $output);
+            if ($last === 0 || --$limit === 0) {
```
morozov: Please change to `--$limit <= 0` to avoid a potential infinite loop.

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
mcfedr: It probably makes little difference to the caller, but as there are two different failures it doesnt do any hard to be more informative - note that this becomes the exit code from the process

mcfedr: Ok, I've changed the text to just be 'ping failed', so it won't be misleading when DBAL 3 throws on different failure types - anyway the caught exception will say its a connection error or something else.

mcfedr:
@greg0ire I've now opened https://github.com/doctrine/DoctrineBundle/pull/1035 with the command added there.

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
greg0ire: I couldn't think of a way to avoid extension here, let's resolve this.

mcfedr:
@greg0ire I think all is good now?

greg0ire (approved):
Yes, sorry!

greg0ire (approved):
Yes, sorry!

mcfedr:
> Yes, sorry!

No problem :)

SenseException (approved):

lib/Doctrine/DBAL/Tools/Console/Command/PingCommand.php
morozov: I still don't understand the meaning of the two different return codes and how they will be supported in `3.0`.

morozov (changes requested):
@mcfedr could you describe the difference between the 1 and 2 exit codes from the user perspective?

morozov:
Pinging a database is no longer supported (https://github.com/doctrine/dbal/pull/4128).

PR #3699, opened 2019-10-15
fixed some misspellings of "Fulfilled"

kylechallis:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | misspellings

#### Summary

fixed some misspellings of "Fulfilled"

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Index.php
+++ b/lib/Doctrine/DBAL/Schema/Index.php
@@ -196,7 +196,7 @@ class Index extends AbstractAsset implements Constraint
      *
      * @return bool
      */
-    public function isFullfilledBy(Index $other)
+    public function isFulfilledBy(Index $other)
     {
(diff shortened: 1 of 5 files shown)


lib/Doctrine/DBAL/Schema/Index.php
```diff
@@ -196,7 +196,7 @@ public function spansColumns(array $columnNames)
      *
      * @return bool
      */
-    public function isFullfilledBy(Index $other)
+    public function isFulfilledBy(Index $other)
```
greg0ire: That's a BC-break. Please create a mispelt method that calls this one, and throws a deprecation notice before doing so.

zera555: Thank you! Will do.

SenseException:
@zera555 Did you  close this PR by accident?

PR #3700, opened 2019-10-16
Resolve an issue where ServerVersion could not be obtained correctly

Yi-pixel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 1. \PDO instance gets ServerVersion

#### Summary

During use, I found that when `$this->_conn` is a `\PDO` instance,
this method can't get the MySQL version of the server correctly,
so  can't select the correct Platform, which leads to some unexpected
situations. 

eg:

When I use \PDO as the connection, there is a json type field in a table in my database.
At this point my MySQL database version is 8.0+ , But it doesn't work

```
Unknown database type json requested, Doctrine\DBAL\Platforms\MySqlPlatform may not support it.
```
This patch is to solve this problem, I am using this problem found in `laravel-ide-helper`.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -467,6 +467,10 @@ class Connection implements DriverConnection
             return $this->_conn->getServerVersion();
         }
 
+        if ($this->_conn instanceof \PDO) {
+            return $this->_conn->getAttribute(\PDO::ATTR_SERVER_VERSION);
+        }
+
         // Unable to detect platform version.
         return null;
     }


lib/Doctrine/DBAL/Connection.php
```diff
@@ -467,6 +467,11 @@ private function getServerVersion()
             return $this->_conn->getServerVersion();
         }
 
+        // if \PDO
+        if($this->_conn instanceof \PDO){
```
greg0ire:
```suggestion
        if ($this->_conn instanceof \PDO){
```

lib/Doctrine/DBAL/Connection.php
```diff
@@ -467,6 +467,11 @@ private function getServerVersion()
             return $this->_conn->getServerVersion();
         }
 
+        // if \PDO
```
greg0ire: That comment is not really useful is it?

lib/Doctrine/DBAL/Connection.php
WonlyCms: I have tried to re-edit and submit, please review.

greg0ire:
> When I use \PDO as the connection

Is the property phpdoc at line 86 wrong then :thinking: ?

greg0ire:
Can you check what type is your connection? Is it a \PDO instance or an instance of the extending class I showed above? Maybe the Laravel integration package is doing weird things and omitted to instantiate a instance of `Connection`

WonlyCms:
> Can you check what type is your connection? Is it a \PDO instance or an instance of the extending class I showed above? Maybe the Laravel integration package is doing weird things and omitted to instantiate a instance of `Connection`. As you can see, we don't do any type check on line 193…


I found the instantiation Connection code in Laravel's source code, and from here I found that Laravel passed PDO directly as a parameter.

https://github.com/laravel/framework/blob/6.x/src/Illuminate/Database/Connection.php#L7

https://github.com/laravel/framework/blob/6.x/src/Illuminate/Database/Connection.php#L896-L910

Can be sure that Laravel is not implemented `vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/ServerInfoAwareConnection.php` and `\Doctrine\DBAL\Driver\Connection`,  So the condition is not established

https://github.com/doctrine/dbal/blob/327eabeabf85a7d9337358a79cc1c9eaab7f76cf/lib/Doctrine/DBAL/Connection.php#L464-L468

Since # is not verified, should I now provide feedback to Laravel or accept my request?

https://github.com/doctrine/dbal/blob/327eabeabf85a7d9337358a79cc1c9eaab7f76cf/lib/Doctrine/DBAL/Connection.php#L192-L196

greg0ire:
I think they chose to do that because there is no easy way to create a `PDOConnection` from an already created PDO object… If that change is hard for Laravel, maybe we could introduce a new class that is able to decorate an existing PDO instance and implement the same interfaces `PDOConnection` implements instead of using extension?

greg0ire:
> but I noticed that in #↓, I need some parameters, I can't get it from the \PDO instance, such as username and password, because Laravel doesn't provide it.

That's why I'm suggesting creating a new implementation of `Connection` in Doctrine, that would use decoration instead of inheritance, and that would not have these parameters in its ctor:

```
class PdoDecoratorConnection implements Connection, ServerInfoOrWhateverTheActualNameIs
{
    public function __construct(\PDO $pdo) {
    {
         …
    }
    // proxyAllMethodsHere
}
```

Issue #3710, opened 2019-10-31
getQuotedName Does Not Quote Columns With Upper and Mixed Cases


paulrrogers:
### Bug Report

Transitioning to Postgresql with mixed case tables and columns is a challenge when the DBAL does not quote them. That's because Pg coerces them to lower case. And if the ORM does quote these mixed case relations then things break. Ideally I imagine it'd consistently quote any asset names which have upper cased letters in their names.

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.5, master

#### Current behaviour

getQuotedName won't quote a mixed or upper case column/table name

#### How to reproduce

`tests/Doctrine/Tests/DBAL/Schema/ColumnTest.php`
`testQuotedColumnName()`
```php
       $columnMixedCase = new Column('fooBar', $string, []);
       $postgresqlPlatform = new PostgreSqlPlatform();

       self::assertEquals('"fooBar"', $columnMixedCase->getQuotedName($postgresqlPlatform));
```

#### Expected behaviour

return would be quoted

PR #3711, opened 2019-10-31
Fixed upper and mixed case columns not quoted.

paulrrogers:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3710

#### Summary

Checks for presence of upper and lowercase letters and if so then quotes the value


morozov:
@paulrrogers than you for the pull request. Since it affects a platform-specific behavior, I'd prefer to have a functional test that would guarantee that the new code works with the target platform. Not just theory by making assertions against an implementation detail.

paulrrogers:
> @paulrrogers thank you for the pull request. Since it affects a platform-specific behavior, I'd prefer to have a functional test that would guarantee that the new code works with the target platform. Not just in theory by making assertions against an implementation detail.

Sorry if I'm not up-to-speed on the terminology here. By 'functional' do you mean a test that covers the integration with the platform? I.e. it covers running queries through a PostgreSQL instance?

morozov:
> By 'functional' do you mean a test that covers the integration with the platform? I.e. it covers running queries through a PostgreSQL instance?

Yes.

paulrrogers:
@morozov I've added a functional test, though not sure why portability tests like `testFetchAllNullColumn` are now failing. Perhaps Doctrine is doing some mocking that has been disrupted or my fix is incomplete?

Can you offer any insight?

morozov:
Not sure why it's failing but the currently provided test is not what I had in mind. Please look at the [`DateExpressionTest`](https://github.com/doctrine/dbal/blob/v2.9.2/tests/Doctrine/Tests/DBAL/Functional/Platform/DateExpressionTest.php) for example. It tests the code that implements a database abstraction by interacting with the real database. That proves that the code works correctly, not just as expected.

I didn't look at the related issue but from the test itself, it's not clear what problem the patch solves. It should be.

paulrrogers:
> …the currently provided test is not what I had in mind. Please look at the [`DateExpressionTest`](https://github.com/doctrine/dbal/blob/v2.9.2/tests/Doctrine/Tests/DBAL/Functional/Platform/DateExpressionTest.php) for example. It tests the code that implements a database abstraction by interacting with the real database. That proves that the code works correctly, not just as expected.

My apologies, didn't realize I forgot to check-in `testQuotingMixedCaseAssets`. Please let me know if that is clear enough.

lib/Doctrine/DBAL/Schema/AbstractAsset.php
```diff
@@ -177,10 +178,12 @@ public function getName()
...
     public function getQuotedName(AbstractPlatform $platform)
     {
-        $keywords = $platform->getReservedKeywordsList();
-        $parts    = explode('.', $this->getName());
+        $keywords     = $platform->getReservedKeywordsList();
+        $isCaseFolded = $platform->isCaseFoldedForUnquotedAssets();
+        $parts        = explode('.', $this->getName());
         foreach ($parts as $k => $v) {
-            $parts[$k] = $this->_quoted || $keywords->isKeyword($v) ? $platform->quoteIdentifier($v) : $v;
+            $caseNeedsQuoting = $isCaseFolded && preg_match('/(\p{Ll}.*\p{Lu}|\p{Lu}.*\p{Ll})/u', $v);
```
SenseException: I'd like to see this line in a separate private method. This way, the regex might not need to run when the first two matches are true in the next line.

PR #3643, opened 2019-08-03
Bumped PHPUnit requrement to ^8.3.3, removed dependency on symfony/phpunit-bridge

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | [build #567334606](https://travis-ci.org/doctrine/dbal/builds/567334606)

Once PHPUnit 8.3 was released (specifically, https://github.com/sebastianbergmann/phpunit/commit/db1dce395c1843f095319a4d9652c2853ea785df), the tests started failing because of symfony/symfony#3287. DBAL no longer triggers new `E_USER_DEPRECATED` errors at runtime since the Symfony community is sensitive about them, so the listener is not really longer relevant.

PR #3719, opened 2019-11-06
Remove phpunit-bridge specific annotations

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | pedantic
| BC Break     | no
| Fixed issues | n/a

#### Summary

The phpunit-bridge has been removed in #3643, so these no longer serve
any purpose. The tests themselves are kept since they check that there
is no crash when using an unknown option, and do not need a phpunit
assertion to do so.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/PDOStatementTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/PDOStatementTest.php
@@ -28,10 +28,6 @@ class PDOStatementTest extends DbalFunctionalTestCase
         $this->connection->getSchemaManager()->dropAndCreateTable($table);
     }
 
-    /**
-     * @group legacy
-     * @expectedDeprecation Using a PDO fetch mode or their combination (%d given) is deprecated and will cause an error in Doctrine 3.0.
-     */
     public function testPDOSpecificModeIsAccepted() : void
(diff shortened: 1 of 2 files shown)


morozov (approved):

tests/Doctrine/Tests/DBAL/Functional/PDOStatementTest.php
```diff
@@ -28,10 +28,6 @@ protected function setUp() : void
         $this->connection->getSchemaManager()->dropAndCreateTable($table);
     }
 
-    /**
-     * @group legacy
```
lcobucci: `@group` is from PHPUnit, though.

greg0ire: Yeah, but `legacy` isn't.

lcobucci: I thought we were using that group to identify the tests we were going to drop eventually because of legacy :laughing:

greg0ire: Wouldn't be a bad idea, but right now I bet it's far from being exhaustive :P

Issue #3717, opened 2019-11-04
Invalid reference in deprecation notices regarding types (DefaultTypes)


Pnoexz:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.0

#### Summary

In [this pull request](https://github.com/doctrine/dbal/pull/3356/files) a new deprecation around Types was introduced. It looks like the author of the PR initially used `DefaultTypes` for the name of the new class, but was later changed to Types, leaving references to the former in @see tags on the deprecation notices. [example](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Types/Type.php#L20)

```
vendor/doctrine/dbal/lib/Doctrine/DBAL/Types/Type.php
21:/** @deprecated Use {@see DefaultTypes::BIGINT} instead. */
24:/** @deprecated Use {@see DefaultTypes::BINARY} instead. */
27:/** @deprecated Use {@see DefaultTypes::BLOB} instead. */
30:/** @deprecated Use {@see DefaultTypes::BOOLEAN} instead. */
33:/** @deprecated Use {@see DefaultTypes::DATE_MUTABLE} instead. */
36:/** @deprecated Use {@see DefaultTypes::DATE_IMMUTABLE} instead. */
39:/** @deprecated Use {@see DefaultTypes::DATEINTERVAL} instead. */
42:/** @deprecated Use {@see DefaultTypes::DATETIME_MUTABLE} instead. */
45:/** @deprecated Use {@see DefaultTypes::DATETIME_IMMUTABLE} instead. */
48:/** @deprecated Use {@see DefaultTypes::DATETIMETZ_MUTABLE} instead. */
51:/** @deprecated Use {@see DefaultTypes::DATETIMETZ_IMMUTABLE} instead. */
54:/** @deprecated Use {@see DefaultTypes::DECIMAL} instead. */
57:/** @deprecated Use {@see DefaultTypes::FLOAT} instead. */
60:/** @deprecated Use {@see DefaultTypes::GUID} instead. */
63:/** @deprecated Use {@see DefaultTypes::INTEGER} instead. */
66:/** @deprecated Use {@see DefaultTypes::JSON} instead. */
69:/** @deprecated Use {@see DefaultTypes::JSON_ARRAY} instead. */
72:/** @deprecated Use {@see DefaultTypes::OBJECT} instead. */
[...]

PR #3723, opened 2019-11-08
Fix annotations

enumag:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3717

### DIFF
--- a/lib/Doctrine/DBAL/Types/Type.php
+++ b/lib/Doctrine/DBAL/Types/Type.php
@@ -18,79 +18,79 @@ use function substr;
  */
 abstract class Type
 {
-    /** @deprecated Use {@see DefaultTypes::BIGINT} instead. */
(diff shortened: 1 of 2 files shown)


andrew-demb:
We creating similar pull requests simultaneously 😁

#3724

enumag:
lol looks like I was 3 mins faster but you caught one additional case that I missed. :grin: 

Closing in favor of yours.

enumag:
Hmm actually no. I think this one is better because of the target branch.

morozov:
Thank you, @enumag.

manu-sparheld:
When can I expect it to be released ?
TIA

morozov:
@manu-sparheld there's no due date for the [2.10.1](https://github.com/doctrine/dbal/milestone/74) release.

PR #3730, opened 2019-11-11
Add bin to .gitattributes

reedy:
Needs symfony/console from `require-dev` so most composer includes probably don't want these files :)

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,6 +5,7 @@
 /.gitignore export-ignore
 /.scrutinizer.yml export-ignore
 /.travis.yml export-ignore
+/bin export-ignore
 /build.properties export-ignore
 /build.xml export-ignore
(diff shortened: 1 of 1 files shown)


andrew-demb:
I think we should check for composer warning about missed bin (specified in composer.json) before merge.

.gitattributes
```diff
@@ -5,6 +5,7 @@
 /.gitignore export-ignore
 /.scrutinizer.yml export-ignore
 /.travis.yml export-ignore
+/bin export-ignore
```
morozov, line 8: This way, the CLI utils will be only available during the development of the DBAL itself but not to other projects. See https://github.com/doctrine/dbal/pull/3956 for example.

PR #3738, opened 2019-11-19
Fix breaks named parameters in Oracle

### LINKED ISSUE #3722: OCI8 binding prepared statements
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.10

#### Summary

After update to 2.10 variable binding not working as before. Various errors are thrown, for example ORA-01036 on select statements, and ORA-01008 on update statements 

#### Previous behaviour

Working with no issues

#### Current behavior

query example:
```sql
update rates
set
    rate_fixing = :rate
where
    to_char(rate_date, 'DD.MM.YYYY') = :d
    and currency = :cur
```
execution code:
```php
         $r = $q->execute(
                    [
                        ':rate'     => round($t['rate'], 6),
                        ':d' => $t['date'],
                        ':cur'   => $t['currency'],
                    ]
                );
```
```log
request.CRITICAL: Uncaught PHP Exception Doctrine\DBAL\DBALException: "An exception occurred while executing 'update rates set     rate_fixing = :rate where     to_char(rate_date, 'DD.MM.YYYY') = :d and currency = :cur' with params [6.69515, "11.09.2019", "USD"]:  Notice: Undefined index: :rate" at G:\apache_root\s4\dan\vendor\doctrine\dbal\lib\Doctrine\DBAL\DBALException.php line 169 {"exception":"[object] (Doctrine\\DBAL\\DBALException(code: 0): An exception occurred while executing 'update rates\r\nset\r\n    rate_fixing = :rate\r\nwhere\r\n    to_char(rate_date, 'DD.MM.YYYY') = :d\r\n    and currency = :cur' with params [6.69515, \"11.09.2019\", \"USD\"]:\n\nNotice: Undefined index: :rate at G:\\apache_root\\s4\\dan\\vendor\\doctrine\\dbal\\lib\\Doctrine\\DBAL\\DBALException.php:169, ErrorException(code: 0): Notice: Undefined index: :rate at G:\\apache_root\\s4\\dan\\vendor\\doctrine\\dbal\\lib\\Doctrine\\DBAL\\Driver\\OCI8\\OCI8Statement.php:281)"} []

```

eisberg:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes

Fixes #3722
Closes #3779

#### Summary

Fix breaks named parameters in Oracle ( see issue  #3722 )

### DIFF
--- a/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
+++ b/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
@@ -276,9 +276,15 @@ class OCI8Statement implements IteratorAggregate, Statement
     /**
      * {@inheritdoc}
      */
-    public function bindParam($column, &$variable, $type = ParameterType::STRING, $length = null)
+    public function bindParam($param, &$variable, $type = ParameterType::STRING, $length = null)
     {
-        $column = $this->_paramMap[$column];
+        if (is_int($param)) {
+            if (! isset($this->_paramMap[$param])) {
+                throw new OCI8Exception(sprintf('Could not find variable mapping with index %d, in the SQL statement', $param));
+            }
+
+            $param = $this->_paramMap[$param];
+        }
 
         if ($type === ParameterType::LARGE_OBJECT) {
             $lob = oci_new_descriptor($this->_dbh, OCI_D_LOB);
@@ -291,11 +297,11 @@ class OCI8Statement implements IteratorAggregate, Statement
             $variable =& $lob;
         }
 
-        $this->boundValues[$column] =& $variable;
+        $this->boundValues[$param] =& $variable;
 
         return oci_bind_by_name(
             $this->_sth,
-            $column,
+            $param,
             $variable,
             $length ?? -1,
             $this->convertParameterType($type)
--- a/tests/Doctrine/Tests/DBAL/Functional/Driver/OCI8/StatementTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Driver/OCI8/StatementTest.php
@@ -38,14 +38,38 @@ class StatementTest extends DbalFunctionalTestCase
     }
 
     /**
+     * Low-level approach to working with parameter binding
+     *
+     * @param mixed[] $params
+     * @param mixed[] $expected
+     *
+     * @dataProvider queryConversionProvider
+     */
+    public function testStatementBindParameters(string $query, array $params, array $expected) : void
+    {
+        $stmt = $this->connection->prepare($query);
+        $stmt->execute($params);
+
+        self::assertEquals(
+            $expected,
+            $stmt->fetch()
+        );
+    }
+
+    /**
      * @return array<string, array<int, mixed>>
      */
     public static function queryConversionProvider() : iterable
     {
         return [
-            'simple' => [
+            'positional' => [
                 'SELECT ? COL1 FROM DUAL',
                 [1],
+                ['COL1' => 1],
+            ],
+            'named' => [
+                'SELECT :COL COL1 FROM DUAL',
+                [':COL' => 1],
                 ['COL1' => 1],
             ],
             'literal-with-placeholder' => [


morozov:
@eisberg could you also add a test?

tests/Doctrine/Tests/DBAL/Driver/OCI8/OCI8StatementTest.php
```diff
@@ -74,6 +74,39 @@ public function testExecute(array $params) : void
...
+     * @param mixed[] $params
+     *
+     * @dataProvider bindParamsDataProvider
+     */
+    public function testBindParam(array $params, string $query) : void
+    {
+        $conn = $this->createMock(OCI8Connection::class);
+        /** @var OCI8Statement|MockObject $statement */
+        $statement = $this->getMockBuilder(OCI8Statement::class)
+            ->onlyMethods([  'prepareQuery', 'bindParamByName'])
```
morozov: Please take a look at `NamedParametersTest`? I'd rather have a functional test when it comes to testing the logic that involves a database (binding named parameters in this case). Is the existing suite missing something?

morozov (changes requested):

tests/Doctrine/Tests/DBAL/Driver/OCI8/OCI8StatementTest.php
eisberg: A set of "NamedParametersTest" is sufficient. But in " NamedParametersTest "the higher-level method"executeQuery" is tested. This method works well, because inside it named placeholders ( :param) are replaced by positioned (? ). And only then does this modified request arrive at the oracle statement input.  Added test in Functional\Driver\OCI8\StatementTest as these fixes concern OCI8StatementTest class.

tests/Doctrine/Tests/DBAL/Functional/Driver/OCI8/StatementTest.php
```diff
@@ -39,6 +39,23 @@ public function testQueryConversion(string $query, array $params, array $expecte
...
+    /**
+     * @param mixed[] $params
+     * @param mixed[] $expected
+     *
+     * @dataProvider queryConversionProvider
+     */
+    public function testQueryPrepare(string $query, array $params, array $expected) : void
+    {
+        $stmt = $this->connection->prepare($query);
+        $stmt->execute($params);
```
morozov: Why does the test named `testQueryPrepare()` prepares _and executes_ the query? How is it different from the one that already does `executeQuery()`?

tests/Doctrine/Tests/DBAL/Functional/Driver/OCI8/StatementTest.php
```diff
@@ -50,6 +67,11 @@ public static function queryConversionProvider() : iterable
                 [1],
                 ['COL1' => 1],
             ],
+            'named' => [
```
morozov: Yeah, this seems to be indeed a missing case. Would it make sense to rename the previous data point from "simple" to "positional" to better reflect the difference between them?

tests/Doctrine/Tests/DBAL/Functional/Driver/OCI8/StatementTest.php
eisberg: `connection->executeQuery` is a high-level method that works correctly with both `named` and `positioned` parameters. In "testQueryPrepare" - oci8statement is prepared, parameters are bound and then it is executed. Here just also there was an error with `named` parameters though Oracle just only supports them

eisberg: I renamed the name of the test to 'testStatementBindParameters' for a better understanding of what is happening

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -90,6 +90,19 @@ class OCI8Statement implements IteratorAggregate, Statement
      */
     private $result = false;
 
+    /**
+     * @param resource $dbh The connection handle.
+     *
+     * @return resource
+     */
+    protected function prepareQuery($dbh, string $query)
```
morozov: Why is this method needed?

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -273,7 +283,7 @@ public function bindValue($param, $value, int $type = ParameterType::STRING) : v
      */
     public function bindParam($param, &$variable, int $type = ParameterType::STRING, ?int $length = null) : void
     {
-        $param = $this->_paramMap[$param];
+        $param = $this->_paramMap[$param] ?? (string) $param;
```
morozov: Why is casting needed? I believe the logic should be about the following:
```php
if (is_int($param)) {
    if (! isset($this->_paramMap[$param])) {
        throw new OCI8Exception(/* ... */);
    }

    $param = $this->_paramMap[$param];
}
```

tests/Doctrine/Tests/DBAL/Functional/Driver/OCI8/StatementTest.php
```diff
@@ -39,17 +41,53 @@ public function testQueryConversion(string $query, array $params, array $expecte
...
+     *
+     * @dataProvider queryConversionProvider
+     */
+    public function testStatementBindParameters(string $query, array $params, array $expected) : void
+    {
+        $stmt         = $this->connection->prepare($query);
+        $hasZeroIndex = array_key_exists(0, $params);
+
+        foreach ($params as $key => $val) {
+            if ($hasZeroIndex && is_int($key)) {
```
morozov: Does this test cover any code path not covered by the data item added to `queryConversionProvider()`. This test seems to duplicate the code being tested and has conditional operators which usually identify a poorly designed test.

morozov (changes requested):

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
eisberg: > Why is casting needed?
Without casting to the string produces a phpstan error.


> I believe the logic should be about the following:

Ok

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
eisberg: i removed it

tests/Doctrine/Tests/DBAL/Functional/Driver/OCI8/StatementTest.php
eisberg: ok, changed it.  

The `testQueryConversion` test does not cover all use cases because it calls the `executeQuery` method on the `connection` object. In this case, the named parameters are transformed into positional ones. As a result, test cases with named parameters are skipped

tests/Doctrine/Tests/DBAL/Functional/Driver/OCI8/StatementTest.php
```diff
@@ -39,17 +39,41 @@ public function testQueryConversion(string $query, array $params, array $expecte
...
+     *
+     * @param mixed[] $params
+     * @param mixed[] $expected
+     *
+     * @dataProvider queryConversionProvider
+     */
+    public function testStatementBindParameters(string $query, array $params, array $expected) : void
+    {
+        $stmt = $this->connection->prepare($query);
+        $stmt->execute($params); // bind params in statement and execute it
```
morozov: >  // bind params in statement and execute it

This is obvious from the method phpDoc. Please remove.

morozov (commented):
> The `testQueryConversion` test does not cover all use cases because it calls the `executeQuery` method on the `connection` object. In this case, the named parameters are transformed into positional ones.

Thanks for bringing this up. We probably need to get rid of this behavior since, on the one hand, it doesn't make any sense for the drivers that support named parameters but don't support the positional ones. On the other hand, it's a driver-specific logic implemented on the wrapper level.

We'll try to take care of that in a different issue. Besides a few nitpicks, looks good.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -273,7 +273,13 @@ public function bindValue($param, $value, int $type = ParameterType::STRING) : v
      */
     public function bindParam($param, &$variable, int $type = ParameterType::STRING, ?int $length = null) : void
     {
-        $param = $this->_paramMap[$param];
+        if (is_int($param)) {
+            if (! isset($this->_paramMap[$param])) {
+                throw new OCI8Exception(sprintf('Could not find variable mapping with index "%s", in the SQL statement', $param));
```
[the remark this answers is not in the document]
eisberg: Ok, fixed

tests/Doctrine/Tests/DBAL/Functional/Driver/OCI8/StatementTest.php
eisberg: removed

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -273,7 +273,13 @@ public function bindValue($param, $value, int $type = ParameterType::STRING) : v
      */
     public function bindParam($param, &$variable, int $type = ParameterType::STRING, ?int $length = null) : void
     {
-        $param = $this->_paramMap[$param];
+        if (is_int($param)) {
+            if (! isset($this->_paramMap[$param])) {
+                throw new OCI8Exception(sprintf('Could not find variable mapping with index "%d", in the SQL statement', $param));
```
morozov: Please see https://github.com/doctrine/dbal/pull/3738#discussion_r349740890:

> Should it be `%d` instead of `"%s"` since we're printing a numeric index?

morozov (changes requested):

lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
```diff
@@ -273,7 +273,13 @@ public function bindValue($param, $value, int $type = ParameterType::STRING) : v
      */
     public function bindParam($param, &$variable, int $type = ParameterType::STRING, ?int $length = null) : void
     {
-        $param = $this->_paramMap[$param];
+        if (is_int($param)) {
+            if (! isset($this->_paramMap[$param])) {
+                throw new OCI8Exception(sprintf('Could not find variable mapping with index %d, in the SQL statement', $param));
```
morozov: The idea behind throwing an exception is that in this case, it wouldn't have worked previously anyways, so we're just making the error more explicit. Need to double-check that before merging.

eisberg: any news?

morozov: @eisberg I forgot about it. Thank you for the reminder.

This is a valid code example that works on `2.9` and a build from this PR:

```php
$stmt = $conn->prepare('SELECT :p FROM DUAL');
$stmt->bindValue('p', 'foo');
$stmt->execute();
echo $stmt->fetchColumn(), PHP_EOL;
// foo
```

This is an example of invalid code that is not supposed to work:

```php
$stmt = $conn->prepare('SELECT :p FROM DUAL');
$stmt->bindValue(1, 'foo');
$stmt->execute();
echo $stmt->fetchColumn(), PHP_EOL;
```

On `2.9` it produces:
> Warning: oci_bind_by_name(): ORA-01036: illegal variable name/number in lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php on line 287
> Uncaught Doctrine\DBAL\Driver\OCI8\OCI8Exception: ORA-01008: not all variables bound

On a build from this patch it produces:
> Uncaught Doctrine\DBAL\Driver\OCI8\OCI8Exception: Could not find variable mapping with index 1, in the SQL statement

Which is perfectly fine. DBAL prevents a knowingly invalid operation on the underlying driver.

/cc @beberlei

morozov:
@eisberg please rebase your patch onto `2.10`. It needs to go there since it's a bugfix. Later, the entire `2.10` branch will be merged into `master`.

morozov (approved):

morozov:
Thank you @eisberg!

Issue #3740, opened 2019-11-20
QueryBuilder binds type to a wrong param when using positional parameters


vdmorozov:
### Bug Report

QueryBuilder tries to bind a type to a wrong param if you use positional parameters, and the type of the first one is not explicit.

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.0

#### Summary

The problem is that the numerical parameters in the QueryBuilder API start with the needle `0`, not with `1` as in the PDO API. That's why we need to use offset in `Doctrine\DBAL\Connection::_bindTypedValues`:
```
if (is_int(key($params))) {
    // Positional parameters
    $typeOffset = array_key_exists(0, $types) ? -1 : 0;
    ...
```

But if you don't specify a type in the first call of `setParameter` method, `$types` doesn't contain key `0`, while you still need to use offset `-1`.

Checking for a `0` key in the `$params` instead of `$types` may be a solution.

#### Current behaviour

Let's say you have a code looking like this:
```
$connection->createQueryBuilder()
    ->insert('contacts')
    ->values(
        [
            'name' => '?',
            'birthday' => '?',
        ]
    )
    ->setParameter(0, (string)$name)
    ->setParameter(1, $birthday, Types::DATE_MUTABLE)
    ->execute();
```

In this example, the `DATE_MUTABLE` type will be bound to the `name` value causing `ConversionException`.

#### Expected behavior

The type `DATE_MUTABLE` should be bound to the second param `birthday`.

SenseException:
This is surely some strange behaviour you found there. Can you please provide a pull request with a failing test that helps to implement a fix?

vdmorozov:
Sure, here it is:
https://github.com/doctrine/dbal/pull/3751

morozov:
Closing as the test in https://github.com/doctrine/dbal/pull/3751 doesn't perform any assertions and works as expected otherwise.

PR #3751, opened 2019-11-26
Test case and fix for issue #3740

vdmorozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3740

#### Summary

Not sure if I've put this test case in a proper place, but I hope it's good enough to illustrate the bug.

A fix is to check if we need to use an offset for types by looking at the first key of params array. It's the only way to know what kind of positioning is used - starting from 0 or 1.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1420,7 +1420,7 @@ class Connection implements DriverConnection
         // Check whether parameters are positional or named. Mixing is not allowed, just like in PDO.
         if (is_int(key($params))) {
             // Positional parameters
-            $typeOffset = array_key_exists(0, $types) ? -1 : 0;
+            $typeOffset = array_key_exists(0, $params) ? -1 : 0;
             $bindIndex  = 1;
             foreach ($params as $value) {
                 $typeIndex = $bindIndex + $typeOffset;
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1352,7 +1352,7 @@ class Connection implements DriverConnection
         // Check whether parameters are positional or named. Mixing is not allowed, just like in PDO.
         if (is_int(key($params))) {
             // Positional parameters
-            $typeOffset = array_key_exists(0, $types) ? -1 : 0;
+            $typeOffset = array_key_exists(0, $params) ? -1 : 0;
             $bindIndex  = 1;
             foreach ($params as $value) {
                 $typeIndex = $bindIndex + $typeOffset;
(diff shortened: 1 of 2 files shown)


vdmorozov:
Looks like new test fails to create table with some databases, but I don't really know how to fix or avoid it.

morozov:
> but I don't really know how to fix or avoid it.

You're using MySQL syntax to create a DB on all platforms which doesn't work. You should use the schema manager and a table abstraction listed. For example:

https://github.com/doctrine/dbal/blob/e1accb21098ef1e70ccae3f8105dc76951b828c7/tests/Doctrine/Tests/DBAL/Functional/ExceptionTest.php#L47-L51

Additionally, your test passed on MySQL and MariaDB which means that the issue was not reproduced.

vdmorozov:
Ok, I'll avoid queries when creating a table. But I'm not sure if it's possible to implement the case itself like this.

>Additionally your test passed on MySQL and MariaDB which means that the issue was not reproduced.

The test passed because the fix is in the same PR. When I checked it locally, it failed perfectly. I can remove the fix from here and create a separate request for it if necessary.

morozov:
> The test passed because the fix is in the same PR. When I checked it locally, it failed perfectly. I can remove the fix from here and create a separate request for it if necessary.

My fault, sorry. I didn't notice there was a fix.

morozov:
> Ok, I'll avoid queries when creating a table.

You can avoid creating a table at all using `AbstractPlatform::getDummySelectSQL('?, ?')`.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1352,7 +1352,7 @@ private function _bindTypedValues(DriverStatement $stmt, array $params, array $t
         // Check whether parameters are positional or named. Mixing is not allowed, just like in PDO.
         if (is_int(key($params))) {
             // Positional parameters
-            $typeOffset = array_key_exists(0, $types) ? -1 : 0;
+            $typeOffset = array_key_exists(0, $params) ? -1 : 0;
```
morozov: Although the fix looks correct, I don't exactly understand the purpose of this offset. Is it needed in order to allow binding numeric parameters as a 0-based array and as 1-based array while having types to be always expressed as a 0-based array? As a feature, it doesn't make much sense. Shouldn't the type indices always correspond to the parameter indices? Otherwise, parameter indices are ignored but the first one isn't, while type indices are all taken into account.

@doctrine/doctrinecore does anyone know/remember the rationale behind this API? The documentation doesn't say anything about parameter and type indices.

vdmorozov:
Just found the same bug in duplicated code in `Connection::resolveParams`

vdmorozov:
Now the test fails on some databases with errors like
>PDOException: SQLSTATE[42P18]: Indeterminate datatype: 7 ERROR:  could not determine data type of parameter $1

or
>db2_bind_param(): Describe Param Failed

Maybe it's related to the use of dummy select. It's better to return to the creation of a well-described test table, I suppose.

PR #3756, opened 2019-11-28
Dropped SQL Server 2008 support

### LINKED ISSUE #3755: Drop support for SQL Server 2008
According to the [website](https://www.microsoft.com/en-us/sql-server/sql-server-2008), the extended support for SQL Server 2008 ended on July 9, 2019.

|    Q        |   A
|------------ | ------
| New Feature | no
| RFC         | no
| BC Break    | yes

Besides being no longer supported, the SQL Server 2008 builds take an order of magnitude more time on AppVeyor than other SQL Server builds. [For instance](https://ci.appveyor.com/project/doctrine/dbal/builds/29172767):

image

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

Fixes #3755.

Additionally, some changes in `.appveyor.yml` were needed in order to work around the build failure. The reason is:
1. Due to https://github.com/chocolatey/choco/issues/1843, `choco search` only returns PHP 7.4.0 as of today:
   ```
   PS C:\Users\Sergei Morozov> choco search php --exact --all-versions
   Chocolatey v0.10.15
   php 7.4.0 [Approved]
   1 packages found.
   ```
2. Combined with `| select-string -pattern 7.3`, it yields an empty list and [fails the build](https://ci.appveyor.com/project/doctrine/dbal/builds/29173708/job/7ci3uidwp28ri59p).
3. Prior to this change, the builds would pass because all runtime dependencies are cached based on the content of `.appveyor.yml`.
4. We cannot switch to PHP 7.4 on AppVeyor because the corresponding `sqlsrv` and `pdo_sqlsrv` binaries are not yet available.


SenseException (approved):

SenseException:
Are we going to create a new PR for having `sqlsrv` and `pdo_sqlsrv` with PHP 7.4 when it is available? Or do you want to try to find another solution?

morozov:
> Are we going to create a new PR for having `sqlsrv` and `pdo_sqlsrv` with PHP 7.4 when it is available? Or do you want to try to find another solution?

Apparently, the [`5.7.1preview` builds](https://windows.php.net/downloads/pecl/releases/pdo_sqlsrv/5.7.1preview/) are compiled for PHP 7.4 but I'd rather solve one problem at a time.

beberlei:
@morozov this needs to get into 2.10 as well as appveyor is mass failing there as well

PR #3368, opened 2018-11-30
[WIP] [3.0] Extract getSequenceNumber() from lastInsertId()

BenMorel:
***Work In Progress**: this requires #3335 and #3367 to be merged first. Their commits are included here; this PR actually starts at commit 69b6f3d.*

*I'm opening this PR early to be able to run CI tests on all platforms, and will rebase it later.*

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | N/A

#### Summary

While talking about `Driver\Connection::lastInsertId()`, @morozov [suggested the idea](https://github.com/doctrine/dbal/pull/3335#pullrequestreview-178029897) that:

> (...) the calls with and without the $name are targeted at the same method but have totally different semantics (named sequences and identity columns). This confusing API was originally borrowed from PDO but we don't have to follow it strictly.

#### Proposed changes

This PR splits the existing `lastInsertId(?string $name = null) : string` method into 2 methods:

- `lastInsertId() : string`
- `getSequenceNumber(string $name) : string`

Naming is a first draft, suggestions welcome (`getLastInsertId()` for consistency?)

[Tests have been extended](https://github.com/doctrine/dbal/blob/94d0416669aff460bd7c6ae54a83d4ddfae49d38/tests/Doctrine/Tests/DBAL/Functional/WriteTest.php#L149-L211) to cover the following scenarios:

- `lastInsertId()` on supported platforms
- `lastInsertId()` on unsupported platforms (exception)
- `lastInsertId()` on all platforms, on a fresh connection, i.e. with no ID available (exception)
- `getSequenceNumber()` with a known sequence name on supported platforms
- `getSequenceNumber()` with an unknown sequence name on all platforms (exception)

PR #3712, opened 2019-11-03
Merge all the work done so far for the 3.0.0 release into master

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

PR #3759, opened 2019-12-01
Extract getSequenceNumber() from lastInsertId()

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | N/A

#### Summary

This PR supersedes #3368. From the original issue:

> **Proposed changes**
> 
>This PR splits the existing `lastInsertId(?string $name = null) : string` method into 2 methods:
> 
> - `lastInsertId() : string`
> - `getSequenceNumber(string $name) : string`
> 
> Naming is a first draft, suggestions welcome (`getLastInsertId()` for consistency?)
> 
> Tests have been extended to cover the following scenarios:
> 
> - `lastInsertId()` on supported platforms
> - `lastInsertId()` on unsupported platforms (exception)
> - `lastInsertId()` on all platforms, on a fresh connection, i.e. with no ID available (exception)
> - `getSequenceNumber()` with a known sequence name on supported platforms
> - `getSequenceNumber()` with an unknown sequence name on all platforms (exception)

This PR has been rebased onto `master`, and only covers the split of `lastInsertId()` into `lastInsertId()` and `getSequenceNumber()`, without the originally proposed changes to the driver interfaces & exceptions.


lib/Doctrine/DBAL/Driver/PDOException.php
```diff
@@ -9,10 +9,7 @@
  */
 class PDOException extends AbstractDriverException
 {
-    /**
-     * @param \PDOException $exception The PDO exception to wrap.
-     */
-    public function __construct(\PDOException $exception)
+    public static function fromNativePDOException(\PDOException $exception)
```
BenMorel: Note: I had to make this a static method to be able to `throw new PDOException('message')` when I do not have a native `PDOException` object to pass. See for example [this line](https://github.com/doctrine/dbal/pull/3759/files#diff-e5f2acb790dca47e2826055870cfa283R104).

greg0ire:
IMO this should be contributed to the stable branch, without a BC-break. Calling the method with an argument should be deprecated in favor of calling the new method.

lib/Doctrine/DBAL/Driver/Connection.php
```diff
@@ -39,9 +39,40 @@ public function quote(string $input) : string;
...
+     * - the sequence does not exist.
+     *
+     * @param string $name The sequence name.
+     *
+     * @return string The current sequence value.
+     *
+     * @throws DriverException If an error occurs.
      */
-    public function lastInsertId(?string $name = null) : string;
+    public function getSequenceNumber(string $name) : string;
```
greg0ire: Could be added on the stable branch as part of a new interface. If that new interface is not implemented, trigger a deprecation so that we can merge both interfaces in the next major release.

BenMorel:
@morozov All tests now pass. Waiting for your review, in particular whether we should consider renaming `lastInsertId()` to `getLastInsertId()` for consistency with `getSequenceNumber()`.

@greg0ire I can understand the motivation behind making this feature available to the `2.x` series, however I cannot myself spend more time to make it happen. If this PR is merged, if should be pretty trivial for someone else to take over and port it to `2.x`.

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
```diff
@@ -135,20 +138,46 @@ public function testUpdate() : void
 
     public function testLastInsertId() : void
     {
-        if (! $this->connection->getDatabasePlatform()->prefersIdentityColumns()) {
```
morozov: This `prefersIdentityColumns()` method is indeed confusing and it's not used anywhere except for tests. We should probably get rid of it (out of scope).

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
```diff
@@ -135,20 +138,46 @@ public function testUpdate() : void
...
-    public function testLastInsertIdSequence() : void
+    public function testLastInsertIdNotSupported() : void
     {
-        if (! $this->connection->getDatabasePlatform()->supportsSequences()) {
-            $this->markTestSkipped('Test only works on platforms with sequences.');
+        if ($this->connection->getDatabasePlatform()->supportsIdentityColumns()) {
+            $this->markTestSkipped('This test targets platforms that don\'t support identity columns.');
+        }
+
+        self::assertEquals(1, $this->connection->insert('write_table', ['test_int' => 2, 'test_string' => 'bar']));
```
morozov: Isn't this redundant?

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
```diff
@@ -255,6 +280,13 @@ public function testEmptyIdentityInsert() : void
             );
         }
 
+        if ($platform instanceof OraclePlatform && $this->connection->getWrappedConnection() instanceof PDOConnection) {
+            $this->markTestSkipped(
+                'Oracle supports emulated identity columns through sequences, ' .
```
morozov: It doesn't look like that. It throws an exception if `lastInsertId()` is called without a name. Neither of Oracle drivers support this API. It should be probably detected based on `AbstractOracleDriver`.

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
```diff
@@ -276,11 +308,15 @@ public function testEmptyIdentityInsert() : void
 
         $this->connection->exec($sql);
 
-        $firstId = $this->lastInsertId($seqName);
+        $firstId = $seqName !== null
```
morozov: No conditionals in tests please.

lib/Doctrine/DBAL/Driver/PDOException.php
```diff
@@ -9,10 +9,7 @@
  */
 class PDOException extends AbstractDriverException
 {
-    /**
-     * @param \PDOException $exception The PDO exception to wrap.
-     */
-    public function __construct(\PDOException $exception)
+    public static function fromNativePDOException(\PDOException $exception) : PDOException
```
morozov: This is something I was also trying to do as part of my other patch which is no longer relevant. I'll review this later once I check my corresponding notes.

morozov (changes requested):
@BenMorel, thank you for continuing to work on this. Please see the comments.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -993,20 +993,33 @@ public function getTransactionNestingLevel() : int
...
+    }
+
+    /**
+     * Returns the current sequence value for the given sequence name.
+     *
+     * If the underlying driver does not support sequences, an exception is thrown.
+     *
+     * @param string $name The sequence name.
+     *
+     * @return string The current sequence value.
```
morozov: I wonder why we should cast the sequence value to a string and can't return it as is (untyped in the API) given that it's always a number.

lib/Doctrine/DBAL/Driver/Connection.php
```diff
@@ -39,9 +39,40 @@ public function quote(string $input) : string;
     public function exec(string $statement) : int;
 
     /**
-     * Returns the ID of the last inserted row or sequence value.
+     * Returns the ID of the last inserted row.
+     *
+     * This method returns a string representing the value of the auto-increment field from the last row inserted into
```
morozov: Same as for sequences, why does it have to be a string?

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
```diff
@@ -116,9 +116,23 @@ public function exec(string $statement) : int
...
      * {@inheritdoc}
      */
-    public function lastInsertId(?string $name = null) : string
+    public function lastInsertId() : string
     {
-        return db2_last_insert_id($this->conn);
+        $lastInsertId = db2_last_insert_id($this->conn);
+
+        if ($lastInsertId === '') {
+            throw new DB2Exception('No identity value was generated by the last statement.');
```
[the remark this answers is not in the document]
BenMorel: Which is exactly why the original PR, #3368, was based on #3367, **which proposed to rework driver exceptions *before* implementing this separation**: a single `DriverException` class should be enough, the driver-specific classes serve no purpose.

I've now worked all the way back to make this PR compatible with the current state of exceptions, because of your [comment ](https://github.com/doctrine/dbal/pull/3367#issuecomment-443899736)back then:

> @Majkl578 drives the rework of exceptions. As far as I understand, design-wise it's not what he had in mind.

lib/Doctrine/DBAL/Driver/OCI8/OCI8Connection.php
```diff
@@ -134,12 +134,16 @@ public function exec(string $statement) : int
     /**
      * {@inheritdoc}
      */
-    public function lastInsertId(?string $name = null) : string
+    public function lastInsertId() : string
     {
-        if ($name === null) {
-            throw new OCI8Exception('The driver does not support identity columns.');
-        }
+        throw new OCI8Exception('The driver does not support identity columns.');
```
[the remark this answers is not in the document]
BenMorel: Please see https://github.com/doctrine/dbal/pull/3759#discussion_r352711671. There should be a single `DriverException`, with factory methods for these use cases. But my previous proposal has been dismissed because @Majkl578 was already working on exceptions.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
```diff
@@ -116,9 +116,23 @@ public function exec(string $statement) : int
...
+
+        return $lastInsertId;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSequenceNumber(string $name) : string
+    {
+        throw new DB2Exception('Sequences on IBM DB2 are not currently supported.');
```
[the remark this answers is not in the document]
BenMorel: This does not really imply that DB2 does not support them, but I understand that this can be a source of confusion. Do you prefer:

> Sequences on IBM DB2 are not currently supported **by the DBAL**.

lib/Doctrine/DBAL/Driver/PDOConnection.php
```diff
@@ -92,17 +92,43 @@ public function quote(string $input) : string
...
+    public function getSequenceNumber(string $name) : string
+    {
+        try {
+            $sequenceNumber = (string) $this->connection->lastInsertId($name);
         } catch (\PDOException $exception) {
-            throw new PDOException($exception);
+            throw PDOException::fromNativePDOException($exception);
+        }
+
+        if (! $this->isValidLastInsertIdOrSequenceNumber($sequenceNumber)) {
```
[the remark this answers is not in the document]
BenMorel: It's up to you really. If the PDO driver already has proper support for sequences on the given platform, I'm not sure whether there's a value in reinventing the wheel?

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
BenMorel: I trust you on this one ;)

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
BenMorel: I would keep it, as `lastInsertId()` may fail for 2 reasons:

- identity columns not supported
- no previous statement has inserted a row

By inserting a row first in the same test, we can guarantee that we're actually testing the first bullet point.

lib/Doctrine/DBAL/Driver/PDOException.php
BenMorel: There is so much room for improvement in so many places, that it's very difficult to organize PRs that don't have dependencies, (when not circular dependencies!) on other PRs, sometimes conflicting ones from other people. Hence my previous comments.

lib/Doctrine/DBAL/Connection.php
BenMorel: I guess that this boils down to 32-bit PHP being still supported, when identity columns / sequences may return 64-bit integers, which can only be accurately represented as strings.

lib/Doctrine/DBAL/Driver/Connection.php
BenMorel: [Same comment](https://github.com/doctrine/dbal/pull/3759#discussion_r352723604) as above.

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
BenMorel: You may be wrong here. [The PDO OCI test failed](https://app.continuousphp.com/git-hub/doctrine/dbal/build/2454fff4-ac55-4feb-8238-bf2b19615841/logs/50ff3103-64d5-4395-a9c8-7194766b2d60), even though the stack trace pointed to:

https://github.com/doctrine/dbal/blob/ea93b5542d15b14db596a7fc758ffe4fef04c3bb/tests/Doctrine/Tests/DBAL/Functional/WriteTest.php#L189

Then:

https://github.com/doctrine/dbal/blob/ea93b5542d15b14db596a7fc758ffe4fef04c3bb/lib/Doctrine/DBAL/Driver/PDOConnection.php#L116

Where you can see that `lastInsertId()` is called with a name.

My suspicion is in the exception message: Oracle supports sequences, but PDO OCI doesn't support them through `lastInsertId()`.

morozov: In any event, the error message "Oracle supports emulated identity columns through sequences" is misleading. The last inserted ID is scoped to a connection and has one value for the entire connection while sequences are persisted in the database and there may be more than one of them. It's hard to imagine how a driver could emulate identity columns through sequences.

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
BenMorel: Fixed in 334cf158f5875701df77f8778faaa40fe73ba2c4.

lib/Doctrine/DBAL/Driver/Connection.php
BenMorel: Not sure what you're expecting from me exactly here?

lib/Doctrine/DBAL/Connection.php
morozov: Correct. And this is up to the underlying DB driver which I don't think we should interfere with.

lib/Doctrine/DBAL/Driver/Connection.php
morozov: I'm expecting the `getSequenceNumber()` to exist in the wrapper statement but not in the driver statements.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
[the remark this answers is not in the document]
morozov: This work has been recently merged as https://github.com/doctrine/dbal/pull/3712.

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
BenMorel: I just followed the [already established convention](https://github.com/doctrine/dbal/blob/299ef4f06645b747fbef7876f87516012eb990ba/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php#L2689). What do you suggest exactly?

morozov: I don't understand what this method means. Besides the boolean value returned by these methods, I don't see any emulation happening on the platform level. This method is not used anywhere except in tests.

I suggest changing the condition to:
```php
if ($driver instanceof AbstractOracleDriver) {
    $this->markTestSkipped('The driver does not support lastInsertId()');
}
```
There's nothing about `lastInsertId()` on the platform level, IMO.

lib/Doctrine/DBAL/Connection.php
BenMorel: So you don't want to use strong typing here? (unless you're willing to target PHP 8.0, in which case you will be able to use [union types](https://wiki.php.net/rfc/union_types_v2) `int|string`).

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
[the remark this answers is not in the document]
BenMorel: Good! Now I would question why `DriverException` is an interface, and why we would need different implementations of it. Not all drivers have their own `DriverException` implementation, and there is, AFAICS, never a need to catch a specific driver exception.

Could you please read the intro of #3367 again, and tell me if you would like to proceed with this change first? I can make it compatible with the current `master` again, but it's quite a lot of work so I need to make sure that this is something you want, first.

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
BenMorel: TBH I only tried to accommodate existing code as much as possible. After more review though, you're right that `usesSequenceEmulatedIdentityColumns()` is only used in tests. Do you want me to attempt to remove it altogether, in a separate PR?

Then `testEmptyIdentityInsertEmulated()` would be gone here.

lib/Doctrine/DBAL/Connection.php
morozov: > So you don't want to use strong typing here?

No, I don't. It doesn't seem to make sense at the abstraction level.

lib/Doctrine/DBAL/Driver/PDOException.php
morozov: I agree. I'll make sure we're not losing control over the scope of the patch.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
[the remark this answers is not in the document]
morozov: I'm not really up to speed on how exactly the exceptions are currently designed although I agree there are some flaws. E.g. the very fact of the existence of a class and an interface named `DriverException` makes me nervous.

> Good! Now I would question why `DriverException` is an interface, and why we would need different implementations of it.

This is true from one PoV: definitely, different implementations for MySQLi, PDO, etc. don't make much sense IMO. On the other hand, different implementations representing different domain-level seem to make sense to me. E.g. `class LastInsertIdNotSupported implements DriverException`.

> Could you please read the intro of #3367 again, and tell me if you would like to proceed with this change first? I can make it compatible with the current `master` again, but it's quite a lot of work so I need to make sure that this is something you want, first.

This is a circular dependency we've been trying to avoid :). For me, it would be easier if you rebased if first. Currently, I don't have enough time to be able to deeply familiarize myself with the approach based on the PR description (thanks for all the details) w/o being able to look at the diff.

Answering your question, I'd like to finish this patch first. If it makes more sense to leave these new exceptions as is, I'd rather leave them as is than switch context to the other ticket.

lib/Doctrine/DBAL/Connection.php
BenMorel: Can't say I'm convinced on this one, but anyway: fixed in 18d245b7b2cdebbd0632175bb9eaa2117b9d173d.

lib/Doctrine/DBAL/Driver/Connection.php
BenMorel: Sorry, you lost me. Do you want to remove `getSequenceNumber()` from the `Driver\Connection` interface, and move it to the `AbstractPlatform`, to be consumed by the `DBAL\Connection` only?

If so, I feel like this is out of scope for this PR. Maybe we should first work on merging it as is (split in 2 methods), then work on moving `getSequenceNumber()` from `Driver\Connection`  to `AbstractPlatform` in another PR.

(Plus, I don't know if we currently have the SQL to get the sequence number on all platforms, some of them may rely on PDO's ability to return a sequence number with `lastInsertId($name)`, which we'll lose if we proceed with such a change.)

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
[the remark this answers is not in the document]
BenMorel: Fixed in 3d93734e77b49b01f847bfaf5a0ea3b1159178d5.

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
morozov: I'd rather have it removed in a separate PR. We can merge it first and then rebase this one.

lib/Doctrine/DBAL/Driver/Connection.php
morozov: > Sorry, you lost me. Do you want to remove `getSequenceNumber()` from the `Driver\Connection` interface, and move it to the `AbstractPlatform`, to be consumed by the `DBAL\Connection` only?

Yes. Do you think this is wrong? I'm really open to a discussion.

> If so, I feel like this is out of scope for this PR

It's not because otherwise, we'll be introducing a driver connection method that will have to be removed later.

> (Plus, I don't know if we currently have the SQL to get the sequence number on all platforms, some of them may rely on PDO's ability to return a sequence number with `lastInsertId($name)`, which we'll lose if we proceed with such a change.)

It's probably only PostgreSQL that supports sequences but we only support it via `pdo_pgsql`. Implementing `getSequenceNumber()` for PostgreSQL should be as trivial as for any other platform.

lib/Doctrine/DBAL/Driver/PDOConnection.php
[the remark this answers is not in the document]
morozov: Because unlike PDO, the concerns of driver and platform are separate in DBAL. If PDO exposed the functionality of generating SQL that we could use on the platform level, we would reuse it. Otherwise, we'll have to implement our own API ourselves.

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
morozov: That's why I'd rather have different exception classes and expect a specific class. If this is out of scope, I'd rather expect a specific error message on the exception than perform an unrelated operation that still doesn't guarantee that the exception is thrown by the code that is intended to be tested.

lib/Doctrine/DBAL/Driver/Connection.php
BenMorel: > Yes. Do you think this is wrong? I'm really open to a discussion.

I don't have a strong opinion about this. What you're saying makes sense, but at the same time I'm personally OK with the way it is now, especially the convenience of using `lastInsertId($name)` when available, even though it's trivial to re-implement it as you said.

> It's not because otherwise, we'll be introducing a driver connection method that will have to be removed later.

I'd say that being hesitant (at least, be open to discussion) on the first point would be in favour of keeping this PR the way it is, and open a new discussion later to see if it's worth moving to AbstractPlatform. Otherwise, we're (I am) venturing into unknown territory and making the current PR less likely to be merged.

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
BenMorel: Makes sense. This takes us back to my suggested modification to driver exceptions, but with different subclassing (I originally intended to have a single `DriverException` class).

Exceptions would not be based on the driver (`DB2Exception`) anymore, but on the intent (`IdentityColumnNotSupportedException`).

This would definitely be clearer. But apart from tests, does this have a real-life use case?

lib/Doctrine/DBAL/Driver/Connection.php
BenMorel: > Yes. Do you think this is wrong?

By the way, off-topic: something that's currently *really* wrong, IMHO, is to have `Connection` implement `Driver\Connection`. It should compose, not implement.

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
BenMorel: No sooner said than done: #3769

morozov: Thank you for filing the ticket. I'm going to take some time and learn the codebase better. I'll get back to you as soon as I have a solid understanding of how exactly I'd like us to proceed.

morozov: > Oracle supports emulated identity columns through sequences

Okay. According to [this respectful source](https://en.wikipedia.org/wiki/Identity_column) an identity column is the one that is made up of values generated by the database. So [Oracle does emulate identity columns through sequences](https://github.com/doctrine/dbal/blob/436f2f6c943937abe8544505630e76be4c468ae1/lib/Doctrine/DBAL/Platforms/OraclePlatform.php#L395) which is a platform feature in the DBAL terminology. 

Although I still fail to see how this is relevant to the `lastInsertId()` API which is a driver feature. The exception message is still misleading to me.

lib/Doctrine/DBAL/Driver/Connection.php
morozov: I agree that those are two different interfaces. Not necessarily that the one should extend the other.

lib/Doctrine/DBAL/Connection.php
morozov: BTW, found some more arguments on the topic from a year ago: https://github.com/doctrine/dbal/pull/3074#issuecomment-382549778.

morozov: BTW, found some more arguments on the topic from a year ago: https://github.com/doctrine/dbal/pull/3074#issuecomment-382549778.

lib/Doctrine/DBAL/Driver/PDOException.php
morozov: > I agree. I'll make sure we're not losing control over the scope of the patch.

Still considering the above, I don't think that this is an acceptable approach. This is yet another case that shows that the current driver exception hierarchy is not maintainable.

I cannot let introduce this method because having it will make master non-releasable until it's removed. Or otherwise, we'd have to maintain this method until the next major release.

With that said, given all other revealed driver exception issues, why don't we focus on that and get back to this project later? Otherwise, it turns into a rabbit hole and requires significant mental effort to keep all considerations in mind.

morozov: > This is something I was also trying to do as part of my other patch which is no longer relevant.

I was unable to find it but it was similar to what you're trying to do.

tests/Doctrine/Tests/DBAL/Functional/WriteTest.php
BenMorel: What do you suggest?

morozov: The same as in https://github.com/doctrine/dbal/pull/3759#discussion_r352818386.

BenMorel: Done in 7ec283d.

lib/Doctrine/DBAL/Driver/PDOException.php
BenMorel: Oh, I wasn't imagining releasing 3.0 *before* reworking the driver exceptions. But we can work on them before if you wish (that was actually the plan in #3368).

If you're OK with this, I'll start with making `DriverException` a single concrete class, and drop the current subclasses / implementations. This will basically be a rebased version of #3367, which is totally out of date now.

We'll then be able to proceed to subclassing `DriverException` for specific *exception reasons*, not *drivers*, later, if required.

morozov: > `DriverException` is already a class and not an interface anymore.

I must have missed it. When did it happen? Again, we have two symbols with the same name `DriverException`:
```sh
$ git grep -P "(class|interface) DriverException"
lib/Doctrine/DBAL/Driver/DriverException.php:interface DriverException extends Throwable
lib/Doctrine/DBAL/Exception/DriverException.php:class DriverException extends DBALException implements DriverExceptionInterface
```

> I will just try to remove driver-specific exceptions and move factory methods to driver connections, which was the original aim. Is that potentially OK with you?

That sounds about right. Ideally, we don't want to have any factory methods but named exceptions instead. E.g. instead of `OCI8Exception::fromErrorInfo()` we may want to have `OCI8Error::new()`. Let's start with a prototype and discuss it on the code level. As I said earlier, I don't have the full picture in my head.

BenMorel: How is `OCIError` different from `OCI8Exception`? My idea is to *remove* `OCI8Exception`, not rename it! Thus the need to move the factory method elsewhere; and I think it belongs in the connection or the statement that throws it.

morozov: The idea is that `OCI8Error` would represent only an error that happened because of `oci_error()`. Otherwise, how would you translate it to an exception? No other errors that have happened in the DBAL's `OCI8` driver will have to be represented with this class. Probably, it could use a better name if it turns out to be confusing.

In any event, I'm not suggesting anything specific, go ahead and draft your vision, and we can discuss it.

BenMorel: What I mean is that *when* we need factory methods for driver-specific exceptions, we should avoid creating subclasses specific to that driver just for the purpose of putting factory methods there. Having an `OC8IException`|`Error` class would imply that you might need to catch it at some point, when in reality you're only interested in catching the base `DriverException`. Therefore I'd move the factory methods elsewhere.

I'll draft a PR soon.

morozov: > Having an `OC8IException`|`Error` class would imply that you might need to catch it at some point.

Why would it imply that? It's just an implementation of the interface that the throwing method would declare in its `@throws` annotation, and this is what you need to be catching.

BenMorel: Because drivers implement interfaces, such as `Driver\Connection`, that explicitly declare which exceptions may be thrown.

`Connection::commit()`, for example, declares throwing a `DriverException`. As a consumer of this interface, you should only be interested in catching this exception, you're not supposed to dig into the internals to find out that there is a (non portable) `OCI8Exception` lying somewhere. Sure, it doesn't *hurt* to have this class internally, as it does extend `DriverException`, but this leaves the door open to someone writing a `catch (OCI8Exception)` in their code, which would be a big code smell, as this would defeat portability, which is the whole point of the DBAL.

Therefore **I'd better not have any driver-specific exception**, and only subclass `DriverException` if there is a legitimate, *cross-driver* use case for it.

If you look at current driver-specific exceptions, you'll notice that **they're only there to provide driver-specific factory methods**. This is not a big enough justification to subclass `DriverException`; the factory methods may live elsewhere.

BenMorel:
@morozov Is there anything specific you want to be changed to finish this PR, or do you just need more time for review? Should I work on removing the driver-specific exceptions in another PR?

morozov:
> Should I work on removing the driver-specific exceptions in another PR?

Yes. See https://github.com/doctrine/dbal/pull/3759#discussion_r354589710.

morozov:
The development of this change should be now unblocked by the recent changes in exception handling in `3.0.x` (see https://github.com/doctrine/dbal/pull/3367#issuecomment-654960605).

lib/Doctrine/DBAL/Driver/Connection.php
morozov: For the reference, this has been addressed in https://github.com/doctrine/dbal/pull/4159.

morozov:
Since #3367 is addressed, @BenMorel, could you try to rebase this so we could reassess where we are? From the static analysis standpoint, this is one of the most annoying APIs in the library (not counting the ones that have been already deprecated and removed).

BenMorel:
@morozov I opened a new PR #4324, please stand by while I refine it, I'll let you know when it's ready for review!
Closing this one in the meantime.

PR #3761, opened 2019-12-01
[GH-1204] Add Platform flag for FK constraints through emulation

beberlei:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1204 doctrine/orm#7841

#### Summary

`SchemaManager::listTableDetails` checks for foreign key support, which SQLite emulates but still responds false to. This leads to bugs in ORMs SchemaTool where it always finds foreign keys are added and thus it re-creates the tables.

We can't change `supportsForeignKeyConstraints()` to true, because `getForeignKeyConstraintCreateSql()` throws an exception that its not supported.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -3148,6 +3148,16 @@ abstract class AbstractPlatform
     }
 
     /**
+     * Wheather the platform can support foreign key constraints through emulation using triggers.
+     *
+     * @return bool
+     */
+    public function supportsForeignKeyConstraintsThroughEmulation() : bool
+    {
+        return false;
+    }
+
+    /**
      * Whether this platform supports onUpdate in foreign key constraints.
      *
      * @return bool
--- a/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
@@ -752,6 +752,14 @@ class SqlitePlatform extends AbstractPlatform
     /**
      * {@inheritDoc}
      */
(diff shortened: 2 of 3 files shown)


lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3147,6 +3147,16 @@ public function supportsForeignKeyConstraints()
         return true;
     }
 
+    /**
+     * Wheather the platform can support foreign key constraints through emulation using triggers.
```
greg0ire: ⛅
```suggestion
     * Whether the platform can support foreign key constraints through emulation using triggers.
```

beberlei:
This doesn't fix the problem though, because SQLite requires the foreign keys to be created in the `CREATE TABLE` statement, but we don't have a dependency resolving, so when we add it there the queries will fail if the constraint references a table that doesn't exist yet.

This is solved in `Comparator` by instructing create table to only create indexes and then creating the foreign keys in seperate SQL statements. This is supported by all other DBs, but sqlite requires them in CREATE TABLE already. This leads to the problem that foreign keys are not declared during the creation of a table in ORM Schema Tool, only subsequently through updates.

beberlei:
OK I realized this approach is wrong.

1. SQLite *does* support foriegn keys, its not an emulation.
2. The problem is that foreign keys *must* be created during `CREATE TABLE` statements. This means there is no way to drop or alter them.
3. Comperator does not order tables based on their dependencies, meaning if we allow creating foreign keys in `CREATE TABLE`, we must have the order of statements right.
4. We need flags on the platform that indicate a complete recreate of the table is needed when something changes. Otherwise SchemaManager will trigger `getCreateForeignKeySQL` and others in SQLite platform that throws exception.

PR #3762, opened 2019-12-01
[GH-1204] Add full support for foreign key constraints for SQLite

beberlei:
Add full support for foreign key in SQLite.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1204 doctrine/orm#7841 doctrine/orm#7930

#### Summary

SQLite has partial foreign key support, which leads to bugs in ORM Schema Tool, because it cannot reliable make use of `AbstractPlatform::supportsForeignKeyConstraints()` and hence always re-creates the same table structure, because the DBAL Comparator it finds a diff in the schema that doesn't exist.

This PR does a few things:

- Sets `SqlitePlatform::supportsForeignKeyConstraints()` to true
- Introduces new flag `supportsCreateDropForeignKeyConstraints()` which is necessary to control in `SchemaDiff::toSql()` how to create the SQL.
- Copies the `CommitOrderCalculator` from ORM to DBAL because its needed in `Schema\Comparator` to calculate the order of new tables to create. The tests are also copied over, the idea is probably that ORM removes the commit order calculator in 2.8 and uses the one from DBAL instead.
- Throw `ForeignKeyConstraintViolationException` when `PRAGMA foreign_keys=On` set and violation occurs. Tests failed after switching supportsForeignKeyConstraints to true.

Even though its big, I would argue this is still considered a bugfix, because returning false from `supportsForeignKeyConstraints()` is actually not wrong, since the code supports foreig key constraints.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -3185,6 +3185,16 @@ abstract class AbstractPlatform
     }
 
     /**
+     * Whether foreign key constraints can be dropped.
+     *
+     * If false, then getDropForeignKeySQL() throws exception.
+     */
+    public function supportsCreateDropForeignKeyConstraints() : bool
+    {
+        return true;
+    }
+
+    /**
      * Whether this platform supports onUpdate in foreign key constraints.
      *
      * @return bool
--- a/lib/Doctrine/DBAL/Schema/SchemaDiff.php
+++ b/lib/Doctrine/DBAL/Schema/SchemaDiff.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\DBAL\Schema;
 
+use Doctrine\DBAL\Internal\DependencyOrderCalculator;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use function array_merge;
 
@@ -137,13 +138,16 @@ class SchemaDiff
         }
 
         $foreignKeySql = [];
-        foreach ($this->newTables as $table) {
-            $sql = array_merge(
-                $sql,
-                $platform->getCreateTableSQL($table, AbstractPlatform::CREATE_INDEXES)
-            );
+        $createFlags   = AbstractPlatform::CREATE_INDEXES;
 
-            if (! $platform->supportsForeignKeyConstraints()) {
+        if (! $platform->supportsCreateDropForeignKeyConstraints()) {
+            $createFlags |= AbstractPlatform::CREATE_FOREIGNKEYS;
+        }
+
+        foreach ($this->getNewTablesSortedByDependencies() as $table) {
+            $sql = array_merge($sql, $platform->getCreateTableSQL($table, $createFlags));
+
+            if (! $platform->supportsCreateDropForeignKeyConstraints()) {
                 continue;
             }
 
@@ -164,5 +168,38 @@ class SchemaDiff
         }
 
         return $sql;
+    }
+
+    /**
+     * Sorts tables by dependencies so that they are created in the right order.
+     *
+     * This is necessary when one table depends on another while creating foreign key
+     * constraints directly during CREATE TABLE.
+     *
+     * @return array<Table>
+     */
+    private function getNewTablesSortedByDependencies()
+    {
+        $calculator = new DependencyOrderCalculator();
+        $newTables  = [];
+
+        foreach ($this->newTables as $table) {
+            $newTables[$table->getName()] = true;
+            $calculator->addNode($table->getName(), $table);
+        }
+
+        foreach ($this->newTables as $table) {
+            foreach ($table->getForeignKeys() as $foreignKey) {
+                $foreignTableName = $foreignKey->getForeignTableName();
+
+                if (! isset($newTables[$foreignTableName])) {
+                    continue;
+                }
+
+                $calculator->addDependency($foreignTableName, $table->getName());
+            }
+        }
+
+        return $calculator->sort();
     }
 }
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -77,3 +77,9 @@ parameters:
         -
             message: '~^Access to undefined constant PDO::PGSQL_ATTR_DISABLE_PREPARES\.$~'
             path: %currentWorkingDirectory%/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
+
+        # False Positive
+        - '~Strict comparison using === between 1 and 2 will always evaluate to false~'
+
+        # Needs Generics
+        - '~Method Doctrine\\DBAL\\Schema\\SchemaDiff::getNewTablesSortedByDependencies\(\) should return array<Doctrine\\DBAL\\Schema\\Table> but returns array<object>.~'
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
@@ -14,6 +14,7 @@ use Doctrine\DBAL\Schema\ColumnDiff;
 use Doctrine\DBAL\Schema\Comparator;
 use Doctrine\DBAL\Schema\ForeignKeyConstraint;
 use Doctrine\DBAL\Schema\Index;
+use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Schema\SchemaDiff;
 use Doctrine\DBAL\Schema\Sequence;
 use Doctrine\DBAL\Schema\Table;
@@ -39,6 +40,7 @@ use function count;
 use function current;
 use function end;
 use function explode;
+use function implode;
 use function in_array;
 use function sprintf;
 use function str_replace;
@@ -1603,5 +1605,50 @@ abstract class SchemaManagerFunctionalTestCase extends DbalFunctionalTestCase
 
         $table = $this->schemaManager->listTableDetails('table_with_comment');
         self::assertSame('Foo with control characters \'\\', $table->getComment());
+    }
+
+    public function testSchemaDiffForeignKeys() : void
+    {
+        $schemaManager = $this->connection->getSchemaManager();
+        $platform      = $this->connection->getDatabasePlatform();
+
+        $table1 = new Table('child');
+        $table1->addColumn('id', 'integer', ['autoincrement' => true]);
+        $table1->addColumn('parent_id', 'integer');
+        $table1->setPrimaryKey(['id']);
+        $table1->addForeignKeyConstraint('parent', ['parent_id'], ['id']);
+
+        $table2 = new Table('parent');
+        $table2->addColumn('id', 'integer', ['autoincrement' => true]);
+        $table2->setPrimaryKey(['id']);
+
+        $diff = new SchemaDiff([$table1, $table2]);
+        $sqls = $diff->toSql($platform);
+
+        foreach ($sqls as $sql) {
+            $this->connection->exec($sql);
+        }
+
+        $schema = new Schema([
+            $schemaManager->listTableDetails('child'),
+            $schemaManager->listTableDetails('parent'),
+        ]);
+
+        $this->assertCount(1, $schema->getTable('child')->getForeignKeys());
+
+        $offlineSchema = new Schema([$table1, $table2]);
+
+        $diff = Comparator::compareSchemas($offlineSchema, $schema);
+
+        foreach ($diff->changedTables as $table) {
+            if (count($table->changedForeignKeys) <= 0 && count($table->addedForeignKeys) <= 0 && count($table->removedForeignKeys) <= 0) {
+                continue;
+            }
+
+            $this->fail(
+                'No changes on foreigh keys should be detected, but we have: ' .
+                implode(', ', $diff->toSql($platform))
+            );
+        }
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Schema/SchemaDiffTest.php
+++ b/tests/Doctrine/Tests/DBAL/Schema/SchemaDiffTest.php
@@ -48,6 +48,10 @@ class SchemaDiffTest extends TestCase
             ->method('getCreateSchemaSQL')
             ->with('foo_ns')
             ->will($this->returnValue('create_schema'));
+
+        $platform->method('supportsCreateDropForeignKeyConstraints')
+            ->will($this->returnValue(true));
+
         if ($unsafe) {
             $platform->expects($this->exactly(1))
                  ->method('getDropSequenceSql')
@@ -95,7 +99,7 @@ class SchemaDiffTest extends TestCase
         $platform->expects($this->exactly(1))
                 ->method('supportsSequences')
                 ->will($this->returnValue(true));
-        $platform->expects($this->exactly(2))
+        $platform->expects($this->any())
                 ->method('supportsForeignKeyConstraints')
                 ->will($this->returnValue(true));
 
--- a/lib/Doctrine/DBAL/Driver/AbstractSQLiteDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractSQLiteDriver.php
@@ -67,6 +67,10 @@ abstract class AbstractSQLiteDriver implements Driver, ExceptionConverterDriver
             return new Exception\ConnectionException($message, $exception);
         }
 
+        if (strpos($exception->getMessage(), 'FOREIGN KEY constraint failed') !== false) {
+            return new Exception\ForeignKeyConstraintViolationException($message, $exception);
+        }
+
         return new Exception\DriverException($message, $exception);
     }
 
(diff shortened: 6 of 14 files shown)


lib/Doctrine/DBAL/Internal/CommitOrderCalculator.php
```diff
@@ -0,0 +1,195 @@
...
+     * Adds a new node (vertex) to the graph, assigning its hash and value.
+     *
+     * @param string $hash
+     * @param object $node
+     *
+     * @return void
+     */
+    public function addNode($hash, $node)
+    {
+        $vertex = new \stdClass();
```
BackEndTea: I'm not too familiar with this project, but maybe its beneficial to introduce some value objects here, instead of using `\stdClass`. That means we can also drop the additional documentation on the properties, and it *should* be slightly more memory efficient.

beberlei: @BackEndTea great idea! I added a `CommitOrderNode` and `CommitOrderEdge` class.

lib/Doctrine/DBAL/Internal/CommitOrderCalculator.php
```diff
@@ -0,0 +1,164 @@
...
+    private $sortedNodeList = [];
+
+    /**
+     * Checks for node (vertex) existence in graph.
+     *
+     * @param string $hash
+     *
+     * @return bool
+     */
+    public function hasNode($hash)
```
morozov: This is not used and not tested. Could this be removed?

lib/Doctrine/DBAL/Internal/CommitOrderCalculator.php
```diff
@@ -0,0 +1,164 @@
...
+
+    /**
+     * Adds a new node (vertex) to the graph, assigning its hash and value.
+     *
+     * @param string $hash
+     * @param object $node
+     *
+     * @return void
+     */
+    public function addNode($hash, $node)
```
morozov: Should the parameter and return types be enforced by type declarations instead of annotations?

lib/Doctrine/DBAL/Internal/CommitOrderCalculator.php
```diff
@@ -0,0 +1,164 @@
...
+use function array_reverse;
+
+/**
+ * CommitOrderCalculator implements topological sorting, which is an ordering
+ * algorithm for directed graphs (DG) and/or directed acyclic graphs (DAG) by
+ * using a depth-first searching (DFS) to traverse the graph built in memory.
+ * This algorithm have a linear running time based on nodes (V) and dependency
+ * between the nodes (E), resulting in a computational complexity of O(V + E).
+ */
+class CommitOrderCalculator
```
morozov: Although this is used for commit order calculation, it's not specialized for that. Would it make sense to rename this class and the related ones in a way that doesn't imply that it's only for commits? E.g. `Graph`, `Node`, `Edge`?

lib/Doctrine/DBAL/Schema/SchemaDiff.php
```diff
@@ -165,4 +169,37 @@ protected function _toSql(AbstractPlatform $platform, $saveMode = false)
...
+
+        foreach ($this->newTables as $table) {
+            foreach ($table->getForeignKeys() as $foreignKey) {
+                $foreignTableName = $foreignKey->getForeignTableName();
+
+                if (! isset($newTables[$foreignTableName])) {
+                    continue;
+                }
+
+                $commitOrderCalculator->addDependency($foreignTableName, $table->getName(), 1);
```
morozov: Currently, this is the only used weight value. Would it make sense to omit it from the implementation or it would be needed for sorting internally anyways?

tests/Doctrine/Tests/DBAL/Internal/CommitOrderCalculatorTest.php
```diff
@@ -0,0 +1,104 @@
...
+        $this->calculator->addDependency($table1->getName(), $table2->getName(), 1);
+        $this->calculator->addDependency($table4->getName(), $table3->getName(), 1);
+        $this->calculator->addDependency($table1->getName(), $table4->getName(), 0);
+
+        $sorted = $this->calculator->sort();
+
+        // There is only multiple valid ordering for this constellation, but
+        // the class4, class1, class2 ordering is important to break the cycle
+        // on the nullable link.
+        $correctOrders = [
```
morozov: Does it mean that the order is not deterministic? Otherwise, we should either make a specific assertion or test all possible outcomes explicitly.

tests/Doctrine/Tests/DBAL/Internal/CommitOrderCalculatorTest.php
```diff
@@ -0,0 +1,104 @@
...
+            [$table4, $table1, $table2, $table3],
+            [$table4, $table1, $table3, $table2],
+            [$table4, $table3, $table1, $table2],
+        ];
+
+        // We want to perform a strict comparison of the array
+        $this->assertContains($sorted, $correctOrders, '');
```
morozov: Is there a need to pass the default value of the 3rd argument explicitly?

morozov (commented):
The code and tests look fine. Just a few questions.

lib/Doctrine/DBAL/Internal/CommitOrderCalculator.php
lcobucci: We should make it `final` too

tests/Doctrine/Tests/DBAL/Schema/SchemaDiffTest.php
```diff
@@ -48,6 +48,11 @@ private function createPlatform(bool $unsafe)
             ->method('getCreateSchemaSQL')
             ->with('foo_ns')
             ->will($this->returnValue('create_schema'));
+
+        $platform->expects($this->any())
+            ->method('supportsCreateDropForeignKeyConstraints')
+            ->will($this->returnValue(true));
```
lcobucci: We can omit `expects()` call when using `->expects($this->any()`.
Also, `->willReturnValue(true)` is much simpler too:

```php
$platform->method('supportsCreateDropForeignKeyConstraints')
         ->willReturnValue(true);
```

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3180,6 +3180,18 @@ public function supportsForeignKeyConstraints()
...
     }
 
+    /**
+     * Whether foreign key constraints can be dropped.
+     *
+     * If false, then getDropForeignKeySQL() throws exception.
+     *
+     * @return bool
+     */
+    public function supportsCreateDropForeignKeyConstraints()
```
lcobucci: We could add type declaration here, right?

lib/Doctrine/DBAL/Internal/CommitOrderCalculator.php
lcobucci: Perhaps we could bring this to `doctrine/persistence` to be reuse by the orm too?

lib/Doctrine/DBAL/Schema/SchemaDiff.php
beberlei: No, the ORM uses it with weights, we can remove he ORM implementation once its here in DBAL.

tests/Doctrine/Tests/DBAL/Internal/CommitOrderCalculatorTest.php
beberlei: I don't know, this is copied from the ORM.

lib/Doctrine/DBAL/Internal/CommitOrderCalculator.php
beberlei: This is used in ORM. I add a test assertion.

tests/Doctrine/Tests/DBAL/Schema/SchemaDiffTest.php
beberlei: This fails `Error: Call to undefined method PHPUnit\Framework\MockObject\Builder\InvocationMocker::willReturnValue()
`

lib/Doctrine/DBAL/Internal/CommitOrderCalculator.php
beberlei: I made it final now, we can move it to persistence in another step, but ORM depends on DBAL, so it can already use it?

lib/Doctrine/DBAL/Schema/SchemaDiff.php
```diff
@@ -165,4 +169,37 @@ protected function _toSql(AbstractPlatform $platform, $saveMode = false)
 
         return $sql;
     }
+
+    /**
+     * Sorts tables by dependencies so that they are created in the right order.
+     *
+     * This is ncessary when one table depends on another while creating foreign key
```
greg0ire:
```suggestion
     * This is necessary when one table depends on another while creating foreign key
```

lib/Doctrine/DBAL/Internal/CommitOrderCalculator.php
morozov: If it's needed in ORM together with the methods like `hasNode()` that are not used by DBAL and probably weights as well, why not move it to a shared library instead?

It feels weird that the code is ported from one library to another as is just because it's needed as is for the library it's currently located in.

lib/Doctrine/DBAL/Internal/CommitOrderCalculator.php
```diff
@@ -0,0 +1,151 @@
...
+            if ($vertex->state !== self::NOT_VISITED) {
+                continue;
+            }
+
+            $this->visit($vertex);
+        }
+
+        $sortedList = $this->sortedNodeList;
+
+        $this->nodeList       = [];
```
BackEndTea: Is it necessary to reset `nodeList` here?

Right now this method is only used in one space, and the object is discarded after that, so it doesn't really matter. But to me its unexpected that sorting would reset the object.

Or is this for memory efficiency?

lib/Doctrine/DBAL/Internal/CommitOrderCalculator.php
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * CommitOrderCalculator implements topological sorting, which is an ordering
+ * algorithm for directed graphs (DG) and/or directed acyclic graphs (DAG) by
+ * using a depth-first searching (DFS) to traverse the graph built in memory.
+ * This algorithm have a linear running time based on nodes (V) and dependency
+ * between the nodes (E), resulting in a computational complexity of O(V + E).
+ */
+final class CommitOrderCalculator
+{
+    public const NOT_VISITED = 0;
```
BackEndTea: I think these consts can be private, as they are not used outside of the class

lib/Doctrine/DBAL/Internal/CommitOrderEdge.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Doctrine\DBAL\Internal;
+
+class CommitOrderEdge
```
BackEndTea: Could these two classes be final?

tests/Doctrine/Tests/DBAL/Schema/SchemaDiffTest.php
lcobucci: My bad, it's `willReturn()` :man_facepalming:

lib/Doctrine/DBAL/Internal/CommitOrderEdge.php
beberlei: They could, but as they are created within the calculator it doesn't make a difference.

lib/Doctrine/DBAL/Internal/CommitOrderCalculator.php
beberlei: As mentioned elsewhere this code is copied from the ORM and kept mostly as-is.

lib/Doctrine/DBAL/Internal/CommitOrderCalculator.php
beberlei: @morozov i don't see how its weird, DBAL is the common denominator of where this code is needed in terms of dependencies. Introducing yet another library is just adding overhead. In addition it would prevent this fix from being merged into 2.10 per the dependencies are locked rule.

morozov: IMO, the "common denominator" approach is wrong and counterproductive. This way, the consumers of the shared logic end up storing and maintaining the logic they need in a place where it doesn't belong.

Eventually, the changes in this logic are driven by the consumers while the owners of the library end up accepting everything "just because it's needed for project X".

In this case, I see two proper solutions:

1. Have `@internal` implementations that solve the problems of each of the libraries independently and evolve independently (looks the most reasonable in this case).
2. Have a shared library that solves this problem in general (look like overkill in this case).

beberlei: We can keep this code as a 100% copy of the ORM code. This is not an attempt to force the ORM to use the class from here in the future. I just happened to need the exact same logic, because its the same problem, hence suggesting it can be re-used by the ORM in the future.

morozov: I disagree. The commit order resolution is an _internal_ (as the namespace suggests) implementation detail of the SQLite driver which is _not_ part of the database abstraction layer. It cannot be planned for reuse by another component. And this argument cannot be used to keep the pieces of the code that this component doesn't need for its _internal_ purposes.

greg0ire:
The new one true base branch is 2.10.x, switching to that.

morozov (dismissed):

morozov:
@beberlei the result looks good. Should we squash at least some commits? The “fix typo”-like are definitely irrelevant for the history.

morozov:
The ContinuousPHP failure is temporary and expected due to the configuration change in https://github.com/doctrine/dbal/pull/3915.

morozov:
@beberlei to avoid double builds w/o making additional configuration changes across CI platforms, please open future PRs from your fork instead of the upstream repository.

PReimers:
@morozov 
Any chance this could make it into the next release?

morozov:
@PReimers yes, but it depends on whether @beberlei finds the time to finish it.

alcaeus:
@morozov if it's just a rebase that involves squashing a bunch of commits, there's nothing to prevent us from doing that ourselves or doing it via the GitHub merge function. I've squashed all commits since all follow-up commits were only due to review changes.

alcaeus (dismissed):

morozov:
@alcaeus the newly added test fails on Oracle:
```
1) Doctrine\Tests\DBAL\Functional\Schema\OracleSchemaManagerTest::testSchemaDiffForeignKeys
--- Expected
+++ Actual
@@ @@
 Array (
+    0 => 'ALTER TABLE child MODIFY (ID ... NULL)'
+    1 => 'ALTER TABLE parent MODIFY (ID... NULL)'
 )

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php:1642
```

beberlei:
@morozov Fixed by rewriting the assertion logic, I couldn't find out why Oracle had the column ID vs id names because of the default casing. Its not related to what the test is testing, so the adjusted assertion takes care of that.

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1604,4 +1606,49 @@ public function testCommentInTable() : void
...
+        ]);
+
+        $this->assertCount(1, $schema->getTable('child')->getForeignKeys());
+
+        $offlineSchema = new Schema([$table1, $table2]);
+
+        $diff = Comparator::compareSchemas($offlineSchema, $schema);
+
+        foreach ($diff->changedTables as $table) {
+            if (count($table->changedForeignKeys) <= 0 && count($table->addedForeignKeys) <= 0 && count($table->removedForeignKeys) <= 0) {
```
morozov, line 1644: Not that it needs to be taken care of immediately but aren't these just three assertions like `assertCount(0, $table->changedForeignKeys)`?

morozov:
> […] Oracle had the column ID vs id names because of the default casing. Its not related to what the test is testing, so the adjusted assertion takes care of that.

Given that on Oracle, DBAL produces unexpected DDL, it looks like a bug (although, unrelated). Right?

morozov (approved):

morozov:
Thanks, @beberlei and @alcaeus.

phpstan.neon.dist
```diff
@@ -77,3 +77,9 @@ parameters:
         -
             message: '~^Access to undefined constant PDO::PGSQL_ATTR_DISABLE_PREPARES\.$~'
             path: %currentWorkingDirectory%/lib/Doctrine/DBAL/Driver/PDOPgSql/Driver.php
+
+        # False Positive
+        - '~Strict comparison using === between 1 and 2 will always evaluate to false~'
```
morozov, line 82: @beberlei how is this a false positive?

https://github.com/doctrine/dbal/blob/85a983c3f8d6447c4441832477def1c28d57bee9/lib/Doctrine/DBAL/Internal/DependencyOrderCalculator.php#L106-L123

It looks like a bug in the code. How is `$vertex->state` expected to change its state from `IN_PROGRESS` to anything else between the first and last lines of the block above?

PR #3769, opened 2019-12-03
Drop AbstractPlatform::usesSequenceEmulatedIdentityColumns()

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | N/A

#### Summary

Drops `AbstractPlatform::usesSequenceEmulatedIdentityColumns()`, as suggested in https://github.com/doctrine/dbal/pull/3759#discussion_r353315457.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -2680,18 +2680,6 @@ abstract class AbstractPlatform
     }
 
     /**
-     * Whether the platform emulates identity columns through sequences.
-     *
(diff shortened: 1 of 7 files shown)


SenseException (approved):

morozov:
@SenseException as it turned out, this patch is not relevant or required to proceed with #3759 as I originally thought (https://github.com/doctrine/dbal/pull/3759#discussion_r354576908). We can leave it open for now and revisit later if/when it becomes relevant.

morozov:
@beberlei I see that this method is used by ORM. What's the use case? Why is it important for the consumer how a certain API is implemented internally?

PR #3771, opened 2019-12-05
Return result from commit()

chosroes:
Add modification from #3588 for 2.9 which support PHP 7.1


morozov:
After the release 2.10.0, 2.9 accepts only security fixes.

PR #3778, opened 2019-12-07
[GH-3777] Don't remove composer lock travis on stable 2.10 branch

### LINKED ISSUE #3777: Don't remove composer.lock in test runs
### Bug Report
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.0

#### Summary

.travis.yml removes the composer.lock and then makes an composer install (being an composer update). This pulls in new dependencies in stable branches that weren't tested against causing errors (PHPunit + stopOnWarning=true).

beberlei:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3777 

#### Summary

Fixes #3777

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -49,7 +49,7 @@ install:
     - ps: |
         # Check if installation is cached
         if (!(Test-Path c:\tools\php)) {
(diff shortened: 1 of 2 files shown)


morozov:
Please remove this too:

https://github.com/doctrine/dbal/blob/70553dac3a60309336d07dc84e489a34216d962c/.appveyor.yml#L120

morozov:
> Don't remove composer lock travis on stable 2.10 branch

As of a couple of weeks ago, we started merging 2.x up into master instead of cherry-picking changes from master to 2.x. This way, this change will end up in master as well. This is fine but the commit message may be a bit misleading: we don't want to remove the lock file on CI regardless of the branch stability because we want PR builds to be stable.

morozov (approved):

PR #625, opened 2014-06-26
Fix pg boolean conversion

lucasvanlierop:
Merge of complementary PR's #527 and #564 + some fixes. Contributing it as an result of the (extended) phpconference.nl 2014 hackathon.

PR #3781, opened 2019-12-14
Transpose PostgreSqlPlatform::$booleanLiterals for optimization

igrizzli:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

Searching by key instead in_array works twice faster on this array.

##### Original version
real	0m0.804s
user	0m0.799s
sys	0m0.004s
```
$booleanLiterals = [
    'true' => [
        't',
        'true',
        'y',
        'yes',
        'on',
        '1',
    ],
    'false' => [
        'f',
        'false',
        'n',
        'no',
        'off',
        '0',
    ],
];

$value = 'true';
for($i = 0; $i < 10000000; $i++) {
    in_array(strtolower(trim($value)), $booleanLiterals['false'], true);
}
```
##### Transposed version
real	0m0.470s
user	0m0.462s
sys	0m0.008s
```
$booleanLiterals = [
    't' => true,
    'true' => true,
    'y' => true,
    'yes' => true,
    'on' => true,
    '1' => true,
    'f' => false,
    'false => false,
    'n' => false,
    'no' => false,
    'off' => false,
    '0' => false,
];

$value = 'true';
for($i = 0; $i < 10000000; $i++) {
    $booleanLiterals[strtolower(trim($value))];
}
```

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -23,7 +23,6 @@ use function array_values;
 use function count;
 use function explode;
 use function implode;
-use function in_array;
 use function is_array;
 use function is_bool;
 use function is_numeric;
@@ -43,24 +42,21 @@ class PostgreSqlPlatform extends AbstractPlatform
     /** @var bool */
     private $useBooleanTrueFalseStrings = true;
 
-    /** @var string[][] PostgreSQL booleans literals */
+    /** @var bool[] PostgreSQL booleans literals */
     private $booleanLiterals = [
-        'true' => [
-            't',
-            'true',
-            'y',
-            'yes',
-            'on',
-            '1',
-        ],
-        'false' => [
-            'f',
-            'false',
-            'n',
-            'no',
-            'off',
-            '0',
-        ],
+        't'     => true,
+        'true'  => true,
+        'y'     => true,
+        'yes'   => true,
+        'on'    => true,
+        '1'     => true,
+
+        'f'     => false,
+        'false' => false,
+        'n'     => false,
+        'no'    => false,
+        'off'   => false,
+        '0'     => false,
     ];
 
     /**
@@ -809,15 +805,9 @@ SQL
             return $callback(true);
         }
 
-        /**
-         * Better safe than sorry: http://php.net/in_array#106319
-         */
-        if (in_array(strtolower(trim($value)), $this->booleanLiterals['false'], true)) {
-            return $callback(false);
-        }
-
-        if (in_array(strtolower(trim($value)), $this->booleanLiterals['true'], true)) {
-            return $callback(true);
+        $boolean = $this->booleanLiterals[strtolower(trim($value))] ?? null;
+        if ($boolean !== null) {
+            return $callback($boolean);
         }
 
         throw new UnexpectedValueException(sprintf(
@@ -896,7 +886,7 @@ SQL
      */
     public function convertFromBoolean($item) : ?bool
     {
-        if (in_array($item, $this->booleanLiterals['false'], true)) {
+        if (($this->booleanLiterals[$item] ?? null) === false) {
             return false;
         }
 
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -39,11 +39,7 @@ use function trim;
  */
 class PostgreSqlPlatform extends AbstractPlatform
 {
-    /** @var bool */
-    private $useBooleanTrueFalseStrings = true;
-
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -809,15 +805,9 @@ private function convertSingleBooleanValue($value, callable $callback)
...
-        /**
-         * Better safe than sorry: http://php.net/in_array#106319
-         */
-        if (in_array(strtolower(trim($value)), $this->booleanLiterals['false'], true)) {
-            return $callback(false);
-        }
-
-        if (in_array(strtolower(trim($value)), $this->booleanLiterals['true'], true)) {
-            return $callback(true);
+        $boolean = $this->booleanLiterals[strtolower(trim($value))] ?? null;
```
morozov: Could it be just an `isset()` check? Although may be out of scope, why do we want to `trim()` the value?

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -896,7 +886,7 @@ static function ($boolean) : ?int {
      */
     public function convertFromBoolean($item) : ?bool
     {
-        if (in_array($item, $this->booleanLiterals['false'], true)) {
+        if (($this->booleanLiterals[$item] ?? null) === false) {
```
morozov: Could it be just an `isset()` check?

morozov (commented):
So… this way, we win 330 nanoseconds per method call? Besides the technical possibility of improvement, is there any practical reason for the change?

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -43,24 +42,21 @@ class PostgreSqlPlatform extends AbstractPlatform
     /** @var bool */
     private $useBooleanTrueFalseStrings = true;
 
-    /** @var string[][] PostgreSQL booleans literals */
+    /** @var bool[] PostgreSQL booleans literals */
     private $booleanLiterals = [
```
morozov: Speaking of the optimization, would it make sense to make it a constant instead of an object property?

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
igrizzli: 1. If we will use isset, we will need to search in array twice.
2. I don't know reason but it comes from here https://github.com/doctrine/dbal/pull/625/commits/b35b682486d76b3eea71ade1bfc4f8a4b59d0a13

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
igrizzli: We should read value, not only check that it exists.

morozov: Yes. An `isset()` check and then read the value. Accessing an element of an array of 12 elements twice is hardly different from doing it once from the performance perspective but makes the logic clear which is more important in this case.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -809,15 +804,9 @@ private function convertSingleBooleanValue($value, callable $callback)
...
-         * Better safe than sorry: http://php.net/in_array#106319
-         */
-        if (in_array(strtolower(trim($value)), $this->booleanLiterals['false'], true)) {
-            return $callback(false);
-        }
-
-        if (in_array(strtolower(trim($value)), $this->booleanLiterals['true'], true)) {
-            return $callback(true);
+        $boolean = self::BOOLEAN_LITERALS[strtolower(trim($value))] ?? null;
+        if ($boolean !== null) {
```
morozov: This `?? null` followed by `!== null` hides the intent of the code. Previously, it was clear: `if (in_array(/* ... */))` and it must remain the same clear: `if (isset(/* ... */))`.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -896,7 +885,7 @@ static function ($boolean) : ?int {
      */
     public function convertFromBoolean($item) : ?bool
     {
-        if (in_array($item, $this->booleanLiterals['false'], true)) {
+        if ((self::BOOLEAN_LITERALS[$item] ?? null) === false) {
```
morozov: This expression must remain as clear as it used to be: `if isset(/* ... */) return`;

morozov (changes requested):

morozov:
Closing as stale. There's no clear performance benefit that would warrant this code change.

PR #3782, opened 2019-12-14
BUGFIX: Fix to handle default values in columnDefinition

ksn135:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | unknown

Fix to handle default values in columnDefinition

#### Summary

For example, you wanna set default value for filed that related to another table
```php
    /**
     * @var Employee $visa
     * @ORM\ManyToOne(targetEntity="Employee")
     * @ORM\JoinColumn(name="visa_id", referencedColumnName="id", nullable=false, columnDefinition="INT NOT NULL DEFAULT 434")
     */
    private $visa = null;
``` 
Without this fix, it's always trying to update schema, even when it's already valid.
```sql
ALTER TABLE contract CHANGE visa_id visa_id INT NOT NULL DEFAULT 434;
```

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Comparator.php
+++ b/lib/Doctrine/DBAL/Schema/Comparator.php
@@ -14,6 +14,7 @@ use function assert;
 use function count;
 use function get_class;
 use function strtolower;
+use function preg_match;
 
 /**
  * Compares two Schemas and return an instance of SchemaDiff.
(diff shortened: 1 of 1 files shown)


morozov (commented):
Please add a cross-platform functional test that reproduces the issue being fixed.

ksn135:
> Please add a cross-platform functional test that reproduces the issue being fixed.

Well, I can reproduce this issue only on MySql 5.6, 5.7 and MariaDB 10.x
Sorry, BUT I have no idea how to write "a cross-platform functional test" for DBAL.

morozov:
@ksn135 please use the tests in the [`Doctrine\Tests\DBAL\Functional`](https://github.com/doctrine/dbal/tree/master/tests/Doctrine/Tests/DBAL/Functional) namespace for example and find something close enough to your case.

We cannot proceed without a test case.

SenseException:
@ksn135 If you need help on how to write the cross-platform test, we're happy to help you.

PR #3786, opened 2019-12-18
Test against PHP 7.4 on AppVeyor

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -21,17 +21,17 @@ environment:
   - db: mssql
     driver: sqlsrv
     db_version: sql2012sp1
-    php: 7.3
+    php: 7.4
     coverage: yes
   - db: mssql
     driver: sqlsrv
     db_version: sql2017
     coverage: no
-    php: 7.3
+    php: 7.4
   - db: mssql
     driver: pdo_sqlsrv
     db_version: sql2017
-    php: 7.3
+    php: 7.4
     coverage: yes
 
 init:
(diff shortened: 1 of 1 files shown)


.appveyor.yml
```diff
@@ -44,7 +44,7 @@ install:
     - ps: |
         # Check if installation is cached
         if (!(Test-Path c:\tools\php)) {
-          appveyor-retry cinst --params '""/InstallDir:C:\tools\php""' --ignore-checksums -y php --version 7.3.12
+          appveyor-retry cinst --params '""/InstallDir:C:\tools\php""' --ignore-checksums -y php --version ((choco search php --exact --all-versions -r | select-string -pattern $env:php | sort { [version]($_ -split '\|' | select -last 1) } -Descending | Select-Object -first 1) -replace '[php|]','')
```
greg0ire: Is there a way to use linebreaks here?

morozov: This is the original implementation being restored (see https://github.com/doctrine/dbal/pull/3756).

While I agree with the suggestion in general, I'm not sure how to break it down in a way that makes it better readable given that the longest expression sits in a sub-sub-shell and there's one 4rd level sub-shell.

greg0ire: Ok, it was in case it would be easy, I don't want to bother you with that.

greg0ire (approved):

PR #3792, opened 2019-12-21
Avoid losing name parts due to namespace splitting

drieschel:
|      Q       |   A
|------------- | -----------
| BC Break     | no
| Fixed issues | 

#### Summary

This pr will hopefully fix a problem when setting an index name with multiple dots (ie my.index.name) over the abstract asset schema class. All parts after the second dot got lost. 

Solution in progress....


lib/Doctrine/DBAL/Schema/AbstractAsset.php
```diff
@@ -48,8 +48,8 @@ protected function _setName(string $name) : void
         }
         if (strpos($name, '.') !== false) {
             $parts            = explode('.', $name);
-            $this->_namespace = $parts[0];
-            $name             = $parts[1];
+            $this->_namespace = array_shift($parts);
+            $name             = implode('.', $parts);
```
greg0ire: I think you can achieve the same by using the third argument of `explode`: `explode('.', $name, 2)`

greg0ire: https://3v4l.org/95JrL

greg0ire: Also, we could have used `list()` here: `[$this->_namespace, $name] = explode('.', $name, 2);`

drieschel: Thanks for the review. I refactored it. What do you think about this version?

greg0ire: I think it's better than the first version you proposed 👍

greg0ire (dismissed):
But this version is really cool :ok_hand:

drieschel:
Thank you @greg0ire :-)

lib/Doctrine/DBAL/Schema/AbstractAsset.php
```diff
@@ -46,10 +46,9 @@ protected function _setName(string $name) : void
             $this->_quoted = true;
             $name          = $this->trimQuotes($name);
         }
+
```
greg0ire: The extra spaces are what make the build fail. Also, you probably want to target a lower branch, I'm going to fix all this for you.

greg0ire (dismissed):

lib/Doctrine/DBAL/Schema/AbstractAsset.php
drieschel: I just saw this comment. Thanks again for your support. If there is something I can do, please let me know.

lib/Doctrine/DBAL/Schema/AbstractAsset.php
```diff
@@ -48,10 +48,9 @@ protected function _setName($name)
...
         if (strpos($name, '.') !== false) {
-            $parts            = explode('.', $name);
-            $this->_namespace = $parts[0];
-            $name             = $parts[1];
+            [$this->_namespace, $name] = explode('.', $name, 2);
```
morozov: How do you know that the dot belongs to the asset but not to the namespace? In SQL syntax, a dot is a separator between database name, schema name, object name, etc. It's surprising to see it used as part of the object name, especially in this generic case (any platform, any object type).

tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
```diff
@@ -185,4 +185,12 @@ public function testOptions()
         self::assertSame('name IS NULL', $idx2->getOption('WHERE'));
         self::assertSame(['where' => 'name IS NULL'], $idx2->getOptions());
     }
+
+    public function testGetNameWithMultipleDots() : void
+    {
+        $name  = 'my.index.name';
```
morozov: Please turn this into a functional test that deploys a schema containing an object that has a dot in its name and introspects it. It needs to be proven that it's practically possible and practically correct.

morozov (changes requested):

lib/Doctrine/DBAL/Schema/AbstractAsset.php
drieschel: Hello @morozov,

look at the original version: The part before the first dot was taken for the namespace. The assumption here is, that the rest should be part of the name.

tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
drieschel: In my case (an index in a mysql database which had dots in its name), the name of the index is not unique anymore when parts of the name are lost. I am not completely sure what else has to be proven here.

greg0ire: > I am not completely sure what else has to be proven here.

That mysql tolerates indices with a name containing a dot, if I understood correctly

drieschel: I see (hopefully). Where should the functional test to be added (.../Functional/Schema/MysqlSchemaManagerTest.php)?

drieschel: @greg0ire, @morozov After a more testing, I realized, that this will not work in the actual way, because the namespace will be taken for the database name (as you said @morozov). What should be done here? Closing the pull request and stick with the convention that index names can't contain dots? Or refactoring it?

I still can create a functional test to show that mysql indices can have dots in the name (without dbal).

morozov: What if the namespace contains dots too?

drieschel: After a lot of testing yesterday, I think that an index does not have a namespace. A schema, table or column can have a namespace (including dots). The name of an asset can not contain dots, except it is an index. Does this make sense?

morozov: > But for index it is, because it does not have a namespace.

Is it true for all supported platforms? There's currently no clear separation in DBAL between indices and constraints, and depending on the platform some of them are scoped to a table and some are scoped to a schema (both of which I believe may have dots in their names if quoted).

It feels like solving this issue will require significant research first.

PR #3793, opened 2019-12-22
Remove superfluous Configuration instance

mhitza:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary
A configuration instance is not strictly required by [DriverManager::getConnection](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/DriverManager.php#L139).

As an upside, it doesn't add an extra warning for people that use Psalm/PHPmd when starting off from the documentation example. As the Configuration class has an `@internal` block those tools threat it as the other meaning for `@internal`, as in class that is [internal to the library and not for public use](https://docs.phpdoc.org/references/phpdoc/tags/internal.html).

### DIFF
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -10,7 +10,6 @@ You can get a DBAL Connection through the
 .. code-block:: php
 
     <?php
-    $config = new \Doctrine\DBAL\Configuration();
     //..
     $connectionParams = array(
         'dbname' => 'mydb',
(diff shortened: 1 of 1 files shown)


morozov (approved):
Looks reasonable. I think the only use case when those parameters need to be passed explicitly is duplicating an existing connection. See `PortabilityTest`.

morozov:
Thanks @mhitza!

PR #3794, opened 2019-12-23
Support ForeignKeyConstraintViolationException in SQLite

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

SQLite [supports foreign keys](https://www.sqlite.org/foreignkeys.html), but `ForeignKeyConstraintViolationException` is not handled by the DBAL.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOSqlite/DriverTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOSqlite/DriverTest.php
@@ -6,7 +6,6 @@ namespace Doctrine\Tests\DBAL\Functional\Driver\PDOSqlite;
 
 use Doctrine\DBAL\Driver as DriverInterface;
 use Doctrine\DBAL\Driver\PDOSqlite\Driver;
-use Doctrine\DBAL\Exception\ForeignKeyConstraintViolationException;
 use Doctrine\Tests\DBAL\Functional\Driver\AbstractDriverTest;
 use function extension_loaded;
 
@@ -46,32 +45,5 @@ class DriverTest extends AbstractDriverTest
     protected static function getDatabaseNameForConnectionWithoutDatabaseNameParameter() : ?string
     {
         return '';
-    }
-
-    public function testForeignKeyConstraintViolationException() : void
-    {
-        $this->connection->exec('PRAGMA foreign_keys = ON');
-
-        $this->connection->exec('
-            CREATE TABLE parent (
-                id INTEGER PRIMARY KEY
-            )
-        ');
-
-        $this->connection->exec('
-            CREATE TABLE child ( 
-                id INTEGER PRIMARY KEY, 
-                parent_id INTEGER,
-                FOREIGN KEY (parent_id) REFERENCES parent(id)
-            );
-        ');
-
-        $this->connection->exec('INSERT INTO parent (id) VALUES (1)');
-        $this->connection->exec('INSERT INTO child (id, parent_id) VALUES (1, 1)');
-        $this->connection->exec('INSERT INTO child (id, parent_id) VALUES (2, 1)');
-
-        $this->expectException(ForeignKeyConstraintViolationException::class);
-
-        $this->connection->exec('INSERT INTO child (id, parent_id) VALUES (3, 2)');
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOSqlite/DriverTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOSqlite/DriverTest.php
@@ -6,6 +6,7 @@ namespace Doctrine\Tests\DBAL\Functional\Driver\PDOSqlite;
 
 use Doctrine\DBAL\Driver as DriverInterface;
 use Doctrine\DBAL\Driver\PDOSqlite\Driver;
+use Doctrine\DBAL\Exception\ForeignKeyConstraintViolationException;
 use Doctrine\Tests\DBAL\Functional\Driver\AbstractDriverTest;
 use function extension_loaded;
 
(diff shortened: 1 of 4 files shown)


morozov:
@BenMorel would it be possible to create or identify an existing functional where a real DB would trigger this exception?

BenMorel:
@morozov Sure thing! Done in a2b4848534618f03bdb08a38f3aa823c3915ef85.

Note that I haven't been able to re-use any of the existing tests, as they're all aimed for platforms supporting `ALTER` constraints, which SQLite does not.

*Side question: is this the reason why [`supportsForeignKeyConstraints()` returns `false`](https://github.com/doctrine/dbal/blob/ef6b84effcdb4948a39661c9706b53e9a728d0e4/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php#L760-L763), even though SQLite does support FKs (when compiled with FK support and when explicitly enabled through `PRAGMA`)?*

So I've created a standalone test in `PDOSqlite\DriverTest`; please let me know if there's a better place for this one.

tests/Doctrine/Tests/DBAL/Functional/Driver/PDOSqlite/DriverTest.php
```diff
@@ -46,4 +47,31 @@ protected static function getDatabaseNameForConnectionWithoutDatabaseNameParamet
     {
         return '';
     }
+
+    public function testForeignKeyConstraintViolationException() : void
```
morozov: Could this test be run against all supported drivers, not only SQLite? We want to make sure that DBAL behaves the same in the same situation regardless of the platform.

BenMorel: There are already tests for FK constraint violations, but they're not run because `SqlitePlatform::supportsForeignKeyConstraints()` returns `false`. If I change it to `true`, many tests start to fail because the whole codebase assumes that when FKs are supported, `ALTER TABLE` statements with FK changes are acceptable, even though they're not for SQLite: FKs can only be set up when creating the table, or when adding a column with the FK attached to it.

That being said, after investigating a bit more, I noticed that these tests in `Functional\ExceptionTest` work fine with SQLite:

- `testForeignKeyConstraintViolationExceptionOnInsert()`
- `testForeignKeyConstraintViolationExceptionOnUpdate()`
- `testForeignKeyConstraintViolationExceptionOnDelete()`
- `testForeignKeyConstraintViolationExceptionOnTruncate()`

...provided that you enable foreign keys with `PRAGMA foreign_keys = ON` first.

So I removed the FK test in `PDOSqlite\DriverTest`, and put an additional condition in the above tests for SQLite, which doesn't fully "support FKs" according to the DBAL (missing the `ALTER` capability), but can still pass these tests successfully!

See d729f7bb9698eaafc0f7adaf89aaaf52a16d5269.

morozov: Isn't it weird that the platform doesn't support foreign keys but does throw a `ForeignKeyConstraintViolationException`?

> Note that I haven't been able to re-use any of the existing tests, as they're all aimed for platforms supporting `ALTER` constraints, which SQLite does not.

Would it be possible to rework them to not rely on `ALTER`? The end goal is to create a table with foreign keys using the current platform capabilities. If it's not possible by means of using the DBAL APIs, it means we're missing something in them.

> *Side question: is this the reason why [`supportsForeignKeyConstraints()` returns `false`](https://github.com/doctrine/dbal/blob/ef6b84effcdb4948a39661c9706b53e9a728d0e4/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php#L760-L763), even though SQLite does support FKs (when compiled with FK support and when explicitly enabled through `PRAGMA`)?*

No idea. Most likely, it just means "not implemented/supported in DBAL".

BenMorel: > Would it be possible to rework them to not rely on ALTER? The end goal is to create a table with foreign keys using the current platform capabilities. If it's not possible by means of using the DBAL APIs, it means we're missing something in them.

I *guess* it should be possible to emulate adding a FK to an existing column by:

- adding another column with FK enabled,
- copying the values from the first column,
- dropping the first column,
- then renaming the new column.

I haven't tested it, though, and for sure I won't have enough time to investigate further and implement it.

Yes, it's kinda weird that the Platform returns `false` for foreign key support, and still handles a FK exception alright; but IMO that's still better than the current situation, and can be improved upon later.

morozov: Not sure I get it why we need to emulate anything in a test and copy the values. We just need to create a table with a foreign key and then perform the operations that violate it to test the exception.

BenMorel: You don't need to emulate it *in tests*, you need to emulate it if you want full support for FKs *as defined by the DBAL*.

For example, `SqlitePlatform` [does not support the following methods](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php#L765-L795):

- `AbstractPlatform::getCreateForeignKeySQL()`
- `AbstractPlatform::getDropForeignKeySQL()`

This is where emulation is required if you want to allow `supportsForeignKeyConstraints()` to return `true`.

For tests, it works fine if you just create a table with the constraints straight away. **And it is indeed what the current PR does**, by enabling `testForeignKeyConstraintViolationExceptionOnXXX()` on SQLite.

morozov: I see. Then the name/meaning of the `supportsForeignKeyConstraints()` method is unclear/misleading:

https://github.com/doctrine/dbal/blob/436f2f6c943937abe8544505630e76be4c468ae1/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php#L2788-L2791

On the one hand, it means that the platform supports generating SQL that implements foreign keys (all platforms except for SQLite), specifically in `get(Pre|Post)AlterTableIndexForeignKeySQL()`.

On the other hand, it means that the foreign keys of a given schema can be introspected and comparison (all platforms without exception): `AbstractSchemaManager::listTableDetails()`.

Given that `supportsForeignKeyConstraints()` still returns `false` for SQLite, how is the above going to work?

BenMorel: > I see. Then the name/meaning of the supportsForeignKeyConstraints() method is unclear/misleading

Agreed. Not sure if this is in the scope of this PR, though.

> Given that supportsForeignKeyConstraints() still returns false for SQLite, how is the above going to work?

As you can see in [the diff](https://github.com/doctrine/dbal/pull/3794/files#diff-f45e6c0d4eabd29618b2fabf8b20413dR83-R85), the `testForeignKeyConstraintViolationExceptionOnXXX()` methods specifically check if the platform is SQLite, before checking for `supportsForeignKeyConstraints()`.

This check, specific to SQLite, would be required anyway to issue the necessary `PRAGMA` statement to enable foreign keys.

The aim of this PR is just to handle the FK exception alright, and these tests check precisely this.

morozov: > Agreed. Not sure if this is in the scope of this PR, though.

No, it's not.

> As you can see in [the diff](https://github.com/doctrine/dbal/pull/3794/files#diff-f45e6c0d4eabd29618b2fabf8b20413dR83-R85), the `testForeignKeyConstraintViolationExceptionOnXXX()` methods specifically check if the platform is SQLite, before checking for `supportsForeignKeyConstraints()`.

The question was about how the schema manager in production code would compare the foreign key definitions given that `AbstractSchemaManager::listTableDetails()` will always return a table definition without foreign keys on SQLite because `supportsForeignKeyConstraints()` will return false.

Although, it's also out of scope.

morozov (approved):

morozov:
Thanks @BenMorel!

PR #3795, opened 2019-12-23
Document MySQL error codes

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

Documents mysql error codes for easier reading.

References:

- https://dev.mysql.com/doc/refman/5.7/en/server-error-reference.html
- https://dev.mysql.com/doc/refman/5.7/en/client-error-reference.html

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
@@ -27,85 +27,148 @@ use function version_compare;
  */
 abstract class AbstractMySQLDriver implements ExceptionConverterDriver, VersionAwarePlatformDriver
 {
+    /**#@+
+     * MySQL server error codes.
+     *
+     * @link https://dev.mysql.com/doc/refman/8.0/en/server-error-reference.html
+     */
+    private const ER_DBACCESS_DENIED_ERROR            = 1044;
+    private const ER_ACCESS_DENIED_ERROR              = 1045;
+    private const ER_NO_DB_ERROR                      = 1046;
+    private const ER_BAD_NULL_ERROR                   = 1048;
+    private const ER_BAD_DB_ERROR                     = 1049;
+    private const ER_TABLE_EXISTS_ERROR               = 1050;
+    private const ER_BAD_TABLE_ERROR                  = 1051;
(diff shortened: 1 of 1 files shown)


BenMorel:
Not sure how to fix the `phpcs` issue:

- when I don't put an empty line before the `return` statement, I get:

> Expected 1 lines before "return", found 0.

- when I do put an empty line before it, I get:

> The CASE body must start on the line following the statement

This contradiction only occurs when a comment is on the same line as the `case`.

morozov:
Why not define the codes as constants instead of comments?

BenMorel:
It's a possibility, it just adds a bit of burden for something that's not going to be reused anywhere IMHO.

If you wish to proceed this way, please tell me where they should go:

- as constants in the `AbstractMySQLDriver` class
- or as constants in the `Doctrine\DBAL\Driver` namespace (and in which file then?)

lcobucci:
@BenMorel I'd suggest private constants in `AbstractMySQLDriver`.

BenMorel:
Here you are!

morozov:
> It's a possibility, it just adds a bit of burden for something that's not going to be reused anywhere IMHO.

It's not about reuse. It's about self-documenting code.

> Here you are!

Thank you.

Could you please also move the `@link` annotations from `convertException()` to the group block around the newly declared constants like here?
https://github.com/doctrine/dbal/blob/c104c687f6a3880fb9ddeb5f9ebc4fae11d12466/lib/Doctrine/DBAL/SQLParserUtils.php#L34-L42

And then squash.

BenMorel:
Is this `/**#@+ ... */ ... /**#@-*/` a PHPdoc standard of some sort? I've never come across this one.

morozov:
> Is this `/**#@+ ... */ ... /**#@-*/` a PHPdoc standard of some sort? I've never come across this one.

This is/was actively used in [PEAR](https://github.com/pear/pear-core/blob/9e9014d4aa662db336eef3121011bdcd27f5e63f/PEAR.php#L21-L34), also I'm not sure if it's part of the phpDoc standard.

BenMorel:
@morozov All done!

morozov (approved):

morozov:
Thanks @BenMorel! :ship:

PR #3799, opened 2019-12-30
Upgrade to PHPStan v0.12

lcobucci:
|      Q       |   A
|------------- | -----------
| Type         | dev improvement
| BC Break     | no

#### Summary

Allowing us to use more advanced checks and have full generic support.

~~This also adds a baseline file, since we have new violations being reported but don't want to address them right now.~~

Adding to `2.10.x` because it doesn't influence production code and help us on #3762

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -383,7 +383,7 @@ class Connection implements DriverConnection
      *
      * @throws DBALException If an invalid platform was specified for this connection.
      */
-    private function detectDatabasePlatform()
+    private function detectDatabasePlatform() : void
     {
         $version = $this->getDatabasePlatformVersion();
 
@@ -504,6 +504,8 @@ class Connection implements DriverConnection
      * @see   isAutoCommit
      *
      * @param bool $autoCommit True to enable auto-commit mode; false to disable it.
+     *
+     * @return void
      */
     public function setAutoCommit($autoCommit)
     {
@@ -1316,7 +1318,7 @@ class Connection implements DriverConnection
     /**
      * Commits all current nesting transactions.
      */
-    private function commitAll()
+    private function commitAll() : void
     {
         while ($this->transactionNestingLevel !== 0) {
             if ($this->autoCommit === false && $this->transactionNestingLevel === 1) {
@@ -1333,6 +1335,8 @@ class Connection implements DriverConnection
 
     /**
      * Cancels any database changes done during the current transaction.
+     *
+     * @return bool
      *
      * @throws ConnectionException If the rollback operation failed.
      */
@@ -1373,6 +1377,8 @@ class Connection implements DriverConnection
             $this->isRollbackOnly = true;
             --$this->transactionNestingLevel;
         }
+
+        return true;
     }
 
     /**
--- a/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
+++ b/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
@@ -120,7 +120,9 @@ class MasterSlaveConnection extends Connection
     }
 
     /**
-     * {@inheritDoc}
+     * @param string|null $connectionName
+     *
+     * @return bool
      */
     public function connect($connectionName = null)
     {
--- a/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
+++ b/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
@@ -8,6 +8,7 @@ use Doctrine\DBAL\ParameterType;
 use stdClass;
 use const DB2_AUTOCOMMIT_OFF;
 use const DB2_AUTOCOMMIT_ON;
+use function assert;
 use function db2_autocommit;
 use function db2_commit;
 use function db2_conn_error;
@@ -23,6 +24,7 @@ use function db2_rollback;
 use function db2_server_info;
 use function db2_stmt_errormsg;
 use function func_get_args;
+use function is_bool;
 
 class DB2Connection implements Connection, ServerInfoAwareConnection
 {
@@ -140,7 +142,10 @@ class DB2Connection implements Connection, ServerInfoAwareConnection
      */
     public function beginTransaction()
     {
-        db2_autocommit($this->conn, DB2_AUTOCOMMIT_OFF);
+        $result = db2_autocommit($this->conn, DB2_AUTOCOMMIT_OFF);
+        assert(is_bool($result));
+
+        return $result;
     }
 
     /**
@@ -151,7 +156,11 @@ class DB2Connection implements Connection, ServerInfoAwareConnection
         if (! db2_commit($this->conn)) {
             throw new DB2Exception(db2_conn_errormsg($this->conn));
         }
-        db2_autocommit($this->conn, DB2_AUTOCOMMIT_ON);
+
+        $result = db2_autocommit($this->conn, DB2_AUTOCOMMIT_ON);
+        assert(is_bool($result));
+
+        return $result;
     }
 
     /**
@@ -162,7 +171,11 @@ class DB2Connection implements Connection, ServerInfoAwareConnection
         if (! db2_rollback($this->conn)) {
             throw new DB2Exception(db2_conn_errormsg($this->conn));
         }
-        db2_autocommit($this->conn, DB2_AUTOCOMMIT_ON);
+
+        $result = db2_autocommit($this->conn, DB2_AUTOCOMMIT_ON);
+        assert(is_bool($result));
+
+        return $result;
     }
 
     /**
--- a/lib/Doctrine/DBAL/Driver/PDOStatement.php
+++ b/lib/Doctrine/DBAL/Driver/PDOStatement.php
@@ -85,7 +85,13 @@ class PDOStatement extends \PDOStatement implements Statement
     }
 
     /**
-     * {@inheritdoc}
+     * @param mixed    $column
+     * @param mixed    $variable
+     * @param int      $type
+     * @param int|null $length
+     * @param mixed    $driverOptions
+     *
+     * @return bool
      */
     public function bindParam($column, &$variable, $type = ParameterType::STRING, $length = null, $driverOptions = null)
     {
--- a/lib/Doctrine/DBAL/Platforms/DrizzlePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/DrizzlePlatform.php
@@ -335,7 +335,10 @@ class DrizzlePlatform extends AbstractPlatform
     }
 
     /**
-     * {@inheritDoc}
+     * @param string      $table
+     * @param string|null $database
+     *
+     * @return string
      */
     public function getListTableForeignKeysSQL($table, $database = null)
     {
@@ -435,6 +438,7 @@ class DrizzlePlatform extends AbstractPlatform
     }
 
     /**
+     * @param string $table
      *
      * @return string
      */
--- a/lib/Doctrine/DBAL/Platforms/DrizzlePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/DrizzlePlatform.php
@@ -435,7 +435,8 @@ class DrizzlePlatform extends AbstractPlatform
     }
 
     /**
-     * {@inheritDoc}
+     *
+     * @return string
      */
     protected function getDropPrimaryKeySQL($table)
     {
--- a/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
@@ -61,6 +61,7 @@ class OraclePlatform extends AbstractPlatform
     }
 
     /**
+     * @param string $type
      *
      * @return string
      */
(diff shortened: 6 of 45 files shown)


morozov:
> […] but don't want to address them right now.

What's the point in using advanced checks and not taking them into account? How exactly is it related to #3762?

lcobucci:
@morozov v0.12 adds generics support, #3762 has a failure [related to it (incompatibility of `array<Doctrine\DBAL\Schema\Table>` and `array<object>`)](https://travis-ci.org/doctrine/dbal/jobs/630728395#L484-L491).

morozov:
Then can we just report/address this issue by using the new tool locally without an update?

lcobucci:
> Then can we just report/address this issue by using the new tool locally without an update?

I don't follow... not updating won't allow us to use the features.

morozov:
I don't follow either. By just baselining all reported errors we're not using it either. It's fine that the new version found an issue in an open PR but it doesn't mean we need to upgrade immediately and ignore all existing issues.

We should fix the ones fixing which won't cause a BC break as we have been always doing.

lcobucci:
The point of ignoring current errors that were discovered by the new version is to allow us to use the generics to unblock #3762 while fixing the issues that won't cause a BC-break in a separate PR (and we'd also prevent new issues from being merged).

We could wait for the fix, sure. I just won't have the time to them today and think that there's no point in waiting for getting help from a smarter static analysis.

morozov:
> I just won't have the time to them today […]

As far as I can tell, #3762 isn't that close to being done to consider the error reported by PHPStan 0.11 a blocker. It can be added to the whitelist and get the future work unblocked. It'd be a much smaller change.

> […] and think that there's no point in waiting for getting help from a smarter static analysis.

As I said, this way we're not getting help from the new PHPStan version by ignoring its input.

lcobucci:
> As I said, this way we're not getting help from the new PHPStan version by ignoring its input.

We would, new PRs would be validated using the new version and new issues would be prevented...

morozov:
> We would, new PRs would be validated using the new version and new issues would be prevented...

Yeah. But it's not that urgent that we need to rush and ignore all existing issues.

beberlei (dismissed):

phpstan-baseline.neon
```diff
@@ -0,0 +1,1027 @@
...
+
+		-
+			message: "#^Method Doctrine\\\\DBAL\\\\Types\\\\ConversionException\\:\\:conversionFailedSerialization\\(\\) has parameter \\$format with no typehint specified\\.$#"
+			count: 1
+			path: lib/Doctrine/DBAL/Types/ConversionException.php
+
+		-
+			message: "#^Method Doctrine\\\\DBAL\\\\Types\\\\ConversionException\\:\\:conversionFailedSerialization\\(\\) has parameter \\$value with no typehint specified\\.$#"
+			count: 1
+			path: lib/Doctrine/DBAL/Types/ConversionException.php
```
morozov: Having this many untriaged errors will complicate the project maintenance in the near future:

1. If someone makes a contribution and sees a new error, how do they know if they should fix it or whitelist? If triaged, some of the non-fixable issues could be generalized and cover the future code changes.
2. PHPStan on `2.10` doesn't cover `tests/` while it does on `master`. Once `2.10` is attempted to be merged to `master`, we'll can see potentially similar number of issues in `tests/`. How are we going to handle that?

morozov (changes requested):

phpstan-baseline.neon
lcobucci: > If someone makes a contribution and sees a new error, how do they know if they should fix it or whitelist? If triaged, some of the non-fixable issues could be generalized and cover the future code changes.

New errors on new code must be fixed.

> PHPStan on 2.10 doesn't cover tests/ while it does on master. Once 2.10 is attempted to be merged to master, we'll can see potentially similar number of issues in tests/. How are we going to handle that?

I'd baseline existing (yet undiscovered) errors.

lcobucci:
@morozov I'm (slowly) addressing the issues to solve your concerns, new commits are coming (eventually :sweat_smile:).

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -120,7 +120,9 @@ public function isConnectedToMaster()
     }
 
     /**
-     * {@inheritDoc}
```
lcobucci: It seems like PHPStan doesn't accept adding info together with `{@inheritDoc}`. Is that correct @ondrejmirtes?

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -606,7 +606,7 @@ public static function udfLocate($str, $substr, $offset = 0)
     /**
      * {@inheritDoc}
      */
-    public function getForUpdateSQL()
```
lcobucci: Todo: rebase things properly :man_facepalming:

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
ondrejmirtes: Yes, it doesn’t merge that. Either you are inheriting the phpDoc, or overriding it.

phpstan-baseline.neon
morozov: > New errors on new code must be fixed.

That's the problem. They may have the same underlying root cause as the other errors that are baselined.

> I'd baseline existing (yet undiscovered) errors.

That's why I'm asking to minimize the number of such errors.

ondrejmirtes: Yes, @morozov is totally right. Baseline only works if new code can be written the correct way. So the level should be lowered or special care should be given to manually ignore common errors (that we want to ignore).

lcobucci: I'm solving the issues, so there's no need for us to keep discussing this :smile:

morozov: > Yes, @morozov is totally right.

What's wrong about this kind of discussion?

ondrejmirtes: I was just excited that someone gets the baseline feature as it was designed :D

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
morozov: FWIW, the `@return` annotation [should be inherited](https://docs.phpdoc.org/guides/inheritance.html#methods) according to the spec regardless of `{@inheritDoc}` that is used to [import the description of a parent element](https://docs.phpdoc.org/guides/inheritance.html#the-inheritdoc-tag).

lib/Doctrine/DBAL/Driver/DrizzlePDOMySql/Connection.php
```diff
@@ -10,12 +10,12 @@ class Connection extends PDOConnection
     /**
      * {@inheritdoc}
      */
-    public function quote($value, $type = ParameterType::STRING)
+    public function quote($input, $type = ParameterType::STRING)
```
morozov: I'd keep it generic as `$value` everywhere. Even if a value doesn't come from input but is defined in the code, it needs to be quoted.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
```diff
@@ -76,9 +76,9 @@ public function requiresQueryForServerVersion()
     /**
      * {@inheritdoc}
      */
-    public function prepare($sql)
+    public function prepare($prepareString)
```
morozov: Let's keep it declared as `$sql` everywhere instead.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
```diff
@@ -140,7 +140,7 @@ public function lastInsertId($name = null)
      */
     public function beginTransaction()
     {
-        db2_autocommit($this->conn, DB2_AUTOCOMMIT_OFF);
+        return db2_autocommit($this->conn, DB2_AUTOCOMMIT_OFF) === true;
```
morozov: Why is `=== true` needed. Does it [return](https://www.php.net/db2_autocommit#refsect1-function.db2-autocommit-returnvalues) anything else than a boolean?

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -85,7 +85,13 @@ public function bindValue($param, $value, $type = ParameterType::STRING)
     }
 
     /**
-     * {@inheritdoc}
+     * @param mixed    $column
+     * @param mixed    $variable
+     * @param int      $type
+     * @param int|null $length
+     * @param mixed    $driverOptions
```
morozov: Although it's [documented](https://www.php.net/manual/en/pdostatement.bindparam) as `mixed`, I've never seen it be anything else than `array`. Most likely, it's a documentation bug. Should we declare it as `array`?

lib/Doctrine/DBAL/Platforms/DrizzlePlatform.php
```diff
@@ -353,10 +356,10 @@ public function getListTableForeignKeysSQL($table, $database = null)
     /**
      * {@inheritDoc}
      */
-    public function getListTableIndexesSQL($table, $database = null)
+    public function getListTableIndexesSQL($table, $currentDatabase = null)
```
morozov: Let's keep it as `$database` everywhere. The "current" aspect is irrelevant in the method signature.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -49,17 +51,19 @@ public static function assertValidIdentifier($identifier)
     /**
      * {@inheritDoc}
      */
-    public function getSubstringExpression($value, $position, $length = null)
+    public function getSubstringExpression($value, $from, $length = null)
```
morozov: Let's keep the argument names consistent with `master`:
https://github.com/doctrine/dbal/blob/27285b19369ccb1a40df66bcf91d535a739b746b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php#L738

When reworking the signatures, I tried to use names from the SQL analogs of the platform methods.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
SenseException: As long as the second argument is given. But if it is omitted, the function will return an integer as far as I understand. Is this to be more explicit about the return value?

morozov: Indeed. It's documented as `return @mixed` in the [stub](https://github.com/JetBrains/phpstorm-stubs/blob/f161ec38d8eb053d726d94c1c2ff961a848780fb/ibm_db2/ibm_db2.php).

I'd rather have an assertion instead of comparison. We shouldn't expect anything else than a boolean from a call with the second parameter.

morozov:
@greg0ire could you please reopen this PR? It'll probably take creating `2.10` from `2.10.x`, pushing it upstream, reopening the PR, retargeting it to `2.10.x` and deleting `2.10` back. IIRC, you did something like this for other accidentally closed ones.

morozov:
FWIW, it is important to upgrade PHPStan sooner than later. The reason is that phpstan/phpstan `0.11` [requires](https://packagist.org/packages/phpstan/phpstan#0.11.19) phpstan/phpdoc-parser `^0.3.5` while slevomat/coding-standard `6.2` (currently, `dev-master`) [will require](https://packagist.org/packages/slevomat/coding-standard#dev-master) phpstan/phpdoc-parser `^0.4`, so we won't be able to upgrade due to the dependency conflict.

ondrejmirtes:
Upgrading PHPStan shouldn't be a lot of hard work. You can upgrade immediately if you introduce the baseline (https://medium.com/@ondrejmirtes/phpstans-baseline-feature-lets-you-hold-new-code-to-a-higher-standard-e77d815a5dff) and fix the new found issues later.

morozov:
@ondrejmirtes I thought that we agreed that the baseline won't work in our case (https://github.com/doctrine/dbal/pull/3799#discussion_r362943325).

ondrejmirtes:
I’ve forgot about that, sorry.

greg0ire:
@morozov sorry, I missed your comment :see_no_evil:

morozov:
I've addressed all remaining issues and file a couple of PHPStan bugs. Going to address the issues in @lcobucci's commits that I pointed out to earlier.

lib/Doctrine/DBAL/Driver/PDOStatement.php
morozov: The usage examples in the same manual contain `$driver_options = PDO::SQLSRV_ENCODING_BINARY` so it looks like it's really `mixed`.

morozov (approved):

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1334,6 +1336,8 @@ private function commitAll()
     /**
      * Cancels any database changes done during the current transaction.
      *
+     * @return bool
+     *
```
greg0ire: Why not use `void` here? There is no `return false`, failures are conveyed with exceptions

morozov: Because it implements `Doctrine\DBAL\Driver\Connection::rollBack() : bool`. The interface an the implementation were reworked in https://github.com/doctrine/dbal/pull/3480.

morozov:
Thanks @lcobucci.

lcobucci:
No, thank YOU @morozov :heart: 
You're doing a fantastic job and I'm sorry for not being able to help you more nowadays. Keep on rockin' :metal:

PR #3800, opened 2019-12-30
Refactor FK exception tests

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

<s>Several tests can be greatly simplified by delegating the tear-down tasks to `finally {}`, instead of catching and re-throwing exceptions.

This works fine with `expectException()`, too.</s>

Moved FK exception tests to their own class with proper `setUp()`/`tearDown()`.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/ExceptionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/ExceptionTest.php
@@ -14,7 +14,6 @@ use Doctrine\DBAL\Platforms\SqlitePlatform;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\Tests\DbalFunctionalTestCase;
-use Throwable;
 use const PHP_OS;
 use function array_merge;
 use function assert;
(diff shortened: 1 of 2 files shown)


BenMorel:
Note: the SQL Server failure on Travis seems unrelated to this change:

> Doctrine\Tests\DBAL\Functional\Platform\DefaultExpressionTest::testCurrentTimestamp  
> Failed asserting that two strings are equal.  
> --- Expected  
> +++ Actual  
> @@ @@  
> -'2019-12-30 14:14:15.770000'  
> +'2019-12-30 14:14:15.766667'

Should I force push the previous commit to restart the build?

tests/Doctrine/Tests/DBAL/Functional/ExceptionTest.php
```diff
@@ -87,27 +86,13 @@ public function testForeignKeyConstraintViolationExceptionOnInsert() : void
...
 
-        try {
             $this->connection->insert('owning_table', ['id' => 2, 'constraint_id' => 2]);
-        } catch (Exception\ForeignKeyConstraintViolationException $exception) {
-            $this->tearDownForeignKeyConstraintViolationExceptionTest();
-
-            throw $exception;
-        } catch (Throwable $exception) {
+        } finally {
             $this->tearDownForeignKeyConstraintViolationExceptionTest();
```
morozov: Would it make more sense to extract this test to a separate class and let PHPUnit take care of calling it? Or just make this routine part of the shared teardown? Calling it manually and having flow control logic in a test still smells.

BenMorel: It does smell, but smells less than before! ;-) That's a good point, though. Moved to their own class with proper `setUp()/tearDown()` in da3a944e8451f0e543a622ed27fd4c20badbf7fb.

SenseException (dismissed):

SenseException:
A rebase is probably needed before this can be merged.

BenMorel:
Rebased & squashed!

morozov (approved):

morozov:
Thanks @BenMorel!

PR #3590, opened 2019-06-04
Make classes final where applicable.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | fixes #3585

PR #3801, opened 2019-12-30
Fix phpstan errors

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

A method declared with a `string` return type attempts to return `null`. This is not caught by tests as this method is never called because `SqlitePlatform` does not support exceptions.

This now throws a `LogicException` instead, as this is not expected to be executed.

#### Update

Following the discussion below, this now fixes many more phpstan errors.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1264,6 +1264,8 @@ class Connection implements DriverConnection
     {
         $this->connect();
 
+        assert($this->_conn !== null);
+
         return $this->_conn;
     }
 
--- a/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
@@ -688,7 +688,7 @@ class SqlitePlatform extends AbstractPlatform
      */
     protected function getPostAlterTableIndexForeignKeySQL(TableDiff $diff) : array
     {
-        if (! $diff->fromTable instanceof Table) {
+        if ($diff->fromTable === null) {
             throw new DBALException('Sqlite platform requires for alter table the table diff with reference to original table schema.');
         }
 
@@ -699,7 +699,7 @@ class SqlitePlatform extends AbstractPlatform
             $tableName = $diff->getName($this);
         }
 
-        foreach ($this->getIndexesInAlteredTable($diff) as $index) {
+        foreach ($this->getIndexesInAlteredTable($diff, $diff->fromTable) as $index) {
             if ($index->isPrimary()) {
                 continue;
             }
@@ -823,7 +823,7 @@ class SqlitePlatform extends AbstractPlatform
         }
 
         $fromTable = $diff->fromTable;
-        if (! $fromTable instanceof Table) {
+        if ($fromTable === null) {
             throw new DBALException('Sqlite platform requires for alter table the table diff with reference to original table schema.');
         }
 
@@ -908,7 +908,7 @@ class SqlitePlatform extends AbstractPlatform
         if (! $this->onSchemaAlterTable($diff, $tableSql)) {
             $dataTable = new Table('__temp__' . $table->getName());
 
-            $newTable = new Table($table->getQuotedName($this), $columns, $this->getPrimaryIndexInAlteredTable($diff), [], $this->getForeignKeysInAlteredTable($diff), $table->getOptions());
+            $newTable = new Table($table->getQuotedName($this), $columns, $this->getPrimaryIndexInAlteredTable($diff, $fromTable), [], $this->getForeignKeysInAlteredTable($diff, $fromTable), $table->getOptions());
             $newTable->addOption('alter', true);
 
             $sql = $this->getPreAlterTableIndexForeignKeySQL($diff);
@@ -1016,11 +1016,11 @@ class SqlitePlatform extends AbstractPlatform
     /**
      * @return string[]
      */
-    private function getColumnNamesInAlteredTable(TableDiff $diff) : array
+    private function getColumnNamesInAlteredTable(TableDiff $diff, Table $fromTable) : array
     {
         $columns = [];
 
-        foreach ($diff->fromTable->getColumns() as $columnName => $column) {
+        foreach ($fromTable->getColumns() as $columnName => $column) {
             $columns[strtolower($columnName)] = $column->getName();
         }
 
@@ -1056,10 +1056,10 @@ class SqlitePlatform extends AbstractPlatform
     /**
      * @return Index[]
      */
-    private function getIndexesInAlteredTable(TableDiff $diff) : array
+    private function getIndexesInAlteredTable(TableDiff $diff, Table $fromTable) : array
     {
-        $indexes     = $diff->fromTable->getIndexes();
-        $columnNames = $this->getColumnNamesInAlteredTable($diff);
+        $indexes     = $fromTable->getIndexes();
+        $columnNames = $this->getColumnNamesInAlteredTable($diff, $fromTable);
 
         foreach ($indexes as $key => $index) {
             foreach ($diff->renamedIndexes as $oldIndexName => $renamedIndex) {
@@ -1120,10 +1120,10 @@ class SqlitePlatform extends AbstractPlatform
     /**
      * @return ForeignKeyConstraint[]
      */
-    private function getForeignKeysInAlteredTable(TableDiff $diff) : array
+    private function getForeignKeysInAlteredTable(TableDiff $diff, Table $fromTable) : array
     {
-        $foreignKeys = $diff->fromTable->getForeignKeys();
-        $columnNames = $this->getColumnNamesInAlteredTable($diff);
+        $foreignKeys = $fromTable->getForeignKeys();
+        $columnNames = $this->getColumnNamesInAlteredTable($diff, $fromTable);
 
         foreach ($foreignKeys as $key => $constraint) {
             $changed      = false;
@@ -1180,11 +1180,11 @@ class SqlitePlatform extends AbstractPlatform
     /**
      * @return Index[]
      */
-    private function getPrimaryIndexInAlteredTable(TableDiff $diff) : array
+    private function getPrimaryIndexInAlteredTable(TableDiff $diff, Table $fromTable) : array
     {
         $primaryIndex = [];
 
-        foreach ($this->getIndexesInAlteredTable($diff) as $index) {
+        foreach ($this->getIndexesInAlteredTable($diff, $fromTable) as $index) {
             if (! $index->isPrimary()) {
                 continue;
             }
--- a/lib/Doctrine/DBAL/Portability/Statement.php
+++ b/lib/Doctrine/DBAL/Portability/Statement.php
@@ -26,7 +26,7 @@ final class Statement implements IteratorAggregate, DriverStatement
     /** @var DriverStatement|ResultStatement */
     private $stmt;
 
-    /** @var int */
+    /** @var int|null */
     private $case;
 
     /** @var int */
@@ -204,6 +204,7 @@ final class Statement implements IteratorAggregate, DriverStatement
         }
 
         if ($fixCase) {
+            assert($this->case !== null);
             $row = array_change_key_case($row, $this->case);
         }
 
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -393,9 +393,9 @@ class QueryBuilder
      * Gets the maximum number of results the query object was set to retrieve (the "limit").
      * Returns NULL if {@link setMaxResults} was not applied to this query builder.
      *
-     * @return int The maximum number of results.
+     * @return int|null The maximum number of results.
      */
-    public function getMaxResults() : int
+    public function getMaxResults() : ?int
     {
         return $this->maxResults;
     }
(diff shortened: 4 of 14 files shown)


tests/Doctrine/Tests/DBAL/Platforms/SqlitePlatformTest.php
```diff
@@ -271,7 +272,7 @@ public function testGeneratesConstraintCreationSql() : void
 
     public function getGenerateForeignKeySql() : string
     {
-        return null;
+        throw new LogicException('Foreign key constraints are not yet supported for SQLite.');
```
morozov: > This is not caught by tests as this method is never called because `SqlitePlatform` does not support exceptions.

I'm curious why this is not flagged by PHPStan. I'm not a fan of fixing issues like this one by one. There may be more issues like this including the production code.

As for the code change, I'd rather fail the test explicitly with `$this->fail()`.

BenMorel: > I'm curious why this is not flagged by PHPStan.

I've seen your work related to phpstan earlier. Did you run it on `tests`, not just `lib`?

> As for the code change, I'd rather fail the test explicitly with $this->fail().

I'll happily change it if you think it's better, but be advised that it makes it harder to understand for static analysis, as the method does not return or throw anything. The throw is hidden in `fail()`.

morozov: > I've seen your work related to phpstan earlier. Did you run it on `tests`, not just `lib`?

It should run on `lib` and `tests`:

https://github.com/doctrine/dbal/blob/644531f360e48d4f222ccfd726a07aed60200b3e/phpstan.neon.dist#L4-L5

> […] it makes it harder to understand for static analysis, as the method does not return or throw anything.

AFAIK the PHPStan extension for PHPUnit is aware of methods like that.

> I'll happily change […]

I'd like to make sure we understand the root cause of why it's not detected before making any change.

BenMorel: > I'd like to make sure we understand the root cause of why it's not detected before making any change.

I'll leave it to you then, I'm not familiar enough with phpstan!

BenMorel: > As for the code change, I'd rather fail the test explicitly with `$this->fail()`.

Done, FWIW.

morozov: Scrutinizer sucks.

> […] the function will implicitly return `null` […]

It is not true. The return will never happen.

SenseException: > I'd like to make sure we understand the root cause of why it's not detected before making any change.

@morozov Are you going for the root cause in this PR or will you create a separate one?

morozov: I'd like to address the root cause. Apparently, PHPStan ignores this issue because of:
https://github.com/doctrine/dbal/blob/f9b3e7908f548a96d4ec62a30c8cb9fed7f95c7f/phpstan.neon.dist#L38-L39

Instead of being applied to a specific class method, it's applied to the entire codebase. Removing this line makes PHPStan find the issue as well as 43 others:
```
$ phpstan a tests/Doctrine/Tests/DBAL/Platforms/SqlitePlatformTest.php

 ------ ------------------------------------------------------------------------------------- 
  Line   SqlitePlatformTest.php                                                               
 ------ ------------------------------------------------------------------------------------- 
  274    Method Doctrine\Tests\DBAL\Platforms\SqlitePlatformTest::getGenerateForeignKeySql()  
         should return string but returns null.                                               
 ------ ------------------------------------------------------------------------------------- 
```
@BenMorel please try to rework the PHPStan configuration so that it's only applied to where it's supposed to and try to fix the rest of the issues.

morozov: The problem is probably because the `|` in the `string|null` type is not escaped, so it's interpreted by the regex engine instead of being used as a literal. We should escape all occurrences of `|` where it's part of the expected error message like it's done it:
https://github.com/doctrine/dbal/blob/f9b3e7908f548a96d4ec62a30c8cb9fed7f95c7f/phpstan.neon.dist#L16-L17

BenMorel: First of all, when I run `phpstan analyze` locally, I get several times this warning:

> Access to undefined constant PDO::PGSQL_ATTR_DISABLE_PREPARES

How do I get rid of it? Should I report a bug with phpstan?

BenMorel: > The problem is probably because the | in the string|null type is not escaped, so it's interpreted by the regex engine instead of being used as a literal.

Good catch! Fixed this in 8f286d31c5b48d58ae71d47a02e3d4fafd279ef5, and now phpstan reports 42 errors. Working on fixing them.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1264,6 +1264,8 @@ public function getWrappedConnection() : DriverConnection
     {
         $this->connect();
 
+        assert($this->_conn !== null);
```
BenMorel: I'm not sure what your policy is in this case: `assert()`? `throw new LogicException()?`

There are several other places, as you'll see in the diff, where we know that the variable will not be null, but phpstan cannot dig deep enough to figure it out by itself.

lib/Doctrine/DBAL/Portability/Statement.php
```diff
@@ -26,7 +26,7 @@ class Statement implements IteratorAggregate, DriverStatement
     /** @var DriverStatement|ResultStatement */
     private $stmt;
 
-    /** @var int */
+    /** @var int|null */
```
BenMorel: This variable [is expected to be nullable](https://github.com/doctrine/dbal/blob/b6e3aa3555e81d3168b4ba96db62ff3f31b93690/lib/Doctrine/DBAL/Portability/Statement.php#L121), however `array_change_key_case()` below expects it to be `int`, never `null`. Hence [another `assert()` below](https://github.com/doctrine/dbal/pull/3801/files#diff-87c7f450dc8a06beead2da33fd34b210R179).

lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
```diff
@@ -53,14 +53,6 @@ protected function _getPortableViewDefinition(array $view) : View
...
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    protected function _getPortableTableDefinition(array $table) : string
-    {
-        return array_shift($table);
-    }
-
```
BenMorel: Removed this one as the function body is the same as the parent class.

BenMorel:
All phpstan errors fixed now! @morozov, waiting for your feedback on assertions, in particular.

lib/Doctrine/DBAL/Portability/Statement.php
morozov: This also seems to be the case for 

https://github.com/doctrine/dbal/blob/88f7feae9a74a1f804956470e75284804944d240/lib/Doctrine/DBAL/Portability/Connection.php#L38-L39

lib/Doctrine/DBAL/Connection.php
morozov: The assertions should be added only if the issue reported by PHPStan is in fact impossible but this is not obvious from the type system. On the other hand, if an exception is added, there should be a test case that covers it being thrown.

In this case, the assertion looks legit. While the API could use some improvement. There should be a method that always _returns_ a connection, not sets it on the object (out of scope).

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -1020,6 +1021,8 @@ private function getColumnNamesInAlteredTable(TableDiff $diff) : array
     {
         $columns = [];
 
+        assert($diff->fromTable !== null);
```
morozov: Instead of adding an assertion against the object state, if we're in a private method, would it make sense to pass the value explicitly and enforce its type in the method signature?

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
```diff
@@ -1058,6 +1061,8 @@ private function getColumnNamesInAlteredTable(TableDiff $diff) : array
      */
     private function getIndexesInAlteredTable(TableDiff $diff) : array
     {
+        assert($diff->fromTable !== null);
```
morozov: Please see if replacing the property access with a method argument would make the API more explicit and eliminate the need for the assertion.

lib/Doctrine/DBAL/Portability/Statement.php
```diff
@@ -176,6 +176,7 @@ protected function fixRow($row, bool $iterateRow, bool $fixCase)
         }
 
         if ($fixCase) {
+            assert($this->case !== null);
```
morozov: Here we could use the same approach as with private methods above, however, this one is protected. We should see what prevents this class from being finalized (https://github.com/doctrine/dbal/pull/3590).

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -393,9 +393,9 @@ public function setMaxResults(int $maxResults) : self
      * Gets the maximum number of results the query object was set to retrieve (the "limit").
      * Returns NULL if {@link setMaxResults} was not applied to this query builder.
      *
-     * @return int The maximum number of results.
+     * @return int|null The maximum number of results.
      */
-    public function getMaxResults() : int
+    public function getMaxResults() : ?int
```
morozov: :+1:

morozov (commented):
@BenMorel I reviewed this partially. Please see my recommendations on using assertions and see if the rest of the assertions follow them.

lib/Doctrine/DBAL/Portability/Statement.php
BenMorel: Indeed, it's null by default (not set in a constructor). And the getter reflects this. Fixed in ff9831475dca77af2e150efd8ad473ad4922b7ce.

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
BenMorel: I think this is doable, as the calling methods seem to have checks for `$fromTable` being not null, for example:

https://github.com/doctrine/dbal/blob/69cd77e007466b8dd9b671c106c419fee5e2d877/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php#L825-L828

This means passing as a parameter a property of the object itself, though, and I feel like it's a smell. Are you sure you want to try this route?

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
BenMorel: Please see above. Doable at first glance. But is it a good practice?

lib/Doctrine/DBAL/Platforms/SqlitePlatform.php
morozov: It is a smell (probably, the method doesn't belong to the class or should be made static) but assertions are also smell.

BenMorel: Alright then! Done in 4815eb8b33777b1dec2666247146c43643d572c9. Not a big deal, after all.

tests/Doctrine/Tests/DBAL/Platforms/SqlitePlatformTest.php
morozov: > Should I report a bug with phpstan?

Just for a future reference: this is a current limitation of PHPStan and the design flaw of PDO. Some PDO extensions define constants on the already existing `PDO` class. If `pdo` is loaded but `pdo_pgsql` isn't, PHPStan cannot add the missing constants from the `pdo_pgsql` stubs because it uses runtime reflection instead of static.

lib/Doctrine/DBAL/Portability/Statement.php
BenMorel: What should we do here then? There are only 4 `assert()` left in this PR; do you want to find a solution to get rid of each of them individually, or is it OK if they stay?

morozov: Let's leave those assertions for now. Please rebase and squash.

morozov:
@BenMorel once this is done, could you please also take a look at https://github.com/doctrine/dbal/blob/f9b3e7908f548a96d4ec62a30c8cb9fed7f95c7f/phpstan.neon.dist#L35-L36

If the `|` is properly escaped, PHPStan produces 14 more errors that need to be fixed or properly suppressed.

lib/Doctrine/DBAL/Portability/Statement.php
BenMorel: Rebased & squashed!

BenMorel:
> @BenMorel once this is done, could you please also take a look at (...)

I'll try to have a look tomorrow!

morozov (approved):

morozov:
Thanks @BenMorel. Great work :+1:

PR #3770, opened 2019-12-05
Add $statement parameter to PDOConnection::query()

nikic:
|      Q       |   A
|------------- | -----------
| Type         | bug-ish
| BC Break     | no
| Fixed issues | 

#### Summary

PHP 8 updates the signature to `PDO::query(string $statement)`, so change `PDOConnection` to be compatible with that (but also older versions).

This doesn't seem relevant for version 3.x anymore (which does not extend PDO), but this should allow testing current Symfony and Laravel versions under PHP 8.

PR #3803, opened 2019-12-30
Backport PDO-related changes from master to 3.0.x

### LINKED ISSUE #3798: Research the possibility to backport PDO-related changes from master to 2.10
Instead of introducing the breaking change suggested in #3770, we may try backporting a few relevant PRs from `master` to `2.x` and release `3.0` that not only introduces the needed breaking changes but also fixes the root cause of the problem and releases some goodies from `master`.

In addition to the changes in #3803, the following needs to be done:

- [x] Fix #3802.

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

Closes #3770.
Closes #3798.

This is a PoC that represents the minimal required changes to support PHP 8 by the DBAL 2 codebase.

**TODO:**
- [x] Backport #3241 (29b5e7f435aa16cb3764281d18ab34a69f577087): The `dbal:import` CLI command is removed since it relies on `PDO::nextRowset()`.
- [x] Backport #2958 (581467fa2d1213dceec536dc2b01103ca2153388): `Doctrine\DBAL\Driver\PDOStatement` no longer extends `PDOStatement`.
- [x] Backport #3080 (7a5ed10255a01b6481617d9b01a8ac9c977213cf):  `Doctrine\DBAL\Driver\PDOConnection` no longer extends `PDO`.
- [x] Backport #3061 (dc47f569f8f3eeadb8efd1b2cf36caafe015d999): Bugfixes in `pdo_sqlsrv` driver.
- [x] Backport #3238 (f9dcb65212d92db4b77ba0f92a08362a23cc716f): PHPStan adjustments for the above.
- [x] Backport #3515 (c5d298c56d93d83e3352fb648f70bbec4b07fcb3): Not strictly required but a nice outcome of the above.
- [x] Backport #3548 (735b5a078ff3447476ac01fc0fc28c8c0f65b8b9): User-provided PDO connection can be no longer supported.
- [x] Make sure that this PR cleanly merges back to master indicating that there are no controversial changes being mistakenly introduced (e.g. due to an invalid conflict resolution).


morozov:
@doctrine/doctrinecore what do you think about branching `3.0` from `2.10` and merging this? This way we'll enable the upstream project to test with PHP 8 using DBAL 3.0.x-dev and could release around the time when PHP 8 goes beta.

nicolas-grekas:
I think that's a good direction. W should ensure that all possible deprecation-free are reported in 2.10.

And I'm not sure about return types for v3. They might be for v4, see how we're doing on the topic on Symfony or on the [FIG](https://www.php-fig.org/bylaws/psr-evolution/)

morozov:
> And I'm not sure about return types for v3. They might be for v4 […]

Yeah. Dragging them (as well as other scalar/strict type changes) into v3 would be too much. I do want to remove the deprecated APIs though but not necessarily all of them. If something appears to require non-trivial handling on the consumer's side, we can agree to postpone some of them by v4.

UPGRADE.md
```diff
@@ -1,3 +1,51 @@
+# Upgrade to 3.0
+
+## BC BREAK User-provided `PDO` instance is no longer supported
+
+In order to share the same `PDO` instances between DBAL and other components, initialize the connection in DBAL and access it using `Connection::getWrappedConnection()->getWrappedConnection()`.
+
+## BC BREAK: the PDO symbols are no longer part of the DBAL API
+
+1. The support of `PDO::PARAM_*`, `PDO::FETCH_*`, `PDO::CASE_*` and `PDO::PARAM_INPUT_OUTPUT` constants in the DBAL API is removed.
```
beberlei: Can we word this differently? First you could mention that `ParameterType` is the alternative, second when users continnue using `PDO` constants, then they will still work, because we use the same values.

This is something we could a.) document on ParameterType to be true b.) add a test for.

beberlei (approved):
This looks good.

I tested the BC break potential with ORM 2.7 and it is extremely minimal. Only some of the manual mocks need to be changed in the tests, the actual lib code just keeps on working. Patch for testing: https://gist.github.com/beberlei/28f1bdd3da76d825d2f9832193beaa3d

UPGRADE.md
morozov: > Can we word this differently? First you could mention that ParameterType is the alternative […]

Will do. Should we do it here or we can file a separate ticket? I'm not sure how detailed it should be described in the upgrade notes. A blog post or a migration guide looks like a more proper place for such information.

> This is something we could a.) document on ParameterType to be true b.) add a test for.

I'd be careful here. The values of the DBAL constants are left as the _current_ values of the PDO constants to avoid unnecessary disruption but it doesn't mean they have to be equal. The PDO constants are not supported which is clearly stated in this very sentence.

> second when users continue using `PDO` constants, then they will still work, because we use the same values.

No, this is no longer supported. If users do so, they are on their own. If PHP decides to change the values of PDO or introduces a new constant whose value collides with the value of ours, we won't be able to rule it out.

morozov:
> Only some of the manual mocks need to be changed in the tests […]

This is why I advocate for using PHPUnit-managed mocks instead of the hard-coded ones. We got rid of all such mocks in DBAL as you can see from the patch.

morozov:
I'm going to merge this without the suggested changes to UPGRADE.md in order to facilitate testing against PHP 8 on Travis (#3814).

PR #3808, opened 2020-01-03
Removed the OCI8Connection::getExecuteMode() method

morozov:
The existing relationship between the connection and its statement violate the ISP principle. Instead of having access only to the execution mode of the connection, statements have access to the entire connection API.

Having a method which is not defined in the driver connection interface makes it impossible to mark the method `final` (#3590).

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

### DIFF
--- a/lib/Doctrine/DBAL/Driver/OCI8/ExecutionMode.php
+++ b/lib/Doctrine/DBAL/Driver/OCI8/ExecutionMode.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Driver\OCI8;
+
+/**
+ * Encapsulates the execution mode that is shared between the connection and its statements.
+ *
+ * @internal This class is not covered by the backward compatibility promise
+ */
+final class ExecutionMode
+{
+    /** @var bool */
+    private $isAutoCommitEnabled = true;
+
+    public function enableAutoCommit() : void
+    {
+        $this->isAutoCommitEnabled = true;
+    }
+
+    public function disableAutoCommit() : void
+    {
+        $this->isAutoCommitEnabled = false;
+    }
+
+    public function isAutoCommitEnabled() : bool
+    {
+        return $this->isAutoCommitEnabled;
+    }
+}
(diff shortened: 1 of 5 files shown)


lib/Doctrine/DBAL/Driver/OCI8/ExecutionMode.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Driver\OCI8;
+
+/**
+ * Encapsulates the execution mode that is shared between the connection and its statements.
```
morozov: This will be marked `@internal`.

morozov:
The ContinuousPHP currently fails due to the issue being addressed by https://github.com/doctrine/dbal/pull/3809.

beberlei (dismissed):

lib/Doctrine/DBAL/Driver/OCI8/ExecutionMode.php
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @internal This class is not covered by the backward compatibility promise
+ */
+final class ExecutionMode
+{
+    /** @var bool */
+    private $isAutoCommitEnabled = true;
```
Ocramius: Seems to always be `true`?

Ocramius: Not sure why this is introduced, if the mode is hardcoded from here on...

Ocramius (changes requested):

lib/Doctrine/DBAL/Driver/OCI8/ExecutionMode.php
morozov: > Seems to always be `true`?

What do you mean? `disableAutoCommit()` called from within `beginTransaction()` sets it to `false` unless I broke something.

Ocramius: Yep, didn't read well enough, sorry! :+1:

Ocramius (approved):

Issue #3802, opened 2019-12-30
OCI8StatementTest fails on PHP 8


morozov:
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10, 3.0-dev

```
$ phpunit tests/Doctrine/Tests/DBAL/Driver/OCI8/OCI8StatementTest.php
PHPUnit 8.4.1 by Sebastian Bergmann and contributors.

FF...

Time: 64 ms, Memory: 8.00 MB

There were 2 failures:

1) Doctrine\Tests\DBAL\Driver\OCI8\OCI8StatementTest::testExecute with data set #0 (array('test', null, 'value'))
Failed asserting that exception of type "TypeError" matches expected exception "Doctrine\DBAL\Driver\OCI8\OCI8Exception". Message was: "oci_execute() expects parameter 1 to be resource, null given" at
lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php:396
tests/Doctrine/Tests/DBAL/Driver/OCI8/OCI8StatementTest.php:81
.

2) Doctrine\Tests\DBAL\Driver\OCI8\OCI8StatementTest::testExecute with data set #1 (array(null, 'test', 'value'))
Failed asserting that exception of type "TypeError" matches expected exception "Doctrine\DBAL\Driver\OCI8\OCI8Exception". Message was: "oci_execute() expects parameter 1 to be resource, null given" at
lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php:396
tests/Doctrine/Tests/DBAL/Driver/OCI8/OCI8StatementTest.php:81
.
```

morozov:
Closed via #3820.

PR #3809, opened 2020-01-03
Dropped handling of one-based numeric arrays of parameters in Statement::execute()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

The logic of handling one-based arrays existed in `OCI8Statement` (57a97eba01156c8ad02debae28620e72217b46bc) and `SQLSrvStatement` (#110) since their initial implementation without any explanation of why it was needed. Apart from that, the APIs with semantics implicitly defined by input are often more error-prone.

This behavior is not portable. Consider the following test:
```php
/**
 * @dataProvider paramsProvider
 */
public function testHasZeroKey(array $params) : void
{
    $sql  = $this->connection->getDatabasePlatform()->getDummySelectSQL('?, ?');
    $stmt = $this->connection->prepare($sql);
    $stmt->execute($params);

    self::assertEquals([1, 2], $stmt->fetch(FetchMode::NUMERIC));
}

public static function paramsProvider() : iterable
{
    return [
        'zero-based' => [
            [0 => 1, 1 => 2],
        ],
        'one-based' => [
            [1 => 1, 2 => 2],
        ],
        'arbitrary-keys' =>
        [
            [28 => 1, 71 => 2],
        ],
    ];
}
```
Currently, it produces the following results:
```
RUNNING TESTS WITH CONFIG ibm-db2.phpunit.xml

EEE (for some reason, the test doesn't work here)                   3 / 3 (100%)

RUNNING TESTS WITH CONFIG mysqli.phpunit.xml

...                                                                 3 / 3 (100%)

RUNNING TESTS WITH CONFIG oci8.phpunit.xml

..E                                                                 3 / 3 (100%)

RUNNING TESTS WITH CONFIG pdo-mysql.phpunit.xml

.EE                                                                 3 / 3 (100%)

RUNNING TESTS WITH CONFIG pdo-oci.phpunit.xml

.EE                                                                 3 / 3 (100%)

RUNNING TESTS WITH CONFIG pdo-pgsql.phpunit.xml

.EE                                                                 3 / 3 (100%)

RUNNING TESTS WITH CONFIG pdo-sqlite.phpunit.xml

.EE                                                                 3 / 3 (100%)

RUNNING TESTS WITH CONFIG pdo-sqlsrv.phpunit.xml

.EE                                                                 3 / 3 (100%)

RUNNING TESTS WITH CONFIG sqlsrv.phpunit.xml

..E                                                                 3 / 3 (100%)
```

As you can see, the 2nd test passes only on `sqlsrv`, `oci8` (where the logic being removed is explicitly implemented) and `mysqli` (which doesn't care about the keys at all because it binds parameters using a `foreach` loop).

Additionally, the test that covers the logic implemented in `OCI8Statement` currently fails on PHP 8 (#3802) and, if kept, would have to be reworked for #3808.

In the past, the test was the source of extra maintenance for no real reason: #3642, #3369, #2854, #3507, #3552.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade to 3.0
 
+## BC BREAK: Dropped handling of one-based numeric arrays of parameters in `Statement::execute()`
+
+The statement implementations no longer detect whether `$params` is a zero- or one-based array. A zero-based numeric array is expected.
+
 ## BC BREAK: `ServerInfoAwareConnection::requiresQueryForServerVersion()` is removed.
 
 The `ServerInfoAwareConnection::requiresQueryForServerVersion()` method has been removed as an implementation detail which is the same for almost all supported drivers.
--- a/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
+++ b/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
@@ -351,10 +351,8 @@ class OCI8Statement implements IteratorAggregate, Statement
     public function execute(?array $params = null) : void
     {
         if ($params) {
-            $hasZeroIndex = array_key_exists(0, $params);
-
             foreach ($params as $key => $val) {
-                if ($hasZeroIndex && is_int($key)) {
+                if (is_int($key)) {
                     $param = $key + 1;
                 } else {
                     $param = $key;
--- a/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
+++ b/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
@@ -204,10 +204,8 @@ final class SQLSrvStatement implements IteratorAggregate, Statement
     public function execute(?array $params = null) : void
     {
         if ($params) {
-            $hasZeroIndex = array_key_exists(0, $params);
-
             foreach ($params as $key => $val) {
-                if ($hasZeroIndex && is_int($key)) {
+                if (is_int($key)) {
                     $this->bindValue($key + 1, $val);
                 } else {
                     $this->bindValue($key, $val);
--- a/lib/Doctrine/DBAL/Driver/Statement.php
+++ b/lib/Doctrine/DBAL/Driver/Statement.php
@@ -68,7 +68,7 @@ interface Statement extends ResultStatement
      * if any, of their associated parameter markers or pass an array of input-only
      * parameter values.
      *
-     * @param mixed[]|null $params An array of values with as many elements as there are
+     * @param mixed[]|null $params A numeric array of values with as many elements as there are
      *                             bound parameters in the SQL statement being executed.
      *
      * @throws DriverException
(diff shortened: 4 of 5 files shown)


beberlei:
@morozov I believe the reason this exists is `->bindParam(1, ...)->bindParam(2, ...)` in QueryBuilder.

morozov:
@beberlei yes, this logic might have been introduced in some drivers because it was needed at the time of introduction but it's not currently needed. `Statement::bindParam()` and `Statement::execute($params)` are different API from the DBAL perspective and the perspective of the most of the drivers (some may implement one via the other but it's an implementation detail). The functional tests like `DataAccessTest::testPrepareWithExecuteParams()` and `::testPrepareWith*()` that still pass prove that.

Additionally, as the test results above show, this behavior is not portable and cannot be relied upon. It works as the test expects only in the drivers that artificially implement this behavior and `mysqli` (where the keys are simply ignored).

beberlei:
I have to honestly say that I am not to deep to give a good assessment, but the fact that this is not portable across different connections is reason enough to get rid of it. My problem is that this is extremely "hidden" for users in a way that it could lead to sutble breaks if someone would rely on it.

morozov:
> My problem is that this is extremely "hidden" for users in a way that it could lead to subtle breaks if someone would rely on it.

This is another reason to get rid of this hidden gem in a major release.

beberlei (dismissed):

morozov:
@beberlei could you reapprove please? I had to resolve a conflict with a just merged PR.

SenseException (changes requested):
These changes also seem to be in sync with the QueryBuilder documentation, where 0 is used for the first parameter. It doesn't though with the Data Retrieval chapter. To prevent the communication of mixed usages, I suggest to adapt the parameter indexes to start with 0: https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/data-retrieval-and-manipulation.html#dynamic-parameters-and-prepared-statements

> OCI8/Oracle only supports named parameters, but Doctrine implements a client side parser to allow positional parameters also.

Could this be somehow related to the zero-index decision?

morozov:
> To prevent the communication of mixed usages, I suggest to adapt the parameter indexes to start with 0

@SenseException, which exactly method do you mean? As pointed out int https://github.com/doctrine/dbal/pull/3809#issuecomment-570781608, 

> `Statement::bindParam()` and `Statement::execute($params)` are different APIs

The former will accept 1-based keys (this did and does work for all DBAL drivers and is true for all underlying drivers that support binding a single parameter natively ([PDO](https://www.php.net/manual/en/pdostatement.bindparam#refsect1-pdostatement.bindparam-parameters), [`ibm_db2`](https://www.php.net/manual/en/function.db2-bind-param.php))), the latter will accept the 0-based ones.

The chapter you're referring to doesn't even mention passing parameters to `Statement::execute()`.

morozov:
@SenseException if you request changes on a PR, please make sure you can participate in the discussion. This PR is a dependency for #3808 and #3802, so by not replying, you're blocking progress on three issues, not just one.

PR #3812, opened 2020-01-05
Fix DebugStack#queries docblock type

ostrolucky:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues |

### DIFF
--- a/lib/Doctrine/DBAL/Logging/DebugStack.php
+++ b/lib/Doctrine/DBAL/Logging/DebugStack.php
@@ -12,7 +12,7 @@ class DebugStack implements SQLLogger
     /**
      * Executed SQL queries.
      *
-     * @var mixed[][]
+     * @var array<int, array<string, mixed>>
      */
     public $queries = [];
 


lib/Doctrine/DBAL/Logging/DebugStack.php
```diff
@@ -12,7 +12,7 @@ class DebugStack implements SQLLogger
     /**
      * Executed SQL queries.
      *
-     * @var mixed[][]
+     * @var mixed[]|mixed[][]
```
morozov: What is the problem being fixed? Doesn't `mixed[]` include `mixed[][]`?

ostrolucky: Different people/tools think different. We need to be more clear about this https://scrutinizer-ci.com/g/doctrine/DoctrineBundle/inspections/5c5ab904-a71a-4a2c-8cfe-2aefd2b4363a/issues/files/Tests/ProfilerTest.php?status=new&orderField=path&order=asc&honorSelectedPaths=0

greg0ire: :thinking: I think a proper fix for this error would be:

```suggestion
     * @var array<int,array<string,string>>
```

If you use `mixed[]|mixed[][]`, you're only really sure it's a `mixed[]`

morozov: > It seems like `array(array('sql' => 'SE...l, 'executionMS' => 1))` of type `array<integer,array<stri...onMS\":\"integer\"}>"}>` is incompatible with the declared type `array<integer,array<integer,*>>` of property `$queries`.

It looks like a bug in Scrutinizer. Let's remove the matching definitions and compare the difference: `array<string>` is incompatible with `array<integer,*>`. The type `string` is definitely compatible with `*` while I'm not sure what the omitted key type means (I thought it implies int).

Could we just replace `mixed[][]` with `array<int,array<string,mixed>>`? It's more specific and may make Scrutinizer happier.

I cannot think of a case when a value of the top-level array is not an array which necessitates adding `mixed[]`.

morozov:
On an unrelated note, this is a case where I share the @beberlei's sentiment about our CI being too hard to deal with. But it's not because of CI but because some of our types expressed via arrays of arrays of arrays are hard to comprehend. In this specific case, we could easily use an array of objects that represent a stack frame and has its properties explicitly defined. This would be beneficial from type safety, type simplicity and memory footprint ([1](https://steemit.com/php/@crell/php-use-associative-arrays-basically-never), [2](https://youtu.be/JBWgvUrb-q8?t=943)).

greg0ire (approved):

morozov (approved):

PR #3816, opened 2020-01-10
Fix documented return type

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

`diffTable()` returns `null` if no differences, not `false`.


morozov (approved):

morozov:
Thanks, @BenMorel!

Issue #3818, opened 2020-01-12
Finalize remaining Connection and Statement classes


morozov:
The following classes could not be finalized as is in #3590:

- [ ] `MysqliConnection` because of its `getWrappedResourceHandle()` which is not used anywhere and is not part of any API convention (see #3534 for a similar feature request).
- [x] `OCI8Connection` because of its `getExecuteMode()` which is used by `OCI8Statement::execute()`. It indicates a violation of the ISP principle (the statement should be given an object that represents the execution mode, not the entire connection). Fixed in https://github.com/doctrine/dbal/pull/3808.
- [x] `OCI8Statement` because of its _static_ `convertPositionalToNamedPlaceholders()` which is `public` only because it's tested directly which indicates a violation of the SRP. It could be extracted to a utility class: https://github.com/doctrine/dbal/blob/f37a88ed99a342080650b969a4dbebea21ca6ab8/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php#L130
   Fixed in https://github.com/doctrine/dbal/pull/3820.
- [x] `PDOConnection` because it has child classes and the `createStatement()` method.
- [x] `PDOSqlsrv\Connection` because of its `createStatement()`.
- [x] `PDOStatement` because it has child classes.

PR #3820, opened 2020-01-12
Made the OCI8Statement class final

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | #3818
| Depends on | #3809

The `OCI8Statement::convertPositionalToNamedPlaceholders()` method is extracted to an internal utility class in order to enable making the class final. The corresponding test methods are extracted from `UtilTest` and `OCI8StatementTest` to the dedicated test class. 

`OCI8StatementTest` has been removed since it doesn't contain any more tests which is a good sign: we're not testing any driver-specific implementation details on the driver class itself.


UPGRADE.md
```diff
@@ -1,5 +1,13 @@
 # Upgrade to 3.0
 
+## BC BREAK: `OCI8Statement::convertPositionalToNamedPlaceholders()` is removed.
+
+The `OCI8Statement::convertPositionalToNamedPlaceholders()` method has been extracted to an internal utility class.
+
+## BC BREAK: Dropped handling of one-based numeric arrays of parameters in `Statement::execute()`
+
+The statement implementations no longer detect whether `$params` is a zero- or one-based array. A zero-based numeric array is expected.
```
Ocramius: We should probably start adding `@psalm-param list<T>` for these

Ocramius (approved):
LGTM 👍

UPGRADE.md
morozov: Should we add Psalm to the build pipeline first?

Ocramius: Sure thing - it's a big amount of work to get it to "like" our codebase though 😅

PR #3821, opened 2020-01-14
[pg] fix getting table information if search_path contains escaped schema name

linniksa:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

I got problem with doctrine migrations with postresql database: 
```There is no column with name 'version' on table 'migration_versions'.```

after research i found that error happiness if `search_path` contains escaped schema name (eg ``` set search_path = '123', public;```)

this patch fix my problem, also simplifies the code and use native processing of search_path from postgres (no need to handle "$user" anymore))

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
@@ -335,6 +335,18 @@ class PostgreSqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
         self::assertFalse($comparator->diffTable($offlineTable, $onlineTable));
     }
 
+    public function testListTableDetailsWhenCurrentSchemaNameQuoted() : void
+    {
+        $this->connection->exec('CREATE SCHEMA "001_test"');
+        $this->connection->exec('SET search_path TO "001_test"');
+
+        try {
+            $this->testListQuotedTable();
+        } finally {
+            $this->connection->close();
+        }
+    }
+
     public function testListTablesExcludesViews() : void
     {
         $this->createTestTable('list_tables_excludes_views');
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -367,7 +367,7 @@ SQL
             [$schema, $table] = explode('.', $table);
             $schema           = $this->quoteStringLiteral($schema);
         } else {
-            $schema = "ANY(string_to_array((select replace(replace(setting,'\"\$user\"',user),' ','') from pg_catalog.pg_settings where name = 'search_path'),','))";
+            $schema = 'ANY(current_schemas(false))';
         }
 
         $table = new Identifier($table);

### CODE REVIEW

**Comment by morozov** [comment]
@linniksa please provide a functional test that would fail without the fix.

**Comment by linniksa** [comment]
@morozov done

File: tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php [hunk 1]
```diff
@@ -348,6 +348,25 @@ public function testListQuotedTable() : void
         self::assertNull($comparator->diffTable($offlineTable, $onlineTable));
     }
 
+    public function testListTableDetailsWithNumberSchema() : void
+    {
+        $this->connection->exec('CREATE SCHEMA IF NOT EXISTS "001_test"');
```
morozov: Would it be possible to rework the test to use the DBAL API (e.g. `(new Schema())->toSQL($platfrom)`) instead of platform-specific SQL and test it across all supported platforms?

We want to make sure that all the supported features are consistently implemented or at least tested across all supported platforms. IIRC, there was a similar issue reported for MySQL (something similar to https://github.com/doctrine/dbal/pull/2772).

tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
```diff
@@ -348,6 +348,25 @@ public function testListQuotedTable() : void
...
+
+        $offlineTable = new Schema\Table('"001_test".test');
+        $offlineTable->addColumn('id', 'integer');
+        $this->schemaManager->dropAndCreateTable($offlineTable);
+
+        $onlineTable = $this->connection->transactional(function () : Table {
+            $this->connection->exec('SET LOCAL search_path = "001_test"');
```
morozov: The Schema Manager API supports the specification of a database in which the table should be looked up. 

https://github.com/doctrine/dbal/blob/7ff15cbcc12bc7e66e9bc56cf3f3a60b58294ded/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php#L156

Would it make sense to implement the support for this parameter for PostgreSQL instead of having to modify the global state?

morozov: > and `PostgreSqlPlatform` actually ignores them

Correct. Why not implement the handling of this parameter in DBAL instead of having to use platform-specific SQL?

> '001_test' is schema not database

Correct. In DBAL, `$database` means schema or database depending on the platform. In the case of `listTableColumns()`, a table is looked up within a schema.

morozov:
I'm fine with merging the patch as is and making all suggested API improvements later but I'd like to get someone who has more experience with PostgreSQL.

morozov (dismissed):

tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
linniksa: probably i have some misunderstanding, but i try to explain my point of view:

> Correct. Why not implement the handling of this parameter in DBAL instead of having to use platform-specific SQL?

because this is not happines in real life: https://github.com/doctrine/migrations/blob/b13df49bd2170a6646469081a7e9713650e1bec5/lib/Doctrine/Migrations/Tracking/TableUpdater.php#L84

in this code doctrine migrations try to find definition of same table used when they call 'select * from migrations_version'

in our cases we use multiple schemas (as i can see in DBAL it calls namespaces) for account segregation for one application which means that every schema in one database has same table structure.

Reason why we not use one database per account is simple - for better performance - PG cannot share connection between databases... but this is a topic for another discussion...

as i see in DBAL only 2 platforms support namespaces - PG and SQL server and as it appears only PG has ability to manually specify order used to search table by short name, for example if i ```SET search_path = one, public``` PG try to search tables in this order:
- pg_temp (if current session uses temporary tables)
- pg_catalog (if you not specify it in search_path)
- one
- public

tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
linniksa: i can simplify this test case to
```php
public function testListTableDetailsWithNumberSchema() : void
{
    $this->connection->exec('CREATE SCHEMA IF NOT EXISTS "001_test"');
    $this->connection->exec('SET search_path = "001_test", public');

    try {
        $this->testListQuotedTable();
    } finally {
        $this->connection->exec('SET search_path TO DEFAULT');
    }
}
```

morozov: Please do.

tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
morozov: I see the use case. I'll let the PR hang for a few days in case if anyone is not okay with it and then will merge. Mind that if the PR targets `master`, it will not be released earlier than in summer.

> because this is not happiness in real life

It's really beyond the scope of the issue. Limitations of a high-level library are not really something that should define the design of the lower-level one. Especially if they belong to the same organization.

tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php
linniksa: done, also i changed target branch to 2.10

morozov (approved):

morozov:
Someone better familiar with PostgreSQL, please take a look.

morozov:
Thanks, @linniksa!

PR #3824, opened 2020-01-16
Fix phpstan errors

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

Fixes more phpstan errors, as suggested in #3801.

### DIFF
--- a/lib/Doctrine/DBAL/Logging/SQLLogger.php
+++ b/lib/Doctrine/DBAL/Logging/SQLLogger.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Doctrine\DBAL\Logging;
 
+use Doctrine\DBAL\Types\Type;
+
 /**
  * Interface for SQL loggers.
  */
@@ -12,9 +14,9 @@ interface SQLLogger
     /**
      * Logs a SQL statement somewhere.
      *
-     * @param string         $sql    The SQL to be executed.
-     * @param mixed[]        $params The SQL parameters.
-     * @param int[]|string[] $types  The SQL parameter types.
+     * @param string                $sql    The SQL to be executed.
+     * @param mixed[]               $params The SQL parameters.
+     * @param int[]|string[]|Type[] $types  The SQL parameter types.
      */
     public function startQuery(string $sql, array $params = [], array $types = []) : void;
 
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -954,7 +954,7 @@ class QueryBuilder
      *         );
      * </code>
      *
-     * @param array<int, mixed> $values The values to specify for the insert query indexed by column names.
+     * @param array<string, mixed> $values The values to specify for the insert query indexed by column names.
      *
      * @return $this This QueryBuilder instance.
      */
--- a/lib/Doctrine/DBAL/Schema/Table.php
+++ b/lib/Doctrine/DBAL/Schema/Table.php
@@ -840,9 +840,9 @@ class Table extends AbstractAsset
     }
 
     /**
-     * @param array<string|int, string> $columns
-     * @param array<int, string>        $flags
-     * @param array<string, mixed>      $options
+     * @param array<int, string>   $columns
+     * @param array<int, string>   $flags
+     * @param array<string, mixed> $options
      *
      * @throws SchemaException
      */
@@ -852,13 +852,7 @@ class Table extends AbstractAsset
             throw IndexNameInvalid::new($indexName);
         }
 
-        foreach ($columns as $index => $value) {
-            if (is_string($index)) {
-                $columnName = $index;
-            } else {
-                $columnName = $value;
-            }
-
+        foreach ($columns as $columnName) {
             if (! $this->hasColumn($columnName)) {
                 throw ColumnDoesNotExist::new($columnName, $this->_name);
             }
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -16,8 +16,8 @@ parameters:
         # https://bugs.php.net/bug.php?id=78126
         - '~^Call to an undefined method PDO::sqliteCreateFunction\(\)\.\z~'
 
-        # https://github.com/phpstan/phpstan/issues/1847
-        - '~^Parameter #2 \$registeredAliases of static method Doctrine\\DBAL\\Query\\QueryException::nonUniqueAlias\(\) expects array<string>, array<int, int|string> given\.\z~'
+        # https://github.com/phpstan/phpstan/issues/2857
+        - '~^Parameter #2 \$registeredAliases of static method Doctrine\\DBAL\\Query\\Exception\\NonUniqueAlias::new\(\) expects array<string>, array<int, int\|string> given\.\z~'
 
         # legacy remnants from doctrine/common
         - '~^Class Doctrine\\Common\\(Collections\\Collection|Persistence\\Proxy) not found\.\z~'
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -558,9 +558,9 @@ class Connection implements DriverConnection
      *
      * Table expression and columns are not escaped and are not safe for user-input.
      *
-     * @param string                                           $table      The SQL expression of the table on which to delete.
-     * @param array<string, mixed>                             $identifier The deletion criteria. An associative array containing column-value pairs.
-     * @param array<int, int|string>|array<string, int|string> $types      The query parameter types.
+     * @param string                        $table      The SQL expression of the table on which to delete.
+     * @param array<string, mixed>          $identifier The deletion criteria. An associative array containing column-value pairs.
+     * @param array<int|string, int|string> $types      The query parameter types.
      *
      * @return int The number of affected rows.
      *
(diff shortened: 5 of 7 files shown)


BenMorel:
@morozov I don't know how to best handle the last 3 errors:

- 2 tests forcibly use a wrong type to ensure that an exception is thrown, but of course this makes phpstan unhappy:

https://github.com/doctrine/dbal/blob/e0a92b05e20d80f48b88f841ec240f0cf1cebbc8/tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php#L624

- Here `array_keys()` is called on an array that only has string keys, but phpstan doesn't understand this, and assumes that the result is an array containing `int|string`:

https://github.com/doctrine/dbal/blob/e0a92b05e20d80f48b88f841ec240f0cf1cebbc8/lib/Doctrine/DBAL/Query/QueryBuilder.php#L1319

Any advice?

lib/Doctrine/DBAL/Logging/SQLLogger.php
```diff
@@ -12,9 +14,9 @@ interface SQLLogger
     /**
      * Logs a SQL statement somewhere.
      *
-     * @param string         $sql    The SQL to be executed.
-     * @param mixed[]        $params The SQL parameters.
-     * @param int[]|string[] $types  The SQL parameter types.
+     * @param string                $sql    The SQL to be executed.
+     * @param mixed[]               $params The SQL parameters.
+     * @param int[]|string[]|Type[] $types  The SQL parameter types.
```
morozov: I believe, here it should be `(int|string|Type)[]`. Unlike the above, the array elements don't have to be of the same type and we used this syntax earlier because it was not supported by PHPStan or PhpStorm. If both are fine which the proposed syntax, let's update the type.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -954,7 +954,7 @@ public function setValue(string $column, string $value) : self
      *         );
      * </code>
      *
-     * @param array<int, mixed> $values The values to specify for the insert query indexed by column names.
+     * @param array<string, mixed> $values The values to specify for the insert query indexed by column names.
```
morozov: It looks like `$values` is a list (i.e. the keys are not used):

https://github.com/doctrine/dbal/blob/8b373f4a5ba339be6c0740256dfa271fca34f8d6/lib/Doctrine/DBAL/Query/QueryBuilder.php#L1191-L1193

Please double-check if this change is needed.

morozov (commented):
> 2 tests forcibly use a wrong type to ensure that an exception is thrown, but of course this makes phpstan unhappy

Drop them. This is what PHPStan is for :-)

> Here `array_keys()` is called on an array that only has string keys, but phpstan doesn't understand this, and assumes that the result is an array containing `int|string`

It should have been fixed as per https://github.com/phpstan/phpstan/issues/1847 but it seems it isn't. It looks like out of the two related suppressions only one is needed (the other is already removed in master, please rebase). Please leave the remaining one for now, file a PHPStan issue that would reproduce our case in their sandbox (same as it was done in https://github.com/phpstan/phpstan/issues/1847) and update the reference to point to the newly created issue in our PHPStan configuration.

BenMorel:
> Drop them. This is what PHPStan is for :-)

Done!

> please rebase

Done.

> It should have been fixed as per phpstan/phpstan#1847 but it seems it isn't.

Strangely, it doesn't happen in their sandbox:

https://phpstan.org/r/eea2b886-3942-425d-8ecd-b50489599ff5

I've double checked, the fix should be included in the phpstan version we're using, too, so I don't understand. Any idea?

lib/Doctrine/DBAL/Logging/SQLLogger.php
BenMorel: PHPStorm doesn't understand it at the moment. What do we do then?

morozov: Keep it as is.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: I'm not following you, the snippet you mention is not where the next call jumps to (`add()`)?

And the change is in line with the description:

> $values The values to specify for the insert query **indexed by column names**.

morozov:
> I've double checked, the fix should be included in the phpstan version we're using, too, so I don't understand. Any idea?

The difference may be in the config. Please put together the simplest possible example that gets flagged with our version and config and then see if it depends on the config. Then report whatever reproducible issue. IIRC, not all issues are reproducible in the sandbox (e.g. https://github.com/phpstan/phpstan/issues/2510).

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: Oh, sorry, I didn't see that there was `array_keys()` used on values. You're right.

BenMorel:
Finally reproduced it, and reported as phpstan/phpstan#2857.

I forgot to include the exception class in the sandbox, obviously. But for some reason phpstan didn't complain about the missing class at all.

lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -852,12 +852,8 @@ private function _createIndex(array $columns, string $indexName, bool $isUnique,
             throw IndexNameInvalid::new($indexName);
         }
 
-        foreach ($columns as $index => $value) {
-            if (is_string($index)) {
-                $columnName = $index;
-            } else {
-                $columnName = $value;
-            }
```
morozov: FWIW, this might be used to support different formats of columns like and `['foo', 'bar' => ['option' => 'value']]`. Not sure if that's still the case but may be worth double-checking.

phpstan.neon.dist
```diff
@@ -17,7 +17,7 @@ parameters:
         - '~^Call to an undefined method PDO::sqliteCreateFunction\(\)\.\z~'
 
         # https://github.com/phpstan/phpstan/issues/1847
-        - '~^Parameter #2 \$registeredAliases of static method Doctrine\\DBAL\\Query\\QueryException::nonUniqueAlias\(\) expects array<string>, array<int, int|string> given\.\z~'
+        - '~^Parameter #2 \$registeredAliases of static method Doctrine\\DBAL\\Query\\QueryException::nonUniqueAlias\(\) expects array<string>, array<int, int\|string> given\.\z~'
```
morozov: Now it should be `Doctrine\DBAL\Query\Exception\NonUniqueAlias::new()` instead of `Doctrine\DBAL\Query\QueryException::nonUniqueAlias()` until the PHPStan issue is fixed (please update the link to the newly created issue).

lib/Doctrine/DBAL/Schema/Table.php
BenMorel: Before deleting this code, I've double-checked manually that all calls pass an `array<int, string>`, so I don't think this is relevant anymore. And this doesn't break any tests, which makes me quite confident.

phpstan.neon.dist
BenMorel: Done in 8272dea2d6f6a7274593094521da50bab0b4de49.

BenMorel:
@morozov All good now! 👍

morozov:
@BenMorel, looks good! Please squash.

morozov:
To get rid of the remaining PHPStan issue, we may replace the associative array representing a join in the Query Builder with an object similar to `DependencyOrderNode` used in https://github.com/doctrine/dbal/pull/3762. This way, we will not only help static analyzers understand which properties and of what types a join has but also may help PHP manage memory more efficiently:
1. https://steemit.com/php/@crell/php-use-associative-arrays-basically-never
2. https://youtu.be/JBWgvUrb-q8?t=943

lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -852,12 +852,8 @@ private function _createIndex(array $columns, string $indexName, bool $isUnique,
...
+        foreach ($columns as $value) {
+            $columnName = $value;
```
morozov: Can we get rid of this intermediary variable? `foreach ($columns as $columnName)`

BenMorel:
> Please squash.

Just for my information, what's wrong with the "Squash & Merge" button on GitHub?

> To get rid of the remaining PHPStan issue, we may replace the associative array representing a join in the Query Builder with an object similar to DependencyOrderNode used in #3762.

I'd love to replace associative arrays with objects. I may try to play with this next.

morozov:
> Just for my information, what's wrong with the "Squash & Merge" button on GitHub?

1. There won't be a merge commit, the changes will be squashed and just committed to the target branch ([ref](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#squash-and-merge-your-pull-request-commits)). We want to have merge commits in the repo for better attribution.
2. If a contributor sings their commits, they will also sign the squashed one. Otherwise, if squashed by GitHub, the resulting commit will be signed by GitHub, not by the contributor.

lib/Doctrine/DBAL/Schema/Table.php
BenMorel: You're right, totally missed it! Fixed.

BenMorel:
@morozov Thanks. Squashed!

morozov (approved):

morozov:
Thanks, @BenMorel!

PR #3829, opened 2020-01-18
Proper classes for QueryBuilder SQL parts

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | none

#### Summary

As suggested by @morozov in https://github.com/doctrine/dbal/pull/3824#issuecomment-575706664:

> To get rid of the remaining PHPStan issue, we may replace the associative array representing a join in the Query Builder with an object similar to DependencyOrderNode used in #3762. This way, we will not only help static analyzers understand which properties and of what types a join has but also may help PHP manage memory more efficiently.

This PR introduces 3 classes, which replace the `$sqlParts` nested associative array:

- `QueryParts`
- `QueryPartFrom`
- `QueryPartJoin`

This is a first draft with both phpunit & phpstan happy.
A few pain points are to be discussed, though.

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -11,6 +11,7 @@ use Doctrine\DBAL\Query\Exception\NonUniqueAlias;
 use Doctrine\DBAL\Query\Exception\UnknownAlias;
 use Doctrine\DBAL\Query\Expression\CompositeExpression;
 use Doctrine\DBAL\Query\Expression\ExpressionBuilder;
+use InvalidArgumentException;
 use function array_key_exists;
 use function array_keys;
 use function array_unshift;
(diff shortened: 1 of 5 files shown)


morozov:
Originally, I wanted to start only with the `Join` object since this is where we had problems with the type system. Why don't we solve one problem at a time? Otherwise, it may be much slower since I already have a lot of questions.

BenMorel:
@morozov I don't feel like this makes it much easier. They all take their roots in `$sqlParts`, and apart from the few comments I've made, it's not that big of a change?

BenMorel:
But you may be right. I'll see if I can make a very small PR with just `QueryPartJoin`.

morozov:
I'm not concerned about the scale of the change. You and I have a history of discovering huge amounts of work by starting with mid-scale changes. We can go full scale if you prefer, but I want to make sure you're not disappointed by the number of back and forths and the time it may take to get done.

BenMorel:
As long as we're getting there, I don't really mind. But you may have a point, we could possibly get there in less time by going with small increments. Let me try a very small PR for the join only, and we'll see much more clearly which path we want to follow.

BenMorel:
@morozov #3830 ready for your review! It's a bite-size chunk :)

BenMorel:
Superseded by #3836.

PR #3830, opened 2020-01-18
Introduce Join class in QueryBuilder

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | none

#### Summary

This is a subset of #3829, that only replaces the associative array for the JOIN part with a proper class.
Note that this change alone doesn't solve the [phpstan issue](https://github.com/doctrine/dbal/blob/6e4599d5e492c7314770c6a8250fd84c211f368a/phpstan.neon.dist#L19-L20).

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -20,7 +20,6 @@ use function implode;
 use function is_array;
 use function is_object;
 use function key;
-use function strtoupper;
 use function substr;
 
 /**
@@ -426,7 +425,7 @@ class QueryBuilder
                 foreach ($sqlPart as $part) {
                     $this->sqlParts[$sqlPartName][] = $part;
                 }
-            } elseif ($isArray && is_array($sqlPart[key($sqlPart)])) {
+            } elseif ($isArray && (is_array($sqlPart[key($sqlPart)]) || is_object($sqlPart[key($sqlPart)]))) {
                 $key                                  = key($sqlPart);
                 $this->sqlParts[$sqlPartName][$key][] = $sqlPart[$key];
             } elseif ($isMultiple) {
@@ -673,12 +672,7 @@ class QueryBuilder
     public function innerJoin(string $fromAlias, string $join, string $alias, ?string $condition = null)
     {
         return $this->add('join', [
-            $fromAlias => [
-                'joinType'      => 'inner',
-                'joinTable'     => $join,
-                'joinAlias'     => $alias,
-                'joinCondition' => $condition,
-            ],
+            $fromAlias => Join::inner($join, $alias, $condition),
         ], true);
     }
 
@@ -702,12 +696,7 @@ class QueryBuilder
     public function leftJoin(string $fromAlias, string $join, string $alias, ?string $condition = null)
     {
         return $this->add('join', [
-            $fromAlias => [
-                'joinType'      => 'left',
-                'joinTable'     => $join,
-                'joinAlias'     => $alias,
-                'joinCondition' => $condition,
-            ],
+            $fromAlias => Join::left($join, $alias, $condition),
         ], true);
     }
 
@@ -731,12 +720,7 @@ class QueryBuilder
     public function rightJoin(string $fromAlias, string $join, string $alias, ?string $condition = null)
     {
         return $this->add('join', [
-            $fromAlias => [
-                'joinType'      => 'right',
-                'joinTable'     => $join,
-                'joinAlias'     => $alias,
-                'joinCondition' => $condition,
-            ],
+            $fromAlias => Join::right($join, $alias, $condition),
         ], true);
     }
 
@@ -770,7 +754,7 @@ class QueryBuilder
      *         ->from('counters', 'c')
      *         ->where('c.id = ?');
      *
-     *     // You can optionally programatically build and/or expressions
+     *     // You can optionally programmatically build and/or expressions
      *     $qb = $conn->createQueryBuilder();
      *
      *     $or = $qb->expr()->orx();
@@ -1315,17 +1299,19 @@ class QueryBuilder
 
         if (isset($this->sqlParts['join'][$fromAlias])) {
             foreach ($this->sqlParts['join'][$fromAlias] as $join) {
-                if (array_key_exists($join['joinAlias'], $knownAliases)) {
-                    throw NonUniqueAlias::new($join['joinAlias'], array_keys($knownAliases));
+                /** @var Join $join */
+                if (array_key_exists($join->alias, $knownAliases)) {
+                    throw NonUniqueAlias::new($join->alias, array_keys($knownAliases));
                 }
-                $sql                             .= ' ' . strtoupper($join['joinType'])
-                    . ' JOIN ' . $join['joinTable'] . ' ' . $join['joinAlias']
-                    . ' ON ' . ((string) $join['joinCondition']);
-                $knownAliases[$join['joinAlias']] = true;
+                $sql                       .= ' ' . $join->type
+                    . ' JOIN ' . $join->table . ' ' . $join->alias
+                    . ' ON ' . ((string) $join->condition);
+                $knownAliases[$join->alias] = true;
             }
 
             foreach ($this->sqlParts['join'][$fromAlias] as $join) {
-                $sql .= $this->getSQLForJoins($join['joinAlias'], $knownAliases);
+                /** @var Join $join */
+                $sql .= $this->getSQLForJoins($join->alias, $knownAliases);
             }
         }
 
--- a/lib/Doctrine/DBAL/Query/Join.php
+++ b/lib/Doctrine/DBAL/Query/Join.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Query;
+
+/**
+ * @internal
+ */
+final class Join
+{
+    /** @var string */
+    public $type;
+
+    /** @var string */
+    public $table;
+
+    /** @var string */
+    public $alias;
+
+    /** @var string|null */
+    public $condition;
+
+    private function __construct(string $type, string $table, string $alias, ?string $condition)
+    {
+        $this->type      = $type;
+        $this->table     = $table;
+        $this->alias     = $alias;
+        $this->condition = $condition;
+    }
+
+    public static function inner(string $table, string $alias, ?string $condition) : Join
+    {
+        return new self('INNER', $table, $alias, $condition);
+    }
+
+    public static function left(string $table, string $alias, ?string $condition) : Join
+    {
+        return new self('LEFT', $table, $alias, $condition);
+    }
+
+    public static function right(string $table, string $alias, ?string $condition) : Join
+    {
+        return new self('RIGHT', $table, $alias, $condition);
+    }
+}


lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1315,17 +1315,19 @@ private function getSQLForJoins(string $fromAlias, array &$knownAliases) : strin
...
+                if (array_key_exists($join->joinAlias, $knownAliases)) {
+                    throw NonUniqueAlias::new($join->joinAlias, array_keys($knownAliases));
                 }
-                $sql                             .= ' ' . strtoupper($join['joinType'])
-                    . ' JOIN ' . $join['joinTable'] . ' ' . $join['joinAlias']
-                    . ' ON ' . ((string) $join['joinCondition']);
-                $knownAliases[$join['joinAlias']] = true;
+                $sql                           .= ' ' . strtoupper($join->joinType)
+                    . ' JOIN ' . $join->joinTable . ' ' . $join->joinAlias
+                    . ' ON ' . ((string) $join->joinCondition);
```
BenMorel: Same comment as in #3829:

> Strangely, AFAICS in the source code (and confirmed by no zero broken tests), `joinCondition` is `string|null`. But casting `null` to `string` here would result in `ON` nothing, which would be a syntax error. I'm not sure what the intention was here.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -426,7 +426,7 @@ public function add(string $sqlPartName, $sqlPart, bool $append = false) : self
                 foreach ($sqlPart as $part) {
                     $this->sqlParts[$sqlPartName][] = $part;
                 }
-            } elseif ($isArray && is_array($sqlPart[key($sqlPart)])) {
+            } elseif ($isArray && (is_array($sqlPart[key($sqlPart)]) || is_object($sqlPart[key($sqlPart)]))) {
```
BenMorel: I had to add this to properly handle the `QueryPartJoin` object, that's not an associative array anymore.
Anyway, this whole `add()` method is doomed to be removed.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1315,17 +1315,19 @@ private function getSQLForJoins(string $fromAlias, array &$knownAliases) : strin
...
-                    throw NonUniqueAlias::new($join['joinAlias'], array_keys($knownAliases));
+                /** @var QueryPartJoin $join */
+                if (array_key_exists($join->joinAlias, $knownAliases)) {
+                    throw NonUniqueAlias::new($join->joinAlias, array_keys($knownAliases));
                 }
-                $sql                             .= ' ' . strtoupper($join['joinType'])
-                    . ' JOIN ' . $join['joinTable'] . ' ' . $join['joinAlias']
-                    . ' ON ' . ((string) $join['joinCondition']);
-                $knownAliases[$join['joinAlias']] = true;
+                $sql                           .= ' ' . strtoupper($join->joinType)
```
morozov: The upper-case value of the join type could be coded in a named constructor instead of the builder.

BenMorel: Done in f1078a65e4e65064021a59b613b0b441c499ad11 and 545481915591df7300339c1665b045ce5a38c5e8.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -701,14 +701,14 @@ public function innerJoin(string $fromAlias, string $join, string $alias, ?strin
...
-            $fromAlias => [
-                'joinType'      => 'left',
-                'joinTable'     => $join,
-                'joinAlias'     => $alias,
-                'joinCondition' => $condition,
-            ],
-        ], true);
+        $queryPartJoin = new QueryPartJoin();
+
+        $queryPartJoin->joinType      = 'left';
```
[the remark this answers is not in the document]
BenMorel: Done in f1078a65e4e65064021a59b613b0b441c499ad11. Please note that I'm keeping public properties, as this class is internal and this is both more readable and more performant.

Is there a way to annotate the class as immutable in phpstan? Psalm has a `@psalm-immutable` annotation for this purpose.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -701,14 +701,14 @@ public function innerJoin(string $fromAlias, string $join, string $alias, ?strin
...
+        $queryPartJoin->joinTable     = $join;
+        $queryPartJoin->joinAlias     = $alias;
+        $queryPartJoin->joinCondition = $condition;
```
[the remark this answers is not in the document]
BenMorel: Done in f1078a65e4e65064021a59b613b0b441c499ad11.

lib/Doctrine/DBAL/Query/QueryPartJoin.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Query;
+
+class QueryPartJoin
+{
+    /** @var string */
+    public $joinType;
```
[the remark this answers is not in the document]
BenMorel: Done in a9b22ef5d31cabb2c3295e67325dc9a84e9c72f8.

lib/Doctrine/DBAL/Query/QueryPartJoin.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Query;
+
+class QueryPartJoin
```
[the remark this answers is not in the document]
BenMorel: Done in a9b22ef5d31cabb2c3295e67325dc9a84e9c72f8.

lib/Doctrine/DBAL/Query/QueryBuilder.php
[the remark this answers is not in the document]
morozov: > Please note that I'm keeping public properties, as this class is internal and this is both more readable and more performant.

Yes, this is how I expected it to be.

> Is there a way to annotate the class as immutable in phpstan?

None that I'm aware of. But we can start gently adding Psalm after we clean up the already identified mess.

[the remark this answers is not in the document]
BenMorel: I've seen projects use both indeed, this may be a good thing.

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: I think we should replace casting with a condition. A `JOIN` without `ON` is syntactically valid but an empty `ON` isn't. It doesn't have to be done now since we'd need a test for it. Mind filing an issue?

BenMorel: The last sticking point is this `ON (empty string)`. Should I change it to only add the `ON` part if the condition is not null? **Was this really a bug that went unnoticed all that time?**

BenMorel: I don't mind either doing it in this PR, or, if you want to keep changes self-contained, you can merge this first and I'll create another PR with the fix and a test.

morozov: > Should I change it to only add the `ON` part if the condition is not null?

Not here but yes, in a separate PR that would have a test that reproduces the actual issue.

> Was this really a bug that went unnoticed all that time?

It looks like that. This logic exists since the initial implementation of the query builder (https://github.com/doctrine/dbal/commit/ae833d153f77a44c3737e7760bebfc992b579cd5).

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -701,14 +695,9 @@ public function innerJoin(string $fromAlias, string $join, string $alias, ?strin
...
-            $fromAlias => [
-                'joinType'      => 'left',
-                'joinTable'     => $join,
-                'joinAlias'     => $alias,
-                'joinCondition' => $condition,
-            ],
-        ], true);
+        $join = Join::left($join, $alias, $condition);
+
+        return $this->add('join', [$fromAlias => $join], true);
```
morozov: Would it make sense to keep the entire body as one expression as it was done before? This way it looks more functional/streamlined/sexy.

morozov:
Code-wise looks good! Do you want to try and remove the whitelisted PHPStan error, then squash everything?

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: > I'll create another PR with the fix and a test.

Once the test confirms that it's a but, the other PR may target `2.10` instead of `master` so that we can release the fix sooner.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Done in a894328e5eb57309c485b3d374867fbbf9020703.

BenMorel:
> Do you want to try and remove the whitelisted PHPStan error

Already tried it, still the same error.

> then squash everything?

Squashed!

morozov (approved):

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Sure, but I'll have to create 2 PRs I guess, as 2.10 will need the fix with the associative array code, and master will need the fix with the Join class?

morozov:
> BC Break: yes

Not sure where it is. Should it be documented in upgrade notes then?

> Already tried it, still the same error.

This is surprising. It may make sense to update the reported issue since unlike the earlier implementation, the proposed uses better-defined types.

BenMorel:
> Not sure where it is. Should it be documented in upgrade notes then?

It's in `getQueryPart(s)()`. The returned array is not compatible with the previous one. I'd rather plan to remove these methods altogether, and document it then.

BenMorel:
> This is surprising. It may make sense to update the reported issue since unlike the earlier implementation, the proposed uses better-defined types.

Strangely, now it doesn't happen in the sandbox anymore:
https://phpstan.org/r/8dbea40a-8f10-441a-8e83-cdbdceb8bb88

But it definitely does in our codebase. TBH, I'd better avoid wasting more time on this phpstan bug, and invest it in improving this codebase instead. Once all the changes will be merged, the error will be gone, as you can see in #3829 (https://github.com/doctrine/dbal/pull/3829/commits/d8eecdf078e055e040a53f74c450546768dbd12e).

morozov:
Thanks, @BenMorel!

morozov:
> I'd better avoid wasting more time on this phpstan bug

```diff
diff --git a/lib/Doctrine/DBAL/Query/QueryBuilder.php b/lib/Doctrine/DBAL/Query/QueryBuilder.php
index 00eba6b55..f20e466d9 100644
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -1298,8 +1298,8 @@ class QueryBuilder
         $sql = '';
 
         if (isset($this->sqlParts['join'][$fromAlias])) {
+            /** @var Join $join */
             foreach ($this->sqlParts['join'][$fromAlias] as $join) {
-                /** @var Join $join */
                 if (array_key_exists($join->alias, $knownAliases)) {
                     throw NonUniqueAlias::new($join->alias, array_keys($knownAliases));
                 }
```
Not a bug but documented behavior: https://github.com/phpstan/phpstan/issues/466#issuecomment-324533979.

BenMorel:
@morozov Well spotted! Fixed in #3835.

PR #3831, opened 2020-01-18
Replaced func_get_args() with variadic arguments

### LINKED ISSUE #3825: Replace func_get_args with variadic arguments where possible in the API
|    Q        |   A
|------------ | ------
| Type | improvement
| BC Break    | yes

#### Summary

Classes like `QueryBuilder`, `ExpressionBuilder` and `AbstractSchemaManager` use `func_get_args()` in their implementations for dealing with variadic method arguments. Those arguments should be defined as such in the method signatures.

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

Fixes #3825

In addition to the method signature changes, the following changes have been made:

1. The implementations of `where()`, `having()` and their `and*` and `or*` counterparts have been refactored to reduce code duplication.
2. The `QueryBuilder` tests that call `$qb->expr()` but don't use the return value have been cleaned up.

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -13,9 +13,9 @@ use Doctrine\DBAL\Query\Expression\CompositeExpression;
 use Doctrine\DBAL\Query\Expression\ExpressionBuilder;
 use function array_key_exists;
 use function array_keys;
-use function array_unshift;
-use function func_get_args;
-use function func_num_args;
+use function array_merge;
+use function array_shift;
+use function count;
 use function implode;
 use function is_array;
 use function is_object;
@@ -454,21 +454,19 @@ class QueryBuilder
      *         ->leftJoin('u', 'phonenumbers', 'p', 'u.id = p.user_id');
      * </code>
      *
-     * @param mixed $select The selection expressions.
+     * @param string ...$expressions The selection expressions.
      *
      * @return $this This QueryBuilder instance.
      */
-    public function select($select = null) : self
+    public function select(string ...$expressions) : self
     {
         $this->type = self::SELECT;
 
-        if (empty($select)) {
+        if (count($expressions) < 1) {
             return $this;
         }
 
-        $selects = is_array($select) ? $select : func_get_args();
-
-        return $this->add('select', $selects);
+        return $this->add('select', $expressions);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -501,21 +499,16 @@ public function distinct() : self
...
         $this->type = self::SELECT;
 
-        if (empty($select)) {
-            return $this;
-        }
-
-        $selects = is_array($select) ? $select : func_get_args();
-
-        return $this->add('select', $selects, true);
+        return $this->add('select', array_merge([$expression], $expressions), true);
```
lcobucci: This is changing the behaviour since the first argument could be an array already.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -872,23 +844,18 @@ public function orWhere($where)
...
+    public function groupBy($expression, ...$expressions) : self
     {
-        if (empty($groupBy)) {
-            return $this;
-        }
-
-        $groupBy = is_array($groupBy) ? $groupBy : func_get_args();
-
-        return $this->add('groupBy', $groupBy, false);
+        return $this->add('groupBy', array_merge([$expression], $expressions), false);
```
lcobucci: [Same here](https://github.com/doctrine/dbal/pull/3831/files#r368245091)

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -898,19 +865,14 @@ public function groupBy($groupBy) : self
...
+    public function addGroupBy($expression, ...$expressions) : self
     {
-        if (empty($groupBy)) {
-            return $this;
-        }
-
-        $groupBy = is_array($groupBy) ? $groupBy : func_get_args();
-
-        return $this->add('groupBy', $groupBy, true);
+        return $this->add('groupBy', array_merge([$expression], $expressions), true);
```
lcobucci: [And here](https://github.com/doctrine/dbal/pull/3831/files#r368245091)

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: I should have probably changed the parameter type in the annotation from `mixed` to `string`. This is already reflected in the upgrade notes and there's another PR coming: https://github.com/morozov/dbal/compare/issues/3825...morozov:issues/3825-NEXT.

morozov: For now, I've changed the annotated parameter type.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -501,21 +499,16 @@ public function distinct() : self
      *         ->leftJoin('u', 'phonenumbers', 'u.id = p.user_id');
      * </code>
      *
-     * @param mixed $select The selection expression.
+     * @param string $expression     The selection expressions.
```
greg0ire:
```suggestion
     * @param string $expression     The selection expression.
```

morozov: Thanks, addressed.

greg0ire (approved):

beberlei (approved):
I think this makes sense, because the array support really only exists because at the time we added this feature the `...` syntax didn't exist yet.

But it will be essential that we `trigger_error(E_USER_DEPRECATED)` here, because this only deprecates half the method, and its not easy to find these in your own code base given how this is probably spread in hundreds of places for a mid-sizied application that uses DBAL QueryBuilder.

morozov:
> […] the array support really only exists because at the time we added this feature the `...` syntax didn't exist yet.

This was my thought too. The current API allows mixing arrays and variadic arguments and will silently drop the latter in favor of the former.

PR #3832, opened 2020-01-18
Fix JOIN with no condition bug

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | none

#### Summary

This fixes the bug raised in #3829 and #3830, where a JOIN with an empty condition would yield a syntax error in the resulting query: https://github.com/doctrine/dbal/pull/3830#discussion_r368217791

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
+++ b/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
@@ -92,9 +92,20 @@ class QueryBuilderTest extends DbalTestCase
 
         $qb->select('u.*', 'p.*')
            ->from('users', 'u')
-           ->Join('u', 'phones', 'p', $expr->eq('p.user_id', 'u.id'));
+           ->join('u', 'phones', 'p', $expr->eq('p.user_id', 'u.id'));
 
         self::assertEquals('SELECT u.*, p.* FROM users u INNER JOIN phones p ON p.user_id = u.id', (string) $qb);
+    }
+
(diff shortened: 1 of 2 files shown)


BenMorel:
@morozov I'll open the same PR against `2.10` (with associative array) once this is merged; unless there's a better way?

morozov:
> I'll open the same PR against `2.10` […]

Please retarget this one instead. I'll convert the implementation during the merge. There will be a conflict anyways and there won't be a significant change.

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
```diff
@@ -87,14 +87,25 @@ public function testSelectWithLeftJoin() : void
         self::assertEquals('SELECT u.*, p.* FROM users u LEFT JOIN phones p ON p.user_id = u.id', (string) $qb);
     }
 
+    public function testSelectWithLeftJoinNoCondition() : void
```
morozov: One test is enough. If something goes wrong, there's no point in seeing three tests failing.

BenMorel:
Retargeted to `2.10`!

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
BenMorel: Done!

morozov (approved):

morozov:
Thanks, @BenMorel!

PR #3833, opened 2020-01-18
Introduce From class in QueryBuilder

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | none

#### Summary

Continuing the work started in #3830, this time with a `From` object.

### DIFF
--- a/lib/Doctrine/DBAL/Query/From.php
+++ b/lib/Doctrine/DBAL/Query/From.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Query;
+
+/**
+ * @internal
+ */
+final class From
+{
+    /** @var string */
+    public $table;
+
+    /** @var string|null */
+    public $alias;
+
+    public function __construct(string $table, ?string $alias = null)
+    {
+        $this->table = $table;
+        $this->alias = $alias;
+    }
+}
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -62,6 +62,7 @@ class QueryBuilder
         'select'   => [],
         'distinct' => false,
         'from'     => [],
+        'table'    => null,
         'join'     => [],
         'set'      => [],
         'where'    => null,
@@ -421,7 +422,7 @@ class QueryBuilder
         $this->state = self::STATE_DIRTY;
 
         if ($append) {
-            if ($sqlPartName === 'orderBy' || $sqlPartName === 'groupBy' || $sqlPartName === 'select' || $sqlPartName === 'set') {
+            if ($sqlPartName === 'orderBy' || $sqlPartName === 'groupBy' || $sqlPartName === 'select' || $sqlPartName === 'set' || $sqlPartName === 'from') {
                 foreach ($sqlPart as $part) {
                     $this->sqlParts[$sqlPartName][] = $part;
                 }
@@ -528,23 +529,15 @@ class QueryBuilder
      *         ->setParameter(':user_id', 1);
      * </code>
      *
-     * @param string $delete The table whose rows are subject to the deletion.
-     * @param string $alias  The table alias used in the constructed query.
+     * @param string $table The table whose rows are subject to the deletion.
      *
      * @return $this This QueryBuilder instance.
      */
-    public function delete(?string $delete = null, ?string $alias = null) : self
+    public function delete(string $table) : self
     {
         $this->type = self::DELETE;
 
-        if (! $delete) {
-            return $this;
-        }
-
-        return $this->add('from', [
-            'table' => $delete,
-            'alias' => $alias,
-        ]);
+        return $this->add('table', $table);
     }
 
     /**
@@ -558,23 +551,15 @@ class QueryBuilder
      *         ->where('c.id = ?');
      * </code>
      *
-     * @param string $update The table whose rows are subject to the update.
-     * @param string $alias  The table alias used in the constructed query.
+     * @param string $table The table whose rows are subject to the update.
      *
      * @return $this This QueryBuilder instance.
      */
-    public function update(?string $update = null, ?string $alias = null) : self
+    public function update(string $table) : self
     {
         $this->type = self::UPDATE;
 
-        if (! $update) {
-            return $this;
-        }
-
-        return $this->add('from', [
-            'table' => $update,
-            'alias' => $alias,
-        ]);
+        return $this->add('table', $table);
     }
 
     /**
@@ -592,19 +577,15 @@ class QueryBuilder
      *         );
      * </code>
      *
-     * @param string $insert The table into which the rows should be inserted.
+     * @param string $table The table into which the rows should be inserted.
      *
      * @return $this This QueryBuilder instance.
      */
-    public function insert(?string $insert = null) : self
+    public function insert(string $table) : self
     {
         $this->type = self::INSERT;
 
-        if (! $insert) {
-            return $this;
-        }
-
-        return $this->add('from', ['table' => $insert]);
+        return $this->add('table', $table);
     }
 
     /**
@@ -624,10 +605,7 @@ class QueryBuilder
      */
     public function from(string $from, ?string $alias = null)
     {
-        return $this->add('from', [
-            'table' => $from,
-            'alias' => $alias,
-        ], true);
+        return $this->add('from', new From($from, $alias), true);
     }
 
     /**
@@ -1125,17 +1103,14 @@ class QueryBuilder
         $knownAliases = [];
 
         // Loop through all FROM clauses
+        /** @var From $from */
         foreach ($this->sqlParts['from'] as $from) {
-            if ($from['alias'] === null || $from['alias'] === $from['table']) {
-                $tableSql = $from['table'];
-
-                /** @var string $tableReference */
-                $tableReference = $from['table'];
+            if ($from->alias === null || $from->alias === $from->table) {
+                $tableSql       = $from->table;
+                $tableReference = $from->table;
             } else {
-                $tableSql = $from['table'] . ' ' . $from['alias'];
-
-                /** @var string $tableReference */
-                $tableReference = $from['alias'];
+                $tableSql       = $from->table . ' ' . $from->alias;
+                $tableReference = $from->alias;
             }
 
             $knownAliases[$tableReference] = true;
@@ -1172,7 +1147,7 @@ class QueryBuilder
      */
     private function getSQLForInsert() : string
     {
-        return 'INSERT INTO ' . $this->sqlParts['from']['table'] .
+        return 'INSERT INTO ' . $this->sqlParts['table'] .
         ' (' . implode(', ', array_keys($this->sqlParts['values'])) . ')' .
         ' VALUES(' . implode(', ', $this->sqlParts['values']) . ')';
     }
@@ -1182,15 +1157,7 @@ class QueryBuilder
      */
     private function getSQLForUpdate() : string
     {
-        $from = $this->sqlParts['from'];
-
-        if ($from['alias'] === null || $from['alias'] === $from['table']) {
-            $table = $from['table'];
-        } else {
-            $table = $from['table'] . ' ' . $from['alias'];
-        }
-
-        return 'UPDATE ' . $table
+        return 'UPDATE ' . $this->sqlParts['table']
             . ' SET ' . implode(', ', $this->sqlParts['set'])
             . ($this->sqlParts['where'] !== null ? ' WHERE ' . ((string) $this->sqlParts['where']) : '');
     }
@@ -1200,15 +1167,7 @@ class QueryBuilder
      */
     private function getSQLForDelete() : string
     {
-        $from = $this->sqlParts['from'];
-
-        if ($from['alias'] === null || $from['alias'] === $from['table']) {
-            $table = $from['table'];
-        } else {
-            $table = $from['table'] . ' ' . $from['alias'];
-        }
-
-        return 'DELETE FROM ' . $table . ($this->sqlParts['where'] !== null ? ' WHERE ' . ((string) $this->sqlParts['where']) : '');
+        return 'DELETE FROM ' . $this->sqlParts['table'] . ($this->sqlParts['where'] !== null ? ' WHERE ' . ((string) $this->sqlParts['where']) : '');
     }
 
     /**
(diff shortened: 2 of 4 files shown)


lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1126,16 +1117,17 @@ private function getFromClauses() : array
...
+                $tableSql = $from->table;
 
                 /** @var string $tableReference */
-                $tableReference = $from['table'];
+                $tableReference = $from->table;
             } else {
-                $tableSql = $from['table'] . ' ' . $from['alias'];
+                $tableSql = $from->table . ' ' . $from->alias;
 
                 /** @var string $tableReference */
```
BenMorel: Cannot remove these `@var` annotations (that were already there), as the actual type is `string|null` and makes phpstan unhappy. I'm not sure what our level of confidence is here in the fact that it will never be `null`?

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1172,7 +1164,10 @@ private function isLimitQuery() : bool
      */
     private function getSQLForInsert() : string
     {
-        return 'INSERT INTO ' . $this->sqlParts['from']['table'] .
+        /** @var From $from */
+        $from = $this->sqlParts['from'][0];
```
BenMorel: I'll quote my PR #3829 here:

> The original code seemed to accept either a single value, or an array of values, in `$this->sqlParts['table']` (each value being an associative array with `'table'` and `'alias'` keys).
>
> But in this method, and a few others, **it seems to assume that there is always a single value**, which is confirmed by the tests passing when I always use `$this->queryParts->from[0]`, and failing if I use another index.
>
> **So it looks like this method will fail badly if there's not exactly one table registered.** This PR **does not change this assumption**, but it would be good to ensure that this assumption is always true when this method is called.

lib/Doctrine/DBAL/Query/From.php
```diff
@@ -0,0 +1,23 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Query;
+
+/**
+ * @internal
+ */
+final class From
+{
+    /** @var string|null */
```
morozov: What's the use case when the table is null?

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -421,7 +421,7 @@ public function add(string $sqlPartName, $sqlPart, bool $append = false) : self
         $this->state = self::STATE_DIRTY;
 
         if ($append) {
-            if ($sqlPartName === 'orderBy' || $sqlPartName === 'groupBy' || $sqlPartName === 'select' || $sqlPartName === 'set') {
+            if ($sqlPartName === 'orderBy' || $sqlPartName === 'groupBy' || $sqlPartName === 'select' || $sqlPartName === 'set' || $sqlPartName === 'from') {
```
morozov: Why is this needed?

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: You can assign the alias to a variable before checking whether it's null. This way, in the `else` case, it couldn't be equal to null.

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: Not sure why changing an array with an object requires this change (same as above).

morozov: > So it looks like this method will fail badly if there's not exactly one table registered.

Most likely, it will, which is a bug. If the object is in an invalid state, it should throw an exception. Multiple `from()` tables seem to be acceptable only for `SELECT`.

morozov: After some more research, it looks like for `select()`, the `'from'` part may take multiple `From`s in a zero-indexed array while for `insert()`, `update()` and `delete()`, it's always one table. The question is, why do we need a `From` with a table and an alias for non-`select()` queries? They don't even use the `FROM` keyword in their syntax.

We may try separating the two concerns into two different query parts (array of `From`s in `'from'`) and a single table in `'table'`.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Because of the `|| is_object(...)` introduced right below in the previous PR. This now takes the wrong code path, and the `From` object gets wrapped in an array inside an array. Re-using this code path for the 'from' part does what we expect here.

Anyway, as seen in #3829, I'd like to get rid of this generic `add()` mess ASAP, and leave to each method such as `from()`, `orderBy()` etc. the responsibility to add data to the main array/object in their own way.

lib/Doctrine/DBAL/Query/From.php
BenMorel: I have no idea. I really just converted the existing code for now.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: What about `DELETE a FROM a, b WHERE ...`?

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Actually in this case, we can be confident that `$table` will never be null, as it can only be created from here:

    public function from(string $from, ?string $alias = null)
    {
        return $this->add('from', new From($from, $alias), true);
    }

So I'd rather separate concerns between `FROM` (with non-nullable `$table`) and `INSERT`/`UPDATE`/`DELETE` as you suggested.

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: > What about `DELETE a FROM a, b WHERE ...`?

Is it currently supported? FWIW, it's not portable. Usually, adding new features during refactoring is not the best idea.

BenMorel: > Is it currently supported? FWIW, it's not portable.

Looks like it isn't, and may be MySQL-specific indeed.

> The question is, why do we need a From with a table and an alias for non-select() queries?
> They don't even use the FROM keyword in their syntax.

`DELETE` does use `FROM`, so I guess it *could* make sense to use it, but explicitly restrict it to a single element.

That being said, if the aim is to have `DELETE`, `INSERT` and `UPDATE` all take a single table, we should indeed probably enforce this in another property (or array key, for now).

And I agree, we probably don't need/want an alias in `DELETE`, `INSERT` and `UPDATE`. First of all, it's not very useful when there's a single table. And, AFAICS, this is not portable either. I tested aliases on db-fiddle on 3 dbs:

| | MySQL | PostgreSQL | SQLite |
| ------------- |:-------------:| -----:| -----:|
| `DELETE` alias | ❌ | ✔️ | ✔️ |
| `INSERT` alias | ❌ | ✔️ | ✔️ |
| `UPDATE` alias | ✔️ | ❌ | ❌ |

So we could just ditch aliases for these, and just add a new `'table'` entry accepting the table name only, as `string|null`?

So basically:

| Query type | Key/Prop | Type
| ------------- |:-------------:| -----:|
| `SELECT` | `'from'` | `From[]` |
| `INSERT`, `UPDATE`, `DELETE` | `'table'` | `string\|null` |

BenMorel: Here's how it could look: 58ae117a0e29270e88a22f5681add92c16a6b48d.

Note how the `QueryBuilder` method signatures change, though:

| Before | After |
| ------------- |:-------------:|
| `delete(?string $delete = null, ?string $alias = null)` | `delete(string $delete)` |
| `update(?string $update = null, ?string $alias = null)` | `update(string $update)` |
| `insert(?string $insert = null)` | `insert(string $insert)` |

I believe this change is necessary for 2 reasons:

- **nullable table name**: currently "supported", but throws an Error when you attempt to get the SQL of a `QueryBuilder` with an empty `insert()`, `update()`, or `delete()`.
- **alias**: as mentioned above, it brings close to zero value to have an alias for `INSERT`, `UPDATE` or `DELETE` statement, involving a single table.

Note that I would also rename the `$delete`, `$update` and `$insert` parameters to `$table` for consistency and readability.

@morozov, what do you think?

morozov: > `DELETE` does use `FROM`, so I guess it *could* make sense to use it, but explicitly restrict it to a single element.

When would it make sense? Maybe we could give the `From` class a better name, e.g. `TableWithAlias`. This way, it will be obvious that it's only for select.

morozov: > @morozov, what do you think?

Makes perfect sense. 

> Note that I would also rename the `$delete`, `$update` and `$insert` parameters to `$table` for consistency and readability.

Please do.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1126,16 +1116,17 @@ private function getFromClauses() : array
 
         // Loop through all FROM clauses
         foreach ($this->sqlParts['from'] as $from) {
-            if ($from['alias'] === null || $from['alias'] === $from['table']) {
-                $tableSql = $from['table'];
+            /** @var From $from */
+            if ($from->alias === null || $from->alias === $from->table) {
```
morozov: This logic could be moved into a `From` method (e.g. `hasAlias()`). It would ignore the non-null alias if it's the same as the table name.

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: Since the separation of concerns is already implemented, is this annotation still needed? As far as I can tell, PHPStan handles the non-null property values correctly within a method: https://phpstan.org/r/700b50cd-9879-4801-9921-f2d5a9c3af3c.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1172,7 +1163,10 @@ private function isLimitQuery() : bool
      */
     private function getSQLForInsert() : string
     {
-        return 'INSERT INTO ' . $this->sqlParts['from']['table'] .
+        /** @var string $table */
```
morozov: Although it's out of scope, but having to make such assertions points to the fact that the builders for each query type should be implemented in separate classes that would expose only the relevant methods and make sure a proper state when `toString()` is called.

Otherwise, all sorts of madness like this are possible:
```php
echo $conn->createQueryBuilder()
    ->from('foo')
    ->values([])
    ->set('x', 'y')

// SELECT  FROM foo
```

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: Another reason to enforce the `$table` argument in statements:
```php
echo $conn->createQueryBuilder()
    ->delete()
    ->from('foo')

// PHP Notice:  Undefined index: alias in lib/Doctrine/DBAL/Query/QueryBuilder.php on line 1205
// PHP Notice:  Undefined index: table in lib/Doctrine/DBAL/Query/QueryBuilder.php on line 1206
// DELETE FROM
```

BenMorel: > Please do.

Done in https://github.com/doctrine/dbal/pull/3833/commits/d5ad2b5cea9203cc212f22420e3787778080e8ca

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: This crossed my mind as well. A fluent interface would deserve a proper separation for each kind of SQL statement. Let's come back to this in another PR, maybe!

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: The first one was a leftover and not required indeed, but the second still failed with phpstan. Moving the `@var From $from` above the `foreach` fixed it (remembered your previous comment).

Fixed in https://github.com/doctrine/dbal/pull/3833/commits/e99069b5abb081736f187f48d733daea12676895

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: I'll keep it this way then.

BenMorel:
@morozov All good for me here. Any more comments?

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1182,13 +1160,8 @@ private function getSQLForInsert() : string
...
     {
-        $from = $this->sqlParts['from'];
-
-        if ($from['alias'] === null || $from['alias'] === $from['table']) {
-            $table = $from['table'];
-        } else {
-            $table = $from['table'] . ' ' . $from['alias'];
-        }
+        /** @var string $table */
+        $table = $this->sqlParts['table'];
```
morozov: This is no longer needed as an individual variable (the same in the rest of the cases).

morozov:
> Any more comments?

Almos there. Please document the breaking changes: the `$table` argument is required in `insert()`, `update()` and `delete()` statements, aliases no longer supported. Maybe something else I'm missing?

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: I'm surprised that phpstan doesn't complain about the unknown type converted to string, but it's OK: https://github.com/doctrine/dbal/pull/3833/commits/6adfcd5c3ecdc888bee32da7636d279832981d4f

BenMorel:
> Please document the breaking changes

Done in https://github.com/doctrine/dbal/pull/3833/commits/7c2fc7079f9fffcccb80c20791b7480a1b61d03b

> Maybe something else I'm missing?

Same as #3830 : `getQueryPart(s)()`. I'll document this once the `QueryParts` object is introduced (next PR, once this one is merged).

morozov:
All good. Please squash.

BenMorel:
@morozov Done!

morozov (approved):

morozov:
Thanks @BenMorel! I like the pace.

PR #3834, opened 2020-01-19
Fix docblock typos in DriverManager docs

CHItA:
|      Q       |   A
|------------- | -----------
| Type         | documentation
| BC Break     | no
| Fixed issues | none

#### Summary

In the api docs the names of the drivers are incorrect. Also, could someone confirm that the `(unstable)` annotations are still correct?

### DIFF
--- a/lib/Doctrine/DBAL/DriverManager.php
+++ b/lib/Doctrine/DBAL/DriverManager.php
@@ -88,20 +88,7 @@ final class DriverManager
      *
      * $params must contain at least one of the following.
      *
-     * Either 'driver' with one of the following values:
-     *
-     *     pdo_mysql
-     *     pdo_sqlite
-     *     pdo_pgsql
-     *     pdo_oci (unstable)
-     *     pdo_sqlsrv
-     *     pdo_sqlsrv
-     *     mysqli
-     *     sqlanywhere
-     *     sqlsrv
-     *     ibm_db2 (unstable)
-     *     drizzle_pdo_mysql
-     *
+     * Either 'driver' with one of the array keys of {@link $_driverMap},
      * OR 'driverClass' that contains the full class name (with namespace) of the
      * driver class to instantiate.
      *


morozov:
@CHItA thanks for the update. The `ibm_db2` is pretty much stable. The `(unstable)` annotation should definitely be removed from there. As for `pdo_oci`, it is usable and we continuously test against it. In the end, it's the consumer's call which driver to use.

I'd remove the `(unstable)` annotation from everywhere and sort the drivers alphabetically.

lib/Doctrine/DBAL/DriverManager.php
```diff
@@ -90,17 +90,17 @@ private function __construct()
      *
      * Either 'driver' with one of the following values:
      *
+     *     drizzle_pdo_mysql
```
morozov: Hm… it actually duplicates the keys of the driver map. Would it make sense to reference if using `{@link $_driverMap}` instead of duplication?

https://github.com/doctrine/dbal/blob/3d6d3542a32d0af28136db9d3a06614b32298a48/lib/Doctrine/DBAL/DriverManager.php#L46-L58

CHItA: True, I am fine with whatever. Although `$_driverMap` doesn't seem to be listed in the [API docs](https://www.doctrine-project.org/api/dbal/latest/Doctrine/DBAL/DriverManager.html), which I would consider to be quite useful.

CHItA: One more note, then I'll be happy to push whatever you find to be the best solution:
a) Which one of these would you prefer? Using a `{@link $_driverMap}` or just linking to the `getAvailableDrivers()` function which returns the keys of this array.
b) The docblock has no return value annotation so if I'm here, I might as well fix that too. Should I?

Also, what is your commit squashing policy? Three commits feel a bit much for a one line documentation change.

morozov: >Which one of these would you prefer?

`{@link $_driverMap}` seems like a more reasonable reference. It contains the actual values we're referring to.

> The docblock has no return value annotation […]

Which one?

> Also, what is your commit squashing policy? Three commits feel a bit much for a one line documentation change.

We usually squash commits unless they provide additional clarity to the changes (not the case here).

morozov (approved):

morozov:
Thanks, @CHItA!

lib/Doctrine/DBAL/DriverManager.php
CHItA: Oh, and `getConnection()` is missing the return statement.

CHItA:
@morozov No worries, thank you for maintaining Doctrine! :)

lib/Doctrine/DBAL/DriverManager.php
morozov: > Oh, and `getConnection()` is missing the return statement.

As for the parameter and return types, there's some work in progress in https://github.com/doctrine/dbal/pull/3799. I'd rather address such issues in a systems way rather than one by one.

PR #3836, opened 2020-01-21
Introduce properties for SQL parts in QueryBuilder

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | none

#### Summary

This supersedes #3829. Now that #3830 and #3833 have introduced `From` and `Join` objects, <s>it's time to replace the `$sqlParts` associative array with an explicit object, namely `QueryParts`</s>.

**Update**: the SQL parts are now part of `QueryBuilder` itself.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -5,10 +5,9 @@
 1. The `select()`, `addSelect()`, `groupBy()` and `addGroupBy()` methods no longer accept an array of arguments. Pass each expression as an individual argument or expand an array of expressions using the `...` operator.
 2. The `select()`, `addSelect()`, `groupBy()` and `addGroupBy()` methods no longer ignore the first argument if it's empty.
 3. The `addSelect()` method can be no longer called without arguments.
-
-## BC BREAK: `QueryBuilder::insert()`, `update()` and `delete()` signatures changed
-
-These methods now require the `$table` parameter, and do not support aliases anymore.
+4. The `insert()`, `update()` and `delete()` methods now require the `$table` parameter, and do not support aliases anymore.
+5. The `add()`, `getQueryPart()`, `getQueryParts()`, `resetQueryPart()` and `resetQueryParts()` methods are removed.
+6. For a `select()` query, the `getSQL()` method now throws an expression if no `SELECT` expressions have been provided.
 
 ## BC BREAK: `OCI8Statement::convertPositionalToNamedPlaceholders()` is removed.
 
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -14,12 +14,10 @@ use Doctrine\DBAL\Query\Expression\ExpressionBuilder;
 use function array_key_exists;
 use function array_keys;
 use function array_merge;
-use function array_shift;
+use function array_unshift;
 use function count;
 use function implode;
-use function is_array;
 use function is_object;
-use function key;
 use function substr;
 
 /**
@@ -54,30 +52,6 @@ class QueryBuilder
      * @var Connection
      */
     private $connection;
-
-    /*
-     * The default values of SQL parts collection
-     */
-    private const SQL_PARTS_DEFAULTS = [
-        'select'   => [],
-        'distinct' => false,
-        'from'     => [],
-        'table'    => null,
-        'join'     => [],
-        'set'      => [],
-        'where'    => null,
-        'groupBy'  => [],
-        'having'   => null,
-        'orderBy'  => [],
-        'values'   => [],
-    ];
-
-    /**
-     * The array of SQL parts collected.
-     *
-     * @var array<string, mixed>
-     */
-    private $sqlParts = self::SQL_PARTS_DEFAULTS;
 
     /**
      * The complete SQL string for this query.
@@ -134,6 +108,83 @@ class QueryBuilder
      * @var int
      */
     private $boundCounter = 0;
+
+    /**
+     * The SELECT parts of the query.
+     *
+     * @var string[]
+     */
+    private $select = [];
+
+    /**
+     * Whether this is a SELECT DISTINCT query.
+     *
+     * @var bool
+     */
+    private $distinct = false;
+
+    /**
+     * The FROM parts of a SELECT query.
+     *
+     * @var From[]
+     */
+    private $from = [];
+
+    /**
+     * The table name for an INSERT, UPDATE or DELETE query.
+     *
+     * @var string|null
+     */
+    private $table;
+
+    /**
+     * The list of joins, indexed by from alias.
+     *
+     * @var array<string, Join[]>
+     */
+    private $join = [];
+
+    /**
+     * The SET parts of an UPDATE query.
+     *
+     * @var string[]
+     */
+    private $set = [];
+
+    /**
+     * The WHERE part of a SELECT, UPDATE or DELETE query.
+     *
+     * @var string|CompositeExpression|null
+     */
+    private $where;
+
+    /**
+     * The GROUP BY part of a SELECT query.
+     *
+     * @var string[]
+     */
+    private $groupBy = [];
+
+    /**
+     * The HAVING part of a SELECT query.
+     *
+     * @var string|CompositeExpression|null
+     */
+    private $having;
+
+    /**
+     * The ORDER BY parts of a SELECT query.
+     *
+     * @var string[]
+     */
+    private $orderBy = [];
+
+    /**
+     * The values of an INSERT query.
+     *
+     * @var array<string, mixed>
+     */
+    private $values = [];
 
     /**
      * Initializes a new <tt>QueryBuilder</tt>.
@@ -218,6 +269,8 @@ class QueryBuilder
      * </code>
      *
      * @return string The SQL query string.
+     *
+     * @throws QueryException If the object doesn't represent a valid query in its current state.
      */
     public function getSQL() : string
     {
@@ -400,49 +453,6 @@ class QueryBuilder
     }
 
     /**
-     * Either appends to or replaces a single, generic query part.
-     *
-     * The available parts are: 'select', 'from', 'set', 'where',
-     * 'groupBy', 'having' and 'orderBy'.
-     *
-     * @param mixed $sqlPart
-     *
-     * @return $this This QueryBuilder instance.
-     */
-    public function add(string $sqlPartName, $sqlPart, bool $append = false) : self
-    {
-        $isArray    = is_array($sqlPart);
-        $isMultiple = is_array($this->sqlParts[$sqlPartName]);
-
-        if ($isMultiple && ! $isArray) {
-            $sqlPart = [$sqlPart];
-        }
-
-        $this->state = self::STATE_DIRTY;
-
-        if ($append) {
-            if ($sqlPartName === 'orderBy' || $sqlPartName === 'groupBy' || $sqlPartName === 'select' || $sqlPartName === 'set' || $sqlPartName === 'from') {
-                foreach ($sqlPart as $part) {
-                    $this->sqlParts[$sqlPartName][] = $part;
-                }
-            } elseif ($isArray && (is_array($sqlPart[key($sqlPart)]) || is_object($sqlPart[key($sqlPart)]))) {
-                $key                                  = key($sqlPart);
-                $this->sqlParts[$sqlPartName][$key][] = $sqlPart[$key];
-            } elseif ($isMultiple) {
-                $this->sqlParts[$sqlPartName][] = $sqlPart;
-            } else {
-                $this->sqlParts[$sqlPartName] = $sqlPart;
-            }
-
-            return $this;
-        }
-
-        $this->sqlParts[$sqlPartName] = $sqlPart;
-
-        return $this;
-    }
-
-    /**
      * Specifies an item that is to be returned in the query result.
      * Replaces any previously specified selections, if any.
      *
@@ -465,7 +475,11 @@ class QueryBuilder
             return $this;
         }
 
-        return $this->add('select', $expressions);
+        $this->select = $expressions;
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -482,7 +496,9 @@ class QueryBuilder
      */
     public function distinct() : self
     {
-        $this->sqlParts['distinct'] = true;
+        $this->distinct = true;
+
+        $this->state = self::STATE_DIRTY;
 
         return $this;
     }
@@ -507,7 +523,11 @@ class QueryBuilder
     {
         $this->type = self::SELECT;
 
-        return $this->add('select', array_merge([$expression], $expressions), true);
+        $this->select = array_merge($this->select, [$expression], $expressions);
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -529,7 +549,11 @@ class QueryBuilder
     {
         $this->type = self::DELETE;
 
-        return $this->add('table', $table);
+        $this->table = $table;
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -551,7 +575,11 @@ class QueryBuilder
     {
         $this->type = self::UPDATE;
 
-        return $this->add('table', $table);
+        $this->table = $table;
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -577,7 +605,11 @@ class QueryBuilder
     {
         $this->type = self::INSERT;
 
-        return $this->add('table', $table);
+        $this->table = $table;
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -590,14 +622,18 @@ class QueryBuilder
      *         ->from('users', 'u')
      * </code>
      *
-     * @param string      $from  The table.
+     * @param string      $table The table.
      * @param string|null $alias The alias of the table.
      *
      * @return $this This QueryBuilder instance.
      */
-    public function from(string $from, ?string $alias = null) : self
+    public function from(string $table, ?string $alias = null) : self
     {
-        return $this->add('from', new From($from, $alias), true);
+        $this->from[] = new From($table, $alias);
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -641,9 +677,11 @@ class QueryBuilder
      */
     public function innerJoin(string $fromAlias, string $join, string $alias, ?string $condition = null) : self
     {
-        return $this->add('join', [
-            $fromAlias => Join::inner($join, $alias, $condition),
-        ], true);
+        $this->join[$fromAlias][] = Join::inner($join, $alias, $condition);
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -665,9 +703,11 @@ class QueryBuilder
      */
     public function leftJoin(string $fromAlias, string $join, string $alias, ?string $condition = null) : self
     {
-        return $this->add('join', [
-            $fromAlias => Join::left($join, $alias, $condition),
-        ], true);
+        $this->join[$fromAlias][] = Join::left($join, $alias, $condition);
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -689,9 +729,11 @@ class QueryBuilder
      */
     public function rightJoin(string $fromAlias, string $join, string $alias, ?string $condition = null) : self
     {
-        return $this->add('join', [
-            $fromAlias => Join::right($join, $alias, $condition),
-        ], true);
+        $this->join[$fromAlias][] = Join::right($join, $alias, $condition);
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -711,7 +753,11 @@ class QueryBuilder
      */
     public function set(string $key, string $value) : self
     {
-        return $this->add('set', $key . ' = ' . $value, true);
+        $this->set[] = $key . ' = ' . $value;
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -743,7 +789,11 @@ class QueryBuilder
      */
     public function where($predicate, ...$predicates) : self
     {
-        return $this->setPredicates('where', $predicate, ...$predicates);
+        $this->where = $this->createPredicate($predicate, ...$predicates);
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -767,7 +817,11 @@ class QueryBuilder
      */
     public function andWhere($predicate, ...$predicates) : self
     {
-        return $this->appendPredicates('where', CompositeExpression::TYPE_AND, $predicate, ...$predicates);
+        $this->where = $this->appendToPredicate($this->where, CompositeExpression::TYPE_AND, $predicate, ...$predicates);
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -791,7 +845,11 @@ class QueryBuilder
      */
     public function orWhere($predicate, ...$predicates) : self
     {
-        return $this->appendPredicates('where', CompositeExpression::TYPE_OR, $predicate, ...$predicates);
+        $this->where = $this->appendToPredicate($this->where, CompositeExpression::TYPE_OR, $predicate, ...$predicates);
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -812,7 +870,11 @@ class QueryBuilder
      */
     public function groupBy(string $expression, string ...$expressions) : self
     {
-        return $this->add('groupBy', array_merge([$expression], $expressions), false);
+        $this->groupBy = array_merge([$expression], $expressions);
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -833,7 +895,11 @@ class QueryBuilder
      */
     public function addGroupBy(string $expression, string ...$expressions) : self
     {
-        return $this->add('groupBy', array_merge([$expression], $expressions), true);
+        $this->groupBy = array_merge($this->groupBy, [$expression], $expressions);
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -857,7 +923,7 @@ class QueryBuilder
      */
     public function setValue(string $column, string $value) : self
     {
-        $this->sqlParts['values'][$column] = $value;
+        $this->values[$column] = $value;
 
         return $this;
     }
@@ -883,7 +949,11 @@ class QueryBuilder
      */
     public function values(array $values) : self
     {
-        return $this->add('values', $values);
+        $this->values = $values;
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -897,7 +967,11 @@ class QueryBuilder
      */
     public function having($predicate, ...$predicates) : self
     {
-        return $this->setPredicates('having', $predicate, ...$predicates);
+        $this->having = $this->createPredicate($predicate, ...$predicates);
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -911,7 +985,11 @@ class QueryBuilder
      */
     public function andHaving($predicate, ...$predicates) : self
     {
-        return $this->appendPredicates('having', CompositeExpression::TYPE_AND, $predicate, ...$predicates);
+        $this->having = $this->appendToPredicate($this->having, CompositeExpression::TYPE_AND, $predicate, ...$predicates);
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -925,52 +1003,51 @@ class QueryBuilder
      */
     public function orHaving($predicate, ...$predicates) : self
     {
-        return $this->appendPredicates('having', CompositeExpression::TYPE_OR, $predicate, ...$predicates);
+        $this->having = $this->appendToPredicate($this->having, CompositeExpression::TYPE_OR, $predicate, ...$predicates);
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
-     * Sets one or more predicates combined by the AND logic as the given query clause.
-     * Replaces any previously specified predicates.
+     * Creates a CompositeExpression from one or more predicates combined by the AND logic.
      *
+     * @param string|CompositeExpression $predicate
      * @param string|CompositeExpression ...$predicates
      *
-     * @return $this This QueryBuilder instance.
+     * @return string|CompositeExpression
      */
-    private function setPredicates(string $clause, ...$predicates) : self
+    private function createPredicate($predicate, ...$predicates)
     {
-        if (count($predicates) > 1) {
-            $predicate = new CompositeExpression(
-                CompositeExpression::TYPE_AND,
-                $predicates
-            );
-        } else {
-            $predicate = array_shift($predicates);
+        if (count($predicates) === 0) {
+            return $predicate;
         }
 
-        return $this->add($clause, $predicate);
+        return new CompositeExpression(CompositeExpression::TYPE_AND, array_merge([$predicate], $predicates));
     }
 
     /**
-     * Appends the given predicates combined by the given type of logic to the given query clause.
+     * Appends the given predicates combined by the given type of logic to the current predicate.
      *
-     * @param string|CompositeExpression ...$predicates
+     * @param string|CompositeExpression|null $currentPredicate
+     * @param string|CompositeExpression      ...$predicates
      *
-     * @return $this This QueryBuilder instance.
+     * @return string|CompositeExpression
      */
-    private function appendPredicates(string $clause, string $type, ...$predicates) : self
+    private function appendToPredicate($currentPredicate, string $type, ...$predicates)
     {
-        $predicate = $this->getQueryPart($clause);
-
-        if ($predicate instanceof CompositeExpression && $predicate->getType() === $type) {
-            $predicate->addMultiple($predicates);
-        } else {
-            $predicate = new CompositeExpression(
-                $type,
-                array_merge([$predicate], $predicates)
-            );
+        if ($currentPredicate instanceof CompositeExpression && $currentPredicate->getType() === $type) {
+            return $currentPredicate->addMultiple($predicates);
         }
 
-        return $this->add($clause, $predicate, true);
+        if ($currentPredicate !== null) {
+            array_unshift($predicates, $currentPredicate);
+        } elseif (count($predicates) === 1) {
+            return $predicates[0];
+        }
+
+        return new CompositeExpression($type, $predicates);
     }
 
     /**
@@ -984,7 +1061,17 @@ class QueryBuilder
      */
     public function orderBy(string $sort, ?string $order = null) : self
     {
-        return $this->add('orderBy', $sort . ' ' . (! $order ? 'ASC' : $order), false);
+        $orderBy = $sort;
+
+        if ($order !== null) {
+            $orderBy .= ' ' . $order;
+        }
+
+        $this->orderBy = [$orderBy];
+
+        $this->state = self::STATE_DIRTY;
+
+        return $this;
     }
 
     /**
@@ -997,57 +1084,13 @@ class QueryBuilder
      */
     public function addOrderBy(string $sort, ?string $order = null) : self
     {
-        return $this->add('orderBy', $sort . ' ' . (! $order ? 'ASC' : $order), true);
-    }
+        $orderBy = $sort;
 
-    /**
-     * Gets a query part by its name.
-     *
-     * @return mixed
-     */
-    public function getQueryPart(string $queryPartName)
-    {
-        return $this->sqlParts[$queryPartName];
-    }
-
-    /**
-     * Gets all query parts.
-     *
-     * @return array<string, mixed>
-     */
-    public function getQueryParts() : array
-    {
-        return $this->sqlParts;
-    }
-
-    /**
-     * Resets SQL parts.
-     *
-     * @param array<int, string>|null $queryPartNames
-     *
-     * @return $this This QueryBuilder instance.
-     */
-    public function resetQueryParts(?array $queryPartNames = null) : self
-    {
-        if ($queryPartNames === null) {
-            $queryPartNames = array_keys($this->sqlParts);
+        if ($order !== null) {
+            $orderBy .= ' ' . $order;
         }
 
-        foreach ($queryPartNames as $queryPartName) {
-            $this->resetQueryPart($queryPartName);
-        }
-
-        return $this;
-    }
-
-    /**
-     * Resets a single SQL part.
-     *
-     * @return $this This QueryBuilder instance.
-     */
-    public function resetQueryPart(string $queryPartName) : self
-    {
-        $this->sqlParts[$queryPartName] = self::SQL_PARTS_DEFAULTS[$queryPartName];
+        $this->orderBy[] = $orderBy;
 
         $this->state = self::STATE_DIRTY;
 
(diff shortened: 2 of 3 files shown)


lib/Doctrine/DBAL/Query/QueryParts.php
```diff
@@ -0,0 +1,74 @@
...
+            }
+        }
+
+        if ($this->where !== null) {
+            $this->where = clone $this->where;
+        }
+
+        if ($this->having !== null) {
+            $this->having = clone $this->having;
+        }
```
BenMorel: **What about `$values`, should they be (deep-)cloned?**

The previous implementation blindly cloned objects and 1-level deep arrays of objects, so it looks like this also applied to `$sqlParts['values']`; was this intentional? I'm not sure.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1012,59 +1029,107 @@ public function orderBy(string $sort, ?string $order = null)
...
 
     /**
-     * Gets a query part by its name.
+     * Gets all query parts.
      *
-     * @return mixed
+     * @internal For tests only.
      */
-    public function getQueryPart(string $queryPartName)
+    public function getQueryParts() : QueryParts
```
BenMorel: **Should we keep this method?** It serves little purpose, it's actually only used in [this test](https://github.com/doctrine/dbal/blob/dbd94c9d591e3d6f980f3a077cfa5b64c600b6d7/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php#L733).
I've marked it as `@internal` for now.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1012,59 +1029,107 @@ public function orderBy(string $sort, ?string $order = null)
...
     {
-        $this->sqlParts[$queryPartName] = self::SQL_PARTS_DEFAULTS[$queryPartName];
+        $this->queryParts->table = null;
+        $this->state             = self::STATE_DIRTY;
 
-        $this->state = self::STATE_DIRTY;
+        return $this;
+    }
+
+    public function resetJoin() : QueryBuilder
```
BenMorel: **Should we introduce all these `resetXXX()` methods?** I've added them to replace `resetQueryPart[s]()`; only [2 of them](https://github.com/doctrine/dbal/blob/dbd94c9d591e3d6f980f3a077cfa5b64c600b6d7/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php#L564-L565) are used in tests at the moment.

BenMorel:
@morozov Do you know [what's wrong with phpcs](https://travis-ci.org/doctrine/dbal/jobs/639936205)? How can I early exit [here](https://github.com/doctrine/dbal/blob/dbd94c9d591e3d6f980f3a077cfa5b64c600b6d7/lib/Doctrine/DBAL/Query/QueryParts.php#L70)?

lib/Doctrine/DBAL/Query/QueryParts.php
morozov: Before figuring this out, let's see why exactly we need a separate class. The builder object always uses the same single instance of parts.

morozov:
> Do you know [what's wrong with phpcs](https://travis-ci.org/doctrine/dbal/jobs/639936205)? How can I early exit [here](https://github.com/doctrine/dbal/blob/dbd94c9d591e3d6f980f3a077cfa5b64c600b6d7/lib/Doctrine/DBAL/Query/QueryParts.php#L70)?

Please try `phpcbf`:
```diff
--- a/lib/Doctrine/DBAL/Query/QueryParts.php
+++ b/lib/Doctrine/DBAL/Query/QueryParts.php
@@ -67,8 +67,10 @@ class QueryParts
             $this->where = clone $this->where;
         }
 
-        if ($this->having !== null) {
-            $this->having = clone $this->having;
+        if ($this->having === null) {
+            return;
         }
+
+        $this->having = clone $this->having;
     }
 }
```

See https://github.com/slevomat/coding-standard/issues/371 for the reference.

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: It's an implementation detail. I'd rather remove it completely. The end result of using the builder is the query, not the parts.

BenMorel:
> Please try phpcbf: (...)

Kills readability in this specific case, if you ask me. Should I proceed anyway?

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: OK to remove [this test](https://github.com/doctrine/dbal/blob/dbd94c9d591e3d6f980f3a077cfa5b64c600b6d7/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php#L733) then?

lib/Doctrine/DBAL/Query/QueryParts.php
BenMorel: Do you want to merge QueryParts into QueryBuilder?

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: Yes.

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: No. In the cases I'm aware of, the `reset*()` methods indicate poor design. E.g. one method builds a query with `JOIN`s and `ORDER BY`, then the other one resets the `SELECT` and `ORDER BY` parts and adds `COUNT(*)`.

Instead, only the needed parts of the query should be built for each case.

Let's keep the existing methods for now but not add any new.

morozov:
> Kills readability in this specific case, if you ask me. Should I proceed anyway?

I agree (hence the issue). However, I'd prefer to format the code according to the standard instead of having to maintain an exception in `phpcs.xml`.

lib/Doctrine/DBAL/Query/QueryParts.php
morozov: Yes.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: > Let's keep the existing methods for now but not add any new.

Which ones? The existing methods are generic `resetQueryPart[s]()`, which we're removing anyway. Should I remove all `resetXXX()` methods then?

morozov: Oh, I thought we had some methods like `resetJoin()`. Let's not introduce any.

BenMorel:
> I agree (hence the issue). However, I'd prefer to format the code according to the standard instead of having to maintain an exception in phpcs.xml.

Fixed in 00b4f898f0bde8e5acbb0d70f27126adff39fcb4.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Done in 682b5bb364ddf26d85f16bf9cf63f1565aeb0c42.

lib/Doctrine/DBAL/Query/QueryParts.php
BenMorel: Done in 42c60625a914dc223d27047afe10b2ea2df793ca.

BenMorel:
> > I agree (hence the issue). However, I'd prefer to format the code according to the standard instead of having to maintain an exception in phpcs.xml.

> Fixed in 00b4f89.

And obsoleted by 42c60625a914dc223d27047afe10b2ea2df793ca. Problem solved.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Done in 6369a4a2cdcd3c240c4a9e09fe5fa5473ffc0ac0.

BenMorel:
@morozov All good for me. Let me know if you see anything else!

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -135,6 +111,83 @@ class QueryBuilder
      */
     private $boundCounter = 0;
 
+    /**
+     * The SELECT parts of the query.
+     *
+     * @var string[]
+     */
+    public $select = [];
```
morozov: In this case, it should be `private`. We made the other class'es properties public to avoid having to introduce getters but here it's not necessary.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -454,7 +464,7 @@ public function add(string $sqlPartName, $sqlPart, bool $append = false) : self
      *         ->leftJoin('u', 'phonenumbers', 'p', 'u.id = p.user_id');
      * </code>
      *
-     * @param mixed $select The selection expressions.
+     * @param string|string[] $select The selection expressions.
```
morozov: This will probably change after a rebase. Please rebase first.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -999,7 +1072,11 @@ public function orHaving($having)
      */
     public function orderBy(string $sort, ?string $order = null)
     {
-        return $this->add('orderBy', $sort . ' ' . (! $order ? 'ASC' : $order), false);
+        $this->orderBy = [$sort . ' ' . (! $order ? 'ASC' : $order)];
```
morozov: Let's remove the default `ASC`. It's implied by the SQL semantics of `ORDER BY` and doesn't need to be explicitly specified. Additionally, please use `if ($order !== null)` instead of `!$order` for better code coverage and type clarity.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1074,14 +1101,14 @@ public function resetQueryPart(string $queryPartName) : self
...
-                  implode(', ', $this->sqlParts['select']);
+        $query = 'SELECT ' . ($this->distinct ? 'DISTINCT ' : '') .
+                  implode(', ', $this->select);
 
-        $query .= ($this->sqlParts['from'] ? ' FROM ' . implode(', ', $this->getFromClauses()) : '')
-            . ($this->sqlParts['where'] !== null ? ' WHERE ' . ((string) $this->sqlParts['where']) : '')
-            . ($this->sqlParts['groupBy'] ? ' GROUP BY ' . implode(', ', $this->sqlParts['groupBy']) : '')
-            . ($this->sqlParts['having'] !== null ? ' HAVING ' . ((string) $this->sqlParts['having']) : '')
-            . ($this->sqlParts['orderBy'] ? ' ORDER BY ' . implode(', ', $this->sqlParts['orderBy']) : '');
+        $query .= ($this->from ? ' FROM ' . implode(', ', $this->getFromClauses()) : '')
```
morozov: I'd replace the ternary operators with a series of `if` statements for more precise code coverage.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1074,14 +1101,14 @@ public function resetQueryPart(string $queryPartName) : self
...
+            . ($this->where !== null ? ' WHERE ' . ((string) $this->where) : '')
```
morozov: The `(string)` casting seems unnecessary.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1147,27 +1173,27 @@ private function isLimitQuery() : bool
...
      * Converts this instance into an UPDATE string in SQL.
      */
     private function getSQLForUpdate() : string
     {
-        return 'UPDATE ' . $this->sqlParts['table']
-            . ' SET ' . implode(', ', $this->sqlParts['set'])
-            . ($this->sqlParts['where'] !== null ? ' WHERE ' . ((string) $this->sqlParts['where']) : '');
+        return 'UPDATE ' . $this->table
+            . ' SET ' . implode(', ', $this->set)
+            . ($this->where !== null ? ' WHERE ' . ((string) $this->where) : '');
```
morozov: Please use an `if` and remove `(string)` if possible.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1147,27 +1173,27 @@ private function isLimitQuery() : bool
...
     }
 
     /**
      * Converts this instance into a DELETE string in SQL.
      */
     private function getSQLForDelete() : string
     {
-        return 'DELETE FROM ' . $this->sqlParts['table'] . ($this->sqlParts['where'] !== null ? ' WHERE ' . ((string) $this->sqlParts['where']) : '');
+        return 'DELETE FROM ' . $this->table . ($this->where !== null ? ' WHERE ' . ((string) $this->where) : '');
```
morozov: Please use an `if` and remove `(string)` if possible.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1256,9 +1282,8 @@ private function getSQLForJoins(string $fromAlias, array &$knownAliases) : strin
     {
         $sql = '';
 
-        if (isset($this->sqlParts['join'][$fromAlias])) {
-            /** @var Join $join */
-            foreach ($this->sqlParts['join'][$fromAlias] as $join) {
+        if (isset($this->join[$fromAlias])) {
```
morozov: Could an eary return be used here?

morozov (commented):
> All good for me. Let me know if you see anything else!

Apart from the nitpicks, looks good. Love the removal of the `add()` implementation.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Oops, obviously. Copy/paste mistake. Fixed.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Done in 2aac6882b1349eb7eb3cf33be8c6b8a4518b10ad.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -135,6 +111,83 @@ class QueryBuilder
...
+     * @var string[]
+     */
+    private $set = [];
+
+    /**
+     * The WHERE part of a SELECT, UPDATE or DELETE query.
+     *
+     * @var CompositeExpression|null
+     */
+    private $where = null;
```
morozov: The `= null` is redundant here.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Indeed, lots of changes upstream! Rebased now.
I love the variadics for the expressions, this is precisely what I was about to tackle next.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Done in 7a43c6a6f5c9bd3a065ea32782eba82fb42b750f.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Indeed, it's implicit, and phpstan doesn't even complain.
Done in 7a43c6a6f5c9bd3a065ea32782eba82fb42b750f.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Done in 08e9edf9fddeb4883dbb7186a438c93611fcc4b7.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Done in 08e9edf9fddeb4883dbb7186a438c93611fcc4b7.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Sure. 81d8c800b5fe36610ee5c735b803d72ed1a405da.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: I'm always advocating these default `null` assignments, even if currently redundant, as this will be the correct way to initialize the property to `null` once typed properties will be introduced (failure to do so would throw an `Error` on access), and as such will ease code conversion at that point. But it looks like I'm failing to get converts.

So, half-heartedly done in 3eea67ff8f992f9701fba778373aa1b8b4610849.

BenMorel:
@morozov All done! Ready for your review.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1059,14 +1108,33 @@ public function resetQueryPart(string $queryPartName) : self
...
     {
-        $query = 'SELECT ' . ($this->sqlParts['distinct'] ? 'DISTINCT ' : '') .
-                  implode(', ', $this->sqlParts['select']);
+        $query = 'SELECT ';
+
+        if ($this->distinct) {
+            $query .= 'DISTINCT ';
+        }
+
+        $query .= implode(', ', $this->select);
```
morozov: Is there a way that this query will be valid if the select block is empty? We probably should throw a `QueryException` here.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1088,8 +1156,7 @@ private function getFromClauses() : array
         $knownAliases = [];
 
         // Loop through all FROM clauses
```
morozov: Looks useless.

BenMorel: What do you suggest?

morozov: Please remove it.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1059,14 +1108,33 @@ public function resetQueryPart(string $queryPartName) : self
      */
     private function getSQLForSelect() : string
     {
-        $query = 'SELECT ' . ($this->sqlParts['distinct'] ? 'DISTINCT ' : '') .
-                  implode(', ', $this->sqlParts['select']);
+        $query = 'SELECT ';
```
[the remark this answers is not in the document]
BenMorel: Done in c7d0e1f8743b687a183e7dbcecae71f602526dea.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Right. Done in ffc65d42e80e87547d56847263fc46c628ccc95b.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Oh, you mean the comment. Done in d3ba7e67c3e9c27b6efef48ed507b36c735a1a7d.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -767,7 +819,21 @@ public function where($predicate, ...$predicates) : self
      */
     public function andWhere($predicate, ...$predicates) : self
     {
-        return $this->appendPredicates('where', CompositeExpression::TYPE_AND, $predicate, ...$predicates);
+        $allPredicates = array_merge([$predicate], $predicates);
+
+        if ($this->where !== null && $this->where->getType() === CompositeExpression::TYPE_AND) {
+            $this->where->addMultiple($allPredicates);
```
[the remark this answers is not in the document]
BenMorel: Sorry, I had a hard time resolving the conflicts while rebasing, and didn't really notice the work you had done factorizing this code.

Fixed in b3a437269003ff4b0e2b0cbf424425b3b362dc25.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -925,52 +1001,44 @@ public function andHaving($predicate, ...$predicates) : self
...
-        if ($predicate instanceof CompositeExpression && $predicate->getType() === $type) {
-            $predicate->addMultiple($predicates);
-        } else {
-            $predicate = new CompositeExpression(
-                $type,
-                array_merge([$predicate], $predicates)
-            );
+        if ($predicate !== null) {
+            if ($predicate->getType() === $type) {
+                return $predicate->addMultiple($predicates);
```
BenMorel: I'm a bit uncomfortable with this mutable `CompositeExpression`, I feel like it's a can of worms that can lead to some nasty bugs.

Example (pure speculation):

- I pass in a single `CompositeExpression` I created, to both `select()` and `having()`;
- This is already a `CompositeExpression`, single argument, so it's directly assigned to `$select` and `$having`;
- Now I `addSelect()`; the original `CompositeExpression` will be mutated, affecting both the value in `$select` and `$having`!

I haven't tested it, but I'm pretty sure there's a bug there, ready to be discovered. Again, let's not fix bugs & refactor in a single PR. If you wish, I can later open a PR to make `CompositeExpression` immutable, which will prevent any such issue.

BenMorel:
@morozov Ready for review, once again.

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: I had a very similar idea while working on this code. Let's put it aside and get back to it later.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -925,52 +1001,44 @@ public function andHaving($predicate, ...$predicates) : self
...
-     * Replaces any previously specified predicates.
+     * Creates a CompositeExpression from one or more predicates combined by the AND logic.
      *
+     * @param string|CompositeExpression $predicate
      * @param string|CompositeExpression ...$predicates
-     *
-     * @return $this This QueryBuilder instance.
      */
-    private function setPredicates(string $clause, ...$predicates) : self
+    private function createPredicate($predicate, ...$predicates) : CompositeExpression
```
morozov: There's probably something wrong in my refactoring or here. If a single argument is passed, it must be returned as-is which means that it's not always a composite expression. We're definitely missing a test for this case.

BenMorel: I don't understand. The result is directly assigned to `$where` or `$having`, so it has to be a `CompositeExpression`. The code checks if a single `CompositeExpression` is passed, in which case it uses it as is. Otherwise, it wraps it in an `AND` expression, which accepts both types.

Not sure what you mean then, feel free to explain further, or provide a failing test case!

morozov: There's no failing test case since the problem is in the internal object state:
```php
$qb = $conn->createQueryBuilder()
    ->select('*')
    ->where('1 = 1');
var_dump($qb->where); // pretend that it's public
```
Inspect the value of the `$where` property. It should be a string, not a `CompositeExpression`.

BenMorel: Currently it will be a `CompositeExpression`, which is what I expect. The property type is `CompositeExpression|null`.

IIRC when I refactored the original code, I inferred the same `CompositeExpression|null` type, a `string` was never assigned to `$sqlParts['where']` as far as I remember.

Anyway, what's wrong here?

morozov: > The property type is `CompositeExpression|null`.

It should be `string|CompositeExpression|null`.

> IIRC when I refactored the original code, I inferred the same `CompositeExpression|null` type, a `string` was never assigned to `$sqlParts['where']` as far as I remember.

Was it before of after the rebase? I may have mistakenly changed in a previous PR.

> Anyway, what's wrong here?

A `CompositeExpression` that has only one value is wrong:

1. It's not a composition, it's just one expression.
2. The and/or logic is not applicable to just one expression. Later, the class constructor should be reworked to `__construct($expression1, $expression2, ...$expressions)` to reflect that.

BenMorel: > Was it before of after the rebase? I may have mistakenly changed in a previous PR.

Before.

> A CompositeExpression that has only one value is wrong

Then back to the drawing board.

BenMorel: Just for the sake of exhaustiveness, there are at least 2 other ways to proceed:

- introduce an `Expression` interface, that would have `SingleExpression` and `CompositeExpression` implementations
- just use `string`, and wrap strings when composing. This has the drawback of adding unnecessary parentheses around expressions

That being said, are you sure you want to follow the `string|CompositeExpression|null` route?

morozov: > - introduce an `Expression` interface, that would have `SingleExpression` and `CompositeExpression` implementations

I see where you're coming from but if we take this approach, what would be the methods exposed on the interface? `__toString()`. And then how will a `SingleExpression` instance be different from a `string` from the consumer’s standpoint?

> - just use `string`, and wrap strings when composing. This has the drawback of adding unnecessary parentheses around expressions

You never know if they are necessary or not and this is a different problem. Consider the case:
```php
$expr->orX('foo = bar AND baz = qux', '1 = 1');
```
We may want to check how it currently works.

> That being said, are you sure you want to follow the `string|CompositeExpression|null` route?

It still looks the most reasonable to me although inelegant.

morozov: Another, purely internal downside of the current implementation (tested on `2.10`):
```php
$qb = $conn->createQueryBuilder()
    ->select('*')
    ->where('1 = 1')
    ->orWhere('2 = 2');
var_dump($qb->getQueryPart('where'));

/*
object(Doctrine\DBAL\Query\Expression\CompositeExpression)#9 (2) {
  ["type":"Doctrine\DBAL\Query\Expression\CompositeExpression":private]=>
  string(2) "OR"
  ["parts":"Doctrine\DBAL\Query\Expression\CompositeExpression":private]=>
  array(2) {
    [0]=>
    object(Doctrine\DBAL\Query\Expression\CompositeExpression)#8 (2) {
      ["type":"Doctrine\DBAL\Query\Expression\CompositeExpression":private]=>
      string(3) "AND"
      ["parts":"Doctrine\DBAL\Query\Expression\CompositeExpression":private]=>
      array(1) {
        [0]=>
        string(5) "1 = 1"
      }
    }
    [1]=>
    string(5) "2 = 2"
  }
}
*/
```
The query obviously has to contain an `OR`-composition of `1 = 1` and `2 = 2` but instead it has a nested `AND`-composition internally.

morozov: Another fine example of the current API (just for the reference):
```php
$qb->select('*')
    ->where($qb->expr()->andX());
echo $qb;
// SELECT * WHERE ()
```

BenMorel: > I see where you're coming from but if we take this approach, what would be the methods exposed on the interface? `__toString()`. And then how will a `SingleExpression` instance be different from a `string` from the consumer’s standpoint?

Indeed, the only public method would be `__toString()`. But `CompositeExpression` would have an additional method to add more expressions, knowing if parentheses are required or not. With a blind `string`, you don't have a choice but to add parentheses everywhere.

But I honestly don't mind, I'll do with `CompositeExpression|string`.

BenMorel: Done in 1a596fb795a0abd597c3aa99db9517f85792e1e3.

morozov: > With a blind `string`, you don't have a choice but to add parentheses everywhere.

So that we're clear, could you provide a code example that reproduces this problem?

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: Filed https://github.com/doctrine/dbal/issues/3844.

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: > Another fine example of the current API (just for the reference) […]

Filed https://github.com/doctrine/dbal/issues/3845.

BenMorel: > So that we're clear, could you provide a code example that reproduces this problem?

With plain strings:

`'a=1 OR b=2'` AND `c=3` => `(a=1 OR b=2) AND (c=3)`
`'a=1 AND b=2'` AND `c=3` => `(a=1 AND b=2) AND (c=3)`

With Expression objects (including a SingleExpression):

`'a=1 OR b=2'` AND `c=3` => `(a=1 OR b=2) AND c=3`
`'a=1 AND b=2'` AND `c=3` => `a=1 AND b=2 AND c=3`

Expression combination could be done on the interface itself:

```php
public function and(Expression $expr);
public function or(Expression $expr);
public function __toString();
```

morozov: How would the approach with expression objects work in this case?
`a=1 AND b=2` OR `c=3`

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -925,52 +1001,53 @@ public function andHaving($predicate, ...$predicates) : self
...
+        if ($currentPredicate === null && count($predicates) === 1) {
+            return $predicates[0];
+        }
+
+        if ($currentPredicate instanceof CompositeExpression && $currentPredicate->getType() === $type) {
+            return $currentPredicate->addMultiple($predicates);
         }
 
-        return $this->add($clause, $predicate, true);
+        if ($currentPredicate !== null) {
```
morozov: There are two inverse `if` statements: `if ($currentPredicate === null)` and `if ($currentPredicate !== null)`. Would it be possible to replace them with an `if`/`else`?

BenMorel: Hardly. There's this `if (...) return` in-between, and if you move the `array_unshift()` above to merge into a `if...else`, you may end up with work being done for nothing if the function returns early.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: > how the expression engine can decide whether a given part needs to be wrapped with parentheses or not given that it can be just a string that contains multiple conditions inside.

If you pass it a string, then obviously it will need parentheses. But if you pass it an Expression, it can decide whether or not the string it contains needs to be wrapped, depending on the expression type.

BenMorel: In the spirit of doing changes in small increments, I suggest we leave `string|CompositeExpression|null` for now. I may propose another PR later that reworks the expressions and fixes the mutability issue at the same time.

BenMorel: Here's a rough PoC: https://github.com/BenMorel/dbal/commit/a85c569f43bf059bd3e8daf37035d6394ca3c972.

I'm not sure whether this is really better TBH. I suggest we leave this for later, we have enough on our plate for now.

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: What I mean is:
```php
if ($currentPredicate instanceof CompositeExpression && $currentPredicate->getType() === $type) {
    return $currentPredicate->addMultiple($predicates);
}

if ($currentPredicate !== null) {
    array_unshift($predicates, $currentPredicate);
} elseif (count($predicates) === 1) {
    return $predicates[0];
}

return new CompositeExpression($type, $predicates);
```

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: > I suggest we leave this for later […]

Agree.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Indeed! Done in 3264682c463a87e17c76a5f0ec5ef13d3604ea2a.

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
```diff
@@ -327,6 +327,9 @@ public function testEmptySelect() : void
 
         self::assertSame($qb, $qb2);
         self::assertEquals(QueryBuilder::SELECT, $qb->getType());
+
+        self::expectException(QueryException::class);
```
morozov: This should be reflected in upgrade notes.

morozov (commented):
Apart from the upgrade notes, looks good. Let's get it merged.

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
BenMorel: ... and documented! Done in a02701c601f7eaf4db99192198ef91670b4095d0.
Should I squash?

UPGRADE.md
```diff
@@ -5,10 +5,9 @@
 1. The `select()`, `addSelect()`, `groupBy()` and `addGroupBy()` methods no longer accept an array of arguments. Pass each expression as an individual argument or expand an array of expressions using the `...` operator.
 2. The `select()`, `addSelect()`, `groupBy()` and `addGroupBy()` methods no longer ignore the first argument if it's empty.
 3. The `addSelect()` method can be no longer called without arguments.
-
-## BC BREAK: `QueryBuilder::insert()`, `update()` and `delete()` signatures changed
-
-These methods now require the `$table` parameter, and do not support aliases anymore.
+4. The `insert()`, `update()` and `delete()` methods now require the `$table` parameter, and do not support aliases anymore.
+5. The `add()`, `getQueryPart()`, `getQueryParts()`, `resetQueryPart()` and `resetQueryParts()` methods are removed.
+6. The `getSQL()` method now throws an expression if no `SELECT` expressions have been provided.
```
morozov: Should it be “For a `select()` query, the `getSQL()` method…”? Otherwise, it's fine to not have `SELECT` expressions defined for other types of queries.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -218,6 +269,8 @@ public function execute()
      * </code>
      *
      * @return string The SQL query string.
+     *
+     * @throws QueryException If the SELECT query is empty.
```
morozov: The "SELECT query is empty" reason will be reflected in the exception message. I'd keep it generic like “If the object doesn't represent a valid query in its current state”.

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
morozov: Yes.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Done.

UPGRADE.md
BenMorel: Done.

BenMorel:
Squashed!

morozov (approved):

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -997,76 +1084,55 @@ public function orderBy(string $sort, ?string $order = null) : self
...
-     */
-    public function resetQueryPart(string $queryPartName) : self
-    {
-        $this->sqlParts[$queryPartName] = self::SQL_PARTS_DEFAULTS[$queryPartName];
+        if ($this->from) {
+            $query .= ' FROM ' . implode(', ', $this->getFromClauses());
+        }
 
-        $this->state = self::STATE_DIRTY;
+        if ($this->where) {
```
morozov: Sorry for the last minute request, but these [issues](https://scrutinizer-ci.com/g/doctrine/dbal/inspections/4ab891ff-68b0-4ff6-a2cd-d5d7833a9c7d/issues/) found by Scrutinizer seem worth addressing. Especially this one.

Since `$this->where` is `string|ComplexExpression|null`, it can be assigned a string `'0'` and ignored. There should be an `if ($this->where !== null)` comparison.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -997,76 +1084,55 @@ public function orderBy(string $sort, ?string $order = null) : self
...
-     *
-     * @return $this This QueryBuilder instance.
+     * @throws QueryException
      */
-    public function resetQueryParts(?array $queryPartNames = null) : self
+    private function getSQLForSelect() : string
     {
-        if ($queryPartNames === null) {
-            $queryPartNames = array_keys($this->sqlParts);
+        if (! $this->select) {
```
morozov: Please use `(if count($this->select) < 1)` instead. If at some moment `$this->select` is replaced with an `interable&coutable` object, this logic will break.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -925,52 +1003,51 @@ public function andHaving($predicate, ...$predicates) : self
...
+    private function createPredicate($predicate, ...$predicates)
     {
-        if (count($predicates) > 1) {
-            $predicate = new CompositeExpression(
-                CompositeExpression::TYPE_AND,
-                $predicates
-            );
-        } else {
-            $predicate = array_shift($predicates);
+        if (! $predicates) {
```
morozov: Please use `if (count($predicates) < 1)` to treat `$predicates` as a generic collection and not depend on it being an array (same as above).

tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
morozov: Please [use](https://scrutinizer-ci.com/g/doctrine/dbal/inspections/4ab891ff-68b0-4ff6-a2cd-d5d7833a9c7d/issues/files/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php?status=new&orderField=path&order=asc&honorSelectedPaths=0) `$this->expectException()` instead.

morozov (changes requested):

morozov:
@BenMorel after getting this done, if you’re still in the mood for static analysis puzzles like this, you may want to look at https://github.com/doctrine/dbal/pull/3799. There was some progress done in the past but it seems stuck.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Good point. This was `CompositeExpression|null` before, so implicit conversion to boolean was OK. Now that's it can be a `string`, it's indeed become a potential issue.

Fixed.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: As long as they're arrays, I'm perfectly fine with implicit conversion to boolean, but you may have a point. I had Scrutinizer's suggestion to use `empty()` though: it lets undefined variables pass through.

Anyway, done.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Done.

BenMorel:
> Please use $this->expectException() instead.

Ugh. All assertions are static, and this method isn't. How can this even work??

Anyway, all fixed, and squashed again.

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: `empty()` is a ticking bomb. Should be avoided, especially in new code.

morozov (approved):

morozov:
Thanks, @BenMorel! All good this time.

PR #3840, opened 2020-01-22
Fixed the QueryBuilder::setMaxResults() signature to accept NULL

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes

#### Summary

The original signature was improperly documented using an annotation which later was converted to an argument type declaration.


morozov:
Once accepted, the relevant bits need to be backported to `2.x`.

greg0ire (approved):

PR #3841, opened 2020-01-22
Enforced argument and return types in QueryBuilder and ExpressionBuilder

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

1. Some of the `ExpressionBuilder` method arguments were mistakenly marked as `mixed` and weren't automatically converted to type declarations. A similar type enforcement was done in #3494 and #3498.
2. The `QueryBuilder` methods returning the instance are enforced to return `self`.

### DIFF
--- a/lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
+++ b/lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
@@ -72,11 +72,11 @@ class ExpressionBuilder
     /**
      * Creates a comparison expression.
      *
-     * @param mixed  $x        The left expression.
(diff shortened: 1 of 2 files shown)


greg0ire (approved):
Looks good but I'm starting to think that maybe some docblocks could be removed entirely… not sure how much value is brought by comments such as "The left expression.". We could use `$left` and `$right` instead of `$x` and `$y`

morozov:
Frankly speaking, I'm not even sure why this `comparison()` method is needed at all. It just concatenates all given values and adds a space in between. In the rest of the cases, the names of the arguments would be more obvious.

PR #3847, opened 2020-01-24
Update Doctrine Coding Standard to 7.0

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
@@ -98,32 +98,40 @@ abstract class AbstractMySQLDriver implements ExceptionConverterDriver, VersionA
         switch ($exception->getCode()) {
             case self::ER_LOCK_DEADLOCK:
                 return new Exception\DeadlockException($message, $exception);
+
             case self::ER_LOCK_WAIT_TIMEOUT:
                 return new Exception\LockWaitTimeoutException($message, $exception);
+
             case self::ER_TABLE_EXISTS_ERROR:
                 return new Exception\TableExistsException($message, $exception);
+
             case self::ER_BAD_TABLE_ERROR:
             case self::ER_NO_SUCH_TABLE:
                 return new Exception\TableNotFoundException($message, $exception);
+
             case self::ER_NO_REFERENCED_ROW:
             case self::ER_ROW_IS_REFERENCED:
             case self::ER_ROW_IS_REFERENCED_2:
             case self::ER_NO_REFERENCED_ROW_2:
             case self::ER_TRUNCATE_ILLEGAL_FK:
                 return new Exception\ForeignKeyConstraintViolationException($message, $exception);
+
             case self::ER_DUP_ENTRY:
             case self::ER_FOREIGN_DUPLICATE_KEY_OLD_UNUSED:
             case self::ER_DUP_ENTRY_AUTOINCREMENT_CASE:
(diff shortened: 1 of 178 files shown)


lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
```diff
@@ -14,12 +14,6 @@
 use ReflectionObject;
 use ReflectionProperty;
 use stdClass;
-use const CASE_LOWER;
-use const DB2_BINARY;
-use const DB2_CHAR;
-use const DB2_LONG;
-use const DB2_PARAM_FILE;
-use const DB2_PARAM_IN;
```
morozov: @carusogabriel is this reordering of `use const` and `use function` expected? Could you please point to the relevant change in the Doctrine or Slevomat coding standard? The old order at least seemed more logical (`const` and `function` blocks sorted alphabetically). I want to make sure it's not a bug.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -146,14 +147,14 @@ public function testEventManagerPassedToPlatform() : void
     {
         $eventManager = new EventManager();
 
-        /** @var AbstractPlatform|MockObject $driver */
         $platform = $this->createMock(AbstractPlatform::class);
+        assert($platform instanceof AbstractPlatform || $platform instanceof MockObject);
```
morozov: This sniff will be disabled in tests. There's no point in making assertions against what PHPUnit generates.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
carusogabriel: Please see https://github.com/doctrine/coding-standard/pull/131

morozov: I see. The order is defined by [PSR-12](https://www.php-fig.org/psr/psr-12/#3-declare-statements-namespace-and-import-statements). Thanks!

lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
```diff
@@ -105,34 +102,28 @@ public function convertException(string $message, DriverExceptionInterface $exce
                 return new Exception\LockWaitTimeoutException($message, $exception);
             case self::ER_TABLE_EXISTS_ERROR:
                 return new Exception\TableExistsException($message, $exception);
-
```
morozov: @carusogabriel could you help me understand this too? The new standard introduces new lines around control statements (must be related to https://github.com/slevomat/coding-standard/issues/472) but at the same time removes them from between `case`s.

Instead of removing the newline, could we enforce it instead?

morozov: It looks like a bug. Reported as https://github.com/slevomat/coding-standard/issues/867.

carusogabriel: This is related to https://github.com/doctrine/coding-standard/commit/e91edf00502b851bec8b0c930b4ca106d494b13b, but looks like it's something misconfigured rather than a bug

morozov: I've excluded this rule temporarily until we find the proper solution.

carusogabriel (approved):
Nice :+1:

Issue #3845, opened 2020-01-23
QueryBuilder produces invalid SQL from an empty CompositeExpression 


morozov:
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.1

```php
$qb = $conn->createQueryBuilder();
$qb->select('*')
    ->where($qb->expr()->andX());
echo $qb;
// SELECT * WHERE ()
```

SenseException:
What's the expected behaviour for this?

morozov:
Either a `QueryException` or no `WHERE` clause. I mostly filed it for documentation purposes. In `master`, I want to make the composite expression to always have at least one (or even better, two) parts. It shouldn't be possible to instantiate an empty object in the first place.

morozov:
Fixed by #3868.

PR #3850, opened 2020-01-26
Prepare CompositeExpression for immutability

### LINKED ISSUE #3844: CompositeExpression is mutable
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.1

If a composite expression is used in different parts of the same query or in different queries, its implicit mutation by the `QueryBuilder::(and|or)(Where|Having)()` methods in one place will affect the other places as well.

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

*Will fix* #3844, once merged into `master` with methods `add*()` removed.

### DIFF
--- a/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
+++ b/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
@@ -49,6 +49,8 @@ class CompositeExpression implements Countable
     /**
      * Adds multiple parts to composite expression.
      *
+     * @deprecated This class will be made immutable. Use with() instead.
+     *
      * @param self[]|string[] $parts
      *
      * @return \Doctrine\DBAL\Query\Expression\CompositeExpression
@@ -64,6 +66,8 @@ class CompositeExpression implements Countable
 
     /**
      * Adds an expression to composite expression.
+     *
+     * @deprecated This class will be made immutable. Use with() instead.
      *
      * @param mixed $part
      *
@@ -82,6 +86,25 @@ class CompositeExpression implements Countable
         $this->parts[] = $part;
 
         return $this;
+    }
+
+    /**
+     * Returns a new CompositeExpression with the given parts added.
+     *
+     * @param self|string $part
+     * @param self|string ...$parts
+     */
+    public function with($part, ...$parts) : self
+    {
+        $that = clone $this;
+
+        $that->parts[] = $part;
+
+        foreach ($parts as $part) {
+            $that->parts[] = $part;
+        }
+
+        return $that;
     }
 
     /**
--- a/tests/Doctrine/Tests/DBAL/Query/Expression/CompositeExpressionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Query/Expression/CompositeExpressionTest.php
@@ -16,7 +16,7 @@ class CompositeExpressionTest extends DbalTestCase
 
         self::assertCount(1, $expr);
 
-        $expr->add('u.group_id = 2');
+        $expr = $expr->with('u.group_id = 2');
 
         self::assertCount(2, $expr);
     }
@@ -40,6 +40,26 @@ class CompositeExpressionTest extends DbalTestCase
         self::assertCount(2, $expr);
 
         $expr->add('u.user_id = 1');
+
+        self::assertCount(3, $expr);
+    }
+
+    public function testWith() : void
+    {
+        $expr = new CompositeExpression(CompositeExpression::TYPE_OR, ['u.group_id = 1']);
+
+        self::assertCount(1, $expr);
+
+        // test immutability
+        $expr->with(new CompositeExpression(CompositeExpression::TYPE_OR, ['u.user_id = 1']));
+
+        self::assertCount(1, $expr);
+
+        $expr = $expr->with(new CompositeExpression(CompositeExpression::TYPE_OR, ['u.user_id = 1']));
+
+        self::assertCount(2, $expr);
+
+        $expr = $expr->with('u.user_id = 1');
 
         self::assertCount(3, $expr);
     }
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -4,6 +4,11 @@
 
 The usage of the `andX()` and `orX()` methods of the `ExpressionBuilder` class has been deprecated. Use `and()` and `or()` instead.
 
+## Deprecated `CompositeExpression` methods
+
+The usage of the `add()` and `addMultiple()` methods of the `CompositeExpression` class has been deprecated. Use `with()` instead, which returns a new instance.
+In the future, the `add*()` methods will be removed and the class will be effectively immutable.
+
 # Upgrade to 2.10
 
 ## Deprecated `Doctrine\DBAL\Event\ConnectionEventArgs` methods
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -823,7 +823,7 @@ class QueryBuilder
         $where = $this->getQueryPart('where');
 
         if ($where instanceof CompositeExpression && $where->getType() === CompositeExpression::TYPE_AND) {
-            $where->addMultiple($args);
+            $where = $where->with(...$args);
         } else {
             array_unshift($args, $where);
             $where = new CompositeExpression(CompositeExpression::TYPE_AND, $args);
@@ -856,7 +856,7 @@ class QueryBuilder
         $where = $this->getQueryPart('where');
 
         if ($where instanceof CompositeExpression && $where->getType() === CompositeExpression::TYPE_OR) {
-            $where->addMultiple($args);
+            $where = $where->with(...$args);
         } else {
             array_unshift($args, $where);
             $where = new CompositeExpression(CompositeExpression::TYPE_OR, $args);
@@ -998,7 +998,7 @@ class QueryBuilder
         $having = $this->getQueryPart('having');
 
         if ($having instanceof CompositeExpression && $having->getType() === CompositeExpression::TYPE_AND) {
-            $having->addMultiple($args);
+            $having = $having->with(...$args);
         } else {
             array_unshift($args, $having);
             $having = new CompositeExpression(CompositeExpression::TYPE_AND, $args);
@@ -1021,7 +1021,7 @@ class QueryBuilder
         $having = $this->getQueryPart('having');
 
         if ($having instanceof CompositeExpression && $having->getType() === CompositeExpression::TYPE_OR) {
-            $having->addMultiple($args);
+            $having = $having->with(...$args);
         } else {
             array_unshift($args, $having);
             $having = new CompositeExpression(CompositeExpression::TYPE_OR, $args);


lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
```diff
@@ -43,47 +47,48 @@ class CompositeExpression implements Countable
      */
     public function __construct(string $type, array $parts = [])
     {
-        $this->type = $type;
-
-        $this->addMultiple($parts);
+        $this->type  = $type;
+        $this->parts = array_values(array_filter($parts, static function ($part) {
+            return ! ($part instanceof self && count($part) === 0);
```
BenMorel: We'll be able to remove these checks/filters once we'll make sure that `CompositeExpression` always contains at least one part. I'll make another PR to fix #3845 and update this code later.

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
```diff
@@ -43,47 +47,48 @@ class CompositeExpression implements Countable
...
-     * Adds an expression to composite expression.
+     * Returns a CompositeExpression with the given part added.
      *
-     * @param self|string $part
+     * This instance is immutable and unaffected by this method call.
      *
-     * @return $this
+     * @param self|string $part
      */
     public function add($part) : self
```
morozov: Let’s rename it to `with()`:
1. A significant change in the method semantics without the signature change is quite dangerous.
2. The "with" naming is quite common for immutable class methods.

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
morozov: Why is this filter needed now?

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
```diff
@@ -43,47 +47,48 @@ class CompositeExpression implements Countable
...
-     * Adds multiple parts to composite expression.
+     * Returns a CompositeExpression with the given parts added.
      *
-     * @param array<int, self|string> $parts
+     * This instance is immutable and unaffected by this method call.
      *
-     * @return $this
+     * @param array<int, self|string> $parts
      */
     public function addMultiple(array $parts = []) : self
```
morozov: Let's remove and make `with()` variadic instead. There's no longer point in having two methods for the same operation.

morozov (changes requested):

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
greg0ire: If we have a new name, this could be contributed to 2.11.x as a new method (and we deprecate `add` and `addMultiple`)

BenMorel: @morozov Agreed.
@greg0ire I don't think this is a good idea, as the semantics of the class differ. You'd end up with a mutable class that has mutator methods and non-mutator methods for the same thing, which would be highly confusing IMO.

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
BenMorel: Because it accepts empty `CompositeExpression` objects, that were filtered so far by `addMultiple()`.

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
greg0ire: Wouldn't the confusion vanish on seeing that mutator methods are deprecated?

BenMorel: IDK. @morozov?

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
BenMorel: Done!

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
morozov: I believe we should introduce the new API in `2.11` w/o declaring the class itself immutable. This way, it will be at the consumer’s discretion to use one of the APIs before upgrading to the next major.

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
```diff
@@ -43,47 +47,32 @@ class CompositeExpression implements Countable
...
-     *
-     * @return $this
-     */
-    public function add($part) : self
-    {
-        if (empty($part)) {
-            return $this;
-        }
+        foreach ($parts as $part) {
+            if ($part instanceof self && count($part) === 0) {
```
morozov: Would it make sense to extract this into `private function isEmptySelf()` to generalize this logic and make its removal easier in the future?

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
BenMorel: Let's do this then. How should we proceed? Something along these lines?

1. retarget this PR on 2.11, just add `with()`
2. merge 2.11 into master
3. create another PR on master to remove mutators?

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
BenMorel: TBH I want to remove this code ASAP (i.e. in my next PR), so I don't think we should spend any more time attempting to make this code look prettier. It won't be easier to remove, either.

This "temporary" code is a small cost to pay to make one change at a time (I hesitated to resolve both #3844 and #3845 in the same PR, but decided not to, due to recent discussions).

morozov: Given that we agreed to retarget this against 2.11, let's do that first. After then, it might be not possible to remove this code ASAP since it would require a breaking change.

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
morozov: Yes, correct. Go ahead.

BenMorel: Done.

greg0ire (commented):
The current code looks good, but there is more to come, right?

BenMorel:
@greg0ire Not in this PR, no! What's missing?

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
greg0ire: @BenMorel maybe I misinterpreted but I think @morozov is telling you that we can't have temporary code on 2.11.x because we want to be able to release 2.11.0 at any given time, and removing the temporary code after that would be a BC-break

greg0ire:
See https://github.com/doctrine/dbal/pull/3850#discussion_r371978543 for why I am confused :stuck_out_tongue:

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
morozov: The extraction of this specific block to a private method won't be a BC break but this logic will have to stay in 2.x forever, so it still makes sense to generalize it.

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
```diff
@@ -84,6 +88,28 @@ public function add($part)
         return $this;
     }
 
+    /**
+     * Returns a CompositeExpression with the given part added.
+     *
+     * This methods returns a new instance. The current instance is unaffected by this method call.
```
morozov: IMO, it's too verbose. “Returns a _new_ CompositeExpression with the given part added” should suffice.

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
```diff
@@ -84,6 +88,28 @@ public function add($part)
         return $this;
     }
 
+    /**
+     * Returns a CompositeExpression with the given part added.
```
morozov: Should it be “parts” instead of “part” since the method is variadic?

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
```diff
@@ -84,6 +88,28 @@ public function add($part)
...
+     *
+     * This methods returns a new instance. The current instance is unaffected by this method call.
+     *
+     * @param self|string ...$parts
+     */
+    public function with(...$parts) : self
```
morozov: The signature should require at least one argument.

tests/Doctrine/Tests/DBAL/Query/Expression/CompositeExpressionTest.php
```diff
@@ -16,30 +16,35 @@ public function testCount() : void
...
         self::assertCount(1, $expr);
 
-        $expr->add('u.group_id = 2');
+        $expr = $expr->with('u.group_id = 2');
 
         self::assertCount(2, $expr);
     }
 
-    public function testAdd() : void
+    public function testAddWith() : void
```
morozov: Should it be `testWith()` corresponding to the method being tested?

tests/Doctrine/Tests/DBAL/Query/Expression/ExpressionBuilderTest.php
```diff
@@ -45,11 +45,7 @@ public function testAnd(array $parts, string $expected) : void
      */
     public function testAndX(array $parts, string $expected) : void
     {
-        $composite = $this->expr->andX();
-
-        foreach ($parts as $part) {
-            $composite->add($part);
-        }
+        $composite = $this->expr->andX()->with(...$parts);
```
morozov: Since this test is about the legacy API, should we leave it untouched to guarantee the old API still works?

tests/Doctrine/Tests/DBAL/Query/Expression/CompositeExpressionTest.php
BenMorel: Both are tested right now, the legacy `add()` and the new `with()`.

tests/Doctrine/Tests/DBAL/Query/Expression/ExpressionBuilderTest.php
BenMorel: Makes sense, I guess. Fixed in 19a850de13858ba4dbc1d7d9bb3386883a4205e5.

tests/Doctrine/Tests/DBAL/Query/Expression/CompositeExpressionTest.php
morozov: Why do we want to have one test for two APIs? It should be structured the way that enables easy removal of the no longer needed test in `master`.

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
BenMorel: Fixed in ea51a40e342861a739108311298f8d8ccdf38ed3 and 2b01c576ebe2206683bceaad0c7d5f26f77850fd.

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
BenMorel: Indeed. Fixed in 8b7fbdafdaf956c0c1b6f6f6bf9a90d440c1c339.

tests/Doctrine/Tests/DBAL/Query/Expression/CompositeExpressionTest.php
BenMorel: Indeed, fixed in 2df2256c731b6cba4d3875c73e04859c670e83be.

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
BenMorel: Updated in d191e953778c24efb0fd4bd3a4752e443563239e.

BenMorel:
> The extraction of this specific block to a private method won't be a BC break but this logic will have to stay in 2.x forever, so it still makes sense to generalize it.

Yes. Done in 2b01c576ebe2206683bceaad0c7d5f26f77850fd.

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
BenMorel: I'm not sure whether we're on the same page here, changes to internal code don't matter for BC, and I just meant I didn't want to spend too much time on stuff that is going to change in `master` again a few days later. But anyway, I've applied the suggestion in 2b01c576ebe2206683bceaad0c7d5f26f77850fd!

BenMorel:
@greg0ire I hope the confusion is fixed. Please tell me if you see more changes to be made!

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
```diff
@@ -84,6 +88,39 @@ public function add($part)
...
+     * Returns a new CompositeExpression with the given parts added.
+     *
+     * @param self|string $part
+     * @param self|string ...$parts
+     */
+    public function with($part, ...$parts) : self
+    {
+        $that = clone $this;
+
+        if (! $this->isEmptySelf($part)) {
```
morozov: I'm not sure if these checks need to be added to the new API. Consumers will opt-in to using it, so they will need to make sure they are not using empty composite expressions. Instead of sweetening the pill for them, we should make sure their code is ready for the upgrade.

greg0ire (approved):

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
BenMorel: That's a good point. At some point, we want to disallow empty expressions. However, we can't do this in 2.x, so what should be do here?

- leave as is: skip empty expressions
- don't make this check: empty expressions will be added to the expression, and produce invalid SQL
- throw an exception?

morozov: I'd just remove the check from here. It's already stated in the code documentation that by the compilation time an expression cannot be empty:

https://github.com/doctrine/dbal/blob/d191e953778c24efb0fd4bd3a4752e443563239e/lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php#L80-L81

BenMorel: Makes sense: no sane code base should ever contain an empty expression anyway, as this would produce invalid SQL. Done in 1341e462da44e4d5734c96d765ef8008029ca2f7.

BenMorel:
Sorry for closing/reopening, I have fat fingers tonight.

Should be all good now. Please note that I reverted `isEmptySelf()`, it was only being used in one place now.

morozov:
Looks neat. Please squash.

BenMorel:
@morozov Squashed.

morozov (approved):

morozov:
@BenMorel is there anything else you'd like to add to 2.x or it's time for an up-merge?

BenMorel:
@morozov All good for me, time for an up-merge!

morozov:
Thanks, @BenMorel!

PR #3851, opened 2020-01-26
Rename andX() / orX() methods

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

`ExpressionBuilder::andX()` and `orX()` have likely been named as such because before PHP 7 and its context sensitive lexer, methods could not be named `and()` or `or()`.

I think it's time to change this for the next major version.

### DIFF
--- a/lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
+++ b/lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
@@ -39,13 +39,27 @@ class ExpressionBuilder
     }
 
     /**
-     * Creates a conjunction of the given boolean expressions.
+     * Creates a conjunction of the given expressions.
      *
-     * Example:
+     * @param string|CompositeExpression ...$expressions Requires at least one defined when converting to string.
+     */
+    public function and(...$expressions) : CompositeExpression
+    {
+        return new CompositeExpression(CompositeExpression::TYPE_AND, $expressions);
+    }
+
+    /**
+     * Creates a disjunction of the given expressions.
      *
-     *     [php]
-     *     // (u.type = ?) AND (u.role = ?)
-     *     $expr->andX('u.type = ?', 'u.role = ?'));
+     * @param string|CompositeExpression ...$expressions Requires at least one defined when converting to string.
+     */
+    public function or(...$expressions) : CompositeExpression
+    {
+        return new CompositeExpression(CompositeExpression::TYPE_OR, $expressions);
+    }
+
+    /**
+     * @deprecated Use `and()` instead.
      *
      * @param mixed $x Optional clause. Defaults = null, but requires
      *                 at least one defined when converting to string.
@@ -54,17 +68,11 @@ class ExpressionBuilder
      */
     public function andX($x = null)
     {
-        return new CompositeExpression(CompositeExpression::TYPE_AND, func_get_args());
+        return $this->and(...func_get_args());
     }
 
     /**
-     * Creates a disjunction of the given boolean expressions.
-     *
-     * Example:
-     *
-     *     [php]
-     *     // (u.type = ?) OR (u.role = ?)
-     *     $qb->where($qb->expr()->orX('u.type = ?', 'u.role = ?'));
+     * @deprecated Use `or()` instead.
      *
      * @param mixed $x Optional clause. Defaults = null, but requires
      *                 at least one defined when converting to string.
@@ -73,7 +81,7 @@ class ExpressionBuilder
      */
     public function orX($x = null)
     {
-        return new CompositeExpression(CompositeExpression::TYPE_OR, func_get_args());
+        return $this->or(...func_get_args());
     }
 
     /**
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,9 @@
+# Upgrade to 2.11
+
+## Deprecated `ExpressionBuilder` methods
+
+The usage of the `andX()` and `orX()` methods of the `ExpressionBuilder` class has been deprecated. Use `and()` and `or()` instead.
+
 # Upgrade to 2.10
 
 ## Deprecated `Doctrine\DBAL\Event\ConnectionEventArgs` methods
--- a/docs/en/reference/query-builder.rst
+++ b/docs/en/reference/query-builder.rst
@@ -332,13 +332,13 @@ Most notably you can use expressions to build nested And-/Or statements:
         ->select('id', 'name')
         ->from('users')
         ->where(
-            $queryBuilder->expr()->andX(
+            $queryBuilder->expr()->and(
                 $queryBuilder->expr()->eq('username', '?'),
                 $queryBuilder->expr()->eq('email', '?')
             )
         );
 
-The ``andX()`` and ``orX()`` methods accept an arbitrary amount
+The ``and()`` and ``or()`` methods accept an arbitrary amount
 of arguments and can be nested in each other.
 
 There is a bunch of methods to create comparisons and other SQL snippets
--- a/tests/Doctrine/Tests/DBAL/Query/Expression/ExpressionBuilderTest.php
+++ b/tests/Doctrine/Tests/DBAL/Query/Expression/ExpressionBuilderTest.php
@@ -29,11 +29,11 @@ class ExpressionBuilderTest extends DbalTestCase
     /**
      * @param string[]|CompositeExpression[] $parts
      *
-     * @dataProvider provideDataForAndX
+     * @dataProvider provideDataForAnd
      */
-    public function testAndX(array $parts, string $expected) : void
+    public function testAnd(array $parts, string $expected) : void
     {
-        $composite = $this->expr->andX();
+        $composite = $this->expr->and();
 
         foreach ($parts as $part) {
             $composite->add($part);
@@ -45,7 +45,7 @@ class ExpressionBuilderTest extends DbalTestCase
     /**
      * @return mixed[][]
      */
-    public static function provideDataForAndX() : iterable
+    public static function provideDataForAnd() : iterable
     {
         return [
             [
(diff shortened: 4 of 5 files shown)


greg0ire:
IMO it would make sense to target 2.10 and deprecate `andX` / `orX` instead of removing them.

BenMorel:
@greg0ire I'm definitely in favour of adding `and()`/`or()` to `2.x` and deprecate `andX()`/`orX()`. I can't do it in `2.10` though, it has to be in a minor version according to semver, not a patch version.

@morozov, is there a `2.11` version planned, and in this case, is there a branch for it?

Note that I would definitely remove them in `3.0`, you don't want to carry BC stuff for too long. One minor version would be enough.

greg0ire:
> is there a 2.11 version planned, and in this case, is there a branch for it?

Also, why is there a 2.10 and 2.10.x branch?

morozov:
> @morozov, is there a `2.11` version planned, and in this case, is there a branch for it?

Yes. It is planned primarily to introduce all needed deprecations and provide a forward compatibility layer. I've just created [`2.11.x`](https://github.com/morozov/dbal/tree/2.11.x) from [`v2.10.1`](https://github.com/doctrine/dbal/commits/v2.10.1).

morozov:
> Note that I would definitely remove them in `3.0`, you don't want to carry BC stuff for too long. One minor version would be enough.

The original purpose of 3.0 is to introduce the minimal needed BC break for compatibility with PHP 8 and other trivial to handle deprecations (e.g. removal of no longer supported DB platforms). We're going to postpone the remaining API changes for 4.0.

morozov:
> Also, why is there a 2.10 and 2.10.x branch?

`2.10` is what we used historically for `2.x` branches. `2.10.x` is created by the @doctrinebot after each release in the `2.10.x` series.

morozov:
@BenMorel please retarget against `2.11.x`.

BenMorel:
> The original purpose of 3.0 is to introduce the minimal needed BC break for compatibility with PHP 8 and other trivial to handle deprecations (e.g. removal of no longer supported DB platforms). We're going to postpone the remaining API changes for 4.0.

Oh,my. When is 4.0 expected? I was hoping all the work done here would be released in the coming months, for a project I wanted based on the latest DBAL.

morozov:
> The original purpose of 3.0 is […]

Sorry, not the best choice of words. 4.0 will be released at the same time when the original 3.0 would be which is undetermined. In order to address compatibility with PHP 8 (https://github.com/doctrine/dbal/pull/3803), we want to make a small 3.0 release and continue working on the rest of the planned breaking changes.

BenMorel:
> I've just created 2.11.x from v2.10.1.

Looks like you only created the branch in your fork, not upstream.

greg0ire:
Indeed. I just created it. @morozov I don't know if there are settings to tweak but I didn't tweak any.

BenMorel:
Retargeted against `2.11.x` and deprecated methods.

morozov:
> Looks like you only created the branch in your fork, not upstream.

Indeed. Thanks, @greg0ire. I added the branch to Scrutinizer and that should be enough.

lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
```diff
@@ -71,6 +71,32 @@ public function andX($x = null)
...
+     * @deprecated Use `and()` instead.
+     *
+     * @param mixed $x Optional clause. Defaults = null, but requires
+     *                 at least one defined when converting to string.
+     *
+     * @return CompositeExpression
+     */
+    public function andX($x = null)
+    {
+        return new CompositeExpression(CompositeExpression::TYPE_AND, func_get_args());
```
morozov: Why duplicate the code instead of calling the new method?

morozov:
In light of https://github.com/doctrine/coding-standard/pull/150, I'm not requesting the addition of `@trigger_error()`.

greg0ire:
In both cases, it should not be added until the deprecation is addressed in all dependent packages if any (and in the code of DBAL itself of course).

morozov:
That's my problem with the approach. It is by definition deprecated as soon as there's another API to use for the same purpose. Documenting it one way or the other depending on the state of the world outside is BS.

lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
BenMorel: It's a one-liner so I hesitated. But you're right. Fixed.

lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
```diff
@@ -52,7 +52,7 @@ public function __construct(Connection $connection)
      *
      * @return CompositeExpression
      */
-    public function andX($x = null)
+    public function and($x = null)
```
BenMorel: Because these are new methods, should we add variadics right now to align the definition of these methods with upcoming 3.0 or 4.0, and ease transition later?

morozov: Definitely. They should look as close to what the consumers will get in the new versions as possible.

morozov: Do we want the first argument to be optional here?

BenMorel: > Definitely. They should look as close to what the consumers will get in the new versions as possible.

OK done!

> Do we want the first argument to be optional here?

That's the job of #3845. I'm getting lost over what to do first, and on which branch.

lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
```diff
@@ -39,41 +39,61 @@ public function __construct(Connection $connection)
...
 
     /**
-     * Creates a conjunction of the given boolean expressions.
+     * Creates a conjunction of the given expressions.
      *
      * Example:
      *
      *     [php]
      *     // (u.type = ?) AND (u.role = ?)
      *     $expr->andX('u.type = ?', 'u.role = ?'));
```
morozov: Should we just get rid of this example? We're too used to getting being covered by static analysis to maintain code examples in comments.

greg0ire:
>  Documenting it one way or the other depending on the state of the world outside is BS.

Well you can document it with `@deprecated` from day one. `@trigger_error` is way more aggressive and can have implication on production systems, so it should be added when you're sure people are not going to be stuck with unfixable deprecations, because it is stronger than just "documenting".

Anyway, when I see the debate in the link above, I think what is going to happen is that there will be deprecations with `@trigger_error`, and packages without, and DBAL is probably going to be one of the packages without because you're clearly the owner of that package and if you're not fully on board with such a decision, it's not ok to push it in this package.

I think we all agree to use `@deprecated` in this PR, and that's great.

lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
morozov: The one-liner will show how the new API will need to be called instead of the old one.

greg0ire:
Please update `docs/en/reference/query-builder.rst` as well.
@morozov should any mention of deprecated be removed entirely, or should we keep them in a dedicated section / use specific rst markup that will allow to document them while discouraging their usage?

morozov:
> […] should any mention of deprecated be removed entirely, or should we keep them in a dedicated section / use specific rst markup that will allow to document them while discouraging their usage?

I haven’t made any documentation changes like this in the past. If it is possible to document both and mark the old API deprecated, that would be nice. Otherwise, we can replace the old examples with the new ones.

lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
BenMorel: Good point, totally missed that. Fixed in 2c69e551c6138cf46fa7c6c3377ed1774e564f65.

BenMorel:
> please update docs/en/reference/query-builder.rst as well.

Good point. Done in 625a70dbf0dc8bc35db0e9aa09d4677d66d40668.

greg0ire (approved):

morozov:
@BenMorel, looks good. Please squash.

lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
```diff
@@ -54,17 +68,11 @@ public function __construct(Connection $connection)
...
 
     /**
-     * Creates a disjunction of the given boolean expressions.
-     *
-     * Example:
-     *
-     *     [php]
-     *     // (u.type = ?) OR (u.role = ?)
-     *     $qb->where($qb->expr()->orX('u.type = ?', 'u.role = ?'));
+     * @deprecated Use `and()` instead.
```
morozov: Copy/paste.

morozov (changes requested):

lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
BenMorel: Caught! Fixed.

BenMorel:
@morozov Squashed!

morozov (approved):

BenMorel:
@morozov Is there anything blocking the merge?

morozov:
Thanks for the reminder. Merged.

BenMorel:
Thank you.

> Do we want the first argument to be optional here?

Should we now make the first argument mandatory in 2.11?

morozov:
Yes, the new API should look like in master to provide forward compatibility.

BenMorel:
It's not done in master yet, but I guess you'll merge it into `master` later, removing the `andX`/`orX`?

morozov:
Yeah. We’ll build FC and deprecate the old APIs first, then merge stuff up and remove the depreciated code.

PR #3853, opened 2020-01-28
Deprecate calling QueryBuilder methods with an array argument

### LINKED ISSUE #3837: Deprecate the usage of arrays in QueryBuilder methods
The usage of array arguments in `QueryBuilder::select()`, `::addSelect()`, `::groupBy()` and `::addGroupBy()` methods needs to be deprecated.

Their support was removed in https://github.com/doctrine/dbal/pull/3831.

BenMorel:
Fixes #3837.

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -451,6 +451,8 @@ class QueryBuilder
      * Specifies an item that is to be returned in the query result.
      * Replaces any previously specified selections, if any.
      *
+     * USING AN ARRAY ARGUMENT IS DEPRECATED. Pass each value as an individual argument.
+     *
      * <code>
      *     $qb = $conn->createQueryBuilder()
      *         ->select('u.id', 'p.id')
@@ -458,11 +460,12 @@ class QueryBuilder
      *         ->leftJoin('u', 'phonenumbers', 'p', 'u.id = p.user_id');
      * </code>
      *
-     * @param mixed $select The selection expressions.
+     * @param string|string[]|null $select The selection expression. USING AN ARRAY OR NULL IS DEPRECATED.
+     *                                     Pass each value as an individual argument.
      *
      * @return $this This QueryBuilder instance.
      */
-    public function select($select = null)
+    public function select($select = null/*, string ...$selects*/)
     {
         $this->type = self::SELECT;
 
@@ -497,6 +500,8 @@ class QueryBuilder
     /**
      * Adds an item that is to be returned in the query result.
      *
+     * USING AN ARRAY ARGUMENT IS DEPRECATED. Pass each value as an individual argument.
+     *
      * <code>
      *     $qb = $conn->createQueryBuilder()
      *         ->select('u.id')
@@ -505,11 +510,12 @@ class QueryBuilder
      *         ->leftJoin('u', 'phonenumbers', 'u.id = p.user_id');
      * </code>
      *
-     * @param mixed $select The selection expression.
+     * @param string|string[]|null $select The selection expression. USING AN ARRAY OR NULL IS DEPRECATED.
+     *                                     Pass each value as an individual argument.
      *
      * @return $this This QueryBuilder instance.
      */
-    public function addSelect($select = null)
+    public function addSelect($select = null/*, string ...$selects*/)
     {
         $this->type = self::SELECT;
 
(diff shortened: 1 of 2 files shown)


morozov:
Could we please annotate the corresponding arguments as variadic so that it's clear what should be used instead?

morozov:
Otherwise, looks good.

BenMorel:
This makes phpstan choke:

```php
$groupBy = is_array($groupBy) ? $groupBy : func_get_args();
```

> Else branch is unreachable because ternary operator condition is always true.

Any idea how to fix this?

morozov:
> Any idea how to fix this?

Could you try adding `treatPhpDocTypesAsCertain: false` to the PHPStan configuration? See https://github.com/phpstan/phpstan/issues/2894#issuecomment-578238459. Alternatively, maybe annotate a second variadic argument instead of making the first one variadic:
```php
/**
 * @param mixed    $value
 * @param ...mixed $values
 */
function f($value/*, ...$values */)
```

BenMorel:
> Could you try adding `treatPhpDocTypesAsCertain: false`

Didn't work. Same errors.

> Alternatively, maybe annotate a second variadic argument instead of making the first one variadic

Doesn't work if the argument is commented out:

> PHPDoc tag @param references unknown parameter: $groupBys

And if it's not commented out, back to square one:

> Call to function is_array() with string will always evaluate to false.

morozov:
Then we’ll need to change the deprecation message:

> USING AN ARRAY ARGUMENT IS DEPRECATED. Pass each value as an individual argument.

BenMorel:
What works is keep documenting the `array`, and the variadic as `string`. Which may make sense, after all:

```php
    /**
     * Adds a grouping expression to the query.
     *
     * USING AN ARRAY ARGUMENT IS DEPRECATED. This feature will be removed in the next major version.
     *
     * (...)
     *
     * @param string|array $groupBy     The grouping expression. USING AN ARRAY IS DEPRECATED.
     * @param string       ...$groupBys
     *
     * @return $this This QueryBuilder instance.
     */
    public function addGroupBy($groupBy, ...$groupBys)
```

morozov:
It will introduce a BC break for extending classes (https://3v4l.org/CSa1P). Let's avoid code changes.

BenMorel:
> It will introduce a BC break

Damn. What about this: 0af792252f4d7c62895e50a33df6500a625770b0?

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -451,18 +451,21 @@ public function add($sqlPartName, $sqlPart, $append = false)
...
      * <code>
      *     $qb = $conn->createQueryBuilder()
      *         ->select('u.id', 'p.id')
      *         ->from('users', 'u')
      *         ->leftJoin('u', 'phonenumbers', 'p', 'u.id = p.user_id');
      * </code>
      *
-     * @param mixed $select The selection expressions.
+     * @param string|string[]|null $select The selection expression. USING AN ARRAY OR NULL IS DEPRECATED.
+     *                                     Pass each value as an individual argument. Pass at least one value.
```
morozov: > Pass at least one value.

This is not necessary. You still will be able to call it without arguments to declare a `SELECT` query. At least it’s the current state.

morozov:
> What about this: 0af7922?

Good with me.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Hmm, true. Actually I think we should force at least one value in `select()` as well; but we can keep this for another time.

BenMorel:
Removed "Pass at least one value." and squashed.

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: I deliberately kept it like for similarity with `insert()`, `update()` and `delete()`. It just defines the query type. Definitely worth revisiting later.

morozov (approved):

morozov:
Thanks, @BenMorel!

BenMorel:
@morozov Have you been able to merge to `master`?

morozov:
Just filed https://github.com/doctrine/dbal/pull/3857. There was a ton of conflicts between `3.0.x` and `master`.

BenMorel:
I knew this would be fun. Are you doing it as a PR to get reviews?

morozov:
It's not that _I_ want to get a review but I want the merge commit to be built in the same environment where it would be built after the merge. Also, we have all release branches protected.

morozov:
@BenMorel, #3857 is merged.

BenMorel:
Good job. Thank you!

PR #3858, opened 2020-01-29
Make CompositeExpression immutable

BenMorel:
Continuing the work in #3850.

### DIFF
--- a/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
+++ b/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
@@ -5,11 +5,15 @@ declare(strict_types=1);
 namespace Doctrine\DBAL\Query\Expression;
 
 use Countable;
+use function array_filter;
+use function array_values;
 use function count;
 use function implode;
 
 /**
  * Composite expression is responsible to build a group of similar expression.
+ *
+ * This class is immutable.
  */
 class CompositeExpression implements Countable
 {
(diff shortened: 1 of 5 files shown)


BenMorel:
@morozov BTW, should this be targeting 3.0 or 4.0? I can see all these breaking changes in the UPGRADE file targeting 3.0, but IIRC you said that there would not be large BC breaks in 3.0? Did I understand this right?

morozov:
These changes should be targeting master (4.0). The upgrade notes in `master` should be corrected have two different sections for `3.0` and `4.0`.

BenMorel:
Do you mean that some of the changes marked as 3.0 in the UPGRADE file should target 4.0? Should you update this file first?

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
```diff
@@ -43,51 +47,10 @@ class CompositeExpression implements Countable
      */
     public function __construct(string $type, array $parts = [])
```
morozov: In a future PR, I'd like the constructor to be made `private` in favor of two named constructors — `and()` and `or()` — which would accept `($part, ...$parts)` and therefore not allow construction of an empty composite expression.

BenMorel: Yes, I've mentioned that this is temporary in https://github.com/doctrine/dbal/pull/3850#discussion_r370997009. I'd like to complete immutability first, then move on to stop accepting empty parts. And you're right, factory methods will be a good idea!

morozov:
> Do you mean that some of the changes marked as 3.0 in the UPGRADE file should target 4.0? 

Yes.

> Should you update this file first?

https://github.com/doctrine/dbal/pull/3859

morozov (approved):

morozov:
Thanks, @BenMorel!

PR #3862, opened 2020-01-30
Specify the platform configuration in composer.json

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | [build #643647287](https://travis-ci.org/doctrine/dbal/builds/643647287)

Having the target platform specified in Composer configuration will help make sure we won't lock the dependencies to a version incompatible with the oldest supported PHP version (see the linked build) without the need to use the oldest supported PHP in the development environment.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,6 +48,9 @@
     },
     "bin": ["bin/doctrine-dbal"],
     "config": {
+        "platform": {
+            "php": "7.2.0"
+        },
         "sort-packages": true
     },
     "autoload": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "120172ae44052999ec9a50b1121414cc",
+    "content-hash": "5917f568d73e1f502700514d88d00b8c",
     "packages": [
         {
             "name": "doctrine/cache",
@@ -3129,8 +3129,10 @@
     "prefer-stable": false,
     "prefer-lowest": false,
     "platform": {
-        "php": "^7.2",
-        "ext-pdo": "*"
+        "php": "^7.2"
     },
-    "platform-dev": []
+    "platform-dev": [],
+    "platform-overrides": {
+        "php": "7.2.0"
+    }
 }


greg0ire (approved):
We should remove the composer.lock IMO, but until then, 👍

SenseException (approved):

morozov:
> We should remove the composer.lock IMO

@greg0ire please see https://github.com/doctrine/dbal/pull/3778. The practice shows that keeping it is the right choice.

How would removing it help to solve the problem of adding incompatible dependencies?

greg0ire:
If we have no composer.lock, and somebody requires a version that requires PHP 7.3, the 7.2 build will fail and prevent the PR from being merged.

morozov:
Instead of relying on CI building against the lowest supported version, this approach does the same in the development environment and does not require the lowest supported PHP version to resolve the dependencies.

greg0ire:
What if the version in the `composer.lock` is not compatible with the highest supported PHP version? You would have the exact same issue, reversed.

Also see @beberlei and I struggling to fix the build the other day because of a similar issue: https://doctrine.slack.com/archives/GERFT2W5T/p1579212730062400?thread_ts=1579212399.059900&cid=GERFT2W5T

morozov:
> What if the version in the `composer.lock` is not compatible with the highest supported PHP version? You would have the exact same issue, reversed.

Then the incompatibility will be detected on CI. For the most packages, the lowest version requirement is by design since they are using certain language features or APIs while the lowest version requirement is implicit and is not by design.

greg0ire:
> Then the incompatibility will be detected on CI. 

That's my point. There might be no incompatibility. There might exist a more recent version of the lib that is compatible with 7.4, yet one that is compatible with 7.2 and not 7.4 is in use.

> while the lowest version requirement

did you mean "while the highest"?

morozov:
> did you mean "while the highest"?

Yes, corrected.

morozov:
> That's my point. There might be no incompatibility.

I see. Makes sense. Let’s reconsider this once it happens.

greg0ire:
I agree that we will be quite unlucky if it happens. Relying on packages that aggressively bump the versions of php (such as `ocramius/package-versions`) increase the likelihood of this happening I think.

PR #3864, opened 2020-01-30
CompositeExpression and()/or() factory methods

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

To properly implement #3845 (forbidding empty `CompositeExpression`), we want to only allow creating expressions through `and()` / `or()` factory methods, that force at least one parameter.

Before doing so in `master`, I'm adding the factory methods to `2.11` and deprecating the constructor.

### DIFF
--- a/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
+++ b/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
@@ -3,6 +3,7 @@
 namespace Doctrine\DBAL\Query\Expression;
 
 use Countable;
+use function array_merge;
 use function count;
 use function implode;
 
@@ -36,6 +37,8 @@ class CompositeExpression implements Countable
     private $parts = [];
 
     /**
+     * @internal Use the and() / or() factory methods.
+     *
      * @param string          $type  Instance type of composite expression.
      * @param self[]|string[] $parts Composition of expressions to be joined on composite expression.
      */
@@ -44,6 +47,16 @@ class CompositeExpression implements Countable
         $this->type = $type;
 
         $this->addMultiple($parts);
+    }
+
+    public static function and($part, ...$parts) : self
+    {
+        return new self(self::TYPE_AND, array_merge([$part], $parts));
+    }
+
+    public static function or($part, ...$parts) : self
+    {
+        return new self(self::TYPE_OR, array_merge([$part], $parts));
     }
 
     /**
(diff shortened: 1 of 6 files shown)


lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
```diff
@@ -58,7 +57,7 @@ public function and($expression, ...$expressions) : CompositeExpression
      */
     public function or($expression, ...$expressions) : CompositeExpression
     {
-        return new CompositeExpression(CompositeExpression::TYPE_OR, array_merge([$expression], $expressions));
+        return CompositeExpression::or($expression, ...$expressions);
```
BenMorel: These methods do nothing more than forwarding to `CompositeExpression` factory methods. Should we deprecate them and remove them in `4.0`?

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
```diff
@@ -36,6 +37,8 @@ class CompositeExpression implements Countable
     private $parts = [];
 
     /**
+     * @deprecated Use the and() / or() factory methods.
```
morozov: We shouldn't deprecate a method that is not going to be removed. It's better to mark it `@internal` instead.

lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
morozov: > Should we deprecate them and remove them in `4.0`?

I don’t think so. `CompositeExpression` is a relatively low-level implementation detail so in most cases the consumers shouldn’t interact with the class directly.

morozov (commented):
> Before doing so in `master`, I'm adding the factory methods to `2.11` and deprecating the constructor.

Should the PR target `2.11.x.`?

BenMorel:
> Should the PR target 2.11.x.?

Oops. Fixed.

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
BenMorel: Done in ff626cab4ae5e0d134a0571bf816f3f0269942bb.

greg0ire:
Maybe the default branch should be changed to 2.11.x ?

greg0ire (approved):

morozov:
Thanks, @BenMorel !

PR #3868, opened 2020-01-31
Disallow empty CompositeExpression

### LINKED ISSUE #3845: QueryBuilder produces invalid SQL from an empty CompositeExpression

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #3845

#### Summary

This PR disallows an empty `CompositeExpression`, fixing #3845.

### DIFF
--- a/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
+++ b/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
@@ -5,9 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\DBAL\Query\Expression;
 
 use Countable;
-use function array_filter;
 use function array_merge;
-use function array_values;
 use function count;
 use function implode;
 
@@ -38,22 +36,20 @@ class CompositeExpression implements Countable
     /**
      * Each expression part of the composite expression.
      *
-     * @var self[]|string[]
+     * @var array<int, self|string>
      */
     private $parts = [];
 
     /**
      * @internal Use the and() / or() factory methods.
      *
-     * @param string          $type  Instance type of composite expression.
-     * @param self[]|string[] $parts Composition of expressions to be joined on composite expression.
+     * @param self|string $part
+     * @param self|string ...$parts
      */
-    public function __construct(string $type, array $parts = [])
+    public function __construct(string $type, $part, ...$parts)
     {
         $this->type  = $type;
-        $this->parts = array_values(array_filter($parts, static function ($part) {
-            return ! ($part instanceof self && count($part) === 0);
-        }));
+        $this->parts = array_merge([$part], $parts);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
```diff
@@ -38,22 +36,20 @@ class CompositeExpression implements Countable
...
     /**
      * @internal Use the and() / or() factory methods.
      *
-     * @param string          $type  Instance type of composite expression.
-     * @param self[]|string[] $parts Composition of expressions to be joined on composite expression.
+     * @param self|string $part
+     * @param self|string ...$parts
      */
-    public function __construct(string $type, array $parts = [])
+    public function __construct(string $type, $part, ...$parts)
```
BenMorel: Do we want to keep the constructor public?

Making it private would require a change to the following logic:

https://github.com/doctrine/dbal/blob/39dd1e53b7e053b76decff77339f73579136e175/lib/Doctrine/DBAL/Query/QueryBuilder.php#L1039-L1052

morozov: It feels like this logic belongs to the expression class itself. It’s already aware of its type and parts and should be able to produce a new instance of self. This way, we wouldn’t even need to expose `getType()` anymore.

But I’m not sure how to put it exactly. Let’s leave it public for now.

morozov (approved):

lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php
BenMorel: What about moving these to static methods in `CompositeExpression`?

Here's how it would look like: https://github.com/BenMorel/dbal/commit/a123732527414b0e5b5b0eecdb6e440e5c2d2a3f (not included in this PR for now).

This way, we can make the constructor private, and remove the `getType()` method. Note that we still have to keep the `TYPE_*` constants public, unless we split `appendToPredicate()` into two distinct methods.

morozov: > What about moving these to static methods in `CompositeExpression`?

Yeah, this is exactly what I had in mind. Let’s take it out of scope since it still needs additional discussion.

BenMorel: All good for me. Let's merge this one, and I'll see what I can do in another PR!

morozov:
Merged. Thanks, @BenMorel!

PR #3869, opened 2020-01-31
Add parameter type hint

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

A factory method accepts `mixed` when only ever `int` will be passed (and fortunately, as it's given as sprintf `%d` parameter).

### DIFF
--- a/lib/Doctrine/DBAL/Driver/Exception/UnknownFetchMode.php
+++ b/lib/Doctrine/DBAL/Driver/Exception/UnknownFetchMode.php
@@ -9,10 +9,7 @@ use function sprintf;
 
 final class UnknownFetchMode extends DBALException
 {
-    /**
-     * @param mixed $fetchMode
-     */
-    public static function new($fetchMode) : self
+    public static function new(int $fetchMode) : self
     {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Driver/Exception/UnknownFetchMode.php
```diff
@@ -9,10 +9,7 @@
 
 final class UnknownFetchMode extends DBALException
 {
-    /**
-     * @param mixed $fetchMode
-     */
-    public static function new($fetchMode) : self
+    public static function new(int $fetchMode) : self
```
BenMorel: The only place where it's being used is:

https://github.com/doctrine/dbal/blob/97e0c2bb7386288ef6954aeb10f990b9231cb517/lib/Doctrine/DBAL/Driver/PDOStatement.php#L208-L215

morozov (approved):

morozov:
Thanks, @BenMorel!

Issue #2979, opened 2018-01-14
getAlterTableName doesn't correctly quote the column name on MySQL


FSMaxB:
I'm not sure how to reproduce this with DBAL directly, but I stumbled upon this problem via the Laravel framework.

Original issue reported with Laravel: https://github.com/laravel/framework/issues/22402

I didn't quite find the code in Laravel that causes this, but I am quite confident that it is caused by a bug in `getAlterTableSQL`.

DBAL seems to create the following queries:
> The query is:
>```SQL
>ALTER TABLE bug CHANGE weird-name good_name VARCHAR(255) NOT NULL;
>```
>
>But it should be:
>```SQL
>ALTER TABLE bug CHANGE `weird-name` good_name VARCHAR(255) NOT NULL;
>```

I guess that the following line needs backticks `\`` around the quoted column name:
https://github.com/doctrine/dbal/blob/9f843bca49a223195921e589173ebaad8980bf0f/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L619

Ocramius:
Is this reproducible with DBAL 2.6?

FSMaxB:
Yes. DBAL 2.6.3, I just checked.

FSMaxB:
Ping @Ocramius

Ocramius:
@FSMaxB what about this?

FSMaxB:
Has anyone looked into this yet?

Ocramius:
@FSMaxB nope - don't think anybody will actively get to it for now.

derekrprice:
Still happening in DBAL v2.10.0.  The solution that @FSMaxB suggested is close but might be slightly naive.  There is already a _quote attribute on the $column objects being passed in, which were generated by DBAL's column diff.  When set, that attribute causes $diff->getName($this)->getQuotedName($this) & $column->getQuotedName($this) to return a correctly quoted identifier.  So, the trick might be getting diff to set those attributes correctly.  Alternately, getQuotedName() appears to automatically quote when it spots an identifier, regardless of the value of _quote.  If it was updated to also quote column names with spaces in them, that would work for me.

Ran into this with Laravel 6.9.0, attempting to alter columns with spaces in the names.  Strangely, the same code works just fine to create the columns.  It's just altering them that is an issue.

derekrprice:
[...]

PR #3872, opened 2020-02-06
Automatically Quote Identifiers with Spaces in Their Names

derekrprice:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2979 

#### Summary
Automatically quote identifiers with spaces in their names.

I know that automatically quoting identifiers is frowned upon, for the reasons described in #146, but `AbstractAsset::getquotedName` already makes an exceptions for keywords, since not quoting those would break the generated SQL anyhow.  In the same way, spaces in column names and other identifiers break the generated SQL, so I think this is justified.


SenseException:
Thank you. Can you please add tests for your additional code?

derekrprice:
I've added tests.  I'd like to see this on the 2.11.x branch too.  I expect that I should create a second pull request for that?

greg0ire:
IMO you should click the EDIT button and change the base to 2.11.x, and then run `git rebase --onto origin/2.11.x origin/master master && git push --force`. It will be a bit confusing because your branch is named `master`, but that should not be an issue.

derekrprice:
I created pull request #3876 instead since I already had this change on a 2.11.x derivative branch, I couldn't edit the source branch, and there are merge conflicts from 2.11.x -> master that I had already resolved anyhow.

derekrprice:
Should I close this PR in favor of the other?

greg0ire:
I think you should indeed. We are switching to a merge up strategy, so your changes should land on master in the end, just no through this PR.

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
```diff
@@ -629,17 +629,30 @@ public function testKeywordList() : void
...
-    public function testQuotedColumnInPrimaryKeyPropagation() : void
+    public function testAutoQuotedColumnInPrimaryKeyPropagation(string $identifier) : void
     {
         $table = new Table('`quoted`');
-        $table->addColumn('create', 'string', ['length' => 255]);
-        $table->setPrimaryKey(['create']);
+        $table->addColumn($identifier, 'string', ['length' => 255]);
+        $table->setPrimaryKey([$identifier]);
 
+        $expected = preg_replace('/create/', $identifier, $this->getQuotedColumnInPrimaryKeySQL());
```
greg0ire: Not a huge fan of this… I would add a new argument defaulting to `create` to `getQuotedColumnInPrimaryKeySQL`.

PR #3876, opened 2020-02-11
Automatically Quote Identifiers with Spaces in Their Names

derekrprice:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2979 

#### Summary
Automatically quote identifiers with spaces in their names.

I know that automatically quoting identifiers is frowned upon, for the reasons described in #146, but `AbstractAsset::getquotedName` already makes an exceptions for keywords, since not quoting those would break the generated SQL anyhow.  In the same way, spaces in column names and other identifiers break the generated SQL, so I think this is justified.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/AbstractAsset.php
+++ b/lib/Doctrine/DBAL/Schema/AbstractAsset.php
@@ -8,6 +8,7 @@ use function crc32;
 use function dechex;
 use function explode;
 use function implode;
+use function preg_match;
 use function str_replace;
 use function strpos;
 use function strtolower;
@@ -180,7 +181,7 @@ abstract class AbstractAsset
         $keywords = $platform->getReservedKeywordsList();
         $parts    = explode('.', $this->getName());
         foreach ($parts as $k => $v) {
-            $parts[$k] = $this->_quoted || $keywords->isKeyword($v) ? $platform->quoteIdentifier($v) : $v;
+            $parts[$k] = $this->_quoted || $keywords->isKeyword($v) || preg_match('/\W/', $v) ? $platform->quoteIdentifier($v) : $v;
         }
 
         return implode('.', $parts);
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
@@ -248,9 +248,9 @@ abstract class AbstractMySQLPlatformTestCase extends AbstractPlatformTestCase
     /**
      * @return string[]
      */
-    protected function getQuotedColumnInPrimaryKeySQL() : array
+    protected function getQuotedColumnInPrimaryKeySQL(string $columnName) : array
     {
-        return ['CREATE TABLE `quoted` (`create` VARCHAR(255) NOT NULL, PRIMARY KEY(`create`)) DEFAULT CHARACTER SET utf8 COLLATE `utf8_unicode_ci` ENGINE = InnoDB'];
+        return ['CREATE TABLE `quoted` (`' . $columnName . '` VARCHAR(255) NOT NULL, PRIMARY KEY(`' . $columnName . '`)) DEFAULT CHARACTER SET utf8 COLLATE `utf8_unicode_ci` ENGINE = InnoDB'];
     }
 
     /**
@@ -272,13 +272,13 @@ abstract class AbstractMySQLPlatformTestCase extends AbstractPlatformTestCase
     /**
      * @return string[]
      */
-    protected function getQuotedColumnInForeignKeySQL() : array
+    protected function getQuotedColumnInForeignKeySQL(string $columnName) : array
     {
         return [
-            'CREATE TABLE `quoted` (`create` VARCHAR(255) NOT NULL, foo VARCHAR(255) NOT NULL, `bar` VARCHAR(255) NOT NULL) DEFAULT CHARACTER SET utf8 COLLATE `utf8_unicode_ci` ENGINE = InnoDB',
-            'ALTER TABLE `quoted` ADD CONSTRAINT FK_WITH_RESERVED_KEYWORD FOREIGN KEY (`create`, foo, `bar`) REFERENCES `foreign` (`create`, bar, `foo-bar`)',
-            'ALTER TABLE `quoted` ADD CONSTRAINT FK_WITH_NON_RESERVED_KEYWORD FOREIGN KEY (`create`, foo, `bar`) REFERENCES foo (`create`, bar, `foo-bar`)',
-            'ALTER TABLE `quoted` ADD CONSTRAINT FK_WITH_INTENDED_QUOTATION FOREIGN KEY (`create`, foo, `bar`) REFERENCES `foo-bar` (`create`, bar, `foo-bar`)',
+            'CREATE TABLE `quoted` (`' . $columnName . '` VARCHAR(255) NOT NULL, foo VARCHAR(255) NOT NULL, `bar` VARCHAR(255) NOT NULL) DEFAULT CHARACTER SET utf8 COLLATE `utf8_unicode_ci` ENGINE = InnoDB',
+            'ALTER TABLE `quoted` ADD CONSTRAINT FK_WITH_RESERVED_KEYWORD FOREIGN KEY (`' . $columnName . '`, foo, `bar`) REFERENCES `foreign` (`' . $columnName . '`, bar, `foo-bar`)',
+            'ALTER TABLE `quoted` ADD CONSTRAINT FK_WITH_NON_RESERVED_KEYWORD FOREIGN KEY (`' . $columnName . '`, foo, `bar`) REFERENCES foo (`' . $columnName . '`, bar, `foo-bar`)',
+            'ALTER TABLE `quoted` ADD CONSTRAINT FK_WITH_INTENDED_QUOTATION FOREIGN KEY (`' . $columnName . '`, foo, `bar`) REFERENCES `foo-bar` (`' . $columnName . '`, bar, `foo-bar`)',
         ];
     }
 
@@ -802,25 +802,25 @@ abstract class AbstractMySQLPlatformTestCase extends AbstractPlatformTestCase
     /**
      * {@inheritdoc}
      */
-    protected function getQuotesReservedKeywordInUniqueConstraintDeclarationSQL() : string
+    protected function getQuotesReservedKeywordInUniqueConstraintDeclarationSQL(string $constraintName) : string
     {
-        return 'CONSTRAINT `select` UNIQUE (foo)';
+        return 'CONSTRAINT `' . $constraintName . '` UNIQUE (foo)';
     }
 
     /**
      * {@inheritdoc}
      */
-    protected function getQuotesReservedKeywordInIndexDeclarationSQL() : string
+    protected function getQuotesReservedKeywordInIndexDeclarationSQL(string $indexName) : string
     {
-        return 'INDEX `select` (foo)';
+        return 'INDEX `' . $indexName . '` (foo)';
     }
 
     /**
      * {@inheritdoc}
      */
-    protected function getQuotesReservedKeywordInTruncateTableSQL() : string
+    protected function getQuotesReservedKeywordInTruncateTableSQL(string $tableName) : string
     {
-        return 'TRUNCATE `select`';
+        return 'TRUNCATE `' . $tableName . '`';
     }
 
     /**
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
@@ -633,22 +633,38 @@ abstract class AbstractPlatformTestCase extends DbalTestCase
     }
 
     /**
-     * @group DBAL-374
+     * @return string[][]
      */
-    public function testQuotedColumnInPrimaryKeyPropagation() : void
+    public function getAutoQuotedIdentifiers() : array
+    {
+        return [
+            ['create'],
+            ['select'],
+            ['a name with spaces'],
+            ['a name with/slash'],
+            ['name-with-dash'],
+        ];
+    }
+
+    /**
+     * @group DBAL-374
+     * @group DBAL-2979
+     * @dataProvider getAutoQuotedIdentifiers
+     */
+    public function testAutoQuotedColumnInPrimaryKeyPropagation(string $columnName) : void
     {
         $table = new Table('`quoted`');
-        $table->addColumn('create', 'string');
-        $table->setPrimaryKey(['create']);
+        $table->addColumn($columnName, 'string');
+        $table->setPrimaryKey([$columnName]);
 
         $sql = $this->platform->getCreateTableSQL($table);
-        self::assertEquals($this->getQuotedColumnInPrimaryKeySQL(), $sql);
+        self::assertEquals($this->getQuotedColumnInPrimaryKeySQL($columnName), $sql);
     }
 
     /**
      * @return string[]
      */
-    abstract protected function getQuotedColumnInPrimaryKeySQL() : array;
+    abstract protected function getQuotedColumnInPrimaryKeySQL(string $columnName) : array;
 
     /**
      * @return string[]
@@ -663,7 +679,7 @@ abstract class AbstractPlatformTestCase extends DbalTestCase
     /**
      * @return string[]
      */
-    abstract protected function getQuotedColumnInForeignKeySQL() : array;
+    abstract protected function getQuotedColumnInForeignKeySQL(string $columnName) : array;
 
     /**
      * @group DBAL-374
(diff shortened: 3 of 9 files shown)


greg0ire:
Don't act on my comment, but it's a good practice to use a branch name that does not look like the target branch, in case someone asks you to retarget your pr.

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
```diff
@@ -632,17 +632,30 @@ public function testKeywordList() : void
...
-    public function testQuotedColumnInPrimaryKeyPropagation() : void
+    public function testAutoQuotedColumnInPrimaryKeyPropagation(string $identifier) : void
     {
         $table = new Table('`quoted`');
-        $table->addColumn('create', 'string');
-        $table->setPrimaryKey(['create']);
+        $table->addColumn($identifier, 'string');
+        $table->setPrimaryKey([$identifier]);
 
+        $expected = preg_replace('/create/', $identifier, $this->getQuotedColumnInPrimaryKeySQL());
```
greg0ire: Not a huge fan of this… I would add a new argument defaulting to `create` to `getQuotedColumnInPrimaryKeySQL`.

derekrprice: It's an abstract function with implementations across 4 or 7 files.  You intend for me to update them all?  There are also four or five functions that I will have to update this way.  I agree that my usage of preg_replace() could be a little fragile, so I don't mind.  I just wasn't sure that I should go so far with a first pass.

greg0ire: > You intend for me to update them all? 

No I don't. That's why I said "defaulting to `create`"

greg0ire: > You intend for me to update them all?

It will indeed be a lot of work, but I think this it will be for the best.

> I just wasn't sure that I should go so far with a first pass.

Yeah good call. Maybe let's wait for another contributor to chime in on this then? If they think we should merge this feature, and they are ok with my improvements for the tests, then you can proceed. @SenseException , maybe you could be of some advice on this?

derekrprice: Incidentally, I am working on this now.  It's not that big a job.  I had considered it initially, but I was worried that it would make my original patch too invasive.

derekrprice: Ok.  I started on the changes but I will put them on hold until I have your go-ahead, then.  Thanks.

derekrprice:
@greg0ire 
> Don't act on my comment, but it's a good practice to use a branch name that does not look like the target branch, in case someone asks you to retarget your pr.

So, I chose this branch name because all-numbers was the only format that I could find that composer was happy to treat as a dev branch and still calculate dependencies from other packages correctly so that I could use the fork with my patch in my project until such time that the patch is accepted.  Is there another naming convention that composer would be happy with that is more appropriate for re-targeting?

greg0ire:
Yes, you could use inline aliases (`dev-my-branch as 1.2.3`

derekrprice:
I'd missed that feature.  Thanks!

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
SenseException: Yes, please don't use the preg_replace approach on this. I would appreciate when you do the improvement that @greg0ire was suggesting. Thank you for your effort in advance. :tulip:

derekrprice: The patch is a lot bigger now, but there you go.  :)

Not sure of the etiquette of resolving conversations.  Should I do it to signify that I think I did what you asked, or should I wait for the original commenter (@greg0ire in this case) to resolve it to mark their satisfaction with the change?

greg0ire: Thanks for asking, you should let the original commenter do the resolving.

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
```diff
@@ -690,88 +702,101 @@ public function testQuotedNameInIndexSQL() : void
 
     /**
      * @group DBAL-374
+     * @group DBAL-2979
+     * @dataProvider getAutoQuotedIdentifiers
+     * @param string $columnName
```
greg0ire: The type declaration is enough
```suggestion
```

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
```diff
@@ -690,88 +702,101 @@ public function testQuotedNameInIndexSQL() : void
...
         $sql = $this->platform->getCreateTableSQL($table, AbstractPlatform::CREATE_FOREIGNKEYS);
-        self::assertEquals($this->getQuotedColumnInForeignKeySQL(), $sql);
+        self::assertEquals($this->getQuotedColumnInForeignKeySQL($columnName), $sql);
     }
 
     /**
      * @group DBAL-1051
+     * @group DBAL-2979
+     * @dataProvider getAutoQuotedIdentifiers
+     * @param string $constraintName
```
greg0ire:
```suggestion
```

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
```diff
@@ -690,88 +702,101 @@ public function testQuotedNameInIndexSQL() : void
...
     }
 
-    abstract protected function getQuotesReservedKeywordInUniqueConstraintDeclarationSQL() : string;
+    abstract protected function getQuotesReservedKeywordInUniqueConstraintDeclarationSQL(string $constraintName = 'select') : string;
 
     /**
      * @group DBAL-2270
+     * @group DBAL-2979
+     * @dataProvider getAutoQuotedIdentifiers
+     * @param string $tableName
```
greg0ire:
```suggestion
```

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
```diff
@@ -690,88 +702,101 @@ public function testQuotedNameInIndexSQL() : void
...
     }
 
-    abstract protected function getQuotesReservedKeywordInTruncateTableSQL() : string;
+    abstract protected function getQuotesReservedKeywordInTruncateTableSQL(string $tableName = 'select') : string;
 
     /**
      * @group DBAL-1051
+     * @group DBAL-2979
+     * @dataProvider getAutoQuotedIdentifiers
+     * @param string $indexName
```
greg0ire:
```suggestion
```

greg0ire:
Please fix the Travis build: https://travis-ci.org/doctrine/dbal/jobs/654431635

derekrprice:
@greg0ire, done, except for one line.  Any idea why Travis would object to a group annotation?  It's the exact same as the one on line 651, but Travis is only squawking about line 708.

Also, the continuous PHP failure looks like it's a problem with the pipeline and not my code, right?

Thanks!

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
```diff
@@ -690,88 +706,97 @@ public function testQuotedNameInIndexSQL() : void
 
     /**
      * @group DBAL-374
+     * @group DBAL-2979
+     * @dataProvider getAutoQuotedIdentifiers
+     * @throws DBALException
```
SenseException: Unexpected `@throws`

```suggestion
```

SenseException (changes requested):
@derekrprice Have you tried if phpcbf can fix it?

derekrprice:
@SenseException, that solved it, thanks.  The only Travis issues now appear to be a problem installing MS SQL Server in the pipeline.
```
The following packages have unmet dependencies:
 msodbcsql17 : Depends: libc6 (>= 2.21) but 2.19-0ubuntu6.13 is to be installed
               Depends: libstdc++6 (>= 4.9) but 4.8.4-2ubuntu1~14.04.3 is to be installed
E: Unable to correct problems, you have held broken packages.
The command "bash ./tests/travis/install-sqlsrv-dependencies.sh" failed and exited with 100 during .
```

derekrprice:
Weird, this PR says that there are still changes requested by @SenseException, but I can't find them.  I committed all the requested changes that I saw.

greg0ire:
> Weird, this PR says that there are still changes requested by @SenseException, but I can't find them. I committed all the requested changes that I saw.

Don't worry, that's normal, it should disappear when he reviews again and approves.

greg0ire:
Did you really mean to target 2.11.x ? If yes, why?

derekrprice:
My Laravel installation is tied to ^2.5, which composer currently interprets as 2.10.1.  So, I did make a mistake.  I was thinking that I should grab the most recent 2.x branch, but unless the 2.11 release is coming pretty soon, I should probably target 2.10.x so that we can use the fix here right away.

greg0ire:
> Would re-targetting to 2.10.x be acceptable?

Yes, please do! I'm going to change the base branch for you , you can `git rebase --onto 96cda30 f9a0563 2.11.2.x && git push --force-with-lease`, and click the edit button to change your base branch

derekrprice:
Done.

greg0ire:
I fixed the build since then, let's trigger the Travis again.

greg0ire:

How to do that?

1. `git rebase -i origin/2.10.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

Also, please reword the resulting commit like so:

```
Quote identifiers with non-word characters
    
Not doing so can break the generated SQL.
Fixes #2979
```


I wanted to do it for you, but for some reason I can't push to your fork.

SenseException (dismissed):
After the mentioned squash, this is :+1:

derekrprice:
Done.

greg0ire (dismissed):

greg0ire:
Retargeting to 2.10 which is the correct branch (see https://github.com/doctrine/dbal/pull/3884#issuecomment-592975527), but does not have my build fix yet.

greg0ire (dismissed):

lib/Doctrine/DBAL/Schema/AbstractAsset.php
```diff
@@ -180,7 +181,7 @@ public function getQuotedName(AbstractPlatform $platform)
         $keywords = $platform->getReservedKeywordsList();
         $parts    = explode('.', $this->getName());
         foreach ($parts as $k => $v) {
-            $parts[$k] = $this->_quoted || $keywords->isKeyword($v) ? $platform->quoteIdentifier($v) : $v;
+            $parts[$k] = $this->_quoted || $keywords->isKeyword($v) || preg_match('/\W/', $v) ? $platform->quoteIdentifier($v) : $v;
```
morozov: This is a slippery slope. An object name needs to be quoted if otherwise it would be interpreted as an SQL expression which is far beyond `preg_match('/\W/', $v)`. Additionally, on some of the supported platforms (e.g. Oracle) quoted object names have different semantics than unquoted ones.

In addition to that, given https://github.com/doctrine/dbal/issues/3152 and https://github.com/doctrine/orm/issues/7815, DBAL is still a leaky abstraction when it comes to handling arbitrary object names.

Before we proceed with any changes like this, I'd like to explicitly define what we do and what we do not expect from DBAL regarding quoting object names.

After that, we’ll need _functional cross-platform_ tests that prove that this promise is fulfilled. Testing SQL without executing it is not enough.

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
```diff
@@ -248,9 +248,9 @@ public function testChangeIndexWithForeignKeys() : void
     /**
      * @return string[]
      */
-    protected function getQuotedColumnInPrimaryKeySQL() : array
+    protected function getQuotedColumnInPrimaryKeySQL(string $columnName = 'create') : array
```
morozov: Why does the parameter need to be optional?

greg0ire:
> Retargeting to 2.10 which is the correct branch (see #3884 (comment)), but does not have my build fix yet.

Actually, re-retargeting to 2.10.x . I am going to clean things up and fix the duplicate branch situation. Sorry for this mess.

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
derekrprice: @greg0ire suggested it.  @greg0ire, I don't think this is needed - can I remove all occurrences?

lib/Doctrine/DBAL/Schema/AbstractAsset.php
derekrprice: @morozovm, my argument here is that object names with non-word characters are going to blow up anyhow, so DBAL might as well quote them, regardless of any potential semantic changes on a given platform.  Quoting an object name and ending up with a case-sensitive field name cannot possibly be worse than the syntax error that occurs running the SQL if you don't quote it.  If it is OK to quote for keywords, then it should be absolutely fine to also do it for non-word characters.

My use case is using the Laravel Schema Builder to dynamically import multiple Access DBs into MySQL.  Laravel does something funny with the quoting, so quoting the field names myself at that level was not enough to work around this.  `preg_match('/\W/', $v)` was more than enough to deal with that use case, and I don't see how functional tests could make this particular patch any better.  Every single field name that this quotes was enough to generate invalid SQL before the patch.  Can you provide me with an example of a SQL expression that _does not_ contain a non-word character?  How about an example of a valid column name that _does_?

Thanks.

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
greg0ire: I suggested to add an optional argument iff you need to change that `create` value? Which you did. If you can make it non-optional and update all the pre-existing method call to pass `create`, that's great. If you won't need the value to change, that's great too.

derekrprice: There was no need for it to be an optional argument because I had already updated all callers (including passing the "default" sometimes).  Fixed.

derekrprice: @greg0ire, should I squash the commits again immediately?

greg0ire: Yes

derekrprice: Done.

lib/Doctrine/DBAL/Schema/AbstractAsset.php
morozov: I agree that there won't be a BC break if we start quoting some names that otherwise would have been interpreted as expressions and fail. My concern is that there are no clear expectations about how DBAL should handle such names and where. And it's not tested in an integration way (i.e. unit tests may claim that it works but it practice it doesn't because such SQL never gets run against a DB).

`preg_match('/\W/', $v)` will work for the names containing whitespaces but it's just a particular case. E.g. the following is still not handled:

```sql
CREATE TABLE `1` (
    `2` INTEGER
)
```

There must be added functional tests handling "special" names around all schema introspection/deployment use cases before we consider it supported.

derekrprice: This doesn't seem like a good reason to hold up this PR to me.  This PR adds additional functionality without breaking anything, fixing SQL that was previously broken.

morozov: Lack of proper tests is a valid reason to not accept code changes. Please stop arguing and collaborate.

derekrprice: I've collaborated for weeks on this PR.  In fact, I think I've been very accommodating, rewriting and expanding this patch for @greg0ire and @SenseException several times.  I agree that integration tests would be more thorough, but you agree that this minor change should not have broken anything and it is being tested as thoroughly as that area of the code was tested previous to my relatively minor change.  Please merge this and do not hold this useful contribution hostage to my making additional useful contributions immediately.

driesvints:
I'm sorry if this might not be the correct place to ask this but would this PR also benefit from including dashes? See https://github.com/laravel/framework/issues/22402

greg0ire:
From https://github.com/doctrine/dbal/pull/3876#discussion_r387263836 : 

> preg_match('/\W/', $v) will work for the names containing whitespaces but it's just a particular case.

So yes, definitely, among other things, and with hard to know consequences that we would feel better about if there was SQL run against a real database at some point.

@derekrprice congrats on taking it this far, you did a big amount of work but sadly, it's not enough, meaning that we can't merge as is. I know it sounds harsh, but this package we maintain has, as of now, 113 million downloads. That's more than Symfony or Laravel, to give an idea of the impact bugs in this package can have. It's used with a variety of databases that requires us to run it with a huge number of CI jobs (see Travis etc.). Although it could have been said in a nicer way, the kind of tests @morozov required are exactly that: required.

lib/Doctrine/DBAL/Schema/AbstractAsset.php
mvorisek: this should be more like `! preg_match('~^\w+$~', ...)`

PR #3881, opened 2020-02-27
Accept Connection interface instead of concrete instance

danielgelling:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

Accept the `Doctrine\DBAL\Driver\Connection` interface instead of the concrete instance in order to improve scalability.


danielgelling:
Closing because interface does not specify all methods used by the QueryBuilder. Considering looking into extending the `Driver\Connection` interface that can be used by by the QueryBuilder.  Your thoughts on putting time into this?

PR #3477, opened 2019-03-03
Drop db support for PostgreSQL 9.3 and MariaDB 10.0

SenseException:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | -
| Related | #3055

#### Summary

PostgreSQL 9.3 is an [unsupported version](https://www.postgresql.org/support/versioning/) since last year. MariaDB 10.0 support [ends in March 2019](https://mariadb.org/about/maintenance-policy/) and considering that the next DBAL major release is still far in the future compared to MariaDB 10.0, I already removed it in this PR.

PR #3883, opened 2020-02-27
Fix broken build

### LINKED ISSUE #3882: Build failure on sql server jobs
There are issues in the pipeline:

```
The following packages have unmet dependencies:
 msodbcsql17 : Depends: libc6 (>= 2.21) but 2.19-0ubuntu6.13 is to be installed
               Depends: libstdc++6 (>= 4.9) but 4.8.4-2ubuntu1~14.04.3 is to be installed
E: Unable to correct problems, you have held broken packages.
The command "bash ./tests/travis/install-sqlsrv-dependencies.sh" failed and exited with 100 during .
```

_Originally posted by @derekrprice in https://github.com/doctrine/dbal/pull/3876#issuecomment-592080819_

I noticed that [apt also complains about some unavailable public keys and says it's going to use old files before that, but for seemingly unrelated packages](https://travis-ci.org/doctrine/dbal/jobs/655882778#L627-L632):

```
W: An error occurred during the signature verification. The repository is not updated and the previous index files will be used. GPG error: http://dl.google.com/linux/chrome/deb stable Release: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 78BD65473CB3BD13
W: An error occurred during the signature verification. The repository is not updated and the previous index files will be used. GPG error: https://packagecloud.io/github/git-lfs/ubuntu trusty InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 6B05F25D762E3157
W: http://ppa.launchpad.net/couchdb/stable/ubuntu/dists/trusty/Release.gpg: Signature by key 15866BAFD9BCC4F3C1E0DFC7D69548E1C17EAB57 uses weak digest algorithm (SHA1)
W: Failed to fetch https://packagecloud.io/github/git-lfs/ubuntu/dists/trusty/InRelease  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 6B05F25D762E3157
W: Failed to fetch http://dl.google.com/linux/chrome/deb/dists/stable/Release.gpg  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 78BD65473CB3BD13
W: Some index files failed to download. They have been ignored, or old ones used instead.
```

greg0ire:
Fixes #3882

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: php
-sudo: false
-dist: trusty
+dist: xenial
 
 cache:
     directories:
@@ -112,6 +111,8 @@ jobs:
     - stage: Test
       php: 7.3
       env: DB=mysql COVERAGE=yes
+      services:
+        - mysql
     - stage: Test
       php: 7.3
       env: DB=mysql.docker MYSQL_VERSION=5.7 COVERAGE=yes
@@ -129,6 +130,8 @@ jobs:
     - stage: Test
       php: 7.3
       env: DB=mysqli COVERAGE=yes
+      services:
+        - mysql
     - stage: Test
       php: 7.3
       env: DB=mysqli.docker MYSQL_VERSION=5.7 COVERAGE=yes
@@ -150,9 +153,11 @@ jobs:
         mariadb: 10.0
     - stage: Test
       php: 7.3
-      env: DB=mariadb MARIADB_VERSION=10.1 COVERAGE=yes
-      addons:
-        mariadb: 10.1
+      env: DB=mariadb.docker MARIADB_VERSION=10.1 COVERAGE=yes
+      services:
+        - docker
+      before_script:
+        - bash ./tests/travis/install-mariadb.sh
     - stage: Test
       php: 7.3
       env: DB=mariadb MARIADB_VERSION=10.2 COVERAGE=yes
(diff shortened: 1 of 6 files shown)


greg0ire:
:thinking: no package on hold before we run `ACCEPT_EULA=Y sudo apt-get install -qy msodbcsql17 unixodbc unixodbc-dev libssl1.0.0`

greg0ire:
I think it has to do with the fact we are still using trusty, which is EOL since 2019

greg0ire:
Switching to bionic fixed the issue, but new issues arise. I'm going to try cherry-picking stuff from master.

greg0ire:
I'm targeting 2.11.x, because that's where the bug was observed, but I'm unsure where this should go, because the branch metadata seems to be outdated:
https://github.com/doctrine/dbal/blob/5fec6ead0354771562be82f868b3699ec8b4923a/.doctrine-project.json#L8-L23

SenseException:
AFAIK 2.10.x is still supported.

greg0ire:
The build is almost green, but I'm going to have to drop MariaDB 10.0:
- cannot be installed
- no longer supported anyway

greg0ire:
Same goes for Postgresql 9.2 and 9.3.

SenseException:
> The build is almost green, but I'm going to have to drop MariaDB 10.0:

>  Same goes for Postgresql 9.2 and 9.3.

Maybe that shouldn't be much of a problem for these unsupported versions. A drop already happened in develop with #3477.

greg0ire:
:thinking: shouldn't that PR have targeted the stable branch?

SenseException:
Dropping support would be a BC break. Develop was the next major version.

SenseException (approved):
One question though: Can mariadb.mysqli.travis.xml be dropped in favor of using mariadb.mysqli.docker.travis.xml?

greg0ire:
Not sure what you mean… that would mean changing the user and port for some builds, right?

morozov (approved):

SenseException:
@greg0ire Yes, it would use Docker for the other MariaDB too.

greg0ire:
I'm going to just merge since this is critical (the build is broken). I marking this as unread to try and do what you suggest.

Issue #3478, opened 2019-03-06
Improve readiness probe stability for containerized databases on CI


morozov:
Please see the details in #3473. Filed based on https://github.com/doctrine/dbal/pull/3473#issuecomment-469663463.

morozov:
Seems fixed by #3884.

PR #3884, opened 2020-02-28
Use Docker consistently

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | not an issue, but https://github.com/doctrine/dbal/pull/3883#pullrequestreview-366003800

#### Summary

When a Docker version of the build is available, use it. Also, remove jobs that test the same version as a docker job.


greg0ire:
No idea if this is actually going to work, let's see…

greg0ire:
Ok it works now. Please review.

SenseException (dismissed):
A few minutes ago, I wanted to write something about `7.4snapshot`, but your recent commit is handling that. :grin:

morozov:
I’m all for this with the exception that it will increase the likelihood of https://github.com/doctrine/dbal/issues/3478 which is already a PITA. The tests against docker-based MySQL (and perhaps MariaDB) images fail from time to time because, despite the readiness probe passing, all subsequent queries fail with "2006 Server gone away".

Before we go all Docker, I want this issue to be addressed or at least mitigated.

greg0ire:
Ok, I'm going to try this: https://stackoverflow.com/questions/25503412/how-do-i-know-when-my-docker-mysql-container-is-up-and-mysql-is-ready-for-taking#comment66272388_29793382

And then … trigger 10 builds?

greg0ire:
Not sure if `mysqladmin` is available in the image, let's see…

greg0ire:
I think I have something more to do to force waiting for the container to be healthy.

greg0ire:
Done, and it seems to work. @morozov what now? Let's restart the build many times?

morozov:
> Let's restart the build many times?

I thought about the same. Maybe 5-10 times to be on the safe side. But before that, since we're generalizing the dependencies, why don't we generalize the `install-*.sh ` scripts for MySQL and MariaDB?

Instead of copy/pasting the code, we could introduce a required CLI argument that would take a Docker image name to be spun up like: `docker-run-mysql-or-mariadb.sh mysql:8.0`.

greg0ire:
I'm on it. I'm going to use an environment variable to specify the image, that way you can still see what the version is in the overview of the jobs.

greg0ire:
Not 100% sure but I think I just made these lines useless:

https://github.com/doctrine/dbal/blob/96d10bd5382250e9fb93729f5b232d3b9e5a46b6/.travis.yml#L16-L17

I think they are not in use for jobs that end with `.docker`, correct?

greg0ire:
Ah that's correct, I thought global and local `before_script` were merged, they aren't, it's an override.

greg0ire:
Green build again :)

Do you think I should get rid of the `.docker` part in `DB` env variables and xml filenames?

greg0ire:
There is an infinite loop now, many containers are unhealthy…

greg0ire:
I added a check that will fail the build on unhealthy container. I noticed docker is run with `sudo`, why is that?

greg0ire:
Ok so it's unhealthy because I badly implemented the addition of the extra command option :)

greg0ire:
> I noticed docker is run with sudo, why is that?

I removed it, and it works… also, I noticed some docker jobs run without any `services` section, so I'm going to drop it as well.

greg0ire:
> I noticed some docker jobs run without any services section, so I'm going to drop try dropping it as well.

Looks like it works too :)

tests/travis/docker-run-mysql-or-mariadb.sh
```diff
@@ -0,0 +1,37 @@
...
+    "$IMAGE" $CMD_OPTIONS
+
+function healthStatus()
+{
+  docker inspect --format "{{json .State.Health.Status }}" rdbms
+}
+
+until [ "$(healthStatus)" == '"healthy"' ]
+do
+  if [ "$(healthStatus)" == '"unhealthy"' ]
```
morozov: Given that there are only three possible statuses, could this be optimized and simplified using the following logic (didn't test it)?
```sh
while true; do
    $status = $(docker inspect --format "{{json .State.Health.Status }}" rdbms)

    if [ "$status" == '"healthy"' ]
    then
        break
    fi

    if [ "$status" == '"unhealthy"' ]
    then
        exit 1
    fi

    echo "Waiting for RDBMS to become ready (status = $status)…"
    sleep 1
done
```

greg0ire: I avoided `status` because it seems to be reserved in some shells (zsh?): vim colors it differently than other names, and it is defined in my shell, although I have many plugins.

greg0ire: Also used a switch case to avoid an infinite loop

greg0ire:
If this one passes it will be 5 green builds in a row.

tests/travis/docker-run-mysql-or-mariadb.sh
```diff
@@ -0,0 +1,42 @@
...
+    --env MYSQL_DATABASE=doctrine_tests \
+    --publish 33306:3306 \
+    --name rdbms \
+    "$IMAGE" $CMD_OPTIONS
+
+while true; do
+  healthStatus=$(docker inspect --format "{{json .State.Health.Status }}" rdbms)
+  case $healthStatus in
+    '"starting"')
+      echo "Waiting for RDBMS to become ready (status = $healthStatus)…">&2
```
morozov: Since you're using the default case to handle unknown statuses, displaying the status here is unnecessary (it’s always the same).

tests/travis/docker-run-mysql-or-mariadb.sh
```diff
@@ -0,0 +1,42 @@
...
+    '"healthy"')
+      echo "Container is healthy">&2
+      break
+      ;;
+    '"unhealthy"')
+      echo "Container is unhealthy">&2
+      exit 1
+      ;;
+    *)
+      echo "Unexpected health status $healthStatus…">&2
```
morozov: The … at the seems added mistakenly. We're not waiting for anything at this point. Also, using `…` with a monospace font looks… um… unusual.

greg0ire:
> If this one passes it will be 5 green builds in a row.

Let's try a 6th one :rocket:

morozov:
> Let's try a 6th one rocket

I'd say, at the current scale, 6 is enough. Great work @greg0ire!

morozov (dismissed):

morozov:
Two more last moment things:

1. I believe the PR should target `2.10`, not `2.10.x`.
2. Please double-check if code coverage configuration needs to be updated https://github.com/doctrine/dbal/blob/a0e23739c7bce97776e32a985bedb63756681547/.scrutinizer.yml#L24.

tests/travis/docker-run-mysql-or-mariadb.sh
greg0ire: I removed it, and yes it's unusal, but I find it more beautiful :P

greg0ire:
This PR now has commits from #3883, I'm going to leave them, because they are needed. It would be great to fix this situation. 2.10.x is created every time a release is done with the automatic release tool, correct? If yes maybe we should adopt this branch and drop 2.10 (and that kind of naming)?

morozov:
>  It would be great to fix this situation. 2.10.x is created every time a release is done with the automatic release tool, correct?

Correct and agree.

> If yes maybe we should adopt this branch and drop 2.10 (and that kind of naming)?

Yes. Not sure what it takes from the website configuration. If you want to take care of this, please do.

morozov:
@greg0ire since the code coverage on Scrutinizer has dropped, could you double-check if we’re not missing any coverage except for the unsupported platforms? Just in case.

greg0ire:
> If you want to take care of this, please do.

See https://github.com/doctrine/dbal/pull/3885

greg0ire:
> Can we keep testing the old platforms using the old distro instead of just removing the jobs?

We probably can, but should we really? I'm really not for dropping things as soon as possible, but supporting platforms that are not supported by the vendor themselves seems to be another extreme.

> @greg0ire since the code coverage on Scrutinizer has dropped, could you double-check if we’re not missing any coverage except for the unsupported platforms? Just in case.

It seems mostly related to IBM DB2… I did touch that job to remove the `services: [docker]` section, but can't see how that would explain this drop in coverage.

greg0ire:
:thinking: no Scrutinizer job now…

morozov:
> I'm really not for dropping things as soon as possible, but supporting platforms that are not supported by the vendor themselves seems to be another extreme.

I agree and I'd rather drop their support in a minor release than be puristic about semver and keep dragging this legacy. Although, this PR doesn't drop support for legacy platforms but instead stops testing still supported code.

> It seems mostly related to IBM DB2… I did touch that job to remove the services: `[docker] section`, but can't see how that would explain this drop in coverage.

It may be a Scrutinizer bug (because it's buggy as hell) but maybe not. If we don't see this issue on other PRs, then most likely this one is causing the trouble.

greg0ire:
> It may be a Scrutinizer bug (because it's buggy as hell) but maybe not. If we don't see this issue on other PRs, then most likely this one is causing the trouble.

I found a setting about tracking branches, and added 2.10.x as a tracked branch. let's trigger a new build.

greg0ire:
> Although, this PR doesn't drop support for legacy platforms but instead stops testing still supported code.

This was done in #3883 actually. How do we usually drop support for legacy platforms?

morozov:
> How do we usually drop support for legacy platforms?

AFAIK, we only do that in the next major as required by semver. Doing it earlier will require at least the team’s approval. Apart from that, the [community](https://twitter.com/srgmrzv/status/1211357385865879552) is quite serious about maintaining the BC promise even if there's no practical sense.

greg0ire:
Ok so this time, no coverage drop. Since 2.10.x had no Scrutinizer job (because of settings), there might have been a coverage drop there, and it might have gone unnoticed. :sweat:

greg0ire:
>  Apart from that, the community is quite serious 

This tweet is unavailable

greg0ire:
> BTW, apparently, it's possible to specify distro on the per-job basis in Scrutinizer.

Did you mean Travis?

greg0ire:
Anyway, there's probably no need to change distros if we use Docker, right?

morozov:
> This tweet is unavailable

Weird. The tweet is available (Dec 29, 2019) but the link seems to work only if clicked from the twitter feed, not from another website.

> Did you mean Travis?

Yes, Travis.

> Ah but that would mean using Docker for Postgres…

Yes. We can do it later if needed.

morozov (approved):

greg0ire:
@SenseException well that was harder than expected :P

SenseException:
Seems like it.

morozov:
> […] well that was harder than expected :P

You shouldn't have requested my review :smile:

PR #3887, opened 2020-03-04
Replace || by concat on postgreSQLPlatform

NiniGeek:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | doctrine/orm#3938

#### Summary
For PostGreSQL, the CONCAT DQL function is translated in concatenation with || operator (which is the default behavior in AbstractPlatform class).

The concatenation with || operator return null if one of the part is null, whereas CONCAT function will simply ignore null values.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -118,6 +118,14 @@ class PostgreSqlPlatform extends AbstractPlatform
     }
 
     /**
+     * {@inheritDoc}
+     */
+    public function getConcatExpression()
+    {
+        return sprintf('CONCAT(%s)', implode(', ', func_get_args()));
+    }
+
+    /**
      * {@inheritdoc}
      */
     protected function getDateArithmeticIntervalExpression($date, $operator, $interval, $unit)
--- a/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
@@ -21,6 +21,7 @@ use function array_unique;
 use function array_values;
 use function count;
 use function explode;
+use function func_get_args;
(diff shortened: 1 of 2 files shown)


SenseException:
The changes are looking good, but I'm not sure if this should be part of 2.10 or for the next major release. If `CONCAT` wasn't used in DBAL because of PostgreSQL's legacy (`||` was there before the SQL function was added), this won't be a bug fix, but a BC break.

I currently would like to see this to be in the next major release if `||` was preferred for compatibility reasons in the past. I also would like to hear what @doctrine/team-dbal have to say about it.

morozov:
Agree with the above.

oleg-andreyev:
It seems that CONCAT function was added only in PostgreSQL 9.1 (https://www.postgresql.org/docs/9.1/functions-string.html#FUNCTIONS-STRING-OTHER) so without checking Platform Version I think it cannot be merged.

morozov:
Since the different implementations `CONCAT()` and `||` have different semantics, before choosing what's right and what's wrong, I'd like to have a cross-platform integration test that covers other platforms. We cannot change the semantics of the API just because the new implementation looks better.

oleg-andreyev:
I've decided to check Travis configuration and found that DBAL is tested against PostgreSQL 9.4
https://github.com/doctrine/dbal/blob/ff1501a12ae604dd3fc910713a3b9c25d30663db/.travis.yml#L100

so can we say that 9.4 is the lowest version DBAL is supporting?

morozov:
@oleg-andreyev DBAL 2 supports at least PostgreSQL 9.1 as the `PostgreSQL91Platform` class implies. We don't test the versions older than 9.4 on Travis because they are not supported by the community and that would require extra effort from the team. But we don't accept any changes that would knowingly break the support.

The support for PostgreSQL versions lower than 9.4 will be dropped in DBAL 3.0 (https://github.com/doctrine/dbal/pull/3912).

PR #3891, opened 2020-03-06
Support named parameters in MySQLi connection->prepare()

MalteWunsch:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues |

#### Summary

According to the [documentation](https://www.doctrine-project.org/projects/doctrine-dbal/en/2.10/reference/data-retrieval-and-manipulation.html#dynamic-parameters-and-prepared-statements), you can use named parameters in `$connection->prepare($query)`. This PR tests and adds support for that with MySQLi connections.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
+++ b/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
@@ -14,6 +14,7 @@ use mysqli_stmt;
 use PDO;
 use function array_combine;
 use function array_fill;
+use function array_filter;
 use function array_key_exists;
 use function array_keys;
 use function assert;
@@ -24,7 +25,6 @@ use function get_resource_type;
 use function is_array;
 use function is_int;
 use function is_resource;
-use function is_string;
 use function sprintf;
 use function str_repeat;
 use function strlen;
@@ -176,9 +176,7 @@ class MysqliStatement implements IteratorAggregate, Statement
      */
     public function execute($params = null)
     {
-        if (is_array($params) && count($params) > 0) {
-            $params = $this->convertNamedToPositionalParams($params);
-        }
+        $params = $this->convertNamedToPositionalParamsIfNeeded($params);
 
         if ($this->_bindedValues !== null) {
             if ($params !== null) {
@@ -194,52 +192,40 @@ class MysqliStatement implements IteratorAggregate, Statement
             throw new MysqliException($this->_stmt->error, $this->_stmt->sqlstate, $this->_stmt->errno);
         }
 
-        if ($this->_columnNames === null) {
-            $meta = $this->_stmt->result_metadata();
-            if ($meta !== false) {
-                $fields = $meta->fetch_fields();
-                assert(is_array($fields));
+        $this->initializeColumnNamesIfNeeded();
 
-                $columnNames = [];
-                foreach ($fields as $col) {
-                    $columnNames[] = $col->name;
-                }
+        if ($this->_columnNames === false) {
+            $this->result = true;
 
-                $meta->free();
-
-                $this->_columnNames = $columnNames;
-            } else {
-                $this->_columnNames = false;
-            }
+            return true;
         }
 
-        if ($this->_columnNames !== false) {
-            // Store result of every execution which has it. Otherwise it will be impossible
-            // to execute a new statement in case if the previous one has non-fetched rows
-            // @link http://dev.mysql.com/doc/refman/5.7/en/commands-out-of-sync.html
-            $this->_stmt->store_result();
+        // Store result of every execution which has it. Otherwise it will be impossible
+        // to execute a new statement in case if the previous one has non-fetched rows
+        // @link http://dev.mysql.com/doc/refman/5.7/en/commands-out-of-sync.html
+        $this->_stmt->store_result();
 
-            // Bind row values _after_ storing the result. Otherwise, if mysqli is compiled with libmysql,
-            // it will have to allocate as much memory as it may be needed for the given column type
-            // (e.g. for a LONGBLOB field it's 4 gigabytes)
-            // @link https://bugs.php.net/bug.php?id=51386#1270673122
-            //
-            // Make sure that the values are bound after each execution. Otherwise, if closeCursor() has been
-            // previously called on the statement, the values are unbound making the statement unusable.
-            //
-            // It's also important that row values are bound after _each_ call to store_result(). Otherwise,
-            // if mysqli is compiled with libmysql, subsequently fetched string values will get truncated
-            // to the length of the ones fetched during the previous execution.
-            $this->_rowBindedValues = array_fill(0, count($this->_columnNames), null);
+        // Bind row values _after_ storing the result. Otherwise, if mysqli is compiled with libmysql,
+        // it will have to allocate as much memory as it may be needed for the given column type
+        // (e.g. for a LONGBLOB field it's 4 gigabytes)
+        // @link https://bugs.php.net/bug.php?id=51386#1270673122
+        //
+        // Make sure that the values are bound after each execution. Otherwise, if closeCursor() has been
+        // previously called on the statement, the values are unbound making the statement unusable.
+        //
+        // It's also important that row values are bound after _each_ call to store_result(). Otherwise,
+        // if mysqli is compiled with libmysql, subsequently fetched string values will get truncated
+        // to the length of the ones fetched during the previous execution.
+        assert(is_array($this->_columnNames));
+        $this->_rowBindedValues = array_fill(0, count($this->_columnNames), null);
 
-            $refs = [];
-            foreach ($this->_rowBindedValues as $key => &$value) {
-                $refs[$key] =& $value;
-            }
+        $refs = [];
+        foreach ($this->_rowBindedValues as $key => &$value) {
+            $refs[$key] =& $value;
+        }
 
-            if (! $this->_stmt->bind_result(...$refs)) {
-                throw new MysqliException($this->_stmt->error, $this->_stmt->sqlstate, $this->_stmt->errno);
-            }
+        if (! $this->_stmt->bind_result(...$refs)) {
+            throw new MysqliException($this->_stmt->error, $this->_stmt->sqlstate, $this->_stmt->errno);
         }
 
         $this->result = true;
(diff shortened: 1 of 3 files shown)


SenseException:
Is this currently still WIP? It seems like the new data provider is breaking tests.

MalteWunsch:
Right, but I'm not sure how to solve that problem.

My intention was to add named parameters support in MySQLi connection->prepare() only, and with MySQLi the tests run well. But the new tests are run with many other drivers / database connections which do not support named parameters there, resulting in them failing.

I don't think I'm able to add support for named parameters to the other drivers, as I don't have these databases at hand.

Should I limit the test to MySQLi and if so: how do I do that? Maybe the test class annotation `@group DDC-1372` indicates in one way or the other which databases are used?

SenseException (changes requested):
> Should I limit the test to MySQLi

You'll find that there are such limitations in the tests if the tested code is only supporting one specific driver. You can find one on the `setUp()` method of `Doctrine\Tests\DBAL\Functional\Driver\Mysqli\ConnectionTest` where it skips every database that isn't using the mysqli driver or doesn't have the mysqli extension. Please also put your test into the same directory and rename it to `MysqliStatementTest`.

MalteWunsch:
Thanks for the hints!

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -54,6 +60,9 @@ class MysqliStatement implements IteratorAggregate, Statement
     /** @var string */
     protected $types;
 
+    /** @var array<string, array<int, int>> maps parameter names to their placeholder number(s). */
+    protected $placeholderNamesToNumbers = [];
```
greg0ire: Let's go private by default, regardless of other properties
```suggestion
    private $placeholderNamesToNumbers = [];
```

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -97,6 +107,35 @@ public function __construct(mysqli $conn, $prepareString)
...
     }
 
+    /**
+     * Converts named placeholders (":parameter") into positional ones ("?"), as MySQL does not support them.
+     *
+     * @param string $query The query string to create a prepared statement of.
+     *
+     * @return string
+     */
+    private function convertNamedToPositionalPlaceholders($query)
```
greg0ire: Type declarations of docblocks please
```suggestion
     */
    private function convertNamedToPositionalPlaceholders(string $query) : string
```

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -97,6 +107,35 @@ public function __construct(mysqli $conn, $prepareString)
...
+            $placeholderName = (string) $placeholderName;
+            if (array_key_exists($placeholderName, $this->placeholderNamesToNumbers) === false) {
+                $this->placeholderNamesToNumbers[$placeholderName] = [];
+            }
+
+            $this->placeholderNamesToNumbers[$placeholderName][] = $placeholderNumber++;
+
+            $placeholderPositionInShortenedQuery = $placeholderPosition - $numberOfCharsQueryIsShortenedBy;
+            $placeholderNameLength               = strlen($placeholderName);
+            $query                               = substr($query, 0, $placeholderPositionInShortenedQuery) . '?' . substr($query, ($placeholderPositionInShortenedQuery + $placeholderNameLength + 1));
```
greg0ire: Let's use linebreaks and remove unneeded parenthesis
```suggestion
            $query                               = substr($query, 0, $placeholderPositionInShortenedQuery) . '?' . 
                substr($query, $placeholderPositionInShortenedQuery + $placeholderNameLength + 1);
```

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -151,59 +192,87 @@ public function execute($params = null)
...
 
         $this->result = true;
 
         return true;
     }
 
+    /**
+     * Converts an array of named parameters, e.g. ['id' => 1, 'foo' => 'bar'] to the corresponding array with
+     * positional parameters referring to the prepared query, e.g. [1 => 1, 2 => 'bar', 3 => 'bar'] for a prepared query
+     * like "SELECT id FROM table WHERE foo = :foo and baz = :foo".
```
greg0ire:
```suggestion
     * such as "SELECT id FROM table WHERE foo = :foo and baz = :foo".
```

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -151,59 +192,87 @@ public function execute($params = null)
...
+     *
+     * @param array<int|string, mixed>|null $params
+     *
+     * @return mixed[]|null more specific: array<int, mixed>, I just don't know an elegant way to convince phpstan
```
greg0ire: Can you elaborate? What's wrong with phpstan here? Please reproduce the issue on https://phpstan.org/ so that we can give it a try.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -151,59 +192,87 @@ public function execute($params = null)
...
+
+        return $positionalParameters;
+    }
+
+    /**
+     * @param mixed[] $array
+     *
+     * @return bool
+     */
+    private function arrayHasOnlyIntegerKeys(array $array)
```
greg0ire:
```suggestion
     */
    private function arrayHasOnlyIntegerKeys(array $array) : bool
```

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -151,59 +192,87 @@ public function execute($params = null)
...
+    /**
+     * Converts an array of named parameters, e.g. ['id' => 1, 'foo' => 'bar'] to the corresponding array with
+     * positional parameters referring to the prepared query, e.g. [1 => 1, 2 => 'bar', 3 => 'bar'] for a prepared query
+     * like "SELECT id FROM table WHERE foo = :foo and baz = :foo".
+     *
+     * @param array<int|string, mixed>|null $params
+     *
+     * @return mixed[]|null more specific: array<int, mixed>, I just don't know an elegant way to convince phpstan
+     */
+    private function convertNamedToPositionalParamsIfNeeded(?array $params = null)
```
greg0ire:
```suggestion
    private function convertNamedToPositionalParamsIfNeeded(?array $params = null) : array
```

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -151,59 +192,87 @@ public function execute($params = null)
...
+    {
+        if ($params === null || count($params) === 0) {
```
greg0ire: The case of `[]` does not need to be singled out IMO.
```suggestion
        if ($params === null) {
```

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
```diff
@@ -450,4 +519,28 @@ public function getIterator()
     {
         return new StatementIterator($this);
     }
+
+    private function initializeColumnNamesIfNeeded()
```
greg0ire:
```suggestion
    private function initializeColumnNamesIfNeeded() : void
```

tests/Doctrine/Tests/DBAL/Functional/Driver/Mysqli/MysqliStatementTest.php
```diff
@@ -0,0 +1,142 @@
...
+                    ['id' => 4, 'foo' => 1, 'bar' => 4],
+                ],
+            ],
+
+            'parameter with string value' => [
+                'SELECT * FROM mysqli_statement_test f WHERE f.foo = :foo',
+                ['foo' => '"\''],
+                [],
+            ],
+        ];
```
greg0ire: What happens if there are more parameters than needed in the query? Or when there are fewer than needed (an exception I hope).

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
MalteWunsch: Minimal example for the phpstan playground:

```
<?php declare(strict_types = 1);

class Example {
    /**
     * @param array<int|string, mixed> $params
     *
     * @return array<int, mixed>
     */
    private function convertNamedToPositionalParamsIfNeeded(array $params): array
    {
        if ($this->arrayHasOnlyIntegerKeys($params)) {
            return $params;
        }

        // removed logic
        return [];
    }
	
    /**
     * @param array<int|string, mixed> $array
     *
     * @return bool
     */
    private function arrayHasOnlyIntegerKeys(array $array): bool
    {
        return count(array_filter(array_keys($array), 'is_int')) === count($array);
    }	
}
```

Then the error is the following: "Method Example::convertNamedToPositionalParamsIfNeeded() should return array<int, mixed> but returns array<int|string, mixed>."

I don't know an elegant way to convince phpstan that the keys in the array are actually  integers. I tried various things like

```
foreach (array_keys($array) as $key) {
    assert(is_int($key));
}
```

without success. The only way I found was creating a new array explicitly with integers as keys, but obviously that's not a good idea with respect to performance/memory.

One could use `array<int|string, mixed>` as return type docblock, being more specific than mixed[]. But for me, this level of exactness felt a little bit like a sham.

Do you have an oppinion - or even better: a solution on that?

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
MalteWunsch: The idea was to deal with any kind of emptyness first, but the empty array behaves well in the following code, too. So I'm happy to go either way.

tests/Doctrine/Tests/DBAL/Functional/Driver/Mysqli/MysqliStatementTest.php
MalteWunsch: Too few parameters:

```
Doctrine\DBAL\DBALException: An exception occurred while executing 'SELECT * FROM mysqli_statement_test f WHERE f.foo = :foo':

Wrong parameter count for mysqli_stmt::bind_param()
```

Too many parameters:

```
Doctrine\DBAL\DBALException: An exception occurred while executing 'SELECT * FROM mysqli_statement_test f WHERE f.foo = :foo' with params [1, 2]:

Undefined index: bar
```

where `bar` is the name of the superfluous named parameter.

Are these exceptions okay? Do you want a test for them?

morozov:
@MalteWunsch named parameters and their conversion to the positional ones are supported at the wrapper connection level, not at the driver connection. Why do you need to implement them at the driver level?

lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
greg0ire: Here is a solution you can use https://phpstan.org/r/c66bb446-e691-4902-a013-e21c9dd9f08f

tests/Doctrine/Tests/DBAL/Functional/Driver/Mysqli/MysqliStatementTest.php
greg0ire: I will answer once and if https://github.com/doctrine/dbal/pull/3891#issuecomment-606299763 is resolved

PR #3892, opened 2020-03-08
Add DBAL Connection Registry

dmaicher:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | -

#### Summary

I am currently working on extracting the DBAL part out of `DoctrineBundle` and into https://github.com/doctrine/dbal-bundle. See https://github.com/doctrine/dbal-bundle/pull/1.

I discussed some details with @stof and @alcaeus on the last SymfonyHackday in Amsterdam.

The new dbal-bundle will need a `ConnectionRegistry` to retrieve registered dbal connections. Since the existing [ConnectionRegistry](https://github.com/doctrine/persistence/blob/master/lib/Doctrine/Persistence/ConnectionRegistry.php) in the persistence namespace is a bit generic (can return any object and is not specificly for DBAL connections) we propose to add this new registry interface (and a PSR-11 container implementation) here.

@stof maybe you can elaborate some more on how this could potentially also be used to simplify the DBAL commands?

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "d06c0dac6585d3779056ae835b684c93",
+    "content-hash": "6d678fda37ae2cbf5e558ee3bc6cf0ab",
     "packages": [
         {
             "name": "doctrine/cache",
@@ -203,6 +203,55 @@
             ],
             "description": "Composer plugin that provides efficient querying for installed package versions (no runtime IO)",
             "time": "2019-11-15T16:17:10+00:00"
+        },
+        {
+            "name": "psr/container",
+            "version": "1.0.0",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/php-fig/container.git",
+                "reference": "b7ce3b176482dbbc1245ebf52b181af44c2cf55f"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/php-fig/container/zipball/b7ce3b176482dbbc1245ebf52b181af44c2cf55f",
+                "reference": "b7ce3b176482dbbc1245ebf52b181af44c2cf55f",
+                "shasum": ""
+            },
+            "require": {
+                "php": ">=5.3.0"
+            },
+            "type": "library",
+            "extra": {
+                "branch-alias": {
+                    "dev-master": "1.0.x-dev"
+                }
+            },
+            "autoload": {
+                "psr-4": {
+                    "Psr\\Container\\": "src/"
+                }
+            },
+            "notification-url": "https://packagist.org/downloads/",
+            "license": [
+                "MIT"
+            ],
+            "authors": [
+                {
+                    "name": "PHP-FIG",
+                    "homepage": "http://www.php-fig.org/"
+                }
+            ],
+            "description": "Common Container Interface (PHP FIG PSR-11)",
+            "homepage": "https://github.com/php-fig/container",
+            "keywords": [
+                "PSR-11",
+                "container",
+                "container-interface",
+                "container-interop",
+                "psr"
+            ],
+            "time": "2017-02-14T16:28:37+00:00"
         }
     ],
     "packages-dev": [
@@ -2435,8 +2484,8 @@
             "authors": [
                 {
                     "name": "Arne Blankerts",
-                    "role": "Developer",
-                    "email": "arne@blankerts.de"
+                    "email": "arne@blankerts.de",
+                    "role": "Developer"
                 }
             ],
             "description": "A small library for converting tokenized PHP source code into XML and potentially other formats",
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,8 @@
         "php": "^7.3",
         "doctrine/cache": "^1.0",
         "doctrine/event-manager": "^1.0",
-        "ocramius/package-versions": "^1.4"
+        "ocramius/package-versions": "^1.4",
+        "psr/container": "^1.0"
     },
     "require-dev": {
         "doctrine/coding-standard": "^7.0",
--- a/lib/Doctrine/DBAL/Registry/ConnectionRegistry.php
+++ b/lib/Doctrine/DBAL/Registry/ConnectionRegistry.php
@@ -1,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Registry;
+
+use Doctrine\DBAL\Connection;
+use InvalidArgumentException;
+
+interface ConnectionRegistry
+{
+    /**
+     * Gets the default connection name.
+     *
+     * @return string The default connection name.
+     */
+    public function getDefaultConnectionName() : string;
+
+    /**
+     * Gets the named connection.
+     *
+     * @param string $name The connection name (null for the default one).
+     *
+     * @throws InvalidArgumentException in case the connection for the given name does not exist.
+     */
+    public function getConnection(?string $name = null) : Connection;
+
+    /**
+     * Gets an array of all registered connections.
+     *
+     * @return array<string, Connection> An array of Connection instances.
+     */
+    public function getConnections() : array;
+
+    /**
+     * Gets all connection names.
+     *
+     * @return array<string> An array of connection names.
+     */
+    public function getConnectionNames() : array;
+}
--- a/lib/Doctrine/DBAL/Registry/Psr11ConnectionRegistry.php
+++ b/lib/Doctrine/DBAL/Registry/Psr11ConnectionRegistry.php
@@ -1,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Registry;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Registry\ConnectionRegistry;
+use InvalidArgumentException;
+use Psr\Container\ContainerInterface;
+use function sprintf;
+
+class Psr11ConnectionRegistry implements ConnectionRegistry
+{
+    /** @var ContainerInterface */
+    private $container;
+
+    /** @var string */
+    private $defaultConnectionName;
+
+    /** @var string[] */
+    private $connectionNames;
+
+    /**
+     * @param string[] $connectionNames
+     */
+    public function __construct(ContainerInterface $container, string $defaultConnectionName, array $connectionNames)
+    {
+        $this->container             = $container;
+        $this->defaultConnectionName = $defaultConnectionName;
+        $this->connectionNames       = $connectionNames;
+    }
+
+    public function getDefaultConnectionName() : string
+    {
+        return $this->defaultConnectionName;
+    }
+
+    public function getConnection(?string $name = null) : Connection
+    {
+        $name = $name ?? $this->defaultConnectionName;
+
+        if (! $this->container->has($name)) {
+            throw new InvalidArgumentException(sprintf('Connection with name "%s" does not exist.', $name));
+        }
+
+        return $this->container->get($name);
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getConnections() : array
+    {
+        $connections = [];
+
+        foreach ($this->connectionNames as $connectionName) {
+            $connections[$connectionName] = $this->container->get($connectionName);
+        }
+
+        return $connections;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function getConnectionNames() : array
+    {
+        return $this->connectionNames;
+    }
+}
--- a/tests/Doctrine/Tests/DBAL/Registry/Psr11ConnectionRegistryTest.php
+++ b/tests/Doctrine/Tests/DBAL/Registry/Psr11ConnectionRegistryTest.php
@@ -1,0 +1,83 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL\Registry;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Registry\ConnectionRegistry;
+use Doctrine\DBAL\Registry\Psr11ConnectionRegistry;
+use InvalidArgumentException;
+use PHPUnit\Framework\MockObject\Stub;
+use PHPUnit\Framework\TestCase;
+use Psr\Container\ContainerInterface;
+use function array_keys;
+
+class Psr11ConnectionRegistryTest extends TestCase
+{
+    /** @var ConnectionRegistry */
+    private $registry;
+
+    /** @var array<string, Connection&Stub> */
+    private $connections;
+
+    protected function setUp() : void
+    {
+        /** @var Connection&Stub $fooConnection */
+        $fooConnection = $this->createStub(Connection::class);
+        /** @var Connection&Stub $barConnection */
+        $barConnection = $this->createStub(Connection::class);
+
+        $this->connections = [
+            'foo' => $fooConnection,
+            'bar' => $barConnection,
+        ];
+
+        /** @var ContainerInterface&Stub $container */
+        $container = $this->createStub(ContainerInterface::class);
+        $container->method('has')
+            ->willReturnCallback(function (string $name) : bool {
+                return isset($this->connections[$name]);
+            });
+
+        $container->method('get')
+            ->willReturnCallback(function (string $name) : Connection {
+                return $this->connections[$name];
+            });
+
+        $this->registry = new Psr11ConnectionRegistry($container, 'bar', array_keys($this->connections));
+    }
+
+    public function testGetDefaultConnection() : void
+    {
+        $this->assertSame($this->connections['bar'], $this->registry->getConnection());
+    }
+
+    public function testGetConnectionByName() : void
+    {
+        $this->assertSame($this->connections['foo'], $this->registry->getConnection('foo'));
+        $this->assertSame($this->connections['bar'], $this->registry->getConnection('bar'));
+    }
+
+    public function testGetNotExistentConnection() : void
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Connection with name "something" does not exist.');
+        $this->registry->getConnection('something');
+    }
+
+    public function testGetDefaultConnectionName() : void
+    {
+        $this->assertSame('bar', $this->registry->getDefaultConnectionName());
+    }
+
+    public function getGetConnections() : void
+    {
+        $this->assertSame($this->connections, $this->registry->getConnections());
+    }
+
+    public function testGetConnectionNames() : void
+    {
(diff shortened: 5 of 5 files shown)


stof:
Well, if we want to simplify the management of commands in the bundle, the idea would be to move the support of ``--connection` allowing to select a connection from the registry into DBAL itself (once the registry is there, there is no issue depending on it in commands, especially if we also provide another "dumb" implementation for the case where you need only one connection that can be wrapped in this registry for compat with commands)

lib/Doctrine/DBAL/ConnectionRegistry.php
```diff
@@ -0,0 +1,40 @@
...
+     * @return string The default connection name.
+     */
+    public function getDefaultConnectionName() : string;
+
+    /**
+     * Gets the named connection.
+     *
+     * @param string $name The connection name (null for the default one).
+     *
+     * @throws InvalidArgumentException in case the connection for the given name does not exist.
```
stof: should we throw a generic InvalidArgumentException or an exception defined in DBAL ?

lib/Doctrine/DBAL/Psr11ConnectionRegistry.php
```diff
@@ -0,0 +1,69 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL;
```
stof: should we put that class at the root, or in a `Registry` subnamespace, to allow grouping implementations if needed ?

lib/Doctrine/DBAL/ConnectionRegistry.php
dmaicher: For now I used this generic exception similar to

https://github.com/doctrine/persistence/blob/master/lib/Doctrine/Persistence/AbstractManagerRegistry.php#L98

But we can also create a more specific `ConnectionNotFoundException` or so

composer.lock
```diff
@@ -769,7 +818,7 @@
                     "homepage": "https://nette.org/contributors"
                 }
             ],
-            "description": "🔍 Nette Finder: find files and directories with an intuitive API.",
+            "description": "? Nette Finder: find files and directories with an intuitive API.",
```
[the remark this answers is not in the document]
dmaicher: no idea :confused: 

just ran `composer req psr/container` again on a clean lock file from master and this is the result for me

```
it@5a2d39b05156:/var/www/dbal  (connection_registry) $ composer --version
Composer version 1.10.1 2020-03-13 20:34:27
```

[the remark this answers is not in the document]
morozov: You can remove `vendor/` and rerun `composer install`. It always works for me in this case.

morozov:
Why is it needed in the DBAL? Would it make sense to maintain it in the bundle instead?

dmaicher:
> Why is it needed in the DBAL? Would it make sense to maintain it in the bundle instead?

@morozov see comment from @stof above: https://github.com/doctrine/dbal/pull/3892#issuecomment-604584671

It could be useful here to allow selecting a connection for the dbal commands.

Some quick changes to show how:

```diff
diff --git a/lib/Doctrine/DBAL/Tools/Console/Command/RunSqlCommand.php b/lib/Doctrine/DBAL/Tools/Console/Command/RunSqlCommand.php
index c7e51933f..4d00db2bc 100644
--- a/lib/Doctrine/DBAL/Tools/Console/Command/RunSqlCommand.php
+++ b/lib/Doctrine/DBAL/Tools/Console/Command/RunSqlCommand.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\DBAL\Tools\Console\Command;
 
+use Doctrine\DBAL\ConnectionRegistry;
 use Doctrine\DBAL\Tools\Dumper;
 use LogicException;
 use RuntimeException;
@@ -23,6 +24,17 @@ use function stripos;
  */
 class RunSqlCommand extends Command
 {
+    /**
+     * @var ConnectionRegistry
+     */
+    private $connectionRegistry;
+
+    public function __construct(ConnectionRegistry $connectionRegistry)
+    {
+        parent::__construct();
+        $this->connectionRegistry = $connectionRegistry;
+    }
+
     protected function configure() : void
     {
         $this
@@ -30,6 +42,7 @@ class RunSqlCommand extends Command
         ->setDescription('Executes arbitrary SQL directly from the command line.')
         ->setDefinition([
             new InputArgument('sql', InputArgument::REQUIRED, 'The SQL statement to execute.'),
+            new InputOption('connection', null, InputOption::VALUE_REQUIRED, 'The database connection', 'default'),
             new InputOption('depth', null, InputOption::VALUE_REQUIRED, 'Dumping depth of result set.', 7),
             new InputOption('force-fetch', null, InputOption::VALUE_NONE, 'Forces fetching the result.'),
         ])
@@ -44,7 +57,7 @@ EOT
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $conn = $this->getHelper('db')->getConnection();
+        $conn = $this->connectionRegistry->getConnection($input->getOption('connection'));
 
         $sql = $input->getArgument('sql');
 
diff --git a/lib/Doctrine/DBAL/Tools/Console/ConsoleRunner.php b/lib/Doctrine/DBAL/Tools/Console/ConsoleRunner.php
index a7a1e987e..1936da9d2 100644
--- a/lib/Doctrine/DBAL/Tools/Console/ConsoleRunner.php
+++ b/lib/Doctrine/DBAL/Tools/Console/ConsoleRunner.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\DBAL\Tools\Console;
 
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\ConnectionRegistry;
 use Doctrine\DBAL\Tools\Console\Command\ReservedWordsCommand;
 use Doctrine\DBAL\Tools\Console\Command\RunSqlCommand;
 use Doctrine\DBAL\Tools\Console\Helper\ConnectionHelper;
@@ -33,23 +34,23 @@ class ConsoleRunner
      *
      * @param array<int, Command> $commands
      */
-    public static function run(HelperSet $helperSet, array $commands = []) : void
+    public static function run(ConnectionRegistry $connectionRegistry, array $commands = []) : void
     {
         $cli = new Application('Doctrine Command Line Interface', Versions::getVersion('doctrine/dbal'));
 
         $cli->setCatchExceptions(true);
-        $cli->setHelperSet($helperSet);
+        //$cli->setHelperSet($helperSet);
 
-        self::addCommands($cli);
+        self::addCommands($connectionRegistry, $cli);
 
         $cli->addCommands($commands);
         $cli->run();
     }
 
-    public static function addCommands(Application $cli) : void
+    public static function addCommands(ConnectionRegistry $connectionRegistry, Application $cli) : void
     {
         $cli->addCommands([
-            new RunSqlCommand(),
+            new RunSqlCommand($connectionRegistry),
             new ReservedWordsCommand(),
         ]);
     }
```

When the `cli-config.php` script then returns a `ConnectionRegistry` and not a `HelperSet` like:

```php
<?php

use Doctrine\DBAL\Driver\PDOMySql\Driver;
use Doctrine\DBAL\Psr11ConnectionRegistry;
use Symfony\Component\DependencyInjection\Container;

$container = new Container();
$container->set('default', new \Doctrine\DBAL\Connection([], new Driver()));

return new Psr11ConnectionRegistry(
    $container,
    'default',
    ['default']
);
```

Then we have native multi-connection support on the dbal commands and not only when wrapped by the doctrine bundle :blush: 

Does this make sense to you?

stof:
And this also means we don't need to extend all commands in bundles anymore to add support for multiple connections (which will allow the possibility to actually make them final in a future major version if needed).

morozov:
> And this also means we don't need to extend all commands in bundles anymore to add support for multiple connections […]

It doesn't sound like valid reasoning to me. An API should be extensible and extended according to the open-close principle. Not modified.

> […] which will allow the possibility to actually make them final in a future major version if needed

We can make the commands explicitly extensible (not via inheritance) if projects need. Or at least agree to not make them final.

Just adding a bunch of new code just because another project needs it looks like a bad idea to me.

alcaeus:
@morozov handling multiple connections in the commands is very painful at the moment, since the connection helper provided by DBAL only knows a static connection. Right now, DoctrineBundle (and future DBALBundle and ORMBundle) need to extend the command to add a `connection` argument, then hook into the `execute` method to inject the selected connection into the helper.

The helper provided by DBAL can only handle a single connection, which is why we have this very helpful suggestion for users of DBAL to get the console commands running ([source](https://github.com/doctrine/dbal/blob/ff1501a12ae604dd3fc910713a3b9c25d30663db/src/Tools/Console/ConsoleRunner.php#L60..L78)):

```
// replace with the mechanism to retrieve DBAL connection in your app
$connection = getDBALConnection();
```

We're telling the user to handle the connection, but we're not giving them any tools to do so. For a long time, we've used the DoctrineBundle to handle these shortcomings, at the expense of anyone not using the bundles to integrate ORM or DBAL into their projects (e.g. because they are not using Symfony).

Having a connection registry that can be used to handle connections along with an CLI argument to select a connection would improve usability for all users, by allowing us to document how to add connections to this registry and how they can be selected in commands (e.g. by passing a `connection` CLI argument).

> Just adding a bunch of new code because another project needs it looks like a bad idea to me.

Sorry, but no. We're not adding a "bunch of new code" because another project needs it. We're adding it because DBAL knows about connections and is technically able to handle multiples of them, but doesn't provide any tools working with them. This is not any project needing this, we're talking about our own projects here that ensure 90% of our users get a sensible set of tools to work with.

Keeping this code in the bundle doesn't make any sense: it will always be Doctrine people maintaining it, but the maintenance burden is increased because any change that should affect a single project (e.g. adding a new CLI command in DBAL) requires work in the bundle (e.g. extending that command with boilerplate code to inject the selected connection). This doesn't make any sense.

lib/Doctrine/DBAL/Psr11ConnectionRegistry.php
alcaeus: Alternative suggestion for a namespace:
```suggestion
namespace Doctrine\DBAL\Connection;
```

tests/Doctrine/Tests/DBAL/Psr11ConnectionRegistryTest.php
```diff
@@ -0,0 +1,78 @@
...
+    private $connections;
+
+    protected function setUp() : void
+    {
+        $this->connections = [
+            'foo' => $this->createMock(Connection::class),
+            'bar' => $this->createMock(Connection::class),
+        ];
+
+        $container = $this->createMock(ContainerInterface::class);
```
greg0ire, line 30:
```suggestion
        $container = $this->createStub(ContainerInterface::class);
```

tests/Doctrine/Tests/DBAL/Psr11ConnectionRegistryTest.php
```diff
@@ -0,0 +1,78 @@
...
+        $container->expects($this->any())
+            ->method('has')
```
greg0ire, line 32:
```suggestion
        $container->method('has')
```

tests/Doctrine/Tests/DBAL/Psr11ConnectionRegistryTest.php
```diff
@@ -0,0 +1,78 @@
...
+            ->willReturnCallback(function (string $name) {
+                return isset($this->connections[$name]);
+            });
+
+        $container->expects($this->any())
+            ->method('get')
```
greg0ire, line 38:
```suggestion
        $container->method('get')
```

tests/Doctrine/Tests/DBAL/Psr11ConnectionRegistryTest.php
```diff
@@ -0,0 +1,78 @@
...
+            ->willReturnCallback(function (string $name) {
```
greg0ire, line 39:
```suggestion
            ->willReturnCallback(function (string $name) : Connection {
```

lib/Doctrine/DBAL/Psr11ConnectionRegistry.php
dmaicher: Then I think I would go for `Registry`. Otherwise I find it confusing that the `Connection` itself is not in the `Connection` namespace :smile: 

Or?

composer.lock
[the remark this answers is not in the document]
dmaicher: I tried that... also does not work for me :confused:

greg0ire:
> It doesn't sound like valid reasoning to me. An API should be extensible and extended according to the open-close principle. Not modified.

That's not how I understand the open-closed principle at all. This PR only adds new types, so there's no OCP violation here IMO. Or do you consider adding a new type modification of the package? To me, it changes nothing to existing behavior here: the API is extended, not modified, so much so that it looks completely BC to me. @dmaicher, you should target a lower branch IMO.

Also, as a developer, if I have several databases, I would very much like to be able to name connections to them and retrieve them from my DI container.

tests/Doctrine/Tests/DBAL/Psr11ConnectionRegistryTest.php
```diff
@@ -0,0 +1,83 @@
...
+
+    protected function setUp() : void
+    {
+        /** @var Connection&Stub $fooConnection */
+        $fooConnection = $this->createStub(Connection::class);
+        /** @var Connection&Stub $barConnection */
+        $barConnection = $this->createStub(Connection::class);
+
+        $this->connections = [
+            'foo' => $fooConnection,
```
greg0ire, line 32: What would be a realistic name? `doctrine.dbal.connection.foo`? If yes maybe there should a decorating implementation that has a `$prefix` argument so that we can still use `getConnection('foo')?

dmaicher: For now I don't think we need to worry about prefixes.

In the Bundle the name will be whatever you configure on the config. So `foo` would be a possible name.

Don't confuse it with the Symfony service name for the connections. Those will be like `doctrine.dbal.foo_connection` for example.

Also see

https://github.com/doctrine/dbal-bundle/pull/1/files#diff-6648b62ba0e1a03e4f23ed65e8efb9f5R70

stof:
@morozov there is no clean way to support multiple connections in commands right now, because DBAL commands are about built around a single connection and don't have an option to select it.
Extending the command in the bundle is the only way we have currently to add that support in the command. And being able to manage multiple connections and choosing the one for which the command is executed is a common need in projects.

I'm all for the open-closed principle. And that's precisely the reason why I suggested to move the support for connection switching in commands to DBAL itself, because adding it from the bundles forces us to break that principle.

tests/Doctrine/Tests/DBAL/Psr11ConnectionRegistryTest.php
stof: @greg0ire DoctrineDBALBundle won't wire this with the full Symfony container, but with a ServiceLocator, mapping connection names to services, not service ids to services. So this is indeed realistic (well, as far as `foo` is a realistic name for your connection)

greg0ire: Oh ok I get it, thanks!

lib/Doctrine/DBAL/Psr11ConnectionRegistry.php
dmaicher: Changed it to `Doctrine\DBAL\Registry` for now

morozov:
Currently, DBAL doesn't have any notion of multiple connections. This is true. And adding a registry here doesn't add this notion. If you want to implement a feature of managing multiple connections, please come up with a proposal and we can discuss it. Just adding a registry solves the problem of the bundle and the projects that use it but it doesn't belong to the DBAL.

stof:
@morozov the goal is to have a follow-up PR moving the support for `--connection` to the console commands implemented in DBAL itself, instead of having to extend all commands in the bundle to support that.

AFAIK, that's the only part of DBAL where it is relevant, because that's the only part of DBAL needing to get the connection. Everything else is happening from the connection already (other projects built on top of DBAL may have places where this is relevant though, and having the registry in DBAL rather than the bundle might also make it easier for them to support it).

dmaicher:
@morozov @stof I will try to come up with another PR that will use the registry on the commands

morozov:
> the goal is to have a follow-up PR moving the support for `--connection` to the console commands […] AFAIK, that's the only part of DBAL where it is relevant

If that's the case, then why do you need a registry at all? Each command invocation will use at most one command given the connection name, e.g. `function (string $name) : Connection`. Please do not over-engineer this unless you get paid for LOC.

stof:
@morozov The advantage of the registry over a callable `?string $name => Connection` is that the signature of the callable cannot be enforced by PHP.
Also, for people using the Symfony DI container, wiring a callable is more complex (we would have to implement the same object, except that we would have `__invoke` rather than `getConnection`).

morozov:
@stof let's not over-think this and solve one problem at a time. Please. If the problem at hand is the CLI command, let's fix it by introducing the necessary minimum of code. I do not see a registry or any API exposed from the DBAL needed to solve the CLI command problem.

dmaicher:
@morozov so you would prefer something like this?

https://github.com/doctrine/dbal/compare/2.11.x...dmaicher:commands_use_registry?expand=1

morozov:
Yeah, something like that but I don't get it who's going to inject the registry into the command. Right now, the command gets the DB from the helper. Could the helper implement `getDefaultConnection()` and `getConnection($name)` instead?

I still fail to see the need for a dedicated registry in the DBAL if it's not going to use it anywhere else than in CLI.

morozov:
Okay… as far as I understand, an application that depends on DBAL and uses its CLI command may manage the DB connections itself and may want to inject those commands into the command explicitly. In this case, instead of a registry, I'd introduce only an interface in DBAL like:
```php
interface ConnectionProvider
{
    function getDefaultConnection() : Connection;
    function getConnection(string $name) : Connection;
}
```
If the application manages connections in an PSR-11 compatible registry, it could wrap it into the interface above and inject. I still don't see much point in implementing the registry in DBAL or supporting PSR-11 for connections. It's a general-purpose DI container while we can afford a more specific interface.

dmaicher:
@morozov yes we can also create a more simplified interface for this. Seems better than using callbacks.

We should then deprecate the `ConnectionHelper`, or? We could have a BC layer with deprecations in 2.11 and then make it mandatory in 3.0?

morozov:
> We should then deprecate the `ConnectionHelper`, or? We could have a BC layer with deprecations in 2.11 and then make it mandatory to inject a provider in 3.0?

Why is it “or”? It's all of the above. The connector provider and the BC layer are implemented, and the helper is deprecated in 2.11, then the BC layer is removed in 3.0.

alcaeus:
> Why is it “or”?

Because it's a very common German thing to end a sentence with "or?" to get confirmation, similar to "right?" in English.

morozov:
Then we're on the same page, or?

dmaicher:
@morozov yes I will propose a new PR :+1:

PR #3893, opened 2020-03-10
Ensure the constructor arguments are passed to custom classes

duncan3dc:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

Currently using `FetchMode::CUSTOM_OBJECT` with constructor arguments doesn't work as the constructor arguments are silently dropped in the statement wrapper.
I had a look around and couldn't find any suggestion this is intentional, it's fixed in `master` as a by-product of  82f369f3ef99645991388cb979c6454bd9560ef2 so this PR proposes a fix for 2.*

### DIFF
--- a/lib/Doctrine/DBAL/Statement.php
+++ b/lib/Doctrine/DBAL/Statement.php
@@ -249,11 +249,7 @@ class Statement implements IteratorAggregate, DriverStatement
      */
     public function fetchAll($fetchMode = null, $fetchArgument = null, $ctorArgs = null)
     {
-        if ($fetchArgument) {
-            return $this->stmt->fetchAll($fetchMode, $fetchArgument);
-        }
-
-        return $this->stmt->fetchAll($fetchMode);
+        return $this->stmt->fetchAll($fetchMode, $fetchArgument, $ctorArgs);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


morozov:
This API is not portable and personally I'd rather drop it than support. See the details in https://github.com/doctrine/dbal/pull/3070.

What makes you want to construct objects by DBAL instead of constructing them in the application?

duncan3dc:
I have an object that does not map to a table, I want to construct it from the result of a complex query. I couldn't see any Doctrine API to achieve this, but `FetchMode::CUSTOM_OBJECT` seems to work exactly as I need (populating the private properties of my object).

If there's another way then this can be closed but could you point me in the right direction please?

morozov:
Why don't you fetch the data and instantiate the object using the `new` keyword? Why do you need the DBAL or the ORM to do that for you?

duncan3dc:
That would require a named constructor with many parameters. I wanted to leverage Doctrine functionality to populate private properties

morozov:
> I wanted to leverage Doctrine functionality to populate private properties.

So without being able to construct the object without DBAL, how are you going to unit-test it?

duncan3dc:
We integration test against a temporary database

ostrolucky:
I think such simple, ready bugfix should be accepted for 2.x. It's fixing bug after all and I can't immediately see what dev resources would it take to merge this, even if support is removed later. Removing support of it is concern of 3.x.

morozov:
Thank you for the reminder, @ostrolucky. That was the plan.

morozov (approved):

morozov:
Thank you @duncan3dc.

PR #3894, opened 2020-03-14
Make sure that the $types array has the same keys $params

### LINKED ISSUE #2897: Incorrect handling of list parameters with sparse types

morozov:
This will allow omitting parameters with unspecified types

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

Fixes #2897.

#### Summary

1. The problem is only solved for positional parameters as originally reported. Solving it for named parameters requires additional refactoring since the named parameter and type keys may or may not have the optional `:` prefix.
2. The addition of `NULL` values to the expected values of `$types` is not considered a BC-break since the `NULL` values are insignificant. Getting rid of them would require additional refactoring and might complicate the implementation as well from the performance standpoint.
3. The data item comments have been converted to keys for better development experience.

### DIFF
--- a/lib/Doctrine/DBAL/SQLParserUtils.php
+++ b/lib/Doctrine/DBAL/SQLParserUtils.php
@@ -4,7 +4,9 @@ namespace Doctrine\DBAL;
 
 use const PREG_OFFSET_CAPTURE;
 use function array_fill;
+use function array_fill_keys;
 use function array_key_exists;
+use function array_keys;
 use function array_merge;
 use function array_slice;
 use function array_values;
@@ -131,6 +133,10 @@ class SQLParserUtils
         $bindIndex      = -1;
 
         if ($isPositional) {
+            // make sure that $types has the same keys as $params
+            // to allow omitting parameters with unspecified types
+            $types += array_fill_keys(array_keys($params), null);
+
             ksort($params);
             ksort($types);
         }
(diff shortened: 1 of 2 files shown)


morozov:
The failure on Oracle is irrelevant and must be caused by the ContinuousPHP environment instability (to be solved separately).

greg0ire:
Is this still a work in progress? If yes, what is left to do?

morozov:
@greg0ire it's ready for review. I removed the `WIP` label.

greg0ire (approved):

PR #3896, opened 2020-03-16
Updated documentation for QueryBuilder::execute() return value type

### LINKED ISSUE #3603: Remove suppression for PHPStan errors fixing which in 2.x would be a BC break
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.0

Something along these lines:
```sh
$ grep "BC break" phpstan.neon.dist 
        # removing it would be BC break
        # declaring $tableName in AbstractSchemaManager::_getPortableTableIndexesList() non-optional will be a BC break
        # changing these would be a BC break, to be done in next major
```

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

Fixes #3603

> https://github.com/doctrine/dbal/blob/68f0d99fcf270525b3df59294f277a6643facbf9/phpstan.neon.dist#L18

Despite the original comment, it's not a BC break since the return value type is not enforced at runtime and is just documentation. We’re not changing the actual return value type which has always been `ResultStatement|int` regardless of the docblock.

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -3,7 +3,7 @@
 namespace Doctrine\DBAL\Query;
 
 use Doctrine\DBAL\Connection;
-use Doctrine\DBAL\Driver\Statement;
+use Doctrine\DBAL\Driver\ResultStatement;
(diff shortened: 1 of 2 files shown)


greg0ire (approved):

greg0ire:
Thanks @morozov !

PR #3901, opened 2020-03-17
Removing duplicate driver from get connection method documentation

jpruiz114:
This is a very simple commit to remove a duplicate driver mentioned in the getConnection method of the DriverManager class. This is not a performance improvement or a bug fix, it's simply something I noticed duplicated in the PHPDocs of the method.


greg0ire (approved):
This problem only affects master, this is the correct branch.

greg0ire:
Thanks @jpruiz114 !

jpruiz114:
Hey @greg0ire glad that I can contribute! It's just a small detail but I will try to help with more significant commits in the future!

Issue #3877, opened 2020-02-18
OracleSchemaManager listTableDetails use index COMMENTS non-existent


Hectorfdv88:
In  Doctrine\DBAL\Schema\OracleSchemaManager.php  line 402

```
public function listTableDetails($tableName) : Table
    {
        $table = parent::listTableDetails($tableName);

        /** @var OraclePlatform $platform */
        $platform = $this->_platform;
        $sql      = $platform->getListTableCommentsSQL($tableName);

        $tableOptions = $this->_conn->fetchAssoc($sql);
        $table->addOption('comment', $tableOptions['COMMENTS']);

        return $table;
    }

```
Please change to

```
public function listTableDetails($tableName) : Table
    {
        $table = parent::listTableDetails($tableName);

        /** @var OraclePlatform $platform */
        $platform = $this->_platform;
        $sql      = $platform->getListTableCommentsSQL($tableName);

        $tableOptions = $this->_conn->fetchAssoc($sql);
        $table->addOption('comment', $tableOptions['comments']);

        return $table;
    }
```

because index COMMENTS is case sensitive, and Oracle return lower case comments.

This is tested un Oracle 18c.

Thanks!

SenseException:
Thank you for opening this issue. Please show what the current result is and what the expected result is supposed to be. As far as I understand you'll get an Notice for an undefined index. Is that correct?
Is this something that differs between Oracle versions?

cesarhein:
I have the same issue. When i run the "php artisan migrate" command in a Laravel project i get this Exception:

   ErrorException  : Undefined index: COMMENTS

  at C:\xampp\htdocs\voyager_oracle_4\vendor\doctrine\dbal\lib\Doctrine\DBAL\Schema\OracleSchemaManager.php:402
    398|         $platform = $this->_platform;
    399|         $sql      = $platform->getListTableCommentsSQL($tableName);
    400| 
    401|         $tableOptions = $this->_conn->fetchAssoc($sql);
  > 402|         $table->addOption('comment', $tableOptions['COMMENTS']);
    403| 
    404|         return $table;
    405|     }
    406| }

  Exception trace:

[...]

PR #3907, opened 2020-03-18
Fix error "Undefined index: COMMENTS"

cesarhein:
Inside "listTableDetails" function you are trying to access the "COMMENTS" index, but the correct way would be "comments". To be 100% sure, we always convert the keys to lower case.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3877

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Schema/OracleSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Schema/OracleSchemaManagerTest.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL\Schema;
+
+use Doctrine\DBAL\Schema\Table;
+use PHPUnit\Framework\TestCase;
+
+class OracleSchemaManagerTest extends TestCase
+{
+    public function testListTableDetails() : void
+    {
+        $table = new Table('test');
+
+        self::assertFalse($table->hasOption('comment'));
+
+        $tableOptions = ['COMMENTS' => 'a comment'];
+        $tableOptionsLowerKeys = array_change_key_case($tableOptions, CASE_LOWER);
+        $table->addOption('comment', $tableOptionsLowerKeys['comments']);
+
+        self::assertTrue($table->hasOption('comment'));
+        self::assertEquals('a comment', $table->getComment());
+    }
+}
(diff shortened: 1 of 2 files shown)


morozov:
@cesarhein thank you for the patch. We’ll need an integration test for it (ideally, a cross-platform one).

tests/Doctrine/Tests/DBAL/Schema/OracleSchemaManagerTest.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL\Schema;
+
+use Doctrine\DBAL\Schema\Table;
+use PHPUnit\Framework\TestCase;
+
+class OracleSchemaManagerTest extends TestCase
```
morozov, line 10: It needs to be an integration test that would call `listTableDetails()` and fail on Oracle thereby reproducing the problem. Unit tests are not really helpful in this case.

morozov (changes requested):

tests/Doctrine/Tests/DBAL/Schema/OracleSchemaManagerTest.php
cesarhein: I get your point, you're right. But I don't have any experience doing functional tests. Could you point me to a similar example within the project to use as a guide? Thank you very much!

morozov: Look at the [`DateExpressionTest`](https://github.com/doctrine/dbal/blob/v2.10.1/tests/Doctrine/Tests/DBAL/Functional/Platform/DateExpressionTest.php) or [`DataAccessTest`](https://github.com/doctrine/dbal/blob/v2.10.1/tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php) for example.

Mind that the `master` branch currently represents DBAL 4.0 with the unknown release date. Since this is a bug fix, you will most likely want to target your PR against the `2.10.x` branch.

imami:
Any update on this issue and this pull request ?

imami:
@doctrine
@morozov This bug is really annoying for years.
Overriding the whole `SchemaManage` class in composer.json is what many developers do just to fix this simple issue.
Is there a near plan to accept this PR or applying any other bug fix?

morozov:
@imami this PR will be accepted as soon as it's ready. I haven't seen any updates on it in a year.

greg0ire:
@imami would you be willing to take over this PR?

morozov:
Closing as stale and missing functional tests.

PR #3906, opened 2020-03-18
Backport removal of legacy platforms and drivers from master

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

**Removed platforms and drivers:**
1. PostgreSQL <9.3, SQL Server < 2008, SQL Anywhere < 16, PDO-based IBM DB2 driver (https://github.com/doctrine/dbal/pull/3055).
2. Drizzle (https://github.com/doctrine/dbal/pull/3040).
3. Azure and sharding (https://github.com/doctrine/dbal/pull/3602).
4. PostgreSQL 9.3, MariaDB 10.0 (https://github.com/doctrine/dbal/pull/3477).
5. SQL Server 2008 (https://github.com/doctrine/dbal/pull/3756).

**TODO:**
- [x] Makes sure it will cleanly merge back to `master` as a validation of the proper backport.
- [x] Once this is merged, update the milestones of the backported PRs and the corresponding issues.

PR #3912, opened 2020-03-20
Made some platform class names reflect the lowest server version they support

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no †

The `PostgreSqlPlatform`, `SQLAnywherePlatform` and `SQLServerPlatform` classes have been renamed to `PostgreSQL94Platform`, `SQLAnywhere16Platform` and `SQLServer2012Platform` respectively.

This way, when deprecating an old platform version, we can deprecate and later remove a specific class. Otherwise, there's no way to express the deprecation on the class level and the upgrade notes contain sentences like: "`XYZPlatform` and `XYZKeywords` now represent XYZ (version)".

I deliberately didn't rename `MySqlPlatform` since there's no clear understanding of which MySQL version it corresponds to. We can remove it when we drop the support for MySQL 5.6.

Additionally, some leftovers of the platform and keyword class are removed (somehow they survived the backport in https://github.com/doctrine/dbal/pull/3906).

† Despite the fact that the class names are being changed, such classes exist in `2.x` and have the same API and purpose.


UPGRADE.md
```diff
@@ -1,15 +1,32 @@
 # Upgrade to 3.0
 
+## BC BREAK: Removed support for SQL Anywhere 12 and older
+
+DBAL now requires SQL Anywhere 16 or newer, support for unmaintained versions has been dropped.
+If you are using any of the legacy versions, you have to upgrade to newer SQL Anywhere version (16+).
```
```suggestion
If you are using any of the legacy versions, you have to upgrade to a newer SQL Anywhere version (16+).
```

morozov: Thanks, @greg0ire. Corrected.

greg0ire (approved):

greg0ire:
Thanks @morozov !

Issue #3918, opened 2020-03-24
Wrong SQL statement causes ORA-22859 error when trying to make nullable a CLOB column  


cesarhein:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.1

#### Summary

In a Oracle DB I'm trying to set as nullable a CLOB column in a table. But the SQL statement generated is wrong.

#### Current behaviour

The SQL statement generated currently is:

**ALTER TABLE table_name MODIFY (column_name CLOB DEFAULT NULL NULL)**

#### How to reproduce

1.- Create a table with a text (CLOB) column, not null, and without default value.
2.- Try to change column making it nullable and with default value = null.
3.- Get Error: ORA-22859

#### Expected behaviour

The correct SQL statement should be:

**ALTER TABLE table_name MODIFY (column_name DEFAULT NULL NULL)**

Same SQL statement but without data type.

#### Explanation

Oracle has a set of limitations on column modification operations. It is not allowed to modify data type for CLOB/NCLOB/BLOB columns even for an empty table.
For these reason, when we put the CLOB data type in the SQL statement, Oracle throw the ORA-22859 error assuming that we are trying to modify the data type.
To avoid this behaviour, we must simply omit the data type. Data type is not necessary for change nullable attribute or default value.

morozov:
The reproducer:
```php
$desiredTable = new Table('table_name');
$desiredTable->addColumn('column_name', 'text', ['notnull' => true]);

$sm = $connection->createSchemaManager();
$sm->createTable($desiredTable);

$actualTable = $sm->listTableDetails('table_name');
$desiredTable->getColumn('column_name')
    ->setNotnull(false)
    ->setDefault(null);
$comparator = $sm->createComparator();
$diff = $comparator->diffTable($actualTable, $desiredTable);
$sm->alterTable($diff);
```

PR #3919, opened 2020-03-25
Fix SQL that modify column attr when type remains

cesarhein:
Omitting data type in SQL statenment. Data type is not necessary for change nullable attribute or default value.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3918 

#### Summary

When generating the SQL statement that modifies column attributes, we check if the data type is being changed. If so, we place the new data type in the query. If it is not being changed, we ignore the data type.
To do that, a new function "getColumnDeclarationWithoutTypeSQL" was created, that is identical to "getColumnDeclarationSQL", with the only difference that it omits the field data type.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
@@ -934,7 +934,7 @@ SQL
      * This declaration is particularly useful when some property of the column changes
      *  (e.g. nullable, default value), but the data type does not change.
      */
-    public function getColumnDeclarationWithoutTypeSQL($name, array $field)
+    private function getColumnDeclarationWithoutTypeSQL($name, array $field)
     {
         if (isset($field['columnDefinition'])) {
             $columnDef = $this->getCustomTypeDeclarationSQL($field);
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -922,6 +928,37 @@ public function getColumnDeclarationSQL($name, array $field)
         return $name . ' ' . $columnDef;
     }
 
+    /**
+     * This function is a copy of getColumnDeclarationSQL.
+     * The only difference is that the data type is not in the column declaration.
+     * This declaration is particularly useful when some property of the column changes
+     *  (e.g. nullable, default value), but the data type does not change.
+     */
+    public function getColumnDeclarationWithoutTypeSQL($name, array $field)
```
SenseException: This method doesn't need to be public.
```suggestion
    private function getColumnDeclarationWithoutTypeSQL($name, array $field)
```

SenseException (changes requested):
* Please format your code to the guidelines of our coding style that can be checked with phpcs
* Please add tests that are covering your code

SenseException:
> Please format your code to the guidelines of our coding style that can be checked with phpcs

What I meant was that your changes had to be adapted to the coding style. Adding new type hints and return types to the public methods in this version are BC breaks. Please revert these back. 

Please also refrain from merging the 2.10.x branch into your branch and prefer a rebase instead. Can you undo the merge commit and rebase from 2.10.x?

PR #3924, opened 2020-03-28
Actualize the content of the .gitattributes file

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

1. Removed no longer existing entries (`build.properties`, `build.xml`)
2. Added new entries (`.doctrine-project.json`, `*.md` files except `README.md`).
3. Sorted using `sort` (ignores case and the leading dot, same as `ls`)

**TODO:**
- [ ] Add `/ci` after merging to `3.0.x`.


greg0ire:
How about also adding such a comment: https://github.com/doctrine/migrations/blob/master/.gitignore#L16 ?

morozov:
While I agree with the statement, I don't think it's the right place for it (same as for the IDE's files).

greg0ire (approved):

PR #3932, opened 2020-04-07
Enable PHPStan strict rules

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

The initial work was done on top of `master` and then backported to `3.0.x` to get the tests improved as well. This is an important addition on top of the existing code quality checks since it helps to find potential bugs and prevents the new code of questionable quality from being introduced.

Additionally, in `master` the test suite will be cleaned up since a lot of assertions are no longer needed thanks to the improved type system (https://github.com/doctrine/dbal/pull/3348).

**Fixed issues:**

The analysis identified a couple of potential issues and deviations from the best practices:
1. The _constraint_ `$name` variable declared in the `foreach` loop overrides the _table_ `$name` method argument that is used down the line. Most likely, a table with such constraints will have an invalid name: https://github.com/doctrine/dbal/blob/8be6d1632696641c06c992b3a41bc973d1ac4583/src/Platforms/SqlitePlatform.php#L329-L331
2. This is unlikely a bug but usually, a loose comparison of the `strpos()` return value looks suspicious: https://github.com/doctrine/dbal/blob/8be6d1632696641c06c992b3a41bc973d1ac4583/src/Schema/OracleSchemaManager.php#L127
3. Also unlikely a bug but depending on the fetch mode, an empty value stored in the cache may be ignored: https://github.com/doctrine/dbal/blob/936e7e3350dd15a2d3de147821dfe5249169ed8b/src/Cache/ResultCacheStatement.php#L84-L85
4. An empty-ish comment like `"0"` set on a table in Oracle may be ignored: https://github.com/doctrine/dbal/blob/8be6d1632696641c06c992b3a41bc973d1ac4583/src/Platforms/OraclePlatform.php#L788-L790
5. This code is barely readable. The only situation when this expression evaluates to `TRUE` is when the counts are not equal which should have been coded explicitly: https://github.com/doctrine/dbal/blob/8be6d1632696641c06c992b3a41bc973d1ac4583/src/Platforms/SQLServer2012Platform.php#L1353

**Technically, BC-breaks:**
1. A zero value of the `$fetchMode` argument in the `Statement` methods is no longer ignored. Not that anyone should have used it but in order to use the default mode, it should be omitted or expressed as `NULL`.
2. The empty values of the "connectstring", "host", "service" and "instancename" parameters of the `oci8` and `pdo_oci` drivers are no longer ignored if passed explicitly.
3. The empty value of the "port" parameter of the `pdo_sqlsrv` driver is no longer ignored if passed explicitly.
4. The following `QueryBuilder` methods still accept null as an optional parameter but will no longer ignore another explicitly passed empty value: `::select()`, `::addSelect()`, `::insert()`, `::update()`, `::delete()`.
5. `Statement::execute()`, `QueryBuilder::groupBy()` and `::addGroupBy()` will no longer an explicitly passed non-array empty value.
6. `AbstractPlatform::getTrimExpression()` will no longer ignore an empty value of `$mode` (use `TrimMode::UNSPECIFIED` or omit); and an empty value of `$char` (use `false` or omit).

 Other empty values in various configurations are not ignored if passed explicitly

**Future improvements identified**

The errors suppressed in the PHPStan configuration identify the by-design issues that need to be fixed in the future:

1. Internal data structures are represented as untyped associative arrays instead of objects (e.g. connection configuration, table/column definitions). A similar improvement was recently implemented in https://github.com/doctrine/dbal/pull/3836.
2. The `AbstractSchemaManager::tryMethod()` implements dynamic method calls and needs to be redesigned and internalized to DBAL.

**Additional, not strictly relevant issues fixed:**
1. The comma in the regular expression pattern doesn't need to be escaped: https://github.com/doctrine/dbal/blob/936e7e3350dd15a2d3de147821dfe5249169ed8b/src/Schema/MySqlSchemaManager.php#L146

**TODO:**
- [x] Rebase on top of https://github.com/doctrine/dbal/pull/3933.
- [x] Rebase on top of https://github.com/doctrine/dbal/pull/3934.
- [x] Document the no longer ignored empty values.
- [ ] When merging up to `master`, revisit the list of whitelisted errors. Some of them will be no longer relevant.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,8 @@
     "require-dev": {
         "doctrine/coding-standard": "^6.0",
         "jetbrains/phpstorm-stubs": "^2019.1",
-        "phpstan/phpstan": "^0.12",
+        "phpstan/phpstan": "^0.12.18",
+        "phpstan/phpstan-strict-rules": "^0.12.2",
         "phpunit/phpunit": "^9.0",
         "symfony/console": "^2.0.5|^3.0|^4.0|^5.0"
     },
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "33232ba3c41c9de6f6bc60c784726991",
+    "content-hash": "194abb042dbd3732c1da46cf05ad7a07",
     "packages": [
         {
             "name": "doctrine/cache",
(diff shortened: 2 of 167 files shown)


greg0ire (approved):
>  Document the no longer ignored empty values.

Is this something you plan to do before merging? Approving in case it isn't

I think Scrutinizer fails because if views the change in SQLSeverSchemaManager as an increase in complexity, but to me the `if` is easier to understand than the ternary

morozov:
> Is this something you plan to do before merging? Approving in case it isn't

I was going to put this in the PR description, no more code changes expected.

> I think Scrutinizer fails because if views the change in SQLSeverSchemaManager as an increase in complexity […]

Yes, and we can't do anything with it than ignore it.

> […] but to me the `if` is easier to understand than the ternary

I agree. Furthermore, the original reason for this was to pass the option down the line only if it's explicitly set. There's no point in passing the default explicitly. Also, it makes line coverage report more specific.

Issue #2925, opened 2017-11-26
Incorrect handling changed primary key


SCIF:
I've received next error message:

```
SQLSTATE[42704]: Undefined object: 7 ERROR:  index "primary" does not exist
```
Searching got me next ticket: https://github.com/doctrine/doctrine2/issues/6271

Here is a dump of diff:
```
Doctrine\DBAL\Schema\TableDiff {#2273
  +name: "sellers"
  +newName: false
  +addedColumns: array:1 [
    "id" => Doctrine\DBAL\Schema\Column {#1972
      #_type: Doctrine\DBAL\Types\IntegerType {#183}
      #_length: null
      #_precision: 0
      #_scale: 0
      #_unsigned: true
      #_fixed: false
      #_notnull: true
      #_default: null
      #_autoincrement: false
      #_platformOptions: array:1 [
        "version" => false
      ]
      #_columnDefinition: null
      #_comment: null
      #_customSchemaOptions: []
      #_name: "id"
      #_namespace: null
      #_quoted: false
    }
  ]
  +changedColumns: []
  +removedColumns: array:1 [
    "user_id" => Doctrine\DBAL\Schema\Column {#1410
      #_type: Doctrine\DBAL\Types\SmallIntType {#171}
      #_length: null
      #_precision: 10
      #_scale: 0
      #_unsigned: false
      #_fixed: false
      #_notnull: true
      #_default: null
      #_autoincrement: false
      #_platformOptions: []
      #_columnDefinition: null
      #_comment: null
      #_customSchemaOptions: []
      #_name: "user_id"
      #_namespace: null
      #_quoted: false
    }
  ]
  +renamedColumns: []
  +addedIndexes: []
  +changedIndexes: array:1 [
    "sellers_pkey" => Doctrine\DBAL\Schema\Index {#1993
      #_columns: array:1 [
        "id" => Doctrine\DBAL\Schema\Identifier {#1994
          #_name: "id"
          #_namespace: null
          #_quoted: false
        }
      ]
      #_isUnique: true
      #_isPrimary: true
      #_flags: []
      -options: []
      #_name: "primary"
      #_namespace: null
      #_quoted: false
    }
  ]
  +removedIndexes: []
  +renamedIndexes: []
  +addedForeignKeys: []
  +changedForeignKeys: []
  +removedForeignKeys: []
  +fromTable: Doctrine\DBAL\Schema\Table {#1426
    #_name: "sellers"
    #_columns: array:3 [
[...]

PR #3936, opened 2020-04-09
Detect index rename/change columns for primary keys as remove/create

goetas:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | https://github.com/doctrine/orm/issues/6271 https://github.com/doctrine/dbal/issues/2925 https://github.com/doctrine/dbal/pull/3184 https://github.com/doctrine/dbal/pull/2929

https://github.com/doctrine/dbal/pull/3936/files?w=1 for a diff without whitespaces

#### Summary

When renaming a primary key, and at the same time changing some columns,  the `Comparator` does not detect the name correctly.

For databases as MySQL where primary keys do not have a name, this does not matter.
For database as PostgreSQL where also primary keys have a name, the generated SQL will try to drop a column that does not exists (It will try to drop the index with the new name).

Given a table:

```php
$table = new Table('users');
$table->addColumn('id', 'integer', ['notnull' => true]);
$table->addColumn('code', 'integer', ['notnull' => true]);
$table->setPrimaryKey(['id'], 'users_pkey');
```

when we change:

```php
$table->setPrimaryKey(['id', 'code'], 'test');
```

**Before** this patch, the generated code for postgres will be

```sql
DROP INDEX "test";
ALTER TABLE users ADD PRIMARY KEY (id, code)
````

This will fail as the index "test" is nonexistent.

**After** this patch, the generated code for postgres will be

```sql
DROP INDEX "users_pkey";
ALTER TABLE users ADD PRIMARY KEY (id, code)
````

### CODE REVIEW

File: tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php [hunk 1]
```diff
@@ -477,6 +479,41 @@ public function testCompareChangedIndexFieldPositions() : void
...
+
+        $platform = new PostgreSQL94Platform();
+        $sql      = $platform->getAlterTableSQL($tableDiff);
+        self::assertSame('DROP INDEX users_pkey', $sql[0]);
+        self::assertSame('ALTER TABLE users ADD PRIMARY KEY (id, code)', $sql[1]);
+
+        $platform = new MySqlPlatform();
+        $sql      = $platform->getAlterTableSQL($tableDiff);
+        self::assertSame('ALTER TABLE users DROP PRIMARY KEY', $sql[0]);
+        self::assertSame('ALTER TABLE users ADD PRIMARY KEY (id, code)', $sql[1]);
```
goetas, line 514: Please let me know if this should be part of this test

goetas:
Ugh..... this is a rabbit hole....

goetas:
How do you deal with the fact that most of the postgres tests are basically wrong?

```php
$table = new Table('users');
$table->addColumn('id', 'integer', ['notnull' => true]);
$table->setPrimaryKey(['id']);
```

The primary key index name is not `primary` but is `users_pkey`.... but assertions all over the code expect that...

goetas:
Nevermind! found the way :)

morozov:
@goetas the change looks reasonable but I believe it should have a functional test (the unit one is optional). The functional test should be run across all supported platforms.

goetas:
Hi, thanks for the feedback.
While writing the functional tests, figure out that is more complicated than the proposed solution. 

Will continue to work on it.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3174,6 +3211,16 @@ public function supportsPrimaryConstraints()
         return true;
     }
 
+    /**
+     * Whether the platform supports named primary key constraints.
+     *
+     * @return bool
+     */
+    public function supportsNamedPrimaryConstraints()
```
morozov, line 3219: I didn't think through this idea enough but it may be a good moment to start introducing platform-level interfaces for features like this. Right now, the features are structured in a way which is rather unfriendly to static analysis:
```php
function supportsTheFeature(): bool;

/**
 * @throws NotImplemented
 */
abstract function doTheFeature()
{
    throw new NotImplemented();
}

// client
if ($platform->supportsTheFeature()) {
    // we are not handling the exception which
    $platform->doTheFeature();
}
```

There's no way for the client to know whether or not to call a certain method on a certain platform and whether or not the documented exception needs to be handled at runtime.

Instead, all the feature's methods could be grouped under an interface so that the client knows that a feature can be used on the platform only if it implements the interface.

Later, the implementation of the features could be even moved out of the gigantic platform classes and composed.

goetas:
Mysql never stops surprising me:

```sql
CREATE TABLE me (id INT NOT NULL);
alter table me add constraint abc primary key (id);
SELECT COLUMN_NAME, CONSTRAINT_NAME FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE WHERE TABLE_NAME = 'me'
```

```
+-------------+-----------------+
| COLUMN_NAME | CONSTRAINT_NAME |
+-------------+-----------------+
| id          | PRIMARY         |
+-------------+-----------------+
1 row in set (0.03 sec)
```

(version 8.0.19)

:sob:

goetas:
well, its green for now! (with postgres, sqlserver and oracle support)

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
```diff
@@ -173,7 +173,7 @@ public function testNonDefaultPKOrder() : void
 CREATE TABLE non_default_pk_order (
     id INTEGER,
     other_id INTEGER,
-    PRIMARY KEY(other_id, id)
+    PRIMARY KEY (other_id, id)
```
morozov, line 176: Please revert the changes like this. It's fine if we want to adjust the formatting (it will need to be justified) but I'd rather do this in a separate PR to not mix two unrelated changes.

goetas: The problem is that currently the codebase is not consistent. Sometimes is `PRIMARY KEY(id)` and sometimes `PRIMARY KEY (id)`. In this PR i have unified the generation of the primary key definition, thus only one way will work

morozov: The unification is fine but I'd rather do it in a separate PR.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
greg0ire: Tell me if I should reopen https://github.com/doctrine/dbal/pull/2839

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
goetas: I can try to split it, but then that PR needs to be merged before this

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
goetas: will use that as reference...

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
greg0ire: How about just a separate commit? We can review commit by commit.

goetas: For me that would help a lot

morozov: A separate commit is fine too. Also, I'd suggest to not have more commits than needed to help the reviewer (i.e. 2 in this case). Having intermediate commits that fail the build sometimes results in horrible `git bisect` experience. If those in-between commits are not valuable from the reviewer's perspective, I'd squash them.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: @greg0ire it's not that black and white. I think the interfaces should be introduced only for something that can be done from the DBAL perspective on one platform and cannot be done on others (e.g. creating sequences). If something is doable in different ways on all platforms (e.g. column comments), the common interface should be defined in the abstract platform.

beberlei: @morozov would the result be that each platform implements 10-20 interfaces? How would that be different from the supports pattern if you do:

```php
if ($platform instanceof AllowsNamedPrimaryKeysPlatform) {
    $platform->doTheThing();
}
```
It would fatal error when not supported, so its also an "exception" and the instanceof check is essentialy the same as a `supports*` call.

morozov: Let's proceed with the `supports()` for now. Having `if`-statements is the biggest problem regardless of whether it's `supports()` or `instanceof`. The client should be able to express their intent (`$platform->doStuff()`), not introspect its capabilities and work around the differences on the client side.

goetas: hope to have some time to continue with this

rwitchell:
how's this progressing?

goetas:
I did not have the time to continue working on this change 😞

mishal:
Any progress with this? Can I help somehow?

morozov:
@mishal as far as I can tell there hasn't been any progress since https://github.com/doctrine/dbal/pull/3936#issuecomment-684534750. You can clone this PR, rebase it on `3.1.x`, address the code review comments, and add integration tests.

mishal:
thanks. 

3.x is no go for us... too many bc breaks.

PR #3939, opened 2020-04-10
Fix parsing column comments on SQLite

### LINKED ISSUE #3937: Column comment incorrectly introspected on SQLite
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.1

#### Summary

If a column without a comment follows a column with a comment that follows another column, then the first column comment is introspected with the value of the second column:

#### How to reproduce

```php
use Doctrine\DBAL\Connection;
use Doctrine\DBAL\Schema\Table;

    'driver' => 'pdo_sqlite',
    'dbname' => 'sqlite::memory:',
]);

$table = new Table('test');
$table->addColumn('col1', 'string', ['length' => 16]);
$table->addColumn('col2', 'string', ['length' => 16, 'comment' => 'Column #2']);
$table->addColumn('col3', 'string', ['length' => 16]);

$sm = $conn->getSchemaManager();
$sm->createTable($table);

$comment = $sm->listTableDetails('test')
    ->getColumn('col1')
    ->getComment();

var_dump($comment);

// string(9) "Column #2"
```

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

Fixes #3937

#### Summary

When a table DDL looks like the following:
```sql
CREATE TABLE column_comments (x VARCHAR(1), y VARCHAR(1) --Some comment
```
Then when parsing the comment for the `x` column, the existing `\(.*?\)` pattern instead of capturing `(1)`, captures `(1), y VARCHAR(1)`, so the `--Some comment` part is parsed as belonging to `x`.

The solution is to not allow the closing parentheses in the pattern to make sure that only the portion prior to the first closing parentheses is captured.

The existing non-greedy behavior of the pattern is not enough since even a non-greedy pattern aims to find a match while in our case there shouldn't be a match.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
@@ -278,4 +278,19 @@ SQL;
         // with an empty table, non autoincrement rowid is always 1
         $this->assertEquals(1, $lastUsedIdAfterDelete);
     }
+
+    public function testOnlyOwnCommentIsParsed() : void
+    {
+        $table = new Table('own_column_comment');
+        $table->addColumn('col1', 'string', ['length' => 16]);
+        $table->addColumn('col2', 'string', ['length' => 16, 'comment' => 'Column #2']);
+        $table->addColumn('col3', 'string', ['length' => 16]);
+
+        $sm = $this->connection->getSchemaManager();
+        $sm->createTable($table);
+
+        $this->assertNull($sm->listTableDetails('own_column_comment')
+            ->getColumn('col1')
+            ->getComment());
+    }
 }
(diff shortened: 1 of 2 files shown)


greg0ire (dismissed):

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
```diff
@@ -278,4 +278,19 @@ public function testPrimaryKeyNoAutoIncrement() : void
         // with an empty table, non autoincrement rowid is always 1
         $this->assertEquals(1, $lastUsedIdAfterDelete);
     }
+
+    public function testOnlyOwnCommentIsParsed() : void
+    {
+        $table = new Table('test');
```
greg0ire, line 284: Travis says that name is already in use

morozov: Damn non-isolated tests. Fixed.

greg0ire (approved):

greg0ire:
Thanks @morozov !

PR #3940, opened 2020-04-10
Update OracleSessionInit.php

OlivierToussaint:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
#### Summary

arguments like OPTIMIZER_INDEX_CACHING & OPTIMIZE_IDEX_COST_ADJ need numeric value

### DIFF
--- a/src/Event/Listeners/OracleSessionInit.php
+++ b/src/Event/Listeners/OracleSessionInit.php
@@ -53,7 +53,11 @@ class OracleSessionInit implements EventSubscriber
             if ($option === 'CURRENT_SCHEMA') {
                 $vars[] = $option . ' = ' . $value;
             } else {
-                $vars[] = $option . " = '" . $value . "'";
+                if (is_numeric($value)) {
+                    $vars[] = $option . " = " . $value;
+                else {
+                    $vars[] = $option . " = '" . $value . "'";
+                }
             }
         }
 
--- a/src/Event/Listeners/OracleSessionInit.php
+++ b/src/Event/Listeners/OracleSessionInit.php
@@ -55,7 +55,7 @@ class OracleSessionInit implements EventSubscriber
             } else {
                 if (is_numeric($value)) {
                     $vars[] = $option . " = " . $value;
-                else {
+                } else {
                     $vars[] = $option . " = '" . $value . "'";
                 }
             }


src/Event/Listeners/OracleSessionInit.php
```diff
@@ -53,7 +53,11 @@ public function postConnect(ConnectionEventArgs $args) : void
             if ($option === 'CURRENT_SCHEMA') {
                 $vars[] = $option . ' = ' . $value;
             } else {
-                $vars[] = $option . " = '" . $value . "'";
+                if (is_numeric($value)) {
+                    $vars[] = $option . " = " . $value;
+                else {
```
andrew-demb, line 58:
```suggestion
                } else {
```

src/Event/Listeners/OracleSessionInit.php
```diff
@@ -53,7 +53,11 @@ public function postConnect(ConnectionEventArgs $args) : void
             if ($option === 'CURRENT_SCHEMA') {
                 $vars[] = $option . ' = ' . $value;
             } else {
-                $vars[] = $option . " = '" . $value . "'";
+                if (is_numeric($value)) {
```
morozov, line 56: Would it make sense to check if the value `is_int()`? Why would someone need to pass `10000` as `"10000"`?

morozov (changes requested):
The patch looks reasonable but we'll need a test for it.

greg0ire:
If there is no BC-break, you should probably be targetting 2.7 (@morozov shouldn't it be labeled with "Bug" rather than with "Improvement"? Also, didn't we make 2.8.x the default in the settings? I thought I did).

OlivierToussaint:
Hello,

You need some context, oracle expert ask me to add some tricks to my app (made with Symfony).

Like that : 
```
OracleOptimizer:
    class: Doctrine\DBAL\Event\Listeners\OracleSessionInit
    arguments:
        - { OPTIMIZER_INDEX_CACHING: 50 }
        - { OPTIMIZER_INDEX_COST_ADJ: 5 }
    tags:
        - { name: doctrine.event_listener, event: postConnect }
```

This configuration don't work because this listener transformes 50 & 5 on string value, oracle accept only int value.

src/Event/Listeners/OracleSessionInit.php
OlivierToussaint: I explain my case in comment you prefert the function is_int() ?

```
 if ($option === 'CURRENT_SCHEMA' || is_int($value)) {
     $vars[] = $option . ' = ' . $value;
 } else {
     $vars[] = $option . " = '" . $value . "'";
 }
```

morozov:
> shouldn't it be labeled with "Bug" rather than with "Improvement"

A bug is something we designed for and thought it would but it doesn't In this case, nobody thought about this case previously, so it's an improvement.

I don't think it's critical to release it in `2.10.x`. It's a session listener that can be copy/pasted by a developer until the new version is released.

> Also, didn't we make 2.10.x the default branch at some point? I thought I did. IMO it should be 2.11.x now

There was a conversion but it wasn't actioned upon. What's the definition of the default branch? Is it for bug fixes? New features? Breaking changes? Could it be defined on a PR template basis?

Definitely, `master` is the worst option for the default one but I'm not sure what's the best.

src/Event/Listeners/OracleSessionInit.php
morozov: The case makes sense. The above does as well.

greg0ire:
> Could it be defined on a PR template basis?

That would be best, awaiting that, looks like we are defaulting to the branch for new features right now. I think that's because hopefully (:crossed_fingers: ), there are not that many bugfixes, and hopefully too, not that many BC-breaks, which means that's the branch that will generated the fewer support for retargets.

Weirdly though, I can't find any example to back up my claim that we are using the branch for new features as a default. I think I remember setting it myself on a repository, but it has probably been set back to master, not sure by who or by what.

morozov:
> I think I remember setting it myself on a repository, but it has probably been set back to master, not sure by who or by what.

Please feel free to set it to whatever you think is appropriate. `2.11.x` should be fine. FWIW, I believe `2.11.0` and `3.0.0` will be released at about the same time, so I'd prefer all new features (not this minor improvement) to go there to reduce the number of up-merges.

greg0ire:
> Please feel free to set it to whatever you think is appropriate. 2.11.x should be fine.

Done

etiennerousseau:
The same problem occures with boolean arguments like GLOBAL_NAMES.

morozov:
> The same problem occurs with boolean arguments like GLOBAL_NAMES.

Makes sense. It may help if we work off the [spec](https://docs.oracle.com/cd/B19306_01/server.102/b14200/statements_2012.htm#i2217986) here and cover all possible cases at once instead of adding `if`-statements one by one:
1. `ASM_POWER_LIMIT` – Integer;
2. `COMMIT_WRITE` – String (a set of keywords, e.g. `IMMEDIATE`);
3. `DB_CREATE_FILE_DEST` – String (enclosed in quotes, e.g. `'/tmp/foo/bar'`, not a keyword);
4. `GLOBAL_NAMES` – Boolean.

There's obviously no way to distinguish between 2 and 3 without listing all supported options or introducing different APIs for defining string literals. Supporting integers, booleans and currently supported string literals is possible though.

morozov:
Closing as stale.

PR #3943, opened 2020-04-11
Do not require hostname for non-persistent MySQL connection and require for persistent

### LINKED ISSUE #3942: MySQLi connection fails or triggers a notice if the hostname is not specified
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.0.0-dev

#### Summary

When attempting to establish a non-persistent connection without a hostname:
```php
echo (new MysqliConnection([], 'root', ''))
    ->query('SELECT VERSION()')
    ->fetch()[0], PHP_EOL;

// Notice: Undefined index: host in src/Driver/Mysqli/MysqliConnection.php on line 52
// 5.7.27
```
When attempting to establish a persistent connection without a hostname:
```php
new MysqliConnection(['persistent' => true], 'root', '');

// Notice: Undefined index: host in src/Driver/Mysqli/MysqliConnection.php on line 52
// MysqliException: php_network_getaddresses: getaddrinfo failed: Temporary failure in name resolution in src/Driver/Mysqli/MysqliConnection.php on line 72
```

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

Fixes #3942.

According to the [documentation](https://www.php.net/mysqli_real_connect), the hostname may be omitted and will default to "localhost". However, given the fact that the persistent connection flag is passed as part of the hostname, the hostname cannot be omitted for persistent connection.

The issue is not reproducible on 2.x because the "Undefined index" notice is suppressed by the error handler:
https://github.com/doctrine/dbal/blob/35beaca3bed81ad4983e58706ae38688de34a853/lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php#L67-L70

### DIFF
--- a/src/Driver/Mysqli/HostRequired.php
+++ b/src/Driver/Mysqli/HostRequired.php
@@ -1,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Driver\Mysqli;
+
+/**
+ * @internal
+ */
+final class HostRequired extends MysqliException
+{
+    public static function forPersistentConnection() : self
+    {
+        return new self('The "host" parameter is required for a persistent connection');
+    }
+}
--- a/src/Driver/Mysqli/MysqliConnection.php
+++ b/src/Driver/Mysqli/MysqliConnection.php
@@ -49,11 +49,16 @@ class MysqliConnection implements PingableConnection, ServerInfoAwareConnection
     {
         $socket = $params['unix_socket'] ?? ini_get('mysqli.default_socket');
         $dbname = $params['dbname'] ?? null;
-        $host   = $params['host'];
         $port   = $params['port'] ?? null;
 
         if (! empty($params['persistent'])) {
-            $host = 'p:' . $host;
+            if (! isset($params['host'])) {
+                throw HostRequired::forPersistentConnection();
+            }
+
+            $host = 'p:' . $params['host'];
+        } else {
+            $host = $params['host'] ?? null;
         }
 
         $flags = $driverOptions[static::OPTION_FLAGS] ?? null;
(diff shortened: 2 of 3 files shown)


src/Driver/Mysqli/HostnameRequired.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Driver\Mysqli;
+
+final class HostnameRequired extends MysqliException
+{
+    public static function fromPersistentConnection() : self
```
greg0ire, line 9: With that naming, I would expect a connection argument. How about this?
```suggestion
    public static function forPersistentConnection() : self
```

src/Driver/Mysqli/HostnameRequired.php
```diff
@@ -0,0 +1,13 @@
...
+    {
+        return new self('Hostname is required for persistent connection');
```
greg0ire, line 11: I'd go with either
```suggestion
        return new self('The host parameter is required for persistent connections');
```

or

```suggestion
        return new self('host parameter required for persistent connection');
```

src/Driver/Mysqli/HostnameRequired.php
morozov: Agree. Let's do that.

src/Driver/Mysqli/HostnameRequired.php
morozov: Not sure about the articles but I'd keep the "connection" singular since the exception is about a single connection.

src/Driver/Mysqli/HostRequired.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Driver\Mysqli;
+
+/**
+ * @internal
```
morozov, line 8: Also, note that I marked it as `@internal` (there's a plan to do the same for the rest of the named exceptions in newer versions). This way, we can rework them at any time w/o breaking other's code who should rely only on higher-level classes and interfaces.

greg0ire (approved):

PR #3949, opened 2020-04-13
Make table and column comments non-nullable strings

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

1. The `$comment` arguments were marked as nullable during the scalar types enforcement in the codebase (https://github.com/doctrine/dbal/pull/3562, https://github.com/doctrine/dbal/pull/3511) because, despite the fact they were annotated as `string` in phpDoc, the values might contain NULL at runtime.
2. In practice a NULL comment and an empty comment are indistinguishable (there either is a comment or there's no).
3. Although the underlying platforms support nullable comments (except for SQLite), they are impossible to implement for commented types that store their type _on_ the comment.
4. The codebase currently contains expressions like `(string) $comment` or `if ($comment === null || $comment === '')` which illustrates that there's no practical difference between them at the abstraction level.
5. The removed functional tests were removed in favor of the newly introduced `ColumnCommentTest` that is cross-platform and more focused on the API rather than on implementation details.


tests/Functional/Schema/ColumnCommentTest.php
```diff
@@ -0,0 +1,129 @@
...
+     *
+     * @dataProvider columnProvider
+     */
+    public function testColumnComment(string $name, string $type, array $options) : void
+    {
+        $this->assertColumnComment('column_comments', $name, $options['comment'] ?? '');
+    }
+
+    /**
+     * @return mixed[][]
```
greg0ire, line 50: [Phpstan is ok with `mixed[][]`](https://phpstan.org/r/0c35f6eb-edf4-4cf9-84a0-0162d19105f8), but [Psalm isn't](https://psalm.dev/r/15456cc5ea).

```suggestion
     * @return iterable<mixed[]>
```

tests/Functional/Schema/ColumnCommentTest.php
```diff
@@ -0,0 +1,129 @@
...
+        self::assertNotNull($diff);
+
+        $sm = $this->connection->getSchemaManager();
+        $sm->alterTable($diff);
+
+        $this->assertColumnComment('column_comments', 'id', $comment2);
+    }
+
+    /**
+     * @return mixed[][]
```
greg0ire, line 106:
```suggestion
     * @return iterable<mixed[]>
```

tests/Functional/Schema/ColumnCommentTest.php
morozov: As far as I understand, `[]` is a list/collection. It doesn't imply a specific implementation. It looks like Psalm thinks differently (cannot find a reference).

I'm fine to adjust but I'd rather do it for the entire codebase and have the rule enforced on CI. E.g. by adding a dependency on Psalm and setting the lowest possible level. Otherwise, it will become a matter of personal preference than a standard.

greg0ire (approved):

tests/Functional/Schema/ColumnCommentTest.php
greg0ire: Ok let's not make this a blocker

PR #3950, opened 2020-04-14
Allow running `dbal:run-sql` using arbitrary connection

delolmo:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no

#### Summary

This slight modification allows running the `dbal:run-sql` command using an arbitrary database connection, which is specially useful in CLI applications that use multiple database connections. Backwards compatibility is maintained by using the default 'db' value for the input option.

The command now includes a `--connection` option, which defaults to 'db'. If specified, a ConnectionHelper is searched for that matches the provided connection name.

### DIFF
--- a/src/Tools/Console/Command/RunSqlCommand.php
+++ b/src/Tools/Console/Command/RunSqlCommand.php
@@ -30,6 +30,7 @@ class RunSqlCommand extends Command
         ->setDescription('Executes arbitrary SQL directly from the command line.')
         ->setDefinition([
             new InputArgument('sql', InputArgument::REQUIRED, 'The SQL statement to execute.'),
+            new InputOption('connection', null, InputOption::VALUE_REQUIRED, 'The database connection on which to execute the SQL statement', 'db'),
             new InputOption('depth', null, InputOption::VALUE_REQUIRED, 'Dumping depth of result set.', 7),
             new InputOption('force-fetch', null, InputOption::VALUE_NONE, 'Forces fetching the result.'),
         ])
@@ -44,7 +45,9 @@ EOT
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $conn = $this->getHelper('db')->getConnection();
+        $connection = $this->getOption('connection');
+        
+        $conn = $this->getHelper($connection)->getConnection();
 
         $sql = $input->getArgument('sql');
 
--- a/src/Tools/Console/Command/RunSqlCommand.php
+++ b/src/Tools/Console/Command/RunSqlCommand.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\DBAL\Tools\Console\Command;
 
+use Doctrine\DBAL\Tools\Console\Helper\ConnectionHelper;
 use Doctrine\DBAL\Tools\Dumper;
 use LogicException;
 use RuntimeException;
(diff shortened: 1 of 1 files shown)


SenseException:
> If specified, a ConnectionHelper is searched for that matches the provided connection name.

How do you add a new `ConnectionHelper` with a new connection in your use case? Can you please provide example code? This might be relevant because adding a new helper for another connection should become part of the documentation in scope of this PR if this gets approved.

morozov:
There's another similar pull request #3956.

delolmo:
> How do you add a new `ConnectionHelper` with a new connection in your use case? Can you please provide example code? This might be relevant because adding a new helper for another connection should become part of the documentation in scope of this PR if this gets approved.

When defining a generic `\Symfony\Component\Console\Application`, a HelperSet may be passed through the `setHelperSet` method. In `doctrine/migrations`, e.g., a `cli-config.php` script is created returning a specific HelperSet which can be modified to include several ConnectionHelper's:

```php
<?php

require 'vendor/autoload.php';

use Doctrine\DBAL\Tools\Console\Helper\ConnectionHelper;
use Symfony\Component\Console\Helper\HelperSet;

$dbParams = include 'migrations-db.php';

$connection = DriverManager::getConnection($dbParams);

return new HelperSet([
    'db' => new ConnectionHelper($connection),
    'db.read' => new ConnectionHelper($connection), // for example
    'db.write' => new ConnectionHelper($connection), // for example
]);
```
For documentation purposes, I was thinking about the following example:

```php
<?php

use Doctrine\DBAL\Tools\Console\Helper\ConnectionHelper;
use Symfony\Component\Console\Application;
use Symfony\Component\Console\Helper\HelperSet;

$dbParams = [...]; // A multidimensional array with connection parameters

$cli = new Application('app name', 'app version');

$cli->setHelperSet(new HelperSet([
    'db.read' => new ConnectionHelper(
        DriverManager::getConnection($dbParams['db.read'])
    ),
    'db.write' => new ConnectionHelper(
        DriverManager::getConnection(($dbParams['db.write'])
    ),
    'pgsql' => new ConnectionHelper(
        DriverManager::getConnection($dbParams['db.pgsql'])
    ),
    'mysql' => new ConnectionHelper(
        DriverManager::getConnection(($dbParams['db.mysql'])
    ),
    // and so on...
]));

$cli->addCommand(RunSqlCommand());

// Continue configuring your cli app, add commands, etc.
```

Regarding #3956, IMO its too much code to solve a very specific use case. Maybe, a new ConnectionProviderHelperSet could be created, but still

delolmo:
BTW I think I mentioned `doctrine/migrations`, but `doctrine/dbal` and `doctrine/orm` use a similar approach to define database connections.

`doctrine/dbal` expects a HelperSet to be returned by the `cli-config.php` file, while `doctrine/orm` uses a [createHelperSet static method](https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/reference/configuration.html#setting-up-the-commandline-tool).

SenseException:
> There's another similar pull request #3956.

Yes, that would drop the need for multiple helpers and extending the helperset if this PR can build on top of the other one.

dmaicher:
Either we go with this approach (which would be backwards compatible without deprecations) or we move forward with https://github.com/doctrine/dbal/pull/3956

@morozov what do you think?

morozov:
@dmaicher, I'd proceed with https://github.com/doctrine/dbal/pull/3956.

morozov:
Closing in favor of #3956.

PR #3951, opened 2020-04-14
Setup static analysis with Psalm

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

This PR adds static analysis with Psalm, configured to the most lenient level, and ignores every error.
I am targeting 2.10.x, because this change does not touch code in `lib`, or dependencies under `require`.

### DIFF
--- a/baseline.xml
+++ b/baseline.xml
@@ -1,0 +1,216 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<files psalm-version="3.11.2@d470903722cfcbc1cd04744c5491d3e6d13ec3d9">
+  <file src="lib/Doctrine/DBAL/Driver/Connection.php">
+    <InvalidParamDefault occurrences="1">
+      <code>int</code>
+    </InvalidParamDefault>
+  </file>
+  <file src="lib/Doctrine/DBAL/Driver/DriverException.php">
+    <MissingImmutableAnnotation occurrences="1">
+      <code>public function getMessage();</code>
+    </MissingImmutableAnnotation>
+  </file>
+  <file src="lib/Doctrine/DBAL/Driver/PDOSqlsrv/Statement.php">
+    <UndefinedConstant occurrences="1">
+      <code>PDO::SQLSRV_ENCODING_BINARY</code>
+    </UndefinedConstant>
+  </file>
+  <file src="lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereConnection.php">
+    <UndefinedFunction occurrences="14">
+      <code>sasql_pconnect($dsn)</code>
+      <code>sasql_connect($dsn)</code>
+      <code>sasql_set_option($this-&gt;connection, 'verbose_errors', false)</code>
+      <code>sasql_set_option($this-&gt;connection, 'auto_commit', 'on')</code>
+      <code>sasql_set_option($this-&gt;connection, 'auto_commit', 'off')</code>
+      <code>sasql_commit($this-&gt;connection)</code>
+      <code>sasql_errorcode($this-&gt;connection)</code>
+      <code>sasql_error($this-&gt;connection)</code>
+      <code>sasql_real_query($this-&gt;connection, $statement)</code>
+      <code>sasql_affected_rows($this-&gt;connection)</code>
+      <code>sasql_insert_id($this-&gt;connection)</code>
+      <code>sasql_escape_string($this-&gt;connection, $input)</code>
+      <code>sasql_rollback($this-&gt;connection)</code>
+      <code>sasql_set_option($this-&gt;connection, 'auto_commit', 'on')</code>
+    </UndefinedFunction>
+  </file>
+  <file src="lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereException.php">
+    <UndefinedFunction occurrences="8">
+      <code>sasql_sqlstate($conn)</code>
+      <code>sasql_sqlstate()</code>
+      <code>sasql_stmt_errno($stmt)</code>
+      <code>sasql_stmt_error($stmt)</code>
+      <code>sasql_errorcode($conn)</code>
+      <code>sasql_error($conn)</code>
+      <code>sasql_errorcode()</code>
+      <code>sasql_error()</code>
+    </UndefinedFunction>
+  </file>
+  <file src="lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereStatement.php">
+    <UndefinedConstant occurrences="1">
+      <code>SASQL_BOTH</code>
+    </UndefinedConstant>
+    <UndefinedFunction occurrences="14">
+      <code>sasql_prepare($conn, $sql)</code>
+      <code>sasql_stmt_bind_param_ex($this-&gt;stmt, $column - 1, $variable, $type, $variable === null)</code>
+      <code>sasql_stmt_reset($this-&gt;stmt)</code>
+      <code>sasql_stmt_field_count($this-&gt;stmt)</code>
+      <code>sasql_stmt_errno($this-&gt;stmt)</code>
+      <code>sasql_stmt_error($this-&gt;stmt)</code>
+      <code>sasql_stmt_execute($this-&gt;stmt)</code>
+      <code>sasql_stmt_result_metadata($this-&gt;stmt)</code>
+      <code>sasql_fetch_assoc($this-&gt;result)</code>
+      <code>sasql_fetch_array($this-&gt;result, SASQL_BOTH)</code>
+      <code>sasql_fetch_object($this-&gt;result)</code>
+      <code>sasql_fetch_row($this-&gt;result)</code>
+      <code>sasql_fetch_object($this-&gt;result)</code>
+      <code>sasql_stmt_affected_rows($this-&gt;stmt)</code>
+    </UndefinedFunction>
+  </file>
+  <file src="lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php">
+    <TooFewArguments occurrences="1">
+      <code>sqlsrv_get_field($this-&gt;stmt, 0)</code>
+    </TooFewArguments>
+  </file>
+  <file src="lib/Doctrine/DBAL/Platforms/AbstractPlatform.php">
+    <InvalidParamDefault occurrences="1">
+      <code>int</code>
+    </InvalidParamDefault>
+  </file>
+  <file src="lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php">
+    <InvalidParamDefault occurrences="1">
+      <code>string</code>
+    </InvalidParamDefault>
+  </file>
+  <file src="lib/Doctrine/DBAL/Tools/Dumper.php">
+    <UndefinedClass occurrences="1">
+      <code>Collection</code>
+    </UndefinedClass>
+  </file>
+  <file src="tests/Doctrine/Tests/DBAL/Driver/PDOPgSql/DriverTest.php">
+    <UndefinedConstant occurrences="5">
+      <code>PDO::PGSQL_ATTR_DISABLE_PREPARES</code>
+      <code>PDO::PGSQL_ATTR_DISABLE_PREPARES</code>
+      <code>PDO::PGSQL_ATTR_DISABLE_PREPARES</code>
+      <code>PDO::PGSQL_ATTR_DISABLE_PREPARES</code>
+      <code>PDO::PGSQL_ATTR_DISABLE_PREPARES</code>
+    </UndefinedConstant>
+  </file>
+  <file src="tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php">
+    <UndefinedDocblockClass occurrences="93">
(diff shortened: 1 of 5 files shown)


greg0ire:
I think it will not show up until merged, but the result can already be preview on [this PR I made in my fork](https://github.com/greg0ire/dbal/pull/1)

greg0ire:
Travis fails because the version of Psalm I installed is too recent. I will add something in `composer.json` to make sure we install something supported on 7.2

composer.json
```diff
@@ -42,14 +42,18 @@
...
     },
     "suggest": {
         "symfony/console": "For helpful console commands such as SQL execution and import of files."
     },
     "bin": ["bin/doctrine-dbal"],
     "config": {
-        "sort-packages": true
+        "sort-packages": true,
+        "platform": {
+            "php": "7.2.15"
```
morozov, line 55: Should this be set to `7.2.0` (see https://github.com/doctrine/dbal/pull/3862)? Otherwise, we may end up updating the dependency constraints that do not support the lowest supported PHP version (which is 7.2.0).

composer.lock
```diff
@@ -877,20 +1358,6 @@
                 "MIT"
             ],
             "description": "PHPStan - PHP Static Analysis Tool",
-            "funding": [
```
morozov, line 880: You're probably using an older version of Composer that doesn't support this type of metadata. Please make sure it's not removed by mistake.

baseline.xml
```diff
@@ -0,0 +1,253 @@
...
+    </InvalidParamDefault>
+  </file>
+  <file src="lib/Doctrine/DBAL/Driver/DriverException.php">
+    <MissingImmutableAnnotation occurrences="1">
+      <code>public function getMessage();</code>
+    </MissingImmutableAnnotation>
+  </file>
+  <file src="lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php">
+    <UndefinedConstant occurrences="3">
+      <code>DB2_AUTOCOMMIT_OFF</code>
```
morozov, line 15: Please see if this can be avoided by using the stubs:

https://github.com/doctrine/dbal/blob/6c789b30a46906c5e669158809076d2405ea4fa9/tests/phpstan-polyfill.php#L6-L12

morozov (commented):
I'm surprised that we have so few errors to whitelist of which most are related to the DB extensions. BTW, why doesn't it report the existing `mixed[][]` return types?

composer.json
greg0ire: I originally used 7.2.0, and that did not work out because `symfony/console` requires 7.2.5, so I resolved to use whatever is used in the 7.2 Travis image we run the tests on.

composer.lock
greg0ire: Thanks for the tip, I'll try again.

baseline.xml
greg0ire: Oh great, will try.

composer.json
morozov: It means that technically the moment we bumped the `symfony/console` constraint, we dropped the support for the older PHP versions. If it's only the locked version that requires PHP 7.2.5, then I suggest we lock it to an older one that doesn't require PHP 7.2.5.

greg0ire:
> I'm surprised that we have so few errors to whitelist of which most are related to the DB extensions. BTW, why doesn't it report the existing mixed[][] return types?

That's probably because I went for the most lenient level (`8`)

composer.json
greg0ire: Here is the message I obtained. I will try again, and will try to downgrade just the minimal amount of package to achieve this.
```
composer update nothing
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - symfony/console v5.0.2 requires php ^7.2.5 -> your PHP version (7.3.15) overridden by "config.platform.php" version (7.2.0) does not satisfy that requirement.
    - symfony/console v5.0.2 requires php ^7.2.5 -> your PHP version (7.3.15) overridden by "config.platform.php" version (7.2.0) does not satisfy that requirement.
    - symfony/console v5.0.2 requires php ^7.2.5 -> your PHP version (7.3.15) overridden by "config.platform.php" version (7.2.0) does not satisfy that requirement.
    - Installation request for symfony/console (locked at v5.0.2, required as ^2.0.5|^3.0|^4.0|^5.0) -> satisfiable by symfony/console[v5.0.2].
```

greg0ire: Ok it was just 2 sf packages, will push shortly.

baseline.xml
greg0ire: Using that file with `autoloader=` gives me the following error:  `PHP Fatal error:  Class mysqli_result must implement interface Traversable as part of either Iterator or IteratorAggregate in Unknown on line 0`. I will try to do something with the `<stubs>` tag

morozov: Alternatively, you may try to include the extensions that dont' have proprietary build dependencies to the docker image that runs Psalm so that they don't need to be stubbed (e.g. `pdo-*`, `mysqli`).

greg0ire: Ok it's fixed now :)

composer.lock
greg0ire: I updated, and it's still removed, but if you look at the diff, it is also added in the `xdebug-handler` package metadata above.

morozov: Yeah, this is because Psalm requires it (unlike PHPUnit!). I'm fine with the changes in the lock file, just wanted to make sure it's not a mistake.

morozov:
> or because the tests directory is not included :P

I see. Is it because we also don't run PHPStan against tests? Do you want to add it in `master`? Mind that some will have to up-merge the new dependency and update the baseline. And that someone will be you @greg0ire :-)

composer.json
morozov: The solution looks good.

greg0ire:
It was both reasons: the issues is reported starting at level 6.

I can add psalm to all branches when doing the upmerges, But I'm not sure if that's what you meant, I'm under the impression that you would like me to add it in master and then do the upmerge, is that correct? And if yes, why?

morozov:
No, once this PR gets merged, we'll need to merge `2.10.x` → `2.11.x` → `3.0.x` → `master` to get the changes propagated. And given that each of the branches is different in the underlying code, each up-merge PR may need to adjust the baseline (e.g. change the paths, include new exceptions or remove the irrelevant ones).

greg0ire:
Ok, I'll take care of it :)

morozov (approved):

morozov:
Great work by the way, @greg0ire! I'm glad that we eventually have Psalm in the build pipeline.

greg0ire:
Thanks me too :)

ondrejmirtes:
@morozov Why is baseline fine for you with Psalm, but not with PHPStan? Linking our discussion here: https://github.com/doctrine/dbal/pull/3799#discussion_r362936413

greg0ire:
@ondrejmirtes we do have a baseline for PHPStan: https://github.com/doctrine/dbal/blob/72d9081c8629cc0d07cb0469f959ced2340ca40c/phpstan.neon.dist#L11-L74

ondrejmirtes:
This is not a baseline in the sense we talked about in the linked discussion :) PHPStan's baseline is generated with `--generate-baseline`: https://phpstan.org/user-guide/baseline

greg0ire:
TIL :)

morozov:
> Why is baseline fine for you with Psalm, but not with PHPStan? Linking our discussion here: [#3799 (comment)](https://github.com/doctrine/dbal/pull/3799#discussion_r362936413)

This is a valid question. I didn't think of it because we were using the lowest possible level and the baseline was way smaller than in #3799. Although, it is getting out of hand (I don't remember seeing `occurrences="93"` in the very first patch).

@greg0ire before we proceed with merging it to master, let's take a pause and see if the baselined errors could be fixed (3.0 allows for BC breaks) and the non-fixable ones are suppressed manually.

greg0ire:
@morozov I don't know if you noticed but the baseline is way smaller for master: https://github.com/doctrine/dbal/pull/3954/files#diff-043e2f3614861e20eb0fb277052a7693

greg0ire:
The first error I noticed does not imply a BC-break, and gets us rid of ~ 100 errors, so I'm going to have a look at 2.10.x first

morozov:
I noticed that but I didn't think of them as the finally required baselines. It's a good sign, on the one hand, meaning that we did a good job there. But on the other hand, we still need to get rid of those many repetitive occurrences of the same error. Otherwise, each time we modify something, we'll hit the same error again.

greg0ire:
Ok so now I'm getting a crash on 2.10.x that I don't get with the deps of 3.0.x… :man_facepalming: 

```
Uncaught TypeError: Return value of PhpParser\Comment::getFilePos() must be of the type int, null returned in /home/greg/dev/doctrine-dbal/vendor/nikic/php-parser/lib/PhpParser/Comment.php:53
Stack trace:
#0 /home/greg/dev/doctrine-dbal/vendor/vimeo/psalm/src/Psalm/Internal/Diff/ClassStatementsDiffer.php(78): PhpParser\Comment->getFilePos()
#1 /home/greg/dev/doctrine-dbal/vendor/vimeo/psalm/src/Psalm/Internal/Diff/AstDiffer.php(53): Psalm\Internal\Diff\ClassStatementsDiffer::Psalm\Internal\Diff\{closure}(Object(PhpParser\Node\Stmt\ClassMethod), Object(PhpParser\Node\Stmt\ClassMethod), '<?php\n\nnamespac...', '<?php\n\nnamespac...', false)
#2 /home/greg/dev/doctrine-dbal/vendor/vimeo/psalm/src/Psalm/Internal/Diff/ClassStatementsDiffer.php(45): Psalm\Internal\Diff\AstDiffer::calculateTrace(Object(Closure), Array, Array, '<?php\n\nnamespac...', '<?php\n\nnamespac...')
#3 /home/greg/dev/doctrine-dbal/vendor/vimeo/psalm/src/Psalm/Internal/Diff/NamespaceStatementsDiffer.php(98): Psalm\Internal\Diff\ClassStatementsDiffer::diff('Doctrine\\DBAL\\S...', Array, Array, '<?php\n\nnamespac...', '<?php\n\nnamespac...')
#4 /home/greg/dev/doctrine-dbal/vendor/vimeo/psalm/src/Psalm/Internal/Diff/FileStatementsDiffer.php(92): Psalm\Internal\Diff\NamespaceStatementsDiffer::diff('Doctrine\\DBAL\\S...', Array, Array, '<?php\n\nnamespac...', '<?php\n\nnamespac...')
#5 /home/greg/dev/doctrine-dbal/vendor/vimeo/psalm/src/Psalm/Internal/Provider/StatementsProvider.php(179): Psalm\Internal\Diff\FileStatementsDiffer::diff(Array, Array, '<?php\n\nnamespac...', '<?php\n\nnamespac...')
#6 /home/greg/dev/doctrine-dbal/vendor/vimeo/psalm/src/Psalm/Internal/Scanner/FileScanner.php(68): Psalm\Internal\Provider\StatementsProvider->getStatementsForFile('/home/greg/dev/...', Object(Psalm\Progress\DefaultProgress))
#7 /home/greg/dev/doctrine-dbal/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Scanner.php(586): Psalm\Internal\Scanner\FileScanner->scan(Object(Psalm\Codebase), Object(Psalm\Storage\FileStorage), false, Object(Psalm\Progress\DefaultProgress))
#8 /home/greg/dev/doctrine-dbal/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Scanner.php(367): Psalm\Internal\Codebase\Scanner->scanFile('/home/greg/dev/...', Array, true)
#9 /home/greg/dev/doctrine-dbal/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Scanner.php(475): Psalm\Internal\Codebase\Scanner->Psalm\Internal\Codebase\{closure}(0, '/home/greg/dev/...')
#10 /home/greg/dev/doctrine-dbal/vendor/vimeo/psalm/src/Psalm/Internal/Codebase/Scanner.php(325): Psalm\Internal\Codebase\Scanner->scanFilePaths(1)
#11 /home/greg/dev/doctrine-dbal/vendor/vimeo/psalm/src/Psalm/Codebase.php(477): Psalm\Internal\Codebase\Scanner->scanFiles(Object(Psalm\Internal\Codebase\ClassLikes), 3)
#12 /home/greg/dev/doctrine-dbal/vendor/vimeo/psalm/src/Psalm/Internal/Analyzer/ProjectAnalyzer.php(519): Psalm\Codebase->scanFiles(3)
#13 /home/greg/dev/doctrine-dbal/vendor/vimeo/psalm/src/psalm.php(588): Psalm\Internal\Analyzer\ProjectAnalyzer->check('/home/greg/dev/...', false)
#14 /home/greg/dev/doctrine-dbal/vendor/vimeo/psalm/psalm(2): require_once('/home/greg/dev/...')
#15 {main}
(Psalm 3.11.2@d470903722cfcbc1cd04744c5491d3e6d13ec3d9 crashed due to an uncaught Throwable)
```

PR #3954, opened 2020-04-14

greg0ire:
No conflicts besides `composer.lock` baseline changed.


greg0ire:
@morozov I need help regarding the 2 last error I get. I don't know how I should fix them:

```
ERROR: ImpureMethodCall - src/Exception/DriverException.php:24:57 - Cannot call an possibly-mutating method Doctrine\DBAL\Driver\DriverException::getCode from a mutation-free context (see https://psalm.dev/203)
        parent::__construct($message, $driverException->getCode(), $driverException);


ERROR: ImpureMethodCall - src/Exception/DriverException.php:29:28 - Cannot call an possibly-mutating method Doctrine\DBAL\Exception\DriverException::getPrevious from a mutation-free context (see https://psalm.dev/203)
        $previous = $this->getPrevious();
```

morozov:
@greg0ire it looks like the problem is that `Doctrine\DBAL\Driver\DriverException::getCode()` is inherited from `Throwable` which is unknown to Psalm. For instance, declaring the method explicitly fixes one issue:
```php
interface DriverException extends Throwable
{
    /* ... */
    public function getCode();
}
```
Although, I'm not sure if it's the right solution since Psalm itself should be aware that the built-in `Exception` is immutable.

It's fine if we suppress them for now since we're just setting up the analysis.

greg0ire:
:thinking: I don't remember why psalm asked me to mark all exceptions as immutable, but when I remove this annotation everywhere, I no longer get the error… this is very confusing.

greg0ire:
Looks like this commit I made is at the origin of the chain reaction: 5a61341fea474990fe2c4fb5e27515c053ccf4a4

greg0ire:
Ok so it all stems from this message I got on 2.10.x, and that I fixed too hastily:

```
ERROR: MissingImmutableAnnotation - lib/Doctrine/DBAL/Driver/DriverException.php:30:5 - Throwable::getMessage is marked immutable, but Doctrine\DBAL\Driver\DriverException::getMessage is not marked immutable (see https://psalm.dev/213)
    /**
     * Returns the driver error message.
     *
     * @return string
     */
    public function getMessage();
```

I'm going to revert that commit, and just mark the method immutable, not the whole class.

psalm.xml
```diff
@@ -0,0 +1,66 @@
...
+    <issueHandlers>
+        <ImpureMethodCall>
+            <errorLevel type="suppress">
+                <!--
+                    Requires a release of
+                    https://github.com/vimeo/psalm/pull/3171
+                -->
+                <file name="src/Exception/DriverException.php"/>
+            </errorLevel>
+        </ImpureMethodCall>
```
greg0ire, line 34: @morozov I added this

greg0ire:
@morozov this one is ready too

morozov (approved):

PR #3955, opened 2020-04-15
Remove baseline

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n / a


greg0ire:
@morozov not sure why this is ignored:

https://github.com/doctrine/dbal/blob/72d9081c8629cc0d07cb0469f959ced2340ca40c/phpstan.neon.dist#L48-L49

AFAIK, Collection is still using the old `Common` namespace, and we depend on that lib. IMO we should add `doctrine/collection` to at least `require-dev` (in the case that dependency is considered "optional"), and remove that exclusion rule from the phpstan config file. WDYT?

greg0ire:
Also, I can't seem to be able to run phpstan locally:
```
vendor/bin/phpstan analyze
Note: Using configuration file /home/greg/dev/doctrine-dbal/phpstan.neon.dist.
PHP Fatal error:  Class mysqli_result must implement interface Traversable as part of either Iterator or IteratorAggregate in Unknown on line 0
Fatal error: Class mysqli_result must implement interface Traversable as part of either Iterator or IteratorAggregate in Unknown on line 0
```

greg0ire:
@morozov I think I fixed everything I could. Please tell me if you would like some commits to move to a PR targeting 2.11.x

morozov:
@greg0ire judging by the if-statement, it's an optional dependency. So instead of a dev-dependency, we could additionally install it for static analysis if we don't want to ignore those errors.

greg0ire:
> So instead of a dev-dependency, we could additionally install it for static analysis if we don't want to ignore those errors.

Installing it as a dev-dependency would make Psalm happy both for a Doctrine dev and for our CI, I'm not following you :thinking:

morozov:
Would it be possible to convert specific baseline errors into more generic ones that wouldn't require future maintenance? E.g. we don't care of all missing `sasql_*` functions and constants regardless of their name and the location. The same is for `PDO::PGSQL_ATTR_DISABLE_PREPARES`.

Ideally, I'd like to get rid of the baseline entirely and turn it into configuration like we do with PHPStan.

greg0ire:
Also, I was expecting phpstan to fail because of useless ignore rules, but it doesn't :thinking:

morozov:
This is because those are disabled (`reportUnmatchedIgnoredErrors: false`). They may or may not be useless depending on the environment (e.g. the missing `PDO::PGSQL_ATTR_DISABLE_PREPARES` constant) and I don't see another way to fix that.

greg0ire:
> Would it be possible to convert specific baseline errors into more generic ones that wouldn't require future maintenance? E.g. we don't care of all missing `sasql_*` functions and constants regardless of their name and the location. The same is for `PDO::PGSQL_ATTR_DISABLE_PREPARES`.
> 
> Ideally, I'd like to get rid of the baseline entirely and turn it into configuration like we do with PHPStan.

A quick look at the docs does not yield much…

morozov:
> Installing it as a dev-dependency would make Psalm happy both for a Doctrine dev and for our CI, I'm not following you

I just don't want to abuse dev-dependencies. This library is not a requirement for development. I'd rather ignore those errors than add the dependency.

greg0ire:
> I just don't want to abuse dev-dependencies. This library is not a requirement for development. I'd rather ignore those errors than add the dependency.

Ok, I will drop both commits.

morozov:
> A quick look at the docs does not yield much…

E.g. we can suppress all `UndefinedFuncion` errors in `lib/Doctrine/DBAL/Driver/SQLAnywhere/`. It's unlikely that we will introduce some other missing functions there than the ones defined by the extension. We may use a similar approach for the rest.

greg0ire:
I agree but I can't find a way to do this here: https://psalm.dev/docs/running_psalm/configuration

greg0ire:
For some reason adding `PDO` to the phpstan stubs did not help. I will revert that.

morozov:
I think it's something like:
```xml
<psalm>
    <issueHandlers>
        <UndefinedFuncion>
            <errorLevel type="suppress">
                <directory name="lib/Doctrine/DBAL/Driver/SQLAnywhere"/>
            </errorLevel>
        </UndefinedFuncion>
    </issueHandlers>
</psalm>
```

greg0ire:
Oh right, I didn't use the right keywords in my research (ignore vs suppress)

greg0ire:
> Ideally, I'd like to get rid of the baseline entirely and turn it into configuration like we do with PHPStan.

> We may use a similar approach for the rest.

Done, I ignored things related to SQLAnywhere.

morozov:
I think we should get rid of the baseline file as such and move all the suppressions in the config. The entries about `PDO::PGSQL_ATTR_DISABLE_PREPARES` will need to be reworked in a similar way. As for the remaining ones, it'd be nice to have a comment for each on why it's being left suppressed (e.g. by design or would require a BC break or a bug in Psalm, etc.). This way we know what we can improve next.

greg0ire:
Ok, I'll do that.

composer.lock
```diff
@@ -816,16 +816,16 @@
         },
         {
             "name": "nikic/php-parser",
-            "version": "v4.3.0",
+            "version": "v4.4.0",
```
morozov, line 819: We do test with the lowest dependencies on Travis, so I'm surprised why just fixing the lock file helps. The build would fail when I tried to update PHPUnit to `^9.1` because of a bug in PHPUnit 9.1.0. Requiring `^9.1.1` solved the problem.

Depending on the exact nature of the issue, should we introduce a constraint or we bump/adjust some other constraint that would eliminate the conflict?

greg0ire: > We do test with the lowest dependencies on Travis, so I'm surprised why just fixing the lock file helps.

This helps fix a crash I had when running Psalm, and we don't run Psalm after running `composer --prefer-lowest`

I think we should not use a conflict when it's something that fails in dev/CI, I can add a `require-dev` constraint if you want me to

morozov: Agree. Let's do a temporary `require-dev`. Additionally, is it worth reporting to Psalm so that they could fix it on their end and we could remove this later?

psalm.xml
```diff
@@ -16,10 +15,48 @@
...
         <file name="vendor/jetbrains/phpstorm-stubs/oci8/oci8.php" />
         <file name="vendor/jetbrains/phpstorm-stubs/pgsql/pgsql.php" />
         <file name="vendor/jetbrains/phpstorm-stubs/sqlsrv/sqlsrv.php" />
     </stubs>
+    <issueHandlers>
+        <InvalidParamDefault>
+            <errorLevel type="suppress">
+                <file name="lib/Doctrine/DBAL/Driver/Connection.php" />
+            </errorLevel>
+        </InvalidParamDefault>
```
greg0ire, line 30: @morozov I don't understand this one… do you?

composer.lock
greg0ire: Requirement added, bug reported here: https://github.com/vimeo/psalm/issues/3155

psalm.xml
morozov: I think it's about the implementors:

https://github.com/doctrine/dbal/blob/35beaca3bed81ad4983e58706ae38688de34a853/lib/Doctrine/DBAL/Connection.php#L823-L826

greg0ire:
```
ERROR: InvalidParamDefault - lib/Doctrine/DBAL/Driver/Connection.php:35:15 - Default value type null for argument 2 of method Doctrine\DBAL\Connection::quote does not match the given type int (see https://psalm.dev/062)
     * @param int   $type
```
But the default parameter is supposed to be an int, not null…

psalm.xml
```diff
@@ -16,10 +15,48 @@
...
+        <TooFewArguments>
+            <errorLevel type="suppress">
+                <!--
+                    Requires a release of
+                    https://github.com/JetBrains/phpstorm-stubs/commit/300ee558a0fc33491a16a32346886ecd8b6726b5
```
morozov, line 35: Let's better reference a pull request (https://github.com/JetBrains/phpstorm-stubs/pull/727) just in case they decide to rebase the branch.

psalm.xml
```diff
@@ -16,10 +15,48 @@
...
+                -->
+                <file name="lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php"/>
+            </errorLevel>
+        </TooFewArguments>
+        <UndefinedConstant>
+            <errorLevel type="suppress">
+                <directory name="lib/Doctrine/DBAL/Driver/SQLAnywhere"/>
+                <!--
+                    Requires a release of
+                    https://github.com/JetBrains/phpstorm-stubs/commit/4dcb1d13a2cca98f0f45eea62b5cff37ad55ea41
```
morozov, line 45: And a reference to the PR (https://github.com/JetBrains/phpstorm-stubs/pull/732) here as well.

psalm.xml
greg0ire: Rebasing master? Who does that (besides Doctrine I mean :trollface: )?

psalm.xml
morozov: This is what the error is above. The default should be changed to `ParameterType::STRING`. I don't see it being a BC break.

psalm.xml
morozov: I should have used “human-readable” approach. Besides, a PR describes the issue better than just a commit.

psalm.xml
greg0ire: Alright let's try that!

greg0ire:
@morozov :ship: ?

morozov (approved):
@greg0ire, :ship:!

vladyslavstartsev:
@morozov I think 2 PRs from phpstorm-stubs that you've mentioned here are released (talking about this https://github.com/JetBrains/phpstorm-stubs/pull/732 and https://github.com/JetBrains/phpstorm-stubs/pull/727 )

morozov:
Thanks, @vladyslavstartsev. Will be addressed in https://github.com/doctrine/dbal/pull/4465.

PR #3956, opened 2020-04-15
allow using multiple connections for CLI commands

dmaicher:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | -

#### Summary

follow-up for https://github.com/doctrine/dbal/pull/3892

This
- adds a `ConnectionProvider` interface and a simple implementation for it
- uses this provider in a backwards-compatible way for the console commands
- deprecates the `ConnectionHelper` that was providing a single connection

### DIFF
--- a/bin/doctrine-dbal.php
+++ b/bin/doctrine-dbal.php
@@ -1,5 +1,6 @@
 <?php
 
+use Doctrine\DBAL\Tools\Console\ConnectionProvider;
 use Doctrine\DBAL\Tools\Console\ConsoleRunner;
 use Symfony\Component\Console\Helper\HelperSet;
 
@@ -41,17 +42,17 @@ if (! is_readable($configFile)) {
     exit(1);
 }
 
-$commands  = [];
-$helperSet = require $configFile;
+$commands                      = [];
+$helperSetOrConnectionProvider = require $configFile;
 
-if (! $helperSet instanceof HelperSet) {
-    foreach ($GLOBALS as $helperSetCandidate) {
-        if ($helperSetCandidate instanceof HelperSet) {
-            $helperSet = $helperSetCandidate;
+if (! $helperSetOrConnectionProvider instanceof HelperSet && ! $helperSetOrConnectionProvider instanceof ConnectionProvider) {
+    foreach ($GLOBALS as $candidate) {
+        if ($candidate instanceof HelperSet) {
+            $helperSetOrConnectionProvider = $candidate;
 
             break;
         }
     }
 }
 
-ConsoleRunner::run($helperSet, $commands);
+ConsoleRunner::run($helperSetOrConnectionProvider, $commands);
--- a/lib/Doctrine/DBAL/Tools/Console/ConsoleRunner.php
+++ b/lib/Doctrine/DBAL/Tools/Console/ConsoleRunner.php
@@ -11,6 +11,10 @@ use Doctrine\DBAL\Version;
 use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Helper\HelperSet;
+use TypeError;
+use function sprintf;
+use function trigger_error;
+use const E_USER_DEPRECATED;
 
 /**
  * Handles running the Console Tools inside Symfony Console context.
@@ -19,6 +23,8 @@ class ConsoleRunner
 {
     /**
      * Create a Symfony Console HelperSet
+     *
+     * @deprecated use a ConnectionProvider instead.
      *
      * @return HelperSet
      */
@@ -30,20 +36,31 @@ class ConsoleRunner
     }
 
     /**
-     * Runs console with the given helperset.
+     * Runs console with the given connection provider or helperset (deprecated).
      *
-     * @param Command[] $commands
+     * @param ConnectionProvider|HelperSet $helperSetOrConnectionProvider
+     * @param Command[]                    $commands
      *
      * @return void
      */
-    public static function run(HelperSet $helperSet, $commands = [])
+    public static function run($helperSetOrConnectionProvider, $commands = [])
     {
         $cli = new Application('Doctrine Command Line Interface', Version::VERSION);
 
         $cli->setCatchExceptions(true);
-        $cli->setHelperSet($helperSet);
 
-        self::addCommands($cli);
+        $connectionProvider = null;
+        if ($helperSetOrConnectionProvider instanceof HelperSet) {
+            @trigger_error(sprintf('Passing an instance of "%s" as the first argument is deprecated. Pass an instance of "%s" instead.', HelperSet::class, ConnectionProvider::class), E_USER_DEPRECATED);
+            $connectionProvider = null;
+            $cli->setHelperSet($helperSetOrConnectionProvider);
+        } elseif ($helperSetOrConnectionProvider instanceof ConnectionProvider) {
+            $connectionProvider = $helperSetOrConnectionProvider;
+        } else {
+            throw new TypeError(sprintf('First argument must be an instance of "%s" or "%s"', HelperSet::class, ConnectionProvider::class));
+        }
+
+        self::addCommands($cli, $connectionProvider);
 
         $cli->addCommands($commands);
         $cli->run();
@@ -52,12 +69,12 @@ class ConsoleRunner
     /**
      * @return void
      */
-    public static function addCommands(Application $cli)
+    public static function addCommands(Application $cli, ?ConnectionProvider $connectionProvider = null)
     {
         $cli->addCommands([
-            new RunSqlCommand(),
+            new RunSqlCommand($connectionProvider),
             new ImportCommand(),
-            new ReservedWordsCommand(),
+            new ReservedWordsCommand($connectionProvider),
         ]);
     }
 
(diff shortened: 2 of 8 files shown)


lib/Doctrine/DBAL/Tools/Console/Command/ConnectionProviderAwareCommand.php
```diff
@@ -0,0 +1,47 @@
...
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Tools\Console\ConnectionProvider\ConnectionProvider;
+use Exception;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use const E_USER_DEPRECATED;
+use function trigger_error;
+
+abstract class ConnectionProviderAwareCommand extends Command
```
dmaicher, line 13: or should we use a trait to share this between the commands?

lib/Doctrine/DBAL/Tools/Console/ConnectionProvider/ArrayConnectionProvider.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Doctrine\DBAL\Tools\Console\ConnectionProvider;
+
+use Doctrine\DBAL\Connection;
+use InvalidArgumentException;
+use function sprintf;
+
+class ArrayConnectionProvider implements ConnectionProvider
```
dmaicher, line 9: didn't come up with a better name yet... any suggestions?

lib/Doctrine/DBAL/Tools/Console/ConnectionProvider/ArrayConnectionProvider.php
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * @param array<string, Connection> $connections
+     */
+    public function __construct(array $connections, string $defaultConnectionName)
+    {
+        $this->connections           = $connections;
+        $this->defaultConnectionName = $defaultConnectionName;
+    }
+
+    public static function createFromSingleConnection(Connection $connection, string $name = 'default') : self
```
dmaicher, line 26: or should we add a separate class for it?

lib/Doctrine/DBAL/Tools/Console/ConnectionProvider/ArrayConnectionProvider.php
dmaicher: replaced it with a `SingleConnectionProvider` for now

lib/Doctrine/DBAL/Tools/Console/ConnectionProvider/ConnectionProvider.php
```diff
@@ -0,0 +1,16 @@
...
+
+use Doctrine\DBAL\Connection;
+use InvalidArgumentException;
+
+interface ConnectionProvider
+{
+    public function getDefaultConnection() : Connection;
+
+    /**
+     * @throws InvalidArgumentException in case a connection with the given name does not exist.
```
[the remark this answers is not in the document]
dmaicher: Added `ConnectionNotFoundException`

bin/doctrine-dbal.php
```diff
@@ -41,17 +42,17 @@
...
-$helperSet = require $configFile;
+$commands                      = [];
+$helperSetOrConnectionProvider = require $configFile;
 
-if (! $helperSet instanceof HelperSet) {
-    foreach ($GLOBALS as $helperSetCandidate) {
-        if ($helperSetCandidate instanceof HelperSet) {
-            $helperSet = $helperSetCandidate;
+if (! $helperSetOrConnectionProvider instanceof HelperSet && ! $helperSetOrConnectionProvider instanceof ConnectionProvider) {
+    foreach ($GLOBALS as $candidate) {
```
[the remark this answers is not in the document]
dmaicher: ok. done

dmaicher:
Actually this change would also affect `doctrine/migrations` and possibly `doctrine/orm` cli tools. They also use this console command helper approach

lib/Doctrine/DBAL/Tools/Console/ConnectionNotFoundException.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace Doctrine\DBAL\Tools\Console;
+
+use OutOfBoundsException;
+
+class ConnectionNotFoundException extends OutOfBoundsException
```
morozov, line 7: We don't use the `Exception` suffix for exceptions. `ConnectionNotFound` is descriptive enough. Please also mark `final`.

lib/Doctrine/DBAL/Tools/Console/ConsoleRunner.php
```diff
@@ -30,20 +36,30 @@ public static function createHelperSet(Connection $connection)
...
-    public static function run(HelperSet $helperSet, $commands = [])
+    public static function run($helperSetOrConnectionProvider, $commands = [])
     {
         $cli = new Application('Doctrine Command Line Interface', Version::VERSION);
 
         $cli->setCatchExceptions(true);
-        $cli->setHelperSet($helperSet);
 
-        self::addCommands($cli);
+        /** @var ConnectionProvider|null $connectionProvider */
```
morozov, line 52: Isn't it `ConnectionProvider|HelperSet|null`?

lib/Doctrine/DBAL/Tools/Console/ConsoleRunner.php
```diff
@@ -30,20 +36,30 @@ public static function createHelperSet(Connection $connection)
...
+        $connectionProvider = $helperSetOrConnectionProvider;
```
morozov, line 53: This assignment looks invalid. We know that the right-hand side value may be of the type `HelperSet` and yet we're assigning it to the variable named `$connectionProvider` and then set back to null in this case.

Please rework.

lib/Doctrine/DBAL/Tools/Console/ConsoleRunner.php
```diff
@@ -30,20 +36,30 @@ public static function createHelperSet(Connection $connection)
...
+        if ($helperSetOrConnectionProvider instanceof HelperSet) {
+            $connectionProvider = null;
+            $cli->setHelperSet($helperSetOrConnectionProvider);
+            @trigger_error(sprintf('Passing an instance of "%s" as the first argument is deprecated. Pass an instance of "%s" instead.', HelperSet::class, ConnectionProvider::class), E_USER_DEPRECATED);
```
morozov, line 57: Should `trigger_error()` happen in the very beginning of the `if`-branch? Even if `$cli->setHelperSet()` throws an exception (regardless of whether it's possible), we want the deprecation to be triggered.

morozov (commented):
Looks good for the most part :+1: Just needs some cleanup.

dmaicher:
@stof I think we don't need this change for the doctrine dbal-bundle.

Its actually only one single command that we need to extend/proxy:

https://github.com/doctrine/dbal-bundle/pull/1/files#diff-3c45b0c1750eeaf9333a6f76e86e91adR15

So @morozov I'm fine if we close this. Not sure the introduced deprecations and overhead for needed fixes on related bundles/libraries are justified. If you still think this feature is useful let me know and I can fix the conflicts and the build so we can finish this PR.

morozov:
> I think we don't need this change for the doctrine dbal-bundle.

No offense, but this is exactly why I was against the original in https://github.com/doctrine/dbal/pull/3892#issuecomment-607605543:

> Just adding a bunch of new code because another project needs it looks like a bad idea to me.

Whether it's adding or non-adding a feature, should be defined by whether it belongs to the project, not by whether the other project needs or doesn't need it.

>  If you still think this feature is useful let me know and I can fix the conflicts and we can finish this PR.

I believe it makes the functionality more extensible (i.e. the possibility to use one or more connections in CLI). Even if the bundle doesn't need it, other projects still may,

morozov:
@dmaicher could you squash everything, please?

dmaicher:
@morozov done :wink:

morozov (approved):

morozov:
Thanks, @dmaicher.

morozov:
@dmaicher would you mind submitting a pull request that removes the deprecated code from `3.0.x`?

PowerKiKi:
I might be missing something, but I think this break the implicit compatibility with `doctrine/orm` when using `cli-config.php` to configure CLI tools of both `dbal` and `orm`.

If I want to avoid `ConnectionHelper`, which has been deprecated by this PR, it seems I have to return directly a `ConnectionProvider` instead of a `HelperSet`. But `orm` package still require a `HelperSet`. That means that the following two commands cannot work at the same time anymore:

```sh
./vendor/bin/doctrine-dbal --version # will work
./vendor/bin/doctrine --version    # will throw exception
```
Moving forward what is the plan to support a single configuration file for both tools ?

morozov:
@PowerKiKi please file an issue in the ORM repository. There isn't much we can do about it here.

PowerKiKi:
I forgot to mention that it also affects `doctrine/migrations`, since `./vendor/bin/doctrine-migrations --version` will fail in the same way as ORM.

So I don't think it's a problem with ORM, nor with Migrations. But it is a problem with DBAL itself, who broke compatibility with all others. Because DBAL now refuse to accept "too much" configuration, then DBAL cannot be used with other tools anymore.

Also I don't see how ORM, or Migrations, could be made to work with only a ConnectionProvider. It requires other things, such as `em` for ORM and `configuration` for Migrations. Because of this PR (and even more so with #4059), it became nearly impossible to share a single config file that return different things depending on who is using that config.

I think this PR went overboard. And instead of replacing HelperSet with a ConnectionProvider, it should have kept the HelperSet and accept a ConnectionProvider in that HelperSet. This would allow for the improvement originally intended while preserving compatibility across the doctrine ecosystem.

Would you accept a PR that restore the HelperSet as the only right way to configure DBAL (and ORM and Migrations), while preserving the new ConnectionProvider in that HelperSet ?

morozov:
> Also I don't see how ORM, or Migrations, could be made to work with only a ConnectionProvider. It requires other things, such as `em` for ORM and `configuration` for Migrations.

`ConnectionProvider` is an interface that defines the API required by the DBAL. Apart from the API defined by the interface, the implementing class can implement any other API required by any other use cases. What's the problem in implementing such classes for the ORM and Migrations?

PowerKiKi:
Fair enough, it can be made to work. I somehow missed the fact that it was an interface, and it could be implemented in other packages too.

Still it seems it would have been better if other packages had similar PR at the same time. Right now there is no way to use latest DBAL with ORM or Migrations, without using something deprecated.

PowerKiKi:
I created the following issues to track progress in their respective packages:

- https://github.com/doctrine/orm/issues/8327
- https://github.com/doctrine/migrations/issues/1070

PR #3960, opened 2020-04-17
Only mark getMessage as mutation free

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

We do not know if children classes do or do not mutate state.
This reverts commit 5a61341fea474990fe2c4fb5e27515c053ccf4a4.
This translates into many issues on master: having to mark much more exceptions as immutable, one of which actually isn't detected as such.

See https://github.com/doctrine/dbal/pull/3954#issuecomment-615350424

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractDriverException.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractDriverException.php
@@ -6,8 +6,6 @@ use Exception;
 
 /**
  * Abstract base implementation of the {@link DriverException} interface.
- *
- * @psalm-immutable
  */
 abstract class AbstractDriverException extends Exception implements DriverException
 {
--- a/lib/Doctrine/DBAL/Driver/DriverException.php
+++ b/lib/Doctrine/DBAL/Driver/DriverException.php
@@ -9,8 +9,6 @@ use Throwable;
  *
  * Driver exceptions provide the SQLSTATE of the driver
  * and the driver specific error code at the time the error occurred.
- *
- * @psalm-immutable
  */
 interface DriverException extends Throwable
 {
@@ -28,6 +26,8 @@ interface DriverException extends Throwable
      * Returns the driver error message.
      *
      * @return string
+     *
+     * @psalm-mutation-free
      */
     public function getMessage();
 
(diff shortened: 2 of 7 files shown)


morozov:
> This translates into many issues on master: having to mark much more exceptions as immutable […]

I don't think this is something we should be avoiding. 

> […] one of which actually isn't detected as such.

I think this is (or may be) because an interface method marked as immutable doesn't guarantee that its implementors are immutable as well. Maybe there is a Psalm annotation that forces the implementors to respect the parent's constraints, I don't know.

I'd rather suppress the error(s) that don't look legit rather than make annotation changes that don't look legit either. Exception classes must be immutable by design.

greg0ire:
Okay.

1. I know why there is an issue, and I fixed it: https://github.com/vimeo/psalm/pull/3171
2. It should work in the CI now, but, not when developing because we are using master in the CI, and I think that's bad (just got bitten by it in sql-formatter because there is a bug with `array_keys` in master, that I have yet to report, but that you can witness here: https://github.com/doctrine/sql-formatter/pull/43/checks?check_run_id=597345249#step:4:76).
3. There are exceptions in 2.10.x that I did not mark as immutable

Here is the course of action I feel would be best:

1. Stabilize things by merging #3961 
2. Get Psalm on all branches including master by merging this PR
3. Wait for a release of Psalm that has my fix and does not have the `array_keys` bug (I don't know if we actually suffer from it here, but chances are high).
4. Revert this PR and mark all exceptions as immutable, and let that flow into master.

morozov:
> Get Psalm on all branches including master by merging this PR

Thank you for the research @greg0ire. I'm still not sure why we need to revert valid annotations instead of suppressing invalid errors. Another approach might be to fork a stable Psalm branch that doesn't have the `array_keys()` bug and apply your fixes on top. We could use it in Composer and on CI before your fixes are released.

greg0ire:
I think the `array_keys` bug is gone from master now (I did another contribution and did not observe it). I'm going to add more `immutable` annotations to 2.10.x

PR #3965, opened 2020-04-19
Sync expected coverage report count

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

See e360ba0 for AppVeyor. For ContinuousPHP, I am not how to track the
history of the pipeline, or if there is one.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -21,7 +21,7 @@ before_commands:
 tools:
     external_code_coverage:
         timeout: 3600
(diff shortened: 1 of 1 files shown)


morozov (approved):
The discrepancy on Appveyor must have been just a human factor. As for ContinuousPHP, their build settings are only accessible via the UI. Recently, I had to disable the builds `pdo_oci` build because the CI broke something again in their Docker environment. Hence, a human factor again.

Issue #3966, opened 2020-04-19
MySQL 'ON UPDATE CURRENT_TIMESTAMP' not supported


djozsef:
### Bug Report

Schema dump does not respect column "extra" properties for "ON UPDATE" default behaviour.

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.x

#### Summary

When dumping schema the `ON UPDATE CURRENT_TIMESTAMP` default behaviour is not respected and not included in the generated SQL.

#### Current behaviour

`ON UPDATE CURRENT_TIMESTAMP` is not appended to default value declaration of the field.

#### How to reproduce

Create a table with at least one `DATETIME` column with `ON UPDATE CURRENT_TIMESTAMP` enabled:
```SQL
CREATE TABLE IF NOT EXISTS `test`.`name_list` (
  `ID` INT(11) NOT NULL AUTO_INCREMENT,
  `name` VARCHAR(255) NULL DEFAULT NULL,
  `date_created` DATETIME NULL DEFAULT CURRENT_TIMESTAMP,
  `date_updated` DATETIME NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  PRIMARY KEY (`ID`))
ENGINE = InnoDB
DEFAULT CHARACTER SET = utf8;
```

Then run migrations schema dump and see that the `ON UPDATE CURRENT_TIMESTAMP` is omitted:
```SQL
CREATE TABLE name_list (
ID INT AUTO_INCREMENT NOT NULL,
name VARCHAR(255) CHARACTER SET utf8 DEFAULT NULL COLLATE `utf8_general_ci`, 
date_created DATETIME DEFAULT CURRENT_TIMESTAMP, 
date_updated DATETIME DEFAULT CURRENT_TIMESTAMP, PRIMARY KEY(ID)) 
DEFAULT CHARACTER SET utf8 COLLATE `utf8_unicode_ci` 
ENGINE = InnoDB COMMENT = '' 
```

#### Expected behaviour

Include `ON UPDATE CURRENT_TIMESTAMP` in default value declarations.

PR #3967, opened 2020-04-19
Added ON UPDATE default field support for MySQL

djozsef:
See referenced issue #3966

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #3966

#### Summary

When dumping schema the `ON UPDATE CURRENT_TIMESTAMP` default behaviour is not respected and not included in the generated SQL.

I added the needed MySQL platform and schema cases to render `ON UPDATE CURRENT_TIMESTAMP` in default value declarations.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -476,7 +476,7 @@ SQL
             $field['default'] = null;
         }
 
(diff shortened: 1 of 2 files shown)


djozsef:
@morozov does this PR worth reworking to be compatible/mergable with `doctrine:2.13.x` and `doctrine:3.1.x`?

morozov:
I wouldn't like to introduce extra conditional logic that parses the default value to support a non-portable feature. A proper way to implement this would be to introduce the support for default expressions (as opposed to values).

PR #3969, opened 2020-04-20
Psalm 7

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

### Summary

Only 3 errors at level 7, but either not fixable or a bit controversial.

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <psalm
     totallyTyped="false"
-    errorLevel="8"
+    errorLevel="7"
     resolveFromConfigFile="true"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="https://getpsalm.org/schema/config"
--- a/psalm.xml
+++ b/psalm.xml
@@ -32,6 +32,15 @@
                 <file name="lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php"/>
             </errorLevel>
         </ConflictingReferenceConstraint>
+        <MethodSignatureMismatch>
+            <errorLevel type="suppress">
+                <!--
+                    This one is clearly deliberate; it's a hack to adapt \PDO to
+                    Doctrine\DBAL\Driver\Connection
+                -->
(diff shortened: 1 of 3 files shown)


psalm.xml
```diff
@@ -23,6 +23,24 @@
         <file name="vendor/jetbrains/phpstorm-stubs/sqlsrv/sqlsrv.php" />
     </stubs>
     <issueHandlers>
+        <ConflictingReferenceConstraint>
+            <errorLevel type="suppress">
+                <!--
+                    This one is just too convoluted for Psalm to figure out, by
+                    its author's own admission
+                -->
+                <file name="lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php"/>
```
morozov, line 32: IIRC, it must be something like this but more complicated:
```php
function test(string &$i) : void {
    $i = null;
}
```
Since we're raising the bar, I'd rather fix the issues that are not false-positives, and this one is most likely not. I'll try to find some time and solve this before we proceed.

greg0ire: Yeah there are 2 methods like this, one can make `$i` a `null` or a `string` the other one only accepts strings. Great if we can rework the code.

morozov: Interestingly, even if I change the condition to `if ($currentLiteralDelimiter === null)` or add `assert(is_string($currentLiteralDelimiter))` before calling the function that expects a string, the error still persists. I don't see an easy way to rework this w/o complicating the implementation a lot (i.e. instead of using references, return structures and then extract their properties into the current state).

At this point, it looks like a shortcoming of Psalm rather than a real issue. Unlike the example in the [article](https://psalm.dev/docs/running_psalm/issues/ConflictingReferenceConstraint/), the null value is only possible in the branch that accepts null.

morozov (approved):

psalm.xml
greg0ire: Yeah I tried this too, then I asked for support [on Slack and was told I could safely ignore that one](https://app.slack.com/client/T39U3FUKV/C8SFXTD2M/thread/C6JKVFPMG-1587419029.172700)

Issue #3561, opened 2019-05-27
MySQL SPATIAL index recreated every migrations:diff with prohibited length


vvoody-:
|    Q        |   A
|------------ | ------
| Version     | v2.9.2

## Current behaviour

Every time I run diff on my db, there is recreation of spatial index. Only difference is length of index in down function.
```php
<?php

declare(strict_types=1);

namespace Migrations;

use Doctrine\DBAL\Schema\Schema;
use Doctrine\Migrations\AbstractMigration;

/**
 * Auto-generated Migration: Please modify to your needs!
 */
final class Version20190527103305 extends AbstractMigration
{
    public function getDescription() : string
    {
        return '';
    }

    public function up(Schema $schema) : void
    {
        // this up() migration is auto-generated, please modify it to your needs
        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');

        $this->addSql('DROP INDEX coordinates_idx ON meal');
        $this->addSql('CREATE SPATIAL INDEX coordinates_idx ON meal (coordinates)');
    }

    public function down(Schema $schema) : void
    {
        // this down() migration is auto-generated, please modify it to your needs
        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');

        $this->addSql('DROP INDEX coordinates_idx ON meal');
        $this->addSql('CREATE SPATIAL INDEX coordinates_idx ON meal (coordinates(32))');
    }
}
```

## Expected behaviour

```
No changes detected in your mapping information.
```

## Posible cause

I was looking where the number 32 come from and find out its `sub_part` in result of `SHOW INDEX FROM meal;` query. ([related code](https://github.com/doctrine/dbal/blob/v2.9.2/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L151))

This `sub_part` value is used [here](https://github.com/doctrine/dbal/blob/v2.9.2/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php#L71) as length of index.

This length should not be included in `CREATE SPATIAL INDEX` query becouse 

> Column prefix lengths are prohibited.

[source](https://dev.mysql.com/doc/refman/8.0/en/create-index.html)

SenseException:
[...]

Issue #3550, opened 2019-05-22
PostgreSQL Extensions support


Knudian:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes/no
| BC Break    | yes/no

#### Summary

Since PostgreSQL can have a lot of extensions, it would be great if, while setting up DBAL, we could provide a list of extensions in use within the database.

This configuration would be implemented the package configuration :
```yml
# In a Symfony 4 project
# [project]/config/packages/doctrine.yaml 
doctrine:
    dbal:
        extensions:
            - pgstattuple
            - pg_prometheus
            - timescaledb
````

---

For information, I have all those available on a simple instance :
```txt
        name        | default_version | installed_version |                               comment                                
--------------------+-----------------+-------------------+----------------------------------------------------------------------
 pgstattuple        | 1.5             |                   | show tuple-level statistics                                          
 bloom              | 1.0             |                   | bloom access method - signature file based index                     
 pgrowlocks         | 1.2             |                   | show row-level locking information                                   
 autoinc            | 1.0             |                   | functions for autoincrementing fields                                
 dict_xsyn          | 1.0             |                   | text search dictionary template for extended synonym processing      
 intarray           | 1.2             |                   | functions, operators, and index support for 1-D arrays of integers   
 amcheck            | 1.0             |                   | functions for verifying relation integrity                           
 uuid-ossp          | 1.1             |                   | generate universally unique identifiers (UUIDs)                      
 pg_stat_statements | 1.6             |                   | track execution statistics of all SQL statements executed            
 xml2               | 1.1             |                   | XPath querying and XSLT                                              
 lo                 | 1.1             |                   | Large Object maintenance                                             
 dict_int           | 1.0             |                   | text search dictionary template for integers                         
 intagg             | 1.1             |                   | integer aggregator and enumerator (obsolete)                         
 plpgsql            | 1.0             | 1.0               | PL/pgSQL procedural language                                         
 moddatetime        | 1.0             |                   | functions for tracking last modification time                        
 chkpass            | 1.0             |                   | data type for auto-encrypted passwords                               
 refint             | 1.0             |                   | functions for implementing referential integrity (obsolete)          
 timetravel         | 1.0             |                   | functions for implementing time travel                               
 btree_gin          | 1.2             |                   | support for indexing common datatypes in GIN                         
 hstore             | 1.4             |                   | data type for storing sets of (key, value) pairs                     
[...]

PR #3970, opened 2020-04-21
Ignore index prefix lengths for spatial indexes

ksaveras:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #3561

#### Summary

MySQL spatial indexes do not support index lengths and this check should be skipped.
If there are MSSQL or PGSQL experts - please comment.

This PR skips index lengths check if both indexes have a spatial flag.

I.e. when an index is created, MySQL 5.7 creates with 32 bytes prefix length. AWS Aurora (MySQL 5.7 compatible) - 24 bytes prefix length.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Index.php
+++ b/lib/Doctrine/DBAL/Schema/Index.php
@@ -223,7 +223,7 @@ class Index extends AbstractAsset implements Constraint
                 return false;
             }
 
-            if (! $this->hasSameColumnLengths($other)) {
+            if (! $this->sameFlag('spatial', $other) && ! $this->hasSameColumnLengths($other)) {
                 return false;
             }
 
@@ -372,5 +372,13 @@ class Index extends AbstractAsset implements Constraint
 
         return array_filter($this->options['lengths'] ?? [], $filter)
             === array_filter($other->options['lengths'] ?? [], $filter);
+    }
+
+    /**
+     * Returns whether the index has the same flag as the other
+     */
+    private function sameFlag(string $flag, Index $other): bool
+    {
+        return $this->hasFlag($flag) && $other->hasFlag($flag);
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
+++ b/tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
@@ -173,4 +173,44 @@ class IndexTest extends TestCase
         self::assertSame('name IS NULL', $idx2->getOption('WHERE'));
         self::assertSame(['where' => 'name IS NULL'], $idx2->getOptions());
     }
+
+    /**
+     * @param string[] $flags1
+     * @param string[] $flags2
+     * @param int[]    $lengths1
+     * @param int[]    $lengths2
+     *
+     * @dataProvider spatialIndexProvider
+     */
+    public function testFulfilledSpatialIndex(
+        array $flags1,
+        array $flags2,
+        array $lengths1,
+        array $lengths2,
+        bool $expected
+    ): void {
+        $index1 = new Index('index1', ['geo_location'], false, false, $flags1, ['lengths' => $lengths1]);
+        $index2 = new Index('index2', ['geo_location'], false, false, $flags2, ['lengths' => $lengths2]);
+
+        self::assertSame($expected, $index1->isFullfilledBy($index2));
+        self::assertSame($expected, $index2->isFullfilledBy($index1));
+    }
+
+    /**
+     * @return mixed[][]
+     */
+    public static function spatialIndexProvider(): iterable
+    {
+        return [
+            'spatial-indexes-without-length' => [['spatial'], ['spatial'], [], [], true],
+            'spatial-indexes-with-2nd-length' => [['spatial'], ['spatial'], [], [32], true],
+            'spatial-indexes-with-length' => [['spatial'], ['spatial'], [32], [32], true],
+            'spatial-indexes-with-different-lengths' => [['spatial'], ['spatial'], [32], [24], true],
+            'spatial-indexes-with-1st-length' => [['spatial'], ['spatial'], [24], [], true],
+            'index-with-length-compare-to-spatial-index' => [[], ['spatial'], [24], [], false],
+            'spatial-with-length-index-compare-to-index' => [['spatial'], [], [24], [], false],
+            'index-compare-to-spatial-with-length-index' => [[], ['spatial'], [], [24], false],
+            'spatial-index-compare-to-index-with-length' => [['spatial'], [], [], [24], false],
+        ];
+    }
 }


cesurapp (dismissed):
This development works for the spatial index.

Apostropher:
Is there an ETA for when this will be merged? I'm currently waiting for this fix.

greg0ire:
Not a PG expert, but it looks like this wouldn't be handled by flags in PG since [the syntax, which requires the PostGis extension is](https://postgis.net/workshops/postgis-intro/indexing.html) :

```sql
CREATE INDEX nyc_census_blocks_geom_idx
  ON nyc_census_blocks
  USING GIST (geom);
```



flags are apparently supposed to appear between `CREATE` and `INDEX`.

PostGIS does not seem to be supported though ([and will never be](https://github.com/doctrine/dbal/issues/3550#issuecomment-496319734)?)

greg0ire:
Also, the linked issue shows that this piece of code is responsible for the length appearing: https://github.com/doctrine/dbal/blob/22800bd651c1d8d2a9719e2a3dc46d5108ebfcc9/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php#L69-L71

Why not change it so that the length is not set at all in case the index is spatial?

lib/Doctrine/DBAL/Schema/Index.php
```diff
@@ -212,7 +214,7 @@ public function isFullfilledBy(Index $other)
                 return false;
             }
 
-            if (! $this->hasSameColumnLengths($other)) {
+            if (! $this->sameFlag(static::FLAG_SPATIAL, $other) && ! $this->hasSameColumnLengths($other)) {
```
morozov, line 217: Why is only the `SPATIAL` flag of interest? Shouldn't it be something like `$this->hasSameFlags($other)`?

tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
```diff
@@ -188,4 +189,40 @@ public function testOptions() : void
...
+        $index2 = new Index('index2', ['geo_location'], false, false, [], ['lengths' => $lengths2]);
+        if ($flag2 !== null) {
+            $index2->addFlag($flag2);
+        }
+
+        self::assertSame($expected, $index1->isFullfilledBy($index2));
+        self::assertSame($expected, $index2->isFullfilledBy($index1));
+    }
+
+    public static function spatialIndexProvider() : Generator
```
morozov, line 215: Please use `iterable` instead since PHPUnit doesn't need a more specific type, and it's just an implementation detail of this given provider.

tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
```diff
@@ -188,4 +189,40 @@ public function testOptions() : void
...
+    /**
+     * @param int[] $lengths1
+     * @param int[] $lengths2
+     *
+     * @dataProvider spatialIndexProvider
+     */
+    public function testFulfilledSpatialIndex(?string $flag1, ?string $flag2, array $lengths1, array $lengths2, bool $expected) : void
+    {
+        $index1 = new Index('index1', ['geo_location'], false, false, [], ['lengths' => $lengths1]);
+        if ($flag1 !== null) {
```
morozov, line 202: Please avoid flow control in tests. Would it be possible to pass flags as an empty or non-empty array?

tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
```diff
@@ -188,4 +189,40 @@ public function testOptions() : void
...
+        self::assertSame($expected, $index1->isFullfilledBy($index2));
+        self::assertSame($expected, $index2->isFullfilledBy($index1));
+    }
+
+    public static function spatialIndexProvider() : Generator
+    {
+        yield ['spatial', 'spatial', [], [], true];
+        yield ['spatial', 'spatial', [], [32], true];
+        yield ['spatial', 'spatial', [32], [32], true];
+        yield ['spatial', 'spatial', [32], [24], true];
```
morozov, line 220: Since this is a data provider for a spatial index test, why do we care about testing the length comparison? If the logic of comparison of the flags _and_ lengths needs to be tested, it's better to do it in a separate test that would not get too much into the details of why the lengths or flags mismatch.

lib/Doctrine/DBAL/Schema/Index.php
ksaveras: We don't want to check other index flags. This is applicable for spatial only. I'm open to suggestions.

tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
ksaveras: PHPUnit does not restrict the data provider return type. Do we have contribution guidelines doc where it describes what types must be used for data providers?

tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
ksaveras: Good idea

tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
ksaveras: We are testing if the method `isFullfilledBy` works as expected in different cases. When spatial index created or migration is generated (or checked) on different platforms: MySQL, AWS Aurora MySQL.

ksaveras:
> Also, the linked issue shows that this piece of code is responsible for the length appearing:
> 
> https://github.com/doctrine/dbal/blob/22800bd651c1d8d2a9719e2a3dc46d5108ebfcc9/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php#L69-L71
> 
> Why not change it so that the length is not set at all in case the index is spatial?

I like this idea.

tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
morozov: No, but we have existing tests that could be used as an example.

lib/Doctrine/DBAL/Schema/Index.php
morozov: Why don't we? Does it mean that if any other flag is added or removed to the index definition, we want to consider them equal and not cause a schema change?

tests/Doctrine/Tests/DBAL/Schema/IndexTest.php
morozov: This is fine but some of these different cases are better to be implemented as a separate test method. Otherwise, if someone breaks the length comparison logic and `testFulfilledSpatialIndex()` fails, it will be misleading feedback from the test suite.

SenseException: Using a text key on data providers would also give more information to a failing test with a data row, but separate test cases provide a more contextual structure for the test class.

lib/Doctrine/DBAL/Schema/Index.php
```diff
@@ -15,6 +15,8 @@
 
 class Index extends AbstractAsset implements Constraint
 {
+    public const FLAG_SPATIAL = 'spatial';
```
SenseException, line 18: The Public scope isn't needed for this constant. There's currently only one place in code where this constant is in use. We probably don't need `'spatial'` as a constant for now.

lib/Doctrine/DBAL/Schema/Index.php
SenseException: I'm also not a fan of `static::`. Is this currently necessary?

SenseException (changes requested):

lib/Doctrine/DBAL/Schema/Index.php
morozov: Agree. Please see https://github.com/doctrine/dbal/pull/3970#discussion_r449196691.

greg0ire:
@ksaveras 2.10.x only accepts security fixes now, retargeting to 2.11.x .

greg0ire:
```
git rebase origin/2.11.x
composer install
vendor/bin/phpcbf
```

should help with the cs issues.

ksaveras:
Thanks @greg0ire 
Do you want to add this globally for all contributors?
https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/setting-guidelines-for-repository-contributors

ksaveras:
@greg0ire I'll close this PR and make a new one based on 2.11 branch.

greg0ire:
No no please don't!

greg0ire:
Let me handle this.

greg0ire:
>  Do you want to add this globally for all contributors?

We already have [a contributing guide](https://www.doctrine-project.org/contribute/index.html), but no link to it, which sounds a bit weird. Good point, I will make a PR later to address that.

greg0ire:
Travis got confused with the 2 PRs, let's close and reopen to trigger a new build

lib/Doctrine/DBAL/Schema/Index.php
```diff
@@ -373,4 +373,14 @@ private function hasSameColumnLengths(self $other): bool
         return array_filter($this->options['lengths'] ?? [], $filter)
             === array_filter($other->options['lengths'] ?? [], $filter);
     }
+
+    /**
+     * Returns whether the index has the same flag as the other
+     *
+     * @param string $flag
+     */
+    private function sameFlag($flag, Index $other): bool
```
morozov, line 382: Please use a scalar type declaration for `$flag` instead of the annotation.

morozov (commented):
Sorry, I misunderstood the boolean logic being tested when requested the changes in https://github.com/doctrine/dbal/pull/3970#discussion_r449198878. Makes sense.

morozov:
> […] when an index is created, MySQL 5.7 creates with 32 bytes prefix length. AWS Aurora (MySQL 5.7 compatible) - 24 bytes prefix length.

While the patch addresses this issue, I can hardly see it being a bug. DBAL does not claim the support for portability between MySQL and Aurora, in which case it would be a bug.

I don't know enough about how index length is applied to spatial indexes, so it is possible that it's a breaking change while the primary focus of `2.11.1` is to fix the existing BC breaks before the 3.0 release.

morozov:
@ksaveras what does MySQL report as the spatial index `sub_part` if the column length is prohibited? Should we, instead of changing the comparison logic, check the introspection logic?

https://github.com/doctrine/dbal/blob/22800bd651c1d8d2a9719e2a3dc46d5108ebfcc9/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php#L66-L71

E.g. not use `sub_part` to populate the `length` of a spatial index.

ottaviano:
Hi, we have the same issue. How can we help getting merged this PR?

morozov:
@ottaviano this patch requires [integration testing](https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/testing.html#integration-tests). This may shed some light on how https://github.com/doctrine/dbal/pull/3970#issuecomment-699514623 should be addressed.

morozov:
Superseded by https://github.com/doctrine/dbal/pull/4525.

PR #3973, opened 2020-04-21

greg0ire:


morozov (dismissed):

greg0ire:
Sorry @morozov, I failed to push the right commit and will need another approval for this one.

greg0ire:
Also, level 6 is already a bit more involving: https://github.com/sebastianbergmann/phpunit/pull/4184 :)

morozov (approved):

morozov:
I see that this Psalm project makes you @greg0ire communicate with and contribute to a lot of other OSS projects (Psalm, PHPUnit, PHP, JetBrains). Nice.

> Also, level 6 is already a bit more involving: [sebastianbergmann/phpunit#4184](https://github.com/sebastianbergmann/phpunit/pull/4184) :)

Does it mean we’ll need to wait for a new PHPUnit release before being able to proceed? Given that PHPUnit itself has Psalm annotations, it's surprising that it doesn't get tested by Psalm as well. Otherwise, those annotations would be already there.

greg0ire:
I was surprised too, because [it does seems to be tested with psalm](https://github.com/sebastianbergmann/phpunit/blob/9d4594e43825b9fc772c49b515add2bbe4903f38/.github/workflows/ci.yml#L49-L53). Not sure why this wasn't caught before.

PR #3974, opened 2020-04-22
Remove unneeded and risky stubs parsing

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

This fixes a crash I have when running phpstan on a machine that does not have mysqli enabled:

```
Note: Using configuration file /home/greg/dev/doctrine-dbal/phpstan.neon.dist.
PHP Fatal error:  Class mysqli_result must implement interface Traversable as part of either Iterator or IteratorAggregate in Unknown on line 0
Fatal error: Class mysqli_result must implement interface Traversable as part of either Iterator or IteratorAggregate in Unknown on line 0
```

### DIFF
--- a/tests/phpstan-polyfill.php
+++ b/tests/phpstan-polyfill.php
@@ -2,12 +2,9 @@
 
 declare(strict_types=1);
 
-(static function () : void {
-    foreach (['ibm_db2', 'mysqli', 'oci8', 'sqlsrv', 'pgsql'] as $extension) {
-        if (extension_loaded($extension)) {
-            continue;
-        }
+// PHPStan does not read global constants from the stubs yet, remove this when it does
+if (defined('OCI_NO_AUTO_COMMIT')) {
+    return;
+}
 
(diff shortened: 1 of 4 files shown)


tests/phpstan-polyfill.php
```diff
@@ -3,11 +3,10 @@
 declare(strict_types=1);
 
 (static function () : void {
```
morozov, line 5: Without local variables being used, this closure is no longer needed.

morozov (commented):
Just curious, how does PHPStan know where our subs are. Or it uses its own?

greg0ire:
> Or it uses its own?

[It uses its own](https://symfony-devs.slack.com/archives/C8SFXTD2M/p1587573341275300?thread_ts=1587572736.274300&cid=C8SFXTD2M)

morozov (approved):

PR #3903, opened 2020-03-17
Upgrade PHPUnit to 9.0

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

**TODO:**
- [x] Backport https://github.com/doctrine/dbal/pull/3750 and https://github.com/doctrine/dbal/pull/3752 (see https://github.com/sebastianbergmann/phpunit/issues/3956).
- [x] Rework the test that relies on `assertContains()` performing non-strict comparison internally (see https://github.com/sebastianbergmann/phpunit/issues/3426).
- [x] Use `expectExceptionMessageMatches()` instead of `expectExceptionMessageRegExp()` ( see https://github.com/sebastianbergmann/phpunit/issues/3957).
- [x] Use `expectNotice()` instead of `expectException(Notice::class)` (see https://github.com/sebastianbergmann/phpunit/issues/3776)

PR #3977, opened 2020-04-25
Psalm 6

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

Pushing although there are still 9 errors left, because they are either blocked, or I need help.
See:
- [x] https://github.com/vimeo/psalm/pull/3221
- [x] https://github.com/vimeo/psalm/issues/3224
- [x] https://github.com/vimeo/psalm/issues/3217
- [x] https://github.com/sebastianbergmann/phpunit/pull/4187
- [x] https://github.com/sebastianbergmann/phpunit/pull/4184
- [x] Rebase on top of https://github.com/doctrine/dbal/pull/3983 , use array shapes directly in `@param`
- [x] Figure out what to do about https://github.com/doctrine/dbal/pull/3977#discussion_r415032908
- [x] Figure out what to do about https://github.com/doctrine/dbal/pull/3977#issuecomment-619357227
- [x] Check if things can be improved with PHPStan following https://github.com/phpstan/phpstan/issues/3228 => can't reproduce now, not sure why
- [x] Wait for a stable release of Psalm
- [x] Wait for a stable release of PHPUnit
- [x] squash what makes sense squashing
- [x] do the merge locally
- [x] prepare [merge up for 2.11.x](https://github.com/doctrine/dbal/pull/4022)
- [x] prepare [merge up for 3.0.x](https://github.com/doctrine/dbal/pull/4025)
- [x] prepare [merge up for master](https://github.com/doctrine/dbal/pull/4026)

I will of course squash many of the commits together in the end.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,7 @@
         "phpunit/phpunit": "^8.5.5",
         "psalm/plugin-phpunit": "^0.10.0",
         "symfony/console": "^2.0.5|^3.0|^4.0|^5.0",
-        "vimeo/psalm": "^3.11"
+        "vimeo/psalm": "^3.11.4"
     },
     "suggest": {
         "symfony/console": "For helpful console commands such as SQL execution and import of files."
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
         "jetbrains/phpstorm-stubs": "^2019.1",
         "nikic/php-parser": "^4.4",
         "phpstan/phpstan": "^0.12",
-        "phpunit/phpunit": "^8.4.1",
+        "phpunit/phpunit": "^8.5.5",
         "psalm/plugin-phpunit": "^0.10.0",
         "symfony/console": "^2.0.5|^3.0|^4.0|^5.0",
         "vimeo/psalm": "^3.11"
--- a/composer.json
+++ b/composer.json
@@ -43,6 +43,7 @@
         "nikic/php-parser": "^4.4",
         "phpstan/phpstan": "^0.12",
         "phpunit/phpunit": "^8.4.1",
+        "psalm/plugin-phpunit": "^0.10.0",
         "symfony/console": "^2.0.5|^3.0|^4.0|^5.0",
         "vimeo/psalm": "^3.11"
     },
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1034,9 +1034,9 @@ class Connection implements DriverConnection
      *
      * This method supports PDO binding types as well as DBAL mapping types.
      *
-     * @param string         $query  The SQL query.
-     * @param mixed[]        $params The query parameters.
-     * @param int[]|string[] $types  The parameter types.
+     * @param string                 $query  The SQL query.
+     * @param array<mixed>           $params The query parameters.
+     * @param array<int|string|null> $types  The parameter types.
      *
      * @return int The number of affected rows.
      *
--- a/lib/Doctrine/DBAL/Driver/DrizzlePDOMySql/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/DrizzlePDOMySql/Driver.php
@@ -33,6 +33,8 @@ class Driver extends \Doctrine\DBAL\Driver\PDOMySql\Driver
 
     /**
      * {@inheritdoc}
+     *
+     * @return DrizzlePlatform
      */
     public function getDatabasePlatform()
     {
@@ -41,6 +43,8 @@ class Driver extends \Doctrine\DBAL\Driver\PDOMySql\Driver
 
     /**
      * {@inheritdoc}
+     *
+     * @return DrizzleSchemaManager
      */
     public function getSchemaManager(\Doctrine\DBAL\Connection $conn)
     {
--- a/lib/Doctrine/DBAL/DriverManager.php
+++ b/lib/Doctrine/DBAL/DriverManager.php
@@ -115,11 +115,15 @@ final class DriverManager
      * <b>driverClass</b>:
      * The driver class to use.
      *
-     * @param mixed[]            $params       The parameters.
-     * @param Configuration|null $config       The configuration to use.
-     * @param EventManager|null  $eventManager The event manager to use.
+     * @param  array{wrapperClass?: class-string<T>} $params
+     * @param Configuration|null                    $config       The configuration to use.
+     * @param EventManager|null                     $eventManager The event manager to use.
      *
      * @throws DBALException
+     *
+     * @phpstan-param mixed[] $params
+     * @psalm-return ($params is array{wrapperClass:mixed} ? T : Connection)
+     * @template T of Connection
      */
     public static function getConnection(
         array $params,
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -2816,8 +2816,8 @@ abstract class AbstractPlatform
     }
 
     /**
-     * @param string      $table
-     * @param string|null $database
+     * @param string $table
+     * @param string $database
      *
      * @return string
      *
--- a/tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
+++ b/tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
@@ -43,6 +43,8 @@ class StatementIteratorTest extends DbalTestCase
     }
 
     /**
+     * @param class-string<Statement> $class
+     *
      * @dataProvider statementProvider()
      */
     public function testStatementIterationCallsFetchOncePerStep(string $class) : void
@@ -80,7 +82,7 @@ class StatementIteratorTest extends DbalTestCase
     }
 
     /**
-     * @return string[][]
+     * @return iterable<array{0: class-string<Statement>}>
      */
     public static function statementProvider() : iterable
     {
--- a/tests/Doctrine/Tests/DBAL/FutureVersionAwarePlatformDriver.php
+++ b/tests/Doctrine/Tests/DBAL/FutureVersionAwarePlatformDriver.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL;
+
+use Doctrine\DBAL\Driver;
+use Doctrine\DBAL\VersionAwarePlatformDriver;
+
+/**
+ * Remove me in 3.0.x
+ */
+interface FutureVersionAwarePlatformDriver extends VersionAwarePlatformDriver, Driver
+{
+}
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
@@ -372,9 +372,7 @@ abstract class AbstractPlatformTestCase extends DbalTestCase
 
     public function testGetCreateTableSqlDispatchEvent() : void
     {
-        $listenerMock = $this->getMockBuilder($this->getMockClass('GetCreateTableSqlDispatchEvenListener'))
-            ->addMethods(['onSchemaCreateTable', 'onSchemaCreateTableColumn'])
-            ->getMock();
+        $listenerMock = $this->createMock(GetCreateTableSqlDispatchEventListener::class);
         $listenerMock
             ->expects($this->once())
             ->method('onSchemaCreateTable');
@@ -396,9 +394,7 @@ abstract class AbstractPlatformTestCase extends DbalTestCase
 
     public function testGetDropTableSqlDispatchEvent() : void
     {
-        $listenerMock = $this->getMockBuilder($this->getMockClass('GetDropTableSqlDispatchEventListener'))
-            ->addMethods(['onSchemaDropTable'])
-            ->getMock();
+        $listenerMock = $this->createMock(GetDropTableSqlDispatchEventListener::class);
         $listenerMock
             ->expects($this->once())
             ->method('onSchemaDropTable');
@@ -413,17 +409,7 @@ abstract class AbstractPlatformTestCase extends DbalTestCase
 
     public function testGetAlterTableSqlDispatchEvent() : void
     {
-        $events = [
-            'onSchemaAlterTable',
-            'onSchemaAlterTableAddColumn',
-            'onSchemaAlterTableRemoveColumn',
-            'onSchemaAlterTableChangeColumn',
-            'onSchemaAlterTableRenameColumn',
-        ];
-
-        $listenerMock = $this->getMockBuilder($this->getMockClass('GetAlterTableSqlDispatchEvenListener'))
-            ->addMethods($events)
-            ->getMock();
+        $listenerMock = $this->createMock(GetAlterTableSqlDispatchEventListener::class);
         $listenerMock
             ->expects($this->once())
             ->method('onSchemaAlterTable');
@@ -1531,4 +1517,29 @@ abstract class AbstractPlatformTestCase extends DbalTestCase
             $this->platform->modifyLimitQuery($query, null, 0)
         );
     }
+}
+
+interface GetCreateTableSqlDispatchEventListener
+{
+    public function onSchemaCreateTable() : void;
+
+    public function onSchemaCreateTableColumn() : void;
+}
+
+interface GetAlterTableSqlDispatchEventListener
+{
+    public function onSchemaAlterTable() : void;
+
+    public function onSchemaAlterTableAddColumn() : void;
+
+    public function onSchemaAlterTableRemoveColumn() : void;
+
+    public function onSchemaAlterTableChangeColumn() : void;
+
+    public function onSchemaAlterTableRenameColumn() : void;
+}
+
+interface GetDropTableSqlDispatchEventListener
+{
+    public function onSchemaDropTable() : void;
 }
--- a/tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
@@ -22,6 +22,9 @@ use function uniqid;
 
 class OraclePlatformTest extends AbstractPlatformTestCase
 {
+    /** @var OraclePlatform */
+    protected $platform;
+
     /**
      * @return mixed[][]
      */
@@ -76,6 +79,9 @@ class OraclePlatformTest extends AbstractPlatformTestCase
         $platform->assertValidIdentifier($identifier);
     }
 
+    /**
+     * @return OraclePlatform
+     */
     public function createPlatform() : AbstractPlatform
     {
         return new OraclePlatform();
--- a/tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php
@@ -348,14 +348,14 @@ class SQLAnywherePlatformTest extends AbstractPlatformTestCase
         self::assertEquals(
             'CONSTRAINT pk PRIMARY KEY CLUSTERED (a, b)',
             $this->platform->getPrimaryKeyDeclarationSQL(
-                new Index(null, ['a', 'b'], true, true, ['clustered']),
+                new Index('', ['a', 'b'], true, true, ['clustered']),
                 'pk'
             )
         );
         self::assertEquals(
             'PRIMARY KEY (a, b)',
             $this->platform->getPrimaryKeyDeclarationSQL(
-                new Index(null, ['a', 'b'], true, true)
+                new Index('', ['a', 'b'], true, true)
             )
         );
     }
@@ -391,12 +391,12 @@ class SQLAnywherePlatformTest extends AbstractPlatformTestCase
             'CONSTRAINT unique_constraint UNIQUE CLUSTERED (a, b)',
             $this->platform->getUniqueConstraintDeclarationSQL(
                 'unique_constraint',
-                new Index(null, ['a', 'b'], true, false, ['clustered'])
+                new Index('', ['a', 'b'], true, false, ['clustered'])
             )
         );
         self::assertEquals(
             'UNIQUE (a, b)',
-            $this->platform->getUniqueConstraintDeclarationSQL(null, new Index(null, ['a', 'b'], true, false))
+            $this->platform->getUniqueConstraintDeclarationSQL('', new Index('', ['a', 'b'], true, false))
         );
     }
 
--- a/tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php
@@ -517,6 +517,9 @@ class SQLAnywherePlatformTest extends AbstractPlatformTestCase
         ));
     }
 
+    /**
+     * @psalm-suppress InvalidArgument
+     */
     public function testCannotGenerateDropIndexSQLWithInvalidIndexParameter() : void
     {
         $this->expectException(InvalidArgumentException::class);
@@ -524,6 +527,9 @@ class SQLAnywherePlatformTest extends AbstractPlatformTestCase
         $this->platform->getDropIndexSQL(['index'], 'table');
     }
 
+    /**
+     * @psalm-suppress InvalidArgument
+     */
     public function testCannotGenerateDropIndexSQLWithInvalidTableParameter() : void
     {
         $this->expectException(InvalidArgumentException::class);
--- a/tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php
@@ -596,7 +596,7 @@ class SQLAnywherePlatformTest extends AbstractPlatformTestCase
         self::assertEquals(
             "REVERSE(SUBSTR(REVERSE(SUBSTR(column, PATINDEX('%[^' + c + ']%', column))), PATINDEX('%[^' + c + ']%', " .
             "REVERSE(SUBSTR(column, PATINDEX('%[^' + c + ']%', column))))))",
-            $this->platform->getTrimExpression('column', null, 'c')
+            $this->platform->getTrimExpression('column', TrimMode::UNSPECIFIED, 'c')
         );
         self::assertEquals(
             "REVERSE(SUBSTR(REVERSE(SUBSTR(column, PATINDEX('%[^' + c + ']%', column))), PATINDEX('%[^' + c + ']%', " .
--- a/tests/Doctrine/Tests/DBAL/Types/TypeTest.php
+++ b/tests/Doctrine/Tests/DBAL/Types/TypeTest.php
@@ -19,7 +19,7 @@ class TypeTest extends TestCase
     }
 
     /**
-     * @return string[][]
+     * @return iterable<string[]>
      */
     public function defaultTypesProvider() : iterable
     {
@@ -28,7 +28,10 @@ class TypeTest extends TestCase
                 continue;
             }
 
-            yield [$constant->getValue()];
+            $constantValue = $constant->getValue();
+            self::assertIsString($constantValue);
+
+            yield [$constantValue];
         }
     }
 }
--- a/tests/Doctrine/Tests/DbalPerformanceTestListener.php
+++ b/tests/Doctrine/Tests/DbalPerformanceTestListener.php
@@ -16,7 +16,7 @@ class DbalPerformanceTestListener implements TestListener
 {
     use TestListenerDefaultImplementation;
 
-    /** @var string[][] */
+    /** @var array<string, array<string, float>> */
     private $timings = [];
 
     public function endTest(Test $test, float $time) : void
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "48526e240461408d90e490246c3b4531",
+    "content-hash": "9157c6d35dd39f5399592fd91126fa07",
     "packages": [
         {
             "name": "doctrine/cache",
@@ -1622,16 +1622,16 @@
         },
         {
             "name": "phpunit/phpunit",
-            "version": "8.4.1",
+            "version": "8.5.5",
             "source": {
                 "type": "git",
                 "url": "https://github.com/sebastianbergmann/phpunit.git",
-                "reference": "366a4a0f2b971fd43b7c351d621e8dd7d7131869"
+                "reference": "63dda3b212a0025d380a745f91bdb4d8c985adb7"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/sebastianbergmann/phpunit/zipball/366a4a0f2b971fd43b7c351d621e8dd7d7131869",
-                "reference": "366a4a0f2b971fd43b7c351d621e8dd7d7131869",
+                "url": "https://api.github.com/repos/sebastianbergmann/phpunit/zipball/63dda3b212a0025d380a745f91bdb4d8c985adb7",
+                "reference": "63dda3b212a0025d380a745f91bdb4d8c985adb7",
                 "shasum": ""
             },
             "require": {
@@ -1675,7 +1675,7 @@
             "type": "library",
             "extra": {
                 "branch-alias": {
-                    "dev-master": "8.4-dev"
+                    "dev-master": "8.5-dev"
                 }
             },
             "autoload": {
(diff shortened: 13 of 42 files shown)


composer.json
```diff
@@ -42,7 +42,8 @@
         "jetbrains/phpstorm-stubs": "^2019.1",
         "nikic/php-parser": "^4.4",
         "phpstan/phpstan": "^0.12",
-        "phpunit/phpunit": "^8.4.1",
+        "phpunit/phpunit": "^8.5@dev",
```
greg0ire, line 45: Let's wait for https://github.com/sebastianbergmann/phpunit/pull/4187

lib/Doctrine/DBAL/Schema/ForeignKeyConstraint.php
```diff
@@ -56,11 +56,11 @@ class ForeignKeyConstraint extends AbstractAsset implements Constraint
...
      * Initializes the foreign key constraint.
      *
-     * @param string[]     $localColumnNames   Names of the referencing table columns.
-     * @param Table|string $foreignTableName   Referenced table.
-     * @param string[]     $foreignColumnNames Names of the referenced table columns.
-     * @param string|null  $name               Name of the foreign key constraint.
-     * @param mixed[]      $options            Options associated with the foreign key constraint.
+     * @param string[]           $localColumnNames   Names of the referencing table columns.
+     * @param Table|string       $foreignTableName   Referenced table.
+     * @param array<string|null> $foreignColumnNames Names of the referenced table columns.
```
greg0ire, line 61: This is the one I'm really not sure about. I wrote it because of this test:
https://github.com/doctrine/dbal/blob/155d028be084ef69110d09938c57c351a2126e5d/tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php#L86-L104

But it causes other issues down the line… I think there might be a bug here.

greg0ire:
In the remaining issues, there's one I don't know what to do about:

```
ERROR: NullArgument - tests/Doctrine/Tests/DBAL/Functional/Driver/SQLSrv/StatementTest.php:30:68 - Argument 1 of Doctrine\DBAL\Driver\Connection::prepare cannot be null, null value provided to parameter with type string (see https://psalm.dev/057)
        $stmt = $this->connection->getWrappedConnection()->prepare(null);
```

https://github.com/doctrine/dbal/blob/155d028be084ef69110d09938c57c351a2126e5d/tests/Doctrine/Tests/DBAL/Functional/Driver/SQLSrv/StatementTest.php#L10-L37

Please help.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1034,9 +1033,9 @@ public function query()
      *
      * This method supports PDO binding types as well as DBAL mapping types.
      *
-     * @param string         $query  The SQL query.
-     * @param mixed[]        $params The query parameters.
-     * @param int[]|string[] $types  The parameter types.
+     * @param string                 $query  The SQL query.
+     * @param mixed[]                $params The query parameters.
+     * @param array<int|string|null> $types  The parameter types.
```
morozov, line 1038: Should `$params` and `$types` have the same "container" type since they both describe parameters? I.e. `mixed[]` → `array<mixed>`.

lib/Doctrine/DBAL/Driver/DrizzlePDOMySql/Driver.php
```diff
@@ -34,15 +34,15 @@ public function createDatabasePlatformForVersion($version)
     /**
      * {@inheritdoc}
      */
-    public function getDatabasePlatform()
+    public function getDatabasePlatform() : DrizzlePlatform
```
morozov, line 37: Isn't it a breaking change?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -787,7 +787,7 @@ public function getModExpression($expression1, $expression2)
      * Returns the SQL snippet to trim a string.
      *
      * @param string      $str  The expression to apply the trim to.
-     * @param int         $mode The position of the trim (leading/trailing/both).
+     * @param int|null    $mode The position of the trim (leading/trailing/both).
```
morozov, line 790: Why null? It doesn't seem to have any special meaning here.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2778,7 +2778,7 @@ public function getListNamespacesSQL()
     }
 
     /**
-     * @param string $database
+     * @param string|null $database
```
morozov, line 2781: Why null? None of the implementations expect null.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -2837,7 +2837,7 @@ public function getListUsersSQL()
     /**
      * Returns the SQL to list all views of a database or user.
      *
-     * @param string $database
+     * @param string|null $database
```
morozov, line 2840: Why null?

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
```diff
@@ -45,6 +46,8 @@ public function testIteratorIterationCallsFetchOncePerStep() : void
     }
 
     /**
+     * @param class-string<IteratorAggregate &Statement> $class
```
morozov, line 49: The whitespace before `&` looks confusing. `ResultStatement` already extends `Traversable`. Is `IteratorAggregate` really required here?

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
```diff
@@ -79,9 +82,9 @@ private function assertIterationCallsFetchOncePerStep(Traversable $iterator, int
     }
 
     /**
-     * @return string[][]
+     * @return Generator<class-string<IteratorAggregate&Statement>[]>
      */
-    public static function statementProvider() : iterable
+    public static function statementProvider() : Generator
```
morozov, line 87: Why does it have to be this specific? `Generator` is an implementation of `iterable` of which the caller doesn't care.

`class-string...[]` is misleading. We're not returning a list of class strings. We're returning a tuple, the first argument of which is a class-string. Adding another argument of another type to the tuple will break this expectation.

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
```diff
@@ -207,6 +207,8 @@ public function testGetDateTimeTypeDeclarationSql() : void
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-return array
```
morozov, line 211: Not sure what this means.

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
```diff
@@ -535,6 +535,8 @@ public function testCreateTableColumnTypeComments() : void
 
     /**
      * @return string[]
+     *
+     * @psalm-return never-return
```
morozov, line 539: Should this be fixed in PHPUnit or in Psalm? If a function unconditionally calls a method that is marked as never-return (first option) or always throws an exception (second option), it should be automatically detected as such.

tests/Doctrine/Tests/DBAL/Types/TypeTest.php
```diff
@@ -28,7 +30,10 @@ public function defaultTypesProvider() : iterable
                 continue;
             }
 
-            yield [$constant->getValue()];
+            $constantValue = $constant->getValue();
+            assert(is_string($constantValue));
```
morozov, line 34: Should we use PHPUnit assertions in tests? They are more specific and are always effective regardless of the runtime configuration.

lib/Doctrine/DBAL/Driver/DrizzlePDOMySql/Driver.php
greg0ire: Woops it is indeed, since the class isn't final… I will use `@return` phpdoc instead.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
greg0ire: I thought having to _specify_ `TrimMode::UNSPECIFIED` would be too ironic. The purpose is to specify a 3rd argument.

```
ERROR: NullArgument - tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php:611:58 - Argument 2 of Doctrine\DBAL\Platforms\SQLAnywherePlatform::getTrimExpression cannot be null, null value provided to parameter with type int (see https://psalm.dev/057)
            $this->platform->getTrimExpression('column', null, 'c')
```

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
greg0ire:
```
ERROR: NullArgument - tests/Doctrine/Tests/DBAL/Platforms/SQLAnywhere12PlatformTest.php:55:50 - Argument 1 of Doctrine\DBAL\Platforms\SQLAnywhere12Platform::getListSequencesSQL cannot be null, null value provided to parameter with type string (see https://psalm.dev/057)
            $this->platform->getListSequencesSQL(null)
```


Looks like SQLAnywhere12 (at least) does https://github.com/doctrine/dbal/blob/155d028be084ef69110d09938c57c351a2126e5d/lib/Doctrine/DBAL/Platforms/SQLAnywhere12Platform.php#L65-L68

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
greg0ire: For the same reason, see an example among many: https://github.com/doctrine/dbal/blob/6c789b30a46906c5e669158809076d2405ea4fa9/lib/Doctrine/DBAL/Platforms/DB2Platform.php#L337-L340

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
greg0ire: I didn't mean to add the whitespace, sorry. I'll look into it

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
greg0ire: Ok, I'll try another syntax which is less misleading. As for being specific, I usually use as specific return types as I can, it can't hurt, but I will change it back.

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
greg0ire: The parent has `@psalm-return never-return`, hence this comment.

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
greg0ire: I think I already fixed everything there is to fix in PHPUnit: https://github.com/sebastianbergmann/phpunit/pull/4187

As for Psalm, I don't know… @muglug, sorry to invoke you here, but can you please comment on this? :pray:

tests/Doctrine/Tests/DBAL/Types/TypeTest.php
greg0ire: Sure

lib/Doctrine/DBAL/DriverManager.php
```diff
@@ -115,11 +115,15 @@ private function __construct()
...
      * The driver class to use.
      *
-     * @param mixed[]            $params       The parameters.
-     * @param Configuration|null $config       The configuration to use.
-     * @param EventManager|null  $eventManager The event manager to use.
+     * @param array{wrapperClass?: class-string<T>} $params       The parameters.
+     * @param Configuration|null                    $config       The configuration to use.
+     * @param EventManager|null                     $eventManager The event manager to use.
+     *
+     * @return T
```
greg0ire, line 122: I'm going to revert this, because if you pass in `null` as a `wrapperClass`, then it returns a `Connection`, and I don't think there is way to express that it return `Connection` if null, or `T` otherwise.

morozov: Would it help if we didn't ignore `NULL and used `array_key_exists()` instead of `isset()`. API-wise, what's the point in passing an option with `NULL` and expecting that it will be ignored?

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
morozov: > I usually use as specific return types as I can

I don't think it's the end goal. The test methods implement the PHPUnit's API that expects data providers to [return a `Traversable`](https://github.com/sebastianbergmann/phpunit/blob/85fefa381541fb3fcbbfe9b5f32be386551cd774/src/Util/Annotation/DocBlock.php#L412). It is fine to be more specific about the contained data type because you want to make sure the items are correct. But there's no point in being more specific about the container type unless you want to create extra coupling between the method signature (`fn (): array`/`fn (): Generator`) and the implementation (`return [];`/`yield ...` respectively).

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: The test is obviously wrong. The method doesn't accept NULL. There's no way in PHP to omit the 2nd optional argument and specify the 3rd one.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: Those methods ignore the passed value. It doesn't mean that they accept `NULL`.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
greg0ire: Ok I'll use `TrimMode::UNSPECIFIED`, I just noticed it is used in the rest of the method.

lib/Doctrine/DBAL/DriverManager.php
greg0ire: I just tried, and it does not help Psalm. These errors come back as soon as I add the `|Connection`, and using `array_key_exists` does not improve that situation:

```
ERROR: UndefinedMethod - tests/Doctrine/Tests/DBAL/Sharding/PoolingShardConnectionTest.php:193:33 - Method Doctrine\DBAL\Connection::getactiveshardid does not exist (see https://psalm.dev/022)
        self::assertNull($conn->getActiveShardId());


ERROR: UndefinedMethod - tests/Doctrine/Tests/DBAL/Sharding/PoolingShardConnectionTest.php:196:38 - Method Doctrine\DBAL\Connection::getactiveshardid does not exist (see https://psalm.dev/022)
        self::assertEquals(0, $conn->getActiveShardId());


ERROR: UndefinedMethod - tests/Doctrine/Tests/DBAL/Sharding/PoolingShardConnectionTest.php:199:38 - Method Doctrine\DBAL\Connection::getactiveshardid does not exist (see https://psalm.dev/022)
        self::assertEquals(1, $conn->getActiveShardId());


ERROR: UndefinedMethod - tests/Doctrine/Tests/DBAL/Sharding/PoolingShardConnectionTest.php:202:33 - Method Doctrine\DBAL\Connection::getactiveshardid does not exist (see https://psalm.dev/022)
        self::assertNull($conn->getActiveShardId());
```

greg0ire: I just asked for support here, just in case: https://symfony-devs.slack.com/archives/C8SFXTD2M/p1588144518359100

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
greg0ire: Ok, I will use `''` in tests that use `null` here.

greg0ire: If that feels wrong, maybe we should use `?string $database` when merging up.

morozov: `''` is better, it's at least of the correct type. `?string $database` is not an option. None of the implementations accept an optional DB name in `master` and this is done deliberately.

greg0ire:
@morozov the failure in Travis is caused by an issue that is fixed on 2.11 because `VersionAwarePlatformDriver` extends `Driver` on that branch. Should I create a short-lived test interface that extends both and use it in the test? I would remove it when merging up.

morozov:
> Should I create a short-lived test interface that extends both and use it in the test?

That should be fine.

lib/Doctrine/DBAL/DriverManager.php
greg0ire: This is going to get fixed: https://github.com/vimeo/psalm/issues/3277 :)

greg0ire: I have it working locally, but now PHPStan is unhappy about it :see_no_evil:

greg0ire: I pushed a new commit that include additional rules to ignor ethe PHPStan issues this creates.

greg0ire:
@morozov phpcs fails unless I use `@psalm-param` instead of `@param` when using array shapes. An alternative is to get `slevomat/coding-standard` to 6.3.3, which can be done by upgrading `phpstan/phpdoc-parser`, `doctrine/coding-standard` and `squizlabs/php_codesniffer`. Should I go ahead and create another PR that upgrades these packages? I might make other things easier.

lib/Doctrine/DBAL/DriverManager.php
```diff
@@ -115,11 +115,16 @@ private function __construct()
...
-     * @param Configuration|null $config       The configuration to use.
-     * @param EventManager|null  $eventManager The event manager to use.
+     * @param array{wrapperClass?: class-string<T>} $params
+     * @param Configuration|null                    $config       The configuration to use.
+     * @param EventManager|null                     $eventManager The event manager to use.
      *
      * @throws DBALException
+     *
+     * @phpstan-param mixed[] $params
+     * @phpstan-return Connection
```
morozov, line 125: This is getting ugly. You need to tell each SA what's happening in its own language. I hope it will force us to introduce simpler and cleaner APIs.

lib/Doctrine/DBAL/DriverManager.php
```diff
@@ -115,11 +115,16 @@ private function __construct()
      * <b>driverClass</b>:
      * The driver class to use.
      *
-     * @param mixed[]            $params       The parameters.
-     * @param Configuration|null $config       The configuration to use.
-     * @param EventManager|null  $eventManager The event manager to use.
+     * @param array{wrapperClass?: class-string<T>} $params
```
morozov, line 118: Can we turn this the other way around and use `@psalm-param` for this definition in order to not distract PHPStan that doesn't understand this?

lib/Doctrine/DBAL/DriverManager.php
morozov: Why is `@phpstan-return` needed? There's no `@return` and there is a return type declaration with the same type. Does PHPStan get confused with `@psalm-return` and needs an override?

morozov:
@greg0ire let's make the needed upgrades and see if we can simplify the annotations. The ones in the driver manager are really concerning.

lib/Doctrine/DBAL/DriverManager.php
greg0ire: Removing the first one causes this:

```
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   Doctrine/DBAL/Id/TableGenerator.php                                                                                                                                               
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  75     Parameter #1 $params of static method Doctrine\DBAL\DriverManager::getConnection() expects array()|array('wrapperClass' => class-string<Doctrine\DBAL\Connection>), array given.  
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 

 ------ -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   Doctrine/DBAL/Schema/SqliteSchemaManager.php                                                                                                                                                            
 ------ -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  59     Parameter #1 $params of static method Doctrine\DBAL\DriverManager::getConnection() expects array()|array('wrapperClass' => class-string<Doctrine\DBAL\Connection>), array('driver' => mixed, 'path' =>  
         string) given.                                                                                                                                                                                          
 ------ -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
```
I will remove `@phpstan-return`, it is not needed, it's a remnant of when I was trying to get phpstan to be quiet about that warning I added in the ignore list.

lib/Doctrine/DBAL/DriverManager.php
greg0ire: I will check, but PHPStan and Psalm read each others' annotation

greg0ire: As I feared, it does not work, I get the error messages I described at https://github.com/doctrine/dbal/pull/3977#discussion_r418601811

morozov: > I will check, but PHPStan and Psalm read each others' annotation

If they read each other's annotations but don't interpret equally then what's the point in that and having the `psalm-` and `phpstan-` prefixes? The difference in interpretation shouldn't be solved here.

greg0ire: I think the point is not to confuse other tools like phpcs+slevomat, and that they started with not interpreting each other's rules at the beginning.

greg0ire:
Down to 4 errors.

greg0ire:
Down to 2 errors, with which I need help please @morozov :pray: see the todo list in the original post for details.

morozov:
> https://github.com/doctrine/dbal/blob/155d028be084ef69110d09938c57c351a2126e5d/tests/Doctrine/Tests/DBAL/Functional/Driver/SQLSrv/StatementTest.php#L29-L30

IIRC, it just needs to be an invalid SQL that will cause failure being prepared. An empty string instead of `null` should work.

lib/Doctrine/DBAL/Schema/ForeignKeyConstraint.php
morozov: In https://github.com/doctrine/dbal/pull/2854, `null` was replaced with an empty string.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
```diff
@@ -33,8 +35,10 @@ public function lastInsertId($name = null)
 
         $stmt = $this->prepare('SELECT CONVERT(VARCHAR(MAX), current_value) FROM sys.sequences WHERE name = ?');
         $stmt->execute([$name]);
+        $lastInsertId = $stmt->fetchColumn();
+        assert(is_string($lastInsertId));
```
morozov, line 39: This doesn't look valid. If the sequence doesn't exist, the query will return `false`.

greg0ire: Not sure why I did this, reverting it doesn't make Psalm fail :thinking:

greg0ire: Ah wait that was Phpstan:

```
 ------ ------------------------------------------------------------------ 
  Line   Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php                     
 ------ ------------------------------------------------------------------ 
  37     Method Doctrine\DBAL\Driver\PDOSqlsrv\Connection::lastInsertId()  
         should return string but returns string|false|null.               
 ------ ------------------------------------------------------------------ 
```

greg0ire: Ok I will assert that it's either a string or `false`

morozov: It's better to assert that it's [not null](https://docs.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-sequences-transact-sql?view=sql-server-ver15) (see the column type).

tests/Doctrine/Tests/DbalPerformanceTestListener.php
```diff
@@ -16,7 +16,7 @@ class DbalPerformanceTestListener implements TestListener
 {
     use TestListenerDefaultImplementation;
 
-    /** @var string[][] */
+    /** @var float[][] */
```
SenseException, line 19:
```suggestion
    /** @var array<string, array<string, float>> */
```

tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
```diff
@@ -356,6 +362,8 @@ public function testGenerateTableWithAutoincrement() : void
 
     /**
      * {@inheritDoc}
+     *
+     * @psalm-return array
```
SenseException, line 366: Does Psalm complain about the return type given in the parent?

SenseException (changes requested):

morozov:
@greg0ire before this PR gets merged, does it need to be 33 commits long? I'd rather see the commits grouped into logical blocks that provide a bigger picture of why certain changes are made rather than “did this”, “did that”.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
greg0ire: This might be an LSP violation actually: this method should only ever return strings according to [the docs](https://www.php.net/manual/en/pdo.lastinsertid.php) (and also the phpstorm stubs)
Should we ignore this particular violation? Or should we return `'0'` if not a string?

tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
greg0ire: It does complain in my IDE, maybe that's a complain at a higher level? Anyway, I think I am going to remove this and use `never-return|string[]` in the parent, see https://github.com/phpstan/phpstan/issues/3335 . WDYT?

greg0ire: In fact, the only version I could get to work on both most recent versions of both tools is that one:
- https://psalm.dev/r/01f47ec513
- https://phpstan.org/r/87a19063-3ef0-4a8a-9c1c-301d662d0090

greg0ire:
> @greg0ire before this PR gets merged, does it need to be 33 commits long? I'd rather see the commits grouped into logical blocks that provide a bigger picture of why certain changes are made rather than “do this”, “do that”.

The big picture is : "fix issues detected by Psalm". That consists in many, many unrelated things, because I'm fixing 136 errors here. Some of them are trivial, some others require a github link, or an explanation, and should not be grouped with others.

tests/Doctrine/Tests/DbalPerformanceTestListener.php
greg0ire: Fixed

tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
greg0ire: Maybe I should report this: https://psalm.dev/r/01f47ec513

greg0ire: I asked for help at https://symfony-devs.slack.com/archives/C8SFXTD2M/p1590518449028600

greg0ire: Ok, here is a commit that should help with Psalm https://github.com/vimeo/psalm/commit/f0a5bd74b6f81c40393d45cd3490a5d2721509c2 :)

morozov:
@greg0ire since there's some work to do, can we do up-merges separately? Specifically, I'm interested in getting `2.11.x` → `3.0.x` to proceed with the removal of the deprecated fetch-related APIs.

greg0ire:
> @greg0ire since there's still some work to do, can we do up-merges separately? Specifically, I'm interested in getting 2.11.x → 3.0.x to proceed with the removal of the deprecated fetch-related APIs.

Yes, sure, my plan of using `--atomic` was a bit unrelatistic I'm afraid. I'm leaving my PRs open because some contains commits that are specific to some branches (new code requiring additional fixes). I don't regret creating them preemptively though, because now I have a better idea of the amount of work required to merge this up.

tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
greg0ire: I updated Psalm, and I can drop the commit about never return :tada: 

But now there are new issues to fix :see_no_evil:

greg0ire:
@morozov @SenseException I pushed a new version that drops the "never-return" commit, it is no longer needed with the latest version of Psalm. I had to push a commit that avoids mocking interfaces that do not exist, I'm not sure you're going to like it but it's either that or ignoring these errors, tell me what you prefer.

tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
SenseException: It was expected after an update :smile: 

Nice work.

SenseException (dismissed):
I'm in favour of the interfaces you provided. These are visible and can be removed once there is another solution for this. Does it makes sense to move those in separate files?

greg0ire:
> Does it makes sense to move those in separate files?

They don't need to be autoloaded from other test files, so I would say no, since it means less clutter in the directory.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
morozov: The documentation on PHP is really vague because it doesn't describe any non-happy paths except for "not implemented". Since it's `2.x`, let's suppress the error instead of modifying the behavior in any way.

tests/Doctrine/Tests/DBAL/FutureVersionAwarePlatformDriver.php
```diff
@@ -0,0 +1,15 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL;
+
+use Doctrine\DBAL\Driver;
+use Doctrine\DBAL\VersionAwarePlatformDriver;
+
+/**
+ * Remove me in 2.11
```
morozov, line 11: Should it be “in 3.0.x"? See https://github.com/doctrine/dbal/pull/3903.

tests/Doctrine/Tests/DBAL/Platforms/AbstractPlatformTestCase.php
```diff
@@ -1532,3 +1518,28 @@ public function testZeroOffsetWithoutLimitIsIgnored() : void
         );
     }
 }
+
+interface GetCreateTableSqlDispatchEventListener
```
morozov, line 1522: I hope one day it gets fixed in the event dispatcher so that instead of just calling random methods dynamically, it lets one define an interface and accepts its implementors as listeners. Otherwise, it looks barely manageable.

tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php
```diff
@@ -875,6 +881,8 @@ protected function getQuotedAlterTableRenameColumnSQL() : array
 
     /**
      * {@inheritdoc}
+     *
+     * @psalm-return never-return
```
morozov, line 885: Is it sill needed?

greg0ire: It isn't, I'll remove it, good catch!

morozov (approved):

greg0ire:
Finally!

pkruithof:
Sorry to comment on an older PR, but this commit causes phpstan to report an error: https://phpstan.org/r/e462e0f9-abc4-44bc-8a9a-80aca2717980

I've asked a support question there, and the author suggested a solution here: https://github.com/phpstan/phpstan/issues/3849#issuecomment-689517922

Would you accept a PR with the second solution? I don't think a third-party library should trigger phpstan errors in a project.

morozov:
@pkruithof please go ahead and submit a PR. @greg0ire could you work with PHPStan on finding the root cause as suggested in the first option?

pkruithof:
I just tried the suggestion but it did not help unfortunately, and I'm kind of out of my depth here. 😞

morozov:
@pkruithof let's start with filing a new issue that would reproduce the problem.

pkruithof:
Done 👉 #4264

PR #3978, opened 2020-04-28
Use composer install in the Psalm job

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

Our policy states that we do not want a new release to suddenly affect
our builds.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -23,11 +23,11 @@ jobs:
         uses: "actions/cache@v1.0.3"
         with:
           path: "~/.composer/cache"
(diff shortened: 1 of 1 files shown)


greg0ire:
Credits to @bendavies for noticing this. sql-formatter and coding-standards are also affected, I will probably open more PRs if this one goes through.

bendavies (commented):
The cache should hash on composer.lock

SenseException (approved):
Looks reasonable to me to use the lock file as long as we have it in git.

PR #3979, opened 2020-04-28
Move phpcs from Travis to GitHub Actions

bendavies:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Moves phpcs from Travis to GitHub Actions.

https://github.com/doctrine/dbal/pull/3978 needs merging first.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -14,7 +14,7 @@ jobs:
         uses: "actions/checkout@v2"
 
       - name: "Install PHP"
-        uses: "shivammathur/setup-php@1.8.1"
+        uses: "shivammathur/setup-php@v2"
         with:
           coverage: "none"
           php-version: "7.4"
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -6,12 +6,12 @@ on:
 
 jobs:
   static-analysis-psalm:
-    name: Static Analysis with Psalm
-    runs-on: ubuntu-latest
+    name: "Static Analysis with Psalm"
+    runs-on: "ubuntu-latest"
 
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v2
+      - name: "Checkout code"
+        uses: "actions/checkout@v2"
 
       - name: "Install PHP"
         uses: "shivammathur/setup-php@1.8.1"
@@ -29,5 +29,5 @@ jobs:
       - name: "Install dependencies with composer"
         run: "composer install --no-interaction --no-progress --no-suggest"
 
-      - name: Psalm
+      - name: "Psalm"
         run: "vendor/bin/psalm"
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -28,11 +28,44 @@ jobs:
         uses: "actions/cache@v1"
         with:
           path: "~/.composer/cache"
-          key: "${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
-          restore-keys: "${{ matrix.php-version }}-composer-locked-"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
 
       - name: "Install dependencies with composer"
         run: "composer install --no-interaction --no-progress --no-suggest"
 
       - name: "Psalm"
         run: "vendor/bin/psalm"
+
+  coding-standards:
+    name: "Coding Standards"
+    runs-on: "ubuntu-latest"
+
+    strategy:
+      matrix:
+        php-version:
+          - "7.4"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          coverage: "none"
+          php-version: "${{ matrix.php-version }}"
+          tools: "cs2pr"
+
+      - name: "Cache dependencies installed with composer"
+        uses: "actions/cache@v1"
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-interaction --no-progress --no-suggest"
+
+      - name: "Run squizlabs/php_codesniffer"
+        run: "vendor/bin/phpcs -q --no-colors --report=checkstyle | cs2pr"
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -20,7 +20,7 @@ jobs:
           php-version: "7.4"
 
       - name: "Cache dependencies installed with composer"
-        uses: "actions/cache@v1.0.3"
+        uses: "actions/cache@v1"
         with:
           path: "~/.composer/cache"
           key: "composer-${{ hashFiles('composer.lock') }}"
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -9,6 +9,11 @@ jobs:
     name: "Static Analysis with Psalm"
     runs-on: "ubuntu-latest"
 
+    strategy:
+      matrix:
+        php-version:
+          - "7.4"
+
     steps:
       - name: "Checkout code"
         uses: "actions/checkout@v2"
@@ -17,14 +22,14 @@ jobs:
         uses: "shivammathur/setup-php@v2"
         with:
           coverage: "none"
-          php-version: "7.4"
+          php-version: "${{ matrix.php-version }}"
 
       - name: "Cache dependencies installed with composer"
         uses: "actions/cache@v1"
         with:
           path: "~/.composer/cache"
-          key: "composer-${{ hashFiles('composer.lock') }}"
-          restore-keys: "composer-"
+          key: "${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "${{ matrix.php-version }}-composer-locked-"
 
       - name: "Install dependencies with composer"
         run: "composer install --no-interaction --no-progress --no-suggest"
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,11 +49,6 @@ jobs:
       env: PHPStan
       install: travis_retry composer install --prefer-dist
       script: vendor/bin/phpstan analyse
-    - stage: Smoke Testing
-      php: 7.3
-      env: PHP_CodeSniffer
-      install: travis_retry composer install --prefer-dist
-      script: vendor/bin/phpcs
 
     - stage: Test
       php: 7.2


.github/workflows/continuous-integration.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         uses: "actions/cache@v1"
         with:
           path: "~/.composer/cache"
-          key: "composer-${{ hashFiles('composer.json') }}"
+          key: "composer-${{ hashFiles('composer.lock') }}"
```
greg0ire: :thinking: I will copy that in #3978 I guess. This commit should probably go since `composer update` is still in use here.

.github/workflows/continuous-integration.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           restore-keys: "${{ matrix.php-version }}-composer-locked-"
 
       - name: "Install dependencies with composer"
-        run: "composer update --no-interaction --no-progress --no-suggest"
+        run: "composer install --no-interaction --no-progress --no-suggest"
```
greg0ire, line 35: Should be fixed up with previous commit

.github/workflows/continuous-integration.yml
```diff
@@ -36,3 +36,36 @@ jobs:
...
         run: "vendor/bin/psalm"
+
+  coding-standards:
+    name: "Coding Standards"
+    runs-on: "ubuntu-latest"
+
+    strategy:
+      matrix:
+        php-version:
+          - "7.2"
```
greg0ire, line 47: Now that you fixed the issue with 7.4 locally, maybe we can use 7.4 here too?

bendavies: well, the issue seems to be in vendors, but let's try.

.github/workflows/continuous-integration.yml
bendavies: done, but will removed once your PR is merged.

.github/workflows/continuous-integration.yml
bendavies: seems fine

greg0ire:
#3978 was merged :)

phpcs.xml.dist
```diff
@@ -2,7 +2,7 @@
 <ruleset>
     <arg name="basepath" value="."/>
     <arg name="extensions" value="php"/>
-    <arg name="parallel" value="80"/>
+    <arg name="parallel" value="8"/>
```
carusogabriel, line 5: Why this change?

bendavies: It seems to me that starting 80 processes is a typo copied from somewhere else. Maybe 8 processes was the original intention.

greg0ire: Looks like phpcs itself is run with a similar value: https://github.com/squizlabs/PHP_CodeSniffer/commit/a57331c6c033ef3c947415898b9921a4237289cf

bendavies: Happy to revert. Just seems odd.

greg0ire: Did you see my edit?

carusogabriel: Interesting. I'll research about `parallel` and `phpcs` to see if this change is right because then we'd need to propagate it to the other Doctrine projects. Thanks for taking a look into that :)

bendavies: shall i revert? normally tools like this set the number of parallel process to the number of available cores, or something like that. 80 just seems wild. not sure why PHPCS is using 75 either.

greg0ire: > normally tools like this set the number of parallel process to the number of available cores

That makes sense to me if your process is CPU bound, but if the process spends 90 percent of its time actually being IO bound, it would make sense to me to multiply the number of cores by some factor, so that the scheduler can give the CPU to a process that is no longer waiting for IO. But again, maybe what I'm saying does not make sense. Can you try locally and tell us what figure results in the fastest processing time for you?

bendavies: I've reverted so we can get this merged but:
```
➜  dbal git:(phpcs-github-actions) ✗ time vendor/bin/phpcs --no-cache --parallel=8
........ 8 / 8 (100%)


Time: 12.85 secs; Memory: 8MB

vendor/bin/phpcs --no-cache --parallel=8  51.49s user 3.37s system 423% cpu 12.945 total
➜  dbal git:(phpcs-github-actions) ✗ time vendor/bin/phpcs --no-cache --parallel=1
............................................................  60 / 456 (13%)
............................................................ 120 / 456 (26%)
............................................................ 180 / 456 (39%)
............................................................ 240 / 456 (53%)
............................................................ 300 / 456 (66%)
............................................................ 360 / 456 (79%)
............................................................ 420 / 456 (92%)
....................................                         456 / 456 (100%)


Time: 28.9 secs; Memory: 34MB

vendor/bin/phpcs --no-cache --parallel=1  28.56s user 0.23s system 99% cpu 28.986 total
➜  dbal git:(phpcs-github-actions) ✗ time vendor/bin/phpcs --no-cache --parallel=8
........ 8 / 8 (100%)


Time: 12.7 secs; Memory: 8MB

vendor/bin/phpcs --no-cache --parallel=8  49.23s user 3.21s system 410% cpu 12.780 total
➜  dbal git:(phpcs-github-actions) ✗ time vendor/bin/phpcs --no-cache --parallel=80
............................................................ 60 / 76 (79%)
................                                             76 / 76 (100%)


Time: 9.2 secs; Memory: 8MB

vendor/bin/phpcs --no-cache --parallel=80  47.74s user 2.92s system 546% cpu 9.264 total
```

greg0ire: Indeed almost no gain, so this is probably more CPU-bound nowadays.

greg0ire (approved):

.github/workflows/continuous-integration.yml
```diff
@@ -6,28 +6,65 @@ on:
...
+        with:
+          path: "~/.composer/cache"
+          key: "${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-interaction --no-progress --no-suggest"
+
+      - name: "Run squizlabs/php_codesniffer"
+        run: "vendor/bin/phpcs"
```
morozov, line 70: Most likely, you'll need to add `--no-colors` here (see https://github.com/mnapoli/pretty/pull/22).

greg0ire (changes requested):
Please add `cs2pr`, I noticed @beberlei did that on ORM (master branch). I'm doing so in a PR to 2.7: https://github.com/doctrine/orm/pull/8116

.github/workflows/continuous-integration.yml
```diff
@@ -6,28 +6,66 @@ on:
...
 
       - name: "Cache dependencies installed with composer"
-        uses: "actions/cache@v1.0.3"
+        uses: "actions/cache@v1"
         with:
           path: "~/.composer/cache"
-          key: "composer-${{ hashFiles('composer.lock') }}"
-          restore-keys: "composer-"
+          key: "${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "${{ matrix.php-version }}-composer-locked-"
```
greg0ire, line 32:
```suggestion
          key: " php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
```

.github/workflows/continuous-integration.yml
```diff
@@ -6,28 +6,66 @@ on:
...
+          coverage: "none"
+          php-version: "${{ matrix.php-version }}"
+          tools: "cs2pr"
+
+      - name: "Cache dependencies installed with composer"
+        uses: "actions/cache@v1"
+        with:
+          path: "~/.composer/cache"
+          key: "${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "${{ matrix.php-version }}-composer-locked-"
```
greg0ire, line 65:
```suggestion
          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
```

greg0ire (dismissed):
:ok_hand:

morozov:
Is there a way to see how `cs2pr` works in this repository before merging the changes? E.g. push a temporary commit that would violate the coding standard.

.github/workflows/continuous-integration.yml
```diff
@@ -6,28 +6,66 @@ on:
 
 jobs:
   static-analysis-psalm:
-    name: Static Analysis with Psalm
-    runs-on: ubuntu-latest
+    name: "Static Analysis with Psalm"
```
morozov, line 9: Are changes like this really needed? I'd rather keep it simple and only add quotes where required (e.g. strings with special characters, numeric strings that must remain strings, etc.).

This file will be maintained by humans who will not see the need to add quotes everywhere.

.github/workflows/continuous-integration.yml
```diff
@@ -6,28 +6,66 @@ on:
...
+          php-version: "${{ matrix.php-version }}"
 
       - name: "Cache dependencies installed with composer"
-        uses: "actions/cache@v1.0.3"
+        uses: "actions/cache@v1"
         with:
           path: "~/.composer/cache"
-          key: "composer-${{ hashFiles('composer.lock') }}"
-          restore-keys: "composer-"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
```
morozov, line 31: This looks like an unnecessary complication. What's the scenario where we're going to run the code style check using different PHP versions?

.github/workflows/continuous-integration.yml
```diff
@@ -6,28 +6,66 @@ on:
...
-      - name: Psalm
+      - name: "Psalm"
         run: "vendor/bin/psalm"
+
+  coding-standards:
+    name: "Coding Standards"
+    runs-on: "ubuntu-latest"
+
+    strategy:
+      matrix:
```
morozov, line 45: I believe the whole point of static analysis is that it's static and doesn't depend on the runtime environment. Having a matrix here looks like overkill to me.

.github/workflows/continuous-integration.yml
bendavies: It is trying to maintain consistency across repos
https://github.com/doctrine/orm/pull/8116/files

.github/workflows/continuous-integration.yml
bendavies: It is trying to maintain consistency across repos
https://github.com/doctrine/orm/pull/8116/files

.github/workflows/continuous-integration.yml
morozov: So what's the reason why we want to have all strings quoted in all repositories?

.github/workflows/continuous-integration.yml
morozov: Could you point me to the thread, please? I could find the discussion regarding the build matrix.

.github/workflows/continuous-integration.yml
bendavies: i suspect inspiration from https://github.com/ergebnis/php-library-template/blob/master/.github/workflows/integrate.yaml

.github/workflows/continuous-integration.yml
bendavies: https://github.com/doctrine/orm/pull/8116#discussion_r418949951

.github/workflows/continuous-integration.yml
bendavies: it's just coding style, and that repo enforces it:
https://github.com/ergebnis/php-library-template/blob/master/.yamllint.yaml

I personally don't like everything quoted, as you point out it's not necessary all the time. but being consistent is the main thing i guess.

cc @localheinz

.github/workflows/continuous-integration.yml
bendavies: sorry, let me elaborate, because that link doesn't help you.

I think, even if the matrix has a single php version, it enables us to use the same `key` everywhere, which helps cache hits. rather than maintaining 
`key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"` where a matrix is used and
`key: "php-7.4-composer-locked-${{ hashFiles('composer.lock') }}"`
where a matrix is not used.

.github/workflows/continuous-integration.yml
localheinz: @morozov 

I am using [`ibiqlik/action-yamllint`](https://github.com/ibiqlik/action-yamllint) to lint YAML files using [`yamllint`](https://github.com/adrienverge/yamllint), and for consistency, I quote YAML strings.

As @bendavies pointed out, it's a matter of taste and consistency. I have to admit that I dislike having to escape a bunch of characters when quoting.

.github/workflows/continuous-integration.yml
morozov: So, by “everywhere” you mean that if named consistently, the same cache entry can be shared between multiple actions? Makes sense in this case.

morozov:
Looks good. Should we test https://github.com/doctrine/dbal/pull/3979#issuecomment-624343839 and squash everything before the merge?

.github/workflows/continuous-integration.yml
bendavies: Multiple action in the same repo yes. Thanks for phrasing it better!

bendavies:
Sure. I can do it later, or feel free to push to my branch if you are available now.

greg0ire:
Screenshot_2020-05-06 Move phpcs from Travis to GitHub Actions by bendavies · Pull Request #3979 · doctrine dbal

greg0ire (approved):

greg0ire:
After merging up all the way until master, let's add this check as well as Psalm as required, shall we?

greg0ire:
Thanks a lot @bendavies !

PR #3980, opened 2020-04-29
Segregated support of unique index and unique constraint

guilhermeblanco:
Added missing commits: 

* https://github.com/doctrine/dbal/commit/400d73c8a1e4f53459a877c8652811ee22c0913b
* https://github.com/doctrine/dbal/commit/e91811cb3fe3101d7d8ac95c45f8bc1ba9bd913a
* https://github.com/doctrine/dbal/commit/a4ecf053186b11bc31fbff79ef36248500383908
* https://github.com/doctrine/dbal/commit/d7d1c397a35735383b34632d3a67bbc3a3371a45
* https://github.com/doctrine/dbal/commit/c71b3ae0425e5c039b047825dbf8351a773e6785
* https://github.com/doctrine/dbal/commit/2f9da39ec645b54d08c95649c99d7b87a32bdc72
* https://github.com/doctrine/dbal/commit/9aa4e79aab0477b4212b01d65de53193d9559e39

## BC BREAK: Doctrine\DBAL\Schema\Table constructor new parameter

Deprecated parameter `$idGeneratorType` removed and added a new parameter `$uniqueConstraints`.
Constructor changed from:

`__construct($tableName, array $columns = [], array $indexes = [], array $fkConstraints = [], $idGeneratorType = 0, array $options = [])`

To the new constructor:

`__construct($tableName, array $columns = [], array $indexes = [], array $uniqueConstraints = [], array $fkConstraints = [], array $options = [])`

### DIFF
--- a/src/Platforms/AbstractPlatform.php
+++ b/src/Platforms/AbstractPlatform.php
@@ -23,6 +23,7 @@ use Doctrine\DBAL\Schema\Index;
 use Doctrine\DBAL\Schema\Sequence;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Schema\TableDiff;
+use Doctrine\DBAL\Schema\UniqueConstraint;
 use Doctrine\DBAL\TransactionIsolationLevel;
 use Doctrine\DBAL\Types;
 use Doctrine\DBAL\Types\Type;
@@ -1495,12 +1496,26 @@ abstract class AbstractPlatform
 
         if (($createFlags & self::CREATE_INDEXES) > 0) {
             foreach ($table->getIndexes() as $index) {
-                if ($index->isPrimary()) {
-                    $options['primary']       = $index->getQuotedColumns($this);
-                    $options['primary_index'] = $index;
-                } else {
+                if (! $index->isPrimary()) {
                     $options['indexes'][$index->getQuotedName($this)] = $index;
+
+                    continue;
                 }
+
+                $options['primary']       = $index->getQuotedColumns($this);
+                $options['primary_index'] = $index;
+            }
+
+            foreach ($table->getUniqueConstraints() as $uniqueConstraint) {
+                $options['uniqueConstraints'][$uniqueConstraint->getQuotedName($this)] = $uniqueConstraint;
+            }
+        }
+
+        if (($createFlags & self::CREATE_FOREIGNKEYS) > 0) {
+            $options['foreignKeys'] = [];
+
+            foreach ($table->getForeignKeys() as $fkConstraint) {
+                $options['foreignKeys'][] = $fkConstraint;
             }
         }
 
@@ -1513,6 +1528,7 @@ abstract class AbstractPlatform
                 && $this->_eventManager->hasListeners(Events::onSchemaCreateTableColumn)
             ) {
                 $eventArgs = new SchemaCreateTableColumnEventArgs($column, $table, $this);
+
                 $this->_eventManager->dispatchEvent(Events::onSchemaCreateTableColumn, $eventArgs);
 
                 $columnSql = array_merge($columnSql, $eventArgs->getSql());
@@ -1539,15 +1555,9 @@ abstract class AbstractPlatform
             $columns[$columnData['name']] = $columnData;
         }
 
-        if (($createFlags & self::CREATE_FOREIGNKEYS) > 0) {
-            $options['foreignKeys'] = [];
-            foreach ($table->getForeignKeys() as $fkConstraint) {
-                $options['foreignKeys'][] = $fkConstraint;
-            }
-        }
-
         if ($this->_eventManager !== null && $this->_eventManager->hasListeners(Events::onSchemaCreateTable)) {
             $eventArgs = new SchemaCreateTableEventArgs($table, $columns, $options, $this);
+
             $this->_eventManager->dispatchEvent(Events::onSchemaCreateTable, $eventArgs);
 
             if ($eventArgs->isDefaultPrevented()) {
@@ -1556,6 +1566,7 @@ abstract class AbstractPlatform
         }
 
         $sql = $this->_getCreateTableSQL($tableName, $columns, $options);
+
         if ($this->supportsCommentOnStatement()) {
             if ($table->hasOption('comment')) {
                 $sql[] = $this->getCommentOnTableSQL($tableName, $table->getOption('comment'));
@@ -1654,8 +1665,8 @@ abstract class AbstractPlatform
         }
 
         $query = 'CREATE TABLE ' . $name . ' (' . $columnListSql;
-
         $check = $this->getCheckDeclarationSQL($columns);
+
         if (! empty($check)) {
             $query .= ', ' . $check;
         }
@@ -2307,25 +2318,27 @@ abstract class AbstractPlatform
      * Obtains DBMS specific SQL code portion needed to set a unique
      * constraint declaration to be used in statements like CREATE TABLE.
      *
-     * @param string $name  The name of the unique constraint.
-     * @param Index  $index The index definition.
+     * @param string           $name       The name of the unique constraint.
+     * @param UniqueConstraint $constraint The unique constraint definition.
      *
      * @return string DBMS specific SQL code portion needed to set a constraint.
      *
      * @throws InvalidArgumentException
      */
-    public function getUniqueConstraintDeclarationSQL($name, Index $index)
+    public function getUniqueConstraintDeclarationSQL($name, UniqueConstraint $constraint)
     {
-        $columns = $index->getColumns();
+        $columns = $constraint->getQuotedColumns($this);
         $name    = new Identifier($name);
 
         if (count($columns) === 0) {
             throw new InvalidArgumentException("Incomplete definition. 'columns' required.");
         }
 
-        return 'CONSTRAINT ' . $name->getQuotedName($this) . ' UNIQUE ('
-            . $this->getColumnsFieldDeclarationListSQL($columns)
-            . ')' . $this->getPartialIndexSQL($index);
+        $constraintFlags = array_merge(['UNIQUE'], array_map('strtoupper', $constraint->getFlags()));
+        $constraintName  = $name->getQuotedName($this);
+        $columnListNames = $this->getColumnsFieldDeclarationListSQL($columns);
+
+        return sprintf('CONSTRAINT %s %s (%s)', $constraintName, implode(' ', $constraintFlags), $columnListNames);
     }
 
     /**
@@ -2348,9 +2361,8 @@ abstract class AbstractPlatform
             throw new InvalidArgumentException("Incomplete definition. 'columns' required.");
         }
 
-        return $this->getCreateIndexSQLFlags($index) . 'INDEX ' . $name->getQuotedName($this) . ' ('
-            . $this->getIndexFieldDeclarationListSQL($index)
-            . ')' . $this->getPartialIndexSQL($index);
+        return $this->getCreateIndexSQLFlags($index) . 'INDEX ' . $name->getQuotedName($this)
+            . ' (' . $this->getIndexFieldDeclarationListSQL($index) . ')' . $this->getPartialIndexSQL($index);
     }
 
     /**
@@ -3022,7 +3034,7 @@ abstract class AbstractPlatform
     /**
      * Returns the name of the sequence for a particular identity column in a particular table.
      *
-     * @see    usesSequenceEmulatedIdentityColumns
+     * @see usesSequenceEmulatedIdentityColumns
      *
      * @param string $tableName  The name of the table to return the sequence name for.
      * @param string $columnName The name of the identity column in the table to return the sequence name for.
--- a/src/Platforms/SqlitePlatform.php
+++ b/src/Platforms/SqlitePlatform.php
@@ -963,8 +963,8 @@ class SqlitePlatform extends AbstractPlatform
                 $table->getQuotedName($this),
                 $columns,
                 $this->getPrimaryIndexInAlteredTable($diff),
+                [],
                 $this->getForeignKeysInAlteredTable($diff),
-                0,
                 $table->getOptions()
             );
             $newTable->addOption('alter', true);
--- a/src/Schema/Index.php
+++ b/src/Schema/Index.php
@@ -11,7 +11,6 @@ use function array_map;
 use function array_search;
 use function array_shift;
 use function count;
-use function is_string;
 use function strtolower;
 
 class Index extends AbstractAsset implements Constraint
@@ -79,18 +78,10 @@ class Index extends AbstractAsset implements Constraint
     }
 
     /**
-     * @param string $column
-     *
-     * @return void
-     *
      * @throws InvalidArgumentException
      */
-    protected function _addColumn($column)
+    protected function _addColumn(string $column): void
     {
-        if (! is_string($column)) {
-            throw new InvalidArgumentException('Expecting a string as Index Column');
-        }
-
         $this->_columns[$column] = new Identifier($column);
     }
 
--- a/src/Schema/SchemaException.php
+++ b/src/Schema/SchemaException.php
@@ -22,7 +22,8 @@ class SchemaException extends DBALException
     public const SEQUENCE_ALREADY_EXISTS  = 80;
     public const INDEX_INVALID_NAME       = 90;
     public const FOREIGNKEY_DOESNT_EXIST  = 100;
-    public const NAMESPACE_ALREADY_EXISTS = 110;
+    public const CONSTRAINT_DOESNT_EXIST  = 110;
+    public const NAMESPACE_ALREADY_EXISTS = 120;
 
     /**
      * @param string $tableName
@@ -144,6 +145,20 @@ class SchemaException extends DBALException
     public static function sequenceDoesNotExist($name)
     {
         return new self("There exists no sequence with the name '" . $name . "'.", self::SEQUENCE_DOENST_EXIST);
+    }
+
+    /**
+     * @param string $constraintName
+     * @param string $table
+     *
+     * @return SchemaException
+     */
+    public static function uniqueConstraintDoesNotExist($constraintName, $table)
+    {
+        return new self(
+            sprintf('There exists no unique constraint with the name "%s" on table "%s".', $constraintName, $table),
+            self::CONSTRAINT_DOESNT_EXIST
+        );
     }
 
     /**
--- a/src/Schema/Table.php
+++ b/src/Schema/Table.php
@@ -51,7 +51,6 @@ class Table extends AbstractAsset
     private $implicitIndexes = [];
 
     /**
-     * @param string                 $name
      * @param Column[]               $columns
      * @param Index[]                $indexes
      * @param UniqueConstraint[]     $uniqueConstraints
@@ -61,14 +60,14 @@ class Table extends AbstractAsset
      * @throws SchemaException
      */
     public function __construct(
-        $name,
+        string $name,
         array $columns = [],
         array $indexes = [],
         array $uniqueConstraints = [],
         array $fkConstraints = [],
         array $options = []
     ) {
-        if (strlen($name) === 0) {
+        if ($name === '') {
             throw InvalidTableName::new($name);
         }
 
--- a/src/Schema/Table.php
+++ b/src/Schema/Table.php
@@ -8,8 +8,11 @@ use Doctrine\DBAL\Schema\Visitor\Visitor;
 use Doctrine\DBAL\Types\Type;
 
 use function array_filter;
+use function array_keys;
 use function array_merge;
 use function in_array;
+use function is_numeric;
+use function is_string;
 use function preg_match;
 use function strlen;
 use function strtolower;
@@ -25,13 +28,13 @@ class Table extends AbstractAsset
     protected $_columns = [];
 
     /** @var Index[] */
-    private $implicitIndexes = [];
-
-    /** @var Index[] */
     protected $_indexes = [];
 
-    /** @var string|false */
-    protected $_primaryKeyName = false;
+    /** @var string|null */
+    protected $_primaryKeyName = null;
+
+    /** @var UniqueConstraint[] */
+    protected $uniqueConstraints = [];
 
     /** @var ForeignKeyConstraint[] */
     protected $_fkConstraints = [];
@@ -44,12 +47,15 @@ class Table extends AbstractAsset
     /** @var SchemaConfig|null */
     protected $_schemaConfig = null;
 
+    /** @var Index[] */
+    private $implicitIndexes = [];
+
     /**
      * @param string                 $name
      * @param Column[]               $columns
      * @param Index[]                $indexes
+     * @param UniqueConstraint[]     $uniqueConstraints
      * @param ForeignKeyConstraint[] $fkConstraints
-     * @param int                    $idGeneratorType
      * @param mixed[]                $options
      *
      * @throws SchemaException
@@ -58,8 +64,8 @@ class Table extends AbstractAsset
         $name,
         array $columns = [],
         array $indexes = [],
+        array $uniqueConstraints = [],
         array $fkConstraints = [],
-        $idGeneratorType = 0,
         array $options = []
     ) {
         if (strlen($name) === 0) {
@@ -74,6 +80,10 @@ class Table extends AbstractAsset
 
         foreach ($indexes as $idx) {
             $this->_addIndex($idx);
+        }
+
+        foreach ($uniqueConstraints as $uniqueConstraint) {
+            $this->_addUniqueConstraint($uniqueConstraint);
         }
 
         foreach ($fkConstraints as $constraint) {
@@ -130,16 +140,15 @@ class Table extends AbstractAsset
     }
 
     /**
-     * @param string[]    $columnNames
-     * @param string|null $indexName
-     * @param string[]    $flags
-     * @param mixed[]     $options
+     * @param string[] $columnNames
+     * @param string[] $flags
+     * @param mixed[]  $options
      *
      * @return self
      *
      * @throws SchemaException
      */
-    public function addIndex(array $columnNames, $indexName = null, array $flags = [], array $options = [])
+    public function addIndex(array $columnNames, ?string $indexName = null, array $flags = [], array $options = [])
     {
         if ($indexName === null) {
             $indexName = $this->_generateIdentifierName(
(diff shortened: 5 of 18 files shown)


lcobucci:
@morozov should be the one checking this. However, as it was mentioned on Slack, I'm pretty sure you should be looking at using DBAL 4.x instead of 3.x

src/Schema/Table.php
```diff
@@ -835,34 +791,245 @@ public function __clone()
...
-    public function getComment() : ?string
+    /**
+     * @return self
+     */
+    protected function _addForeignKeyConstraint(ForeignKeyConstraint $constraint)
     {
-        return $this->_options['comment'] ?? null;
+        $constraint->setLocalTable($this);
+
+        if (strlen($constraint->getName()) > 0) {
```
SenseException, line 915:
```suggestion
        if ($constraint->getName() !== '') {
```

src/Schema/Table.php
```diff
@@ -835,34 +791,245 @@ public function __clone()
...
+    }
+
+    /**
+     * Normalizes a given identifier.
+     *
+     * Trims quotes and lowercases the given identifier.
+     *
+     * @return string The normalized identifier.
+     */
+    private function normalizeIdentifier(?string $identifier)
```
SenseException, line 1013:
```suggestion
    private function normalizeIdentifier(?string $identifier) : string
```

src/Schema/Table.php
```diff
@@ -835,34 +791,245 @@ public function __clone()
...
+    }
+
+    /**
+     * Returns only columns that have specified names
+     *
+     * @param string[] $columnNames
+     *
+     * @return Column[]
+     */
+    private function filterColumns(array $columnNames, bool $reverse = false)
```
SenseException, line 1029:
```suggestion
    private function filterColumns(array $columnNames, bool $reverse = false) : array
```

src/Schema/Index.php
```diff
@@ -58,31 +57,27 @@ public function __construct($indexName, array $columns, $isUnique = false, $isPr
...
 
     /**
-     * @param string $column
-     *
      * @return void
      *
      * @throws InvalidArgumentException
      */
-    protected function _addColumn($column)
+    protected function _addColumn(string $column)
```
SenseException, line 79:
```suggestion
    protected function _addColumn(string $column) : void
```

SenseException (changes requested):
* Can we remove underscore prefix in method and property names at least in new classes?
* Since this is a major release branch that had no release yet, can return values be added to methods (at least those with scopes less than public)?
* As mentioned in a comment it looks like there are missing tests.

src/Platforms/AbstractPlatform.php
```diff
@@ -3078,10 +3097,30 @@ public function usesSequenceEmulatedIdentityColumns()
...
 
+    /**
+     * Gets the sequence name prefix based on table information.
+     *
+     * @param string      $tableName
+     * @param string|null $schemaName
+     *
+     * @return string
+     */
+    public function getSequencePrefix($tableName, $schemaName = null)
```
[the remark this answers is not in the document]
guilhermeblanco: This is being used in ORM, and was conceptually supposed to be part of sequence name resolution/definition.

Usage:
* https://github.com/doctrine/orm/blob/master/lib/Doctrine/ORM/Mapping/Builder/ValueGeneratorMetadataBuilder.php#L139
* https://github.com/doctrine/orm/blob/master/tests/Doctrine/Tests/ORM/Functional/Ticket/DDC2825Test.php#L62

src/Schema/Table.php
```diff
@@ -835,34 +791,245 @@ public function __clone()
...
-    public function setComment(?string $comment) : self
+    /**
+     * @return self
+     */
+    protected function _addUniqueConstraint(UniqueConstraint $constraint)
     {
-        // For keeping backward compatibility with MySQL in previous releases, table comments are stored as options.
-        $this->addOption('comment', $comment);
+        $mergedNames = array_merge([$this->getName()], $constraint->getColumns());
+        $name        = strlen($constraint->getName()) > 0
```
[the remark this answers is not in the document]
guilhermeblanco: Constraint name can be null at this stage, so `!== ''` wouldn't work here.

src/Schema/Table.php
guilhermeblanco: This method was being moved to follow the logical class declaration order:
* public constants
* protected constants
* private constants
* public static attributes
* protected static attributes
* private static attributes
* public attributes
* protected attributes
* private attributes
* constructor
* static public methods
* static protected methods
* static private methods
* magic methods
* public methods
* protected methods
* private methods
No work was done here.

guilhermeblanco:
> * Can we remove underscore prefix in method and property names at least in new classes?

There is only one class added to project (`UniqueConstraint`). I've fixed the only method that was using underscore.

> * Since this is a major release branch that had no release yet, can return values be added to methods (at least those with scopes less than public)?

I've moved a lot of methods following logical declarations. I only added 2 public methods and added type hint and return types there.

> * As mentioned in a comment it looks like there are missing tests.

Fixed that adding tests to `AbstractPlatform::getSequencePrefix()`

.gitignore
```diff
@@ -1,3 +1,4 @@
+.idea
```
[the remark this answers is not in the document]
greg0ire: I'd rather we use [a global gitinore file](https://greg0ire.fr/git-gud/#/23)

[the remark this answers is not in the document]
guilhermeblanco: Fixed

src/Platforms/AbstractPlatform.php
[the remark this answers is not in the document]
guilhermeblanco: Added tests

src/Schema/Table.php
SenseException: This was due to consistency to a previous of my comments. Because my suggested change wasn't needed and therefore not applied, this one also can stay as it is.

SenseException (dismissed):
Looks good. :+1:  Only the coding style issues seem to prevent a full build.

src/Platforms/AbstractPlatform.php
```diff
@@ -1550,15 +1551,30 @@ public function getCreateTableSQL(Table $table, $createFlags = self::CREATE_INDE
...
+
+                    continue;
                 }
+
+                $options['primary']       = $index->getQuotedColumns($this);
+                $options['primary_index'] = $index;
+            }
+
+            foreach ($table->getUniqueConstraints() as $uniqueConstraint) {
+                /** @var UniqueConstraint $uniqueConstraint */
```
morozov, line 1568: Given the type of the collection being iterated over, is this type hint needed?

https://github.com/doctrine/dbal/blob/de1f02b628082bf3956d1d64941b4480efc81be1/src/Schema/Table.php#L710-L712

src/Platforms/AbstractPlatform.php
```diff
@@ -2365,25 +2377,33 @@ public function getCheckDeclarationSQL(array $definition)
...
-            . $this->getIndexFieldDeclarationListSQL($index)
-            . ')' . $this->getPartialIndexSQL($index);
+        $flags = ['UNIQUE'];
+
+        if ($constraint->hasFlag('clustered')) {
+            $flags[] = 'CLUSTERED';
+        }
+
+        $constraintName  = $name->getQuotedName($this);
+        $constraintName  = ! empty($constraintName) ? $constraintName . ' ' : '';
```
morozov, line 2403: The trailing whitespace is not part of the constraint name. Would it make sense to rename `$flags` to something like `$tokens`, push the name there and then join everything?

src/Platforms/AbstractPlatform.php
morozov, line 3108: Please use type declarations for new APIs.

src/Platforms/AbstractPlatform.php
morozov: Please avoid `empty()`. It's forbidden by PHPStan in master.

src/Platforms/OraclePlatform.php
```diff
@@ -380,11 +380,11 @@ public function getListSequencesSQL($database)
     /**
      * {@inheritDoc}
      */
-    protected function _getCreateTableSQL($table, array $columns, array $options = [])
+    protected function _getCreateTableSQL($tableName, array $columns, array $options = [])
```
morozov, line 383: Would it make sense to use `$name` for the table name argument? `$columns` and `$options` also belong to the table but don't have this prefix.

src/Platforms/SqlitePlatform.php
```diff
@@ -319,9 +319,9 @@ public function getForeignKeyDeclarationSQL(ForeignKeyConstraint $foreignKey)
     /**
      * {@inheritDoc}
      */
-    protected function _getCreateTableSQL($name, array $columns, array $options = [])
+    protected function _getCreateTableSQL($tableName, array $columns, array $options = [])
```
morozov, line 322: Please revert.

src/Schema/SchemaException.php
```diff
@@ -145,6 +146,20 @@ public static function sequenceDoesNotExist($sequenceName)
...
+    /**
+     * @param string $constraintName
+     * @param string $table
+     *
+     * @return \Doctrine\DBAL\Schema\SchemaException
+     */
+    public static function uniqueConstraintDoesNotExist($constraintName, $table)
+    {
+        return new self(
+            sprintf("There exists no unique constraint with the name '%s' on table '%s'.", $constraintName, $table),
```
morozov, line 158: Please use double quotes to quote names (see https://github.com/doctrine/dbal/pull/3536).

src/Schema/Table.php
```diff
@@ -21,14 +24,14 @@ class Table extends AbstractAsset
...
     /** @var Index[] */
     protected $_indexes = [];
 
-    /** @var string|false */
-    protected $_primaryKeyName = false;
+    /** @var string|null */
+    protected $_primaryKeyName = null;
+
+    /** @var UniqueConstraint[] */
+    protected $_uniqueConstraints = [];
```
morozov, line 34: This should not have the leading underscore.

src/Schema/Table.php
```diff
@@ -41,18 +44,27 @@ class Table extends AbstractAsset
...
      * @param ForeignKeyConstraint[] $fkConstraints
-     * @param int                    $idGeneratorType
      * @param mixed[]                $options
      *
      * @throws DBALException
      */
-    public function __construct($tableName, array $columns = [], array $indexes = [], array $fkConstraints = [], $idGeneratorType = 0, array $options = [])
-    {
+    public function __construct(
+        $tableName,
```
morozov, line 61: Should this be `string $name`?

src/Schema/Table.php
```diff
@@ -139,19 +142,39 @@ public function addIndex(array $columnNames, $indexName = null, array $flags = [
...
     }
 
+    /**
+     * @param string[] $columnNames
+     * @param string[] $flags
+     * @param mixed[]  $options
+     *
+     * @return self
+     */
+    public function addUniqueConstraint(array $columnNames, ?string $indexName = null, array $flags = [], array $options = [])
```
morozov, line 152: This could use a return type hint.

src/Schema/Table.php
```diff
@@ -597,49 +503,74 @@ public function removeForeignKey($constraintName)
...
-     * Returns ordered list of columns (primary keys are first, then foreign keys, then the rest)
+     * Returns whether this table has a unique constraint with the given name.
      *
-     * @return Column[]
+     * @param string $constraintName
+     *
+     * @return bool
      */
-    public function getColumns()
+    public function hasUniqueConstraint($constraintName)
```
morozov, line 512: `(string $name) : bool`

morozov (commented):
Since this is a new API and a breaking change, I'd expect thew new API to be documented in the PR description and the BC break documented in upgrade notes.

When we decided to release DBAL 3.0 earlier for compatibility with PHP 8, we agreed that it should only have trivial to handle breaking changes. There doesn't seem to be much BC break here, so it should be good IMO.

@guilhermeblanco please make sure all new methods have native parameter and return type declarations.

src/Schema/SchemaException.php
guilhermeblanco: All other methods here are using single quotes, I'm just following the standard.

src/Schema/Table.php
guilhermeblanco: This might break somewhere else. No worth doing it here as part of this PR.

src/Platforms/OraclePlatform.php
guilhermeblanco: I've improved the code into something better. Marking as resolved.

guilhermeblanco:
@morozov all updated. Please merge if appropriate.

src/Platforms/OraclePlatform.php
morozov: What do you mean?

src/Schema/SchemaException.php
morozov: We need to make sure that the standard set in master is followed as well during merging up.

src/Schema/Table.php
morozov: Let's agree that the starter of the discussion resolves it. The very first instruction of the method checks the length of the table name. What else than a string can it be to work? Do you expect consumers to pass an object that implements `__toString()`?

src/Schema/Table.php
```diff
@@ -682,17 +602,15 @@ public function getColumn($columnName)
      */
     public function getPrimaryKey()
     {
-        if ($this->_primaryKeyName !== false) {
-            return $this->getIndex($this->_primaryKeyName);
-        }
-
-        return null;
+        return $this->_primaryKeyName !== null
+            ? $this->getIndex($this->_primaryKeyName)
```
morozov, line 606: Please keep it the old way for more accurate code coverage.

src/Schema/Table.php
```diff
@@ -704,7 +622,23 @@ public function getPrimaryKeyColumns()
...
-        return $primaryKey->getColumns();
+        return $this->filterColumns($primaryKey->getColumns());
+    }
+
+    /**
+     * Returns the foreign key columns
+     *
+     * @return Column[]
+     */
+    public function getForeignKeyColumns()
```
morozov, line 633: Should it declare the return type?

src/Schema/Table.php
```diff
@@ -758,6 +692,16 @@ public function getIndexes()
         return $this->_indexes;
     }
 
+    /**
+     * Returns the unique constraints.
+     *
+     * @return UniqueConstraint[]
+     */
+    public function getUniqueConstraints()
```
morozov, line 700: Missing return type declaration.

src/Schema/Table.php
```diff
@@ -835,34 +779,243 @@ public function __clone()
...
+    public function getComment() : ?string
+    {
+        return $this->_options['comment'] ?? null;
+    }
+
     /**
-     * Normalizes a given identifier.
+     * @return int
+     */
+    protected function _getMaxIdentifierLength()
```
morozov, line 798: Why does the method declaration need to be moved?

src/Platforms/OraclePlatform.php
guilhermeblanco: Sorry, comment was misplaced. I was referring to constraintName padding a space at the end. This one is resolved actually,.

src/Schema/Table.php
guilhermeblanco: I've ran a test and it works ok on both DBAL and ORM.
Personally, my changes have no correlation to the parameter $name in Table class, and this discussion was outside of scope of this PR (which leads to longer cycle time). But anyway, I've done the change as requested.

src/Schema/Table.php
guilhermeblanco: Code coverage works properly if you separate ternary lines (which I did). Also, ternaries are faster than if-guard clauses.

src/Schema/Table.php
guilhermeblanco: This was a moved method, it'll break BC by doing it if anyone extends Table.

src/Schema/Table.php
guilhermeblanco: Done

src/Schema/Table.php
guilhermeblanco: This method was being moved to follow the logical class declaration order:

* public constants
* protected constants
* private constants
* public static attributes
* protected static attributes
* private static attributes
* public attributes
* protected attributes
* private attributes
* constructor
* static public methods
* static protected methods
* static private methods
* public methods
* protected methods
* private methods
* magic methods

guilhermeblanco: For a better wording, the rule to thumb is: Non-public methods should always be at the end of the class declaration.

morozov: Let's do it in a separate PR. Ideally, it should be done consistently and enforced by the standard.

In this PR, it unnecessarily complicates the review since it introduces new methods and shuffles around the existing ones.

morozov: @carusogabriel is there something like above in the standard or it could be implemented there?

src/Schema/SchemaException.php
guilhermeblanco: Done

src/Schema/Table.php
guilhermeblanco: @morozov is that a requirement to have this merged? I'm trying to make ORM `master` point to `3.0.x` instead of `missed-commits`. I did update some methods (don't think this one in particular) because phpstan was not happy.

carusogabriel: @morozov Yeap, there's: https://github.com/slevomat/coding-standard#slevomatcodingstandardclassesclassstructure-

I don't think that'll make to the next major version of `doctrine/coding-standard`, as that release is already big enough. I'll propose to v9.

morozov: > […] is that a requirement to have this merged?

I believe it is. Otherwise, I cannot wrap my head around the scope of the PR.

tests/Functional/ExceptionTest.php
```diff
@@ -300,7 +301,7 @@ public function testConnectionExceptionSqLite() : void
         }
 
         // mode 0 is considered read-only on Windows
-        $mode = PHP_OS === 'Linux' ? 0444 : 0000;
+        $mode = in_array(PHP_OS, ['Linux', 'Darwin']) ? 0444 : 0000;
```
carusogabriel, line 304: What about inverting the logic to simplify?
```suggestion
        $mode = PHP_OS_FAMILY === 'Windows' ? 0444 : 0000;
```

And use a new constant introduced in PHP 7.2: https://php.net/manual/reserved.constants#PHP_OS_FAMILY

tests/Platforms/SqlitePlatformTest.php
```diff
@@ -824,4 +824,9 @@ public function testGetCreateTableSQLWithColumnCollation() : void
             'Column "no_collation" will use the default collation (BINARY) and "column_collation" overwrites the collation on this column'
         );
     }
+
+    public function testGetSequencePrefixWithSchema() : void
+    {
+        self::assertEquals('bar__foo', $this->platform->getSequencePrefix('foo', 'bar'));
```
carusogabriel, line 830: Here should be 
```suggestion
        self::assertSame('bar__foo', $this->platform->getSequencePrefix('foo', 'bar'));
```

Maybe we can propose this stricter check with `phpstan/phpstan-unit`.

tests/Functional/Schema/MySqlSchemaManagerTest.php
```diff
@@ -46,8 +46,8 @@ public function testSwitchPrimaryKeyColumns() : void
         $primaryKey = $table->getPrimaryKeyColumns();
 
         self::assertCount(2, $primaryKey);
-        self::assertContains('bar_id', $primaryKey);
-        self::assertContains('foo_id', $primaryKey);
+        self::assertArrayHasKey('bar_id', $primaryKey);
+        self::assertArrayHasKey('foo_id', $primaryKey);
```
carusogabriel, line 50: Why this change?

src/Platforms/AbstractPlatform.php
```diff
@@ -3078,10 +3092,25 @@ public function usesSequenceEmulatedIdentityColumns()
         return false;
     }
 
+    /**
+     * Gets the sequence name prefix based on table information.
+     */
+    public function getSequencePrefix(string $tableName, ?string $schemaName = null) : string
```
morozov, line 3098: This looks like a leaky abstraction to me. Especially because it's not used by DBAL itself. Given that all the internally used APIs are public, should ORM implement this logic instead?

src/Schema/Table.php
guilhermeblanco: I've moved back all methods to their original (not logical) location.

morozov (changes requested):
The addition of `Platform::getSequencePrefix()` is really questionable. Although this code is ported from the former `develop`, I don't see any PR/discussion through which it would get merged there.

tests/Functional/Schema/MySqlSchemaManagerTest.php
guilhermeblanco: Because before the methods were returning an array of column names. Now it's returning a HashMap/array indexed by column name => Column

src/Platforms/AbstractPlatform.php
guilhermeblanco: ORM is the one that would be leaking abstraction if calling this method.
Conceptually, ORM is telling DBAL "build me a primary key sequence for this table under this schema". The fact that ORM needs to do more than that is a bad abstraction on its own.
DBAL should defer to platforms that once a PK sequence is requested, a sequence named "bar__foo_seq" should be created if schemas are not supported or "bar.foo_seq" if they are.

I migrated the method to the project that owns it. A later stage would be encapsulate it and handle it accordingly internally without exposing this API.

guilhermeblanco:
@morozov answered you in the specific comment about why it was moved.

src/Platforms/AbstractPlatform.php
morozov: > ORM is the one that would be leaking abstraction if calling this method.

As far as I understand, what you're saying is ORM already relies on implementation details of DBAL. This is a given fact and it cannot be used to make DBAL _leak_ even more of its implementation details.

The very fact that there's a string in ORM that "bar__foo_seq" is built with "__" defined DBAL and "_" defined in ORM looks horrible.

I would agree that the DBAL abstraction used by ORM needs to be _improved_ but the fact that it's already bad cannot be used as a reason to make it even worse.

morozov:
@guilhermeblanco given the lack of progress on this issue in the past months and the fact that we're approaching the release, I'm inclined to move it out of the release milestone.

Please let me know if you plan to finish the work before the release.

morozov:
I removed the `3.0.0` and unmarked the PR as a blocker since the primary focus of the `3.0.0` release is PHP 8 support.

guilhermeblanco:
There is no further progress on this PR aside for waiting for merge.
We spoke in depth about this issue, and there are more functionality added to this PR than a method moving. The missing part was merging it. I relocated the method `getSequencePrefix` as described as a TODO on ORM (reference: https://github.com/doctrine/orm/blob/2.8.x/lib/Doctrine/ORM/Mapping/ClassMetadataInfo.php#L3380) , and this seems to be cause of this not being merged.
It's no responsibility to ORM to determine the sequence prefix to be used depending of the Platform being used, so ORM is not the place to have this code. If there is further missing parts to the functionality being moved, then at least the method at the proper repository. The fact that this is not landing on DBAL 3.0.0 means that ORM 3.0.0 will not support this major version at all, which is a really bad setback. 

We've spoken already about the role of DBAL and ORM. We have originally established that DBAL was a support library for ORM, not a fully independent library. If this have changed since the agreement in 2009, then it needs to be throughly discussed and determined how that will happen. As it current stands, DBAL is a support library for ORM, so moving the questioned method is valid and no further work is needed.

morozov:
I don't see `getSequencePrefix()` being part of a public DB abstraction API. Until this issue is resolved, this won't be merged.

vladyslavstartsev:
@morozov it's there

https://github.com/doctrine/dbal/blob/5c7a0e23a5a43dd4a22c79974257d5b815bb0bfc/src/Platforms/AbstractPlatform.php#L3110

`AbstractPlatform::getSequencePrefix()`

morozov:
> I don't see `getSequencePrefix()` being part of a public DB abstraction API.

I mean that I don't agree with this method being introduced.

vladyslavstartsev:
@guilhermeblanco  can you please rebase? So that version 3.0 can be released.

guilhermeblanco:
I'm on it... should be ready later today or tomorrow

morozov:
@guilhermeblanco do you need help with this?

guilhermeblanco:
I worked on that yesterday, but the amount of caused conflicts (and their pointed blocks) are too hard to resolve.
I feel the easier approach seems to be cherry pick the commits on top of current 3.0.x branch.
If you have time, feel free to start. I can wrap it up tonight if you can't make it to the end.

morozov:
@guilhermeblanco I rebased this and reorganized the commits in the way that each individual commit passes `phpcs` thereby eliminating commits like “Fixed code style”.

Please make the needed code changes and do some more cleanup. E.g.: 
1. The “Normalized parameter name across platforms” is no longer relevant since the same was done in #4172 and #4224.
2. Commits like “Type hinted Table constructor” and “Added missing return method” are not really relevant from the code change history and can be combined with others.
3. If “Moved back methods” revert the changes made in the same PR, it's better to not introduce these changes in the first place in order to not cause potential merge conflicts.
4. Normally, each commit should pass all build stages in order to be really atomic. This can be accomplished via `git rebase -i upstream/master exec $cmd` where `$cmd` is `phpcs`, `phpunit`, `phpstan a`, `psalm`, etc. This will help to bisect potential issues and not deal with knowingly broken states. If you don't want to do this, please squash everything.

morozov:
@guilhermeblanco given your lack of availability, how do you think we proceed here? Should I just remove `getSchemaPrefix()` and count that you handle it later in the ORM or you wanted to make more changes? Note that we're already 3 weeks behind the schedule and it makes us maintain an extra `2.10.x` release.

SenseException:
I'm fine with the changes. If getting `getSequencePrefix()` moved back to ORM is the last step of this PR that needs to be done, then I'm :+1: with this PR and it can be merged after that.

Handling `getSequencePrefix()` in DBAL can be done in a separate PR in coordination with an ORM PR.

morozov (approved):

PR #3983, opened 2020-05-02
Bump cs libs

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

Using the latest coding style should make #3977 a lot easier


lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
```diff
@@ -36,34 +36,28 @@ public function convertException($message, DriverException $exception)
                 return new Exception\LockWaitTimeoutException($message, $exception);
             case '1050':
                 return new Exception\TableExistsException($message, $exception);
-
```
morozov, line 39: This is an undesired change. See https://github.com/doctrine/coding-standard/pull/169.

phpcs.xml.dist
```diff
@@ -21,6 +21,10 @@
         <exclude name="SlevomatCodingStandard.Classes.DisallowLateStaticBindingForConstants.DisallowedLateStaticBindingForConstant"/>
     </rule>
 
+    <rule ref="Squiz.NamingConventions.ValidVariableName.PublicHasUnderscore">
```
morozov, line 42: Did this rule change its name? It must have been already ignored.

phpcs.xml.dist
```diff
@@ -21,11 +21,33 @@
         <exclude name="SlevomatCodingStandard.Classes.DisallowLateStaticBindingForConstants.DisallowedLateStaticBindingForConstant"/>
     </rule>
 
-    <rule ref="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingParameterTypeHint">
+    <!-- See https://github.com/squizlabs/PHP_CodeSniffer/issues/2937 -->
+    <rule ref="Squiz.Arrays.ArrayDeclaration.ValueNoNewline">
```
morozov, line 25: Should this be just excluded from the ruleset instead of being disabled for every path individually?

phpcs.xml.dist
```diff
@@ -21,11 +21,33 @@
...
+
+    <rule ref="Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps">
+        <!-- This test contains a fixture class that should use snake case  -->
+        <exclude-pattern>tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php</exclude-pattern>
+    </rule>
+    <rule ref="Squiz.NamingConventions.ValidVariableName.NotCamelCaps">
+        <!-- The dumper has special variables, and the test contains references to print_r  -->
+        <exclude-pattern>lib/Doctrine/DBAL/Tools/Dumper.php</exclude-pattern>
+        <exclude-pattern>tests/Doctrine/Tests/DBAL/Tools/DumperTest.php</exclude-pattern>
+        <!-- This files uses the return value db2_server_info(), which does not follow conventions  -->
```
morozov, line 38: If something like the usage of `$server->DBMS_VER` is flagged, then it's a bug in the standard. The caller cannot change the API being called. The naming convention is only violated by the defining side.

lib/Doctrine/DBAL/SQLParserUtils.php
```diff
@@ -71,7 +71,7 @@ private static function getPositionalPlaceholderPositions(string $statement) : a
             $statement,
             '?',
             self::POSITIONAL_TOKEN,
-            static function (string $_, int $placeholderPosition, int $fragmentPosition, array &$carry) : void {
+            static function (string $devNull, int $placeholderPosition, int $fragmentPosition, array &$carry) : void {
```
morozov, line 74: Please suppress (see https://github.com/squizlabs/PHP_CodeSniffer/issues/2837).

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
```diff
@@ -74,7 +74,7 @@ private function configureStatement(MockObject $stmt, int &$calls) : void
 
     private function assertIterationCallsFetchOncePerStep(Traversable $iterator, int &$calls) : void
     {
-        foreach ($iterator as $i => $_) {
+        foreach ($iterator as $i => $devNull) {
```
morozov, line 80: Please suppress.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -166,8 +166,8 @@ public function testEventManagerPassedToPlatform() : void
     {
         $eventManager = new EventManager();
 
-        /** @var AbstractPlatform|MockObject $driver */
         $platform = $this->createMock(AbstractPlatform::class);
+        assert($platform instanceof AbstractPlatform && $platform instanceof MockObject);
```
morozov, line 170: It feels awkward to make assertions against the values produced by the test framework. As of https://github.com/sebastianbergmann/phpunit/commit/69d23b99e1fa301a28656ea83d68c384cd1a5c4e, it may be no longer needed.

lib/Doctrine/DBAL/Platforms/PostgreSQL100Platform.php
```diff
@@ -16,6 +16,9 @@ protected function getReservedKeywordsClass() : string
         return PostgreSQL100Keywords::class;
     }
 
+    /**
+     * {@inheritDoc}
```
morozov, line 20: What exactly do we need to explicitly inherit?

morozov (changes requested):

lib/Doctrine/DBAL/Platforms/PostgreSQL100Platform.php
greg0ire: the type information

tests/Doctrine/Tests/DBAL/ConnectionTest.php
greg0ire: I'm going to grep `instanceof MockObject` and remove occurences. It should result in better type information since currently, `||` is used, mostly.

phpcs.xml.dist
greg0ire: I just reported this. I will improve the comment with a link to the bug report.

phpcs.xml.dist
greg0ire: Doesn't look like it. I checked and couldn't find a rule with a similar name, and besides, if the name changes, phpcs refuses to run because it does not recognize the old name.

lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
greg0ire: The rule that requires this change is `SlevomatCodingStandard.ControlStructures.JumpStatementsSpacing.IncorrectLinesCountAfterLastControlStructure`

greg0ire: Oh I get what you mean now. I should ignore it until 8.0.0

greg0ire: Done.

phpcs.xml.dist
morozov: Why are we only introducing the suppression right now? Wasn't this requirement already part of the standard?

lib/Doctrine/DBAL/Platforms/PostgreSQL100Platform.php
morozov: I already brought it up somewhere but [`{@inheritDoc}`](https://docs.phpdoc.org/latest/guides/inheritance.html#the-inheritdoc-tag) is irrelevant to the type information which is [inherited automatically](https://docs.phpdoc.org/latest/guides/inheritance.html#methods). While this may be a known limitation of the slevomat/doctrine standard, I don't like this turning into a cargo cult.

phpcs.xml.dist
greg0ire: I guess not: when I checkout `origin/2.10.x`, remove all the exclude rules from `phpcs.xml.dist`, and run phpcs against `lib/Doctrine/DBAL/Event/Listeners/OracleSessionInit.php`, which is an offender for this rule, I get no complaints.

lib/Doctrine/DBAL/Platforms/PostgreSQL100Platform.php
greg0ire: > Currently some applications have DocBlocks containing just the {@inheritDoc} inline tag to indicate that their complete contents should be inherited. This usage breaks with the PHPDoc Standard as summaries cannot contain inline tags and inheritance is automatic; you do not need to define a special tag for it.

If I understand this correctly, this should never happen:
```
/**
 * {@inheritDoc}
 */
```

> However, it does make clear that an element has been explicitly documented (and thus not forgotten). As such we are working to include a new (normal) tag in the PHPDoc Standard @inheritDoc that will serve that purpose.

but this can:

```
/**
 * @inheritDoc
 */
```

The former happens all over the place in this lib… maybe we should merge as is and this should fix afterwards? I reported the issue here: https://github.com/slevomat/coding-standard/issues/991

greg0ire: Weirdly, the build currently fails because of extra `{@inheritDoc}`: https://travis-ci.org/github/doctrine/dbal/jobs/682337729#L242-L243 Something is off…

morozov: > […] maybe we should merge as is and this should fix afterwards?

Yeah. Just wanted to make sure we're not blindly adding some annotations and removing exactly the same ones in other places just to pass the check.

greg0ire: > Something is off…

I just needed to rebase

phpcs.xml.dist
greg0ire: @morozov maybe you are thinking of this: https://github.com/doctrine/coding-standard/blob/c665baebb119e087b421e282a8565519be4c6c00/lib/Doctrine/ruleset.xml#L486-L488

The name of this one suggests it's about private properties only.

morozov: Oh, I didn't realize there are different rules for different types of visibility. Thank you.

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
```diff
@@ -531,7 +532,7 @@ public function getAlterTableSQL(TableDiff $diff)
         }
 
         foreach ($diff->changedColumns as $columnDiff) {
-            /** @var $columnDiff \Doctrine\DBAL\Schema\ColumnDiff */
+            assert($columnDiff instanceof ColumnDiff);
```
morozov, line 535: This assertion, as well as the original doc block, may be redundant.

lib/Doctrine/DBAL/Schema/Table.php
```diff
@@ -240,7 +242,7 @@ public function renameIndex($oldIndexName, $newIndexName = null)
     public function columnsAreIndexed(array $columnNames)
     {
         foreach ($this->getIndexes() as $index) {
-            /** @var $index Index */
+            assert($index instanceof Index);
```
morozov, line 245: This may be redundant.

phpcs.xml.dist
```diff
@@ -14,18 +14,67 @@
...
         <exclude name="SlevomatCodingStandard.ControlStructures.ControlStructureSpacing.IncorrectLinesCountAfterLastControlStructure"/>
         <exclude name="SlevomatCodingStandard.Classes.DisallowLateStaticBindingForConstants.DisallowedLateStaticBindingForConstant"/>
+        <!-- See https://github.com/squizlabs/PHP_CodeSniffer/issues/2937 -->
+        <exclude name="Squiz.Arrays.ArrayDeclaration.ValueNoNewline"/>
     </rule>
 
-    <rule ref="SlevomatCodingStandard.TypeHints.TypeHintDeclaration.MissingParameterTypeHint">
+    <rule ref="Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps">
+        <!-- This test contains a fixture class that should use snake case  -->
+        <exclude-pattern>tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php</exclude-pattern>
```
morozov, line 33: Apparently, the CS is already updated in `master` (#3847), and this issue is addressed in a slightly different way.

In order to avoid unnecessary merge conflicts, what if we restart by backporting the CS update from master and applying all the necessary changes on top?

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -710,8 +705,8 @@ public function testPassingExternalPDOMeansConnectionIsConnected() : void
 
     public function testCallingDeleteWithNoDeletionCriteriaResultsInInvalidArgumentException() : void
     {
-        /** @var Driver $driver */
-        $driver  = $this->createMock(Driver::class);
+        $driver = $this->createMock(Driver::class);
+        assert($driver instanceof Driver);
```
morozov, line 709: This must be redundant.

tests/Doctrine/Tests/DBAL/Driver/Mysqli/MysqliConnectionTest.php
```diff
@@ -44,10 +44,10 @@ public function testDoesNotRequireQueryForServerVersion() : void
 
     public function testRestoresErrorHandlerOnException() : void
     {
-        $handler         = static function () : bool {
+        $handler        = static function () : bool {
```
morozov, line 47: Would it make sense to add an empty line after the `$handler` declaration? It doesn't make much sense (even from the standard PoV) to keep the assignment of a multi-line expression aligned with the next one.

tests/Doctrine/Tests/DBAL/Functional/MasterSlaveConnectionTest.php
```diff
@@ -83,8 +84,8 @@ public function testInheritCharsetFromMaster() : void
                 unset($params['slaves'][$index]['charset']);
             }
 
-            /** @var MasterSlaveConnection $conn */
             $conn = DriverManager::getConnection($params);
+            assert($conn instanceof MasterSlaveConnection);
```
morozov, line 88: Should it be a PHPUnit assertion?

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
```diff
@@ -23,8 +24,8 @@ class StatementIteratorTest extends DbalTestCase
      */
     public function testGettingIteratorDoesNotCallFetch(string $class) : void
     {
-        /** @var IteratorAggregate|MockObject $stmt */
         $stmt = $this->createPartialMock($class, ['fetch', 'fetchAll', 'fetchColumn']);
+        assert($stmt instanceof IteratorAggregate || $stmt instanceof MockObject);
```
morozov, line 28: I am really concerned about adding such assertions in tests:

1. If the pre-condition is not met, the test will most likely fail, there's no point in adding native PHP assertions.
2. If we want to keep the assertion, it must be done via the PHPUnit API.

In this specific test, the must be of both types (two assertions), not any of them.

lib/Doctrine/DBAL/Platforms/PostgreSQL100Platform.php
greg0ire: @morozov as suggested in https://github.com/slevomat/coding-standard/issues/991#issuecomment-623069823 , I think we should wrap this up and merge as is.

morozov (commented):
Let's double-check all the newly introduced `assert()`. Many of them look redundant.

lib/Doctrine/DBAL/Platforms/PostgreSQL100Platform.php
morozov: Agree. We can keep the discussion in the above issue.

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
greg0ire: :thinking: wait I thought I removed those

phpcs.xml.dist
greg0ire: Ok, I'll cherry-pick a84752c4382a23ea34cec1cfcfb883417b1cac47, then 78546ce9be08752d9ea2e621b9df816098969e3b (that should result in just an extra rule),  . I will not cherry-pick the commit about automated fixes and will instead run `/vendor/bin/phpcbf`, and finally I will cherry-pick a92fa15198aac398068deb81d331cd79f4f9bb1c

greg0ire: > that should result in just an extra rule

It actually doesn't, since some line breaks are already missing on 2.10.x

lib/Doctrine/DBAL/Platforms/PostgreSqlPlatform.php
greg0ire: Even psalm says so (as an info level)

tests/Doctrine/Tests/DBAL/Driver/StatementIteratorTest.php
greg0ire: I removed it

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1552,7 +1552,7 @@ public function getCreateTableSQL(Table $table, $createFlags = self::CREATE_INDE
 
         if (($createFlags&self::CREATE_INDEXES) > 0) {
             foreach ($table->getIndexes() as $index) {
-                /** @var $index Index */
+                assert($index instanceof Index);
```
morozov, line 1555: Is this required?

lib/Doctrine/DBAL/Schema/Visitor/DropSchemaSqlCollector.php
```diff
@@ -82,19 +83,19 @@ public function getQueries()
     {
         $sql = [];
 
-        /** @var ForeignKeyConstraint $fkConstraint */
         foreach ($this->constraints as $fkConstraint) {
+            assert($fkConstraint instanceof ForeignKeyConstraint);
```
morozov, line 87: Not necessarily for now, but would it be possible to document the property as `SplObjectStorage<ForeignKeyConstraint>` and avoid runtime assertions? I may have tried earlier but not sure.

phpcs.xml.dist
```diff
@@ -45,13 +65,43 @@
...
+
+    <rule ref="Squiz.NamingConventions.ValidVariableName.MemberNotCamelCaps">
+        <!-- This test contains a fixture class that should use snake case  -->
+        <exclude-pattern>tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php</exclude-pattern>
+    </rule>
+
+    <!-- https://github.com/squizlabs/PHP_CodeSniffer/issues/2837 -->
+    <rule ref="Squiz.NamingConventions.ValidVariableName.NotCamelCaps">
+        <!--
+            This files uses the return value db2_server_info(), which does not follow conventions
```
morozov, line 93: Should it be “this _file_ uses the return value _of_…”?

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -829,8 +819,8 @@ public function testConnectionParamsArePassedToTheQueryCacheProfileInExecuteCach
             ->with($query, $params, $types, $this->params)
             ->will($this->returnValue(['cacheKey', 'realKey']));
 
-        /** @var Driver $driver */
         $driver = $this->createMock(Driver::class);
+        assert($driver instanceof Driver);
```
morozov, line 823: This is redundant.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -871,8 +860,8 @@ public function testShouldNotPassPlatformInParamsToTheQueryCacheProfileInExecute
 
         $connectionParams['platform'] = $this->createMock(AbstractPlatform::class);
 
-        /** @var Driver $driver */
         $driver = $this->createMock(Driver::class);
+        assert($driver instanceof Driver);
```
morozov, line 864: This is redundant.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -885,8 +874,8 @@ public function testThrowsExceptionWhenInValidPlatformSpecified() : void
         $connectionParams             = $this->params;
         $connectionParams['platform'] = new stdClass();
 
-        /** @var Driver $driver */
         $driver = $this->createMock(Driver::class);
+        assert($driver instanceof Driver);
```
morozov, line 878: This is redundant.

tests/Doctrine/Tests/DBAL/DBALExceptionTest.php
```diff
@@ -17,27 +18,27 @@ class DBALExceptionTest extends DbalTestCase
 {
     public function testDriverExceptionDuringQueryAcceptsBinaryData() : void
     {
-        /** @var Driver $driver */
         $driver = $this->createMock(Driver::class);
-        $e      = DBALException::driverExceptionDuringQuery($driver, new Exception(), '', ['ABC', chr(128)]);
+        assert($driver instanceof Driver);
```
morozov, line 22: This is redundant.

tests/Doctrine/Tests/DBAL/DBALExceptionTest.php
```diff
@@ -17,27 +18,27 @@ class DBALExceptionTest extends DbalTestCase
...
+        $e = DBALException::driverExceptionDuringQuery($driver, new Exception(), '', ['ABC', chr(128)]);
         self::assertStringContainsString('with params ["ABC", "\x80"]', $e->getMessage());
     }
 
     public function testDriverExceptionDuringQueryAcceptsResource() : void
     {
-        /** @var Driver $driver */
         $driver = $this->createMock(Driver::class);
-        $e      = DBALException::driverExceptionDuringQuery($driver, new Exception(), 'INSERT INTO file (`content`) VALUES (?)', [1 => fopen(__FILE__, 'r')]);
+        assert($driver instanceof Driver);
```
morozov, line 30: This is redundant.

tests/Doctrine/Tests/DBAL/DBALExceptionTest.php
```diff
@@ -17,27 +18,27 @@ class DBALExceptionTest extends DbalTestCase
...
+        $e = DBALException::driverExceptionDuringQuery($driver, new Exception(), 'INSERT INTO file (`content`) VALUES (?)', [1 => fopen(__FILE__, 'r')]);
         self::assertStringContainsString('Resource', $e->getMessage());
     }
 
     public function testAvoidOverWrappingOnDriverException() : void
     {
-        /** @var Driver $driver */
         $driver = $this->createMock(Driver::class);
+        assert($driver instanceof Driver);
```
morozov, line 38: This is redundant.

tests/Doctrine/Tests/DBAL/DBALExceptionTest.php
```diff
@@ -17,27 +18,27 @@ class DBALExceptionTest extends DbalTestCase
...
 
-        /** @var InnerDriverException $inner */
         $inner = $this->createMock(InnerDriverException::class);
+        assert($inner instanceof InnerDriverException);
```
morozov, line 41: This is redundant.

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
```diff
@@ -953,8 +954,8 @@ private function beforeFetchClassTest() : void
             return;
         }
 
-        /** @var PDOConnection $connection */
         $connection = $this->connection->getWrappedConnection();
+        assert($connection instanceof PDOConnection);
```
morozov, line 958: Please use a PHPUnit assertion if needed.

tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
```diff
@@ -844,7 +841,7 @@ public static function getReturnsGetListTableColumnsSQL() : iterable
 WHERE    c.table_name = 'test'
 ORDER BY c.column_id
 SQL
-                ,
+,
```
morozov, line 844: Is this expected?

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
greg0ire: Nope, good point

lib/Doctrine/DBAL/Schema/Visitor/DropSchemaSqlCollector.php
greg0ire: Documenting it as `SplObjectStorage<ForeignKeyConstraint, Table>` makes Psalm happy, but Phpstan gives the following error: `PHPDoc tag @var for property                                                                   
         Doctrine\DBAL\Schema\Visitor\DropSchemaSqlCollector::$constraints contains unresolvable type.`

tests/Doctrine/Tests/DBAL/Platforms/OraclePlatformTest.php
greg0ire: :thinking: no… it looks weird to me, but phpcs complains if I don't do it…

```
 844 | ERROR | [x] Expected 0 spaces before comma; 15 found
     |       |     (Squiz.Arrays.ArrayDeclaration.SpaceBeforeComma)
```

Looks like a bug…

greg0ire: Hey look who's there when you type "heredoc nowdoc SpaceBeforeComma" on the phpcs repo! https://github.com/squizlabs/PHP_CodeSniffer/issues/2165

greg0ire: I think a nice solution is to avoid putting the comma on its own line

greg0ire: Ah in fact it's not nice at all, because it does not work on 7.3, reverting this.

greg0ire: Reported here: https://github.com/squizlabs/PHP_CodeSniffer/issues/2951

lib/Doctrine/DBAL/Schema/Visitor/DropSchemaSqlCollector.php
morozov: Alternatively, it might be something like `SplObjectStorage&Iterable<ForeignKeyConstraint>`.

morozov:
Looks good, @greg0ire :+1: Please group/squash the commits.

greg0ire:
Done! Down from 16 to 10 commits

morozov (approved):

greg0ire:
I'm ignoring Scrutinizer on this one! Thanks @morozov for your review!

Issue #3975, opened 2020-04-22
v2.10.2 - SQLSTATE[HY000]: General error: Extraneous additional parameters


oojacoboo:
### Bug Report

When using the Doctrine PDO connection object to bind params to a prepared statement, the following error is being caused.

`SQLSTATE[HY000]: General error: Extraneous additional parameters`

This issue did not exist with 2.10.1.  I also believe it's related to `$stmt->fetchAll(\PDO::FETCH_ASSOC)`.

|    Q        |   A
|------------ | ------
| Version     | 2.10.2

greg0ire:
Here is the list of changes between the 2 versions: https://github.com/doctrine/dbal/compare/v2.10.1...2.10.2

Do you think it might have to do with https://github.com/doctrine/dbal/commit/24cefe5d83a8e0ccc49b18fe774b11105a7fbcd9 ?

Is there a stack trace you could share?

oojacoboo:
Here is the trace:

```
[
  ...
  {
      "file": "/srv/www/vendor/doctrine/dbal/lib/Doctrine/DBAL/Statement.php",
      "line": 252,
      "function": "fetchAll",
      "class": "PDOStatement",
      "type": "->",
      "args": [
        2,
        null,
        null
      ]
    }
  ],
  "message": "SQLSTATE[HY000]: General error: Extraneous additional parameters",
  "file": "/srv/www/vendor/doctrine/dbal/lib/Doctrine/DBAL/Statement.php",
  "line": 252,
  "class": "PDOException",
  "user": [],
  "tags": [],
  "extra": [],
  "timestamp": "2020-04-22 18:28:31 UTC",
  "level": "critical",
  "environment": "dev",
  "app_version": "4.0.0-stage+20200419180027"
}
```

And yes, that could certainly be related.  It looks like Propel dealt with an issue related to this years ago- found some issues on github.

greg0ire:
Possible next steps:

1. Confirming that by editing the code and just removing the 3rd argument
2. Checking the `fetchAll` docs to see if there is anything about this
3. Contributing a PR that passes the third argument only when necessary, depending on the first one

Carpenter0100:
i am running into the same issue and can confirm that this PR is the cause: 24cefe5

greg0ire:
@duncan3dc please have a look at this

DanielSchwiperich:
Can also confirm and set a conflict for 2.10.2 in composer.json which fixed our pipeline so far

[...]

PR #3984, opened 2020-05-03
Only pass the additional arguments if they are not null

duncan3dc:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3975

#### Summary

Following on from #3893 this PR ensures the additional arguments are only passed when they are non-null, as issues were reported in #3975

### DIFF
--- a/lib/Doctrine/DBAL/Statement.php
+++ b/lib/Doctrine/DBAL/Statement.php
@@ -249,7 +249,15 @@ class Statement implements IteratorAggregate, DriverStatement
      */
     public function fetchAll($fetchMode = null, $fetchArgument = null, $ctorArgs = null)
     {
-        return $this->stmt->fetchAll($fetchMode, $fetchArgument, $ctorArgs);
+        if ($ctorArgs !== null) {
+            return $this->stmt->fetchAll($fetchMode, $fetchArgument, $ctorArgs);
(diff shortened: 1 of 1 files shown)


greg0ire (commented):
Can you please add some tests?

duncan3dc:
Any advice on how you'd like that done?

It looks like you're using PHPUnit for mocking here, which doesn't support verifying that no argument was passed, it always substitutes in the default value anyway. So mocking doesn't appear to be an option.

It looks like you set up some databases, would you prefer an integration test with one of these database servers? If so, which one and where should the test case live?

greg0ire:
Since the issue arises in the form of an SQL error, an integration test makes more sense. You can run it against all DBs, and it should live under `tests/Doctrine/Tests/DBAL/Functional` I think. I wrote one for the first time a few weeks ago, maybe it will help: https://github.com/doctrine/dbal/pull/3994

On a somewhat related note, I see that kind of question coming back again and again: https://github.com/doctrine/dbal/pull/4012#issuecomment-632009994

We should really work on providing that kind of guide :sweat_smile:

duncan3dc:
Closing as @BenMorel is tackling the tricky tests in #4173

PR #3989, opened 2020-05-05
Remove useless rules

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

This removes ignore rule that no longer seem useful. I left 2 rules that don't have an effect on my machine though. First, this one:

https://github.com/doctrine/dbal/blob/4c03ed81471c62f178581eb325339a3c34f3b71a/phpcs.xml.dist#L120-L124

It's about not adding strict types, that we don't have on this branch, but that we have in other branches. Removing it would make up merges needlessly difficult and seems pointless since it's going to appear again.

Then we have this one:

https://github.com/doctrine/dbal/blob/4c03ed81471c62f178581eb325339a3c34f3b71a/phpcs.xml.dist#L126-L130

It does nothing on my machine but I suspect it might be because I don't have an extension that provides sqlsrv functions, and it looks like being a built-in function has a special effect:

https://github.com/squizlabs/PHP_CodeSniffer/blob/e9ebf5253a3e6dd3cc4b584ce54b1f4c34fcb1f3/src/Standards/Squiz/Sniffs/PHP/LowercasePHPFunctionsSniff.php#L70-L75

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -98,11 +98,6 @@
         <exclude-pattern>lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php</exclude-pattern>
     </rule>
 
-    <!-- sqlsrv functions are documented in upper case but reflected in lower case -->
-    <rule ref="Squiz.PHP.LowercasePHPFunctions.CallUppercase">
-        <exclude-pattern>lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php</exclude-pattern>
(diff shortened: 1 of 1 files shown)


greg0ire:
ping @morozov

morozov (approved):

morozov:
Thanks for the reminder. I completely forgot about it.

greg0ire:
No worries :)

PR #3993, opened 2020-05-05
Use proper check in acceptForeignKey()

### LINKED ISSUE #3990: Symfony CI red after "[GH-1204] Add full support for foreign key constraints in SQLite Platform and Schema"
Commit https://github.com/doctrine/dbal/commit/85a983c3f8d6447c4441832477def1c28d57bee9 broke the CI of Symfony

see eg https://travis-ci.org/github/symfony/symfony/jobs/683266277

`./phpunit src/Symfony/Bridge/Doctrine` on any branch is the reproducer for now.

greg0ire:
|      Q       |   A
|------------- | -----------
| BC Break     | no
| Fixed issues | #3990 

#### Summary
Checking that the platform supports foreign key is not the right case
here, because we should check if we can create them by themselves, after
the table is created. It is not the case for Sqlite.

Closes #3990

# How can I test this?

```shell
composer config repositories.greg0ire vcs https://github.com/greg0ire/doctrine-dbal
composer require doctrine/dbal "dev-check-for-alter-support as 2.10.2"
```

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Visitor/CreateSchemaSqlCollector.php
+++ b/lib/Doctrine/DBAL/Schema/Visitor/CreateSchemaSqlCollector.php
@@ -55,7 +55,7 @@ class CreateSchemaSqlCollector extends AbstractVisitor
      */
     public function acceptForeignKey(Table $localTable, ForeignKeyConstraint $fkConstraint)
     {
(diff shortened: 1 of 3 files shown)


greg0ire:
Okay, it somehow found and associated the builds from previous PR with it… I need to create a 3rd PR :facepalm:

PR #3994, opened 2020-05-05
Use proper check in acceptForeignKey()

### LINKED ISSUE #3990: Symfony CI red after "[GH-1204] Add full support for foreign key constraints in SQLite Platform and Schema"

greg0ire:
|      Q       |   A
|------------- | -----------
| BC Break     | no
| Fixed issues | #3990 

#### Summary
Checking that the platform supports foreign key is not the right case
here, because we should check if we can create them by themselves, after
the table is created. It is not the case for Sqlite.

Closes #3990

# How can I test this?

```shell
composer config repositories.greg0ire vcs https://github.com/greg0ire/doctrine-dbal
composer require doctrine/dbal "dev-check-for-alter-support as 2.10.2"
```

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Visitor/CreateSchemaSqlCollector.php
+++ b/lib/Doctrine/DBAL/Schema/Visitor/CreateSchemaSqlCollector.php
@@ -55,7 +55,7 @@ class CreateSchemaSqlCollector extends AbstractVisitor
      */
     public function acceptForeignKey(Table $localTable, ForeignKeyConstraint $fkConstraint)
     {
-        if (! $this->platform->supportsForeignKeyConstraints()) {
+        if (! $this->platform->supportsCreateDropForeignKeyConstraints()) {
             return;
         }
 
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/ForeignKeyTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/ForeignKeyTest.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL\Functional\Schema;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Tests\DbalFunctionalTestCase;
+
+class ForeignKeyTest extends DbalFunctionalTestCase
+{
+    public function testCreatingATableWithAForeignKey() : void
+    {
+        $schema = new Schema();
+
+        $referencedTable = $schema->createTable('referenced_table');
+        $referencedTable->addColumn('id', 'integer');
+        $referencedTable->setPrimaryKey(['id']);
+
+        $referencingTable = $schema->createTable('referencing_table');
+        $referencingTable->addColumn('referenced_id', 'integer');
+        $referencingTable->addForeignKeyConstraint(
+            $referencedTable,
+            ['referenced_id'],
+            ['id']
+        );
+
+        foreach ($schema->toSql($this->connection->getDatabasePlatform()) as $sql) {
+            $this->connection->exec($sql);
+        }
+
+        self::assertCount(
+            1,
+            $this->connection->getSchemaManager()->listTableForeignKeys('referencing_table')
+        );
+    }
+}
--- a/tests/Doctrine/Tests/DBAL/Schema/Visitor/CreateSchemaSqlCollectorTest.php
+++ b/tests/Doctrine/Tests/DBAL/Schema/Visitor/CreateSchemaSqlCollectorTest.php
@@ -29,7 +29,7 @@ class CreateSchemaSqlCollectorTest extends TestCase
                     'getCreateSchemaSQL',
                     'getCreateSequenceSQL',
                     'getCreateTableSQL',
-                    'supportsForeignKeyConstraints',
+                    'supportsCreateDropForeignKeyConstraints',
                     'supportsSchemas',
                 ]
             )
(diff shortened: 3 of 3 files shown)


greg0ire:
It still finds the old builds… let's try an amend

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
```diff
@@ -294,4 +294,26 @@ public function testOnlyOwnCommentIsParsed() : void
             ->getColumn('col1')
             ->getComment());
     }
+
+    /**
+     * @doesNotPerformAssertions
```
lcobucci, line 299: Just a note: `@doesNotPerformAssertions` also disregards coverage (conscious decision on PHPUnit's side). The recommended way to test that an exception hasn't been thrown is calling `$this->addToAssertionCount(1)` after the method that throws exceptions.

greg0ire: TIL

lcobucci: @greg0ire https://github.com/sebastianbergmann/phpunit/pull/3348 :smile:

morozov: I'm really confused about having `@doesNotPerformAssertions` here. Is the only outcome of actions in the test scenario is not having an exception thrown? I would understand such expectations in a validation scenario (no exception means “valid”) but not here.

greg0ire: @morozov my goal here was to have a failing test for an existing bug, I'm not really trying to assert anything, I want to fix a crash, and make sure it does not come back. Tell me if this file is not the right place for that (since it is in `Functional`, it might not be, but I'm not sure where I should put this).

morozov: > I want to fix a crash, and make sure it does not come back.

IMO, the purpose of a test is to ensure that the code works, not that it doesn't crash. I.e. should be worded in terms of assertions against produced results.

How do you know that just the absence of failure is enough in the reported case?

greg0ire: Because this is what is complained about… they are not complaining that the fk does not get created properly for instance.

Anyway, I started by asserting that the generated SQL contains something about foreign key, but then I thought that even that might not work so I executed the SQL and asserted that there is a foreign key defined on it, tell me if I went overboard.

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
```diff
@@ -294,4 +294,40 @@ public function testOnlyOwnCommentIsParsed() : void
             ->getColumn('col1')
             ->getComment());
     }
+
+    public function testCreatingATableWithAForeignKey() : void
+    {
```
greg0ire, line 299: Few tests here have a guard clause about `sqlite3`, should I add one?

morozov: Is there anything specific to SQLite in this scenario? If yes, then I think the setup of the functional platform-specific test should require the corresponding platform to be active.

Otherwise, the test should be part of a generic suite.

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
morozov: > Because this is what is complained about… they are not complaining that the fk does not get created properly for instance.

I don't think this is the right mindset. But I believe we're on the same page now.

> […] tell me if I went overboard

The approach looks fine. The only question is, is this case is platform-specific or we can move it into a platform-independent suite and get covered on all platforms. The fact that people tend to write platform-specific tests because they experience issues using a given platform is the primary reason why we have inconsistency in platform support.

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
greg0ire: I think this is platform independent, and I'm not sure what a platform-independent suite is. I moved it to a file that inherits from `DbalFunctionTestCase`, let's see how this pans out. My only doubt is about the foreign key name, I have no idea what computes it and if it will be the same for every platform.

greg0ire:
Just pushed a version with a way, way less specific assertion.

greg0ire:
@morozov any idea why it fails for IBM DB2 and only for that RDBMS?

morozov:
> […] any idea why it fails for IBM DB2 and only for that RDBMS?

I cannot check it right now. It's fine to mark it incomplete for DB2 and fix later but I think you'll need the @beberlei’s approval to get it merged.

morozov:
Well, actually, the problem seems to be in the case-sensitivity of DB2 when using quoted names.

This is what the SQL creating the tables looks like (note that one name is quoted and the other isn't):
```sql
CREATE TABLE referenced (id INTEGER NOT NULL, PRIMARY KEY(id))
CREATE TABLE "referencing" (referenced_id INTEGER NOT NULL)
-- ...
ALTER TABLE "referencing" ADD CONSTRAINT FK_E844F4C85A38C91 FOREIGN KEY (referenced_id) REFERENCES referenced (id)
```
And then how the FK is selected:

https://github.com/doctrine/dbal/blob/d6f232e4066b2d6ee6d75371c03657d3dd8db7cc/lib/Doctrine/DBAL/Platforms/DB2Platform.php#L393

I don't think that quoting object names by default on DB2 or Oracle is the right thing to do, so it's definitely a bug in DBAL.

greg0ire:
Ok, I'll just pick another name for this PR I guess! Thanks a lot for the investigation.

morozov (approved):

greg0ire:
The tests that fails is unrelated (and seems time based)

beberlei (approved):

PR #3997, opened 2020-05-06
Changing the position of the emptied variable

dfrnks:
When use fetch or fetchColumn and don't pass for all registers if while they don't close the cursor and don't save in the cache

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues |

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -922,7 +922,7 @@ class Connection implements DriverConnection
 
         return $stmt;
     }
-    
+
     /**
      * Executes a caching query.
      *
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -564,6 +564,8 @@ class Connection implements DriverConnection
      * @param int[]|string[] $types     The query parameter types.
      *
      * @return mixed[]|false False is returned if no rows are found.
+     *
+     * @throws DBALException
      */
     public function fetchArray($statement, array $params = [], array $types = [])
     {
@@ -687,6 +689,8 @@ class Connection implements DriverConnection
      * @param int $level The level to set.
      *
      * @return int
+     *
+     * @throws DBALException
      */
     public function setTransactionIsolation($level)
     {
@@ -699,6 +703,8 @@ class Connection implements DriverConnection
      * Gets the currently active transaction isolation level.
      *
      * @return int The current transaction isolation level.
+     *
+     * @throws DBALException
      */
     public function getTransactionIsolation()
     {
@@ -814,6 +820,8 @@ class Connection implements DriverConnection
      * @param string $str The name to be quoted.
      *
      * @return string The quoted name.
+     *
+     * @throws DBALException
      */
     public function quoteIdentifier($str)
     {
@@ -840,6 +848,8 @@ class Connection implements DriverConnection
      * @param int[]|string[] $types  The query parameter types.
      *
      * @return mixed[]
+     *
+     * @throws DBALException
      */
     public function fetchAll($sql, array $params = [], $types = [])
     {
@@ -1200,6 +1210,8 @@ class Connection implements DriverConnection
      * @return void
      *
      * @throws ConnectionException
+     *
+     * @throws DBALException
      */
     public function setNestTransactionsWithSavepoints($nestTransactionsWithSavepoints)
     {
@@ -1275,6 +1287,8 @@ class Connection implements DriverConnection
      *
      * @throws ConnectionException If the commit failed due to no active transaction or
      *                                            because the transaction was marked for rollback only.
+     *
+     * @throws DBALException
      */
     public function commit()
     {
@@ -1348,6 +1362,8 @@ class Connection implements DriverConnection
      * @return bool
      *
      * @throws ConnectionException If the rollback operation failed.
+     *
+     * @throws DBALException
      */
     public function rollBack()
     {
@@ -1400,6 +1416,8 @@ class Connection implements DriverConnection
      * @return void
      *
      * @throws ConnectionException
+     *
+     * @throws DBALException
      */
     public function createSavepoint($savepoint)
     {
@@ -1418,6 +1436,8 @@ class Connection implements DriverConnection
      * @return void
      *
      * @throws ConnectionException
+     *
+     * @throws DBALException
      */
     public function releaseSavepoint($savepoint)
     {
@@ -1440,6 +1460,8 @@ class Connection implements DriverConnection
      * @return void
      *
      * @throws ConnectionException
+     *
+     * @throws DBALException
      */
     public function rollbackSavepoint($savepoint)
     {
@@ -1518,6 +1540,8 @@ class Connection implements DriverConnection
      * @param string $type  The name of the DBAL mapping type.
      *
      * @return mixed The converted value.
+     *
+     * @throws DBALException
      */
     public function convertToDatabaseValue($value, $type)
     {
@@ -1532,6 +1556,8 @@ class Connection implements DriverConnection
      * @param string $type  The name of the DBAL mapping type.
      *
      * @return mixed The converted type.
+     *
+     * @throws DBALException
      */
     public function convertToPHPValue($value, $type)
     {
@@ -1550,6 +1576,8 @@ class Connection implements DriverConnection
      * @param int[]|string[]                  $types  The parameter types (PDO binding types or DBAL mapping types).
      *
      * @return void
+     *
+     * @throws DBALException
      */
     private function _bindTypedValues($stmt, array $params, array $types)
     {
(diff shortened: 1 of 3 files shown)


tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
```diff
@@ -194,6 +195,28 @@ public function testFetchAllColumn() : void
...
+        $stmt->closeCursor();
+
+        $stmt = $this->connection->executeCacheQuery($query, [], [], $qcp);
+
+        $iterator = $stmt->/** @scrutinizer ignore-call */
+                            getIterator();
+
+        self::assertTrue($iterator instanceof ArrayIterator);
+
+        self::assertEquals(1, $iterator->count());
```
SenseException, line 217:
```suggestion
        self::assertCount(1, $iterator);
```

SenseException (changes requested):

tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
```diff
@@ -194,6 +195,28 @@ public function testFetchAllColumn() : void
...
+
+        $stmt = $this->connection->executeCacheQuery($query, [], [], $qcp);
+        $stmt->/** @scrutinizer ignore-call */
+                getIterator();
+        $stmt->closeCursor();
+
+        $stmt = $this->connection->executeCacheQuery($query, [], [], $qcp);
+
+        $iterator = $stmt->/** @scrutinizer ignore-call */
+                            getIterator();
```
greg0ire, line 213: What did Scrutinizer have to say about this? There are currently no usages of such an annotation in the codebase, after all these years of development, hence my question.

greg0ire:
> When use fetch or fetchColumn and don't pass for all registers if while they don't close the cursor and don't save in the cache

I do not understand. Can you please elaborate?

dfrnks:
> > When use fetch or fetchColumn and don't pass for all registers if while they don't close the cursor and don't save in the cache
> 
> I do not understand. Can you please elaborate?

When this query is made

        $qb = $db
            ->createQueryBuilder()
            ->select('*')
            ->from("base_config");

        $stmt   = $db->executeCacheQuery($qb, [], [], new QueryCacheProfile(3600));
        $config = $stmt->fetch();
        $stmt->closeCursor();

It is not saved in the cache, because it has not been passed in all registers and the variable $emptied in lib/Doctrine/DBAL/Cache/ResultCacheStatement.php is never true when we call closeCursor.

It will only be true if you pass all registers with while for example and then will save in the cache when we call closeCursor.

But is not the case of this query, this query only need the first register. And otherwise it will not influence if someone needs to pass all the registers, it will save in the cache too.

tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
dfrnks: The Scrutinizer said that don't have getIterator in ResultStatement in this is true, but here we call ResultCacheStatement indeed

greg0ire:
Pardon my ignorance but… what is a register? What are you referring to?

tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
dfrnks: Thanks

dfrnks:
> Pardon my ignorance but… what is a register? What are you referring to?

Sorry, I mean, without this pull request you need to do like this for save the cache using the fetch method

        while ($row = $stmt->fetch()) {
            $data[] = $row;
        }
       $stmt->closeCursor();

You need get all rows.

If you get just the first $row, they don't save the cache.

        $row = $stmt->fetch();
        $stmt->closeCursor();

And I think don't matter if you get one row or all rows to save the cache, it necessary save in the cache always.

tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
greg0ire: Scrutinizer is not wrong, sometimes it returns something else:

https://github.com/doctrine/dbal/blob/4c03ed81471c62f178581eb325339a3c34f3b71a/lib/Doctrine/DBAL/Connection.php#L954-L965

The thing is, `ArrayStatement` also has a `getIterator` method. Try changing this line:
https://github.com/doctrine/dbal/blob/4c03ed81471c62f178581eb325339a3c34f3b71a/lib/Doctrine/DBAL/Connection.php#L934

Use `ResultCacheStatement|ArrayStatement`, and you should be able to remove these annotations.

SenseException (approved):
:+1: 

@greg0ire I haven't checked all tests in the repository, but is it possible that we don't have a test that checks the cache behaviour with parameters?

tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
```diff
@@ -165,7 +165,7 @@ public function testDontFinishNoCache() : void
 
         $this->hydrateStmt($stmt, FetchMode::NUMERIC);
 
-        self::assertCount(2, $this->sqlLogger->queries);
+        self::assertCount(1, $this->sqlLogger->queries);
```
SenseException, line 168: Wait, isn't this test supposed to have 2 queries executed?

SenseException (changes requested):

greg0ire:
@SenseException what do you mean by "with parameters"?

tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
dfrnks: Not really, a second query he took from the cache

SenseException: It previously expected 2 queries in the test and I want to know what's the reason for this test. I'll need help from @beberlei, who implemented this test in #69 a long time ago.

SenseException:
@greg0ire Every `Connection::executeQuery()` with a `QueryCacheProfile` has an empty array for `$params`. Depending on an omitted key in `QueryCacheProfile` the `$params` are also used for the cache key creation and I wonder if this caching is covered by an integration test. This has nothing to do with this PR, but maybe needs to be covered by another PR. Or I haven't found an existing test covering this case yet.

tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
dfrnks: The test was incorrect because, just will save the query in $this->sqlLogger->queries when the query are execute in the database.

https://github.com/doctrine/dbal/blob/4c03ed81471c62f178581eb325339a3c34f3b71a/lib/Doctrine/DBAL/Connection.php#L894-L897

When the query are in the cache, they return before

https://github.com/doctrine/dbal/blob/4c03ed81471c62f178581eb325339a3c34f3b71a/lib/Doctrine/DBAL/Connection.php#L886-L897

The reason are the cache doesn't work before, when we use fetch

SenseException: I agree, which is why I approved if at first, but if there's another reason why @beberlei expected 2 in the past, I want to know and would like to be on the save side.

morozov: This was recently changed in https://github.com/doctrine/dbal/pull/4048. It's an improvement, not a bugfix, so this change shouldn't be made in `2.10.x.`.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -931,9 +931,10 @@ public function executeQuery($query, array $params = [], $types = [], ?QueryCach
      * @param int[]|string[]    $types  The types the previous parameters are in.
      * @param QueryCacheProfile $qcp    The query cache profile.
      *
-     * @return ResultStatement
+     * @return ResultCacheStatement|ArrayStatement
```
morozov, line 934: Please revert. The type hint should refer to an interface, not to a specific implementation.

tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
```diff
@@ -194,6 +194,24 @@ public function testFetchAllColumn() : void
...
+
+        $stmt = $this->connection->executeCacheQuery($query, [], [], $qcp);
+        $stmt->getIterator();
+        $stmt->closeCursor();
+
+        $stmt = $this->connection->executeCacheQuery($query, [], [], $qcp);
+
+        $iterator = $stmt->getIterator();
+
+        self::assertCount(1, $iterator);
```
morozov, line 212: This is not a valid behavior to expect. The fact that the iterator internally calls `fetchAll()` is a bug (see https://github.com/doctrine/dbal/pull/2718) and shouldn't be relied upon to expect that the statement is cached just because an iterator over its results was instantiated.

morozov (changes requested):
This is likely resolved by https://github.com/doctrine/dbal/pull/4048.

dfrnks:
With I understanding, my pull request isn't necessary anymore because they are resolved the other way in #4048 right? @morozov

morozov:
@dfrnks I believe so. You can check your case using the `2.11.x` branch.

Issue #4003, opened 2020-05-09
Pretty print sql


reedy:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes

#### Summary
While it doesn't matter for a computer, for humans it's easier to read things with whitespace etc.

Is there any chance of being able to get a "pretty print" mode, or some sort of SQL formatter to add newlines, whitespace etc for readability purposes?

andrew-demb:
https://github.com/doctrine/sql-formatter

reedy:
> https://github.com/doctrine/sql-formatter

Aha, thanks. Any chance of maybe getting that documented somewhere? :) Or is it, and I've just managed to completely miss it?

greg0ire:
> Aha, thanks. Any chance of maybe getting that documented somewhere? :) Or is it, and I've just managed to completely miss it?

Well it comes with a fairly complete README IMO.

greg0ire:
> Or even in suggest in composer.json? 

:thinking: I'm not sure… in what circumstances do you get SQL from the DBAL?

reedy:
> > Aha, thanks. Any chance of maybe getting that documented somewhere? :) Or is it, and I've just managed to completely miss it?
> 
> Well it comes with a fairly complete README IMO.

"it" being what exactly? https://github.com/doctrine/dbal/blob/2.10.x/README.md is pretty limited ;)

The problem is discoverability

> > Or even in suggest in composer.json?
> 
> 🤔 I'm not sure… in what circumstances do you get SQL from the DBAL?

Isn't that the whole point of the library? To build SQL?

greg0ire:
I was referring to the README of https://github.com/doctrine/sql-formatter so, so very obviously ;)

> Isn't that the whole point of the library? To build SQL?

The whole point to me would be executing it. Rarely do I get SQL from the DBAL (but it is still possible)

reedy:
> I was referring to the README of https://github.com/doctrine/sql-formatter so, so very obviously 😉

I think we crossed wires a little. I was meaning documenting it somewhere in here/the docs for DBAL - "Hey, if you want to pretty print your SQL use https://github.com/doctrine/sql-formatter"

Not going into the extent of how to use it etc. Mostly discoverability. If you know the right keywords you can probably find the sql formatter repo.. If you don't, you might not

Hence the 'suggest' suggestion:

`"doctrine/sql-formatter": "SQL statement formatter. Can be used for syntax highlighting too"`

[...]

PR #4005, opened 2020-05-09
Add doctrine/sql-formatter to composer suggest

reedy:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Adds doctrine/sql-formatter to composer.json suggest for visibility

Follows up on discussion in #4003

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -49,6 +49,7 @@
         "vimeo/psalm": "^3.11"
     },
     "suggest": {
+        "doctrine/sql-formatter": "SQL statement formatter. Can be used for syntax highlighting too,",
         "symfony/console": "For helpful console commands such as SQL execution and import of files."
     },
     "bin": ["bin/doctrine-dbal"],


greg0ire (approved):

composer.json
```diff
@@ -49,6 +49,7 @@
         "vimeo/psalm": "^3.11"
     },
     "suggest": {
+        "doctrine/sql-formatter": "SQL statement formatter. Can be used for syntax highlighting too,",
```
morozov, line 52: I don't see it being used by the library. What does the suggestion mean?

reedy: For discoverability purposes

morozov: > Suggested packages that can enhance or work well with this package.

I'm not really sold. The formatted doesn't enhance the work of DBAL and can work well with formatting any SQL, not only the one generated by DBAL.

An example of appropriate suggestions for DBAL would be `ext-mysql`, `ext-oci8`, etc. that the library can use but they are not required.

reedy: Sure. But a "suggest" doesn't mean it only works with something, does it?

Maybe it should be wired in? ;)

The problem as per https://github.com/doctrine/dbal/issues/4003 is it wasn't obvious at all that you had a formatter available.

reedy: Also, by the definition of "work well with this package"... this seems to meet that

morozov: It doesn't work with the package. It works with SQL. Does it mean that any package that generates SQL should suggest this formatter? Since the formatting is about developer experience, should DBAL suggest all development tools like Xdebug, etc.?

This looks totally unnecessary to me.

reedy:
I give up

Issue #3744, opened 2019-11-23
Named statement parameters are unnecessarily converted to positional on Oracle


morozov:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | 2.10.0

#### Summary

The current library design around handling named and positional parameters in prepared statements is a bit of a mess.

#### Current behaviour

1. When a statement containing named parameters is executed via the wrapper connection's `executeQuery()`, the parameters get _unconditionally_ converted to the positional ones via `SQLParserUtils::expandListParameters()`.
2. The `OCI8Statement` class converts positional parameters back to the named ones via `OCI8Statement::convertPositionalToNamedPlaceholders()`.

Therefore:

1. Parameter names get lost in translation: `SELECT :COL1 COL1 FROM DUAL` → `SELECT :param1 COL1 FROM DUAL`.
2. Extra overhead on query parsing and conversion back and forth.
3. The wrapper connection implements driver-specific logic which is wrong by design.

#### How to reproduce

Currently, there's no mechanism to observe the queries in the form in which they hit the underlying DB driver. See the data flow in the new data item added to `queryConversionProvider()` added in #3738.

#### Expected behavior

1. The wrapper statement doesn't contain any conversion logic.
2. The driver statement converts the unsupported parameters to the supported ones.
3. No unnecessary conversion happens.

RafaelKr:
I just ran into this while writing a migration. It throws with `An exception occurred while executing a query: SQLSTATE[HY093]: Invalid parameter number: mixed named and positional parameters`.

My query to update Postgres JSONB columns:
```sql
UPDATE
    "my_table"
SET
    data[(:person)][(:key)] = NULL
WHERE
    type='financing'
    AND data[(:person)][(:key)]['#scalar'] = :removed_case_value
```
used via:
```php
$this->addSql($sql, [
    'person' => $person,
    'key' => $key,
    'removed_case_value' => \sprintf('"%s"', $removedCaseValue),
]);
```

results in:
```sql
UPDATE
    "my_table"
SET
    data[(:person)][(:key)] = NULL
WHERE
    type='financing'
    AND data[(:person)][(:key)]['#scalar'] = ?
```
with parameters:
```php
[
  0 => '"value"',
]
```

**Edit:**
I think there is another issue coming into play here. The Parser doesn't recognize named parameters wrapped inside square brackets (`(:person)` and `(:key)`). It's choosing the `sqlPattern`
[...]

Issue #3591, opened 2019-06-05
Separate DBAL and driver interfaces


morozov:
Please refer to the conversation in https://github.com/doctrine/dbal/pull/3582#discussion_r289579071. More details to be added.

While currently the wrapper and driver connections and statements implement the same connection and statement interfaces, they don't implement them in the same way and completely. At the same time, the interfaces represent a superset of the contract of the wrapper and driver classes.

1. The wrapper `Connection` class has two pairs of methods: `executeQuery()` + `executeUpdate()` and `query()` + `exec()` where the pairs are almost identical to the second. The first pair is mostly used at the DBAL level but the second is needed to implement the driver-level interface.
2. Driver statements have to support all fetch modes exposed via the abstraction (e.g. mixed or class) and therefore implement them instead of having the abstraction implementing the fetch modes.

morozov:
Closed via https://github.com/doctrine/dbal/pull/4159.

PR #4007, opened 2020-05-13
Replace the concept of statement fetch modes with the explicit API

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | hell yes

The approach of using `FetchMode` when fetching data from a statement inherited from PDO has certain downsides:
1. The type of the return value of a given method (e.g. `fetch()`) is determined at runtime (it can be a `mixed`, `array<int,mixed>`, `array<string,mixed>` and even `array<int|string,mixed>`) and is unclear for static analysis.
2. Besides the mode passed explicitly to a `fetch()` method, it can be set on the statement via `setFetchMode()`, so if a component gets passed an existing `ResultStatement()`, it cannot be certain of the fetched data type unless it specifies it explicitly.
3. Even if the component executes the statement itself, it cannot be certain about the fetched data type because it can be set globally at the connection level.
4. Having the fetch mode expressed via a set of constants requires their additional validation in the method and throwing an exception if the mode is invalid or unsupported. This puts us in an awkward position: a lot of methods can throw an invalid mode exception but it's not handled anywhere.
5. Unlike calling fetch methods where a method can be passed explicitly, it's impossible to pass the fetch mode to the statement when iterating over it. It's only possible to set it _on_ the statement and therefore modify the shared state that others may rely on.

The proposed solution is to define an explicit fetch-oriented API that driver-level statements have to implement. This API will define a dedicated method for each combination of the following:
1. Row presentation mode (associative array, numeric array, the value of the first column).
2. Row traversal mode (one row, all rows, iterator).

| Statement   | Numeric |      Associative       | One/Column |      All       |  Iterator    |
| :--------- | :----: | :----: | :----: | :------------------: | :-----------: |
| `mysqli` | ✓ |   |   |   |   |
| `pdo` | ✓ | ✓ | ✓ | ✓ | ✓ |
| `sqlsrv`     | ✓ | ✓ |   |   |   |
| `oci`    | ✓ | ✓ |   ✓    | ✓ |   |
| `ibm_db2` | ✓ | ✓ | | | |
| Cache    |   | ✓ | |          ✓           |       ✓       |

The above matrix illustrates the driver capabilities that I want to be exposed via the new API and is meant to justify its verbosity:

1. **Numeric**. Support for fetching rows as numeric arrays.
2. **Associative**. Support for fetching rows as associative arrays.
3. **One/Column**. Support for fetching one value or a column.
4. **All**. Support for all rows in one call.
5. **Iterator**. Support for iteration over rows.

As one can see, the numeric and associative modes are supported by most of the drivers and they are really the essential primitives for fetching data. The “one”, “all” and “iterator” capabilities are supported only by a few but I believe it's important to be able to use their in-driver implementation instead of having to emulate them in DBAL even if the driver supports it (see a similar problem in #3744).

**Breaking changes:**
1. The `FetchMode` class and the `setFetchMode()` methods are removed as well as the entire concept of the fetch mode.
2. The `fetch()` method is replaced with `fetchNumeric()`, `fetchAssociative()` and `fetchOne()` (like [`fetchOne()` in ZF1](https://framework.zend.com/manual/1.12/en/zend.db.adapter.html#zend.db.adapter.select.fetchone)).
3. The `fetchAll()` method is replaced with `fetchAllNumeric()`, `fetchAllAssociative()` and… should it be `fetchColumn()` like [`fetchCol()` in ZF1](https://framework.zend.com/manual/1.12/en/zend.db.adapter.html#zend.db.adapter.select.fetchcol)?
4. `Connection::fetchAssoc()` and `::fetchArray()` are renamed to `::fetchAssociative()` and `::fetchNumeric()` respectively for consistency.
5. In order to make the Driver and the DBAL APIs distinct, a DBAL-level `ResultStatement` interface is introduced. It extends the one at the Driver level (therefore, it can gain new DBAL-level features w/o having to implement them at the Driver level).
6. The statements are no longer traversable. Only DBAL-level statements may be iterated via `iterateNumeric()`, `iterateAssociative()` or `iterateColumn()`.
7. The `StatementIterator` class is removed.
8. Fetching in the “mixed”/“both” mode is no longer supported. It's not a limitation of the new design. I just haven't seen a single case in my entire life where it would be needed. Well… the caching layer could use it but it doesn't.

**TODO:**
- [x] Add the new methods to the wrapper connection class for consistency.
- [x] Provide the upgrade path and forward compatibility layer.
- [x] Adjust code comments and documentation.
- [x] Document the change in `UPGRADE.md`.

**The next steps will be:**
1. Separate driver-level API from the wrapper-level (#3591). This will allow independent evolution of the wrapper-level API w/o the need to make any changes to the drivers (e.g. add new fetch modes).
2. Replace `ResultStatement` with the `Result` interface. This will allow avoiding awkward situations like:
   1. A statement exists but cannot be fetched from because it wasn't executed.
   2. A component is given a statement to fetch the data from but it can bind new parameters to it and execute it.


greg0ire:
Why not provide the upgrade path upfront and target 2.11.x? Since we do merges up now, it makes more sense to use that flow to keep branches in sync IMO, with a first PR that includes a BC layer and deprecations, a merge up PR, and a PR that drops deprecated code.

src/Driver/ResultStatement.php
```diff
@@ -37,40 +37,83 @@ public function columnCount() : int;
     public function rowCount() : int;
 
     /**
-     * Sets the fetch mode to use while iterating this statement.
+     * Returns the next row of a result set as a numeric array or FALSE if there are no more rows.
      *
-     * @param int $fetchMode Controls how the next row will be returned to the caller.
-     *                       The value must be one of the {@link \Doctrine\DBAL\FetchMode} constants.
+     * @return array<int,mixed>|false
```
greg0ire, line 42: I think PDO returns `false` in its API because it makes methods usable with a `while` statement.
If we are being strict though, it implies casting an array to a boolean in most cases. How about we use `null` instead here? `null` means absence of data, so I think it makes more sense from a semantic PoV, and as a bonus, we can have a return type declaration here.

src/Driver/FetchUtils.php
```diff
@@ -0,0 +1,113 @@
...
+
+        return $row[0];
+    }
+
+    /**
+     * @return array<int,array<int,mixed>>
+     *
+     * @throws DriverException
+     */
+    public static function fetchAllNumericByOne(ResultStatement $stmt) : array
```
greg0ire, line 35: Why the `ByOne` suffix here and below? I don't understand what it means.

src/Driver/OCI8/OCI8Statement.php
```diff
@@ -219,120 +207,131 @@ public function execute(?array $params = null) : void
...
+    }
 
-        if ($row === false) {
-            return false;
+    /**
+     * @return array<mixed>
+     */
+    private function fetchAll(int $mode, int $fetchStructure) : array
+    {
+        // do not try fetching from the statement if it's not expected to contain result
```
greg0ire, line 291:
```suggestion
        // do not try fetching from the statement if it's not expected to contain results
```

greg0ire (commented):
This new API is great, many bugs will be caught earlier thanks to it I think :+1:

src/Driver/ResultStatement.php
morozov: This is a valid point. Although `null` is indeed more semantically correct, the reason why PDO uses `false` to indicate the end of result is because `$stmt->fetch(PDO::FETCH_COLUMN)` may return `null` as a value.

Regardless of whether return `?array` or `array|false` in `fetchNumeric()` and `fetchAssociative()`, we'll have to return `mixe|false` in `fetchOne()`. The fact that all three can be used in a `while ()` statement makes me believe that consistency is more important than purity in each individual case.

What do you think?

src/Driver/FetchUtils.php
morozov: These methods implement `fetchAll*()` statement methods by fetching each row individually (one by one). Not sure how to put it clear and compact at the same time.

src/Driver/OCI8/OCI8Statement.php
morozov: It will be “the result” since it refers to the `$result` property or “statement result” in general (like `mysqli_result`).

src/Driver/ResultStatement.php
greg0ire: It makes sense, let's keep it the way you wrote it 👍

src/Driver/FetchUtils.php
greg0ire: It's an implementation detail, isn't it? Why does it need to be reflected in the name of the method? And if it really does, since all methods in that class seem to have that same behavior, let's maybe reflect it in the class name? `OneByOneFetcher`, `IteratingFetcher`, something like that?

morozov: > It's an implementation detail, isn't it? Why does it need to be reflected in the name of the method?

That's a good point. In the beginning, I thought we might have different implementations of the same fetch mode for different cases (e.g. “fetch one value from a numeric array” vs “fetch one value from an associative array”) but it didn't happen. We can remove the suffixes and leave just the methods matching the statement method names.

Would that be okay?

greg0ire: Sounds perfect!

morozov:
> Why not provide the upgrade path upfront and target 2.11.x?

Makes sense. I'd like to finish all the work here first, and before merging it, extract the relevant parts into a patch against `2.11.x`.

morozov:
Not sure how to proceed with the forward compatibility layer for `2.x`. We can easily introduce the new methods in the wrapper connection class but what about the driver layer? Unlike a class where we can add methods, adding methods to the driver connection interface is a BC break on its own. If we just add them to the driver connection classes, then in order to use them, the code that depends on the interface will have to abuse its knowledge of the specific classes that implement the old interface and the new methods. Any ideas, @greg0ire?

greg0ire:
You could create a new interface, and have the code that depends on the interface check if the new one is implemented, and complain if it isn't.

morozov:
> This plan leaves us with one issue: naming.

The problem is with the design, not just with naming. See my notes in the description regarding `ResultStatement`. I already started prototyping the new API but the changes will have a lot in common with this one, so I decided to not proceed before this one is done.

morozov:
Thanks for the code example, @greg0ire. It helps to understand the idea significantly. Let's see if it's feasible to apply it at the scale of all components that depend on the `Statement` interface.

greg0ire:
The other PR I just merged seems to introduce conflicts

src/Cache/ResultCacheStatement.php
```diff
@@ -240,63 +129,45 @@ public function fetchOne()
...
-            $data = $this->statement->fetchAllAssociative();
-        } else {
-            $data = $this->statement->fetchAll(FetchMode::ASSOCIATIVE);
-        }
+        $this->store(
+            $this->statement->fetchAllAssociative()
+        );
 
-        return $this->store($data);
+        return array_map('array_values', $this->data);
```
morozov, line 136: This will need to be backported to `2.11.x` (found by PHPStan and enforced by the refactored `ResultCacheTest`).

morozov:
@greg0ire it's ready now.

docs/en/reference/caching.rst
```diff
@@ -42,7 +42,7 @@ object is closed:
 
     <?php
     $stmt = $conn->executeCacheQuery($query, $params, $types, new QueryCacheProfile(0, "some key"));
-    $data = $stmt->fetchAll();
+    $data = $stmt->fetchAllAssociative();
```
greg0ire, line 45: Shouldn't the new docs be backported to the lower branch?

src/ForwardCompatibility/Driver/ResultStatement.php
```diff
@@ -1,59 +0,0 @@
...
-namespace Doctrine\DBAL\ForwardCompatibility\Driver;
-
-use Doctrine\DBAL\Driver\DriverException;
-use Doctrine\DBAL\Driver\ResultStatement as BaseResultStatement;
-
-/**
- * Forward compatibility extension for the ResultStatement interface.
- */
-interface ResultStatement extends BaseResultStatement
-{
```
greg0ire, line 14: Should we maybe mark this as internal in lower branches? Since we remove it here, we don't want people to type hint against it. If it's ok to type hint against it, maybe we should keep it, deprecate it, and empty it.

docs/en/reference/caching.rst
morozov: Yes.

src/ForwardCompatibility/Driver/ResultStatement.php
morozov: Let's not mark it internal so that people don't feel wrong about using it, and let's empty and deprecate it instead of removal in `3.0.x`.

Most likely, this interface will go away or get renamed/repurposed before `2.11.x` is released but I'm fine with keeping it always releasable.

greg0ire (approved):
Not sure if it's the right day to use that emoji, but… :rocket:

lcobucci:
@morozov nice job, mate :+1: 

The API is much nicer and explicit but it has a huge impact on the ORM, so I think it's crucial to align these changes with @guilhermeblanco.

morozov:
@lcobucci thanks for the heads up. I hope the impact is huge in terms of the number of LOC that need to be changed, not in terms of complexity. As far as I can tell, most of the code can be fixed almost by find and replace.

Since the same API is already implemented in `2.11.x`, I believe we can proceed with the merge, and if there are any specific issues with the new API that need to be addressed, we'll have to address them starting `2.11.x` anyways.

In the meanwhile, merging this will unblock me in reworking the driver/wrapper and statement/result interfaces.

Issue #4010, opened 2020-05-17
SchemaDiff unable to output diff for SqlitePlatform with foreign key changes 


camilleterol:
### Bug Report
|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.11.x

#### Summary

When I output a diff using SchemaDiff from the migrations:diff command, I get the following error thrown at me:

```
In SqlitePlatform.php line 805:
                                                                                                                 
  [Doctrine\DBAL\DBALException]                                                                                  
  Sqlite platform does not support alter foreign key, the table must be fully recreated using getAlterTableSQL.
```

#### Current behaviour

I cannot generate a diff with SchemaDiff for the SqlitePlatform if there has been any foreign key changes to the mapping.

#### How to reproduce

Alter the mapping in such a way that the diff will try to add / delete foreign keys on existing table.

#### Expected behaviour

The diff should be generated without any problems.

camilleterol:
Seems related to #3990

PR #4011, opened 2020-05-17
Fix a condition that would not allow diffing of foreign keys for Sqlite

camilleterol:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | #1204, #4010 

#### Summary

After the changes introduced by #1204, the SchemaDiff would throw an error when trying to diff an Sqlite-backed schema with foreign keys modifications.


greg0ire:
This looks a lot like https://github.com/doctrine/dbal/pull/3994 … maybe there are more occurrences of `supportsForeignKeyConstraints` that should be re-evaluated?

In any case, please add a test.

camilleterol:
Thanks for the swift reply. I'll look into adding a test for this specific use case.

greg0ire:
`git rebase --onto origin/2.10.x origin/2.11.x 2.11.x` would have done the trick ;)

PR #4016, opened 2020-05-20
Prevent column flagging as diff just because type differs (when one type is parent to other)

jesparic:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | No direct issue but builds on change made for #3427 

#### Summary
The ability to extend the dbal type system allows us to define our own Enum implementations. Often these extend from the StringType built-in type class. Everything works as it should except for one caveat - when generating migration diffs the Comparator::diffColumn() will detect that the two types are not identical and add the column to the generated migration diff every time. This leaves us with 2 main choices:
1. Manually remove the diffs from the generated migration files each time (not ideal and prevents us from using the diff in CI tasks effectively)
2. Add Type::requiresSQLCommentHint() == true method to our extension class

Although a perfectly reasonable solution, in some cases #2 is less than ideal (i.e., if we don't want to make modifications to the schema columns). For example, when porting a legacy app where the source of truth (and legacy migration system) is still the old app. In this case, we use the diff not to migrate the changes, but instead to make our schema continually match the legacy app while the migration work continues (i.e., the legacy app can and will still change).

To satisfy the above use case, I have presented this pull request. I have added a small modification to the Comparator::diffColumn() method which basically checks if the second check type is a subclass of the parent and that the parent is a basic string type. If this is the case, then the 'type' diff is not added; i.e., Only detected differences in options will then cause the schema diff to report change to the column

To summarise, I think implementing an enum as an extension of the string class is a very common case and I don't see a strong reason to enforce/require `requiresSQLCommentHint()` for the type in this case (Doctrine is perfectly able to detect and hydrate the type correctly at runtime)

Kind regards,
Richard

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Comparator.php
+++ b/lib/Doctrine/DBAL/Schema/Comparator.php
@@ -13,6 +13,7 @@ use function array_unique;
 use function assert;
 use function count;
 use function get_class;
+use function is_a;
 use function strtolower;
 
 /**
@@ -426,7 +427,11 @@ class Comparator
 
         $changedProperties = [];
 
-        if (get_class($properties1['type']) !== get_class($properties2['type'])) {
+        // Exception here helps with specific case of base string-type and a custom enum-type (inheriting from string).
+        // When doing migration diffs, there is no need to flag the column as different (unless options differ too)
+        $typeClass1 = get_class($properties1['type']);
+        $typeClass2 = get_class($properties2['type']);
+        if ($typeClass1 !== $typeClass2 && ! $this->isStringParentOf($typeClass1, $typeClass2)) {
             $changedProperties[] = 'type';
         }
 
@@ -506,6 +511,15 @@ class Comparator
         }
 
         return array_unique($changedProperties);
+    }
+
+    private function isStringParentOf(string $type1, string $type2) : bool
+    {
+        if ($type1 !== Types\StringType::class) {
+            return false;
+        }
+
+        return is_a($type2, $type1, true);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


SenseException (commented):
I'm not sure about this feature being in 2.10. If this will be considered to be added to DBAL, I would suggest the next minor version.

lib/Doctrine/DBAL/Schema/Comparator.php
```diff
@@ -426,7 +427,11 @@ public function diffColumn(Column $column1, Column $column2)
 
         $changedProperties = [];
 
-        if (get_class($properties1['type']) !== get_class($properties2['type'])) {
+        // Exception here helps with specific case of base string-type and a custom enum-type (inheriting from string).
+        // When doing migration diffs, there is no need to flag the column as different (unless options differ too)
+        $typeClass1 = get_class($properties1['type']);
+        $typeClass2 = get_class($properties2['type']);
+        if ($typeClass1 !== $typeClass2 && ! $this->isStringParentOf($typeClass1, $typeClass2)) {
```
morozov, line 434: Some types require a comment hint:

https://github.com/doctrine/dbal/blob/dbc623d40ad18c24274735c1c85fdaa1dee3323f/lib/Doctrine/DBAL/Types/ArrayType.php#L69-L72

Even if the two types are implemented by a child and the parent classes, they may produce different DDL and require a migration.

jesparic: Hi all. Thank you for taking time to look at this PR and sorry for the very late response!

In hindsight, I think this change was a bit naive and too niche to be part of the core mechanism as you have suggested. The tricky part for me is that the dbal migration code is really closed off (no dependency injection so I can't really get at the code in any sane way) and pretty much impossible to modify in the way I need above in user-land. I think my options are to:
1. - Just add the comments to the enum columns to help dbal out (and the risks that dual migration managers entail). This may not be too bad but I envision issues when the current migration system e.g., removes/renames a column
2. - Parse and remove unwanted enum/string diffs in a post process
3. - Continue to maintain a custom fork until we can switch the migrations source of truth over to Doctrine

I'm prob gonna go with the latter..

Happy to close this PR. Thanks again for your time

PR #4019, opened 2020-05-23
Deprecated the concept of the fetch mode

morozov:
|      Q       |   A
|------------- | -----------
| Type         | feature, deprecation
| BC Break     | no

**Proposed changes:**
1. The `FetchMode` class, the `UnknownFetchMode` exception and the `setFetchMode()` methods are deprecated as well as the entire concept of the fetch mode.
2. The `fetch()` method is deprecated in favor of `fetchNumeric()`, `fetchAssociative()` and `fetchOne()` (like [`fetchOne()` in ZF1](https://framework.zend.com/manual/1.12/en/zend.db.adapter.html#zend.db.adapter.select.fetchone)).
3. The `fetchAll()` method is deprecated in favor of `fetchAllNumeric()` and `fetchAllAssociative()`. There's no currently replacement for `fetchAll(FETCH_MODE::COLUMN)`. In a future major version, `fetchColumn()` will be used as a replacement (like [`fetchCol()` in ZF1](https://framework.zend.com/manual/1.12/en/zend.db.adapter.html#zend.db.adapter.select.fetchcol)).
4. `Connection::fetchAssoc()` and `::fetchArray()` are deprecated in favor of `::fetchAssociative()` and `::fetchNumeric()` respectively for consistency.
5. In order to make the Driver and the DBAL APIs distinct, two forward-compatible result statement interfaces are introduced. The one at the DBAL level extends the one at the Driver level (therefore, it can gain new DBAL-level features w/o having to implement them at the driver level).
6. Iteration over statements is deprecated. Only DBAL-level statements may be iterated via `iterateNumeric()`, `iterateAssociative()` or `iterateColumn()`.
7. The `StatementIterator` class is deprecated.
8. Fetching in the “mixed”/“both” mode is deprecated.

See https://github.com/doctrine/dbal/pull/4007.

**TODO:**
- [x] ~~See if the deprecated methods can be partially reimplemented using the new ones (the new ones don't support mixed fetch mode).~~ Too much work for throwaway code.
- [x] Add tests for the newly added APIs.
- [x] Add tests to ensure that forward-incompatible implementations still work.
- [x] Update slevomat/coding-standard in order to avoid false detection of the code like the following (cannot find the corresponding fixed issue):
   ```php
   if ($condition) {
       yield from [];
   } else {
       yield from [];
   }
   ```

The existing tests are changed to the necessary minimum (only the expected method invocations) and do not represent intentional breaking changes.

### DIFF
--- a/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
+++ b/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
@@ -4,9 +4,12 @@ namespace Doctrine\DBAL\Cache;
 
 use ArrayIterator;
 use Doctrine\Common\Cache\Cache;
+use Doctrine\DBAL\Driver\DriverException;
+use Doctrine\DBAL\Driver\FetchUtils;
 use Doctrine\DBAL\Driver\ResultStatement;
 use Doctrine\DBAL\Driver\Statement;
 use Doctrine\DBAL\FetchMode;
+use Doctrine\DBAL\ForwardCompatibility\Driver\ResultStatement as ForwardCompatibleResultStatement;
 use InvalidArgumentException;
 use IteratorAggregate;
 use PDO;
@@ -28,7 +31,7 @@ use function reset;
  * Also you have to realize that the cache will load the whole result into memory at once to ensure 2.
  * This means that the memory usage for cached results might increase by using this feature.
  */
-class ResultCacheStatement implements IteratorAggregate, ResultStatement
+class ResultCacheStatement implements IteratorAggregate, ResultStatement, ForwardCompatibleResultStatement
 {
     /** @var Cache */
     private $resultCache;
@@ -105,6 +108,8 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement
 
     /**
      * {@inheritdoc}
+     *
+     * @deprecated Use one of the fetch- or iterate-related methods.
      */
     public function setFetchMode($fetchMode, $arg2 = null, $arg3 = null)
     {
@@ -115,6 +120,8 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement
 
     /**
      * {@inheritdoc}
+     *
+     * @deprecated Use iterateNumeric(), iterateAssociative() or iterateColumn() instead.
      */
     public function getIterator()
     {
@@ -125,6 +132,8 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement
 
     /**
      * {@inheritdoc}
+     *
+     * @deprecated Use fetchNumeric(), fetchAssociative() or fetchOne() instead.
      */
     public function fetch($fetchMode = null, $cursorOrientation = PDO::FETCH_ORI_NEXT, $cursorOffset = 0)
     {
@@ -165,6 +174,8 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement
 
     /**
      * {@inheritdoc}
+     *
+     * @deprecated Use fetchAllNumeric(), fetchAllAssociative() or fetchColumn() instead.
      */
     public function fetchAll($fetchMode = null, $fetchArgument = null, $ctorArgs = null)
     {
@@ -184,6 +195,8 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement
 
     /**
      * {@inheritdoc}
+     *
+     * @deprecated Use fetchOne() instead.
      */
     public function fetchColumn($columnIndex = 0)
     {
@@ -191,6 +204,64 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement
 
         // TODO: verify that return false is the correct behavior
         return $row[$columnIndex] ?? false;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fetchNumeric()
+    {
+        $row = $this->doFetch();
+
+        if ($row === false) {
+            return false;
+        }
+
+        return array_values($row);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fetchAssociative()
+    {
+        return $this->doFetch();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fetchOne()
+    {
+        return FetchUtils::fetchOne($this);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fetchAllNumeric() : array
+    {
+        if ($this->statement instanceof ForwardCompatibleResultStatement) {
+            $data = $this->statement->fetchAllAssociative();
+        } else {
+            $data = $this->statement->fetchAll(FetchMode::ASSOCIATIVE);
+        }
+
+        return $this->store($data);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fetchAllAssociative() : array
+    {
+        if ($this->statement instanceof ForwardCompatibleResultStatement) {
+            $data = $this->statement->fetchAllAssociative();
+        } else {
+            $data = $this->statement->fetchAll(FetchMode::ASSOCIATIVE);
+        }
+
+        return $this->store($data);
     }
 
     /**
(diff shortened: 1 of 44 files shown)


tests/Doctrine/Tests/DBAL/Schema/MySqlSchemaManagerTest.php
```diff
@@ -23,19 +23,23 @@ class MySqlSchemaManagerTest extends TestCase
...
     {
-        $eventManager  = new EventManager();
-        $driverMock    = $this->createMock(Driver::class);
-        $platform      = $this->createMock(MySqlPlatform::class);
+        $eventManager = new EventManager();
+        $driverMock   = $this->createMock(Driver::class);
+
+        $platform = $this->createMock(MySqlPlatform::class);
+        $platform->method('getListTableForeignKeysSQL')
+            ->willReturn('');
```
morozov, line 31: This needs to be mocked because the newly called method has a scalar parameter type hint but the method being mocked doesn't have a return type hint, so PHPUnit cannot mock it correctly. This is to be removed in `master`.

lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
```diff
@@ -210,4 +312,51 @@ public function rowCount()
...
+
+        return false;
+    }
+
+    /**
+     * @param  array<int,array<mixed>> $data
+     *
+     * @return array<int,array<mixed>>
+     */
+    private function store(array $data, bool $isArray) : array
```
greg0ire, line 349: :thinking: what's the point of that second argument? It's always `true`, and the first argument is declared as `array` in the signature

morozov: It was needed before the introduction of `FetchUtils` to implement the "column" fetch modes. It's no longer needed, I’ll remove it.

morozov:
Here's some food for thought:
https://github.com/doctrine/dbal/blob/115f20e12172f10f77b01cb680ff9a97fc6a6c4d/lib/Doctrine/DBAL/Driver/PDOStatement.php#L281-L283

1. We're getting rid of the fetch mode as such and from the APIs that allow setting it as a shared state.
2. We're introducing `iterate*()` methods at the driver statement level because some drivers natively implement it.
3. In fact, PDO is the only such driver, and "natively" here is a stretch since PDO itself is a wrapper. Being able to leverage this driver capability requires relying on the shared state that we want to avoid.

With that said, I think we should not introduce `iterate*` methods at the driver level and leave them only at the DBAL level.

greg0ire (approved):

derrabus:
> There's no currently replacement for `fetchAll(FETCH_MODE::COLUMN)`.

If I had such a piece of code in my application and wanted to remove calls to deprecated functionality, what would be my upgrade path here?

morozov:
> If I had such a piece of code in my application and wanted to remove calls to deprecated functionality, what would be my upgrade path here?

There's no immediate upgrade path. You'll have to keep using the deprecated API until the next major release where it will be implemented by `fetchColumn()`.

greg0ire:
@morozov I didn't spot this issue, is that why you suggested to name the new method `fetchCol()`? It's not a perfect name because of the abbrevation, but it will make the upgrade experience better (and maybe we can then migrate to `fetchColumn()` in 4.x)

derrabus:
First of all, thank you very much for you contribution. I really think that this is a good step forward to improve the public API of DBAL. :heart: 

And I totally understand that the new behavior is a way better fit for `fetchColumn()` that the current one. However, I really think that having a method change its behavior without prior notice is an unnecessary source of errors. My fear is that it'll prevent an incremental upgrade path which would be interesting especially for large codebases and might ultimately lead to reluctance to upgrade.

What we could do to improve the upgrade path here:

* If you really really want to keep that method name:
    * Introduce a flag to `fetchColumn()` which will make the function behave "the new way" in 2.11 already.
    * Trigger a deprecation error in 2.11, if the flag is not explicitly set to true.
    * In 3.0, remove the old behavior and throw if the flag is explicitly set to false.
    * In 3.1, trigger a deprecation if the flag is still passed.
    * In 4.0, remove the flag.
* If we consider the name `fetchColumn()` burned:
    * Find another nice name for your new behavior, like `fetchSingleColumn()`. Yeah, naming things is hard.
    * In 2.11, trigger a deprecation error is `fetchColumn()` is called.
    * In 3.0, remove `fetchColumn()`.

From my point of view, the second path should be easier to maintain, so my personal suggestion would be to pick that one. And I believe that we can come up with a nice method name.

morozov:
Thank you for the suggestion @derrabus.

> And I believe that we can come up with a nice method name.

I believe this is the name I'd like to see in the public API. So I'd prefer to have it eventually and migrate towards it somehow. Also, I wouldn't like to have to rely on runtime deprecations.

How about this:
1. Introduce `fetchFirstColumn()` in `2.11.x` that would fulfill the role of `fetchAll(FETCH_MODE::COLUMN)`.
2. Drop `fetchColumn()` in `3.0`.
3. Deprecate `fetchFirstColumn()` and reintroduce `fetchColumn()` in `3.last`. Depending on how things go, we may decide that the existing name is even better is stop here.
4. Otherwise, drop `fetchFirstColumn()` in `4.0` and enjoy using `fetchColumn()`.

This way, eventually we will have a nice name instead of something else, and there will be enough time between `2.11.0` and `3.last` to get rid of the old `fetchColumn()` in their code.

What do you say?

morozov:
> @morozov I didn't spot this issue, is that why you suggested to name the new method `fetchCol()`? It's not a perfect name because of the abbrevation, but it will make the upgrade experience better (and maybe we can then migrate to `fetchColumn()` in 4.x)

Sorry @greg0ire I just noticed your message. No, `fetchCol()` was used as a reference to ZF1 where "fetch column" fetches all values of a column as opposed to PDO and DBAL where "fetch column" fetches just one value.

Let's rename it to something else than `fetchColumn()` in `2.11.x`. Actually, I like `fechCol()` more than `fetchFirstColumn()` as a temporary name because it's uglier and we'll want to rework it.

derrabus:
> How about this:
> 
>     1. Introduce `fetchFirstColumn()` in `2.11.x` that would fulfill the role of `fetchAll(FETCH_MODE::COLUMN)`.
> 
>     2. Drop `fetchColumn()` in `3.0`.
> 
>     3. Deprecate `fetchFirstColumn()` and reintroduce `fetchColumn()` in `3.last`. Depending on how things go, we may decide that the existing name is even better is stop here.
> 
>     4. Otherwise, drop `fetchFirstColumn()` in `4.0` and enjoy using `fetchColumn()`.

That sounds like a reasonable plan. And I like `fetchFirstColumn()` as method name for the new behavior.

> Actually, I like `fechCol()` more than `fetchFirstColumn()` as a temporary name because it's uglier and we'll want to rework it.

I'm unsure if we should intentionally pick an ugly name, tbh. The abbreviation would be especially odd since the PDO-style abbreviated methods `fetchAssoc()` and `fetchNum()` are about to be replaced by the non-abbreviated versions `fetchAssociative()` and `fetchNumeric()`.

morozov:
@greg0ire, @derrabus, please see #4037.

tswcode:
I just stumbled upon the deprecated methods.
Everything is cleary documentated, but i cannot find the special PDO fetch method for "FETCH_KEY_PAIR".
Let's assume a sql, summarizing entity ids and a special count. Then I need the workflow like the old one:

`$counts = $sqlStatement->fetchAll(PDO::FETCH_KEY_PAIR, 0);
        $sqlStatement->free();`

The result has been an array with the ids as keys and counts as values.
Will you include this in the future or do I have to do implement my own method for this?

morozov:
@tswcode `PDO::FETCH_KEY_PAIR` was never supported by DBAL. It was and is supported by PDO but not by other drivers. I was going to start implementing it but then realized that it's just as simple as:
```php
foreach ($statement->iterateNumeric() as [$key => $value]) {
    // your code goes here
}
```
At the same time, implementing it in the DBAL will require it to handle the error when the resulting record set contains fewer than two columns (not sure how PDO handles that). If it's implemented in the userland, it's clear that the user should do the error handling.

Is there a case when the above doesn't work for you and you need it implemented in DBAL?

morozov:
```php
$conn = new PDO('sqlite::memory:');
$stmt = $conn->query('SELECT 1');

$stmt->fetchAll(PDO::FETCH_KEY_PAIR);

// PDOStatement::fetchAll(): SQLSTATE[HY000]: General error: PDO::FETCH_KEY_PAIR fetch mode requires the result set to contain extactly 2 columns
```
Looks like PDO is strict enough about the number of columns. I think we can implement this in DBAL for all drivers as well.

tswcode:
@morozov Indeed it could be implemented in userland, yet I always try to avoid boilerplate code, when using this fetch mode. That would be great, if the DBAL would provide a method, so that i can replace the deprecated one.

morozov:
@tswcode could you file a feature request please?

tmotyl:
Btw, do you consider providing rector rule to help project migrate to doctrine v3 where the deprecated methods will be dropped?

morozov:
No, but it would be nice if somebody kicked that off. As far as I understand, Rector has some rules for other Doctrine projects, so the development would have to be done in its repo.

jokaorgua:
Could anyone please advise how to change current deprecated code without losing functionality?

```
$query = $this->connection->createQueryBuilder()
    ->select()....

$result = $query->execute()->fetchAll(FetchMode::STANDARD_OBJECT);
```

how to fetch an array of objects after this merge?

derrabus:
```php
$result = array_map(
    static fn (array $row) => (object) $row,
    $query->execute()->fetchAllAssociative()
)
```

jokaorgua:
> ```
> $result = array_map(
>     static fn (array $row) => (object) $row,
>     $query->execute()->fetchAllAssociative()
> );
> ```

thanks. but this seems like a hack. are there any docs where it can be found why developers decided to remove the ability to fetch objects?

morozov:
@jokaorgua please see https://github.com/doctrine/dbal/pull/3070.

derrabus:
> thanks. but this seems like a hack.

It is a hack to unblock codebases that rely on object fetching. In the long run, that code should be migrated to use associative arrays instead.

PR #4020, opened 2020-05-23
Deprecated Connection::project(), Statement::errorCode() and errorInfo()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no

See https://github.com/doctrine/dbal/pull/3507, https://github.com/doctrine/dbal/pull/3823.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,8 +1,16 @@
 # Upgrade to 2.11
 
+## Deprecated `Connection::project()`
+
+The `Connection::project()` method is deprecated. Implement data transformation outside of DBAL.
+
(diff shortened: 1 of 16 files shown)


greg0ire (approved):
I guess `@deprecated` cannot be inherited?

morozov:
I believe [it can't](https://docs.phpdoc.org/latest/guides/inheritance.html). From the common sense PoV, if you're looking for some bit of information (e.g. description) on some symbol and cannot find it there, you may refer to a parent class or interface, you'll try to look it up. But if you're not looking for it (like in the case of `@deprecated`), you will not refer to a parent.

PR #4027, opened 2020-05-25
@link in convertException gives 404

tomasnorre:
The links in convertException give 404, I'm not 100% if I updated them correct, or if the e.g. should be updated to 8.0 already. 

https://dev.mysql.com/doc/refman/.0/en/server-error-reference.html
https://dev.mysql.com/doc/refman/8.0/en/client-error-reference.html

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues |


SenseException (changes requested):
Since there is no alias for the latest version in the URL I suggest to use 8.0.

SenseException (approved):
Thank you @tomasnorre

tomasnorre:
You're welcome.

PR #4028, opened 2020-05-25
Allow php 8

greg0ire:
Targeting 3.0.x on that particular one, because there will never be support for php 8 on 2.x (right?)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         {"name": "Jonathan Wage", "email": "jonwage@gmail.com"}
     ],
     "require": {
-        "php": "^7.3",
+        "php": "^7.3 || ^8.0",
         "doctrine/cache": "^1.0",
         "doctrine/event-manager": "^1.0",
         "ocramius/package-versions": "^1.4"
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "893d32dea2abc8c66017ee58e435fac8",
+    "content-hash": "375d7b1967e204d71959234a23055326",
     "packages": [
         {
             "name": "doctrine/cache",
@@ -1816,6 +1816,16 @@
                 "testing",
                 "xunit"
             ],
+            "funding": [
+                {
+                    "url": "https://phpunit.de/donate.html",
+                    "type": "custom"
+                },
+                {
+                    "url": "https://github.com/sebastianbergmann",
+                    "type": "github"
+                }
+            ],
             "time": "2020-04-03T14:40:04+00:00"
         },
         {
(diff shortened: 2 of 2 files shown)


composer.json.orig
```diff
@@ -0,0 +1,77 @@
...
+    "homepage": "https://www.doctrine-project.org/projects/dbal.html",
+    "license": "MIT",
+    "authors": [
+        {"name": "Guilherme Blanco", "email": "guilhermeblanco@gmail.com"},
+        {"name": "Roman Borschel", "email": "roman@code-factory.org"},
+        {"name": "Benjamin Eberlei", "email": "kontakt@beberlei.de"},
+        {"name": "Jonathan Wage", "email": "jonwage@gmail.com"}
+    ],
+    "require": {
+<<<<<<< HEAD
```
SenseException, line 35: Whoops.

SenseException (changes requested):

composer.json.orig
greg0ire: whoops indeed :sweat_smile:

morozov:
> Targeting 3.0.x on that particular one, because there will never be support for php 8 on 2.x (right?)

Right. Support for PHP 8 requires a breaking change.

.travis.yml
```diff
@@ -44,7 +44,17 @@ after_script:
     fi
 
 jobs:
+  allow_failures:
+    - php: nightly
+
   include:
+    -   stage: Test
+        php: nightly
```
morozov, line 52: Given that we already build against `nightly`, why is this specific job needed?

.travis.yml
```diff
@@ -44,7 +44,17 @@ after_script:
     fi
 
 jobs:
+  allow_failures:
+    - php: nightly
```
morozov, line 48: This is already configured:

https://github.com/doctrine/dbal/blob/54a6f7b3585e0cda5d5613200ea6bc6fadf04e95/.travis.yml#L230-L232

greg0ire:
Sorry, I didn't notice that `nightly` was already handled here… nothing to change besides the version constraint.

SenseException (approved):
Some green nightly builds :man_dancing:

morozov (approved):

morozov:
Should we then remove `--ignore-platform-reqs`?

greg0ire:
> Should we then remove --ignore-platform-reqs?

We can try replacing that with `composer config platform.php 7.4.99`, because we still need to fool our own dependencies into thinking this is 8.

morozov:
> because we still need to fool our own dependencies into thinking this is 8

I don't get this. Isn't PHP nightly the PHP 8?

greg0ire:
Sorry, I meant: into thinking this is 7. Some might not be 8-compatible yet.

morozov:
Okay, not it makes sense.

greg0ire:
So do we stay with `--ignore-platform-reqs`, or do we use the above trick?

morozov:
`--ignore-platform-reqs` seems to be the proper way.

PR #4034, opened 2020-05-27
Additional changes based on the discussion in #4007

### LINKED ISSUE #4033: ResultCacheStatement::fetchAll() doesn't respect fetch mode on subsequent calls
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.2

When a cached result set is used with `fetchAll()` and the fetch mode is different from the one used to populate the cache, the original fetch mode is used instead.

```php
    'driver' => 'pdo_mysql',
    'user'   => 'root',
    'dbname' => 'mysql',
]);

$conn->getConfiguration()
    ->setResultCacheImpl(new ArrayCache());

$qcp = new QueryCacheProfile(0, 'test');

$stmt = $conn->executeQuery("SELECT 'foo' bar", [], [], $qcp);
$stmt->fetchAll(FetchMode::NUMERIC);
$stmt->closeCursor();

$stmt = $conn->executeQuery("SELECT 'foo' bar", [], [], $qcp);
var_dump($stmt->fetchAll(FetchMode::ASSOCIATIVE));
```

#### Current behaviour

```
array(1) {
  [0] =>
  array(1) {
    [0] =>
    string(3) "foo"
  }
}
```

#### Expected behaviour

```
array(1) {
  [0] =>
  array(1) {
    'bar' =>
    string(3) "foo"
  }
}
```

morozov:
|      Q       |   A
|------------- | -----------
| BC Break     | yes/no
| Fixed issues | #4033

1. Backport the documentation changes as suggested in https://github.com/doctrine/dbal/pull/4007#discussion_r431331657.
2. Backport the fix for the issue discovered in https://github.com/doctrine/dbal/pull/4007#discussion_r431285890 and the corresponding changes in the tests that cover the issue.
3. Fix #4033 since it prevents the refactored tests from passing on `2.x`.

### DIFF
--- a/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
+++ b/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
@@ -13,6 +13,7 @@ use Doctrine\DBAL\ForwardCompatibility\Driver\ResultStatement as ForwardCompatib
 use InvalidArgumentException;
 use IteratorAggregate;
 use PDO;
+use function array_map;
 use function array_merge;
 use function array_values;
 use function assert;
@@ -55,7 +56,7 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement, Forwar
      */
     private $emptied = false;
 
-    /** @var mixed[] */
+    /** @var array<int,array<string,mixed>> */
     private $data;
 
     /** @var int */
@@ -247,7 +248,9 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement, Forwar
             $data = $this->statement->fetchAll(FetchMode::ASSOCIATIVE);
         }
 
-        return $this->store($data);
+        $this->store($data);
+
+        return array_map('array_values', $this->data);
     }
 
     /**
@@ -261,7 +264,9 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement, Forwar
             $data = $this->statement->fetchAll(FetchMode::ASSOCIATIVE);
         }
 
-        return $this->store($data);
+        $this->store($data);
+
+        return $this->data;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
```diff
@@ -179,18 +180,26 @@ public function fetch($fetchMode = null, $cursorOrientation = PDO::FETCH_ORI_NEX
      */
     public function fetchAll($fetchMode = null, $fetchArgument = null, $ctorArgs = null)
     {
-        $data = $this->statement->fetchAll($fetchMode, $fetchArgument, $ctorArgs);
```
morozov, line 182: This is the root cause of https://github.com/doctrine/dbal/issues/4033.

tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
```diff
@@ -243,7 +244,8 @@ public function testChangeCacheImpl() : void
     private function hydrateStmt(ResultStatement $stmt, int $fetchMode = FetchMode::ASSOCIATIVE) : array
     {
         $data = [];
-        while ($row = $stmt->fetch($fetchMode)) {
+
+        foreach ($stmt->fetchAll($fetchMode) as $row) {
```
morozov, line 248: This is how https://github.com/doctrine/dbal/issues/4033 was discovered.

lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
```diff
@@ -179,18 +180,26 @@ public function fetch($fetchMode = null, $cursorOrientation = PDO::FETCH_ORI_NEX
...
+            foreach ($data as $i => $row) {
+                $data[$i] = array_merge($row, array_values($row));
+            }
+        } elseif ($fetchMode === FetchMode::COLUMN) {
+            foreach ($data as $i => $row) {
+                $data[$i] = reset($row);
+            }
+        }
+
+        return $data;
```
greg0ire, line 202: How about this:

```suggestion
        return array_map(function(array $row) use ($fetchMode)  {
            switch ($fetchMode)
            {
                case FetchMode::NUMERIC:
                    return array_values($row);

                case FetchMode::MIXED:
                    return array_merge($row, array_values($row));
                
                case FetchMode::COLUMN:
                    return reset($row);
                
                default:   
                    return $row;
            }
        }, $data);
```
?

greg0ire (approved):
Feel free to ignore my suggestion and merge, it's nothing important.

lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
morozov: It will require a switch on each row instead of an if/else once.

PR #4036, opened 2020-05-28
Added specific handling of string type query parameters

dvaeversted:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

This change introduces specific handling of strings, which will assist the SQL server query optimizer in choosing the best possible execution plan.

The current behaviour causes all string parameters to be sent to the SQL server as NVARCHAR(4000).
This causes an unfortunate situation for the query optimizer, that will create a sub-optimal execution plan from the assumption that the parameters are big data strings.

In production we saw multiple instances of a factor 10 or more changes in query time, with the proposed change.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
+++ b/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
@@ -28,6 +28,7 @@ use function sqlsrv_prepare;
 use function sqlsrv_rows_affected;
 use function SQLSRV_SQLTYPE_VARBINARY;
 use function stripos;
+use function strlen;
 use const SQLSRV_ENC_BINARY;
 use const SQLSRV_ERR_ERRORS;
 use const SQLSRV_FETCH_ASSOC;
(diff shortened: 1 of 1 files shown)


SenseException (changes requested):
Please create a test that is covering your code and use case. This also breaks existing tests which might be because of BC breaks.

dvaeversted:
I have corrected the patch, to fix the failure seen in the unit tests.
Apparently we never have a query where we send in a NULL as a string parameter in our environment, hence i missed that case.

Regarding a specific test for the introduced change, i am unsure how to actually write a specific test for it since the previous way it was done will yield the same result seen from the client.
It is affecting the query optimizer in the SQL server to make a better execution plan, i also believe that the function itself is sufficiently covered by the existing unit tests as illustrated by the previous failing unit tests.

SenseException:
I want to achieve that removing your new introduced code will also have tests failing. The other cases, like e.g. large objects, were covered in functional tests. I wonder if this one can be handled with an unit test since this change doesn't influence the result. I'll have to look deeper into this for an idea how we can test this.

greg0ire:
Don't mind me, just testing CodeCov's reports.

PR #4039, opened 2020-05-30
Refactor portability statement into a functional composition

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

The current implementation of the portability statement contains quite a lot of code repetition, especially introduced at #4019. The reason is that there's no clear separation of concerns of data fetching and data transformation which is aggravated by the number of possible configurations.

In order to make the code more extensible, the statement is split into a wrapper that fetches the data from the underlying statement and a converter that converts the data. Unlike the current design, the converter is created once per connection and is then reused for all fetches.

The other benefit that the separation of concerns gives is that the converter can be fully unit-tested w/o the need to mock the underlying statement or set up data fixtures.

### DIFF
--- a/src/Portability/Converter.php
+++ b/src/Portability/Converter.php
@@ -1,0 +1,247 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Portability;
+
+use function array_change_key_case;
+use function array_map;
+use function array_reduce;
+use function is_string;
+use function rtrim;
+
+final class Converter
+{
+    /** @var callable */
+    private $convertNumeric;
+
+    /** @var callable */
+    private $convertAssociative;
+
+    /** @var callable */
+    private $convertOne;
+
+    /** @var callable */
+    private $convertAllNumeric;
+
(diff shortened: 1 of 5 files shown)


src/Portability/Converter.php
```diff
@@ -0,0 +1,246 @@
...
+    {
+        return static function (array $array) use ($function) : array {
+            return array_map($function, $array);
+        };
+    }
+
+    /**
+     * Creates a composition of the given set of functions
+     *
+     * @param callable ...$functions The functions to compose
```
greg0ire, line 232: I think you are going to have to use https://psalm.dev/docs/annotating_code/type_syntax/callable_types/ to fix the Psalm issue below

morozov: I ended up changing the signature of the `array_reduce()`'s `$callback` so that it can handle `$carry = null`, and now Psalm seems happy.

greg0ire (approved):
"convert" is better naming than "fix" :ok_hand:

Issue #3963, opened 2020-04-18
Additions to Platform


aboks:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary
Hi DBAL team, we're currently in the process of migrating our own database abstraction (MySQL/MariaDB and SQLite only) to Doctrine DBAL. While doing so, we noticed that we have some abstraction for functionality that differs between MySQL and SQLite, for which there is no method yet on the DBAL platform classes. We're wondering if you would accept PR's for these cases. They would be things like:

* Syntax to use for `INSERT IGNORE INTO` (SQLite requires `INSERT OR IGNORE INTO`)
* Syntax to use for `GROUP_CONCAT` (MySQL uses `GROUP_CONCAT(%s SEPARATOR %s)`, SQLite uses `GROUP_CONCAT(%s, %s)`)
* Syntax to use for `ANALYZE TABLE` (SQLite uses just `ANALYZE`)
* Whether the database supports `SELECT ... FOR UPDATE`
* Whether the database supports `SELECT ... LOCK IN SHARE MODE`
* Whether the database supports `UPDATE ... ORDER BY`
* Whether the database supports multi-row inserts
* Syntax to use for `LEAST`/`GREATEST` (SQLite uses `MIN`/`MAX`)
* Whether the database supports `CONCAT_WS`
* Whether the database supports `CREATE TABLE LIKE`
* Whether the database supports `INSERT ... ON DUPLICATE KEY UPDATE`
* Whether the database supports `UPDATE ... JOIN ...`

If you'd welcome those, I will create separate issues/PR's for each of them. In general, against which branch would you like such PR's? Any guidelines or specific things to keep in mind when adding functionality to the platform layer?

morozov:
Closing as non-actionable.

PR #4041, opened 2020-06-01
Add getAggregateConcatExpression to platform

aboks:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #3963 (only partially)

#### Summary

This new method returns the expression for aggregating values by concatenating them with a separator between the individual values. This corresponds to something like GROUP_CONCAT/LIST/LISTAGG/STRING_AGG. Since there does not really seem to be a standard, the function is generically named and does not have a default implementation in AbstractPlatform.

This is one of the features mentioned in #3963.

### DIFF
--- a/src/Platforms/AbstractPlatform.php
+++ b/src/Platforms/AbstractPlatform.php
@@ -561,7 +561,10 @@ abstract class AbstractPlatform
      * @param string      $separator SQL expression producing the separator.
      * @param string|null $orderBy   Optional SQL expression (and direction) to order the grouped rows by.
      */
-    abstract public function getAggregateConcatExpression(string $value, string $separator, ?string $orderBy) : string;
+    public function getAggregateConcatExpression(string $value, string $separator, ?string $orderBy) : string
+    {
+        throw NotSupported::new(__METHOD__);
+    }
 
     // scalar functions
 
--- a/src/Platforms/AbstractPlatform.php
+++ b/src/Platforms/AbstractPlatform.php
@@ -554,6 +554,14 @@ abstract class AbstractPlatform
         return 'SUM(' . $value . ')';
     }
 
+	/**
+	 * Returns the SQL snippet to get the concatenation of the values in a set with a separator.
+	 *
+	 * @param string $value SQL expression producing the value.
+	 * @param string $separator SQL expression producing the separator.
+	 */
+	abstract public function getAggregateConcatExpression(string $value, string $separator) : string;
+
     // scalar functions
 
     /**
--- a/src/Platforms/AbstractPlatform.php
+++ b/src/Platforms/AbstractPlatform.php
@@ -561,7 +561,7 @@ abstract class AbstractPlatform
      * @param string      $separator SQL expression producing the separator.
      * @param string|null $orderBy   Optional SQL expression (and direction) to order the grouped rows by.
      */
-    public function getAggregateConcatExpression(string $value, string $separator, ?string $orderBy = null) : string
+    public function getAggregateConcatExpression(string $value, string $separator, ?string $orderBy = null): string
     {
         throw NotSupported::new(__METHOD__);
     }
--- a/src/Platforms/AbstractPlatform.php
+++ b/src/Platforms/AbstractPlatform.php
@@ -554,13 +554,13 @@ abstract class AbstractPlatform
         return 'SUM(' . $value . ')';
     }
 
-	/**
-	 * Returns the SQL snippet to get the concatenation of the values in a set with a separator.
-	 *
-	 * @param string $value SQL expression producing the value.
-	 * @param string $separator SQL expression producing the separator.
-	 */
-	abstract public function getAggregateConcatExpression(string $value, string $separator) : string;
+    /**
+     * Returns the SQL snippet to get the concatenation of the values in a set with a separator.
+     *
+     * @param string $value     SQL expression producing the value.
+     * @param string $separator SQL expression producing the separator.
+     */
+    abstract public function getAggregateConcatExpression(string $value, string $separator) : string;
 
     // scalar functions
 
--- a/src/Platforms/AbstractPlatform.php
+++ b/src/Platforms/AbstractPlatform.php
@@ -561,7 +561,7 @@ abstract class AbstractPlatform
      * @param string      $separator SQL expression producing the separator.
      * @param string|null $orderBy   Optional SQL expression (and direction) to order the grouped rows by.
      */
-    public function getAggregateConcatExpression(string $value, string $separator, ?string $orderBy) : string
+    public function getAggregateConcatExpression(string $value, string $separator, ?string $orderBy = null) : string
     {
         throw NotSupported::new(__METHOD__);
     }
--- a/src/Platforms/AbstractPlatform.php
+++ b/src/Platforms/AbstractPlatform.php
@@ -557,10 +557,11 @@ abstract class AbstractPlatform
     /**
      * Returns the SQL snippet to get the concatenation of the values in a set with a separator.
      *
-     * @param string $value     SQL expression producing the value.
-     * @param string $separator SQL expression producing the separator.
+     * @param string      $value     SQL expression producing the value.
+     * @param string      $separator SQL expression producing the separator.
+     * @param string|null $orderBy   Optional SQL expression (and direction) to order the grouped rows by.
      */
-    abstract public function getAggregateConcatExpression(string $value, string $separator) : string;
+    abstract public function getAggregateConcatExpression(string $value, string $separator, ?string $orderBy) : string;
 
     // scalar functions
 
--- a/tests/Functional/Platform/AggregateConcatExpressionTest.php
+++ b/tests/Functional/Platform/AggregateConcatExpressionTest.php
@@ -25,7 +25,7 @@ class AggregateConcatExpressionTest extends FunctionalTestCase
         $concatenated = $this->connection->query($sql)->fetchOne();
         $parts        = explode(',', $concatenated);
 
-        self::assertEquals(['foo', 'bar', 'baz'], $parts);
+        self::assertEqualsCanonicalizing(['foo', 'bar', 'baz'], $parts);
     }
 
     public function testAggregateConcatWithOrder() : void
--- a/tests/Functional/Platform/AggregateConcatExpressionTest.php
+++ b/tests/Functional/Platform/AggregateConcatExpressionTest.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\DBAL\Tests\Functional\Platform;
 
+use Doctrine\DBAL\ParameterType;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Tests\FunctionalTestCase;
 use function explode;
@@ -16,9 +17,9 @@ class AggregateConcatExpressionTest extends FunctionalTestCase
         $table = new Table('aggregate_concat_test');
         $table->addColumn('value', 'string', ['length' => 10]);
         $this->connection->getSchemaManager()->dropAndCreateTable($table);
-        $this->connection->insert('aggregate_concat_test', ['value' => 'foo']);
-        $this->connection->insert('aggregate_concat_test', ['value' => 'bar']);
-        $this->connection->insert('aggregate_concat_test', ['value' => 'baz']);
+        $this->connection->insert('aggregate_concat_test', ['value' => 'foo'], ['value' => ParameterType::STRING]);
+        $this->connection->insert('aggregate_concat_test', ['value' => 'bar'], ['value' => ParameterType::STRING]);
+        $this->connection->insert('aggregate_concat_test', ['value' => 'baz'], ['value' => ParameterType::STRING]);
 
         $platform = $this->connection->getDatabasePlatform();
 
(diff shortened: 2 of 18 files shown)


morozov:
Thank you for your contribution, @aboks. This new API will need to be covered by a functional test to make sure the new behavior is portable. See `DateExpressionTest` for example.

aboks:
Thanks for the hint, I'll look into it.

aboks:
I added a functional test. @morozov do you have any idea why the test seems to fail for DB2 on Travis? Other functional tests have similar `insert()`-calls, and they seem to work fine.

morozov:
> Other functional tests have similar `insert()`-calls, and they seem to work fine.

I can tell that an SQL Server and Oracle builds fail too due to invalid SQL. It is possible that the DB2 one fails due to invalid syntax as well. I'll double-check and see if we can report that better than just "Describe Param Failed" which comes directly from the `ibm_db2` extension.

tests/Functional/Platform/AggregateConcatExpressionTest.php
```diff
@@ -0,0 +1,34 @@
...
+        $this->connection->getSchemaManager()->dropAndCreateTable($table);
+        $this->connection->insert('aggregate_concat_test', ['value' => 'foo'], ['value' => ParameterType::STRING]);
+        $this->connection->insert('aggregate_concat_test', ['value' => 'bar'], ['value' => ParameterType::STRING]);
+        $this->connection->insert('aggregate_concat_test', ['value' => 'baz'], ['value' => ParameterType::STRING]);
+
+        $platform = $this->connection->getDatabasePlatform();
+
+        $sql = sprintf('SELECT %s FROM aggregate_concat_test', $platform->getAggregateConcatExpression('value', $this->connection->quote(',')));
+
+        // The order is not guaranteed, so we can only check if the parts are the same
```
morozov, line 28: Is it not guaranteed on a specific platform or it's by design? This way the abstracted out function is barely usable, and when the order is important (which is often the case), the consumer will have to resort to using raw SQL.

aboks: Unfortunately not all platforms support an order (SQLite being at least one of them). I figured that passing an order as an extra optional parameter could be a possible future extension of this method. That would mean that a platform like SQLite would have to throw `NotSupported` if that parameter is set, though.

While some use cases definitely require ordering, there are also cases where having no ordering is fine, or sorting can happen afterwards. E.g. when you have tables `posts` and `tags` with a many-to-many relationship between them, concat aggregation could give you all posts with all their tags in one single database query. Otherwise such an operation would require multiple queries, or doing something fetch-join-like and manual deduplication of rows.

morozov: In your example with posts and tags, using the string aggregation function will only work if only the tag names are needed. If anything else than a name is needed, this query will have to be reworked to using two queries: one retrieves all posts and their identifiers, the other retrieves all tag information by the identifiers collected by the first query.

> Otherwise such an operation would require multiple queries, or doing something fetch-join-like and manual deduplication of rows.

Performance-wise, those two simple queries will be most likely faster than the one with the join and group by.

While I don't have a strong objection, it looks like the API proposed helps solve a problem in a way that is not extensible and will require a rework w/o using this API if needed.

morozov: Another example is, let's say each post has more than one author and more than one tag. As long as you're only showing tags in the list, you can string-aggregate them but as soon as you also need to show the authors, you'll have to add a secondary query.

aboks: I could try to add an optional `$orderBy` parameter to this method if you think it would make this feature much more useful. That would mean it must throw `NotSupported` if `$orderBy` is set and the platform does not support an order, though.

> Another example is, let's say each post has more than one author and more than one tag. As long as you're only showing tags in the list, you can string-aggregate them but as soon as you also need to show the authors, you'll have to add a secondary query.

I agree that concat-aggregation surely has its limitations. I just checked our codebase, and we only use it in a few select places. Still, given that all major DB platforms support it, it seems useful to have an abstraction for in the DBAL platform.

morozov:
@doctrine/doctrinecore, any opinions on whether the DBAL should have this functionality? See my concerns in the discussion above.

ostrolucky:
I like it. Order is really not important for most use cases (but support can be added in future).

src/Platforms/AbstractPlatform.php
```diff
@@ -553,6 +553,14 @@ public function getSumExpression(string $value) : string
         return 'SUM(' . $value . ')';
     }
 
+    /**
+     * Returns the SQL snippet to get the concatenation of the values in a set with a separator.
+     *
+     * @param string $value     SQL expression producing the value.
+     * @param string $separator SQL expression producing the separator.
+     */
+    abstract public function getAggregateConcatExpression(string $value, string $separator) : string;
```
stof, line 562: adding new abstract methods is a BC break. You should make it a concrete method throwing `NotSupported`.

greg0ire:
I've had a look at the usage of `GROUP_CONCAT` in my projects and indeed, the order didn't matter. If think we might want to add support for it in the future, we should think about it now, because adding a new argument like OP proposed would be a BC break for extending classes.

No super strong opinion on whether we should have the feature or not, but I've used both `GROUP_CONCAT` and `STRING_AGG` in the past, so I'd say it could be useful.

src/Platforms/AbstractPlatform.php
greg0ire: Once this is fixed, the PR should be retargeting to 2.11.x IMO, so that ppl can benefit from it earlier.

src/Platforms/AbstractPlatform.php
```diff
@@ -556,10 +556,14 @@ public function getSumExpression(string $value) : string
...
      * Returns the SQL snippet to get the concatenation of the values in a set with a separator.
      *
-     * @param string $value     SQL expression producing the value.
-     * @param string $separator SQL expression producing the separator.
+     * @param string      $value     SQL expression producing the value.
+     * @param string      $separator SQL expression producing the separator.
+     * @param string|null $orderBy   Optional SQL expression (and direction) to order the grouped rows by.
      */
-    abstract public function getAggregateConcatExpression(string $value, string $separator) : string;
+    public function getAggregateConcatExpression(string $value, string $separator, ?string $orderBy) : string
```
greg0ire:
```suggestion
    public function getAggregateConcatExpression(string $value, string $separator, ?string $orderBy = null) : string
```

aboks: Wow, you're almost as fast as PhpStan and Psalm! :)

tests/Functional/Platform/AggregateConcatExpressionTest.php
aboks: Some research showed that all platforms except SQLite support ordering, so I just added an optional `$orderBy` parameter.

src/Platforms/AbstractPlatform.php
aboks: Would the retarget to 2.11.x be something to be done when merging, or should I rebase my branch onto 2.11.x already?

aboks:
It seems like the syntax for SQL Server was only added in version 2017. Would it be preferred to create a new `SQLServer2017Platform` subclass, or would it be better to somehow inject the version number into the existing platform class?

morozov:
The reasons why I wouldn't like to accept this new API in the library:

1. It implements a small subset of what most of the platforms are capable of (i.e. no ordering) and is not extensible by design.
2. The API itself is applicable in a very narrow set of cases (see the comments above) which as a developer I'd rather be forced to solve in an extensible way than being tricked into using a shortcut that will then make me rewrite this code.
3. It's dangerous due to its trimming nature on MySQL.

ostrolucky:
I think you are being unfair and I feel I need to support OP a bit:

> It implements a small subset of what most of the platforms are capable of (i.e. no ordering) and is not extensible by design.

It implements ordering, as you wished, even if not every DBMS supports that, but majority do. So you can't say it implements small subset. And I didn't see proposals how to make this feature more extensible? What do you mean by that and why does this function need extensibility in different way than other existing driver functions?

> The API itself is applicable in a very narrow set of cases (see the comments above)

You again say something that I see is opposite. Perhaps you didn't, but me and lot of other people needed group_concat lot of times and were limited by DBAL (and hence ORM) not having it. Limitation of displaying concatenated single field only in this column is well understood here, but it's better to execute 1 query with 10 joins and 5 group_concats than 5 slightly different queries each with same 10 joins and having to heavily post process the result for each query in php. And frankly it's not a problem to slightly change the query when you find out you need 2 field there. What problem is though, is when you don't even have an option to decide which approach you can use. It is indeed quite limiting feeling when you find out how many features DBAL does not support. But we can improve this.

> It's dangerous due to its trimming nature on MySQL.

We would have no DBAL if we started to throw out every feature that works slightly different across RDBMs, that's just not avoidable. This is not something that DBAL should be concerned about, similarly as DBAL is not concerned by different default order of results, or mysql with different `sql_mode`. This project will always be leaky abstraction.

morozov:
@ostrolucky isn't it wonderful that all of us can have and express their opinion?

greg0ire:
Oof… let's keep the tone friendly please :pray: 

Let's also take this as an occasion to clarify what makes a good candidate for abstraction, and what doesn't?

1. It's indeed not extensible… in the code of my application, we use `DISTINCT` in some occurrences, which means this PR wouldn't help as much as it could.
2. "I'd rather be forced to solve in an extensible way" => what would that look like? Should we pass a final value object that can have additional properties in the future rather than scalars to `getAggregateConcatExpression`? Did you have a specific solution in mind? Something that would avoid the API end up being the lowest common denominator?
3. Sorry, but what is this about? Does mysql trim aggregated strings ?

ostrolucky:
Users need to process the result of such query with `explode()` anyway, it's not hard to wrap that in `array_unique`, so I wouldn't say missing `DISTINCT` is limiting. But if I see it right, users willing to sacrify cross-db compatibility can do just `$this->platform->getAggregateConcatExpression('DISTINCT column1', ";")`, right?

greg0ire:
I suppose that would work indeed 👍

morozov:
> 2\. "I'd rather be forced to solve in an extensible way" => what would that look like?

What I mean is instead of using string aggregation in the query, I'd prefer using a secondary query. Using the blog analogy, I'd retrieve the list of posts in the first query, retrieve the list of tags using the blog IDs in the second query, retrieved the list of author names in the third query and so on. This approach is pretty much composable, portable and optimal from the performance perspective.

> Should we pass a final value object that can have additional properties in the future rather than scalars to `getAggregateConcatExpression`?

I didn't mean the extensibility of the `getAggregateConcatExpression()` method itself. I meant the extensibility of the string aggregation approach. As soon as your logic becomes a bit more complex than "concatenate strings", you'll not be able to use this API. Its usage is an immediate technical debt.

> Did you have a specific solution in mind? Something that would avoid the API end up being the lowest common denominator?

I believe it was in the context of the API extensibility.

> 1. in the code of my application, we use `DISTINCT` in some occurrences

Using the blog analogy again, if two different authors have the same name (less possible in the blog case but is possible in other cases, e.g. large company employees), would you just squash them together because they happen to have the same name? The approach above is safer in this case.

> But if I see it right, users willing to sacrify cross-db compatibility can do just $this->platform->getAggregateConcatExpression('DISTINCT column1', ";"), right?

So why implement a portable implementation if the users will have to sacrifice portability?

ostrolucky:
> So why implement a portable implementation if the users will have to sacrifice portability?

They don't have to sacrify portability. All they need is to not use distinct. It's their choice.

You know your standpoint would make sense if we were maintaining database. There you could argue that we don't want to implement group_concat because it's too limited feature. But not here in DBAL, group_concat likes were implemented in all databases so apparently database maintainers themselves see value in it. DBAL's job is to just expose it.

aboks:
Note that it would also be possible to allow adding a `DISTINCT` in a portable way by adding an extra optional parameter to the method signature. Of course that might mean risking a `NotSupported` if the underlying platform does not support any form of `DISTINCT`.

src/Platforms/AbstractPlatform.php
greg0ire: You should rebase

morozov:
> Note that it would also be possible to allow adding a `DISTINCT` in a portable way by adding an extra optional parameter to the method signature. Of course that might mean risking a `NotSupported` if the underlying platform does not support any form of `DISTINCT`.

Distinct is a workaround of the problem that the query fetches more data than it needs and will cause performance issues once the database grows. This by no means should be added to the API.

aboks:
I rebased onto the latest master, fixed the remaining CS issues, and created a separate platform version for SQL Server 2017 (as earlier versions do not support an aggregate concat expression). Everything seems to work fine now, except the test failure with DB2 on Travis, which I still don't understand (`DateExpressionTest` and `ColumnTest` do about the same and work fine). If anyone knows what's going on there, please let me know.

aboks:
@morozov Can you explain why this was closed?

morozov:
@aboks please see https://github.com/doctrine/dbal/pull/4041#issuecomment-641580368.

PR #4047, opened 2020-06-02
Integrate with Codecov

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Send coverage to Codecov so that we can compare with Coveralls.

Todo:
- [x] travis
- [x] appveyor
- [x] update badge in the README ?

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -20,22 +20,18 @@ environment:
   - db: mssql
     driver: sqlsrv
     db_version: sql2008r2sp2
-    coverage: yes
     php: 7.3
   - db: mssql
     driver: sqlsrv
     db_version: sql2012sp1
-    coverage: yes
     php: 7.3
   - db: mssql
     driver: sqlsrv
     db_version: sql2017
-    coverage: no
     php: 7.3
   - db: mssql
     driver: pdo_sqlsrv
     db_version: sql2017
-    coverage: yes
     php: 7.3
 
 init:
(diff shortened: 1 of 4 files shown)


greg0ire:
I reported the AppVeyor issue at https://community.codecov.io/t/appveyor-find-parameter-format-not-correct/1506

.appveyor.yml
```diff
@@ -123,8 +127,19 @@ test_script:
...
       if ($LastExitCode -ne 0) {
         $host.SetShouldExit($LastExitCode)
       }
+
+after_test:
+  - ps: |
+      Invoke-WebRequest -Uri 'https://codecov.io/bash' -OutFile codecov.sh
+      $env:APPVEYOR="true"
+      $env:CI="true"
+      bash codecov.sh -f "clover.xml"
```
thomasrockhu, line 145: Hi, Tom from Codecov here. I'm not sure this will fix the issue, but I have a sneak suspicion it might.
```suggestion
      bash codecov.sh -f clover.xml
```

greg0ire: Thanks, let's try!

greg0ire: Looks like it doesn't help…

morozov: FWIW, I successfully switched to Codecov and PCOV on Appveyor in https://github.com/diff-sniffer/diff-sniffer/pull/34 and https://github.com/diff-sniffer/diff-sniffer/pull/35.

greg0ire: Thanks it was helpful!
1. I remove the env variables, didn't help
2. I switched to multiline syntax, and got an error message I did not see before: `Invoke-WebRequest -Uri 'https://codecov.io/bash' -OutFile codecov.sh
'Invoke-WebRequest' is not recognized as an internal or external command,
operable program or batch file`.
3. I used the same download command as you did, and now it works!

greg0ire:
Note that the numbers are not the same than with Coveralls: it says 71% vs 73% with CodeCov.

.appveyor.yml
```diff
@@ -20,18 +20,22 @@ environment:
...
     php: 7.3
   - db: mssql
     driver: sqlsrv
     db_version: sql2012sp1
+    coverage: yes
     php: 7.3
   - db: mssql
     driver: sqlsrv
     db_version: sql2017
+    coverage: no
```
morozov, line 33: This is only needed if Xdebug is used. Since we're switching to PCOV, would it make sense to collect code coverage just always on AppVeyor?

The configuration of Travis could be changed in a subsequent PR.

morozov (approved):
Thanks, @greg0ire. It's a very desired change.

greg0ire:
Looks like we have flaky tests:

```
1) Doctrine\Tests\DBAL\Functional\Platform\DefaultExpressionTest::testCurrentTime
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'16:08:42'
+'16:08:41'
C:\projects\dbal\tests\Doctrine\Tests\DBAL\Functional\Platform\DefaultExpressionTest.php:76
C:\projects\dbal\tests\Doctrine\Tests\DBAL\Functional\Platform\DefaultExpressionTest.php:45
FAILURES!
```

morozov:
> Looks like we have flaky tests:

The test performs an atomic insertion query, so the flakiness happens in the DB itself and I have no idea how to avoid that. The DB interprets the `CURRENT_TIMESTAMP` expression from the SQL and the `CURRENT_TIMESTAMP` from the DDL at different points in time.

greg0ire:
I have no idea either… doesn't seem to happen that much though.

PR #4048, opened 2020-06-06
Make caching layer not rely on closeCursor()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

The current behavior, although it's documented in the form of a test, is counterintuitive:
https://github.com/doctrine/dbal/blob/c08d3bc8df1f2d048e2912c06110f6dc6faf688e/tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php#L136

This test and the fact that all other tests call `$stmt->closeCursor()` imply that the statement result will be only cached if the statement cursor is closed.

The meaning of `closeCursor()` has nothing to deal with caching:
https://github.com/doctrine/dbal/blob/c08d3bc8df1f2d048e2912c06110f6dc6faf688e/lib/Doctrine/DBAL/Driver/ResultStatement.php#L13-L18

It should be used in certain environments when not the entire result set is fetched in order to prevent errors like this:

> SQLSTATE[HY000]: General error: 2014 Cannot execute queries while other unbuffered queries are active

In the case when the full result set is fetched, nothing should prevent the caching layer from caching the result. It happens at the end of `fetchAll()` or iterative `fetch()` operations.

### DIFF
--- a/docs/en/reference/caching.rst
+++ b/docs/en/reference/caching.rst
@@ -35,15 +35,13 @@ the default cache instance:
     new QueryCacheProfile(0, "some key", $cache);
 
 In order for the data to actually be cached its necessary to ensure that the entire
-result set is read (the easiest way to ensure this is to use ``fetchAll``) and the statement
-object is closed:
+result set is read (the easiest way to ensure this is to use one of the ``fetchAll*()`` methods):
 
 ::
 
     <?php
     $stmt = $conn->executeCacheQuery($query, $params, $types, new QueryCacheProfile(0, "some key"));
     $data = $stmt->fetchAllAssociative();
-    $stmt->closeCursor(); // at this point the result is cached
 
 .. warning::
 
--- a/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
+++ b/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
@@ -50,14 +50,7 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement, Forwar
     /** @var ResultStatement */
     private $statement;
 
-    /**
-     * Did we reach the end of the statement?
-     *
-     * @var bool
-     */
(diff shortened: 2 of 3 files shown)


greg0ire (commented):
Great idea! I used the caching layer a few weeks ago and had a hard time making it work, until I read the docs more carefully. I think they should be changed, namely `https://github.com/doctrine/dbal/blob/2.11.x/docs/en/reference/caching.rst`

beberlei:
I haven't seen this code for a long while, but does this ever read from the cache? `fetchAll` immediately delegates to the nested statement for example.

morozov:
> I haven't seen this code for a long while, but does this ever read from the cache? `fetchAll` immediately delegates to the nested statement for example.

This class is responsible only for reading from the underlying statement and storing the result in the cache if there was no hit. Reading from cache is implemented in `Connection::executeCacheQuery()`: https://github.com/doctrine/dbal/blob/acf84c3dfaa29e158f3c8e8a3b513c5fb210ff1e/lib/Doctrine/DBAL/Connection.php#L1204-L1207

morozov:
> […] I read the docs more carefully. I think they should be changed […]

Thanks for pointing out, @greg0ire. Actually, the following bit can be also removed from the documentation as of https://github.com/doctrine/dbal/pull/3070:

https://github.com/doctrine/dbal/blob/96f8cabf4fed81d00dd7a1b1cde3c76f9e2f8a5f/docs/en/reference/caching.rst#L48-L50

greg0ire (dismissed):

greg0ire:
Please squash your commits unless the squash button is enabled.

greg0ire (dismissed):

morozov:
This had to be rebased due to a conflict with https://github.com/doctrine/dbal/pull/4050.

greg0ire (approved):

PR #4049, opened 2020-06-06
Replace forward-compatible ResultStatement interfaces with Result

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no

This patch backports the interfaces being introduced in #4045. Unfortunately, since `Statement::execute()` already returns `bool` there's no way to implement a forward-compatible behavior.

Before:
```php
$statement->closeCursor();
```
After:
```php
$statement->free();
```

### DIFF
--- a/lib/Doctrine/DBAL/Driver/Result.php
+++ b/lib/Doctrine/DBAL/Driver/Result.php
@@ -83,9 +83,7 @@ interface Result
     public function columnCount();
 
     /**
-     * Closes the cursor, enabling the statement to be executed again.
-     *
-     * @return bool TRUE on success or FALSE on failure.
+     * Discards the non-fetched portion of the result, enabling the originating statement to be executed again.
      */
-    public function closeCursor();
+    public function free(): void;
 }
--- a/lib/Doctrine/DBAL/Driver/Result.php
+++ b/lib/Doctrine/DBAL/Driver/Result.php
@@ -4,12 +4,10 @@ declare(strict_types=1);
 
 namespace Doctrine\DBAL\Driver;
 
-use Doctrine\DBAL\Driver\ResultStatement as BaseResultStatement;
-
 /**
  * Driver-level result statement execution result.
  */
-interface Result extends BaseResultStatement
+interface Result
 {
     /**
      * Returns the next row of the result as a numeric array or FALSE if there are no more rows.
@@ -64,4 +62,19 @@ interface Result extends BaseResultStatement
      * @throws DriverException
      */
     public function fetchFirstColumn(): array;
+
+    /**
+     * Returns the number of columns in the result
+     *
+     * @return int The number of columns in the result. If the columns cannot be counted,
+     *             this method must return 0.
+     */
+    public function columnCount();
+
+    /**
+     * Closes the cursor, enabling the statement to be executed again.
+     *
+     * @return bool TRUE on success or FALSE on failure.
+     */
+    public function closeCursor();
 }
--- a/lib/Doctrine/DBAL/Driver/Result.php
+++ b/lib/Doctrine/DBAL/Driver/Result.php
@@ -2,18 +2,17 @@
 
 declare(strict_types=1);
 
-namespace Doctrine\DBAL\ForwardCompatibility\Driver;
+namespace Doctrine\DBAL\Driver;
 
-use Doctrine\DBAL\Driver\DriverException;
 use Doctrine\DBAL\Driver\ResultStatement as BaseResultStatement;
 
 /**
- * Forward compatibility extension for the ResultStatement interface.
+ * Driver-level result statement execution result.
  */
-interface ResultStatement extends BaseResultStatement
+interface Result extends BaseResultStatement
 {
     /**
-     * Returns the next row of a result set as a numeric array or FALSE if there are no more rows.
+     * Returns the next row of the result as a numeric array or FALSE if there are no more rows.
      *
      * @return array<int,mixed>|false
      *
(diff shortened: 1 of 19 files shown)


greg0ire:
> Unfortunately, since Statement::execute() already returns bool there's no way to implement a forward-compatible behavior.

Sounds indeed complicated maybe @derrabus would have an idea?

Only ugly solutions come to mind:
- relying on global state (some constant)
- supporting an extra argument to specify you want the new behavior
- renaming the method

derrabus:
Thank you for the ping. Splitting statement and result into different classes is a reasonable change. Thanks again for working on this.

So the question is, how to deal with the changed return type of `Statement::execute()`? It would be great if we could craft a forward-compatible solution here.

> Only ugly solutions come to mind:
> 
> * relying on state (some global constant, or some local property)
> 
> * supporting an extra argument to specify you want the new behavior
> 
> * renaming the method

All three would probably work. Right now, I would be in favor of version 2.

Another idea could be that `Connection::prepare()` is replaced by a new method that returns a statement with the new behavior. (Not a suggestion, just brainstorming)

I could offer to work on forward-compatible behavior for `execute()` once this PR is merged.

derrabus:
Would it make sense to backport the new result classes from #4045 as well, so that the old statement classes would act as proxies to the new result classes for all functionality that has been moved?

morozov:
> So the question is, how to deal with the changed return type of `Statement::execute()`? It would be great if we could craft a forward-compatible solution here.

Could you provide a real example of the boolean return value of `execute()` being used by the caller? As far as I can tell, it's a rudiment that exists because of DBAL extending PDO APIs. The failure to execute a statement is in fact reported by throwing an exception.

morozov:
> Would it make sense to backport the new result classes from #4045 as well, so that the old statement classes would act as proxies to the new result classes for all functionality that has been moved?

This was the original intent but given that we cannot return the result from `execute()`, what's the point of changing statement implementations? Right now, the statement itself implements the `Result` interface, so the consuming code can be changed to expect this type instead of the old one.

derrabus:
> Could you provide a real example of the boolean return value of `execute()` being used by the caller?

In a pure DBAL environment, it is pointless to check for the return value because, as you pointed out, `true` is the only value to expect here.

However, I could imagine cases where the connection is passed to some library functionality that operates on generic PDO connections and thus needs to handle cases where `PDO::ATTR_ERRMODE` is not set to `PDO::ERRMODE_EXCEPTION`.

That code would probably look like this:

```
if (!$someStatement->execute($someParams)) {
    // Do some error handling.
}
```

That code would still work even if `execute()` would suddenly return an object. A piece of code that would break would be this one:

```
if (true !== $someStatement->execute($someParams)) {
    // Do some error handling.
}
```

I wouldn't write that kind of code, but I've seen projects where this was considered best practice. I expect the impact of the return type change to be pretty low, if that was the answer you're looking for. Yet, it's a breaking change.

derrabus:
> > Would it make sense to backport the new result classes from #4045 as well, so that the old statement classes would act as proxies to the new result classes for all functionality that has been moved?
> 
> This was the original intent but given that we cannot return the result from `execute()`, what's the point of changing statement implementations?

You're right, you wouldn't gain much. My strongest argument would be that you'd keep the diff between 2.11 and 3.0 smaller which could be helpful for the maintenance later on.

morozov:
> My strongest argument would be that you'd keep the diff between 2.11 and 3.0 smaller which could be helpful for the maintenance later on.

To me, it looks like too much effort for such little gain.

Let's take the driver-level statement and result as an example. Since the new `Result` interface implements only part of the result-related `Statement` interface (e.g. the result doesn't implement fetching in the mixed mode), the proxy will have either have to emulate it on top of the `Result` API or implement it itself. The latter will require sharing the underlying statement resource between the result and the proxy.

In the case of the wrapper-level, it's even more complicated: the wrapper statement wraps a driver-level statement that may or may not implement the new interface. In this case, proxying to the `Result` is only possible if the wrapped statement implements `Result` and the method being called is part of the `Result` interface. Otherwise, see above.

morozov:
> However, I could imagine cases where the connection is passed to some library functionality that operates on generic PDO connections and thus needs to handle cases where `PDO::ATTR_ERRMODE` is not set to `PDO::ERRMODE_EXCEPTION`.

This doesn't look like a valid case. The fact that a library does `if (true !== $stmt->execute())` means that it expects a non-true value to be returned in the case of an error but it will get an exception from DBAL.

> I expect the impact of the return type change to be pretty low, if that was the answer you're looking for. Yet, it's a breaking change.

I totally agree. Not sure how we got into this argument. We're not going to change the return value in `2.x` for the sake of providing an FC layer.

lib/Doctrine/DBAL/Driver/Result.php
```diff
@@ -65,4 +62,31 @@ public function fetchAllAssociative(): array;
...
+     * @return int The number of rows.
+     */
+    public function rowCount();
+
+    /**
+     * Returns the number of columns in the result set
+     *
+     * @return int The number of columns in the result set represented
+     *                 by the PDOStatement object. If there is no result set,
+     *                 this method should return 0.
```
greg0ire, line 84:
```suggestion
     *             by the PDOStatement object. If there is no result set,
     *             this method should return 0.
```

derrabus:
> This doesn't look like a valid case. The fact that a library does `if (true !== $stmt->execute())` means that it expects a non-true value to be returned in the case of an error but it will get an exception from DBAL.

Yes, but you won't necessarily get an exception from PDO. This is why code like this might be in place in libraries that are not aware of DBAL and operate on generic PDO connections.

It's still not good code and it might be an edge-case. But it's the best example I could come up with.

lib/Doctrine/DBAL/Driver/Result.php
morozov: One day we will also scrape out all those PDO references from non-PDO related code.

greg0ire (approved):

PR #4053, opened 2020-06-06
Report coverage unconditionally on Travis

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

pcov is fast enough for this to be acceptable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,7 @@ cache:
 
 before_install:
   - phpenv config-rm xdebug.ini || true
-  - |
-    if [ "x$COVERAGE" == "xyes" ]; then
-      pecl install pcov-1.0.0
-    fi
+  - pecl install pcov
 
 before_script:
   - |
@@ -24,25 +21,17 @@ install:
   - travis_retry composer -n install --prefer-dist
 
 script:
-  - |
-    if [ "x$COVERAGE" == "xyes" ]; then
-       ./vendor/bin/phpunit --configuration tests/travis/$DB.travis.xml --coverage-clover clover.xml
-    else
-       ./vendor/bin/phpunit --configuration tests/travis/$DB.travis.xml
-    fi
+  - ./vendor/bin/phpunit --configuration tests/travis/$DB.travis.xml --coverage-clover clover.xml
 
 after_success:
-  - |
-    if [ "x$COVERAGE" == "xyes" ]; then
-      bash <(curl -s https://codecov.io/bash)
-    fi
+  - bash <(curl -s https://codecov.io/bash)
 
 jobs:
   include:
 
     - stage: Smoke Testing
       php: 7.3
-      env: DB=sqlite COVERAGE=yes
+      env: DB=sqlite
 
     - stage: Test
       php: 7.2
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -8,10 +8,7 @@ cache:
 
 before_install:
   - phpenv config-rm xdebug.ini || true
-  - |
-    if [ "x$COVERAGE" == "xyes" ]; then
-      pecl install pcov-1.0.0
-    fi
+  - pecl install pcov-1.0.0
```
morozov, line 11: Can we do just `pecl install pcov` and use the latest stable? It may be required for compatibility with the latest PHP versions.

greg0ire:
This raises the question about the Smoke Testing step… should I keep it and remove the sqlite job from test? That way if something is wrong with the unit tests we don't run a bazillion tests.

morozov:
> This raises the question about the Smoke Testing step… should I keep it and remove the sqlite job from test? That way if something is wrong with the unit tests we don't run a bazillion test jobs for nothing

Definitely. That's the idea. If there's a unit test failure, you don't want to wait 10 minutes until the DB2 job completes. Ideally, I'd like to not run tests until all static analysis phases succeed (which was the case before we moved some jobs to Actions) but I'm not sure how this could be implemented.

greg0ire:
I think this will be achieved by moving to GA, with [job dependencies](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobs). BTW @bendavies , are you still planning to work on that? :pray: :crossed_fingers:

morozov:
Looks good :+1: Please squash.

greg0ire:
> Looks good +1 Please squash.

Should I really? I mean I deliberately didn't, since it does not have to do with running coverage unconditionally… using the latest stable could be something we would want to do or revert independently, or maybe there is a link I don't see?

morozov:
Without the second commit, the first one fails on PHP 7.4.

greg0ire:
Oh all right then, didn't notice.

morozov (approved):

greg0ire:
Ok now continousphp is flaky too…

bendavies:
> I think this will be achieved by moving to GA, with [job dependencies](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobs). BTW @bendavies , are you still planning to work on that?

I do have a branch that moves postgres and myswl tests to GitHub actions.
It needs work.
I can submit it I'd you'd like to work on it together.

morozov:
While I like the idea as such, before we go full steam ahead, could we establish some expectations of the end result? E.g. Travis currently allows us to run 10 jobs in parallel (I believe it's a favor since the [default for OSS is 5](https://twitter.com/travisci/status/651856122559774722), may be outdated) and we execute ~40-50 jobs per build.

Will we be able to do the same on GitHub?

bendavies:
> While I like the idea as such, before we go full steam ahead, could we establish some expectations of the end result? E.g. Travis currently allows us to run 10 jobs in parallel (I believe it's a favor since the [default for OSS is 5](https://twitter.com/travisci/status/651856122559774722), may be outdated) and we execute ~40-50 jobs per build.
> 
> Will we be able to do the same on GitHub?

I think 20 as per https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#usage-limits

There might be other reasons that would prevent migrating.
Last time I checked, secrets were not shared with forks. Are any secrets used?

morozov:
> Last time I checked, secrets were not shared with forks. Are any secrets used?

I believe this is by design on any public CI. No secrets are currently used.

Issue #4052, opened 2020-06-06
RFC: Rename MasterSlaveConnection


albe:
I would like to suggest to rename the MasterSlaveConnection extension to something more sensible.
My suggestion would be `PrimaryReplicaConnection`, since that's one version of naming alternatives for the concept that is widely used (https://en.wikipedia.org/wiki/Master/slave_(technology)#Terminology_concerns).

Since this part of the code is primarily concerned about managing connections between writable and readable database replicas it could also be named `WriteReadConnection`, though IMO that could be a bit more misleading.

IF this and a specific naming is agreed upon by the team, I would take on the code part of this change.

For b/c an alias for the old class name could be kept around at least for a bit of time.

greg0ire:
Thanks for proposing this change. I would go with `PrimaryReplicaConnection`. If we are going with an alias, this might help: https://dev.to/greg0ire/how-to-deprecate-a-type-in-php-48cf

Disclaimer though, this method caused us a lot of headaches with `doctrine/persistence`, because types we deprecated were used in signatures. I'm not sure that would be likely here though.

An alternative would be to just not use aliases and copy/paste/improve the class (making it `final`, using private properties, using type declarations). We would then deprecate `MasterSlaveConnection` (and maybe stop maintenance on it?).

beberlei:
yes lets do it, rename it and keep the old class + deprecate + remove in the next major.

albe:
Thanks for the quick feedback and action! Guess this is indeed a good opportunity to not just rename, but deprecate and create an improved version of this connection strategy.

beberlei:
Closed with #7047798

PR #4054, opened 2020-06-06
[GH-4052] Deprecate MasterSlaveConnection and rename to PrimaryReplicaConnection

beberlei:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #4052 

#### Summary

Renames `MasterSlaveConncetion` to `PrimaryReplicaConnection` and all params or connection keys from `master` to `primary` and from `slaves/slave` to `replica`.

A completly BC implementation calling the new class is provided as well and throws deprecation messages.

Usage:

```php
$conn = DriverManager::getConnection(array(
   'wrapperClass' => 'Doctrine\DBAL\Connections\PrimaryReplicaConnection',
   'driver' => 'pdo_mysql',
   'primary' => array('user' => '', 'password' => '', 'host' => '', 'dbname' => ''),
   'replica' => array(
       array('user' => 'replica1', 'password', 'host' => '', 'dbname' => ''),
       array('user' => 'replica2', 'password', 'host' => '', 'dbname' => ''),
   )
));
```

Open question: Should the master => primary and slave/salves => replica renames for constructor params and connect method trigger deprecations on MasterSlaveConnection only, or also when wrongly used on PrimaryReplicaConnection?

### DIFF
--- a/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
+++ b/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
@@ -4,88 +4,21 @@ namespace Doctrine\DBAL\Connections;
 
 use Doctrine\Common\EventManager;
 use Doctrine\DBAL\Configuration;
-use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Driver;
-use Doctrine\DBAL\Driver\Connection as DriverConnection;
-use Doctrine\DBAL\Event\ConnectionEventArgs;
-use Doctrine\DBAL\Events;
 use InvalidArgumentException;
 
-use function array_rand;
-use function assert;
-use function count;
-use function func_get_args;
+use function sprintf;
+use function trigger_error;
+
+use const E_USER_DEPRECATED;
 
 /**
- * Master-Slave Connection
- *
- * Connection can be used with master-slave setups.
- *
- * Important for the understanding of this connection should be how and when
- * it picks the slave or master.
- *
- * 1. Slave if master was never picked before and ONLY if 'getWrappedConnection'
- *    or 'executeQuery' is used.
- * 2. Master picked when 'exec', 'executeUpdate', 'insert', 'delete', 'update', 'createSavepoint',
- *    'releaseSavepoint', 'beginTransaction', 'rollback', 'commit', 'query' or
- *    'prepare' is called.
- * 3. If master was picked once during the lifetime of the connection it will always get picked afterwards.
- * 4. One slave connection is randomly picked ONCE during a request.
- *
- * ATTENTION: You can write to the slave with this connection if you execute a write query without
- * opening up a transaction. For example:
- *
- *      $conn = DriverManager::getConnection(...);
- *      $conn->executeQuery("DELETE FROM table");
- *
- * Be aware that Connection#executeQuery is a method specifically for READ
- * operations only.
- *
- * This connection is limited to slave operations using the
- * Connection#executeQuery operation only, because it wouldn't be compatible
- * with the ORM or SchemaManager code otherwise. Both use all the other
- * operations in a context where writes could happen to a slave, which makes
- * this restricted approach necessary.
- *
- * You can manually connect to the master at any time by calling:
- *
- *      $conn->connect('master');
- *
- * Instantiation through the DriverManager looks like:
- *
- * @example
- *
- * $conn = DriverManager::getConnection(array(
- *    'wrapperClass' => 'Doctrine\DBAL\Connections\MasterSlaveConnection',
- *    'driver' => 'pdo_mysql',
- *    'master' => array('user' => '', 'password' => '', 'host' => '', 'dbname' => ''),
- *    'slaves' => array(
- *        array('user' => 'slave1', 'password', 'host' => '', 'dbname' => ''),
- *        array('user' => 'slave2', 'password', 'host' => '', 'dbname' => ''),
- *    )
- * ));
- *
- * You can also pass 'driverOptions' and any other documented option to each of this drivers to pass additional information.
+ * @deprecated Use PrimaryReadReplicaConnection instead
  */
-class MasterSlaveConnection extends Connection
+class MasterSlaveConnection extends PrimaryReadReplicaConnection
 {
     /**
-     * Master and slave connection (one of the randomly picked slaves).
-     *
-     * @var DriverConnection[]|null[]
-     */
-    protected $connections = ['master' => null, 'slave' => null];
-
-    /**
-     * You can keep the slave connection and then switch back to it
-     * during the request if you know what you are doing.
-     *
-     * @var bool
-     */
-    protected $keepSlave = false;
-
-    /**
-     * Creates Master Slave Connection.
+     * Creates Primary Replica Connection.
      *
      * @param mixed[] $params
      *
@@ -93,32 +26,40 @@ class MasterSlaveConnection extends Connection
      */
     public function __construct(array $params, Driver $driver, ?Configuration $config = null, ?EventManager $eventManager = null)
     {
-        if (! isset($params['slaves'], $params['master'])) {
-            throw new InvalidArgumentException('master or slaves configuration missing');
+        $this->deprecated(self::class, PrimaryReadReplicaConnection::class);
+
+        if (isset($params['master'])) {
+            $this->deprecated('Params key "master"', '"primary"');
+
+            $params['primary'] = $params['master'];
+            unset($params['master']);
         }
 
-        if (count($params['slaves']) === 0) {
-            throw new InvalidArgumentException('You have to configure at least one slaves.');
+        if (isset($params['slaves'])) {
+            $this->deprecated('Params key "slaves"', '"replica"');
+
+            $params['replica'] = $params['slaves'];
+            unset($params['slaves']);
         }
 
-        $params['master']['driver'] = $params['driver'];
-        foreach ($params['slaves'] as $slaveKey => $slave) {
-            $params['slaves'][$slaveKey]['driver'] = $params['driver'];
-        }
+        if (isset($params['keepSlave'])) {
+            $this->deprecated('Params key "keepSlave"', '"keepReplica"');
 
-        $this->keepSlave = (bool) ($params['keepSlave'] ?? false);
+            $params['keepReplica'] = $params['keepSlave'];
+            unset($params['keepSlave']);
+        }
 
         parent::__construct($params, $driver, $config, $eventManager);
     }
 
     /**
-     * Checks if the connection is currently towards the master or not.
-     *
-     * @return bool
+     * Checks if the connection is currently towards the primary or not.
      */
-    public function isConnectedToMaster()
+    public function isConnectedToMaster(): bool
     {
-        return $this->_conn !== null && $this->_conn === $this->connections['master'];
+        $this->deprecated('isConnectedtoMaster()', 'isConnectedToPrimary()');
+
+        return $this->isConnectedToPrimary();
     }
 
     /**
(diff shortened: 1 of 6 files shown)


morozov (dismissed):
While I'm fully onboard with the rename, the things I don't like about the current and the new implementation are:

1. It inherits the wrapper connection instead of using composition (not sure if it's possible with the current design).
2. It changes the semantic of the `connect()` method by adding an optional argument.

While the above is not currently a blocker, I'm curious if anyone had any thoughts about redesigning it rather than just renaming.

morozov:
@beberlei are you sure we want to introduce a deprecation in `2.10.x`? It probably should be `2.11.x`.

greg0ire:
> While the above is not currently a blocker, I'm curious if anyone had any thoughts about redesigning it rather than just renaming.

I had the same thought in #4052 : 

> An alternative would be to just not use aliases and copy/paste/improve the class (making it final, using private properties, using type declarations). We would then deprecate MasterSlaveConnection (and maybe stop maintenance on it?).

greg0ire:
I'm working on a alternate implementation so that you can see what I mean.

greg0ire:
See #4055

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,407 @@
...
+    {
+        return $this->_conn !== null && $this->_conn === $this->connections['primary'];
+    }
+
+    /**
+     * @param string|null $connectionName
+     *
+     * @return bool
+     */
+    public function connect($connectionName = null, bool $triggerDeprecationWhenConnectionNameUsed = true)
```
greg0ire, line 129:
```suggestion
    public function connect($connectionName = null, bool $triggerDeprecationWhenConnectionNameUsed = true)
```

Is the second argument used anywhere? It makes Psalm fail.

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -4,378 +4,95 @@
...
         }
 
-        if (isset($this->connections[$connectionName])) {
-            $this->_conn = $this->connections[$connectionName];
-
-            if ($forceMasterAsSlave && ! $this->keepSlave) {
-                $this->connections['slave'] = $this->_conn;
-            }
+        if (isset($params['slaves'])) {
+            $this->deprecated("Params key 'slaves'", "'replica'");
```
stof, line 39: I'd rather use quotes than apostrophs around the key names in the message (I know that devs think about `'` being a single quote, but that's not what it is in English sentences).

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -4,378 +4,95 @@
...
         if ($connectionName === 'master') {
-            return $params['master'];
-        }
-
-        $config = $params['slaves'][array_rand($params['slaves'])];
+            $connectionName = 'primary';
 
-        if (! isset($config['charset']) && isset($params['master']['charset'])) {
-            $config['charset'] = $params['master']['charset'];
+            $this->deprecated('connect("master")', 'connect("primary")');
```
stof, line 75: this should hint at `ensureConnectedToPrimary()` instead. and I even suggest not supporting `connect('primary')` at all (so that the new class does not need to add a parameter in `connect()` compared to the parent Connection class)

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -4,378 +4,95 @@
...
-            $logger->startQuery($args[0]);
-        }
-
-        $statement = $this->_conn->query(...$args);
-
-        $statement->setFetchMode($this->defaultFetchMode);
-
-        if ($logger) {
-            $logger->stopQuery();
+            $this->deprecated('connect("slave")', 'connect("replica")');
```
stof, line 81: should hint at `ensureConnectedToReplica()` instead.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,407 @@
...
+ * 4. One replica connection is randomly picked ONCE during a request.
+ *
+ * ATTENTION: You can write to the replica with this connection if you execute a write query without
+ * opening up a transaction. For example:
+ *
+ *      $conn = DriverManager::getConnection(...);
+ *      $conn->executeQuery("DELETE FROM table");
+ *
+ * Be aware that Connection#executeQuery is a method specifically for READ
+ * operations only.
```
stof, line 42: I would add a sentence hinting at `Connection#executeUpdate` for write operations (which can be inserts and deletes, not only updates)

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,407 @@
...
+
+        parent::__construct($params, $driver, $config, $eventManager);
+    }
+
+    /**
+     * Checks if the connection is currently towards the primary or not.
+     *
+     * @return bool
+     */
+    public function isConnectedToPrimary()
```
stof, line 119: As this is a new method, you could add the `: bool` return type without any BC issue.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
stof, line 129: Do we actually need to keep support for `$connectionName` at all in this new class ? The BC layer could live in the deprecated child class.

stof: The method accepting a connection name in that class could be a private method, called both from `connect()` (with `'replica` as value for the connection name, and `$requestedConnectionChange` set to `false`), and from the new `ensureConnectedTo*` methods.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,407 @@
...
+
+        return $config;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function executeUpdate($query, array $params = [], array $types = [])
+    {
+        $this->connect('primary');
```
stof, line 254: Use `ensureConnectedToPrimary()` rather than the deprecated API.

stof: same in all other methods

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,407 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    public function close()
+    {
+        unset($this->connections['primary'], $this->connections['replica']);
+
+        parent::close();
+
+        $this->_conn       = null;
```
stof, line 308: isn't this already done in `parent::close()` ?

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,407 @@
...
+     * {@inheritDoc}
+     */
+    public function query()
+    {
+        $this->connect('primary');
+        assert($this->_conn instanceof DriverConnection);
+
+        $args = func_get_args();
+
+        $logger = $this->getConfiguration()->getSQLLogger();
```
stof, line 382: Why duplicating the parent method here rather than calling it ?

beberlei: this is just copied code from 2011 when this class was introduced.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
beberlei: this is just copied code from 2011 when this class was introduced.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
beberlei: Made this change.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
beberlei: done

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
beberlei: This went away with the changes @stof suggested

tests/Doctrine/Tests/DBAL/Functional/PrimaryReadReplicaConnectionTest.php
```diff
@@ -0,0 +1,243 @@
...
+class PrimaryReadReplicaConnectionTest extends DbalFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $platformName = $this->connection->getDatabasePlatform()->getName();
+
+        // This is a MySQL specific test, skip other vendors.
+        if ($platformName !== 'mysql') {
```
greg0ire, line 32: @stof said this was supposed to work with PG too, should we maybe change this line?

beberlei: @greg0ire yes thats a future todo imho, the test actually only works with MySQL because of the charset test.

greg0ire (dismissed):

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -4,378 +4,95 @@
...
-
-        $statement->setFetchMode($this->defaultFetchMode);
-
-        if ($logger) {
-            $logger->stopQuery();
+            $this->deprecated('connect("slave")', 'ensureConnectedToReplica()');
         }
 
-        return $statement;
+        return parent::connect($connectionName);
```
stof, line 84: This is broken. When passing `slave` or `master` as connection name, you must call the alternative method instead of calling `parent::connect($connectionName);` with a renamed connection name, as you changed the parent class.

`parent::connect()` should be called only for remaining cases

lib/Doctrine/DBAL/Connections/PrimaryReadReplicaConnection.php
```diff
@@ -0,0 +1,413 @@
...
+
+    /**
+     * @param string|null $connectionName
+     *
+     * @return bool
+     */
+    public function connect($connectionName = null)
+    {
+        if ($connectionName !== null) {
+            throw new InvalidArgumentException('Passing a connection name as first argument is not supported anymore. Use ensureConnectedToPrimary()/ensureConnectedToReplica() instead.');
```
stof, line 133: Do we even need that argument and that exception ? This is a new class.

tests/Doctrine/Tests/DBAL/Functional/MasterSlaveConnectionTest.php
```diff
@@ -1,243 +0,0 @@
...
-use function strlen;
-use function strtolower;
-use function substr;
-
-use const CASE_LOWER;
-
-/**
- * @group DBAL-20
- */
-class MasterSlaveConnectionTest extends DbalFunctionalTestCase
```
stof, line 23: these tests should be kept (as legacy tests), to ensure that the deprecated class still works fine.

lib/Doctrine/DBAL/Connections/PrimaryReadReplicaConnection.php
beberlei: @stof yes i believe, the problem is that passing to many arguments does not throw an error in strict_types=0, as such if someone would wrongly convert their code and keep one `connect("master")` or slave somewhere they would never notice.

stof (dismissed):

lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
```diff
@@ -4,378 +4,95 @@
...
-     */
-    public function prepare($statement)
+    private function deprecated(string $thing, string $instead): void
     {
-        $this->connect('master');
-
-        return parent::prepare($statement);
+        @trigger_error(
+            sprintf(
+                '%s is deprecated since doctrine/dbal 2.10 and will be removed in 3.0, use %s instead.',
```
stof, line 91: @beberlei this should be updated to `2.11` instead.

stof (approved):

morozov:
@beberlei or @greg0ire could you merge this up, please? I'm not sure I will resolve the conflicts correctly.

greg0ire:
I'm on it!

greg0ire:
https://github.com/doctrine/dbal/pull/4063

morozov:
@greg0ire could you please also file a PR that removes the deprecated class from `3.0.x`?

Issue #3593, opened 2019-06-06
Connection::getParams() leaks implementation details


morozov:
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.9.3

The fact that the `Connection::getParams()` method exists makes it impossible to make any changes in the structure of the connection `$params` array w/o breaking the API. In #3568, we already removed `::getHost()`, `::getPort()`, `::getUsername()` and `::getPassword()`.

We need to see if it's possible to get rid of the method entirely or at least reduce its usage and mark it as `@internal`.

PR #4055, opened 2020-06-07
Deprecate master/slave terminology and underlying feature

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #4052

#### Summary

`MasterSlaveConnection` is left untouched, deprecated, copy/pasted/improved
Alternative to #4054

Targeting 2.11.x because of the deprecation.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,10 @@
 # Upgrade to 2.11
 
+## Deprecated `MasterSlaveConnection`
+
+The naming is offensive, the implementation very questionable. Use each
+endpoint connection explicitly.
+
 ## Deprecated `ArrayStatement` and `ResultCacheStatement` classes.
 
 The `ArrayStatement` and `ResultCacheStatement` classes are deprecated. In a future major release they will be renamed and marked internal as implementation details of the caching layer.
--- a/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
+++ b/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
@@ -15,6 +15,10 @@ use function array_rand;
 use function assert;
 use function count;
 use function func_get_args;
+use function sprintf;
+use function trigger_error;
+
+use const E_USER_DEPRECATED;
 
 /**
  * Master-Slave Connection
@@ -52,6 +56,8 @@ use function func_get_args;
  *      $conn->connect('master');
  *
  * Instantiation through the DriverManager looks like:
+ *
+ * @deprecated use regular connections instead
  *
  * @example
  *
@@ -93,6 +99,11 @@ class MasterSlaveConnection extends Connection
      */
     public function __construct(array $params, Driver $driver, ?Configuration $config = null, ?EventManager $eventManager = null)
     {
+        @trigger_error(sprintf(
+            'Class "%s" is deprecated since doctrine/dbal 2.11 and will be removed in 3.0',
+            self::class
+        ), E_USER_DEPRECATED);
+
         if (! isset($params['slaves'], $params['master'])) {
             throw new InvalidArgumentException('master or slaves configuration missing');
         }


lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,359 @@
...
+            $this->_eventManager->dispatchEvent(Events::postConnect, $eventArgs);
+        }
+
+        return true;
+    }
+
+    /**
+     * Connects to a specific connection.
+     */
+    private function connectTo(string $connectionName): DriverConnection
```
greg0ire, line 182: If you think somebody relies on `connectTo` or `chooseConnectionConfiguration`, we could make them public. Doing so after this is merged and release would not be a BC break though.

albe (dismissed):

lib/Doctrine/DBAL/DriverManager.php
```diff
@@ -142,17 +142,29 @@ public static function getConnection(
 
         $params = self::parseDatabaseUrl($params);
 
-        // URL support for MasterSlaveConnection
+        // @todo: deprecated, notice thrown by connection constructor
```
SenseException, line 145: Will the `Connection` constructor get a deprecation message in this PR?

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,359 @@
...
+
+    /**
+     * Checks if the connection is currently towards the primary or not.
+     */
+    public function isConnectedToPrimary(): bool
+    {
+        return $this->_conn !== null && $this->_conn === $this->connections['primary'];
+    }
+
+    public function connect(?string $connectionName = null): bool
```
morozov, line 127: Please see https://github.com/doctrine/dbal/pull/4054#pullrequestreview-425775128. I am concerned about using a proper class name and leaving all downsides of the old API. That means that if we decide to fix this downsides in a BC-breaking way, we'll have to use another class name. It's better to try an fix both the class name and the API problem in one shot.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,359 @@
...
+ *    'primary' => array('user' => '', 'password' => '', 'host' => '', 'dbname' => ''),
+ *    'replica' => array(
+ *        array('user' => 'replica1', 'password', 'host' => '', 'dbname' => ''),
+ *        array('user' => 'replica2', 'password', 'host' => '', 'dbname' => ''),
+ *    )
+ * ));
+ *
+ * You can also pass 'driverOptions' and any other documented option to each of this drivers to pass additional information.
+ */
+final class PrimaryReplicaConnection extends Connection
```
morozov, line 71: Since this class implements public methods outsides the ones it extends, marking it final makes it impossible to extend (see the [article](https://ocramius.github.io/blog/when-to-declare-classes-final/)). Why is this needed?

lib/Doctrine/DBAL/DriverManager.php
greg0ire: Ah right that's a part I forgot to backport from #4054

greg0ire: Unsure about the `@todo` part, I copied it from #4054, is this a way we have to remember to remove things on 3.0.x?

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
greg0ire: > It changes the semantic of the connect() method by adding an optional argument.

https://github.com/doctrine/dbal/blob/0ad4c165318128c50d9a425c8c15520c094f465b/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php#L128

There's no new argument, the only existing argument was already optional, wasn't it?

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
greg0ire: Indeed, I will remove it.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
greg0ire: Oh right you were talking about the one in the interface, I get it now… how do we fix this?

lib/Doctrine/DBAL/DriverManager.php
SenseException: I don't know. I thought `@deprecated` and/or `trigger_error()` is usually in use.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
morozov: Something like that. I haven't use this component, so I don't know what the intended use is. E.g. why does a single connection need to be stateful and switch between different underlying connection? If a given component is about to make a query or execute a statement, how does it know if it needs to switch or the connection is already switched to the proper endpoint?

greg0ire: @beberlei @lstrojny might be able to help with this?

The more I read about this, the less I see why you would need to be able to `$this->connect('replica')`, since `replica` is the default, and since `If Primary was picked once during the lifetime of the connection it will always get picked afterwards.`

> If a given component is about to make a query or execute a statement, how does it know if it needs to switch or the connection is already switched to the proper endpoint?

It could use `isConnectedToPrimary()` for that, but apparently it shouldn't even need to, since `Primary picked when 'exec', 'executeUpdate', 'insert', 'delete', 'update', 'createSavepoint', 'releaseSavepoint', 'beginTransaction', 'rollback', 'commit', 'query' or l'prepare' is called.`

So the only use case I see for `$this->connect('primary')` is when you know you are going to perform a call to `executeQuery` and then calls that require using the primary connection, and you want to make sure you only ever connect to one things for performance reasons.

morozov: I see two ways of implementing the primary/replica connection consumer in an application:

1. The actual endpoint being used for a specific statement is abstracted out and is chosen based on whether the statement is read or write. In this case, the application should depend on a `Connection` interface (no additional methods exposed).
2. The application wants to control which endpoint to use for a specific statement. This way, it's given two or more independent connections that don't expose any additional methods either.

All this `$this->connect('primary' /* ATTENTION, MAGIK STRING LITERAL! */)` and `$this->forceUsingPrimary(/* ATTENTION, MOAR MAGIK! */)` looks like devilry to me.

greg0ire: With 1. we would drop the argument and `isConnectedToPrimary()`, so I could restore `final`, right? I love that keyword, it usually means fewer headaches with BC-breaks, that why I tend to use it (maybe a bit too much, as you pointed out).
2. This could be fixed by moving this grouping of connections as a concern for a DIC. The doctrine bundle would name connections according to a pattern, like `connection_foo_primary`, `connection_foo_replica0`, `connection_foo_replica1`… If that's not convenient enough, a service locator could be created to have `$container->get('primary')`, `$connection->get('replica0')`, etc. (where `$container` is a service locator that contains only these 3 connections). 

I'm going to push another commit showing how this could look like. If we do that, I think we should also drop `$keepReplica`: 


```
    /**
     * You can keep the replica connection and then switch back to it
     * during the request if you know what you are doing.
     *
     * @var bool
     */
    private $keepReplica = false;
```

greg0ire: There is going to be an issue with the tests though: they heavily rely on `isConnectedToPrimary`

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,359 @@
...
+{
+    /**
+     * Primary and Replica connection (one of the randomly picked replicas).
+     *
+     * @var DriverConnection[]|null[]
+     */
+    private $connections = ['primary' => null, 'replica' => null];
```
morozov, line 78: As far as I understand, it's one primary and many replicas. Why do we need to keep them under one array with magic keys? It's fine if we don't resolve it in this ticket.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
morozov: If somebody needs to choose the connection explicitly, they should operate those connections manually w/o this wrapper.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
morozov, line 71: This class still cannot be extended since there's no wrapper connection interface yet. Although I'm fine if we still want to make it final.

morozov (commented):
API-wise looks good. We should be able to address the implementation details issues later.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
morozov: It could be tested properly if this was a wrapper over multiple other connections that could be mocked. The "is connected to" is an implementation detail, not the behavior that needs to be tested.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
greg0ire: I'm not sure I understand this sentence the way you intend me to. But I just noticed that although some methods are also defined in the parent, they don't appear in any interface, like for instance `rollbackSavepoint`, so according to the article you linked to, it should still not be final. I'll remove that keyword.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
greg0ire: Why? And who would that somebody be? This method is `private`, so it can't be used by somebody outside this class if that's who you were thinking.

morozov: > If you think somebody relies on `connectTo` or `chooseConnectionConfiguration`, we could make them public.

It was the answer to your comment. I don't think that this should be made part of the public API.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
greg0ire: > Or, looking from another angle, where does the requirement of being connected only to one endpoint at a time comes from?

I'd say the belief that performance will be bad or results will be inaccurate if you perform a write on the primary db and immediately read from the replica. I have no idea if this holds true or not.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
greg0ire: I can resolve it in this ticket, I think that will be the easiest part TBH :P

greg0ire: In fact it's just the one replica (picked at random), as stated in the comment above. Your comment still applies though.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,361 @@
...
+            throw new InvalidArgumentException('You have to configure at least one replica.');
+        }
+
+        $params['primary']['driver'] = $params['driver'];
+        foreach ($params['replica'] as $replicaKey => $replica) {
+            $params['replica'][$replicaKey]['driver'] = $params['driver'];
+        }
+
+
+        parent::__construct($params, $driver, $config, $eventManager);
```
morozov, line 110: This is another place where my eyes are bleeding. The parent class doesn't understand the "replica" key but for some reason, we're passing it there. Shouldn't the concept of multiple connections be encapsulated to this very class?

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    private function switchToPrimary(): bool
```
morozov, line 113: Is the returned boolean used anywhere?

greg0ire: Nope, let's drop it.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,361 @@
...
+
+        return $statement;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function prepare($statement): Statement
+    {
+        $this->switchToPrimary();
```
morozov, line 357: This primary/replica thing in DBAL is a really poor and leaky abstraction. Why do all statements need to be prepared on the primary endpoint?

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
greg0ire: :thinking: uh yeah that's nasty. Are you suggesting `unset`-ing `replica` and `primary`, storing them in some other property, or not calling the parent constructor altogether?

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,361 @@
...
+    public function rollbackSavepoint($savepoint): void
+    {
+        $this->switchToPrimary();
+
+        parent::rollbackSavepoint($savepoint);
+    }
+
+    public function query(): Statement
+    {
+        $this->switchToPrimary();
```
morozov, line 331: And it's even worse here. Why are we switching to the primary endpoint to execute a _query_?

greg0ire: I won't like, I have no idea…

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
morozov: We cannot avoid extending the connection because there's no DBAL-level connection interface to implement. Unsetting and managing replicate within the class would help migrate to composition later.

greg0ire:
@morozov this looks nasty too:

https://github.com/doctrine/dbal/blob/0ad4c165318128c50d9a425c8c15520c094f465b/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php#L34-L41

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,361 @@
...
+        $params = $this->getParams();
+
+        $driverOptions = $params['driverOptions'] ?? [];
+
+        $connectionParams = $this->chooseConnectionConfiguration($connectionName, $params);
+
+        $user     = $connectionParams['user'] ?? null;
+        $password = $connectionParams['password'] ?? null;
+
+        return $this->_driver->connect($connectionParams, $user, $password, $driverOptions);
```
morozov, line 196: And this is another place where the abstraction leaks. Shouldn't the parent class implement the logic of converting the DBAL parameters to a `Driver::connect()` call?

https://github.com/doctrine/dbal/blob/b0ca9434e5429079f02f661f5aefea2021cd1c75/src/Connection.php#L350-L354

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
greg0ire: :thinking: would this be the reason: https://github.com/doctrine/dbal/blob/0ad4c165318128c50d9a425c8c15520c094f465b/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php#L43-L47

Not sure what it refers to.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,361 @@
...
+    /**
+     * Connects to a specific connection.
+     */
+    private function connectTo(string $connectionName): DriverConnection
+    {
+        $params = $this->getParams();
+
+        $driverOptions = $params['driverOptions'] ?? [];
+
+        $connectionParams = $this->chooseConnectionConfiguration($connectionName, $params);
```
morozov, line 191: So does this mean that every time we "switch" between the primary and replica endpoints, the replica will be randomly chosen and the old connection will be closed? What's the rationale behind this behavior?

greg0ire: :thinking: we are not supposed to switch from the primary to the replica, are we?

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,361 @@
...
+            }
+
+            return false;
+        }
+
+        if ($connectionName === 'primary') {
+            $this->primary = $this->_conn = $this->connectTo($connectionName);
+
+            // Set replica connection to primary to avoid invalid reads
+            $this->replica = $this->primary;
```
morozov, line 169: This is insane. Why instead of just keeping two connections separate and independent on the state, we assign one to the other? From the code readability standpoint, this is somewhat like "Monday = Tuesday".

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,361 @@
...
+
+        return $this->_driver->connect($connectionParams, $user, $password, $driverOptions);
+    }
+
+    /**
+     * @param mixed[] $params
+     *
+     * @return mixed[]
+     */
+    private function chooseConnectionConfiguration(string $connectionName, array $params): array
```
morozov, line 204: This method accepts only two possible values of `$connectionName` and then branches based on the value. Shouldn't it be split into two methods?

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
greg0ire: Should we just `unset($this->replica)`?

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
```diff
@@ -0,0 +1,361 @@
...
+
+        return true;
+    }
+
+    /**
+     * Connects to a specific connection.
+     */
+    private function connectTo(string $connectionName): DriverConnection
+    {
+        $params = $this->getParams();
```
morozov, line 187: This is BTW one of the places where I couldn't get rid of the call to `getParams()` (https://github.com/doctrine/dbal/issues/3593). The purpose of the connection is to transmit the data between the client and the server, not expose the connection parameters via the public API.

Having these implementation details exposed as is via the API makes refactoring of the internals close to impossible.

morozov:
Given the number of WTFs that this code produces, I'd personally prefer to just deprecate it for now. If we don't manage to implement a proper replacement, the old code could be just copy/pasted in the project that needs this functionality. But looking at the behaviors that it implements, I hope nobody in their right mind uses it in production.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
morozov: This is really questionable behavior. If I insert a "user logged in" event in one table and then query a heavy report from the other, why should the report use the master connection?

greg0ire:
@morozov I agree. In my company, we have a RW endpoint and an RO endpoint, we just created 2 connections, and don't use this class, and it works just fine. These features could be brought back at some point with a brand new design, that would have to avoid this unholy mix of composition and inheritance.

morozov:
> In my company, we have a RW endpoint and an RO endpoint, we just created 2 connections, and don't use this class, and it works just fine.

This is how I see it should be implemented. The existing implementation has too many questionable assumptions to be the one and the only that is shipped with DBAL.

greg0ire:
Ah I need to add an upgrade note for this.

UPGRADE.md
```diff
@@ -1,5 +1,10 @@
 # Upgrade to 2.11
 
+## Deprecated `MasterSlaveConnection`
+
+The naming is offensive, the implementation very questionable. Use regular
+connections and switch between them explicitely.
```
morozov, line 6: Should it be “explicitly”? I'd also avoid “switch”. Let's use “Use each endpoint connection explicitly”.

morozov (dismissed):

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
beberlei: We cannot detect if a query is truely only a read operation, it would require parsing the query string. 

ORM uses executeUpdate() or exec(), or directly prepare()/execute() to do writes. 

But executeQuery() is semantically about reads, so we can be sure that when this method is used, its a almost 100% read operation (not 100%, because users can just perform a write query with executeQuery even though it doesnt make sense).

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
beberlei: the query function can be used for any kind of SQL, including write operations. This cannot be detected without parsing the SQL statement, which is a rabbit hole. This is why for safety reason the connectoin assumes that when using $connection->query(), you might be performing a write operation.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
beberlei: The step from primary to replica is not really a common use-case, but what you describe could be an example. If the application does a single write early in the request and you can isolate that and don't need to care about replication lag with this write, then you could switch back to a replica. The idea would be:

```php
$connection->connect('primary');
$connection->executeUpdate("....");
$connection->connect("replica");
```

morozov: Why not just do `$primaryConnection->executeUpdate(); $replicaConnection->doWhatever();`? Why keep one connection as a shared state and switch between them?

morozov (approved):

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
albe: The answer is probably "convenience" (at the cost of some usefulness). Like always with such things ;)

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
lstrojny: Disclaimer: it’s more than 10 years that I did work on the original implementation.

The original design premise was "read from a replication, write to primary". This is fine until the first write, then we need to assume the replication to be out of sync, since we cannot control the specific replication algorithm from a database abstraction perspective.

I am nowadays of the opinion of a) this code should have never made it into Doctrine and b) what a seperate project (which may or may not live inside of Doctrine) could provide is building blocks for good multi-connection support. Typically the topologies of replicated or distributed setups are quite application specific, as are the specific configurations and therefore guarantees so providing flexible building blocks over a single implementation, albeit "conservative" in it’s choices, would be a better way.

morozov: Here you go, @beberlei. Thank you for the confirmation, @lstrojny.

lib/Doctrine/DBAL/Connections/PrimaryReplicaConnection.php
morozov: This is why the developer should be in control of which connection to use for a given statement. Not the abstraction layer.

beberlei:
Closing in favor of #4054

PR #4061, opened 2020-06-10
Deprecated platform-specific portability mode constants

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

The way that the Portability Connection API is structured is quite cryptic:

1. Besides the user-facing portability mode constants (e.g. `PORTABILITY_EMPTY_TO_NULL`), it exposes the platform-specific `PORTABILITY_ORACLE` constants.
2. The meaning of the platform constants is "which of the user-provided bits need to be left intact because this behavior indeed may need to be ported from the current platform".
3. These constants are expressed in the form of absolute values which require interpretation.
4. Why most of the platforms include all flags but `PORTABILITY_OTHERVENDORS` doesn't include `PORTABILITY_RTRIM` is a mystery.

#### Proposal
1. Remove the platform mode constants from the public API as internal to the portability layer.
2. Instead of describing what _may_ need to be ported on a specific platform, describe what _does not_ need to be ported which is much easier to understand (e.g. Oracle cannot tell the difference between an empty string and `NULL`).

### DIFF
--- a/lib/Doctrine/DBAL/Portability/Connection.php
+++ b/lib/Doctrine/DBAL/Portability/Connection.php
@@ -23,6 +23,10 @@ class Connection extends \Doctrine\DBAL\Connection
     public const PORTABILITY_EMPTY_TO_NULL = 4;
     public const PORTABILITY_FIX_CASE      = 8;
 
+    /**#@+
+     *
+     * @deprecated Will be removed as internal implementation details.
+     */
     public const PORTABILITY_DB2          = 13;
     public const PORTABILITY_ORACLE       = 9;
     public const PORTABILITY_POSTGRESQL   = 13;
@@ -31,6 +35,7 @@ class Connection extends \Doctrine\DBAL\Connection
     public const PORTABILITY_DRIZZLE      = 13;
     public const PORTABILITY_SQLANYWHERE  = 13;
     public const PORTABILITY_SQLSRV       = 13;
+    /**#@-*/
 
     /** @var int */
     private $portability = self::PORTABILITY_NONE;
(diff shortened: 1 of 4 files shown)


lib/Doctrine/DBAL/Portability/Connection.php
```diff
@@ -23,6 +23,10 @@ class Connection extends \Doctrine\DBAL\Connection
     public const PORTABILITY_EMPTY_TO_NULL = 4;
     public const PORTABILITY_FIX_CASE      = 8;
 
+    /**#@+
```
greg0ire, line 26: First time I see this notation… where can I read more about it?

lib/Doctrine/DBAL/Portability/Connection.php
```diff
@@ -47,23 +52,9 @@ public function connect()
...
-                } elseif ($this->getDatabasePlatform()->getName() === 'db2') {
-                    $params['portability'] &= self::PORTABILITY_DB2;
-                } elseif ($this->getDatabasePlatform()->getName() === 'mssql') {
-                    $params['portability'] &= self::PORTABILITY_SQLSRV;
-                } else {
-                    $params['portability'] &= self::PORTABILITY_OTHERVENDORS;
-                }
+                $platform = $this->getDatabasePlatform();
+
+                $params['portability'] = (new OptimizeFlags())($platform, $params['portability']);
```
greg0ire, line 57: When I see that variable I wrongly think it is going to be used more than once, let's get rid of it?

```suggestion
                $params['portability'] = (new OptimizeFlags())(
                    $this->getDatabasePlatform(),
                    $params['portability']
                );
```

lib/Doctrine/DBAL/Portability/Connection.php
morozov: See https://github.com/doctrine/dbal/pull/3795#issuecomment-568921124.

greg0ire: TIL

greg0ire (approved):

PR #4063, opened 2020-06-11

greg0ire:


lib/Doctrine/DBAL/Portability/Connection.php
```diff
@@ -0,0 +1,127 @@
+<?php
```
morozov, line 1: This file should be moved to `src/Portability/Connection.php`. Apparently, depending on the number of changes, Git doesn't always understand the rename.

morozov (changes requested):

lib/Doctrine/DBAL/Portability/Connection.php
greg0ire: Woops! Yes, I saw it managed to do it for some other files.

greg0ire: Oh and actually I applied the changes to the destination file, bug forgot to remove the source file :facepalm:

morozov: In this case, I usually accept the files in `lib/` and move them manually to `src/`. Then I have the actual diff and can fine-tune it. Shouldn't be happening that often though.

morozov (approved):

morozov:
Thanks, @greg0ire!

PR #4066, opened 2020-06-11
Fix DSN detection in the IBM DB2 driver

### LINKED ISSUE #4065: The IBM DB2 driver uses the dbname connectino parameter improperly
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.2

#### Summary

According to the [documentation](https://www.php.net/db2_connect),

> For a cataloged connection to a database, `database` represents the database alias in the DB2 client catalog.
> 
> For an uncataloged connection to a database, `database` represents a complete connection string in the following format:
> 
> `DATABASE=database;HOSTNAME=hostname;PORT=port;PROTOCOL=TCPIP;UID=username;PWD=password;`

In the `ibm_db2` extension, the detection of an uncataloged connection (i.e. a DSN) is [implemented](https://github.com/php/pecl-database-ibm_db2/blob/master/ibm_db2.c#L2803-L2804) as:
```c
/* If the string contains a =, use SQLDriverConnect */
if (strstr(database, "=") != NULL)
```

The `ibm_db2` driver in the DBAL uses a different approach which is incompatible with the underlying driver's expectations in some cases:

https://github.com/doctrine/dbal/blob/4074ae933de3bef5130bf99a20c797f6b8a4adc7/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Driver.php#L21

#### Current behaviour

1. If a connection specifies "localhost" as the host and an uncataloged database (i.e. just the database name), the connection will fail.
2. If a connection specifies a non-localhost host and a DSN, the connection will also fail.

The above makes it necessary to specify a full DSN in the PHPUnit configuration as well as duplicate the host, port, and user credentials:

https://github.com/doctrine/dbal/blob/194f710b130e45b70227ca366a2005a7c90a0f4a/tests/travis/ibm_db2.travis.xml#L13-L18

#### How to reproduce

```php
function test(array $params): void
{
    $conn = DriverManager::getConnection($params);

    try {
        echo $conn->fetchColumn('SELECT \'OK\' FROM sysibm.sysdummy1'), PHP_EOL;
    } catch (Exception $e) {
        echo $e->getMessage(), PHP_EOL;
    }
}

// Localhost + DSN
test([
    'driver'   => 'ibm_db2',
    'host'     => 'localhost',
    'dbname'   => 'UID=db2inst1;PWD=Passw0rd;DATABASE=doctrine',
]);

// Localhost + DB name
test([
    'driver'   => 'ibm_db2',
    'host'     => 'localhost',
    'user'     => 'db2inst1',
    'password' => 'Passw0rd',
    'dbname'   => 'doctrine',
]);

// Non-localhost + DSN
test([
    'driver'   => 'ibm_db2',
    'host'     => '127.0.1.1',
    'dbname'   => 'UID=db2inst1;PWD=Passw0rd;DATABASE=doctrine',
]);

// Non-localhost + DB name
test([
    'driver'   => 'ibm_db2',
    'host'     => '127.0.1.1',
    'user'     => 'db2inst1',
    'password' => 'Passw0rd',
    'dbname'   => 'doctrine',
]);

// OK
// [IBM][CLI Driver] SQL1531N  The connection failed because the name specified with the DSN connection string keyword could not be found in either the db2dsdriver.cfg configuration file or the db2cli.ini configuration file.  Data source name specified in the connection string: "DOCTRINE". SQLCODE=-1531
// [IBM][CLI Driver] SQL30082N  Security processing failed with reason "5" ("USERID MISSING").  SQLSTATE=08001 SQLCODE=-30082
// OK
````
#### Expected behaviour

All the connections above succeed.

### PR DESCRIPTION
**Pull request by morozov**
|      Q       |   A
|------------- | -----------
| Type         | bug, improvement
| BC Break     | no

Fixes https://github.com/doctrine/dbal/issues/4065.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Driver.php
+++ b/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Driver.php
@@ -4,6 +4,8 @@ namespace Doctrine\DBAL\Driver\IBMDB2;
 
 use Doctrine\DBAL\Driver\AbstractDB2Driver;
 
+use function strpos;
+
 /**
  * IBM DB2 Driver.
  */
@@ -18,7 +20,7 @@ class DB2Driver extends AbstractDB2Driver
             $params['protocol'] = 'TCPIP';
         }
 
-        if ($params['host'] !== 'localhost' && $params['host'] !== '127.0.0.1') {
+        if (strpos($params['dbname'], '=') === false) {
             // if the host isn't localhost, use extended connection params
             $params['dbname'] = 'DRIVER={IBM DB2 ODBC DRIVER}' .
                      ';DATABASE=' . $params['dbname'] .
--- a/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Driver.php
+++ b/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Driver.php
@@ -4,8 +4,6 @@ namespace Doctrine\DBAL\Driver\IBMDB2;
 
 use Doctrine\DBAL\Driver\AbstractDB2Driver;
 
-use function strpos;
-
 /**
  * IBM DB2 Driver.
  */
@@ -16,27 +14,16 @@ class DB2Driver extends AbstractDB2Driver
      */
     public function connect(array $params, $username = null, $password = null, array $driverOptions = [])
     {
-        if (! isset($params['protocol'])) {
-            $params['protocol'] = 'TCPIP';
-        }
+        $params['user']     = $username;
+        $params['password'] = $password;
+        $params['dbname']   = DataSourceName::fromConnectionParameters($params)->toString();
 
-        if (strpos($params['dbname'], '=') === false) {
-            // if the host isn't localhost, use extended connection params
-            $params['dbname'] = 'DRIVER={IBM DB2 ODBC DRIVER}' .
-                     ';DATABASE=' . $params['dbname'] .
-                     ';HOSTNAME=' . $params['host'] .
-                     ';PROTOCOL=' . $params['protocol'] .
-                     ';UID=' . $username .
-                     ';PWD=' . $password . ';';
-            if (isset($params['port'])) {
-                $params['dbname'] .= 'PORT=' . $params['port'];
-            }
-
-            $username = null;
-            $password = null;
-        }
-
-        return new DB2Connection($params, (string) $username, (string) $password, $driverOptions);
+        return new DB2Connection(
+            $params,
+            (string) $username,
+            (string) $password,
+            $driverOptions
+        );
     }
 
     /**
--- a/lib/Doctrine/DBAL/Driver/IBMDB2/DataSourceName.php
+++ b/lib/Doctrine/DBAL/Driver/IBMDB2/DataSourceName.php
@@ -1,0 +1,75 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Driver\IBMDB2;
+
+use function implode;
+use function sprintf;
+use function strpos;
+
+/**
+ * IBM DB2 DSN
+ */
+final class DataSourceName
+{
+    /** @var string */
+    private $string;
+
+    private function __construct(string $string)
+    {
+        $this->string = $string;
+    }
+
+    public function toString() : string
+    {
+        return $this->string;
+    }
+
+    /**
+     * Creates the object from an array representation
+     *
+     * @param array<string,mixed> $params
+     */
+    public static function fromArray(array $params) : self
+    {
+        $chunks = [];
+
+        foreach ($params as $key => $value) {
+            $chunks[] = sprintf('%s=%s', $key, $value);
+        }
+
+        return new self(implode(';', $chunks));
+    }
+
+    /**
+     * Creates the object from the given DBAL connection parameters.
+     *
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: lib/Doctrine/DBAL/Driver/IBMDB2/DB2Driver.php [hunk 1]
```diff
@@ -14,27 +14,16 @@ class DB2Driver extends AbstractDB2Driver
      */
     public function connect(array $params, $username = null, $password = null, array $driverOptions = [])
     {
-        if (! isset($params['protocol'])) {
-            $params['protocol'] = 'TCPIP';
-        }
+        $params['user']     = $username;
+        $params['password'] = $password;
```
morozov, line 18: User credentials and driver options are passed to the driver both via individual parameters and as part of `$params` which is awkward:
https://github.com/doctrine/dbal/blob/48625f1bc761f1f8f70fe6630a0f3a3204d9a6ff/src/Connection.php#L352-L356

This can be improved by simplifying the signature of `Driver::connect()` to just `Driver::connect($params)` (a separate issue is to be filed). Once it's done, the lines above will be removed.

lib/Doctrine/DBAL/Driver/IBMDB2/DataSourceName.php
```diff
@@ -0,0 +1,75 @@
...
+
+    public function toString() : string
+    {
+        return $this->string;
+    }
+
+    /**
+     * Creates the object from an array representation
+     *
+     * @param array<string,mixed> $params
```
greg0ire, line 32: Let's document the possible keys with an array shape?

lib/Doctrine/DBAL/Driver/IBMDB2/DataSourceName.php
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public static function fromArray(array $params) : self
+    {
+        $chunks = [];
+
+        foreach ($params as $key => $value) {
+            $chunks[] = sprintf('%s=%s', $key, $value);
```
greg0ire, line 39: Should we throw or maybe trigger a deprecation if an unknown key is passed? Or do we need the door open for more params than the one appearing from lines 59 to 64?

lib/Doctrine/DBAL/Driver/IBMDB2/DataSourceName.php
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        return new self(implode(';', $chunks));
+    }
+
+    /**
+     * Creates the object from the given DBAL connection parameters.
+     *
+     * @param mixed[] $params
```
greg0ire, line 48: An array shape would be nice here too.

lib/Doctrine/DBAL/Driver/IBMDB2/DataSourceName.php
morozov: This method doesn't limit the keys that can be used.

lib/Doctrine/DBAL/Driver/IBMDB2/DataSourceName.php
morozov: No. DBAL is not the source of truth about which parameters the underlying driver or the platform support.

lib/Doctrine/DBAL/Driver/IBMDB2/DataSourceName.php
morozov: By array shape, do you mean `array<k,v>` or `array{k1:v1,k2:v2,…}`? Although in https://github.com/doctrine/dbal/pull/4066#discussion_r439251232 it's already `array<string,mixed>`.

lib/Doctrine/DBAL/Driver/IBMDB2/DataSourceName.php
greg0ire: Array shapes do not limit the keys that can be used either

lib/Doctrine/DBAL/Driver/IBMDB2/DataSourceName.php
greg0ire: I mean the latter.

lib/Doctrine/DBAL/Driver/IBMDB2/DataSourceName.php
morozov: So what will be the behavior if the key being passed is not mandatory or optional?

In this specific case, it's still a map of any string to any value that can be cast to a string. But instead of defining it as an array shape, I'd rather have an object/structure. If we don't do that soon, we may define an array shape later when bumping the Psalm level and it demands it. This way, we'll make sure that our type annotations are valid.

greg0ire: > So what will be the behavior if the key being passed is not mandatory or optional?

Not the one I thought apparently… I could have sworn this was OK: https://psalm.dev/r/cd931bdda1

greg0ire (approved):

lib/Doctrine/DBAL/Driver/IBMDB2/DataSourceName.php
morozov: No worries. I wasn't sure about how Psalm works either, so now we're on the same page :+1:

PR #4067, opened 2020-06-12
Simplify test suite configuration

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

1. None of the supported drivers require all the connection parameters required by the test suite configuration. Most of them have default values for most of the parameters (e.g. `mysqli` can connect even w/o any parameters). Having to specify them all in a local development environment is unnecessarily tedious.
2. All so-called temporary connections look identical to the primary ones with the exception that the temporary one doesn't usually have a database name specified (creating a database is the primary purpose of the temporary connection). Instead of requiring the parameters to be specified explicitly, in the absence of the parameters, the test suite can fall back to the primary connection parameters and ignore the database.
3. There's no exact match between the test suite configuration parameter names and the connection parameters. E.g. the the `driver` is configured as `db_type`, `dbname` is configured as `db_name`. It would be nice to have a direct correspondence.

BTW, `tmp_host`, `tmp_port`, etc. also look redundant.

## Change summary
1. `(tmp)db_type` → `(tmp)db_driver`
2. `(tmp)db_username` → `(tmp)db_user`. This isn't nice but it corresponds to the connection parameter name.
3. `(tmp)db_name` → `(tmp)db_dbname`. This isn't nice either, but this is because we're not configuring a database (`db_`), we're configuring a test database _connection_. We may get rid of the `db_` prefix in parameter names or replace it with something else (e.g. `test_`).
4. Connection parameters are reordered logically: `driver`, `host`, `port`, `user`, `password`.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -22,24 +22,24 @@
     <php>
         <ini name="error_reporting" value="-1" />
 
-        <!-- "Real" test database -->
+        <!-- Test connection parameters -->
         <!-- Uncomment, otherwise SQLite runs
-        <var name="db_type" value="pdo_mysql"/>
+        <var name="db_driver" value="pdo_mysql"/>
         <var name="db_host" value="localhost" />
-        <var name="db_username" value="root" />
-        <var name="db_password" value="" />
-        <var name="db_name" value="doctrine_tests" />
         <var name="db_port" value="3306"/>
+        <var name="db_user" value="root" />
+        <var name="db_password" value="" />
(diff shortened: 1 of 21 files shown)


morozov:
The code coverage slightly dropped because the `pdo_pgsql` and `*sqlsrv` drivers no longer use the port to build the DSN. It's better to test building DSN later in a unit test rather than have it in the connection parameters to have unintended coverage.

greg0ire (approved):

PR #4068, opened 2020-06-12
Deprecate Driver::getDatabase()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no

The actual removal was already done in #3606 but it wasn't accompanied by deprecation.


greg0ire (approved):

greg0ire:
Is there a way to restart the AppVeyor build and only it?

morozov:
> Is there a way to restart the AppVeyor build and only it?

Yeah. I think you need to be a repo admin to do that.

PR #4077, opened 2020-06-13
Remove SQL Anywhere platform and driver

### LINKED ISSUE #4076: The SQLAnywhere driver and platform are effectively unmaintained
#### Summary

As of #2652 (3 years ago) there was effectively no progress made in improving the processes around SQLAnywhere support:

1. The activity around the driver (bug reports, fixes, etc.) is close to zero.
2. During active refactoring, the changes in the driver code are tested by integration tests. Given the integration purpose of the driver. The results of (partial) static analysis and unit tests are not sufficient.
3. There's no maintainer of the driver within the organization.
4. The PHP extension is not available on PECL, so we cannot properly statically analyze the driver code (https://github.com/JetBrains/phpstorm-stubs/pull/382).
5. SQLAnywhere is not supported by PHPStorm, so developing it would require using some additional software (no idea what that might be).
6. There are some unofficial Docker images on GitHub but I couldn't test them because of no GUI clients.

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

Fixes #4076.

The removal of the platform and the driver code should not be taken negatively. The DBAL APIs are extensible enough to maintain the removed components in a different project if needed.

Earlier, the support for the Azure platform was removed for the same reason (see #3602).


Ocramius (approved):

greg0ire (approved):

greg0ire:
Yay! Coverage increased!

beberlei:
@deeky666 let me know if you want to move the code to a new „contrib“ doctrine repo that you want to maintain.

beberlei (approved):

imankurpatel000:
@morozov @beberlei You mentioned that we can maintain the removed component in a separate repo. Is there is any example repo already maintained for some other database driver. If possible I would like to help support the SQL Anywhere driver and platform, If not in the main repo then at least on a separate repo. We are using Doctrine with SQL Anywhere right now and I would hate to be in a situation where doctrine DBAL 3.0 is released and we don't have any way to use it with SQL Anywhere.

For the following issue that you mentioned.

> The PHP extension is not available on PECL, so we cannot properly statically analyze the driver code (JetBrains/phpstorm-stubs#382).

We have enterprise support from SAP. Maybe we can push them to publish the PHP extension on PECL. Currently, the source code and extensions are published on the following link.
https://wiki.scn.sap.com/wiki/display/SQLANY/The+SAP+SQL+Anywhere+PHP+Module

If we can get the SAP to publish the PHP extension on PECL and then I think we can also help solve (JetBrains/phpstorm-stubs#382) which should give the SQLAnywhere support in PHPStorm.

morozov:
> Is there is any example repo already maintained for some other database driver.

I'm not aware of any. But I imagine that it's just a repository that contains all the driver and platform code. You can use the `driverClass` and `platform` [configuration](https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#driver) parameters when connecting to SQL Anywhere.

> Maybe we can push them to publish the PHP extension on PECL. Currently, the source code and extensions are published on the following link.

The problem was not only about the PHP extension and the API stubs. The major problem is that there's no Docker image for the database server that we could use on CI and test the code continuously.

imankurpatel000:
@morozov I got it. I understand that without a Docker image of SQL Anywhere database server it will be pretty complicated to setup CI flow for SQL Anywhere. I will ask SAP if they have any plans to publish a docker image. Even if SAP agrees to publish a docker image, I think it is probably going to take a while.

Meanwhile, I will try to set up a custom service container for SQL Anywhere in GitHub actions to make it work somehow. I don't know if that's even possible but If I manage to get the CI workflow set up for SQL Anywhere, do you think it is possible to add back support for SQL Anywhere?

morozov:
In theory, I'm not against having just another supported DB platform. In practice, it depends on all the factors that lead us to the decision to drop the support.

imankurpatel000:
@efarrar Is there anything you can do to help here. I know you were the one who was involved in getting the SQL Anywhere PHP modules released at the following link.
https://wiki.scn.sap.com/wiki/display/SQLANY/The+SAP+SQL+Anywhere+PHP+Module

I think there are 2 things that you can help us with.

1. Move the SQL Anywhere PHP module source code to Github, all the other SQL Anywhere client drivers are already in Github.
2. Publish the SQL Anywhere PHP extension on PECL instead of a custom link.

imankurpatel000:
@morozov I totally understand. I am just trying to see if I can do anything to help you guys add back support for SQL Anywhere and also help maintain it in future. 
Though if for whatever reason it doesn't work out then I guess we will just keep it in a separate repo as you mentioned.

PR #4078, opened 2020-06-14
Bump PHP requirement to 7.3 as of DBAL 2.11.0

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

Active support for PHP 7.2 by the PHP community [ended on November 30, 2019](https://www.php.net/supported-versions.php). Since the `2.11.x` release series of the DBAL will primary contain deprecations and the forward compatibility layer with the DBAL 3 that will not support PHP 7.3, there's no point in supporting it in `2.11.x` either.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,43 +34,6 @@ jobs:
       env: DB=sqlite
 
     - stage: Test
-      php: 7.2
-      env: DB=mysql.docker IMAGE=mysql:8.0
-    - stage: Test
-      php: 7.2
-      env: DB=mysqli.docker IMAGE=mysql:8.0
-    - stage: Test
-      php: 7.2
-      env: DB=mariadb.docker IMAGE=mariadb:10.3
-    - stage: Test
-      php: 7.2
-      env: DB=mariadb.mysqli.docker IMAGE=mariadb:10.3
-    - stage: Test
-      php: 7.2
-      env: DB=pgsql POSTGRESQL_VERSION=11.0
-      sudo: required
-      before_script:
-        - bash ./tests/travis/install-postgres-11.sh
-    - stage: Test
-      php: 7.2
-      env: DB=sqlite
-    - stage: Test
-      php: 7.2
-      env: DB=sqlsrv
-      sudo: required
-      before_script:
-        - bash ./tests/travis/install-sqlsrv-dependencies.sh
-        - bash ./tests/travis/install-mssql-sqlsrv.sh
-        - bash ./tests/travis/install-mssql.sh
-    - stage: Test
-      php: 7.2
-      env: DB=pdo_sqlsrv
-      sudo: required
-      before_script:
-        - bash ./tests/travis/install-sqlsrv-dependencies.sh
-        - bash ./tests/travis/install-mssql-pdo_sqlsrv.sh
-        - bash ./tests/travis/install-mssql.sh
-    - stage: Test
       php: 7.3
       env: DB=mysql.docker IMAGE=mysql:5.7
     - stage: Test
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         {"name": "Jonathan Wage", "email": "jonwage@gmail.com"}
     ],
     "require": {
-        "php": "^7.2",
+        "php": "^7.3",
         "ext-pdo": "*",
         "doctrine/cache": "^1.0",
         "doctrine/event-manager": "^1.0"
@@ -54,7 +54,7 @@
     "config": {
         "sort-packages": true,
         "platform": {
-            "php": "7.2.0"
+            "php": "7.3.0"
         }
     },
     "autoload": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "a5c06fdb9d602a1498b1eba6aa45f1d4",
+    "content-hash": "84bb64fae33632b7c7f562447fda9b7d",
     "packages": [
         {
             "name": "doctrine/cache",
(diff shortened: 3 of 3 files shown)


greg0ire (approved):
The changes look good.
When do we plan to release 2.11.0? It makes sense to me to keep unreleased versions in the green zone of https://www.php.net/supported-versions.php, but I'm wondering if we have a Doctrine-wide policy for this. It would be great if we had this kind of reference somewhere: https://github.com/sonata-project/dev-kit/blob/master/project/CONTRIBUTING.md#dependency-changes , but that would mean getting everyone to agree on something :sweat_smile: 

 If we merge this, we should probably also drop 7.2 in `doctrine/orm`.

morozov:
> When do we plan to release 2.11.0?

The plan is to release it with DBAL 3.0.0 at the time when PHP 8 beta 1 is released. If we can find the specific dates for that, we can update the milestone dates.

> I'm wondering if we have a Doctrine-wide policy for this

No, but it's an important discussion. I like how often PHPUnit releases their major versions and I believe they have their policy rules written down somewhere. They define the release cadence and the PHP version support.

greg0ire:
I updated the milestones dates  based on https://wiki.php.net/todo/php80

> Regardless of that, why does ORM need to support an unsupported PHP version? :wink: 

If I had to guess I would say it has to somehow do with Symfony: [the LTS version still provides support for 7.1](https://packagist.org/packages/symfony/symfony#v4.4.10), and will do so until November 2022, November 2023 if you count security updates, which is the reason why DoctrineBundle also does… and then it raises the question for more packages? Or maybe it has to somehow do with the fact that ORM has a non-dev dependency on `symfony/console`? I don't know…

For a moment I almost thought we did have a Doctrine-wide policy: https://www.doctrine-project.org/policies/releases.html I feel rick rolled.

morozov:
> I updated the milestones dates based on https://wiki.php.net/todo/php80

Thanks!

This is the [PHPUnit release process](https://github.com/sebastianbergmann/phpunit/wiki/Development-and-Release-Process) we could use for inspiration: major release every year, minor release every two months.

beberlei:
@greg0ire @morozov btw i am pushing on the deprecations thing mainly to get that ready for 2.11 as well so that you know

beberlei:
@morozov its really hard sticking to a schedule for releases when you can't really guarantee that you have time or motivation to work on something. I remember we tried fixed schedules with Doctrine a long time ago, and it quickly evaporated, it can even create pressure that demotivates even more.

beberlei:
I think supporting old PHP versions is fine, especially if you don't merge bugs back to old branches, then users might be forced to use the latest version of your library to get bugfixes, but they might have a 3 or 5 year upgrade schedule. In addition PHP maintenance schedule is essentially a "lie", because Linux distributions are willing to guarantee much longer maintenance.

greg0ire:
> I think supporting old PHP versions is fine, especially if you don't merge bugs back to old branches

An idea I had about that was to allow PRs on old branches that we currently label as unmaintained as long as they are cherry-picks of commits already merged in maintained branch, or if they only apply on said branches. IMO it could be a rare-enough occurrence. I agree with you that it would be preferable to do at least one of those things (supporting old PHP versions or allowing backports)

morozov:
> @morozov its really hard sticking to a schedule for releases when you can't really guarantee that you have time or motivation to work on something. I remember we tried fixed schedules with Doctrine a long time ago, and it quickly evaporated, it can even create pressure that demotivates even more.

Point taken. We can take it easy and just aim at releasing more often w/o any strict schedule. It will at least break the bad habit of trying to do everything before the release and this everything never gets done.

> I think supporting old PHP versions is fine

It's fine if we want to. Personally I don't. It's usually the corporate consumers who lag behind the PHP schedule and make money by using the results of our work. I'm fine with supporting a commercial LTS branch but it will have nothing to do with the regular releases.

> especially if you don't merge bugs back to old branches

Currently, we do. We expect bugfixes to be targeted at the lowest supported versions.

> In addition PHP maintenance schedule is essentially a "lie", because Linux distributions are willing to guarantee much longer maintenance.

It's also done for this corporate part of the ecosystem, not for OSS.

morozov:
> An idea I had about that was to allow PRs on old branches that we currently label as unmaintained as long as they are cherry-picks of commits already merged in maintained branch

As long as we accept commits to the old branch it means that we maintain it.

> I agree with you that it would be preferable to do at least one of those things (supporting old PHP versions or allowing backports)

I'd rather be on the “pushing the ecosystem forward” end of the spectrum rather on the LTS one.

beberlei:
@morozov i wasn't suggesting we should stay with 7.2 for DBAL 2.11 btw, I do agree more with the sentiment then previously that "why you want your dependencies on the latest version if you cannot upgrade PHP at the same pace". But for many of my libraries I also don't care about supporting 7.0 or 7.1 still for example, it is not usually that big of a burden. "maintaining old versions is fine" was more a general statement :)

A policy of cherry-picking critical fixes backwards is probably better as @greg0ire suggested, as this would also allow users to do most of the work themselves if they know we review, merge + release to older versions for tickets that are "Bug" and released in higher versions.

morozov:
> A policy of cherry-picking critical fixes backwards is probably better […]

I agree that technically it's easier to backport some bugfixes to the not-fully-supported branches than have them as the primary ones where those fixes initially land.

But before we proceed with this discussion, we'll need a definition of a critical issue. We already have a policy of backporting security fixes to the latest non-actively supported version but I don't remember a single case like this in the past 3 years.

As for the non-security related ones, how can an issue be critical if it does exist in a no longer supported release but wasn't noticed there in its entire lifetime? Most likely it's either an edge case or has a reasonable workaround.

greg0ire:
IMO we should allow backporting any fix people want to backport, as long as it is a backport. Also, and this is key, we should not do that backporting ourselves as soon as we merge a fix. Instead, we should let people open PRs for the fixes they so badly need to get. The only fixes we should maybe backport automatically are security fixes.

morozov:
> Also, and this is key, we should not do that backporting ourselves as soon as we merge a fix. Instead, we should let people open PRs for the fixes they so badly need to get.

And who then will maintain the build infrastructure for those builds and people? I.e. Travis may stop supporting the distro that the build is done on, we may stop using Scrutinized or other services in the newer branches, etc.

greg0ire:
I'd say that this would also be the burden of contributors. If they can't get a green build because there is no available distro, we just don't merge, they are the ones that have to find a solution. If there is no solution (e.g. no reasonable way to install such an old version of PHP on Travis or whatever), then maintenance just stops "naturally". The philosophy of this workflow is "we will allow you to get unstuck if you do all the work on our terms".

morozov:
This sounds fair. I think I could live with that.

PR #4079, opened 2020-06-14
Forward compatibility with PHPUnit 9.3

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | [job #9388.31](https://travis-ci.org/github/doctrine/dbal/jobs/698103923)

#### Summary

Some of the forward compatibility issues detected by the build failed on `3.0.x` can be fixed in `2.11.x` and up-merged. Once https://github.com/doctrine/dbal/pull/4078 is merged, we can proceed with updating PHPUnit in `2.11.x` as well.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Schema/SchemaTest.php
+++ b/tests/Doctrine/Tests/DBAL/Schema/SchemaTest.php
@@ -378,27 +378,19 @@ class SchemaTest extends TestCase
             ->method('acceptSchema')
             ->with($schema);
 
-        $visitor->expects($this->at(1))
+        $visitor->expects(self::exactly(2))
             ->method('acceptTable')
-            ->with($schema->getTable('baz'));
+            ->withConsecutive(
+                [$schema->getTable('baz')],
+                [$schema->getTable('bla.bloo')]
+            );
 
-        $visitor->expects($this->at(2))
-            ->method('acceptTable')
-            ->with($schema->getTable('bla.bloo'));
-
-        $visitor->expects($this->exactly(2))
-            ->method('acceptTable');
-
-        $visitor->expects($this->at(3))
+        $visitor->expects(self::exactly(2))
             ->method('acceptSequence')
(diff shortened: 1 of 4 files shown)


tests/Doctrine/Tests/DBAL/Schema/SchemaTest.php
```diff
@@ -386,9 +386,6 @@ public function testVisitsVisitor(): void
             ->method('acceptTable')
             ->with($schema->getTable('bla.bloo'));
 
-        $visitor->expects($this->exactly(2))
-            ->method('acceptTable');
-
```
greg0ire, line 391: Why is this change necessary?

morozov: Because combined with `$this->at()` above it will cause the following warning on PHPUnit 9.3:

> More than one invocation handler has been configured for Doctrine\DBAL\Schema\Visitor\Visitor::acceptTable()

At the time of making the change, I thought that the combination of `$this->at()` calls will also define the expected count of calls but it doesn't limit the upper boundary. E.g. if I add one more call to the mock in the code being tested, with this condition remove, the test will pass but it should fail.

Let me see if there's a better way to fix this.

morozov: Filed https://github.com/sebastianbergmann/phpunit/issues/4291.

greg0ire: Very interesting outcome!

greg0ire (approved):

tests/Doctrine/Tests/DBAL/Schema/SchemaTest.php
morozov: Yeah, it's always fun to test with dev-stability dependencies.

PR #4081, opened 2020-06-16
Simplify Driver::connect() signature

### LINKED ISSUE #3947: mysqli::$sqlstate is inaccessible on a failed/non-initialized connection since PHP 8
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.0.0-dev

#### Summary

1. https://bugs.php.net/bug.php?id=79469
2. https://travis-ci.org/github/doctrine/dbal/jobs/673655058:
   > Doctrine\DBAL\Tests\Driver\Mysqli\MysqliConnectionTest::testRestoresErrorHandlerOnException
   > Error: Property access is not allowed yet

### LINKED ISSUE #4069: The Driver::connect() method signature is redundant
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.2

User credentials and driver options are passed to the driver both via individual parameters and as part of `$params`:
https://github.com/doctrine/dbal/blob/48625f1bc761f1f8f70fe6630a0f3a3204d9a6ff/src/Connection.php#L352-L356

On the one hand, it's unclear to the implementors of `Driver::connect()`, whether the parameters should be taken from `$params` or from the individual parameters. On the other hand, the code components that call `connect()` need to mimic the behavior of the `Connection` class:

1. https://github.com/doctrine/dbal/blob/d6f232e4066b2d6ee6d75371c03657d3dd8db7cc/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php#L195-L198
2. https://github.com/doctrine/dbal/blob/3d8c57560d46f9cdd299e6e26de49f2ec25b349f/tests/Functional/Driver/AbstractDriverTest.php#L34-L37

<details>
 <summary>More examples:</summary>

3. https://github.com/doctrine/dbal/blob/48625f1bc761f1f8f70fe6630a0f3a3204d9a6ff/tests/Functional/Schema/SchemaManagerFunctionalTestCase.php#L120-L123
4. https://github.com/doctrine/dbal/blob/48625f1bc761f1f8f70fe6630a0f3a3204d9a6ff/tests/Functional/Schema/SqliteSchemaManagerTest.php#L53-L56
5. https://github.com/doctrine/dbal/blob/f88dc282080826528fe7326cb9a7c465aa3074d7/tests/Doctrine/Tests/DBAL/Functional/Driver/Mysqli/ConnectionTest.php#L63-L68
6. https://github.com/doctrine/dbal/blob/f88dc282080826528fe7326cb9a7c465aa3074d7/tests/Doctrine/Tests/DBAL/Driver/PDOPgSql/DriverTest.php#L96-L101
7. https://github.com/doctrine/dbal/blob/f88dc282080826528fe7326cb9a7c465aa3074d7/tests/Doctrine/Tests/DBAL/Functional/Driver/PDOSqlsrv/DriverTest.php#L47-L52
</details>

By being this verbose, this API doesn't solve any design problem. On the contrary, once a driver receives these parameters individually, it may have to pack them back or regroup otherwise before processing:

1. https://github.com/doctrine/dbal/blob/747b0420188e8775c71c0190abb62333eda52cf6/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Driver.php#L17-L19
2. https://github.com/doctrine/dbal/blob/4074ae933de3bef5130bf99a20c797f6b8a4adc7/lib/Doctrine/DBAL/Driver/SQLSrv/Driver.php#L34-L40
3. https://github.com/doctrine/dbal/blob/4074ae933de3bef5130bf99a20c797f6b8a4adc7/lib/Doctrine/DBAL/Driver/SQLAnywhere/Driver.php#L78-L84

Since `Driver::connect()` plays the role of an adapter between the DBAL `Connection::connect()` and the Driver `Connection::__construct()`, it shouldn't expect `Connection::connect()` to do any adaptation.

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

API design considerations:

1. `Driver::connect()` is an adapter between the DBAL `Connection::connect()` and the `Driver\Connection::__construct()` APIs. It should do all conversion of the parameters and not leak `array $params` into the connection.
2. `Driver\Connection::__construct()` is a tiny wrapper over the underlying driver's "connect" function. Its API should be as close to the "connect" function as possible but not too close to have to leak the underlying connection resource back to the driver like it happens in some PDO drivers:
   1. https://github.com/doctrine/dbal/blob/ae8c7208f64fa4647709df0232ea8514a4299169/src/Driver/PDOPgSql/Driver.php#L38
   2. https://github.com/doctrine/dbal/blob/ae8c7208f64fa4647709df0232ea8514a4299169/src/Driver/PDOSqlite/Driver.php#L43-L47
   In order to avoid the leakage but not to have too many parameters in MySQLi connection constructor either, a concept of connection initializers has been introduced. It can be thought of as an implementation of the Builder or Functional Options pattern.

**Additional changes**:
1. The custom error handler in MySQLi connection has been removed in favor of an error suppression operator. The error handler would yield the same results but required maintenance and a dedicated unit test that doesn't pass on Windows. Additionally, it masked improper usage of `$connection->sqlstate` in case of a connection failure (should help fixing #3947).
2. `$supportedDriverOptions` have been removed from the MySQLi connection. Whether the option is supported will be determined by the call to `mysqli_options()`. The previous approach led to improper error handling around `mysqli_options()` because an error was impossible to reproduce.

Although the code implementing secure MySQLi connection has been modified, it's not tested as it never was. It should be possible to set up a build job using [cyprien/mysql-tls](https://hub.docker.com/r/cyprien/mysql-tls/) on Travis later.

Fixes #4069

### DIFF
--- a/src/Driver/Mysqli/Driver.php
+++ b/src/Driver/Mysqli/Driver.php
@@ -3,15 +3,115 @@
 namespace Doctrine\DBAL\Driver\Mysqli;
 
 use Doctrine\DBAL\Driver\AbstractMySQLDriver;
+use Doctrine\DBAL\Driver\Mysqli\Initializer\Charset;
+use Doctrine\DBAL\Driver\Mysqli\Initializer\Options;
+use Doctrine\DBAL\Driver\Mysqli\Initializer\Secure;
+
+use function count;
 
 class Driver extends AbstractMySQLDriver
 {
     /**
      * {@inheritdoc}
      */
-    public function connect(array $params, $username = null, $password = null, array $driverOptions = [])
+    public function connect(array $params)
     {
-        return new MysqliConnection($params, (string) $username, (string) $password, $driverOptions);
+        if (! empty($params['persistent'])) {
+            if (! isset($params['host'])) {
+                throw HostRequired::forPersistentConnection();
+            }
+
+            $host = 'p:' . $params['host'];
+        } else {
+            $host = $params['host'] ?? null;
+        }
+
+        $flags = null;
+
+        $preInitializers = $postInitializers = [];
+
+        if (isset($params['driver_options'])) {
+            $driverOptions = $params['driver_options'];
+
+            if (isset($driverOptions[MysqliConnection::OPTION_FLAGS])) {
+                $flags = $driverOptions[MysqliConnection::OPTION_FLAGS];
+                unset($driverOptions[MysqliConnection::OPTION_FLAGS]);
+            }
+
+            $preInitializers = $this->withOptions($preInitializers, $driverOptions);
+        }
+
+        $preInitializers  = $this->withSecure($preInitializers, $params);
+        $postInitializers = $this->withCharset($postInitializers, $params);
+
+        return new MysqliConnection(
+            $host,
+            $params['user'] ?? null,
+            $params['password'] ?? null,
+            $params['dbname'] ?? null,
+            $params['port'] ?? null,
+            $params['unix_socket'] ?? null,
+            $flags,
+            $preInitializers,
+            $postInitializers
+        );
+    }
+
+    /**
+     * @param list<Initializer> $initializers
+     * @param array<int,mixed>  $options
+     *
+     * @return list<Initializer>
+     */
+    private function withOptions(array $initializers, array $options): array
+    {
+        if (count($options) !== 0) {
+            $initializers[] = new Options($options);
+        }
+
+        return $initializers;
+    }
+
+    /**
+     * @param list<Initializer>   $initializers
+     * @param array<string,mixed> $params
+     *
+     * @return list<Initializer>
+     */
+    private function withSecure(array $initializers, array $params): array
+    {
+        if (
+            isset($params['ssl_key']) ||
+            isset($params['ssl_cert']) ||
+            isset($params['ssl_ca']) ||
+            isset($params['ssl_capath']) ||
+            isset($params['ssl_cipher'])
+        ) {
+            $initializers[] = new Secure(
+                $params['ssl_key']    ?? null,
+                $params['ssl_cert']   ?? null,
+                $params['ssl_ca']     ?? null,
+                $params['ssl_capath'] ?? null,
+                $params['ssl_cipher'] ?? null
+            );
+        }
+
+        return $initializers;
+    }
+
+    /**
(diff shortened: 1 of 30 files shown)


src/Driver/Mysqli/Driver.php
```diff
@@ -3,15 +3,115 @@
...
+            $preInitializers,
+            $postInitializers
+        );
+    }
+
+    /**
+     * @param array<int,Initializer> $initializers
+     * @param array<int,mixed>       $options
+     *
+     * @return array<int,Initializer>
```
greg0ire, line 64: Should we use `list<>` for those?

src/Driver/Mysqli/MysqliConnection.php
```diff
@@ -40,53 +24,41 @@ class MysqliConnection implements PingableConnection, ServerInfoAwareConnection
     private $conn;
 
     /**
-     * @param mixed[] $params
-     * @param string  $username
-     * @param string  $password
-     * @param mixed[] $driverOptions
+     * @param Initializer[] $preInitializers
```
greg0ire, line 27: `iterable` seems to be enough, we don't really need those to be arrays

greg0ire (commented):
Are there parts of this that could be backported to 2.11 to make the upgrade path smoother?

morozov:
> Are there parts of this that could be backported to 2.11 to make the upgrade path smoother?

There cannot be FC-layer provided for the changes in the `Driver::connect()` signature. The very fact that the drivers start using the "user", "password" and "driver_options" keys of `$params` that they used to ignore, even as a fallback, is a BC break.

Do you have something else in mind?

greg0ire:
> Do you have something else in mind?

No I read this just before getting to work and did not have the time to figure something out. Not sure if that's a big deal though, I don't think `Driver` classes are directly used downstream, are they?

greg0ire (commented):
As a great man asked me recently,

> Does this removal need an upgrade note?

morozov:
> Not sure if that's a big deal though, I don't think `Driver` classes are directly used downstream, are they?

It's absolutely not a big deal. Unless someone maintains their own driver, they shouldn't be concerned about this change at all.

greg0ire (approved):

PR #4084, opened 2020-06-17
Allow using on PHP 7.3 with Composer 2

nicolas-grekas:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

In order to support Composer 2, `ocramius/package-versions` requires PHP 7.4.
This effectively means that using it in "require" forces ppl on PHP 7.3 to use Composer 1.

@seldaek created the `composer/package-versions-deprecated` to unlock this situation.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,9 +32,9 @@
     ],
     "require": {
         "php": "^7.3 || ^8.0",
+        "composer/package-versions-deprecated": "^1.8",
         "doctrine/cache": "^1.0",
-        "doctrine/event-manager": "^1.0",
-        "ocramius/package-versions": "^1.4"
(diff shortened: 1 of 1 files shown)


greg0ire:
Why not contribute this to 2.10.x?

morozov:
> Why not contribute this to 2.10.x? Not sure how urgent this is, but we expect to release 3.0.0 in a little less than 2 months

I believe 2.10.x doesn't use `ocramius/package-versions`.

greg0ire (approved):
Neither does 2.11.x, this is indeed the correct branch 👍

morozov:
Thanks, @nicolas-grekas!

PR #4085, opened 2020-06-18
The IBM DB2 driver Exception class must implement the DriverException interface

### LINKED ISSUE #4044: DB2Exception doesn't implement the DriverException interface
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.2

#### Summary

The fact that the class doesn't implement the interface makes it impossible to work with the driver in a portable way (i.e. to catching a specific interface and use the error code and SQLSTATE).

#### Current behaviour

https://github.com/doctrine/dbal/blob/fedac50334c7a0d7006fd621124c1b466b96ba78/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Exception.php#L5-L12

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

Fixes #4044

### DIFF
--- a/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
+++ b/lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
@@ -3,6 +3,9 @@
 namespace Doctrine\DBAL\Driver\IBMDB2;
 
 use Doctrine\DBAL\Driver\Connection;
+use Doctrine\DBAL\Driver\IBMDB2\Exception\ConnectionError;
+use Doctrine\DBAL\Driver\IBMDB2\Exception\ConnectionFailed;
+use Doctrine\DBAL\Driver\IBMDB2\Exception\PrepareFailed;
 use Doctrine\DBAL\Driver\ServerInfoAwareConnection;
 use Doctrine\DBAL\ParameterType;
 use stdClass;
@@ -21,7 +24,7 @@ use function db2_pconnect;
 use function db2_prepare;
 use function db2_rollback;
 use function db2_server_info;
-use function db2_stmt_errormsg;
+use function error_get_last;
 use function func_get_args;
 use function is_bool;
 
@@ -52,7 +55,7 @@ class DB2Connection implements Connection, ServerInfoAwareConnection
         }
 
         if ($conn === false) {
-            throw new DB2Exception(db2_conn_errormsg());
+            throw ConnectionFailed::new();
         }
 
         $this->conn = $conn;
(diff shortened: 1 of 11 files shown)


tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
```diff
@@ -255,13 +257,51 @@ public function testFetchAllWithMissingTypes(): void
             $this->markTestSkipped('mysqli and sqlsrv actually supports this');
         }
 
+        if (
+            $this->connection->getDriver() instanceof DB2Driver
+        ) {
+            $this->markTestSkipped(
+                'ibm_ibm2 may or may not report the error depending on the PHP version and the connection state'
```
morozov, line 264: See https://travis-ci.org/github/doctrine/dbal/jobs/699519713

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Connection.php
```diff
@@ -83,8 +86,9 @@ public function requiresQueryForServerVersion()
     public function prepare($sql)
     {
         $stmt = @db2_prepare($this->conn, $sql);
-        if (! $stmt) {
-            throw new DB2Exception(db2_stmt_errormsg());
+
+        if ($stmt === false) {
+            throw PrepareFailed::new(error_get_last()['message']);
```
morozov, line 91: In this case, `db2_stmt_errormsg()` and `db2_conn_errormsg()` return empty value but a warning is triggered.

tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
```diff
@@ -255,13 +257,51 @@ public function testFetchAllWithMissingTypes(): void
...
         $sql            = 'SELECT test_int, test_datetime FROM fetch_table WHERE test_int = ? AND test_datetime = ?';
 
         $this->expectException(DBALException::class);
 
-        $this->connection->fetchAll($sql, [1, $datetime]);
+        $fetch($this->connection, $sql, [1, $datetime]);
+    }
+
+    /**
+     * @return iterable<string,array<int,mixed>>
```
greg0ire, line 278:
```suggestion
     * @return iterable<string,array{0:callable}>
```

morozov: We're getting into a danger zone here. Techhically, it's something like `iterable<string,array{0:callable(string,<array<int|string,mixed>,<array<int|string,int|string|Type|null>>)}>`. Not sure how specific we want to be here.

greg0ire: Ugh yeah let's keep it as is.

greg0ire (approved):

PR #4089, opened 2020-06-19
Remove the CommentedType class from tests

morozov:
This class was removed by https://github.com/doctrine/dbal/commit/deb2f699828c8b669ca2e186bba4b6089037ffed but for some reason, it's still in the source tree.


greg0ire:
The only way I could find this commit in the git log is by using `--follow`… weird

greg0ire (approved):

PR #4093, opened 2020-06-19
Expect only DriverException from Result::fetch*() methods

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

The test can expect a `DriverException` after #4044 was fixed.

### DIFF
--- a/tests/Functional/StatementTest.php
+++ b/tests/Functional/StatementTest.php
@@ -2,13 +2,13 @@
 
 namespace Doctrine\DBAL\Tests\Functional;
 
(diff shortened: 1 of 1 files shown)


morozov:
@SenseException the failure is expected. When I was filing the PR, I thought that only the IBMDB2 driver misbehaves but OCI8 misbehaves as well. It needs to be fixed as well.

SenseException:
Is this a draft then?

greg0ire (approved):

PR #4094, opened 2020-06-20
Bump Psalm level to 5

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -95,3 +95,8 @@ parameters:
             message: '~Method Doctrine\\DBAL\\Driver\\PDOSqlsrv\\Connection\:\:lastInsertId\(\) should return string but returns string\|false\|null\.~'
             paths:
                 - %currentWorkingDirectory%/lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
+
+        -
+            message: '~Method Doctrine\\DBAL\\Portability\\Connection::prepare\(\) should return Doctrine\\DBAL\\Statement but returns Doctrine\\DBAL\\Portability\\Statement\.~'
+            paths:
+                - %currentWorkingDirectory%/lib/Doctrine/DBAL/Portability/Connection.php
--- a/tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL630Test.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL630Test.php
@@ -3,6 +3,7 @@
 namespace Doctrine\Tests\DBAL\Functional\Ticket;
 
 use Doctrine\DBAL\DBALException;
+use Doctrine\DBAL\Driver\PDOConnection;
 use Doctrine\DBAL\ParameterType;
 use Doctrine\Tests\DbalFunctionalTestCase;
 use PDO;
@@ -38,7 +39,7 @@ class DBAL630Test extends DbalFunctionalTestCase
     protected function tearDown() : void
     {
         if ($this->running) {
-            $this->connection->getWrappedConnection()->setAttribute(PDO::ATTR_EMULATE_PREPARES, false);
+            $this->getWrappedConnection()->setAttribute(PDO::ATTR_EMULATE_PREPARES, false);
         }
 
         parent::tearDown();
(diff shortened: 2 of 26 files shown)


tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL630Test.php
```diff
@@ -170,4 +171,12 @@ public static function booleanTypeConversionWithoutPdoTypeProvider() : iterable
...
         ];
     }
+
+    private function getWrappedConnection() : PDOConnection
+    {
+        $connection = $this->connection->getWrappedConnection();
+        self::assertInstanceOf(PDOConnection::class, $connection);
+
+        return $connection;
+    }
```
greg0ire, line 181: Have you considered redeclaring the property like this?

```php
/** @var PDOConnection */
protected $connection;
```

lib/Doctrine/DBAL/Driver/SQLAnywhere/SQLAnywhereConnection.php
```diff
@@ -137,7 +137,13 @@ public function lastInsertId($name = null)
             return sasql_insert_id($this->connection);
         }
 
-        return $this->query('SELECT ' . $name . '.CURRVAL')->fetchColumn();
+        $value = $this->query('SELECT ' . $name . '.CURRVAL')->fetchColumn();
+
+        if ($value === false) {
+            return '';
+        }
```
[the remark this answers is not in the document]
morozov: Yeah, this should be suppressed. I'm not sure if the “BC break” is the right term here because the new value corresponds to the documented return type while the old one didn't.

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL630Test.php
morozov: `protected $connection` is always the wrapper connection. It's the wrapped connection class that we need to make an assertion about.

tests/Doctrine/Tests/DBAL/Functional/Driver/IBMDB2/DB2StatementTest.php
```diff
@@ -29,6 +30,7 @@ protected function setUp() : void
     public function testExecutionErrorsAreNotSuppressed() : void
     {
         $stmt = $this->connection->prepare('SELECT * FROM SYSIBM.SYSDUMMY1 WHERE \'foo\' = ?');
+        self::assertInstanceOf(Statement::class, $stmt);
```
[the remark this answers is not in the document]
morozov: It actually is.

phpstan.neon.dist
```diff
@@ -95,3 +95,8 @@ parameters:
             message: '~Method Doctrine\\DBAL\\Driver\\PDOSqlsrv\\Connection\:\:lastInsertId\(\) should return string but returns string\|false\|null\.~'
             paths:
                 - %currentWorkingDirectory%/lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
+
+        -
+            message: '~Method Doctrine\\DBAL\\Portability\\Connection::prepare\(\) should return Doctrine\\DBAL\\Statement but returns Doctrine\\DBAL\\Portability\\Statement\.~'
```
morozov, line 100: This is caused by the issue earlier identified in https://github.com/doctrine/dbal/pull/4045:

> The portability layer is currently implemented on two levels of abstraction at the same time: most of its components implement driver-level interfaces but the `Connection` class extends the DBAL connection.

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL630Test.php
greg0ire: Oh right sorry!

greg0ire (approved):

greg0ire:
Thanks @morozov !

morozov:
The assertion was done to prove the method contract. It's expected to return a `DriverConnection` but the `$_conn` property is nullable. Thus we let the static analyzers know that by the end of the call to `connect()`, the property contains a connection.

Your use case violates the API.

PR #4097, opened 2020-06-22

morozov:


morozov:
@greg0ire, any ideas on how to fix the PHPStan issues? Looks like it cannot infer the value of the "wrapperClass" element because it's passed as part of the return value of `createPrimaryReadReplicaConnectionParams(): mixed[]`. I tried annotating its return type as following but it didn't help:
```php
/**
 * @return array{wrapperClass: class-string<T>}
 *
 * @template T of PrimaryReadReplicaConnection
 */
```

greg0ire (approved):

PR #4098, opened 2020-06-22
MSSQL create statements not respecting primary index name

joeltperry:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | 

#### Summary

Providing a name for primary index does nothing on MSSQL and then can cause unexpected results when referring back to the created index later.


SenseException:
@joelperry93 Can you please also provide a test that covers your fix and an example how to reproduce the issue?

joelperry93:
The tests that are broken cover the changes. I think the expected values of those tests are wrong

PR #4100, opened 2020-06-23
Deprecate inconsistently and ambiguously named driver-level classes

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

1. `Driver` and driver-level `Connection` and `Statement` classes have been renamed for consistency.
2. The `PDO`-related classes have been additionally moved under the `PDO` namespace.
3. The `Driver\DriverException` class has been renamed to `Driver\Exception`. On the one hand, the class name duplicated the name of its namespace; on the other, there's another `DriverException` class in the codebase.
4. The `Driver\AbstractDriverException` class has been renamed to `Driver\AbstractException` and marked internal.
5. Abstract driver-level exception classes have been deprecated in favor of the `Driver\Exception` interface and named exception classes.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
@@ -5,7 +5,19 @@ namespace Doctrine\DBAL\Driver;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Driver;
-use Doctrine\DBAL\Exception;
+use Doctrine\DBAL\Driver\DriverException as DeprecatedDriverException;
+use Doctrine\DBAL\Exception\ConnectionException;
+use Doctrine\DBAL\Exception\DeadlockException;
+use Doctrine\DBAL\Exception\DriverException;
+use Doctrine\DBAL\Exception\ForeignKeyConstraintViolationException;
+use Doctrine\DBAL\Exception\InvalidFieldNameException;
+use Doctrine\DBAL\Exception\LockWaitTimeoutException;
+use Doctrine\DBAL\Exception\NonUniqueFieldNameException;
+use Doctrine\DBAL\Exception\NotNullConstraintViolationException;
+use Doctrine\DBAL\Exception\SyntaxErrorException;
+use Doctrine\DBAL\Exception\TableExistsException;
+use Doctrine\DBAL\Exception\TableNotFoundException;
+use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
 use Doctrine\DBAL\Platforms\MariaDb1027Platform;
 use Doctrine\DBAL\Platforms\MySQL57Platform;
 use Doctrine\DBAL\Platforms\MySQL80Platform;
@@ -19,7 +31,7 @@ use function stripos;
 use function version_compare;
 
 /**
- * Abstract base implementation of the {@link Doctrine\DBAL\Driver} interface for MySQL based drivers.
+ * Abstract base implementation of the {@link Driver} interface for MySQL based drivers.
  */
 abstract class AbstractMySQLDriver implements Driver, ExceptionConverterDriver, VersionAwarePlatformDriver
 {
@@ -29,44 +41,44 @@ abstract class AbstractMySQLDriver implements Driver, ExceptionConverterDriver, 
      * @link https://dev.mysql.com/doc/refman/8.0/en/client-error-reference.html
      * @link https://dev.mysql.com/doc/refman/8.0/en/server-error-reference.html
      */
-    public function convertException($message, DriverException $exception)
+    public function convertException($message, DeprecatedDriverException $exception)
     {
         switch ($exception->getErrorCode()) {
             case '1213':
-                return new Exception\DeadlockException($message, $exception);
+                return new DeadlockException($message, $exception);
 
             case '1205':
-                return new Exception\LockWaitTimeoutException($message, $exception);
+                return new LockWaitTimeoutException($message, $exception);
 
             case '1050':
-                return new Exception\TableExistsException($message, $exception);
+                return new TableExistsException($message, $exception);
 
             case '1051':
             case '1146':
-                return new Exception\TableNotFoundException($message, $exception);
+                return new TableNotFoundException($message, $exception);
 
             case '1216':
             case '1217':
             case '1451':
             case '1452':
             case '1701':
-                return new Exception\ForeignKeyConstraintViolationException($message, $exception);
+                return new ForeignKeyConstraintViolationException($message, $exception);
 
             case '1062':
             case '1557':
             case '1569':
             case '1586':
-                return new Exception\UniqueConstraintViolationException($message, $exception);
+                return new UniqueConstraintViolationException($message, $exception);
 
             case '1054':
             case '1166':
             case '1611':
-                return new Exception\InvalidFieldNameException($message, $exception);
+                return new InvalidFieldNameException($message, $exception);
 
             case '1052':
             case '1060':
             case '1110':
-                return new Exception\NonUniqueFieldNameException($message, $exception);
+                return new NonUniqueFieldNameException($message, $exception);
 
             case '1064':
             case '1149':
@@ -80,7 +92,7 @@ abstract class AbstractMySQLDriver implements Driver, ExceptionConverterDriver, 
             case '1541':
             case '1554':
             case '1626':
-                return new Exception\SyntaxErrorException($message, $exception);
+                return new SyntaxErrorException($message, $exception);
 
             case '1044':
             case '1045':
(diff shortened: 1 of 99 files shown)


greg0ire:
:thinking: how are any of these items not BC-breaks?

morozov:
>  how are any of these items not BC-breaks? Do we consider all these classes and interfaces internal?

No, the original classes and interfaces remain in place but they are deprecated and extend their new versions. It's a good time to do the rename before most of these classes (or all by design) become `final`.

lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
```diff
@@ -29,44 +40,44 @@ abstract class AbstractMySQLDriver implements Driver, ExceptionConverterDriver,
      * @link https://dev.mysql.com/doc/refman/8.0/en/client-error-reference.html
      * @link https://dev.mysql.com/doc/refman/8.0/en/server-error-reference.html
      */
-    public function convertException($message, DriverException $exception)
+    public function convertException($message, Exception $exception)
```
greg0ire, line 43: Small BC-break here: classes extending this one will narrow the type, which breaks the LSP. I'm not saying it has to be a blocker, but maybe in these 103 files there are similar BC-breaks that are more serious, so I raise this in case you did not have that in mind.

Slightly off-topic, have we ever considered using roave's BC-checker on this package?

morozov: > Small BC-break here: classes extending this one will narrow the type, which breaks the LSP. I'm not saying it has to be a blocker, but maybe in these 103 files there are similar BC-breaks that are more serious, so I raise this in case you did not have that in mind.

This is a valid point, I didn't have that in mind. To be absolutely on the safe side, we may only introduce the new classes but not change any method signatures that accept them (we could change the return types but we probably don't have many of them in `2.11.x`).

> Slightly off-topic, have we ever considered using roave's BC-checker on this package?

It might be a good time to give it a try.

greg0ire: Is there still a point in introducing the new classes? Can they be used anywhere?

[I have a solution to that kind of issue](https://dev.to/greg0ire/how-to-deprecate-a-type-in-php-48cf), and we went with it when migrating the `Doctrine\Common\Persistence` namespace to just `Doctrine\Persistence`, and it was painful (Andreas wrote several blog posts about it). We vowed never to do that again, so I'm not sure how this is going to pan out. Maybe you can find a subset of this PR that consists in classes never used in method signatures that can be overriden outside of the DBAL?

This is why I tend to `final` everything…

morozov: Most of these classes can be finalized (https://github.com/doctrine/dbal/pull/3590, https://github.com/doctrine/dbal/issues/3818) but not all of them. We indeed can/should mark all the new ones that can be marked final right now instead of waiting.

morozov:
This is what the output of the BC tool looks like:
```
$ roave-backward-compatibility-check --from=upstream/2.11.x
[BC] SKIPPED: Roave\BetterReflection\Reflection\ReflectionClass "Symfony\Component\Console\Helper\Helper" could not be found in the located source
...
[BC] REMOVED: These ancestors of Doctrine\DBAL\Driver\IBMDB2\Exception\ConnectionFailed have been removed: ["Doctrine\\DBAL\\Driver\\IBMDB2\\DB2Exception","Doctrine\\DBAL\\Driver\\AbstractDriverException"]
[BC] CHANGED: Doctrine\DBAL\Driver\IBMDB2\Exception\ConnectionFailed was marked "@internal"
[BC] REMOVED: These ancestors of Doctrine\DBAL\Driver\IBMDB2\Exception\ConnectionError have been removed: ["Doctrine\\DBAL\\Driver\\IBMDB2\\DB2Exception","Doctrine\\DBAL\\Driver\\AbstractDriverException"]
[BC] CHANGED: Doctrine\DBAL\Driver\IBMDB2\Exception\ConnectionError was marked "@internal"
[BC] REMOVED: These ancestors of Doctrine\DBAL\Driver\IBMDB2\Exception\StatementError have been removed: ["Doctrine\\DBAL\\Driver\\IBMDB2\\DB2Exception","Doctrine\\DBAL\\Driver\\AbstractDriverException"]
[BC] CHANGED: Doctrine\DBAL\Driver\IBMDB2\Exception\StatementError was marked "@internal"
[BC] REMOVED: These ancestors of Doctrine\DBAL\Driver\IBMDB2\Exception\PrepareFailed have been removed: ["Doctrine\\DBAL\\Driver\\IBMDB2\\DB2Exception","Doctrine\\DBAL\\Driver\\AbstractDriverException"]
[BC] CHANGED: Doctrine\DBAL\Driver\IBMDB2\Exception\PrepareFailed was marked "@internal"
```
All the IBMDB2-related errors can be ignored since these classes were introduced in `2.11.x` (https://github.com/doctrine/dbal/pull/4085).

greg0ire (approved):

lib/Doctrine/DBAL/Driver/AbstractSQLServerDriver/Exception/PortWithoutHost.php
```diff
@@ -2,16 +2,16 @@
...
-use Doctrine\DBAL\Driver\AbstractDriverException;
+use Doctrine\DBAL\Driver\AbstractException;
 
 /**
  * @internal
  *
  * @psalm-immutable
  */
-final class PortWithoutHost extends AbstractDriverException
+final class PortWithoutHost extends AbstractException
```
stof, line 14: this is a BC break, because catching `AbstractDriverException` will not catch this exception anymore (remember that *using* the exception classes is not only done in instantiations but also in `catch` clauses)

greg0ire: The class is marked as `internal` though.

morozov: @greg0ire is right. Consumers should be only catching `Driver\DriverException`.

greg0ire: Ah but AbstractDriverException itself is not marked as `internal`, so they might be, might they not?

morozov: I see. Let's revert this line. It looks like the only change that needs to be reverted. The rest of the exceptions changed in a similar way were introduced in 2.11.x.

lib/Doctrine/DBAL/Driver/PDOException.php
```diff
@@ -3,7 +3,7 @@
 namespace Doctrine\DBAL\Driver;
 
 /**
- * Tiny wrapper for PDOException instances to implement the {@link DriverException} interface.
+ * @deprecated Use {@link Exception} instead
```
stof, line 6: this is missing a use statement to make the `Exception` reference valid.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Connection.php
```diff
@@ -14,7 +14,7 @@
 /**
  * Sqlsrv Connection implementation.
  */
-class Connection extends PDOConnection
+class Connection extends BaseStatement
```
stof, line 17: This is a BaseConnection, not a BaseStatement

lib/Doctrine/DBAL/Driver/AbstractSQLServerDriver/Exception/PortWithoutHost.php
morozov: So this might be something that the BackwardCompatibilityCheck doesn't yet detect: even if the class is internal if the new version is not a subtype of the old version, it's a BC break. Even if it's not an exception.

stof: well, for internal classes which are not an exception, that's fine (as it is not covered by the BC promise). For exception, that's a bit more complex: due to the catching semantic, even an internal exception class must still stay a subtype of all its non-internal types of the old version.

morozov: How is catching an exception different from accepting an argument of a given type by a function?

stof: hmm, indeed. So caring about non-internal parents of internal classes indeed applies to other cases too.

morozov: Reported as https://github.com/Roave/BackwardCompatibilityCheck/issues/256.

PR #4103, opened 2020-06-24
Invalid queries are generated

4c0n:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes

#### Summary

When changing a column type the platform options are not considered, in version 2.9 for MySQL/MariaDB the `charset` key was not passed as platform option in a column object. Since 2.10 the `charset` key is passed and it is causing generation of invalid queries.

In this scenario the query generated to alter the table is:
```ALTER TABLE to_be_altered CHANGE some_column_name some_column_name INT CHARACTER SET utf8 DEFAULT NULL COLLATE `utf8_unicode_ci` ```

Obviously the charset should not be provided/processed for integer type columns.

I'd like to fix this, but I'm a little lost on where to do that. For example we could check the platform options when setting a type to the column and remove the irrelevant ones, but that would require specific logic for specific platform options in the column class.

Maybe introduce a `ColumnTypeModifier` class and have the logic in a separate place? Or am I missing something?

### DIFF
--- a/tests/Doctrine/Tests/DBAL/BugScenarioTest.php
+++ b/tests/Doctrine/Tests/DBAL/BugScenarioTest.php
@@ -7,6 +7,7 @@ namespace Doctrine\Tests\DBAL;
 use Doctrine\DBAL\Schema\Comparator;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\DBAL\Types\Types;
+use Doctrine\Tests\DbalFunctionalTestCase;
 
 final class BugScenarioTest extends DbalFunctionalTestCase
 {
--- a/tests/Doctrine/Tests/DBAL/BugScenarioTest.php
+++ b/tests/Doctrine/Tests/DBAL/BugScenarioTest.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace Doctrine\Tests;
+namespace Doctrine\Tests\DBAL;
 
 use Doctrine\DBAL\Schema\Comparator;
 use Doctrine\DBAL\Types\Type;
--- a/tests/Doctrine/Tests/DBAL/BugScenarioTest.php
+++ b/tests/Doctrine/Tests/DBAL/BugScenarioTest.php
@@ -40,5 +40,7 @@ final class BugScenarioTest extends DbalFunctionalTestCase
         $tableDiff  = $comparator->diffTable($originalTable, $modifiedTable);
 
         $schemaManager->alterTable($tableDiff);
+
+        $this->assertTrue(true);
     }
 }
--- a/tests/Doctrine/Tests/BugScenarioTest.php
+++ b/tests/Doctrine/Tests/BugScenarioTest.php
@@ -1,0 +1,43 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests;
+
+use Doctrine\DBAL\Schema\Comparator;
+use Doctrine\DBAL\Types\Type;
+use Doctrine\DBAL\Types\Types;
+
+final class BugScenarioTest extends DbalFunctionalTestCase
+{
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        $schemaManager = $this->connection->getSchemaManager();
+        $schema = $schemaManager->createSchema();
+
+        $originalTable = $schema->createTable('to_be_altered');
+        $column = $originalTable->addColumn('some_column_name', Types::TEXT);
+        $column->setNotnull(false);
+
+        $schemaManager->createTable($originalTable);
+    }
+
+    public function testInvalidQueryIsGenerated(): void
+    {
+        $schemaManager = $this->connection->getSchemaManager();
+        $schema = $schemaManager->createSchema();
+
+        $originalTable = $schema->getTable('to_be_altered');
+        $modifiedTable = clone $originalTable;
+
+        $column = $modifiedTable->getColumn('some_column_name');
+        $column->setType(Type::getType(Types::INTEGER));
+
+        $comparator = new Comparator();
+        $tableDiff = $comparator->diffTable($originalTable, $modifiedTable);
+
+        $schemaManager->alterTable($tableDiff);
+    }
+}
--- a/tests/Doctrine/Tests/BugScenarioTest.php
+++ b/tests/Doctrine/Tests/BugScenarioTest.php
@@ -10,24 +10,24 @@ use Doctrine\DBAL\Types\Types;
 
 final class BugScenarioTest extends DbalFunctionalTestCase
 {
-    public function setUp(): void
+    public function setUp() : void
     {
         parent::setUp();
 
         $schemaManager = $this->connection->getSchemaManager();
-        $schema = $schemaManager->createSchema();
+        $schema        = $schemaManager->createSchema();
 
         $originalTable = $schema->createTable('to_be_altered');
-        $column = $originalTable->addColumn('some_column_name', Types::TEXT);
+        $column        = $originalTable->addColumn('some_column_name', Types::TEXT);
         $column->setNotnull(false);
 
         $schemaManager->createTable($originalTable);
     }
 
-    public function testInvalidQueryIsGenerated(): void
+    public function testInvalidQueryIsGenerated() : void
     {
         $schemaManager = $this->connection->getSchemaManager();
-        $schema = $schemaManager->createSchema();
+        $schema        = $schemaManager->createSchema();
 
         $originalTable = $schema->getTable('to_be_altered');
         $modifiedTable = clone $originalTable;
@@ -36,7 +36,7 @@ final class BugScenarioTest extends DbalFunctionalTestCase
         $column->setType(Type::getType(Types::INTEGER));
 
         $comparator = new Comparator();
-        $tableDiff = $comparator->diffTable($originalTable, $modifiedTable);
+        $tableDiff  = $comparator->diffTable($originalTable, $modifiedTable);
 
         $schemaManager->alterTable($tableDiff);
     }

### CODE REVIEW

**Comment by greg0ire** [comment]
> Since 2.10 the charset key is passed

Can you please pinpoint the commit that does this change? That way we may involve relevant people in the discussion.

**Comment by 4c0n** [comment]
@greg0ire that would be this commit: https://github.com/doctrine/dbal/commit/73f38e722421fed5f00ff3fba76c80cf1a7545b4

**Comment by greg0ire** [comment]
Oh wow it's quite old, it's strange that this issue wasn't reported sooner… @AlterTable can you please have a look at this?

**Comment by 4c0n** [comment]
@greg0ire Thank you for suggesting that @AlterTable should have a look at it, however I'm not sure it will really help.

The thing is, when you request the column from an existing table it is totally correct that the platform option is present.

However when we change the type it becomes invalid. Not necessarily related to that change as you can see the build fails for pretty much every rdbms except for SQLite. (Different reasons though).

My point is that even is this is a bug that came to be because of that change, we should focus on fixing the problem at hand as the change is correct in my opinion. The fact that it worked before was luck on my end I suppose.
Regardless of that, we still have the case of not being able to intuitively changing the column type in this way (from text to int).

Judging by the builds there seem to be different reasons for it to fail depending on the rdbms. So let's focus on them one by one?

For example using MySQL/MariaDB the query that is generated is invalid (something we would probably like to prevent dbal from doing?)

The simple solution would be to add some form of mapping to the column class to determine how to update the platform options when the type is modified. Which I'm more than willing to implement, however I'd hate for me to implement that and then finding out during a code review on the PR, that it is not the way you guys (maintainers) would like to see the problem fixed.

So if somebody could just free a tiny amount of time to think this over, than it could be fixed this week.

**Comment by greg0ire** [comment]
I don't the DBAL well enough to advise you with architecture concerns, so I request a review from someone that should be able to do that :slightly_smiling_face:

File: tests/Doctrine/Tests/DBAL/BugScenarioTest.php [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+    public function testInvalidQueryIsGenerated() : void
+    {
+        $schemaManager = $this->connection->getSchemaManager();
+        $schema        = $schemaManager->createSchema();
+
+        $originalTable = $schema->getTable('to_be_altered');
+        $modifiedTable = clone $originalTable;
+
+        $column = $modifiedTable->getColumn('some_column_name');
+        $column->setType(Type::getType(Types::INTEGER));
```
morozov, line 37: IMO this is where the existing design falls apart. It's like `$car->setType(Book::class)`.  It works in some cases and doesn't in others. There's just one more case when it doesn't work.

> I'd like to fix this, but I'm a little lost on where to do that. For example we could check the platform options when setting a type to the column and remove the irrelevant ones, but that would require specific logic for specific platform options in the column class.

Here's another example:
```php
$column = $table->addColumn('test', Types::TEXT, ['length' => 128]);

$column->setType(Type::getType(Types::INTEGER));
```
What's the expected column type and length given that both types support the "length" option?

4c0n: @morozov I totally agree, it would nicer to have another layer making `Column` just an abstraction. A column should be a type, not specify one.

So we could have an `IntegerColumn` or a `TextColumn` and have to transform it, if we want to change it using a `IntegerToTextColumnTransformer` of `TextToIntegerColumnTransformer` (Maybe all can available through a `ColumnTypeModifier` facade/helper for convience). Do you like that idea?

I guess a change like that would require a major version release though.

morozov: > A column should be a type, not specify one.

I'd start with documenting the existing architectural layers, what properties they have, what problems they solve and how they should be regrouped in order to solve this problem as well. E.g.:

1. We have columns which can have types and attributes that are currently applied to any type. Columns are used for DDL.
2. We have types which are used to defined columns and to bind parameters to prepared statements.
3. Some column attributes seem to belong to a type in the DDL context (e.g. string length). But others belong to a column _of_ a given type (e.g. charset that is not part of the type but is a property of a column of type string).

I think it's too premature to discuss the solution of the problem w/o the problem statement and the existing design laid out.

4c0n: Alright I think you described the problem pretty well already, but I can dig into how `Column` is used exactly within `DBAL` and create some sort of report/proposal. It will take me some time though.

4c0n:
@greg0ire @morozov 

**How is Column used with DBAL?**
_Instantiation_:
AbstractSchemaManager and its derived classes create Column objects when listTableColumns() is called. This happens in the _getPortableTableColumnDefinition() method.

Table also creates Column objects in the addColumn() method.
Of course a Table holds a collection of Columns.

_Events_:
There are a number of events argument classes that hold one or more Column objects. They just provide encapsulation and don't really use the Column themselves.

_Platforms_:
AbstractPlatform and its derived classes use Column objects to map to and from column definitions and SQL statements.

_Validator_:
The reserved keywords validator uses Column objects to be able to add a good validation message.

_Outside of DBAL_:
We know that the in the DDL for entities for the ORM uses the same type of properties in its own Column class and is likely using this class to convert that data to this Column class and uses it to do what the name of this package implies: "DataBase Abstraction Layer".
Who knows really, it is a standalone package that could be used by anyone in whatever way they please.


**Which Types exist?**
- array
- bigint
- binary
- blob
- boolean
- date
- date_immutable
- dateinterval
- datetime
- datetime_immutable
- datetimetz
- datetimetz_immutable
- decimal
- float
- guid
- integer
- json
- object
- simple_array
- smallint
- string
- text
- time
- time_immutable
- json_array (deprecated)

**Which properties relate to which types?**
- _type:
  - All

- _length:
  - string
  - binary
  - guid
  - array
  - blob
  - text
  - simple_array
  - object

- _precision:
  - decimal

- _scale:
  - decimal

- _unsigned:
  - integer

- _fixed:
  - string
  - binary
  - guid

- _notnull:
  - All

- _default:
  - All

- _autoincrement:
  - numeric types

**Which platform options exist for which platform?**
- collation:
  - Drizzle
  - MySql
  - PostgreSql
  - Sqlite
  - SQLServer

- charset:
  - MySql

- jsonb:
  - PostgreSql

Charset and collation both relate to text types, since they both specify a character set and character set based format. Both are not necessarily for columns only. In the case of MySql for example, it is possible to specify this on schema, table and column level.


**Which custom schema options relate to which types?**
I can't find any real usage of this method within DBAL.


Now back to the original story, when using the MySql platform, when using DBAL to retrieve the table information and then modify the type of a column from text to integer type, the charset is still in the platform options and is causing the generation of invalid queries.

Currently we can state that Column is a very generic object, that in this context contains quite a bit of properties that are not  related to it and in some cases or not even optional, in the end the properties get merged with the platform options and custom schema options into one big array. The array is then used by the platform layer to produce statements.

There is no specification of that array, nor is a structure enforced for that array.  It might be common in PHP, but I still think it's a bit of code smell (might be personal). I also don't like that if there is a collision in keys, that the property value is for example overwritten by the value of the platform option or custom schema option (whatever the custom schema options may be meant for, I couldn't find out).

My suggestion is define a Column layer, so that only the properties that apply to that specific column are available, using abstraction we can have common parameters. It is a generic solution to a specific problem, as the error is in this form cannot ever occur again. It also solves my original problem, as the type in that case can no longer be modified. The column just has to removed from the table and a new one with the same name has to be added. To make that more convenient we could introduce a class that can "transform" column objects (basically by replacing them).

morozov:
> There is no specification of that array, nor is a structure enforced for that array. It might be common in PHP, but I still think it's a bit of code smell (might be personal).

It is a problem (absolutely objective). The Column object doesn't properly protect its invariants and allows to put itself in an invalid state (type=int, charset=xyz).

> My suggestion is define a Column layer, so that only the properties that apply to that specific column are available, using abstraction we can have common parameters.

The way I see it is we define a column type. A column can have a name and type. Or maybe the column doesn't even need to have a name but the table will have columns indexed by name. In any event, we need to define the hierarchy of column types, the attributes they may hold, and the ways to convert one type to another.

> The column just has to removed from the table and a new one with the same name has to be added.

I believe it doesn't directly translate to `DROP COLUMN` and `ADD COLUMN`. If the column and the type are two different interfaces, we may keep the `$column->setType()` interface but instead of being just a setter, it will actually replace one type with the other and construct the new one based on the old one.

> To make that more convenient we could introduce a class that can "transform" column objects (basically by replacing them).

Let's define types and their attributes and relationships and then an API that converts types. Should it only convert one type to another or also take care of changing attributes of the same type? Both will eventually result in `ALTER TABLE MODIFY …`.

4c0n:
Okay first things first, I made a class diagram:
[Class Diagram](https://svgshare.com/s/MgV)

We kind of have to keep the name property, as sometimes DBAL gives us a collection of columns that is not a table. Changing that would be good, but beyond the scope of this problem. 
About the removing and adding columns from the `Table` class. It shouldn't translate to `DROP` and `MODIFY` directly as that could result in preventable data loss. In the end the Comparator just generates sort of diff and uses it to generate a unit of work.  Anyway since we're dealing with types now and not columns it's a different story all together. So let's forget about that for now.

So this way the (existing) type objects are turned into value objects that hold the appropriate properties. If we then want to modify the column type we can simply call `setType()` on the column with a new type that has all the options/arguments that we need.

morozov:
I'm not exactly sure how to interpret this diagram. Specifically,

1. Does it represent the existing state of things or the end goal? Some of the types do currently exist but some don't.
2. The purpose of some types is unclear: CollatedGuidType, NonDefaultCharsetType — what are they?

What we may benefit from is defining layers: some types may be reproduced just on a DB (e.g. a fixed-length string) but some can be only defined at the application level and built on top of the DB ones (e.g. the SimpleArrayType is probably stored in varchar). Some application-level types share the same underlying DB type and differ only in serialization/deserialization (e.g. date and immutable date).

To me, the nullability, the default value and autoincrement are not properties of the type. They are properties of the column.

Some types like Drizzle are no longer supported. The big/small integer types are purely MySQL dialects. IMO the more fundamental properties of a numeric type are scale and precision. Numbers of a zero scale are integers.

Given the class hierarchy and the interfaces on the diagram, how should the type conversion work?

4c0n:
> I'm not exactly sure how to interpret this diagram. Specifically,

It is a class diagram representing the types laid out by the properties that are specific to that type. So it would be a diagram of the "desired" situation.

> Does it represent the existing state of things or the end goal? Some of the types do currently exist but some don't.
> The purpose of some types is unclear: CollatedGuidType, NonDefaultCharsetType — what are they?

They basically solve the problem that for example in MySQL the type supports the collation property. So in the case of MySQL the `CollatedGuidType` would be used (supporting collation), but for systems that do not support collation on that column type the `GuidType` would have to be used. That way there is no possible way a property is used to generate an invalid query, as that property does not exist. 

> What we may benefit from is defining layers: some types may be reproduced just on a DB (e.g. a fixed-length string) but some > can be only defined at the application level and built on top of the DB ones (e.g. the SimpleArrayType is probably stored in 
> varchar). Some application-level types share the same underlying DB type and differ only in serialization/deserialization (e.g. 
> date and immutable date).

I'm suprised this does not exist already, would be good to have in the docs? You're basically asking me to reverse engineer this whole package and those platform classes ain't pretty. The mapping of these types resides in the platform layer, our focus is currently on the column layer. If you'd like an overview of all the types and how they map to specific rdbms types and what options/properties they support per version, than I must say I'm sorry, but that is too much for me and reaches way beyond the scope of my issue.

> To me, the nullability, the default value and autoincrement are not properties of the type. They are properties of the column.

Okay, so if that is the case, how would you handle the autoincrement property on a `BooleanType` and why?
As far as I've seen the auto increment is used for integer type columns  only, so it would be a little odd IMO. Also unfortunately not all columns can have a default value in all rdbms. For example in MySQL a `TEXT` and `BLOB` type  column cannot have a default value. As for nonnull, I agree which is why it is part of the base type.

> Some types like Drizzle are no longer supported. The big/small integer types are purely MySQL dialects. IMO the more 

I took the types that are available in the version that we're talking about, the one used in this PR. The classes exist and might by in use, hence the reason they're part of this diagram.

> fundamental properties of a numeric type are scale and precision. Numbers of a zero scale are integers.

That is not how it currently works, there is a clear separation between a numeric/decimal type and an integer based type. See for example the type mapping for MySQL:

```
            'tinyint'       => 'boolean',
            'smallint'      => 'smallint',
            'mediumint'     => 'integer',
            'int'           => 'integer',
            'integer'       => 'integer',
            'bigint'        => 'bigint',
            'tinytext'      => 'text',
            'mediumtext'    => 'text',
            'longtext'      => 'text',
            'text'          => 'text',
            'varchar'       => 'string',
            'string'        => 'string',
            'char'          => 'string',
            'date'          => 'date',
            'datetime'      => 'datetime',
            'timestamp'     => 'datetime',
            'time'          => 'time',
            'float'         => 'float',
            'double'        => 'float',
            'real'          => 'float',
            'decimal'       => 'decimal',
            'numeric'       => 'decimal',
            'year'          => 'date',
            'longblob'      => 'blob',
            'blob'          => 'blob',
            'mediumblob'    => 'blob',
            'tinyblob'      => 'blob',
            'binary'        => 'binary',
            'varbinary'     => 'binary',
            'set'           => 'simple_array',
```

> Given the class hierarchy and the interfaces on the diagram, how should the type conversion work?

Considering all type/platform specific options are now part of the type and not the column, I think it would be as easy as to just call `setType()` on the `Column` with a different type object.

4c0n:
@morozov ping

morozov:
@4c0n, right now I'm on vacation and I wasn't able to thoroughly read your last comment.

It looks like this issue will require a lot of effort which I personally am not ready to commit to. You can try working on improving the type system alone or in a group, and I hope somebody of the maintainers (including myself) will be able to find to provide feedback.

4c0n:
@greg0ire @SenseException @morozov Alright, I get it, I'm not ready to put in that much work either, although the difference is that I'm not involved in this project. You could try throwing it out there in your team perhaps? Maybe you maintainers have some meetings or get together and discuss future plans and/or some form of roadmap?

Also this would definitely require a major version bump, and I can't wait that long.

Shame though, a lot of time went to waste looking into/explaining this.

4c0n:
For now we could roll with the simple fix:
https://github.com/doctrine/dbal/pull/4154

PR #4106, opened 2020-06-25
respect to platform restrictions when creating autoincrement identifier name

karakayasemi:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

`OraclePlatform` does not respect to 30 character max element name restriction in `getAutoincrementIdentifierName` method. This situation leads to `ORA-00972: identifier is too long` error in Oracle databases like here https://github.com/owncloud/twofactor_totp/issues/140#issue-499219646. This PR aims to fix this bug.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/OraclePlatform.php
@@ -587,6 +587,22 @@ END;';
     }
 
     /**
+     * Adds suffix to identifier,
+     *
+     * if the new string exceeds max identifier length,
+     * keeps $suffix, cuts from $identifier as much as the part exceeding.
+     */
+    private function addSuffix(string $identifier, string $suffix) : string
+    {
+        $maxPossibleLengthWithoutSuffix = $this->getMaxIdentifierLength() - strlen($suffix);
+        if (strlen($identifier) > $maxPossibleLengthWithoutSuffix) {
+            $identifier = substr($identifier, 0, $maxPossibleLengthWithoutSuffix);
+        }
+
+        return $identifier . $suffix;
+    }
+
+    /**
      * Returns the autoincrement primary key identifier name for the given table identifier.
      *
      * Quotes the autoincrement primary key identifier name
@@ -598,7 +614,7 @@ END;';
      */
     private function getAutoincrementIdentifierName(Identifier $table)
     {
-        $identifierName = $table->getName() . '_AI_PK';
+        $identifierName = $this->addSuffix($table->getName(), '_AI_PK');
 
         return $table->isQuoted()
             ? $this->quoteSingleIdentifier($identifierName)
@@ -966,7 +982,7 @@ SQL
         $table = new Identifier($tableName);
 
         // No usage of column name to preserve BC compatibility with <2.5
-        $identitySequenceName = $table->getName() . '_SEQ';
+        $identitySequenceName = $this->addSuffix($table->getName(), '_SEQ');
 
         if ($table->isQuoted()) {
             $identitySequenceName = '"' . $identitySequenceName . '"';
--- a/tests/Doctrine/Tests/DBAL/Functional/Platform/PlatformRestrictionsTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Platform/PlatformRestrictionsTest.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Functional\Platform;
+
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\Tests\DbalFunctionalTestCase;
+use function str_repeat;
+
+/**
+ * This class holds tests that make sure generated SQL statements respect to platform restrictions
+ * like maximum element name length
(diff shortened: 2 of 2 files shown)


morozov:
Thanks for the contribution, @karakayasemi. Please add a functional test or tweak an existing one that will cover your case on all supported platforms. Looks like it's not tested anywhere.

karakayasemi:
Since generated names include table name, shortening generated name by cutting its exceeding part from the at the end of the string can result in the same name with the table name. So, I kept the suffix and cut the base string as needed to adjust the length.

As you suggest, a functional test that covers the max identifier length case has been added. @morozov please review once more. Thank you.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -963,10 +971,17 @@ public function usesSequenceEmulatedIdentityColumns()
      */
     public function getIdentitySequenceName($tableName, $columnName)
     {
-        $table = new Identifier($tableName);
+        $table                = new Identifier($tableName);
+        $identitySequenceName = $table->getName();
+
+        // max possible length without '_SEQ' suffix
+        $maxPossibleLengthWithoutSuffix = $this->getMaxIdentifierLength() - 4;
```
morozov, line 978: Could this repeating code be extracted in a separate private method e.g. `addSuffix(string $identifier, string $suffix): string`?

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
```diff
@@ -1074,9 +1089,9 @@ public function getTimeFormatString()
      */
     public function fixSchemaElementName($schemaElementName)
     {
-        if (strlen($schemaElementName) > 30) {
+        if (strlen($schemaElementName) > $this->getMaxIdentifierLength()) {
```
morozov, line 1092: This method doesn't seem to be used anywhere. Why is it needed to modify it?

tests/Doctrine/Tests/DBAL/Functional/Platform/PlatformRestrictionsTest.php
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * This class holds tests that make sure generated SQL statements respect to platform restrictions
+ * like maximum element name length
+ */
+class PlatformRestrictionsTest extends DbalFunctionalTestCase
+{
+    protected function setUp() : void
+    {
+        parent::setUp();
+    }
```
morozov, line 19: This method is unnecessary.

tests/Doctrine/Tests/DBAL/Functional/Platform/PlatformRestrictionsTest.php
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Tests element names that are at the boundary of the identifier length limit.
+     * Ensures generated auto-increment identifier name respects to platform restrictions.
+     */
+    public function testMaxIdentifierLengthLimitWithAutoIncrement() : void
+    {
+        $allowedCharacters = 'abcdefghijklmnopqrstuvwxyz';
+        $platform          = $this->connection->getDatabasePlatform();
+        $tableName         = $this->generateRandomString($allowedCharacters, $platform->getMaxIdentifierLength());
```
morozov, line 29: Why does it have to be a random string? `str_repeat('x', $platform->getMaxIdentifierLength())` should suffice.

morozov (commented):
Looks good for the most part, just needs some cleanup :+1:

tests/Doctrine/Tests/DBAL/Functional/Platform/PlatformRestrictionsTest.php
karakayasemi: removed.

tests/Doctrine/Tests/DBAL/Functional/Platform/PlatformRestrictionsTest.php
karakayasemi: Changed with `str_repeat`.

karakayasemi:
@morozov it is ready for review again. Thank you for your help.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
morozov: This is irrelevant to the fix and is not used anywhere. I'm going to deprecate this method and remove it later. Please revert.

morozov:
Looks good. Please revert the irrelevant changes and squash.

lib/Doctrine/DBAL/Platforms/OraclePlatform.php
karakayasemi: The changes in the mentioned method have been reverted.

morozov (approved):

karakayasemi:
@morozov the irrelevant change has been reverted and commits have been squashed.

morozov:
Thanks, @karakayasemi!

PR #4119, opened 2020-06-27
Deprecate PingableConnection in favor of handling lost connection

morozov:
|      Q       |   A
|------------- | -----------
| BC Break     | no

1. The "ping" functionality in the DBAL is really questionable. The only way to check if the connection is alive is to execute a query. And even if one query succeeds or fails, it doesn't guarantee that another one will succeed or fail as well. Additionally, it's only supported by `mysql`-based drivers.
2. Instead, the wrapper connection can handle a `ConnectionLost` exception if it's reported by the driver. This means the behavior is happening automatically from now on.
3. Statement constructors are marked internal. It will allow to remove the dependency of the wrapper `Statement` on the wrapper `Connection` similarly to how it was done with the `oci8` driver in #3808. The new `Connection::handle*()` internal methods are temporary and will be moved to an `ExceptionHandler` after refactoring.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1925,6 +1925,8 @@ class Connection implements DriverConnection
      * It is responsibility of the developer to handle this case
      * and abort the request or reconnect manually:
      *
+     * @deprecated
+     *
      * @return bool
      *
      * @example
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -13,6 +13,7 @@ use Doctrine\DBAL\Driver\Connection as DriverConnection;
 use Doctrine\DBAL\Driver\PingableConnection;
 use Doctrine\DBAL\Driver\ResultStatement;
 use Doctrine\DBAL\Driver\ServerInfoAwareConnection;
+use Doctrine\DBAL\Exception\ConnectionLost;
 use Doctrine\DBAL\Exception\InvalidArgumentException;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Query\Expression\ExpressionBuilder;
@@ -614,7 +615,7 @@ class Connection implements DriverConnection
 
             return $stmt->fetch(FetchMode::ASSOCIATIVE);
         } catch (Throwable $e) {
-            throw DBALException::driverExceptionDuringQuery($this->_driver, $e, $query);
+            $this->handleExceptionDuringQuery($e, $query, $params, $types);
         }
     }
 
@@ -641,7 +642,7 @@ class Connection implements DriverConnection
 
             return $stmt->fetch(FetchMode::NUMERIC);
         } catch (Throwable $e) {
-            throw DBALException::driverExceptionDuringQuery($this->_driver, $e, $query);
+            $this->handleExceptionDuringQuery($e, $query, $params, $types);
         }
     }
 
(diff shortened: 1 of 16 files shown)


greg0ire:
I restarted the MySQL jobs but they still fail.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1954,4 +1957,60 @@ public function ping()
...
+     *
+     * @param string                                           $sql
+     * @param array<int, mixed>|array<string, mixed>           $params
+     * @param array<int, int|string>|array<string, int|string> $types
+     *
+     * @throws DBALException
+     *
+     * @psalm-return never-return
+     */
+    public function handleExceptionDuringQuery(Throwable $e, $sql, array $params = [], array $types = []): void
```
greg0ire, line 1972: It's internal. Should we maybe also make it `final`?
```suggestion
    final public function handleExceptionDuringQuery(Throwable $e, $sql, array $params = [], array $types = []): void
```

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1954,4 +1957,60 @@ public function ping()
             return false;
         }
     }
+
+    /**
+     * @internal
+     *
+     * @param string                                           $sql
```
greg0ire, line 1964: Why not use a type declaration here?

morozov: Because of a copy/paste from some other signature.

lib/Doctrine/DBAL/Connection.php
morozov: Make sense.

morozov: Unfortunately, it won't work because of the existing test that now has to mock this method to pass:

https://github.com/doctrine/dbal/blob/0a5bac1c2ce1d1b68944541e6a74e2325d274847/tests/Doctrine/Tests/DBAL/StatementTest.php#L134-L136

greg0ire (approved):

healerz:
Hi @morozov 
I have a question regarding this change. In our project, we use `ping()` to ensure the connection is not lost:
```
if (false === $this->ping($connection)) {
  $connection->close();
  $connection->connect();
}
```
Correct me if I'm wrong, but if i understand this change correctly, the difference is that the connection gets closed properly if the connection is lost. But it doesn't executes the query again, in such a case. So it only ensures that the next query succeeds again. 
To me, this is a BC, not in terms of the behavior, but regarding the provided functionality. Or what is the alternative to ensure the connection is not lost due to a connection timeout, so we can savely execute our actual query?

morozov:
> Correct me if I'm wrong, but if I understand this change correctly, the difference is that the connection gets closed properly if the connection is lost. But it doesn't execute the query again, in such a case. So it only ensures that the next query succeeds again.

Sounds correct.

> What is the alternative to ensure the connection is not lost due to a connection timeout, so we can safely execute our actual query?

There's no way to ensure that it's not lost. Even if the query you use to probe the connection succeeds, there's no guarantee that the real query will.

> Or is the solution to execute a dummy query and catch the ConnectionLost exception?

See the above. The solution is to catch the `ConnectionLost` exception and handle it depending on the query. E.g. it is safe to re-execute a `SELECT` but isn't safe to re-execute an `INSERT` w/o first checking the current state of the database (the connection may fail after the row has been inserted).

healerz:
@morozov thx for your reply. Makes all sense.

flmommens:
I catch the exception and immediately retry my SELECT query but get the same exception again.

morozov:
@flmommens please file a separate issue and provide all the details of the failure. The basic case is covered with this test: https://github.com/doctrine/dbal/blob/0a5bac1c2ce1d1b68944541e6a74e2325d274847/tests/Doctrine/Tests/DBAL/Functional/Connection/ConnectionLostTest.php#L31-L51

vudaltsov:
@morozov, why don't you check `instanceof ConnectionException`?

vudaltsov:
I think, I got it. `ConnectionException` is mainly thrown when credentials are invalid, so reconnecting will not help.

PR #4127, opened 2020-06-28
Rework extension detection in tests

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

1. Use `@requires extension` instead of `if (extension_loaded())` as a more declarative/idiomatic way.
2. Remove the test that requires `sqlite3` which is not supported.


greg0ire (approved):
Nice!

PR #4129, opened 2020-06-28
Remove ExceptionConverterDriver

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

See https://github.com/doctrine/dbal/pull/4118.


morozov:
@greg0ire after filing this PR, I realized that I want to go further and move the exception conversion logic into a separate interface that the driver would instantiate similarly to `getDatabasePlatform()`. The conversion logic is the biggest reason why we have abstract driver classes, so moving it out would be a good first step towards getting rid of the inheritance in drivers and turning them into thin facades (not sure if that's the right term though).

Implementing this will require some more cleanup of the driver exceptions which I'm currently working on, and it also depends on these changes as well.

What do you think if we merge this as is and rework in the next few days?

greg0ire (approved):
Sounds good

PR #4130, opened 2020-06-28
Missing driver exception annotations

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

1. All driver-level methods that may throw an exception are documented with a `@throws` annotation.
2. `OCI8\Connection::getServerVersion()` no longer throws an `UnexpectedValueException`.
3. `Mysqli\Statement::bindTypedParameters()` and `PDO\Statement::convertParamType()` no longer throw an `InvalidArgumentException`.
4. Driver-level methods are no longer allowed to throw any other exceptions that `Driver\Exeption`.

### DIFF
--- a/src/Driver/Mysqli/Exception/NonStreamResourceUsedAsLargeObject.php
+++ b/src/Driver/Mysqli/Exception/NonStreamResourceUsedAsLargeObject.php
@@ -13,10 +13,12 @@ use function sprintf;
  *
  * @psalm-immutable
  */
-final class FailedReadingStreamOffset extends AbstractException
+final class NonStreamResourceUsedAsLargeObject extends AbstractException
 {
     public static function new(int $parameter): self
     {
-        return new self(sprintf('Failed reading the stream resource for parameter #%d.', $parameter));
+        return new self(
+            sprintf('The resource passed as a LARGE_OBJECT parameter #%d must be of type "stream"', $parameter)
+        );
     }
 }
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,20 @@
 # Upgrade to 3.0
 
+## BC BREAK: More driver-level methods are allowed to throw a Driver\Exception.
+
+The following driver-level methods are allowed to throw a Driver\Exception:
+
+- `Connection::prepare()`
+- `Connection::lastInsertId()`
+- `Connection::beginTransaction()`
+- `Connection::commit()`
+- `Connection::rollBack()`
+- `ServerInfoAwareConnection::getServerVersion()`
+- `Statement::bindParam()`
+- `Statement::bindValue()`
+- `Result::rowCount()`
+- `Result::columnCount()`
+
 ## The `ExceptionConverterDriver` interface is removed
 
 All drivers must implement the `convertException()` method which is now part of the `Driver` interface.
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -5,6 +5,8 @@
 The following driver-level methods are allowed to throw a Driver\Exception:
 
 - `Connection::prepare()`
+- `Connection::query()`
+- `Connection::exec()`
 - `Connection::lastInsertId()`
 - `Connection::beginTransaction()`
 - `Connection::commit()`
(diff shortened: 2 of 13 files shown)


greg0ire (dismissed):

morozov:
@SenseException could you review this, please? There was needed a rebase due to a conflict in `UPGRADE.md`.

morozov:
Ping @greg0ire.

src/Driver/Mysqli/Exception/NonStreamResourceUsedAsLargeObject.php
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * @internal
+ *
+ * @psalm-immutable
+ */
+final class NonStreamResourceUsedAsLargeObject extends AbstractException
+{
+    public static function new(int $parameter): self
+    {
+        return new self(sprintf('A non-stream resource passed as a LARGE_OBJECT parameter #%d', $parameter));
```
greg0ire, line 20: This sentence feels a bit weird to me, I'd either use
```suggestion
        return new self(sprintf('A non-stream resource was passed as a LARGE_OBJECT parameter #%d', $parameter));
```

or 

```suggestion
        return new self(sprintf('Non-stream resource passed as a LARGE_OBJECT parameter #%d', $parameter));
```

morozov: What if we get rid of the past tense here and change it to something like: 'The resource passed as a LARGE_OBJECT parameter must be of type "stream"?' It will be more about how to fix the problem, not just the problem statement.

morozov: As a side note, have you seen any articles/guides specifically on composing error messages? E.g. in navigation it seems acceptable to sacrifice the grammatical correctness of sentence in favor of brevity: do not lean on door). And many other computer programs I saw (especially in the past) would use this principle as well.

morozov: As another side note, we probably can get rid of this exception and let the underlying driver decide whether it can bind this resource to the statement or not. Then the message (if it's clear enough) may come directly from the driver (should be easier to google).

greg0ire (approved):

src/Driver/Mysqli/Exception/NonStreamResourceUsedAsLargeObject.php
greg0ire: > As a side note, have you seen any articles/guides specifically on composing error messages? E.g. in navigation it seems acceptable to sacrifice the grammatical correctness of sentence in favor of brevity: do not lean on door). And many other computer programs I saw (especially in the past) would use this principle as well.

I didn't but it's an interesting topic for sure!

morozov: Just [googled](https://forum.wordreference.com/threads/do-not-lean-on-door.2337460/) it for myself and will leave it here :

> Signs often have reduced grammar, as do newspaper headlines, captions of photographs, and so on. They may omit articles and auxiliary verbs, for example. We're so used to this that it might actually sound strange to use a complete sentence in such situations.

This is by no means a justification for my illiteracy.

greg0ire: Haha I stumbled upon this blog post that refers to us while looking this up: http://rosstuck.com/formatting-exception-messages

This one is quite interesting: https://softwareengineering.stackexchange.com/questions/29433/how-to-write-a-good-exception-message, but more about the contents than the form.

This one advocates against brevity: http://amortizedcost.net/writing-good-exception-message/

IMO there might have been a point in brevity in the past, when memory was an issue, but nowadays we might want to revisit that thought and start using multiline exception messages, which detail the context, the problem and solutions as suggested in the second link.

src/Driver/Mysqli/Exception/NonStreamResourceUsedAsLargeObject.php
```diff
@@ -0,0 +1,24 @@
...
+ * @internal
+ *
+ * @psalm-immutable
+ */
+final class NonStreamResourceUsedAsLargeObject extends AbstractException
+{
+    public static function new(int $parameter): self
+    {
+        return new self(
+            sprintf('The resource passed as a LARGE_OBJECT parameter #%d must be of type "stream"', $parameter)
```
greg0ire, line 21: Should we add the actual resource type to the error message?

PR #4132, opened 2020-06-29
Deprecate AbstractPlatform::fixSchemaElementName()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no

The method is not used anywhere except for tests (discovered by #4106).

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade to 2.11
 
+## `AbstractPlatform::fixSchemaElementName()` is deprecated.
+
+The method is not used anywhere except for tests.
+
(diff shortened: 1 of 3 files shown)


SenseException (changes requested):
`SQLAnywherePlatform` is also having that method and needs a deprecation. https://github.com/doctrine/dbal/blob/e8510d10f32d9924535b9583226c485d31f10e19/lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php#L73

morozov:
I deliberately omitted it because the class itself is deprecated.

SenseException (approved):
That should be sufficient.

Issue #4141, opened 2020-07-03
Add support for MYSQLI_OPT_READ_TIMEOUT option


siimsoni:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

Add support for undocumented `MYSQLI_OPT_READ_TIMEOUT` option. Comparable to `MYSQL_OPT_READ_TIMEOUT` in [MySQL C API](https://dev.mysql.com/doc/refman/8.0/en/mysql-options.html).

> Add support to specify timeout in seconds for each attempt to read from the server.

The currently supported `MYSQLI_OPT_CONNECT_TIMEOUT` specifies timeout in seconds for establishing the connection, but not for the time it takes to read the data. There doesn't appear to be support for write timeouts.

There is an open [PHP bug report](https://bugs.php.net/bug.php?id=76497) to add documentation for this option.

siimsoni:
Closed with df465cc

Sunil4imr:
Hello,
In symfony, I've been trying to add MYSQLI_OPT_READ_TIMEOUT in cofig.yml. it's not giving any error. but it didn't work for me.
i used it like this. is it a right way to use it in cofig.yml file?

> 
    doctrine:
            dbal:
                driver: pdo_mysql
                host: "%database_host%"
                dbname: "%database_name%"
                user: "%database_user%"
                password: "%database_password%"
                options:    
                    MYSQLI_OPT_READ_TIMEOUT : 1

siimsoni:
This is an option for `mysqli` driver.
https://www.doctrine-project.org/projects/doctrine-dbal/en/2.10/reference/configuration.html#mysqli

Sunil4imr:
I added mysqli driver but still  it's not working

        doctrine:
            dbal:
                driver: mysqli
                host: "%database_host%"
                port: "%database_port%"
                dbname: "%database_name%"
                user: "%database_user%"
                password: "%database_password%"
                server_version: '15.1'
                options:    
                      MYSQLI_OPT_READ_TIMEOUT : 1    
it's showing this error.....
[...]

PR #4142, opened 2020-07-03
Add support for MYSQLI_OPT_READ_TIMEOUT

siimsoni:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #4141

#### Summary

Add support for MYSQLI_OPT_READ_TIMEOUT.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
+++ b/lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
@@ -11,7 +11,6 @@ use Doctrine\DBAL\Driver\ServerInfoAwareConnection;
 use Doctrine\DBAL\ParameterType;
 use mysqli;
 
-use function defined;
 use function floor;
 use function func_get_args;
 use function in_array;
@@ -28,6 +27,7 @@ use function stripos;
 use const MYSQLI_INIT_COMMAND;
 use const MYSQLI_OPT_CONNECT_TIMEOUT;
 use const MYSQLI_OPT_LOCAL_INFILE;
+use const MYSQLI_OPT_READ_TIMEOUT;
 use const MYSQLI_READ_DEFAULT_FILE;
 use const MYSQLI_READ_DEFAULT_GROUP;
 use const MYSQLI_SERVER_PUBLIC_KEY;
@@ -243,14 +243,12 @@ class MysqliConnection implements ConnectionInterface, PingableConnection, Serve
         $supportedDriverOptions = [
             MYSQLI_OPT_CONNECT_TIMEOUT,
             MYSQLI_OPT_LOCAL_INFILE,
+            MYSQLI_OPT_READ_TIMEOUT,
             MYSQLI_INIT_COMMAND,
             MYSQLI_READ_DEFAULT_FILE,
             MYSQLI_READ_DEFAULT_GROUP,
+            MYSQLI_SERVER_PUBLIC_KEY,
         ];
-
-        if (defined('MYSQLI_SERVER_PUBLIC_KEY')) {
-            $supportedDriverOptions[] = MYSQLI_SERVER_PUBLIC_KEY;
-        }
 
         $exceptionMsg = "%s option '%s' with value '%s'";
 


greg0ire:
Please associate your email address with your Github account, or change the

greg0ire:
Just to understand, this will timeout if the connection is successful but SQL request takes too long to execute, or are there other things involved when talking to mysql that might cause it to kick in before? Do you know if there are equivalents for PDO?

siimsoni:
Yes, this will terminate the connection when a query exceeds timeout period.

For example, with a 4 second timeout you can run 2 queries each 3 seconds. If the first query takes 4 seconds instead, the connection will be terminated, and running the second query would require a reconnect.

I don't think PDO supports driver specific options:

> PDO::ATTR_TIMEOUT: Specifies the timeout duration in seconds. Not all drivers support this option, and its meaning may differ from driver to driver. For example, sqlite will wait for up to this time value before giving up on obtaining an writable lock, but **other drivers may interpret this as a connect or a read timeout interval**. Requires int.

https://www.php.net/manual/en/pdo.setattribute.php

siimsoni:
To expand on the previous point about PDO, the MySQL driver for PDO only sets `MYSQL_OPT_CONNECT_TIMEOUT`.
https://github.com/php/php-src/blob/c932cc841ab1bd80dbb1465f29d58302c95e8c78/ext/pdo_mysql/mysql_driver.c#L630

greg0ire:
Thanks for digging this up, I also asked because a colleague asked me about this today if all days :)

siimsoni:
I'm possibly wrong here, but related to missing tests tag, isn't option passing already covered in `Doctrine\Tests\DBAL\Functional\Driver\Mysqli\ConnectionTest::testDriverOptions`?

greg0ire:
I'm on my phone so it's hard for me to check, but codecov says your patch is covered, let me remove that tag

greg0ire (dismissed):
LGTM, please check the docs for any place where this option should be mentioned though

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
```diff
@@ -240,6 +241,10 @@ private function setDriverOptions(array $driverOptions = []) : void
             $supportedDriverOptions[] = MYSQLI_SERVER_PUBLIC_KEY;
         }
 
+        if (defined('MYSQLI_OPT_READ_TIMEOUT')) {
```
morozov, line 244: Why is this condition needed?

morozov (commented):
I don't say this often but I think it's okay if we don't add a test here. The `$supportedOptions` array is removed as of #4081, so the consumer will be able to specify any options that the underlying driver supports.

If you want to add a test, it will have to be a functional one (see `ConnectionLostTest` for example).

morozov:
Since this is a new feature, please retarget against `2.11.x`.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
siimsoni: The option appears to be added in PHP 7.2.

morozov:
>> I'm possibly wrong here, but related to missing tests tag, isn't option passing already covered in `Doctrine\Tests\DBAL\Functional\Driver\Mysqli\ConnectionTest::testDriverOptions`?
>
> I'm on my phone so it's hard for me to check, but codecov says your patch is covered, let me remove that tag

The fact that the new code is covered by existing tests can be only used as evidence that the changes don't break existing functionality. It cannot be used as a reason to not add a test. Otherwise, using the same arguments, this code can be removed w/o breaking any test.

The test is still missing but it's not required here just because there's no logic associated with it and it will be removed in `3.0.x` anyways.

morozov (changes requested):
See above.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
morozov: The requirement for PHP 7.2 or newer is rock-solid.

siimsoni:
> LGTM, please check the docs for any place where this option should be mentioned though

The related documentation appears to only link to relevant PHP documentation. This option is missing documentation there, and this is covered by [open bug report](https://bugs.php.net/bug.php?id=76497).

> If you want to add a test, it will have to be a functional one (see ConnectionLostTest for example).

The way I see it is that DBAL doesn't make assumptions about what happens when this option is specified, and the functional behavior should be tested in respective project that maintains it (MySQL C API or the server).

morozov (approved):

morozov:
Thank you, @siimsoni.

siimsoni:
> Do you know if there are equivalents for PDO?

Turns out when using `mysqlnd` (instead of `libmysqlclient`), which typically is the case, there is a better way to specify this. That also works with PDO:
https://www.php.net/manual/en/mysqlnd.config.php#ini.mysqlnd.net-read-timeout

greg0ire:
Interesting! And it can be set at runtime too, good to know!

PR #4144, opened 2020-07-03
Deprecate classes in Driver\PDO* namespaces

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

Now that we marked connection constructors internal in https://github.com/doctrine/dbal/pull/4139 and reworked PDO classes into a composition in https://github.com/doctrine/dbal/pull/4138, we effectively have `PDOSqlsrv\Driver::connect()` calling `PDO\Connection::__construct()` which is `@internal` to the `PDO` namespace. Psalm at level 4 may not like it (https://psalm.dev/r/d385c2021c, https://github.com/sebastianbergmann/phpunit/issues/4309).

The solution is to move all PDO-based components under the PDO namespace and thereby legitimize their usage of the internal PDO APIs. Additionally, it will make the driver namespace structure cleaner.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,17 @@
 # Upgrade to 2.11
 
+## PDO-related classes outside of the PDO namespace are deprecated
+
+The following outside of the PDO namespace have been deprecated in favor of their counterparts in the PDO namespace:
+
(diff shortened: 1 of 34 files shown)


greg0ire (approved):
TIL `@internal` works on a namespace basis

PR #4150, opened 2020-07-05
Update doctrine/coding-standard to 8.1

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

When upgrading to Coding Standard the 8.0 in https://github.com/doctrine/dbal/pull/4050, I decided to upgrade only `2.11.x` in order to not touch dependencies in `2.10.x` but in practice having different coding standards in different supported versions creates extra merge conflicts during up-merges. Furthermore, the coding standard is a development dependency bumping which won't affect the package dependents.


greg0ire (approved):
It's only dev dependencies so this is ok 👍

Issue #4104, opened 2020-06-24
Invalid queries are generated


4c0n:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.10.x

#### Summary

When changing a column type the platform options are not considered, in version 2.9 for MySQL/MariaDB the charset key was not passed as platform option in a column object. Since 2.10 the charset key is passed and it is causing generation of invalid queries.

#### Current behaviour

In this scenario the query generated to alter the table is:
'ALTER TABLE to_be_altered CHANGE some_column_name some_column_name INT CHARACTER SET utf8 DEFAULT NULL COLLATE utf8_unicode_ci'

Obviously the charset should not be provided/processed for integer type columns.

#### How to reproduce

I opened a PR containing the failing case: https://github.com/doctrine/dbal/pull/4103
See the builds for more information.

#### Expected behaviour

At least for the MySQL dialect the query should be generated without 'CHARACTER SET utf8` when changing to integer type.

morozov:
Closing based on https://github.com/doctrine/dbal/pull/4103#issuecomment-658021097.

PR #4154, opened 2020-07-08
Feature/do not use charset collation for mysql column types that do not support it

4c0n:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #4104 

#### Summary

Remove the `charset` and `collation` platform options for column types that cannot use it.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -2,7 +2,6 @@
 
 namespace Doctrine\DBAL\Platforms;
 
-use Doctrine\DBAL\Schema\Column;
 use Doctrine\DBAL\Schema\ForeignKeyConstraint;
 use Doctrine\DBAL\Schema\Identifier;
 use Doctrine\DBAL\Schema\Index;
@@ -12,7 +11,6 @@ use Doctrine\DBAL\TransactionIsolationLevel;
 use Doctrine\DBAL\Types\BlobType;
 use Doctrine\DBAL\Types\StringType;
 use Doctrine\DBAL\Types\TextType;
-use Doctrine\DBAL\Types\Type;
 use InvalidArgumentException;
 use function array_diff_key;
 use function array_merge;
@@ -597,7 +595,7 @@ SQL
                 continue;
             }
 
-            $type = $column->getType();
+            $type = get_class($column->getType());
             if ($type !== StringType::class && $type !== TextType::class) {
                 $options = $column->getPlatformOptions();
                 unset($options['charset']);
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -18,6 +18,7 @@ use function array_unique;
 use function array_values;
 use function count;
 use function func_get_args;
+use function get_class;
 use function implode;
 use function in_array;
 use function is_numeric;
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\DBAL\Platforms;
 
+use Doctrine\DBAL\Schema\Column;
 use Doctrine\DBAL\Schema\ForeignKeyConstraint;
 use Doctrine\DBAL\Schema\Identifier;
 use Doctrine\DBAL\Schema\Index;
@@ -9,7 +10,9 @@ use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Schema\TableDiff;
 use Doctrine\DBAL\TransactionIsolationLevel;
 use Doctrine\DBAL\Types\BlobType;
+use Doctrine\DBAL\Types\StringType;
 use Doctrine\DBAL\Types\TextType;
+use Doctrine\DBAL\Types\Type;
 use InvalidArgumentException;
 use function array_diff_key;
 use function array_merge;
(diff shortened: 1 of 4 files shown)


SenseException (commented):
What is the issue here that made you create this PR? Did you run into errors or other problems? It seems that these changes are a programmatic removal of a misconfiguration.

4c0n:
@SenseException please see: https://github.com/doctrine/dbal/issues/4104

4c0n:
@SenseException Did you have the chance to look at the issue?

SenseException (changes requested):
Please note my previous comment.

We also need one or more tests that covers the query part where the alteration won't be taken on non-text and non-text types. I don't think that the type-specific handlings should be done in `Column`. _charset_ and _collation_ are added in the platforms of DBAL like `AbstractPlatform` and the db specific platforms. Can you please take a look on the SQL generation there instead of `Column`?

4c0n:
@SenseException I updated the PR following your request. If this is okay for you I'll add some tests.

4c0n:
@SenseException Is it okay for you this way? Please let me know so that I can add the tests.

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -594,6 +596,13 @@ public function getAlterTableSQL(TableDiff $diff)
                 continue;
             }
 
+            $type = get_class($column->getType());
+            if ($type !== StringType::class && $type !== TextType::class) {
```
SenseException, line 600: You can use `$type->getName()` instead of `get_class`.

SenseException (changes requested):
Thanks for the changes and your time.

I'd like to ask @doctrine/doctrinecore what they think about the fix for removing charsets. There's currently only MySQL covered by this but it might be interesting for the other platforms too. DBAL has already some event listeners for other type of events and I wonder if it makes sense to introduce an `onSchemaAlterTableChangeColumn` event to handle charsets for all platforms.

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
morozov: I would rather compare the classes. Similarly to other interfaces’ `getName()` methods, this one may be deprecated and removed in the future.

morozov:
> I'd like to ask @doctrine/doctrinecore what they think about the fix for removing charsets.

I don't understand the question. Are you proposing to remove charsets as a core feature and implement their support in event listeners? It looks like currently they are implemented as a platform option which seems the right place. Another question is, what's the use case for specifying the charset/collation on the per-table or per-column basis? In most cases, it should be enough to specify them once for the entire database.

SenseException:
No, I meant that the here implemented removal of the charset for the fix of the mentioned issue could be put into an event instead of the platform classes. I don't know if events would be a fitting spot in case this fix is supposed to work for every platform.

> Another question is, what's the use case for specifying the charset/collation on the per-table or per-column basis? In most cases, it should be enough to specify them once for the entire database.

I had no use case yet where a different charset could be needed in columns that differs from table or database. It's still supported though.

4c0n:
@SenseException Can we just focus on the issue at hand? Which is MySQL only at this point. I'd really like to get this sorted so that I can start enjoying newer versions of dbal again. I've spend weeks having this issue open and making suggestions, pr's etc... @morozov abandoned the issue because he was going on holiday and didn't want to commit to a big refactor. And is now questioning the use case of this. Without following up on your response.

Now this is the simple fix for the problem at hand and I'd like to suggest opening (a) new issue(s) for the remaining platforms. It's always possible to refactor this change if/when you decide to do something about that. Otherwise the YAGNI principle comes to mind.

From here on I'm sorry to say that I'm very much done spending time on this. I really love helping out and contributing to open source projects, but if things have to go like this, it seems like a big waste of time. Take it or leave it. If it doesn't get merged anytime soon I'll just implement some silly/ugly workaround in my own code base.

Thanks for your time and efforts so far.

SenseException:
@4c0n Since no one else added suggestions to this review I'd like to ask you to add the tests for your changes.

4c0n:
@SenseException Thanks, I'll add the tests soon.

4c0n:
@SenseException I've added the test for the change. Please have a look.

tests/Doctrine/Tests/DBAL/Platforms/MySqlPlatformTest.php
```diff
@@ -20,4 +26,32 @@ public function testHasCorrectDefaultTransactionIsolationLevel() : void
...
+        $tableDiff  = $comparator->diffTable($tableA, $tableB);
+
+        $this->platform->getAlterTableSQL($tableDiff);
+
+        self::assertGreaterThan(0, count($tableDiff->changedColumns));
+
+        foreach ($tableDiff->changedColumns as $columnDiff) {
+            $column = $columnDiff->column;
+
+            self::assertFalse($column->hasPlatformOption('charset'));
```
SenseException, line 54: Can you please assert the resulting SQL queries that are created by `getAlterTableSQL()`? You'll find similar tests in `AbstractMySQLPlatformTestCase`, so you don't need to check the columns definitions.

SenseException (changes requested):

tests/Doctrine/Tests/DBAL/Platforms/MySqlPlatformTest.php
4c0n: @SenseException The purpose of this (unit) test is to test that the platform option is removed (which could be quite a bit more isolated using mocks I might add). Which is exactly what my change does under certain conditions.
Why are you asking me to implement a test to verify the complete behavior of the method of this derived platform class? I guess it is something the author of the method should have done.
The way I see it, my change is covered and it even increased the overall code coverage. If anything I'd say the test isn't testing the code in enough isolation and covering too much.
Anyway, like I explained before (and you did not acknowledge BTW), I'm done spending time on this.

SenseException: Thank you for your time you spent on this PR and your thoughts you put into this. I'll keep this PR open in case someone else wants to take over.

PR #4157, opened 2020-07-09
Rework the portability layer to act as a middleware

### LINKED ISSUE #4156: The portability spans two different architecture layers and doesn't properly implement the wrapper API
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.0.0-dev

As reported by PHPStan in https://github.com/doctrine/dbal/pull/4094 (`2.10.x`),

https://github.com/doctrine/dbal/blob/15d9be0b3c87e05f428f8c2beb1fb9555e7b75bf/phpstan.neon.dist#L99-L102

In `3.0.x` where the driver and wrapper APIs started diverging (https://github.com/doctrine/dbal/pull/4045), it results in the following:
```php
    'driver' => 'pdo_sqlite',
]);

foreach ($conn->prepare('SELECT 1')->execute()->iterateColumn() as $value) {
    var_dump($value);
}

// string(1) "1"

    'driver' => 'pdo_sqlite',
    'wrapperClass' => \Doctrine\DBAL\Portability\Connection::class,
    'portability'  => \Doctrine\DBAL\Portability\Connection::PORTABILITY_ALL,
    'fetch_case'   => \Doctrine\DBAL\ColumnCase::LOWER,
]);

foreach ($conn->prepare('SELECT 1')->execute()->iterateColumn() as $value) {
    var_dump($value);
}

// Fatal error: Call to undefined method Doctrine\DBAL\Portability\Result::iterateColumn()
```

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug, improvement
| BC Break     | yes

Fixes #4156.

### Importance

Apart from the fatal error reported in the linked issue, the current implementation blocks the decoupling of the driver and wrapper APIs. The wrapper-level `Connection::prepare()` declares its return type as `DriverStatement` (which is not intended but required by the driver-level `Connection` interface) however it returns a `Statement` (a more specific and feature-rich type):

https://github.com/doctrine/dbal/blob/98c23feb379994bdc5440a6d65e8933ed3d9a5f3/src/Connection.php#L908-L912

 Even if it stops implementing the driver connection interface, it cannot declare its return type as `Statement` because the `Portability\Connection` class currently extends it and returns a driver-level statement:

https://github.com/doctrine/dbal/blob/48625f1bc761f1f8f70fe6630a0f3a3204d9a6ff/src/Portability/Statement.php#L9-L12

### The Driver Middleware concept

The concept of a driver middleware will allow extending the behavior of the DBAL connection by wrapping the underlying driver. Once the driver is wrapped, the wrapper can also wrap the connection and the statement objects effectively extending any of the behaviors. The middleware objects are registered via configuration (see example in the test).

Apart from the portability layer, it can be used to reimplement the caching layer which is currently implemented as part of the wrapper connection class.

### Portability layer reworked

The portability layer has been reworked to act as middleware and implement only the driver-level APIs.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -158,9 +158,11 @@ The method no longer accepts the `$username`, `$password` and `$driverOptions` a
 
 This class was deprecated in favor of `PrimaryReadReplicaConnection`
 
-## Removed `Portability\Connection::PORTABILITY_{PLATFORM}` constants`
+## BC BREAK: Changes in the portability layer
 
-The platform-specific portability constants were internal implementation details which are longer relevant.
+1. The platform-specific portability constants (`Portability\Connection::PORTABILITY_{PLATFORM}`) were internal implementation details which are no longer relevant.
+2. The `Portability\Connection` class no longer extends the DBAL `Connection`.
+3. The `Portability\Class` class has been made final.
 
 ## BC BREAK changes in fetching statement results
 
--- a/src/Portability/Driver.php
+++ b/src/Portability/Driver.php
@@ -1,0 +1,90 @@
+<?php
+
+namespace Doctrine\DBAL\Portability;
+
+use Doctrine\DBAL\ColumnCase;
+use Doctrine\DBAL\Connection as DBALConnection;
+use Doctrine\DBAL\Driver as DriverInterface;
+use Doctrine\DBAL\Driver\API\ExceptionConverter;
+use Doctrine\DBAL\Driver\PDO;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+
+use const CASE_LOWER;
+use const CASE_UPPER;
+
+final class Driver implements DriverInterface
+{
+    /** @var DriverInterface */
+    private $driver;
+
+    /** @var int */
+    private $mode;
+
+    /** @var int */
+    private $case;
+
+    public function __construct(DriverInterface $driver, int $mode, int $case)
+    {
+        $this->driver = $driver;
+        $this->mode   = $mode;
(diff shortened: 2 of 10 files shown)


UPGRADE.md
```diff
@@ -158,9 +158,11 @@ The method no longer accepts the `$username`, `$password` and `$driverOptions` a
 
 This class was deprecated in favor of `PrimaryReadReplicaConnection`
 
-## Removed `Portability\Connection::PORTABILITY_{PLATFORM}` constants`
+## BC BREAK: Changes in the portability layer
 
-The platform-specific portability constants were internal implementation details which are longer relevant.
+1. The platform-specific portability constants (`Portability\Connection::PORTABILITY_{PLATFORM}`) were internal implementation details which are longer relevant.
```
greg0ire, line 163: Let's take this as an occasion to fix this sentence?

```suggestion
1. The platform-specific portability constants (`Portability\Connection::PORTABILITY_{PLATFORM}`) were internal implementation details which are no longer relevant.
```

src/Portability/Driver.php
```diff
@@ -0,0 +1,90 @@
...
+
+    public function __construct(DriverInterface $driver, int $mode, int $case)
+    {
+        $this->driver = $driver;
+        $this->mode   = $mode;
+        $this->case   = $case;
+    }
+
+    /** @var DriverInterface */
+    private $driver;
```
greg0ire, line 31: I don't think you intended to put that property here, did you?

greg0ire (approved):

PR #4159, opened 2020-07-10
Do not implement driver-level interfaces by wrapper-level classes

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

There are certain downsides of having the wrapper-level classes have to implement the driver-level interfaces:

1. Alongside the `executeQuery()` and `executeUpdate()` methods, the wrapper connection has to implement `query()` and `exec()` just because the interface declares them. Although, API-wise the latter two are just the subset of the former.
2. Prior to PHP 7.4, the [return type contravariance](https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters) is not supported (https://3v4l.org/dNsho) which means that the wrapper-level classes cannot declare their return type specifically enough and have to declare the type just as prescribed by the lower-level interface.
3. The fact that the wrappers implement the driver interfaces makes it harder to evolve the wrapper layer independently of the driver one. E.g. we cannot add a new (even optional) argument to any of the wrapper methods that are declared in the driver interface without a BC break.
4. Unlike the driver-level classes that are allowed to throw driver exceptions and only them, the wrapper-level classes are not allowed to throw driver-level exceptions but are allowed to throw wrapper-level exceptions. This leads to the following issues:
   1. If a consumer uses a wrapper class as a driver interface implementation, it will expect a driver-level exception to be thrown but will get a wrapper-level one.
   2. If a consumer uses a wrapper class as such, it will expect a wrapper-level exception to be thrown but the static analysis will identify it as not handling the driver-level exceptions that are allowed by the driver-level interfaces but in fact never happen.

   Strictly speaking, at this point the wrapper layer _violates_ the interface that it's forced to implement.

As for the upsides, I haven't seen a case where a consumer should be able to use a wrapper connection as a driver one. The fact that not a single test had to be changed confirms that.

**TODO**:
- [x] Add a deprecation in `2.11.x` (#4165).

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade to 3.0
 
+## BC BREAK: Changes in the wrapper-level API ancestry
+
+The wrapper-level `Connection` and `Statement` classes no longer implement the corresponding driver-level interfaces.
+
 ## BC BREAK: Removed DBALException factory methods
 
 The following factory methods of the DBALException class have been removed:
--- a/docs/en/reference/architecture.rst
+++ b/docs/en/reference/architecture.rst
@@ -16,11 +16,7 @@ interfaces are implemented by concrete drivers. For all PDO based
 drivers, ``PDO`` and ``PDOStatement`` are the implementations of
 these interfaces. Thus, for PDO-based drivers, a
 ``Doctrine\DBAL\Connection`` wraps a ``PDO`` instance and a
-``Doctrine\DBAL\Statement`` wraps a ``PDOStatement`` instance. Even
-more, a ``Doctrine\DBAL\Connection`` *is a*
-``Doctrine\DBAL\Driver\Connection`` and a
-``Doctrine\DBAL\Statement`` *is a*
-``Doctrine\DBAL\Driver\Statement``.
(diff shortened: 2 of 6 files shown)


greg0ire (dismissed):

greg0ire (commented):
Hold on, I think https://github.com/doctrine/dbal/blob/3.0.x/docs/en/reference/architecture.rst has to be changed, for instance this sentence:

> Even more, a Doctrine\DBAL\Connection is a Doctrine\DBAL\Driver\Connection and a Doctrine\DBAL\Statement is a Doctrine\DBAL\Driver\Statement.

morozov:
> […] https://github.com/doctrine/dbal/blob/3.0.x/docs/en/reference/architecture.rst has to be changed, for instance this sentence:
> 
> > Even more, a Doctrine\DBAL\Connection is a Doctrine\DBAL\Driver\Connection and a Doctrine\DBAL\Statement is a Doctrine\DBAL\Driver\Statement.

Yeah, this is no longer true, so I'll just remove it. There are some other bits outdated as of `3.0.0-dev` but I'll update them in a separate PR.

greg0ire (approved):

PR #4163, opened 2020-07-11
Deprecate duplicate and ambiguous wrapper connection methods

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

### Deprecation of `Connection::query()` and `::exec()`

Once the wrapper connection has been freed from the responsibilities of the driver connection (https://github.com/doctrine/dbal/pull/4159), the `query()` and `exec()` methods that effectively (and poorly) duplicate `executeQuery()` and `executeUpdate()` can be deprecated.

### Deprecation of `Connection::executeUpdate()` in favor of `::executeStatement()`

The name of the `executeUpdate()` method is quote misleading since apart from the `UPDATE` statements, it should be used for `INSERT`, `DELETE` and all sorts of DML, DDL, TCL and other statements. As generic alternative `executeStatement()` looks more intuitive.

### Notes in conjunction with PrimaryReplicaConnection

Note that `PrimaryReplicaConnection::query()` ensured connection to the primary instance while `executeQuery()` doesn't.

Depending on the desired behavior:

- If the statement doesn't have to be executed on the primary instance, use `executeQuery()`.
- If the statement has to be executed on the primary instance and yields rows (e.g. `SELECT`), prepend `executeQuery()` with `ensureConnectedToPrimary()`.
- Otherwise, use `executeStatement()`.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -749,7 +749,7 @@ class Connection implements DriverConnection
 
         $this->addIdentifierCondition($identifier, $columns, $values, $conditions);
 
-        return $this->executeUpdate(
+        return $this->executeStatement(
             'DELETE FROM ' . $tableExpression . ' WHERE ' . implode(' AND ', $conditions),
             $values,
             is_string(key($types)) ? $this->extractTypeValues($columns, $types) : $types
@@ -777,7 +777,7 @@ class Connection implements DriverConnection
     {
         $this->transactionIsolationLevel = $level;
 
-        return $this->executeUpdate($this->getDatabasePlatform()->getSetTransactionIsolationSQL($level));
+        return $this->executeStatement($this->getDatabasePlatform()->getSetTransactionIsolationSQL($level));
     }
 
     /**
@@ -827,7 +827,7 @@ class Connection implements DriverConnection
         $sql = 'UPDATE ' . $tableExpression . ' SET ' . implode(', ', $set)
                 . ' WHERE ' . implode(' AND ', $conditions);
 
-        return $this->executeUpdate($sql, $values, $types);
+        return $this->executeStatement($sql, $values, $types);
     }
 
     /**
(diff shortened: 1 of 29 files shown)


tests/Doctrine/Tests/DBAL/Tools/Console/RunSqlCommandTest.php
```diff
@@ -79,7 +79,7 @@ public function testSelectStatementsPrintsResult(): void
 
     public function testUpdateStatementsPrintsAffectedLines(): void
     {
-        $this->expectConnectionExecuteUpdate();
+        $this->expectConnectionexecuteStatement();
```
greg0ire, line 82:
```suggestion
        $this->expectConnectionExecuteStatement();
```

tests/Doctrine/Tests/DBAL/Tools/Console/RunSqlCommandTest.php
```diff
@@ -90,11 +90,11 @@ public function testUpdateStatementsPrintsAffectedLines(): void
         self::assertDoesNotMatchRegularExpression('@array.*1.*@', $this->commandTester->getDisplay());
     }
 
-    private function expectConnectionExecuteUpdate(): void
+    private function expectConnectionexecuteStatement(): void
```
greg0ire, line 93:
```suggestion
    private function expectConnectionExecuteStatement(): void
```

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1294,19 +1298,44 @@ public function query()
...
+     *  - DDL statements: CREATE, DROP, ALTER, etc.
+     *  - DCL statements: GRANT, REVOKE, etc.
+     *  - Session control statements: ALTER SESSION, SET, DECLARE, etc.
+     *  - Other statements that don't yield a row set.
+     *
+     * This method supports PDO binding types as well as DBAL mapping types.
+     *
+     * @param string                 $sql    The statement SQL.
+     * @param array<mixed>           $params The query parameters.
+     * @param array<int|string|null> $types  The parameter types.
```
greg0ire, line 1319: Those periods at the end of non-sentences are enforced by our CS, right? I really don't get why we do this…

morozov: This looks more like an exception, however, I believe a full stop is enforced in exception messages. Let me see if I can get rid of them here.

greg0ire: I must have been confusing with some other coding standard

greg0ire (approved):

PR #4167, opened 2020-07-12
Remove deprecated methods from the wrapper Connection

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

#### Summary

1. Calls like `$conn->query()->fetch*()` have been reworked to `$conn->fetch*()` before the method removal.
2. The rest of the calls to `$conn->query()` have been replaced with `$conn->executeQuery()` and `$conn->executeStatement()` depending on whether the statement yields rows.
3. Calls to `$conn->exec()` and `$conn->executeUpdate()` have been replaced with `$conn->executeStatement()`.
4. The overridden versions of the `insert()`, `update()` and `delete()` methods have been removed form `PrimaryReplicaConnection` since they all are implemented via `executeStatement()` in the parent class which ensures connection to the primary instance.

**TODO**:
- [x] Unlike `PrimaryReplicaConnection::query()`, `PrimaryReplicaConnection::executeQuery()` doesn't automatically connect to the primary instance. I have no idea why the method semantics are different in the first place. It looks like a design flaw to me. This should be reflected in the deprecation note in `2.11.x`.
   > Note that `PrimaryReplicaConnection::query()` ensures connection to the primary instance while `executeQuery()` doesn't.
   > 
   > Depending on the desired behavior:
   > 
   > - If the statement doesn't have to be executed on the primary instance, use `executeQuery()`.
   > - If the statement has to be executed on the primary instance and yields rows (e.g. `SELECT`), prepend `executeQuery()` with `ensureConnectedToPrimary()`.
   > - Otherwise, use `executeStatement()`.

   Updated the upgrade notes in https://github.com/doctrine/dbal/pull/4175.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,11 @@
 # Upgrade to 3.0
 
+## BC BREAK: removed wrapper `Connection` methods
+
+The following methods of the `Connection` class have been removed:
+
+1. `query()`.
+
 ## BC BREAK: Changes in the wrapper-level API ancestry
 
 The wrapper-level `Connection` and `Statement` classes no longer implement the corresponding driver-level interfaces.
@@ -50,7 +56,6 @@ The following classes have been renamed:
 The following driver-level methods are allowed to throw a Driver\Exception:
 
 - `Connection::prepare()`
-- `Connection::query()`
 - `Connection::exec()`
 - `Connection::lastInsertId()`
 - `Connection::beginTransaction()`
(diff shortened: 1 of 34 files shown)


morozov:
@greg0ire could you take a look please?

greg0ire:
Sure! Your message ends with three backticks, did you maybe mean to write more?

greg0ire (approved):

greg0ire:
Also, do you mean to fix the todo in another PR?

morozov:
> Your message ends with three backticks, did you maybe mean to write more?

No, it was a code block originally.

> Also, do you mean to fix the todo in another PR?

Yes, it will have to be done in `2.11.x`.

PR #4169, opened 2020-07-17
Do not query result cache twice in case of cache miss

### LINKED ISSUE #4178: DDC-2043: Extra cache operation in DBAL\Cache\ResultCacheStatement.php
Jira issue originally created by user bogdan.albei:

This is the closeCursor() method in DBAL\Cache\ResultCacheStatement.php:

```
public function closeCursor()
    {
        $this->statement->closeCursor();
        if ($this->emptied && $this->data !== null) {
            $data = $this->resultCache->fetch($this->cacheKey);
            if ( ! $data) {
                $data = array();
            }
            $data[$this->realKey] = $this->data;

            $this->resultCache->save($this->cacheKey, $data, $this->lifetime);
            unset($this->data);
        }
    }
```

We are using Memcache and I noticed an extra GET operation on all cache misses. In the code above I believe the fetch call is not necessary and that the code would do the same without it. 
Also, may I ask why is the SQL used as a key in the cached data?

k0ka:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #4178 

#### Summary

If the data is not found in the query cache, it will be queried again to append a new key to it. This PR removes the second fetching by passing the first result to ResultCacheStatement.

First cache fetch at `Connection::executeCacheQuery()`: https://github.com/doctrine/dbal/blob/8c9e92ab38825be2f525d50fb1d9a0ab8ca0e1bd/lib/Doctrine/DBAL/Connection.php#L1201
Second cache fetch at `ResultCacheStatement::saveToCache()`: https://github.com/doctrine/dbal/blob/8c9e92ab38825be2f525d50fb1d9a0ab8ca0e1bd/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php#L339

### DIFF
--- a/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
+++ b/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
@@ -52,6 +52,9 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement, Result
     /** @var ResultStatement */
     private $statement;
 
+    /** @var array<string,mixed>|null */
+    private $fetchedData;
+
     /** @var array<int,array<string,mixed>>|null */
     private $data;
 
@@ -59,17 +62,19 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement, Result
     private $defaultFetchMode = FetchMode::MIXED;
 
     /**
-     * @param string $cacheKey
-     * @param string $realKey
-     * @param int    $lifetime
+     * @param string                   $cacheKey
+     * @param string                   $realKey
+     * @param int                      $lifetime
+     * @param array<string,mixed>|null $fetchedData
      */
-    public function __construct(ResultStatement $stmt, Cache $resultCache, $cacheKey, $realKey, $lifetime)
+    public function __construct(ResultStatement $stmt, Cache $resultCache, $cacheKey, $realKey, $lifetime, ?array $fetchedData = null)
     {
         $this->statement   = $stmt;
         $this->resultCache = $resultCache;
         $this->cacheKey    = $cacheKey;
         $this->realKey     = $realKey;
         $this->lifetime    = $lifetime;
+        $this->fetchedData = $fetchedData;
     }
 
     /**
@@ -336,13 +341,16 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement, Result
             return;
         }
 
-        $data = $this->resultCache->fetch($this->cacheKey);
-        if (! $data) {
-            $data = [];
+        if ($this->fetchedData === null) {
+            $this->fetchedData = $this->resultCache->fetch($this->cacheKey);
         }
 
-        $data[$this->realKey] = $this->data;
+        if (! $this->fetchedData) {
+            $this->fetchedData = [];
+        }
 
-        $this->resultCache->save($this->cacheKey, $data, $this->lifetime);
+        $this->fetchedData[$this->realKey] = $this->data;
+
+        $this->resultCache->save($this->cacheKey, $this->fetchedData, $this->lifetime);
     }
 }
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1207,10 +1207,19 @@ class Connection implements DriverConnection
             } elseif (array_key_exists($realKey, $data)) {
                 $stmt = new ArrayStatement([]);
             }
+        } else {
+            $data = [];
         }
 
         if (! isset($stmt)) {
-            $stmt = new ResultCacheStatement($this->executeQuery($sql, $params, $types), $resultCache, $cacheKey, $realKey, $qcp->getLifetime());
+            $stmt = new ResultCacheStatement(
+                $this->executeQuery($sql, $params, $types),
+                $resultCache,
+                $cacheKey,
+                $realKey,
+                $qcp->getLifetime(),
+                $data
+            );
         }
 
         $stmt->setFetchMode($this->defaultFetchMode);


lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
```diff
@@ -52,24 +52,29 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement, Result
...
-     * @param string $cacheKey
-     * @param string $realKey
-     * @param int    $lifetime
+     * @param string                   $cacheKey
+     * @param string                   $realKey
+     * @param int                      $lifetime
+     * @param array<string,mixed>|null $fetchedData
      */
-    public function __construct(ResultStatement $stmt, Cache $resultCache, $cacheKey, $realKey, $lifetime)
+    public function __construct(ResultStatement $stmt, Cache $resultCache, $cacheKey, $realKey, $lifetime, $fetchedData = null)
```
greg0ire, line 70:
```suggestion
    public function __construct(ResultStatement $stmt, Cache $resultCache, $cacheKey, $realKey, $lifetime, ?array $fetchedData = null)
```

lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
```diff
@@ -52,24 +52,29 @@ class ResultCacheStatement implements IteratorAggregate, ResultStatement, Result
...
     private $defaultFetchMode = FetchMode::MIXED;
 
     /**
-     * @param string $cacheKey
-     * @param string $realKey
-     * @param int    $lifetime
+     * @param string                   $cacheKey
+     * @param string                   $realKey
+     * @param int                      $lifetime
+     * @param array<string,mixed>|null $fetchedData
```
greg0ire, line 68: Should this new argument become mandatory in 3.0.x?

greg0ire (commented):
Please rebase, there is a conflict. Please also squash your commits.

lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
k0ka: I think, yes.
However, the class is renamed to CachingResult in 3.0

morozov: Why don't we implement a proper fix in `3.0.x` first and then backport in a BC-way? Otherwise, it's not clear what the new intended logic is.

lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
```diff
@@ -336,13 +341,16 @@ private function saveToCache(): void
...
         }
 
-        $data = $this->resultCache->fetch($this->cacheKey);
-        if (! $data) {
-            $data = [];
+        if ($this->fetchedData === null) {
+            $this->fetchedData = $this->resultCache->fetch($this->cacheKey);
+        }
+
+        if (! $this->fetchedData) {
```
morozov, line 348: Please avoid implicit conversion to boolean. It's not allowed in newer branches.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1207,10 +1207,19 @@ public function executeCacheQuery($sql, $params, $types, QueryCacheProfile $qcp)
             } elseif (array_key_exists($realKey, $data)) {
                 $stmt = new ArrayStatement([]);
             }
+        } else {
+            $data = [];
```
morozov, line 1211: Why is this needed?

morozov (changes requested):
There are three new code execution branches and not a single test. There must be a unit test that covers the logic of interaction with the external cache.

lib/Doctrine/DBAL/Connection.php
k0ka: It is needed to distinguish "data from query cache is not loaded" and "data from query cache is loaded and empty". 
This data was be set to empty array in ResultCacheStatement before - https://github.com/doctrine/dbal/blob/2.11.x/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php#L341

lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
k0ka: Ok. This line was written before me - I've just change the variable name.

https://github.com/doctrine/dbal/blob/2.11.x/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php#L340

lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
k0ka: I don't know how you releasing process is working. I consider this bug as a minor so it doesn't make sense to break backward compatibility.

morozov: Let's at least make it clear in the parameter description that this parameter will be mandatory in `3.0`. Otherwise, it's not clear why fetching the data has to be implemented twice.

k0ka: I feel confused about how to do it all in 2.11.

I have opened the new https://github.com/doctrine/dbal/pull/4189 for 3.0.x version with BC Break and mandatory $fetchedData field.

morozov: I'm not that well familiar with the design of the caching layer and I still don't get the design: instead of fetching the data from the result, why does the result now use the fetched data _and_ still can fetch from the result via `fetch*()` methods? Is this still necessary?

k0ka: The result cache is fetched in `ResultCacheStatement` only if no `fetchedData` is provided for constructor. It is done for backward compatibility - someone could use `ResultCacheStatement` before and it should continue working without the last parameter.

DBAL uses this class only in one place and it always sends `fetchedData`. So, result cache won't be fetched 2 times if DBAL is used ordinarily.

morozov: > The result cache is fetched in `ResultCacheStatement` only if no `fetchedData` is provided to the constructor.

Let's then continue the discussion in https://github.com/doctrine/dbal/pull/4189 where BC is of no concern. The new class still depends on the `Result`.

PR #4173, opened 2020-07-19
Fix Third parameter not allowed for PDO::FETCH_COLUMN

### LINKED ISSUE #3975: v2.10.2 - SQLSTATE[HY000]: General error: Extraneous additional parameters

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3975 

#### Summary

I'm getting the following error with PHP 7.4 (tested on 7.4.5 and 7.4.8):

> PDOException: SQLSTATE[HY000]: General error: Third parameter not allowed for PDO::FETCH_COLUMN in (...)/vendor/doctrine/dbal/lib/Doctrine/DBAL/Statement.php:256

This only affects `PDOStatement`, that does not seem to accept `null` anymore to skip the 3rd parameter.

This fix calls the method with 2 parameters only, when the third argument is `null`.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/ExternalPDOInstanceTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/ExternalPDOInstanceTest.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Functional;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\PDO\SQLite\Driver as PDOSqliteDriver;
+use Doctrine\DBAL\FetchMode;
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\Tests\DbalTestCase;
+use Doctrine\Tests\TestUtil;
+use PDO;
+
+class ExternalPDOInstanceTest extends DbalTestCase
+{
+    /** @var Connection */
+    protected $connection;
+
+    protected function setUp(): void
+    {
+        if (! TestUtil::getConnection()->getDriver() instanceof PDOSqliteDriver) {
+            $this->markTestSkipped('External PDO instance tests are only run on PDO SQLite for now');
+        }
+
+        $pdo = new PDO('sqlite::memory:');
+
+        $this->connection = new Connection(['pdo' => $pdo], new PDOSqliteDriver());
+
+        $table = new Table('stmt_fetch_all');
+        $table->addColumn('a', 'integer');
+        $table->addColumn('b', 'integer');
+
+        $this->connection->getSchemaManager()->createTable($table);
+
+        $this->connection->insert('stmt_fetch_all', [
+            'a' => 1,
+            'b' => 2,
+        ]);
+    }
+
+    public function testFetchAllWithOneArgument(): void
+    {
+        $stmt = $this->connection->prepare('SELECT a, b FROM stmt_fetch_all');
+        $stmt->execute();
+
+        self::assertEquals([[1, 2]], $stmt->fetchAll(FetchMode::NUMERIC));
+    }
+
+    public function testFetchAllWithTwoArguments(): void
+    {
+        $stmt = $this->connection->prepare('SELECT a, b FROM stmt_fetch_all');
+        $stmt->execute();
+
+        self::assertEquals([2], $stmt->fetchAll(FetchMode::COLUMN, 1));
+    }
+
+    public function testFetchAllWithThreeArguments(): void
+    {
+        $stmt = $this->connection->prepare('SELECT a, b FROM stmt_fetch_all');
+        $stmt->execute();
+
+        [$obj] = $stmt->fetchAll(FetchMode::CUSTOM_OBJECT, StatementTestModel::class, ['foo', 'bar']);
+
+        $this->assertInstanceOf(StatementTestModel::class, $obj);
+
+        self::assertEquals(1, $obj->a);
+        self::assertEquals(2, $obj->b);
+        self::assertEquals('foo', $obj->x);
+        self::assertEquals('bar', $obj->y);
+    }
+}
--- a/lib/Doctrine/DBAL/Statement.php
+++ b/lib/Doctrine/DBAL/Statement.php
@@ -264,7 +264,15 @@ class Statement implements IteratorAggregate, DriverStatement, Result
      */
     public function fetchAll($fetchMode = null, $fetchArgument = null, $ctorArgs = null)
     {
-        return $this->stmt->fetchAll($fetchMode, $fetchArgument, $ctorArgs);
+        if ($ctorArgs !== null) {
+            return $this->stmt->fetchAll($fetchMode, $fetchArgument, $ctorArgs);
+        }
+
+        if ($fetchArgument !== null) {
+            return $this->stmt->fetchAll($fetchMode, $fetchArgument);
+        }
+
+        return $this->stmt->fetchAll($fetchMode);
     }
 
     /**
--- a/tests/Doctrine/Tests/DBAL/Functional/StatementTestModel.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/StatementTestModel.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace Doctrine\Tests\DBAL\Functional;
+
+class StatementTestModel
+{
+    public function __construct(string $x, string $y)
+    {
+        $this->x = $x;
+        $this->y = $y;
+    }
+
+    /** @var int */
+    public $a;
+
+    /** @var int */
+    public $b;
+
+    /** @var string */
+    public $x;
+
+    /** @var string */
+    public $y;
+}


greg0ire:
Triggering a new AppVeyor build…

greg0ire (commented):
Maybe there should be a test asserting that `fetchAll` can be used without a third argument?

Also, I could swear I have seen something similar to this PR somewhere, but I can't find it… maybe you remember @morozov ?

morozov:
> Maybe there should be a test asserting that `fetchAll` can be used without a third argument?

I believe this bug is only possible with an externally initialized PDO connection which is a dirty hack deprecated in `2.11.x`. The DBAL’s wrapper over the PDO statement already does this. A test is a must.

> Also, I could swear I have seen something similar to this PR somewhere, but I can't find it… maybe you remember @morozov ?

I remember seeing something similar too. Let me try to find it.

morozov:
@greg0ire I believe you were talking about #3984.

greg0ire:
@BenMorel tell us if you need help writing the test. Also, please have a look at #3984, it handles one more case.

greg0ire:
cc @duncan3dc

BenMorel:
I'm happy to take over #3984, unless @duncan3dc completes his PR by adding tests!

duncan3dc:
@BenMorel that would be super if you have the time

BenMorel:
Alright, I took over #3984, and added tests. Actually it looks like `Statement::fetchAll()` was never tested with 2 and 3 arguments, so this is a useful addition!

My tests highlighted an issue, though: Travis [shows](https://travis-ci.org/github/doctrine/dbal/builds/710543023) that mysqli, sqlsrv and ibm_db2 seem to ignore the second parameter for `FetchMode::COLUMN`, and return the first column's value instead.

Any idea why?

morozov:
> Any idea why?

I suspect that DBAL exposes these parameters only because PDO has them and DBAL literally implements the PDO APIs by extending the corresponding classes. At the same time, the underlying drivers don't have a similar API and nobody really cared about using them. FWIW, all these parameters will be removed in `3.0.x`.

BenMorel:
Makes sense, but what should we do in the meantime? It's hard to test something that's broken!

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
```diff
@@ -50,6 +50,60 @@ public function testStatementIsReusableAfterClosingCursor(): void
         self::assertEquals(2, $id);
     }
 
+    public function testFetchAllWithOneArgument(): void
+    {
+        $table = new Table('stmt_fetch_all');
```
morozov, line 55: You don't have to create a temporary table just to fetch one row. You can try using `TestUtil::generateResultSetQuery()` instead.

morozov:
> […] what should we do in the meantime?

You don't have to cover all drivers. It's fine if you exclude the ones that don't implement the API being tested.

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
BenMorel: Where is this method located? I can't find it anywhere!

morozov: Ah, sorry. It was only introduced in `2.11.x` as part of https://github.com/doctrine/dbal/pull/4019. Feel free to leave your test as is or backport it.

BenMorel:
All good for me now, ready for review!

morozov:
The test doesn't seem to reproduce the problem being fixed:
```
$ git checkout 2.10.x
$ git checkout d23530c08a53eacff089123e87443b7224293893 -- tests/Doctrine/Tests/DBAL/Functional/StatementTest.php tests/Doctrine/Tests/DBAL/Functional/StatementTestModel.php
$ phpunit -c pdo-sqlite.phpunit.xml tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
PHPUnit 8.5.5 by Sebastian Bergmann and contributors.

.......................                                                           23 / 23 (100%)

Time: 74 ms, Memory: 12.00 MB

OK (23 tests, 34 assertions)
```

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
```diff
@@ -50,6 +54,75 @@ public function testStatementIsReusableAfterClosingCursor(): void
...
+    }
+
+    /**
+     * @depends testFetchAllWithOneArgument
+     */
+    public function testFetchAllWithTwoArguments(): void
+    {
+        $driver = $this->connection->getDriver();
+
+        if ($driver instanceof IBMDB2Driver) {
```
morozov, line 82: Should this explicitly allow only PDO-based drivers? They could be identified by the `PDOConnection`.

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
```diff
@@ -50,6 +54,75 @@ public function testStatementIsReusableAfterClosingCursor(): void
...
+        if ($driver instanceof MysqliDriver) {
+            $this->markTestSkipped('Not supported on Mysqli');
+        }
+
+        if ($driver instanceof OCI8Driver) {
+            $this->markTestSkipped('Not supported on OCI8');
+        }
+
+        if ($driver instanceof PDOOracleDriver) {
+            $this->markTestSkipped('Not supported on PDO Oracle');
```
morozov, line 95: Why not? I suspect that the test may fail because Oracle by default upper-cases the column names but it's not not supported. This is one of the reasons to not use this feature.

BenMorel: Indeed [it fails](https://github.com/doctrine/dbal/runs/904835107) with:

> Failed asserting that null matches expected 1.

What's the problem with simply skipping PDO Oracle?

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
BenMorel: Good idea. Let me know what you think of PDO Oracle and I'll refactor this.

BenMorel:
> The test doesn't seem to reproduce the problem being fixed

I'm not getting the error anymore either. I definitely was. I'm not sure what I'm missing?

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
morozov: The misleading message. This functionality is supported on `pdo_oci`.

morozov:
> I'm not sure what I'm missing?

Probably, https://github.com/doctrine/dbal/pull/4173#issuecomment-661960732.

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
BenMorel: True. What do you suggest?

BenMorel:
> Probably, #4173 (comment).

Should I create a separate test class that initializes a connection this way? The problem is, it will run whatever the configured database driver.

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
morozov: Given that you'll need to use an external PDO connection, you may try to use `PDO::ATTR_CASE` on the connection you're passing to make it convert all column names to the lower case as the portability wrapper does.

morozov:
I genuinely don't understand why you want to fix the behavior multiple aspects of which are deprecated:

1. Passing an external PDO is deprecated.
2. The object fetch mode is deprecated.

BenMorel:
> 1. Passing an external PDO is deprecated.

I know that now, but my app is heavily using PDO together with ORM/DBAL (for example through my [BulkInserter](https://github.com/brick/db) package), and that stuff would need to be refactored to use the DBAL instead, or a separate DB connection.

> 2. The object fetch mode is deprecated.

I'm only using the object fetch mode in tests. When I encountered the bug, I was simply using `PDO::FETCH_COLUMN`, which is not deprecated AFAIK, and breaks as well.

> You can try whatever approach that makes this test fail w/o the fix. We can discuss the actual implementation once it's submitted.

I will, as soon as I can.

BenMorel:
@morozov Getting back to this.

I'm now creating a `Connection` with an externally initialized `PDO` instance. I'm only doing it for SQLite, using an in-memory connection, that doesn't require any configuration. I'm detecting if we're running on the `PDOSqliteDriver`, to avoid repeating the tests for other drivers.

I'm not running this test on other PDO drivers, as creating connections with external PDO instances would require duplicating the driver's logic to create the DSN, etc., which I feel is way overkill for what should be a quick fix for an issue that won't exist in 3.0.

The test does now reproduce the problem being fixed:

```
$ git checkout 2.10.x
$ git checkout 9d08344119df519ddb76a9b0a38bbc9dccf70757 -- tests/Doctrine/Tests/DBAL/Functional/StatementTest.php tests/Doctrine/Tests/DBAL/Functional/StatementTestModel.php
$ vendor/bin/phpunit tests/Doctrine/Tests/DBAL/Functional/StatementTest.php

....................ESS                                           23 / 23 (100%)

Time: 372 ms, Memory: 8.00 MB

There was 1 error:

1) Doctrine\Tests\DBAL\Functional\StatementTest::testFetchAllWithOneArgumentOnExternalPDO
Exception: [PHPUnit\Framework\Error\Warning] PDOStatement::fetchAll(): SQLSTATE[HY000]: General error: Extraneous additional parameters
```

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
```diff
@@ -320,4 +323,65 @@ public function testFetchInColumnMode(): void
...
+        $stmt = $connection->prepare('SELECT a, b FROM stmt_fetch_all');
+        $stmt->execute();
+
+        self::assertEquals([[1, 2]], $stmt->fetchAll(FetchMode::NUMERIC));
+
+        return $connection;
+    }
+
+    /**
+     * @depends testFetchAllWithOneArgumentOnExternalPDO
```
morozov, line 358: Why does the test for one mode depend on the test for another? If it relies on the setup, then it should be extracted and shared. Ideally, it should be a separate throwaway test with a proper setup implementation.

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
```diff
@@ -320,4 +323,65 @@ public function testFetchInColumnMode(): void
 
         self::assertEquals(1, $result);
     }
+
+    public function testFetchAllWithOneArgumentOnExternalPDO(): Connection
+    {
+        $driver = $this->connection->getDriver();
+
+        if (! $driver instanceof PDOSqliteDriver) {
+            $this->markTestSkipped('External PDO connection tests run only with an SQLite configuration.');
```
morozov, line 332: Why does this test require SQLite?

morozov (commented):
You'll have to retarget this to `2.11.x` since `2.10.x` only accepts security fixes.

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
BenMorel: I've moved the tests to their own class, with proper `setUp()`.

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
BenMorel: As I mentioned in https://github.com/doctrine/dbal/pull/4173#issuecomment-697987861, I think it's overkill to test every PDO driver here, so I've set it up for in-memory SQLite only. And I thought it's more logical to only actually run this test when we're running with an SQLite config, then?

Actually, I've had to remove this check as I'm not extending `DbalFunctionalTestCase` anymore, so I don't have access to the currently configured driver. May I'd need to re-introduce a condition before running these tests, either reading from the global config, or at least check that PDO SQLite is installed?

BenMorel:
Rebased & ready for review.

greg0ire:
@BenMorel there is a CS issue

BenMorel:
@greg0ire Fixed!

greg0ire (dismissed):

BenMorel:
@greg0ire Sorry, just pushed a new commit!

greg0ire (dismissed):
We should remember to squash merge this, 16 commits is a bit much :P

bytehead (dismissed):
💪 Thanks, works in my case!

tests/Doctrine/Tests/DBAL/Functional/ExternalPDOInstanceTest.php
```diff
@@ -0,0 +1,65 @@
...
+use PDO;
+
+class ExternalPDOInstanceTest extends DbalTestCase
+{
+    /** @var Connection */
+    protected $connection;
+
+    protected function setUp(): void
+    {
+        $pdo = new PDO('sqlite::memory:');
```
morozov, line 19: > Actually, I've had to remove this check as I'm not extending DbalFunctionalTestCase anymore, so I don't have access to the currently configured driver.

What will happen here if a run the test suite with say `sqlsrv` configuration and don't have `pdo_sqlite` installed? All existing functional tests have proper checks for the installed extensions and current configuration. Why do you need it to not be a functional one?

> May I'd need to re-introduce a condition before running these tests, either reading from the global config, or at least check that PDO SQLite is installed?

I'd write a proper functional test that:
1. Skips if the currently used connection is not PDO.
2. Extract PDO from the connection.
3. Create a new connection with external PDO.
4. Proceed with testing.

tests/Doctrine/Tests/DBAL/Functional/StatementTest.php
morozov: > I think it's overkill to test every PDO driver here […]

Overkill in what? Run time, development time, maintenance? The test suite is designed to run all functional tests with all supported configurations. You don't have to reimplement that from scratch.

tests/Doctrine/Tests/DBAL/Functional/ExternalPDOInstanceTest.php
BenMorel: Good point, I forgot that `PDOConnection` *extends* `PDO` in 2.x, so I can indeed get the PDO instance from the existing `Connection` to create a new one. Done!

BenMorel:
@morozov Ready for review!

tests/Doctrine/Tests/DBAL/Functional/ExternalPDOInstanceTest.php
BenMorel: Actually, while the tests do pass, **they also pass on 2.10** now I'm afraid. This is because the PDO we extract from the Connection is a PDOConnection, not a raw PDO. So we're back to square one: there does not seem to be a way to run this test for all PDO drivers, without creating our own PDO instances, which requires duplicating the PDO creation logic from the driver (DSN etc.) in the tests.

This is what I called *overkill* in development time, for a quick fix that will be gone in 3.0 anyway.

morozov: > This is because the PDO we extract from the Connection is a PDOConnection, not a raw PDO.

Uh... you're right. This is why we're getting rid of this untestable inheritance. Please restore the previous version and squash.

> This is what I called _overkill_ in development time

I can't agree. Fixing something that is not supposed to be used may be overkill. Proper testing is not.

BenMorel: > Please restore the previous version and squash.

Done! And re-introduced the PDO SQLite driver check to only run tests on CI that target SQLite, and prevent it from running when SQLite might not be available.

morozov (approved):

morozov:
Thanks, @BenMorel.

BenMorel:
Thank you!

PR #4181, opened 2020-07-22
Fetch parent commits

greg0ire:
CI platforms do not run their jobs on the commit you push. Instead, they
run it on a merge commit so that both your changes and the latest
version of the target branch are taken into account.
Before uploading the coverage report, CodeCov's bash uploader attempts
to find the original commit hash, and succeeds on Travis and AppVeyor,
but not on Github, because the method it uses does not work unless the
parent commits are fetched too, which results in Github posting to CodeCov
with a different hash than Travis or AppVeyor.
Setting the fetch depth fixes that issue.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -121,6 +121,8 @@ jobs:
     steps:
       - name: "Checkout"
         uses: "actions/checkout@v2"
+        with:
+          fetch-depth: 2
 
       - name: "Install PHP"
         uses: "shivammathur/setup-php@v2"
@@ -163,6 +165,8 @@ jobs:
     steps:
       - name: "Checkout"
         uses: "actions/checkout@v2"
+        with:
+          fetch-depth: 2
 
       - name: "Install PHP"
         uses: "shivammathur/setup-php@v2"


morozov (approved):
Great work, @greg0ire!

morozov:
BTW, will this work if a PR has 10 commits?

greg0ire:
I think so: you want to compute the coverage for the last one and only this one.

greg0ire:
Not setting a milestone since this kind of changes is irrelevant to the end-user and should not appear in the changelog IMO. Is that how we usually handle this?

morozov:
I usually set a milestone for everything. It’s better to exclude insignificant changes from the changelog manually IMO.

greg0ire:
Ok then.

MrPetovan:
> The only situation where this would not work would be if the merge commit had more than one parent, which cannot happen IMO.

I see you aren't familiar with git's octopus merge strategy. https://git-scm.com/docs/git-merge#_merge_strategies

It isn't common, but it is possible to merge more than two parents.

greg0ire:
@MrPetovan Hey mate! I'm super familiar with it ;) In fact, I used it yesterday. What I'm saying is that it will definitely not be performed in the context of a PR, by Github, before being transmitted to CI platforms like Travis, AppVeyor or Github itself. Here is such a commit: https://github.com/doctrine/dbal/commit/7c3efc9ffbad51e7ac097bc82c195817be2fc24f

MrPetovan:
And you are absolutely correct about that in this context! I'm leaving my comment for other people who may not be as familiar as you are with it.

MrPetovan:
> Here is such a commit: [7c3efc9](https://github.com/doctrine/dbal/commit/7c3efc9ffbad51e7ac097bc82c195817be2fc24f)

Interestingly, in the GitHub UI it shows as a merged with 2 parents, one of which is also a merge with two parents.

greg0ire:
That's expected, the first one is always the commit from the head branch, which is most of the time a merge commit (at least in the dbal repo, other like the ORM might use squash merge or rebase). The second one will always be the last commit pushed on the PR branch.

PR #4180, opened 2020-07-22
Remove test that covers OCI8 classes

greg0ire:
This should result in a coverage drop.

PR #4184, opened 2020-07-23
Trigger Github workflow for merges

greg0ire:
A merge is viewed by Github as a push to the branch where the merge
occurs, see https://github.community/t/trigger-workflow-only-on-pull-request-merge/17359/2
This means we should have coverage reports for merge commits from
Github, which should result in saner numbers.

I will rebase #4180 after that.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,8 +1,7 @@
 
 name: "Continuous Integration"
 
-on:
-  pull_request:
+on: ["pull_request", "push"]
 
 jobs:
   static-analysis-phpstan:


greg0ire:
After this gets merged, we should no longer get the +3.16% coverage you can see here.

morozov (approved):

morozov:
I would never have thought that it's that easy to screw up these two highly automated technologies: GitHub Actions and Codecov.

greg0ire:
@thomasrockhu this and #4181 will be of some interest to you I think ;) There could be some improvements like warnings in the output if the parent cannot be determined in the case of #4181

thomasrockhu:
@greg0ire 👏 👏 👏 incredible. Thanks for the work here, I'll update our documentation and see what we can do in terms of error messaging.

PR #4189, opened 2020-07-24
Do not fetch result cache twice in case of cache miss

k0ka:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | #4178

#### Summary

It is https://github.com/doctrine/dbal/pull/4169 for 3.0.x version. BC Break doesn't really matters, as it's already broken by renaming of `CachingResult` class

### DIFF
--- a/src/Cache/CachingResult.php
+++ b/src/Cache/CachingResult.php
@@ -43,18 +43,23 @@ class CachingResult implements Result
     /** @var array<int,array<string,mixed>>|null */
     private $data;
 
+    /** @var array<string, mixed> */
+    private $fetchedData;
+
     /**
-     * @param string $cacheKey
-     * @param string $realKey
-     * @param int    $lifetime
+     * @param string               $cacheKey
+     * @param string               $realKey
+     * @param int                  $lifetime
+     * @param array<string, mixed> $fetchedData
      */
-    public function __construct(Result $result, Cache $cache, $cacheKey, $realKey, $lifetime)
+    public function __construct(Result $result, Cache $cache, $cacheKey, $realKey, $lifetime, array $fetchedData)
     {
-        $this->result   = $result;
-        $this->cache    = $cache;
-        $this->cacheKey = $cacheKey;
-        $this->realKey  = $realKey;
-        $this->lifetime = $lifetime;
+        $this->result      = $result;
+        $this->cache       = $cache;
+        $this->cacheKey    = $cacheKey;
+        $this->realKey     = $realKey;
+        $this->lifetime    = $lifetime;
+        $this->fetchedData = $fetchedData;
     }
 
     /**
@@ -170,14 +175,8 @@ class CachingResult implements Result
             return;
         }
 
-        $data = $this->cache->fetch($this->cacheKey);
+        $this->fetchedData[$this->realKey] = $this->data;
 
-        if ($data === false) {
-            $data = [];
-        }
-
-        $data[$this->realKey] = $this->data;
-
-        $this->cache->save($this->cacheKey, $data, $this->lifetime);
+        $this->cache->save($this->cacheKey, $this->fetchedData, $this->lifetime);
     }
 }
--- a/tests/Cache/CachingResultTest.php
+++ b/tests/Cache/CachingResultTest.php
@@ -1,0 +1,77 @@
+<?php
+
+namespace Doctrine\DBAL\Tests\Cache;
+
+use Doctrine\Common\Cache\Cache;
+use Doctrine\DBAL\Cache\CachingResult;
+use Doctrine\DBAL\Driver\Result;
+use PHPUnit\Framework\TestCase;
+
+class CachingResultTest extends TestCase
+{
+    /** @var string */
+    private $cacheKey = 'cacheKey';
+
+    /** @var string */
+    private $realKey = 'realKey';
+
+    /** @var int */
+    private $lifetime = 3600;
+
+    /** @var array<string, mixed> */
+    private $resultData = ['id' => 123, 'field' => 'value'];
+
+    /** @var array<string, mixed> */
+    private $cachedData;
+
+    /** @var Cache */
+    private $cache;
+
+    /** @var Result */
+    private $result;
+
+    protected function setUp(): void
+    {
+        $this->result = $this->createMock(Result::class);
+        $this->result->expects(self::exactly(2))
+            ->method('fetchAssociative')
+            ->will($this->onConsecutiveCalls($this->resultData, false));
+
+        $this->cache = $this->createMock(Cache::class);
+        $this->cache->expects(self::exactly(1))
+            ->method('save')
+            ->willReturnCallback(function (string $id, $data, int $ttl): void {
+                $this->assertEquals($this->cacheKey, $id, 'The cache key should match the given one');
+                $this->assertEquals($this->lifetime, $ttl, 'The cache key ttl should match the given one');
+                $this->cachedData = $data;
+            });
+    }
+
+    public function testShouldSaveResultToCache(): void
+    {
+        $cachingResult = new CachingResult(
+            $this->result,
+            $this->cache,
+            $this->cacheKey,
+            $this->realKey,
+            $this->lifetime,
+            ['otherRealKey' => 'resultValue']
+        );
+
+        do {
+            $row = $cachingResult->fetchAssociative();
+        } while ($row !== false);
+
+        $this->assertContains(
+            $this->resultData,
+            $this->cachedData[$this->realKey],
+            'CachingResult should cache data from the given result'
+        );
+
+        $this->assertEquals(
+            'resultValue',
+            $this->cachedData['otherRealKey'],
+            'CachingResult should not change other keys from cache'
+        );
+    }
+}
--- a/tests/Functional/ResultCacheTest.php
+++ b/tests/Functional/ResultCacheTest.php
@@ -225,6 +225,64 @@ class ResultCacheTest extends FunctionalTestCase
         self::assertCount(1, $layerCache->fetch('testcachekey'));
     }
 
+    public function testCacheQueriedOnlyOnceForCacheMiss(): void
+    {
+        $layerCache = $this->createMock(ArrayCache::class);
+        $layerCache->expects(self::once())
+            ->method('fetch')
+            ->willReturn(false);
+
+        $layerCache->expects(self::once())
+            ->method('save');
+
+        $result = $this->connection->executeQuery(
+            'SELECT * FROM caching WHERE test_int > 500',
+            [],
+            [],
+            new QueryCacheProfile(0, 'testcachekey', $layerCache)
+        );
+        $result->fetchAllAssociative();
+    }
+
+    public function testDifferentQueriesMayBeSavedToOneCacheKey(): void
+    {
+        $layerCache = new ArrayCache();
+
+        $this->executeTwoCachedQueries($layerCache);
+        self::assertCount(2, $this->sqlLogger->queries, 'Two queries executed');
+        self::assertCount(2, $layerCache->fetch('testcachekey'), 'Both queries are saved to cache');
+
+        $this->executeTwoCachedQueries($layerCache);
+        self::assertCount(2, $this->sqlLogger->queries, 'Consecutive queries are fetched from cache');
+
+        $layerCache->delete('testcachekey');
+        $this->executeTwoCachedQueries($layerCache);
+        self::assertCount(
+            4,
+            $this->sqlLogger->queries,
+            'Deleting one cache key leads to deleting cache for both queris'
+        );
+    }
+
+    private function executeTwoCachedQueries(ArrayCache $cache): void
+    {
+        $result = $this->connection->executeQuery(
+            'SELECT * FROM caching WHERE test_int > 500',
+            [],
+            [],
+            new QueryCacheProfile(0, 'testcachekey', $cache)
+        );
+        $result->fetchAllAssociative();
+
+        $result = $this->connection->executeQuery(
+            'SELECT * FROM caching WHERE test_int > 400',
+            [],
+            [],
+            new QueryCacheProfile(0, 'testcachekey', $cache)
+        );
+        $result->fetchAllAssociative();
+    }
+
     public function testFetchColumn(): void
     {
         $query = $this->connection->getDatabasePlatform()
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -1013,6 +1013,8 @@ class Connection
             } elseif (array_key_exists($realKey, $data)) {
                 $result = new ArrayResult([]);
             }
+        } else {
+            $data = [];
         }
 
         if (! isset($result)) {
@@ -1021,7 +1023,8 @@ class Connection
                 $resultCache,
                 $cacheKey,
                 $realKey,
-                $qcp->getLifetime()
+                $qcp->getLifetime(),
+                $data
             );
         }
 


src/Cache/CachingResult.php
```diff
@@ -43,18 +43,23 @@ class CachingResult implements Result
...
-     * @param string $cacheKey
-     * @param string $realKey
-     * @param int    $lifetime
+     * @param string               $cacheKey
+     * @param string               $realKey
+     * @param int                  $lifetime
+     * @param array<string, mixed> $fetchedData
      */
-    public function __construct(Result $result, Cache $cache, $cacheKey, $realKey, $lifetime)
+    public function __construct(Result $result, Cache $cache, $cacheKey, $realKey, $lifetime, $fetchedData)
```
morozov, line 55: See https://github.com/doctrine/dbal/pull/4169#discussion_r460100582.

k0ka: `$fetchedData` - is data, fetched from the result cache by `$cacheKey` key. It is a hash of `$realKey` to serialized `$result`.
One result cache data may store more than 1 result caches, that are distinguished by realKey. 

`CachingResult` is updating one of these keys with the underlying `Result`. 

In order to keep all other keys, it should have the full data, stored in cache. It is fetching it now
https://github.com/doctrine/dbal/blob/824d2eb0af416006f8a5317d08882ea204f608f9/src/Cache/CachingResult.php#L173
and than appends its key to fetched array
https://github.com/doctrine/dbal/blob/824d2eb0af416006f8a5317d08882ea204f608f9/src/Cache/CachingResult.php#L179

However, this array was fetched from cache before 
https://github.com/doctrine/dbal/blob/824d2eb0af416006f8a5317d08882ea204f608f9/src/Connection.php#L1006
and might be just passed to the constructor of `CachingResult`

This PR does exactly this - removes 2nd fetching from CachingResult and passes fetched data from Connection.

morozov: @k0ka thank you for the details. I'm still trying to get to the bottom of this. The last time I touched this layer, I described the responsibility of this class as following (https://github.com/doctrine/dbal/pull/4048#issuecomment-640071573):

> This class is responsible only for reading from the underlying statement and storing the result in the cache if there was no hit.

Am I wrong, or, otherwise, why does the class try to fetch from the cache if there knowingly was no hit?

morozov: In any event, as I asked earlier in https://github.com/doctrine/dbal/pull/4169#pullrequestreview-454380957, this change will require a unit test that will also help find the most optimal solution for this problem. Right now, it looks dirty that besides the `Result` that contains the fetched data, the object needs to receive the data fetched from that result.

k0ka: > 
> Am I wrong, or, otherwise, why does the class try to fetch from the cache if there knowingly was no hit?

Because one cache key may contain more than one result.

k0ka: > it looks dirty that besides the `Result` that contains the fetched data, the object needs to receive the data fetched from that result.

I have added unit and functional test to show you how it works now.

tests/Functional/ResultCacheTest.php
```diff
@@ -225,6 +225,64 @@ public function testFetchAllSavesCache(): void
...
+        $result = $this->connection->executeQuery(
+            'SELECT * FROM caching WHERE test_int > 500',
+            [],
+            [],
+            new QueryCacheProfile(0, 'testcachekey', $layerCache)
+        );
+        $result->fetchAllAssociative();
+    }
+
+    public function testDifferentQueriesMayBeSavedToOneCacheKey(): void
```
morozov, line 247: This test passes even without the fix. Is it really needed?

src/Cache/CachingResult.php
morozov: > I have added unit and functional test to show you how it works now.

Thanks for the test. I understand the logic better now.

> Because one cache key may contain more than one result.

Do you know why we need to store more than one result in a cache entry? I.e. why not store each key individually and use the cache key as a sort of namespace for hashing the keys?

tests/Functional/ResultCacheTest.php
```diff
@@ -225,6 +225,64 @@ public function testFetchAllSavesCache(): void
         self::assertCount(1, $layerCache->fetch('testcachekey'));
     }
 
+    public function testCacheQueriedOnlyOnceForCacheMiss(): void
+    {
+        $layerCache = $this->createMock(ArrayCache::class);
+        $layerCache->expects(self::once())
```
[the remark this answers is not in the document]
k0ka: It can't be a unit test as it covers the full use case. Right now, there are two invocations of `fetch` - in different places of the library. This test checks, that `fetch` from real cache is executed only once. 

In my PR it is executed in the `Connection::executeCacheQuery` method. However, it might be executed anywhere and moved somewhere else in the future.

tests/Functional/ResultCacheTest.php
k0ka: It shows how many queries may be stored in one cache key. This logic wasn't explicitly shown in other tests and you weren't aware of it, so I think it's better to leave this test. However, it might be taken out to another PR, as it doesn't directly relates to this one.

src/Cache/CachingResult.php
k0ka: > Do you know why we need to store more than one result in a cache entry?

I don't know exactly, but I think it's done to be able to easily clear cache entries. You don't have to know exact sql and all parameters to purge its result.

morozov (commented):
@doctrine/team-dbal anyone wants to take over this? I don't really understand how the caching layer is supposed to be used nor I have time/desire to learn.

tests/Functional/ResultCacheTest.php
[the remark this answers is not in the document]
k0ka: I may change it to `ArrayCache`, but:
1) it will depend on the `ArrayCache` implementation, which is a part of another project - `Doctrine/Common`.
2) essentially `ArrayCache` is a mock by itself. I can hardly imagine why this class might be used in the production environment.
3) `dbal` must accept any object, that implements `Doctrine\Common\Cache\Cache` interface and work with it. This class is external relative to `dbal`.

Functional testing performs tests of the whole library by sending input to its end-user endpoint and examining output (without checking internal structure). In our case, cache-object is just another type of the input. And we're checking whether `dbal` uses it correctly.

k0ka:
This PR is staying without review and comments for more than a month. What can I do to push it?

src/Cache/CachingResult.php
SenseException:
```suggestion
    public function __construct(Result $result, Cache $cache, $cacheKey, $realKey, $lifetime, array $fetchedData)
```

tests/Cache/CachingResultTest.php
```diff
@@ -0,0 +1,77 @@
...
+            ->will($this->onConsecutiveCalls($this->resultData, false));
+
+        $this->cache = $this->createMock(Cache::class);
+        $this->cache->expects(self::exactly(1))
+            ->method('save')
+            ->will($this->returnCallback(function ($id, $data, $ttl): void {
+                $this->assertEquals($this->cacheKey, $id, 'The cache key should match the given one');
+                $this->assertEquals($this->lifetime, $ttl, 'The cache key ttl should match the given one');
+                $this->cachedData = $data;
+            }));
```
SenseException, line 47:
```suggestion
            ->willReturnCallback(function ($id, $data, $ttl): void {
                $this->assertEquals($this->cacheKey, $id, 'The cache key should match the given one');
                $this->assertEquals($this->lifetime, $ttl, 'The cache key ttl should match the given one');
                $this->cachedData = $data;
            });
```

tests/Functional/ResultCacheTest.php
```diff
@@ -225,6 +225,64 @@ public function testFetchAllSavesCache(): void
...
+            ->method('fetch')
+            ->willReturn(false);
+
+        $layerCache->expects(self::once())
+            ->method('save');
```
SenseException, line 236: Would you recommend to check which values will be saved here?

SenseException (changes requested):
> I don't really understand how the caching layer is supposed to be used

Neither do I, but maybe I can review this based on how the cache code is working.

tests/Functional/ResultCacheTest.php
k0ka: No. It's a functional test, so it should not depend on the internal structure.

We're checking that only one fetch() is executed as it may take significant amount of time to query cache - in case of remote cache server. So from the external point of view, dbal can save anything as long as it fetches it only once per query.

SenseException (dismissed):
@doctrine/team-dbal

greg0ire:
You can drop the "retry test" commit, and should squash the 2 remaining commits together. Be aware that retrying a build can be done by closing and reopening the PR

tests/Cache/CachingResultTest.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        $this->result = $this->createMock(Result::class);
+        $this->result->expects(self::exactly(2))
+            ->method('fetchAssociative')
+            ->will($this->onConsecutiveCalls($this->resultData, false));
+
+        $this->cache = $this->createMock(Cache::class);
+        $this->cache->expects(self::exactly(1))
+            ->method('save')
+            ->willReturnCallback(function ($id, $data, $ttl): void {
```
greg0ire, line 43:
```suggestion
            ->willReturnCallback(function (string $id, $data, int $ttl): void {
```

tests/Functional/ResultCacheTest.php
morozov: > We're checking that only one fetch() is executed

This is the concern of a unit test — cover the implementation details of a given unit. A functional test cannot cover the number of calls to an idempotent method by definition.

tests/Cache/CachingResultTest.php
```diff
@@ -0,0 +1,77 @@
...
+            $this->cache,
+            $this->cacheKey,
+            $this->realKey,
+            $this->lifetime,
+            ['otherRealKey' => 'resultValue']
+        );
+
+        do {
+            $row = $cachingResult->fetchAssociative();
+        } while ($row !== false);
```
morozov, line 63: Why does a test need flow control logic? Do we not know how many rows are expected to be fetched here?

morozov (commented):
This patch requires additional work on tests. I understand that the author has put some effort in writing them and they pass and they cover the code but the way they mix the unit and functional concerns may become a problem during their maintenance.

Please have the unit and functional concerns separated.

tests/Functional/ResultCacheTest.php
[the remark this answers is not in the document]
greg0ire: What's your take on "Don't mock what you don't own?"
To me, as you pointed out in 2. `ArrayCache` is already a mock, but with a few differences:
a) it's easier to use, you don't have to write any expectations, only `new ArrayCache()`
b) it's closer to reality although it probably won't be used in production (or won't it? you don't know that, could be useful in a long-running process)
c) it is not controlled by you, and a change in it may break your tests; and if that's the case _you want to know_ you don't want to wait until production to notice it breaks.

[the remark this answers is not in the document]
k0ka: I don't get this question and how it is related to this topic.

Let me explain one more time:

`dbal` has the "contract": it will accept any object that implements `Doctrine\Common\Cache\Cache` interface and will use it independenty of it's internal structure to store and retrieve cache entries. In most unit and functional tests `Doctrine\Common\Cache\ArrayCache` is used. But it's done just for simplicity. `dbal` must be able to work with any cache provider. 

I'm mocking my own object, that implements `Doctrine\Common\Cache\Cache` interface, because in this particular case it's easier. 

Why `Doctrine\Common\Cache\ArrayCache` is harder to use? The only way to get the number of fetches from it is to retrieve statistics by the `getStats()` method. But consider running this code:

```php
$cache = new ArrayCache();
$cache->fetch('test_key');
self::assertEquals(1, $cache->getStats()[Cache::STATS_MISSES]);
```

It will fail. The right value for `Cache::STATS_MISSES` after one `fetch()` is 2. It's because of their internal structure (the first cache miss is due to fetching namespace version - but it doesn't matter here). So instead of my current test code 

```php
$layerCache->expects(self::once())->method('fetch')
```

I will have to write the following code 

```php
self::assertEquals(2, $layerCache->getStats()[Cache::STATS_MISSES]);
```

It's completely unclear why I'm using 2 here, while the whole idea of this patch is to remove the double fetch. And if they remove this "namespace version" feature in the `ArrayCache` (as it useless in this case), this test will fail. 

TL;DR:
1) This test is functional as it treats `dbal` as a whole and doesn't depend on its internal structure
2) Mock is used because it's easier to calculate number of real fetches. While `ArrayCache` is quite simple, its statistics is distorted.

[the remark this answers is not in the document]
greg0ire: Oh I see, you really need a mock since you are adding expectations to it.

> essentially ArrayCache is a mock by itself.

It's more like a stub in fact, that's what got me confused. It's easier to use as a stub, but can't be used as a mock since you can't know which methods were called,  and instead just get some distorted stats.

tests/Functional/ResultCacheTest.php
k0ka: In the wikipedia definition there is nothing about "cannot cover the number of calls": `Functional testing is a quality assurance process and a type of black-box testing that bases its test cases on the specifications of the software component under test. Functions are tested by feeding them input and examining the output`

The functional test is something that tests the product on the whole (black box), the unit test is intended to test a small part of the product knowing of its internal structure (white box).

`Doctrine\Common\Cache\Cache` is external to `dbal`. It's a complicated input. Another type of such an input is RDBMS. In most functional tests you are counting the number of sql queries that was sent to RDBMS. These sql queries are the output that you are examining.

In our case, we are concerned over the number of cache fetches. So we're counting how many times `dbal` has requested anything from the cache provider. 

Also, `Doctrine\Common\Cache\Cache::fetch()` is not idempotent - even in the simple case of `Memcached` it may be changed by any other process (however, I don't understand why idempotence does matter here).

tests/Cache/CachingResultTest.php
k0ka: In this particular test, we know that the 2nd `fetchAssociative` should be `false`. But that is not what is tested here. This block is a standard way to fetch all rows. 

I may use `Doctrine\DBAL\Driver\FetchUtils::fetchAllAssociative` instead of this block if you like.

SenseException (dismissed):

tests/Functional/ResultCacheTest.php
```diff
@@ -225,6 +225,64 @@ public function testFetchAllSavesCache(): void
...
+
+        $this->executeTwoCachedQueries($layerCache);
+        self::assertCount(2, $this->sqlLogger->queries, 'Consecutive queries are fetched from cache');
+
+        $layerCache->delete('testcachekey');
+        $this->executeTwoCachedQueries($layerCache);
+        self::assertCount(
+            4,
+            $this->sqlLogger->queries,
+            'Deleting one cache key leads to deleting cache for both queris'
```
greg0ire, line 263:
```suggestion
            'Deleting one cache key leads to deleting cache for both queries'
```

tests/Functional/ResultCacheTest.php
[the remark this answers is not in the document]
morozov: Why does interaction between the units of code (the cache and the underlying statement) need to be tested in a functional way? It's obviously a unit-level requirement, not a functional requirement.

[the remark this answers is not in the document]
k0ka: > Why does interaction between the units of code

Any interaction is an "interaction between units of code". The question is where these units of code reside. Cache is not part of DBAL.


> A cache is a non-functional unit by definition since it's not meant to change the behavior of the system. 

Cache DOES change the behavior of the system. The system is working differently depending on what the cache object returns. 

I see you don't understand (or don't want to understand) why this test is functional. I can delete it if you want. Or I can copy it to the "tests/Cache" directory.

tests/Functional/ResultCacheTest.php
greg0ire: @k0ka please don't mark conversations you did not start as resolved.

greg0ire: I can't speak for @morozov but to me, even if the cache component is external to the dbal, it's still an implementation detail and as such, is inside the black box. The user of the black box does not know about it IMO.

tests/Functional/ResultCacheTest.php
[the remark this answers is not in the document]
greg0ire: > I see you don't understand (or don't want to understand)

Please treat others more kindly, especially when they take that much time to answer to you. Assuming either ignorance or malice like this will not help.

As for your definition of behavior, it's debatable. To me, it would be more about direct inputs and outputs / the public interface of the system. Whether the system calls systems external systems to create the output should not be relevant in a functional test IMO.

tests/Functional/ResultCacheTest.php
SenseException: There already is a discussion about unit and functional attributes of `testCacheQueriedOnlyOnceForCacheMiss` above, which is why I resolve this conversation to prevent talking about similar topics in two different places.

PR #4209, opened 2020-08-17
Test MySQLi connection via TLS on Travis

### LINKED ISSUE #4122: MySQL connection via TLS is not continuously tested

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

Fixes #4122. The newly added test configuration will provide additional code coverage.

### DIFF
--- a/tests/travis/docker-run-mysql-or-mariadb.sh
+++ b/tests/travis/docker-run-mysql-or-mariadb.sh
@@ -40,3 +40,11 @@ while true; do
       ;;
   esac
 done
+
+if [[ "$TLS" == "yes" ]]
+then
+  for file in "ca.pem" "client-cert.pem" "client-key.pem"
+  do
+    docker cp "rdbms:/var/lib/mysql/$file" .
+  done
+fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -187,6 +187,9 @@ jobs:
       env: DB=mysqli.docker IMAGE=mysql:8.0
     - stage: Test
       php: 7.4
+      env: DB=mysqli-tls.docker IMAGE=mysql:8.0 TLS=yes
+    - stage: Test
+      php: 7.4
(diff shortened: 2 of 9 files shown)


tests/travis/docker-run-mysql-or-mariadb.sh
```diff
@@ -40,3 +40,11 @@ while true; do
       ;;
   esac
 done
+
+if [[ "$TLS" == "yes" ]]
+then
+  for i in "ca.pem" "client-cert.pem" "client-key.pem"
+  do
+   docker cp "rdbms:/var/lib/mysql/$i" .
```
greg0ire, line 48:
```suggestion
  for file in "ca.pem" "client-cert.pem" "client-key.pem"
  do
   docker cp "rdbms:/var/lib/mysql/$file" .
```

tests/travis/docker-run-mysql-or-mariadb.sh
```diff
@@ -40,3 +40,11 @@ while true; do
       ;;
   esac
 done
+
+if [[ "$TLS" == "yes" ]]
```
greg0ire, line 44: I noticed this in several of our packages, is that a standard way of doing this? I mean we could use set `TLS` to `true` which is a program that always succeeds, and then write:
```suggestion
if ${TLS:false}
```

greg0ire (approved):

tests/travis/docker-run-mysql-or-mariadb.sh
morozov: Honestly, I haven't ever seen the syntax like `if ${TLS:false}` being used used but have seen `if [[ "$TLS" == "yes" ]]` or even `if [[ "x$TLS" == "xyes" ]]`. AFAIK, it's all about compatibility with other shells, etc.

I don't think any of us wants to spend our time on coming up with a Bash coding standard right now :-)

PR #4212, opened 2020-08-18
Documentation on `DBAL\Driver\Statement::bindValue` is incorrect

### LINKED ISSUE #4210: Documentation on DBAL\Driver\Statement.php is incorrect
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

The documentation for [DBAL/Driver/Statement::bindValue](https://github.com/doctrine/dbal/blob/2.10.x/lib/Doctrine/DBAL/Driver/Statement.php#L26) is incorrect.

Per the [implementation](https://github.com/doctrine/dbal/blob/2.10.x/lib/Doctrine/DBAL/Statement.php#L87) this function can take a string, int, or type instance.

We should change this so that static analyzers like phpstan don't complain about this mismatch:

```
 ------ -----------------------------------------------------------------------------------------------------
  34     Parameter #3 $type of method Doctrine\DBAL\Driver\Statement::bindValue() expects int, string given.
 ------ -----------------------------------------------------------------------------------------------------

```

gjdanis:
This change updates the documentation on the `bindValue` function. Per the implementation of `DBAL\Statement` this function can take a string, int, or `Type` instance.

Fixes #4210

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #4210

#### Summary

This change updates the documentation on the `bindValue` to fix problems with static analyzers. Since this doesn't appear to be a problem in 3.x versions of this library, I've only fixed the issue in 2.10.x.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/Statement.php
+++ b/lib/Doctrine/DBAL/Driver/Statement.php
@@ -3,6 +3,7 @@
 namespace Doctrine\DBAL\Driver;
 
 use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Types\Type;
 
 /**
  * Statement interface.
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Driver/Statement.php
```diff
@@ -23,8 +23,8 @@ interface Statement extends ResultStatement
      *                          this will be a parameter name of the form :name. For a prepared statement
      *                          using question mark placeholders, this will be the 1-indexed position of the parameter.
      * @param mixed      $value The value to bind to the parameter.
-     * @param int        $type  Explicit data type for the parameter using the {@link ParameterType}
-     *                          constants.
+     * @param mixed      $type  Explicit data type for the parameter using the {@link ParameterType}
```
morozov, line 26: This change is invalid. All implementations of this interface must accept integers. A specific implementation (`Doctrine\DBAL\Statement`) accepts other types but they are not part of the interface.

gjdanis: Thanks @morozov - is the documentation incorrect then on `Doctrine\DBAL\Connection::prepare`? FWIW changing the documentation on `prepare` to return a `Doctrine\DBAL\Statement` fixes the issue for us.

morozov: It looks like it's already fixed in https://github.com/doctrine/dbal/commit/15d9be0b3c87e05f428f8c2beb1fb9555e7b75bf but not yet released.

gjdanis: Are there plans to create a release in the near future?

morozov: We're waiting for #3980 to be resolved in order to release `2.10.3`, `2.11.0`, and `3.0.0`. We can release `2.10.3` earlier but there's no reason for that right now.

PR #3905, opened 2020-03-18
Deprecate the usage of the legacy platforms and drivers

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

Fixes #3855. The actual removal in DBAL 3.0 is being backported in #3906.

PR #4213, opened 2020-08-18
Deprecate the Synchronizer package

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no

#### Summary

The `SchemaSynchronizer` API was introduced as part of the support for sharding (#162) which is now deprecated (https://github.com/doctrine/dbal/pull/3905). This code hasn't received any functional changes in the last 8 years (12f381f1254f756114267e1a1e71d0a783bfaf1c) and doesn't seem to be used by any open source projects.

According to the tests and the implementation, `SingleDatabaseSynchronizer` is just a wrapper over other schema-related APIs and doesn't provide any additional functionality.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade to 2.11
 
+## Deprecated `Synchronizer` package
+
+The `Doctrine\DBAL\Schema\Synchronizer\SchemaSynchronizer` interface and all its implementations are deprecated.
+
 ## Deprecated usage of wrapper-level components as implementations of driver-level interfaces
 
 The usage of the wrapper `Connection` and `Statement` classes as implementations of the `Driver\Connection` and `Driver\Statement` interfaces is deprecated.
--- a/docs/en/reference/sharding.rst
+++ b/docs/en/reference/sharding.rst
@@ -269,8 +269,8 @@ you have to sort the data in the application.
     $sql = "SELECT * FROM customers";
     $rows = $shardManager->queryAll($sql, $params);
 
-Schema Operations: SchemaSynchronizer Interface
------------------------------------------------
+Schema Operations: SchemaSynchronizer Interface (deprecated)
+------------------------------------------------------------
 
 Schema Operations in a sharding architecture are tricky. You have to perform
 them on all databases instances (shards) at the same time. Also Doctrine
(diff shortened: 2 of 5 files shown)


morozov:
From https://github.com/doctrine/dbal/pull/162:
> It is self-contained and does not affect the normal DBAL code.

It's a perfect example of why “let's add this code in the library because another project may need to use it” is a bad idea.

greg0ire:
There still seems to be references to this in docs: https://github.com/doctrine/dbal/blob/2.11.x/docs/en/reference/sharding.rst#schema-operations-schemasynchronizer-interface Should these docs have been removed, or maybe marked as deprecated in #3905 ? IMO we should not advertise deprecated features.

morozov:
> IMO we should not advertise deprecated features.

Well, we're not necessarily advertising them, it's just documentation. I'll make sure it's removed during the up-merge.

greg0ire:
I mean that we should maybe remove it from 2.11.x too, or at least make it clear that sharding is deprecated in the docs

greg0ire (dismissed):

morozov:
> I mean that we should maybe remove it from 2.11.x too […]

Removing documentation for the existing code looks odd to me.

> […] or at least make it clear that sharding is deprecated in the docs

We haven't been doing this for any other deprecations, so we're at least consistent here. Otherwise, there's a lot more to update than just this.

greg0ire:
Nice resource, thanks for sharing it!
After a quick look, I'd say that https://github.com/doctrine/dbal/blob/2.10.x/docs/en/reference/sharding.rst is more of a How-to guide, while
https://github.com/doctrine/dbal/blob/2.10.x/docs/design/SHARDING.md looks like an explanation, but there might be other types leaking in both documents. What would you say the right approach is for each type?

morozov:
> What would you say the right approach is for each type?

I suggest we take the same approach as for sharding: don't make any documentation changes during the deprecation (#3905) and removed the documentation during the code removal (#3906).

If we want to change this process, it's out of the scope of this issue and should be done separately.

morozov:
On the other hand, we do have some APIs (e.g. platforms) marked as deprecated in the documentation. So I went ahead and updated the documentation on `SchemaSynchronizer`.

greg0ire (approved):

PR #4172, opened 2020-07-17
Clean up inconsistent or redundant argument names

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

Fixes #4171

1. The `@group legacy` and `@expectedDeprecation` annotations are irrelevant as of https://github.com/doctrine/dbal/pull/3643.
2. The rest of the `@group` annotations are inconsistent across the test suite and aren't used anywhere.
3. Arguments have been renamed for consistency between the base declaration and extended implementation.
4. Argument names have been simplified where possible (e.g. `dropTable($tableName)` → `dropTable($name)`).
5. All references to a “field” in the context of the table schema have been replaced with a “column” as more appropriate.

PR #4215, opened 2020-08-19
Remove test group configuration leftovers

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

The `performance` test group was removed earlier in #3923 (this change should have been made at that time). All other groups were removed in #4172.

The `locking_functional` group seems to have never existed:
```
$ git log -G locking_functional -- tests/Doctrine/
```


greg0ire (approved):
> The locking_functional group seems to have never existed:

:smile:

PR #4217, opened 2020-08-20
Remove the Synchronizer package

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

The package is deprecated in https://github.com/doctrine/dbal/pull/4213.


greg0ire (commented):
> I suggest we take the same approach as for sharding: don't make any documentation changes during the deprecation (#3905) and removed the documentation during the code removal (#3906).

No docs to remove?

morozov:
Apparently, it's already removed in https://github.com/doctrine/dbal/pull/3906.

greg0ire (approved):

PR #4223, opened 2020-08-24
Manually merge 2.10.x into 2.11.x

greg0ire:
I did a little experiment in this merge up: instead of doing an evil merge, I did the required changes in a separate commit, but I move said commit before the merge, so that there is no commit that doesn't pass the build. That commit, which needs to be reviewed, is 12346644c

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,7 @@
         "phpunit/phpunit": "^8.5.5",
         "psalm/plugin-phpunit": "^0.10.0",
         "symfony/console": "^2.0.5|^3.0|^4.0|^5.0",
-        "vimeo/psalm": "^3.11.4"
+        "vimeo/psalm": "^3.14.2"
     },
     "suggest": {
         "symfony/console": "For helpful console commands such as SQL execution and import of files."
--- a/composer.lock
+++ b/composer.lock
@@ -158,16 +158,16 @@
     "packages-dev": [
         {
             "name": "amphp/amp",
-            "version": "v2.4.4",
+            "version": "v2.5.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/amphp/amp.git",
-                "reference": "1e58d53e4af390efc7813e36cd215bd82cba4b06"
+                "reference": "f220a51458bf4dd0dedebb171ac3457813c72bbc"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/amphp/amp/zipball/1e58d53e4af390efc7813e36cd215bd82cba4b06",
-                "reference": "1e58d53e4af390efc7813e36cd215bd82cba4b06",
+                "url": "https://api.github.com/repos/amphp/amp/zipball/f220a51458bf4dd0dedebb171ac3457813c72bbc",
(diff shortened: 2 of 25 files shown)


morozov:
> I did a little experiment in this merge up: instead of doing an evil merge, I did the required changes in a separate commit, but I move said commit before the merge, so that there is no commit that doesn't pass the build.

I’ve been doing the same for some time as well: if a change that is needed for the merge build to pass can be made in the target branch prior to the merge, I'll make it there. Although, depending on the number of changes, conflicts, etc. this process may be quite non-straightforward. Any know-how from you @greg0ire?

morozov (approved):

greg0ire:
> Any know-how from you @greg0ire?

Not really, I did the merge, fixed the issues, then rebased on 2.11.x and re-did the merge… feels a bit clumsy and might not work every time indeed, but glad to see we had the same idea :)

morozov:
What I do is:

1. Do the merge, resolve conflicts, and make all the needed changes to make the build pass to in the merge commit.
2. Once I realize that some changes should be made prior to the merge, I make another clone of the repo (just by copying the project).
3. Switch back to the target branch and port all the needed changes from the merge commit.
4. Do the merge again but instead of resolving conflicts, just copy the entire tree from the clone.
5. Done.

In general, the clone and copy approach could be replaced with `git checkout <merge-sha> -- .` but it doesn't delete the files that should no longer exist in the tree. This is the clumsy part in my case..

greg0ire:
I have the same issue with 3.0.x, so I will try again and pay more attention to see if `git rerere` kicks in… I think it didn't last time, but I'm not 100% sure. If it does, it should could help. If it doesn't that would be because the parent commit differ but I don't think that should be an issue for `git rerere`

greg0ire:
It does kick in:

```
Resolved 'composer.lock' using previous resolution.
Resolved 'src/Connections/PrimaryReadReplicaConnection.php' using previous resolution.
Resolved 'src/Platforms/SQLServer2012Platform.php' using previous resolution.
```

morozov:
Why do you need to re-resolve the same conflict more than once? Didn't this conflict exist between say 2.10 and 2.11 and it doesn't exist between 2.11 and 3.0?

greg0ire:
I proceed like this:
1. merge and fix obvious conflicts
2. notice more things need to be fixed for Psalm to run cleanly
3. fix those and commit
4. rebase on the target branch, dropping everything but that last commit. At that point, I have lost the merge commit
5. run the merge again, run into the same conflict, let rerere resolve them

> Didn't this conflict exist between say 2.10 and 2.11 and it doesn't exist between 2.11 and 3.0?

The conflict is different that conflicts between 2.10 and 2.11 indeed

morozov:
> run the merge again, run into the same conflict, let rerere resolve them

This is what I've been fixing just by copying the tree. By the time the same conflict happens again, I already know what the expected result is, regardless of whether `rerere` kicks in.

greg0ire:
Might be better because `rerere` does not seem to record deletion, which I had to do again with `git mergetool`. That was done quickly, but still…

PR #4226, opened 2020-08-25

greg0ire:


src/Platforms/AbstractPlatform.php
```diff
@@ -731,18 +731,21 @@ public function getNowExpression(): string
...
+     * SQLite only supports the 2 parameter variant of this function.
+     *
+     * @param string   $string An sql string literal or column name/alias.
+     * @param int      $start  Where to start the substring portion.
+     * @param int|null $length The substring portion length.
+     *
+     * @return string
      */
-    public function getSubstringExpression(string $string, string $start, ?string $length = null): string
+    public function getSubstringExpression($string, $start, $length = null)
```
greg0ire, line 742: Ah I think this is where all the issues come from, isn't it? :sweat_smile: bad merge resolution.

greg0ire: We don't, that's why I say "bad merge resolution", and why this diff is marked as outdated :smile: sorry for the confusion.

morozov: Sorry, I didn't notice the Outdated pill.

greg0ire: No worries :)

morozov (approved):

PR #4227, opened 2020-08-25
Use an include list for export-ignore'd stuff

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

Using an exclude list involves more maintenance: you have to think of
adding new files/directories to the list, and you have to think of
removing them from the list.

Note that this effectively removes `psalm.xml`, `README.md`, `UPGRADE.md` and `SECURITY.md` from the generated archive, which can be changed of course.

The result can be seen [on my fork](https://github.com/greg0ire/dbal/releases/tag/gitattributes_simplified), and should be compared with https://github.com/doctrine/dbal/releases/tag/2.10.2

When merging up, we will have to be careful to replace `lib` with `src`, otherwise, no more source code, which wouldn't look too good.


.gitattributes
```diff
@@ -1,14 +1,7 @@
...
-/composer.lock export-ignore
-/docs export-ignore
-/phpcs.xml.dist export-ignore
-/phpstan.neon.dist export-ignore
-/phpunit.xml.dist export-ignore
-/run-all.sh export-ignore
-/tests export-ignore
+/* export-ignore
+/bin -export-ignore
+/bin/** -export-ignore
```
morozov, line 3: Are there two expressions per directory required? What happens if we remove the first one?

morozov (commented):
> The result can be seen [on my fork](https://github.com/greg0ire/dbal/releases/tag/gitattributes_simplified), and should be compared with https://github.com/doctrine/dbal/releases/tag/2.10.2

As far as I can tell, this is how GitHub interprets `.gitattributes` which is not necessarily the case for Packagist or Satis (I know there is some discrepancy between how the two handle these rules).

While I absolutely agree with this change, I believe it could use some more testing.

.gitattributes
greg0ire: Yes it is required, there will  not be a `bin` directory if you include the first one, because that's equivalent to saying `bin` is ignored for the export, although every file in it is isn't. That's at least what I observed when I tried.

greg0ire:
> As far as I can tell, this is how GitHub interprets .gitattributes which is not necessarily the case for Packagist or Satis (I know there is some discrepancy between how the two handle these rules).

Not sure if that's really GitHub that is at work here, IMO it just delegates to `git archive`. Besides, how do Packagist or Satis relate to this? Do they take `.gitattributes` into account for anything?

morozov:
> IMO it just delegates to `git archive`.

I believe this is how it should work but see https://github.com/composer/satis/issues/471.

greg0ire:
> Also, Composer\Package\Archiver\GitExcludeFilter::parseGitAttributesLine().

Was just going to post it, indeed they have that functionality…

morozov:
Regardless of the outcome of https://github.com/composer/composer/issues/9153, this change may affect those poor souls who use self-hosted Satis and will receive a broken package that may meet their `2.10.x` version constraint.

greg0ire:
Yeah… let's close I guess :disappointed:

greg0ire:
Cc @VincentLanglet

Issue #4225, opened 2020-08-25
Sporadic test failure on PDO_OCI


morozov:
```
There was 1 error:

1) Doctrine\DBAL\Tests\Functional\Schema\OracleSchemaManagerTest::testDropsDatabaseWithActiveConnections
Exception: [Doctrine\DBAL\Exception\DriverException] An exception occurred while executing 'DROP USER test_drop_database CASCADE':

SQLSTATE[HY000]: General error: 1940 OCIStmtExecute: ORA-01940: cannot drop a user that is currently connected
 (/opt/oracle/php-src-php-7.4.9/ext/pdo_oci/oci_driver.c:348)
```

According to the test name, it is meant to drop a database with active connections but under certain circumstances, it fails to do so.

At the same time, it's not clear why this functionality is needed in the first place.

greg0ire:
That functionality was introduced in https://github.com/doctrine/dbal/pull/849 
@deeky666 , maybe you remember why this was needed?

morozov:
> This PR closes all active connections before dropping the database if dropping the database failed before.

IMO this is a fundamentally wrong approach, even from the DBAL's own perspective. If DBAL loses its connection, it will attempt to recreate it again before the next query. If a service that uses the database being dropped gets its connection closed, it will attempt to restore it and proceed thereby interfering with the process of dropping the database.

A proper approach is to either have a maintenance window during the database is inaccessible to its clients or perform a rolling migration.

greg0ire:
Should we drop the flaky test? It's annoying to have to restart the builds every time.

morozov:
I can mark it as incomplete for Oracle in https://github.com/doctrine/dbal/pull/4230 based on your earlier findings.

PR #4230, opened 2020-08-26
Deprecate the functionality of dropping client connections when dropping a database

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no

See https://github.com/doctrine/dbal/issues/4225#issuecomment-680654156.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,10 @@
 # Upgrade to 2.11
 
+## Deprecated the functionality of dropping client connections when dropping a database
+
+The corresponding `getDisallowDatabaseConnectionsSQL()` and `getCloseActiveDatabaseConnectionsSQL` methods
+of the `PostgreSqlPlatform` class have been deprecated.
+
 ## Deprecated `Synchronizer` package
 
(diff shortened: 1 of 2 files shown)


UPGRADE.md
```diff
@@ -1,5 +1,10 @@
 # Upgrade to 2.11
 
+## Deprecated the functionality of dropping client connections when dropping a database
+
+The corresponding `getDisallowDatabaseConnectionsSQL()` and `getCloseActiveDatabaseConnectionsSQL` methods
+of the `PostgreSqlPlatform` class have been deprecated.
```
greg0ire, line 6: Why only Postgres? #4225 is about Oracle.

morozov: The deprecation covers all platforms. Specifically, the ones that override `SchemaManager::dropDatabase()` and implement error handling there. Those include Oracle, Postgres, and others.

But only Postgres has public methods that are used in such implementation. Those methods are also deprecated.

greg0ire (dismissed):

greg0ire (approved):

PR #4236, opened 2020-08-27
Invalid sp_rename with case sensitive DB (sp_RENAME -> sp_rename)

redgnar:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

In MSSQLs DBs that use case sensitive collation, the command sp_RENAME will not work. It should be sp_rename, it is proper command name

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
         "doctrine/coding-standard": "^8.1",
         "jetbrains/phpstorm-stubs": "^2019.1",
         "nikic/php-parser": "^4.4",
-        "phpstan/phpstan": "^0.12.31",
+        "phpstan/phpstan": "^0.12.40",
         "phpunit/phpunit": "^8.5.5",
         "psalm/plugin-phpunit": "^0.10.0",
         "symfony/console": "^2.0.5|^3.0|^4.0|^5.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "c7a5a6e253d55b762c8113554a570902",
+    "content-hash": "0f85745785af604940cdf6c65474abe0",
     "packages": [
         {
             "name": "doctrine/cache",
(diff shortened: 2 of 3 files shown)


greg0ire:
Please associate your email address with your Github account, or change the

greg0ire:
> It should be sp_rename, it is proper command name


Source: https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-rename-transact-sql?view=sql-server-ver15

greg0ire (dismissed):

redgnar:
Ok 👍

morozov (changes requested):
This change requires steps to reproduce and an integration test that fails without the fix. This way, we'll make sure that DBAL is tested with this previously untested configuration and likely find other similar issues.

If a test is impossible to implement in the current environment, consider adding another AppVeyor or Travis build.

redgnar:
> This change requires steps to reproduce and an integration test that fails without the fix. This way, we'll make sure that DBAL is tested with this previously untested configuration and likely find other similar issues.
> 
> If a test is impossible to implement in the current environment, consider adding another AppVeyor or Travis build.

Such test requires mssql instance with CS collation for example Polish_CS_AS. If you have such db you can change something like column:

`create table example
(
	id nvarchar(50) not null
		primary key,
	testcol nvarchar(50)
)
go`

`exec sp_RENAME 'example.testcol', testcol_new, 'COLUMN'
go`

You will receive error that there is no procedure sp_RENAME.

Really I don't know how to reproduce it in AppVeyor or Travis. I don't know how you handle it in dbal (calling test on real db automatically). Do you have some manual or can you help me with that ?

redgnar:
I think you can reproduce this error by changing collation of doctrine_tests db to Latin1_General_100_CS_AS

redgnar:
> This change requires steps to reproduce and an integration test that fails without the fix. This way, we'll make sure that DBAL is tested with this previously untested configuration and likely find other similar issues.
> 
> If a test is impossible to implement in the current environment, consider adding another AppVeyor or Travis build.

I think you can reproduce this error by changing collation of doctrine_tests db to Latin1_General_100_CS_AS

greg0ire:
How does one do that? Should it be done here: https://github.com/doctrine/dbal/blob/2.10.x/tests/travis/install-mssql.sh

Or should the config block that uses that file be duplicated and an `configure-mssql-cs.sh` be created, so that we have one job with a CS db, another with a CI db?

redgnar:
There is docker env MSSQL_COLLATION. You can set it to "Latin1_General_100_CS_AS". It should be enought to reproduce error.

greg0ire:
Ok so maybe you can duplicate this block:
https://github.com/doctrine/dbal/blob/1efdc639b66392c396937775b4f426e129688462/.travis.yml#L214-L221

And set that variable in the environment, and in the script, pass it to docker if it is defined?

redgnar:
And add it to this PR or create new one ?

redgnar:
I can see there are other problems with CS as well

redgnar:
I fixed other issues with CS. SysObjects and SysColumns. Should be sysobjects and syscolumns

greg0ire:
The code looks good, but you have 3 commits that are all labelled the same way. I think there should be maybe 2 commits, one with the renames (plural), and the second that would setup the new build job.

PR #4237, opened 2020-08-29
2.10.x reproduce cs error

redgnar:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

Reproducing CS error on MSSQL - with sp_RENAME


greg0ire:
Nice! Looks like the env var works, and you found other issues!

Issue #3038, opened 2018-03-07
[Question] Partial Indexes on SQL Server


fabiang:
Hallo,

Is there any reason why the platform definition for SQL Server returns false for `supportsPartialIndexes()`? As far as I can see SQL Server [supports them since 2008](https://www.mssqltips.com/sqlservertip/1785/sql-server-filtered-indexes-what-they-are-how-to-use-and-performance-advantages/). Microsoft just named them "filtered indexes".

What confuses me is that new indexes are always created with `WHERE fieldname IS NOT NULL`, where from the code the WHERE clause should only be added when `supportsPartialIndexes()` is true.

Thanks

Ocramius:
@fabiang they aren't supported by our platform implementation: if you can apply the partial filter tests existing for the PostgreSQL platform also to SQL Server, then please do open a PR with the added tests and this one method changed.

PR #4240, opened 2020-08-31
Added partial index support for SQL Server and SQLite

cziegenberg:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #3038

#### Summary

Added support for partial indexes in SQL Server (supported since version 2008) and SQLite (supported since version 3.8.0). Also had to adjust the unique constraint generation in SQL Server so that the WHERE part is correctly added.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServer2008Platform.php
@@ -11,6 +11,15 @@ namespace Doctrine\DBAL\Platforms;
 class SQLServer2008Platform extends SQLServer2005Platform
 {
     /**
+     * {@inheritdoc}
+     */
+    public function supportsPartialIndexes()
+    {
+        // Partial indexes are supported as "filtered indexes" since version 2008.
+        return true;
+    }
+
+    /**
      * {@inheritDoc}
      */
     public function getListTablesSQL()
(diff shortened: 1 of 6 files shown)


morozov (changes requested):
The code looks good so far but it seems to be not covered by any functional tests meaning that there's no guarantee that this code actually works.

There is a similar test for PostgreSQL:
https://github.com/doctrine/dbal/blob/3f3551aadb20c7ed58ee9c7f06ed74770588272f/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php#L355

It could be moved up to the parent class and executed against all platforms that support partial indexes.

Additionally, please retarget against `2.11.x`. The `2.10.x` branch only accepts bug fixes.

masacc:
Hello, any updates on this PR ? Thank you

cziegenberg:
I'm still working on it, but there is a lot to do to do that correctly and there are many open questions.

1) There is a problem when saving a partial index and reading it again. When it's saved, it's optimized by the database, so the input doesn't equal the output (in realistic use cases). The current test for Postgres is a very simple example to make the test work, but with real code, especially to support multiple databases, it would always fail. The index would always be detected as changed.

2) This gets even worse with SQL Server, because there the indexes are always partial indexes, because there are some conditions added to them automatically. (Why? That's not documented. And if I understand it correctly, this should not work with Version 2005, although DBAL tries to support it...) Even if you know that, some of these added conditions may be duplicated, because they are part of the own conditions, and then they are removed by the database.

So how to compare the index conditions correctly? Save a comment with the original index condition? Or do some normalization before comparing the conditions (which doesn't seem to exist in DBAL)?

I will close this now, but it would be great to get some input regarding the points above - especially why the indexes are automatically modified in SQL Server (to perhaps remove this).

Issue #1411, opened 2012-03-08
DBAL-235: Column order is not synchronized by Comparator


doctrinebot:
Jira issue originally created by user artem:

Comparator does not synchronize column order. It would be great if it did.

mvorisek:
Column order is important when doiing table scan of large tables, columns at the beginning are faster to retrive.

Another reason is much easier diff & backup.

I belive, this should be fixed even if for MySQL only.

For other tables, we may do something like:
Postgresql: https://stackoverflow.com/questions/285733/how-do-i-alter-the-position-of-a-column-in-a-postgresql-database-table#39781155 (with full table lock first)
Sqlite: https://stackoverflow.com/questions/20574113/sqlite3-how-to-reorder-columns-in-a-table#20574263

mfn:
> columns at the beginning are faster to retrive

Do you have some further information regarding this?

mvorisek:
> 
> 
> > columns at the beginning are faster to retrive
> 
> Do you have some further information regarding this?

personal experience, I belive it is true when the fetched column has one or more variable column before

PR #4242, opened 2020-09-02
Update column position on schema update for MySQL

mvorisek:
finish after 4276

|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #1411

#### Summary

This PR adds support to Comparator to detect columns that needs to be reordered.

Please help with the MySQL alter table impl., ie. to generate `FIRST` or `AFTER` after each alter column sql.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -3367,6 +3367,14 @@ abstract class AbstractPlatform
     }
 
     /**
+     * Does this platform support column reordering in alter table?
+     */
+    public function supportsUpdateColumnOrder(): bool
+    {
+        return false;
+    }
+
+    /**
      * Gets the format string, as accepted by the date() function, that describes
      * the format of a stored datetime value of this platform.
      *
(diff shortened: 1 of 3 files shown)


lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -3366,6 +3366,14 @@ public function supportsColumnCollation()
         return false;
     }
 
+    /**
+     * Does this platform support column reordering in alter table?
+     */
+    public function supportsUpdateColumnOrder(): bool
```
mvorisek, line 3372: or `supportsColumnRerder`?

mvorisek:
@greg0ire @morozov please advise if this change is welcomed.

To finish this I think we need:
- pass previous column name to a column info created in `Table::getColumns()`
- later we can add to orm `orm:schema-tool:update` `--full-update` which will always generate full ALTER CHANGE query - I did tests and at least MySQL processes it instantly if the DB schema is already fully up to date - this way, we can cover DB specifics like column order or collation which are currently not detected by the diff in orm

greg0ire:
I have no objections to this feature being added.

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
morozov: How is it supposed to be used? The approach of adding `supports*` methods implies that the actual platform-specific logic will be implemented outside of the platform in the calling code. It might be better to define the API in terms of actions, e.g. `reorderColumns()`.

PR #4246, opened 2020-09-02
SQLite: Fix wrongly detected reference constraint name on schema change

MyIgel:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no?
| Fixed issues | #4243

#### Summary
This fixes #4243 as other parts of the code expect the constraint name to be `null`, for example `\Doctrine\DBAL\Schema\ForeignKeyConstraint::__construct` if no actual name exists

PR #4247, opened 2020-09-02
SQLite: Escape numbers as keywords

### LINKED ISSUE #4243: SQLite Foreign Key Behavior BC Break
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.10.3

#### Summary

A change in SQLite foreign key constraint behavior is a breaking change that was released as a patch release.

See: https://github.com/laravel/framework/issues/34093

MyIgel:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #4243

#### Summary
This fixes #4243 by escaping numbers

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/Keywords/SQLiteKeywords.php
+++ b/lib/Doctrine/DBAL/Platforms/Keywords/SQLiteKeywords.php
@@ -2,6 +2,8 @@
 
 namespace Doctrine\DBAL\Platforms\Keywords;
 
+use function is_numeric;
+
 /**
  * SQLite Keywordlist.
  */
@@ -13,6 +15,14 @@ class SQLiteKeywords extends KeywordList
     public function getName()
     {
         return 'SQLite';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isKeyword($word)
+    {
+        return parent::isKeyword($word) || is_numeric($word);
     }
 
     /**


lib/Doctrine/DBAL/Platforms/Keywords/SQLiteKeywords.php
```diff
@@ -15,6 +15,14 @@ public function getName()
         return 'SQLite';
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function isKeyword($word)
+    {
+        return parent::isKeyword($word) || is_numeric($word);
```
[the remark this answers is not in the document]
MyIgel: Fixing it directly is faster than starting the container ;)

lib/Doctrine/DBAL/Platforms/Keywords/SQLiteKeywords.php
```diff
@@ -15,6 +17,14 @@ public function getName()
...
```
morozov, line 25: Is this an alternative to #4246? `5` doesn't look like an SQLite keyword, so this change doesn't look right. As pointed out earlier, the problem is that an invalid key name is generated in the first place, not that it's not escaped.

MyIgel: Yes and no, no one prevents a user to use numbers as FK names so it can break for them in the future. The reason its here is to not clutter \Doctrine\DBAL\Schema\AbstractAsset::getQuotedName with some numerical checks, esp. as i don't know how other databases handle this and at the end this was the fastest (aka least code change/restructuring) change

morozov: We're currently reevaluating the approach to quoting names leaning towards the user being responsible for quoting since the existing magic is unreliable anyways.

> no one prevents a user to use numbers as FK names so it can break for them in the future

I believe it never worked, so it's a good enough preventing measure.

Regardless of that, it doesn't look like a bugfix so `2.10.x` is not the right place for this change.

MyIgel: At least it would be the smallest hotfix for `2.10.x` as its currently broken for anyone changing tables so i would consider it a bugfix? But i totally agree with the point that it definitely does not feel right to add it there (but again, its the "minimum change"/the change that i am currently comfortable to do in this project :grin:)

beberlei:
This is a band-aid and its better to fix the root cause see #4246 and #4256

PR #4248, opened 2020-09-02
MasterSlaveConnection::query() throws driver-specific exceptions (e.g. PDOException etc)

supersmile2009:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no, but it depends*
| Fixed issues | #2769, #3118

#### Summary
This PR fixes how driver exceptions are handled in Doctrine/DBAL/Connections/MasterSlaveConnection::query() and prevents driver-specific exceptions from bubbling. Driver exception conversion was added in https://github.com/doctrine/dbal/commit/4cc70c177bfe4872f446b7de3f13e4df24ad9c8a, but somehow `MasterSlaveConnection` was forgotten.
This PR reduces code duplication by calling the parent method, so no more chance of forgetting to apply some change to `MasterSlaveConnection` if `Doctrine\DBAL\Connection::query()` is modified. At the end of the day except for the proper exception handling parent's method behavior is the same.

#### Alternative PRs

Unlike #3313, this PR addresses only the core issue - exception not being converted. That PR suggests catching `PDOException` in addition to `DBALException`, which is completely unnecessary if exception is properly converted. Also unlike #3313, this PR adds a localized unit test, while the other PR tests merely a side-effect on the `ping` method.

#### BC Break
Any bug fix is a change of behavior, which can be a potential BC break for someone. Any of the 3rd party code that relies on current buggy behavior which in fact violates the contract would get broken by this change.
So in practice it's a BC break for some projects that probably tried to implement some workarounds for this bug. But ignoring exploitation of undocumented/buggy behavior when making stuff work according to the spec is a common thing when evaluating if a change breaks BC. The only exception that is supposed to be thrown out of the `query` method is `DBALException` as advertised in the docblock. Therefore I wouldn't consider it a BC break.

#### Other
I can see 2.11.x branch has some significant changes around the code affected by this PR. I could create a separate PR for 2.11 if you consider this solution good enough.

### DIFF
--- a/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
+++ b/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php
@@ -12,7 +12,6 @@ use Doctrine\DBAL\Events;
 use InvalidArgumentException;
 
 use function array_rand;
-use function assert;
 use function count;
 use function func_get_args;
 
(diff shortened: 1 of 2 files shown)


morozov:
@beberlei, as the one who insisted on keeping this functionality in the library, please provide your review.

beberlei:
Superseded by https://github.com/doctrine/dbal/pull/4456

PR #4253, opened 2020-09-04
Deprecate DBAL\DBALException in favor of DBAL\Exception

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no

Similarly to the deprecation of `Driver\DriverException` in favor of `Driver\Exception` #4100, the `DBAL\DBALException` name is unnecessarily verbose and is being deprecated in favor of `DBAL\Exception`.

There's no BC break according to `roave-backward-compatibility-check`.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -20,7 +20,6 @@ use Doctrine\DBAL\Query\Expression\ExpressionBuilder;
 use Doctrine\DBAL\Query\QueryBuilder;
 use Doctrine\DBAL\Schema\AbstractSchemaManager;
 use Doctrine\DBAL\Types\Type;
-use Exception;
 use Throwable;
 use Traversable;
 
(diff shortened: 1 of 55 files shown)


SenseException (changes requested):
Is there a reason why `OracleSchemaManager` and `SQLServerSchemaManager` still catches `DBALException`?

morozov:
> Is there a reason why `OracleSchemaManager` and `SQLServerSchemaManager` still catches `DBALException`?

Yes, BC. They may be extended and throwing the old exception.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1481,10 +1478,6 @@ public function transactional(Closure $func)
             $this->commit();
 
             return $res;
-        } catch (Exception $e) {
-            $this->rollBack();
-
-            throw $e;
```
greg0ire, line 1487: Good catch! :drum:

greg0ire (approved):

SenseException (approved):

PR #4255, opened 2020-09-05
Revert full support for foreign key constraints for SQLite

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

Fixes #4243.

PR #4256, opened 2020-09-06
[GH-4243] more sqlite schema manager fixes

beberlei:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #4243 

#### Summary

The previous fix #4246 did not fully fix the bug, because it missed handling of `constraint_name` == NULL in a downstream function. This lead to foreign keys without name grouped together.

Patch includes one regression example copied from Taylor in the original Laravel issue, but could use another one from @greg0ire example with the documents table, pending the original SQL to create it from Laravel blueprint.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Comparator.php
+++ b/lib/Doctrine/DBAL/Schema/Comparator.php
@@ -14,6 +14,7 @@ use function array_unique;
 use function assert;
 use function count;
 use function get_class;
+use function strlen;
 use function strtolower;
 
 /**
@@ -284,12 +285,19 @@ class Comparator
             foreach ($toFkeys as $key2 => $constraint2) {
                 if ($this->diffForeignKey($constraint1, $constraint2) === false) {
                     unset($fromFkeys[$key1], $toFkeys[$key2]);
-                } else {
-                    if (strtolower($constraint1->getName()) === strtolower($constraint2->getName())) {
-                        $tableDifferences->changedForeignKeys[] = $constraint2;
-                        $changes++;
-                        unset($fromFkeys[$key1], $toFkeys[$key2]);
-                    }
+                    continue 2;
+                }
+
+                // only named foreign key constraints can change, otherwise it appears as removed+added.
+                if (
+                    strlen($constraint1->getName()) > 0 &&
+                    strlen($constraint2->getName()) > 0 &&
+                    strtolower($constraint1->getName()) === strtolower($constraint2->getName())
+                ) {
+                    $tableDifferences->changedForeignKeys[] = $constraint2;
+                    $changes++;
+                    unset($fromFkeys[$key1], $toFkeys[$key2]);
+                    continue 2;
                 }
             }
         }
--- a/lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
@@ -408,8 +408,8 @@ class SqliteSchemaManager extends AbstractSchemaManager
         $list = [];
         foreach ($tableForeignKeys as $value) {
             $value = array_change_key_case($value, CASE_LOWER);
-            $name  = $value['constraint_name'];
-            if (! isset($list[$name])) {
+            $id    = $value['id'];
+            if (! isset($list[$id])) {
                 if (! isset($value['on_delete']) || $value['on_delete'] === 'RESTRICT') {
                     $value['on_delete'] = null;
                 }
@@ -418,8 +418,8 @@ class SqliteSchemaManager extends AbstractSchemaManager
                     $value['on_update'] = null;
                 }
 
-                $list[$name] = [
-                    'name' => $name,
+                $list[$id] = [
+                    'name' => $value['constraint_name'],
                     'local' => [],
                     'foreign' => [],
                     'foreignTable' => $value['table'],
@@ -430,8 +430,8 @@ class SqliteSchemaManager extends AbstractSchemaManager
                 ];
             }
 
-            $list[$name]['local'][]   = $value['from'];
-            $list[$name]['foreign'][] = $value['to'];
+            $list[$id]['local'][]   = $value['from'];
+            $list[$id]['foreign'][] = $value['to'];
         }
 
         $result = [];
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
@@ -257,4 +257,62 @@ SQL;
             ->getColumn('col1')
             ->getComment());
     }
+
+    /**
+     * This test requires specific SQLite code to trigger the error and could
+     * not be added in a more general way to the base SchemaManagerFunctionalTestCase
+     * to make sure the regression does not happpen on other databases that
+     * do not have support for creating and dropping foreign keys. We do not have
+     * another database at the moment other than SQLite without this support at
+     * the moment however.
+     */
+    public function testUnnamedForeignKeyConstraintHandlingRegression(): void
+    {
+        $sql = <<<SQL
+DROP TABLE IF EXISTS "unnamedfk_referenced1";
+DROP TABLE IF EXISTS "unnamedfk_referenced2";
+DROP TABLE IF EXISTS "unnamedfk_referencing";
+CREATE TABLE "unnamedfk_referenced1" (
+"id1" integer not null,
+"id2" integer not null,
+primary key("id1", "id2")
+);
+CREATE TABLE "unnamedfk_referenced2" (
+"id" integer not null primary key autoincrement
+);
+CREATE TABLE "unnamedfk_referencing" (
+    "id" integer not null primary key autoincrement,
+    "reference1_id1" integer not null,
+    "reference1_id2" integer not null,
+    "reference2_id" integer not null,
+    foreign key("reference1_id1", "reference1_id2") references "unnamedfk_referenced1"("id1", "id2") on delete cascade,
+    foreign key("reference2_id") references "unnamedfk_referenced2"("id") on delete cascade
+)
+SQL;
+
+        $this->connection->exec($sql);
+
+        $sm          = $this->connection->getSchemaManager();
+        $onlineTable = $sm->listTableDetails('unnamedfk_referencing');
+
+        $offlineTable = new Table('unnamedfk_referencing');
+        $offlineTable->addColumn('id', 'integer');
+        $offlineTable->addColumn('reference1_id1', 'integer');
+        $offlineTable->addColumn('reference1_id2', 'integer');
+        $offlineTable->addColumn('reference2_id', 'integer');
+
+        $comparator = new Schema\Comparator();
+        $diff       = $comparator->diffTable($offlineTable, $onlineTable);
+
+        $sqls = $this->connection->getDatabasePlatform()->getAlterTableSQL($diff);
+
+        foreach ($sqls as $sql) {
+            $this->connection->exec($sql);
+        }
+
+        $onlineTableAfter = $sm->listTableDetails('unnamedfk_referencing');
+        $diff             = $comparator->diffTable($onlineTable, $onlineTableAfter);
+
+        $this->assertFalse($diff);
+    }
 }
--- a/tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
+++ b/tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
@@ -1317,4 +1317,36 @@ class ComparatorTest extends TestCase
         self::assertCount(1, $actual->changedTables['table2']->addedForeignKeys, 'FK to table3 should be added.');
         self::assertEquals('table3', $actual->changedTables['table2']->addedForeignKeys[0]->getForeignTableName());
     }
+
+    public function testCompareForeignKeysOfTable(): void
+    {
+        $fromTable = new Table('table');
+        $fromTable->addColumn('id', 'integer');
+        $fromTable->addColumn('reference1_id1', 'integer');
+        $fromTable->addColumn('reference1_id2', 'integer');
+        $fromTable->addColumn('reference2_id', 'integer');
+        $fromTable->addColumn('reference3_id1', 'integer');
+        $fromTable->addColumn('reference3_id2', 'integer');
+        $fromTable->addColumn('reference4_id', 'integer');
+
+        $fromTable->addForeignKeyConstraint('table_ref1', ['reference1_id1'], ['id1'], [], 'fk1');
+        $fromTable->addForeignKeyConstraint('table_ref2', ['reference2_id'], ['id']);
+
+        $toTable = clone $fromTable;
+        $toTable->removeForeignKey('fk1');
+        $toTable->addForeignKeyConstraint(
+            'table_ref1',
+            ['reference1_id1', 'reference1_id2'],
+            ['id1', 'id2'],
+            [],
+            'fk1'
+        );
+        $toTable->addForeignKeyConstraint('table_ref3', ['reference3_id1', 'reference3_id2'], ['id1', 'id2']);
+        $toTable->addForeignKeyConstraint('table_ref4', ['reference4_id'], ['id']);
+
+        $tableDiff = (new Comparator())->diffTable($fromTable, $toTable);
+
+        self::assertCount(2, $tableDiff->addedForeignKeys);
+        self::assertCount(1, $tableDiff->changedForeignKeys);
+    }
 }


greg0ire:
The blueprint code for the documents table is here: https://github.com/uuf6429/doc-sqlite-poc/blob/b9b5bd688342ca7d0b75addf8d54579e50a8c4d7/database/migrations/2030_09_02_3_create_documents_table.php

That repo also comes with an SQLite database versioned in it. Using sqliteman to describe the documents table, I obtain the following piece of SQL:

```sql
CREATE TABLE "documents" (
    "hash" varchar not null,
    "id" integer not null primary key autoincrement,
    "created_at" datetime null,
    "updated_at" datetime null,
    "title" varchar not null,
    "ek" varchar not null,
    "iv" varchar not null,
    "deleted_at" datetime null,
    "type_id" integer not null,
    "group_id" integer null,
    foreign key("type_id") references "types"("id") on delete cascade,
    foreign key("group_id") references "groups"("id") on delete cascade
)
```

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
```diff
@@ -257,4 +257,42 @@ public function testOnlyOwnCommentIsParsed(): void
...
+        $sqls = $this->connection->getDatabasePlatform()->getAlterTableSQL($diff);
+
+        foreach ($sqls as $sql) {
+            $this->connection->exec($sql);
+        }
+
+        $onlineTableAfter = $sm->listTableDetails('bug4243_posts');
+        $diff             = $comparator->diffTable($onlineTable, $onlineTableAfter);
+
+        $this->assertFalse($diff);
```
greg0ire, line 296: I'm unfamiliar with how this works… is `$onlineTable` supposed to be mutated at some point to be the same as `$onlineTableAfter`, i.e. a table with just an `id` column and nothing more?

beberlei: no, $onlineTable is the one create from the first exec(). the offline table has only the id, so the iteration over $sqls essentially recreates the existing table, because sqlite does not really do an ALTER.

Then after the applied SQLs, the onlineTableAfter should be exactly the same as the onlineTable.

lib/Doctrine/DBAL/Schema/Comparator.php
```diff
@@ -284,12 +284,19 @@ public function diffTable(Table $table1, Table $table2)
             foreach ($toFkeys as $key2 => $constraint2) {
                 if ($this->diffForeignKey($constraint1, $constraint2) === false) {
                     unset($fromFkeys[$key1], $toFkeys[$key2]);
-                } else {
-                    if (strtolower($constraint1->getName()) === strtolower($constraint2->getName())) {
-                        $tableDifferences->changedForeignKeys[] = $constraint2;
-                        $changes++;
-                        unset($fromFkeys[$key1], $toFkeys[$key2]);
-                    }
+                    continue 2;
```
beberlei, line 287: The `continue 2;` are new, but i believe they were missing in the previous code. If the "fromKey" from the outer loop was matched to a "toKey", then we can skip to the next "fromKey", and don't need to match the current fromKey to another "toKey".

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
greg0ire: Ok so in `$comparator->diffTable($offlineTable, $onlineTable);$sqls = $this->connection->getDatabasePlatform()->getAlterTableSQL($diff);`, `$offlineTable` is the starting point, and `$onlineTable` is the goal, is that it? So we are pretending `$offlineTable` is what we have in the DB, when in fact we have more columns, and it works anyway because we start by dropping the table?

beberlei:
AppVeyor fail is unrelated

greg0ire:
> Fixed issues | #4243

Not entirely, but it's definitely a step forward 👍

After that  we still have to figure out the issue of implicit indexes, that fails even with this patch:

```
git clone git@github.com:uuf6429/doc-sqlite-poc.git
cd doc-sqlite-poc
composer config repositories.beberlei vcs https://github.com/beberlei/dbal
composer require doctrine/dbal "dev-GH-4243-MoreSqliteSchemaManagerFixes as 2.10.3"
vendor/bin/phpunit tests/DatabaseTest.php
PHPUnit 8.5.8 by Sebastian Bergmann and contributors.

E                                                                   1 / 1 (100%)

Time: 189 ms, Memory: 10.00 MB

There was 1 error:

1) Tests\DatabaseTest::testThatTableExists
Illuminate\Database\QueryException: SQLSTATE[HY000]: General error: 1 no such index: IDX_A2B07288FE54D947 (SQL: DROP INDEX IDX_A2B07288FE54D947)

```

greg0ire (dismissed):

lib/Doctrine/DBAL/Schema/Comparator.php
```diff
@@ -284,12 +284,19 @@ public function diffTable(Table $table1, Table $table2)
...
+                }
+
+                // only named foreign key constraints can change, otherwise it appears as removed+added.
+                if (
+                    $constraint1->getName() &&
```
morozov, line 292: Implicit casting to boolean is not allowed in newer branches. Please express the logic explicitly.

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -406,10 +406,11 @@ protected function _getPortableViewDefinition($view)
     protected function _getPortableTableForeignKeysList($tableForeignKeys)
     {
         $list = [];
+        $i    = 0;
         foreach ($tableForeignKeys as $value) {
             $value = array_change_key_case($value, CASE_LOWER);
-            $name  = $value['constraint_name'];
-            if (! isset($list[$name])) {
+            $index = $value['constraint_name'] ?: $i++;
```
morozov, line 412: Why is it necessary to mix integers and strings as array keys? Since numeric index names are technically possible (https://github.com/doctrine/dbal/pull/4247), it may create unnecessary conflicts. As far as I see, the keys are not used once the array is constructed.

morozov: Short ternary operator is not allowed by PHPStan starting `3.0.x`. Please be more specific about what the empty `$value['constraint_name']` actually is.

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
```diff
@@ -257,4 +257,52 @@ public function testOnlyOwnCommentIsParsed(): void
...
+    "hash" varchar not null,
+    "id" integer not null primary key autoincrement,
+    "created_at" datetime null,
+    "updated_at" datetime null,
+    "title" varchar not null,
+    "ek" varchar not null,
+    "iv" varchar not null,
+    "deleted_at" datetime null,
+    "type_id" integer not null,
+    "group_id" integer null,
```
morozov, line 280: Are all the columns above relevant to the test? Please reduce the declaration to the necessary minimum.

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
```diff
@@ -257,4 +257,52 @@ public function testOnlyOwnCommentIsParsed(): void
             ->getColumn('col1')
             ->getComment());
     }
+
+    public function testUnnamedForeignKeyConstraintHandling(): void
+    {
+        $sql = <<<SQL
+CREATE TABLE "bug4243_types" (
```
morozov, line 264: Would it be possible to not have references to the bug in the test? The reference is available through Git history. The uniqueness of the table names needs to be guaranteed within one test file.

It would be even better if the tables were named according to their roles in the test (e.g. `referencing`, `referenced`).

morozov:
The change from https://github.com/greg0ire/dbal/commit/68fe0000ad539d34ffc80390b94dfe89339c3334 is a BC break. I wouldn't consider it for `2.11.x` (and for any newer release either) even in theory.

morozov:
> AppVeyor fail is unrelated

Offtopic: it fails regularly on AppVeyor due to not being able to submit code coverage to CodeCov. The question is, does it only happen on AppVeyor due to connectivity issues between AppVeyor and CodeCov (but not between GitHub and CodeCov or Travis and CodeCov) or because it's the only place where we fail the build that cannot submit code coverage.

This all instability results in fluctuation of the code coverage and misleading coverage diffs.

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
beberlei: @morozov you are right, its better to use `$value['id']` which is the identifier of the foreign key on the table. This would allow unnamed foreign keys with multiple columns, which the approach with $i does not support. Good catch.

The PHPStan change to disallow `?:` but allow `??` makes no sense to me, but ok :)

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
beberlei: the connection is reused accross the whole test setup based on `DbalFunctionalTestCase`, which would lead to potential clashes in table names. The namespacing of the table name is necessary or not?

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
beberlei: @greg0ire exactly.

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
morozov: Actually, I was wrong (confused with something else). The tests are responsible for the uniqueness of their tables. For instance:

https://github.com/doctrine/dbal/blob/3f3551aadb20c7ed58ee9c7f06ed74770588272f/tests/Doctrine/Tests/DBAL/Functional/TemporaryTableTest.php#L15-L18

In this case, adding `DROP TABLE IF EXISTS` in the beginning of the script will suffice.

lib/Doctrine/DBAL/Schema/Comparator.php
morozov: This change in the code that is used by all platforms is only covered by an SQLite test. Would it be possible to rework the new test to use the DBAL APIs and run across platforms?

greg0ire:
> The change from greg0ire@68fe000 is a BC break. I wouldn't consider it for 2.10.x (and for any newer release either) even in theory.

Me neither to be clear: it's just here to demonstrate that the issue that will be left after this is merged has to do with implicit indexes.

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -408,8 +408,8 @@ protected function _getPortableTableForeignKeysList($tableForeignKeys)
         $list = [];
         foreach ($tableForeignKeys as $value) {
             $value = array_change_key_case($value, CASE_LOWER);
-            $name  = $value['constraint_name'];
-            if (! isset($list[$name])) {
+            $index = $value['constraint_name'] ?? $value['id'];
```
morozov, line 411: Why does the list need to have the constraint name as the key? There's still a potential for collision (the identifiers are numeric strings: `0`, `1`, `2`…).

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
```diff
@@ -257,4 +257,51 @@ public function testOnlyOwnCommentIsParsed(): void
             ->getColumn('col1')
             ->getComment());
     }
+
+    public function testUnnamedForeignKeyConstraintHandling(): void
+    {
+        $sql = <<<SQL
+CREATE TABLE "unnamedfk_referenced1" (
```
morozov, line 264: Namespacing table names doesn't yield enough test isolation. If the test is run against a persistent database, it will be impossible to run it twice. Use `DROP TABLE IF EXISTS`.

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
morozov: It makes perfect sense: it disallows implicit casting to boolean.

morozov: > This would allow unnamed foreign keys with multiple columns, which the approach with $i does not support.

Indeed.

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
```diff
@@ -257,4 +257,54 @@ public function testOnlyOwnCommentIsParsed(): void
             ->getColumn('col1')
             ->getComment());
     }
+
+    public function testUnnamedForeignKeyConstraintHandling(): void
```
morozov, line 261: Would it be possible to move it to the abstract class and test across platforms? The more we have platform-specific tests, the lass consistent the abstraction is.

beberlei: The platform feature added only being used by sqlite, it would be skipped on all other platforms. The bug was an exception due to simple logic error, no need to make it more functional, execution of the sql is irrelevant for the fix, only not triggering the exception is, the test wouldnt need the assertions actually

lib/Doctrine/DBAL/Schema/Comparator.php
greg0ire: Wouldn't applying https://github.com/doctrine/dbal/pull/4256/files#r484110298 be enough to do this? Or does "use the DBAL APIs" mean something else?

morozov: @greg0ire, yes, it's exactly the same. I just reiterated because the previous comment wasn't taken into account.

greg0ire: I tried it in https://github.com/doctrine/dbal/pull/4262 , and it works only for SQLite although I used a high level API.

morozov: It is fine if it only works with SQLite right now, we can mark the test incomplete for other platforms. But this way, we at least identified the inconsistency in the API (which is a bug) which will be documented in the form of an incomplete test.

lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php
```diff
@@ -408,8 +408,8 @@ protected function _getPortableTableForeignKeysList($tableForeignKeys)
         $list = [];
         foreach ($tableForeignKeys as $value) {
             $value = array_change_key_case($value, CASE_LOWER);
-            $name  = $value['constraint_name'];
-            if (! isset($list[$name])) {
+            $index = $value['id'];
```
morozov, line 411: Since the array key is no longer a combination of the name and id, would it make sense to rename the variable to `$id` in order to not introduce new terms?

morozov (commented):
It looks like both new tests pass on `2.10.x` without the changes in `lib/`.

beberlei:
@morozov thats bad, for the compare foreign key that is expected, because I added it as a guard for the new `continue 2;` statements, but for the functional test that means the meaning was lost while converting it from the specific Sqlite test to the generic test. Will look what happened, but I have an idea.

beberlei:
@morozov it has to be an sqlite specific test, because the schema that breaks it cannot be created with the Schema abstraction, it will always generate a constraint name.

morozov:
@beberlei I think I understand what you're saying. To make it documented, could you post the SQL that needs to be generated to fail the test (it should exist in the earlier versions of this PR) and the closest possible SQL that could be produced by the Schema API. This way, it will be clear why it's an SQLite-only scenario.

beberlei:
@morozov i would rather revert to the old test with the SQL and document it, wouldn't it be better to have a working regression test for this?

The test is only relevant on SQLite in anyways, because the other databases did not have this bug, because they don't have supportsCreateDropForeignKey returning false. So its not problematic that we don't cover the abstraction here, as its more a regression test for Sqlite.

It seems i deprecated the necessary method on Table `addUnnamedForeignKeyConstraint`, though it wouldn't work anymore anyways at it automatically generates a name now. It would be possible using another deprecated method, but thats also a funky way:

```php
$referencingTable->addNamedForeignKeyConstraint(null, ...);
```

beberlei:
@greg0ire shouldn't the SQL in https://github.com/doctrine/dbal/pull/4256/files#diff-ffff5b2690b3aab7bb35b745b6ca89b9R261 already lead to the DROP INDEX problem as well? wondering how to expand the test to get that failure as well.

morozov:
> i would rather revert to the old test with the SQL and document it

Absolutely. I just wanted it to be clarified in the discussion of this PR. So, as I understand, it can't be tested via the abstraction because it doesn't allow to create an anonymous FK for SQLite? If that's the case, then it's a clear enough explanation. Let's have it clearly stated in the PR header and possibly in the comment to the test.

morozov:
The next question (not blocking the work on the fix) is, why does DBAL force naming the object (e.g. an FK) if the underlying platform allows it to be anonymous? Even if it's to simplify the schema management, there's still a wrong assumption that such objects should be identified by name. They should be identified by value (which is implemented to some extent for indices).

beberlei:
@morozov are the Postgresql failures unrelated? https://travis-ci.org/github/doctrine/dbal/jobs/725648695 they suddenly started to pop up, 2.10.x last commit is green.

morozov:
@beberlei I've never seen these specific failures on Postgres. I restarted the job to see if they persist.

morozov:
Is it expected that `testCompareForeignKeysOfTable()` passes even without the change in the comparator?

greg0ire:
> @greg0ire shouldn't the SQL in https://github.com/doctrine/dbal/pull/4256/files#diff-ffff5b2690b3aab7bb35b745b6ca89b9R261 already lead to the DROP INDEX problem as well? wondering how to expand the test to get that failure as well.

Having any FK should be enough I think, the way this bug happens is DBAL wrongly supposes the index is supposed to exist because who says FK in SQLite says DBAL-implicit index, but if the table is created with Laravel blueprint that will not be the case. Maybe there should be a separate test case where we attempt the same thing that is attempted in https://github.com/uuf6429/doc-sqlite-poc/blob/master/database/migrations/2030_09_02_4_alter_documents_table.php

morozov:
Irrelevant as of #4255.

PR #4262, opened 2020-09-08
Carry #4256

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #4243, partially

#### Summary

Let's see if the test passes if moved to the abstract functional test class


greg0ire:
So we have Oracle that complains with "SQL command not properly ended"
Mysql is either unhappy with `DROP TABLE IF EXISTS` or with the quotes around the table name
PostgreSQL does not understand `autoincrement`

I know understand what you meant by DBAL APIs @morozov … I suppose this should be done with the schema manager?

greg0ire:
Gave it another try with the schema manager API and got other issues…

morozov:
The [first query that fails on MySQL](https://travis-ci.org/github/doctrine/dbal/jobs/725368838#L746) looks like:
```sql
ALTER TABLE unnamedfk_referencing ADD PRIMARY KEY (id)
```
which is obviously wrong because the table already has the primary key:
https://github.com/doctrine/dbal/blob/8b04de0b35b37c9b49897bd2d0793aee8116a448/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php#L1632-L1634

This test will also fail if run on top of unchanged `2.10.x` code (i.e. it's not a regression but is likely a bug).

morozov:
Schema comparison on PostgreSQL seems to be [not working properly](https://travis-ci.org/github/doctrine/dbal/jobs/725368842#L532) as well:
```sql
ALTER TABLE unnamedfk_referencing ADD CONSTRAINT fk_79b66fb8f4a2af526dabfee8 FOREIGN KEY (reference1_id1, reference1_id2) REFERENCES unnamedfk_referenced1 (id1, id2) ON DELETE CASCADE NOT DEFERRABLE INITIALLY IMMEDIATE
```
The foreign key is already there: https://github.com/doctrine/dbal/blob/8b04de0b35b37c9b49897bd2d0793aee8116a448/tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php#L1638-L1643

tests/Doctrine/Tests/DBAL/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1616,6 +1616,64 @@ public function testSchemaDiffForeignKeys(): void
...
+        $onlineTable = $sm->listTableDetails('unnamedfk_referencing');
+
+        $offlineTable = new Table('unnamedfk_referencing');
+        $offlineTable->addColumn('id', 'integer');
+        $offlineTable->addColumn('reference1_id1', 'integer');
+        $offlineTable->addColumn('reference1_id2', 'integer');
+        $offlineTable->addColumn('reference2_id', 'integer');
+
+        $comparator = new Comparator();
+        $diff       = $comparator->diffTable($offlineTable, $onlineTable);
```
morozov, line 1664: Not sure I understand the logic:

1. We compare the offline table (which doesn't have the foreign keys) with the online one (which does).
2. The diff will obviously have the missing foreign keys.
3. By executing the diff SQL, we apply this diff to the online schema (which already has the foreign keys) which fails.

What's the expected result?

morozov: The only reason why the test passes on SQLite is that its schema manager, instead of adding FKs to the table, creates a new table and drops the existing one: https://github.com/doctrine/dbal/blob/b5e15999e8f626b9a50edef142fffc7be57d0211/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php#L961-L981

greg0ire: See https://github.com/doctrine/dbal/pull/4256/#discussion_r484050582 for the explanation about the logic . This is a resolved discussion, I'm not sure how Github behaves when you link to that, if it does not work it was my first comment on that PR.

>  it works anyway because we start by dropping the table

I see we came to the same conclusion. This logic is unlikely to work with other RDBMS than SQLite.

morozov: I see that the updated #4256 no longer has this logic but if it did, it would mean we were fixing something that doesn't make sense in the first place (i.e. we didn't replicate the issue properly in the form of a test or it's not a BC break because, despite the fact it worked in an earlier version, it wasn't meant to).

morozov: > This logic is unlikely to work with other RDBMS than SQLite since they support `ALTER TABLE` with foreign keys

Regardless of the underlying platform, I don't understand what this “logic” is about. If you compute a diff between `a` and `b` and then apply to `b`, what's the intent, and what is the expected result? Purely logically.

greg0ire: Maybe @beberlei can answer that, to me it's not fully clear to be honest.

PR #4266, opened 2020-09-10
Update SchemaDiff.php

ta-tikoma:
|      Q       |   A
|------------- | -----------
| BC Break     | yes/no
| Fixed issues | 

#### Summary

When I create migration for sqlite I have error:

Sqlite platform does not support alter foreign key, the table must be fully recreated using getAlterTableSQL.


morozov (changes requested):
This may be irrelevant as of #4255. Otherwise, please provide a test case.

ta-tikoma:
possibly. I hope

PR #4268, opened 2020-09-10
Handle both ways of PDO reporting failed connection

### LINKED ISSUE #4267: PostgreSQL tests fail on Travis with PHP 7.4 and PHP 7.4
The failures must be the result of fixing bug [#64705](https://bugs.php.net/bug.php?id=64705) (https://github.com/php/php-src/commit/2fe2e5b48f8cf9a45ac2530dd0e2dca8f4287e01) and are reproducible on PHP 7.3.22+ and PHP 7.4.10+.

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

Fixes #4267.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractPostgreSQLDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractPostgreSQLDriver.php
@@ -31,7 +31,9 @@ abstract class AbstractPostgreSQLDriver implements Driver, ExceptionConverterDri
      */
     public function convertException($message, DriverException $exception)
     {
-        switch ($exception->getSQLState()) {
+        $sqlState = $exception->getSQLState();
+
+        switch ($sqlState) {
             case '40001':
             case '40P01':
                 return new Exception\DeadlockException($message, $exception);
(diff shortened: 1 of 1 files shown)


morozov:
`DriverExeption` in DBAL 2.x (prior to https://github.com/doctrine/dbal/pull/4113) is a total mess. In case of a PostgreSQL connection failure, on PHP 7.4.9, `$exception->getSQLState()` would return `(int) 7` but on PHP 7.4.10 it's `(string) '08006'`.

greg0ire (approved):

lib/Doctrine/DBAL/Driver/AbstractPostgreSQLDriver.php
```diff
@@ -31,7 +31,9 @@ abstract class AbstractPostgreSQLDriver implements Driver, ExceptionConverterDri
      */
     public function convertException($message, DriverException $exception)
     {
-        switch ($exception->getSQLState()) {
+        $sqlState = $exception->getSQLState();
+
+        switch ($sqlState) {
```
phil-davis, line 36: It looks like this was not an essential part of the fix. `$sqlState` is only used by the `switch` statement anyway. But creating a local variable will do no harm.

morozov: @phil-davis, you're right. It was introduced in the original version of this PR (https://github.com/doctrine/dbal/commit/5a7479c29481949c92784850c164411b328f527c) which then had to be reworked. This chunk could have been reverted if noticed earlier.

PR #4270, opened 2020-09-11
Gdanis add varchar type

### LINKED ISSUE #4263: Support true VARCHAR binding to prevent implicit conversions?
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

PHP 7.2 added support for specifying different bind types for unicode and non-unicode string data: https://wiki.php.net/rfc/extended-string-types-for-pdo

This is a major improvement over the default (PDO::PARAM_STR) which binds strings as unicode data. In the past, the inability for PHP to handle this has led to performance problems in certain cases. For example if we have something like:

```php
$statement = $connection->prepare(
   'SELECT * FROM ... WHERE myVarcharColumn = :value' -- will bind as N'...'
);

$statement->bindValue(':value', $myVarcharValue);
$statement->execute();
```

If there's an index on `myVarcharColumn` SQL won't be able to use the index effectively since the binding will be an NVARCHAR string and force an [implicit conversation](http://blog.sqlgrease.com/are-varchar-or-char-hurting-your-performance-due-to-implicit-conversions/).

Does the team have a plan to support this binding in the future? We actually implemented this internally in a wrapper we have DBAL. I'm happy to open a PR if the team supports this feature:

```php
use Doctrine\DBAL\Types\StringType;
use PDO;

class VarcharType extends StringType
{
    public const NAME = 'varchar';

    public function getSQLDeclaration(array $field, AbstractPlatform $platform)
    {
        $length = $field['length'] ?? $platform->getVarcharDefaultLength();
        return 'VARCHAR(' . $length . ')';
    }

    public function getBindingType()
    {
        return PDO::PARAM_STR | PDO::PARAM_STR_CHAR;
    }

    public function getName()
    {
        return static::NAME;
    }
}
```

gjdanis:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #4263 

#### Summary

Implements the changes proposed in #4263.


src/Driver/Encodings.php
```diff
@@ -0,0 +1,24 @@
...
+namespace Doctrine\DBAL\Driver;
+
+final class Encodings
+{
+    /**
+     * Represents whether a value should be bound with ASCII encoding
+     *
+     * @see \PDO::PARAM_STR_CHAR
+     */
+    public const ASCII = 536870912;
```
morozov, line 14: Since there's only one possible value of the encoding (UTF8 will be used implicitly), would it make sense to just introduce another parameter type? See https://github.com/doctrine/dbal/pull/3149 for the reference.

src/Driver/Encodings.php
```diff
@@ -0,0 +1,24 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Driver;
+
+final class Encodings
+{
+    /**
+     * Represents whether a value should be bound with ASCII encoding
+     *
+     * @see \PDO::PARAM_STR_CHAR
```
morozov, line 12: The existing code has references to PDO only for backward compatibility. The new code doesn't have to be and it definitely doesn't have to borrow values from PDO.

src/Types/VarCharType.php
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Doctrine\DBAL\Types;
+
+use Doctrine\DBAL\Driver\Encodings;
+use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+
+use function sprintf;
+
+class VarCharType extends StringType
```
morozov, line 13: Would it make sense to name it as `AsciiStringType`? I believe it can be used with other column types than `VARCHAR` (e.g. `CHAR`).

src/Types/VarCharType.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getSQLDeclaration(array $column, AbstractPlatform $platform): string
```
morozov, line 18: Why does it have to be overridden?

tests/Functional/Driver/PDO/SQLSrv/DriverTest.php
```diff
@@ -77,4 +78,31 @@ public function testDriverOptions(): void
                 ->getAttribute(PDO::ATTR_CASE)
         );
     }
+
+    /**
+     * @dataProvider stringDataProvider
+     */
+    public function testDriverStringBinding(string $bindingType, string $expectedSqlType): void
```
morozov, line 85: I'd rather create a new functional `StringBindingTest` that would run with all drivers that support ASCII binding (`sqlsrv` and `pdo_sqlsrv`).

tests/Functional/Driver/PDO/SQLSrv/DriverTest.php
```diff
@@ -77,4 +78,31 @@ public function testDriverOptions(): void
...
+    {
+        $statement = $this->connection->prepare(
+            'SELECT sql_variant_property(:parameter, \'BaseType\') AS type'
```
morozov, line 88: The usage of named parameters is unnecessary here. They are emulated for most of the drivers and extend the scope of the test.

tests/Functional/Driver/PDO/SQLSrv/DriverTest.php
```diff
@@ -77,4 +78,31 @@ public function testDriverOptions(): void
...
+        );
+
+        $statement->bindValue(':parameter', 'TEST', $bindingType);
+        $results = $statement->execute()->fetchAssociative();
```
morozov, line 92: You can remove `AS type` from the query and use `fetchOne()` instead.

tests/Functional/Driver/PDO/SQLSrv/DriverTest.php
```diff
@@ -77,4 +78,31 @@ public function testDriverOptions(): void
...
+        self::assertEquals($expectedSqlType, $results['type']);
+    }
+
+    /**
+     * @return array<int, array<int, string>>
+     */
+    public static function stringDataProvider(): array
+    {
+        return [
+            [Types::VARCHAR, 'varchar'],
```
morozov, line 104: I'd rather use the parameter types here (`ParameterType::STRING` and `ParameterType::ASCII_STRING`). DBAL types is a higher-level abstraction that doesn't need to be tested at the driver level.

tests/Types/VarCharTest.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        self::assertEquals($this->type->getBindingType(), ParameterType::STRING | Encodings::ASCII);
+    }
+
+    /**
+     * @param array<string, int|bool> $column
+     *
+     * @dataProvider sqlDeclarationDataProvider
+     */
+    public function testReturnsSQLDeclaration(string $expectedSql, array $column): void
```
morozov, line 38: Does it even need to be tested? The sole purpose of this type is to override the default way of binding variables to the statement. The SQL declaration should be identical to the String type. Isn't it?

morozov:
Pease retarget against `3.0.x`. `master` is currently a placeholder for post-3.0 work.

gjdanis:
Thanks for the review @morozov -- implemented them in a new PR that targets v3.0.x: https://github.com/doctrine/dbal/pull/4274

PR #4274, opened 2020-09-14
Support ASCII parameter binding

### LINKED ISSUE #4263: Support true VARCHAR binding to prevent implicit conversions?

gjdanis:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #4263 

#### Summary

Implements the changes proposed in #4263.

### DIFF
--- a/docs/en/reference/types.rst
+++ b/docs/en/reference/types.rst
@@ -157,6 +157,13 @@ or ``null`` if no data is present.
     This can lead to type inconsistencies when reverse engineering the
     type from the database.
 
+ascii_string
+++++++++++++
+
+Similar to the ``string`` type but for binding non-unicode data. This type
+should be used with database vendors where a binding type mismatch
+can trigger an implicit cast and lead to performance problems.
+
 text
 ++++
 
@@ -606,6 +613,9 @@ Please also notice the mapping specific footnotes for additional information.
 |                   |               | **SQL Server**           | *all*   | ``NVARCHAR(n)`` [3]_                                     |
 |                   |               |                          |         +----------------------------------------------------------+
 |                   |               |                          |         | ``NCHAR(n)`` [4]_                                        |
++-------------------+---------------+--------------------------+---------+----------------------------------------------------------+
+| **ascii_string**  | ``string``    | **SQL Server**           |         | ``VARCHAR(n)``                                           |
+|                   |               |                          |         | ``CHAR(n)``                                              |
 +-------------------+---------------+--------------------------+---------+----------------------------------------------------------+
 | **text**          | ``string``    | **MySQL**                | *all*   | ``TINYTEXT`` [17]_                                       |
 |                   |               |                          |         +----------------------------------------------------------+
--- a/lib/Doctrine/DBAL/Driver/PDOSqlsrv/Statement.php
+++ b/lib/Doctrine/DBAL/Driver/PDOSqlsrv/Statement.php
@@ -17,11 +17,20 @@ class Statement extends PDO\Statement
      */
     public function bindParam($param, &$variable, $type = ParameterType::STRING, $length = null, $driverOptions = null)
     {
-        if (
-            ($type === ParameterType::LARGE_OBJECT || $type === ParameterType::BINARY)
-            && $driverOptions === null
-        ) {
-            $driverOptions = \PDO::SQLSRV_ENCODING_BINARY;
+        switch ($type) {
+            case ParameterType::LARGE_OBJECT:
+            case ParameterType::BINARY:
+                if ($driverOptions === null) {
+                    $driverOptions = \PDO::SQLSRV_ENCODING_BINARY;
+                }
+
+                break;
+
+            case ParameterType::ASCII:
+                $type          = ParameterType::STRING;
+                $length        = 0;
+                $driverOptions = \PDO::SQLSRV_ENCODING_SYSTEM;
+                break;
         }
 
         return parent::bindParam($param, $variable, $type, $length, $driverOptions);
--- a/tests/Doctrine/Tests/DBAL/Functional/ParameterTypes/AsciiTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/ParameterTypes/AsciiTest.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL\Functional\ParameterTypes;
+
+use Doctrine\DBAL\Driver\AbstractSQLServerDriver;
+use Doctrine\DBAL\ParameterType;
+use Doctrine\Tests\DbalFunctionalTestCase;
+
+class AsciiTest extends DbalFunctionalTestCase
+{
+    public function testAsciiBinding(): void
+    {
+        if (! $this->connection->getDriver() instanceof AbstractSQLServerDriver) {
+            self::markTestSkipped('Driver does not support ascii string binding');
+        }
+
+        $statement = $this->connection->prepare('SELECT sql_variant_property(?, \'BaseType\')');
+
+        $statement->bindValue(1, 'test', ParameterType::ASCII);
+        $statement->execute();
+
+        $results = $statement->fetchOne();
+
+        self::assertEquals('varchar', $results);
+    }
+}
--- a/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
+++ b/lib/Doctrine/DBAL/Driver/Mysqli/MysqliStatement.php
@@ -38,6 +38,7 @@ class MysqliStatement implements IteratorAggregate, StatementInterface, Result
 {
     /** @var string[] */
     protected static $_paramTypeMap = [
+        ParameterType::ASCII        => 's',
         ParameterType::STRING       => 's',
         ParameterType::BINARY       => 's',
         ParameterType::BOOLEAN      => 'i',
--- a/lib/Doctrine/DBAL/Driver/PDOStatement.php
+++ b/lib/Doctrine/DBAL/Driver/PDOStatement.php
@@ -30,6 +30,7 @@ class PDOStatement extends \PDOStatement implements StatementInterface, Result
         ParameterType::NULL         => PDO::PARAM_NULL,
         ParameterType::INTEGER      => PDO::PARAM_INT,
         ParameterType::STRING       => PDO::PARAM_STR,
+        ParameterType::ASCII        => PDO::PARAM_STR,
         ParameterType::BINARY       => PDO::PARAM_LOB,
         ParameterType::LARGE_OBJECT => PDO::PARAM_LOB,
         ParameterType::BOOLEAN      => PDO::PARAM_BOOL,
--- a/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
+++ b/lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvStatement.php
@@ -34,6 +34,7 @@ use function SQLSRV_SQLTYPE_VARBINARY;
 use function stripos;
 
 use const SQLSRV_ENC_BINARY;
+use const SQLSRV_ENC_CHAR;
 use const SQLSRV_ERR_ERRORS;
 use const SQLSRV_FETCH_ASSOC;
 use const SQLSRV_FETCH_BOTH;
@@ -303,6 +304,14 @@ class SQLSrvStatement implements IteratorAggregate, StatementInterface, Result
                         &$variable,
                         SQLSRV_PARAM_IN,
                         SQLSRV_PHPTYPE_STRING(SQLSRV_ENC_BINARY),
+                    ];
+                    break;
+
+                case ParameterType::ASCII:
+                    $params[$column - 1] = [
+                        &$variable,
+                        SQLSRV_PARAM_IN,
+                        SQLSRV_PHPTYPE_STRING(SQLSRV_ENC_CHAR),
                     ];
                     break;
 
--- a/lib/Doctrine/DBAL/ParameterType.php
+++ b/lib/Doctrine/DBAL/ParameterType.php
@@ -50,6 +50,11 @@ final class ParameterType
     public const BINARY = 16;
 
     /**
+     * Represents an ASCII string data type
+     */
+    public const ASCII = 17;
+
+    /**
      * This class cannot be instantiated.
      *
      * @codeCoverageIgnore
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -246,6 +246,17 @@ abstract class AbstractPlatform
     }
 
     /**
+     * Returns the SQL snippet used to declare a column that can
+     * store characters in the ASCII character set
+     *
+     * @param mixed[] $column
+     */
+    public function getAsciiStringTypeDeclarationSQL(array $column): string
+    {
+        return $this->getVarcharTypeDeclarationSQL($column);
+    }
+
+    /**
      * Returns the SQL snippet used to declare a VARCHAR column type.
      *
      * @param mixed[] $column
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -1194,6 +1194,20 @@ SQL
     /**
      * {@inheritDoc}
      */
+    public function getAsciiStringTypeDeclarationSQL(array $column): string
+    {
+        $length = $column['length'] ?? null;
+
+        if (! isset($column['fixed'])) {
+            return sprintf('VARCHAR(%d)', $length);
+        }
+
+        return sprintf('CHAR(%d)', $length);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
     protected function getVarcharTypeDeclarationSQLSnippet($length, $fixed)
     {
         return $fixed
--- a/lib/Doctrine/DBAL/Types/AsciiStringType.php
+++ b/lib/Doctrine/DBAL/Types/AsciiStringType.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Types;
+
+use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+
+final class AsciiStringType extends StringType
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getSQLDeclaration(array $column, AbstractPlatform $platform)
+    {
+        return $platform->getAsciiStringTypeDeclarationSQL($column);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getBindingType()
+    {
+        return ParameterType::ASCII;
+    }
+
+    public function getName(): string
+    {
+        return Types::ASCII_STRING;
+    }
+}
--- a/lib/Doctrine/DBAL/Types/Type.php
+++ b/lib/Doctrine/DBAL/Types/Type.php
@@ -99,6 +99,7 @@ abstract class Type
      */
     private const BUILTIN_TYPES_MAP = [
         Types::ARRAY                => ArrayType::class,
+        Types::ASCII_STRING         => AsciiStringType::class,
         Types::BIGINT               => BigIntType::class,
         Types::BINARY               => BinaryType::class,
         Types::BLOB                 => BlobType::class,
--- a/lib/Doctrine/DBAL/Types/Types.php
+++ b/lib/Doctrine/DBAL/Types/Types.php
@@ -10,6 +10,7 @@ namespace Doctrine\DBAL\Types;
 final class Types
 {
     public const ARRAY                = 'array';
+    public const ASCII_STRING         = 'ascii_string';
     public const BIGINT               = 'bigint';
     public const BINARY               = 'binary';
     public const BLOB                 = 'blob';
--- a/tests/Doctrine/Tests/DBAL/Functional/Types/AsciiStringTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Types/AsciiStringTest.php
@@ -1,0 +1,70 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL\Functional\Types;
+
+use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\Tests\DbalFunctionalTestCase;
+
+class AsciiStringTest extends DbalFunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $table = new Table('ascii_table');
+        $table->addColumn('id', 'ascii_string', [
+            'length' => 3,
+            'fixed' => true,
+        ]);
+
+        $table->addColumn('val', 'ascii_string', ['length' => 4]);
+        $table->setPrimaryKey(['id']);
+
+        $sm = $this->connection->getSchemaManager();
+        $sm->dropAndCreateTable($table);
+    }
+
+    public function testInsertAndSelect(): void
+    {
+        $id1 = 'id1';
+        $id2 = 'id2';
+
+        $value1 = 'val1';
+        $value2 = 'val2';
+
+        $this->insert($id1, $value1);
+        $this->insert($id2, $value2);
+
+        self::assertSame($value1, $this->select($id1));
+        self::assertSame($value2, $this->select($id2));
+    }
+
+    private function insert(string $id, string $value): void
+    {
+        $result = $this->connection->insert('ascii_table', [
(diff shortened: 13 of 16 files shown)


.gitignore
```diff
@@ -1,11 +1,12 @@
...
-reports/
-dist/
-download/
-vendor/
-/*.phpunit.xml
-/phpunit.xml
-/.phpcs-cache
-/phpstan.neon
+/.phpunit.result.cache
+.idea/
```
greg0ire, line 2: Please drop that commit. I'd rather not open the door to adding rules for every editor out there. Instead, developers should globally ignore their editor directory if any once in their life and be done with it.

gjdanis: Thanks, @greg0ire I'll do that.

Unrelated to this, is there a way to run the integration tests locally? The test I've added does return the expected `varchar` on a database I have with this version of the driver, so I'm not sure what's going wrong here.

greg0ire: There is no easy, documented way to run all tests for all databases if that's what you want, but you can still look at `.travis.yaml`, pick and job and runs what it runs.

gjdanis: Thanks @greg0ire - I think I've fixed the issue (misunderstood what `fetchOne` returns).

Maybe in a subsequent PR we could add a docker-compose configuration to make it easier to run integration tests locally?

greg0ire: > Maybe in a subsequent PR we could add a docker-compose configuration to make it easier to run integration tests locally?

IIRC, that kind of proposal is rejected on grounds that it it does not also run in the CI, it's going to break without us noticing. But I think some documentation on this, maybe in CONTRIBUTING.md would be welcome

greg0ire:
You should use `git rebase -i origin/3.0.x` and drop any commit related to `.idea`, as asked in https://github.com/doctrine/dbal/pull/4274#discussion_r488220780

Also, please associate your email address with your Github account, or change the

.gitignore
```diff
@@ -1,11 +1,11 @@
...
+build/
+logs/
+reports/
+dist/
+download/
+vendor/
+/*.phpunit.xml
+/phpunit.xml
+/.phpcs-cache
+/phpstan.neon
```
greg0ire, line 11: Why does this file appear in the diff?

ci/travis/install-mssql-pdo_sqlsrv.sh
```diff
@@ -4,4 +4,4 @@ set -ex
 
 echo "Installing extension"
 
-pecl install pdo_sqlsrv-5.7.0preview
+pecl install pdo_sqlsrv-5.8.1
```
greg0ire, line 7: This looks like a good change, but does it related to your changes? Shouldn't it be committed separately?

src/ParameterType.php
```diff
@@ -41,6 +41,11 @@ final class ParameterType
      */
     public const BINARY = 16;
 
+    /**
+     * Represents a ascii string data type
```
greg0ire, line 45:
```suggestion
     * Represents an ascii string data type
```

tests/Types/AsciiStringTest.php
```diff
@@ -0,0 +1,53 @@
...
+     * @dataProvider sqlDeclarationDataProvider
+     */
+    public function testReturnsSQLDeclaration(string $expectedSql, array $column): void
+    {
+        $declarationSql = $this->type->getSQLDeclaration($column, $this->platform);
+        self::assertEquals($expectedSql, $declarationSql);
+    }
+
+    /**
+     * @return array<int, array<int, string|array<string, mixed>>>
```
greg0ire, line 44:
```suggestion
     * @return array<int, array{string, array<string, mixed>}>
```

tests/Functional/Types/StringBindingTest.php
```diff
@@ -0,0 +1,32 @@
...
+        $statement->bindValue(1, 'test', ParameterType::ASCII_STRING);
+        $results = $statement->execute()->fetchOne();
+
+        self::assertEquals('varchar', $results);
+    }
+
+    private function driverSupportsAsciiBinding(): bool
+    {
+        return $this->connection->getDriver() instanceof Driver
+            || $this->connection->getDriver() instanceof \Doctrine\DBAL\Driver\PDO\SQLSrv\Driver;
```
greg0ire, line 30: This could also be implemented for MySQL, right?

>  MySQL and Microsoft SQL Server use it to store Unicode data. 

https://wiki.php.net/rfc/extended-string-types-for-pdo

greg0ire (commented):
docs/en/reference/types.rst should be updated, shouldn't it?

tests/Functional/Types/StringBindingTest.php
gjdanis: Yes possibly - do you have a good way to test this? The method described [here](https://stackoverflow.com/questions/1223211/get-the-type-of-a-variable-in-mysql) doesn't work unfortunately.

greg0ire: I don't, sorry. I just meant to clarify why you didn't do it in this PR (which is fine)

gjdanis: Ah perfect - I can open a subsequent issue on this.

src/Types/AsciiStringType.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getSQLDeclaration(array $column, AbstractPlatform $platform)
+    {
+        $length = $column['length'] ?? null;
+
+        if (! isset($column['fixed'])) {
+            return sprintf('VARCHAR(%d)', $length);
```
morozov, line 22: The SQL should be generated by the platform, not the type. Otherwise, it's not extensible.

I propose the following:
1. Introduce `AbstractPlatform::getAsciiStringTypeDeclarationSQL()` that would call `AbstractPlatform::getVarcharTypeDeclarationSQL()`.
2. Override the implementation in `SQLServer2012Platform`.

tests/Functional/Types/StringBindingTest.php
```diff
@@ -0,0 +1,32 @@
...
+
+        $statement->bindValue(1, 'test', ParameterType::ASCII_STRING);
+        $results = $statement->execute()->fetchOne();
+
+        self::assertEquals('varchar', $results);
+    }
+
+    private function driverSupportsAsciiBinding(): bool
+    {
+        return $this->connection->getDriver() instanceof Driver
```
morozov, line 29: You can check for `AbstractSQLServerDriver` instead.

tests/Types/AsciiStringTest.php
```diff
@@ -0,0 +1,53 @@
...
+{
+    /** @var AbstractPlatform|MockObject */
+    private $platform;
+
+    /** @var AsciiStringType */
+    private $type;
+
+    protected function setUp(): void
+    {
+        $this->platform = $this->createMock(AbstractPlatform::class);
```
morozov, line 23: Instead of mocking the platform, I believe the test should use the SQLServer2012Platform that will actually implement the generation of the desired SQL.

tests/Functional/Types/StringBindingTest.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Tests\Functional\Types;
```
morozov, line 5: This test is no longer about the data type, it's about the `ParameterType`. Might make sense to rename to `\Doctrine\DBAL\Tests\Functional\ParameterTypes\AsciiStringTest.php`.

tests/Functional/Types/StringBindingTest.php
```diff
@@ -0,0 +1,32 @@
...
+
+use Doctrine\DBAL\Driver\SQLSrv\Driver;
+use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Tests\FunctionalTestCase;
+
+class StringBindingTest extends FunctionalTestCase
```
morozov, line 11: There's no longer string binding. It's an `AsciiStringTest`.

morozov (commented):
Great work so far!

Besides a couple of comments below, it'd be nice to have another functional test at the driver level that would demonstrate the newly introduced type will work on all platforms, not only on the ones that actually support the new binding type.

See [`BinaryTest`](https://github.com/doctrine/dbal/blob/f2f611779a7b044cf06db6915d5d9f6dd958a3c2/tests/Functional/Types/BinaryTest.php) for the reference.

tests/Types/AsciiStringTest.php
gjdanis: I think we can actually test the SQL declaration in `SQLServerPlatformTest`. This test then becomes somewhat not needed.

Added this in the latest push if you want to take a look.

gjdanis: Unrelated to this do you have any idea why this pipeline step is failing? https://ci.appveyor.com/project/doctrine/dbal/builds/35249676/job/8ytj6a8vv8a8lcv0

It seems like the tests are passing but failing to download some code coverage analysis. Do you have any visibility into whether these failures are related to my changes?

morozov: It's a known issue, unrelated to your changes.

src/Platforms/AbstractPlatform.php
```diff
@@ -185,6 +185,19 @@ private function initializeAllDoctrineTypeMappings()
...
 
+    /**
+     * Returns the SQL snippet used to declare a column that can
+     * store characters in the ASCII character set
+     *
+     * @param mixed[] $column
+     *
+     * @return string
+     */
+    public function getAsciiTypeDeclarationSQL(array $column)
```
morozov, line 196: Let's keep the naming consistent. It should be `getAsciiStringTypeDeclarationSQL()` (https://github.com/doctrine/dbal/pull/4274#discussion_r489842653).

src/Types/AsciiType.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Types;
+
+use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+
+class AsciiType extends StringType
```
morozov, line 10: Please name `AsciiStringType`.

tests/Functional/Types/AsciiTest.php
```diff
@@ -0,0 +1,70 @@
...
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Tests\Functional\Types;
+
+use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\Tests\FunctionalTestCase;
+
+class AsciiTest extends FunctionalTestCase
```
morozov, line 11: Please name `AsciiStringTypeTest` according to the class being tested.

tests/Platforms/SQLServerPlatformTest.php
```diff
@@ -46,4 +46,26 @@ public function testGeneratesTypeDeclarationForDateTimeTz(): void
     {
         self::assertEquals('DATETIMEOFFSET(6)', $this->platform->getDateTimeTzTypeDeclarationSQL([]));
     }
+
+    /**
+     * @param array<string, mixed> $column
+     *
+     * @dataProvider sqlDeclarationDataProvider
+     */
+    public function testAsciiSQLDeclaration(string $expectedSql, array $column): void
```
morozov, line 55: This could be moved to the `AbstractPlatformTest` and cover all platforms. Only the data provider will have to be overridden here (see other platform tests).

tests/Types/AsciiTest.php
```diff
@@ -0,0 +1,43 @@
...
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Tests\Types;
+
+use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Types\AsciiType;
+use PHPUnit\Framework\TestCase;
+
+class AsciiTest extends TestCase
```
morozov, line 12: Please name `AsciiStringTypeTest` according to the class being tested.

gjdanis: @morozov should be name this new type `AsciiType` or `AsciiStringType`? I tend to prefer the former but let me know either way.

morozov: I'd prefer `AsciiStringType` and `'ascii_string'`:
1. They extend the String type.
2. Technically, there may be other ASCII-based types (e.g. ASCII Text).

src/ParameterType.php
```diff
@@ -41,6 +41,11 @@ final class ParameterType
      */
     public const BINARY = 16;
 
+    /**
+     * Represents an ascii string data type
+     */
+    public const ASCII_STRING = 17;
```
morozov, line 47: No, this should stay `ASCII`. This is a binding type that could be reused by different data types (e.g. `'ascii_text'`). The `'ascii_string'` is a data type.

src/Types/Type.php
```diff
@@ -21,6 +21,7 @@ abstract class Type
      */
     private const BUILTIN_TYPES_MAP = [
         Types::ARRAY                => ArrayType::class,
+        Types::ASCII_STRING                => AsciiStringType::class,
```
morozov, line 24: This needs better alignment.

docs/en/reference/types.rst
```diff
@@ -583,6 +590,9 @@ Please also notice the mapping specific footnotes for additional information.
 |                   |               |                          |         +----------------------------------------------------------+
 |                   |               |                          |         | ``NCHAR(n)`` [4]_                                        |
 +-------------------+---------------+--------------------------+---------+----------------------------------------------------------+
+| **ascii**         | ``string``    | **SQL Server**           |         | ``VARCHAR(n)``                                           |
```
morozov, line 593: Should it be `ascii_string` as a data type?

docs/en/reference/types.rst
```diff
@@ -157,6 +157,13 @@ or ``null`` if no data is present.
     This can lead to type inconsistencies when reverse engineering the
     type from the database.
 
+ascii_string
+++++++++++++
+
+Similar to the ``string`` type but for binding non unicode data. This type
```
morozov, line 163: Should this be “non-unicode”?

src/ParameterType.php
```diff
@@ -41,6 +41,11 @@ final class ParameterType
      */
     public const BINARY = 16;
 
+    /**
+     * Represents an ascii string data type
```
morozov, line 45: It's better to use “ASCII” here since it's an acronym.

morozov (commented):
Looks good. Please address the remaining comments and squash. It should be good to go.

morozov (dismissed):

greg0ire:
In your example:

```php
$statement = $connection->prepare(
   'SELECT * FROM ... WHERE myVarcharColumn = :value' -- will bind as N'...'
);
```

What exactly dictates that `:value` will be bound as `N'…'`?

gjdanis:
@greg0ire the driver code prepends the `N` character. If you use SQL Server Profiler you can see that the SQL that is sent to the server includes it. The actual code for this can be found here: https://github.com/microsoft/msphpsql/blob/5f76f838e5d44ca951a44bec6be1af05fae4bd93/source/pdo_sqlsrv/pdo_dbh.cpp#L1565

[Here's](https://github.com/microsoft/msphpsql/issues/1018) my conversation with the developers of the driver about this isssue. There's also an [RFC](https://wiki.php.net/rfc/extended-string-types-for-pdo) about this issue from the PDO maintainers. This issue is why they added the extended string types.

greg0ire (dismissed):
Ok I think I got this right thanks to your explanation: when using the new ascii type, `PDO::SQLSRV_ENCODING_SYSTEM` will be used which means, "bind as 8-bit chars", which in turns means the driver will not add the `N` that would result in the implicit cast of the column values to utf-8

greg0ire:
Should this target 2.11.x or is there a BC-break somewhere in there?

gjdanis:
@greg0ire I think it could target 2.11.x but a previous conversation with @morozov asked that I move it to 3.0.x: https://github.com/doctrine/dbal/pull/4270#issuecomment-691362520

Let me know if I should retarget.

greg0ire:
He was asking you to target to 3.0.x, but that retarget was from master… @morozov should we re-evaluate this? I see no reason to postpone this to 3.0.x when we could have it in 2.11.x .

src/Platforms/AbstractPlatform.php
```diff
@@ -185,6 +185,19 @@ private function initializeAllDoctrineTypeMappings()
...
 
+    /**
+     * Returns the SQL snippet used to declare a column that can
+     * store characters in the ASCII character set
+     *
+     * @param mixed[] $column
+     *
+     * @return string
+     */
+    public function getAsciiStringTypeDeclarationSQL(array $column)
```
greg0ire, line 196: Any reason not to use a type declartion here?

```suggestion
    public function getAsciiStringTypeDeclarationSQL(array $column): string
```

src/Platforms/SQLServer2012Platform.php
```diff
@@ -1247,6 +1247,20 @@ public function getDateTimeTzTypeDeclarationSQL(array $column)
         return 'DATETIMEOFFSET(6)';
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public function getAsciiStringTypeDeclarationSQL(array $column)
```
greg0ire, line 1253:
```suggestion
    public function getAsciiStringTypeDeclarationSQL(array $column): string
```

src/Types/AsciiStringType.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Types;
+
+use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+
+class AsciiStringType extends StringType
```
greg0ire, line 10: @morozov I see we never use `final` for types, but maybe we should start?

greg0ire (commented):
A few extra comments while you wait for a decision on the target branch

src/Types/AsciiStringType.php
morozov: Absolutely. The requirement for the return type declaration is enforced by the coding standard only in `4.x` but the use of final classes isn't enforced even there. It might be a good topic to explore further at the standard level.

src/Platforms/AbstractPlatform.php
morozov: > Any reason not to use a type declaration here?

I believe it's because all other methods don't have it and it's not enforced by the standard. Although it's a good thing to use them for new code.

morozov:
> @morozov should we re-evaluate this? I see no reason to postpone this to 3.0.x when we could have it in 2.11.x .

I don't see a problem having it in `2.11.x` I asked for `3.0.x` because I didn't foresee it being ready before `2.11.0` is released.

@gjdanis if you want to retarget the PR, please do not open a new one. Edit and rebase this one instead.

greg0ire (dismissed):
@gjdanis I addressed the changes for you, LGTM

greg0ire:
Wait why did you change the branch back to 3.0.x?

greg0ire:
Ok this is why… I'll do the rebase for you.

gjdanis:
@greg0ire I got a lot of conflicts so I changed the base branch back. Was trying to clean this up locally but it's a little messy. If you know of a better way to retarget please go ahead and do so.

greg0ire:
There are way more conflicts than I expected, the code looks nothing like 3.0.x… let's change the base branch again I guess :sweat_smile: … 3.0.0 is not that far away anyway :crossed_fingers:

greg0ire (dismissed):
LGTM, you tell us if you want to give the retarget another attempt or if you can wait for 3.0.0

morozov:
Note that if we retarget at `2.11.x` and it turns out to be a lot of work, we'll also have to do the inverse of that work again when merging the result up. Although I'm fine with this and I can help to retarget it as well.

gjdanis:
@greg0ire, @morozov -- what do you think about merging this PR into 3.0.x and opening a new PR for 2.11.x? When we merge 2.11.x into 3.0.x we can just drop the PR that implements this change in 2.11.x, since the code will already be there.

Let me know what you want to do. I'm happy to implement the changes in 2.11.x but I think it would be a lot to do in this PR.

morozov:
@gjdanis it would be at least the same amount of work or even more. If you want this to be released in `2.11.0`, we should retarget this PR against `2.11.x`. I can do this myself.

gjdanis:
@morozov I feel like there should be 2.11.x version of this code so that large codebases that are reliant on 2.0 versions of this library can benefit from the change without having to refactor their code to be 3.0.x compliant. If you're offering to do this work, I'll take you up on the offer 👍

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Statement.php
```diff
@@ -17,11 +17,20 @@ class Statement extends PDO\Statement
...
+            case ParameterType::LARGE_OBJECT:
+            case ParameterType::BINARY:
+                if ($driverOptions === null) {
+                    $driverOptions = \PDO::SQLSRV_ENCODING_BINARY;
+                }
+
+                break;
+
+            case ParameterType::ASCII:
+                $type          = ParameterType::STRING;
```
morozov, line 30: I changed `PDO::PARAM_STR` to `ParameterType::STRING` during the rebase since the direct usage of `PDO` binding types in calls to DBAL is deprecated (although may work since they use the same underlying vaules).

tests/Doctrine/Tests/DBAL/Functional/ParameterTypes/AsciiTest.php
```diff
@@ -0,0 +1,28 @@
...
+    public function testAsciiBinding(): void
+    {
+        if (! $this->connection->getDriver() instanceof AbstractSQLServerDriver) {
+            self::markTestSkipped('Driver does not support ascii string binding');
+        }
+
+        $statement = $this->connection->prepare('SELECT sql_variant_property(?, \'BaseType\')');
+
+        $statement->bindValue(1, 'test', ParameterType::ASCII);
+        $statement->execute();
```
morozov, line 22: This had to be reworked during the rebase because in `2.x` `Statement::execute()` doesn't return a result.

morozov:
@gjdanis please see if the rebased PR looks do you.

lib/Doctrine/DBAL/Driver/PDOSqlsrv/Statement.php
gjdanis: Makes sense 👍

tests/Doctrine/Tests/DBAL/Functional/ParameterTypes/AsciiTest.php
gjdanis: Looks good thanks!

gjdanis (commented):
@morozov changes LGTM - thanks for the work to rebase!

morozov (approved):

morozov:
Thanks, @gjdanis!

greg0ire:
It sounds like a good plan to me!

greg0ire:
Strategy for the merge-up:
1. merge 2.11.x as it was before merging this PR (so `git merge cce615a9e`)
2. merge this PR as it was before the rebase ( `git merge 3b8c8f9f6`)
3. merge 2.11.x as it is/was after merging this PR, but discard changes (`git merge -s ours 2f3d9bb61`)
4. if 2.11.x moved since then, merge it.

PR #4275, opened 2020-09-19

greg0ire:
Not gonna lie, this was super hard because of:
- the exception rename
- the sqlite thingy revert
- things related to exceptions that happened in the meantime on 3.0.x

Please review super carefully. If it turns out not to be correct and too hard to fix, let's try doing the items above one by one. I used imerge for the merge, which helped a bit I think, but still…


phpstan.neon.dist
```diff
@@ -19,6 +19,10 @@ parameters:
         # TODO: remove in 4.0.0
         - '~^Parameter #2 \$registeredAliases of static method Doctrine\\DBAL\\Query\\QueryException::nonUniqueAlias\(\) expects array<string>, array<int, int\|string> given\.\z~'
 
+        # legacy variadic-like signature
+        # TODO: remove in 3.0.0
+        - '~^Method Doctrine\\DBAL(\\.*)?Connection::query\(\) invoked with \d+ parameters?, 0 required\.\z~'
```
morozov, line 24: This looks being introduced mistakenly.

morozov (commented):
I haven't reviewed it completely but it looks like at least a couple of changes are missing:
- https://github.com/doctrine/dbal/compare/ce85bd8719b18b61c3171f3324df281c7716aec4...2.11.x#diff-58432246b30eea57e32e0912074f5807
- https://github.com/doctrine/dbal/compare/ce85bd8719b18b61c3171f3324df281c7716aec4...2.11.x#diff-51b2b592af5d0b614210f1e1f9d5eeb0

The above is the comparison of the latest state of `2.11.x` merged into `3.0.x` (`git merge-base 2.11.x 3.0.x`) and the current state of `2.11.x`: https://github.com/doctrine/dbal/compare/ce85bd8719b18b61c3171f3324df281c7716aec4...2.11.x.

Please compare the diff above and the PR and see if something else is missing.

UPD: note that GitHub renders the links above incorrectly. Use the URLs from the message source.

greg0ire:
@morozov as outlined in https://github.com/doctrine/dbal/pull/4274#issuecomment-695181074, I'm not merging 2.11.x here, but a prior state of it, so I believe the correct diff to compare against would be https://github.com/doctrine/dbal/compare/ce85bd8719b18b61c3171f3324df281c7716aec4...cce615a9e

greg0ire:
Here are the 4 classes that do not appear in the diff of this PR, but do appear in the diff I linked:

- `AbstractSQLAnywhereDriver`: does not exist on 3.0.x
- `AbstractSQLServerDriver`: does not throw on 3.0.x
- `AbstractSQLiteDriver`: already does not have the code that gets removed
- `Mysqli/Driver`: no longer throws a DBALException

morozov:
OK, so the PR is only expected to not include the changes from https://github.com/doctrine/dbal/pull/4274 but the second chunk mentioned in the review comment doesn't belong there. Honestly, I don't see much benefit in taking the merge work apart. It adds confusion and complexity and doesn't reduce the amount of work to be done.

Why don't we merge everything at once as we've been doing so far?

greg0ire:
> Why don't we merge everything at once as we've been doing so far?

I thought it would be best to do that because steps 2 and 3 of the plan I described make this non-trivial. 

> the second chunk mentioned in the review comment doesn't belong there

not sure what you are talking about: 

- https://github.com/doctrine/dbal/pull/4275#discussion_r491469198 ? in this case it's the chunk mentioned in the second review comment, and I removed it.
- ce85bd8...2.11.x#diff-51b2b592af5d0b614210f1e1f9d5eeb0 ? You said that one was missing from the diff, so how can it "not belong here"?

I will try the next steps, hopefully it will make things easier.

morozov:
The chunk missing from the current PR (the changes in `AbstractPostgreSQLDriver` from https://github.com/doctrine/dbal/pull/4268) does not belong to https://github.com/doctrine/dbal/pull/4274, so it should be here but it doesn't seem to.

greg0ire:
> Please compare the diff above and the PR and see if something else is missing.

Ok I think I understood where I messed up. Will tell you when I am done with that.

greg0ire:
Done, please resume your review.

morozov:
Now this PR introduces the `getAsciiStringTypeDeclarationSQL()` method from #4274 but via the merge commit instead of having 7b58fd22bbd90b77dbabc892d28bc7500b5a5189 in the ancestry tree.

greg0ire:
@morozov fixed, sorry

morozov:
This is still not “Merge 2.11.x into 3.0.x”. The 0d9f0ad5e723d85194bd1a3a5691825d8bd452ac commit doesn't belong `2.11.x`, and there's 3 merge commits at the end.

Please redo it the normal way. Instead of resolving the conflicts again, you can use the tree in its current state.

greg0ire:
> The 0d9f0ad commit doesn't belong to 2.11.x

This was on purpose. Since gjdanis had already done the work for 3.0.x , I thought it would be simpler to merge this work, which 0d9f0ad represents, and then use the `ours` strategy to let git know that this is equivalent to merging what gjdanis did on 2.11.x . The 3 merge commits are:
- merge 2.11.x as it was before gjdanis' contribution
- merge gjdanis' contribution as it was before you rebased it on 2.11.x
- merge -s ours 2.11.x current state.

> Instead of resolving the conflicts again, you can use the tree in its current state.

Sounds easy enough. Done.

morozov:
> Since gjdanis had already done the work for 3.0.x , I thought it would be simpler to merge this work, which 0d9f0ad represents, and then use the ours strategy to let git know that this is equivalent to merging what gjdanis did on 2.11.x .

I get the idea but it's not “Merge … into …”, it's something else which we don't normally do. It could be done behind the curtains and documented since it's an implementation detail of the PR, not the intent.

Apart from that, I made a couple of changes during the backport, so just using the original commit couldn't be backported, there should have been another one on top (which you had I believe). This whole dance complicates the process hugely and unnecessarily IMO.

src/Driver/PDO/SQLSrv/Statement.php
```diff
@@ -28,11 +28,20 @@ public function __construct(PDOStatement $statement)
...
+            case ParameterType::LARGE_OBJECT:
+            case ParameterType::BINARY:
+                if ($driverOptions === null) {
+                    $driverOptions = PDO::SQLSRV_ENCODING_BINARY;
+                }
+
+                break;
+
+            case ParameterType::ASCII:
+                $type          = PDO::PARAM_STR;
```
morozov, line 41: This is what I was talking about. Look that the merged PR: https://github.com/doctrine/dbal/pull/4274/files#r491219932.

greg0ire: Ok, I'll add this change, but not the other one, since it looks more simple the way it is on 3.0.x

morozov: That other one should stay as is, right.

tests/Schema/SchemaDiffTest.php
```diff
@@ -108,8 +104,8 @@ private function createPlatform(bool $unsafe)
                 ->will(self::returnValue(true));
         $platform->expects(self::exactly(1))
                 ->method('supportsSequences')
-                ->will(self::returnValue(true));
-        $platform->expects(self::any())
+                ->will($this->returnValue(true));
+        $platform->expects($this->exactly(2))
```
morozov, line 108: These should remain `self::`.

morozov:
These tests should use `Exception`:
- https://github.com/doctrine/dbal/blob/9eb0219291aec995a75a40984be6f77449ab9b0b/tests/DBALExceptionTest.php#L29-L31
- https://github.com/doctrine/dbal/blob/9eb0219291aec995a75a40984be6f77449ab9b0b/tests/DBALExceptionTest.php#L40-L42

morozov:
The issue in the `SchemaDiffTest` must be the last one. The rest looks good.

morozov (approved):

morozov:
> Not gonna lie, this was super hard

Thanks, @greg0ire!

PR #4276, opened 2020-09-19
Generate alter add/change column with explicit position for MySQL

mvorisek:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | Alter add/change column now has now explicit position, so new/updated column is updated to the same position like when the table is created freshly.<br>- https://github.com/doctrine/orm/issues/8258<br>- https://github.com/doctrine/migrations/issues/558<br>- https://stackoverflow.com/questions/34873327/doctrine-migration-add-column-after-other-one<br>- https://stackoverflow.com/questions/24148568/doctrineschemaupdate-doesnt-respect-column-order

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -569,7 +569,8 @@ SQL
                 'comment' => $this->getColumnComment($column),
             ]);
 
-            $queryParts[] = 'ADD ' . $this->getColumnDeclarationSQL($column->getQuotedName($this), $columnArray);
+            $queryParts[] = 'ADD ' . $this->getColumnDeclarationSQL($column->getQuotedName($this), $columnArray)
+                    . $this->getColumnDeclarationPositionSQL($columnArray['name'], $diff->toTable);
         }
 
         foreach ($diff->removedColumns as $column) {
@@ -599,7 +600,8 @@ SQL
 
             $columnArray['comment'] = $this->getColumnComment($column);
             $queryParts[]           =  'CHANGE ' . ($columnDiff->getOldColumnName()->getQuotedName($this)) . ' '
-                    . $this->getColumnDeclarationSQL($column->getQuotedName($this), $columnArray);
+                    . $this->getColumnDeclarationSQL($column->getQuotedName($this), $columnArray)
+                    . $this->getColumnDeclarationPositionSQL($columnArray['name'], $diff->toTable);
         }
 
         foreach ($diff->renamedColumns as $oldColumnName => $column) {
@@ -611,7 +613,8 @@ SQL
             $columnArray            = $column->toArray();
             $columnArray['comment'] = $this->getColumnComment($column);
             $queryParts[]           =  'CHANGE ' . $oldColumnName->getQuotedName($this) . ' '
-                    . $this->getColumnDeclarationSQL($column->getQuotedName($this), $columnArray);
+                    . $this->getColumnDeclarationSQL($column->getQuotedName($this), $columnArray)
+                    . $this->getColumnDeclarationPositionSQL($columnArray['name'], $diff->toTable);
         }
 
         if (isset($diff->addedIndexes['primary'])) {
@@ -648,6 +651,31 @@ SQL
         }
 
         return array_merge($sql, $tableSql, $columnSql);
+    }
+
+    protected function getColumnDeclarationPositionSQL(string $columnName, ?Table $toTable): ?string
+    {
+        if ($toTable === null) {
+            return null;
+        }
+
+        $columnFound = false;
+        $prevColumn  = null;
+        foreach ($toTable->getColumns() as $k => $column) {
+            if ($k === $columnName) {
+                $columnFound = true;
+
+                break;
+            }
+
+            $prevColumn = $column;
+        }
+
+        if (! $columnFound) {
+            throw new InvalidArgumentException('Column name not found');
+        }
+
+        return $prevColumn === null ? ' FIRST' : ' AFTER ' . $prevColumn->getQuotedName($this);
     }
 
     /**
@@ -740,8 +768,11 @@ SQL
 
             $column->setAutoincrement(false);
 
-            $sql[] = 'ALTER TABLE ' . $tableName . ' MODIFY ' .
-                $this->getColumnDeclarationSQL($column->getQuotedName($this), $column->toArray());
+            $columnArray = $column->toArray();
+
+            $sql[] = 'ALTER TABLE ' . $tableName . ' MODIFY '
+                . $this->getColumnDeclarationSQL($column->getQuotedName($this), $columnArray)
+                . $this->getColumnDeclarationPositionSQL($columnArray['name'], $diff->toTable);
 
             // original autoincrement information might be needed later on by other parts of the table alteration
             $column->setAutoincrement(true);
@@ -778,8 +809,11 @@ SQL
                 // before we can drop and recreate the primary key.
                 $column->setAutoincrement(false);
 
-                $sql[] = 'ALTER TABLE ' . $table . ' MODIFY ' .
-                    $this->getColumnDeclarationSQL($column->getQuotedName($this), $column->toArray());
+                $columnArray = $column->toArray();
+
+                $sql[] = 'ALTER TABLE ' . $table . ' MODIFY '
+                    . $this->getColumnDeclarationSQL($column->getQuotedName($this), $columnArray)
+                    . $this->getColumnDeclarationPositionSQL($columnArray['name'], $diff->toTable);
 
                 // Restore the autoincrement attribute as it might be needed later on
                 // by other parts of the table alteration.
--- a/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
@@ -672,7 +672,7 @@ SQL
         }
 
         if (! $columnFound) {
-            throw new InvalidArgumentException('Column name not found');
+            throw new InvalidArgumentException('Column name "' . $columnName . '" not found');
         }
 
         return $prevColumn === null ? ' FIRST' : ' AFTER ' . $prevColumn->getQuotedName($this);
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
@@ -14,6 +14,8 @@ use Doctrine\DBAL\Types\Type;
 use Doctrine\DBAL\Types\Types;
 use Doctrine\Tests\Types\MySqlPointType;
 
+use function array_keys;
+
 class MySqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
 {
     protected function supportsPlatform(AbstractPlatform $platform): bool
@@ -214,6 +216,32 @@ class MySqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
         self::assertNull($onlineTable->getColumn('def_blob')->getDefault());
         self::assertNull($onlineTable->getColumn('def_blob_null')->getDefault());
         self::assertFalse($onlineTable->getColumn('def_blob_null')->getNotnull());
+    }
+
+    public function testAlterTableExplicitColumnPosition(): void
+    {
+        $tableName = 'test_column_position';
+        $table     = new Table($tableName);
+        $table->addColumn('id', 'integer');
+        $table->addColumn('a', 'integer');
+        $table->addColumn('b', 'integer');
+
+        $this->schemaManager->createTable($table);
+
+        self::assertSame(['id', 'a', 'b'], array_keys($this->schemaManager->listTableColumns($tableName)));
+
+        $table2 = new Table($tableName);
+        $table2->addColumn('id', 'float');
+        $table2->addColumn('new', 'float');
+        $table2->addColumn('b', 'float');
+        $table2->addColumn('a', 'float');
+
+        $comparator = new Comparator();
+        $diff       = $comparator->diffTable($table, $table2);
+        self::assertNotFalse($diff);
+        $this->schemaManager->alterTable($diff);
+
+        self::assertSame(['id', 'new', 'b', 'a'], array_keys($this->schemaManager->listTableColumns($tableName)));
     }
 
     public function testColumnCharset(): void
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
@@ -272,7 +272,7 @@ class MySqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
         $diff = $fromSchema->getMigrateToSql($toSchema, $this->connection->getDatabasePlatform());
         self::assertContains(
             'ALTER TABLE test_column_charset_change CHANGE col_string'
-                . ' col_string VARCHAR(100) CHARACTER SET ascii NOT NULL',
+                . ' col_string VARCHAR(100) CHARACTER SET ascii NOT NULL FIRST',
             $diff
         );
     }
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
@@ -13,6 +13,7 @@ use Doctrine\DBAL\Types\BlobType;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\DBAL\Types\Types;
 use Doctrine\Tests\Types\MySqlPointType;
+use InvalidArgumentException;
 
 use function array_keys;
 
@@ -242,6 +243,29 @@ class MySqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
         $this->schemaManager->alterTable($diff);
 
         self::assertSame(['id', 'new', 'b', 'a'], array_keys($this->schemaManager->listTableColumns($tableName)));
+    }
+
+    public function testAlterTableExplicitColumnPositionColumnNotFoundException(): void
+    {
+        $tableName = 'test_column_position_exception';
+        $table     = new Table($tableName);
+        $table->addColumn('id', 'integer');
+
+        $this->schemaManager->createTable($table);
+
+        $table2 = new Table($tableName);
+        $table2->addColumn('id', 'float');
+        $table2WithoutAColumn = clone $table2;
+        $table2->addColumn('a', 'float');
+
+        $comparator = new Comparator();
+        $diff       = $comparator->diffTable($table, $table2);
+        self::assertNotFalse($diff);
+        $diff->toTable = $table2WithoutAColumn;
+
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Column name "a" not found');
+        $this->schemaManager->alterTable($diff);
     }
 
     public function testColumnCharset(): void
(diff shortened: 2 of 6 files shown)


morozov:
Thank you for the patch, @mvorisek. Before proceeding, please make sure the new logic is covered with a functional test that demonstrates the desired end result.

mvorisek:
> Thank you for the patch, @mvorisek. Before proceeding, please make sure the new logic is covered with a functional test that demonstrates the desired end result.

functional test added, all working as expected

mvorisek:
> Thank you for the patch, @mvorisek. Before proceeding, please make sure the new logic is covered with a functional test that demonstrates the desired end result.

@morozov functional test added

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
```diff
@@ -650,6 +653,31 @@ public function getAlterTableSQL(TableDiff $diff)
...
+                $columnFound = true;
+
+                break;
+            }
+
+            $prevColumn = $column;
+        }
+
+        if (! $columnFound) {
+            throw new InvalidArgumentException('Column name "' . $columnName . '" not found');
```
morozov, line 675: Would it be possible to avoid this exception happening by using a more suitable method signature? Each column is either first or goes after some other column. See https://github.com/doctrine/dbal/pull/4145 for example. Otherwise, this will need a functional test that demonstrates the exception being thrown.

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
```diff
@@ -260,7 +286,7 @@ public function testColumnCharsetChange(): void
         $diff = $fromSchema->getMigrateToSql($toSchema, $this->connection->getDatabasePlatform());
         self::assertContains(
             'ALTER TABLE test_column_charset_change CHANGE col_string'
-                . ' col_string VARCHAR(100) CHARACTER SET ascii NOT NULL',
+                . ' col_string VARCHAR(100) CHARACTER SET ascii NOT NULL FIRST',
```
morozov, line 289: Why do we expect the SQL to contain the `FIRST`/`AFTER` qualifiers if the column position doesn't change? I believe these should be only generated when necessary.

tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
```diff
@@ -146,6 +146,7 @@ public function testCompareMissingField(): void
         );
         $expected->fromSchema                        = $schema1;
         $expected->changedTables['bugdb']->fromTable = $schema1->getTable('bugdb');
+        $expected->changedTables['bugdb']->toTable   = $schema2->getTable('bugdb');
```
morozov, line 149: The need to pass the `toTable` is a BC break. Why is it necessary in this case?

mvorisek: `toTable` is newly defined, what does it break?

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
mvorisek: we generate this for ALTER TABLE only, for TABLE CREATE, position is not generated

if position is the same, we regenerate whole column definition

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
mvorisek: we can call `$toTable->getColumn()` before the foreach loop which should check if the column exists or not

please sudgest ideal approach to this

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
morozov: Can we generate the column position only if it's changed?

tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
morozov: Why do you need to set it in an existing test?

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
morozov: It might work if instead of having the `$toTable` in the `SchemaDiff`, the `ColumnDiff` had a property like `$afterColumn` that would contain the column name and be non-empty only if the column position changed. This assumes that a similar SQL syntax could be used on other supported platforms. This would also solve the problem of having the `FIRST` and `AFTER` fragments unnecessarily generated.

mvorisek: `ColumnDiff` is used also for a new columns?

tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
mvorisek: because we are diffing whole object, I would say added property instead of breaking change

morozov: The need to modify existing tests to keep them passing after a change identifies a BC break.

morozov:
As long as this new feature requires breaking or non-breaking API changes, I'd like to make sure that the new API could be used to implement the same feature for other platforms. Otherwise, it may require a BC break in the future.

@mvorisek could you do some research and see which SQL syntax could be used for the rest of the platforms to implement this in the future?

mvorisek:
> As long as this new feature requires breaking or non-breaking API changes, I'd like to make sure that the new API could be used to implement the same feature for other platforms. Otherwise, it may require a BC break in the future.

the proposed data structure (with `$toTable` - there is already simillar use of `$fromTable`) we can implement it everywhere

> could you do some research and see which SQL syntax could be used for the rest of the platforms to implement this in the future?

I already did, other DB platforms require table recreation, ie. create table, insert into select, drop table

morozov:
> I already did, other DB platforms require table recreation, ie. create table, insert into select, drop table

In this case, I'm not sure we should be adding this feature.

mvorisek:
> > I already did, other DB platforms require table recreation, ie. create table, insert into select, drop table
> 
> In this case, I'm not sure we should be adding this feature.

This feature - reflecting column order defined by `Table` - has been asked by many users and I think we should do as much as we can to support it.

I think:
- for MySQL, the proposed impl. is very clean and it does allow MySQL users to update DB like if it was created freshly
- for other DBs, this feature can be implemented by recreate and reimport - eg. this PR lands a concept for it and someone else can implement this feature for other engines

@morozov can we agree on this and merge?

wdyt about adding `public function supportsColumnRerder(): bool` into `AbstractPlatform`, so user can check easily if reorder is supported or not? (and once supported in another platform, this method can be simply overrided in BC fashion)

morozov:
> This feature - reflecting column order defined by `Table` - has been asked by many users and I think we should do as much as we can to support it.

What's the use case for this? When does the order of columns in a table matter to the application that uses it?

> for MySQL, the proposed impl. is very clean

I disagree. I expressed my concerns above.

> for other DBs, this feature can be implemented by recreate and reimport - eg. this PR lands a concept for it and someone else can implement this feature for other engines

This is not completely true. We cannot introduce a new API that works for one DB platform and expect it to work for others that knowingly don't support the way of implementing the feature. That's why I asked to check how other platforms behave in this regard.

> wdyt about adding `public function supportsColumnRerder(): bool` into `AbstractPlatform`

It's not the best idea. Every time a "supports" method is added to the API, it forces the platform-specific logic to be implemented outside of the platform. It invalidates the purpose of the API which is to encapsulate the details of each its implementation behind a common interface.

> so user can check easily if reorder is supported or not

What if it's not supported? If the code can handle this case for other platforms, why can't it handle for MySQL?

For now, this could be reliably used in the code:

```php
if ($platform instanceof MySQLPlatform) {
    // reordering supported
}
```

tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
mvorisek: because we are diffing a whole object, which is something not expected to be tested outside DBAL

this change does not break anything in user application except the resulting column order is now like the table was created freshly

if added optional property is considered as BC break, then ok, it is breaking

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
mvorisek: I can add a condition, if the previous column is the same for fromTable and toTable, then to not add the position. Tests may however need more changes as currently, we expect that SQL column definition is always fully regenerated.

Let solve this as the last step before merge, firstly, I want to make sure `toTable` property will stay

mvorisek:
> > This feature - reflecting column order defined by `Table` - has been asked by many users and I think we should do as much as we can to support it.
> 
> What's the use case for this? When does the order of columns in a table matter to the application that uses it?

Basically any applications that use the database/column order like:
- any end user app that fetches/display data using `select *`
- database administration/development tools

the later point is very important from developers UX perspective and this PR improves it hugely, as column order is now like when a all tables would be created freshly - good for visual orientation, diffs, ...

> > for MySQL, the proposed impl. is very clean
> 
> I disagree. I expressed my concerns above.

I know, please reply to https://github.com/doctrine/dbal/pull/4276#discussion_r493095413 , I will then solve the impl. according to that, I think, `toTable` is clean approach and it goes well with the exisitng `fromTable`. If have a better idea, please propose, `ColumnDiff` can not be used as column may be new and `Column` itself is not the right place to contain `afterColumn` as `Column` as managing this data structure will be much harder instead of "just ordered list" like now

> > for other DBs, this feature can be implemented by recreate and reimport - eg. this PR lands a concept for it and someone else can implement this feature for other engines
> 
> This is not completely true. We cannot introduce a new API that works for one DB platform and expect it to work for others that knowingly don't support the way of implementing the feature. That's why I asked to check how other platforms behave in this regard.

now I propose for MySQL only, later, see above, it can be implemented with multiple SQL queries like "RENAME TABLE", "CREATE TABLE", "INSERT INTO SELECT", "DROP TABLE"

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
mvorisek: `ColumnDiff` can not be used as we need position also for `Column`. @morozov I would like to finish this PR asap, please advise, if the `$toTable` is ok - used like `$fromTable` already used - or if you have a better alternative. Thanks.

morozov:
The link to https://github.com/doctrine/dbal/pull/4276#discussion_r493095413 doesn't take me anywhere. Could you please quote your question?

> database administration/development tools

IMO it's an edge case that we can potentially support in a subset of supported platforms. Given how unclear the current approach is, it's not a priority for me personally at the moment. I'm fine if somebody else from the project team works on this and helps solve the issues I identified above.

mvorisek:
@greg0ire can you please review and comment especially on added `$toTable` property?

mvorisek:
@greg0ire can you please review and provide your feedback?

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
greg0ire: > ColumnDiff can not be used as we need position also for Column

I do not understand what this means. Can you please elaborate?

greg0ire:
> This feature has been asked by many users and I think we should do as much as we can to support it.

Can you please link to Github tickets in the "fixed issues" row of the table in your original message?

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
mvorisek: > 
> 
> > ColumnDiff can not be used as we need position also for Column
> 
> I do not understand what this means. Can you please elaborate?

if column already exists, then `ColumnDiff` instance is returned from the comparator

if the column is new (not renamed/modified), then there is no `ColumnDiff` for it

mvorisek:
> 
> 
> > This feature has been asked by many users and I think we should do as much as we can to support it.
> 
> Can you please link to Github tickets in the "fixed issues" row of the table in your original message?

Added. From my side, this PR is complete, but if you have any ideas, I am ready to implement them so we can release this asap. Thanks.

greg0ire:
>  From my side, this PR is complete

>  Otherwise, this will need a functional test that demonstrates the exception being thrown.

I think you still have to address this, both codecov and Sergei mentioned it.

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
greg0ire: I think `toTable` should stay since the other solution with `ColumnDiff` instance will not handle the case of new columns, which means we can resume this conversation. Let's generate the position only when necessary please.

mvorisek:
> 
> 
> > From my side, this PR is complete
> 
> > Otherwise, this will need a functional test that demonstrates the exception being thrown.
> 
> I think you still have to address this, both codecov and Sergei mentioned it.

test added, any additional feedback?

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
mvorisek: exception is covered by added test

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
mvorisek: > Let's generate the position only when necessary please.

I understand we want to generate minimalistic diff, but I think we do not want it. ALTER TABLE ADD/MODIFY column is always generated fully even if there is a minor column definition change like column comment. Adding position based on the original table will violate this definition.

mvorisek: @morozov @greg0ire can we agree on this and can we merge this PR?

greg0ire: > can we agree on this

If it's always generated fully indeed then it's ok with me.

> can we merge this PR

In theory yes, in practice the tests fail

mvorisek: CI fixed

greg0ire (dismissed):

mvorisek:
@greg0ire I fixed merge conflict, please rereview and please include this in the next upcoming 2.11 release.

greg0ire (dismissed):

mvorisek:
all feedback has been fixed, can we merge now?

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
```diff
@@ -216,6 +219,52 @@ public function testDoesNotPropagateDefaultValuesForUnsupportedColumnTypes(): vo
...
+        $this->schemaManager->createTable($table);
+
+        $table2 = new Table($tableName);
+        $table2->addColumn('id', 'float');
+        $table2WithoutAColumn = clone $table2;
+        $table2->addColumn('a', 'float');
+
+        $comparator    = new Comparator();
+        $diff          = $comparator->diffTable($table, $table2);
+        $diff->toTable = $table2WithoutAColumn;
```
morozov, line 261: Once the diff is generated from the two tables, why should it be possible to change the referenced table via the API?

morozov (commented):
@greg0ire what's your take on whether this behavior should be added in the first place? See https://github.com/doctrine/dbal/pull/4276#issuecomment-698699160.

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
mvorisek: this is test - to test exception https://github.com/doctrine/dbal/pull/4276/files#diff-548b552a74ea52eab68318520c213a142ba77527a4b965c4d3bd50eeba1afa9aR675

morozov: This test was meant to be added to document the case when this exception can happen because IMO it shouldn't be possible in the first place (https://github.com/doctrine/dbal/pull/4276/files#r493065750). The test doesn't show a valid case. Which means that there is no justification for this API to throw an exception.

lib/Doctrine/DBAL/Platforms/MySqlPlatform.php
morozov: @mvorisek please make sure you don't resolve the conversations that you didn't start unless you're resolving one on the starter's behalf.

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
mvorisek: When the diff object is built using Comparator::diff which will add the input, then this exception can not happen.

However, the `toTable` is public property - like `fromTable` already present - and if the table is replaced by the user with different table that has not all the column names set - this exception is legit and this test cover it.

greg0ire:
> @greg0ire what's your take on whether this behavior should be added in the first place? See #4276 (comment).

I think you know far more than I do about the philosophy of this project, and if you're not sure, I'm even less sure than you are… maybe @beberlei can advise on this?

@beberlei, is it commonplace to accept PRs that add features that can never be implemented on other platforms?

SenseException:
>> I already did, other DB platforms require table recreation, ie. create table, insert into select, drop table
>
> In this case, I'm not sure we should be adding this feature.

We have features that are not supported by other platforms, but might be added at some point. Here we can't implement it to certain platforms because there a recreation is needed. I'm not really sold yet to improve one platform that might be the only one with this feature, because the mentioned benefits aren't big ones.

While doing some research I found something in the PostgreSQL docs that has one technical benefit mentioned in http://wiki.postgresql.org/wiki/Alter_column_position:

> 1. physical layout can be optimized by putting fixed size columns at the start of the table

But I'm not sure if this is also the same benefit in MySQL.

mvorisek:
@greg0ire please reaprove, I rebased it from 2.11.x to 2.12.x branch.

> > 1. physical layout can be optimized by putting fixed size columns at the start of the table
> 
> But I'm not sure if this is also the same benefit in MySQL.

see https://stackoverflow.com/questions/894522/is-there-any-reason-to-worry-about-the-column-order-in-a-table - there is some performance improvement with controlled columns placement

Is there anything I can do more to get this PR merged?

mvorisek:
@greg0ire please reaprove, I rebased it from 2.11.x to 2.12.x branch.

> > 1. physical layout can be optimized by putting fixed size columns at the start of the table
> 
> But I'm not sure if this is also the same benefit in MySQL.

see https://stackoverflow.com/questions/894522/is-there-any-reason-to-worry-about-the-column-order-in-a-table - there is some performance improvement with controlled columns placement

Is there anything I can do more to get this PR merged?

mvorisek:
@greg0ire please reaprove, I rebased it from 2.11.x to 2.12.x branch.

> > 1. physical layout can be optimized by putting fixed size columns at the start of the table
> 
> But I'm not sure if this is also the same benefit in MySQL.

see https://stackoverflow.com/questions/894522/is-there-any-reason-to-worry-about-the-column-order-in-a-table - there is some performance improvement with controlled columns placement

Is there anything I can do more to get this PR merged?

mvorisek:
@greg0ire please reaprove, I rebased it from 2.11.x to 2.12.x branch.

> > 1. physical layout can be optimized by putting fixed size columns at the start of the table
> 
> But I'm not sure if this is also the same benefit in MySQL.

see https://stackoverflow.com/questions/894522/is-there-any-reason-to-worry-about-the-column-order-in-a-table - there is some performance improvement with controlled columns placement

Is there anything I can do more to get this PR merged?

mvorisek:
solved privately, contact me in PM if you are a company and want to pay to open source it, more work needed than this PR

PR #4278, opened 2020-09-21
Remove deprecated APIs

morozov:
|      Q       |   A
|------------- | -----------
| BC Break     | yes

1. A coding standard violation is fixed (looks introduced by https://github.com/doctrine/dbal/pull/4275).
2. `DBALException` was deprecated in https://github.com/doctrine/dbal/pull/4253.
3. `AbstractPlatform::supportsForeignKeyOnUpdate()` is removed. Originally it was removed in https://github.com/doctrine/dbal/pull/4234 but then mistakenly reintroduced in https://github.com/doctrine/dbal/pull/4275.
4. Back-ported #3255 from `master`.


greg0ire (approved):
I think I read somewhere that the CodeCov message only appears on lines with exceptions (technically, you exit the method kinda before reaching its end).

morozov:
I believe it's just that code is [not covered](https://codecov.io/gh/doctrine/dbal/src/f8395ef4023797a42c09e5e030038b3a2f5c21de/src/DBALException.php).

Issue #4283, opened 2019-08-12
Doctrine fails to declare Extended Property on SQL-Server array column (when table name is quoted) 


metaturso:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | possibly
| Version     | ORM v2.6.3, DBAL v2.9.2
| Database Server | Microsoft SQL-Server 2017
| Doctrine Driver | `pdo_sqlsrv`

#### Summary

The following error happens when using Doctrine's `SchemaTool` to construct an entity requiring Extended Properties which also has a quoted table name:

```
'An exception occurred while executing 'EXEC sp_addextendedproperty N'MS_Description', N'(DC2Type:array)', N'SCHEMA', 'dbo', N'TABLE', '[the-table]', N'COLUMN', chairs':
                                                                                                                                 
SQLSTATE[42000]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server] Object is invalid. Extended properties are not permitted on 'dbo.[the-table].chairs', or the object does not exist.'
```

The user has sufficient permissions to add Extended Properties to the database metadata. If I understand this right, this problem is happening because of the quoted table name being passed to the stored procedure as a string (i.e. the databases recognises `[the-table] !== "[the-table]"`.)

Stopping Doctrine from quoting the table name **fixes the problem**, alas I can't afford to rename the table.

#### Current behavior

The database server refuses to execute the query because it does not recognise the referenced object.

#### How to reproduce

1. Declare an entity whose name is a quoted symbol, e.g. 
```php
/**
 *@Table(name="`the-table`")
 */
class TheTable 
{
}
```
2. Add a column requiring the use of Extended Properties by Doctrine

```php
class TheTable
{
 /**
  * @ORM\Column(type="array", name="chairs")
  */
 private $chairs;
}
```

3. Use the Schema Tool to create the tables

```php
$schemaTool = new SchemaTool($entityManager);
$schemaTool->createSchema([
 $entityManager->getMetadataForClass(TheTable::class)
]); 
```

#### Expected behavior

Doctrine can successfully attach the Extended Properties that it needs to the column.

SenseException:
Depending on the test this seems to be a DBAL issue.

PR #4284, opened 2020-09-22
Fixes invalid T-SQL column comment for quoted table names.

metaturso:
|      Q           |   A
|------------- | -----------
| Type            | bug
| BC Break     | no
| Fixed issues | #4283 

#### Summary

This merge request fixes the issue whereby a quoted table or schema name was getting converted into a string literal, resulting in a T-SQL statement referencing an invalid table name like `N'[table_name]'`. 

The issues happened whenever the entity table name contained [Quoted Reserved Words](https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/reference/basic-mapping.html#quoting-reserved-words), for example:

```php
/**
 * @ORM\Table(name="`the-model-table`")
 */
class TheModel
{
}
```

The issue is fixed within the `SQLServer2012Platform` base class by inserting the schema or table names as T-SQL identifiers when generating the call to `sp_addextendedproperty` for quoted names. 

In other words, because Doctrine converts a quoted name like `` `table-name` `` into a valid T-SQL identifier like `[table-name]` we needn't wrap the identifier into a string literal before passing it to `getAddExtendedPropertySQL`.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL4283Test.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL4283Test.php
@@ -1,0 +1,147 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL\Functional\Ticket;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver;
+use Doctrine\DBAL\Platforms\SQLServer2012Platform;
+use Doctrine\DBAL\Schema\Column;
+use Doctrine\DBAL\Schema\ColumnDiff;
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Schema\SQLServerSchemaManager;
+use Doctrine\DBAL\Schema\TableDiff;
+use Doctrine\DBAL\Types\Type;
+use Doctrine\Tests\DbalFunctionalTestCase;
+
+/**
+ * Fixes DBAL issue #4283: Doctrine fails to declare Extended Properties on SQL-Server when table name is quoted.
+ *
+ * The problem is in the SQLServerPlatform which incorrectly quotes the SCHEMA and TABLE object names
+ * when they contain Reserved Words.
+ *
+ * The fix is to keep the SCHEMA and TABLE as identifiers in the T-SQL query.
+ *
+ * TODO: 1. Trigger the bug through {@see SQLServerPlatform::getAlterColumnCommentSQL}
+ * TODO: 2. It's possible that the COLUMN object needs the same treatment.
+ *
+ * @link https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/reference/basic-mapping.html#quoting-reserved-words
+ * @link https://github.com/doctrine/dbal/issues/4283
+ */
+class DBAL4283Test extends DbalFunctionalTestCase
+{
+    /* @var Schema */
+    private $schema;
+
+    /* @var SQLServer2012Platform */
+    private $platform;
+
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        $this->platform = new SQLServer2012Platform();
+
+        $driver        = $this->createStub(Driver::class);
+        $connection    = new Connection([], $driver);
+        $schemaManager = new SQLServerSchemaManager($connection, $this->platform);
+
+        $metadataSchemaConfig = $schemaManager->createSchemaConfig();
+        $metadataSchemaConfig->setExplicitForeignKeyIndexes(false);
+
+        $this->schema = new Schema([], [], $metadataSchemaConfig);
+    }
+
+    /**
+     * @dataProvider tableNameProvider
+     */
+    public function testAddExtendedPropertyToColumn(string $tableName, string $schemaSQL, string $tableSQL): void
+    {
+        self::assertEquals(
+            sprintf(
+                "EXEC sp_addextendedproperty N'MS_Description', N'(DC2Type:array)', "
+                . "N'SCHEMA', %s, N'TABLE', %s, N'COLUMN', column_name",
+                $schemaSQL,
+                $tableSQL
+            ),
+            $this->getAddExtendedPropertyQuery($tableName)
+        );
+    }
+
+    /**
+     * @dataProvider tableNameProvider
+     */
+    public function testRemoveExtendedPropertyFromColumn(string $tableName, string $schemaSQL, string $tableSQL): void
+    {
+        self::assertEquals(
+            sprintf(
+                "EXEC sp_dropextendedproperty N'MS_Description', "
+                . "N'SCHEMA', %s, N'TABLE', %s, N'COLUMN', column_name",
+                $schemaSQL,
+                $tableSQL
+            ),
+            $this->getDropExtendedPropertyQuery($tableName)
+        );
+    }
+
+    /**
+     * @dataProvider tableNameProvider
+     */
+    public function testUpdateExtendedPropertyFromColumn(string $tableName, $schemaSQL, string $tableSQL): void
+    {
+        self::assertEquals(
+            sprintf(
+                "EXEC sp_updateextendedproperty N'MS_Description', "
+                . "N'SCHEMA', %s, N'TABLE', %s, N'COLUMN', column_name",
+                $schemaSQL,
+                $tableSQL
+            ),
+            $this->getUpdateExtendedPropertyQuery($tableName)
+        );
+    }
+
+    public function getAddExtendedPropertyQuery(string $tableName): string
+    {
+        $table = $this->schema->createTable($tableName);
+        $table->addColumn('column_name', 'array');
+
+        return $this->platform->getCreateTableSQL($table)[1];
+    }
+
+    public function getDropExtendedPropertyQuery(string $tableName): string
+    {
+        // Change column type to something that needn't an Extended Property declaration.
+        $table = $this->schema->createTable($tableName);
+        $table->addColumn('column_name', 'array');
+
+        $newColumn = new Column('column_name', Type::getType('string'));
+        $tableDiff = new TableDiff($tableName, [], [
+            new ColumnDiff($table->getName(), $newColumn, [], $table->getColumn('column_name'))
+        ]);
+
+        return $this->platform->getAlterTableSQL($tableDiff)[1];
+    }
+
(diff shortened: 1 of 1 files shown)


src/Platforms/SQLServer2012Platform.php
```diff
@@ -363,19 +363,24 @@ public function getCreatePrimaryKeySQL(Index $index, $table): string
      * as column comments are stored in the same property there when
      * specifying a column's "Description" attribute.
      *
-     * @param string $tableName  The quoted table name to which the column belongs.
+     * @param string $tableName The quoted table name to which the column belongs.
      * @param string $columnName The quoted column name to create the comment for.
-     * @param string $comment    The column's comment.
```
metaturso, line 368: Untabified by accident. Should I undo this change?

metaturso: Never mind, the PHPCS caught this as an error. I'm going to undo this.

greg0ire:
I think you should target 2.11.x

metaturso:
I rebased my branch onto `2.11.x`. (`git rebase --onto 2.11.x master fix/tsql-comment-column`) and it's possible that this didn't play nice with AppVeyor's mergeability checks. 

I'm happy to open a fresh MR targeting the desired branch if you  prefer.

There's a few places left in `SQLServerPlatform` that need the same fix (e.g. callers of `get*ExtendedPropertySQL` methods.) I'll write tests and fix for those too.

How would you feel about a method which initialises both variables at once, e.g. `[$schemaSQL, $tableSQL] = $this->doodad($tableName)`?

greg0ire:
> I'm happy to open a fresh MR targeting the desired branch if you prefer.

Please don't, it's great the way it is 👍

I will close and reopen, it might help with the checks.

tests/Functional/Ticket/DBAL4283Test.php
```diff
@@ -0,0 +1,77 @@
...
+{
+    public function testAddingExtendedPropertyToUnquotedTableName(): void
+    {
+        $createTableSQL = $this->getCreateTableSQL('table_name');
+
+        self::assertEquals(
+            "EXEC sp_addextendedproperty N'MS_Description', N'(DC2Type:array)', "
+            . "N'SCHEMA', 'dbo', N'TABLE', 'table_name', N'COLUMN', column_name",
+            $createTableSQL
+        );
```
greg0ire, line 35:
```suggestion
        self::assertEquals(
            "EXEC sp_addextendedproperty N'MS_Description', N'(DC2Type:array)', "
            . "N'SCHEMA', 'dbo', N'TABLE', 'table_name', N'COLUMN', column_name",
            $this->getCreateTableSQL('table_name')
        );
```

tests/Functional/Ticket/DBAL4283Test.php
```diff
@@ -0,0 +1,77 @@
...
+    public function testAddingExtendedPropertyToQuotedTableName(): void
+    {
+        $createTableSQL = $this->getCreateTableSQL('`quoted-table-name`');
+
+        self::assertEquals(
+            "EXEC sp_addextendedproperty N'MS_Description', N'(DC2Type:array)', "
+            . "N'SCHEMA', 'dbo', N'TABLE', [quoted-table-name], N'COLUMN', column_name",
+            $createTableSQL
+        );
+    }
```
greg0ire, line 47:
```suggestion
        self::assertEquals(
            "EXEC sp_addextendedproperty N'MS_Description', N'(DC2Type:array)', "
            . "N'SCHEMA', 'dbo', N'TABLE', [quoted-table-name], N'COLUMN', column_name",
            $this->getCreateTableSQL('`quoted-table-name`')
        );
    }
```

tests/Functional/Ticket/DBAL4283Test.php
```diff
@@ -0,0 +1,77 @@
...
+
+    public function testAddingExtendedPropertyToQuotedTableInSchema(): void
+    {
+        $createTableSQL = $this->getCreateTableSQL('`custom.quoted-table-name`');
+
+        self::assertEquals(
+            "EXEC sp_addextendedproperty N'MS_Description', N'(DC2Type:array)', "
+            . "N'SCHEMA', [custom], N'TABLE', [quoted-table-name], N'COLUMN', column_name",
+            $createTableSQL
+        );
```
greg0ire, line 57:
```suggestion
        self::assertEquals(
            "EXEC sp_addextendedproperty N'MS_Description', N'(DC2Type:array)', "
            . "N'SCHEMA', [custom], N'TABLE', [quoted-table-name], N'COLUMN', column_name",
            $this->getCreateTableSQL('`custom.quoted-table-name`')
        );
```

tests/Functional/Ticket/DBAL4283Test.php
```diff
@@ -0,0 +1,77 @@
...
+    }
```
greg0ire, line 58: Would it make sense to use a data provider instead? All 3 tests have the same logic, it would be great to be able to see just what changes in one place.

tests/Functional/Ticket/DBAL4283Test.php
```diff
@@ -0,0 +1,77 @@
...
+
+    public function getCreateTableSQL(string $quotedName): string
+    {
+        $driver   = $this->createMock(Driver::class);
```
greg0ire, line 62: Since you perform no expectations, this is a stub:

```suggestion
        $driver   = $this->createStub(Driver::class);
```

morozov:
This change requires a functional test that runs the updated SQL queries by a real database, not making assertions against the generated SQL and using mocks.

metaturso:
Couldn't agree more!

The proposed unit test is good enough to show the problem exists not prove that it's fixed. In that sense, it helped expose the fact that the ALTER and DROP column/table cases are also bugged.

I won't get another chance to work on this probably until next weekend, when I plan to begin writing the new integration test against a real SQL Server instance before attempting a fix. I'll try to push the new version of the unit failing unit test sooner than that to arouse discussion about a clean fix.

In the meantime, could you let me know your requirements or preferences for this new test? For example, could you point me to an existing test that I can or should use as reference for level of abstraction, structure, assertion style, etc., just to get an idea?

On a related note, is the current Travis CI pipeline already configured to run SQL Server tests?

Thanks in advance for your time.

morozov:
> In the meantime, could you let me know your requirements or preferences for this new test? For example, could you point me to an existing test that I can or should use as reference for level of abstraction, structure, assertion style, etc., just to get an idea?

As far as I understand the problem, you'll need to create a table with a name that has to be quoted. Look for the tests that call `dropAndCreateTable()` under [tests/Doctrine/Tests/DBAL/Functional](https://github.com/doctrine/dbal/tree/2.11.x/tests/Doctrine/Tests/DBAL/Functional).

> On a related note, is the current Travis CI pipeline already configured to run SQL Server tests?

Yes, see the [Travis](https://travis-ci.org/github/doctrine/dbal/builds/729378606) and [AppVeyor](https://ci.appveyor.com/project/doctrine/dbal/builds/35345923) builds.

mvorisek:
I confirm this bug, @metaturso can you fix the PR so it can be merged quickly?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -360,13 +360,18 @@ public function getCreatePrimaryKeySQL(Index $index, $table)
...
         if (strpos($tableName, '.') !== false) {
             [$schemaSQL, $tableSQL] = explode('.', $tableName);
-            $schemaSQL              = $this->quoteStringLiteral($schemaSQL);
-            $tableSQL               = $this->quoteStringLiteral($tableSQL);
-        } else {
-            $schemaSQL = "'dbo'";
-            $tableSQL  = $this->quoteStringLiteral($tableName);
+        }
+
+        if (strpos($tableName, '[') === false) {
```
mvorisek, line 369: I think stripping the `[` `]` quotes would be better, as we can then quote it using `'` like we would normally do.

mvorisek, line 369: I think we should `preg_replace('~^\[|\]$~s', '', $tableName)` instead and quote as literal as we would normally do.

String literal seeems to be the official way to escape: https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-addextendedproperty-transact-sql?view=sql-server-ver15#b-adding-an-extended-property-to-a-column-in-a-table

mvorisek:
@metaturso please see https://github.com/doctrine/dbal/pull/4360 , I think we need to fix in a little more abstract way - please provide feedback, if you want to change anything or help with the tests, please send a PR directly to my branch. Michael

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
metaturso: > @metaturso please see #4360 , I think we need to fix in a little more abstract way - please provide feedback, if you want to change anything or help with the tests, please send a PR directly to my branch. Michael

@mvorisek thanks for stepping in with a patch for this, chap. Your fix in PR #4360 is reasonably self-contained and I'm happy to ditch my changes in favour of yours. 

There's definitely something about the amount of (un)escaping that's done at multiple levels that doesn't completely sit right with me, but fixing the data-flow is a much more pervasive change that's likely to break B/C or third-party code so I can't offer a better alternative.

I didn't get a chance to sit down and write the promised integration test. I'll commit and push the fixed version of the unit test, which runs against 2.11.x version of the codebase, during my lunch break. Feel free to steal this from my PR if that's of any use to you.

Should you decide not to use my test, please don't forget to cover for #4283 by adding test cases or fixture using i) table names with a custom schema name and ii) containing special characters or reserved keywords to your pull request.

metaturso:
I've applied all suggestions to the unit test and updated it to cover the *add*, *update* and *drop* Extended Property cases, @morozov and @greg0ire thanks for your feedback and suggestions. 

All changes to the SQL Server platform class have been removed from this branch because @mvorisek has already raised a comprehensive fix to the original issue (thanks for that.) The failing unit test has been left here for his convenience.

I reckon this PR can now be closed in favour of #4360.

mvorisek:
@metaturso please submit the tests as PR into https://github.com/mvorisek/dbal/tree/fix_mssql_comment_with_escaped_column branch, so your work can be integrated into #4360

metaturso:
Will do shortly. Close this PR at your discretion.

PR #4286, opened 2020-09-22
Fix BC break: QueryBuilder::andWhere() etc. should ignore empty strings

### LINKED ISSUE #4282: Skip empty expression parts
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.11.0

#### Summary

The commit 1d503253ff57987161f10d4f2c81b5a9889d48a4 introduced a breaking change.

[CompsiteExpression->with()](https://github.com/doctrine/dbal/blob/2.11.x/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php#L119) directly adds `parts`, whereas the old [CompositeExpression->addMultiple()](https://github.com/doctrine/dbal/blob/2.11.x/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php#L80) utilized [CompositeExpression->add()](https://github.com/doctrine/dbal/blob/2.11.x/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php#L98) which had an `empty`-check.

- has a guard
see https://github.com/doctrine/dbal/blob/2.11.x/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php#L100

- has no guard
see https://github.com/doctrine/dbal/blob/2.11.x/lib/Doctrine/DBAL/Query/Expression/CompositeExpression.php#L126

#### Current behaviour

`parts` with empty-string result in `AND ()` which is an invalid where-expression.

#### How to reproduce

- Instantiate QueryBuilder instance.
- provide an empty string to `andWhere`

```php
$queryBuilder->andWhere('', "foo = 'bar'")->getSQL();
```

#### Expected behaviour

Do not add empty where conditions in `CompositeExpression->with()` as `CompositeExpression->add()` did.

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #4282

#### Summary

The previous behaviour of these methods has been restored: they now ignore empty strings passed as parameters.

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -8,6 +8,7 @@ use Doctrine\DBAL\ParameterType;
 use Doctrine\DBAL\Query\Expression\CompositeExpression;
 use Doctrine\DBAL\Query\Expression\ExpressionBuilder;
 
+use function array_filter;
 use function array_key_exists;
 use function array_keys;
 use function array_unshift;
@@ -829,6 +830,7 @@ class QueryBuilder
     public function andWhere($where)
     {
         $args  = func_get_args();
+        $args  = array_filter($args); // https://github.com/doctrine/dbal/issues/4282
         $where = $this->getQueryPart('where');
 
         if ($where instanceof CompositeExpression && $where->getType() === CompositeExpression::TYPE_AND) {
@@ -862,6 +864,7 @@ class QueryBuilder
     public function orWhere($where)
     {
         $args  = func_get_args();
+        $args  = array_filter($args); // https://github.com/doctrine/dbal/issues/4282
         $where = $this->getQueryPart('where');
 
         if ($where instanceof CompositeExpression && $where->getType() === CompositeExpression::TYPE_OR) {
@@ -1010,6 +1013,7 @@ class QueryBuilder
     public function andHaving($having)
     {
         $args   = func_get_args();
+        $args   = array_filter($args); // https://github.com/doctrine/dbal/issues/4282
         $having = $this->getQueryPart('having');
 
         if ($having instanceof CompositeExpression && $having->getType() === CompositeExpression::TYPE_AND) {
@@ -1033,6 +1037,7 @@ class QueryBuilder
     public function orHaving($having)
     {
         $args   = func_get_args();
+        $args   = array_filter($args); // https://github.com/doctrine/dbal/issues/4282
         $having = $this->getQueryPart('having');
 
         if ($having instanceof CompositeExpression && $having->getType() === CompositeExpression::TYPE_OR) {
--- a/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
+++ b/tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php
@@ -949,4 +949,104 @@ class QueryBuilderTest extends DbalTestCase
 
         $qb->getSQL();
     }
+
+    public function testAndWhereEmptyStringStartingWithEmptyExpression(): void
+    {
+        $qb = new QueryBuilder($this->conn);
+
+        $qb->select('id')
+            ->from('foo');
+
+        $qb->andWhere('', 'a = b');
+
+        self::assertSame('SELECT id FROM foo WHERE a = b', $qb->getSQL());
+    }
+
+    public function testAndWhereEmptyStringStartingWithNonEmptyExpression(): void
+    {
+        $qb = new QueryBuilder($this->conn);
+
+        $qb->select('id')
+            ->from('foo')
+            ->where('a = b');
+
+        $qb->andWhere('', 'c = d');
+
+        self::assertSame('SELECT id FROM foo WHERE (a = b) AND (c = d)', $qb->getSQL());
+    }
+
+    public function testOrWhereEmptyStringStartingWithEmptyExpression(): void
+    {
+        $qb = new QueryBuilder($this->conn);
+
+        $qb->select('id')
+            ->from('foo');
+
+        $qb->orWhere('', 'a = b');
+
+        self::assertSame('SELECT id FROM foo WHERE a = b', $qb->getSQL());
+    }
+
+    public function testOrWhereEmptyStringStartingWithNonEmptyExpression(): void
+    {
+        $qb = new QueryBuilder($this->conn);
+
+        $qb->select('id')
+            ->from('foo')
+            ->where('a = b');
+
+        $qb->orWhere('', 'c = d');
+
+        self::assertSame('SELECT id FROM foo WHERE (a = b) OR (c = d)', $qb->getSQL());
+    }
+
+    public function testAndHavingEmptyStringStartingWithEmptyExpression(): void
+    {
+        $qb = new QueryBuilder($this->conn);
+
+        $qb->select('id')
+            ->from('foo');
+
+        $qb->andHaving('', 'a = b');
+
+        self::assertSame('SELECT id FROM foo HAVING a = b', $qb->getSQL());
+    }
+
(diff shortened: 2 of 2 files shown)


lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1393,4 +1399,21 @@ public function __clone()
...
+     * @param array<mixed> $args
+     *
+     * @return array<mixed>
+     *
+     * @todo remove for 3.0
+     */
+    private static function filterEmptyArgs(array $args): array
+    {
+        return array_values(array_filter($args, static function ($value) {
+            return ! empty($value);
```
morozov, line 1416: Isn't it the default behavior of `array_filter` without a callback?

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1393,4 +1399,21 @@ public function __clone()
...
+     * Filters out empty values from the given argument list.
+     * Fixes a BC break introduces in 2.11: https://github.com/doctrine/dbal/issues/4282
+     *
+     * @param array<mixed> $args
+     *
+     * @return array<mixed>
+     *
+     * @todo remove for 3.0
+     */
+    private static function filterEmptyArgs(array $args): array
```
morozov, line 1413: Why does the method need to be static?

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -829,6 +831,7 @@ public function where($predicates)
     public function andWhere($where)
     {
         $args  = func_get_args();
+        $args  = self::filterEmptyArgs($args);
```
morozov, line 834: As far as I understand, we need to cover both branches with the `with()` and `::and()` because the latter used to be implemented by `->addMultiple()` → `->add()` as well. Would it be possible to cover it with a test as well?

Once implemented, please replicate the test for the rest of the changed methods.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Because phpcs wants it so!

morozov: Please double-check. I can't reproduce this by making the method non-static.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Indeed! Updated. Should we keep the `filterEmptyArgs()` anyway, even if it's very simple, just as documentation?

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -1393,4 +1399,19 @@ public function __clone()
...
+     *
+     * @param array<mixed> $args
+     *
+     * @return array<mixed>
+     *
+     * @todo remove for 3.0
+     */
+    private static function filterEmptyArgs(array $args): array
+    {
+        return array_values(array_filter($args));
```
morozov, line 1415: Is `array_values()` necessary here? The array keys are not used anywhere explicitly, and PHP seems to ignore the keys of the array passed via `...`.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Tests added!

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: If it ends up being just `array_filter()`, it could be used directly w/o being wrapped in a documenting method. But I don't have a strong opinion.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Indeed! Because PHP errors when non-numeric keys are used, I was pretty sure that it would error if non-sequentials keys were used... I was wrong. Fixed.

The above question stands even more then: should we keep `filterEmptyArgs()` as documentation? Or should we use `array_filter()` inline? But we lose the link to the issue and the TODO.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Okay, I've inlined it with a link to the issue and a TODO. See if you're happy with that!

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: It's [here](https://github.com/BenMorel/dbal/runs/1151940985):
> [error]Closure not using "$this" should be declared static.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -829,6 +830,7 @@ public function where($predicates)
     public function andWhere($where)
     {
         $args  = func_get_args();
+        $args  = array_filter($args); // TODO remove for 3.0 https://github.com/doctrine/dbal/issues/4282
```
morozov, line 833: Instead of a TODO, it's better to file a release blocker for 3.0. It's fine to leave a comment explaining why this is needed though.

BenMorel: I don't know what a release blocker is. Is it an issue that mentions that this code should be removed?

morozov: Yes, please file an issue. I'll mark it as a blocker and assign it to the 3.0.0 milestone. This way, we'll have it tracked in the issue tracker, not in the source code.

BenMorel: Anyway I've removed the TODO now. Should I squash?

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: Are you sure the four added tests cover the four updated methods with two execution branches in each?

BenMorel: There are 8 tests added!

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: Yep, looks good. I also ran the test on the previous release to make the patch doesn't fix something that's not broken:
```
$ git checkout 2.10.2
HEAD is now at aab745e7b Release v2.10.2

$ git diff 2.11.x 19f33f7072cbbd6edf103551298ff12a068209c2 -- tests | patch
patching file tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php

$ phpunit tests/Doctrine/Tests/DBAL/Query/QueryBuilderTest.php

.............................................................................     77 / 77 (100%)

Time: 1 second, Memory: 12.00 MB

OK (77 tests, 111 assertions)
```

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: I'm sorry. I think I saw what I expected to see.

lib/Doctrine/DBAL/Query/QueryBuilder.php
morozov: Looks good.

lib/Doctrine/DBAL/Query/QueryBuilder.php
BenMorel: Squashed!

morozov (approved):

morozov:
Thanks, @BenMorel!

PR #4287, opened 2020-09-22
Restore PDOStatement::quote() for backward compatibility

### LINKED ISSUE #4281: BC Break after removing method \Doctrine\DBAL\Driver\PDOConnection::quote
The problem appeared in releases 2.10.3 [commit](https://github.com/doctrine/dbal/pull/3287/commits/85bc13deb0abbbf3a99347ab9b757c8c1a1ebd96)

The problem is that now, with strict mode enabled, nothing but string cannot be passed to \Doctrine\DBAL\Driver\PDOConnection::quote. These changes break backward compatibility and should not have been recorded as minor changes.

morozov:
Fixes #4281.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/PDOConnection.php
+++ b/lib/Doctrine/DBAL/Driver/PDOConnection.php
@@ -5,6 +5,7 @@ namespace Doctrine\DBAL\Driver;
 use Doctrine\DBAL\Driver\Connection as ConnectionInterface;
 use Doctrine\DBAL\Driver\PDO\Exception;
 use Doctrine\DBAL\Driver\PDO\Statement;
+use Doctrine\DBAL\ParameterType;
 use PDO;
 use PDOException;
 use PDOStatement;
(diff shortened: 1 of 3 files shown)


greg0ire (approved):
The code looks good

> well, int is a valid type according to the API

Is it really? I don't know what phpdoc we inherit but the signature says this should be a string: https://www.php.net/manual/en/pdo.quote.php

Also, it's a bit weird that PHPstan does not complain for this test… maybe it's a new rule, or maybe it's a level above the one we currently use.

morozov:
> Is it really? I don't know what phpdoc we inherit but the signature says this should be a string:

It is. This is the phpdoc `PDOConnection` inherits that says it's a `mixed`: https://github.com/doctrine/dbal/blob/b5e15999e8f626b9a50edef142fffc7be57d0211/lib/Doctrine/DBAL/Driver/Connection.php#L31-L39

Take this code for example:
```php
$conn = DriverManager::getConnection(
    'driver' => 'pdo_sqlite',
    'dbname' => 'sqlite::memory:',
)->getWrappedConnection();
```

For the caller, `$conn` above is just a `Driver\Connection` (which accepts `mixed`). The caller doesn't know that it extends PDO (which accepts `string`).

Furthermore, some other drivers explicitly handle the non-string values in `quote()`:
https://github.com/doctrine/dbal/blob/b5e15999e8f626b9a50edef142fffc7be57d0211/lib/Doctrine/DBAL/Driver/OCI8/OCI8Connection.php#L129-L133

SenseException (approved):
Since 2.10.x won't get this fix, this PR looks goot to me.

greg0ire:
Thanks for the explanation @morozov , very clear!

PR #4292, opened 2020-09-23
Add support for read-only SQLite databases

Ma27:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | n/a

#### Summary

Since PHP 7.3 it's posible to open SQLite databases in read-only
mode[1]. I figured that it makes sense to add a driverOption for
`pdo_sqlite` to support this.

The idea behind it is to declare `read_only: true` in e.g. a sf-bundle
config without having to use the constants in the config files.

[1] https://github.com/php/php-src/commit/ed2f6510da7c68b5690c60344cbb9fe73241592a

### DIFF
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -172,6 +172,10 @@ pdo_sqlite
 -  ``memory`` (boolean): True if the SQLite database should be
    in-memory (non-persistent). Mutually exclusive with ``path``.
    ``path`` takes precedence.
+-  ``driverOptions`` (array): Arbitrary options that are directly
+   passed to ``PDO`` including some custom DBAL options:
+      - ``readOnly`` (boolean): Whether to open the database in
+        read-only mode.
 
 pdo_mysql
 ^^^^^^^^^
--- a/src/Driver/PDO/SQLite/Driver.php
+++ b/src/Driver/PDO/SQLite/Driver.php
@@ -5,6 +5,7 @@ namespace Doctrine\DBAL\Driver\PDO\SQLite;
 use Doctrine\DBAL\Driver\AbstractSQLiteDriver;
 use Doctrine\DBAL\Driver\PDO\Connection;
 use Doctrine\DBAL\Platforms\SqlitePlatform;
+use PDO;
 
 use function array_merge;
 
@@ -30,6 +31,14 @@ final class Driver extends AbstractSQLiteDriver
                 $driverOptions['userDefinedFunctions']
             );
             unset($driverOptions['userDefinedFunctions']);
+        }
+
+        if (isset($driverOptions['readOnly'])) {
+            if ($driverOptions['readOnly']) {
+                $driverOptions[PDO::SQLITE_ATTR_OPEN_FLAGS] = PDO::SQLITE_OPEN_READONLY;
+            }
+
+            unset($driverOptions['readOnly']);
         }
 
         $connection = new Connection(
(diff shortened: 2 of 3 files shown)


greg0ire:
> BC Break | yes

Can you point out the BC-break? Also why did you target `master` and not `3.0.x`?

Ma27:
Oof sorry, this is obviously not a BC break, fixed the PR message.
Will retarget the branch tomorrow :)

Ma27:
Rebased onto `3.0.x` and fixed the CS violations.
Anything else I missed? :)

greg0ire (commented):
Terrific commit message!

I think the new `readOnly` option should be documented at https://github.com/doctrine/dbal/blob/2.11.x/docs/en/reference/configuration.rst#pdo_sqlite

Ma27:
Updated. Unfortunately I'm not fluent in reStructuredText so you may have a better suggestion on how to document this :)

morozov:
Since the DBAL already supports passing driver options, why does it need to declare its own version of the parameter?

Ma27:
I think that this is a helpful convenience flag and way easier to keep in mind than how exactly to configure PDO to open a sqlite in read-only mode. Also, I already opened a follow-up PR in DoctrineBundle that adds a `read_only` argument to the bundle-config to get rid of this `!php/const PDO::foobar` hackery for that (which makes it probably clearer what I'm thinking of :))

morozov:
> I think that this is a helpful convenience flag and way easier to keep in mind than how exactly […]

This is arguable. How do I remember if it's `readOnly`, `readonly`, `read_only`, or `read-only`?

DBAL provides only the configuration implemented in PHP which it directly passes to PDO. PDO uses constants for configuration which are better than named array elements because:
1. Constants are discoverable via static analysis (e.g. by IDEs).
2. An invalid reference to a constant will be reported by PHP immediately, while a non-matching string parameter will be just ignored by the library.

> get rid of this `!php/const PDO::foobar` hackery

This is exactly how it should be done IMO.

PDO has a lot of constants, and there's no point to make up another better version of it.

morozov:
We can reconsider this if this configuration parameter gets promoted to the abstraction level (e.g. supported by multiple drivers). Closing for now.

Ma27:
> DBAL provides only the configuration implemented in PHP which it directly passes to PDO. PDO uses constants for configuration which are better than named array elements because:

OTOH DBAL provides - from my understanding at least - a high-level abstraction to configure database connections and I'd consider this argument also part for such a configuration (such as `path`, `memory` etc that are used to construct the DSN).

I know that this is not part of the DSN itself, but I'd consider this a relevant param for the connection I'd like to have exposed as well (especially for e.g. a sf-bundle where I can introspect those kinds of options and have them validated) - although this is just a driver option. Though I totally understand if you have a different opinion as maintainer of this library, I just want to clarify my point :) 

> We can reconsider this if this configuration parameter gets promoted to the abstraction level (e.g. supported by multiple drivers).

You mean if hypothetically all other DB drivers would consume this argument as well?

I'd like to raise the question though how to move forward from here: would it make sense to only add this to DoctrineBundle given my reasoning (opened https://github.com/doctrine/DoctrineBundle/pull/1213 as follow-up back when I filed this PR) or should I close this as well?

morozov:
> OTOH DBAL provides - from my understanding at least - a high-level abstraction to configure database connections […]

Right, and the SQLite _driver-specific_ option is not part of that abstraction. DBAL just allows passing driver-specific options to the underlying driver to fine-tune the connection.

> You mean if hypothetically all other DB drivers would consume this argument as well?

Yes.

> would it make sense to only add this to DoctrineBundle

I'm not closely familiar with this package but the logic you're trying to implement belongs to the layer that supports non-PHP configurations and needs to translate the "readOnly" string to the PHP constant.

Ma27:
Thanks for the clarification!

PR #4293, opened 2020-09-24
Add Result::fetchAllKeyValue() and ::iterateKeyValue()

### LINKED ISSUE #4289: Add a fetch mode methods for "PDO::FETCH_KEY_PAIR"
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

Let's assume a sql, summarizing entity ids and a special count. Then I need the workflow like the default PDO does:

<pre lang=php>
public function countOrders()
{
    $sql = 'SELECT `e`.`id`, COUNT(`o`.`id`) 
        FROM `event` AS `e` 
        LEFT JOIN `order` AS `o` ON `o`.`event_id` = `e`.`id`
        WHERE `o`.`deleted` = 0
        GROUP BY `e`.`id` ';

    $dbConnection = $this->getEntityManager()->getConnection();
    $sqlStatement = $dbConnection->prepare($sql);
    $sqlStatement->execute();

    $counts = $sqlStatement->fetchAll(PDO::FETCH_KEY_PAIR); 
    $sqlStatement->free();

    return $counts;
}
</pre>

The result was an array with the ids as keys and counts as values, like:
<pre lang=php>
array (2) [
    11 => integer 10
    22 => integer 5
]
</pre>

Since the old fetch methods are deprecated, it would be great to implement 
an adequate method to the PDO feature to reduce boilerplate code.

Here is an quote from @morozov of #4019 
<pre lang=php>
$conn = new PDO('sqlite::memory:');
$stmt = $conn->query('SELECT 1');

$stmt->fetchAll(PDO::FETCH_KEY_PAIR);

// PDOStatement::fetchAll(): SQLSTATE[HY000]: General error: PDO::FETCH_KEY_PAIR fetch mode requires the result set to contain extactly 2 columns
</pre>

As @morozov mentioned:
> Looks like PDO is strict enough about the number of columns. I think we can implement this in DBAL at the wrapper level which will automatically cover all supported drivers.

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

Fixes #4289.

The behavior of the newly added methods is a replacement of `PDO::FETCH_KEY_PAIR` which is implemented at the wrapper level and is available for all supported drivers.

Unlike PDO which requires the result to have exactly two columns, DBAL will require at least two columns in order to allow the ROWNUM columns used by the platforms that do not natively support LIIMIT queries (Oracle, IBM DB2).

### DIFF
--- a/docs/en/reference/data-retrieval-and-manipulation.rst
+++ b/docs/en/reference/data-retrieval-and-manipulation.rst
@@ -379,6 +379,25 @@ Execute the query and fetch all results into an array:
     )
     */
 
+fetchAllKeyValue()
+~~~~~~~~~~~~~~~~~~
+
+Execute the query and fetch the first two columns into an associative array as keys and values respectively:
+
+.. code-block:: php
+
+    <?php
+    $users = $conn->fetchAllKeyValue('SELECT username, password FROM user');
+
+    /*
+    array(
+      'jwage' => 'changeme',
+    )
+    */
+
+.. note::
+   All additional columns will be ignored and are only allowed to be selected by DBAL for its internal purposes.
+
 fetchNumeric()
 ~~~~~~~~~~~~~~
 
@@ -424,6 +443,21 @@ Retrieve associative array of the first result row.
     */
 
 There are also convenience methods for data manipulation queries:
+
+iterateKeyValue()
+~~~~~~~~~~~~~~~~~
+
+Execute the query and iterate over the first two columns as keys and values respectively:
+
+.. code-block:: php
+
+    <?php
+    foreach ($conn->iterateKeyValue('SELECT username, password FROM user') as $username => $password) {
+        // ...
+    }
+
+.. note::
+   All additional columns will be ignored and are only allowed to be selected by DBAL for its internal purposes.
 
 delete()
 ~~~~~~~~~
--- a/tests/Functional/Connection/FetchTest.php
+++ b/tests/Functional/Connection/FetchTest.php
@@ -2,6 +2,8 @@
 
 namespace Doctrine\DBAL\Tests\Functional\Connection;
 
+use Doctrine\DBAL\Exception\NoKeyValue;
+use Doctrine\DBAL\Platforms\SQLServer2012Platform;
 use Doctrine\DBAL\Tests\FunctionalTestCase;
 use Doctrine\DBAL\Tests\TestUtil;
 
@@ -77,6 +79,41 @@ class FetchTest extends FunctionalTestCase
         ], $this->connection->fetchAllAssociative($this->query));
     }
 
+    public function testFetchAllKeyValue(): void
+    {
+        self::assertEquals([
+            'foo' => 1,
+            'bar' => 2,
+            'baz' => 3,
+        ], $this->connection->fetchAllKeyValue($this->query));
+    }
+
+    /**
+     * This test covers the requirement for the statement result to have at least two columns,
+     * not exactly two as PDO requires.
+     */
+    public function testFetchAllKeyValueWithLimit(): void
+    {
+        $platform = $this->connection->getDatabasePlatform();
+
+        if ($platform instanceof SQLServer2012Platform) {
+            self::markTestSkipped('See https://github.com/doctrine/dbal/issues/2374');
+        }
+
+        $query = $platform->modifyLimitQuery($this->query, 1, 1);
+
+        self::assertEquals(['bar' => 2], $this->connection->fetchAllKeyValue($query));
+    }
+
+    public function testFetchAllKeyValueOneColumn(): void
+    {
+        $sql = $this->connection->getDatabasePlatform()
+            ->getDummySelectSQL();
+
+        $this->expectException(NoKeyValue::class);
+        $this->connection->fetchAllKeyValue($sql);
+    }
+
     public function testFetchFirstColumn(): void
     {
         self::assertEquals([
@@ -111,6 +148,24 @@ class FetchTest extends FunctionalTestCase
                 'b' => 3,
             ],
         ], iterator_to_array($this->connection->iterateAssociative($this->query)));
+    }
+
+    public function testIterateKeyValue(): void
+    {
+        self::assertEquals([
+            'foo' => 1,
+            'bar' => 2,
+            'baz' => 3,
+        ], iterator_to_array($this->connection->iterateKeyValue($this->query)));
+    }
+
+    public function testIterateKeyValueOneColumn(): void
+    {
+        $sql = $this->connection->getDatabasePlatform()
+            ->getDummySelectSQL();
+
+        $this->expectException(NoKeyValue::class);
+        iterator_to_array($this->connection->iterateKeyValue($sql));
     }
 
     public function testIterateColumn(): void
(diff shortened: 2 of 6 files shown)


docs/en/reference/data-retrieval-and-manipulation.rst
```diff
@@ -379,6 +379,22 @@ Execute the query and fetch all results into an array:
     )
     */
 
+fetchAllKeyValue()
+~~~~~~~~~~~~~~~~~~~~~
```
greg0ire, line 383:
```suggestion
~~~~~~~~~~~~~~~~~~
```

docs/en/reference/data-retrieval-and-manipulation.rst
```diff
@@ -425,6 +441,18 @@ Retrieve associative array of the first result row.
 
 There are also convenience methods for data manipulation queries:
 
+iterateKeyValue()
+~~~~~~~~~~~~~~~~~~~~~
```
greg0ire, line 445:
```suggestion
~~~~~~~~~~~~~~~~~
```

greg0ire (commented):
> Unlike PDO which requires the result to have exactly two columns, DBAL will require at least two columns in order to allow the ROWNUM columns used by the platforms that do not natively support LIIMIT queries (Oracle, IBM DB2).

Can you please elaborate on this? I'm unfamiliar with these platforms. Is that kind of column always returned in a result set?

morozov:
> Can you please elaborate on this? I'm unfamiliar with these platforms. Is that kind of column always returned in a result set?

Yes. It's appended to the inner query for counting rows and used by the outer query for filtering.

For instance, on DB2:
```php
    'driver'   => 'ibm_db2',
    'user'     => 'db2inst1',
    'password' => 'Passw0rd',
    'dbname'   => 'doctrine',
]);

$sql = $conn->getDatabasePlatform()
    ->modifyLimitQuery('SELECT TEST_INT, TEST_STRING from FETCH_TABLE', 1, 1);

echo $sql, PHP_EOL; // formatted for readability
/*
SELECT db22.*
FROM (
    SELECT db21.*, ROW_NUMBER() OVER () AS DC_ROWNUM FROM (
        SELECT TEST_INT, TEST_STRING from FETCH_TABLE
    ) db21
) db22
WHERE db22.DC_ROWNUM >= 2
  AND db22.DC_ROWNUM <= 2
*/


$data = $conn->fetchAllAssociative($sql);
var_dump($data);
/*
array(1) {
  [0] =>
  array(3) {
    'TEST_INT' =>
    int(2)
    'TEST_STRING' =>
    string(3) "foo"
    'DC_ROWNUM' =>
    string(1) "2"
  }
}
*/
```
Despite the fact that we selected two columns, the result contains three.

greg0ire:
I see, the DBAL does that here: https://github.com/doctrine/dbal/blob/0d4e1a8b29dd987704842f0465aded378f441dca/lib/Doctrine/DBAL/Platforms/DB2Platform.php#L816-L820

Thanks for the explanation, this looks good.

Good point about adding a test also, I think you discovered a bug :sweat_smile:

greg0ire (dismissed):

morozov:
> I think you discovered a bug :sweat_smile:

Yeah, 4½ years ago (https://github.com/doctrine/dbal/issues/2374) :sweat_smile:

morozov:
@tswcode does it look like what you're looking for?

tswcode:
Indeed, that looks promising. I will test it as soon as I am back at work (on monday).  :)

tests/Functional/Connection/FetchTest.php
```diff
@@ -77,6 +79,41 @@ public function testFetchAllAssociative(): void
         ], $this->connection->fetchAllAssociative($this->query));
     }
 
+    public function testFetchAllKeyValue(): void
+    {
+        self::assertEquals([
```
SenseException, line 84: Depending on the answer about a defined behaviour of queries with more than 2 columns, tests need to be created to check that the methods behave as expected with more than two columns.

SenseException (dismissed):
My thoughts about the behaviour of these methods are written to the related lines of code.

I'm not sure about missing tests for the catched exceptions for `rowCount()` and `columnCount()`. There might be value that a `DriverException` is only internally handled and that the methods are converting them to the expected exceptions.

docs/en/reference/data-retrieval-and-manipulation.rst
```diff
@@ -379,6 +379,22 @@ Execute the query and fetch all results into an array:
     )
     */
 
+fetchAllKeyValue()
+~~~~~~~~~~~~~~~~~~
+
+Execute the query and fetch the first two columns into an associative array as keys and values respectively:
```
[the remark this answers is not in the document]
morozov: > What happens to the rest of the columns?

Nothing. As well as nothing happens to other columns when the first column is fetched via `fetchFirstColumn()`.

> it should be put into the documentation

How would you put it?

tests/Functional/Connection/FetchTest.php
morozov: This case is covered by `testFetchAllKeyValueWithLimit()`, and this is the sole reason why allow more than two columns. Do you think we should test this case more explicitly?

morozov:
> I'm not sure about missing tests for the catched exceptions for `rowCount()` and `columnCount()`. There might be value that a `DriverException` is only internally handled and that the methods are converting them to the expected exceptions.

The tests are missing because there's no clear understanding of how these methods could throw an exception in any of the existing drivers. Not that it's impossible, they just were never added.

The general concept is that the driver-level methods may throw driver exceptions while the wrapper should convert them into a generic or specialized wrapper-level exception (e.g. `ConnectionLost`). The only reason I'm making this change now is that the new fetch mode calls `columnCount()` that doesn't handle the documented driver exception. But the a method in the wrapper layer cannot throw a driver exception, so it must be handled. The only right place to handle the exception from the driver `columnCount()` is in the wrapper `columnCount()`.

There's absolutely a lot of work on exception handling (#4160 is just the beginning).

tests/Functional/Connection/FetchTest.php
SenseException: If an explicit more than 2 columns are the same as with your DB2 use case, then there isn't any additional test necessary

docs/en/reference/data-retrieval-and-manipulation.rst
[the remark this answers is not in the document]
SenseException:
```
.. note::
   Any additional column to the key and value columns will be ignored by the method.
```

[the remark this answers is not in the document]
morozov: While I agree that this note is better than nothing, I wouldn't like the users to rely on this behavior just because it's currently implemented. How about this:
> All additional columns will be ignored and are only allowed to be selected by DBAL for its internal purposes.

greg0ire (approved):

tswcode:
Unfortunately I was busy this week, I just noticed that this feature has been added to the branch 3.0.x.

@morozov Could you also add it to the branch 2.11.x ?

morozov:
@tswcode see #4338.

PR #4294, opened 2020-09-24
Remove Abstraction\Result

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

The interface is proposed for deprecation in https://github.com/doctrine/dbal/pull/4291.


src/Cache/CachingResult.php
```diff
@@ -5,7 +5,8 @@
 use Doctrine\Common\Cache\Cache;
 use Doctrine\DBAL\Driver\Exception;
 use Doctrine\DBAL\Driver\FetchUtils;
-use Doctrine\DBAL\Driver\Result;
+use Doctrine\DBAL\Driver\Result as DriverResult;
+use Doctrine\DBAL\Result;
```
SenseException, line 9: Is having a `Doctrine\DBAL\Result` in `$result` on purpose instead of a `Driver\Result` and planned for 3.0?

SenseException (dismissed):
If the answer to my previous question is "yes", this then this is ready for merge IMO.

src/Cache/CachingResult.php
morozov: Yes. The caching layer only works with the wrapper-level result. Prior to this change, the wrapper result would implement the driver result interface, so this signature would work. As of this change, it's no longer true.

Additional references:
1. The same is true for the wrapper Connection and Statement classes as of https://github.com/doctrine/dbal/pull/4159.
2. We may rework the caching layer into a driver-level middleware where it belongs but it's not a top priority for me. See https://github.com/doctrine/dbal/pull/4157, there's basically a similar issue here.

SenseException (approved):

greg0ire (approved):

PR #4296, opened 2020-09-25
Increase indent in definition lists

greg0ire:
Although the official docs [1] show examples of definition list with an
indent of 2 spaces, it seems that doctrine/rst-parser has been written
with an indent of 4 spaces in mind. [2]

This should ensure the definition lists are properly parsed as such,
right now they are parsed as paragraphs.

[1] https://docutils.sourceforge.io/docs/user/rst/quickref.html#definition-lists
[2] https://github.com/doctrine/rst-parser/blob/d97da75a8a2cab33453711f01b90303f5c3845db/lib/Parser/LineDataParser.php#L154

Please tell me if you think I should report this as a bug to `doctrine/rst-parser` instead, or if you think we are just dealing with what could be called Doctrine-flavored RST here.

### DIFF
--- a/docs/en/explanation/implicit-indexes.rst
+++ b/docs/en/explanation/implicit-indexes.rst
@@ -6,13 +6,13 @@ with names such as ``IDX_885DBAFAA76ED395``? In this document, we will
 distinguish three types of indexes:
 
 user-defined indexes
-  indexes you did ask for
+    indexes you did ask for
 
 DBAL-defined indexes
-  indexes you did not ask for, created on your behalf by the DBAL
+    indexes you did not ask for, created on your behalf by the DBAL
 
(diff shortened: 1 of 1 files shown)


morozov (approved):
The existing markup looks valid according to the reference. There are other structures that use 4 spaces instead of two, so I have no idea of the potential side effects of using 4 spaces where the standard expects two are.

I'll approve if you want to proceed.

SenseException (approved):
> I have no idea of the potential side effects of using 4 spaces where the standard expects two are.

Neither do I. There's a project with dozens of pages in rst and the indent of 4 had no negative impact.

greg0ire:
Ok let's proceed then.

PR #4298, opened 2020-09-26
Ignore index prefix lengths for spatial indexes

ksaveras:
|      Q       |   A
|------------- | -----------
| BC Break     | no
| Fixed issues | #3561 

#### Summary

MySQL spatial indexes do not support index lengths and this check should be skipped.
If there are MSSQL or PGSQL experts - please comment.

This PR skips index lengths check if both indexes have a spatial flag.

I.e. when an index is created, MySQL 5.7 creates with 32 bytes prefix length. AWS Aurora (MySQL 5.7 compatible) - 24 bytes prefix length.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Index.php
+++ b/lib/Doctrine/DBAL/Schema/Index.php
@@ -223,7 +223,7 @@ class Index extends AbstractAsset implements Constraint
                 return false;
             }
 
-            if (! $this->hasSameColumnLengths($other)) {
+            if (! $this->sameFlag('spatial', $other) && ! $this->hasSameColumnLengths($other)) {
                 return false;
             }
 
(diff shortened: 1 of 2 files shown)


greg0ire:
Why did you create another PR?

ksaveras:
> Why did you create another PR?

https://github.com/doctrine/dbal/pull/3970#issuecomment-699475726 
Don't want to make more mess with merge conflicts.

greg0ire:
So you make more mess with more PRs, and we lose all the history of the comments?

greg0ire:
Any mess can be undone by force-pushing something correct, which is what I just did.

ksaveras:
Awesome. Thanks for the help.

PR #4309, opened 2020-09-28
Use cache action v2

greg0ire:


greg0ire:
@morozov I noticed that the default branch is now 3.0.x, is that intended? Or was it a consequence of using automatic-releases?

morozov:
It's not intended. It must have been changed by the release workflow. I reverted manually after the previous release but didn't yet notice the need this time.

morozov (approved):

greg0ire:
> haven't yet noticed the need this time.

Are you suggesting we shouldn't change it back this time? According to https://github.com/doctrine/dbal/pull/4309 it should be if you feel otherwise please elaborate.

morozov:
We should.

greg0ire:
Okay, done.

SenseException (approved):

PR #4310, opened 2020-09-28
Migrate jobs away from Travis to Github Actions 

greg0ire:
Todo:

- [x] mssql
- [x] sqlite
- [x] ibm_db2 (this one looks really nasty)
- [x] setup dependency on sqlite job
- [x] decide what to do about PostgreSQL 9.2 (I can't get it to work)
- [x] add some 7.3 jobs for MariaDB / PG ?
- [x] remove scripts from `tests/travis`?
- [x] find replacement for cron that runs with dev dependencies
- [x] make relevant jobs required
- [x] make sure code coverage hasn't decreased
- [x] test only the latest version of a platform when not using the latest version of PHP


greg0ire:
@localheinz @bendavies please also review :pray:

.github/workflows/continuous-integration.yml
```diff
@@ -156,3 +206,334 @@ jobs:
...
+      matrix:
+        php-version:
+          - "7.4"
+        postgres-version:
+          - "9.3"
+          - "9.4"
+          - "9.5"
+          - "9.6"
+          - "10"
+          - "11.1"
```
bendavies, line 225: 11.1 is very old. just specify `11` for the latest?

bendavies: add 12?

bendavies: add 13?

bendavies:
very nice

.github/workflows/continuous-integration.yml
greg0ire: That was introduced in 0a165b465ddbc4312be469e8a6b762877a61c87a , I don't know the reason for that but maybe @morozov recalls?

tests/travis/install-sqlsrv-dependencies.sh
```diff
@@ -1,10 +0,0 @@
-#!/usr/bin/env bash
-
-set -ex
-
-echo Installing driver dependencies
-
-curl https://packages.microsoft.com/keys/microsoft.asc | sudo apt-key add -
-curl https://packages.microsoft.com/config/ubuntu/16.04/prod.list | sudo tee /etc/apt/sources.list.d/mssql.list
-sudo apt-get update
-ACCEPT_EULA=Y sudo apt-get install -qy msodbcsql17 unixodbc unixodbc-dev libssl1.0.0
```
greg0ire, line 10: I wrote the equivalent of this at first, only to see that all the packages above were already present somehow.

.github/workflows/continuous-integration.yml
greg0ire: I also think we could drop 9.2 (I can't get it to work but IIRC it's not supported anymore). @morozov , what should the final list look like?

morozov: There is code in DBAL for PostgreSQL 9.2 which is technically still supported by DBAL 2.x and is continuously tested on Travis. If for some reason we can't test it on GitHub Actions, can we leave this build on Travis? We can get rid of it as soon as we stop supporting 2.x or manage to migrate to GitHub Actions.

morozov:
This is impressive work, @greg0ire!

.github/workflows/continuous-integration.yml
morozov: Why don't you add the failing PostgreSQL 9.2 build to the PR to possibly have another pair of eyes on it?

greg0ire: > Why don't you add the failing PostgreSQL 9.2 build to the PR to possibly have another pair of eyes on it?

Sure, done, please take a look 👍

If we can't find a solution, then for sure we can leave the build on Travis.

greg0ire:
The issue with the PG 9.2 is that we don't have the logs of the service… Let me try to remove the healtcheck, that way we can proceed with any next step we want. I will add a step that waits 1 mn, and then another that shows the logs.

greg0ire:
Well… I didn't expect this! Looks like the issue is with the healthcheck itself!

greg0ire:
Works like a charm w/o the delay too… maybe we should merge as is (after a squash and some cleanup of the commented out code of course). If we get issues down the line, we can still add the healtcheck back, and mark the 9.2 build as not required until we figure it out.

bendavies:
@greg0ire  `pg_isready ` was added in postgresql 9.3

greg0ire:
Yup that's what I figured in https://github.com/doctrine/dbal/pull/4310#issuecomment-700913661 :smile:

bendavies:
Ah sorry!
This should work:
```
--health-cmd "pg_isready || psql -U postgres -c 'SELECT 1'"
```

greg0ire:
Nice! Let's try that!

greg0ire:
Works great too apparently :)

I'll apply your suggestions regarding the list of versions, we can still change that later if 11.1 is really required for some reason but I doubt it.

greg0ire:
@morozov can you recommend something regarding " add some 7.3 jobs for MariaDB / PG ?"

morozov:
> @morozov can you recommend something regarding " add some 7.3 jobs for MariaDB / PG ?"

What's the question? If it's about which exactly configurations to test, https://github.com/doctrine/dbal/pull/3347#issuecomment-439231396 may help.

greg0ire:
It does help, thanks! I'll be AFK for a week but feel free to take over if you want to.

morozov:
I think I can wait another week. Have a good time off!

.github/workflows/continuous-integration.yml
```diff
@@ -156,3 +206,337 @@ jobs:
...
+      postgres:
+        image: postgres:${{ matrix.postgres-version }}
+        env:
+          POSTGRES_PASSWORD: postgres
+
+        options: >-
+          --health-cmd "pg_isready || psql -U postgres -c 'SELECT 1'"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
```
morozov, line 240: Where do these numbers come from? I don't know what the defaults are but somehow we got away with not specifying anything like that for the Oracle builds, and they work. At the same time, some similar options seem to cause issues in https://github.com/doctrine/doctrine-website/pull/367.

.github/workflows/continuous-integration.yml
```diff
@@ -156,3 +206,337 @@ jobs:
...
+          - collation: Latin1_General_100_CS_AS
+            php-version: "7.4"
+            driver: "pdo_sqlsrv"
+
+    services:
+      mssql:
+        image: microsoft/mssql-server-linux:2017-latest
+        env:
+          ACCEPT_EULA: 'Y'
+          SA_PASSWORD: 'Doctrine2018'
```
morozov, line 441: Any reason to use single quotes here? I believe we should use quotes only where necessary, e.g. in the strings like `"7.4"` or `"3306:3306"`.

.github/workflows/continuous-integration.yml
greg0ire: They come from https://docs.github.com/en/free-pro-team@latest/actions/guides/creating-postgresql-service-containers, I will try removing them.

.github/workflows/dev-dependencies.yml
```diff
@@ -0,0 +1,41 @@
+
+name: "Scheduled checks"
```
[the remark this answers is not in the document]
greg0ire: I was not very inspired… what would you suggest here?

ci/github/install-db2-ibm_db2.sh
```diff
@@ -21,7 +21,6 @@ echo "Installing extension"
     cd ibm_db2-*
     phpize
     ./configure --with-IBM_DB2=/tmp/dsdriver
-    make -j `nproc`
-    make install
-    echo -e 'extension=ibm_db2.so\nibm_db2.instance_name=db2inst1' > ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/ibm_db2.ini
+    make -j $(nproc)
```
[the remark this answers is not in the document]
greg0ire: I don't think it should, since the `nproc` part is supposed to make this portable.

.github/workflows/continuous-integration.yml
```diff
@@ -69,9 +69,58 @@ jobs:
...
+      - name: "Run PHPUnit"
+        run: "vendor/bin/phpunit -c ci/github/phpunit.sqlite.xml --coverage-clover=coverage.xml"
+
+      - name: "Upload Code Coverage"
+        uses: "codecov/codecov-action@v1"
+
   phpunit-oci8:
     name: "PHPUnit on OCI8"
     runs-on: "ubuntu-latest"
+    needs: phpunit-sqlite
```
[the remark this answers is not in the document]
greg0ire: That would make `phpunit-sqlite` inconsistently named with other jobs. I will try to create a job named `smoke-check` and make it depend on the `phpunit-sqlite` job. That way, it will no longer look weird, and we can add more checks as part of the smoke check.

.github/workflows/continuous-integration.yml
```diff
@@ -156,3 +206,337 @@ jobs:
 
       - name: "Upload Code Coverage"
         uses: "codecov/codecov-action@v1"
+
+  phpunit-postgres:
+    name: "PHPUnit with PostgreSQL"
+    runs-on: "ubuntu-latest"
```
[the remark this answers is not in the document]
greg0ire: That was not deliberate, I will use a fixed version.

.github/workflows/continuous-integration.yml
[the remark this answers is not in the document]
greg0ire: There does not seem to be such a thing as empty jobs, so I'll just use your suggestion.

.github/workflows/continuous-integration.yml
```diff
@@ -156,3 +206,337 @@ jobs:
...
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-interaction --no-progress --no-suggest"
+
+      - name: "Run PHPUnit"
+        run: "vendor/bin/phpunit -c ci/github/phpunit.mysql-mariadb.${{ matrix.driver }}.xml --coverage-clover=coverage.xml"
```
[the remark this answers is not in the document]
greg0ire: There is a difference in one of the lines: `<var name="db_driver" value="pdo_mysql"/>` vs `<var name="db_driver" value="mysqli"/>`. Note that this file is not just for MariaDB, it is also used with MySQL

.github/workflows/continuous-integration.yml
```diff
@@ -156,3 +206,337 @@ jobs:
...
+          - "5.7"
+          - "8.0"
+        driver:
+          - "mysqli"
+          - "pdo_mysql"
+        config-file:
+          - "mysql-mariadb"
+        include:
+          - mysql-version: '5.7'
+            custom-entrypoint: ''
```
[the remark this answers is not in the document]
greg0ire: Good point, I'll try removing it

.github/workflows/continuous-integration.yml
```diff
@@ -156,3 +206,337 @@ jobs:
...
+          - mysql-version: '8.0'
+            # https://stackoverflow.com/questions/60902904/how-to-pass-mysql-native-password-to-mysql-service-in-github-actions
+            custom-entrypoint: >-
+              --entrypoint sh mysql:8 -c "exec docker-entrypoint.sh mysqld --default-authentication-plugin=mysql_native_password"
```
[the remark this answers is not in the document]
greg0ire: It does look weird, let's try without.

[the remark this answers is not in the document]
greg0ire: Looks like it is necessary. It now fails on this line:
```
/usr/bin/docker create --name fe434eb3cdfc480593fc56b0f182aac4_mysql80_d128bc --label 1e5c35 --network github_network_c8c7dc3b30ba49acaa4fd92828200ea4 --network-alias mysql -p 3306:3306 --health-cmd "mysqladmin ping --silent" -e MYSQL_ALLOW_EMPTY_PASSWORD=yes -e MYSQL_DATABASE=doctrine_tests --entrypoint sh -c "exec docker-entrypoint.sh mysqld --default-authentication-plugin=mysql_native_password" -e GITHUB_ACTIONS=true -e CI=true mysql:8.0
```

Previously , it worked fine with

```
 /usr/bin/docker create --name a888c831f23f44debb23bebeef3c0d90_mysql80_6f6e5e --label 1e5c35 --network github_network_c76a6ca279684713bd8355ad3f5a74bb --network-alias mysql -p 33306:3306 --health-cmd "mysqladmin ping --silent" -e MYSQL_ALLOW_EMPTY_PASSWORD=yes -e MYSQL_DATABASE=doctrine_tests --entrypoint sh mysql:8 -c "exec docker-entrypoint.sh mysqld --default-authentication-plugin=mysql_native_password" -e GITHUB_ACTIONS=true -e CI=true mysql:8.0
```

The error message is `invalid argument "exec docker-entrypoint.sh mysqld --default-authentication-plugin=mysql_native_password" for "-c, --cpu-shares" flag: strconv.ParseInt: parsing "exec docker-entrypoint.sh mysqld --default-authentication-plugin=mysql_native_password": invalid syntax`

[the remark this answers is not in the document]
greg0ire: The usage is `docker create [OPTIONS] IMAGE [COMMAND] [ARG...]`
I think the entrypoint is simply sh, and `-c "exec docker-entrypoint.sh mysqld --default-authentication-plugin=mysql_native_password"` ends up in `COMMAND`? @mvorisek can you maybe help? :pray:

.github/workflows/dev-dependencies.yml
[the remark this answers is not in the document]
morozov: I'd rename the file to `scheduled-checks.yml`.

ci/github/install-db2-ibm_db2.sh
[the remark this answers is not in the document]
morozov: I mean the change from `` `command` `` to `$(command)`. Both are valid Bash syntaxes.

.github/workflows/continuous-integration.yml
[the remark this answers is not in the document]
morozov: These are different drivers, each of which can be used both with MariaDB or MySQL. Since the configuration in the file doesn't mention either of the platforms specifically, its name shouldn't either. It should only mention the driver (which it already does).

.github/workflows/continuous-integration.yml
morozov: If you want to use quotes consistently, let's then use double quotes by default.

.github/workflows/continuous-integration.yml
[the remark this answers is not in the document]
morozov: FWIW, I'm fine with it if it's needed.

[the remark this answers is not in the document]
mvorisek: > @mvorisek can you maybe help? 🙏

please see my SO post https://stackoverflow.com/questions/60902904/how-to-pass-mysql-native-password-to-mysql-service-in-github-actions , I use this GitHub Actions hack successfully for https://github.com/atk4/data and several other repos. To answer your question - see https://github.com/doctrine/dbal/pull/4310/checks?check_run_id=1235719344#step:2:77 , ie. how GitHub actually builds the Docker command, which is documented in Docker doc

.github/workflows/dev-dependencies.yml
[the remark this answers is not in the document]
mvorisek: I think it is not needed at all, see https://github.com/atk4/data/blob/develop/.github/workflows/unit-tests.yml#L9

the main CI workflow can be simply run on multiple events incl. schedule/cron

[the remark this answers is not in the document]
greg0ire: I'm trying to reproduce the current Travis setup, and I don't think it runs the whole workflow.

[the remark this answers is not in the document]
morozov: If it simplifies the setup, then why not?

.github/workflows/continuous-integration.yml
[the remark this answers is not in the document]
greg0ire: I saw your SO post, I even linked to it in the configuration file ;)

Anyway, it's needed so I'll revert to the previous configuration.

.github/workflows/dev-dependencies.yml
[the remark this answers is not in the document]
greg0ire: Okay, let's try that then.

greg0ire:
@morozov the IBM DB2 job fails :facepalm: 

It says 

> PHP Warning:  PHP Startup: Unable to load dynamic library 'ibm_db2.so' (tried: /usr/lib/php/20180731/ibm_db2.so (/usr/lib/php/20180731/ibm_db2.so: cannot open shared object file: No such file or directory), /usr/lib/php/20180731/ibm_db2.so.so (/usr/lib/php/20180731/ibm_db2.so.so: cannot open shared object file: No such file or directory)) in Unknown on line 0

The install steps shows another directory: `Installing shared extensions:     /usr/lib/php/20190902/`

Any idea what's going on?

.github/workflows/continuous-integration.yml
```diff
@@ -1,12 +1,18 @@
 
 name: "Continuous Integration"
```
mvorisek, line 2:
```suggestion
name: "CI"
```

image

long names as not displayed completely... Maybe also rename the workflow file to `ci.yml` then

greg0ire: > long names as not displayed completely...

They are:

Screenshot_2020-10-10 Migrate jobs away from Travis to Github Actions by greg0ire · Pull Request #4310 · doctrine dbal

mvorisek: image

not everywhere...

greg0ire: Indeed. Hopefully Github will fix that in the future. In the meantime, I'd rather not use abbreviations, which would mean being inconsistent with workflows that come from `doctrine/.github` and don't use abbreviations either.

morozov:
> Any idea what's going on?

Looks like the extension is not installed. We install it from source and use the client library hosted by a 3rd party. See the Travis implementation.

greg0ire:
> Looks like the extension is not installed. We install it from source and use the client library hosted by a 3rd party. See the Travis implementation.

My understanding is that it is installed, but the directory where it is installed (`/usr/lib/php/20190902/`) isn't looked at when using `ubuntu-latest`. Instead, PHP looks into `/usr/lib/php/20180731`… that's the part I don't understand.

.github/workflows/continuous-integration.yml
```diff
@@ -1,12 +1,18 @@
...
 
-on: ["pull_request", "push"]
+on:
+  pull_request:
+    branches: '**'
+  push:
+    branches: '**'
```
morozov, line 8: This is why when somebody opens a PR from the upstream repo, we get all jobs run twice. Why don't we restrict these to only the release branches?

.github/workflows/continuous-integration.yml
```diff
@@ -156,3 +212,360 @@ jobs:
...
+          - "13"
+
+    services:
+      postgres:
+        image: postgres:${{ matrix.postgres-version }}
+        env:
+          POSTGRES_PASSWORD: postgres
+
+        options: >-
+          --health-cmd "pg_isready || psql -U postgres -c 'SELECT 1'"
```
morozov, line 243: Note to self: the fallback can be removed when merging up to `3.0.x`.

.github/workflows/continuous-integration.yml
```diff
@@ -156,3 +212,360 @@ jobs:
...
+        if: ${{ endsWith(matrix.config-file, 'tls') }}
+
+      - name: "Run PHPUnit"
+        run: "vendor/bin/phpunit -c ci/github/${{ matrix.config-file-prefix }}.${{ matrix.driver }}.xml --coverage-clover=coverage.xml"
+
+      - name: "Upload Code Coverage"
+        uses: "codecov/codecov-action@v1"
+
+  phpunit-mssql:
+    name: "PHPUnit with MSSQL"
```
morozov, line 412: I believe MSSQL is a legacy name while SQL Server is what's used now.

.github/workflows/continuous-integration.yml
```diff
@@ -69,9 +75,58 @@ jobs:
...
+    name: "PHPUnit with SQLite"
+    runs-on: "ubuntu-20.04"
+
+    strategy:
+      matrix:
+        php-version:
+          - "7.3"
+          - "7.4"
+        deps:
+          - fixed
```
morozov, line 88: My OCD is still unhappy. What's your take on quoting stings?

.github/workflows/continuous-integration.yml
```diff
@@ -156,3 +212,360 @@ jobs:
...
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          php-version: "${{ matrix.php-version }}"
+          coverage: "pcov"
+          ini-values: extension=ibm_db2.so, ibm_db2.instance_name=db2inst1
+
+      - name: "Install ibm_db2 extension"
+        run: "bash ./ci/github/install-db2-ibm_db2.sh ${{ matrix.php-version }}"
```
morozov, line 521: Would it make sense to make the script executable and just runt it as is? The `./` prefix looks redundant since we don't have it available in multiple paths.

.github/workflows/continuous-integration.yml
greg0ire: IIRC `no` can be wrongly converting to `false`. Let's not take any chances and quote everything.

.github/workflows/continuous-integration.yml
mvorisek: impossible - I already spent a lot of time figuring this out

.github/workflows/continuous-integration.yml
morozov: FWIW, this is what the version that quotes only the necessary pieces looks like:

<details>
  <summary>Click to expand!</summary>
  
```yaml
name: Continuous Integration
on:
  pull_request:
    branches: '**'
  push:
    branches: '**'
  schedule:
  - cron: 42 3 * * *
jobs:
  development-deps:
    name: PHPUnit with SQLite and development dependencies
    runs-on: ubuntu-20.04
    steps:
    - name: Checkout
      uses: actions/checkout@v2
    - name: Install PHP
      uses: shivammathur/setup-php@v2
      with:
        php-version: ${{ matrix.php-version }}
    - name: Cache dependencies installed with composer
      uses: actions/cache@v2
      with:
        key: php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock')
          }}
        path: ~/.composer/cache
        restore-keys: php-${{ matrix.php-version }}-composer-locked-
    - name: Lower minimum stability
      run: composer config minimum-stability dev
    - name: Install development dependencies with composer
      run: composer update --no-interaction --no-progress --no-suggest --prefer-dist
    - name: Run PHPUnit
      run: vendor/bin/phpunit -c ci/github/phpunit.sqlite.xml
    strategy:
      matrix:
        php-version:
        - "7.4"
  phpunit-ibm-db2:
    name: PHPUnit with IBM DB2
    needs: phpunit-smoke-check
    runs-on: ubuntu-18.04
    services:
      ibm_db2:
        env:
          DB2INST1_PASSWORD: Doctrine2018
          DBNAME: doctrine
          LICENSE: accept
        image: ibmcom/db2:11.5.0.0
        options: --privileged=true
        ports:
        - 50000:50000
    steps:
    - name: Perform healthcheck from the outside
      run: docker logs -f ${{ job.services.ibm_db2.id }} | sed '/(*) Setup has completed./
        q'
    - name: Create temporary tablespace
      run: docker exec ${{ job.services.ibm_db2.id }} su - db2inst1 -c 'db2 CONNECT
        TO doctrine && db2 CREATE USER TEMPORARY TABLESPACE doctrine_tbsp PAGESIZE
        4 K'
    - name: Checkout
      uses: actions/checkout@v2
      with:
        fetch-depth: 2
    - name: Install PHP
      uses: shivammathur/setup-php@v2
      with:
        coverage: pcov
        ini-values: extension=ibm_db2.so, ibm_db2.instance_name=db2inst1
        php-version: ${{ matrix.php-version }}
    - name: Install ibm_db2 extension
      run: bash ./ci/github/install-db2-ibm_db2.sh ${{ matrix.php-version }}
    - name: Cache dependencies installed with composer
      uses: actions/cache@v2
      with:
        key: php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock')
          }}
        path: ~/.composer/cache
        restore-keys: php-${{ matrix.php-version }}-composer-locked-
    - name: Install dependencies with composer
      run: composer install --no-interaction --no-progress --no-suggest
    - name: Run PHPUnit
      run: vendor/bin/phpunit -c ci/github/phpunit.ibm_db2.xml --coverage-clover=coverage.xml
    - name: Upload Code Coverage
      uses: codecov/codecov-action@v1
    strategy:
      matrix:
        php-version:
        - "7.3"
  phpunit-mariadb:
    name: PHPUnit with MariaDB
    needs: phpunit-smoke-check
    runs-on: ubuntu-20.04
    services:
      mariadb:
        env:
          MYSQL_ALLOW_EMPTY_PASSWORD: true
          MYSQL_DATABASE: doctrine_tests
        image: mariadb:${{ matrix.mariadb-version }}
        options: --health-cmd "mysqladmin ping --silent"
        ports:
        - 3306:3306
    steps:
    - name: Checkout
      uses: actions/checkout@v2
      with:
        fetch-depth: 2
    - name: Install PHP
      uses: shivammathur/setup-php@v2
      with:
        coverage: pcov
        extensions: ${{ matrix.driver }}
        php-version: ${{ matrix.php-version }}
    - name: Cache dependencies installed with composer
      uses: actions/cache@v2
      with:
        key: php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock')
          }}
        path: ~/.composer/cache
        restore-keys: php-${{ matrix.php-version }}-composer-locked-
    - name: Install dependencies with composer
      run: composer install --no-interaction --no-progress --no-suggest
    - name: Run PHPUnit
      run: vendor/bin/phpunit -c ci/github/phpunit.${{ matrix.driver }}.xml --coverage-clover=coverage.xml
    - name: Upload Code Coverage
      uses: codecov/codecov-action@v1
    strategy:
      matrix:
        driver:
        - mysqli
        - pdo_mysql
        mariadb-version:
        - "10.0"
        - "10.1"
        - "10.2"
        - "10.3"
        php-version:
        - "7.4"
  phpunit-mssql:
    name: PHPUnit with MSSQL
    needs: phpunit-smoke-check
    runs-on: ubuntu-20.04
    services:
      mssql:
        env:
          ACCEPT_EULA: "Y"
          MSSQL_COLLATION: ${{ matrix.collation }}
          SA_PASSWORD: Doctrine2018
        image: microsoft/mssql-server-linux:2017-latest
        options: --health-cmd "echo quit | /opt/mssql-tools/bin/sqlcmd -S 127.0.0.1
          -l 1 -U sa -P Doctrine2018"
        ports:
        - 1433:1433
    steps:
    - name: Checkout
      uses: actions/checkout@v2
      with:
        fetch-depth: 2
    - name: Install PHP
      uses: shivammathur/setup-php@v2
      with:
        coverage: pcov
        extensions: ${{ matrix.driver }}-5.7.0preview
        php-version: ${{ matrix.php-version }}
        tools: pecl
    - name: Cache dependencies installed with composer
      uses: actions/cache@v2
      with:
        key: php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock')
          }}
        path: ~/.composer/cache
        restore-keys: php-${{ matrix.php-version }}-composer-locked-
    - name: Install dependencies with composer
      run: composer install --no-interaction --no-progress --no-suggest
    - name: Run PHPUnit
      run: vendor/bin/phpunit -c ci/github/phpunit.${{ matrix.driver }}.xml --coverage-clover=coverage.xml
    - name: Upload Code Coverage
      uses: codecov/codecov-action@v1
    strategy:
      matrix:
        collation:
        - Latin1_General_100_CI_AS
        driver:
        - sqlsrv
        - pdo_sqlsrv
        include:
        - collation: Latin1_General_100_CS_AS
          driver: sqlsrv
          php-version: "7.4"
        - collation: Latin1_General_100_CS_AS
          driver: pdo_sqlsrv
          php-version: "7.4"
        php-version:
        - "7.3"
        - "7.4"
  phpunit-mysql:
    name: PHPUnit with MySQL
    needs: phpunit-smoke-check
    runs-on: ubuntu-20.04
    services:
      mysql:
        image: mysql:${{ matrix.mysql-version }}
        options: --health-cmd "mysqladmin ping --silent" -e MYSQL_ALLOW_EMPTY_PASSWORD=yes
          -e MYSQL_DATABASE=doctrine_tests ${{ matrix.custom-entrypoint }}
        ports:
        - 3306:3306
    steps:
    - name: Checkout
      uses: actions/checkout@v2
      with:
        fetch-depth: 2
    - name: Install PHP
      uses: shivammathur/setup-php@v2
      with:
        coverage: pcov
        extensions: ${{ matrix.driver }}
        php-version: ${{ matrix.php-version }}
    - name: Cache dependencies installed with composer
      uses: actions/cache@v2
      with:
        key: php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock')
          }}
        path: ~/.composer/cache
        restore-keys: php-${{ matrix.php-version }}-composer-locked-
    - name: Install dependencies with composer
      run: composer install --no-interaction --no-progress --no-suggest
    - if: ${{ endsWith(matrix.config-file, 'tls') }}
      name: Copy TLS-related files
      run: docker cp "${{ job.services.mysql.id }}:/var/lib/mysql/ca.pem" . && docker
        cp "${{ job.services.mysql.id }}:/var/lib/mysql/client-cert.pem" . && docker
        cp "${{ job.services.mysql.id }}:/var/lib/mysql/client-key.pem" .
    - name: Run PHPUnit
      run: vendor/bin/phpunit -c ci/github/${{ matrix.config-file-prefix }}.${{ matrix.driver
        }}.xml --coverage-clover=coverage.xml
    - name: Upload Code Coverage
      uses: codecov/codecov-action@v1
    strategy:
      matrix:
        config-file-prefix:
        - phpunit
        driver:
        - mysqli
        - pdo_mysql
        include:
        - mysql-version: "5.7"
        - custom-entrypoint: --entrypoint sh mysql:8 -c "exec docker-entrypoint.sh
            mysqld --default-authentication-plugin=mysql_native_password"
          mysql-version: "8.0"
        - config-file-prefix: phpunit.mysql-tls
          driver: mysqli
          mysql-version: "8.0"
          php-version: "7.4"
        mysql-version:
        - "5.7"
        - "8.0"
        php-version:
        - "7.3"
        - "7.4"
  phpunit-oci8:
    name: PHPUnit on OCI8
    needs: phpunit-smoke-check
    runs-on: ubuntu-20.04
    services:
      oracle:
        image: wnameless/oracle-xe-11g-r2
        ports:
        - 1521:1521
    steps:
    - name: Checkout
      uses: actions/checkout@v2
      with:
        fetch-depth: 2
    - name: Install PHP
      uses: shivammathur/setup-php@v2
      with:
        coverage: pcov
        extensions: oci8
        php-version: ${{ matrix.php-version }}
    - name: Cache dependencies installed with composer
      uses: actions/cache@v2
      with:
        key: php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock')
          }}
        path: ~/.composer/cache
        restore-keys: php-${{ matrix.php-version }}-composer-locked-
    - name: Install dependencies with composer
      run: composer install --no-interaction --no-progress --no-suggest
    - name: Run PHPUnit
      run: vendor/bin/phpunit -c ci/github/phpunit.oci8.xml --coverage-clover=coverage.xml
    - name: Upload Code Coverage
      uses: codecov/codecov-action@v1
    strategy:
      matrix:
        php-version:
        - "7.4"
  phpunit-pdo-oci:
    name: PHPUnit on PDO_OCI
    needs: phpunit-smoke-check
    runs-on: ubuntu-20.04
    services:
      oracle:
        image: wnameless/oracle-xe-11g-r2
        ports:
        - 1521:1521
    steps:
    - name: Checkout
      uses: actions/checkout@v2
      with:
        fetch-depth: 2
    - name: Install PHP
      uses: shivammathur/setup-php@v2
      with:
        coverage: pcov
        extensions: pdo_oci
        php-version: ${{ matrix.php-version }}
    - name: Cache dependencies installed with composer
      uses: actions/cache@v1
      with:
        key: php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock')
          }}
        path: ~/.composer/cache
        restore-keys: php-${{ matrix.php-version }}-composer-locked-
    - name: Install dependencies with composer
      run: composer install --no-interaction --no-progress --no-suggest
    - name: Run PHPUnit
      run: vendor/bin/phpunit -c ci/github/phpunit.pdo-oci.xml --coverage-clover=coverage.xml
    - name: Upload Code Coverage
      uses: codecov/codecov-action@v1
    strategy:
      matrix:
        php-version:
        - "7.4"
  phpunit-postgres:
    name: PHPUnit with PostgreSQL
    needs: phpunit-smoke-check
    runs-on: ubuntu-20.04
    services:
      postgres:
        env:
          POSTGRES_PASSWORD: postgres
        image: postgres:${{ matrix.postgres-version }}
        options: --health-cmd "pg_isready || psql -U postgres -c 'SELECT 1'"
        ports:
        - 5432:5432
    steps:
    - name: Checkout
      uses: actions/checkout@v2
      with:
        fetch-depth: 2
    - name: Install PHP
      uses: shivammathur/setup-php@v2
      with:
        coverage: pcov
        php-version: ${{ matrix.php-version }}
    - name: Cache dependencies installed with composer
      uses: actions/cache@v2
      with:
        key: php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock')
          }}
        path: ~/.composer/cache
        restore-keys: php-${{ matrix.php-version }}-composer-locked-
    - name: Install dependencies with composer
      run: composer install --no-interaction --no-progress --no-suggest
    - name: Run PHPUnit
      run: vendor/bin/phpunit -c ci/github/phpunit.pgsql.xml --coverage-clover=coverage.xml
    - name: Upload Code Coverage
      uses: codecov/codecov-action@v1
    strategy:
      matrix:
        php-version:
        - "7.4"
        postgres-version:
        - "9.2"
        - "9.3"
        - "9.4"
        - "9.5"
        - "9.6"
        - "10"
        - "11"
        - "12"
        - "13"
  phpunit-smoke-check:
    name: PHPUnit with SQLite
    runs-on: ubuntu-20.04
    steps:
    - name: Checkout
      uses: actions/checkout@v2
      with:
        fetch-depth: 2
    - name: Install PHP
      uses: shivammathur/setup-php@v2
      with:
        coverage: pcov
        php-version: ${{ matrix.php-version }}
    - name: Cache dependencies installed with composer
      uses: actions/cache@v2
      with:
        key: php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock')
          }}
        path: ~/.composer/cache
        restore-keys: php-${{ matrix.php-version }}-composer-locked-
    - if: ${{ matrix.deps == 'fixed' }}
      name: Install dependencies with composer
      run: composer install --no-interaction --no-progress --no-suggest
    - if: ${{ matrix.deps == 'low' }}
      name: Install lowest possible dependencies with composer
      run: composer update --no-interaction --no-progress --no-suggest --prefer-dist
        --prefer-lowest
    - name: Run PHPUnit
      run: vendor/bin/phpunit -c ci/github/phpunit.sqlite.xml --coverage-clover=coverage.xml
    - name: Upload Code Coverage
      uses: codecov/codecov-action@v1
    strategy:
      matrix:
        deps:
        - fixed
        include:
        - deps: low
          php-version: "7.3"
        php-version:
        - "7.3"
        - "7.4"
  static-analysis-phpstan:
    name: Static Analysis with PHPStan
    runs-on: ubuntu-20.04
    steps:
    - name: Checkout code
      uses: actions/checkout@v2
    - name: Install PHP
      uses: shivammathur/setup-php@v2
      with:
        coverage: none
        php-version: ${{ matrix.php-version }}
        tools: cs2pr
    - name: Cache dependencies installed with composer
      uses: actions/cache@v1
      with:
        key: php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock')
          }}
        path: ~/.composer/cache
        restore-keys: php-${{ matrix.php-version }}-composer-locked-
    - name: Install dependencies with composer
      run: composer install --no-interaction --no-progress --no-suggest
    - name: Run a static analysis with phpstan/phpstan
      run: vendor/bin/phpstan analyse --error-format=checkstyle | cs2pr
    strategy:
      matrix:
        php-version:
        - "7.4"
  static-analysis-psalm:
    name: Static Analysis with Psalm
    runs-on: ubuntu-20.04
    steps:
    - name: Checkout code
      uses: actions/checkout@v2
    - name: Install PHP
      uses: shivammathur/setup-php@v2
      with:
        coverage: none
        php-version: ${{ matrix.php-version }}
    - name: Cache dependencies installed with composer
      uses: actions/cache@v2
      with:
        key: php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock')
          }}
        path: ~/.composer/cache
        restore-keys: php-${{ matrix.php-version }}-composer-locked-
    - name: Install dependencies with composer
      run: composer install --no-interaction --no-progress --no-suggest
    - name: Run a static analysis with vimeo/psalm
      run: vendor/bin/psalm --show-info=false --stats --output-format=github --threads=4
    strategy:
      matrix:
        php-version:
        - "7.4"
```
</details>

.github/workflows/continuous-integration.yml
```diff
@@ -156,3 +216,360 @@ jobs:
...
+          - "pdo_mysql"
+        config-file-prefix:
+          - "phpunit"
+        include:
+          - mysql-version: "5.7"
+          - mysql-version: "8.0"
+            # https://stackoverflow.com/questions/60902904/how-to-pass-mysql-native-password-to-mysql-service-in-github-actions
+            custom-entrypoint: >-
+              --entrypoint sh mysql:8 -c "exec docker-entrypoint.sh mysqld --default-authentication-plugin=mysql_native_password"
+          - config-file-prefix: "phpunit.mysql-tls"
```
morozov, line 364: Why do we need to have the "phpunit" prefix configurable? Isn't it supposed to be used consistently everywhere? The resulting filename "phpunit.mysql-tls.mysqli.xml" looks redundant. Should it be a suffix instead? E.g. "phpunit.mysqli.tls.xml".

Another thing is that personally I'm not a fan of using a dot to separate parts of the file name because syntactically, it's a separator of the name and extension. Since we're renaming and restructuring files anyways, would it make sense to name them like  "phpunit/mysqli.xml", "phpunit/mysqli-tls.xml", etc. (i.e. put in a subdirectory instead of having a fixed prefix).

.github/workflows/continuous-integration.yml
greg0ire: I think we elected to always use strings in other workflows, so let's stick with that, shall we?

.github/workflows/continuous-integration.yml
greg0ire: What do you mean? I pushed something, it seems to work…

.github/workflows/continuous-integration.yml
morozov: Yes, let's quote everything using double-quotes.

.github/workflows/continuous-integration.yml
```diff
@@ -152,7 +212,364 @@ jobs:
...
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-interaction --no-progress --no-suggest"
+
+      - name: "Run PHPUnit"
+        run: "vendor/bin/phpunit -c ci/github/phpunit/pgsql.xml --coverage-clover=coverage.xml"
```
morozov, line 275: It should be "pdo-pgsql" since there is a non-PDO extension which the DBAL currently doesn't support.

.github/workflows/continuous-integration.yml
morozov: > the branch here is the target branch, right?

It should be. At least it is on Travis and AppVeyor where this does work.

.github/workflows/continuous-integration.yml
```diff
@@ -152,7 +212,364 @@ jobs:
...
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          php-version: "${{ matrix.php-version }}"
+          coverage: "pcov"
+          ini-values: "extension=ibm_db2.so, ibm_db2.instance_name=db2inst1"
+
+      - name: "Install ibm_db2 extension"
+        run: "ci/github/install-db2-ibm_db2.sh ${{ matrix.php-version }}"
```
morozov, line 525: This script is named this weirdly because the name was autogenerated on Travis from the build matrix parameters. Let's rename it to `ext/install-ibm-db2.sh`.

.github/workflows/continuous-integration.yml
```diff
@@ -152,7 +212,364 @@ jobs:
...
+        with:
+          path: "~/.composer/cache"
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer install --no-interaction --no-progress --no-suggest"
+
+      - name: "Run PHPUnit"
+        run: "vendor/bin/phpunit -c ci/github/phpunit/ibm_db2.xml --coverage-clover=coverage.xml"
```
morozov, line 538: Let's rename it to `ibm-db2.xml`.

.github/workflows/continuous-integration.yml
```diff
@@ -152,7 +212,364 @@ jobs:
...
+        with:
+          fetch-depth: 2
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          php-version: "${{ matrix.php-version }}"
+          coverage: "pcov"
+          tools: "pecl"
+          extensions: "${{ matrix.driver }}-5.7.0preview"
```
greg0ire, line 464: I have to find a way to use an underscore here.

greg0ire: Either that or I revert to `pdo_`

greg0ire: I cannot find a clean solution @morozov … can you? Please note that inside the phpunit files, the `db_driver` values uses underscore. I'll revert to underscores (same for `ibm-db2`)

morozov: Let's use underscore for PHP extensions and the corresponding filename fragments.

greg0ire:
> One more thing is that using dashes and underscores in the same naming schema looks untidy. 

in that schema `pdo_mysql` is an element, `tls` is another. But `pdo` and `sqlsrv` alone are not. I think using underscores when refering to a driver is better.

greg0ire:
Hang on, there is an issue with the tls job, it says `No tests executed!`

.github/workflows/continuous-integration.yml
greg0ire: Done and undone

.github/workflows/continuous-integration.yml
greg0ire: Done and undone

.github/workflows/continuous-integration.yml
greg0ire: Done

greg0ire:
Fixed the issue above

.github/workflows/continuous-integration.yml
morozov: It still should be `ext/install-ibm_db2.sh`.

.github/workflows/continuous-integration.yml
```diff
@@ -152,7 +212,364 @@ jobs:
...
+        uses: "actions/checkout@v2"
+        with:
+          fetch-depth: 2
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          php-version: "${{ matrix.php-version }}"
+          coverage: "pcov"
+          extensions: "${{ matrix.driver }}"
```
morozov, line 322: To make it clear that `pdo_sqlite` is namely the extension, should it be `matrix.extension`?

.github/workflows/continuous-integration.yml
mvorisek: > Why don't we restrict these to only the release branches?

pull request - must be for all branches, otherwise CI results will not be displayed in PR if from another repo
push - we must not limit, otherwise it CI will not run in user's repo

so as long as GH Actions are free, this should not be an issue if it is run twice (once for push, once for PR)

morozov: > so as long as GH Actions are free, this should not be an issue if it is run twice (once for push, once for PR)

It has been the exact issue in the past with Travis. Such builds would take twice longer for no reason.

.github/workflows/continuous-integration.yml
greg0ire: :thinking: do you want me to rename all occurrences of `driver` in the file to `extension`? If not, what is special about sqlite?

greg0ire: I renamed all occurrences.

morozov: Yes, all occurrences. As it turns out, all our filenames are based on extension names.

.github/workflows/continuous-integration.yml
```diff
@@ -152,7 +212,364 @@ jobs:
...
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          php-version: "${{ matrix.php-version }}"
+          coverage: "pcov"
+          ini-values: "extension=ibm_db2.so, ibm_db2.instance_name=db2inst1"
+
+      - name: "Install ibm_db2 extension"
+        run: "ci/github/ext/install-db2-ibm_db2.sh ${{ matrix.php-version }}"
```
morozov, line 525: It still should be `ext/install-ibm_db2.sh`. We're not installing the DB2 itself here, we're only installing the `ibm_db2` PHP extension.

morozov:
Looks good so far. Are you going to address the rest of the TODO items, @greg0ire? I added one more yesterday.

.github/workflows/continuous-integration.yml
greg0ire: Oh right didn't spot this sorry.

greg0ire:
The last one should be addressed by you before merging

greg0ire:
> add some 7.3 jobs for MariaDB / PG ?

@morozov can I get your input on this? Are there enough jobs to your liking?

.github/workflows/continuous-integration.yml
greg0ire: Addressed.

greg0ire:
I see there a coverage drop, for some reason `OracleSchemaManager` is no longer considered as covered, which is really weird because that's one of the jobs I did not touch.

Also, `assert` seems to be disabled, so lines with `assert` are not considered as covered.

greg0ire:
:thinking: downgrading to `ubuntu-latest` seems to fix the `pdo_sqlsrv` issue, but not the Oracle issue…

morozov:
> Also, `assert` seems to be disabled, so lines with `assert` are not considered as covered.

Interesting. In this case, they shouldn't be considered coverable either (something to run by PHPUnit).

morozov:
> > add some 7.3 jobs for MariaDB / PG ?
> 
> @morozov can I get your input on this? Are there enough jobs to your liking?

From the comment linked above:

> All other supported PHP versions ([…]), test each driver with the latest version of the platform.

So we'll need to test `pdo_pgsql`, `mysqli` and `pdo_mysql` with the latest versions of PostgreSQL and MariadDB respectively.

greg0ire:
Ok, I added a few more for MariaDB. Pg and MySQL latest versions are already tested.

greg0ire:
Flagging the jobs shows that oci reports are never uploaded…

greg0ire:
Ok, data point: when I remove all new jobs, I get coverage for the oracle builds.

greg0ire:
Re-adding just the SQlite is enough to break it.

greg0ire:
Doesn't work either, let's add a dependency from SQlite to Oracle, so that sqlite builds are run after.

greg0ire:
After making sure every job uses the cache action@v2 seems to fix the issue I get the Oracle builds but not all of the SQlite ones…

greg0ire:
I really don't get it… I'm going to file a bug report.

greg0ire:
@thomasrockhu can you please take a look? :pray:

thomasrockhu:
@greg0ire responding in the community thread

morozov:
@greg0ire since the problem with the code coverage is caused by the number of uploaded reports, should we try to reduce the number of builds? E.g. 10+ MariaDB builds on PHP 7.4 alone is too much.

greg0ire:
> @greg0ire since the problem with the code coverage is caused by the number of uploaded reports, should we try to reduce the number of builds? E.g. 10+ MariaDB builds on PHP 7.4 alone is too much.

I tried adding just the postgres reports the other day, it didn't work, so I think we will hit the limit pretty fast.

greg0ire:
Ah I think I'm onto something!

greg0ire:
Yaaaaay! +2% coverage!

greg0ire:
I added a todo item to reduce the number of builds according to your recommendation @morozov

greg0ire:
I think that will not change things much. Please tell me precisely what rule I should apply to decrease the number of builds. Should we maybe have some rule like "test the lowest supported platform version, and the 3 latest minor versions, but not all versions in between"?

morozov:
> I turned on assertions, it's still code that is better off being run than not run: what if an assertion is wrong?

Yeah. CI is the right place to have them enabled.

> Please tell me precisely what rule I should apply to decrease the number of builds. Should we maybe have some rule like "test the lowest supported platform version, and the 3 latest minor versions, but not all versions in between"?

I didn't realize we have that many different MariaDB versions tested, so the existing 12 jobs do not technically violate the approach defined in https://github.com/doctrine/dbal/pull/3347#issuecomment-439231396. I believe we can leave them as is for now and rely on support for some older versions being dropped in `3.0.x`.

> […] please note that MariaDB is not tested with 7.3, so maybe there are more todo items to add

I wouldn't bother right now unless those jobs provide additional code coverage or reveal new issues.

morozov (approved):

greg0ire:
Ok, I will now fiddle with the branch settings and finally merge this! :tada:

morozov:
> Should we maybe have some rule like "test the lowest supported platform version, and the 3 latest minor versions, but not all versions in between"?

Actually what we could do later as optimization is to reduce the set of specific platform versions being tested to the necessary minimum that our code is aware of. I.e. test the oldest version, the newest version, and all versions for which we have version-specific code. For MariaDB, it would be 10.0 (oldest), 10.2 (`MariaDb1027Platform`), and 10.5 (newest).

morozov:
Awesome work, @greg0ire. You rock!

thomasrockhu:
Seriously awesome work @greg0ire for the `upload-artifact` stuff!

.github/workflows/continuous-integration.yml
```diff
@@ -83,6 +143,362 @@ jobs:
...
+
+      - name: "Run PHPUnit"
+        run: "vendor/bin/phpunit -c ci/github/phpunit/${{ matrix.extension }}.xml --coverage-clover=coverage.xml"
+
+      - name: "Upload coverage file"
+        uses: "actions/upload-artifact@v2"
+        with:
+          name: "${{ github.job }}-${{ matrix.mariadb-version }}-${{ matrix.extension }}-${{ matrix.php-version }}.coverage"
+          path: "coverage.xml"
+
```
bendavies, line 296: extra new line here ;)

.github/workflows/continuous-integration.yml
```diff
@@ -95,7 +511,78 @@ jobs:
...
-        run: "vendor/bin/phpunit -c ci/github/phpunit.pdo-oci.xml --coverage-clover=coverage.xml"
+        run: "vendor/bin/phpunit -c ci/github/phpunit/ibm_db2.xml --coverage-clover=coverage.xml"
+
+      - name: "Upload coverage file"
+        uses: "actions/upload-artifact@v2"
+        with:
+          name: "${{ github.job }}-${{ matrix.php-version }}.coverage"
+          path: "coverage.xml"
+
+
```
bendavies, line 522: and here

bendavies:
seriously impressive work! 💪

.github/workflows/continuous-integration.yml
greg0ire: @bendavies please send a PR :P

PR #4312, opened 2020-09-29
Add enum to doctrineTypeMapping in MySqlPlatform

sami-akkawi:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

When creating a diff with a DB that uses enum in some table columns, the diff fails because 'enum' is not in the doctrineTypeMapping array. This solves this issue.


morozov:
This change will require a functional test.

PR #4317, opened 2020-10-02
Reintroduce parts of Legacy API for some more time

beberlei:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

With this PR we are re-introducing the PDO compatible API and common APIs of the DBAL "hotpath" again to keep them for a bit longer. This is too give users more time to migrate off of them, especially since many libraries and frameworks expose Doctrine DBAL via their plugin APIs.

### DIFF
--- a/src/Result.php
+++ b/src/Result.php
@@ -7,7 +7,10 @@ namespace Doctrine\DBAL;
 use Doctrine\DBAL\Driver\Exception as DriverException;
 use Doctrine\DBAL\Driver\Result as DriverResult;
 use Doctrine\DBAL\Exception\NoKeyValue;
+use LogicException;
 use Traversable;
+
+use function func_num_args;
 
 class Result
 {
@@ -241,5 +244,61 @@ class Result
         if ($columnCount < 2) {
             throw NoKeyValue::fromColumnCount($columnCount);
         }
+    }
+
+    /**
+     * BC layer for a wide-spread use-case of old DBAL APIs
+     *
+     * @deprecated This API is deprecated and will be removed after 2022
+     *
+     * @return mixed
+     */
+    public function fetch(int $mode = FetchMode::ASSOCIATIVE)
+    {
+        if (func_num_args() > 1) {
+            throw new LogicException('Only invocations with one argument are still supported by this legecy API.');
+        }
+
+        if ($mode === FetchMode::ASSOCIATIVE) {
+            return $this->fetchAssociative();
+        }
+
+        if ($mode === FetchMode::NUMERIC) {
+            return $this->fetchNumeric();
+        }
+
+        if ($mode === FetchMode::COLUMN) {
+            return $this->fetchOne();
+        }
+
+        throw new LogicException('Only fetch modes declared on Doctrine\DBAL\FetchMode are supported by legacy API.');
+    }
+
+    /**
+     * BC layer for a wide-spread use-case of old DBAL APIs
+     *
+     * @deprecated This API is deprecated and will be removed after 2022
+     *
+     * @return list<mixed>
+     */
+    public function fetchAll(int $mode = FetchMode::ASSOCIATIVE): array
+    {
+        if (func_num_args() > 1) {
+            throw new LogicException('Only invocations with one argument are still supported by this legecy API.');
+        }
+
+        if ($mode === FetchMode::ASSOCIATIVE) {
+            return $this->fetchAllAssociative();
+        }
+
+        if ($mode === FetchMode::NUMERIC) {
+            return $this->fetchAllNumeric();
+        }
+
+        if ($mode === FetchMode::COLUMN) {
+            return $this->fetchFirstColumn();
+        }
+
+        throw new LogicException('Only fetch modes declared on Doctrine\DBAL\FetchMode are supported by legacy API.');
     }
 }
--- a/tests/Functional/LegacyAPITest.php
+++ b/tests/Functional/LegacyAPITest.php
@@ -1,0 +1,176 @@
+<?php
+
+namespace Doctrine\DBAL\Tests\Functional;
+
+use Doctrine\DBAL\FetchMode;
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\Tests\FunctionalTestCase;
+use LogicException;
+
+use function array_change_key_case;
+use function array_map;
+
+use const CASE_LOWER;
+
+class LegacyAPITest extends FunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $table = new Table('legacy_table');
+        $table->addColumn('test_int', 'integer');
+        $table->addColumn('test_string', 'string');
+        $table->setPrimaryKey(['test_int']);
+
+        $sm = $this->connection->getSchemaManager();
+        $sm->dropAndCreateTable($table);
+
+        $this->connection->insert('legacy_table', [
+            'test_int' => 1,
+            'test_string' => 'foo',
+        ]);
+    }
+
+    public function tearDown(): void
+    {
+        $this->connection->executeStatement('DELETE FROM legacy_table WHERE test_int > 1');
+    }
+
+    public function testFetchWithAssociativeMode(): void
+    {
+        $sql = 'SELECT test_int FROM legacy_table WHERE test_int = 1';
+
+        $stmt = $this->connection->executeQuery($sql);
+        $row  = array_change_key_case($stmt->fetch(FetchMode::ASSOCIATIVE), CASE_LOWER);
+        self::assertEquals(1, $row['test_int']);
+    }
+
+    public function testFetchWithNumericMode(): void
+    {
+        $sql = 'SELECT test_int FROM legacy_table WHERE test_int = 1';
+
+        $stmt = $this->connection->executeQuery($sql);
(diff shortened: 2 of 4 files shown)


src/Result.php
```diff
@@ -242,4 +243,52 @@ private function ensureHasKeyValue(): void
...
+    /**
+     * BC layer for a wide-spread use-case of old DBAL APIs
+     *
+     * @deprecated This API is deprecated and will be removed after 2022
+     *
+     * @return array<array-key,mixed>|false
+     */
+    public function fetch(int $mode = FetchMode::ASSOCIATIVE)
+    {
+        if ($mode === FetchMode::ASSOCIATIVE) { // PDO::FETCH_ASSOC
```
morozov, line 256: Would it make sense to move the references to PDO to the `FetchMode` class and use a docblock with `{@link PDO::FETCH_ASSOC}` for better discoverability?

src/Result.php
```diff
@@ -242,4 +243,52 @@ private function ensureHasKeyValue(): void
             throw NoKeyValue::fromColumnCount($columnCount);
         }
     }
+
+    /**
+     * BC layer for a wide-spread use-case of old DBAL APIs
+     *
+     * @deprecated This API is deprecated and will be removed after 2022
+     *
+     * @return array<array-key,mixed>|false
```
morozov, line 252: `array<array-key,…>` looks unnecessarily verbose, it's effectively `array<mixed>`. However, is it even always an array? E.g. `fetch(FetchMode::COLUMN)` will return one value.

src/Result.php
```diff
@@ -242,4 +243,52 @@ private function ensureHasKeyValue(): void
...
+     */
+    public function fetch(int $mode = FetchMode::ASSOCIATIVE)
```
morozov, line 254: As far as I understand, we can afford to have a type declaration here because the extending classes that don't provide a type declaration won't violate LSP but the same is not true about the return types. Right?

src/Result.php
```diff
@@ -242,4 +243,52 @@ private function ensureHasKeyValue(): void
...
+
+        if ($mode === FetchMode::NUMERIC) { // PDO::FETCH_NUM
+            return $this->fetchNumeric();
+        }
+
+        if ($mode === FetchMode::COLUMN) { // PDO::FETCH_COLUMN
+            return $this->fetchOne();
+        }
+
+        throw new RuntimeException('only fetch mode PDO::FETCH_ASSOC and PDO::FETCH_NUM are supported by legacy API.');
```
morozov, line 268: I believe it should be a `LogicalException` (the error is in the code logic, not at runtime).

tests/Functional/LegacyAPITest.php
```diff
@@ -0,0 +1,120 @@
...
+
+use function array_change_key_case;
+use function array_map;
+
+use const CASE_LOWER;
+
+class LegacyAPITest extends FunctionalTestCase
+{
+    /** @var bool */
+    private static $generated = false;
```
morozov, line 17: It's better to use `dropAndCreateTable()` and make sure that the setup always produces the same sate just by that.

src/Result.php
morozov, line 268: Please have these paths also covered by a test.

src/Result.php
```diff
@@ -242,4 +243,52 @@ private function ensureHasKeyValue(): void
...
+    }
+
+    /**
+     * BC layer for a wide-spread use-case of old DBAL APIs
+     *
+     * @deprecated This API is deprecated and will be removed after 2022
+     *
+     * @return array<array-key,mixed>|false
```
morozov, line 276: Isn't it `list<mixed>|false`? The array keys are always numeric and sequential starting with `0`.

src/Result.php
morozov, line 268: Is the exception message starting with a lower-case letter on purpose? You must have written some Golang recently :-)

tests/Functional/LegacyAPITest.php
```diff
@@ -0,0 +1,120 @@
...
+        }, $rows);
+
+        self::assertEquals([['test_int' => 1]], $rows);
+
+        $stmt = $this->connection->executeQuery($sql);
+        $rows = $stmt->fetchAll(FetchMode::NUMERIC);
+        self::assertEquals([[0 => 1]], $rows);
+
+        $stmt = $this->connection->executeQuery($sql);
+        $rows = $stmt->fetchAll(FetchMode::COLUMN);
```
morozov, line 83: Should this be a different test or at least a data provider item? If the test for one mode fails, there's no reason to not test the other(s).

tests/Functional/LegacyAPITest.php
```diff
@@ -0,0 +1,120 @@
...
+        $stmt = $this->connection->executeQuery($sql);
+        $row  = array_change_key_case($stmt->fetch(FetchMode::ASSOCIATIVE), CASE_LOWER);
+        self::assertEquals(1, $row['test_int']);
+
+        $stmt = $this->connection->executeQuery($sql);
+        $row  = $stmt->fetch(FetchMode::NUMERIC);
+        self::assertEquals(1, $row[0]);
+
+        $stmt = $this->connection->executeQuery($sql);
+        $row  = $stmt->fetch(FetchMode::COLUMN);
```
morozov, line 61: Please have modes tested independently.

src/Result.php
beberlei: Yes, i added the comments before re-introducing `FetchMode` and forgot to clean this up.

src/Result.php
beberlei: My thought process was, that this is a new API so we can add whatever type declarations, however the return value is really `mixed` so i didn't declare it.

src/Result.php
beberlei: Right, changed

src/Result.php
morozov: In this case, should return type declarations be added everywhere else where possible?

src/Result.php
```diff
@@ -242,4 +243,60 @@ private function ensureHasKeyValue(): void
...
+
+        throw new LogicException('only fetch mods declared on Doctrine\DBAL\FetchMode are supported by legacy API.');
+    }
+
+    /**
+     * BC layer for a wide-spread use-case of old DBAL APIs
+     *
+     * @deprecated This API is deprecated and will be removed after 2022
+     *
+     * @return mixed
```
morozov, line 280: It's always an array of something (could be enforced at the language level), `list<mixed>` would be more specific.

tests/Functional/LegacyAPITest.php
beberlei: Ah ok, i was just replicating what `DataAccessTest` does.

src/Result.php
beberlei: in DBAL in general or what do you mean?

src/Result.php
```diff
@@ -242,4 +245,60 @@ private function ensureHasKeyValue(): void
...
+     */
+    public function fetch(int $mode = FetchMode::ASSOCIATIVE)
+    {
+        if (func_num_args() > 1) {
+            throw new LogicException('only invocations with one argument are still supported by this legecy API.');
```
morozov, line 259: The message should start with an upper-case letter.

src/Result.php
```diff
@@ -242,4 +245,60 @@ private function ensureHasKeyValue(): void
...
+
+        if ($mode === FetchMode::NUMERIC) {
+            return $this->fetchNumeric();
+        }
+
+        if ($mode === FetchMode::COLUMN) {
+            return $this->fetchOne();
+        }
+
+        throw new LogicException('only fetch mods declared on Doctrine\DBAL\FetchMode are supported by legacy API.');
```
morozov, line 274: The message should start with an upper-case letter. The "mods" should be "modes".

src/Result.php
```diff
@@ -242,4 +245,60 @@ private function ensureHasKeyValue(): void
...
+     * BC layer for a wide-spread use-case of old DBAL APIs
+     *
+     * @deprecated This API is deprecated and will be removed after 2022
+     *
+     * @return list<mixed>
+     */
+    public function fetchAll(int $mode = FetchMode::ASSOCIATIVE)
+    {
+        if (func_num_args() > 1) {
+            throw new LogicException('only invocations with one argument are still supported by this legecy API.');
```
morozov, line 287: The message should start with an upper-case letter.

src/Result.php
```diff
@@ -242,4 +245,60 @@ private function ensureHasKeyValue(): void
...
+
+        if ($mode === FetchMode::NUMERIC) {
+            return $this->fetchAllNumeric();
+        }
+
+        if ($mode === FetchMode::COLUMN) {
+            return $this->fetchFirstColumn();
+        }
+
+        throw new LogicException('only fetch mods declared on Doctrine\DBAL\FetchMode are supported by legacy API.');
```
morozov, line 302: The message should start with an upper-case letter. The "mods" should be "modes".

beberlei:
@morozov added the typehints and cleanuped everything, last point is the test, i used `DataAccessTest` as a template to create the table only once. Do you have an alternative testcase you want this to be based upon?

morozov:
@beberlei, see `BlobTest` for example.

morozov (approved):

morozov:
Thanks, @beberlei.

PR #4319, opened 2020-10-03
Bump Psalm level to 4

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

### DIFF
--- a/lib/Doctrine/DBAL/DriverManager.php
+++ b/lib/Doctrine/DBAL/DriverManager.php
@@ -12,7 +12,6 @@ use Doctrine\DBAL\Driver\SQLAnywhere;
 use Doctrine\DBAL\Driver\SQLSrv;
 
 use function array_keys;
-use function array_map;
 use function array_merge;
 use function assert;
 use function class_implements;
@@ -22,6 +21,7 @@ use function is_subclass_of;
 use function parse_str;
 use function parse_url;
 use function preg_replace;
+use function rawurldecode;
 use function str_replace;
 use function strpos;
 use function substr;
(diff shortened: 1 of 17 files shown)


psalm.xml
```diff
@@ -42,6 +42,40 @@
                 <file name="lib/Doctrine/DBAL/Driver/SQLSrv/SQLSrvConnection.php"/>
             </errorLevel>
         </FalsableReturnStatement>
+        <ForbiddenCode>
+            <errorLevel type="suppress">
+                <!-- See https://github.com/doctrine/dbal/pull/3934 -->
+                <file name="lib/Doctrine/DBAL/Logging/EchoSQLLogger.php"/>
+
+                <!-- The call to var_dump() here is by design -->
```
SenseException, line 50: If this is by design, shouldn't it be handled in the file itself in the affected line of code to provide this context/decision?

SenseException (changes requested):
I don't understand the relation of the linked PRs in the psalm.xml and how they are connected? Are e.g. the backport removals somehow the reason why these files need to be excluded because they can't be handled right now (BC)?

lib/Doctrine/DBAL/DriverManager.php
```diff
@@ -412,23 +418,22 @@ private static function parseSqliteDatabaseUrlPath(array $url, array $params): a
...
-     * @param mixed[] $params The connection parameters to resolve.
+     * @param string|null $scheme The connection URL scheme, if available
+     * @param mixed[]     $params The connection parameters to resolve.
      *
      * @return mixed[] The resolved connection parameters.
      *
      * @throws Exception If parsing failed or resolution is not possible.
      */
-    private static function parseDatabaseUrlScheme(array $url, array $params): array
+    private static function parseDatabaseUrlScheme($scheme, array $params): array
```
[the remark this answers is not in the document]
morozov: This is intentional. I deliberately didn't want to introduce any, even theoretical, risk in a patch release. I wouldn't think twice if it was a minor release.

psalm.xml
morozov: Do you suggest to use in-file annotations instead of suppressing the error here? I'd like to avoid having annotations unrelated to the code if possible. Additionally, this entire file is a dumper, so no matter where in the file `var_dump()` is called, it's by design.

morozov:
> I don't understand the relation of the linked PRs in the psalm.xml and how they are connected? Are e.g. the backport removals somehow the reason why these files need to be excluded because they can't be handled right now (BC)?

Each issue suppressed in the configuration has a reference to the PR where it's fixed in a BC-breaking way, and there's no other way to fix it. After these changes get merged up, those suppressions may be removed (in 3.0 or 4.0 where the corresponding changes were merged).

psalm.xml
SenseException: > Additionally, this entire file is a dumper, so no matter where in the file var_dump() is called, it's by design.

I guess this answer counts for letting it stay in psalm.xml.
A lot of the suppressed rules are supposed to be fixed at one time or another, but I tend to use inline annotation when the affected code looks like this by design (or can't be disabled more selective in the config with the analysis tool of choice).

SenseException (approved):

Issue #3746, opened 2019-11-25
VersionAwarePlatformDriver should extend Driver


morozov:
According to https://github.com/sebastianbergmann/phpunit/issues/3955, the fact that some of our tests mock a combination of `Driver` and `VersionAwarePlatformDriver` indicates poor design.

Indeed, in its essence, `VersionAwarePlatformDriver` is a subtype of `Driver`, not a separate type and should be defined accordingly.

PR #4323, opened 2020-10-04
Clean up redundant implements

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

`ServerInfoAwareConnection` extends `Connection`, so it is redundant to implement both.

---

BTW, why was `ServerInfoAwareConnection` changed to extend `Connection`? I found the old design much better, when interfaces were independent, and connection classes could compose their own set of implementations. What if you need to introduce `SomethingElseAwareConnection` in the future? Will it extend `Connection` as well? Sure, you'll still be able to implement both of them and it will work, but will look weird if you ask me.

### DIFF
--- a/src/Driver/OCI8/Connection.php
+++ b/src/Driver/OCI8/Connection.php
@@ -2,7 +2,6 @@
 
 namespace Doctrine\DBAL\Driver\OCI8;
 
-use Doctrine\DBAL\Driver\Connection as ConnectionInterface;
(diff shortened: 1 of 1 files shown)


morozov:
@BenMorel see the rationale in https://github.com/doctrine/dbal/issues/3746.

morozov (approved):

morozov:
Thanks, @BenMorel!

Issue #3166, opened 2018-05-26
Improve exception hiearchy


Majkl578:
> 1. All transaction-related exceptions extend ConnectionException which probably could renamed to TransactionException.
> 2. NonUniqueFieldNameException directly extending ServerException is weird.
> 3. TableExistsException extending ServerException is weird too.
> 4. ServerException extending DriverException
> 5. More like these.

This should be follow-up of #3131 once merged.

morozov:
Some remaining notes from the previous issue: https://github.com/doctrine/dbal/pull/3131#issuecomment-388572889.

PR #4324, opened 2020-10-04
Throw exception from lastInsertId() if there is no identity value

### LINKED ISSUE #3596: Rework Connection::lastInsertId()
Instead of the optional parameter, there should be two separate methods: one which returns the last inserted record ID if supported by the driver, the other returns the current value of the given sequence if supported by the platform.

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | N/A

#### Summary

Supersedes #3759 and #3368.

**Proposed changes**

This PR splits the existing `lastInsertId(?string $name = null): string` method into 2 methods:

- `getLastInsertId() : int|string`
- `getSequenceNumber(string $name): int|string`

Tests have been extended to cover the following scenarios:

- `getLastInsertId()` on supported platforms
- `getLastInsertId()` on supported platforms, on a fresh connection, i.e. with no ID available (`NoIdentityValue` exception)
- `getLastInsertId()` on unsupported platforms (`IdentityColumnsNotSupported` exception)
- `getSequenceNumber()` with a known sequence name on supported platforms
- `getSequenceNumber()` with an unknown sequence name on supported platforms (`SequenceDoesNotExist` exception)
- `getSequenceNumber()` on unsupported platforms (`SequencesNotSupported` exception)


src/Driver/PDO/Connection.php
```diff
@@ -87,17 +91,80 @@ public function quote(string $value): string
...
+
+        try {
+            $sequenceNumber = $this->connection->lastInsertId($name);
         } catch (PDOException $exception) {
+            [$sqlState] = $exception->errorInfo;
+
+            // PDO PGSQL throws an 'Undefined table' when the sequence does not exist
+            if ($sqlState === '42P01' && $this->connection->getAttribute(PDO::ATTR_DRIVER_NAME) === 'pgsql') {
+                throw SequenceDoesNotExist::new($name, $exception);
+            }
```
BenMorel, line 149: I'm not comfortable with having these checks in the driver Connection rather than in the `ExceptionConverter`, but we need some context when converting to an exception, as `42P01` can either mean a `TableNotFoundException` or a `SequenceDoesNotExist`.

src/Driver/PDO/SQLite/Connection.php
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getSequenceNumber(string $name)
+    {
+        // SQLite does not support sequences. However, PDO::lastInsertId() ignores the name parameter, and returns
+        // the last insert ID even if a sequence name is given. We expect an exception in that case.
+        throw SequencesNotSupported::new();
+    }
```
BenMorel, line 59: Do we really want to make `Driver\PDO\Connection` final? Because of this, we have to create decorator classes just to make `lastInsertId()` or `getSequenceNumber()` throw, with all other methods just proxying to the decorated driver connection.

Sure, we could create an `AbstractConnectionDecorator`, but then, why override the decorator's method rather than the Connection's method directly?

Another option is to not create driver-specific connection classes for this purpose, but then we end up with [this](https://github.com/doctrine/dbal/pull/4324/files#diff-d4cd4042dbac00d3f1359475b4168d72R99-R103), which I doubt is good design.

src/Driver/PDO/Connection.php
```diff
@@ -87,17 +91,80 @@ public function quote(string $value): string
...
+    /**
+     * {@inheritDoc}
+     */
+    public function lastInsertId()
     {
+        $driverName = $this->connection->getAttribute(PDO::ATTR_DRIVER_NAME);
+
+        if ($driverName === 'oci') {
+            throw IdentityColumnsNotSupported::new();
+        }
```
BenMorel, line 103: Just as demonstration of an alternative to subclassing. Please see [this comment](https://github.com/doctrine/dbal/pull/4324/files#r499686504).

BenMorel:
@morozov Ready for discussion!

src/Connection.php
```diff
@@ -1094,23 +1098,46 @@ public function getTransactionNestingLevel(): int
...
+    }
+
+    /**
+     * Returns the current sequence value for the given sequence name.
+     *
+     * If the underlying driver does not support sequences, an exception is thrown.
+     *
+     * @param string $name The sequence name.
+     *
+     * @return int|string The sequence number, as an integer or a numeric string.
```
morozov, line 1129: I remember having discussions about what return type should these methods have but I don't remember if we came to any conclusion. It definitely cannot be an `int` because the value may not fit into `int32` but it could be a numeric `string` enforced at the method declaration level. Why `int|string`? Should we document the decision-making process?

src/Connection.php
```diff
@@ -1094,23 +1098,46 @@ public function getTransactionNestingLevel(): int
...
      */
-    public function lastInsertId(?string $name = null): string
+    public function getSequenceNumber(string $name)
     {
         try {
-            return $this->getWrappedConnection()->lastInsertId($name);
+            return $this->getWrappedConnection()->getSequenceNumber($name);
+        } catch (DriverException\SequencesNotSupported $e) {
+            throw new SequencesNotSupported($e->getMessage(), $e);
+        } catch (DriverException\SequenceDoesNotExist $e) {
```
morozov, line 1139: Same as above, the specialized wrapper-level exception classes are redundant IMO.

src/Driver/Exception/IdentityColumnsNotSupported.php
```diff
@@ -13,8 +14,8 @@
  */
 final class IdentityColumnsNotSupported extends AbstractException
 {
-    public static function new(): self
+    public static function new(?Throwable $previous = null): self
```
morozov, line 17: What's the use case for the `$previous` exception? It doesn't seem to be used anywhere.

src/Driver/PDO/Connection.php
```diff
@@ -87,17 +91,80 @@ public function quote(string $value): string
...
 
-    public function lastInsertId(?string $name = null): string
+    /**
+     * {@inheritDoc}
+     */
+    public function lastInsertId()
     {
+        $driverName = $this->connection->getAttribute(PDO::ATTR_DRIVER_NAME);
+
+        if ($driverName === 'oci') {
```
morozov, line 101: Should this be a new connection class?

src/Driver/PDO/Connection.php
```diff
@@ -87,17 +91,80 @@ public function quote(string $value): string
...
         try {
-            if ($name === null) {
-                return $this->connection->lastInsertId();
+            $lastInsertId = $this->connection->lastInsertId();
+        } catch (PDOException $exception) {
+            [$sqlState] = $exception->errorInfo;
+
+            // PDO PGSQL throws a 'lastval is not yet defined in this session' error when no identity value is
+            // available, with SQLSTATE 55000 'Object Not In Prerequisite State'
+            if ($sqlState === '55000' && $this->connection->getAttribute(PDO::ATTR_DRIVER_NAME) === 'pgsql') {
```
morozov, line 112: This looks like logic for the `ExceptionConverter`.

src/Driver/PDO/Connection.php
```diff
@@ -87,17 +91,80 @@ public function quote(string $value): string
...
+
+        return $sequenceNumber;
+    }
+
+    /**
+     * @param int|string $value
+     */
+    private function isValidLastInsertIdOrSequenceNumber($value): bool
+    {
+        // pdo_mysql & pdo_sqlite return 0 or '0', pdo_sqlsrv returns ''
```
morozov, line 166: Looks like a lot of driver-specific logic in a driver-agnostic place. I'm curious if it can have false-positive effects. E.g. an auto-incremented field is incremented form `-1` to `0`.

Just a thought, would it be safer if we only failed if it's not a numeric string? Otherwise, if neither the user nor the DB driver care about the proper logic, why should we?

src/Driver/PDO/SQLSrv/Connection.php
```diff
@@ -42,15 +43,29 @@ public function exec(string $sql): int
...
+    }
 
-        return $this->prepare('SELECT CONVERT(VARCHAR(MAX), current_value) FROM sys.sequences WHERE name = ?')
+    /**
+     * {@inheritDoc}
+     */
+    public function getSequenceNumber(string $name)
+    {
+        $sequenceNumber =
+            $this->prepare('SELECT CONVERT(VARCHAR(MAX), current_value) FROM sys.sequences WHERE name = ?')
```
morozov, line 60: I'm curious why this `CONVERT(VARCHAR(MAX), ...)` is needed and is it related to the `string|int` type.

src/Driver/PDO/SQLite/Connection.php
morozov: I don't see much problem in creating a decorator.

src/Connection.php
```diff
@@ -1094,23 +1098,46 @@ public function getTransactionNestingLevel(): int
...
-     * @return string A string representation of the last inserted ID.
+     * @throws Exception
+     */
+    public function lastInsertId()
+    {
+        try {
+            return $this->getWrappedConnection()->lastInsertId();
+        } catch (DriverException\IdentityColumnsNotSupported $e) {
+            throw new IdentityColumnsNotSupported($e->getMessage(), $e);
+        } catch (DriverException\NoIdentityValue $e) {
```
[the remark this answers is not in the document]
BenMorel: I agree, except for testability. Looking back to https://github.com/doctrine/dbal/pull/3759#discussion_r353325041, I think we may get away with testing for a specific exception message indeed, but this is slightly less clean.

src/Connection.php
BenMorel: Please see https://github.com/doctrine/dbal/pull/3759#discussion_r353246669

src/Driver/Exception/IdentityColumnsNotSupported.php
BenMorel: Correct. Cleaned up in 5ff918d2f00ab65c235836791f7ba7acafeb6312.

src/Driver/PDO/Connection.php
BenMorel: Yes, this was left as a demonstration for an alternative to subclassing. See https://github.com/doctrine/dbal/pull/4324/files#r499686504

src/Driver/PDO/Connection.php
BenMorel: Please see [this comment](https://github.com/doctrine/dbal/pull/4324/files/83e4265f9d3172e507d6b33d06156fcc66dd6fea#r499680978).

`ExceptionConverter` lacks some context (what method was being called when the error occurred): a given error code may for example represent a Table not found, or a Sequence does not exist.

If you're not willing to introduce wrapper-level exception classes for identity columns / sequences -related exceptions, then we can just remove this code altogether and get away with it.

src/Driver/PDO/Connection.php
BenMorel: > E.g. an auto-incremented field is incremented form -1 to 0.

I'm not sure if any DB allows negative auto-increments? MySQL at leasts fails:

```
ALTER TABLE ai AUTO_INCREMENT=-1;
#1064 - You have an error in your SQL syntax (...)
```

> Otherwise, if neither the user nor the DB driver care about the proper logic, why should we?

Consistency. If one driver fails with an exception when a value is not available, I think all of them should. If this driver-specific logic is a problem, maybe the individual checks can be moved to the driver-specific connection class: `PDO\SQLSrv\Connection` would only check for an empty string, while others would only check for `0`, etc.?

src/Driver/PDO/SQLite/Connection.php
BenMorel: > I don't see much problem in creating a decorator.

Do you mean an abstract decorator, or a decorator for each Connection class?

Creating a decorator for each Connection class looks cumbersome to me, and creates too much distraction when reading the Connection classes' code: you can't clearly separate the custom logic from the boilerplate logic. 

On the other hand, if your suggestion is to create an abstract decorator: what's the point of making a class `final` to then create an abstract decorator to extend from? That's an extra class for... nothing if you ask me.

src/Connection.php
BenMorel: Do we need driver-level exception classes then? We could just throw a generic driver exception, because now they don't need to be caught.

src/Driver/PDO/SQLSrv/Connection.php
BenMorel: Looks like this was to respect the `: string` return type, but then I'm not sure why the cast was not done in PHP? I'll try to rework this if you confirm that we're going the `int|string` route.

src/Driver/PDO/SQLite/Connection.php
morozov: My suggestion is to create a class for each connector. It may look cumbersome because PHP as a language doesn't make it straightforward, not because of the composition approach itself.

src/Driver/PDO/Connection.php
morozov: So are you suggesting to move this error handling to the exception converter or to leave it here?

src/Driver/PDO/Connection.php
morozov: This would work too. I don't think that sub-classing in this case is much better than if-statements. If it doesn't affect the API, we can change it as soon as needed.

src/Connection.php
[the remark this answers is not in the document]
morozov: We can and absolutely should have different driver-level exceptions and test this at the driver level but we don't necessarily need different wrapper-level exceptions.

src/Connection.php
morozov: Thanks for digging it out. Let's have it declared as `int|string`.

src/Connection.php
morozov: > Do we need driver-level exception classes then?

1. Testability. We can expect a certain exception class in a certain case.
2. Ease of construction. The named constructor can accept exception-specific parameters and encapsulate the creation of the exception message.

Note that all driver-level exceptions are marked as `@internal`, so if you try to catch a specific class in your code, some code analyzers will tell you that you're doing it wrong.

src/Driver/PDO/Connection.php
morozov: I'm good with this approach as well. Let's start simple, by having all the logic in one class, and introduce sub-classes if necessary.

src/Driver/PDO/Connection.php
morozov: > Consistency.

This may provide a false sense of consistency. As it's already pointed out in the documentation, it is recommended to use this method right after the insert, otherwise the behavior depends on the platform. Why do we want to enforce consistency in the case which is not recommended and where full consistency is not possible?

src/Driver/PDO/SQLSrv/Connection.php
morozov: > I'll try to rework this if you confirm that we're going the int|string route.

Let's do.

src/Driver/PDO/Connection.php
BenMorel: If we want to move it to the exception converter, we have to add a bit of context to the API, but I am not sure in which form: name of the calling method? An enum value like `QUERY` / `LAST_INSERT_ID` / `GET_SEQUENCE_NUMBER`? Or just leave this code here, but it's not pretty. TBH I don't know what's the best approach, or even if there is one.

src/Driver/PDO/Connection.php
BenMorel: Unless I didn't get you, this contradicts https://github.com/doctrine/dbal/pull/4324#discussion_r500568276. Which way do you want to go?

src/Connection.php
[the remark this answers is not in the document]
BenMorel: OK, I guess I'll remove the wrapper-level exceptions then. Just to be sure, why should `TableNotFoundException` exist at wrapper level, when `SequenceDoesNotExist` (inconsistently named, BTW) shouldn't?

src/Driver/PDO/Connection.php
morozov: The linked comment clarifies that I was suggesting an individual class per driver, not an abstract decorator. Although I originally suggested to use driver-specific classes, I'm fine if we start without them.

src/Connection.php
BenMorel: > 1. Testability. We can expect a certain exception class in a certain case.

<s>As far as testability is concerned, this would only be useful if we specifically tested each driver in isolation; as it stands, tests target the wrapper, so we have no way to catch the driver-level exception.</s> Actually we do: we can test the previous exceptions and see if any of them is the driver-level exception. Is that acceptable in tests?

> 2. Ease of construction. The named constructor can accept exception-specific parameters and encapsulate the creation of the exception message.

If this is the only justification, we can also create factory methods on a generic driver-level class for these use cases, if we want to avoid creating exceptions that will never be caught anywhere.

> Note that all driver-level exceptions are marked as @internal, so if you try to catch a specific class in your code, some code analyzers will tell you that you're doing it wrong.

You should have no way to catch driver-level exceptions anyway, as they'll will have been converted to wrapper-level exceptions before they can reach userland code. Unless I missed something?

src/Connection.php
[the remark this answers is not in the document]
morozov: > why should TableNotFoundException exist at wrapper level

I can't answer that. It was added 7 years ago in https://github.com/doctrine/dbal/pull/478. IMO, it shouldn't as well.

src/Driver/PDO/Connection.php
BenMorel: Getting an outdated value (i.e. for an older INSERT) is one thing we can't avoid, and there will be inconsistencies between drivers.

But getting an empty- or zero-value is something that, IMO, should consistently throw an exception. You *do* want your code to fail if for any reason, you're calling this method after an INSERT to a table that doesn't have an auto-increment field: you don't want it to silently continue with a zero value.

morozov: I agree. My point was that a zero value is only possible in the case of a logical error, which we declare to not handle consistently. Even this zero value may or may not happen depending on the currently used driver, so we won't be able to throw this exception consistently with the same invalid logic but different drivers.

I'm not strongly opposed to handling zero values, I'm just trying to come up with a clear understanding of what our API contract is.

src/Driver/PDO/SQLSrv/Connection.php
BenMorel: Done in befc0dd4937a53fb8abfa16e9c5403a30ee0d5c5!

src/Connection.php
BenMorel: Alright, nothing to change here then! Resolving.

src/Connection.php
morozov: > [...] we can test the previous exceptions and see if any of them is the driver-level exception. Is that acceptable in tests?

As long as there no dedicated wrapper-level exception, I think yes. Some tests work with the drivers directly, it may be even better if we want to not run the code that is not part of the behavior being tested.

> If this is the only justification, we can also create factory methods on a generic driver-level class for these use cases, if we want to avoid creating exceptions that will never be caught anywhere.

How is the exception produced by a factory method meant to be identified? Also, the single responsibility principle that the generic driver-level class will violate.

> You should have no way to catch driver-level exceptions anyway, as they'll will have been converted to wrapper-level exceptions before they can reach userland code. Unless I missed something?

Right, but if not marked internal explicitly they are automatically part of the public API of the library, but I want to be flexible in changing how the exceptions are constructed w/o the consumers of the API being impacted by the changes.

BenMorel: > How is the exception produced by a factory method meant to be identified?

That's the whole point: if we don't catch the specific driver exception (`SequencesNotSupported`, for example), the exception does not need to be identified;.

> Also, the single responsibility principle that the generic driver-level class will violate.

I don't think factory methods are a violation of the SRP, they all create a driver exception with a specific error message.

But, anyway, I think keeping internal exceptions for identity/sequences is a good idea.

> Right, but if not marked internal explicitly they are automatically part of the public API of the library, but I want to be flexible in changing how the exceptions are constructed w/o the consumers of the API being impacted by the changes.

Sure, I will remove the public exceptions:

- remove the wrapper-level exceptions
- make driver-level exceptions `@internal`
- change the tests so that they check the previous exceptions looking for the internal driver exception

morozov: > if we don't catch the specific driver exception (`SequencesNotSupported`, for example), the exception does not need to be identified

It needs to be identified for testing. I want to make sure that the driver exception that occurs during a test is exactly the `SequenceDoesNotExist` exception, not whatever connection/syntax error.

As for `SequencesNotSupported`, it's a totally different story. The fact that the user finds out that they cannot use a certain API only at runtime is a design smell. The `AbstractPlatform` class contains ~200 public methods which makes it quite a god object, and this is a side effect of that.

BenMorel: Removed the public exceptions in b6f1c06f6552e69917dc9adb8ab7f3522b65394e, please tell me if you're happy with the `expectPreviousException()` thing.

BenMorel:
@morozov I'd like to come back to this. Before diving again in this PR, should I retarget `3.0.x` instead of `master`?

morozov (commented):
> I'd like to come back to this.

That would be awesome!

> Before diving again in this PR, should I retarget `3.0.x` instead of `master`?

`3.0.0` is already released.

BenMorel:
> `3.0.0` is already released.

Oh wow, I didn't expect it anytime soon. I had quite a few things I would have liked to propose for 3.0, but looks like they'll have to wait for 4.0 now. Any rough ETA on this one? I mean, can we expect it in a reasonable timeframe like a year from now, or will we have to wait for multiple years like Doctrine 2 to 3?

src/Driver/PDO/Connection.php
BenMorel: I think it's good enough to leave it there for now; adding complexity / extra parameters to the ExceptionConverter is a bit premature IMO, and may do more harm than good.

src/Driver/PDO/SQLite/Connection.php
BenMorel: I'm not convinced, but I'll leave the decorator there, then. If you happen to change your mind, we can remove `final` from the  `Driver\PDO\Connection`.

src/Driver/PDO/Connection.php
BenMorel: Actually, let's start without decorators. A few extra conditions in `PDO\Connection` don't look too bad, and add much less clutter than creating a decorator for each platform. Please let me know if you're still happy with that.

src/Driver/Mysqli/Connection.php
```diff
@@ -118,9 +118,23 @@ public function exec(string $sql): int
...
 
-    public function lastInsertId(?string $name = null): string
+    /**
+     * {@inheritDoc}
+     */
+    public function lastInsertId()
+    {
+        $lastInsertId = $this->conn->insert_id;
+
+        if ($lastInsertId === 0 || $lastInsertId === '0') {
```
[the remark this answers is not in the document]
BenMorel: [Yes](https://www.php.net/manual/en/mysqli.insert-id.php).

> If the number is greater than maximal int value, mysqli_insert_id() will return a string.

I actually just [filed a Psalm PR](https://github.com/vimeo/psalm/pull/4577).

src/Driver/PDO/Connection.php
BenMorel: I guess our API contract is:

- you are not guaranteed to get a value if you don't call `getLastInsertId()` right after an `INSERT` statement
- if you do not get a value, you get an exception

Which the current code fulfills well.

BenMorel:
@morozov Ready for review again. The main points that you may want to have a look at are:

- [PDO driver-specific checks](https://github.com/doctrine/dbal/pull/4324/files#diff-1b2ab1855552d12552ffcda6afd7f3773aaf435d258004f46c2dcbd7f1f371edR134-R146) in `Driver\PDO\Connection::getLastInsertId()` and `getSequenceNumber()`, rather than creating a decorator for each PDO driver just to throw an exception
- [PGSQL-specific exception checks](https://github.com/doctrine/dbal/pull/4324/files#diff-1b2ab1855552d12552ffcda6afd7f3773aaf435d258004f46c2dcbd7f1f371edR113) that we cannot move to the `ExceptionConverter` without modifying its signature somehow to accept a context (the same error code can mean 2 different things depending on the operation that was being performed when it was thrown); I think that adding these 2 checks right in the `Driver\PDO\Connection` class is a pragmatic way to avoid adding unnecessary complexity to the `ExceptionConverter`
- [Check for underlying driver exception in tests](https://github.com/doctrine/dbal/pull/4324/files#diff-cc769e60784447c54cd5424584c82fe100c547591900434e87c1c197b31856c4R436-R459), as you don't want to introduce wrapper-level exceptions for identity / sequence errors, but do want to introduce driver-level exceptions to be used in tests
- settle on whether an exception should be thrown when no identity value / sequence number is available (https://github.com/doctrine/dbal/pull/4324#discussion_r525029676)
- and finally naming: I suggest using `getLastInsertId()` instead of `lastInsertId()`, which is already reflected in the code

src/Driver/Mysqli/Connection.php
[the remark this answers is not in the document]
morozov: But zero isn't greater than `PHP_INT_MAX`.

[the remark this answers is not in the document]
BenMorel: Good point. I'll try without this check.

morozov:
@BenMorel thanks for the update. I've been short on time for OSS lately and I cannot promise I'll look at it immediately but I'll do as soon as I can.

src/Driver/Mysqli/Connection.php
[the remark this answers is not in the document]
BenMorel: Looks like you were right. Fixed in da1fd377ed9b269317c49d3d8c6e94d520539121!

src/Driver/PDO/Connection.php
```diff
@@ -88,17 +92,86 @@ public function quote(string $value): string
...
+        }
+
+        if ($driverName === 'sqlite') {
+            // SQLite does not support sequences. However, PDO::lastInsertId() ignores the name parameter, and returns
+            // the last insert ID even if a sequence name is given. We expect an exception in that case.
+            throw SequencesNotSupported::new();
+        }
+
+        if ($driverName === 'oci') {
+            throw new SequencesNotSupported('Oracle supports sequences, but the PDO OCI driver does not support them.');
```
morozov, line 145: This is where it becomes clear that sequences are a platform-level concern, not a driver/connection-level concern. I believe the `getSequenceNumber()` should be available only at the wrapper level. The platform level then would implement something like `getGetSequenceNumberSQL(string $name): string` which the wrapper would execute.

There is no point of having different implementation and different behavior of sequences on the same platform depending on the driver.

src/Driver/PDO/Connection.php
```diff
@@ -88,17 +92,86 @@ public function quote(string $value): string
...
+        }
+
+        try {
+            $sequenceNumber = $this->connection->lastInsertId($name);
         } catch (PDOException $exception) {
+            [$sqlState] = $exception->errorInfo;
+
+            // PDO PGSQL throws an 'Undefined table' when the sequence does not exist
+            if ($sqlState === '42P01' && $this->connection->getAttribute(PDO::ATTR_DRIVER_NAME) === 'pgsql') {
```
morozov, line 154: The approach of implementing sequences only at the wrapper level will also allow for more reasonable error handling which is implemented at the wrapper level (see `ExceptionConverter` and `DriverException` subclasses).

src/Driver/PDO/Connection.php
```diff
@@ -88,17 +92,86 @@ public function quote(string $value): string
...
+        return $sequenceNumber;
+    }
+
+    /**
+     * @param int|string $value
+     */
+    private function isValidLastInsertIdOrSequenceNumber($value): bool
+    {
+        // pdo_mysql & pdo_sqlite return 0 or '0', pdo_sqlsrv returns ''
+        return $value !== 0 && $value !== '0' && $value !== '';
```
morozov, line 174: By the moment when this code is executed, we know the driver name. Should we make this code explicitly dependent on the driver?

morozov:
@BenMorel, just FYI, I squashed and rebased your PR to accommodate the changes from https://github.com/doctrine/dbal/pull/4691. This way, the scope of this PR boils down to changing the return type of `lastInsertId()` (we could probably leave the method unchanged) and improving error handling.

I have a couple more ideas on improving the code, will document later.

src/Driver/PDO/Connection.php
```diff
@@ -88,13 +90,45 @@ public function quote(string $value): string
...
 
-    public function lastInsertId(): string
+    /**
+     * {@inheritDoc}
+     */
+    public function lastInsertId()
     {
+        $driverName = $this->connection->getAttribute(PDO::ATTR_DRIVER_NAME);
+
+        if ($driverName === 'oci') {
```
morozov: Instead of presuming the driver's incapability of fulfilling the API, we should let the driver explicitly report that. See: https://github.com/doctrine/dbal/blob/f18aee0a2b209662370de14f3a5268553c9b239e/tests/Functional/WriteTest.php#L366-L369

tests/Functional/WriteTest.php
```diff
@@ -307,4 +323,30 @@ public function testDeleteWhereIsNull(): void
...
+            $e = $exception;
+            do {
+                if ($e instanceof $expectedExceptionClass) {
+                    $this->addToAssertionCount(1);
+
+                    return;
+                }
+
+                $e = $e->getPrevious();
+            } while ($e);
```
morozov: This way, the test allows the driver exception to be at any nesting level which is kinda odd: we know exactly where it's supposed to be.

morozov:
@BenMorel I did some more cleanup, and with the reduced scope, I believe it's good to go. The part about exceptions could still use some improvement but it's orthogonal to this patch, but it's orthogonal to this patch and should be taken care of by https://github.com/doctrine/dbal/issues/3166.

morozov (approved):

morozov:
Thanks for the patch!

PR #4325, opened 2020-10-05
Remove implicit ext-json dependency

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

The codebase contains [calls to `json_encode()`](https://github.com/doctrine/dbal/blob/master/src/Connection.php#L1632), yet this is not documented in the requirements.

The `json` extension [can be disabled](https://stackoverflow.com/a/61230713/759866) in PHP 7!

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -36,14 +36,21 @@ use function array_map;
 use function assert;
 use function bin2hex;
 use function count;
+use function extension_loaded;
+use function gettype;
 use function implode;
+use function is_array;
 use function is_int;
-use function is_resource;
+use function is_scalar;
 use function is_string;
-use function json_encode;
 use function key;
+use function mb_check_encoding;
+use function ord;
+use function preg_match;
 use function preg_replace;
 use function sprintf;
+use function strlen;
+use function var_export;
 
 /**
  * A database abstraction-level connection that implements features like events, transaction isolation levels,
@@ -1597,7 +1604,7 @@ MESSAGE;
             $messageFormat,
             $sql,
             $params !== []
-                ? sprintf(' with params %s', $this->formatParameters($this->resolveParams($params, $types)))
+                ? sprintf(' with params %s', $this->formatParameter($this->resolveParams($params, $types)))
                 : '',
             $e->getMessage()
         );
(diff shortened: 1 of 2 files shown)


morozov:
The JSON extension is used in literally [a couple of places](https://github.com/doctrine/dbal/search?q=json_encode&type=code) of which only the usage for the JSON type is really justified. This doesn't look a good reason to make it a core dependency. I'd rather rework the Exception class to not depend on it. As for the JSON type, its usage is optional, so it doesn't necessitate adding the dependency IMO (same as there's no dependency on any of the DB drivers).

michaljusiega:
Since PHP 8.0 JSON extension will be always available: https://wiki.php.net/rfc/always_enable_json

BenMorel:
> This doesn't look like a good reason to make it a core dependency. I'd rather rework the Exception class to not depend on it.

I agree, but as it stands, it *is* a core dependency at the moment, due to this usage. I'll try to refactor the Exception class then. Leaving this issue open until then.

BenMorel:
OK, I actually did it in this PR. Ready for review, @morozov!

This would maybe benefit from being moved to a separate class, like `ParameterFormatter`?

src/Connection.php
```diff
@@ -1617,27 +1624,71 @@ public function createQueryBuilder(): QueryBuilder
...
     {
-        return '[' . implode(', ', array_map(static function ($param): string {
-            if (is_resource($param)) {
-                return (string) $param;
+        if ($param === null) {
+            return 'null';
+        }
+
+        if (is_scalar($param)) {
+            if (is_string($param) && (! $this->isValidUTF8($param) || $this->containsNonPrintableASCIIChars($param))) {
```
morozov, line 1639: After sleeping on it, I'm not sure if DBAL should be in the business of formatting prepared statement parameters at all.

The responsibility of the exception is to tell what went wrong by its error message and encapsulate the context for future handling or troubleshooting. By imprinting prepared statement parameters into the message, it may do a disservice: it makes it impossible to log the error message w/o logging the potentially sensitive information stored in the exception context (e.g. tokens, password hashes, personally identifiable information, etc).

Let's try to explore the approach of storing the parameters with the exception w/o making them part of the message. This way, the handling part will be free to process them in an appropriate way (e.g. ignore all parameters except for passwords and tokens) or ignore them.

GrahamCampbell:
This should probably target 2.11.x?

src/Connection.php
BenMorel: Prototype available in #4387!

morozov:
Closing in favor of https://github.com/doctrine/dbal/pull/4387.

PR #4326, opened 2020-10-05
Fix typo in constant name

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | none

#### Summary

`DOENST` => `DOESNT`

### DIFF
--- a/src/Schema/Exception/SequenceDoesNotExist.php
+++ b/src/Schema/Exception/SequenceDoesNotExist.php
@@ -17,7 +17,7 @@ final class SequenceDoesNotExist extends SchemaException
     {
         return new self(
             sprintf('There exists no sequence with the name "%s".', $sequenceName),
-            self::SEQUENCE_DOENST_EXIST
+            self::SEQUENCE_DOESNT_EXIST
         );
     }
 }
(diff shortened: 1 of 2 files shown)


SenseException:
@doctrine/team-dbal Should this still target `3.0.x`? I suppose it also needs a deprecation path from the latest 2 version.

BenMorel:
I still don't get the difference between `master` and `3.0.x` I'm afraid. But you're right, this constant also exists in 2.x anyway so it should at least be mentioned in the changelog. I'm not sure whether it's a big deal to provide a deprecation path for this, but I'm all ears.

@morozov?

src/Schema/Exception/SequenceDoesNotExist.php
```diff
@@ -17,7 +17,7 @@ public static function new(string $sequenceName): self
     {
         return new self(
             sprintf('There exists no sequence with the name "%s".', $sequenceName),
-            self::SEQUENCE_DOENST_EXIST
+            self::SEQUENCE_DOESNT_EXIST
```
morozov, line 20: This exception has a unique class by which it can be identified both in code and in logs. Why does it need a numeric error code?

BenMorel: No worries. I thought this was a no-brainer until I realized that this was introduced in 2.x, not 3.x.
I wouldn't worry that much about the BC break in 3.x though, considering the amount of breaking changes, this one is minor.

Do you want to close this anyway?

morozov: Yes, I’d close it.

BenMorel: Alright!

PR #4337, opened 2020-10-12
Rename Comparator::diffTable() params - from/to instead of 1/2

mvorisek:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | no issue, pure refactor

#### Summary

swapping input of `Comparator::diffTable` will produce completely different result - the first gven table is expected to be the "from" table and the 2nd table is expected to be the "to" table

the names are also consistent with `Comparator::compare` (which uses `$fromSchema`, `$toSchema` names)

no more changes in `Comparator` class needed - `Comparator::diffSequence` uses `$sequence1` and `$sequence2` names, but the inputs can be swapped, so these names should stay

### DIFF
--- a/lib/Doctrine/DBAL/Schema/Comparator.php
+++ b/lib/Doctrine/DBAL/Schema/Comparator.php
@@ -188,24 +188,24 @@ class Comparator
     }
 
     /**
-     * Returns the difference between the tables $table1 and $table2.
+     * Returns the difference between the tables $fromTable and $toTable.
      *
      * If there are no differences this method returns the boolean false.
      *
      * @return TableDiff|false
      */
-    public function diffTable(Table $table1, Table $table2)
+    public function diffTable(Table $fromTable, Table $toTable)
     {
         $changes                     = 0;
-        $tableDifferences            = new TableDiff($table1->getName());
-        $tableDifferences->fromTable = $table1;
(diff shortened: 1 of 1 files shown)


greg0ire:

mvorisek:
@greg0ire squashed

greg0ire (dismissed):
Let's do this before it becomes a BC-break 👍

morozov (dismissed):

greg0ire (approved):

greg0ire:
Rebasing because of new pipeline

greg0ire:
Thanks @mvorisek !

@morozov Codecov is behaving a bit weird, it seems to need time to process the ~50 reports we send, then it show the right percentage diff in Github's UI, but in its own UI, it's correct in some places, outdated and alarming in some others, for example the Coverage change tab is wrong and so is the Overview graph cc @thomasrockhu

thomasrockhu:
@greg0ire, this is indeed strange. Would you be able to open a ticket in our community boards? I'll take a look.

greg0ire:
Looks like it's fixed now, it just seems to take a lot of time. Should I still open the ticket? It will be hard for you to investigate anything I'm afraid.

thomasrockhu:
@greg0ire, ah got it. I'll keep an eye out, but if you see it happening, please feel free to open.

Issue #3871, opened 2020-02-04
Partial unique index on datetime field generates same migration every time


tejerka:
#### Summary

I add a partial unique index on a datetime field and since every execution of
`bin/console doctrine:migration:diff` create the same file.

#### Current behaviour

my entity annotation :

```php
/**
 * @ORM\Table(name="pages", uniqueConstraints={
 *     @ORM\UniqueConstraint(columns={"page_id", "valid_until"}, options={"where": "valid_until IS NULL"})
 * })
 */
```

the diff migration file generate every time :

```php
<?php

declare(strict_types=1);

namespace App\DoctrineMigrations;

use Doctrine\DBAL\Schema\Schema;
use Doctrine\Migrations\AbstractMigration;

/**
 * Auto-generated Migration: Please modify to your needs!
 */
final class Version20200204091939 extends AbstractMigration
{
    public function getDescription() : string
    {
        return '';
    }

    public function up(Schema $schema) : void
    {
        // this up() migration is auto-generated, please modify it to your needs
        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');

        $this->addSql('DROP INDEX UNIQ_615E7D21C4663E4FDEF8F78');
        $this->addSql('CREATE UNIQUE INDEX UNIQ_615E7D21C4663E4FDEF8F78 ON pages (page_id, valid_until) WHERE valid_until IS NULL');
    }

    public function down(Schema $schema) : void
    {
        // this down() migration is auto-generated, please modify it to your needs
        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'postgresql', 'Migration can only be executed safely on \'postgresql\'.');

        $this->addSql('CREATE SCHEMA public');
        $this->addSql('DROP INDEX uniq_615e7d21c4663e4fdef8f78');
        $this->addSql('CREATE UNIQUE INDEX uniq_615e7d21c4663e4fdef8f78 ON pages (page_id, valid_until) WHERE valid_until IS NULL');
    }
}
```

I figured out that the diff come from the behaviour of the `samePartialIndex` function in [Index.php](https://github.com/doctrine/dbal/blob/5fec6ead0354771562be82f868b3699ec8b4923a/lib/Doctrine/DBAL/Schema/Index.php#L285) file.

I add the code below in `samePartialIndex` function :

```php
[...]

PR #4339, opened 2020-10-13
Added filtered index support for SQL Server

cziegenberg:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #3038

#### Summary

Added support for filtered (or partial) indexes in SQL Server (supported since version 2008). This is the second version after the closed pull request #4240. It contains the implementation for SQL Server only. I will try to add the implementation for SQLite later, in a separate pull request.

The implementation has the following limitation (same as with other platforms like PostgreSQL, see #3871):
The given "where" option may be modified by the server. So when reading it again for comparison, the result may differ, which can result in an unnecessary update when creating a migration. This depends on how the "where" option is defined and especially occurrs in case of unique constraints, because for each indexed column in SQL Server a "IS NOT NULL" filter is added automatically by DBAL (for compatibility with other platforms and the SQL standard). This has to be solved globally (i.e. by saving the original value as a comment or adding a normalization step before the comparison) and is independent of this pull request.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SQLServerSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SQLServerSchemaManagerTest.php
@@ -6,6 +6,7 @@ use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Platforms\SQLServer2012Platform;
 use Doctrine\DBAL\Schema\Column;
 use Doctrine\DBAL\Schema\ColumnDiff;
+use Doctrine\DBAL\Schema\Comparator;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Schema\TableDiff;
 use Doctrine\DBAL\Types\Type;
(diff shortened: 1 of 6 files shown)


cziegenberg:
@morozov As requested, I created an updated pull request for the right branch and with additional tests. Would be great if you could review it once again. Thanks.

morozov:
@cziegenberg since `2.11.0` has been released, `2.11.x` only accepts bugfixes and the improvements for the upgrade path to 3.0. Please retarget against `3.0.x` (no need to file another PR, use the Edit button at the top of the page).

tests/Doctrine/Tests/DBAL/Functional/Schema/SQLServerSchemaManagerTest.php
```diff
@@ -404,4 +405,30 @@ public function testPkOrdering(): void
         self::assertEquals('colB', $columns[0]);
         self::assertEquals('colA', $columns[1]);
     }
+
+    public function testPartialIndexes(): void
+    {
+        if (! $this->schemaManager->getDatabasePlatform()->supportsPartialIndexes()) {
```
morozov, line 411: Would it be possible to move this test one level higher, to the `SchemaManagerFunctionalTestCase` and reuse for all platforms that support partial indexes?

Since it's a functional test, it should cover the API on all platforms that declare support for it, and we already have one in https://github.com/doctrine/dbal/blob/3f3551aadb20c7ed58ee9c7f06ed74770588272f/tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php#L355

cziegenberg: No, not at the moment, because of the limitation/bug #3871 - Postgres and SQL Server automatically add (different) brackets, type hints etc., so the returned where condition depends on the platform (for the same input).

cziegenberg:
Sorry, I seems like I did something wrong. I changed the target branch, but could not change the source for this pull request, although I switched the branch on my site...?! So the only thing I could do was to create a new pull request (#4342). So closing this one...

BTW: Good to see, that all the old SQL Server code has been removed now.

PR #4341, opened 2020-10-14
Add Statement::fetchAllIndexedAssociative() and ::iterateIndexedAssociative()

### LINKED ISSUE #4206: FetchMode::GROUP, FetchMode::UNIQUE are missing
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.7

#### Summary

My code spits this: `PHP Deprecated: Using a PDO fetch mode or their combination (196610 given) is deprecated and will cause an error in Doctrine DBAL 3.0`, so I'm trying to [upgrade](https://github.com/doctrine/dbal/blob/v2.7.0/UPGRADE.md#deprecation-direct-usage-of-the-pdo-apis-in-the-dbal-api)

This should be the part where my code is unfit:

`$stmt->fetchAll(\PDO::FETCH_GROUP|\PDO::FETCH_UNIQUE|\PDO::FETCH_ASSOC)`

As suggested in the upgrade.MD, I tried to replace it with this:

`$stmt->fetchAll(FetchMode::GROUP|FetchMode::UNIQUE|FetchMode::ASSOCIATIVE)`

But it doesn't work, because the `FetchMode` class doesn't define neither FetchMode::GROUP nor FetchMode::UNIQUE 

What am I missing here?

Thanks!

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

Closes #4206.

The fetch mode is analogous to the PDO's `FETCH_GROUP|FETCH_UNIQUE|FETCH_ASSOC`. Based on the `fetchIndexed()` name proposed in https://github.com/doctrine/dbal/issues/4206#issuecomment-706617488, I decided to name the method as `fetchAllIndexedAssociative()` for the following reasons:
1. It should have the "all" prefix for consistency with the rest of `fetchAll*()` methods.
2. It should have an "associative" suffix to be more specific and allow for the future addition of `fetchAllIndexedNumeric()` (although it's less likely to be requested).
3. It's more developer-friendly than `fetchAllAssociativeIndexed()` from the code completion standpoint since it shares fewer leading characters with `fetchAllAssociative()`.

### DIFF
--- a/docs/en/reference/data-retrieval-and-manipulation.rst
+++ b/docs/en/reference/data-retrieval-and-manipulation.rst
@@ -398,6 +398,26 @@ Execute the query and fetch the first two columns into an associative array as k
 .. note::
    All additional columns will be ignored and are only allowed to be selected by DBAL for its internal purposes.
 
+fetchAllAssociativeIndexed()
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Execute the query and fetch the data as an associative array where the key represents the first column and the value is
+an associative array of the rest of the columns and their values:
+
+.. code-block:: php
+
+    <?php
+    $users = $conn->fetchAllAssociativeIndexed('SELECT id, username, password FROM user');
+
+    /*
+    array(
+        1 => array(
+          'username' => 'jwage',
+          'password' => 'changeme',
+        )
+    )
+    */
+
 fetchNumeric()
 ~~~~~~~~~~~~~~
 
@@ -458,6 +478,19 @@ Execute the query and iterate over the first two columns as keys and values resp
 
 .. note::
    All additional columns will be ignored and are only allowed to be selected by DBAL for its internal purposes.
+
+iterateAssociativeIndexed()
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Execute the query and iterate over the result with the key representing the first column and the value being
+an associative array of the rest of the columns and their values:
+
+.. code-block:: php
+
+    <?php
+    foreach ($conn->iterateAssociativeIndexed('SELECT id, username, password FROM user') as $id => $data) {
+        // ...
+    }
 
 delete()
 ~~~~~~~~~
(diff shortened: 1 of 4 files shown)


docs/en/reference/data-retrieval-and-manipulation.rst
```diff
@@ -459,6 +479,19 @@ Execute the query and iterate over the first two columns as keys and values resp
 .. note::
    All additional columns will be ignored and are only allowed to be selected by DBAL for its internal purposes.
 
+fetchAllAssociativeIndexed()
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
greg0ire, line 483:
```suggestion
iterateAssociativeIndexed()
~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1013,6 +1014,32 @@ public function fetchAllKeyValue(string $query, array $params = [], array $types
...
+     * @throws Exception
+     */
+    public function fetchAllAssociativeIndexed(string $query, array $params = [], array $types = []): array
+    {
+        $stmt = $this->executeQuery($query, $params, $types);
+
+        $data = [];
+
+        foreach ($stmt->fetchAll(FetchMode::ASSOCIATIVE) as $row) {
+            $data[array_shift($row)] = $row;
```
greg0ire, line 1037: Is it possible to obtain a float from that call to `array_shift`? If yes, should we assert the result of the call is a valid array key? Bit weird that SA tools do not complain here… Using a float as a key is possible but has surprising results: https://3v4l.org/E1jGM

morozov: When floats are used as array keys, they are cast to an integer. I don't think DBAL should do much about that. It can be not only float but NULL. It's up to the developer who uses this API to decide what they are using as a key.

greg0ire (approved):

PR #4342, opened 2020-10-14
Added filtered index support for SQL Server

cziegenberg:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #3038

#### Summary

Added support for filtered (or partial) indexes in SQL Server (supported since version 2008).

The implementation has the following limitation (same as with other platforms like PostgreSQL, see #3871):
The given "where" option may be modified by the server. So when reading it again for comparison, the result may differ, which can result in an unnecessary update when creating a migration. This depends on how the "where" option is defined and especially occurrs in case of unique indexes (but not unique constraints), because for each indexed column in SQL Server a "IS NOT NULL" filter is added automatically by DBAL (for compatibility with other platforms and the SQL standard). This has to be solved globally (i.e. by saving the original value as a comment or adding a normalization step before the comparison) and is independent of this pull request.

### DIFF
--- a/src/Platforms/SQLServer2012Platform.php
+++ b/src/Platforms/SQLServer2012Platform.php
@@ -44,6 +44,15 @@ class SQLServer2012Platform extends AbstractPlatform
     /**
      * {@inheritdoc}
      */
+    public function supportsPartialIndexes()
+    {
+        // Partial indexes are supported as "filtered indexes" since version 2008.
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function getCurrentDateSQL()
     {
         return $this->getConvertExpression('date', 'GETDATE()');
@@ -498,13 +507,14 @@ SQL
      */
     private function _appendUniqueConstraintDefinition($sql, Index $index)
     {
-        $fields = [];
+        $concat = $index->hasOption('where') ? ' AND ' : ' WHERE ';
 
         foreach ($index->getQuotedColumns($this) as $field) {
-            $fields[] = $field . ' IS NOT NULL';
+            $sql   .= $concat . $field . ' IS NOT NULL';
+            $concat = ' AND ';
         }
 
-        return $sql . ' WHERE ' . implode(' AND ', $fields);
+        return $sql;
     }
 
     /**
@@ -1021,7 +1031,8 @@ SQL
                            WHEN '1' THEN 'clustered'
                            WHEN '2' THEN 'nonclustered'
                            ELSE NULL
-                       END AS flags
+                       END AS flags,
+                       idx.filter_definition AS [where]
                 FROM sys.tables AS tbl
                 JOIN sys.schemas AS scm ON tbl.schema_id = scm.schema_id
                 JOIN sys.indexes AS idx ON tbl.object_id = idx.object_id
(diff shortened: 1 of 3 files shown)


src/Platforms/SQLServer2012Platform.php
```diff
@@ -504,7 +513,17 @@ private function _appendUniqueConstraintDefinition($sql, Index $index)
             $fields[] = $field . ' IS NOT NULL';
         }
 
-        return $sql . ' WHERE ' . implode(' AND ', $fields);
+        if (count($fields) > 0) {
```
mvorisek, line 516: this, is much, much harder to read than before

tests/Functional/Schema/SQLServerSchemaManagerTest.php
```diff
@@ -404,4 +405,30 @@ public function testPkOrdering(): void
         self::assertEquals('colB', $columns[0]);
         self::assertEquals('colA', $columns[1]);
     }
+
+    public function testPartialIndexes(): void
+    {
+        if (! $this->schemaManager->getDatabasePlatform()->supportsPartialIndexes()) {
+            self::markTestSkipped('Platform does not support partial indexes.');
```
mvorisek, line 412: do we still support MSSQL before 2008?

cziegenberg: Yes, you're right. All supported SQL Server versions support it now. Will remove it.

src/Platforms/SQLServer2012Platform.php
cziegenberg: I will check this. At least the check for partial index support can be removed now.

tests/Functional/Schema/SQLServerSchemaManagerTest.php
cziegenberg: Fixed

src/Platforms/SQLServer2012Platform.php
cziegenberg: Fixed

cziegenberg:
I just saw that some checks have been added in the meantime, although all check have been passed before. Can you merge it without these additional checks or how do I trigger their execution?

mvorisek:
@cziegenberg please see https://github.com/doctrine/dbal/pull/4360 , I think we need to fix in a little more abstract way - please provide feedback, if you want to change anything or help with the tests, please send a PR directly to my branch. Michael

mvorisek:
> 
> 
> I just saw that some checks have been added in the meantime, although all check have been passed before. Can you merge it without these additional checks or how do I trigger their execution?

I think you need to rebase your PR on the latest commit of branch you target

cziegenberg:
Okay, all new checks passed too.

morozov:
Please see https://github.com/doctrine/dbal/pull/4339#discussion_r504321722.

cziegenberg:
> Please see [#4339 (comment)](https://github.com/doctrine/dbal/pull/4339#discussion_r504321722).

I replied to it and also mentioned this general problem in the summary. The problem is, that the output of the stored condition is platform dependent and in most cases doesn't match the input. For example brackets are added (PG), identifiers are modified (MS), types are added (PG), duplicate conditions are removed (MS) and - in case of SQL Server - some additional conditions are added automatically by DBAL for unique indexes. I don't see a chance to create a test that works in both platforms, without a general normalization step for the SQL returned by the platform and the user defined condition before comparing them (see #3871).

morozov:
> The problem is, that the output of the stored condition is platform dependent and in most cases doesn't match the input.

This is indeed a problem. Apart from the existing tests, this change requires a test that make sure that once a table with the newly implemented indexes is deployed and introspected, it doesn't cause false diffs.

devim-aryzhkov:
Guys, how are you? Bug is very actual

masacc:
Hello, this PR was closed without being merged. Has it been transferred to another PR for branch 3.1.x (I didn't find it) ?

morozov:
No, the PR got closed after I deleted the `3.0.x` branch. I cannot reopen it since the original branch is gone as well. The PR will have to be recreated.

PR #4343, opened 2020-10-14
Rename MySql... to MySQL... and PostgreSqlSchemaManager to PostgreSQLSchemaManager

mvorisek:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes (class names are case insensitive, but it the class will not load with autoload if referenced first with the old case)
| Fixed issues | todo from `MySqlPlatform` class

#### Summary

Renamed classes:
- `Doctrine\DBAL\Platforms\MySqlPlatform` -> `Doctrine\DBAL\Platforms\MySQLPlatform`
- `Doctrine\DBAL\Schema\MySqlSchemaManager` -> `Doctrine\DBAL\Schema\MySQLSchemaManager`
- `Doctrine\DBAL\Schema\PostgreSqlSchemaManager` -> `Doctrine\DBAL\Schema\PostgreSQLSchemaManager`
- test classes

PS: I am a huge fan of PascalCase defined with snake case, ie. if `mysql_platform` is the most sensual form, then the class should be named `MysqlPlatform`. But to keep things consistent, I only fixed one TODO from code and make `MySql` to be consistent with other names.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -187,7 +187,7 @@ The `Doctrine\DBAL\Driver::getName()` has been removed.
  * Removed `AbstractPlatform::getSQLResultCasing()`, `::prefersSequences()` and `::supportsForeignKeyOnUpdate()` methods.
  * Removed `PostgreSqlPlatform::getDisallowDatabaseConnectionsSQL()` and `::getCloseActiveDatabaseConnectionsSQL()` methods.
  * Removed `MysqlSessionInit` listener.
- * Removed `MysqlPlatform::getCollationFieldDeclaration()`.
+ * Removed `MySQLPlatform::getCollationFieldDeclaration()`.
  * Removed `AbstractPlatform::getIdentityColumnNullInsertSQL()`.
  * Removed `AbstractPlatform::fixSchemaElementName()`.
  * Removed `Table::addUnnamedForeignKeyConstraint()` and `Table::addNamedForeignKeyConstraint()`.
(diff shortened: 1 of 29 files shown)


mvorisek:
@morozov @greg0ire

-there are also other things to rename (`MySql(?!Plat)`).

- ok to target `2.11.x`?

greg0ire:
It's ok with me 👍

mvorisek:
PR done

greg0ire:
> yes (class names are case insensitive, but it the class will not load with autoload if referenced first with the old case)

This means you should target 3.0.x

mvorisek:
> 
> 
> > yes (class names are case insensitive, but it the class will not load with autoload if referenced first with the old case)
> 
> This means you should target 3.0.x

are the renamed classes even intended to be instantiated directly - if there are ment to be used by `Driver::getDatabasePlatform()`, then there is no BC break

morozov:
> @morozov @greg0ire is this welcomed and ok to target `2.11.x`?

Since `2.11.0` has been released, `2.11.x` only accepts bug fixes and the improvements to the upgrade path to `3.0`. Please retarget against `3.0.x`.

morozov (approved):

greg0ire:
Thanks @mvorisek !

morozov:
While this change is welcomed, it made me hate macOS even more. Before it propagates through all supported branches, it makes it hard to switch between them due to the case insensitivity of APFS. Merging the branch with one case to the branch with another is even more challenging.

PR #4347, opened 2020-10-15
[2.12] PHP 8 compatibility

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | kinda
| Fixed issues | N/A

#### Summary

This PR makes DBAL 2.12 compatible with PHP 8.

The main problem with PHP 8 is that certain PDO classes have changed their method signatures in PHP 8. Since DBAL extends those classes, those changed signatures have to be applied to those child classes as well. That is of course a BC break if the calling code itself extends DBAL's classes.

With this PR, I suggest to move the affected methods into traits that exist in a php 7 and a php 8 version. This way, the BC break introduced with php 8 is handed down if and only if we're actually on php 8.

I am aware that DBAL 3.0 is supposed to become the php 8 compatible version. So why am I so eager to fix the 2.x series?

I'm really looking forward to 3.0 because it moves away from the tight coupling to PDO which makes perfect sense – especially if we look at the problem this PR aims to fix. But DBAL 3 has no stable release yet and ORM has not been migrated to DBAL 3 yet. PHP 8 on the other hand has just reached RC2 and plans the first stable release in six weeks. I really doubt that the ecosystem around Doctrine will be ready for php 8 if we make DBAL 3 a precondition for php 8.

Finally, I'd really like to test the applications I am working on with php 8 and this small change I'm proposing here unblocks me for the moment.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -23,6 +23,7 @@ jobs:
         php-version:
           - "7.3"
           - "7.4"
+          - "8.0"
         deps:
           - "fixed"
         include:
@@ -183,6 +184,9 @@ jobs:
           - "11"
           - "12"
           - "13"
+        include:
+          - php-version: "8.0"
+            postgres-version: "13"
 
     services:
       postgres:
@@ -247,6 +251,13 @@ jobs:
         extension:
           - "mysqli"
           - "pdo_mysql"
+        include:
+          - php-version: "8.0"
+            mariadb-version: "10.5"
+            extension: "mysqli"
+          - php-version: "8.0"
+            mariadb-version: "10.5"
+            extension: "pdo_mysql"
 
     services:
       mariadb:
@@ -304,6 +315,7 @@ jobs:
       matrix:
         php-version:
           - "7.4"
+          - "8.0"
         mysql-version:
           - "5.7"
           - "8.0"
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         {"name": "Jonathan Wage", "email": "jonwage@gmail.com"}
     ],
     "require": {
-        "php": "^7.3",
+        "php": "^7.3 || ^8",
         "ext-pdo": "*",
         "doctrine/cache": "^1.0",
         "doctrine/event-manager": "^1.0"
--- a/lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php
+++ b/lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Doctrine\DBAL\Driver;
+
+use PDOStatement;
+
+use function func_get_args;
+
+use const PHP_VERSION_ID;
+
+if (PHP_VERSION_ID >= 80000) {
+    /**
+     * @internal
+     */
+    trait PDOQueryImplementation
+    {
+        /**
+         * @return PDOStatement
+         */
+        public function query(?string $query = null, ?int $fetchMode = null, mixed ...$fetchModeArgs)
+        {
+            return $this->doQuery($query, $fetchMode, ...$fetchModeArgs);
+        }
+    }
+} else {
+    /**
+     * @internal
+     */
+    trait PDOQueryImplementation
+    {
+        /**
+         * @return PDOStatement
+         */
+        public function query()
+        {
+            return $this->doQuery(...func_get_args());
+        }
+    }
+}
--- a/lib/Doctrine/DBAL/Driver/PDOStatement.php
+++ b/lib/Doctrine/DBAL/Driver/PDOStatement.php
@@ -26,6 +26,8 @@ use const E_USER_DEPRECATED;
  */
 class PDOStatement extends \PDOStatement implements StatementInterface, Result
 {
+    use PDOStatementImplementations;
+
     private const PARAM_TYPE_MAP = [
         ParameterType::NULL         => PDO::PARAM_NULL,
         ParameterType::INTEGER      => PDO::PARAM_INT,
@@ -52,34 +54,6 @@ class PDOStatement extends \PDOStatement implements StatementInterface, Result
      */
     protected function __construct()
     {
-    }
-
-    /**
-     * {@inheritdoc}
-     *
-     * @deprecated Use one of the fetch- or iterate-related methods.
-     */
-    public function setFetchMode($fetchMode, $arg2 = null, $arg3 = null)
-    {
-        $fetchMode = $this->convertFetchMode($fetchMode);
-
-        // This thin wrapper is necessary to shield against the weird signature
-        // of PDOStatement::setFetchMode(): even if the second and third
-        // parameters are optional, PHP will not let us remove it from this
-        // declaration.
-        try {
-            if ($arg2 === null && $arg3 === null) {
-                return parent::setFetchMode($fetchMode);
-            }
-
-            if ($arg3 === null) {
-                return parent::setFetchMode($fetchMode, $arg2);
-            }
-
-            return parent::setFetchMode($fetchMode, $arg2, $arg3);
-        } catch (PDOException $exception) {
-            throw Exception::new($exception);
-        }
     }
 
     /**
@@ -167,39 +141,6 @@ class PDOStatement extends \PDOStatement implements StatementInterface, Result
     /**
      * {@inheritdoc}
      *
-     * @deprecated Use fetchAllNumeric(), fetchAllAssociative() or fetchFirstColumn() instead.
-     */
-    public function fetchAll($fetchMode = null, $fetchArgument = null, $ctorArgs = null)
-    {
-        $args = func_get_args();
-
-        if (isset($args[0])) {
-            $args[0] = $this->convertFetchMode($args[0]);
-        }
-
-        if ($fetchMode === null && $fetchArgument === null && $ctorArgs === null) {
-            $args = [];
-        } elseif ($fetchArgument === null && $ctorArgs === null) {
-            $args = [$fetchMode];
-        } elseif ($ctorArgs === null) {
-            $args = [$fetchMode, $fetchArgument];
-        } else {
-            $args = [$fetchMode, $fetchArgument, $ctorArgs];
-        }
-
-        try {
-            $data = parent::fetchAll(...$args);
-            assert(is_array($data));
-
-            return $data;
-        } catch (PDOException $exception) {
-            throw Exception::new($exception);
-        }
-    }
-
-    /**
-     * {@inheritdoc}
-     *
      * @deprecated Use fetchOne() instead.
      */
     public function fetchColumn($columnIndex = 0)
@@ -262,6 +203,66 @@ class PDOStatement extends \PDOStatement implements StatementInterface, Result
     public function free(): void
     {
         parent::closeCursor();
+    }
+
+    /**
+     * @param mixed ...$args
+     */
+    private function doSetFetchMode(int $fetchMode, ...$args): bool
+    {
+        $fetchMode = $this->convertFetchMode($fetchMode);
+
+        // This thin wrapper is necessary to shield against the weird signature
+        // of PDOStatement::setFetchMode(): even if the second and third
+        // parameters are optional, PHP will not let us remove it from this
+        // declaration.
+        $slice = [];
+
+        foreach ($args as $arg) {
+            if ($arg === null) {
+                break;
+            }
+
+            $slice[] = $arg;
+        }
+
+        try {
+            return parent::setFetchMode($fetchMode, ...$slice);
+        } catch (PDOException $exception) {
+            throw Exception::new($exception);
+        }
+    }
+
+    /**
+     * @param mixed ...$args
+     *
+     * @return mixed[]
+     */
+    private function doFetchAll(...$args): array
+    {
+        if (isset($args[0])) {
+            $args[0] = $this->convertFetchMode($args[0]);
+        }
+
+        $slice = [];
+
+        foreach ($args as $arg) {
+            if ($arg === null) {
+                break;
+            }
+
+            $slice[] = $arg;
+        }
+
+        try {
+            $data = parent::fetchAll(...$slice);
+        } catch (PDOException $exception) {
+            throw Exception::new($exception);
+        }
+
+        assert(is_array($data));
+
+        return $data;
     }
 
     /**
--- a/lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php
+++ b/lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php
@@ -1,0 +1,73 @@
+<?php
+
+namespace Doctrine\DBAL\Driver;
+
+use function func_get_args;
+
+use const PHP_VERSION_ID;
+
+if (PHP_VERSION_ID >= 80000) {
+    /**
+     * @internal
+     */
+    trait PDOStatementImplementations
+    {
+        /**
+         * @deprecated Use one of the fetch- or iterate-related methods.
+         *
+         * @param int   $mode
+         * @param mixed ...$args
+         *
+         * @return bool
+         */
+        public function setFetchMode($mode, ...$args)
+        {
+            return $this->doSetFetchMode($mode, ...$args);
+        }
+
+        /**
+         * @deprecated Use fetchAllNumeric(), fetchAllAssociative() or fetchFirstColumn() instead.
+         *
+         * @param int|null $mode
+         * @param mixed    ...$args
+         *
+         * @return mixed[]
+         */
+        public function fetchAll($mode = null, ...$args)
+        {
+            return $this->doFetchAll($mode, ...$args);
+        }
+    }
+} else {
+    /**
+     * @internal
+     */
+    trait PDOStatementImplementations
+    {
+        /**
+         * @deprecated Use one of the fetch- or iterate-related methods.
+         *
+         * @param int   $fetchMode
+         * @param mixed $arg2
+         * @param mixed $arg3
+         */
+        public function setFetchMode($fetchMode, $arg2 = null, $arg3 = null): bool
+        {
+            return $this->doSetFetchMode(...func_get_args());
+        }
+
+        /**
+         * @deprecated Use fetchAllNumeric(), fetchAllAssociative() or fetchFirstColumn() instead.
+         *
+         * @param int|null $fetchMode
+         * @param mixed    $fetchArgument
+         * @param mixed    $ctorArgs
+         *
+         * @return mixed[]
+         */
+        public function fetchAll($fetchMode = null, $fetchArgument = null, $ctorArgs = null)
+        {
+            return $this->doFetchAll(...func_get_args());
+        }
+    }
+}
--- a/lib/Doctrine/DBAL/Statement.php
+++ b/lib/Doctrine/DBAL/Statement.php
@@ -10,6 +10,7 @@ use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
 use IteratorAggregate;
 use PDO;
+use PDOStatement;
 use Throwable;
 use Traversable;
 
@@ -134,6 +135,10 @@ class Statement implements IteratorAggregate, DriverStatement, Result
     {
         $this->params[$param] = $variable;
         $this->types[$param]  = $type;
+
+        if ($this->stmt instanceof PDOStatement) {
+            $length = $length ?? 0;
+        }
 
         return $this->stmt->bindParam($param, $variable, $type, $length);
     }
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -22,6 +22,7 @@
         <exclude name="SlevomatCodingStandard.Classes.SuperfluousExceptionNaming"/>
         <exclude name="SlevomatCodingStandard.Classes.DisallowLateStaticBindingForConstants.DisallowedLateStaticBindingForConstant"/>
         <exclude name="SlevomatCodingStandard.ControlStructures.ControlStructureSpacing.IncorrectLinesCountAfterLastControlStructure"/>
+        <exclude name="SlevomatCodingStandard.ControlStructures.RequireNullCoalesceEqualOperator.RequiredNullCoalesceEqualOperator"/>
 
         <exclude name="PSR2.Methods.MethodDeclaration.Underscore"/>
         <!-- https://github.com/slevomat/coding-standard/issues/867 -->
@@ -29,6 +30,11 @@
         <!-- See https://github.com/squizlabs/PHP_CodeSniffer/issues/2937 -->
         <exclude name="Squiz.Arrays.ArrayDeclaration.ValueNoNewline"/>
         <exclude name="Squiz.NamingConventions.ValidVariableName.PublicHasUnderscore"/>
+    </rule>
+
+    <rule ref="Generic.Classes.DuplicateClassName.Found">
+        <exclude-pattern>*/lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php</exclude-pattern>
+        <exclude-pattern>*/lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php</exclude-pattern>
     </rule>
 
     <!-- Disable the rules that will require PHP 7.4 -->
@@ -57,6 +63,8 @@
     </rule>
 
     <rule ref="PSR1.Classes.ClassDeclaration.MultipleClasses">
+        <exclude-pattern>*/lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php</exclude-pattern>
+        <exclude-pattern>*/lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php</exclude-pattern>
         <exclude-pattern>*/tests/*</exclude-pattern>
     </rule>
 
@@ -113,6 +121,11 @@
     <!-- The override opts out the caller from the strict mode for backward compatibility -->
     <rule ref="Generic.CodeAnalysis.UselessOverridingMethod.Found">
         <exclude-pattern>lib/Doctrine/DBAL/Driver/PDOConnection.php</exclude-pattern>
+    </rule>
+
+    <rule ref="SlevomatCodingStandard.Classes.UnusedPrivateElements.UnusedMethod">
+        <exclude-pattern>*/lib/Doctrine/DBAL/Driver/PDOConnection.php</exclude-pattern>
+        <exclude-pattern>*/lib/Doctrine/DBAL/Driver/PDOStatement.php</exclude-pattern>
     </rule>
 
     <!-- See https://github.com/slevomat/coding-standard/issues/770 -->
--- a/psalm.xml
+++ b/psalm.xml
@@ -32,6 +32,13 @@
                 <file name="lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php"/>
             </errorLevel>
         </ConflictingReferenceConstraint>
+        <DuplicateClass>
+            <errorLevel type="suppress">
+                <!-- These files contain a php 7 and a php 8 version of the same trait -->
+                <file name="lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php"/>
+                <file name="lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php"/>
+            </errorLevel>
+        </DuplicateClass>
         <FalsableReturnStatement>
             <errorLevel type="suppress">
                 <!--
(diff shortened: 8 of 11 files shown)


greg0ire (commented):
Nice! Please add some CI jobs to make sure the PHP 8 code is covered

derrabus:
Do you already have CI jobs for php 8 for higher branches that I could backport?

greg0ire:
> Do you already have CI jobs for php 8 for higher branches that I could backport?

We have some, but for Travis, and as you may know I just merged #4310 . I still need to merge it into 3.0.x. Until then, no GA jobs to backport, sorry!

beberlei (changes requested):
This is a good idea, i was under the impression its impossible, tried my share of things but didn't come up with this workaroud :) 

It would reduce a lot of the pressure to have PHP 8 support in DBAL 2.x already, we started on work for DBAL 3 in ORM 2.8, but its probably a long stretch to finish that in 6 weeks, including all the other parts of the ecosystem (fixtures, migrations, DoctrineBundle, Laminas module etc...)

As for the phpcs violations, a few have to be ignored here to disable our strict standard.

lib/Doctrine/DBAL/Driver/PDOConnectionTrait.php
```diff
@@ -0,0 +1,62 @@
...
+if (PHP_VERSION_ID >= 80000) {
+    /**
+     * @internal
+     */
+    trait PDOConnectionTrait
+    {
+        /**
+         * {@inheritdoc}
+         *
+         * @return PDOStatement
```
greg0ire, line 23: https://github.com/doctrine/dbal/blob/29bf3156bc0474b7ee61d354892e4f2de8ca83c2/psalm.xml#L70

lib/Doctrine/DBAL/Driver/PDOStatementTrait.php
```diff
@@ -0,0 +1,139 @@
...
+use function func_get_args;
+use function is_array;
+
+use const PHP_VERSION_ID;
+
+if (PHP_VERSION_ID >= 80000) {
+    /**
+     * @internal
+     */
+    trait PDOStatementTrait
```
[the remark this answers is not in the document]
derrabus: I can also split this into two traits if naming gets too hard. 😉

phpcs.xml.dist
```diff
@@ -22,6 +22,7 @@
         <exclude name="SlevomatCodingStandard.Classes.SuperfluousExceptionNaming"/>
         <exclude name="SlevomatCodingStandard.Classes.DisallowLateStaticBindingForConstants.DisallowedLateStaticBindingForConstant"/>
         <exclude name="SlevomatCodingStandard.ControlStructures.ControlStructureSpacing.IncorrectLinesCountAfterLastControlStructure"/>
+        <exclude name="SlevomatCodingStandard.ControlStructures.RequireNullCoalesceEqualOperator.RequiredNullCoalesceEqualOperator"/>
```
derrabus, line 25: 💡 The `??=` operator does not exist in php 7.3.

derrabus:
PHP CodeSniffer is happy now, but I need some help with silencing Psalm. I'm unfamiliar with the configuration of that tool. 😓

derrabus:
CI jobs on php 8 with SQLite, MySQL and Postgres were successful. I've tried to enable DB2 and MSSQL as well, but apparently the corresponding php extensions are not ready yet.

composer.json
```diff
@@ -32,7 +32,7 @@
         {"name": "Jonathan Wage", "email": "jonwage@gmail.com"}
     ],
     "require": {
-        "php": "^7.3",
+        "php": "^7.3 || ^8",
```
mvorisek, line 35:
```suggestion
        "php": "^7.3 || (^8 && <8.1)",
```

lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php
```diff
@@ -0,0 +1,63 @@
...
+            } catch (PDOException $exception) {
+                throw Exception::new($exception);
+            }
+        }
+    }
+} else {
+    /**
+     * @internal
+     */
+    // @codingStandardsIgnoreLine PSR1.Classes.ClassDeclaration.MultipleClasses Generic.Classes.DuplicateClassName.Found
```
morozov, line 41: Please move it to the configuration file.

lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php
```diff
@@ -0,0 +1,140 @@
...
+        /**
+         * {@inheritdoc}
+         *
+         * @deprecated Use one of the fetch- or iterate-related methods.
+         */
+        public function setFetchMode($mode, ...$args)
+        {
+            $mode = $this->convertFetchMode($mode);
+
+            $filteredArgs = array_filter($args, static function (mixed $value): bool {
```
morozov, line 30: Does the `mixed` type have to be used here? As far as I understand, it still gets parsed on PHP 7 but is interpreted as a class named `mixed` which is not what we want even if this code is not executed.

morozov: Does this implementation have to be different from the one for PHP 7? If three arguments are passed, and the first two are NULL while the third is not, the value of the third argument will be passed as the first one.

lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php
```diff
@@ -0,0 +1,140 @@
...
+            } catch (PDOException $exception) {
+                throw Exception::new($exception);
+            }
+        }
+    }
+} else {
+    /**
+     * @internal
+     */
+    // @codingStandardsIgnoreLine PSR1.Classes.ClassDeclaration.MultipleClasses Generic.Classes.DuplicateClassName.Found
```
morozov, line 76: Please move it to the configuration file.

morozov (commented):
Is there a need to have an individual implementation for each version? As far as I understand, the underlying logic doesn't need to change, only the method signatures do. In this case, it should be possible to have the logic implemented in one trait and extended by two others which defined the expected signature.

.github/workflows/continuous-integration.yml
```diff
@@ -183,6 +184,9 @@ jobs:
           - "11"
           - "12"
           - "13"
+        include:
+          - php-version: "8.0"
+            postgres-version: "12"
```
morozov, line 189: Why not the latest version?

morozov:
I see that PHP 8 has been added to some build jobs but not to all of them (`oci8`, `pdo_oci`, `sqlsrv`, `pdo_sqlsrv`, MariaDB are missing it). Is there a plan to update those? Testing with SQL Server will require bumping the extension version to `5.9.0-preview1`.

lib/Doctrine/DBAL/Statement.php
```diff
@@ -135,6 +136,10 @@ public function bindParam($param, &$variable, $type = ParameterType::STRING, $le
         $this->params[$param] = $variable;
         $this->types[$param]  = $type;
 
+        if ($this->stmt instanceof PDOStatement) {
+            $length = $length ?? 0;
```
morozov, line 140: This looks like a BC break regardless of the version.

composer.json
morozov: @mvorisek please make sure you provide an explanation if you're suggesting a change.

morozov:
I plan to release `2.11.2` tomorrow since it contains a couple of BC fixes. I anticipate resolving this PR will take a longer than that. Let me know if you think I should hold off the release.

On the other hand, this change itself is significant enough and not a bug fix that it may deserve another unplanned minor release.

greg0ire:
> PHP CodeSniffer is happy now, but I need some help with silencing Psalm. I'm unfamiliar with the configuration of that tool. sweat

I pushed a commit showing how to ignore one of the errors. You can run `vendor/bin/psalm` to find out the type of each error (for some reason Github Actions does not show them)

lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php
derrabus: > Does the `mixed` type have to be used here?
No. It's pure eye candy because we're in a code branch that will only be executed in PHP 8. I can remove it if you don't like it. 😃 

> Does this implementation have to be different from the one for PHP 7?

We have different parameters, so the implementation is different. If the current logic is too confusing, I'm open to suggestions.

> If three arguments are passed, and the first two are NULL while the third is not, the value of the third argument will be passed as the first one.

No, what I'm checking here is if there are any additional non-null parameters passed. If that is the case, I'm passing the additional arguments as they are, otherwise I'm passing no additional arguments at all.

> UPD: I see that the original `$args` gets eventually passed down the line, so the intent is is even less clear.

The method signature might have changed, but PDO is still picky about passing unwanted `null` arguments.

morozov: > We have different parameters, so the implementation is different.

The only change in PHP 8 I'm aware of is the signature change of `PDOStatement::query()` (https://github.com/doctrine/dbal/pull/), there must have been other signature changes (e.g. https://github.com/php/php-src/commit/4a5e17f9afedb6bd750ae8a89a0cf705f738c7e4). Please document all that changes that this patch attempts to adress.

.github/workflows/continuous-integration.yml
derrabus: My intention was to not slow the CI down by running tests for all Postgres-Versions with php 8 again. But I don't have strong opinions on which Postgres version to choose here. We can change this.

derrabus:
> I see that PHP 8 has been added to some build jobs but not to all of them

As I wrote earlier, I tried to enable the SQL Server tests, but the extension did not compile properly. Same for DB2.

MariaDB shouldn't be a problem since the MySQL tests are working already. I've giving it a try. I'll look into Oracle as well.

> Is there a plan to update those?

I wouldn't make this a blocker for this PR. As soon as the extensions are ready, they can be added to the CI.

lib/Doctrine/DBAL/Statement.php
derrabus: It isn't. The `$length` parameter has to be an `int`. If you pass `null`, it is silently cast to zero on php 7. On php 8 you'll get a type error. If you prefer, I can additionally check if we're on php 8, but I don't think that this is necessary.

greg0ire:
> I wouldn't make this a blocker for this PR.

Agreed, let's not make it a blocker, your code is already covered by the jobs you added.

derrabus:
Sutting up the OCI8 extension did not work either. I don't know if this is actually a problem of the GitHub action that is used to setup PECL extensions or if the extensions themselves don't compile yet. But that's something that could be done in a follow-up PR.

But the good news is: MariaDB is tested with php 8 now. :-)

derrabus:
> On the other hand, this change itself is significant enough and not a bug fix that it may deserve another unplanned minor release.

I am unfamiliar with your versioning policy. Personally, I think releasing this as a 2.11 bugfix should be fine. But if you want to bump to 2.12 because of this change, I'm also cool with it.

lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php
derrabus: The methods that changed their signature in an incompatible way are
* `PDO::query`
* `PDOStatement::setFetchMode()`
* `PDOStatement::fetchAll()`

Apart from that, a lot of functions and methods provided by php have been upgraded with scalar type declarations. This change also affects PDO and is basically why I had to fix that one call to `bindParam()`.

greg0ire:
I'd say let's move this to 2.12.0 because otherwise if there is anything wrong with this PR, the nasty people who report the bugs are going to make sure to point out this should have not have shipped in a bugfix release. Let's not delay 2.12.0 at all though, and release it right after 2.11.2, shall we?

derrabus:
Sounds like a plan.

composer.json
mvorisek: Add explicit support for php 8.0.x only, we should not declare like 8.1 support if not tested against this version (with which this core package might be incompatible at the current state)

derrabus: We haven't tested against php 7.5 either, yet we declare compatibility. I think we're safe here.

derrabus:
I have updated UPGRADE.md with the signature changed.

lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php
derrabus: I have changed `array_filter` call, hoping that the new implementation makes the intent of that code block clearer.

greg0ire (dismissed):
@morozov if you agree with the plan please create branch 2.12.x and retarget :pray:

lib/Doctrine/DBAL/Statement.php
morozov: There can be a class that extends the PDO statement which this code will call and it may expect NULL.

morozov:
> @morozov if you agree with the plan please create branch 2.12.x and retarget 🙏

Done.

lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php
```diff
@@ -0,0 +1,147 @@
...
+     * @internal
+     */
+    trait PDOStatementImplementations
+    {
+        /**
+         * {@inheritdoc}
+         *
+         * @deprecated Use one of the fetch- or iterate-related methods.
+         */
+        public function setFetchMode($fetchMode, $arg2 = null, $arg3 = null)
```
dbrumann, line 91: I was a bit irritated by the argument names, but looking at the method signature for `PDOStatement::setFetchMode()` I get why it is what it is 🙈

dbrumann (dismissed):

lib/Doctrine/DBAL/Statement.php
derrabus: That class would be violating the contract of `PDOStatement` if `null` led to a different (non-exceptional) behavior than zero. But yes, it's possible. Do you have a suggestion how to move forward here?

morozov: > The `$length` parameter has to be an `int`.

You're right. I was under the impression that it became non-nullable in PHP 8. No additional changes needed.

lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php
```diff
@@ -0,0 +1,62 @@
...
+     * @internal
+     */
+    trait PDOQueryImplementation
+    {
+        /**
+         * {@inheritdoc}
+         *
+         * @return PDOStatement
+         */
+        public function query(?string $query = null, ?int $fetchMode = null, mixed ...$fetchModeArgs)
```
morozov, line 25: Are the first two arguments nullable? They aren't [documented](https://www.php.net/manual/en/pdo.query.php) as such.

lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php
```diff
@@ -0,0 +1,147 @@
...
+     * @internal
+     */
+    trait PDOStatementImplementations
+    {
+        /**
+         * {@inheritdoc}
+         *
+         * @deprecated Use one of the fetch- or iterate-related methods.
+         */
+        public function setFetchMode($mode, ...$args)
```
morozov, line 26: Should it be `int $mode` as [documented](https://www.php.net/manual/en/pdostatement.setfetchmode.php)?

lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php
morozov: Please do not resolve a conversation started by someone else unless you're resolving it on their behalf.

morozov: > The methods that changed their signature in an incompatible way [...] Apart from that, a lot of functions and methods provided by php have been upgraded with scalar type declarations.

It still doesn't answer why we need different implementations in order to satisfy just the difference in method signatures. I mean specifically handling `$args` in `setFetchMode()` but it may be more.

lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php
derrabus: Not in PDO. But I need to fulfill `Doctrine\DBAL\Driver\Connection::query()` as well, so I cannot add mandatory parameters.

lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php
derrabus: Same problem, I needed to find a signature that is compatible to the parent PDO class as well as the implemented DBAL interface.

If we wanted `int $mode`, I would need to add the `int` type declaration to `Doctrine\DBAL\Driver\ResultStatement::setFetchMode()` as well.

lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php
derrabus: The php 7 version of each trait is just C&P of the current implementation.

lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php
derrabus: The main difference between the two implementations is that the php 7 version of the signature allows up to two additional arguments while the php 8 signature has a variadic part. Both have in common that PDO will throw an error if additional arguments are passed although the selected fetch mode does not accept any.

The way that DBAL currently solves that problem is to only pass on the additional arguments is they aren't null. My attempt to preserve that logic was to search the variadic part for non-null values. If it doesn't contain any, I throw the variadic part away. If you know a better way, I'm open to suggestions.

lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php
derrabus: This is by the way something we can safely do. I was just reluctant to change the existing interfaces. I'll let you decide which way to go.

lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php
morozov: Makes sense.

lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php
morozov: It sounds safe but since it's not required, I'm fine with the current state. Thanks for the explanation.

lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php
morozov: This is the approach I had in mind (the pseudo-code):
```php
if (PHP_VERSION_ID >= 80000) {
    trait PDOStatementInterface
    {
        use PDOStatementImplementation;

        public function fetchAll($mode = null, ...$args)
        {
            return $this->doFetchAll($mode, ...$args);
        }
    }
} else {
    trait PDOStatementInterface
    {
        use PDOStatementImplementation;

        public function fetchAll($fetchMode = null, $fetchArgument = null, $ctorArgs = null)
        {
            $args = func_get_args();
            array_shift($args);

            $this->doFetchAll($fetchMode, ...$args);
        }
    }
}

trait PDOStatementImplementation
{
    private function doFetchAll(...$args)
    {
        // the actual implementation
    }
}
```

derrabus: Yes, we can do this, I suppose. But I wouldn't create a separate trait for `doFetchAll()` but declare that method inside the actual `PDOStatement` class. I'll give it a try.

derrabus: I've pushed a commit. Please have a look.

morozov (commented):
> […] I wouldn't create a separate trait for `doFetchAll()` but declare that method inside the actual `PDOStatement` class.

True. It's not necessary. Doesn't the result look better? It does to me.

psalm.xml
```diff
@@ -140,6 +156,12 @@
                 <file name="lib/Doctrine/DBAL/Tools/Dumper.php"/>
             </errorLevel>
         </UndefinedClass>
+        <UndefinedVariable>
+            <errorLevel type="suppress">
+                <file name="lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php"/>
+                <file name="lib/Doctrine/DBAL/Driver/PDOStatementImplementations.php"/>
+            </errorLevel>
+        </UndefinedVariable>
```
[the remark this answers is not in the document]
derrabus: I don't need it anymore.

lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php
```diff
@@ -0,0 +1,43 @@
...
+use const PHP_VERSION_ID;
+
+if (PHP_VERSION_ID >= 80000) {
+    /**
+     * @internal
+     */
+    trait PDOQueryImplementation
+    {
+        /**
+         * {@inheritdoc}
```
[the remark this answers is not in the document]
derrabus: The method inside this trait is used to override a method from `PDO`. That's where it inherits documentation from.

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -264,6 +203,55 @@ public function free(): void
...
+
+        $hasFetchModeArgs = array_reduce(
+            $args,
+            static function (bool $hasFetchModeArgs, $value): bool {
+                return $hasFetchModeArgs || $value !== null;
+            },
+            false
+        );
+        if (! $hasFetchModeArgs) {
+            $args = [];
```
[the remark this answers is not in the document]
derrabus: I'll try to find an implementation closer to the current one, hang on.

lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php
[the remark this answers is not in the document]
morozov: It doesn't seem necessary. `{@inheritdoc}` is used to inherit the description (which we can't source from PDO) or suppress the coding standard errors about missing parameter documentation. Neither seems to be the case.

What do we lose if this is removed?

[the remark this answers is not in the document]
derrabus: If I remove it, Psalm complains about the `mixed` type declaration. 🙈

lib/Doctrine/DBAL/Driver/PDOStatement.php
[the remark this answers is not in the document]
derrabus: Have another look, please.

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -264,6 +206,71 @@ public function free(): void
...
+     * @param mixed $ctorArgs
+     * @param mixed ...$additionalArgs
+     *
+     * @return mixed[]
+     */
+    private function doFetchAll(
+        ?int $fetchMode = null,
+        $fetchArgument = null,
+        $ctorArgs = null,
+        ...$additionalArgs
```
morozov, line 248: What are these additional arguments?

derrabus: They are part of the new `setFetchMode()`/`fetchAll()` signatures. I don't know if they're used inside PDO for anything yet. I'm just passing them down if they're there. Dropping them felt wrong here.

derrabus:
Thanks for the review so far, @morozov. The traits are smaller now and probably better to maintain that way. I'm heading to bed now, but I can continue to work on the PR tomorrow.

lib/Doctrine/DBAL/Driver/PDOStatement.php
morozov: > I don't know if they're used inside PDO for anything yet.

I don't think they do but if they add a 4th argument (which is extremely unlikely), then it may behave similar to the previous ones and would have to be filtered out. Or at least we cannot guarantee the opposite. With that said, I see to approaches:
1. Just handle the 3 arguments as they are handled now.
2. Handle all arguments in a generic way:
    ```php
    function fetch(...$args) {
        if (isset($args[0])) {
            $args[0] = convertMode($args[0]);
        }

        $slice = [];

        foreach ($args as $arg) {
            if ($arg === null) {
                break;
            }

            $slice[] = $arg.
        }

        parent::fetch(...$slice);
    }
    ````

lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php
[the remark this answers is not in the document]
morozov: Are you sure? I don't see any errors from PHP_CodeSniffer or Psalm if I remove this annotation.

lib/Doctrine/DBAL/Driver/PDOStatement.php
derrabus: > I don't think they do but if they add a 4th argument (which is extremely unlikely), then it may behave similar to the previous ones and would have to be filtered out.

Filtering the currently existing parameters is only necessary for BC reasons: Doctrine's PDOStatement allows to pass `null` values where PDO's own implementation would throw errors. This does not apply to new parameters that we're about to introduce.

> Handle all arguments in a generic way:

The code you suggested breaks quite a lot of tests. I'd rather not change the logic here.

> Just handle the 3 arguments as they are handled now.

If you don't like that variadic part, we can of course just ignore it. My motivation was to stay as close to the upstream interface as possible.

lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php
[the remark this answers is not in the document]
derrabus: When I remove all `{@inheritdoc}` annotations from the traits, this is what I get.

Bildschirmfoto 2020-10-19 um 16 49 00

Please, let's just keep them.

lib/Doctrine/DBAL/Driver/PDOStatement.php
derrabus: My bad, your implementation works now. I'll change it.

derrabus:
I've rebased the PR to fix the merge conflicts. Build is green and I think all comments have been addressed. I'd say, the PR is ready.

lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php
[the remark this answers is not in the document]
morozov: https://psalm.dev/docs/running_psalm/issues/ReservedWord/ - Psalm treats the `mixed` type from PHP 8 as a class name in PHP 7. You can either remove it or fix/suppress at the Psalm level.

There is no point in having an `{@inheritdoc}` annotation if there's nothing to inherit. Even more to suppress a static analysis issue.

psalm.xml
```diff
@@ -131,6 +151,9 @@
                 <!-- See https://github.com/doctrine/dbal/pull/3562 -->
                 <file name="lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php"/>
                 <file name="lib/Doctrine/DBAL/Schema/SqliteSchemaManager.php"/>
+
+                <!-- See https://github.com/doctrine/dbal/pull/4347 -->
```
morozov, line 155: It doesn't explain why this suppression is being introduced. All other referenced GitHub pull requests and issues actually fix those suppressed violations.

morozov (commented):
Code-wise looks good now. A couple of notes are still remaining.

psalm.xml
derrabus: That particular line isn't needed anymore, I figured. I'll see what I can do about the other ones.

derrabus: Check again, please.

morozov: Removing is fine.

lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php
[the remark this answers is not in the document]
derrabus: Well then. `{@inheritdoc}`s removed, error suppressed.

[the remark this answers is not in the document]
derrabus: All right, but PHPCS didn't like that. May I add the blocks again since apparently we did inherit something there?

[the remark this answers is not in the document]
morozov: You'll need to add the missing parameter annotations. When PHPCS notices an `{@inheritdoc}` it just doesn't report the missing annotations but having this annotation doesn't mean that they are inherited from somewhere.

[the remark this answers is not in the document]
derrabus: Here you go.

morozov (approved):
Looks good 👍  @beberlei has to give his approval as well since he requested some changes initially.

beberlei (approved):

driesvints:
Thanks @derrabus!

greg0ire:
Thanks a lot @derrabus ! Awesome work!

greg0ire:
@morozov let's release 2.12.0 ?

morozov:
Before we release `2.12.x` and declare the PHP 8 support, we need to make sure that all extensions currently tested on CI are tested with PHP 8 as well (which should be all except for `ibm_db2`).

greg0ire:
I will make a PR so that everyone can see the CI failures.

martinssipenko:
Please leave a link to that PR here.

greg0ire:
Here you go: https://github.com/doctrine/dbal/pull/4361

derrabus:
Thank you very much everyone. Having a PHP 8 compatible DBAL is great news. 🎉

Issue #4295, opened 2020-09-24
MasterSlaveConnection was BC Broken in 2.11.0 affecting doctrine/doctrine-bundle


kralos:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.11.0

#### Summary

`Doctrine\DBAL\Connections\MasterSlaveConnection` has changed how it handles `$params` (`__construct` arg 1). Specifically, it now unsets the legacy parameters `master` / `slaves`. This is a BC break because `Doctrine\DBAL\Connection::getParams` is `public`. Anything relying on these parameters will now break however the major version of `doctrine/dbal` was not bumped.

#### Current behaviour

For example `doctrine/doctrine-bundle`'s `Command`s use the following code to extract connection params:

```php
$params = $connection->getParams();
if (isset($params['master'])) {
    $params = $params['master'];
}
```

These commands now fail:

```
03:26:50 ERROR     [console] Error thrown while running command "{command}". Message: "{message}" ["exception" => InvalidArgumentException { …},"extra" => ["command" => "-v doctrine:database:create","message" => "Connection does not contain a 'path' or 'dbname' parameter and cannot be created."]]

In CreateDatabaseDoctrineCommand.php line 82:

  [InvalidArgumentException]                                                         
  Connection does not contain a 'path' or 'dbname' parameter and cannot be created.  

```

#### How to reproduce

```php
$masterParam = [/***/];
$slavesParam = [/***/];
$conn = new \Doctrine\DBAL\Connections\MasterSlaveConnection([
    'master' => $masterParam,
    'slaves' => $slavesParam,
]);
\assert($conn->getParams()['master'] === $masterParam); // fails, used to pass in doctrine/dbal 2.10.4
\assert($conn->getParams()['slaves'] === $slavesParam); // fails, used to pass in doctrine/dbal 2.10.4
```

#### Expected behaviour

`Doctrine\DBAL\Connections\MasterSlaveConnection` should not `unset($params['master']);` / `unset($params['slaves']);`

greg0ire:
On the one hand, this is indeed a BC-break, on the other hand, [as pointed out by @stof](https://github.com/doctrine/DoctrineBundle/pull/1215#issuecomment-699046566), `Connection::getParams()` is marked as internal:

https://github.com/doctrine/dbal/blob/6ff72e4596ad8c597dedd49ac3249d727d71ce62/lib/Doctrine/DBAL/Connection.php#L221-L228

[...]

PR #4354, opened 2020-10-17
Deduplicate explicitly passed args (driverOptions, user, password)

mvorisek:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes - when custom driver does not use the explicit param
| Fixed issues | no

#### Summary

arguments should not be duplicated when passed explicitly

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -359,6 +359,10 @@ class Connection implements DriverConnection
         $user          = $this->params['user'] ?? null;
         $password      = $this->params['password'] ?? null;
 
+        unset($this->params['driverOptions']);
+        unset($this->params['user']);
+        unset($this->params['password']);
+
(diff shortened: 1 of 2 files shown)


mvorisek:
if considered BC breaking, let's merge it into upcoming 2.12

greg0ire:
> if considered BC breaking, let's merge it into upcoming 2.12

You mean 3.0.x, right?

Regarding unsetting parameters and BC-breaks, see https://github.com/doctrine/dbal/issues/4295

mvorisek:
> You mean 3.0.x, right?

It seems it is not used consistently, some drivers use password from array, some from arguments.

Do you support the idea of completely removing user/password param? Then definitely into 3.0.x.

> Regarding unsetting parameters and BC-breaks, see #4295

we can revert the legacy fix for 3.0.x

mvorisek:
closing this as it will be quite breaking (at least internally) and will require quite a lof of work - if someone wants to pursue, go ahead!

PR #4355, opened 2020-10-17

morozov:


morozov:
The Psalm action is still misbehaving. @greg0ire could you do some research?

Also, the OCI-based extensions started failing to install all of a sudden.

greg0ire:
I reported the OCI issue here: https://github.com/shivammathur/setup-php/issues/312

When I checkout this branch and run `composer install && vendor/bin/psalm`, this is what I get:

```
ERROR: UndefinedMethod - src/Tools/Console/ConsoleRunner.php:29:13 - Method PackageVersions\Versions::rootpackagename does not exist (see https://psalm.dev/022)
            Versions::rootPackageName()
```

Running `composer update ocramius/package-versions composer/package-versions-deprecated` fixes the issue locally for some reason, so that's what I will force push to this branch in a moment.

greg0ire:
Well… that did not help it seems :(

greg0ire:
The Psalm issue is a special kind of nasty caused by a piece of code I can find locally but not on https://github.com/composer/package-versions-deprecated. It looks like this:

`@psalm-param key-of<self::VERSIONS> $packageName` and is located in `src/PackageVersions/Versions.php`

The only matching commit for that string is https://github.com/composer/package-versions-deprecated/commit/15b57b9cac3ecba44a193394a9e2ed852138f05c , and doesn't even apply on `src/PackageVersions/Versions.php`

greg0ire:
The OCI-based jobs are fixed.

greg0ire:
@morozov maybe we should revert part of 3ee27c32a7 ? Why did you do that change?

morozov:
@greg0ire thanks for all the findings and fixes. I believe the issue is in the package-versions package but it's an installer / code generator bug. Look for occurrences of "__root__" in the Composer source code. Declaring the return type as `self::VERSIONS` (by the way, is it `self` everywhere) and then returning an invalid value would likely have the same problem (just not in our source code).

> @morozov maybe we should revert part of [3ee27c3](https://github.com/doctrine/dbal/commit/3ee27c32a75bc3ea379395b81973e1fdeae72632) ?

I'll try to suppress just this very error in our `psalm.xml`.

> Why did you do that change?

Because it introduced `Versions::rootPackageName()` which fixed this very same problem the previous time.

greg0ire:
This is what is used by Psalm: `composer install --no-scripts --no-progress`.
I thought the `--no-scripts` part would explain that bug, but I still cannot reproduce it locally with that command line. It does have an effect on the generated file though.

morozov:
> This is what is used by Psalm: `composer install --no-scripts --no-progress`.
> I thought the `--no-scripts` part would explain that bug, but I still cannot reproduce it locally with that command line. It does have an effect on the generated file though.

Yeah, I tried all the same yesterday with the same result.

> Can you maybe try adding a step that runs `composer install --no-progress`?

I'd rather suppress the error which would clearly indicate that there is a bug somewhere.

morozov:
The suppression worked: https://github.com/doctrine/dbal/blob/3c845853b0eabba9277f40fb7a79b22233b875d5/psalm.xml#L26-L33

greg0ire (approved):
Great! Let's push this!

Issue #4279, opened 2020-09-21
transactional() doesn't correctly manage nested transactions with savepoints upon MySQL deadlock


aboks:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.10.2

#### Summary

Using DBAL with MySQL, if a deadlock occurs inside a nested `transactional()` (with savepoints to emulate nested transactions), the DBAL connection tries to rollback the savepoint. However, MySQL automatically rolls back the transaction (and thus releases all savepoints) upon a deadlock. Therefore the attempt to rollback the savepoint fails with `SQLSTATE[42000]: Syntax error or access violation: 1305 SAVEPOINT DOCTRINE2_SAVEPOINT_2 does not exist`.

#### Current behaviour

The DBAL connection does not correctly take into account that the deadlock rolled back the transaction and all savepoints, and thus tries to rollback to the nonexisting savepoint. This fails with

```
Fatal error: Uncaught PDOException: SQLSTATE[42000]: Syntax error or access violation: 1305 SAVEPOINT DOCTRINE2_SAVEPOINT_2 does not exist in /var/www/html/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php:41
Stack trace:
#0 /var/www/html/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php(41): PDO->exec()
#1 /var/www/html/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(1439): Doctrine\DBAL\Driver\PDOConnection->exec()
#2 /var/www/html/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(1371): Doctrine\DBAL\Connection->rollbackSavepoint()
#3 /var/www/html/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(1181): Doctrine\DBAL\Connection->rollBack()
#4 /var/www/html/deadlock.php(43): Doctrine\DBAL\Connection->transactional()
#5 {main}

Next Doctrine\DBAL\Driver\PDOException: SQLSTATE[42000]: Syntax error or access violation: 1305 SAVEPOINT DOCTRINE2_SAVEPOINT_2 does not exist in /var/www/html/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php on line 43
```

#### How to reproduce

[...]

PR #4356, opened 2020-10-18
Testing Guidelines

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

This is a placeholder and a temporary reference for the testing requirements and guidelines.

**TODO**:
- [x] Add more details and examples to facilitate decision making.
- [x] Highlight the importance of testing on all platforms rather than on a specific one where the issue being fixed was observed. In addition to making sure that all platforms implement the same logic, they can be used as a reference for using the DBAL API.
- [x] Document the cases that may be exempt from automated integration testing (e.g. transactions, locks, race conditions, etc).
- [x] If a test covers a fix, it should fail without the fix.
- [x] Add a link to the sidebar.

### DIFF
--- a/docs/en/reference/testing.rst
+++ b/docs/en/reference/testing.rst
@@ -1,0 +1,123 @@
+Testing Guidelines
+===================
+
+To ensure high quality, all components of the Doctrine DBAL library are extensively covered with tests.
+
+Having the code covered with tests and running all tests against each individual code change helps prevent
+breakages of the library logic when its code changes.
+
+Additionally, when code changes are accompanied by new tests, the tests:
+
+1. Help understand what problem the given code change is trying to solve.
+2. Make sure that the problem being solved needs to be solved in the DBAL.
+3. Document the proper usage of the DBAL APIs.
+
+Requirements
+------------
+
+1. Each pull request that adds new or changes the existing logic must have tests.
+
+   .. note::
+
+       Modifications to the keyword lists under the ``Doctrine\DBAL\Platforms\Keywords`` namespace
+       don't have to be covered with tests.
+
+2. The test that covers certain logic must fail without this logic implemented.
+
+Types of Tests
+--------------
+
+Doctrine DBAL primarily uses unit and integration tests.
+
+Unit Tests
+~~~~~~~~~~
+
+Unit tests are meant to cover the logic of a given unit (e.g. a class or a method) including the logic
+of its interaction with other units. In this case, the other units could be mocked.
+
+Unit tests are most welcomed for testing the logic that the DBAL itself defines (e.g. logging, caching, data types).
+
+In this case, the DBAL is the source of truth about what this logic is and the test plays the role of its description.
+
+Integration Tests
+~~~~~~~~~~~~~~~~~
+
+Integration (a.k.a. functional) tests are required when the behavior under test is dictated by the logic
+defined outside of the DBAL. It could be:
+
+- The underlying database platform.
+- The underlying database driver.
+- SQL syntax and the standard as such.
+
+It is important to have integration tests for the cases above. Unlike unit tests, they make the external components
+the source of truth and help make sure that the logic implemented in the DBAL is correct even if the external components
+change (e.g. a new version of a database platform is supported).
+
+When are Integration Tests not Required?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Some cases cannot be reproduced with the existing integration testing suite. It could be the scenarios that involve
+multiple concurrent database connections, transactions, locking, performance-related issues, etc.
+
+In such cases, it is still important that a pull request fixing the issues is accompanied by a free-form reproducer
+that demonstrates the issue being fixed.
+
+Recommendations on Writing Tests
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Tests in Doctrine DBAL are located under the ``tests`` directory and implemented on top of PHPUnit. Use its
+`documentation <https://phpunit.de/documentation.html>`_ to get started.
+
+Writing Integration Tests
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Integration tests are located under the ``tests/Doctrine/Tests/DBAL/Functional`` directory. Unlike unit tests,
+they require a real database connection to test their logic against.
+
+It is recommended to use ``Doctrine\DBAL\Tests\FunctionalTestCase`` as the base class for integration tests.
+Based on the configuration, it will automatically create and connect to the test database.
+
+Data Fixtures in Integration Tests
+++++++++++++++++++++++++++++++++++
+
+To test selecting and fetching data from the database, the test may create the necessary schema and populate it
+with the test data. To create database tables, instead of checking if the table exists, it is recommended
+to use ``AbstractSchemaManager::dropAndCreateTable()``. This way, the table will be dropped and created every time
+providing better isolation between the test runs.
+
+Testing Different Database Platforms
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Although most of the issues are originally discovered on a specific database platform,
+the integration tests for all issues should be implemented by default at the database abstraction level
+and run against all the platforms that support the API being tested.
+
+This allows us to ensure that the same scenario that was found failing on one platform also works on others. Or otherwise,
+the same issue could be reproduced on the platforms where it wasn't originally tested.
+
+If the newly added test fails on other platforms, and fixing it is out of the scope, the test can be explicitly marked
+as incomplete which will identify the issue.
+
+Examples of such tests could be found under the ``Doctrine\Tests\DBAL\Functional\Platform`` namespace.
+
+Using Unit and Integration Tests Together
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+For example, the ``AbstractPlatform::modifyLimitQuery()`` method has both unit and integration tests.
+
+1. Unit test cases for each platform (``Doctrine\Tests\DBAL\Platforms\*PlatformTest``) have a test that calls
+   ``$platform->modifyLimitQuery()`` and asserts that the resulting SQL looks as expected.
+   These tests cannot guarantee that the generated SQL is valid syntactically and semantically but they guarantee
+   that the code works as designed. They provide fast feedback because they don't require a database connection
+   and can test all platforms in a single test suite run.
+2. There is an integration test ``Doctrine\Tests\DBAL\Functional\ModifyLimitQueryTest`` which calls
+   ``$platform->modifyLimitQuery()`` and executes the generated queries on a real database to which the test suite
+   is connected. This test guarantees that the generated queries are valid but it's much slower and works
+   only with one database at a time.
+
+As you can see, both approaches have their strengths and weaknesses and can complement each other.
+
+.. warning::
+
+    Do not mix the unit and the integration approaches in one test. Each of the approaches has its area of application
+    and purpose. Mixing them makes it harder to identify the reason and the impact of a failing mixed-type test.
--- a/docs/en/sidebar.rst
+++ b/docs/en/sidebar.rst
@@ -20,5 +20,6 @@
     reference/caching
     reference/known-vendor-issues
     reference/upgrading
+    reference/testing
 
     explanation/implicit-indexes


SenseException:
In its current state it would also fit well to other projects than dbal. Could this be used as a generic guideline on the website or will this diverge more from the other projects?

morozov:
My goal is to have the DBAL documentation first. Once it's done content-wise, we can see how it could be restructured into the generic and DBAL-specific sections. I'm open to suggestions and contributions.

morozov:
I could use some help with the examples (if needed) and feedback on whether some ideas need to be clarified.

docs/en/reference/testing.rst
```diff
@@ -0,0 +1,74 @@
+Testing Guidelines
+===================
+
+In order to ensure high quality, all components the Doctrine DBAL library are extensively covered with tests.
```
greg0ire, line 4:
```suggestion
In order to ensure high quality, all components of the Doctrine DBAL library are extensively covered with tests.
```

docs/en/reference/testing.rst
```diff
@@ -0,0 +1,74 @@
...
+Additionally, tests:
+
+1. Help maintainers understand what problem the given code change is trying to solve.
+2. Make sure that the problem being solved needs to be solved in the DBAL.
+3. Document the proper usage of the DBAL APIs.
+
+Requirements
+------------
+
+1. Each pull request that adds new or changes the existing logic must have tests.
```
greg0ire, line 18: That sentence looks correct but sounds a bit awkward, I'd rather have some repetition here.
```suggestion
1. Each pull request that adds new logic or changes the existing logic must have tests.
```

docs/en/reference/testing.rst
```diff
@@ -0,0 +1,74 @@
...
+2. The test that covers certain logic must fail without the changes in the code it covers.
+
+Unit Tests
+----------
+
+Unit test are meant to cover the logic of a given unit (e.g. a class or a method) including the logic of its interaction
```
greg0ire, line 24:
```suggestion
Unit tests are meant to cover the logic of a given unit (e.g. a class or a method) including the logic of its interaction
```

docs/en/reference/testing.rst
```diff
@@ -0,0 +1,74 @@
...
+with other units. In this case, the other units could be mocked.
+
+Unit tests are mostly welcomed for testing the logic that the DBAL itself defines (e.g. logging). They also could be
+used as a secondary tier of testing for the logic that requires integration testing.
+
+Integration Tests
+-----------------
+
+Integration (a.k.a. functional) test are required when the behavior under the test is dictated by the logic
```
greg0ire, line 33:
```suggestion
Integration (a.k.a. functional) tests are required when the behavior under test is dictated by the logic
```

docs/en/reference/testing.rst
```diff
@@ -0,0 +1,74 @@
...
+multiple concurrent database connections, transactions, locking, performance-related issues, etc.
+
+In such cases, it is still important that a pull request fixing the issues is accompanied with a free-form test that
+reproduces the issue being fixed.
+
+Testing Different Database Platforms
+------------------------------------
+
+Despite the fact that most of the issues are originally discovered on a specific database platform,
+the integration tests for all issues by default should be implemented at the database abstraction level
```
greg0ire, line 57:
```suggestion
the integration tests for all issues should be implemented by default at the database abstraction level
```

docs/en/reference/testing.rst
```diff
@@ -0,0 +1,74 @@
...
+2. The test that covers certain logic must fail without the changes in the code it covers.
+
+Unit Tests
+----------
+
+Unit test are meant to cover the logic of a given unit (e.g. a class or a method) including the logic of its interaction
+with other units. In this case, the other units could be mocked.
+
+Unit tests are mostly welcomed for testing the logic that the DBAL itself defines (e.g. logging). They also could be
+used as a secondary tier of testing for the logic that requires integration testing.
```
greg0ire, line 28: That last sentence could be clarified with an example.

docs/en/reference/testing.rst
```diff
@@ -0,0 +1,74 @@
...
+
+1. The underlying database platform.
+2. The underlying database driver.
+3. SQL syntax and the standard as such.
+
+It is important to have integration tests for the cases above. Unlike unit tests, they make the external components
+the source of truth and help make sure that the logic implemented in the DBAL is correct even if the external components
+change (e.g. a new version of a database platform is supported).
+
+When Integration Testing Is Not Required
```
greg0ire, line 44: I'm not a native speaker, but this feels a bit awkward too. I don't have a better suggestion though

docs/en/reference/testing.rst
```diff
@@ -0,0 +1,74 @@
...
+----------------------------------------
+
+Some cases cannot be reproduced with the existing integration testing suite. It could be the scenarios that involve
+multiple concurrent database connections, transactions, locking, performance-related issues, etc.
+
+In such cases, it is still important that a pull request fixing the issues is accompanied with a free-form test that
```
greg0ire, line 50: What would such a test look like?

morozov: It could be a script like in https://github.com/doctrine/dbal/issues/4279, a gist, a PoC project, a `docker-compose` setup (to be added to the text).

I don't think I'll be able to provide a specific example of each option but we can update the documentation once we find some good examples.

docs/en/reference/testing.rst
morozov: I'm open to suggestions. You get the idea from the text under the header.

SenseException: "When are Integration Tests not required?"
"When Integration Tests are not required"

docs/en/reference/testing.rst
```diff
@@ -0,0 +1,74 @@
...
+as incomplete which will clearly identify the issue.
+
+Examples of such tests could be found under the the
+`Doctrine\Tests\DBAL\Functional\Platform <https://github.com/doctrine/dbal/tree/2.11.3/tests/Doctrine/Tests/DBAL/Functional/Platform>`_
+namespace.
+
+Other Notes
+-----------
+
+Do not mix the unit and the integration approaches in one test. Each of the approaches has its own area of application
```
SenseException, line 73: This needs the paths and separation strategies shown or explained, that are used in the repo.

morozov: Could you elaborate on what you mean? I can actually add the details about how exactly the functional tests are written: the namespace, the base class and the auto-instantiated connection, something about paying more attention to the cleanup/teardown, but it would all go to the integration testing section.

As for the unit testing, there's not much to add.

Could you ask me some questions that I may try to answer?

docs/en/reference/testing.rst
```diff
@@ -0,0 +1,95 @@
...
+1. Help maintainers understand what problem the given code change is trying to solve.
+2. Make sure that the problem being solved needs to be solved in the DBAL.
+3. Document the proper usage of the DBAL APIs.
+
+Requirements
+------------
+
+1. Each pull request that adds new logic or changes the existing logic must have tests.
+
+    (!) Exceptions to the rule:
```
morozov, line 20: This is meant visually highlighted (I'll figure out the RST syntax for this later).

docs/en/reference/testing.rst
morozov: I moved this under the "Using Unit and Integration Tests Together" section. The wording is to be revisited.

docs/en/reference/testing.rst
```diff
@@ -0,0 +1,95 @@
...
+change (e.g. a new version of a database platform is supported).
+
+When are Integration Tests not required?
+----------------------------------------
+
+Some cases cannot be reproduced with the existing integration testing suite. It could be the scenarios that involve
+multiple concurrent database connections, transactions, locking, performance-related issues, etc.
+
+In such cases, it is still important that a pull request fixing the issues is accompanied with a free-form test that
+reproduces the issue being fixed.
```
greg0ire, line 56: Ok, how about this then:

```suggestion
In such cases, it is still important that a pull request fixing the issues is accompanied with a free-form reproducer that demonstrates the issue being fixed.
```

? I think the word `test` should be avoided here.

morozov: If "test" is misleading in this case, let's use "reproducer".

docs/en/reference/testing.rst
SenseException: Yes, I meant the namespace, the base class and that stuff. I hadn't the cleanup in mind, but if you think this should be added too, then feel free to add it. `Doctrine\Tests\DBAL\Functional` is kind of hidden inside the namespaces/directories, so questions could be:

* Where do I put integration tests?
* Where do I put/setup fixtures for the integration tests? (Or: where does the data for integration tests come from?)

Because we already have examples in the project itself, there can be a link to one of those, so it doesn't have to be a detailed explanation.

For some people testing is something new and a link to the PHPUnit docs could be added too.

morozov:
I haven't been finding much passion lately to finish this document, so in order to not let it rot, let's give it another round of proofreading and try to not increase the scope.

The primary idea I want to convey is the need to write integration tests across all supported platforms as I have been asking contributors for dozens and dozens of time. Everybody can contribute to this guide afterwards and add the other important bits.

docs/en/reference/testing.rst
```diff
@@ -0,0 +1,125 @@
...
+Examples of such tests could be found under the
+`Doctrine\Tests\DBAL\Functional\Platform <https://github.com/doctrine/dbal/tree/2.11.3/tests/Doctrine/Tests/DBAL/Functional/Platform>`_
+namespace.
+
+Using Unit and Integration Tests Together
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+For example, the ``AbstractPlatform::modifyLimitQuery()`` method has both unit and integration tests.
+
+1. Unit test cases for each platform (``Doctrine\Tests\DBAL\Platforms\*PlatformTest.php``) have a test that calls
```
morozov, line 110: The ".php" needs to be removed since it's a class name, not a file name.

docs/en/reference/testing.rst
```diff
@@ -0,0 +1,125 @@
...
+and run against all the platforms that support the API being tested.
+
+This allows us to ensure that the same scenario that was found failing on one platform also works on others. Or otherwise,
+the same issue could be reproduced on the platforms where it wasn't originally tested.
+
+If the newly added test fails on other platforms, and fixing it is out of the scope, the test can be explicitly marked
+as incomplete which will identify the issue.
+
+Examples of such tests could be found under the
+`Doctrine\Tests\DBAL\Functional\Platform <https://github.com/doctrine/dbal/tree/2.11.3/tests/Doctrine/Tests/DBAL/Functional/Platform>`_
```
morozov, line 102: This doesn't seem to be properly rendered, at least on GitHub.

docs/en/reference/testing.rst
```diff
@@ -0,0 +1,123 @@
...
+`documentation <https://phpunit.de/documentation.html>`_ to get started.
+
+Writing Integration Tests
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Integration tests are located under the ``tests/Doctrine/Tests/DBAL/Functional`` directory. Unlike unit tests,
+they require a real database connection to test their logic against.
+
+It is recommended to use ``Doctrine\DBAL\Tests\FunctionalTestCase`` as the base class for integration tests.
+Base on the configuration, it will automatically create and connect to the test database.
```
greg0ire, line 78:
```suggestion
Based on the configuration, it will automatically create and connect to the test database.
```

greg0ire (dismissed):

docs/en/reference/testing.rst
```diff
@@ -0,0 +1,123 @@
...
+
+Unit Tests
+~~~~~~~~~~
+
+Unit tests are meant to cover the logic of a given unit (e.g. a class or a method) including the logic
+of its interaction with other units. In this case, the other units could be mocked.
+
+Unit tests are most welcomed for testing the logic that the DBAL itself defines (e.g. logging, caching, data types).
+
+In this case, the DBAL is the source of truth about what this logic should like and the test behaves as its description.
```
SenseException, line 40:
```suggestion
In this case, the DBAL is the source of truth about what this logic should be like and the test behaves as its description.
```

SenseException (changes requested):
> I haven't been finding much passion lately to finish this document, so in order to not let it rot, let's give it another round of proofreading and try to not increase the scope.

Okay :+1: . Just have a small change, otherwise this is ready for a merge IMO.

docs/en/reference/testing.rst
morozov: Not sure what exactly I wanted to say by "should like", I replaced it with just "is".

greg0ire (approved):

SenseException (approved):

morozov:
@SenseException how do we get this deployed to the website?

Issue #4357, opened 2020-10-18
The support for quoting identifiers is fundamentally broken


morozov:
The DBAL takes the responsibility for automatic quoting of identifiers but it doesn't work in many cases:

1. The list of keywords is never complete and up to date.
2. It's impossible to quote a keyword or another identifier that needs quoting manually.

The details are to follow.

morozov:
Object names are normalized assuming that all target platforms are case-insensitive about object names:

https://github.com/doctrine/dbal/blob/0e398ec80cc877acb7fdde653503638b78f30b8c/src/Schema/Table.php#L984-L991

This is not true at least for PostgreSQL, Oracle, IBM DB2 and SQL Server (if the database uses a case sensitive collation).
```php
$connection = DriverManager::getConnection([
    'driver'   => 'oci8',
    'host'     => 'localhost',
    'user'     => 'system',
    'password' => 'oracle',
    'dbname'   => 'XE',
]);

$connection->executeStatement('CREATE TABLE test ("id" NUMBER, ID NUMBER)');
$sm = $connection->createSchemaManager();
$sm->listTableDetails('test');
// Doctrine\DBAL\Schema\SchemaException: The column 'id' on table 'test' already exists.
```

morozov:
The schema implementation assumes that the object names need to be quoted only if they are keywords or in some other platform-specific cases. For instance:
```php
$connection = DriverManager::getConnection([
    'driver'   => 'oci8',
    'host'     => 'localhost',
    'user'     => 'system',
    'password' => 'oracle',
    'dbname'   => 'XE',
]);

$connection->executeStatement('CREATE TABLE test ("X Y" NUMBER)');
$sm = $connection->createSchemaManager();
$table = $sm->listTableDetails('test');
$online = clone $table;
$id = $table->getColumn('x y');
$id->setType(Type::getType('string'));
$diff = $sm->createComparator()->diffTable($online, $table);
$sm->alterTable($diff);
// ALTER TABLE test MODIFY (X Y VARCHAR2(255) DEFAULT NULL)
// ORA-00907: missing right parenthesis
```
An identifier obtained by introspecting the schema should be always quoted since it was introspected in its original case.

morozov:
But it's not only about proprietary platforms. A similar issue is reproducible on any platform, e.g. MySQL:
```php
$connection = DriverManager::getConnection([
    'driver' => 'mysqli',
    'host'   => '127.0.0.1',
    'user'   => 'root',
[...]

PR #4360, opened 2020-10-19
Fix identifier escaping when used as string literal for T-SQL

mvorisek:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #4283

#### Summary

Alternative to closes https://github.com/doctrine/dbal/pull/4284 to fix it correctly and with proper functional tests.

The issue was that MSSQL expects string literal (quoted by `'`) for `EXEC sp_addextendedproperty` instead of identifier name (quoted by `[]`).

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -26,6 +26,7 @@ use function is_bool;
 use function is_numeric;
 use function is_string;
 use function preg_match;
+use function preg_replace;
 use function sprintf;
 use function str_replace;
 use function stripos;
@@ -787,17 +788,22 @@ class SQLServerPlatform extends AbstractPlatform
         ];
     }
 
+    private function quoteSingleIdentifierAsStringLiteral(string $levelName): string
+    {
+        return $this->quoteStringLiteral(preg_replace('~^\[|\]$~s', '', $levelName));
+    }
+
     /**
      * Returns the SQL statement for adding an extended property to a database object.
      *
      * @link http://msdn.microsoft.com/en-us/library/ms180047%28v=sql.90%29.aspx
      *
      * @param string      $name       The name of the property to add.
-     * @param string|null $value      The value of the property to add.
-     * @param string|null $level0Type The type of the object at level 0 the property belongs to.
-     * @param string|null $level0Name The name of the object at level 0 the property belongs to.
-     * @param string|null $level1Type The type of the object at level 1 the property belongs to.
-     * @param string|null $level1Name The name of the object at level 1 the property belongs to.
+     * @param string      $value      The value of the property to add.
+     * @param string      $level0Type The type of the object at level 0 the property belongs to.
+     * @param string      $level0Name The name of the object at level 0 the property belongs to.
+     * @param string      $level1Type The type of the object at level 1 the property belongs to.
+     * @param string      $level1Name The name of the object at level 1 the property belongs to.
      * @param string|null $level2Type The type of the object at level 2 the property belongs to.
      * @param string|null $level2Name The name of the object at level 2 the property belongs to.
      *
@@ -805,19 +811,25 @@ class SQLServerPlatform extends AbstractPlatform
      */
     public function getAddExtendedPropertySQL(
         $name,
-        $value = null,
-        $level0Type = null,
-        $level0Name = null,
-        $level1Type = null,
-        $level1Name = null,
-        $level2Type = null,
-        $level2Name = null
+        $value,
+        string $level0Type,
+        string $level0Name,
+        string $level1Type,
+        string $level1Name,
+        ?string $level2Type = null,
+        ?string $level2Name = null
     ) {
-        return 'EXEC sp_addextendedproperty ' .
-            'N' . $this->quoteStringLiteral($name) . ', N' . $this->quoteStringLiteral((string) $value) . ', ' .
-            'N' . $this->quoteStringLiteral((string) $level0Type) . ', ' . $level0Name . ', ' .
-            'N' . $this->quoteStringLiteral((string) $level1Type) . ', ' . $level1Name . ', ' .
-            'N' . $this->quoteStringLiteral((string) $level2Type) . ', ' . $level2Name;
+        return 'EXEC sp_addextendedproperty'
+            . ' N' . $this->quoteStringLiteral($name) . ', N' . $this->quoteStringLiteral((string) $value)
+            . ', N' . $this->quoteStringLiteral($level0Type)
+            . ', ' . $this->quoteSingleIdentifierAsStringLiteral($level0Name)
+            . ', N' . $this->quoteStringLiteral($level1Type)
+            . ', ' . $this->quoteSingleIdentifierAsStringLiteral($level1Name)
+            . ($level2Type !== null || $level2Name !== null
+                ? ', N' . $this->quoteStringLiteral((string) $level2Type)
+                  . ', ' . $this->quoteSingleIdentifierAsStringLiteral((string) $level2Name)
+                : ''
+              );
     }
 
     /**
@@ -826,10 +838,10 @@ class SQLServerPlatform extends AbstractPlatform
      * @link http://technet.microsoft.com/en-gb/library/ms178595%28v=sql.90%29.aspx
      *
      * @param string      $name       The name of the property to drop.
-     * @param string|null $level0Type The type of the object at level 0 the property belongs to.
-     * @param string|null $level0Name The name of the object at level 0 the property belongs to.
-     * @param string|null $level1Type The type of the object at level 1 the property belongs to.
-     * @param string|null $level1Name The name of the object at level 1 the property belongs to.
+     * @param string      $level0Type The type of the object at level 0 the property belongs to.
+     * @param string      $level0Name The name of the object at level 0 the property belongs to.
+     * @param string      $level1Type The type of the object at level 1 the property belongs to.
+     * @param string      $level1Name The name of the object at level 1 the property belongs to.
      * @param string|null $level2Type The type of the object at level 2 the property belongs to.
      * @param string|null $level2Name The name of the object at level 2 the property belongs to.
      *
@@ -837,18 +849,24 @@ class SQLServerPlatform extends AbstractPlatform
      */
     public function getDropExtendedPropertySQL(
         $name,
-        $level0Type = null,
-        $level0Name = null,
-        $level1Type = null,
-        $level1Name = null,
-        $level2Type = null,
-        $level2Name = null
+        string $level0Type,
+        string $level0Name,
+        string $level1Type,
+        string $level1Name,
+        ?string $level2Type = null,
+        ?string $level2Name = null
     ) {
-        return 'EXEC sp_dropextendedproperty ' .
-            'N' . $this->quoteStringLiteral($name) . ', ' .
-            'N' . $this->quoteStringLiteral((string) $level0Type) . ', ' . $level0Name . ', ' .
-            'N' . $this->quoteStringLiteral((string) $level1Type) . ', ' . $level1Name . ', ' .
-            'N' . $this->quoteStringLiteral((string) $level2Type) . ', ' . $level2Name;
+        return 'EXEC sp_dropextendedproperty'
+            . ' N' . $this->quoteStringLiteral($name)
+            . ', N' . $this->quoteStringLiteral($level0Type)
+            . ', ' . $this->quoteSingleIdentifierAsStringLiteral($level0Name)
+            . ', N' . $this->quoteStringLiteral($level1Type)
+            . ', ' . $this->quoteSingleIdentifierAsStringLiteral($level1Name)
+            . ($level2Type !== null || $level2Name !== null
+                ? ', N' . $this->quoteStringLiteral((string) $level2Type)
+                  . ', ' . $this->quoteSingleIdentifierAsStringLiteral((string) $level2Name)
+                : ''
+              );
     }
 
     /**
@@ -857,11 +875,11 @@ class SQLServerPlatform extends AbstractPlatform
      * @link http://msdn.microsoft.com/en-us/library/ms186885%28v=sql.90%29.aspx
      *
      * @param string      $name       The name of the property to update.
-     * @param string|null $value      The value of the property to update.
-     * @param string|null $level0Type The type of the object at level 0 the property belongs to.
-     * @param string|null $level0Name The name of the object at level 0 the property belongs to.
-     * @param string|null $level1Type The type of the object at level 1 the property belongs to.
-     * @param string|null $level1Name The name of the object at level 1 the property belongs to.
+     * @param string      $value      The value of the property to update.
+     * @param string      $level0Type The type of the object at level 0 the property belongs to.
+     * @param string      $level0Name The name of the object at level 0 the property belongs to.
+     * @param string      $level1Type The type of the object at level 1 the property belongs to.
+     * @param string      $level1Name The name of the object at level 1 the property belongs to.
      * @param string|null $level2Type The type of the object at level 2 the property belongs to.
      * @param string|null $level2Name The name of the object at level 2 the property belongs to.
      *
@@ -869,19 +887,25 @@ class SQLServerPlatform extends AbstractPlatform
      */
     public function getUpdateExtendedPropertySQL(
         $name,
-        $value = null,
-        $level0Type = null,
-        $level0Name = null,
-        $level1Type = null,
-        $level1Name = null,
-        $level2Type = null,
-        $level2Name = null
+        $value,
+        string $level0Type,
+        string $level0Name,
+        string $level1Type,
+        string $level1Name,
+        ?string $level2Type = null,
+        ?string $level2Name = null
     ) {
-        return 'EXEC sp_updateextendedproperty ' .
-            'N' . $this->quoteStringLiteral($name) . ', N' . $this->quoteStringLiteral((string) $value) . ', ' .
-            'N' . $this->quoteStringLiteral((string) $level0Type) . ', ' . $level0Name . ', ' .
-            'N' . $this->quoteStringLiteral((string) $level1Type) . ', ' . $level1Name . ', ' .
-            'N' . $this->quoteStringLiteral((string) $level2Type) . ', ' . $level2Name;
+        return 'EXEC sp_updateextendedproperty'
+            . ' N' . $this->quoteStringLiteral($name) . ', N' . $this->quoteStringLiteral((string) $value)
+            . ', N' . $this->quoteStringLiteral($level0Type)
+            . ', ' . $this->quoteSingleIdentifierAsStringLiteral($level0Name)
+            . ', N' . $this->quoteStringLiteral($level1Type)
+            . ', ' . $this->quoteSingleIdentifierAsStringLiteral($level1Name)
+            . ($level2Type !== null || $level2Name !== null
+                ? ', N' . $this->quoteStringLiteral((string) $level2Type)
+                  . ', ' . $this->quoteSingleIdentifierAsStringLiteral((string) $level2Name)
+                : ''
+              );
     }
 
     /**
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -360,21 +360,18 @@ class SQLServerPlatform extends AbstractPlatform
     protected function getCreateColumnCommentSQL($tableName, $columnName, $comment)
     {
         if (strpos($tableName, '.') !== false) {
-            [$schemaSQL, $tableSQL] = explode('.', $tableName);
-            $schemaSQL              = $this->quoteStringLiteral($schemaSQL);
-            $tableSQL               = $this->quoteStringLiteral($tableSQL);
+            [$schemaName, $tableName] = explode('.', $tableName, 2);
         } else {
-            $schemaSQL = "'dbo'";
-            $tableSQL  = $this->quoteStringLiteral($tableName);
+            $schemaName = $this->getDefaultSchemaName();
         }
 
         return $this->getAddExtendedPropertySQL(
             'MS_Description',
-            $comment,
+            (string) $comment,
             'SCHEMA',
-            $schemaSQL,
+            $schemaName,
             'TABLE',
-            $tableSQL,
+            $tableName,
             'COLUMN',
             $columnName
         );
@@ -724,21 +721,18 @@ class SQLServerPlatform extends AbstractPlatform
     protected function getAlterColumnCommentSQL($tableName, $columnName, $comment)
     {
         if (strpos($tableName, '.') !== false) {
-            [$schemaSQL, $tableSQL] = explode('.', $tableName);
-            $schemaSQL              = $this->quoteStringLiteral($schemaSQL);
-            $tableSQL               = $this->quoteStringLiteral($tableSQL);
+            [$schemaName, $tableName] = explode('.', $tableName, 2);
         } else {
-            $schemaSQL = "'dbo'";
-            $tableSQL  = $this->quoteStringLiteral($tableName);
+            $schemaName = $this->getDefaultSchemaName();
         }
 
         return $this->getUpdateExtendedPropertySQL(
             'MS_Description',
-            $comment,
+            (string) $comment,
             'SCHEMA',
-            $schemaSQL,
+            $schemaName,
             'TABLE',
-            $tableSQL,
+            $tableName,
             'COLUMN',
             $columnName
         );
@@ -763,20 +757,17 @@ class SQLServerPlatform extends AbstractPlatform
     protected function getDropColumnCommentSQL($tableName, $columnName)
     {
         if (strpos($tableName, '.') !== false) {
-            [$schemaSQL, $tableSQL] = explode('.', $tableName);
-            $schemaSQL              = $this->quoteStringLiteral($schemaSQL);
-            $tableSQL               = $this->quoteStringLiteral($tableSQL);
+            [$schemaName, $tableName] = explode('.', $tableName, 2);
         } else {
-            $schemaSQL = "'dbo'";
-            $tableSQL  = $this->quoteStringLiteral($tableName);
+            $schemaName = $this->getDefaultSchemaName();
         }
 
         return $this->getDropExtendedPropertySQL(
             'MS_Description',
             'SCHEMA',
-            $schemaSQL,
+            $schemaName,
             'TABLE',
-            $tableSQL,
+            $tableName,
             'COLUMN',
             $columnName
         );
@@ -1689,14 +1680,19 @@ class SQLServerPlatform extends AbstractPlatform
 
     protected function getCommentOnTableSQL(string $tableName, ?string $comment): string
     {
-        return sprintf(
-            "
-                EXEC sys.sp_addextendedproperty @name=N'MS_Description',
-                  @value=N%s, @level0type=N'SCHEMA', @level0name=N'dbo',
-                  @level1type=N'TABLE', @level1name=N%s
-            ",
-            $this->quoteStringLiteral((string) $comment),
-            $this->quoteStringLiteral($tableName)
+        if (strpos($tableName, '.') !== false) {
+            [$schemaName, $tableName] = explode('.', $tableName, 2);
+        } else {
+            $schemaName = $this->getDefaultSchemaName();
+        }
+
+        return $this->getAddExtendedPropertySQL(
+            'MS_Description',
+            (string) $comment,
+            'SCHEMA',
+            $schemaName,
+            'TABLE',
+            $tableName
         );
     }
 
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -799,11 +799,11 @@ class SQLServerPlatform extends AbstractPlatform
      * @link http://msdn.microsoft.com/en-us/library/ms180047%28v=sql.90%29.aspx
      *
      * @param string      $name       The name of the property to add.
-     * @param string      $value      The value of the property to add.
-     * @param string      $level0Type The type of the object at level 0 the property belongs to.
-     * @param string      $level0Name The name of the object at level 0 the property belongs to.
-     * @param string      $level1Type The type of the object at level 1 the property belongs to.
-     * @param string      $level1Name The name of the object at level 1 the property belongs to.
+     * @param string|null $value      The value of the property to add.
+     * @param string|null $level0Type The type of the object at level 0 the property belongs to.
+     * @param string|null $level0Name The name of the object at level 0 the property belongs to.
+     * @param string|null $level1Type The type of the object at level 1 the property belongs to.
+     * @param string|null $level1Name The name of the object at level 1 the property belongs to.
      * @param string|null $level2Type The type of the object at level 2 the property belongs to.
      * @param string|null $level2Name The name of the object at level 2 the property belongs to.
      *
@@ -811,20 +811,20 @@ class SQLServerPlatform extends AbstractPlatform
      */
     public function getAddExtendedPropertySQL(
         $name,
-        $value,
-        string $level0Type,
-        string $level0Name,
-        string $level1Type,
-        string $level1Name,
-        ?string $level2Type = null,
-        ?string $level2Name = null
+        $value = null,
+        $level0Type = null,
+        $level0Name = null,
+        $level1Type = null,
+        $level1Name = null,
+        $level2Type = null,
+        $level2Name = null
     ) {
         return 'EXEC sp_addextendedproperty'
             . ' N' . $this->quoteStringLiteral($name) . ', N' . $this->quoteStringLiteral((string) $value)
-            . ', N' . $this->quoteStringLiteral($level0Type)
-            . ', ' . $this->quoteSingleIdentifierAsStringLiteral($level0Name)
-            . ', N' . $this->quoteStringLiteral($level1Type)
-            . ', ' . $this->quoteSingleIdentifierAsStringLiteral($level1Name)
+            . ', N' . $this->quoteStringLiteral((string) $level0Type)
+            . ', ' . $this->quoteSingleIdentifierAsStringLiteral((string) $level0Name)
+            . ', N' . $this->quoteStringLiteral((string) $level1Type)
+            . ', ' . $this->quoteSingleIdentifierAsStringLiteral((string) $level1Name)
             . ($level2Type !== null || $level2Name !== null
                 ? ', N' . $this->quoteStringLiteral((string) $level2Type)
                   . ', ' . $this->quoteSingleIdentifierAsStringLiteral((string) $level2Name)
@@ -838,10 +838,10 @@ class SQLServerPlatform extends AbstractPlatform
      * @link http://technet.microsoft.com/en-gb/library/ms178595%28v=sql.90%29.aspx
      *
      * @param string      $name       The name of the property to drop.
-     * @param string      $level0Type The type of the object at level 0 the property belongs to.
-     * @param string      $level0Name The name of the object at level 0 the property belongs to.
-     * @param string      $level1Type The type of the object at level 1 the property belongs to.
-     * @param string      $level1Name The name of the object at level 1 the property belongs to.
+     * @param string|null $level0Type The type of the object at level 0 the property belongs to.
+     * @param string|null $level0Name The name of the object at level 0 the property belongs to.
+     * @param string|null $level1Type The type of the object at level 1 the property belongs to.
+     * @param string|null $level1Name The name of the object at level 1 the property belongs to.
      * @param string|null $level2Type The type of the object at level 2 the property belongs to.
      * @param string|null $level2Name The name of the object at level 2 the property belongs to.
      *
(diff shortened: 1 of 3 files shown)


lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -360,24 +360,20 @@ public function getCreatePrimaryKeySQL(Index $index, $table)
      */
     protected function getCreateColumnCommentSQL($tableName, $columnName, $comment)
```
mvorisek, line 361: I belive, there is the same issue with `getAlterColumnCommentSQL`, `getDropColumnCommentSQL`, `getCommentOnTableSQL`

-> we should update out functional tests to test at least create/alter/drop comment with quote table name for all platforms

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -360,24 +360,20 @@ public function getCreatePrimaryKeySQL(Index $index, $table)
...
-            $tableSQL  = $this->quoteStringLiteral($tableName);
+            [$schemaName, $tableName] = explode('.', $tableName, 2);
         }
 
         return $this->getAddExtendedPropertySQL(
             'MS_Description',
             $comment,
             'SCHEMA',
-            $schemaSQL,
+            $this->quoteStringLiteral(preg_replace('~^\[|\]$~s', '', $schemaName)),
```
mvorisek, line 372: from https://medium.com/@selcukkilinc23/dynamic-sql-samples-134c0b8851d
> What is sysname?
> 
> Note: The sysname data type is used for table columns, variables, and stored procedure parameters that store object names. It is basically the same as using nvarchar(128) NOT NULL.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -230,12 +230,11 @@ public function getDropIndexSQL($index, $table = null)
 
         return sprintf(
             <<<SQL
-IF EXISTS (SELECT * FROM sysobjects WHERE name = '%s')
-    ALTER TABLE %s DROP CONSTRAINT %s
-ELSE
-    DROP INDEX %s ON %s
-SQL
-            ,
+            IF EXISTS (SELECT * FROM sys.sysobjects WHERE name = '%s')
```
mvorisek: normalize all system FROMs to use `sys.` prefix

mvorisek:
@greg0ire @morozov can we merge?

greg0ire:
TL;DR of the issue, the blog post and your diff please?
Also, there are no tests, why is that?

mvorisek:
@greg0ire https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-dropextendedproperty-transact-sql?view=sql-server-ver15 shows the `sp_addextendedproperty` accepts `sysname` datatype which is nothing else than a string datatype - therefore the link

everything is covered by static tests - you want one functional test for `getCreateColumnCommentSQL`?

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL4283Test.php
```diff
@@ -0,0 +1,142 @@
...
+use Doctrine\DBAL\Schema\ColumnDiff;
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Schema\SQLServerSchemaManager;
+use Doctrine\DBAL\Schema\TableDiff;
+use Doctrine\DBAL\Types\Type;
+use Doctrine\Tests\DbalFunctionalTestCase;
+
+use function sprintf;
+
+class DBAL4283Test extends DbalFunctionalTestCase
```
greg0ire, line 16: If that was really a functional test it would not only work for some jobs of the build matrix, and there would be something to skip it in other jobs. IMO, the test should demonstrate that the bug described at #4283 , which occurs when executing an SQL query does not happen anymore. This means you will have to execute that query at some point.

mvorisek: yes - this is test that I merged from another PR

The issue is that DBAL accepts unquoted/quoted names and that is not tested extensively enought across databases.

I replace it completely with cross-vendor functional test. I also verified, that the original issue is fixed by reverting the fix.

mvorisek:
@greg0ire @morozov one last issue - tests with OCI are failing - can you please help with fixing this issue?

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL4283Test.php
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * @return iterable<string[]>
+     */
+    public function columnNameProvider(): iterable
+    {
+        return [
+            ['basic'],
+            [$this->doubleQuoteName('basic')],
+            [$this->doubleQuoteName('and')],
```
mvorisek, line 84: reserved word...

mvorisek:
Function test is testing all platforms, however Oracle and PostgreSQL platform does not support quoted names completely and thus unsupported cases are skipped. I think it will require quite a lot of work and should be fixed in different PR anyway.

@greg0ire @morozov please review, the targetted MSSQL bug is fully fixed.

greg0ire:
The test should be in the same commit as the code it tests, because it's unlikely you will revert that commit and not the other. The commit message starts with "Fix". That means it should also say:
- what is wrong with the current behavior;
- what you are doing;
- why that fixes the issue.

Please also rebase on `2.12.x`, as `2.11.x` is no longer maintained.

mvorisek:
> 
> 
> The test should be in the same commit as the code it tests, because it's unlikely you will revert that commit and not the other. The commit message starts with "Fix". That means it should also say:
> 
>     * what is wrong with the current behavior;
> 
>     * what you are doing;
> 
>     * why that fixes the issue.

done

also, I would propose to reconfigure this repo to merge PRs by "allow squash merge" only, see GH repo config:

image

I do not feel ok to loose all devel history by squashing it all the time manually, also, much easier to review and team work

> 
> Please also rebase on `2.12.x`, as `2.11.x` is no longer maintained.

done incl. target branch update

greg0ire:
Todo:

- [x] The test should be in the same commit as the code it tests
- [ ] Include something describing what is wrong with the current behavior in the commit message body
- [ ] Describe what you are doing in the commit message subject
- [ ] Describe why that fixes the issue in the commit message body
- [x] rebase on 2.12.x

> also, I would propose to reconfigure this repo to merge PRs by "allow squash merge" only, see GH repo config:

Sorry, sometimes we merge PRs without squashing, for instance if you fixed a CS issue in the same file you edit, but the CS issue is completely unrelated to your change, I would still accept it, but I'd rather have it is in a separate commit, that way if your PR is wrong, we can still revert one commit and keep the other one.

mvorisek:
@greg0ire please take over the todo

image

you should have write access to the source branch

greg0ire:
@mvorisek I'm not asking because I want to nag you, I'm asking because I don't understand everything about this PR, and having a good commit message will make my review easier. I don't want to go through 3 links and reverse engineer an explanation from your diff, as my inbox is full of PRs that have better explanations, I'm choosing to give them the priority. Here is an example of a good commit if you need one: https://github.com/doctrine/dbal/commit/9346bbd16d0fdb6c9d821bfa6e8b41bd8f5581c1

mvorisek:
image

I understand, but I really has nothing more to add - I fixed one bug.

greg0ire:
It's incredible how you are able to write tons of codes and tests and link to blog posts and issues and post screenshots with nice red squares but flat out give up when it comes to explaining the issue, and explaining how and why your fix works in a few sentences. Maybe the issue is that you think I know SQL Server? I have never used that database error. Let me show you what I think I understood, then you can correct me and then I will be able to build a good commit message from our discussion. I will use numbered bullets so that it's easy for you to comment.

1. The issue is that when a table contains a special character, Doctrine quotes it, and that results in SQL Server being unabled to find the table… because it looks for a table that literally contains quotes? That does not seem very clear already, the error message is `Extended properties are not permitted on 'dbo.[the-table].chairs'`, not `Extended properties are not permitted on 'dbo."[the-table]".chairs'`. I'm unfamiliar with SQL Server, so I'll just assume having square brackets is normal, do tell me if it has to do with the issue.
2. SQL Server has a special data type called `sysname` that it uses among other things, for table names.
3. Your PR prefixes all `sysobjects` with `sys.`, maybe that's part of the fix? I can't see how.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -1687,31 +1703,34 @@ private function generateIdentifierName($identifier)
...
-SQL
-            ,
+            SELECT
+              p.value AS [table_comment]
+            FROM
+              sys.tables AS tbl
+              INNER JOIN sys.extended_properties AS p ON p.major_id=tbl.object_id AND p.minor_id=0 AND p.class=1
+            WHERE
+              (tbl.name=N%s and SCHEMA_NAME(tbl.schema_id)=N'dbo' and p.name=N'MS_Description')
+            SQL,
```
greg0ire: This change is welcome, but in a separate commit that explains why you do it. Are there other changes in this PR that are not related to the fix? If yes, please move them to a separate commit, it will make it easier for me to understand. Ideally, you should have 1 commit with only what is strictly necessary to fix the issue + the tests, and all the rest should be in separate commits.

tests/Doctrine/Tests/DBAL/Platforms/AbstractSQLServerPlatformTestCase.php
```diff
@@ -656,7 +656,7 @@ public function testCreateTableWithSchemaColumnComments(): void
         $expectedSql = [
             'CREATE TABLE testschema.test (id INT NOT NULL, PRIMARY KEY (id))',
             "EXEC sp_addextendedproperty N'MS_Description', N'This is a comment', "
-                . "N'SCHEMA', 'testschema', N'TABLE', 'test', N'COLUMN', id",
+                . "N'SCHEMA', 'testschema', N'TABLE', 'test', N'COLUMN', 'id'",
```
greg0ire, line 659: This is also very suspicious. If your PR is a fix, why are you changing existing tests? To be clear, it's fine if your PR is not only a fix, but if that's the case, you need to have several commits. Use `git add -p`.

mvorisek: PR fixes escaping, this is consistent with the title itself. Tests that tested generated SQLs were adjusted, of course, this is needed :)

greg0ire: I thought things like table and column names should not be be quoted at all, otherwise they are no longer of the `sysname` datatype. This seems to do the opposite. I'm confused.

mvorisek: great - if name will contain `,`, what will happen?

I just added a test to proove, that we need it this way.

Anyway, `sysname` is string datatype - mentioned in that blog post - that is specific in one thing only - defined/optimized to have the same properties as underlaying definition type of identifier, ie. it allows all the characters, supported lengh etc.

mvorisek: I can also very easily proove, that the whole DBAL is super buggy when I add test with `.` (I already tried and I just notice how much hot coded this project actually is - check yourself, search for `explode('.', ` - this is of course wrong when does split names like `[table.stilltable.name]` / `` `table.stilltable.name` `` ;-))

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
mvorisek: @greg0ire you are great guy, I am maintainer of several projects as well, ok, you have twice the yearly contributions I have, but please, if this is needed, please do it so. We are not inventing a rocket and as you can see in the functional test, Oracle and PostgreSQL platforms does not even meet the standard/abstract functionality (accept quoted identifier) - for these reasons, I left this for someone else to fix in a different/larger PR.

Also a reson, why I created https://github.com/doctrine/dbal/issues/4385 5 hours ago 🚀

mvorisek:
@greg0ire can this be merged?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
greg0ire: > if this is needed, please do it so.

This part is needed, but maybe there are other changes in this PR that are note related to the fix either? If not do you understand why I would like to separate those? It will make it easier for me to understand which part of the PR actually fixes the bug you say you are fixing. Again, use `git add -p`

tests/Doctrine/Tests/DBAL/Platforms/AbstractSQLServerPlatformTestCase.php
greg0ire: > great - if name will contain ,, what will happen?

Something bad I suppose… I don't have much time to play riddles. If the solution to the issue is quoting everything that is conceivably quotable, then you should say so in your commit message explicitly. having `id` instead of `'id'` is not broken, there is nothing to fix here. Fixing issue #4283 is a _side effect_ of something you are doing, but for some reason you are refusing to explain what it is you are doing precisely in your commit message.

greg0ire:
> @greg0ire can this be merged?

I try hard not to merge things I don't fully understand. By the looks of this PR you seem to think that:
1. I am very clever.
2. I know the DBAL like the back of my hand.
3. I am familiar with SQL Server.

All three assertions are false, which means you will have to make commits that are crystal clear, which is better for the DBAL's commit history anyway.

tests/Doctrine/Tests/DBAL/Platforms/AbstractSQLServerPlatformTestCase.php
mvorisek: You missed "when used as string literal" from title - this is fix for #4283.

mvorisek:
> having id instead of 'id' is not broken, there is nothing to fix here.

this is super clean fix, please, read title again!

greg0ire:
Ok let's read it together:
Fix identifier escaping => ok sounds greate, go on
when used as string literal => when what is used as a string literal? The escaping? How can escaping be used a string literal? that does not make sense at all…

mvorisek:
makes - typical example is SELECT query from MySQL INFORMATION_SCHEMA.COLUMNS

mvorisek:
@greg0ire so is it clear to you?

`` `name` `` - identifier `name escaped in SQL as native identifier - used, when identifier ONLY is expected in SQL grammar
`'name'` - identifier `name` escaped in SQL as string literal - used, when passed as value in SQL grammer, like to query it from information schema tables, pass to function, trigger, ...

greg0ire:
I think I might start to understand… your PR makes sure backticks or whatever the equivalent for SQL Server is (what is it? I have no idea, please tell me.) when an SQL object (table, columns) is expected, while using a string literal when the _name_ of an SQL object is expected, is that it?

mvorisek:
@greg0ire yes

SQL Server (Sever with capital S) === T-SQL queryable server, product by Microsoft

SQL object (table, columns) === identifier

that's it!

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
greg0ire: Ok, why not, but what does that change? If you confirm to me that it is not directly related to the fix, I will move it to a separate commit.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -361,21 +361,18 @@ public function getCreatePrimaryKeySQL(Index $index, $table)
     protected function getCreateColumnCommentSQL($tableName, $columnName, $comment)
     {
         if (strpos($tableName, '.') !== false) {
-            [$schemaSQL, $tableSQL] = explode('.', $tableName);
-            $schemaSQL              = $this->quoteStringLiteral($schemaSQL);
-            $tableSQL               = $this->quoteStringLiteral($tableSQL);
+            [$schemaName, $tableName] = explode('.', $tableName, 2);
```
greg0ire, line 364: That looks related to the fix: we don't quote the identifier because it should not be passed as a string literal

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -794,37 +785,48 @@ protected function getRenameIndexSQL($oldIndexName, Index $index, $tableName)
         ];
     }
 
+    private function quoteSingleIdentifierAsStringLiteral(string $levelName): string
+    {
+        return $this->quoteStringLiteral(preg_replace('~^\[|\]$~s', '', $levelName));
```
greg0ire, line 793: So this removes brackets. Please explain why, it's the first time I see brackets in SQL (is that specific to the T-SQL thing you mentioned earlier?)

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -794,37 +785,48 @@ protected function getRenameIndexSQL($oldIndexName, Index $index, $tableName)
...
-        $level1Name = null,
-        $level2Type = null,
-        $level2Name = null
+        $value,
+        string $level0Type,
+        string $level0Name,
+        string $level1Type,
+        string $level1Name,
+        ?string $level2Type = null,
+        ?string $level2Name = null
```
greg0ire: Good change too, but not related directly to the fix I think, that change could be done independently, right?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -794,37 +785,48 @@ protected function getRenameIndexSQL($oldIndexName, Index $index, $tableName)
...
-            'N' . $this->quoteStringLiteral((string) $level2Type) . ', ' . $level2Name;
+        return 'EXEC sp_addextendedproperty'
+            . ' N' . $this->quoteStringLiteral($name) . ', N' . $this->quoteStringLiteral((string) $value)
+            . ', N' . $this->quoteStringLiteral($level0Type)
+            . ', ' . $this->quoteSingleIdentifierAsStringLiteral($level0Name)
+            . ', N' . $this->quoteStringLiteral($level1Type)
+            . ', ' . $this->quoteSingleIdentifierAsStringLiteral($level1Name)
+            . ($level2Type !== null || $level2Name !== null
+                ? ', N' . $this->quoteStringLiteral((string) $level2Type)
+                  . ', ' . $this->quoteSingleIdentifierAsStringLiteral((string) $level2Name)
```
greg0ire, line 830: The only line related to the fix are the ones with `quoteSingleIdentifierAsStringLiteral`, right?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -906,7 +920,9 @@ public function getListTablesSQL()
     {
         // "sysdiagrams" table must be ignored as it's internal SQL Server table for Database Diagrams
         // Category 2 must be ignored as it is "MS SQL Server 'pseudo-system' object[s]" for replication
-        return "SELECT name FROM sysobjects WHERE type = 'U' AND name != 'sysdiagrams' AND category != 2 ORDER BY name";
+        return 'SELECT name FROM sys.sysobjects'
+            . ' WHERE type = \'U\' AND name != \'sysdiagrams\' AND category != 2'
+            . ' ORDER BY name';
```
greg0ire: Ok so now we have backslashes and concatenation. Please use nowdoc instead.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
mvorisek: yes

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
mvorisek: yes - `` ` `` in MySQL, `[` (and `]`) in T-SQL, just different SQL grammar

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
mvorisek: feel free to do so - but it is highly related with the fixed issue #

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
mvorisek: I think `getCommentOnTableSQL` at least as well

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
greg0ire: Ok, this part looks super ugly and fragile, do we really need it? Why are there brackets in the name in the first place?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
mvorisek: fine, fix it

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
mvorisek: Strictly needed, because name can already be escaped, because of DBAL escaping approach now which allows it

greg0ire: You mean the DBAL adds those brackets somewhere?

mvorisek: no in general - DBAL allows name to be specified in unquoted and quoted string form, related with https://github.com/doctrine/dbal/issues/4385, which we will definitely not solve in this PR

mvorisek:
@greg0ire can you please adjust the commits to satisfy your needs and merge today? Thanks.

greg0ire:
When I see a mention of "ASAP", today, etc. in a FOSS context I tend to put the offending person at the bottom of my priority list. If you are not ready to do the changes I asked yourself, then you should at least be patient about it. I will go through my inbox and if I have time and energy left after that, then maybe I will do the commits the way they should be. If you want things to go faster, you can work yourself towards that.

metaturso:
@greg0ire I'll try to answer your questions, chap. 

## What's the problem?

Doctrine [gives programmers the ability to escape a table whose name is either a reserved SQL keyword or an invalid object name](https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/reference/basic-mapping.html#quoting-reserved-words) to avoid generating invalid DDL (and DML) statements. 

When Doctrine runs on Microsoft SQL Server it uses so called "Extended Properties" to declare extra metadata on columns. For example, to indicate that a column contains JSON data to be decoded during hydration.

Doctrine currently fails to declare (amend and remove) extended properties during migrations because the generated DDL statement references a bogus table name called _"[group]"_ instead of the table called _"group"_ (for example.) 

Keeping in mind the following differences:

1. `group` is a keyword part of the standard SQL grammar.
2. `[group]` identifies a database object called `group`. 
3. `'[group]'` is just a word wrapped in square brackets.

The problem is that Doctrine is generating a query that should contain 2  but contains 3 instead.

## Is this a minimal fix?

@mvorisek I don't know that this MR should be merged in its current state. It's also missing the requested integration tests.

It may be possible to fix this bug completely within the confines of the `SqlServerPlatform::get*ExtendedPropertySQL()` methods or their direct caller by escaping the identifier only when the original name was wrapped in Doctrine's backticks.

I wouldn't feel comfortable taking this current patch forward because it's too pervasive and could negatively impact backward-compatibility.

## Is checking for already escaped identifiers the right approach?

I think not. Doctrine passes table and column names around as plain strings, at some point it's forced to inspect the string to check whether it contains an escaped identifier or not.

Personally I'd rather see Doctrine differentiate between quoted and unquoted identifiers using data types. For example by passing around objects like `QuotedName("user")` instead of strings like `` "`user`" ``.

Whether using `sysname` is a necessary condition to fix this... I'm too thick to say. From my experience an escaped identifier like `[table].[column]` works just fine, then again I am not one who publishes on Medium.

metaturso:
@mvorisek you did a ton of work in this MR and added an integration test (I apologise for failing to notice before writing my previous comment.) You even spotted and fixed the problem in a couple data dictionary query methods (`getListTableMetadataSQL` and `getCommentOnTableSQL` IIRC,) that's definitely better now!

As reviewers much wiser than myself have pointed out, however, there's a number of changes in this MR that aren't immediately needed to address the original problem. This work should focus on the single issue at hand, which is that _Doctrine's `SqlServerPlatform` class generates the wrong DDL to manage and query Extended Properties for tables having "invalid" names or living in a non-standard schema_ (#4283,) and leave the bigger headache posed by inconsistent quoting throughout Doctrine's internal layers for later (#4357.)

I think I understand your frustration. It's evident that you've done your homework and already have this all figured out ticketed in progression and that you're ready to carry this plan out, maybe even get a t-shirt rewarded in the process. 

Alas Doctrine isn't a toy project and hasty changes cost money in the real world, don't get bent out of shape because you have to explain the root cause of a problem three different ways and justify the presence of superfluous code.

If you've lost interest and would rather somebody else completed this for you, may I step in to help you distil the changes down into a minimal form and groom the commit/patch history as requested? It would be a real shame to see all this effort go to waste :( at the end of the day the work is pretty much all there!

greg0ire:
Thanks @metaturso for the very thorough explanations, it helps a lot!

> may I step in to help you distil the changes down into a minimal form and groom the commit/patch history as requested? It would be a real shame to see all this effort go to waste :( at the end of the day the work is pretty much all there!

That would be great! And thanks for expressing what I didn't: although this is clearly a great amount of work, it's hard for me to merge it confidently. The amount of downloads this package has warrants a lot of that reluctance, and if you were to indeed groom that commit history, that would help a great deal!

> Is checking against already escaped identifiers the right approach?

I didn't understand a lot in that PR but I'm sure that the use of `preg_replace` here is only a symptom of a bigger issue indeed. Fully agree, this is not the right approach.

mvorisek:
@metaturso I very strongly belive in one - identifiers, that needs to be passed as/to string must be firstly splitted/unquoted. I also belive, that I coded very powerfull functional test (with reserved keywords and characters like dots or dashes).

Everything else is up too the maintainers. If you personally think you have a simpler solution and you want to help with the solution, feel free to do so.

mvorisek:
@greg0ire above, I have explained why we need to use string quotes for `EXEC sp_addextendedproperty`. Can you please review this PR so we can integrate it quickly? Thanks.

greg0ire:
Sorry, I don't have the required knowledge and practice with SQL Server, or the required energy to review this properly. You will have to find another maintainer.

mvorisek:
@greg0ire can you please ping someone with the knowledge?

greg0ire:
Surely you can do that yourself.

mvorisek:
@greg0ire I am rebased this PR on the latest `2.12.x` branch and updated the individual commits to reflect the exact changes made. Please review.

mvorisek:
I rebased this PR on the latest `2.12.x` branch and updated the individual commits to reflect the exact changes made. @greg0ire please review.

bartfeenstra:
I can't comment on either Doctrine or SQL Server, but I can confirm that PR branch as it exists now does fix the problem in the context of generated migrations, as table names that are reserved keywords are no longer escaped when they are used as strings.

mvorisek:
@greg0ire I done my bests to fix this issue, cover it with extensive testing and document it with clear description and commit messages. Is there anything else I can do for this PR to be merged? Thanks.

greg0ire:
Yes:

> You will have to find another maintainer.

mvorisek:
@yitam, I am reaching you as you are the major contributor to the official `microsoft/msphpsql`, can you please take a quick look at this PR and submit your feedback?

@morozov Please review as well. Thanks!

yitam:
I'm not familiar with Doctrine. For SQL Server indeed the square brackets are often used to delimit [identifiers](https://docs.microsoft.com/sql/relational-databases/databases/database-identifiers), which is necessary if the column or table name is a reserved keyword or contains special characters such as a space or hyphen, but there are other choices. Please check [QUOTENAME (Transact-SQL)](https://docs.microsoft.com/sql/t-sql/functions/quotename-transact-sql) for more details.

mvorisek:
@metaturso can you please submit your review on this PR?

mvorisek:
> 
> 
> Yes:
> 
> > You will have to find another maintainer.

What else can I do to get this PR merged? Btw. it is a bug fix incl. full functional tests prooving the fix.

mvorisek:
> 
> 
> Yes:
> 
> > You will have to find another maintainer.

What else can I do to get this PR merged? Btw. it is a bug fix incl. full functional tests prooving the fix.

ping @greg0ire @morozov

mvorisek:
> 
> 
> Yes:
> 
> > You will have to find another maintainer.

What else can I do to get this PR merged? Btw. it is a bug fix incl. full functional tests prooving the fix.

ping @greg0ire @morozov @beberlei

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
morozov: `2` assumes that the schema name cannot contain dots but the table can. Why?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -797,37 +789,48 @@ protected function getRenameIndexSQL($oldIndexName, Index $index, $tableName)
...
     public function getAddExtendedPropertySQL(
         $name,
-        $value = null,
-        $level0Type = null,
-        $level0Name = null,
-        $level1Type = null,
-        $level1Name = null,
-        $level2Type = null,
-        $level2Name = null
+        $value,
```
morozov: This is a breaking change.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -797,37 +789,48 @@ protected function getRenameIndexSQL($oldIndexName, Index $index, $tableName)
...
-            'N' . $this->quoteStringLiteral((string) $level0Type) . ', ' . $level0Name . ', ' .
-            'N' . $this->quoteStringLiteral((string) $level1Type) . ', ' . $level1Name . ', ' .
-            'N' . $this->quoteStringLiteral((string) $level2Type) . ', ' . $level2Name;
+        return 'EXEC sp_addextendedproperty'
+            . ' N' . $this->quoteStringLiteral($name) . ', N' . $this->quoteStringLiteral((string) $value)
+            . ', N' . $this->quoteStringLiteral($level0Type)
+            . ', ' . $this->quoteSingleIdentifierAsStringLiteral($level0Name)
+            . ', N' . $this->quoteStringLiteral($level1Type)
+            . ', ' . $this->quoteSingleIdentifierAsStringLiteral($level1Name)
+            . ($level2Type !== null || $level2Name !== null
```
morozov, line 828: Is it a valid case when one of the values is null but the other is not? I assume the pairs of arguments are optional but the arguments within a pair aren't.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -836,29 +839,35 @@ public function getAddExtendedPropertySQL(
...
      */
     public function getDropExtendedPropertySQL(
         $name,
-        $level0Type = null,
-        $level0Name = null,
-        $level1Type = null,
-        $level1Name = null,
-        $level2Type = null,
-        $level2Name = null
+        string $level0Type,
```
morozov: This is a breaking change.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -867,31 +876,37 @@ public function getDropExtendedPropertySQL(
...
     public function getUpdateExtendedPropertySQL(
         $name,
-        $value = null,
-        $level0Type = null,
-        $level0Name = null,
-        $level1Type = null,
-        $level1Name = null,
-        $level2Type = null,
-        $level2Name = null
+        $value,
```
morozov: This is a breaking change.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
mvorisek: see https://3v4l.org/UrgeT - this is only improvement of the original `explode`, which can return more than 2 result (and the 3rd and later result was silently discarded previously)

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
mvorisek: Previously, `null` was allowed for every value, thus the original behaviour is kept.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
mvorisek: Is this breaking change welcomed?

morozov: Not in a patch release. Please compare with `4.0.x`, there must be some similar changes there.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
morozov: Correct. But if this is being modified, we need to understand what the expected usage of this API is and prevent unexpected ones. E.g. if we believe it shouldn't be used by passing only one argument from a pair, we can deprecate it and forbid later.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
morozov: The question still is: why do you think the previously discarded part after the dot should be attributed to the table name?

mvorisek: Because the data of "the previously discarded part after the dot" come from `$tableName` :)

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
mvorisek: Discussed in https://github.com/doctrine/dbal/pull/4360#discussion_r602897895 , not changing it in this PR.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
mvorisek: Discussed in https://github.com/doctrine/dbal/pull/4360#discussion_r602897895 , not changing it in this PR.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
mvorisek: Keeping original behaviour, not changing it in this PR.

mvorisek:
@morozov the feedback has been addressed, it there anything left?

morozov (commented):
In my opinion, this whole magic about smashing the schema name and the table together and then exploding/parsing with regular expressions is absolutely unreliable.

What if the schema name is "example.com" and the table name is "inventory"?

A proper API must operate with object (schema, table) names and quote, and concatenate them while building SQL. The API works that worse the other way around may pass the tests that cover the happy but won't work in real cases.

mvorisek:
@morozov Please tell me where introduced newly, this PR fixes a bug related with wrong column names escaping for `EXEC sp_addextendedproperty` prodedure. Exploding name by `.` is the original behaviour.

morozov:
I see that this PR relies on the reworked implementation using `explode()` which was wrong and is still wrong. I can't tell if this is relevant to the fix because this PR:

1. Did contain the irrelevant previously (argument type changes in `getAddExtendedPropertySQL()`).
2. Does contain the corresponding type casting which is no longer necessary.

It's challenging to review such changes since it takes extra effort to get to their essence.

Apart from that, I find the way you behave in the community hard to deal with:
1. You propose changes that have questionable value and/or take unnecessarily much time and energy to discuss ([ref](https://github.com/doctrine/dbal/pulls?q=is%3Apr+is%3Aclosed+author%3Amvorisek)).
2. Given the contributions above, I don't believe you're credible to work on such complex code changes.
3. You create an unnecessary urgency to review your changes (https://github.com/doctrine/dbal/pull/4276#discussion_r496172368) and tag people that has no relation to the work you're doing (https://github.com/doctrine/dbal/pull/4360#issuecomment-776597926).

With that said, I'm not personally interested in reviewing this. You can continue working with other DBAL maintainers on getting your changes approved.

mvorisek:
@morozov I have added functional tests covering reserved keywords in column names for all database platforms (first commit).

I belive I fixed the issue cleanly and in a minimalistic way. If you think the issue can be fixed better (and still fully passing added tests), please advise.

To my crediatability. I understand you may "rate me" by several closed PRs into this repo, but I am proud to say this is pure coincidence. I am strong contributor to another backend repos and database abstraction libs like `atk4/dsql`, `atk4/data` in which I actually spotted the issue.

morozov:
The DBAL currently has very limited support for quoting identifiers (https://github.com/doctrine/dbal/issues/4357). Most importantly, there's no distinction at the API level between object names (e.g. `foo-bar`) and the SQL representing them: `[foo-bar]`, `"foo-bar"`, `` `foo-bar` ``etc. Even when an identifier is passed as `` `foo-bar` ``, it's not clear whether it's `foo-bar` represented as SQL or the backticks are part of the name (which is also valid).

I don't see much point in complicating the implementation of the API that was not supposed to be used for this purpose in the first place.

mvorisek:
@morozov Your reaction seems brutal to me. With a lot of efford (several iterations of feedback, 10+ hours spent) I was able to fix the issue with very good functional testing FOR EVERY PLATFORM.

> I don't see much point in complicating the implementation of the API 

What is wrong with fixed ~80 lines (+ added tests, + normalized quotes in existing tests)?

Here is example of the current issue: https://github.com/atk4/data/runs/2261267667#step:12:24

beberlei:
These kind of changes to quoting fundamentals are very high risk for Doctrine. In conjunction with our limited understanding of SQLServer platform, this PR is something we all hesitate to merge, defensive maintenance is critical for a database library.

Sorry for getting to the conclusion so late, i certainly have been there myself after having to revert a patch on sqlite foreign keys that took weeks to finish last year for example.

As a point in feedback, this kind of deep change to a driver needs careful research and your PR description leaves a lot open wirh respect to quoting original documentation, resources explaining the issue with examples and your own evaluation of them w.r.t Doctrine. We have asked for this several times and we habe also invested hours reviewing this change but still far away from 100% understanding.

mvorisek:
I am still convinced this is a bug and it should be fixed.

Guys, @greg0ire , @morozov, @beberlei , let me discuss the possible impacts:

Pros:
- issue #4283 is fixed (ie. names with unlisted keywords, already quoted names)
- all existing usage continues to work (no single changed test usage)
- new test cases work with all platforms and full table/column lifecycle

To consider:
- expected sql queries are changed to a different quoting (see https://github.com/doctrine/dbal/pull/4360/files#diff-037ab13274497cf54e95b2386c69077b5a7838858d83085cb349a0b8a5b5e6b0R797 , all arguments except the last one are already using the proposed escaping `'`)
- `getCreateColumnCommentSQL` (and `getAlter...` variants) methods are now stricter:
  - `explode` now does not ignore name with more dot symbols (I belive error is better than silent ignore, if not, this change is definitely not required, I can remove the `2` explode limit)
  - `getAddExtendedPropertySQL` previously accepted splitted but already SQL escaped name parts (but comment was/is accepted unescaped, now normalized to accept all parts unescaped and escaping is done inside - always, thus stronger)
  - no other behaviour changed

Cons:
- except discussed cases above I do not see any real BC break

Can we agree on these points and is there anything else to discuss/address?

PR #4361, opened 2020-10-19
Test all extensions with PHP8

greg0ire:
Jobs that use other versions of PHP are temporarily removed so that it's
easier to focus and iterate.

Failures:
- [x] [OCI8 installs successfully, but might contain API changes?](https://github.com/doctrine/dbal/pull/4361/checks?check_run_id=1277115408#step:7:106)
- [x] [sqlsrv fails to install](https://github.com/doctrine/dbal/pull/4361/checks?check_run_id=1277115981#step:4:18)

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -76,6 +76,7 @@ jobs:
       matrix:
         php-version:
           - "7.4"
+          - "8.0"
 
     services:
       oracle:
@@ -125,6 +126,7 @@ jobs:
       matrix:
         php-version:
           - "7.4"
+          - "8.0"
 
     services:
       oracle:
@@ -315,7 +317,6 @@ jobs:
       matrix:
         php-version:
           - "7.4"
-          - "8.0"
         mysql-version:
           - "5.7"
           - "8.0"
@@ -344,6 +345,16 @@ jobs:
             php-version: "7.4"
             mysql-version: "8.0"
             extension: "mysqli"
+          - php-version: "8.0"
+            mysql-version: "8.0"
+            extension: "mysqli"
+            custom-entrypoint: >-
+              --entrypoint sh mysql:8 -c "exec docker-entrypoint.sh mysqld --default-authentication-plugin=mysql_native_password"
+          - php-version: "8.0"
+            mysql-version: "8.0"
+            extension: "pdo_mysql"
+            custom-entrypoint: >-
+              --entrypoint sh mysql:8 -c "exec docker-entrypoint.sh mysqld --default-authentication-plugin=mysql_native_password"
 
     services:
       mysql:
@@ -405,6 +416,7 @@ jobs:
         php-version:
           - "7.3"
           - "7.4"
+          - "8.0"
         extension:
           - "sqlsrv"
           - "pdo_sqlsrv"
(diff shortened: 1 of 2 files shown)


greg0ire:
[The releases page for `sqlsrv`](https://github.com/Microsoft/msphpsql/releases) shows PHP8 support in version `5.9.0-preview1`, let's try that one.

greg0ire:
Also, let's remove all the successful jobs.

morozov:
> Message was: "oci_execute(): Argument #1 ($statement) must be of type resource, null given"

It looks like `php_oci8` just enforced its argument types as well as many other extensions.

greg0ire:
The solution for `sqlsrv` works. We will just have to tweak the configuration so that it picks the right version of `sqlsrv` for each version of PHP, or maybe we can use 5.9.0 everywhere, dunno…

morozov:
> maybe we can use 5.9.0 everywhere

This is fine.

greg0ire:
Link to the faulty call: https://github.com/doctrine/dbal/blob/b407f8b064440009194175318d5754d6e1dfde43/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php#L405

Note the error suppression operator lol

greg0ire:
>     Message was: "oci_execute(): Argument #​1 ($statement) must be of type resource, null given"

Love how it tells you the name of the argument to avoid confusion because 0 vs 1 numbering :star_struck:

morozov:
> Note the error suppression operator lol

It's not a problem per se if the API always returns false in the case of an error and we implement proper handling. Although it is a problem in cases like this (PHP-level error).

greg0ire:
Yeah, it would be great to have an exception instead of an error…

morozov:
> I tried fooling `oci_execute`, but it looks like the type checks are pretty thorough now…

I'd rather have this test refactored. IIRC, that's the one that had to be reworked many times just because of how creepy it is.

greg0ire:
I just looked at the comment and that part of the test is irrelevant anyway, asserting the type of the exception is not relevant anyway.

morozov:
See where it ended up: https://github.com/doctrine/dbal/pull/3820.

greg0ire:
Oh right, RIP `OCI8StatementTest.php`.

> I'd rather have this test refactored. 

I pushed a commit, but I don't think that's what you mean by "refactored" :P

morozov:
> I don't think that's what you mean by "refactored" :P

No, just skip it on PHP 8.

greg0ire:
The OCI8 job seems stuck :/

greg0ire:
Ok so now we have 78 checks :sweat_smile:

morozov:
> Ok so now we have 78 checks 😅

This is overkill. We just need to test each driver with the latest platform version on PHP. There's no much point to test all platform versions with any version than the stable (currently 7.4).

morozov:
Just in case, I'm not concerned about the number of jobs as such (as long as GA is fine with it). But I am concerned about the build stability: there's a higher chance that one of the 78 jobs will decide to fail for no reason, and right now (unlike Travis and AppVeyor), GA only allows to restart all jobs (not just the failed ones.)

greg0ire:
Oh I forgot that there were already tests defined under `include` for some platforms. Fixed.

greg0ire:
Down to 59, we can apply your ideas on 2.11.x to reduce that number even further.

morozov (approved):

morozov:
By the way, @BenMorel put together a [script](https://gist.github.com/BenMorel/67e5d94c864af9b9ae1d3a1db8010b98) some time ago that I believe still can be used as a reference for the algorithm to generate the matrix.

PR #4362, opened 2020-10-19
Adds exception thrown by execute() method

toby-griffiths:
|      Q       |   A
|------------- | -----------
| Type         | improvement (minor)
| BC Break     | no
| Fixed issues | n/a

#### Summary

My IDE was telling me I was trying to catch a non-existent exception, however it's clear by digging down a level that the Exception is throw by this class, so this simply annotates that.

Perhaps this should catch it and throw a QueryException?  Happy to update if it should be.  Please just let me know.

Also, I raised this against the 2.12 branch which is the latest branch I could find this file in, as I'm currently using 2.11.1.  I hope this was OK?

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -4,6 +4,7 @@ namespace Doctrine\DBAL\Query;
 
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Driver\ResultStatement;
+use Doctrine\DBAL\Exception;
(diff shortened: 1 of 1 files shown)


SenseException (dismissed):

greg0ire:
I rebased because new jobs were added to the pipeline, and were never run on this PR. Also, I want to know if the list of required jobs is accurate, there might be some extra required jobs, or some jobs that should be required but are not.

morozov (approved):

greg0ire:
Thanks @toby-griffiths !

Issue #4359, opened 2020-10-19
Bug with PostgreSQL schema manager after update on 2.11.2


andreybolonin:
<img width="361" alt="Screen Shot 2020-10-19 at 10 39 12 AM" src="https://user-images.githubusercontent.com/2576509/96415484-5a07ff80-11f7-11eb-9b11-78bcc69bd318.png">

sasule:
I have the same problem.
My configuration: Postgre13 and php 7.4.11.

digitv:
I confirm the same behaior on our app (Laravel 6) deploy with bitbucket pipepines.
PHP 7.4.5
PostgreSQL10

shEtendard:
```
assert(preg_match(
            '/FOREIGN KEY \((.+)\) REFERENCES (.+)\((.+)\)/',
            $tableForeignKey['condef'],
            $values
        ) !== 0);
```
https://www.php.net/manual/en/function.preg-match.php
preg_match() returns 1 if the pattern matches given subject, 0 if it does not, or FALSE if an error occurred.

agustingomes:
I can confirm the same error in the [PostgreSqlSchemaManager.php](https://github.com/doctrine/dbal/blob/b88a45cbc395e850b2d958dcd7b9d49d09ff23e0/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php#L165).

In my case it was with php 7.4.11 and Postgres 10.6

SindlaXYZ:
Same error after 2.11.2 update (from 2.11.1).

PHP 7.4.11 & PostgreSQL 12.4

pelikan7:
The code inside assert() does not run at all, leaving $values as undefined variable

bensquire:
Can also confirm the problem, 2.4.12 from 2.4.11, PostgresSQL, PHP 7.4.11.

jarofgreen:
Confirmed, downgrading to 2.11.1 fixes it fine.

In my case I got it in Symfony 5.1 when running:

    ./bin/console doctrine:schema:drop --force --quiet

morozov:
The issue is reproducible via the existing test suite with disabled assertions (`zend.assertions=-1`):
```
$ phpunit -c pdo-pgsql.phpunit.xml tests/Doctrine/Tests/DBAL/Functional/Schema/PostgreSqlSchemaManagerTest.php

PHPUnit 9.4.0 by Sebastian Bergmann and contributors.

[...]

PR #4366, opened 2020-10-20
Run the test suite with assertions off

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

Enabling assertions to get more coverage was a bad idea, because these
are supposed to be disabled in production, which means you could have
code working in the CI, and not working in production. We do not want
that.

See https://github.com/doctrine/dbal/issues/4359

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -40,7 +40,6 @@ jobs:
         with:
           php-version: "${{ matrix.php-version }}"
           coverage: "pcov"
-          ini-values: "zend.assertions=1"
 
       - name: "Cache dependencies installed with composer"
         uses: "actions/cache@v2"
@@ -94,7 +93,6 @@ jobs:
           php-version: "${{ matrix.php-version }}"
           extensions: "oci8"
           coverage: "pcov"
-          ini-values: "zend.assertions=1"
 
       - name: "Cache dependencies installed with composer"
         uses: "actions/cache@v2"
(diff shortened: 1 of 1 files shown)


morozov:
> Enabling assertions to get more coverage was a bad idea, because these are supposed to be disabled in production, which means you could have code working in the CI, and not working in production. We do not want that.

If not have assertions on CI and not have them in production, then where are they supposed to be enabled?

greg0ire:
You're right I'm attempting to fix the issue with a wrong solution. Maybe the right solution would be to somehow find a way to limit calls to `assert()` to simple expressions that do not change anything.

simonwelsh:
If an INI setting can change what code gets run (and the use of `assert()` at all causes this) then CI should run with that setting on and with it off

greg0ire:
@simonwelsh I considered it, but if you want to use that to catch the bug we had with PG, then you will have to run it for the Pg job, so following that logic you might end up with twice as many CI jobs,  and we already have ~ 50.

morozov:
@greg0ire I believe this should be solved at the static analysis level. Maybe Psalm implements this at a higher level than 3 which we're currently at, or it could be an improvement request.

The idea is that if there is an assertion around non-pure expression, it's a bug. As simple as that.

greg0ire:
I had the same idea but didn't connect it to the concept of pure expressions. Fully agree, this is the right solution, and I'm glad we live in a time where static analysis in PHP is advanced enough to make this a realistic feature request.

PR #4348, opened 2020-10-16
Bump Psalm level to 3

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

PR #4370, opened 2020-10-20
Fix undefined variable when assert() is disabled

nicolas-grekas:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

Issue introduced in 8c1e546448e92370702b5570d7dd7014b7c494e2 from #4348

### DIFF
--- a/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
@@ -154,11 +154,12 @@ class PostgreSqlSchemaManager extends AbstractSchemaManager
             $onDelete = $match[1];
         }
 
-        assert(preg_match(
+        $match = preg_match(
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Schema/PostgreSqlSchemaManager.php
```diff
@@ -154,11 +154,12 @@ protected function _getPortableTableForeignKeyDefinition($tableForeignKey)
             $onDelete = $match[1];
         }
 
-        assert(preg_match(
+        $match = preg_match(
```
lcobucci, line 157: This executes the regex now, though, which shouldn't be done - according to the previous code.

morozov:
Thanks for the PR, @nicolas-grekas. The issue has been addressed in https://github.com/doctrine/dbal/pull/4364 (`2.11.x`) and just needs to be merged up.

nicolas-grekas:
Thanks to you!

PR #4374, opened 2020-10-20
Reduce number of build jobs

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

Test the newest version, the oldest version, and versions for which we
have version-specific code.
This is based on platform classes that have a version number in their
name.


.github/workflows/continuous-integration.yml
```diff
@@ -178,13 +178,7 @@ jobs:
           - "7.4"
         postgres-version:
```
greg0ire, line 179: I noticed that we have a class named `lib/Doctrine/DBAL/Platforms/PostgreSQL91Platform.php` in our codebase, should it be removed?

greg0ire:
https://www.githubstatus.com/ says now is not the best time to work on this :sweat_smile:

greg0ire:
This results in 12 fewer checks :)

morozov (approved):

morozov:
@greg0ire please see if it's possible to optimize it even further by not testing all platform versions with PHP 7.3 as it's done for PHP 8 in https://github.com/doctrine/dbal/pull/4361.

PR #4381, opened 2020-10-22

greg0ire:


composer.json
```diff
@@ -31,7 +31,8 @@
         {"name": "Jonathan Wage", "email": "jonwage@gmail.com"}
     ],
     "require": {
-        "php": "^7.3 || ^8.0",
+        "php": "^7.3 || ^8",
+        "ext-pdo": "*",
```
morozov, line 35: This should be reverted. The change in the first line is non-essential, the addition of the second line is wrong.

src/Query/QueryBuilder.php
```diff
@@ -3,6 +3,7 @@
 namespace Doctrine\DBAL\Query;
 
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\ResultStatement;
```
morozov, line 6: This looks like a mistake.

morozov (changes requested):

src/Query/QueryBuilder.php
greg0ire: Indeed, and it's weird phpcs has nothing to complain about this :thinking:

composer.json
SenseException: Wrong for 3.0 or does this also affect 2.12?

morozov: Only wrong for `3.0.x`.

morozov (approved):

PR #4387, opened 2020-10-24
Store query params in driver exceptions

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

This is an attempt at https://github.com/doctrine/dbal/pull/4325#discussion_r500588926.

This PR **removes the text representation of the bound parameters in the query exception message**, and stores these params together with their types in the `DriverException` instead. This shifts the responsibility of (not) storing/displaying bound parameters explicitly to the consumer of the exception.

Just like #4325, this removes the dependency on ext-json (plus a couple lines of code!).


tests/Connection/ExceptionHandlingTest.php
```diff
@@ -33,29 +34,37 @@ protected function setUp(): void
...
-            $this->createMock(DriverException::class),
+        $convertedException = $this->connection->convertExceptionDuringQuery(
+            $driverException,
             'INSERT INTO file (`content`) VALUES (?)',
             [
                 1 => fopen(__FILE__, 'r'),
             ]
         );
+
+        self::assertIsResource($convertedException->getQueryParams()[1]);
```
BenMorel, line 68: I've refactored these tests, but I'm not sure if they really bring any value now. We should probably just create a test that ensures that the params & types are properly passed to the `DriverException` and that's it.

BenMorel:
Oops, after running Psalm, I realized that `DriverException` is marked as immutable. I cannot easily change this as it cascades to other errors, which makes me believe that you do want to keep exceptions immutable.

Thus I only see 2 solutions:

- pass the params & types down to the `ExceptionConverter` and back to the `DriverException`; this would be really cumbersome but would work;
- change `setQueryParams()` and `setQueryParamTypes()` to `withQueryParams()`, that returns a copy of the exception; I've tried this path in 4ee8fe1796c4aa08399aac03632ad4df81ed2a80 but the exception being uncloneable, this makes the implementation a bit of a mess: I had to suppress phpstan & psalm warnings, so I can't say I'm happy with this.

src/Exception/DriverException.php
```diff
@@ -32,4 +47,43 @@ public function getSQLState(): ?string
...
+
+    /**
+     * @internal This is only meant to be called by the DBAL.
+     *
+     * @param array<mixed>                $params
+     * @param array<Type|int|string|null> $types
+     *
+     * @psalm-suppress ImpurePropertyAssignment
+     */
+    public function withQueryParams(array $params, array $types): self
```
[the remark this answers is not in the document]
BenMorel: I don't like it either, that's just the solution that required the least amount of modifications. I'll try to rework this according to your suggestions!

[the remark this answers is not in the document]
BenMorel: Actually we can't just introduce an `Exception\Query` class: all subclasses of `DBAL\Exception\DriverException` (created in the exception converters) may potentially need the SQL, params & types.

**I think adding the properties to `DriverException` is therefore the way to go**, but I will need to change the `ExceptionConverter::convert()` signature to accept the SQL, params & types as well, which is cumbersome. Is that acceptable?

An alternative to avoid the bloat while still passing all the information is to replace `string $message` in both `ExceptionConverter::convert()` and the `DriverException` constructor with an object that will contain the exception message, the SQL, the params & the types:

```php
class DriverExceptionInfo {
    private string $message;
    private string $sql;
    private array $params;
    private array $types;
}
```

Another alternative would be to return a class-string in `ExceptionConverter::convert()` instead of a class instance (as I don't think exception converters do anything else than `new *Exception()`, do they?), this way we don't need to pass the SQL params etc. to the converters at all.

Thoughts?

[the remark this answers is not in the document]
morozov: > ```php
> class DriverExceptionInfo {
>     private string $message;
>     private string $sql;
>     private array $params;
>     private array $types;
> }
> ```

This is what I meant by the `Query` class:
```php
class Query {
    private string $sql;
    private array $params;
    private array $types;
}
```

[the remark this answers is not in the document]
BenMorel: Sorry, I read your comment too fast. Do you want to follow this route then? (this will be a breaking change to the `ExceptionConverter`)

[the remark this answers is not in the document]
morozov: Yeah, the driver converter will have to accept an optional query. I don't think it's a problem.

src/Connection.php
```diff
@@ -18,6 +18,7 @@
 use Doctrine\DBAL\Driver\Statement as DriverStatement;
 use Doctrine\DBAL\Exception\CommitFailedRollbackOnly;
 use Doctrine\DBAL\Exception\ConnectionLost;
+use Doctrine\DBAL\Exception\DriverException as DBALDriverException;
```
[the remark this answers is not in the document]
BenMorel: Done. But IMHO this highlights an issue with exception naming.

tests/Connection/ExceptionHandlingTest.php
BenMorel: I've removed this test altogether. The `getQuery()` part is tested elsewhere.

src/Exception/DriverException.php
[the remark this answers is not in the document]
BenMorel: Done!

src/Query.php
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @var array<Type|int|string|null>
+     */
+    private $types;
+
+    /**
+     * @param array<mixed>                $params
+     * @param array<Type|int|string|null> $types
+     *
+     * @psalm-suppress ImpurePropertyAssignment
```
BenMorel, line 41: I haven't found a way around this: Psalm is right that `Type` is not immutable.

src/Query.php
```diff
@@ -0,0 +1,70 @@
...
+namespace Doctrine\DBAL;
+
+use Doctrine\DBAL\Types\Type;
+
+/**
+ * A SQL query together with its bound parameters.
+ *
+ * @psalm-immutable
+ */
+class Query
```
BenMorel, line 14: Not sure if you expected `Query` in the `Doctrine\DBAL` namespace?

BenMorel:
@morozov Please check this new version!

src/Query.php
morozov, line 14: This could be declared as `final`.

src/Query.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL;
+
+use Doctrine\DBAL\Types\Type;
+
+/**
+ * A SQL query together with its bound parameters.
```
morozov, line 10: It should be an "An SQL" (because /ˌɛsˌkjuːˈɛl/) for consistency with the rest of the documentation.

src/Query.php
```diff
@@ -0,0 +1,70 @@
...
+{
+    /**
+     * The SQL query.
+     *
+     * @var string
+     */
+    private $sql;
+
+    /**
+     * The parameters bound to the query, if the exception occurred during a query.
```
morozov, line 24: The "if the exception occurred during a query" is irrelevant because otherwise this object won't be instantiated.

src/Query.php
morozov, line 41: Is it still required?

src/Connection.php
```diff
@@ -1593,71 +1543,49 @@ public function createQueryBuilder(): QueryBuilder
...
-        DriverException $e,
+        Driver\Exception $e,
         string $sql,
         array $params = [],
         array $types = []
-    ): Exception {
+    ): DriverException {
         $messageFormat = <<<'MESSAGE'
-An exception occurred while executing "%s"%s:
+An exception occurred while executing the query:
```
morozov, line 1552: Without the SQL being part of the exception message, this becomes just a static prefix. I don't think there are any other cases where a driver exception is instantiated.

What if we remove the $message argument from everywhere and instantiate the driver exception from the underlying exception and, optionally, a query.

morozov (commented):
Looks like we're moving in the right direction.

morozov:
@beberlei do you think we could include this into 3.0 instead of `master`? While it changes the behavior developers may be used to, it eliminates the possibility of leaking sensitive data which cannot be fixed w/o a BC break.

src/Query.php
morozov: Ah, no, I expected it under `Doctrine\DBAL\Exception`

src/Query.php
morozov: I see. Makes sense.

src/Connection.php
BenMorel: Makes sense. BTW,  if you really want to avoid leaking sensitive data, you must not repeat the underlying driver exception message that may contain some SQL:

```php
$connection->query("SELECT * 'sentitive data'");
```

> SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near **''sentitive data''** at line 1

Should we just always use a generic message like `'An exception occurred during query.'`? Not very informative, but secure.

src/Query.php
BenMorel: Yes, because `Type` is not immutable.

src/Query.php
BenMorel: Sorry, copy/paste mistake. Fixed in 05909e844c39ba011e6345dab19a9a09a8a47b27.

src/Query.php
BenMorel: Done in https://github.com/doctrine/dbal/pull/4387/commits/0c840ce893aced99b093d2bb0be55d7e124cdad8.

src/Query.php
BenMorel: Right, done in https://github.com/doctrine/dbal/pull/4387/commits/0c840ce893aced99b093d2bb0be55d7e124cdad8.

src/Connection.php
morozov: Let's take it one step at a time. I agree with the statement above but I don't want to creep the scope of this case.

BenMorel: Basically, we have 2 ways to create driver exceptions:

- [Connection::convertExceptionDuringQuery()](https://github.com/doctrine/dbal/blob/4b8386c987e37f92c7e299279c0a5896e1bb4aa6/src/Connection.php#L1595)
- [Connection::convertException()](https://github.com/doctrine/dbal/blob/4b8386c987e37f92c7e299279c0a5896e1bb4aa6/src/Connection.php#L1622)

Should we just use 2 generic messages here:

- "An exception occurred during query"
- "An exception occurred in driver"

And *not* include the original exception message?

BenMorel: > I agree with the statement above but I don't want to creep the scope of this case.

Sorry, not sure what you're suggesting?

morozov: I'm suggesting to not make any changes outside the scope of this case. This case is about getting rid of JSON encoding by putting all the query information into the exception object instead of imprinting it in the message. A side effect of this is better handling of the potentially sensitive statement parameters. We can focus on improving it further in a separate case.

BenMorel: Alright, so if I understand you correctly, you want to keep including the underlying exception message in the `DriverException` message?

morozov: Correct. But I'm not sure if the static prefix is needed any longer. Could we just pass the underlying exception to the `DriverException` and make it extract the message and assign it to itself?

BenMorel: Sure we can. It all depends on what you want in this exception message. If you really want to make it bulletproof secure, you just put the prefix and nothing else. Otherwise the original exception message, with or without the prefix.

A couple thoughts, unrelated to each other:

- Even if you want to keep the prefix, you don't need to pass it down the `ExceptionConverter`, you can just hardcode it in the `DriverException` constructor, and choose which prefix to use depending on whether `$query` is set;
- TBH I'm not sure how insecure putting sensitive data in the exception message is; we should advocate best practices and not displaying exception messages anyway. And if you're using a third-party error reporting system, you'll usually trust it not to share sensitive data it might intercept. My two cents.

morozov: > And if you're using a third-party error reporting system, you'll usually trust it not to share sensitive data it might intercept. My two cents.

It's not about trust. E.g. according to GDPR you should be able to erase customer data from the database and you have a procedure for that. If this data lands in the logs, you should be able to process your logs as well which is absolutely not necessary.

morozov: Since it's not obvious how to proceed in this case, let's leave it as is.

morozov: > You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ''sentitive data'' at line 1

I'm curious if the same is reproducible with a syntactically valid query. A syntax error is an indicator of something utterly broken in the code. Allowing such code to deal with sensitive data and blaming logging for leaking it is not the best strategy.

src/Query.php
BenMorel: @morozov I can see that you resolved this, do you want to keep `Query` in the `Doctrine\DBAL` namespace after all?

src/Connection.php
BenMorel: > Since it's not obvious how to proceed in this case, let's leave it as is.

I pushed 1b0820032ab75e6a2d70680f202cb93dfc620f1a to test creating the exception message in `DriverException`, IMO it looks much cleaner. I can revert it if you wish, though.

morozov (commented):
Logic-wise looks good to me, only the upgrade notes are missing. Could you try to retarget against `3.0.x`?

morozov:
@doctrine/team-dbal, please be informed. With this change, the query and its parameters will be no longer part of the exception message but will be accessible via `DriverException::getQuery()`. It will help prevent leaking potentially sensitive data to logs, etc. but will require changes in the tooling to display these bits during development.

BenMorel:
@morozov Rebased on `3.0.x`, waiting for feedback on wording!

src/Connection.php
morozov: No, it looks good.

morozov:
@BenMorel I believe the wording is not that critical but the upgrade notes are still missing.

BenMorel:
@morozov Done!

By the way, shouldn't `ExceptionConverter` and `DriverException::__construct()` be internal?

morozov:
> By the way, shouldn't `ExceptionConverter` and `DriverException::__construct()` be internal?

That would make sense, I think. I believe all exception constructors should be internal (not in this patch) since only the DBAL is supposed to instantiate them.

BenMorel:
> That would make sense, I think. I believe all exception constructors should be internal (not in this patch) since only the DBAL is supposed to instantiate them.

Should I make at least these 2 `@internal`? Or we'll have to revert the changelog later.

morozov:
> Should I make at least these 2 `@internal`? Or we'll have to revert the changelog later.

Yeah, let's do, and update the upgrade notes accordingly. The wording changes look good to me as well, so it may be also the time to squash everything.

BenMorel:
@morozov Done!

src/Driver/API/ExceptionConverter.php
```diff
@@ -6,7 +6,11 @@
 
 use Doctrine\DBAL\Driver\Exception;
 use Doctrine\DBAL\Exception\DriverException;
+use Doctrine\DBAL\Query;
 
+/**
+ * @internal
```
morozov, line 12: The interface shouldn't be made internal. Driver developers will implement it in their drivers. Only the specific classes can. Although, it won't prevent the changes like this from being the breaking ones.

src/Exception/DriverException.php
```diff
@@ -4,23 +4,41 @@
...
 use Doctrine\DBAL\Driver\Exception as TheDriverException;
 use Doctrine\DBAL\Exception;
+use Doctrine\DBAL\Query;
 
 use function assert;
 
 /**
  * Base class for all errors detected in the driver.
  *
+ * @internal
```
morozov, line 14: The class itself shouldn't be made internal. Consumers of the DBAL API may catch it. Only the constructor may be made internal.

morozov (commented):
I'm sorry. I misread your question in https://github.com/doctrine/dbal/pull/4387#issuecomment-719044234 and thought you were asking about constructors in both cases.

src/Driver/API/ExceptionConverter.php
BenMorel: Ok, then it doesn't make much sense. I'll revert this one.

src/Exception/DriverException.php
BenMorel: Oops, my bad. That was what I intended.

BenMorel:
@morozov I fixed this in https://github.com/doctrine/dbal/pull/4387/commits/06732c5789c689960a537a209344fd8b36fe00a1, please let me know if everything's ok now and I'll squash.

morozov:
Looks good.

BenMorel:
Squashed!

morozov (dismissed):

morozov (approved):

morozov:
Thanks, @BenMorel!

Issue #1295, opened 2011-08-01
DBAL-141: PDO Connection Failure through TNS - PDOOracle/Driver.php line 56


doctrinebot:
Jira issue originally created by user digitalmechanic:

On line 56 of the file ./vendor/doctrine-dbal/lib/Doctrine/DBAL/Driver/PDOOracle/Driver.php:

The line reads:
$dsn .= '))(CONNECT_DATA=(SID=' . $params['dbname'] . ')))'; should read  $dsn .= '))(CONNECT_DATA=(SERVICE_NAME=' . $params['dbname'] . ')))';

If both SID and SERVICE_NAME need to be available for use, then logic should be inserted to determine which method is being used.

PR #4393, opened 2020-10-31
Remove warning comment from PDOOracle driver

mvorisek:
|      Q       |   A
|------------- | -----------
| Type         | comment only
| BC Break     | no
| Fixed issues | no

#### Summary

I am no longer aware of any segfaults, if so, please be specific.

Comment added in - https://github.com/doctrine/dbal/commit/5be161335fcd3a16c19c4014cc94472e9473d0b5 which title links to https://github.com/doctrine/dbal/issues/1295

### DIFF
--- a/lib/Doctrine/DBAL/Driver/PDOOracle/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/PDOOracle/Driver.php
@@ -10,10 +10,6 @@ use PDOException;
 /**
  * PDO Oracle driver.
  *
- * WARNING: PDO Oracle is not maintained by Oracle or anyone in the PHP community,
- * which leads us to the recommendation to use the "oci8" driver to connect
- * to Oracle instead.
- *
  * @deprecated Use {@link PDO\OCI\Driver} instead.
  */
 class Driver extends AbstractOracleDriver
--- a/lib/Doctrine/DBAL/Driver/PDOOracle/Driver.php
+++ b/lib/Doctrine/DBAL/Driver/PDOOracle/Driver.php
@@ -10,8 +10,7 @@ use PDOException;
 /**
  * PDO Oracle driver.
  *
- * WARNING: This driver gives us segfaults in our testsuites on CLOB and other
- * stuff. PDO Oracle is not maintained by Oracle or anyone in the PHP community,
+ * WARNING: PDO Oracle is not maintained by Oracle or anyone in the PHP community,
  * which leads us to the recommendation to use the "oci8" driver to connect
  * to Oracle instead.
  *


mvorisek:
there is something bad with CI

mvorisek:
Hi @beberlei, that comment is from yours commit https://github.com/doctrine/dbal/commit/5be161335fcd3a16c19c4014cc94472e9473d0b5#diff-1c5108ac2719f535f6603abecd44a515c25b065494276cd10913f9a3734c31c1R27 . I know, 9 years ago, but can you provide your feedback on it, is there still any known segfault?

greg0ire:
I reported one of the CI issues here: https://github.com/shivammathur/setup-php/issues/325

mvorisek:
I belive we can merge this, if PHP 7.1 support will be reintroduced, than this change (and every other needed) will be reverted in that PR for 2.x

greg0ire:
We could, but probably won't in order to not make this harder than it should be. I expect the PR to be either merged or rejected in a reasonable time frame, we can decide what should happen to this PR after, there is absolutely no rush.

mvorisek:
yes, 4394 - delete the last 3 comment here

greg0ire:
@shivammathur patched the issue, let's try again.

greg0ire:
Okay it works. The next issue seems to be PHP 8 + pdo_mysql: it fails with both MariaDB and MySQL, only with that driver and that version of PHP.

lib/Doctrine/DBAL/Driver/PDOOracle/Driver.php
```diff
@@ -10,8 +10,7 @@
 /**
  * PDO Oracle driver.
  *
- * WARNING: This driver gives us segfaults in our testsuites on CLOB and other
- * stuff. PDO Oracle is not maintained by Oracle or anyone in the PHP community,
+ * WARNING: PDO Oracle is not maintained by Oracle or anyone in the PHP community,
```
morozov, line 13: As well as the previous version of this warning, it may become invalid over time. I'd rather let the end-user decide which of the drivers they want to use and remove this warning entirely.

greg0ire, line 13: That statement seems wrong too. This driver seems to be maintained directly in php-src, and the last commit was in October this year: https://github.com/php/php-src/blob/master/ext/pdo_oci/oci_driver.c

mvorisek: removed

mvorisek: done

greg0ire (approved):

morozov:
Thanks, @mvorisek.

PR #4394, opened 2020-10-31
Strong type $sql param in logger interfaces

mvorisek:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no - LSP not violated by declaring stricter iface
| Fixed issues | no

### DIFF
--- a/lib/Doctrine/DBAL/Logging/DebugStack.php
+++ b/lib/Doctrine/DBAL/Logging/DebugStack.php
@@ -32,7 +32,7 @@ class DebugStack implements SQLLogger
     /**
      * {@inheritdoc}
      */
-    public function startQuery($sql, ?array $params = null, ?array $types = null)
+    public function startQuery(string $sql, ?array $params = null, ?array $types = null)
     {
         if (! $this->enabled) {
             return;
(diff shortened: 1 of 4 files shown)


greg0ire:
If #4386 gets merged, this will be a BC-break because PHP 7.1 does not support dropping type declarations when extending a class.

mvorisek:
> 
> 
> If #4386 gets merged, this will be a BC-break because PHP 7.1 does not support dropping type declarations when extending a class.

That is correct - https://3v4l.org/Asimv - one small reason to never reintroduce PHP 7.1 support

mvorisek:
I belive we can merge this, if PHP 7.1 support will be reintroduced, than this change (and every other needed) will be reverted in that PR for 2.x

greg0ire:
We could, but probably won't in order to not make this harder than it should be. I expect the PR to be either merged or rejected in a reasonable time frame, we can decide what should happen to this PR after, there is absolutely no rush.

morozov:
Strict scalar types are already implemented for the entire codebase in `master`. Implementing them in `2.x` would be a breaking change, regardless of the supported PHP versions.

mvorisek:
@morozov in master yes, in 3.x no - what is the difference between these branches, is master planned for 4.x and when will it be roughly released? Thanks.

greg0ire:
Yes `master` is currently 4.x, we cannot promise that everything that is on master will be released in 4.0.0 though (we originally planned to release it as 3.0.0).

PR #4396, opened 2020-10-31
Fix php8 mysql mariadb

greg0ire:
The only thing that I could find that looks related is this: https://github.com/php/php-src/commit/50752401a7c16f8e1f4bbe7ef52bc8fb5b17418f

I think getting a warning in lower PHP versions instead of an exception although we configure PDO to throw might be a bug, and I think the commit above might have for side effect to fix that bug.


greg0ire:
Not sure if I picked the right labels, CI might be wrong.

morozov (approved):
👍

PR #4400, opened 2020-11-01
LockMode::NONE should not set WITH (NOLOCK)

### LINKED ISSUE #4391: Clarify LockMode::NONE
I feel like there is some confusion with `LockMode::NONE`.

In all platforms but SQLServer and SQLAnywhere, `LockMode::NONE` is equivalent to not giving a lock hint.

I these 2 platforms however, `LockMode::NONE` adds a `WITH (NOLOCK)` hint, which as I understand it, allows dirty reads, effectively changing the isolation level to READ UNCOMMITTED for this table.

**Is this really what was intended?**

In the ORM for example, using `null` as a lock mode in `EntityManager::find()` is different from using `LockMode::NONE` , but only on these 2 platforms:

| Lock mode  | SQLServer/SQLAnywhere | Other platforms |
| ------------- | ------------- | ------------- |
| `null`  | *no hint*  | *no hint* |
| `LockMode::NONE`  | `WITH(NOLOCK)` | *no hint*  |

And the read semantics are therefore different across platforms for `LockMode::NONE`, which I believe is a bug.

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #4391

#### Summary

This fixes the issue detailed in #4391, with SQL Server and SQL Anywhere setting `WITH (NOLOCK)` for `LockMode::NONE`, which effectively means using a `READ UNCOMMITTED` isolation level at table level, which is not the contract of `LockMode::NONE`.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
@@ -52,7 +52,7 @@ class SQLAnywherePlatform extends AbstractPlatform
     {
         switch (true) {
             case $lockMode === LockMode::NONE:
-                return $fromClause . ' WITH (NOLOCK)';
+                return $fromClause;
 
             case $lockMode === LockMode::PESSIMISTIC_READ:
                 return $fromClause . ' WITH (UPDLOCK)';
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -1574,7 +1574,7 @@ SQL
     {
         switch (true) {
             case $lockMode === LockMode::NONE:
-                return $fromClause . ' WITH (NOLOCK)';
+                return $fromClause;
 
             case $lockMode === LockMode::PESSIMISTIC_READ:
                 return $fromClause . ' WITH (HOLDLOCK, ROWLOCK)';
--- a/tests/Doctrine/Tests/DBAL/Functional/LockMode/NoneTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/LockMode/NoneTest.php
@@ -1,0 +1,101 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL\Functional\LockMode;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\OCI8;
+use Doctrine\DBAL\Exception;
+use Doctrine\DBAL\LockMode;
+use Doctrine\DBAL\Platforms\SqlitePlatform;
+use Doctrine\DBAL\Platforms\SQLServerPlatform;
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\TransactionIsolationLevel;
+use Doctrine\Tests\DbalFunctionalTestCase;
+use Doctrine\Tests\TestUtil;
+
+class NoneTest extends DbalFunctionalTestCase
+{
+    /** @var Connection */
+    private $connection2;
+
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        if ($this->connection->getDriver() instanceof OCI8\Driver) {
+            // https://github.com/doctrine/dbal/issues/4417
+            self::markTestSkipped('This test fails on OCI8 for a currently unknown reason');
+        }
+
+        if ($this->connection->getDatabasePlatform() instanceof SQLServerPlatform) {
+            // Use row versioning instead of locking on SQL Server (if we don't, the second connection will block when
+            // attempting to read the row created by the first connection, instead of reading the previous version);
+            // for some reason we cannot set READ_COMMITTED_SNAPSHOT ON when not running this test in isolation,
+            // there may be another connection active at this point; temporarily forcing to SINGLE_USER does the trick.
+            $db = $this->connection->getDatabase();
+            $this->connection->executeStatement('ALTER DATABASE ' . $db . ' SET SINGLE_USER WITH ROLLBACK IMMEDIATE');
+            $this->connection->executeStatement('ALTER DATABASE ' . $db . ' SET READ_COMMITTED_SNAPSHOT ON');
+            $this->connection->executeStatement('ALTER DATABASE ' . $db . ' SET MULTI_USER');
+        }
+
+        $table = new Table('users');
+        $table->addColumn('id', 'integer');
+        $table->setPrimaryKey(['id']);
+
+        $this->connection->getSchemaManager()->dropAndCreateTable($table);
+
+        $this->connection2 = TestUtil::getConnection();
+
+        if ($this->connection2->getSchemaManager()->tablesExist('users')) {
+            return;
+        }
+
+        if ($this->connection2->getDatabasePlatform() instanceof SqlitePlatform) {
+            self::markTestSkipped('This test cannot run on SQLite using an in-memory database');
+        }
+
+        self::fail('Separate connections do not seem to talk to the same database');
+    }
+
+    public function tearDown(): void
+    {
+        parent::tearDown();
+
+        if ($this->connection2->isTransactionActive()) {
+            $this->connection2->rollBack();
+        }
+
+        $this->connection2->close();
+
+        $this->connection->getSchemaManager()->dropTable('users');
+
+        if (! $this->connection->getDatabasePlatform() instanceof SQLServerPlatform) {
+            return;
+        }
+
+        $db = $this->connection->getDatabase();
+        $this->connection->executeStatement('ALTER DATABASE ' . $db . ' SET READ_COMMITTED_SNAPSHOT OFF');
+    }
+
+    public function testLockModeNoneDoesNotBreakTransactionIsolation(): void
+    {
+        try {
+            $this->connection->setTransactionIsolation(TransactionIsolationLevel::READ_COMMITTED);
+            $this->connection2->setTransactionIsolation(TransactionIsolationLevel::READ_COMMITTED);
+        } catch (Exception $e) {
+            self::markTestSkipped('This test must be able to set a transaction isolation level');
+        }
+
+        $this->connection->beginTransaction();
+        $this->connection2->beginTransaction();
+
+        $this->connection->insert('users', ['id' => 1]);
+
+        $query = 'SELECT id FROM users';
+        $query = $this->connection2->getDatabasePlatform()->appendLockHint($query, LockMode::NONE);
+
+        self::assertFalse($this->connection2->fetchOne($query));
+    }
+}
--- a/tests/Doctrine/Tests/DBAL/Platforms/AbstractSQLServerPlatformTestCase.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/AbstractSQLServerPlatformTestCase.php
@@ -374,14 +374,14 @@ abstract class AbstractSQLServerPlatformTestCase extends AbstractPlatformTestCas
         }
 
         $sql = 'SELECT m0_.NOMBRE AS NOMBRE0, m0_.FECHAINICIO AS FECHAINICIO1, m0_.FECHAFIN AS FECHAFIN2'
-            . ' FROM MEDICION m0_ WITH (NOLOCK)'
+            . ' FROM MEDICION m0_'
             . ' INNER JOIN ESTUDIO e1_ ON m0_.ESTUDIO_ID = e1_.ID'
             . ' INNER JOIN CLIENTE c2_ ON e1_.CLIENTE_ID = c2_.ID'
             . ' INNER JOIN USUARIO u3_ ON c2_.ID = u3_.CLIENTE_ID'
             . ' WHERE u3_.ID = ? ORDER BY m0_.FECHAINICIO DESC';
 
         $alteredSql = 'SELECT TOP 15 m0_.NOMBRE AS NOMBRE0, m0_.FECHAINICIO AS FECHAINICIO1, m0_.FECHAFIN AS FECHAFIN2'
-            . ' FROM MEDICION m0_ WITH (NOLOCK)'
+            . ' FROM MEDICION m0_'
             . ' INNER JOIN ESTUDIO e1_ ON m0_.ESTUDIO_ID = e1_.ID'
             . ' INNER JOIN CLIENTE c2_ ON e1_.CLIENTE_ID = c2_.ID'
             . ' INNER JOIN USUARIO u3_ ON c2_.ID = u3_.CLIENTE_ID'
--- a/tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php
+++ b/tests/Doctrine/Tests/DBAL/Platforms/SQLAnywherePlatformTest.php
@@ -266,7 +266,7 @@ class SQLAnywherePlatformTest extends AbstractPlatformTestCase
             [null, ''],
             [false, ''],
             [true, ''],
-            [LockMode::NONE, ' WITH (NOLOCK)'],
+            [LockMode::NONE, ''],
             [LockMode::OPTIMISTIC, ''],
             [LockMode::PESSIMISTIC_READ, ' WITH (UPDLOCK)'],
             [LockMode::PESSIMISTIC_WRITE, ' WITH (XLOCK)'],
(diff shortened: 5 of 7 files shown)


greg0ire:
Please improve your commit message according to [the contributing guide](https://www.doctrine-project.org/contribute/index.html#working-on-topic-branches).

BenMorel:
@greg0ire Done!

SenseException (dismissed):
If this is handled as a bugfix, this looks :+1: after reading a bit more about `WITH (NOLOCK)` behavior.

BenMorel:
Does this PR need approval from someone else?

morozov:
@BenMorel I believe this change cannot be covered with an integration test but it still requires a reproducer that could be run manually across all supported platforms to confirm that the fix is valid. See the testing guidelines (#4356) and #4279 for an example.

BenMorel:
@morozov Would it be OK if I add a test that uses 2 separate connections (using `TestUtil::getConnection()`) to ensure that SQL executed by one uncommitted transaction is not visible by another transaction?

morozov:
@BenMorel, Absolutely. If the scenario in question can be reproduced using the existing APIs, please add a test for it.

BenMorel:
To ensure that we're realling fixing something that's broken, I created a failing test case in #4415; it fails as expected on SQL Server, but **does anyone know why [it's also failing on Oracle](https://github.com/doctrine/dbal/pull/4415/checks?check_run_id=1371468375)?**

Basically, given 2 transactions in `READ COMMITTED` isolation, `T2` is allowed to see a row that's been inserted by `T1` but not yet committed. Bug?

BenMorel:
Sorry, force-pushed a commit there and the link to the OCI8 build was broken. Fixed.

morozov:
> Bug?

No idea. You can mark this test incomplete on Oracle since it's out of scope.

As far as I understand the code change, it makes the behavior of `null` and `NONE` identical. Which raises the question of whether we should support both and whether the call to "append" with "none" makes any sense.

BenMorel:
> You can mark this test incomplete on Oracle since it's out of scope.

OK, and I'll open a bug to keep track of this, to ensure that we're not overlooking something big here.

> As far as I understand the code change, it makes the behavior of null and NONE identical. Which raises the question of whether we should support both and whether the call to "append" with "none" makes any sense.

Yep, that's why I'm advocating to remove support for `null`, and use `LockMode::NONE` as the new default. The call to "append" with NONE is fine IMO, as long as the method returns the original string untouched.

morozov:
Both points sound good.

BenMorel:
**Does anyone have a SQL Server installation at hand?**

I'm struggling to create a working test case here. The problem is: by default, in `READ COMMITTED` isolation, SQL Server blocks all reads on a row that's being updated by another transaction, so my test case blocks indefinitely on the second connection.

I've read [here](https://www.mssqltips.com/sqlservertip/6368/sql-server-readcommittedsnapshot-database-option-and-read-commited-transaction-isolation-level/) that we can set the following to make SQL Server use row versioning instead of locking:

> ALTER DATABASE doctrine SET READ_COMMITTED_SNAPSHOT ON;

This should, in theory, make the query in the second transaction return immediately, but it still blocks. This is contrary to what I can read in the linked article.

I would appreciate a helping hand here!

morozov:
> Does anyone have a SQL Server installation at hand?

You can spin up a server from the same Docker image as we use on CI: https://github.com/doctrine/dbal/blob/7b5d241e0ae981d00855412b28ca68d5a4477a60/.github/workflows/continuous-integration.yml#L429

morozov:
@BenMorel please avoid experimentation on CI. Each build job of this PR has locked and timed out after one hour. Given that we run 4 jobs sequentially on AppVeyor, a single build effectively blocked the pipeline for 4 hours.

BenMorel:
@morozov I'm sorry, but I could not repeat the lock locally using the Docker image; at least, not while executing queries manually on the CLI, where it works as expected.

I'll try to do my best to investigate further locally and attempt to not block CI again.

BenMorel:
@morozov, I finally got it to work, with quite a lot of local setup & trial-and-error.

I can't say it's really pretty, but I think it's as pretty as an integration test with 2 connections can be, especially on this quite complex RDBMS.

We now have the same tests failing on SQL server in #4415, and passing here.

Let me know if you need anything else to get this merged! :+1:

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
```diff
@@ -0,0 +1,100 @@
...
+            // https://github.com/doctrine/dbal/issues/4417
+            self::markTestSkipped('This test fails on OCI8 for a currently unknown reason');
+        }
+
+        if ($this->connection->getDatabasePlatform() instanceof SQLServerPlatform) {
+            // Use row versioning instead of locking on SQL Server (if we don't, the second connection will block when
+            // attempting to read the row created by the first connection, instead of reading the previous version);
+            // for some reason we cannot set READ_COMMITTED_SNAPSHOT ON when not running this test in isolation,
+            // there may be another connection active at this point; temporarily forcing to SINGLE_USER does the trick.
+            $name = $this->connection->fetchOne('SELECT db_name()');
```
morozov, line 37: It should be the same as `$this->connection->getDatabase()`.

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
```diff
@@ -0,0 +1,100 @@
...
+            $this->connection->executeStatement('ALTER DATABASE ' . $name . ' SET SINGLE_USER WITH ROLLBACK IMMEDIATE');
+            $this->connection->executeStatement('ALTER DATABASE ' . $name . ' SET READ_COMMITTED_SNAPSHOT ON');
+            $this->connection->executeStatement('ALTER DATABASE ' . $name . ' SET MULTI_USER');
+        }
+
+        $table = new Table('users');
+        $table->addColumn('id', 'integer');
+        $table->setPrimaryKey(['id']);
+
+        $this->connection->getSchemaManager()->createTable($table);
```
morozov, line 47: It should `dropAndCreateTable()` instead. Otherwise, if the test crashes badly, the table will exist on the subsequent run.

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
```diff
@@ -0,0 +1,100 @@
...
+
+        $this->connection2 = TestUtil::getConnection();
+
+        if ($this->connection2->getSchemaManager()->tablesExist('users')) {
```
morozov, line 51: What is this logic for? Is it an assertion?

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
```diff
@@ -0,0 +1,100 @@
...
+            return;
+        }
+
+        if ($this->connection2->getDatabasePlatform() instanceof SqlitePlatform) {
```
morozov, line 55: This could be moved to the beginning of the method.

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
```diff
@@ -0,0 +1,100 @@
...
+            self::markTestSkipped('This test cannot run on SQLite using an in-memory database');
```
morozov, line 56: What about the non-in-memory database?

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
```diff
@@ -0,0 +1,100 @@
...
+        $name = $this->connection->fetchOne('SELECT db_name()');
+        $this->connection->executeStatement('ALTER DATABASE ' . $name . ' SET READ_COMMITTED_SNAPSHOT OFF');
+    }
+
+    public function testLockModeNoneDoesNotBreakTransactionIsolation(): void
+    {
+        try {
+            $this->connection->setTransactionIsolation(TransactionIsolationLevel::READ_COMMITTED);
+            $this->connection2->setTransactionIsolation(TransactionIsolationLevel::READ_COMMITTED);
+        } catch (Exception $e) {
```
morozov, line 83: Should we catch a specific exception in this case? Otherwise, the test may get skipped due to any exception.

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
```diff
@@ -0,0 +1,100 @@
...
+
+        $this->connection->beginTransaction();
+        $this->connection2->beginTransaction();
+
+        $this->connection->insert('users', ['id' => 1]);
+
+        $query = 'SELECT id FROM users';
+        $query = $this->connection2->getDatabasePlatform()->appendLockHint($query, LockMode::NONE);
+
+        self::assertSame([], $this->connection2->fetchAllNumeric($query));
```
morozov, line 95: Could it be `self::assertFalse($this->connection2->fetchOne($query))`?

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
```diff
@@ -0,0 +1,100 @@
...
+
+        $this->connection->commit();
```
morozov, line 97: This is not the right place for this code. It may not be reached if the assertion fails.

The transaction in the default connection should be rolled back automatically: https://github.com/doctrine/dbal/blob/9eb0219291aec995a75a40984be6f77449ab9b0b/tests/FunctionalTestCase.php#L62-L67

lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
```diff
@@ -51,9 +51,6 @@ class SQLAnywherePlatform extends AbstractPlatform
     public function appendLockHint($fromClause, $lockMode)
     {
         switch (true) {
-            case $lockMode === LockMode::NONE:
-                return $fromClause . ' WITH (NOLOCK)';
```
morozov, line 55: It should return just the `$fromClause` instead. On the contrary, the `default` case should throw an exception in a future major release.

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
```diff
@@ -0,0 +1,100 @@
...
+use Doctrine\DBAL\Exception;
+use Doctrine\DBAL\LockMode;
+use Doctrine\DBAL\Platforms\SqlitePlatform;
+use Doctrine\DBAL\Platforms\SQLServerPlatform;
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\TransactionIsolationLevel;
+use Doctrine\Tests\DbalFunctionalTestCase;
+use Doctrine\Tests\TestUtil;
+
+class LockModeTest extends DbalFunctionalTestCase
```
morozov, line 18: Would it make sense to move it under `LockMode\NoneTest`? It's unlikely that this test case could accommodate the tests for any other modes.

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
BenMorel: I hate the fact that PHPCS forces me to return early here, IMO it doesn't convey the meaning of this logic indeed. It's to ensure that both connections are actually dealing with the same persistent storage behind the scenes (i.e., not in-memory SQLite).

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
BenMorel: Not really, as SQLite may not only run from memory, but from a file. I need to know at this point whether we're running in-memory (table `users` does not exist on connection 2), or from a file (table exists).

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
BenMorel: It would pass the if (table exists) check and would execute the test.

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
BenMorel: It does not throw a specific exception, but a generic `Doctrine\DBAL\Exception`:

https://github.com/doctrine/dbal/blob/5f94000488373d45445fd6fcd495cb9a4bdbe09c/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php#L3005

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
BenMorel: It's the same thing, yes. I'll change it if you wish.

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
BenMorel: Good point. I'll fix this.

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
BenMorel: Good idea.

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
morozov: Please do.

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
morozov: What exactly error code/message do we expect, and what platform does it happen on?

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
morozov: Now I understand the logic but it's not that obvious. I'm curious if it could be factored more explicitly or at least have some comments that explain it.

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
BenMorel: Fixed in f6ad92af525842417a503675aad6ad0939c1a41d.

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
BenMorel: Fixed in f6ad92af525842417a503675aad6ad0939c1a41d.

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
BenMorel: Updated in f6ad92af525842417a503675aad6ad0939c1a41d.

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
BenMorel: Done in f6ad92af525842417a503675aad6ad0939c1a41d.

lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
BenMorel: Done in https://github.com/doctrine/dbal/pull/4400/commits/c20d7daeb9c7b6c8b6aef8002949d695c4af8c5a. Do you want me to write a todo in the default clause, to throw an exception in 3.x?

tests/Doctrine/Tests/DBAL/Functional/LockModeTest.php
BenMorel: Done in 93af9af11db9fc73623cf07d7cd327c3733ee486.

BenMorel:
@morozov All your feedback has been processed. Please review.

lib/Doctrine/DBAL/Platforms/SQLAnywherePlatform.php
morozov: No TODO needed, you can file an issue for 3.0.

morozov:
Looks good! Please squash, and let's get it merged.

BenMorel:
@morozov Squashed, and I opened #4428 to keep track of https://github.com/doctrine/dbal/pull/4400#discussion_r522239670.

morozov (approved):

morozov:
Thanks, @BenMorel!

BenMorel:
Thank you, could you please merge 2.12.x into 3.0.x, so that I can take care of #4428?

morozov:
@BenMorel I already filed https://github.com/doctrine/dbal/pull/4429.

PR #4403, opened 2020-11-02
Remove redundant phpstan param from DriverManager::getConnection()

simPod:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     |no
| Fixed issues | 

#### Summary

This effectively prevented phpstan from inferring type of `T` template.

> Unable to resolve the template type T in call to method static method Doctrine\DBAL\DriverManager::getConnection()

### DIFF
--- a/lib/Doctrine/DBAL/DriverManager.php
+++ b/lib/Doctrine/DBAL/DriverManager.php
@@ -119,7 +119,6 @@ final class DriverManager
      *
      * @throws Exception
      *
-     * @phpstan-param mixed[] $params
      * @psalm-return ($params is array{wrapperClass:mixed} ? T : Connection)
      * @template T of Connection
      */
--- a/tests/Doctrine/Tests/DBAL/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/ConnectionTest.php
@@ -36,7 +36,7 @@ class ConnectionTest extends DbalTestCase
     /** @var Connection */
     private $connection;
 
-    /** @var string[] */
+    /** @var array{wrapperClass?: class-string<Connection>} */
     protected $params = [
         'driver' => 'pdo_mysql',
         'host' => 'localhost',


greg0ire:
> This effectively prevented phpstan from inferring type of T template.

IIRC that was the goal, it did not understand the annotations I added when I added them. Glad to see that has been fixed :)

greg0ire (dismissed):

simPod:
> it did not understand the annotations I added when I added them

Sounds legit, I also often type differently for psalm and phpstan

morozov:
Thanks, @simPod. Please squash and double-check if the same will work on `3.0.x` (strict PHPStan rules) and `master`, e.g. cherry-pick your commit on top of those branches and run both PHPStan and Psalm locally.

simPod:
There are some errors but I'd rather fix them individually on each branch.

morozov:
> There are some errors but I'd rather fix them individually on each branch.

It doesn't work this way. We regularly merge the changes from the older branch up to `master`. If the build fails during the up merge, it will have to be fixed in an emergency way. We need to be sure that when we apply certain changes in 2.12, we know how to handle it in the newer ones.

simPod:
I did this for master as well https://github.com/doctrine/dbal/pull/4419. Don't know how to do it in one PR properly, same as in #4413

morozov (approved):

morozov:
FYI: I reverted it during the merge of `3.0.x` to `master` (https://github.com/doctrine/dbal/pull/4432).

This change opens a can of worms: the connection parameters array contains a lot of different keys of different types which may or may not be used in certain scenarios. The `array{?key: type}` notation implies that there are no other keys than "key" which is not true.

simPod:
> This change opens a can of worms: the connection parameters array contains a lot of different keys of different types which may or may not be used in certain scenarios. The array{?key: type} notation implies that there are no other keys than "key" which is not true.

I agree, if I was the one who introduced it I'd definitely list all possible keys. 

> I reverted it during the merge of 3.0.x to master

Should I do something? Not sure what it implies.

morozov:
> I agree, if I was the one who introduced it I'd definitely list all possible keys.

If only it was that easy :-)

> Should I do something? Not sure what it implies.

It was just a note for your information.

andrew-demb:
>  while Psalm allows specifying named array types, it doesn't allow sharing them across files

@morozov Psalm allows import and reuse them - please see https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-import-type

morozov:
@andrew-demb that's new to me. Thanks for the pointer!

PR #4408, opened 2020-11-04
Refactor - access AbstractPlatform::_eventManager prop by getter only

mvorisek:
|      Q       |   A
|------------- | -----------
| Type         | refactor only
| BC Break     | no
| Fixed issues | no

#### Summary

Refactor is motivated as with pure access by getter it is much simpler to wrap/proxy a platform.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -169,7 +169,7 @@ abstract class AbstractPlatform
     /**
      * Gets the EventManager used by the Platform.
      *
-     * @return EventManager
+     * @return EventManager|null
      */
     public function getEventManager()
     {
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -1443,9 +1443,9 @@ public function getDropTableSQL($table)
             );
         }
 
-        if ($this->_eventManager !== null && $this->_eventManager->hasListeners(Events::onSchemaDropTable)) {
+        if ($this->getEventManager() !== null && $this->getEventManager()->hasListeners(Events::onSchemaDropTable)) {
```
mvorisek, line 1446: let me know if I should assign to variable here first

lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
```diff
@@ -169,7 +169,7 @@ public function setEventManager(EventManager $eventManager)
     /**
      * Gets the EventManager used by the Platform.
      *
-     * @return EventManager
+     * @return EventManager|null
```
mvorisek, line 172: legit change?

PR #4412, opened 2020-11-06
Upgrade phpstan and psalm

simPod:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     |no
| Fixed issues | 

#### Summary

In order to fix issues related but not explicitly caused by https://github.com/doctrine/dbal/pull/4403


tests/Functional/Connection/FetchTest.php
```diff
@@ -176,7 +176,7 @@ public function testIterateKeyValueOneColumn(): void
             ->getDummySelectSQL();
 
         $this->expectException(NoKeyValue::class);
-        iterator_to_array($this->connection->iterateKeyValue($sql));
+        $assign = iterator_to_array($this->connection->iterateKeyValue($sql));
```
greg0ire, line 179: I think at least Psalm might be complaining about this at some point in time or when using a stricter level. [There is even a blog post about that](https://psalm.dev/articles/better-unused-variable-detection)
I had the same issue on a paid project with my colleague @dFayet , and resorted to an ignore rule on the `tests` directory. @morozov , what's your opinion on this?

morozov: A proper way to fix this might be to rework the usage of `iterator_to_array()` to a set of lower-level iterator functions like `reset()`, `next()` to trigger the fetch. Not sure how PHPStan decides whether the return value of a given function should be used or not.

Alternatively, this could be temporarily suppressed in the configuration.

morozov (changes requested):
Please retarget against `2.12.x`.

tests/Functional/Connection/FetchTest.php
simPod: I tried to change it

simPod: ok the error disappeared. any idea why?

composer.json
```diff
@@ -40,9 +40,11 @@
     "require-dev": {
         "doctrine/coding-standard": "^8.1",
         "jetbrains/phpstorm-stubs": "^2019.1",
-        "phpstan/phpstan": "^0.12.40",
+        "phpstan/phpstan": "^0.12.54",
+        "phpstan/phpstan-phpunit": "^0.12.16",
+        "phpstan/phpstan-strict-rules": "^0.12.5",
```
morozov, line 45: The two dependencies above should not be introduced. They are only used in the newer versions of the DBAL.

phpstan.neon.dist
```diff
@@ -77,6 +77,13 @@ parameters:
             paths:
                 - %currentWorkingDirectory%/lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
 
+        # Fixing the issue may cause a BC break.
+        # TODO: fix in 3.0.0
```
morozov, line 81: It won't be fixed in 3.0.0. Please remove the TODO.

tests/Functional/Connection/FetchTest.php
morozov: What error?

composer.json
```diff
@@ -40,9 +40,11 @@
...
         "phpunit/phpunit": "^9.4",
-        "psalm/plugin-phpunit": "^0.10.0",
+        "psalm/plugin-phpunit": "^0.12.0",
         "symfony/console": "^2.0.5|^3.0|^4.0|^5.0",
         "vimeo/psalm": "^3.17.2"
```
morozov, line 49: This PR claims to update Psalm but it doesn't.

greg0ire: It does, in the `composer.lock`. But then https://github.com/doctrine/dbal/blob/faf8ddd7e09e495d890a7579f842e5b6fc24aa4a/.github/workflows/static-analysis.yml#L62 should probably be changed too.

morozov: It should bump the constraint too.

.github/workflows/static-analysis.yml
```diff
@@ -59,4 +59,4 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Psalm
-        uses: docker://vimeo/psalm-github-actions:3.17.2
+        uses: docker://vimeo/psalm-github-actions:3.18.0
```
greg0ire, line 62: Why are you not using the exact same version as in `composer.json`?

composer.json
```diff
@@ -40,11 +40,11 @@
...
         "doctrine/coding-standard": "^8.1",
         "jetbrains/phpstorm-stubs": "^2019.1",
-        "phpstan/phpstan": "^0.12.40",
+        "phpstan/phpstan": "^0.12.54",
         "phpunit/phpunit": "^9.4",
-        "psalm/plugin-phpunit": "^0.10.0",
+        "psalm/plugin-phpunit": "^0.12.0",
         "symfony/console": "^2.0.5|^3.0|^4.0|^5.0",
-        "vimeo/psalm": "^3.17.2"
+        "vimeo/psalm": "^3.18.2"
```
greg0ire, line 47: @morozov should we drop the `^` part here? that way we avoid discrepancies between local runs and CI (at the cost of a warning from Composer I think)

.github/workflows/static-analysis.yml
simPod: it is not tagged https://hub.docker.com/r/vimeo/psalm-github-actions/tags

composer.json
ciaranmcnulty: The ^ doesn’t block later versions so why a warning?

greg0ire: I checked and you are right, there is no warning. Maybe that changed, or I have faulty memory. That said, I was proposing to remove the `^`, that's when I thought you would get a warning, you won't get a warning with it that's for sure.

.github/workflows/static-analysis.yml
morozov: Why not upgrade directly to 4.1.1?

composer.json
morozov: > Maybe that changed, or I have faulty memory.

No, Composer would definitely trigger a warning in some cases if some dependency constraint was declared as a specific version.

> should we drop the `^` part here?

Why now? The specific version is locked in the lock file which we still commit to Git. I believe it should be done separately from this specific upgrade.

greg0ire: I wasn't require we do it now, but in general. Now would be as good a time as any.

.github/workflows/static-analysis.yml
simPod: done

.github/workflows/static-analysis.yml
```diff
@@ -59,4 +59,6 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Psalm
-        uses: docker://vimeo/psalm-github-actions:3.17.2
+        uses: docker://vimeo/psalm-github-actions:4.1.1
+        env:
+          REQUIRE_DEV: "true"
```
morozov, line 64: What is this flag for?

.gitignore
```diff
@@ -9,3 +9,4 @@ vendor/
 /phpunit.xml
 /.phpcs-cache
 /phpstan.neon
+/psalm.xml.dist
```
morozov, line 12: Why exclude Psalm configuration from version control?

simPod: mb, should have been no dist

.github/workflows/static-analysis.yml
simPod: See https://github.com/psalm/psalm-github-actions/pull/17

composer.json
morozov: > Now would be as good a time as any.

The way how we manage development dependencies is for the most part initiated by the work being done in #4386 which seems stuck. Once we decide to make a change, we'll make it for all dependencies at once.

This patch comes from an external contributor and just updates a dependency. I don't see much point in increasing the scope.

morozov (dismissed):

composer.lock
```diff
@@ -4,32 +4,33 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "0e4428b976a1e1835abdff64014ad723",
+    "content-hash": "9147cdc691d11f68d6fbae8012f036bd",
     "packages": [
         {
             "name": "doctrine/cache",
-            "version": "v1.7.1",
+            "version": "1.10.2",
```
morozov, line 11: It looks like you updated all locked dependencies. Would it make sense to just update psalm with dependencies?

simPod: Should be k now

morozov (approved):

SenseException (approved):

morozov:
Thanks, @simPod.

simPod:
Thanks for patience with reviews :)

PR #4414, opened 2020-11-08
ResultCacheStatement::fetchAllAssociative does not store results in cache

### LINKED ISSUE #4406: ResultCacheStatement::fetchAllAssociative does not store results in cache
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | 2.12.0

#### Summary

Using `fetchAllAssociative` to fetch query results on a Symfony project using `doctrine/dbal` (no ORM) with a Redis cache seems to not use the cache. But `fetchAll` does.  

#### Current behaviour

The query is executed at each run.

#### How to reproduce

```php
$sql = <<<SQL
    SELECT
        *
    FROM my_table
    WHERE id = :id
SQL;

$connection = ;// instance of Doctrine\DBAL\Connection
$stmt = $connection->executeCacheQuery(
    $sql,
    ['id' => 1],
    ['id' => PDO::PARAM_INT],
    new QueryCacheProfile(
        360,
        'my_key'
    )
);
$results = $stmt->fetchAllAssociative(); // This will always execute a query
```

This problem also shows up when using `fetchAllNumeric`, `fetchAssociative`, and `fetchNumeric`, but not when using `fetchAll`

#### Expected behaviour

The query should be executed at the same occurence with `fetchAllAssociative|fetchAllNumeric|fetchAssociative|fetchNumeric` as with `fetchAll`

I've tried to dig a bit in the code, I saw 2 things:
- `fetchAllAssociative` and `fetchAllNumeric`:
  I am a bit less confident there as it seems to have "nested statement calls" (I do not know how it is called, look for example at lines [256](https://github.com/doctrine/dbal/blob/cad295a6b777b759fe3dc0be0cc2a3d7e87aa8e1/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php#L256) and [258](https://github.com/doctrine/dbal/blob/cad295a6b777b759fe3dc0be0cc2a3d7e87aa8e1/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php#L258]), but they both calls a method [store()](https://github.com/doctrine/dbal/blob/cad295a6b777b759fe3dc0be0cc2a3d7e87aa8e1/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php#L328) that will also not save data in cache.

-  `fetchAssociative` and `fetchNumeric`:
   They both call method `doFetch`, but this methods will not store the result if it differs from `false`
https://github.com/doctrine/dbal/blob/cad295a6b777b759fe3dc0be0cc2a3d7e87aa8e1/lib/Doctrine/DBAL/Cache/ResultCacheStatement.php#L314-L320

**Update**: `fetchAssociative` and `fetchNumeric` are probably not an issue, according to the doc

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

Fixes #4406.

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
@@ -4,6 +4,7 @@ namespace Doctrine\Tests\DBAL\Functional;
 
 use Doctrine\Common\Cache\ArrayCache;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
+use Doctrine\DBAL\Cache\ResultCacheStatement;
 use Doctrine\DBAL\Driver\ResultStatement;
 use Doctrine\DBAL\FetchMode;
 use Doctrine\DBAL\Logging\DebugStack;
@@ -212,7 +213,10 @@ class ResultCacheTest extends DbalFunctionalTestCase
         self::assertCount(2, $this->sqlLogger->queries);
     }
 
-    public function testFetchAllSavesCache(): void
+    /**
+     * @dataProvider fetchAllProvider
+     */
+    public function testFetchingAllRowsSavesCache(callable $fetchAll): void
     {
         $layerCache = new ArrayCache();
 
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
```diff
@@ -212,7 +213,10 @@ public function testDontFinishNoCache(): void
         self::assertCount(2, $this->sqlLogger->queries);
     }
 
-    public function testFetchAllSavesCache(): void
+    /**
+     * @dataProvider fetchAllProvider
+     */
+    public function testFetchAllSavesCache(callable $fetchAll): void
```
greg0ire, line 219:
```suggestion
    public function testMethodSavesCache(callable $fetchAll): void
```

tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
```diff
@@ -212,7 +213,10 @@ public function testDontFinishNoCache(): void
         self::assertCount(2, $this->sqlLogger->queries);
     }
 
-    public function testFetchAllSavesCache(): void
+    /**
+     * @dataProvider fetchAllProvider
```
greg0ire, line 217: This is no longer solely about `fetchAll()`

```suggestion
     * @dataProvider cacheSavingMethodProvider
```

tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
```diff
@@ -222,11 +226,42 @@ public function testFetchAllSavesCache(): void
...
+
+        $fetchAll($stmt);
 
         self::assertCount(1, $layerCache->fetch('testcachekey'));
     }
 
+    /**
+     * @return iterable<string,list<mixed>>
+     */
+    public static function fetchAllProvider(): iterable
```
greg0ire, line 238:
```suggestion
    public static function cacheMethodSavingProvider(): iterable
```

tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
morozov: All methods being tested fetch all records, that's why they save cache, they are not the methods saving cache. From the fetching a single `fetchAll()` method, the test transitioned to testing all methods fetching all records.

I'd just rename the test and leave everything as is.

greg0ire (approved):

greg0ire:
Thanks @morozov !

PR #4416, opened 2020-11-09
Update .gitattributes

bytestream:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Exclude `ci` from Composer

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,5 @@
 /.appveyor.yml export-ignore
+/ci export-ignore
 /composer.lock export-ignore
 /docs export-ignore
 /.doctrine-project.json export-ignore


.gitattributes
```diff
@@ -13,3 +13,4 @@
 /tests export-ignore
 /.travis.yml export-ignore
 /UPGRADE.md export-ignore
+/ci export-ignore
```
morozov, line 16: Please use alphabetical order (before `/composer.lock`).

morozov (changes requested):

.gitattributes
bytestream: Sure. I've updated it.

morozov (approved):

morozov:
@greg0ire I recall there was an attempt to only include certain files into the package as opposed to constantly excluding the new ones. Do you remember how it ended?

morozov:
Thanks, @bytestream.

PR #4422, opened 2020-11-11
QueryBuilder must use 0-based positional parameter keys

### LINKED ISSUE #4421: QueryBuilder uses 1-based positional parameter keys
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    |no
| Version     | 2.12.1

#### Summary

According to the documentation,

https://github.com/doctrine/dbal/blob/84328cd947706210caebcaea3ca0394b3ebc4673/docs/en/reference/query-builder.rst#L47-L48

The usage of non-zero-based keys has been explicitly deprecated in https://github.com/doctrine/dbal/pull/4411. The issue was discovered during the work on https://github.com/doctrine/dbal/pull/4397 which enforces this deprecation. 

#### Current behaviour

When using `QueryBuilder::creartePositionalParameter()`, the keys are created starting from `1`.

#### How to reproduce

```php
$qb->select($qb->createPositionalParameter('FOO'));

var_dump($qb->getParameters());
// array(1) {
//   [1] =>
//   string(3) "FOO"
// }
```

#### Expected behaviour

```php
// array(1) {
//   [0] =>
//   string(3) "FOO"
// }
```

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

Fixes #4421.

Although this bug exists in `2.x`, I don't think we have to fix it in a patch release. It's sufficient to fix it in `3.0.x`.

### DIFF
--- a/src/Query/QueryBuilder.php
+++ b/src/Query/QueryBuilder.php
@@ -1330,8 +1330,8 @@ class QueryBuilder
      */
     public function createPositionalParameter($value, $type = ParameterType::STRING)
     {
-        $this->boundCounter++;
         $this->setParameter($this->boundCounter, $value, $type);
+        $this->boundCounter++;
 
         return '?';
     }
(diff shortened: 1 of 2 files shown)


greg0ire (commented):
Should the phpdoc of `getParameters` be changed to use `list` then?

morozov:
No, because those can be named parameters as well or even a mix which is not allowed by the documentation but not enforced programmatically.

SenseException (approved):

PR #4424, opened 2020-11-11
Mark SQLParserUtils internal

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

`SQLParserUtils` is not meant for external usage and will be removed in #4397 in favor of another API. In order to be able to finish this change after the 3.0 release, we need to make sure the class is not part of the public API.


greg0ire (approved):
Ok with adding this annotation and considering this not a BC-break but some soft warning for users of this class, but that does not mean we can change anything on 2.x , right?

morozov:
> that does not mean we can change anything on 2.x , right?

Right. 2.x stays unchanged.

PR #4425, opened 2020-11-12

morozov:


tests/Functional/ResultCacheTest.php
```diff
@@ -218,11 +221,36 @@ public function testFetchAllSavesCache(): void
...
         );
-        $result->fetchAllAssociative();
+
+        $fetchAll($result);
 
         self::assertCount(1, $layerCache->fetch('testcachekey'));
     }
 
+    /**
+     * @return iterable<string,list<mixed>>
```
greg0ire, line 231: Not sure why we didn't use `iterable<string, array{callable}>` here, but this is probably not the place to fix this.

greg0ire (approved):

tests/Functional/ResultCacheTest.php
morozov: I didn't know of this syntax.

PR #4431, opened 2020-11-14
Use individual properties in Configuration

BenMorel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | -

#### Summary

This PR moves away from a global `Configuration::$_attributes` associative array, towards individual properties.

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -15,61 +15,71 @@ class Configuration
     private $middlewares = [];
 
     /**
-     * The attributes that are contained in the configuration.
-     * Values are default values.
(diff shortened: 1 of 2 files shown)


src/Configuration.php
```diff
@@ -23,61 +23,79 @@ class Configuration
     protected $_attributes = [];
```
morozov, line 23: Should it be removed?

BenMorel: OMG, yes of course. Done.

morozov (approved):

morozov:
Thanks, @BenMorel!

PR #4435, opened 2020-11-15
Update documentation after the 3.0.0 release

morozov:
1. Add 3.0 to the build matrix.
2. Rename `master` to `4.0-dev` for clarity, since this is about releases, not branches.

### DIFF
--- a/.doctrine-project.json
+++ b/.doctrine-project.json
@@ -9,17 +9,16 @@
             "name": "3.0",
             "branchName": "3.0.x",
             "slug": "latest",
-            "upcoming": true
-        },
-        {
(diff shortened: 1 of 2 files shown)


.doctrine-project.json
```diff
@@ -9,18 +9,17 @@
             "name": "3.0",
             "branchName": "3.0.x",
             "slug": "latest",
```
greg0ire, line 11: Should we add `3.1.x` as well?

morozov: I think it’s too early.

greg0ire (approved):

Issue #4388, opened 2020-10-26
dbal/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php should override getLengthExpression() and use CHAR_LENGTH()


cmanley:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | x.y.z

#### Summary

For multibyte character encodings in MySQL, the CHAR_LENGTH() function returns the character length.
So it seems to me that dbal/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php should override getLengthExpression() and use CHAR_LENGTH() instead of LENGTH() as inherited from the parent class.

#### Current behaviour

Incorrect calculation of string lengths for strings containing multibyte characters.

#### How to reproduce

Place a single multibyte character in a mysql field with utf8mb4 encoding and then calculate it's length with getLengthExpression().

#### Expected behaviour

1

morozov:
Should be fixed by https://github.com/doctrine/dbal/pull/4855.

PR #4442, opened 2020-11-18
Implementation CHAR_LENGTH expression for Mysql

MacGruber91:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #4388 

#### Summary
Added new method to `Doctrine\DBAL\Platforms\AbstractPlatform` - `getCharLengthExpression`, implemented only for `Doctrine\DBAL\Platforms\MySqlPlatform`. Method makes it possible to get the length in characters.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/AbstractPlatform.php
@@ -770,7 +770,7 @@ abstract class AbstractPlatform
      *
      * @throws Exception If not supported on this platform.
      */
-    public function getCharLengthExpression(string $column): string
+    public function getCharLengthExpression($column)
     {
         throw Exception::notSupported(__METHOD__);
     }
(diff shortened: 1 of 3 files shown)


morozov:
Since it's only supported by MySQL, what's the point in implementing it in the DBAL? How would using this API be different than building the SQL expression explicitly?

MacGruber91:
I agree, for this, you can use API. But it improves the developer experience.

morozov:
> […] it improves the developer experience.

This is arguable. As a developer who targets my code at more platforms than MySQL, I may use this new method because it works on MySQL and only learn that it's only MySQL after running tests against other platforms.

The introduction of a `LogicException` like in this case is a clear indicator of worse developer experience.

MacGruber91:
The method with an exception shows that way for calculating the length of a string in characters is not supported or has a different specificity on other platforms

PR #4446, opened 2020-11-22
Update incompatible method signatures in PDO Connection and Statement definition to comply with LSP violations RFC in PHP 8.

Hailong:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | n/a 

### Update incompatible method signatures in PDO Connection and Statement definition to comply with LSP violations RFC in PHP 8. [https://wiki.php.net/rfc/lsp_errors](https://wiki.php.net/rfc/lsp_errors)

The issue can be reproduced and verified with the [Symfony Demo](https://github.com/symfony/demo) project.

Build PHP 8 and setup the Symfony Demo project, we will see the fetal error be thrown which complains the incompatible method signature. After applying the change in this PR, the errors will be fixed and CRUD works.

I ran unit tests with PHP 8, while it seems this case is not covered.

### DIFF
--- a/lib/Doctrine/DBAL/Cache/ArrayStatement.php
+++ b/lib/Doctrine/DBAL/Cache/ArrayStatement.php
@@ -62,8 +62,11 @@ class ArrayStatement implements IteratorAggregate, ResultStatement
     /**
      * {@inheritdoc}
      */
-    public function setFetchMode($fetchMode, $arg2 = null, $arg3 = null)
+    public function setFetchMode($fetchMode, ...$args)
     {
+        $arg2 = $args[0] ?? null;
+        $arg3 = $args[1] ?? null;
+
         if ($arg2 !== null || $arg3 !== null) {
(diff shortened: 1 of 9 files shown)


andrew-demb:
PR target versoon 2.10.x doesn't support  PHP 8.
https://github.com/doctrine/dbal/blob/ec4e1b90e6dbe81e10e556150be16b368acad67c/composer.json#L35

AFAIK 2.12.x already fixed to be compatible with PHP 8

Hailong:
> PR target versoon 2.10.x doesn't support PHP 8.
> https://github.com/doctrine/dbal/blob/ec4e1b90e6dbe81e10e556150be16b368acad67c/composer.json#L35
> 
> AFAIK 2.12.x already fixed to be compatible with PHP 8

I just check the 2.12.x branch quickly, it seems not fixed yet. While as long as it is on track we should be good. Feel free to close this one, since I'm not able to close it myself.

Issue #2265, opened 2015-12-17
(mssql) Problem with method addColumn when i need to set default in existing table with data


tjura:
When i use code $schema->getTable('users')->addColumn('sex' , 'integer')->setDefault(0)
Doctrine built two Queries: 
1. Alter table...
and second
2. Add constraint

Migrations throws errow on first query.

Error shows up only when i have data in table users.

I found solution, adding custom SQLServer2012Platform but there is a lot of private method that i need to copy.

A simillar problem happend when a try drop column with constraint.

deeky666:
@inferno1992 can you please provide exact details on how to reproduce this issue? I'm afraid the information you provided is insufficient for a decent investigation.

kalinowcyk:
I've got the same problem. 
This is how to reproduce: 
1) I have a table filled with data, database is MSSQL Server,
2) I've made a migration like this:
```
<?php

namespace Application\Migrations;

use Doctrine\DBAL\Migrations\AbstractMigration;
use Doctrine\DBAL\Schema\Schema;

/**
 * Auto-generated Migration: Please modify to your needs!
 */
class Version20160906134946 extends AbstractMigration
{
    /**
     * @param Schema $schema
     */
    public function up(Schema $schema)
    {
        // this up() migration is auto-generated, please modify it to your needs
		$schema->getTable('my_table')->addColumn('my_column', 'integer')->setNotnull(false)->setDefault(0);

    }

    /**
     * @param Schema $schema
     */
    public function down(Schema $schema)
    {
        // this down() migration is auto-generated, please modify it to your needs

    }
}
```
3) When I run this migration, I've got an error:
```
Migration 20160906134946 failed during Execution. Error An exception occurred while executing 'ALTER TABLE my_table ADD my_column INT NOT NULL':

SQLSTATE[42000]: [Microsoft][ODBC Driver 11 for SQL Server][SQL Server]ALTER TABLE only allows columns to be added that can contain nulls, or have a DEFAULT definition specified, or the column being added is an identity or timestamp column, or alternatively if none of the previous conditions are satisfied the table must be empty to allow addition of this column. Column 'my_column' cannot be added to non-empty table 'my_table' because it does not satisfy these conditions.
```

morozov:
Closing as fixed by https://github.com/doctrine/dbal/pull/4447. Please reopen if it's not.

PR #4447, opened 2020-11-24
Ensure the default is declared in the same statement for new columns

duncan3dc:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2265

#### Summary

When adding columns to tables that already contain data, SQL Server requires that the default constraint is declared in the same statement, Doctrine currently generates a separate `ADD CONSTRAINT` statement, which means the add column statement errors with something like:
```
SQLSTATE [42000, 4901]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]
ALTER TABLE only allows columns to be added that can contain nulls, or have a DEFAULT
definition specified, or the column being added is an identity or timestamp column,
or alternatively if none of the previous conditions are satisfied the table must be
empty to allow addition of this column. Column 'addcolumn' cannot be added
to non-empty table 'mytable' because it does not satisfy these conditions.
```

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -484,11 +484,14 @@ SQL
             }
 
             $columnDef    = $column->toArray();
-            $queryParts[] = 'ADD ' . $this->getColumnDeclarationSQL($column->getQuotedName($this), $columnDef);
-
+            $addColumnSql = 'ADD ' . $this->getColumnDeclarationSQL($column->getQuotedName($this), $columnDef);
             if (isset($columnDef['default'])) {
-                $queryParts[] = $this->getAlterTableAddDefaultConstraintClause($diff->name, $column);
+                $addColumnSql .= ' CONSTRAINT ' .
+                    $this->generateDefaultConstraintName($diff->name, $column->getQuotedName($this)) .
+                    $this->getDefaultValueDeclarationSQL($columnDef);
             }
+
+            $queryParts[] = $addColumnSql;
 
             $comment = $this->getColumnComment($column);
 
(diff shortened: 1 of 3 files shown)


greg0ire (dismissed):

morozov:
Thank you for the patch, @duncan3dc. In addition to the unit test, it would be good to have an integration test that confirms that the fixed SQL also works and produces the expected results. See the [testing guidelines](https://www.doctrine-project.org/projects/doctrine-dbal/en/2.12/reference/testing.html#integration-tests).

duncan3dc:
Of course, I've had a go at that in https://github.com/doctrine/dbal/pull/4447/commits/2ea06d11f277958ff50b2dc0c4f073720c22a99e it fails against the 2.12.x branch and passes with this PR

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -484,12 +484,15 @@ public function getAlterTableSQL(TableDiff $diff)
             }
 
             $columnDef    = $column->toArray();
-            $queryParts[] = 'ADD ' . $this->getColumnDeclarationSQL($column->getQuotedName($this), $columnDef);
-
+            $addColumnSql = 'ADD ' . $this->getColumnDeclarationSQL($column->getQuotedName($this), $columnDef);
             if (isset($columnDef['default'])) {
```
morozov, line 488: This is out of the scope of this issue but it looks like this code incorrectly handles the `DEFAULT NULL` case.

morozov (commented):
Looks good. Please squash into one commit since the code change and both tests belong together.

morozov (approved):

morozov:
Thanks, @duncan3dc!

PR #2284, opened 2016-01-08
Added parameter `default_dbname` to pdo_pgsql driver, used to override the default database

kimhemsoe:

PR #4449, opened 2020-11-25
Document the shape of the connection parameters array

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

This pull request documents the shape of the connection parameters array and addresses some issues discovered by the static analysis. It also enforces documentation of the newly added parameters (I hope it's not going to happen) and identifies a certain potential for improvement:

1. Parameters currently is a mess of DBAL-level (host, port), wrapper-level (primary/replica, sharding), and driver-level (charset, memory) elements.
2. There's no way to tell whether a given parameter is currently used or not (e.g. due to a typo or being driver/wrapper-specific).

Notes:

1. The `OverrideParams` type is added temporarily until Psalm implements the support for self-referencing types (https://github.com/vimeo/psalm/issues/4653#issuecomment-731768265).
2. PHPStan currently interprets `@psalm-*` annotations but doesn't understand user-defined types. Therefore, each added annotation has to be doubled with a `@phpstan-` one with a simpler type.
3. What the hell is `default_dbname` and why was it introduced in https://github.com/doctrine/dbal/pull/2284?

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -432,23 +432,21 @@ class Connection implements DriverConnection
 
                 // The database to connect to might not yet exist.
                 // Retry detection without database name connection parameter.
-                $databaseName           = $this->params['dbname'];
-                $this->params['dbname'] = null;
+                $params = $this->params;
+
+                unset($this->params['dbname']);
 
                 try {
                     $this->connect();
                 } catch (Throwable $fallbackException) {
                     // Either the platform does not support database-less connections
                     // or something else went wrong.
-                    // Reset connection parameters and rethrow the original exception.
-                    $this->params['dbname'] = $databaseName;
-
                     throw $originalException;
+                } finally {
+                    $this->params = $params;
                 }
 
-                // Reset connection parameters.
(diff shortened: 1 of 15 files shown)


greg0ire:
> What the hell is default_dbname and why was it introduced in #2284?

I think it might be the extra layer Postgres and only Postgres has. See https://stackoverflow.com/a/1157008/353612

greg0ire (approved):

morozov:
> I think it might be the extra layer Postgres and only Postgres has.

Most of the DB platforms have a notion of the schema (e.g. Oracle, SQL Server). In Oracle, the schema is a synonym of the user; in MySQL, it's a synonym of the database. DBAL already supports defining them separately. I'll file a deprecation PR since it doesn't make any sense. Most likely, the problem attempted to be solved can be solved w/o introducing an extra configuration parameter.

bendavies:
i'm thinking `@psalm-type Params` is missing `url`. what do you think?

morozov:
As far as I understand, nothing prevents it from being used, it's just of unknown type. Adding it shouldn't hurt.

PR #4450, opened 2020-11-25

morozov:
Notice the difference in the number of newly suppressed Psalm errors between https://github.com/doctrine/dbal/pull/4449 (10) and here (2).


greg0ire (approved):
> Notice the difference in the number of newly suppressed Psalm errors between #4449 and here.

Indeed, nice!

PR #4452, opened 2020-11-27
Resolve exceptions in php8 related to php/php-src#6157

mikepage:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

#### Summary

In a Symfony project with Doctrine migrations and multiple transactions in a single command, the following exception in thrown with MySQL connections.

In Connection.php line 1761:
There is no active transaction 

In my understanding this is occurring because PHP 8 now throws an error when committing/rolling back a transaction which was already autocommitted. (php/php-src@990bb34).

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1700,7 +1700,8 @@ class Connection implements DriverConnection
 
             try {
                 $result = $connection->commit();
-            } catch (\PDOException $e) {}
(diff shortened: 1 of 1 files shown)


morozov:
> In my understanding this is occurring because PHP 8 now throws an error when committing/rolling back a transaction which was already autocommitted.

Why do you want to commit a transaction that was already autocommitted?

Issue #3290, opened 2018-09-17
Oracle does not support streams for the CLOB columns


morozov:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.8.0

#### Summary

See the [build failure](https://app.continuousphp.com/git-hub/doctrine/dbal/build/ee788937-49f1-4be2-a652-d8dc7fe72712/logs/a9bb62f2-3809-4d79-93b4-f24dc1719040) on ContinuousPHP. The Oracle driver handles `ParameterType::LARGE_OBJECT` (which represents both BLOB and CLOB parameters) by creating a `OCI_TEMP_BLOB` resource which doesn't work for CLOB columns:

https://github.com/doctrine/dbal/blob/3979efeedfda1eda12d30781344583b0ece0a644/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php#L304

Which results in the follwing error: 
> ORA-00932: inconsistent datatypes: expected CLOB got BLOB.

Apparently, the same issue ([#57095](https://bugs.php.net/bug.php?id=57095)) exits in `pdo_oci` since 2006.

morozov:
I don't think it's fixable. OCI requires manual reading from and writing to the BLOB handle (i.e. start a transaction, query, read/write, commit). While the DBAL API is designed with the expectation that the drivers will do that automatically (all within a single API call).

It could be only accomplished if the DBAL owned the transaction boundaries (e.g. like the ORM does).

PR #4473, opened 2020-12-28
CLOB type for ORACLEDB included.

rodrigogarbin:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #3290

#### Summary

For oci there is a difference in the types BLOB and CLOB, the class OCI8Statement treated only the BLOB fields. For this, a new type was created in the ParameterType class, which in the OCI8Statement class is converted to OCI_B_CLOB.
For parameter bind with type CLOB, ParameterType :: CLOB must be used.

For CLOB fields smaller than the maximum size of the String type I had no problems, however for large texts it was returning an error, this made me implement this improvement.

### DIFF
--- a/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
+++ b/lib/Doctrine/DBAL/Driver/OCI8/OCI8Statement.php
@@ -306,6 +306,16 @@ class OCI8Statement implements IteratorAggregate, StatementInterface, Result
             $variable =& $lob;
         }
 
+        if ($type === ParameterType::CLOB) {
+            $lob = oci_new_descriptor($this->_dbh, OCI_D_LOB);
+
+            assert($lob !== false);
+
+            $lob->writeTemporary($variable, OCI_TEMP_CLOB);
+
+            $variable = $lob;
+        }
+
         $this->boundValues[$param] =& $variable;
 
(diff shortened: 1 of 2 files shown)


jzonta:
I have the same problem. I hope this PR will be approved.

jzonta (approved):

jzonta (dismissed):
I tested and approved this code. I hope this PR accepted. Congratulations @rodrigogarbin

morozov:
Thanks for the patch, @rodrigogarbin. A few notes:

1. I believe this change warrants a deprecation of `ParameterType::LARGE_OBJECT` in favor of `ParameterType::::CLOB` (added here) and `ParameterType::BLOB` (to be introduced). This way, there will be no illusion that the "large object" binding type can handle both BLOBs and CLOBs across supported platforms.
2. This code change will require an integration (see [testing guidelines](https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/testing.html)).
3. There's no plan to tag any new `2.x` releases, potentially with the exception of improving the upgrade path to `3.0`. Since it's an improvement that adds a new API, it should target the `3.1.x` branch.

morozov:

Issue #4437, opened 2020-11-17
MSSQL adds a redundant ORDER BY clause when with subquery with ORDER BY


trusek:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.12

#### Summary

I have a query where in the ORDER BY clause I have a subquery with an ORDER BY clause, when I limit the query I get an incorrect query with redundant ORDER BY clause

#### Current behaviour

Input sql
``` sql
SELECT col1 FROM test ORDER BY ( SELECT col2 from test ORDER BY col2 )
```
Output sql
``` sql
SELECT col1 FROM test ORDER BY ( SELECT col2 from test ORDER BY col2 ) ORDER BY (SELECT 0) OFFSET 0 ROWS FETCH NEXT 10 ROWS ONLY
```

#### How to reproduce

``` php
/** @var SQLServerPlatform */
$platform;
$querySql = 'SELECT col1 FROM test ORDER BY ( SELECT col2 from test ORDER BY col2 )';
$sql = $platform->modifyLimitQuery($querySql, 10);
```

#### Expected behaviour

``` sql
SELECT col1 FROM test ORDER BY ( SELECT col2 from test ORDER BY col2 ) OFFSET 0 ROWS FETCH NEXT 10 ROWS ONLY
```

PR #4475, opened 2021-01-07
Merge 2.12.x up into 3.0.x

greg0ire:


composer.json
```diff
@@ -37,14 +37,14 @@
...
+        "jetbrains/phpstorm-stubs": "2019.3",
+        "phpstan/phpstan": "0.12.57",
         "phpstan/phpstan-strict-rules": "^0.12.2",
-        "phpunit/phpunit": "^9.4",
-        "psalm/plugin-phpunit": "^0.13.0",
+        "phpunit/phpunit": "9.5.0",
+        "psalm/plugin-phpunit": "0.13.0",
         "symfony/console": "^2.0.5|^3.0|^4.0|^5.0",
-        "vimeo/psalm": "^4.1"
+        "vimeo/psalm": "4.2.1"
```
greg0ire, line 47: @nicwortel why did you use 4.3.1 on your PR to 2.12.x? 2.12.x was using 4.1.1, and 3.0.x is using 4.2.1. Using 4.3.1 here results in some errors.

An interesting consequence of removing composer.lock is that now we can see discrepancies in tools version between branches.

nicwortel: You are right... my bad :facepalm: 

I could have sworn that I used the exact version as it was locked by the `composer.lock`, but I must have messed up switching between versions at some point.

It seems like I accidentally also moved doctrine/coding-standard from 8.1.0 to 8.2.0.

greg0ire: No worries, I'll fix the issues!

greg0ire: Ah I'm going to be forced to upgrade to 4.3.2 because we use `strtok()` with one argument in the codebase, and 4.3.1 takes issue with that (https://github.com/vimeo/psalm/issues/4874)

greg0ire:
Looks like I messed up somewhere, there is a failure that seems related to #4437, @trusek can you maybe help? What I did is I removed the following block, because I think `SQLServer2012Platform` is the only SQL Server platform on this branch.
```
        if (
            $platform instanceof SQLServerPlatform
            && ! $platform instanceof SQLServer2012Platform
        ) {
            $subquery = preg_replace('/^SELECT\s/i', 'SELECT TOP 1 ', $subquery);
        }
```

trusek:
Hey @greg0ire it looks like changes to `SQLServer2012Platform` from this [commit](https://github.com/doctrine/dbal/pull/4475/commits/7aac0847552735af95e409d8691ebabde8db5670) have not been applied.

And If the SqlServer 2012 version is the latest supported version, this mentioned block of code can be safely removed

greg0ire:
Thanks @trusek ! I don't know how I managed this one :sweat_smile:

morozov (approved):

PR #4476, opened 2021-01-08
Merge 3.1.x up into master

greg0ire:


greg0ire:
I need to merge 3.0 in 3.1 first

greg0ire:
@morozov do you understand what's up with AppVeyor? I don't…

I think the part that fails is `appveyor-retry cinst --params '""/InstallDir:C:\tools\php""' --ignore-checksums -y php --version ((choco search php --exact --all-versions -r | select-string -pattern $env:php | sort { [version]($_ -split '\|' | select -last 1) } -Descending | Select-Object -first 1) -replace '[php|]','')`

because later we get the following: 

```
Package name is required. Please pass at least one package name to install.
Command "cinst --params ""/InstallDir:C:\tools\php"" --ignore-checksums -y php --version" failed with exit code 1. Retrying 3 of 3
```

It does not make a lot of sense though, because `php` looks like a package name to be, but I think the part between parenthesis is supposed to output something and does not.

morozov:
@greg0ire last time I dealt with that, I spun up a Windows VM, installed Chocolatey there, and debugged those commands locally. At that time, there was a bug in `choco search` that made it not return anything although it was supposed to return a list. I believe it may be a problem as well.

greg0ire:
Thanks for digging that up! I added steps that run the `choco search` parts by themselves in more detail, hopefully it will help see if this is a similar issue.

greg0ire:
Ok so `choco search php --exact --all-versions -r` outputs `php|8.0.1`, but `choco search php --exact --all-versions -r | select-string -pattern $env:php` outputs nothing at all…

greg0ire:
Fixed by upgrading to the beta version.

morozov:
Please get rid of https://github.com/doctrine/dbal/pull/4476/commits/9a41eae1cb9dbf2cbd3573bdfbc07451c1d615bb.

greg0ire:
ping @morozov

morozov (approved):

PR #4478, opened 2021-01-11
Exclude duplicate SQL statements from migrations (schemadiff->toSql)

orim65:
This is a bug fix for the command doctrine:migrations:diff. A simple way to reproduce the buggy behaviour is to
1) create two entities and a unidirectional one-to-one relationship between them, 
2) update the database schema, 
3) rename the table that is foreign in the key constraint, 
4) run migrations:diff. 
The result is a migration that always fails because of a duplicate drop statement for the same foreign key.

Things taken into consideration in the bug fix:
 - The to-be-orphan foreign key must be dropped before the table is dropped.
 - A new key cannot be created before the new table has been created.
 - When the new key is created, the old key must not be dropped again because it has already been dropped.

Regarding the correctness of the result, it is the latter SQL drop statement that should be removed to achieve a executable migration and desired schema update. It is currently added in the code that adds ALTER TABLE statements based on a TableDiff but unaware of the other parts of SchemaDiff. So I reckon the best place for the 
fix is in SchemaDiff.php where all the SQL statements are gathered. I've added new condition for the incoming ALTER TABLE statements: only include them if not yet included. This condition would also apply to other kind of statements but I haven't come across any bugs in any other SQL statements.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | none raised

#### Summary

Added a check to prevent migrations:diff from adding duplicate alter table statements to a migration script.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/SchemaDiff.php
+++ b/lib/Doctrine/DBAL/Schema/SchemaDiff.php
@@ -163,9 +163,10 @@ class SchemaDiff
 
         foreach ($this->changedTables as $tableDiff) {
             foreach ($platform->getAlterTableSQL($tableDiff) as $alterTableSQL) {
-                if (!in_array($alterTableSQL, $sql)) {
-                    $sql[] = $alterTableSQL;
+                if (in_array($alterTableSQL, $sql)) {
+                    continue;
                 }
+                $sql[] = $alterTableSQL;
             }
         }
 
--- a/lib/Doctrine/DBAL/Schema/SchemaDiff.php
+++ b/lib/Doctrine/DBAL/Schema/SchemaDiff.php
@@ -162,7 +162,11 @@ class SchemaDiff
         }
 
         foreach ($this->changedTables as $tableDiff) {
-            $sql = array_merge($sql, $platform->getAlterTableSQL($tableDiff));
+            foreach ($platform->getAlterTableSQL($tableDiff) as $alterTableSQL) {
+                if (!in_array($alterTableSQL, $sql)) {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Schema/SchemaDiff.php
```diff
@@ -162,7 +163,13 @@ protected function _toSql(AbstractPlatform $platform, $saveMode = false)
         }
 
         foreach ($this->changedTables as $tableDiff) {
-            $sql = array_merge($sql, $platform->getAlterTableSQL($tableDiff));
+            foreach ($platform->getAlterTableSQL($tableDiff) as $alterTableSQL) {
+                if (in_array($alterTableSQL, $sql)) {
```
mvorisek, line 167: I belive this is unsafe. If generated queries need like drop/recreate ID twice (any binary same SQL query in general needed twice), this will break it.

My question is why do we need this PR at all?

orim65: It's broken in its current state, not just unsafe. There's no migration that requires the same SQL statement to appear more than once. Generating duplicate statements only generates redundancy, or in this case, a broken migration.

morozov:

Miro-re:
This is a bug fix that hasn't been reviewed by anyone who read the commit message. Should this bug be posted on the issues page? 

The migrations bundle is quite buggy but there are people who can help fix the bugs. Closing bug-fix PR's without a real review doesn't really encourage anyone to report or fix more bugs.

Issue #4335, opened 2020-10-11
Removal of getSQLResultCasing without alternative breaks ORM assumptions


beberlei:
### BC Break Report

The ORM currently does not quote any identifier. This means the default case for unquoted columns of each database is used, which is a.) depends on server config for MySQL, but by default its case used at creation b.) lowercase for postgresql c.) uppercase for Oracle.

To allow the ORM to support arbitrary databases, the platform must need to know what this behavior is of each platform, so that when it fetches the result of an SQL query without quoted identifiers it can access the values correctly.

When you declare a field in the ORM with for example `<field name="name" />` then the quoting strategy uses `AbstractPlatform::getSQLResultCasing` to determine if the column for this field is `name` or `NAME`.

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 3.0.0

#### Summary

`AbstractPlatform::getSQLResultCasing` was removed, is platform specific behavior needed in ORM. ORM can't offer this abstraction itself as it depends on Platform.

morozov:
As discussed offline, instead of mimicking the behavior of a specific database platform by using the API above, consumers may use the portability layer and normalize the behavior instead. @beberlei please let me know if you expect more work to be done on this issue.

morozov:
As discussed offline, using the portability layer is the proper way to address this issue in consumers.

greg0ire:
@morozov to be clear, this means calls to `getSQLResultCasing` should be removed from the ORM, and people should be advised to use https://github.com/doctrine/dbal/blob/3.1.x/docs/en/reference/portability.rst ?

morozov:
In my understanding, the ORM itself currently relies on the column casing, so it should be using the portability middleware instead.

greg0ire:
Ok, so in `Doctrine\ORM\Configuration::__construct()`, there should be the following call: `$this->setMiddlewares([new \Doctrine\DBAL\PortabilityMiddleware(NOT_SURE_WHAT_TO_PASS_HERE, NOT_SURE_WHAT_TO_PASS_HERE_EITHER)])`?

beberlei:
Referencing https://github.com/doctrine/orm/issues/8887 here as the first step in the 2.x branches.

morozov:
The arguments should be `(Connection::PORTABILITY_FIX_CASE, ColumnCase::LOWER)`.

PR #4481, opened 2021-01-14
Expose support for transactional DDL or lack thereof

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | n/a
#### Summary

Clients may need that information, notably in the context of SQL
migrations.

At the moment, I only know about this being an issue with MySQL, there might be more platforms having issues with this.

The MySQL docs I linked in the code say the following:

>  Similarly, DDL statements are not transactional, and therefore a transaction is (almost) never started for a DDL statement. But there's a difference between a DDL statement and an administrative statement: the DDL statement always commits the current transaction (if any) before proceeding; the administrative statement doesn't. 

I did not add unit tests for this method since there is no complexity, do tell me if you feel they are needed anyway.

### DIFF
--- a/src/Platforms/AbstractPlatform.php
+++ b/src/Platforms/AbstractPlatform.php
@@ -3119,6 +3119,14 @@ abstract class AbstractPlatform
     }
 
     /**
+     * Whether it is safe to wrap DDL statements in transactions
+     */
+    public function supportsTransactionalDDL(): bool
+    {
+        return true;
+    }
+
+    /**
      * Whether the platform supports savepoints.
      *
      * @return bool
--- a/src/Platforms/MySQLPlatform.php
+++ b/src/Platforms/MySQLPlatform.php
@@ -338,6 +338,14 @@ class MySQLPlatform extends AbstractPlatform
     }
 
     /**
+     * @see https://dev.mysql.com/doc/internals/en/transactions-current-situation.html
+     */
+    public function supportsTransactionalDDL(): bool
+    {
+        return false;
+    }
+
+    /**
      * {@inheritDoc}
      */
     public function getListTablesSQL()


greg0ire:
Blocked by https://github.com/doctrine/dbal/pull/4482

stof:
According to https://www.datasunrise.com/blog/professional-info/how-popular-rdbmss-deal-with-ddl-commands-in-transactions/ Oracle does not support it either.

stof:
> I did not add unit tests for this method since there is no complexity, do tell me if you feel they are needed anyway.

should we actually have a functional test validating that this method does not lie ? This could be done by starting a transaction, creating a table, rolling back the transaction and then checking if the table exists or no (if transactional DDL works, there should be no such table left)

src/Platforms/AbstractPlatform.php
```diff
@@ -3118,6 +3118,14 @@ public function supportsTransactions()
         return true;
     }
 
+    /**
+     * Whether it is safe to wrap DDL statements in transactions
+     */
+    public function supportsTransactionalDDL(): bool
```
morozov, line 3124: The API in terms of "supports" as opposed to "do something" violates the "Tell, Don't Ask" principle. It moves the responsibility of handling the difference in implementation details to the consumer of the API and creates an unnecessary coupling.

A proper API, in this case, should look like e.g. `beginTransactioWithDDL()` which would fail for MySQL.

greg0ire: I don't think this is compatible with @stof's plan from https://github.com/doctrine/migrations/issues/1104#issuecomment-759380867 : 
The migrations package would still need a way to decide whether to call `beginTransaction()` (for the case of a migration purely about data), or `beginTransactionWithDDL()` (for the main case).

morozov: If the migration becomes transactional or non-transactional based on the underlying platform, it invalidates the purpose of the DBAL: the end-user will experience different results on different platforms. They test it on Postgres, it works as expected; they test it on MySQL, it applies partially in case of a failure. Is it the desired behavior?

greg0ire: One could argue on what might be considered different results. I think it might be sensible to say people using `extends AbstractionMigration` expect their migration to just work and don't particularly care whether is transactional or not, while people overriding `isTransactional` to let it return `true` really care about their migration being transactional. Your comment might make us reconsider if the behavior of the code that generates migrations should be changed and only this one, so that `AbstractionMigration::isTransactional()` still returns `true` in every case, but is overrided to return `false` when MySQL is detected. @stof, what do you think?

greg0ire: cc @goetas

kimhemsoe: Postgres have a few edge cases as well ex. "_ALTER TYPE ... ADD VALUE (the form that adds a new value to an enum type) cannot be executed inside a transaction block._" in older versions and in newer versions "_If ALTER TYPE ... ADD VALUE (the form that adds a new value to an enum type) is executed inside a transaction block, the new value cannot be used until after the transaction has been committed._" 

Don't know if that is useful information.

stof: @kimhemsoe  given that DBAL does not support native enum types anyway, I think that's OK. Also, in newer versions, that might still be OK if your migration does not need to use the new value. And if it does need it, the solution is the same than today: override `AbstractionMigration::isTransactional()` to disable the transactional behavior for that particular migration.

@morozov the whole point of that API is that we would like doctrine/migrations to provide a smart default. When transactional DDL is supported by the platform, this is the sane default (as it avoids having to deal with half-executed migrations in case of failures, which are a huge pain). We don't want to default to non-transactional migrations for those projects (which don't care about the case of mysql as they generated their migrations for postgresql anyway).
The current behavior of doctrine/migrations is to use a transaction, relying on the fact that mysql was not failing for such case (it was silently dropping the transactional behavior, leaving you with the drawbacks of non-transactional migrations, which you cannot avoid on mysql anyway). But PDO became stricter in PHP 8, breaking the case where you explicit commit a transaction while mysql already committed earlier due to DDL and so is not in a transaction anymore. And so mysql with PHP 8 will need to run doctrine/migrations in a non-transactional way for things to work. If we cannot provide a smart default, it requires editing every generated migrations, either for mysql-based projects or for all other projects (depending on the selected default).

morozov: With https://github.com/doctrine/dbal/pull/4481#discussion_r560147527, this API makes even less sense: whether or not one can execute the migration in a transactional way depends not only on the target platform but also on the content of the migration.

> When transactional DDL is supported by the platform […]

Apparently, it doesn't work this way.

> We don't want to default to non-transactional migrations for those projects (which don't care about the case of mysql as they generated their migrations for postgresql anyway).

It should be the consumer's choice: whether they need it transactional or not. If they need, they shouldn't be able to run certain DDL statements or any DDL statements on a certain platform, but in all other cases, the transaction nature of the migration will be guaranteed. Otherwise, the tool doesn't provide any guarantees, and the consumer will have to be prepared to handling the failures themselves.

> But PDO became stricter in PHP 8, breaking the case where you explicit commit a transaction while mysql already committed earlier due to DDL and so is not in a transaction anymore.

Why not just suppress this error as expected on the migrations side?

And again, the point of an abstraction is to provide a _common ground_ between the supported platforms and drivers. What this API does is exposes certain _differences_ of those making _others_ have to handle them.

stof: > Why not just suppress this error as expected on the migrations side?

silently suppressing this error would imply that we keep default to marking all generated migrations as transactional, and we silently switch to non-transactional transaction for mysql. I don't see how this is better than an explicit smart default.

morozov: Then I don't understand what the proposed "smart default" is.

stof: Well, our smart default would be "use transactional migrations for platforms where this is possible, as that's a much better experience". But having that as a default implementation of `isTransactional()` would allow any check of `isTransactional` to report the right value, instead of the current behavior where executing migrations on mysql shows them as transactional in the PHP side while they are silently not transactional (and that "silently" would require adding hacks to keep it working on PHP 8, as I said).

stof: note that the DBAL platform already has a bunch of similar `supports*` method, allowing the ORM to make some smart decisions for the mapping based on supported features for instance (by default, generated ids are using sequences on postgresql while they use a auto-incremented field on mysql for instance, as that's the recommended approach on each platform)

morozov: > Well, our smart default would be "use transactional migrations for platforms where this is possible, as that's a much better experience".

IMO, this "smart default" is a disservice: if a developer expects their migration to work in a transactional way on PostgresSQL, they will be surprised that it doesn't work the same on MySQL (or in some cases in PostgreSQL as well). It means that if they will have to be prepared to deal with a non-transactional migration. And if that's the case, why do they need to use transactions in other platforms?

> note that the DBAL platform already has a bunch of similar `supports*` method

As I said, this is an anti-pattern that breaks the abstraction and moves the responsibility of handling certain differences between the platform to the consumer of the abstraction.

stof: Generated migrations are not cross-platform anyway (the generated migration file contains the platform-specific SQL queries generated by the platform, as migrations can then be tweaked by devs, which is almost always necessary when migrating a non-empty table).

And postgresql will not silently make it non-transactional. It will fail the migration if you use a feature not supported in transaction (which can then be solved by explicitly making that migration non-transactional in your code).

> It means that if they will have to be prepared to deal with a non-transactional migration. And if that's the case, why do they need to use transactions in other platforms?

Most projects are not dealing with multiple platforms. they run on one (and if they run on multiple, they cannot use doctrine/migrations directly anyway as they need to find a way to have different migrations for each one).

> And if that's the case, why do they need to use transactions in other platforms?

If you never make mistakes when writing your migrations, there is indeed no need. But that's kind of hard to achieve as requirement.
When running non-transactional migrations, recovering from a failure is a manual process where you need to find out which part of the migrations have been applied before the failure and revert those parts only. And when migrating a prod database, that can force you to have downtime because your DB is neither in the old state nor in the new one (and so you have no code able to use that state). When it is about a local DB containing fixtures, I often found it easier to wipe out the whole DB and recreate it to recover (years ago, when I was still using mysql).

greg0ire:
Reusing this PR to see if an AppVeyor issue spotted elsewhere also affects this branch.

src/Platforms/AbstractPlatform.php
morozov: This entire discussion is about Migrations, not the DBAL. If Migrations need to implement a different behavior based on the platform, it should be solved in Migrations. The DBAL is about common things between the platforms, not about the different things.

morozov: On a more constructive note: the problem I've been observing about the ORM and now Migrations is that the _consumers_ of the DBAL APIs are trying to force the introduction of certain APIs in the DBAL that make sense only for them. All the `supports*` and `prefers*` platform methods are a sign of that.

We need to find a different approach to extending the platform hierarchy to the needs of a given project w/o stuffing new responsibilities into the DBAL. This way, each project will be able to maintain that extended API according to its needs and the release cycle w/o creating unnecessary coupling.

The closest possible solution that comes to mind is the Visitor pattern. The DBAL may define the `PlatformVisitor` interface where each platform could identify itself, and the implementor of the visitor could define the actual platform-dependent logic. The only downside of that is the interface will have to remain intact between the major releases and won't be able to accommodate new platforms. Also it's unclear how fine-grained it should be (e.g. whether MySQL, MySQL 8 and MariaDB are different platforms from this PoV). In any event, it's just an example.

stof: well, if you suggest that other Doctrine projects need to implement some code which has a switch on the platform to perform the logic, I find this even worse: adding a new platform (in DBAL or as a separate package providing an implementation) becomes a BC break for these projects. The *list of platform implementations* becomes part of the API for such case.
I think this idea that DBAL APIs should only do things and not report capabilities of the platform is too limiting (or would require upstreaming lots of logic into DBAL itself instead of allowing downstream projects do their thing).

> If Migrations need to implement a different behavior based on the platform, it should be solved in Migrations. The DBAL is about common things between the platforms, not about the different things.

That's precisely what's being done today: doing the different logic based on the capabilities of the platform.. But for that, we need to know those capabilities. And if the DBAL platform stops supporting that, these projects would need to implement a separate concept of platforms which either needs to be configured explicitly (harder for user as they have to configure multiple times they use Postgresql and they could end up with a mismatch if they are not aware of that) or needs the DBAL platform to return the corresponding implementation (no-go for the coupling direction) or requires these projects to support a fixed list of DBAL platforms (the ones for which it knows the correspondence).

kimhemsoe: > This is absolutely fine. Transaction handling and data migrations are two different [bounded contexts](https://verraes.net/2014/02/buzzword-free-bounded-contexts/). The same model (a platform) may have different representations in each context to solve certain problems. The only question is how to extend the model that belongs to one context in another. The best approach I see for now is `if ($platform instanceof MySQLPlatform)` in Migrations.

@morozov I can see where you going and i would agree with you if it was not because Migrations, ORM and DBAL is all under the same umbrella called Doctrine. It should be possible to implement a new driver, platform and schema for a unsupported vendor in dbal without without the need to also hack into Migrations, ORM and etc. to make the driver work in those. That have always, in my humble opinion, been one of the goals behind DBAL and Doctrine as a whole.

morozov: I heard this idea expressed by others and I disagree with the approach. While we should work together on solving problems, each library should have its boundaries. The above is a justification of bad design decisions based on the organizational structure.

 Here's another example of such an issue: https://github.com/doctrine/dbal/issues/4335 and another one: https://github.com/doctrine/dbal/pull/3980#pullrequestreview-405379261. Both of them were solved the right way w/o crippling the DBAL API.

ostrolucky:
I've created https://github.com/doctrine/dbal/pull/4544 which is the test case better demonstrating the issue. Interestingly, issue is in `pdo_mysql`, but not `mysqli`. We need DBAL to handle this seamlessly if it's undesired for DBAL to add API that consumer could use to solve this problem.

greg0ire:
Closing as this has been resolved in doctrine/migrations

PR #4487, opened 2021-01-24
Add cursor support for OCI8

pilot114:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no

#### Summary

Its PR add support bind cursors for OCI8 driver.
Example usage:

```
$conn = Doctrine\DBAL\DriverManager::getConnection($params, $config);

$stmt = $conn->prepare('BEGIN proc_name(o_result => :o_result); END;');
$stmt->bindParam('o_result', $cursor, ParameterType::CURSOR);
$stmt->execute();

/** @var $result Doctrine\DBAL\Driver\OCI8\Result */
$result = $cursor->execute();
$rows = $result->fetchAllAssociative();
```

### DIFF
--- a/src/Driver/OCI8/Cursor.php
+++ b/src/Driver/OCI8/Cursor.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Doctrine\DBAL\Driver\OCI8;
+
+class Cursor extends Statement
+{
+    public function __construct($dbh, ExecutionMode $executionMode, $sth = null)
+    {
+        $this->_dbh          = $dbh;
+        $this->executionMode = $executionMode;
+        $this->_sth          = $sth ?: oci_new_cursor($dbh);
+    }
+
+    public function getStatement()
+    {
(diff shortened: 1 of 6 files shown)


morozov:
One of the DBAL goals is to provide an abstraction layer on top of multiple database platforms so the end-user can write their code in a platform-agnostic way using the DBAL APIs. So far, the API being proposed is implemented only for one platform while other supported platforms in theory may implement it as well (e.g. PostgreSQL, SQL Server).

Before we proceed, we need to see how portable this API is.

pilot114:
We know that not all databases support this functionality. What solution do you see in the general case?

morozov:
The easiest way forward should be to "steal" the connection handle from the driver-level connection via reflection and use it directly via the `oci8` API. Note that the backward compatibility promise doesn't cover this case, so you'll have to be careful when upgrading the DBAL.

PR #4489, opened 2021-01-27
Introduce fetch* methods in query builder

### LINKED ISSUE #4461: Introduce fetch* methods in QueryBuilder for SELECT queries
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

Since DBAL deprecates concept of fetch mode in favor usage specific to every fetch mode Connection methods, there are no easy way to avoid deprecated staff with QueryBuilder usage.

As an alternative to deprecated API DBAL may provide  new methods in QueryBuilder - `fetchAllAssociative()` and other, related to new fetch modes.

Implementation for fetch associative method in QueryBuilder may looks like this:
 
```php
public function fetchAllAssociative(): array
{
    // assert current state contains SELECT query
 
    return $this->connection->fetchAllAssociative($this->getSQL(), $this->params, $this->paramTypes);
}
```

andrew-demb:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #4461 

#### Summary

Added new `fetch*` methods (all non-deprecated fetch* methods from Connection class) to allow the use of better typing in fetching data.

Please let me know if I should write some tests and what to use in the description of the methods.

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -844,9 +844,9 @@ class Connection
      * to the first column and the values being an associative array representing the rest of the columns
      * and their values.
      *
-     * @param string                                           $query  SQL query
-     * @param list<mixed>|array<string, mixed>                 $params Query parameters
-     * @param array<int, int|string>|array<string, int|string> $types  Parameter types
+     * @param string                                                               $query  SQL query
+     * @param list<mixed>|array<string, mixed>                                     $params Query parameters
+     * @param array<int, int|string|Type|null>|array<string, int|string|Type|null> $types  Parameter types
      *
      * @return array<mixed,array<string,mixed>>
      *
--- a/src/Query/QueryBuilder.php
+++ b/src/Query/QueryBuilder.php
@@ -198,6 +198,108 @@ class QueryBuilder
     }
 
     /**
+     * Prepares and executes an SQL query and returns the first row of the result
+     * as an associative array.
+     *
+     * @return array<string, mixed>|false False is returned if no rows are found.
+     *
+     * @throws Exception
+     */
+    public function fetchAssociative()
+    {
+        return $this->connection->fetchAssociative($this->getSQL(), $this->params, $this->paramTypes);
+    }
+
+    /**
+     * Prepares and executes an SQL query and returns the first row of the result
+     * as a numerically indexed array.
+     *
+     * @return array<int, mixed>|false False is returned if no rows are found.
+     *
+     * @throws Exception
+     */
+    public function fetchNumeric()
+    {
+        return $this->connection->fetchNumeric($this->getSQL(), $this->params, $this->paramTypes);
+    }
+
+    /**
+     * Prepares and executes an SQL query and returns the value of a single column
+     * of the first row of the result.
+     *
+     * @return mixed|false False is returned if no rows are found.
+     *
+     * @throws Exception
+     */
+    public function fetchOne()
+    {
+        return $this->connection->fetchOne($this->getSQL(), $this->params, $this->paramTypes);
+    }
+
+    /**
+     * Prepares and executes an SQL query and returns the result as an array of numeric arrays.
+     *
+     * @return array<int,array<int,mixed>>
+     *
+     * @throws Exception
+     */
+    public function fetchAllNumeric(): array
+    {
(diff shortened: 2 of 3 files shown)


morozov:
> Please let me know if I should write some tests

Yes, please do. It could be a set of unit tests that all use the same `QueryBuilder` instance producing a couple of rows and a couple of columns and testing all the new methods. Something similar to [`ConverterTest`](https://github.com/doctrine/dbal/commit/7204a7df65225db28296d23d87bf958cb92c5ece#diff-05d06b0345acb4bb5df8e3283d6a516405a87e87837bde26ce23c1eef5ac9dcc).

> [...] what to use in the description of the methods.

You can copy-paste the description of the corresponding `Statement`/`Result` methods or leave them empty.

andrew-demb:
> It could be a set of unit tests that all use the same `QueryBuilder` instance producing a couple of rows and a couple of columns and testing all the new methods. Something similar to [`ConverterTest`](https://github.com/doctrine/dbal/commit/7204a7df65225db28296d23d87bf958cb92c5ece#diff-05d06b0345acb4bb5df8e3283d6a516405a87e87837bde26ce23c1eef5ac9dcc).

To be in sync - are you suggest a "fill" query builder with some data and use the mock connection with a strictly expected query to be bypassed to the needed method?

Example (expected query and other will be parametrized from data provider):
```php
    protected function setUp(): void
    {
        $this->conn = $this->createMock(Connection::class);
    }

    public function testFetchAssociative(): void
    {
        $qb = new QueryBuilder($this->conn);

        $this->conn->expects($this->once())
            ->method('fetchAssociative')
            ->with('SELECT some FROM table');

        $qb->select('some')
            ->from('table')
            ->fetchAssociative();
    }
```

Thank you for help.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -999,9 +999,9 @@ public function fetchAllAssociative(string $query, array $params = [], array $ty
      * Prepares and executes an SQL query and returns the result as an associative array with the keys
      * mapped to the first column and the values mapped to the second column.
      *
-     * @param string                                           $query  SQL query
-     * @param array<int, mixed>|array<string, mixed>           $params Query parameters
-     * @param array<int, int|string>|array<string, int|string> $types  Parameter types
+     * @param string                                                               $query  SQL query
+     * @param array<int, mixed>|array<string, mixed>                               $params Query parameters
+     * @param array<int, int|string|Type|null>|array<string, int|string|Type|null> $types  Parameter types
```
andrew-demb: Static analysis failed, because query builder may use `Type` objects passed to these Connection methods.

morozov:
Yeah, the above looks good. You can also add the assertions that not only the query but the parameters and types are passed downstream, and the value returned by the statement is returned. This is as much as these new methods do.

morozov:
As for the target branch, there's no plan to tag any new `2.x` releases potentially with the exception of improving the upgrade path to `3.0`. This improvement doesn't fall into the "upgrade path" category, so I believe it should target `3.1.x`.

andrew-demb:
Tests added.
Ready for review

morozov:
Looks good, @andrew-demb. Please squash, and I can merge it. Or I can squash it before the merge.

PowerKiKi:
I think this PR lacks the addition of a new method `executeStatement()`, to be really complete. It could be done in a second PR, but I feel it would be more coherent to add all methods together, don't you think ?

morozov (approved):

morozov:
> I think this PR lacks the addition of a new method `executeStatement()`, to be really complete. It could be done in a second PR, but I feel it would be more coherent to add all methods together, don't you think ?

This PR doesn't deprecate anything so it's complete by just adding new methods. If you want to work on the deprecation, in addition to `executeStatement()` we'll need to add `executeQuery()` as well. Otherwise, there will be no way to obtain a `Result` from the built query.

andrew-demb:
@morozov thank you for squash

@PowerKiKi sorry, I missed your messages before. I think a new deprecation and more methods may require an additional discussion.
Lets do it in additional PR.

PR #4491, opened 2021-01-30
Remove internal flag from TypeRegistry

ostrolucky:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

This class is almost 2 years old and never received BC breaking change. We need to rely on its API so we can build DI integration in DoctrineBundle

I've chosen 2.12 because since there are no changes here between 2.12 and higher, it doesn't matter, unless we intend to do BC break in this class in 2.x but not 3.x (super unlikely). But it does matter for SA analysis for people still using DBAL 2.x.

Related: https://github.com/doctrine/dbal/issues/2841, https://github.com/doctrine/DoctrineBundle/issues/1246

### DIFF
--- a/lib/Doctrine/DBAL/Types/Type.php
+++ b/lib/Doctrine/DBAL/Types/Type.php
@@ -195,9 +195,6 @@ abstract class Type
      */
     abstract public function getName();
 
-    /**
-     * @internal This method is only to be used within DBAL for forward compatibility purposes. Do not use directly.
-     */
     final public static function getTypeRegistry(): TypeRegistry
     {
         if (self::$typeRegistry === null) {
--- a/lib/Doctrine/DBAL/Types/TypeRegistry.php
+++ b/lib/Doctrine/DBAL/Types/TypeRegistry.php
@@ -12,8 +12,6 @@ use function in_array;
 /**
  * The type registry is responsible for holding a map of all known DBAL types.
  * The types are stored using the flyweight pattern so that one type only exists as exactly one instance.
- *
- * @internal TypeRegistry exists for forward compatibility, its API should not be considered stable.
  */
 final class TypeRegistry
 {


SenseException (dismissed):
The API that should not be considered stable is stable? I don't know about the background why it wasn't considered stable though, but it seems that prophecy didn't come true.

morozov:
As far as I understand from https://github.com/doctrine/dbal/pull/3354, `TypeRegistry` is an internal implementation detail of the `Type` class whose API could and still can be used to register types and it hasn't been deprecated in favor of `TypeRegistry`. Why do consumers need to use the alternative internal API?

If the plan to replace `Type` with `TypeRegistry`, would it make sense to do as planned?

ostrolucky:
> Why do consumers need to use the alternative internal API?

To be able to register instance of the classes. Type class is doing `new $className()` everywhere - that's unusable with DI. So that's why we need to use alternative API, because current API is bad. People are forced to use TypeRegistry, that's the only way how to register instance. But it's marked internal for some reason, even though it's stable for quite a while.

> If the plan to replace Type with TypeRegistry, would it make sense to do as planned?

I think it would, but what work and what timeframe are we talking? I thought there was some blocker or unknowns if nothing was being done about this plan further, even though pieces were put in place. If this is only about replacing `Type::` calls with `Type::getTypeRegistry()->`, I assume that should be pretty easy, but I'm not sure that's the plan here. Anything else might be complicated if BC is preserved and I don't want to wait much longer for this.

morozov:
> But it's marked internal for some reason, even though it's stable for quite a while.

Internal is not an antonym of stable.

> If this is only about replacing `Type::` calls with `Type::getTypeRegistry()->`, I assume that should be pretty easy, but I'm not sure that's the plan here.

I don't think anybody but @Majkl578 knows. I'd expect the type registry to be a dependency of the connection. Otherwise, it's not much better than the current static class.

morozov (approved):

SenseException (approved):

PR #4493, opened 2021-02-02
Use explicit `@return` when implementing base PHP types

nicolas-grekas:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

This PR replaces `{@inheritdoc}` that relate to internal PHP types by explicit `@return` annotations.

This removes the need to rely on external knowledge to get some understanding of the annotations.

Not that similar explicit annotations are already used in some places in the codebase.


nicolas-grekas:
Closing as all methods are deprecated.

PR #3421, opened 2019-01-02
Issue #950: Fixed cardinality issue where subquery could return more than one row

jwilson-cee:
Fixed cardinality issue where subquery could return more than than one row in getListTableForeignKeysSQL function

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #950 

#### Summary

Replaced "=" with "IN" in the query in the PostgreSqlPlatform->getListTableForeignKeysSQL() function to avoid getting a cardinality violation error thrown

PR #4494, opened 2021-02-08
Crash when there are 2 tables on different schemas

shemgp:
This fixes the crash:
```
   Doctrine\DBAL\Exception\DriverException  : An exception occurred while executing 'SELECT quote_ident(r.conname) as conname, pg_catalog.pg_get_constraintdef(r.oid, true) as condef
                  FROM pg_catalog.pg_constraint r
                  WHERE r.conrelid =
                  (
                      SELECT c.oid
                      FROM pg_catalog.pg_class c, pg_catalog.pg_namespace n
                      WHERE n.nspname NOT IN ('pg_catalog', 'information_schema', 'pg_toast') AND c.relname = 'migrations' AND n.nspname = ANY(current_schemas(false)) AND n.oid = c.relnamespace
                  )
                  AND r.contype = 'f'':

SQLSTATE[21000]: Cardinality violation: 7 ERROR:  more than one row returned by a subquery used as an expression
```
which is because I have two same named table in two schemas (public and kub). Laravel created `migration` tables in both schemas. 

So since the DDL of both tables are the same, the fix using `LIMIT 1` is just fine for me.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues |

### DIFF
--- a/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
+++ b/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php
@@ -41,8 +41,8 @@ abstract class AbstractMySQLDriver implements Driver, ExceptionConverterDriver, 
      *
      * @deprecated
      *
-     * @link https://dev.mysql.com/doc/refman/8.0/en/client-error-reference.html
-     * @link https://dev.mysql.com/doc/refman/8.0/en/server-error-reference.html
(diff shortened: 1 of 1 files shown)


morozov:
The original issue https://github.com/doctrine/dbal/issues/950 is still open. The proper solution is to make sure that the query uses the schema name explicitly.

Linking https://github.com/doctrine/dbal/pull/3421, https://github.com/doctrine/dbal/pull/3422 for reference.

PR #4505, opened 2021-02-22
Use better example for data retrieval

xelan:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

The existing example proposes/implies clear-text storage of passwords, which is highly discouraged. Furthermore, such an example sheds a negative light on the Doctrine project if security-conscious developers evaluate whether they should use the library.


greg0ire (approved):

SenseException (approved):

SenseException:
Good point. Thanks for the changes @xelan.

PR #4518, opened 2021-02-28
Remove Connection::$_schemaManager and ::getSchemaManager()

### LINKED ISSUE #4515: Memory/resource leaks in the wrapper connection
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.12.1

If a wrapper connection is instantiated, establishes a driver-level connection and then is unreferenced, it doesn't get garbage-collected by PHP until GC kicks in (in the best case) and doesn't close the underlying connection.

**Example 1**
On each iteration, the previously instantiated connection should be destructed and close its driver connection but it isn't the case:
```php
for ($i = 0; $i < 5; $i++) {
        'driver' => 'pdo_pgsql',
        'host'   => 'localhost',
        'port'   => '5432',
        'user'   => 'postgres',
        'password' => 'Passw0rd',
        'dbname' => 'postgres',
    ]);

    echo $conn->fetchOne('SELECT SUM(numbackends) FROM pg_stat_database'), PHP_EOL;
}

// 1
// 2
// 3
// 4
// 5
```
Adding `$conn->close()` to the loop solves the problem however it doesn't do anything than removing the reference to the driver-level connection: https://github.com/doctrine/dbal/blob/de5464df915ea7a2d42a07086f72119c3d1e4bfa/lib/Doctrine/DBAL/Connection.php#L772-L775

This means that during iteration the object doesn't get destructed.

**Example 2**

There seems to be an internal circular reference between the wrapper and some of its components that doesn't let the connection to be destructed (see `refcount=2`):

```php
$conn1 = DriverManager::getConnection([
    'driver' => 'pdo_pgsql',
]);
xdebug_debug_zval('conn1');
// conn1: (refcount=2, is_ref=0)=class Doctrine\DBAL\Connection { ... }

$conn2 = new stdClass();
xdebug_debug_zval('conn2');
// conn2: (refcount=1, is_ref=0)=class stdClass {  }
```

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | #4515

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,9 @@
 # Upgrade to 4.0
 
+## BC BREAK: Removed `Connection::$_schemaManager` and `Connection::getSchemaManager()`
+
+The `Connection` and `AbstractSchemaManager` classes used to have a reference on each other effectively making a circular reference. Use `createSchemaManager()` to instantiate a schema manager.
+
 ## BC BREAK: Removed `Connection::$_expr` property
 
 The `Connection` and `ExpressionBuilder` classes used to have a reference on each other effectively making a circular reference. Use `getExpressionBuilder()` to instantiate an expression builder.
--- a/docs/en/reference/schema-manager.rst
+++ b/docs/en/reference/schema-manager.rst
@@ -5,15 +5,15 @@ A Schema Manager instance helps you with the abstraction of the
 generation of SQL assets such as Tables, Sequences, Foreign Keys
 and Indexes.
 
-To retrieve the ``SchemaManager`` for your connection you can use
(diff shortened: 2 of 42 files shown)


SenseException:
The last 3.x branch should introduce a `createSchemaManager()` and deprecate `getSchemaManager()`. `createSchemaManager()` is a more appropriate name for 4.0.x.

morozov:
Agree. We should do the same for `getExpressionBuilder()`.

greg0ire (approved):

SenseException (approved):

PR #4525, opened 2021-03-01
[Mysql] ignore index length for spatial indexes

ottaviano:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3561

#### Summary

Skips index length option for SPATIAL index on MySql (cf initial issue and this PR #3970 for more details)

### DIFF
--- a/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
+++ b/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
@@ -91,7 +91,10 @@ class MySqlSchemaManager extends AbstractSchemaManager
                 $v['flags'] = ['SPATIAL'];
             }
 
-            $v['length'] = isset($v['sub_part']) ? (int) $v['sub_part'] : null;
+            // Ignore prohibited prefix `length` for spatial index
+            if (strpos($v['index_type'], 'SPATIAL') === false) {
+                $v['length'] = isset($v['sub_part']) ? (int) $v['sub_part'] : null;
+            }
 
             $tableIndexes[$k] = $v;
         }
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
@@ -111,6 +111,7 @@ class MySqlSchemaManagerTest extends SchemaManagerFunctionalTestCase
         $indexes = $this->schemaManager->listTableIndexes('spatial_index');
         self::assertArrayHasKey('s_index', $indexes);
         self::assertTrue($indexes['s_index']->hasFlag('spatial'));
+        self::assertSame([0 => null], $indexes['s_index']->getOption('lengths'));
     }
 
     public function testIndexWithLength(): void


lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
```diff
@@ -85,14 +85,15 @@ protected function _getPortableTableIndexesList($tableIndexes, $tableName = null
                 $v['primary'] = false;
             }
 
+            $v['length'] = isset($v['sub_part']) ? (int) $v['sub_part'] : null;
+
             if (strpos($v['index_type'], 'FULLTEXT') !== false) {
                 $v['flags'] = ['FULLTEXT'];
             } elseif (strpos($v['index_type'], 'SPATIAL') !== false) {
+                unset($v['length']);
```
morozov: Instead of setting and then unsetting length, should it be set only when needed? E.g. in a separate `if`-statement.

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
```diff
@@ -102,6 +102,7 @@ public function testSpatialIndex(): void
         $table->addColumn('point', 'point');
         $table->addIndex(['point'], 's_index');
         $table->addOption('engine', 'MyISAM');
+        $table->addOption('length', 42);
```
morozov: According to the documentation:
> Column prefix lengths are prohibited. The full width of each column is indexed.

Why are we passing a prohibited property? According to the issue description, it looks like MySQL would report the index length even if it wasn't specified during creation. Otherwise, why would we need to ignore it during introspection?

ottaviano: I thought if `length` was present in options, we should ignore it:
```php
@ORM\Index(name="point_inx", columns={"point"}, flags={"spatial"}, options={"length": 42})
```

lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php
ottaviano: fixed 👍

morozov:
@ottaviano, looks good. I just ran the new test w/o the fix, and it fails as expected:
```
$ phpunit -c mysqli.phpunit.xml --filter testSpatialIndex tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
PHPUnit 9.5.0 by Sebastian Bergmann and contributors.

F                                                                                   1 / 1 (100%)

Time: 00:00.221, Memory: 16.00 MB

There was 1 failure:

1) Doctrine\Tests\DBAL\Functional\Schema\MySqlSchemaManagerTest::testSpatialIndex
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
 Array &0 (
-    0 => null
+    0 => 32
 )

/Users/smorozov/Projects/dbal/tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php:114

FAILURES!
Tests: 1, Assertions: 3, Failures: 1.
```

morozov (approved):

tests/Doctrine/Tests/DBAL/Functional/Schema/MySqlSchemaManagerTest.php
morozov: As the test below shows, MySQL will return the index length even if it's not a partial index which conflicts with the DBAL semantics.

morozov:
Thanks, @ottaviano.

ottaviano:
Thank you, @morozov for review and fast merge and @ksaveras for initialising the fix of this issue 👍

ottaviano:
hi @morozov, would you know if the new release 2.12.2 will be created soon?
we are waiting for it to make a big update 🚀

morozov:
The plan is to release it right before [2.13.0](https://github.com/doctrine/dbal/milestone/88).

Issue #4280, opened 2020-09-21
Static analysis issue in 2.11


enumag:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.11.0

#### Summary

`ResultStatement::fetchAll()` is deprecated with message "Use fetchAllNumeric(), fetchAllAssociative() or fetchFirstColumn() instead."

The problem is that these new methods are not on ResultStatement interface and the `@return` annotation on `Connection::executeQuery()` was not changed which causes PHPStan to report an issue.

#### How to reproduce

When I change this code:

```
		$result = $connection->executeQuery($sql)->fetchAll();
```

according to the deprecation notice:

```
		$result = $connection->executeQuery($sql)->fetchAllAssociative();
```

PHPStan then reports:

```
Call to an undefined method Doctrine\DBAL\Driver\ResultStatement::fetchAllAssociative().
```

#### Expected behaviour

No PHPStan issues - `executeQuery` should have `@return Result`.

greg0ire:
Useful links: 

https://github.com/doctrine/dbal/blob/5d2282389438138b91b31742f7b7f4779e1a2122/lib/Doctrine/DBAL/Connection.php#L1130-L1134

https://github.com/doctrine/dbal/blob/5d2282389438138b91b31742f7b7f4779e1a2122/lib/Doctrine/DBAL/Driver/ResultStatement.php#L76-L97

https://github.com/doctrine/dbal/blob/2.11.x/lib/Doctrine/DBAL/Driver/Result.php

> executeQuery should have @return Result.

Maybe it should have `@return Result&ResultStatement`?

enumag:
Or possibly `@return Statement`? (`Doctrine\DBAL\Statement`)

greg0ire:
Why? `Doctrine\DBAL\Statement` does not implement `Doctrine\DBAL\Driver\ResultStatement`

enumag:
Ah sorry, I meant `Doctrine\DBAL\Driver\Statement`. But then again that doesn't implement `Result` either. These statements and result classes/interfaces are quite confusing to be honest.

morozov:
It's not reported by PHPStan on the library itself because during the rework I replaced all code blocks like `$connection->executeQuery($sql)->fetchAllAssociative()` with `$connection->fetchAllAssociative($sql)`.

While this is being fixed, you can do the same.

morozov:
[...]

PR #4513, opened 2021-02-27
Bump Psalm level to 2

morozov:
|      Q       |   A
|------------- | -----------
| Type         |improvement
| BC Break     | no

This patch shouldn't contain any breaking changes but I cannot confirm that with `roave-backward-compatibility-check`:
```
$ roave-backward-compatibility-check --from=upstream/2.12.x
Comparing from 1d6771f300bfdf5578634a3e1f0a8fe067775d16 to a1986e07dbb63b46a446f186cdf93db6c89eadb9...
Loading composer repositories with package information
<warning>Warning from https://repo.packagist.org: You are using an outdated version of Composer. Composer 2 is now available and you should upgrade. See https://getcomposer.org/2</warning>
Updating dependencies
Package operations: 2 installs, 0 updates, 0 removals
  - Installing doctrine/cache (1.10.2): Loading from cache
  - Installing doctrine/event-manager (1.1.1): Loading from cache
Writing lock file
2 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
Loading composer repositories with package information
<warning>Warning from https://repo.packagist.org: You are using an outdated version of Composer. Composer 2 is now available and you should upgrade. See https://getcomposer.org/2</warning>
Updating dependencies
Package operations: 2 installs, 0 updates, 0 removals
  - Installing doctrine/cache (1.10.2): Loading from cache
  - Installing doctrine/event-manager (1.1.1): Loading from cache
Writing lock file
2 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
fish: 'roave-backward-compatibility-ch…' terminated by signal SIGSEGV (Address boundary error)
```

Fails with a segfault on PHP 7.4.15, both on Linux and macOS.

PR #4529, opened 2021-03-04
Improve forward compatibility between 2.x and 3.0

mdumoulin:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #4280

#### Summary

The current Doctrine\DBAL\Connection executeQuery(...) and executeCacheQuery() return a ResultStatement interface object. This PR try to improve forward compatibility between v2.x & v3.0 by adding some of missing features to Result returned by Connection :
- fetchNumeric()
- fetchAssociative()
- fetchOne()
- fetchAllNumeric()
- fetchAllAssociative()
- fetchAllKeyValue()
- fetchAllAssociativeIndexed()
- fetchFirstColumn()
- iterateNumeric()
- iterateAssociative()
- iterateKeyValue()
- iterateAssociativeIndexed()
- iterateColumn()
- rowCount()
- columnCount()
- free()

I used this version as reference : https://github.com/doctrine/dbal/blob/3.0.x/src/Result.php 

##### Static analysis error
Here under, methods which throw a static analysis error in phpstan but actually work. This is all methods defined in Abstraction\Result interface :

```php
/** @var Doctrine\DBAL\Connection $connection */
$connection->executeQuery('SELECT "something"')->fetchNumeric();
$connection->executeQuery('SELECT "something"')->fetchAssociative();
$connection->executeQuery('SELECT "something"')->fetchOne();
$connection->executeQuery('SELECT "something"')->fetchAllNumeric();
$connection->executeQuery('SELECT "something"')->fetchAllAssociative();
$connection->executeQuery('SELECT "something"')->fetchFirstColumn();
$connection->executeQuery('SELECT "something"')->iterateNumeric();
$connection->executeQuery('SELECT "something"')->iterateAssociative();
$connection->executeQuery('SELECT "something"')->iterateNumeric();
$connection->executeQuery('SELECT "something"')->iterateColumn();
```

The outpout pattern when running phpstan is the following :
```text
Call to an undefined method Doctrine\DBAL\Driver\ResultStatement::fetchOne().
```

#### Not available methods
Here under, methods which throw undefined method fatal & static analyse errors. This is all methods added in Abstraction\V3Result interface :

```php
$connection->executeQuery('SELECT "something"')->fetchAllKeyValue();
$connection->executeQuery('SELECT "something"')->fetchAllAssociativeIndexed();
$connection->executeQuery('SELECT "something"')->iterateKeyValue();
$connection->executeQuery('SELECT "something"')->iterateAssociativeIndexed();
```

The output pattern when trying to use one of those methods :
```text
Fatal error: Uncaught Error: Call to undefined method Doctrine\DBAL\Driver\PDO\Statement::fetchAllKeyValue() in /var/www/test.php on line 41

Error: Call to undefined method Doctrine\DBAL\Driver\PDO\Statement::fetchAllKeyValue() in /var/www/test.php on line 41

Call Stack:
    0.0006     415576   1. {main}() /var/www/test.php:0
```

#### Implementation key points

The strategy is to :
- Add a new interfaces with missing V3 Result features
- Make Connection return an object implementing ResultStatement & V3Result interfaces

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -4,7 +4,6 @@ namespace Doctrine\DBAL;
 
 use Closure;
 use Doctrine\Common\EventManager;
-use Doctrine\DBAL\Abstraction\Result;
 use Doctrine\DBAL\Cache\ArrayStatement;
 use Doctrine\DBAL\Cache\CacheException;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
@@ -19,6 +18,7 @@ use Doctrine\DBAL\Exception\NoKeyValue;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Query\Expression\ExpressionBuilder;
 use Doctrine\DBAL\Query\QueryBuilder;
+use Doctrine\DBAL\Result as BaseResult;
 use Doctrine\DBAL\Schema\AbstractSchemaManager;
 use Doctrine\DBAL\Types\Type;
 use Throwable;
@@ -616,13 +616,11 @@ class Connection implements DriverConnection
     public function fetchAssociative(string $query, array $params = [], array $types = [])
     {
         try {
-            $stmt = $this->executeQuery($query, $params, $types);
+            $stmt = $this->ensureForwardCompatibilityStatement(
+                $this->executeQuery($query, $params, $types)
+            );
 
-            if ($stmt instanceof Result) {
-                return $stmt->fetchAssociative();
-            }
-
-            return $stmt->fetch(FetchMode::ASSOCIATIVE);
+            return $stmt->fetchAssociative();
         } catch (Throwable $e) {
             $this->handleExceptionDuringQuery($e, $query, $params, $types);
         }
@@ -643,13 +641,11 @@ class Connection implements DriverConnection
     public function fetchNumeric(string $query, array $params = [], array $types = [])
     {
         try {
-            $stmt = $this->executeQuery($query, $params, $types);
+            $stmt = $this->ensureForwardCompatibilityStatement(
+                $this->executeQuery($query, $params, $types)
+            );
 
-            if ($stmt instanceof Result) {
-                return $stmt->fetchNumeric();
-            }
-
-            return $stmt->fetch(FetchMode::NUMERIC);
+            return $stmt->fetchNumeric();
         } catch (Throwable $e) {
             $this->handleExceptionDuringQuery($e, $query, $params, $types);
         }
@@ -670,13 +666,11 @@ class Connection implements DriverConnection
     public function fetchOne(string $query, array $params = [], array $types = [])
     {
         try {
-            $stmt = $this->executeQuery($query, $params, $types);
+            $stmt = $this->ensureForwardCompatibilityStatement(
+                $this->executeQuery($query, $params, $types)
+            );
 
-            if ($stmt instanceof Result) {
-                return $stmt->fetchOne();
-            }
-
-            return $stmt->fetch(FetchMode::COLUMN);
+            return $stmt->fetchOne();
         } catch (Throwable $e) {
             $this->handleExceptionDuringQuery($e, $query, $params, $types);
         }
@@ -956,13 +950,11 @@ class Connection implements DriverConnection
     public function fetchAllNumeric(string $query, array $params = [], array $types = []): array
     {
         try {
-            $stmt = $this->executeQuery($query, $params, $types);
+            $stmt = $this->ensureForwardCompatibilityStatement(
+                $this->executeQuery($query, $params, $types)
+            );
 
-            if ($stmt instanceof Result) {
-                return $stmt->fetchAllNumeric();
-            }
-
-            return $stmt->fetchAll(FetchMode::NUMERIC);
+            return $stmt->fetchAllNumeric();
         } catch (Throwable $e) {
             $this->handleExceptionDuringQuery($e, $query, $params, $types);
         }
@@ -982,13 +974,11 @@ class Connection implements DriverConnection
     public function fetchAllAssociative(string $query, array $params = [], array $types = []): array
     {
         try {
-            $stmt = $this->executeQuery($query, $params, $types);
+            $stmt = $this->ensureForwardCompatibilityStatement(
+                $this->executeQuery($query, $params, $types)
+            );
 
-            if ($stmt instanceof Result) {
-                return $stmt->fetchAllAssociative();
-            }
-
-            return $stmt->fetchAll(FetchMode::ASSOCIATIVE);
+            return $stmt->fetchAllAssociative();
         } catch (Throwable $e) {
             $this->handleExceptionDuringQuery($e, $query, $params, $types);
         }
@@ -1061,13 +1051,11 @@ class Connection implements DriverConnection
     public function fetchFirstColumn(string $query, array $params = [], array $types = []): array
     {
         try {
-            $stmt = $this->executeQuery($query, $params, $types);
+            $stmt = $this->ensureForwardCompatibilityStatement(
+                $this->executeQuery($query, $params, $types)
+            );
 
-            if ($stmt instanceof Result) {
-                return $stmt->fetchFirstColumn();
-            }
-
-            return $stmt->fetchAll(FetchMode::COLUMN);
+            return $stmt->fetchFirstColumn();
         } catch (Throwable $e) {
             $this->handleExceptionDuringQuery($e, $query, $params, $types);
         }
@@ -1087,15 +1075,11 @@ class Connection implements DriverConnection
     public function iterateNumeric(string $query, array $params = [], array $types = []): Traversable
     {
         try {
-            $stmt = $this->executeQuery($query, $params, $types);
+            $stmt = $this->ensureForwardCompatibilityStatement(
+                $this->executeQuery($query, $params, $types)
+            );
 
-            if ($stmt instanceof Result) {
-                yield from $stmt->iterateNumeric();
-            } else {
-                while (($row = $stmt->fetch(FetchMode::NUMERIC)) !== false) {
-                    yield $row;
-                }
-            }
+            yield from $stmt->iterateNumeric();
         } catch (Throwable $e) {
             $this->handleExceptionDuringQuery($e, $query, $params, $types);
         }
@@ -1116,15 +1100,11 @@ class Connection implements DriverConnection
     public function iterateAssociative(string $query, array $params = [], array $types = []): Traversable
     {
         try {
-            $stmt = $this->executeQuery($query, $params, $types);
+            $stmt = $this->ensureForwardCompatibilityStatement(
+                $this->executeQuery($query, $params, $types)
+            );
 
-            if ($stmt instanceof Result) {
-                yield from $stmt->iterateAssociative();
-            } else {
-                while (($row = $stmt->fetch(FetchMode::ASSOCIATIVE)) !== false) {
-                    yield $row;
-                }
-            }
+            yield from $stmt->iterateAssociative();
         } catch (Throwable $e) {
             $this->handleExceptionDuringQuery($e, $query, $params, $types);
         }
@@ -1189,15 +1169,11 @@ class Connection implements DriverConnection
     public function iterateColumn(string $query, array $params = [], array $types = []): Traversable
     {
         try {
-            $stmt = $this->executeQuery($query, $params, $types);
+            $stmt = $this->ensureForwardCompatibilityStatement(
+                $this->executeQuery($query, $params, $types)
+            );
 
-            if ($stmt instanceof Result) {
-                yield from $stmt->iterateColumn();
-            } else {
-                while (($value = $stmt->fetch(FetchMode::COLUMN)) !== false) {
-                    yield $value;
-                }
-            }
+            yield from $stmt->iterateColumn();
         } catch (Throwable $e) {
             $this->handleExceptionDuringQuery($e, $query, $params, $types);
         }
@@ -1235,7 +1211,7 @@ class Connection implements DriverConnection
      * @param array<int, mixed>|array<string, mixed>                               $params Query parameters
      * @param array<int, int|string|Type|null>|array<string, int|string|Type|null> $types  Parameter types
      *
-     * @return ResultStatement The executed statement.
+     * @return ResultStatement&BaseResult The executed statement.
      *
      * @throws Exception
      */
@@ -1281,7 +1257,7 @@ class Connection implements DriverConnection
             $logger->stopQuery();
         }
 
-        return $stmt;
+        return $this->ensureForwardCompatibilityStatement($stmt);
     }
 
     /**
@@ -1291,7 +1267,7 @@ class Connection implements DriverConnection
      * @param array<int, mixed>|array<string, mixed>                               $params Query parameters
      * @param array<int, int|string|Type|null>|array<string, int|string|Type|null> $types  Parameter types
      *
-     * @return ResultStatement
+     * @return ResultStatement&BaseResult
      *
      * @throws CacheException
      */
@@ -1332,7 +1308,19 @@ class Connection implements DriverConnection
 
         $stmt->setFetchMode($this->defaultFetchMode);
 
-        return $stmt;
+        return $this->ensureForwardCompatibilityStatement($stmt);
+    }
+
+    /**
+     * @return ResultStatement&BaseResult
+     */
+    private function ensureForwardCompatibilityStatement(ResultStatement $stmt)
+    {
+        if ($stmt instanceof BaseResult) {
+            return $stmt;
+        }
+
+        return new ForwardCompatibility\Result($stmt);
     }
 
     /**
--- a/tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
@@ -4,8 +4,7 @@ namespace Doctrine\Tests\DBAL\Functional;
 
 use Doctrine\Common\Cache\ArrayCache;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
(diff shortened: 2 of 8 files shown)


lib/Doctrine/DBAL/Cache/ArrayStatement.php
```diff
@@ -19,7 +19,7 @@
 /**
  * @deprecated
  */
-class ArrayStatement implements IteratorAggregate, ResultStatement, Result
+class ArrayStatement implements IteratorAggregate, ResultStatement
```
morozov: This is a breaking change. Please try using [Roave/BackwardCompatibilityCheck](https://github.com/Roave/BackwardCompatibilityCheck) (see https://github.com/doctrine/dbal/pull/4513 for example).

mdumoulin: Thanks for your quick return. It should be alright for BC break part now.

lib/Doctrine/DBAL/Cache/ArrayStatement.php
```diff
@@ -19,7 +20,7 @@
 /**
  * @deprecated
  */
-class ArrayStatement implements IteratorAggregate, ResultStatement, Result
+class ArrayStatement implements IteratorAggregate, ResultStatement, V3Result
```
morozov: This is still a breaking change.

mdumoulin: Sorry, I don't understand why. This class still implements Result through V3Result. By the way, I used roave_bc_check tool and it did not throw BC break warning anymore. Can you explain it to me please ? 

Thanks again for your quick feedback.

morozov: Sorry, you're right. I didn't notice that `V3Result` extends `Result`. I'll try to review the changes more thoroughly in the next few days. In the meanwhile, please fix the Psalm failure and provide a couple of examples of the code that gets analyzed incorrectly w/o these changes.

lib/Doctrine/DBAL/Abstraction/V3Result.php
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Doctrine\DBAL\Abstraction;
+
+use Doctrine\DBAL\Driver\Exception;
+use Traversable;
+
+/**
+ * @see https://github.com/doctrine/dbal/blob/3.0.x/src/Result.php
+ */
+interface V3Result extends Result
```
beberlei: Couldn't this already be an interface called `Doctrine\DBAL\Result`? 

It could be documented as a docblock that it will turn from interface into class in 3.0.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1235,7 +1194,7 @@ public function prepare($sql)
      * @param array<int, mixed>|array<string, mixed>                               $params Query parameters
      * @param array<int, int|string|Type|null>|array<string, int|string|Type|null> $types  Parameter types
      *
-     * @return ResultStatement The executed statement.
+     * @return ResultStatement&V3Result The executed statement.
```
beberlei: I would rather as mentioned above this being `Doctrine\DBAL\Result`, because that is the type it will become in 3.0

beberlei (commented):
This is very important, however I think we should be bolder and move the `V3Result` to be `Doctrine\DBAL\Result` otherwise the API will change again in 3.0 and you cannot rely on typing the result tpye, example:

```php
use Doctrine\DBAL\Result;

public function performQuery()
{
     $result = $this->connection->executeQuery('SELECT * FROM orders');
     return $this->convertQueryResult($result);
}

private function convertQueryResult(Result $result)
{
}
```

lib/Doctrine/DBAL/Abstraction/V3Result.php
morozov: This should work. Otherwise, I'd rather use a more specific term like `ForwardCompatibility` and the corresponding namespace (see #4019) to simplify the cleanup.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1332,6 +1291,18 @@ public function executeCacheQuery($sql, $params, $types, QueryCacheProfile $qcp)
 
         $stmt->setFetchMode($this->defaultFetchMode);
 
+        return $this->createV3CompatStatement($stmt);
+    }
+
+    /**
+     * @return ResultStatement&V3Result
+     */
+    private function createV3CompatStatement(ResultStatement $stmt)
```
morozov: Please use "forward compatibility" instead of "v3". Also, the method doesn't unconditionally create a statement as the name implies. Would something like `ensureStatementForwardCompatibility()` make sense.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1332,6 +1291,18 @@ public function executeCacheQuery($sql, $params, $types, QueryCacheProfile $qcp)
...
+    {
+        if (! $stmt instanceof V3Result) {
```
morozov: Negation can be avoided to improve readability:
```php
if ($stmt instanceof V3Result) {
    return $stmt;
}

return new V3CompatStatement($stmt);
```

lib/Doctrine/DBAL/Driver/V3CompatStatement.php
```diff
@@ -0,0 +1,203 @@
...
+
+namespace Doctrine\DBAL\Driver;
+
+use Doctrine\DBAL\Abstraction\V3Result;
+use Doctrine\DBAL\Exception;
+use IteratorAggregate;
+use PDO;
+use Traversable;
+
+class V3CompatStatement implements IteratorAggregate, ResultStatement, V3Result
```
morozov: Please move to a separate `ForwardCompatibility` namespace.

lib/Doctrine/DBAL/Driver/V3CompatStatement.php
```diff
@@ -0,0 +1,203 @@
...
+    {
+        return $this->stmt->setFetchMode($fetchMode, $arg2, $arg3);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function fetch($fetchMode = null, $cursorOrientation = PDO::FETCH_ORI_NEXT, $cursorOffset = 0)
+    {
+        return $this->stmt->fetch($fetchMode, $cursorOrientation, $cursorOffset);
```
morozov: I don't know if it's applicable to this case but in some cases, just passing the arguments down the line doesn't work because the underlying driver relies on the number of arguments passed, not their values: https://github.com/doctrine/dbal/blob/84553e361d7ab15b16c45e1a08572d07e8a992e0/lib/Doctrine/DBAL/Driver/PDOStatement.php#L127-L136

The fact that the build is passing means that most likely this case isn't tested.

lib/Doctrine/DBAL/Driver/V3CompatStatement.php
```diff
@@ -0,0 +1,203 @@
...
+    {
+        return FetchUtils::iterateColumn($this);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function rowCount()
+    {
+        throw Exception::notSupported('rowCount');
```
morozov: Doesn't it make the implementation incompatible with the old one? It is possible to count rows on the statement in 2.x.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -618,11 +619,7 @@ public function fetchAssociative(string $query, array $params = [], array $types
         try {
             $stmt = $this->executeQuery($query, $params, $types);
 
-            if ($stmt instanceof Result) {
```
morozov, line 621: What if `executeQuery()` is overridden in a custom class and doesn't implement the newly introduced wrapping?

lib/Doctrine/DBAL/Driver/FetchUtils.php
```diff
@@ -57,6 +62,44 @@ public static function fetchAllAssociative(Result $result): array
...
     }
 
+    /**
+     * Returns an array containing the values of the first column of the result.
+     *
+     * @return array<mixed,mixed>
+     *
+     * @throws Exception
+     */
+    public static function fetchAllAssociativeIndexed(Result $result): array
```
morozov: I'd rather have these methods implemented in a separate class under the `ForwardCompatibility` namespace. This way, it will be clear that it won't need to be merged up. Also, it reduces the risk of accidentally introducing an API that isn't supported in `3.x`.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
```diff
@@ -427,6 +428,22 @@ public function fetchAllAssociative(): array
         return FetchUtils::fetchAllAssociative($this);
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function fetchAllKeyValue(): array
```
morozov: Drivers do not implement this method in `3.x`. Why does it need to be implemented in `2.x`?

lib/Doctrine/DBAL/Abstraction/V3Result.php
mdumoulin: Seems a good idea to name it after the name used in 3.0.

lib/Doctrine/DBAL/Connection.php
mdumoulin: Done

lib/Doctrine/DBAL/Connection.php
mdumoulin: Done

lib/Doctrine/DBAL/Driver/V3CompatStatement.php
mdumoulin: Done

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
mdumoulin: It is not available in the driver implementation, but feature is ensured by the Result class : https://github.com/doctrine/dbal/blob/3.0.x/src/Result.php#L120-L131

It seems to me we have to ensure that user can write the following :
```php
$result = $connection->executeQuery(...)->fetchAllKeyValue();
```

lib/Doctrine/DBAL/Driver/FetchUtils.php
mdumoulin: Added methods are used exactly in the same way that previously existing ones : fetchOne(), fetchAllNumeric(), fetchAllAssociative() ... etc. Wouldn't it be weird to split them in 2 separate classes ?

lib/Doctrine/DBAL/Connection.php
mdumoulin: You are right, it will break this kind of extension. I propose to use the ensureForwardCompatibleStatement() that should do the work.

lib/Doctrine/DBAL/Driver/V3CompatStatement.php
mdumoulin: The class decorate the ResultStatement which does not contain this method but some subclasses implements it. Maybe a simple method_exists() check can do the job ?

lib/Doctrine/DBAL/Driver/V3CompatStatement.php
mdumoulin: Not sure to fully understand in which case it could cause an issue. Anyway, all drivers statements of the library have been migrated on forward compatible Result interface, so they will not be wrapped in the compatibility statement object.

https://github.com/doctrine/dbal/blob/c4e4ce67bde3d407db27b157306373d52a6b0bec/lib/Doctrine/DBAL/Connection.php#L1308-L1315

lib/Doctrine/DBAL/Driver/FetchUtils.php
morozov: The point is that drivers must implement only a subset of the methods while the ones you're adding are implemented at the wrapper level. You're adding wrapper-specific methods to the driver level.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Statement.php
morozov: It creates a BC break in 3.0.0 because these methods won't exist there. The driver layer must implement only a subset of `fetch*()` methods and none of the `iterate*()` ones.

mdumoulin: Ok, I did not see that but it make perfect senses. I will remove the support of Doctrine\PDO\Result interface by Drivers and use an object to wrap the Driver\Result as it is done in 3.0.

lib/Doctrine/DBAL/Cache/ArrayStatement.php
```diff
@@ -4,12 +4,13 @@
 
 use ArrayIterator;
 use Doctrine\DBAL\Driver\FetchUtils;
-use Doctrine\DBAL\Driver\Result;
 use Doctrine\DBAL\Driver\ResultStatement;
 use Doctrine\DBAL\FetchMode;
+use Doctrine\DBAL\Result;
```
morozov: An array statement must implement the driver-level API, not the wrapper level one.

lib/Doctrine/DBAL/Cache/ArrayStatement.php
```diff
@@ -215,6 +216,22 @@ public function fetchAllAssociative(): array
...
+     */
+    public function fetchAllKeyValue(): array
+    {
+        return FetchUtils::fetchAllKeyValue($this);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fetchAllAssociativeIndexed(): array
```
morozov: This method doesn't belong here.

lib/Doctrine/DBAL/Cache/ResultCacheStatement.php
```diff
@@ -35,7 +37,7 @@
  *
  * @deprecated
  */
-class ResultCacheStatement implements IteratorAggregate, ResultStatement, Result
+class ResultCacheStatement implements IteratorAggregate, ResultStatement, ForwardCompatibilityResult
```
morozov: It shouldn't implement the wrapper-level result.

mdumoulin:
@morozov Drivers statements no longer implements wrapper level Result. It break some unit tests because of ResultStatement wrapper. The following code will no longer work for example.

```php
$statement = $connection->executeQuery('SELECT "something"');

if ($statement instanceof ArrayStatement) {
    // True in 2.12.x version, false in this branch
}
```

I can't see any generic solution for this kind of edge case. Maybe we should let the user enable the Statement wrapper on purpose. We will avoid the risk of regression and allow to resolve deprecation on 2.x. It can be implemented by adding a ForwardCompatibility\Connection wrapper for example.

morozov:
> Maybe we should let the user enable the Statement wrapper on purpose. We will avoid the risk of regression and allow to resolve deprecation on 2.x. It can be implemented by adding a ForwardCompatibility\Connection wrapper for example.

If it's an opt-in layer, then there will be no improvement for static analysis, because the analyzers won't be able to detect the opt-in. As for the examples that currently don't work, e.g.:
```php
$connection->executeQuery('SELECT "something"')->fetchNumeric();
```
There's already an alternative:
```php
$connection->fetchNumeric('SELECT "something"');
```

beberlei:
Yes this edge case break is ok imho, the api is returning a statement, expecting a specific statement is not part of the interface

tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
```diff
@@ -238,25 +238,25 @@ public function testFetchingAllRowsSavesCache(callable $fetchAll): void
     public static function fetchAllProvider(): iterable
     {
         yield 'fetchAll' => [
-            static function (ResultCacheStatement $statement): void {
+            static function (ForwardCompatibility\Result $statement): void {
```
morozov: The expected `ResultCacheStatement` type was unnecessarily specific. Let's loosen it to `Driver\ResultStatement`. If it had been implemented this way, it wouldn't require a change now.

morozov (commented):
Looks good.

tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php
```diff
@@ -238,25 +238,25 @@ public function testFetchingAllRowsSavesCache(callable $fetchAll): void
...
         yield 'fetchAllNumeric' => [
-            static function (ResultCacheStatement $statement): void {
+            static function (ForwardCompatibility\Result $statement): void {
                 $statement->fetchAllNumeric();
             },
         ];
 
         yield 'fetchFirstColumn' => [
-            static function (ResultCacheStatement $result): void {
+            static function (ForwardCompatibility\Result $result): void {
```
morozov: https://github.com/doctrine/dbal/pull/4529#discussion_r593785306 applies to all the above cases.

mdumoulin: Not sure to understand here. ResultStament interface does not contain the fetchAllAssociative(), fetchAllNumeric() &  fetchFirstColumn() methods. I can use the Doctrine\DBAL\Result interface, if we prefer not relying on concrete object.

morozov: Sorry, it should have been `Driver\Result`.

morozov: Alright, so the expected type was not unnecessarily specific: it was the type that implemented both of the old and new APIs. It still implements them but now the `fetch*()` methods return another class that implements both interfaces as well. Looks good 👍

morozov:
@mdumoulin It looks good. I'd like all the changes to be squashed in one commit. Please do or otherwise, I'll do it later before the merge.

mdumoulin:
@morozov Great ! I just squashed all my commits in a single one.

morozov (approved):

morozov:
Thanks, @mdumoulin!

mdumoulin:
@morozov @beberlei Thanks again for your quick feedback & review.

Issue #4533, opened 2021-03-06
The $driverOptions argument of PDO\Statement#bindParam() is inconsistently accessible to the caller


morozov:
|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.0.x

The issue can be reproduced without custom code on 3.0 but is relevant to any DBAL version. `PDO\Statement#bindParam()` adds the `$driverOptions` argument which isn't part of the interface being implemented:
https://github.com/doctrine/dbal/blob/789d900236e04ddd22d3ef56b42a9ba0a3ee781a/src/Driver/PDO/Statement.php#L65
https://github.com/doctrine/dbal/blob/4cc12da9c6bbf3bcb5222fe56fd1f4d3bfa07bbc/src/Driver/Statement.php#L59

If a PDO statement is wrapped in another driver-level statement implementation, the `$driverOptions` argument becomes inaccessible to the caller:
```php
function test(?Configuration $config) {
    $conn = DriverManager::getConnection([
        'driver'   => 'pdo_sqlsrv',
        'user'     => 'sa',
        'password' => 'Passw0rd',
        'dbname'   => 'doctrine',
    ], $config);

    $var = 'Test';

    $stmt = $conn->getWrappedConnection()->prepare("SELECT sql_variant_property(?, 'BaseType')");
    $stmt->bindParam(1, $var, PDO::PARAM_STR, 0, PDO::SQLSRV_ENCODING_SYSTEM);

    echo $stmt->execute()->fetchOne(), PHP_EOL;
}

test(null);
// varchar

test((new Configuration())->setMiddlewares([
    new Doctrine\DBAL\Portability\Middleware(
        Doctrine\DBAL\Portability\Connection::PORTABILITY_ALL,
        Doctrine\DBAL\ColumnCase::LOWER
    )
]));
// nvarchar
```

morozov:
Closed via #4546.

PR #4534, opened 2021-03-06
Deprecate Statement::bindParam($driverOptions)

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation

See https://github.com/doctrine/dbal/issues/4533.

The parameter is only supported by the PDO-based drivers and only at the driver level. It's usage in PDO is barely [documented](https://www.php.net/manual/en/pdostatement.bindparam#refsect1-pdostatement.bindparam-parameters). The usage of the parameter is not covered by the existing tests.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,6 +1,11 @@
 # Upgrade to 3.1
 
-## Deprecated `Connection::$_schemaManager` and `Connection::getSchemaManager()` 
+## Deprecated `$driverOptions` argument of `PDO\Statement::bindParam()` and `PDO\SQLSrv\Statement::bindParam()`
+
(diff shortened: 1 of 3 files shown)


greg0ire (dismissed):

SenseException (commented):
Does the deprecation need to be referenced in the docs? https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/data-retrieval-and-manipulation.html#using-prepared-statements

morozov:
> Does the deprecation need to be referenced in the docs?

I don't see `$driverOptions` being mentioned on this page. Why reference the deprecation of something that was never even documented?

greg0ire (approved):

SenseException (approved):
I wasn't sure if this was indirectly related in some way to the docs.

PR #4535, opened 2021-03-06
Deprecations

beberlei:
Turns all the possible deprecated APIs into using the doctrine/deprecations API.

This includes deprecations on everything that can be done with code at runtime and is missing only the deprecation of `fetch` methods on the `Statement` class, which need this PR https://github.com/doctrine/dbal/pull/4529 for forwards compatibility to suggest a migration from `Statement::fetch*` to `Connection::executeQuery->fetch*()`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,7 @@
         "php": "^7.3 || ^8",
         "ext-pdo": "*",
         "doctrine/cache": "^1.0",
+        "doctrine/deprecations": "^0.5.3",
         "doctrine/event-manager": "^1.0"
     },
     "require-dev": {
--- a/lib/Doctrine/DBAL/Configuration.php
+++ b/lib/Doctrine/DBAL/Configuration.php
@@ -5,6 +5,7 @@ namespace Doctrine\DBAL;
 use Doctrine\Common\Cache\Cache;
 use Doctrine\DBAL\Logging\SQLLogger;
 use Doctrine\DBAL\Schema\AbstractAsset;
+use Doctrine\Deprecations\Deprecation;
 
 use function preg_match;
 
@@ -79,6 +80,12 @@ class Configuration
      */
     public function setFilterSchemaAssetsExpression($filterExpression)
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/3316',
+            'Configuration::setFilterSchemaAssetsExpression() is deprecated, use setSchemaAssetsFilter() instead.'
+        );
+
         $this->_attributes['filterSchemaAssetsExpression'] = $filterExpression;
         if ($filterExpression) {
             $this->_attributes['filterSchemaAssetsExpressionCallable']
@@ -97,6 +104,12 @@ class Configuration
      */
     public function getFilterSchemaAssetsExpression()
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/3316',
+            'Configuration::getFilterSchemaAssetsExpression() is deprecated, use getSchemaAssetsFilter() instead.'
+        );
+
         return $this->_attributes['filterSchemaAssetsExpression'] ?? null;
     }
 
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -2142,6 +2142,12 @@ class Connection implements DriverConnection
      */
     public function ping()
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/4119',
+            'Retry and reconnecting lost connections now happens automatically, ping() will be removed in DBAL 3.'
+        );
+
         $connection = $this->getWrappedConnection();
 
         if ($connection instanceof PingableConnection) {
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1582,6 +1582,12 @@ class Connection implements DriverConnection
      */
     public function errorCode()
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/3507',
+            'Connection::errorCode() is deprecated, use getCode() or getSQLState() on Exception instead.'
+        );
+
         return $this->getWrappedConnection()->errorCode();
     }
 
@@ -1592,6 +1598,12 @@ class Connection implements DriverConnection
      */
     public function errorInfo()
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/3507',
+            'Connection::errorInfo() is deprecated, use getCode() or getSQLState() on Exception instead.'
+        );
+
         return $this->getWrappedConnection()->errorInfo();
     }
 
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1376,6 +1376,12 @@ class Connection implements DriverConnection
      */
     public function project($sql, array $params, Closure $function)
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/3823',
+            'Connection::project() is deprecated without replacement, implement data projections in your own code.'
+        );
+
         $result = [];
         $stmt   = $this->executeQuery($sql, $params);
 
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -21,6 +21,7 @@ use Doctrine\DBAL\Query\Expression\ExpressionBuilder;
 use Doctrine\DBAL\Query\QueryBuilder;
 use Doctrine\DBAL\Schema\AbstractSchemaManager;
 use Doctrine\DBAL\Types\Type;
+use Doctrine\Deprecations\Deprecation;
 use Throwable;
 use Traversable;
 
@@ -542,6 +543,12 @@ class Connection implements DriverConnection
      */
     public function setFetchMode($fetchMode)
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/4019',
+            'Default Fetch Mode configuration is deprecated, use explicit Connection::fetch*() APIs instead.'
+        );
+
         $this->defaultFetchMode = $fetchMode;
     }
 
@@ -561,6 +568,12 @@ class Connection implements DriverConnection
      */
     public function fetchAssoc($sql, array $params = [], array $types = [])
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/4019',
+            'Connection::fetchAssoc() is deprecated, use Connection::fetchAssociative() API instead.'
+        );
+
         return $this->executeQuery($sql, $params, $types)->fetch(FetchMode::ASSOCIATIVE);
     }
 
@@ -578,6 +591,12 @@ class Connection implements DriverConnection
      */
     public function fetchArray($sql, array $params = [], array $types = [])
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/4019',
+            'Connection::fetchArray() is deprecated, use Connection::fetchNumeric() API instead.'
+        );
+
         return $this->executeQuery($sql, $params, $types)->fetch(FetchMode::NUMERIC);
     }
 
@@ -598,6 +617,12 @@ class Connection implements DriverConnection
      */
     public function fetchColumn($sql, array $params = [], $column = 0, array $types = [])
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/4019',
+            'Connection::fetchColumn() is deprecated, use Connection::fetchOne() API instead.'
+        );
+
         return $this->executeQuery($sql, $params, $types)->fetchColumn($column);
     }
 
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -1399,6 +1399,12 @@ class Connection implements DriverConnection
      */
     public function query()
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/4163',
+            'Connection::query() is deprecated, use Connection::executeQuery() instead.'
+        );
+
         $connection = $this->getWrappedConnection();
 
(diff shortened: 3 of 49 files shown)


lib/Doctrine/DBAL/Configuration.php
```diff
@@ -79,6 +80,12 @@ public function setResultCacheImpl(Cache $cacheImpl)
      */
     public function setFilterSchemaAssetsExpression($filterExpression)
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/3316',
+            'Configuration::setFilterSchemaAssetsExpression() is deprecated, use setSchemaAssetsFilter() instead.'
```
morozov, line 86: Would it make sense to use `('%s()', __METHOD__)` instead of the actual method name? This may simplify the process of composing such messages but on the other hand, will make the code less readable. I'm fine with either approach.

lib/Doctrine/DBAL/Driver/PDOException.php
```diff
@@ -43,6 +44,13 @@ public function __construct(\PDOException $exception)
      */
     public function getErrorCode()
     {
+        /** @psalm-suppress ImpureMethodCall */
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/4112',
+            'Driver\AbstractException::getErrorCode() is deprecated, use getSQLState() or getCode() instead.'
```
morozov, line 51: With the `__METHOD__` approach, this would be a fully qualified class name. Otherwise, it's inconsistent with the rest of the deprecation messages which mostly contain a class name without the namespace.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -1374,6 +1433,12 @@ public function project($sql, array $params, Closure $function)
      */
     public function query()
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/4163',
+            'Connection::query() is deprecated, use Connection::executeQuery() instead.'
```
morozov, line 1439: I think it's okay to omit the class name in `executeQuery()` since it's the same class that contains the deprecated method. It would be consistent with the other messages.

lib/Doctrine/DBAL/Driver/PDOStatement.php
```diff
@@ -274,10 +271,13 @@ private function convertParamType(int $type): int
     {
         if (! isset(self::PARAM_TYPE_MAP[$type])) {
             // TODO: next major: throw an exception
-            @trigger_error(sprintf(
-                'Using a PDO parameter type (%d given) is deprecated and will cause an error in Doctrine DBAL 3.0',
+            Deprecation::trigger(
+                'doctrine/dbal',
+                'https://github.com/doctrine/dbal/pull/3088',
+                'Using a PDO parameter type (%d given) is deprecated, ' .
+                'use \Doctrine\DBAL\Types\Types constants instead.',
```
morozov, line 278: One could use `Types::class` instead. From the standpoint of referencing classes, this is also inconsistent with the rest.

morozov (commented):
Looks good overall. Would it be possible to rebase the "Housekeeping: psalm" commit and address these issues by those commits that introduce them? Otherwise, we'll have a broken state in between.

lib/Doctrine/DBAL/Configuration.php
beberlei: Yes, my thought was that decomposing the messages for constant strings is a bit annoying and that the FQCN don't add value why i did it this way, not sure about it though. I think it would help if its readable for humans, especially since the alternative is written out anyways.

lib/Doctrine/DBAL/Driver/PDOStatement.php
beberlei: Yeah weird feeling with that one was that "Types" might have not been read as a class by users, where in the other cases with the ::method its more obvious.

morozov (dismissed):

lib/Doctrine/DBAL/Connection.php
```diff
@@ -2099,6 +2188,12 @@ public function createQueryBuilder()
      */
     public function ping()
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/4119',
+            'Retry and reconnnecting lost connections now happens automatically, ping() will be removed in DBAL 3.'
```
greg0ire:
```suggestion
            'Retry and reconnecting lost connections now happens automatically, ping() will be removed in DBAL 3.'
```

composer.json
```diff
@@ -35,6 +35,7 @@
         "php": "^7.3 || ^8",
         "ext-pdo": "*",
         "doctrine/cache": "^1.0",
+        "doctrine/deprecations": ">=0.3.0",
```
morozov: Why is the change from `^0.2` to this necessary? Is `doctrine/deprecations` expected to be backward-compatible forever?

morozov (changes requested):

composer.json
beberlei: My bad, i was expecting `^0.2.0` to also install `^0.3.0` but during  the `0.` branch the caret is actually more strict and `^0.2.0` means `>=0.2.0,<0.3.0` and not `<1.0.0`. I guess that means as long as deprecations doesn't break BC new features should get into mini versions for now, or we should release `1.0` very quickly.

beberlei:
I have updated https://github.com/doctrine/dbal/pull/4535/commits/5d20e3cec0643789ec728b2d15a288784b63fd24 for the `CompositeExpression` deprecation into one commit, because i needed to add a few flags that trigger or not depending on who is the caller. This is quite annoying and I am wondering about https://github.com/doctrine/deprecations/pull/4 in that context.

composer.json
morozov: With the major version being `0`, the minor part is promoted to major. Probably the same applies to the patch promoted to being the minor. I think it's okay to introduce features in patches before a stable version is released because ([ref](https://semver.org/spec/v0.1.0.html#spec-item-5)):
> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

morozov:
https://github.com/doctrine/deprecations/pull/4 doesn't look like a proper solution to the problem. It shouldn't be necessary for a library to suppress its own calls to its own deprecated APIs. In fact, not calling a certain method because it's deprecated may be a BC break on its own because somebody may have extended/implemented this method, and their extension is no longer effective. So not reporting deprecated calls within the library should be the default behavior.

morozov (dismissed):

lib/Doctrine/DBAL/Configuration.php
```diff
@@ -79,6 +80,12 @@ public function setResultCacheImpl(Cache $cacheImpl)
      */
     public function setFilterSchemaAssetsExpression($filterExpression)
     {
+        Deprecation::triggerIfCalledFromOutside(
```
morozov: This method doesn't seem to be used internally. Tests don't count, they are not part of the library. I didn't expect this many methods having to rely on `triggerIfCalledFromOutside()`.

beberlei: Ah, so if a method is not used internally anymore then we could use the unconditional trigger indeed.

i will need to go over the methods more carefully to decide which one to use.

beberlei: @morozov i went over each call again and converted quite a few back to using `trigger`.

lib/Doctrine/DBAL/Driver/IBMDB2/DB2Driver.php
```diff
@@ -35,6 +36,12 @@ public function connect(array $params, $username = null, $password = null, array
      */
     public function getName()
     {
+        Deprecation::triggerIfCalledFromOutside(
```
morozov: This method seems to be called only from tests, should be `trigger()` (the same is for the rest of the implementations).

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
```diff
@@ -135,6 +136,12 @@ public function getServerVersion()
      */
     public function requiresQueryForServerVersion()
     {
+        Deprecation::triggerIfCalledFromOutside(
```
morozov, line 139: This method seems to be called only from tests, should be `trigger()` (the same is for the rest of the implementations).

lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
```diff
@@ -71,6 +72,12 @@ public function or($expression, ...$expressions): CompositeExpression
      */
     public function andX($x = null)
     {
+        Deprecation::triggerIfCalledFromOutside(
```
morozov: This method seems to be called only from tests, should be `trigger()`.

lib/Doctrine/DBAL/Query/Expression/ExpressionBuilder.php
```diff
@@ -84,6 +91,12 @@ public function andX($x = null)
      */
     public function orX($x = null)
     {
+        Deprecation::triggerIfCalledFromOutside(
```
morozov: This method seems to be called only from tests, should be `trigger()`.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -483,6 +484,15 @@ public function select($select = null/*, string ...$selects*/)
             return $this;
         }
 
+        if (is_array($select)) {
+            Deprecation::triggerIfCalledFromOutside(
```
morozov: `QueryBuilder` isn't used by the DBAL, its API is only called from tests, should be `trigger()`.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -533,6 +543,15 @@ public function addSelect($select = null/*, string ...$selects*/)
...
```
morozov: `QueryBuilder` isn't used by the DBAL, its API is only called from tests, should be `trigger()`.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -911,6 +930,15 @@ public function groupBy($groupBy/*, string ...$groupBys*/)
             return $this;
         }
 
+        if (is_array($groupBy)) {
+            Deprecation::triggerIfCalledFromOutside(
```
morozov: `QueryBuilder` isn't used by the DBAL, its API is only called from tests, should be `trigger()`.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -940,6 +968,15 @@ public function addGroupBy($groupBy/*, string ...$groupBys*/)
...
```
morozov: `QueryBuilder` isn't used by the DBAL, its API is only called from tests, should be `trigger()`.

morozov (commented):
> i went over each call again and converted quite a few back to using `trigger`.

That's quite a few but I think we can do better. Please see the comments.

lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php
beberlei: `Connection::getServerVersion` calls it

beberlei:
Failure is fixed in deprecatios 0.5.3 but i cannot retrigger the appveyor build.

composer.json
```diff
@@ -35,6 +35,7 @@
         "php": "^7.3 || ^8",
         "ext-pdo": "*",
         "doctrine/cache": "^1.0",
+        "doctrine/deprecations": "^0.5.2",
```
morozov: Should this be bumped to `^0.5.3`, since otherwise, the build will fail?

morozov (commented):
> Failure is fixed in deprecatios 0.5.3 but i cannot retrigger the appveyor build.

It warrants a bump of the Deprecations version.

Looks good! Just needs the changes in the last few commits moved under the corresponding original commits.

morozov (dismissed):

lib/Doctrine/DBAL/Connection.php
```diff
@@ -269,6 +277,13 @@ public function getHost()
      */
     public function getPort()
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/issues/3580',
+            'Connection::getPort() is deprecated, get database host from application config ' .
```
greg0ire:
```suggestion
            'Connection::getPort() is deprecated, get the database server port from application config ' .
```

lib/Doctrine/DBAL/Connection.php
```diff
@@ -281,6 +296,13 @@ public function getPort()
      */
     public function getUsername()
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/issues/3580',
+            'Connection::getUsername() is deprecated, get database host from application config ' .
```
greg0ire:
```suggestion
            'Connection::getUsername() is deprecated, get the username from application config ' .
```

lib/Doctrine/DBAL/Connection.php
```diff
@@ -257,6 +258,13 @@ public function getDatabase()
      */
     public function getHost()
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/issues/3580',
+            'Connection::getHost() is deprecated, get database host from application config ' .
```
greg0ire:
```suggestion
            'Connection::getHost() is deprecated, get the database server host from application config ' .
```

lib/Doctrine/DBAL/Connection.php
```diff
@@ -293,6 +315,13 @@ public function getUsername()
      */
     public function getPassword()
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/issues/3580',
+            'Connection::getPassword() is deprecated, get database host from application config ' .
```
greg0ire:
```suggestion
            'Connection::getPassword() is deprecated, get the password from application config ' .
```

lib/Doctrine/DBAL/Logging/EchoSQLLogger.php
```diff
@@ -13,6 +15,15 @@
  */
 class EchoSQLLogger implements SQLLogger
 {
+    public function __construct()
+    {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/3935',
+            'EchoSQLLogger is deprecated without replacement, move code into your project if you rely on it.'
```
greg0ire:
```suggestion
            'EchoSQLLogger is deprecated without replacement, move the code into your project if you rely on it.'
```

morozov (approved):

greg0ire (approved):

PR #4543, opened 2021-03-11
Remove incomplete details of QueryBuilder::setParameter($type)

andrew-demb:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

The argument description doesn't mention a more flexible way to configure binding parameter type - use a constant from `Doctrine\DBAL\Types\Types`.

~~PR proposes to mention this way of type configuration~~
UPD: instead of mention all ways to configure parameter type, the description updated to the short form, which is consistent with a codebase

### DIFF
--- a/lib/Doctrine/DBAL/Query/QueryBuilder.php
+++ b/lib/Doctrine/DBAL/Query/QueryBuilder.php
@@ -269,7 +269,7 @@ class QueryBuilder
      *
      * @param int|string           $key   Parameter position or name
      * @param mixed                $value Parameter value
-     * @param int|string|Type|null $type  One of the {@link ParameterType} constants or DBAL type
+     * @param int|string|Type|null $type  Parameter type
      *
      * @return $this This QueryBuilder instance.
      */


src/Query/QueryBuilder.php
```diff
@@ -371,7 +372,7 @@ public function getSQL()
      *
      * @param int|string           $key   Parameter position or name
      * @param mixed                $value Parameter value
-     * @param int|string|Type|null $type  One of the {@link ParameterType} constants or DBAL type
+     * @param int|string|Type|null $type  One of the {@link ParameterType} constants, {@link Types} constants or DBAL type instance
```
morozov: There are dozens of places in the codebase where this union type is used. Is it the only place where this change is needed?

andrew-demb: In the other places `ParameterType` class is not mentioned :)

We can use the same full description in all places, if we need to suggest it to user always.

Should we?

morozov: I'd rather remove the details from here.

morozov:
@andrew-demb please squash and update the commit message (see the updated PR title). Also, please retarget against the oldest supported version where this is applicable.

andrew-demb:
@morozov Done. Thank you for your suggestions.

morozov (approved):

morozov:
Thanks, @andrew-demb.

PR #4483, opened 2021-01-21
Improve heredoc indentation

mvorisek:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | only CS update

#### Summary

Update CS, split of https://github.com/doctrine/dbal/pull/4360

PR #4548, opened 2021-03-16
Deprecations in schema- and namespace-related APIs

### LINKED ISSUE #4503: Inconsistent SchemaManager API across platforms

```php
```

morozov:
|      Q       |   A
|------------- | -----------
| BC Break     | no

See #4503.

1. `PostgreSQLSchemaManager::getExistingSchemaSearchPaths()` and `::determineExistingSchemaSearchPaths()` have been marked internal and will be made protected in `4.0.0`. They are only used by the schema manager itself and are not part of the API.
2. `AbstractSchemaManager::dropSchema()` has been added to implement the test properly.
3. `AbstractSchemaManager::listSchemaNames()` has been added as a replacement for `AbstractSchemaManager::listNamespaceNames()` and `PostgreSQLSchemaManager::getSchemaNames()`.

**Reasoning behind the naming changes**:

The "namespace" term is [internal](https://www.postgresql.org/docs/13/catalog-pg-namespace.html) to PostgreSQL and isn't part of the ANSI SQL vocabulary:

> A namespace is the structure underlying SQL schemas

Schemas are supported by most if not all platforms supported by the DBAL.

**Reasoning behind the design changes**:

The new `AbstractSchemaManager::listSchemaNames()` method have been implemented directly in the schema manager instead of using the `getListSchemataSQL()` → `getPortableSchemaList()` → `getPortableSchemaDefinition()` approach. This approach has the following flaws:

1. Although `getListSchemataSQL(): string` would be part of the abstract API, it could be used only within a given platform or even its version, so it's not really part of the abstraction. Also, such methods are untestable unless tested as part of the schema manager.
2. `getPortableSchemaDefinition()` would expect a `list<string,mixed>` and return a `list<string>` but again, it couldn't accept a row set from another platform. Furthermore, neither the API nor default the default implementation of such methods make any sense: https://github.com/doctrine/dbal/blob/f18aee0a2b209662370de14f3a5268553c9b239e/src/Schema/AbstractSchemaManager.php#L710-L718

`AbstractSchemaManager::dropSchema()` has been implemented using the implementation from `AbstractPlatform` in order to be able to reuse it go generate migrations.

### DIFF
--- a/src/Schema/PostgreSQLSchemaManager.php
+++ b/src/Schema/PostgreSQLSchemaManager.php
@@ -73,6 +73,8 @@ class PostgreSQLSchemaManager extends AbstractSchemaManager
      *
      * This is a PostgreSQL only function.
      *
+     * @internal The method should be only used from within the PostgreSQLSchemaManager class hierarchy.
+     *
      * @return string[]
      */
     public function getExistingSchemaSearchPaths()
@@ -88,6 +90,8 @@ class PostgreSQLSchemaManager extends AbstractSchemaManager
      * Sets or resets the order of the existing schemas in the current search path of the user.
      *
      * This is a PostgreSQL only function.
+     *
+     * @internal The method should be only used from within the PostgreSQLSchemaManager class hierarchy.
      *
      * @return void
      */
--- a/src/Schema/PostgreSQLSchemaManager.php
+++ b/src/Schema/PostgreSQLSchemaManager.php
@@ -38,14 +38,29 @@ class PostgreSQLSchemaManager extends AbstractSchemaManager
     /**
      * Gets all the existing schema names.
      *
+     * @deprecated Use {@link listSchemaNames()} instead.
+     *
      * @return string[]
      *
      * @throws Exception
      */
     public function getSchemaNames()
     {
+        return $this->listNamespaceNames();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function listSchemaNames(): array
+    {
         return $this->_conn->fetchFirstColumn(
-            "SELECT nspname FROM pg_namespace WHERE nspname !~ '^pg_.*' AND nspname != 'information_schema'"
+            <<<'SQL'
+SELECT schema_name
+FROM   information_schema.schemata
+WHERE  schema_name NOT LIKE 'pg\_%'
+AND    schema_name != 'information_schema'
+SQL
         );
     }
 
@@ -97,7 +112,7 @@ class PostgreSQLSchemaManager extends AbstractSchemaManager
      */
     public function determineExistingSchemaSearchPaths()
     {
-        $names = $this->getSchemaNames();
+        $names = $this->listSchemaNames();
         $paths = $this->getSchemaSearchPaths();
 
         $this->existingSchemaPaths = array_filter($paths, static function ($v) use ($names): bool {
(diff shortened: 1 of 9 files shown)


src/Schema/PostgreSQLSchemaManager.php
```diff
@@ -38,14 +38,29 @@ class PostgreSQLSchemaManager extends AbstractSchemaManager
...
+    public function listSchemaNames(): array
     {
         return $this->_conn->fetchFirstColumn(
-            "SELECT nspname FROM pg_namespace WHERE nspname !~ '^pg_.*' AND nspname != 'information_schema'"
+            <<<'SQL'
+SELECT schema_name
+FROM   information_schema.schemata
+WHERE  schema_name NOT LIKE 'pg\_%'
+AND    schema_name != 'information_schema'
+SQL
```
greg0ire, line 63: PHP 7.3 allows you to use indention for nowdoc here and below :slightly_smiling_face:

tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -159,26 +158,40 @@ public function testListDatabases(): void
...
 
-            $namespaces = $this->schemaManager->listNamespaceNames();
-            $namespaces = array_map('strtolower', $namespaces);
-        }
+        self::assertContains('test_create_schema', $method($this->schemaManager));
+    }
 
-        self::assertContains('test_create_schema', $namespaces);
+    /**
+     * @return iterable<list<mixed>>
```
greg0ire, line 180: dataproviders return signature should match the `return iterable<array{…}>`, where the `…` part is build from the phpdoc of the method that use it, so here `return iterable<array{callable(AbstractSchemaManager): list<string>}>`?

src/Schema/PostgreSQLSchemaManager.php
morozov: I have mixed feelings about this feature:
1. The left margin of the block is not clearly visible.
2. It's usually a program within a program, so it's fine that it's not indented the same as the outer one.

So far, it's consistent with the rest of the codebase (except for https://github.com/doctrine/dbal/pull/4483). Let's reconsider this at the coding standard level first.

tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
morozov: > ERROR: InvalidReturnType - tests/Functional/Schema/SchemaManagerFunctionalTestCase.php:180:16 - The declared return type 'iterable<mixed, array{callable(Doctrine\DBAL\Schema\AbstractSchemaManager):list<string>}>' for Doctrine\DBAL\Tests\Functional\Schema\SchemaManagerFunctionalTestCase::listSchemaNamesMethodProvider is incorrect, got 'Generator<int, array{impure-Closure(Doctrine\DBAL\Schema\AbstractSchemaManager):array<array-key, string>|impure-Closure(Doctrine\DBAL\Schema\AbstractSchemaManager):list<string>}, mixed, void>' (see https://psalm.dev/011)
     * @return iterable<array{callable(AbstractSchemaManager): list<string>}>

I agree in general but it would be much better if it was validated by the static analysis. Otherwise, it helps only when implementing the data provider but isn't used by the caller.

Similarly to the above comment, let's get it sorted out at the project level and apply consistently. Psalm at level 1 should be able to report (and maybe fix) some of such issues.

greg0ire (dismissed):

greg0ire (approved):

PR #4550, opened 2021-03-17
Update test DB config: rename privileged db and require explicit driver

acoulton:
|      Q       |   A
|------------- | -----------
| Type         | improvement (to unit testing)
| BC Break     | not in the public API - method renamed in internal test helper and may require changes to contributors' local phpunit.xml config.
| Fixed issues | N/A

#### Summary

As discussed with @greg0ire this brings the naming / behaviour of the TestUtil class and the database config parameters up to date with the changes in doctrine/orm#8532 for consistency between projects. See commit message for full details.

### DIFF
--- a/ci/github/phpunit/sqlite.xml
+++ b/ci/github/phpunit/sqlite.xml
@@ -7,6 +7,12 @@
          failOnRisky="true"
          failOnWarning="true"
 >
+    <php>
+        <!-- use an in-memory sqlite database -->
+        <var name="db_driver" value="pdo_sqlite"/>
+        <var name="db_memory" value="true"/>
+    </php>
+
     <testsuites>
         <testsuite name="Doctrine DBAL Test Suite">
             <directory>../../../tests</directory>
--- a/ci/github/phpunit/sqlite.xml
+++ b/ci/github/phpunit/sqlite.xml
@@ -8,7 +8,6 @@
          failOnWarning="true"
 >
     <php>
-        <!-- use an in-memory sqlite database -->
         <var name="db_driver" value="pdo_sqlite"/>
         <var name="db_memory" value="true"/>
     </php>
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -22,24 +22,37 @@
     <php>
         <ini name="error_reporting" value="-1" />
 
-        <!-- Test connection parameters -->
-        <!-- Uncomment, otherwise SQLite runs
+        <!-- "Real" test database -->
+        <var name="db_driver" value="pdo_sqlite"/>
+        <var name="db_memory" value="true"/>
+        <!-- to use another database driver / credentials, provide them like so:
         <var name="db_driver" value="pdo_mysql"/>
         <var name="db_host" value="localhost" />
-        <var name="db_port" value="3306"/>
         <var name="db_user" value="root" />
         <var name="db_password" value="" />
         <var name="db_dbname" value="doctrine_tests" />
-        -->
+        <var name="db_port" value="3306"/>-->
         <!--<var name="db_event_subscribers" value="Doctrine\DBAL\Event\Listeners\OracleSessionInit">-->
 
-        <!-- Privileged user connection parameters. Used to create and drop the test database -->
-        <var name="tmpdb_driver" value="pdo_mysql"/>
-        <var name="tmpdb_host" value="localhost" />
-        <var name="tmpdb_port" value="3306"/>
-        <var name="tmpdb_user" value="root" />
-        <var name="tmpdb_password" value="" />
-        <var name="tmpdb_dbname" value="doctrine_tests_tmp" />
+        <!--
+         At the start of each test run, we will drop and recreate the test database.
+
+         By default we assume that the `db_` config above has unrestricted access to the provided database
+         platform.
+
+         If you prefer, you can provide a restricted user above and a separate `privileged_db` config
+         block to provide details of a privileged connection to use for the setup / teardown actions.
+
+         Note that these configurations are not merged - if you specify a `privileged_db_driver` then
+         you must also specify all the other options that your driver requires.
+
+        <var name="privileged_db_driver" value="pdo_mysql"/>
+        <var name="privileged_db_host" value="localhost" />
+        <var name="privileged_db_user" value="root" />
+        <var name="privileged_db_password" value="" />
+        <var name="privileged_db_dbname" value="doctrine_tests_tmp" />
+        <var name="privileged_db_port" value="3306"/>
+        -->
     </php>
 
     <testsuites>
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -28,10 +28,11 @@
         <!-- to use another database driver / credentials, provide them like so:
         <var name="db_driver" value="pdo_mysql"/>
         <var name="db_host" value="localhost" />
+        <var name="db_port" value="3306"/>
         <var name="db_user" value="root" />
         <var name="db_password" value="" />
         <var name="db_dbname" value="doctrine_tests" />
-        <var name="db_port" value="3306"/>-->
+        -->
         <!--<var name="db_event_subscribers" value="Doctrine\DBAL\Event\Listeners\OracleSessionInit">-->
 
         <!--
--- a/tests/Doctrine/Tests/TestUtil.php
+++ b/tests/Doctrine/Tests/TestUtil.php
@@ -141,7 +141,7 @@ class TestUtil
     {
         if (! isset($GLOBALS['db_driver'])) {
             throw new UnexpectedValueException(
-                'You must provide database connection params including a db_driver value. See phpunit.xml.dist for details'
+                'You must provide db connection params including a db_driver value. See phpunit.xml.dist for details'
             );
         }
 
--- a/tests/Doctrine/Tests/TestUtil.php
+++ b/tests/Doctrine/Tests/TestUtil.php
@@ -27,46 +27,6 @@ use const STDOUT;
 
 /**
  * TestUtil is a class with static utility methods used during tests.
- *
- * @psalm-type OverrideParams = array{
- *     charset?: string,
- *     dbname?: string,
- *     default_dbname?: string,
- *     driver?: key-of<\Doctrine\DBAL\DriverManager::DRIVER_MAP>,
- *     driverClass?: class-string<\Doctrine\DBAL\Driver>,
- *     driverOptions?: array<mixed>,
- *     host?: string,
- *     password?: string,
- *     path?: string,
- *     pdo?: \PDO,
- *     platform?: \Doctrine\DBAL\Platforms\AbstractPlatform,
- *     port?: int,
- *     user?: string,
- * }
- * @psalm-type ConnectionParams = array{
- *     charset?: string,
- *     dbname?: string,
- *     default_dbname?: string,
- *     driver?: key-of<\Doctrine\DBAL\DriverManager::DRIVER_MAP>,
- *     driverClass?: class-string<\Doctrine\DBAL\Driver>,
- *     driverOptions?: array<mixed>,
- *     host?: string,
- *     keepSlave?: bool,
- *     keepReplica?: bool,
- *     master?: OverrideParams,
- *     memory?: bool,
- *     password?: string,
- *     path?: string,
- *     pdo?: \PDO,
- *     platform?: \Doctrine\DBAL\Platforms\AbstractPlatform,
- *     port?: int,
- *     primary?: OverrideParams,
- *     replica?: array<OverrideParams>,
- *     sharding?: array<string,mixed>,
- *     slaves?: array<OverrideParams>,
- *     user?: string,
- *     wrapperClass?: class-string<Connection>,
- * }
  */
 class TestUtil
 {
@@ -97,6 +57,7 @@ class TestUtil
             self::$initialized = true;
         }
 
+        /** @psalm-suppress ArgumentTypeCoercion */
         $conn = DriverManager::getConnection(self::getTestConnectionParameters());
 
         self::addDbEventSubscribers($conn);
@@ -106,6 +67,7 @@ class TestUtil
 
     public static function getPrivilegedConnection(): Connection
     {
+        /** @psalm-suppress ArgumentTypeCoercion */
         return DriverManager::getConnection(self::getPrivilegedConnectionParameters());
     }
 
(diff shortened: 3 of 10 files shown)


acoulton:
@greg0ire sorry - I can't work out how to get psalm to accept the arrays returned from the connection parameter methods, as it doesn't consider the generic `array<string, mixed>` valid for the detailed list of keys and values that are documented in the typehint for `DriverManager::getConnection()`. 

I've had a very quick look at the psalm docs to see how to handle it but that's not shed any light. I'm also not sure why this isn't reporting the same issue with the return / param types on the version I added to ORM - psalm was quite happy there.

Any tips?

greg0ire:
> I've had a very quick look at the psalm docs to see how to handle it but that's not shed any light. I'm also not sure why this isn't reporting the same issue with the return / param types on the version I added to ORM - psalm was quite happy there.

Psalm is still at a very high level (high is lenient for Psalm) on ORM, that's why. To fix your issue, if you know that the parameters are supposed to match the parameters of the method you call, you can copy/paste

https://github.com/doctrine/dbal/blob/4fa884c65043e530a20d5f9ac868400e7f3bc2ab/src/DriverManager.php#L45-L68 or what you know to be the correct subset of it as a `psalm-return` phpdoc for `getTestConnectionParameters`

acoulton:
Ah, that makes sense about the difference to ORM. 

Copy-pasting that docblock feels like it introduces a reasonable bit of duplication though, and is potentially misleading about what those methods return since they don't actually enforce any type-checking of the values in the array?

For example `memory` for the sqlite driver happens to be bool because that's what it's set to in the phpunit.xml config, but it could just as easily have been set to a string. Similarly `port` is probably an int, but that's not guaranteed, and we don't do anything to validate the `driver` value is valid or that the subset of values we have makes sense for the requested driver...

Should TestUtil actually be validating these properties itself? I feel like the answer is probably no, that's the responsibility of the DriverManager and the individual drivers? Otherwise there's quite a bit of duplication and coupling in these generic data-mapping methods. I suppose in a totally clean architecture the validation would be in a separate shared layer which could take `array<string, mixed>` and guarantee to return a strongly typed array / object - but that's not how the interfaces are structured at the moment.

If I've understood right, before psalm can be made stricter on orm these typehints would also have to be carried back to e.g. `EntityManager::create` since that ultimately passes the array on to `DriverManager::getConnection`. And then in turn third-party projects would need to typehint their own config loader / dependency factories if they wanted to use psalm themselves? I think it's fairly common - and arguably correct in the current architecture - to just take an arbitrary array of config and assume the DriverManager will complain at runtime if it's not valid.

I'm fairly new to psalm, so I may well have misunderstood, but it sounds to me like actually this is highlighting that the typing on DriverManager::getConnection should be relaxed rather than these methods made stricter?

greg0ire:
That duplication is indeed a code smell, and I don't think there is any validation anywhere. Not sure how to solve this, but maybe having a value object that could be created as early as possible would help here. We would have a method to build it from `array<string,mixed>` that would throw if anything looks wrong and would then be able to trust that object. Right now, I'm not sure we can actually relax anything in `DriverManager::getConnection`: the code just assumes that parameters are of the documented type. We could relax types if we performed checks.

phpunit.xml.dist
```diff
@@ -22,24 +22,37 @@
...
+        <var name="db_memory" value="true"/>
+        <!-- to use another database driver / credentials, provide them like so:
         <var name="db_driver" value="pdo_mysql"/>
         <var name="db_host" value="localhost" />
-        <var name="db_port" value="3306"/>
         <var name="db_user" value="root" />
         <var name="db_password" value="" />
         <var name="db_dbname" value="doctrine_tests" />
-        -->
+        <var name="db_port" value="3306"/>-->
```
morozov: Why is this change needed?

ci/github/phpunit/sqlite.xml
```diff
@@ -7,6 +7,12 @@
          failOnRisky="true"
          failOnWarning="true"
 >
+    <php>
+        <!-- use an in-memory sqlite database -->
```
morozov: The two following lines are pretty much self-explanatory. The comment looks redundant.

tests/Doctrine/Tests/TestUtil.php
```diff
@@ -85,6 +76,10 @@ private static function initializeDatabase(): void
 
         $testConn = DriverManager::getConnection($testConnParams);
 
+        // Note, writes direct to STDERR to prevent phpunit detecting output - otherwise this would cause either an
+        // "unexpected output" warning or a failure on the first test case to call this method.
+        fwrite(STDERR, sprintf("\nUsing DB driver %s\n", get_class($testConn->getDriver())));
```
morozov: The driver is a required part of the test configuration. Why does it need to be logged?

tests/Doctrine/Tests/TestUtil.php
```diff
@@ -213,11 +193,6 @@ private static function mapConnectionParameters(array $configuration, string $pr
         return $parameters;
     }
 
-    public static function getPrivilegedConnection(): Connection
```
morozov, line 216: Why is this change needed?

phpunit.xml.dist
```diff
@@ -22,24 +22,37 @@
...
 
-        <!-- Privileged user connection parameters. Used to create and drop the test database -->
-        <var name="tmpdb_driver" value="pdo_mysql"/>
-        <var name="tmpdb_host" value="localhost" />
-        <var name="tmpdb_port" value="3306"/>
-        <var name="tmpdb_user" value="root" />
-        <var name="tmpdb_password" value="" />
-        <var name="tmpdb_dbname" value="doctrine_tests_tmp" />
+        <!--
+         At the start of each test run, we will drop and recreate the test database.
```
morozov, line 39: This comment and "if you prefer" seem to contradict each other. The test suite needs a privileged connection to create the test database. So if a user prefers to not provide the privileged connection configuration, the test suite won't work.

This is why all platforms have both configurations.

ci/github/phpunit/sqlite.xml
acoulton: fair point, I'll remove

phpunit.xml.dist
acoulton: For consistency with the equivalent file content in doctrine/orm. Obviously makes no functional difference but I think it makes sense that any diff between the projects is intentional, to make it easier to spot / apply changes in future.

phpunit.xml.dist
acoulton: Apologies but that is not correct. As discussed in doctrine/orm#8532 neither dbal nor orm actually *use* a separate `privileged_db` connection in any of the out-of-the-box configurations (except Oracle in dbal).

If `privileged_db` values are not specified, the TestUtil class falls back to using the `db_xx` values, which are assumed to have root-like access. Note that this has not changed in this pull, only the naming of the `tmpdb_` values and the documentation to clarify that these are entirely optional and only required if the user wants/needs to provide a different set of credentials to create/drop the DB than the ones that will be used for the rest of the tests.

tests/Doctrine/Tests/TestUtil.php
acoulton: So that in CI (for example if moving to a new platform / changing configuration structure) there is some way to verify that a given job is running tests with the expected driver. Although the risk of this is much lower now that the fallback option has been removed, it may still be possible (e.g. if a matrix variable somehow selects the wrong phpunit config file).

Given the potential consequences of e.g. only running the sqlite memory tests in CI for a prolonged period, it feels worth the extra line of console output to provide a bit more safety / visibility that the configuration has been applied as expected.

tests/Doctrine/Tests/TestUtil.php
acoulton: Again just to bring this file in sync with the doctrine/orm equivalent - the method has not been removed, just moved up the file so that the diff between the two projects only contains expected / intentional differences.

phpunit.xml.dist
morozov: Currently, the parameters are specified in the logical order: driver (how should I connect), host (to which machine do I connect), port (to which port on this machine should I connect). Please revert.

tests/Doctrine/Tests/TestUtil.php
morozov: It makes the review harder. There are components in existing coding standards that enforce the or of members in the class. Before it's standardized, there's no guarantee that it will remain like this, so let's not make this change now.

tests/Doctrine/Tests/TestUtil.php
morozov: This way, it would make sense to log all the connection details, not just the driver. Otherwise, if some of the parameters are misspelled, they will be silently ignored.

phpunit.xml.dist
morozov: > neither dbal nor orm actually use a separate privileged_db connection in any of the out-of-the-box configurations (except Oracle in dbal).

Are you talking about the logic introduced in https://github.com/doctrine/dbal/pull/4067? 
> All so-called temporary connections look identical to the primary ones with the exception that the temporary one doesn't usually have a database name specified (creating a database is the primary purpose of the temporary connection). Instead of requiring the parameters to be specified explicitly, in the absence of the parameters, the test suite can fall back to the primary connection parameters **and ignore the database**.

So these are two different connections built from the same set of parameters. But it's not possible for every driver, so it's not a matter of the user preference. The proposed wording is misleading.

phpunit.xml.dist
acoulton: Fair enough - though the non-logical order is not mine, it's the one that's been in the config file for orm for years. Reverted here, I can do a PR to bring orm up to date once you're happy with the overall diff on this.

ci/github/phpunit/sqlite.xml
acoulton: done

tests/Doctrine/Tests/TestUtil.php
acoulton: I'm not convinced that logging all the details brings any benefit.

* If the ignored parameters were required to make the connection at all (e.g. it hasn't been given / has a default db host it can't connect to), the driver will throw
* If they were required post-connection to run a test (e.g. an explicit dbname) that test will fail
* If they weren't required for anything, their absence is not relevant

I suppose if it was common to run in an environment with multiple different servers all running at once (e.g. you have mysql 5.7 on one port and mysql 8 on another, and two different phpunit.xml files to connect to the appropriate version) there'd be a case for knowing you were connected to the right port. But that would be an unusual setup, and a long way from how CI is configured at the moment.

So the only silent failure case I can really see is if you're using a completely different driver to the expected one - which is what was happening in the ORM CI.

IMHO therefore there's no benefit to the extra verbosity of logging all the connection details, and it makes it harder to spot the one that matters.

tests/Doctrine/Tests/TestUtil.php
acoulton: OK but presumably when those are standardized they'll be applied across both ORM and DBAL so will produce the same change in both. And will result in a massive diff like the recent coding standards ones anyway, so this method moving will be fairly irrelevant in the bigger picture.

For me as an outsider, trying to debug why the ORM CI was not working as expected when DBAL was, the huge diff between the TestUtil classes in each project meant it took me a lot longer to spot the problem. And I suspect contributed to the issue arising in the first place since it wouldn't have been easy to see that the two projects couldn't just use the same phpunit / CI config files.

Hence I think it makes a lot more sense to minimise the diff between projects even if it means a few extra LOC changed in this PR.

But if you'd like me to move it back I can.

phpunit.xml.dist
acoulton: No offence intended, I have no idea of the history of the logic - though from a skim of the PR you mention, yes.

The wording I've proposed is my best effort to understand the meaning and function of parameters that were IMO confusingly named and not documented anywhere at all. It appears from doctrine/orm#8532 that there are core team members that were also unclear on why/how these work, and that this was probably the root cause of ORM not running CI against any real databases for months through a period of massive automated changes to the codebase.

I can see that the Oracle driver uses two different user accounts (with every other aspect of the `privileged_` and `db_` params identical). I have no idea why it can't just use `system` for both connections the same way every other driver just uses `root` or equivalent.

Obviously if you can explain it better I'd welcome alternative wording, but I'm not sure I'll be able to figure that out on my own.

acoulton:
@greg0ire getting psalm to be happy with these connection parameters is proving very challenging. It's not as simple as copying the `@psalm-type` declaration from the DriverManager and referencing that because:

* That declaration references imported classes / classes in the same namespace that resolve differently in TestUtil, so e.g. `driver?: key-of<self::DRIVER_MAP>,` has to be changed to `driver?: key-of<Doctrine\DBAL\DriverManager::DRIVER_MAP>` and similarly for a bunch of other classes
* Once you get the declaration itself valid, psalm is happy with the ::getConnection calls but then generates two errors relating to the the inferred return type of `::mapConnectionParameters` not matching the declared type. See https://github.com/doctrine/dbal/runs/2288582601
* At that point I started trying to run it locally, it turns out it's not enough to just subset the list because a) mapConnectionParams copies all the values to driverOptions but leaves them in the outer array, rather than knowing where they should all go, and b) it obviously doesn't validate the driver is one of the expected values etc.

Short of adding a lot of runtime validation and knowledge to the `::mapConnectionParams` method - which I think would be a very bad coupling/complexity - I can't find a way round this. 

I think the only sensible approach is to just suppress the psalm warning on the args that are passed to DriverManager::getConnection, knowing that it will internally validate / ignore things as appropriate at runtime. Unless you can find a better solution?

greg0ire:
I just discovered `@psalm-import-type` in the docs, maybe that can be of some help: https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-import-type ?

Update: I think we still might end up with suppress rules, but those would be earlier in the call tree, and we would be able to guarantee that e.g. `mapConnectionParameters` return an array in the desired shape, which would be nice. It might mean fewer suppressing annotations, I'm not sure yet.

greg0ire:
I pushed a new commit, please let me know what you think, and feel free to drop it if you don't like it.

acoulton:
@greg0ire ah that's much cleaner, cheers. I can squash all these up once the other outstanding comments are resolved.

acoulton:
@morozov did you close this intentionally? It looks like potentially github has mass-closed a large number of PRs when you deleted the 2.12.x branch.

If you and @greg0ire still want this I can rebase and open a new PR for 2.13.x - just let me know.

morozov:
@acoulton the PR got closed because I removed the branch. Please retarget against `3.1.x`. Sorry for the inconvenience.

Personally, I don't see where this change is going so I'm not sure it's worth spending the time.

acoulton:
@morozov OK fair enough, I only put the time into it because @greg0ire in doctrine/orm#8532 asked me to. I was only really concerned about minimising the risk of a config error causing the CI process to give false confidence again in future.

But I'm pretty busy at the moment so I'm not overly fussed about pushing it through.

I do still think that for new contributors (and evidently some core members) there's a significant lack of clarity about the purpose / function of the `tmpdb_` params and it would be worth someone who understands them documenting them at some point, to avoid confusion.

Issue #2663, opened 2017-02-24
Comparator fails for BLOBS


thomashohn:
In the TYPO3 project we use the diffColumn method in the class Doctrine\DBAL\Schema. Unfortunately this method fails to detect changes in BLOB columns. So changing a column from BLOB to another BLOB type fails.

Ocramius:
@thomashohn this needs a test case, as the issue as reported is too superficial.

thomashohn:
Create the following table:
create table test (
 ablob blob
)

Try to change the column to ablob medium blob - the diffColumn doesn't catch this case.

chrisguitarguy:
Seems like this happens because the the [`MySqlPlatform` uses `length` to determine blob size](https://github.com/doctrine/dbal/blob/03e17189c48a92bff9c5fd7d9242d13a15780d15/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L1104-L1123), but `length` [is only diffed for a few types](https://github.com/doctrine/dbal/blob/03e17189c48a92bff9c5fd7d9242d13a15780d15/lib/Doctrine/DBAL/Schema/Comparator.php#L447-L459).

Defcon0:
Is there any progress here? We‘re also experiencing this issue in our contao cms instance :-(

ausi:
> this needs a test case, as the issue as reported is too superficial.

I created pull request #4551 to fix this issue, it includes several test cases.

morozov:
This bug will be indirectly covered by the `Doctrine\DBAL\Tests\Functional\Schema\MySQL::testLobLengthIncrementOverLimit` test in https://github.com/doctrine/dbal/pull/4746.

Additionally,
```php
$connection->executeStatement('DROP TABLE IF EXISTS test');
$connection->executeStatement('CREATE TABLE test (ablob BLOB)');

$sm = $connection->createSchemaManager();
$desiredTable = $sm->listTableDetails('test');

$connection->executeStatement('ALTER TABLE test MODIFY COLUMN ablob MEDIUMBLOB');

$actualTable = $sm->listTableDetails('test');

$comparator = $sm->createComparator();
$diff = $comparator->diffTable($desiredTable, $actualTable);
var_dump($diff);
/*
class Doctrine\DBAL\Schema\TableDiff#49 (14) {
    ...
}
*/
```

PR #4551, opened 2021-03-23
Fix Comparator column diff for blob length

ausi:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #2663

#### Summary

In MySQL the `blob` type depends on the `length` to use one of `TINYBLOB`, `BLOB`, `MEDIUMBLOB` or `LONGBLOB`. When diffing two `blob` columns the length is currently ignored which results in missing changes of the underlying database column type.

UPDATE: It is the same with the `text` type which uses one of `TINYTEXT`, `TEXT`, `MEDIUMTEXT` or `LONGTEXT`.

### DIFF
--- a/src/Schema/Comparator.php
+++ b/src/Schema/Comparator.php
@@ -13,6 +13,7 @@ use function array_unique;
 use function assert;
 use function count;
 use function get_class;
+use function max;
 use function strtolower;
 
 /**
@@ -480,9 +481,9 @@ class Comparator
                 $changedProperties[] = 'fixed';
             }
         } elseif ($properties1['type'] instanceof Types\BlobType || $properties1['type'] instanceof Types\TextType) {
-            // check if value of length is set at all, default value assumed otherwise.
-            $length1 = $properties1['length'] ?: 0;
-            $length2 = $properties2['length'] ?: 0;
+            // check if value of length is set at all, -1 gets normalized to 0.
+            $length1 = max(0, $properties1['length'] ?: 0);
+            $length2 = max(0, $properties2['length'] ?: 0);
             if ($length1 !== $length2) {
                 $changedProperties[] = 'length';
             }
--- a/src/Schema/Comparator.php
+++ b/src/Schema/Comparator.php
@@ -480,7 +480,10 @@ class Comparator
                 $changedProperties[] = 'fixed';
             }
         } elseif ($properties1['type'] instanceof Types\BlobType || $properties1['type'] instanceof Types\TextType) {
-            if ($properties1['length'] !== $properties2['length']) {
+            // check if value of length is set at all, default value assumed otherwise.
+            $length1 = $properties1['length'] ?: 0;
+            $length2 = $properties2['length'] ?: 0;
+            if ($length1 !== $length2) {
                 $changedProperties[] = 'length';
             }
         } elseif ($properties1['type'] instanceof Types\DecimalType) {
--- a/src/Schema/Comparator.php
+++ b/src/Schema/Comparator.php
@@ -482,8 +482,8 @@ class Comparator
             }
         } elseif ($properties1['type'] instanceof Types\BlobType || $properties1['type'] instanceof Types\TextType) {
             // check if value of length is set at all, -1 gets normalized to 0.
-            $length1 = max(0, $properties1['length'] ?: 0);
-            $length2 = max(0, $properties2['length'] ?: 0);
+            $length1 = max(0, $properties1['length'] ?? 0);
+            $length2 = max(0, $properties2['length'] ?? 0);
             if ($length1 !== $length2) {
                 $changedProperties[] = 'length';
             }
--- a/src/Schema/Comparator.php
+++ b/src/Schema/Comparator.php
@@ -479,7 +479,7 @@ class Comparator
             if ($properties1['fixed'] !== $properties2['fixed']) {
                 $changedProperties[] = 'fixed';
             }
-        } elseif ($properties1['type'] instanceof Types\BlobType) {
+        } elseif ($properties1['type'] instanceof Types\BlobType || $properties1['type'] instanceof Types\TextType) {
             if ($properties1['length'] !== $properties2['length']) {
                 $changedProperties[] = 'length';
             }
--- a/src/Schema/Comparator.php
+++ b/src/Schema/Comparator.php
@@ -479,6 +479,10 @@ class Comparator
             if ($properties1['fixed'] !== $properties2['fixed']) {
                 $changedProperties[] = 'fixed';
             }
+        } elseif ($properties1['type'] instanceof Types\BlobType) {
+            if ($properties1['length'] !== $properties2['length']) {
+                $changedProperties[] = 'length';
+            }
         } elseif ($properties1['type'] instanceof Types\DecimalType) {
             if (($properties1['precision'] ?? 10) !== ($properties2['precision'] ?? 10)) {
                 $changedProperties[] = 'precision';
--- a/src/Schema/DB2SchemaManager.php
+++ b/src/Schema/DB2SchemaManager.php
@@ -78,10 +78,6 @@ class DB2SchemaManager extends AbstractSchemaManager
                 $fixed  = true;
                 break;
 
-            case 'clob':
-                $length = $tableColumn['length'];
-                break;
-
             case 'decimal':
             case 'double':
             case 'real':
--- a/tests/Platforms/AbstractMySQLPlatformTestCase.php
+++ b/tests/Platforms/AbstractMySQLPlatformTestCase.php
@@ -808,6 +808,57 @@ abstract class AbstractMySQLPlatformTestCase extends AbstractPlatformTestCase
         ];
     }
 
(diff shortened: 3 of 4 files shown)


greg0ire (dismissed):
The patch looks good, and the tests fail as they should when reverting changes in `lib`:

```
There were 4 failures:

1) Doctrine\Tests\DBAL\Platforms\MariaDb1027PlatformTest::testAlterTableChangeBlobColumnLength
Failed asserting that false is not false.

/home/gregoire/dev/doctrine-dbal/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php:840

2) Doctrine\Tests\DBAL\Platforms\MySQL57PlatformTest::testAlterTableChangeBlobColumnLength
Failed asserting that false is not false.

/home/gregoire/dev/doctrine-dbal/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php:840

3) Doctrine\Tests\DBAL\Platforms\MySqlPlatformTest::testAlterTableChangeBlobColumnLength
Failed asserting that false is not false.

/home/gregoire/dev/doctrine-dbal/tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php:840

4) Doctrine\Tests\DBAL\Schema\ComparatorTest::testCompareChangedBlobColumn
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
     'removedNamespaces' => Array ()
     'newTables' => Array ()
     'changedTables' => Array (
-        'foo' => Doctrine\DBAL\Schema\TableDiff Object (...)
     )
     'removedTables' => Array ()
     'newSequences' => Array ()

/home/gregoire/dev/doctrine-dbal/tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php:1126
```

About the target branch, I'm unsure if there will be more releases of 2.12.x in the future, but it appears this branch has been contributed recently though. @morozov will know more about this.

morozov:
Yes, all bugfixes currently go to 2.12.x. It will be released right before 2.13.0.

lib/Doctrine/DBAL/Schema/Comparator.php
```diff
@@ -482,6 +482,10 @@ public function diffColumn(Column $column1, Column $column2)
             if ($properties1['fixed'] !== $properties2['fixed']) {
                 $changedProperties[] = 'fixed';
             }
+        } elseif ($properties1['type'] instanceof Types\BlobType) {
+            if ($properties1['length'] !== $properties2['length']) {
```
morozov: Looks like it will introduce a false-positive diff when comparing say `BLOB(1)` and `BLOB(2)` which both would translate to a `TINYBLOB` in MySQL.

While the fix seems to be within the current comparator design, the design obviously has flaws. A proper approach might be to add platform awareness to the comparator API and compare the SQL generated against a given platform. It will also likely eliminate the need for the hack introduced in https://github.com/doctrine/dbal/pull/3382.

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
```diff
@@ -808,6 +808,57 @@ protected function getQuotedAlterTableChangeColumnLengthSQL(): array
...
+        self::assertEquals(
+            $this->getAlterTableChangeBlobColumnLength(),
+            $this->platform->getAlterTableSQL($diff)
+        );
+    }
+
+    /**
+     * @return string[]
+     */
+    protected function getAlterTableChangeBlobColumnLength(): array
```
morozov: Why does it have to be a separate method?

tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
```diff
@@ -1050,6 +1051,31 @@ public function testCompareChangedColumn(): void
         self::assertEquals($expected, Comparator::compareSchemas($oldSchema, $newSchema));
     }
 
+    public function testCompareChangedColumnLength(): void
+    {
+        $oldSchema = new Schema();
+
+        $tableFoo = $oldSchema->createTable('foo');
+        $tableFoo->addColumn('id', 'string', ['length' => 127]);
```
morozov: The issue is about BLOB columns. Why does the test compare string column lengths?

morozov (commented):
Does the same logic apply to the TEXT/CLOB columns?

tests/Doctrine/Tests/DBAL/Schema/ComparatorTest.php
ausi: There was no test yet that tested the length changes for string columns. I mainly added this test to confirm my expectation about how the API works. I can move this test to its own pull request, or remove it alltogehter, however you prefer ☺️

tests/Doctrine/Tests/DBAL/Platforms/AbstractMySQLPlatformTestCase.php
ausi: I wanted to use the same coding style as other similar tests. Should I integrate it into the `testAlterTableChangeBlobColumnLength()` method instead?

lib/Doctrine/DBAL/Schema/Comparator.php
ausi: Agreed, but changing the comparator API cannot be done in a bugfix release I guess?

ausi:
> Does the same logic apply to the TEXT/CLOB columns?

Yes, added in a85737f782d9c6e830c0551214708a062382d17b

lib/Doctrine/DBAL/Schema/Comparator.php
morozov: `2.12.2` and `2.13.0` should be released at about the same time, so the release schedule shouldn't be a concern. If it's a non-breaking change (e.g. that adds a new optional parameter to the API), it could be released in `2.13.0`. Why does it have to be a bugfix release?

ausi: > Why does it have to be a bugfix release?

I would consider it a bug because the MySQL diff for blob/text columns doesn’t work currently.

> Looks like it will introduce a false-positive diff when comparing say `BLOB(1)` and `BLOB(2)`

Wouldn’t such false positives also occur for many other types already? Like `binary(fixed=true)` vs `binary(fixed=false)` on PostgreSQL/SQLite or `smallint` vs `integer` on SQLite?

lib/Doctrine/DBAL/Schema/DB2SchemaManager.php
```diff
@@ -76,10 +76,6 @@ protected function _getPortableTableColumnDefinition($tableColumn)
                 $fixed  = true;
                 break;
 
-            case 'clob':
-                $length = $tableColumn['length'];
-                break;
-
```
ausi: This needs to be removed in order for `Db2SchemaManagerTest::testDiffListTableColumns()` to succeed.

I think this is correct because `DB2Platform` does currently not handle lengths for `blob` and `text` types, see https://github.com/doctrine/dbal/blob/b366d93c7950af1fb66ed28eea072e1400f229b5/lib/Doctrine/DBAL/Platforms/DB2Platform.php#L127-L130 and https://github.com/doctrine/dbal/blob/b366d93c7950af1fb66ed28eea072e1400f229b5/lib/Doctrine/DBAL/Platforms/DB2Platform.php#L63-L66 and this change normalizes the behavior for `blob` and `text` to work the same now.

Toflar:
Any chance this is going to be merged into 3.1 if conflicts are resolved?
I just ran into this bug again :/

morozov:
> Any chance this is going to be merged into 3.1 if conflicts are resolved?

I do not intend to merge it in the current state: https://github.com/doctrine/dbal/pull/4551#discussion_r599719511.

ausi:
> While the fix seems to be within the current comparator design, the design obviously has flaws. A proper approach might be to add platform awareness to the comparator API and compare the SQL generated against a given platform. It will also likely eliminate the need for the hack introduced in #3382.

I will try to work on this. I would add a constructor to the `Comparator` class with an optional argument for passing an `AbstractPlatform` object.

Should I mark using the comparator without a platform as deprecated?

Is this considered to be a new feature and based on the `3.2.x` branch or a bugfix based on `3.1.x` instead?

morozov:
> I would add a constructor to the `Comparator` class with an optional argument for passing an `AbstractPlatform` object.

I think it'd make more sense to pass the platform to all methods explicitly. This will keep the comparator free of dependencies.

> Should I mark using the comparator without a platform as deprecated?

I wouldn't bother about deprecation and backward compatibility for now. I'd start with prototyping a solution with the new API and see if it solves the problem and reduces the number of platform-specific edge cases to be handled in the code. Once this is confirmed, we can think of the rest.

> Is this considered to be a new feature and based on the `3.2.x` branch or a bugfix based on `3.1.x` instead?

It's definitely a new feature.

ausi:
Done: #4659

morozov:
Closing in favor of https://github.com/doctrine/dbal/pull/4659.

PR #4574, opened 2021-04-01
Add test case for date literals instead of testing date columns only

ostrolucky:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | https://github.com/doctrine/orm/pull/8573

#### Summary

This is a follow-up from https://github.com/doctrine/orm/pull/8573. What we found was that DBAL doesn't handle this correctly for Pgsql

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/DataAccessTest.php
@@ -557,29 +557,34 @@ class DataAccessTest extends DbalFunctionalTestCase
         ];
     }
 
-    public function testDateArithmetics(): void
+    /**
+     * @param array<array-key, mixed> $parameters
+     *
+     * @dataProvider getDateArithmeticInputProvider
+     */
+    public function testDateArithmetics(string $input, array $parameters): void
     {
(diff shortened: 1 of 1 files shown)


ostrolucky:
@greg0ire is this accurate enough test for what you saw?

greg0ire:
The error message looks the same, so it certainly looks that way yes 👍

morozov:
It looks like PostgreSQL needs explicit [casting](https://stackoverflow.com/questions/28283807/invalid-input-syntax-for-type-interval/28283893#28283893) of a string literal when it represents a date. When it's a column, it already contains a date and so casting is not needed. The same applies to Oracle (`TO_DATE()`) and IBM DB2.

Strictly speaking, it's not a bug: https://github.com/doctrine/dbal/blob/872167f787c302d20b0c8c72ea18e76299505ab2/src/Platforms/AbstractPlatform.php#L874

The SQL expression provided in the test produces a string literal that some platforms don't cast to a date automatically. It's more of a reasonable improvement.

morozov:
Closing as the PR in its current state is not mergeable. Could be reopened once the logic expected by the test is implemented.

PR #4577, opened 2021-04-02
Remove deprecated schema- and namespace-related APIs

### LINKED ISSUE #4503: Inconsistent SchemaManager API across platforms

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

See #4503.

### DIFF
--- a/src/Schema/PostgreSQLSchemaManager.php
+++ b/src/Schema/PostgreSQLSchemaManager.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\DBAL\Schema;
 
+use Doctrine\DBAL\Exception;
 use Doctrine\DBAL\Platforms\PostgreSQL94Platform;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\DBAL\Types\Types;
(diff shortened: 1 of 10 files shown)


src/Schema/PostgreSQLSchemaManager.php
```diff
@@ -93,36 +71,34 @@ public function getSchemaSearchPaths(): array
...
-     *
-     * @internal The method should be only used from within the PostgreSQLSchemaManager class hierarchy.
+     * Gets names of all existing schemas in the current user's search path.
      *
      * @return array<int, string>
+     *
+     * @throws Exception
      */
-    public function getExistingSchemaSearchPaths(): array
+    protected function getExistingSchemaSearchPaths(): array
```
greg0ire: Should these also become final?

greg0ire (dismissed):
Approving in case the answer to my question is no.

src/Schema/PostgreSQLSchemaManager.php
morozov: The purpose of this API is not really clear to me: there's a single use case which uses only the first element of the path: https://github.com/doctrine/dbal/blob/d19875a0f127c1cbeb77e160619ec36e096dc8d3/src/Schema/PostgreSQLSchemaManager.php#L177-L180
So I'd expect some more rework here in the future.

But yeah, `final` for the wrapper makes sense regardless. I'd leave the "determine" method as is (in case if somebody hacks their own logic in there).

greg0ire (approved):

PR #4578, opened 2021-04-02
Predictable `QueryBuilder::executeQuery()` and `QueryBuilder::executeStatement()`

### LINKED ISSUE #4461: Introduce fetch* methods in QueryBuilder for SELECT queries

PowerKiKi:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | https://github.com/doctrine/dbal/issues/4461#issuecomment-778875457

#### Summary

This deprecates `QueryBuilder::execute()`, because its return type is
unpredictable and raises issues with static analysis tools such as PHPStan.

Instead you should use either `QueryBuilder::executeQuery()` or
`QueryBuilder::executeStatement()`, depending on whether the queryBuilder is a query (SELECT)
or a statement (INSERT, UPDATE, DELETE).

You might also consider the use of the new shortcut methods, such as:

- `fetchAllAssociative()`
- `fetchAllAssociativeIndexed()`
- `fetchAllKeyValue()`
- `fetchAllNumeric()`
- `fetchAssociative()`
- `fetchFirstColumn()`
- `fetchNumeric()`
- `fetchOne()`

This commit is a direct follow-up to https://github.com/doctrine/dbal/issues/4461
where those shortcut methods where introduced.

### DIFF
--- a/src/Query/QueryBuilder.php
+++ b/src/Query/QueryBuilder.php
@@ -301,7 +301,33 @@ class QueryBuilder
     }
 
     /**
+     * Executes an SQL query (SELECT) and returns a Result.
+     *
+     * @throws Exception
+     */
+    public function executeQuery(): Result
+    {
+        return $this->connection->executeQuery($this->getSQL(), $this->params, $this->paramTypes);
+    }
+
+    /**
+     * Executes an SQL statement and returns the number of affected rows.
+     *
+     * Should be used for INSERT, UPDATE and DELETE
+     *
+     * @return int The number of affected rows.
+     *
+     * @throws Exception
+     */
+    public function executeStatement(): int
+    {
+        return $this->connection->executeStatement($this->getSQL(), $this->params, $this->paramTypes);
+    }
+
+    /**
      * Executes this query using the bound parameters and their types.
+     *
+     * @deprecated Use {@link executeQuery()} or {@link executeStatement()} instead.
      *
      * @return Result|int
      *
@@ -310,8 +336,20 @@ class QueryBuilder
     public function execute()
     {
         if ($this->type === self::SELECT) {
+            Deprecation::trigger(
+                'doctrine/dbal',
+                'https://github.com/doctrine/dbal/pull/4578',
+                'QueryBuilder::execute() is deprecated, use QueryBuilder::executeQuery() for SQL queries instead.'
+            );
+
             return $this->connection->executeQuery($this->getSQL(), $this->params, $this->paramTypes);
         }
+
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/4578',
+            'QueryBuilder::execute() is deprecated, use QueryBuilder::executeStatement() for SQL statements instead.'
+        );
 
         return $this->connection->executeStatement($this->getSQL(), $this->params, $this->paramTypes);
     }
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -36,6 +36,23 @@ Use `Connection::createSchemaManager()` instead.
 The usage of `Connection::$_expr` and `Connection::getExpressionBuilder()` is deprecated.
 Use `Connection::createExpressionBuilder()` instead.
 
+## Deprecated `QueryBuilder::execute()`
+
+The usage of `QueryBuilder::execute()` is deprecated. Use either `QueryBuilder::executeQuery()` or
+`QueryBuilder::executeStatement()`, depending on whether the queryBuilder is a query (SELECT) or a statement (INSERT,
+UPDATE, DELETE).
+
+You might also consider the use of the new shortcut methods, such as:
+
+- `fetchAllAssociative()`
+- `fetchAllAssociativeIndexed()`
+- `fetchAllKeyValue()`
+- `fetchAllNumeric()`
+- `fetchAssociative()`
+- `fetchFirstColumn()`
+- `fetchNumeric()`
+- `fetchOne()`
+
 # Upgrade to 3.0
 
 ## BC BREAK: leading colon in named parameter names not supported
--- a/tests/Query/QueryBuilderTest.php
+++ b/tests/Query/QueryBuilderTest.php
@@ -7,6 +7,7 @@ use Doctrine\DBAL\ParameterType;
 use Doctrine\DBAL\Query\Expression\ExpressionBuilder;
 use Doctrine\DBAL\Query\QueryBuilder;
 use Doctrine\DBAL\Query\QueryException;
+use Doctrine\DBAL\Result;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\DBAL\Types\Types;
 use PHPUnit\Framework\MockObject\MockObject;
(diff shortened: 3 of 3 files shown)


beberlei:
If we do this change, then we could look to add this `Statement` in general maybe even in 2.x, because it would help us solve the issue with `Statement::execute()` returning a boolean at the moment. The API could be in general `Statement::executeStatement` (yeah naming) and `Statement::executeQuery`.

morozov:
> yeah naming

The naming problem is not on our end. The "statement" on the left-hand side is a _prepared_ statement. The "statement" on the right-hand side is a SQL _statement_ (DDL/DML/DCL) as opposed to an SQL _query_. Having it named as [`Statement::executeUpdate()`](https://docs.oracle.com/javase/7/docs/api/java/sql/Statement.html#executeUpdate(java.lang.String)) would be worse IMO.

morozov:
Looks good to me. @beberlei do you want to backport it to `2.x` or it's not necessary for the upgrade path improvement?

beberlei:
@morozov the query builder change is ok for 3.1.x I believe, since the API for QueryBuilder::execute() does not return a boolean, it is already ok to use, would be a new migration step in the future for users. I don't believe there is pressure moving this to 2.13

beberlei:
#4580 would be the one that is more interesting for the 2.x branches

morozov (commented):
@PowerKiKi please add a runtime deprecation using `Deprecation::trigger()`.

PowerKiKi:
I added the `Deprecation::trigger()` call as required.

src/Query/QueryBuilder.php
```diff
@@ -335,6 +335,13 @@ public function executeStatement(): int
      */
     public function execute()
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/4578',
+            'QueryBuilder::execute is deprecated, instead use QueryBuilder::executeQuery (for SELECT) or' .
```
morozov: This could be moved under the `if` statement and made more specific (without the "or").

PowerKiKi: Good idea, I pushed it

greg0ire: You should run `vendor/bin/phpcbf`

PowerKiKi: done

src/Query/QueryBuilder.php
```diff
@@ -300,19 +300,57 @@ public function fetchFirstColumn(): array
...
      *
      * @throws Exception
      */
     public function execute()
     {
         if ($this->type === self::SELECT) {
+            Deprecation::trigger(
+                'doctrine/dbal',
+                'https://github.com/doctrine/dbal/pull/4578',
+                'QueryBuilder::execute is deprecated, instead use QueryBuilder::executeQuery'
```
morozov: Please reference methods using parentheses like it's done in other places. I'd rephrase it to:
```
QueryBuilder::execute() is deprecated, use QueryBuilder::executeQuery() for SQL queries instead.
```

src/Query/QueryBuilder.php
```diff
@@ -300,19 +300,57 @@ public function fetchFirstColumn(): array
...
+            );
+
             return $this->connection->executeQuery($this->getSQL(), $this->params, $this->paramTypes);
         }
 
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/4578',
+            'QueryBuilder::execute is deprecated, instead use QueryBuilder::executeStatement'
```
morozov:
```
QueryBuilder::execute() is deprecated, use QueryBuilder:: executeStatement() for SQL statements instead.
```

src/Query/QueryBuilder.php
```diff
@@ -300,19 +300,57 @@ public function fetchFirstColumn(): array
         return $this->connection->fetchFirstColumn($this->getSQL(), $this->params, $this->paramTypes);
     }
 
+    /**
+     * Executes an SQL query (SELECT) and returns a Result.
```
morozov, line 304: This may be misleading since it works also for `DESCRIBE`, `SHOW` and so on. Let's rephrase to:
```
Executes an SQL query that returns a Result.
```

src/Query/QueryBuilder.php
```diff
@@ -300,19 +300,57 @@ public function fetchFirstColumn(): array
...
+     */
+    public function executeQuery(): Result
+    {
+        return $this->connection->executeQuery($this->getSQL(), $this->params, $this->paramTypes);
+    }
+
+    /**
+     * Executes an SQL statement and returns the number of affected rows.
+     *
+     * Should be used for INSERT, UPDATE and DELETE
```
morozov, line 316: It's not limited to DML (`INSERT`, etc.), it could be also DDL and other non-DQL statements. See [JDBC](https://docs.oracle.com/javase/7/docs/api/java/sql/Statement.html) for reference.

morozov (commented):
Looks good overall. Please squash at the end.

src/Query/QueryBuilder.php
PowerKiKi: I'd agree with that on the `Connection` class, but AFAIK, there is no way to build anything else than a `SELECT` query with a `QueryBuilder`. So I don't think we should document an impossible use-case.

WDYT ?

see: https://github.com/doctrine/dbal/blob/461b3b70012182777977ad8bef105f1cfaeedcf9/src/Query/QueryBuilder.php#L40-L46

src/Query/QueryBuilder.php
PowerKiKi: Idem, the `QueryBuilder` cannot build anything else than DML. Should we really mention that ?

src/Query/QueryBuilder.php
PowerKiKi: done

src/Query/QueryBuilder.php
PowerKiKi: done

src/Query/QueryBuilder.php
morozov: You're right, I didn't take the scope into account. Sorry for the confusion.

src/Query/QueryBuilder.php
morozov: No, you're right.

morozov (approved):

morozov:
Thanks, @PowerKiKi.

PR #4580, opened 2021-04-02
Deprecate Statement::execute in favor of Statement::executeStatement/executeQuery

### LINKED ISSUE #4569: Doctrine DBAL 2.13 $statement->execute() returns bool (instead of Result)
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.13.0

#### Summary

The release notes for DBAL 2.13 (https://www.doctrine-project.org/2021/03/29/dbal-2.13.html) stated that "_DBAL 3.0 extracts all fetch-methods from the Statement API and moved them to a new Result API that is returned from Statement::execute. **We have backported this API to 2.13_**"
It also gives an example that with 2.13 you could/should use code like this:
```
$result = $statement->execute();
while (($row = $result->fetchAssociative()) !== false) {
}
```
#### Current behaviour

$statement->execute() still returns only a boolean value instead of a resultset.

#### How to reproduce

- Upgrade to 2.13.0
- try to use fetch() on the result of $statement->execute()

#### Expected behaviour

$statement->execute() should return a resultset instead of bool.

beberlei:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixes        | https://github.com/doctrine/dbal/issues/4569

#### Summary

Adopt the `Connection` API with differentation of `executeQuery` and `executeStatement` also for `Statement`.

This would allow to return `Result` or `int` respectively instead of `Statement::execute` returning a boolean.

Old code:

```php
$sql = 'SELECT * FROM test WHERE id = ?';
$stmt = $connection->prepare($sql);
$stmt->bindValue(1, $id);

$stmt->execute();
$row = $stmt->fetch();
```

New Code:

```php
$sql = 'SELECT * FROM test WHERE id = ?';
$stmt = $connection->prepare($sql);
$stmt->bindValue(1, $id);

$row = $stmt->executeQuery()->fetchAssociative();
```

This deprecation is going to be spread out across mre versions, as DBAL 3.x is already released with Statement::execute as well, we can only remove it earliest in DBAL 4.

### DIFF
--- a/lib/Doctrine/DBAL/Statement.php
+++ b/lib/Doctrine/DBAL/Statement.php
@@ -7,6 +7,7 @@ use Doctrine\DBAL\Driver\Exception;
 use Doctrine\DBAL\Driver\Statement as DriverStatement;
 use Doctrine\DBAL\Exception\NoKeyValue;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Result as BaseResult;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\Deprecations\Deprecation;
 use IteratorAggregate;
@@ -179,6 +180,34 @@ class Statement implements IteratorAggregate, DriverStatement, Result
         }
 
         return $stmt;
+    }
+
+    /**
+     * Executes the statement with the currently bound parameters and return result.
+     *
+     * @param mixed[]|null $params
+
+     * @throws Exception
+     */
+    public function executeQuery(?array $params = null): BaseResult
+    {
+        $this->execute($params);
+
+        return new ForwardCompatibility\Result($this);
+    }
+
+    /**
+     * Executes the statement with the currently bound parameters and return affected rows.
+     *
+     * @param mixed[]|null $params
+
+     * @throws Exception
+     */
+    public function executeStatement(?array $params = null): int
+    {
+        $this->execute($params);
+
+        return $this->rowCount();
     }
 
     /**
--- a/lib/Doctrine/DBAL/Statement.php
+++ b/lib/Doctrine/DBAL/Statement.php
@@ -148,6 +148,8 @@ class Statement implements IteratorAggregate, DriverStatement, Result
     /**
      * Executes the statement with the currently bound parameters.
      *
+     * @deprecated Statement::execute() is deprecated, use Statement::executeQuery() or executeStatement() instead
+     *
      * @param mixed[]|null $params
      *
      * @return bool TRUE on success, FALSE on failure.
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Statement.php
```diff
@@ -181,6 +190,34 @@ public function execute($params = null)
...
     }
 
+    /**
+     * Executes the statement with the currently bound parameters and return result.
+     *
+     * @param mixed[]|null $params
+
+     * @throws Exception
+     */
+    public function executeQuery(?array $params = null): BaseResult
```
morozov: I'm curious if we could get rid of the nullable type here. There's no difference between not passing an argument and passing an empty array (at least in the non-PDO drivers where we implement it explicitly). As for the PDO ones, it could be checked separately.

lib/Doctrine/DBAL/Statement.php
```diff
@@ -181,6 +190,34 @@ public function execute($params = null)
...
+        $this->execute($params);
+
+        return new ForwardCompatibility\Result($this);
+    }
+
+    /**
+     * Executes the statement with the currently bound parameters and return affected rows.
+     *
+     * @param mixed[]|null $params
+
```
morozov: A `*` is missing.

lib/Doctrine/DBAL/Statement.php
```diff
@@ -155,6 +158,12 @@ public function bindParam($param, &$variable, $type = ParameterType::STRING, $le
      */
     public function execute($params = null)
     {
+        Deprecation::triggerIfCalledFromOutside(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/4580',
```
morozov, line 163: I think it's preferable to specify the issue link which would describe the reason for the deprecation, not the PR link where the deprecation was made.

lib/Doctrine/DBAL/Statement.php
```diff
@@ -181,6 +190,34 @@ public function execute($params = null)
         return $stmt;
     }
 
+    /**
+     * Executes the statement with the currently bound parameters and return result.
+     *
+     * @param mixed[]|null $params
+
```
morozov: A `*` is missing.

lib/Doctrine/DBAL/Statement.php
beberlei: #4569 only reports this as a bug, because ::execute still returns a boolean. This PR here has all the summary about why this change makes sense and examples of new/old, I would prefer this as a user trying to understand the deprecation. Should I open a new issue that copies over the summary?

morozov: No, let's keep it simple. But let's give it some thought in the meanwhile. Technically, both the PR and the issue are discoverable via Git history. But as an end-user, I might be more interested in "why" it was deprecated (what the problem with this API is, the issue) rather than "how it was deprecated" (the code change). Just thinking out loud.

morozov (commented):
Looks good.

morozov (approved):

PR #4581, opened 2021-04-02
Fix truncated oracle fetch result

### LINKED ISSUE #4553: DBAL FETCH silently returns truncated results when database emits an error after initial prefetch
#### Summary

If one uses a FETCH to loop over a data set that is larger than the set prefetch size, it is possible that the result set will silently be truncated without an exception being thrown in PHP.  DBAL should screen for DB errors for each fetch and emit a PHP exception.  

This can become a problem when using distributed databases that are replicated from a central database source.  Datasets could become invalidated when querying data on a replicated database that started before a DB refresh.  The symptom of this issue will be that result sets can be truncated to any increment of the set prefetch size depending on when the results get invalidated on the DB.  The rest of the PHP program could then be working with unexpected truncated result sets.  Currently the only way to catch truncated result errors is to wrap the fetch method with a custom warning handler that screens for DB errors being emitted as PHP warnings (The PHP OCI8 driver only converts DB errors during a fetch to PHP warnings).

#### Current behavior

DBAL fetch will ignore db errors on subsequent fetches and can return truncated results without emitting a PHP exception.

#### How to reproduce

*Example reproduction of issue with Oracle*:
* Create a DB connection and assign the Doctrine\DBAL\Connection object to the $dbh variable
* Create a pipelined function on the database that returns numbers up to your defined php.ini oci8.default_prefetch size plus 10 more.  This will cause the result set to require at least 2 DB calls.  For example, if your defined prefetch value is 100, than this function will return 110 rows with each row containing a sequential number (e.g. 0 => ['column_value' => 1], 1 => ['column_value' => 2], ... ).
   ```php
   $dbh->executeQuery(
       "CREATE TYPE return_numbers AS TABLE OF NUMBER(11)"
   );
   $dbh->executeQuery("
       CREATE OR REPLACE FUNCTION test_oracle_fetch_failure 
       RETURN return_numbers PIPELINED 
       AS 
           v_number_list return_numbers; 
       BEGIN 
           SELECT ROWNUM r 
           BULK COLLECT INTO v_number_list 
           FROM DUAL 
           CONNECT BY ROWNUM <= " . ini_get("oci8.default_prefetch") . " + 10;
   
           FOR i IN 1 .. v_number_list.COUNT 
           LOOP 
               PIPE ROW (v_number_list(i)); 
           END LOOP; 
   
           RETURN; 
       END;
   ");
   ```
* Execute a select statement on the pipelined function
   ```php
   echo "Your default OCI8 prefetch value is " . ini_get("oci8.default_prefetch") . PHP_EOL;
   $statement = $dbh->prepare("SELECT * FROM TABLE(test_oracle_fetch_failure())");
   $statement->execute();
   ```
* Access the first row result causing the prefetch to be executed to locally cache rows up to the default prefetch number (often defaults to 100 rows)
   ```php
   $firstRow = $statement->fetch();
   echo $firstRow['column_value'] . PHP_EOL // 1
   ```
* Invalidate the original data set by updating the pipelined function on the DB to return numbers up to your defined php.ini prefetch size plus 20
   ```php
   $dbh->executeQuery("
       CREATE OR REPLACE FUNCTION test_oracle_fetch_failure 
       RETURN return_numbers PIPELINED 
       AS 
           v_number_list return_numbers; 
       BEGIN 
           SELECT ROWNUM r 
           BULK COLLECT INTO v_number_list 
           FROM DUAL 
           CONNECT BY ROWNUM <= " . ini_get("oci8.default_prefetch") . " + 20;
   
           FOR i IN 1 .. v_number_list.COUNT 
           LOOP 
               PIPE ROW (v_number_list(i)); 
           END LOOP; 
   
           RETURN; 
       END;
   ");
   ```
* Attempt to access the remaining rows.  However, after iterating through the first batch of locally cached prefetched rows, PHP OCI8 driver will see the oracle error "ORA-04068: existing state of packages has been discarded" when it needs to prefetch the next batch of rows above the first set of prefetched rows and only emit a PHP warning.  For example, if your prefetch value is 100, than this will only return the first 100 rows that were originally prefetched, and fail to fetch the remaining 10 rows. 
   ```php
   // Even though one is expecting this to echo out numbers up to 10 higher than your default prefetch size, 
   // the following will only echo numbers up to your default prefetch size (often 100 rows) and no more. 
   // No exception will be thrown; only a PHP warning.
   while ($row = $statement->fetch()) {
       echo $row['column_value'] . PHP_EOL;
   }
   ```
* DBAL fetch does not screen for PHP warnings and DBAL fetch will behave as if all results have been returned without error; the PHP program will continue on without exception even though only a truncated result set was returned instead of the full result set expected. 

#### Expected behavior

DBAL should screen for DB errors on each fetch and emit a PHP exception instead of allowing truncated result sets.

amenning:
|      Q       |   A
|------------- | -----------
| Type         | bug
| Related issues | #4553

#### Summary

This adds a test that demonstrates the issue where a result set
larger than the oci8 default prefetch value is queried, but a truncated
row set is returned due to the dataset being invalidated during the
fetch sequence.  No php exception is thrown.

### DIFF
--- a/src/Driver/OCI8/Result.php
+++ b/src/Driver/OCI8/Result.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\DBAL\Driver\OCI8;
 
 use Doctrine\DBAL\Driver\FetchUtils;
+use Doctrine\DBAL\Driver\OCI8\Exception\PHPError;
 use Doctrine\DBAL\Driver\Result as ResultInterface;
 
 use function oci_cancel;
@@ -12,7 +13,11 @@ use function oci_fetch_all;
 use function oci_fetch_array;
 use function oci_num_fields;
 use function oci_num_rows;
+use function preg_match_all;
+use function restore_error_handler;
+use function set_error_handler;
 
+use const E_WARNING;
 use const OCI_ASSOC;
 use const OCI_FETCHSTATEMENT_BY_COLUMN;
 use const OCI_FETCHSTATEMENT_BY_ROW;
@@ -115,10 +120,52 @@ final class Result implements ResultInterface
      */
     private function fetch(int $mode)
     {
-        return oci_fetch_array(
-            $this->statement,
-            $mode | OCI_RETURN_NULLS | OCI_RETURN_LOBS
+        set_error_handler(
+            static function (int $errno, string $errstr, string $errfile, int $errline): bool {
+                /*
+                 * Screen PHP error messages for any error that the OCI8 extension
+                 * doesn't report via oci_error()
+                 *
+                 * @link http://www.dba-oracle.com/t_error_code_list.htm
+                 */
+                if (preg_match_all('/ORA-(\d+)/', $errstr, $matches) === 0) {
+                    // If no ORA error codes are present in error message,
+                    // skip this error and bubble this error up to
+                    // the normal error handler
+                    return false;
+                }
+
+                foreach ($matches[1] as $oraErrorCode) {
+                    switch ($oraErrorCode) {
+                        case '04061':
+                        case '04065':
+                        case '04068':
+                            throw PHPError::new(
+                                'There was an error before all rows could be fetched.',
+                                $errno,
+                                $errstr,
+                                $errfile,
+                                $errline
+                            );
+                    }
+                }
+
+                // If ORA error codes in message do not match specified blocklist
+                // ORA error codes, skip this error and bubble this error up to
+                // the normal error handler
+                return false;
+            },
+            E_WARNING
         );
+
+        try {
+            return oci_fetch_array(
+                $this->statement,
+                $mode | OCI_RETURN_NULLS | OCI_RETURN_LOBS
+            );
+        } finally {
+            restore_error_handler();
+        }
     }
 
     /**
--- a/tests/Functional/Driver/OCI8/ResultTest.php
+++ b/tests/Functional/Driver/OCI8/ResultTest.php
@@ -1,0 +1,174 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Tests\Functional\Driver\OCI8;
+
+use Doctrine\DBAL\Driver\OCI8\Driver;
+use Doctrine\DBAL\Exception\DriverException;
+use Doctrine\DBAL\Tests\FunctionalTestCase;
+use Doctrine\DBAL\Tests\TestUtil;
+use Generator;
+
+use function ini_get;
+use function sprintf;
+
+/**
+ * @requires extension oci8
+ */
+class ResultTest extends FunctionalTestCase
+{
+    /**
+     * Database connection parameters for functional test case
+     *
+     * @var array<string,mixed>
+     */
+    private $connectionParams;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->connectionParams = TestUtil::getConnectionParams();
+
+        if ($this->connection->getDriver() instanceof Driver) {
+            return;
+        }
+
+        self::markTestSkipped('oci8 only test.');
+    }
+
+    protected function tearDown(): void
+    {
+        $this->connection->executeQuery(sprintf(
+            'DROP FUNCTION test_oracle_fetch_failure',
+            $this->connectionParams['user']
+        ));
+        $this->connection->executeQuery(sprintf(
+            'DROP TYPE return_numbers',
+            $this->connectionParams['user']
+        ));
+
+        parent::tearDown();
+    }
+
+    /**
+     * This test will recreate the case where a data set that is larger than the
+     * oci8 default prefetch is invalidated on the database after a fetch has begun,
+     * but before the fetch has completed.
+     *
+     * Note that this test requires 2 separate user connections so that the
+     * pipelined function can be changed mid fetch.
+     *
+     * @dataProvider dataProviderForTestTruncatedFetch
+     */
+    public function testTruncatedFetch(
+        bool $invalidateDataMidFetch
+    ): void {
+        if ($invalidateDataMidFetch) {
+            $this->expectException(DriverException::class);
+            $this->expectErrorMessageMatches(
+                '/^An exception occurred in the driver: '
+                . 'There was an error before all rows could be fetched. '
+                . 'Error number: 2, Error String: oci_fetch_array.*/'
+            );
+        }
+
+        // Create a pipelined funtion that returns 10 rows more than the
+        // oci8 default prefetch
+        $this->createReturnTypeNeededForPipelinedFunction();
+        $expectedTotalRowCount = (int) ini_get('oci8.default_prefetch') + 10;
+        $this->createOrReplacePipelinedFunction($expectedTotalRowCount);
+
+        // Create a separate connection from that used to create/update the function
+        // This must be a different user with permissions to change the given function
+        $separateConnection = TestUtil::getPrivilegedConnection();
+
+        // Query the pipelined function to get initial dataset
+        $statement = $separateConnection->prepare(sprintf(
+            'SELECT * FROM TABLE(%s.test_oracle_fetch_failure())',
+            $this->connectionParams['user']
+        ));
+        $result    = $statement->execute();
+
+        // Access the first result to cause the first X rows to be prefetched
+        // as defined by oci8.default_prefetch (often 100 rows)
+        $firstNumber = $result->fetchOne();
+
+        if ($invalidateDataMidFetch) {
+            // Invalidate the original dataset by changing the pipelined function
+            // after the initial prefetch that caches locally the first X results
+            $this->createOrReplacePipelinedFunction($expectedTotalRowCount + 10);
+        }
+
+        while ($number = $result->fetchOne()) {
+            // Attempt to access all remaining rows from the original fetch
+            // The rows locally cached from the default prefetch will first be used
+            // but when the result attempts to get the remaining 10 rows beyond
+            // the first prefetch, nothing will be returned
+            //
+            // PHP oci8 oci_fetch_array will issue a PHP E_WARNING when the 2nd prefetch occurs
+            // oci_fetch_array(): ORA-04068: existing state of packages has been discarded
+            // ORA-04061: existing state of function "ROOT.TEST_ORACLE_FETCH_FAILURE" has been invalidated
+            // ORA-04065: not executed, altered or dropped function "ROOT.TEST_ORACLE_FETCH_FAILURE"
+            //
+            // If there was no issue, this should have returned rows totalling 10
+            // higher than the oci8 default prefetch
+            continue;
+        }
+
+        $this->assertEquals(
+            $expectedTotalRowCount,
+            $result->rowCount(),
+            sprintf(
+                'Expected to have %s total rows fetched but only found %s rows fetched',
+                $expectedTotalRowCount,
+                $result->rowCount()
+            )
+        );
+    }
+
+    public function dataProviderForTestTruncatedFetch(): Generator
+    {
+        yield 'it should return all rows if no data invalidation occurs'
+            => [false];
+
+        yield 'it should convert oci8 data invalidation error to DriverException'
+            => [true];
+    }
+
+    private function createReturnTypeNeededForPipelinedFunction(): void
+    {
+        $this->connection->executeQuery(
+            'CREATE TYPE return_numbers AS TABLE OF NUMBER(11)'
+        );
+    }
+
+    /**
+     * This will create a pipelined function that returns X rows with
+     * each row returning a single column_value of that row's row number.
+     * The total number of rows returned is equal to $totalRowCount.
+     */
+    private function createOrReplacePipelinedFunction(int $totalRowCount): void
+    {
+        $this->connection->executeQuery(sprintf(
+            'CREATE OR REPLACE FUNCTION test_oracle_fetch_failure
+            RETURN return_numbers PIPELINED
+            AS
+                v_number_list return_numbers;
+            BEGIN
+                SELECT ROWNUM r
+                BULK COLLECT INTO v_number_list
+                FROM DUAL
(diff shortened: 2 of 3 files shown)


src/Driver/OCI8/Exception/TruncatedQueryResult.php
```diff
@@ -0,0 +1,20 @@
...
+namespace Doctrine\DBAL\Driver\OCI8\Exception;
+
+use Doctrine\DBAL\Driver\AbstractException;
+
+/**
+ * @internal
+ *
+ * @psalm-immutable
+ */
+final class TruncatedQueryResult extends AbstractException
```
morozov: I'd name this class as `PHPError` meaning that this exception could be constructed from any PHP error that the OCI8 extension doesn't report via `oci_error()`. It could be reused for other cases since it's just a named constrictor.

src/Driver/OCI8/Result.php
```diff
@@ -115,10 +121,14 @@ public function free(): void
      */
     private function fetch(int $mode)
     {
-        return oci_fetch_array(
+        set_error_handler($this->getFetchErrorHandler(), E_WARNING);
+        $result = oci_fetch_array(
             $this->statement,
             $mode | OCI_RETURN_NULLS | OCI_RETURN_LOBS
         );
+        restore_error_handler();
```
morozov: The error handler must be finally restored, because otherwise it won't be called of an exception is thrown:
```php
set_error_handler();

try {
    return oci_fetch_array();
} finally {
    restore_error_handler();
}
```

src/Driver/OCI8/Result.php
```diff
@@ -136,4 +146,27 @@ private function fetchAll(int $mode, int $fetchStructure): array
 
         return $result;
     }
+
+    private function getFetchErrorHandler(): callable
```
morozov: There's no need to create a new error handler for each fetch since it's stateless. It could be `private handleError(int $errno, ...)` instead.

src/Driver/OCI8/Result.php
```diff
@@ -136,4 +146,27 @@ private function fetchAll(int $mode, int $fetchStructure): array
...
+            string $errfile,
+            int $errline
+        ): bool {
+            if (stripos($errstr, 'ORA-') === false) {
+                return true;
+            }
+
+            throw TruncatedQueryResult::new(sprintf(
+                'There was an error before all rows could be fetched. '
+                . 'Error number: %s, Error String: %s, Error File: %s, Error Line: %s',
```
morozov: Please more the `sprintf()` logic to the exception constructor (`new()`). The point of those named classes is to make exception instantiation as straightforward as possible by keeping all the error formatting logic away from the caller.

tests/Functional/Driver/OCI8/ResultTest.php
```diff
@@ -0,0 +1,168 @@
...
+        // oci8 default prefetch
+        $this->createReturnTypeNeededForPipelinedFunction();
+        $expectedTotalRowCount = (int) ini_get('oci8.default_prefetch') + 10;
+        $this->createPipelinedFunction($expectedTotalRowCount);
+
+        // Create a separate connection from that used to create/update the function
+        // This must be a different user with permissions to change the given function
+        $separateConnection = $this->createSeparateConnection([
+            'user'     => 'system',
+            'password' => 'oracle',
```
morozov: User credentials shouldn't be hardcoded. Are they different from the ones specified in the test configuration?

src/Driver/OCI8/Result.php
```diff
@@ -136,4 +146,27 @@ private function fetchAll(int $mode, int $fetchStructure): array
...
+
+    private function getFetchErrorHandler(): callable
+    {
+        return static function (
+            int $errno,
+            string $errstr,
+            string $errfile,
+            int $errline
+        ): bool {
+            if (stripos($errstr, 'ORA-') === false) {
```
morozov: I wouldn't trigger an exception on any warning. I'd create a white list of the error codes that should be converted to an exception. More codes could be added to the list later but each would have to have a functional test like here.

For instance `pdo_oci` is guilty of converting informational messages to an error ([bug#77120](https://bugs.php.net/bug.php?id=77120)).

We could do the following:
```php
if (preg_match('/ORA-(\d+)/', $errstr, $matches) === 0) {
    return false;
}

switch ($matches[1]) {
    case 12345:
    case 67890:
        // handle error

    default:
        return false;
}
```

tests/Functional/Driver/OCI8/ResultTest.php
```diff
@@ -0,0 +1,168 @@
...
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->connectionParams = TestUtil::getConnectionParams();
+
+        if ($this->connection->getDriver() instanceof Driver) {
+            return;
+        }
+
+        self::markTestSkipped('oci8 only test.');
```
morozov, line 37: This is out of scope for now but this case is relevant for `pdo_oci` as well.

morozov (commented):
Looks good overall. Please see inline comments.

Since it's a non-breaking improvement, it could target `3.1.x` instead of `4.0.x`.

tests/Functional/Driver/OCI8/ResultTest.php
amenning: Yeah, this test requires two separate connections so that the pipelined function can be changed mid-fetch.  The default test user is 'doctrine' for the DBAL CI tests.  This hard coded user is an existing system user on the same CI database instance.  I was trying to get 2 separate connections for the default test user "doctrine" to work but couldn't get DBAL to all those two connections to work independently.  Is there a way to create two connections using the same user credentials that can work independently with DBAL?

Otherwise, I wasn't sure how to update the documentation in the phpunit.xml.dist to indicate that just for OCI8 tests, a developer would need to add credentials for two users.

src/Driver/OCI8/Result.php
```diff
@@ -115,10 +120,18 @@ public function free(): void
      */
     private function fetch(int $mode)
     {
-        return oci_fetch_array(
-            $this->statement,
-            $mode | OCI_RETURN_NULLS | OCI_RETURN_LOBS
-        );
+        set_error_handler([$this, 'handleError'], E_WARNING);
```
amenning, line 123: The coding style check doesn't register this as using the handleError method.  Is there a better way to call this to pass that coding style check?

Coding Standards (7.4)
Class Result contains unused private method handleError().

tests/Functional/Driver/OCI8/ResultTest.php
amenning: Would you like me to expand this PR to include pdo_oci?

src/Driver/OCI8/Result.php
```diff
@@ -115,10 +120,23 @@ public function free(): void
      */
     private function fetch(int $mode)
     {
-        return oci_fetch_array(
-            $this->statement,
-            $mode | OCI_RETURN_NULLS | OCI_RETURN_LOBS
+        set_error_handler(
+            function (int $errno, string $errstr, string $errfile, int $errline): bool {
+                return Result::handleError($errno, $errstr, $errfile, $errline);
```
morozov: You can move the implementation right here. It's not that big of a deal. You can also suppress the code style error (which is a false-positive) in `phpcs.xml`.

src/Driver/OCI8/Result.php
```diff
@@ -115,10 +120,23 @@ public function free(): void
...
+            },
+            E_WARNING
         );
+
+        try {
+            $result = oci_fetch_array(
```
morozov: You can return right away, there's no need for a variable.

tests/Functional/Driver/OCI8/ResultTest.php
morozov: If you don't mind. Please include pdo_oci here and see if it also needs fixing. We can revert if it turns out that fixing it takes too much effort.

src/Driver/OCI8/Result.php
```diff
@@ -136,4 +154,46 @@ private function fetchAll(int $mode, int $fetchStructure): array
...
+                case '04068':
+                    throw PHPError::new(
+                        'There was an error before all rows could be fetched.',
+                        $errno,
+                        $errstr,
+                        $errfile,
+                        $errline
+                    );
+                default:
+                    break;
```
morozov: It looks like the `foreach` loop will either throw an exception or break at the very first iteration. Why do we need `preg_match_all()` then?

If we need to look for any known error, the loop shouldn't break by default. If we want to consider only the first encountered error code, there shouldn't be a loop.

amenning: Good catch, I added in the break to deal with a code style failure.  Since the error message could contain multiple ORA codes, like the example given, we want to loop over all codes in case the first code doesn't match the block list.  I'll change that to properly continue the loop.

tests/Functional/Driver/OCI8/ResultTest.php
amenning: I created a test for the pdo_oci but no code changes were needed.  The pdo_oci driver successfully throws the following exception below during the fetch, which gets converted into a DriverException.  I'll omit that test unless you want that added in.

```
Caused by
Doctrine\DBAL\Exception\DriverException: An exception occurred in the driver: SQLSTATE[HY000]: General error: 4068 OCIStmtFetch: ORA-04068: existing state of packages has been discarded
ORA-04061: existing state of function "ROOT.TEST_ORACLE_FETCH_FAILURE" has been invalidated
ORA-04065: not executed, altered or dropped function "ROOT.TEST_ORACLE_FETCH_FAILURE"
 (/ext/pdo_oci/oci_statement.c:489)

...

Caused by
Doctrine\DBAL\Driver\PDO\Exception: SQLSTATE[HY000]: General error: 4068 OCIStmtFetch: ORA-04068: existing state of packages has been discarded
ORA-04061: existing state of function "ROOT.TEST_ORACLE_FETCH_FAILURE" has been invalidated
ORA-04065: not executed, altered or dropped function "ROOT.TEST_ORACLE_FETCH_FAILURE"
 (/ext/pdo_oci/oci_statement.c:489)

```

tests/Functional/Driver/OCI8/ResultTest.php
amenning: I was able to switch to use the privileged user as the second connection

tests/Functional/Driver/OCI8/ResultTest.php
```diff
@@ -0,0 +1,177 @@
...
+                $expectedTotalRowCount,
+                $result->rowCount()
+            )
+        );
+    }
+
+    public function dataProviderForTestTruncatedFetch(): Generator
+    {
+        yield 'it should return all rows if no data invalidation occurs'
+            => ['invalidateDataMidFetch' => false];
```
morozov: The argument name may improve readability in the short term but it's not enforced at the testing framework level. I'd rather get rid of it and left the parameters as `[false]`.

tests/Functional/Driver/OCI8/ResultTest.php
morozov: Thanks for the research. Let's leave the test as is for now. We can update it later.

tests/Functional/Driver/OCI8/ResultTest.php
```diff
@@ -0,0 +1,177 @@
...
+            return;
+        }
+
+        self::markTestSkipped('oci8 only test.');
+    }
+
+    protected function tearDown(): void
+    {
+        $this->connection->executeQuery(sprintf(
+            'DROP FUNCTION %s.test_oracle_fetch_failure',
```
morozov: Is a fully qualified function necessary here? As far as I understand, it belongs to the same user as the connection. A fully qualified name is only necessary when a user's function is referenced from the privileged connection.

src/Driver/OCI8/Result.php
```diff
@@ -136,4 +158,47 @@ private function fetchAll(int $mode, int $fetchStructure): array
...
         return $result;
     }
+
+    /**
+     * Screen PHP error messages for any error that the OCI8 extension
+     * doesn't report via oci_error()
+     *
+     * @link http://www.dba-oracle.com/t_error_code_list.htm
+     */
+    public static function handleError(
```
morozov: Please move the implementation under the callback set by `set_error_handler()`. There's no need to keep a dedicated public function.

src/Driver/OCI8/Result.php
```diff
@@ -136,4 +158,47 @@ private function fetchAll(int $mode, int $fetchStructure): array
...
+                case '04065':
+                case '04068':
+                    throw PHPError::new(
+                        $preambleMessage,
+                        $errno,
+                        $errstr,
+                        $errfile,
+                        $errline
+                    );
+                default:
```
morozov: This block could be omitted. It doesn't do anything.

ci/github/phpunit/.phpunit.result.cache
```diff
@@ -0,0 +1 @@
+C:37:"PHPUnit\Runner\DefaultTestResultCache":379:{a:2:{s:7:"defects";a:0:{}s:5:"times";a:2:{s:146:"Doctrine\DBAL\Tests\Functional\Driver\OCI8\ResultTest::testTruncatedFetch with data set "it should return all rows if no data invalidation occurs"";d:0.891;s:155:"Doctrine\DBAL\Tests\Functional\Driver\OCI8\ResultTest::testTruncatedFetch with data set "it should convert oci8 data invalidation error to DriverException"";d:0.074;}}}
```
morozov: This isn't meant to be committed.

morozov (commented):
Looks good. @amenning please remove the cache file and squash commits.

amenning:
I removed that file from the commit and squashed them into one commit.  All checks passed.

morozov (approved):

morozov:
Thanks, @amenning!

amenning:
Thank you for quick reviews and excellent guidance.

PR #4582, opened 2021-04-02
Fix usage of deprecated fetch method

acrobat:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

When upgrading a library to 2.13 and trying to fix all the deprecations, this always came up in the results. I've replaced the deprecated fetch method so the internal code doesn't trigger any "self deprecations"


beberlei:
There are hundreds of self calls to currently deprecated APIs in 2.*, its something that is fixed in 3.0 when the code is cleaned up, but not really possible and sensible in 2.x already.

See https://github.com/doctrine/dbal/blob/3.0.x/src/Schema/SqliteSchemaManager.php#L523 - this is already changed in the released 3.0

PR #4585, opened 2021-04-03
Fix MSSQL getXxxExtendedPropertySQL - fix #4283 issue

### LINKED ISSUE #4283: Doctrine fails to declare Extended Property on SQL-Server array column (when table name is quoted)

mvorisek:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #4283

#### Summary

Fix escaping for `EXEC sp_addextendedproperty` SQL function which accept column name as value instead of identifier.

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -15,6 +15,7 @@ use InvalidArgumentException;
 use function array_merge;
 use function array_unique;
 use function array_values;
+use function assert;
 use function count;
 use function crc32;
 use function dechex;
@@ -790,7 +791,10 @@ class SQLServerPlatform extends AbstractPlatform
 
     private function quoteSingleIdentifierAsStringLiteral(string $levelName): string
     {
-        return $this->quoteStringLiteral(preg_replace('~^\[|\]$~s', '', $levelName));
+        $levelName = preg_replace('~^\[|\]$~s', '', $levelName);
+        assert(is_string($levelName));
+
+        return $this->quoteStringLiteral($levelName);
     }
 
     /**
--- a/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
+++ b/lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
@@ -26,6 +26,7 @@ use function is_bool;
 use function is_numeric;
 use function is_string;
 use function preg_match;
+use function preg_replace;
 use function sprintf;
 use function str_replace;
 use function stripos;
@@ -787,17 +788,22 @@ class SQLServerPlatform extends AbstractPlatform
         ];
     }
 
+    private function quoteSingleIdentifierAsStringLiteral(string $levelName): string
+    {
+        return $this->quoteStringLiteral(preg_replace('~^\[|\]$~s', '', $levelName));
+    }
+
     /**
      * Returns the SQL statement for adding an extended property to a database object.
      *
      * @link http://msdn.microsoft.com/en-us/library/ms180047%28v=sql.90%29.aspx
      *
      * @param string      $name       The name of the property to add.
-     * @param string|null $value      The value of the property to add.
-     * @param string|null $level0Type The type of the object at level 0 the property belongs to.
-     * @param string|null $level0Name The name of the object at level 0 the property belongs to.
-     * @param string|null $level1Type The type of the object at level 1 the property belongs to.
-     * @param string|null $level1Name The name of the object at level 1 the property belongs to.
+     * @param string      $value      The value of the property to add.
+     * @param string      $level0Type The type of the object at level 0 the property belongs to.
+     * @param string      $level0Name The name of the object at level 0 the property belongs to.
+     * @param string      $level1Type The type of the object at level 1 the property belongs to.
+     * @param string      $level1Name The name of the object at level 1 the property belongs to.
      * @param string|null $level2Type The type of the object at level 2 the property belongs to.
      * @param string|null $level2Name The name of the object at level 2 the property belongs to.
      *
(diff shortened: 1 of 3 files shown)


aimeos:
@mvorisek Any progress on this issue?

mvorisek:
> @mvorisek Any progress on this issue?

PR is RTM from my side

aimeos:
@mvorisek Is there a PR for 3.1.x too?

greg0ire:
@aimeos no need: 2.13.x is regularly merged up in 3.1.x

aimeos:
@morozov @greg0ire  Any chance to merge that PR soon?

greg0ire:
I started reviewing it, then gave up… please rework your commits: one of them reverts what another did, another adds tests, when really there should be one commit with both the tests and the code. Another one mixes questionable style changes with what this PR is all about, making it unnecessarily hard to review it. There is also a merge commit.

IMO there should be 2 commits, maybe 3 but not 6: one for the style changes (dots at the end of line for multiline concatenation, which for some reason not stated, you did not like), and another for the actual fix.

aimeos:
@greg0ire https://github.com/doctrine/dbal/pull/4585/files only shows the real changes that are applied at the end

greg0ire:
@aimeos don't you think I know that?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -360,21 +362,18 @@ public function getCreatePrimaryKeySQL(Index $index, $table)
     protected function getCreateColumnCommentSQL($tableName, $columnName, $comment)
     {
         if (strpos($tableName, '.') !== false) {
-            [$schemaSQL, $tableSQL] = explode('.', $tableName);
-            $schemaSQL              = $this->quoteStringLiteral($schemaSQL);
-            $tableSQL               = $this->quoteStringLiteral($tableSQL);
+            [$schemaName, $tableName] = explode('.', $tableName, 2);
```
morozov, line 365: This implies that a schema name cannot have a dot.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -360,21 +362,18 @@ public function getCreatePrimaryKeySQL(Index $index, $table)
...
         } else {
-            $schemaSQL = "'dbo'";
-            $tableSQL  = $this->quoteStringLiteral($tableName);
+            $schemaName = $this->getDefaultSchemaName();
         }
 
         return $this->getAddExtendedPropertySQL(
             'MS_Description',
-            $comment,
+            (string) $comment,
```
morozov, line 372: Why is this necessary?

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
```diff
@@ -796,6 +789,14 @@ protected function getRenameIndexSQL($oldIndexName, Index $index, $tableName)
         ];
     }
 
+    private function quoteSingleIdentifierAsStringLiteral(string $levelName): string
+    {
+        $levelName = preg_replace('~^\[|\]$~s', '', $levelName);
```
morozov, line 794: Why would one need to pass an explicitly quoted identifier when using SQL Server? Unlike Oracle, quoting doesn't affect the semantics of the identifier.

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL4283Test.php
```diff
@@ -0,0 +1,115 @@
...
+     * @dataProvider columnNameProvider
+     */
+    public function testColumnCommnentOperations(string $columnName): void
+    {
+        if ($this->connection->getDatabasePlatform() instanceof OraclePlatform) {
+            if ($columnName === 'basic') {
+                // Oracle platform does not support other than UC column names
+                $columnName = 'BASIC';
+            } else {
+                $this->markTestSkipped('Oracle platform does not support quoted column names');
```
morozov, line 46: What exactly does this mean?

tests/Doctrine/Tests/DBAL/Functional/Ticket/DBAL4283Test.php
```diff
@@ -0,0 +1,115 @@
...
+            }
+
+            // Oracle platform does not support other than UC table names
+            $tableName = 'DBAL4283TBL';
+        } elseif ($this->connection->getDatabasePlatform() instanceof PostgreSqlPlatform) {
+            // PostgreSQL platform does not support quoted table names
```
morozov, line 52: What does this mean?

tests/Doctrine/Tests/DBAL/Platforms/AbstractSQLServerPlatformTestCase.php
```diff
@@ -656,7 +656,7 @@ public function testCreateTableWithSchemaColumnComments(): void
         $expectedSql = [
             'CREATE TABLE testschema.test (id INT NOT NULL, PRIMARY KEY (id))',
             "EXEC sp_addextendedproperty N'MS_Description', N'This is a comment', "
-                . "N'SCHEMA', 'testschema', N'TABLE', 'test', N'COLUMN', id",
+                . "N'SCHEMA', 'testschema', N'TABLE', 'test', N'COLUMN', 'id'",
```
morozov, line 659: The changes from here below seem reasonable.

lib/Doctrine/DBAL/Platforms/SQLServerPlatform.php
mvorisek: This is how it was, slight improvement when there are more dots, the remaining parts are not silently discarded.

mvorisek:
non-stale

PR #4592, opened 2021-04-09
Merge 3.1.x into 4.0.x

morozov:


morozov:
@greg0ire please look at the Coding Standards failure. Without the lock file, even when the direct dependencies have very strict constraints, the indirect ones (PHP_CodeSniffer) can have breaking changes from the consumer PoV within a minor version range.

For instance, in https://github.com/squizlabs/PHP_CodeSniffer/commit/0aeb095248f307cc7bac7b0268ae4d5d3440f300 an error code that we suppress was changed. We either need to add a dev-dependency on versions `< 3.6.0`, or add a dev-dependency on versions `>= 3.6.0` and rework the suppression, or suppress both the old and the new error, but there's no elegant solution.

greg0ire:
Maybe it should be a direct dependency? After all, we are referencing it directly a lot in https://github.com/doctrine/dbal/blob/4.0.x/phpcs.xml.dist

The same principle that applies to php and can be automated with https://packagist.org/packages/maglnet/composer-require-checker should also apply to phpcs.xml.dist IMO. Let's fix this by adding that dependency as a dev requirement and pin it then?

greg0ire (approved):

PR #4593, opened 2021-04-09
Add direct dependency on PHP_CodeSniffer

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

See the build failure of https://github.com/doctrine/dbal/pull/4592.

PHP_CodeSniffer 3.6.0 introduces changes of some error codes suppressed in the codebase (e.g. https://github.com/squizlabs/PHP_CodeSniffer/commit/0aeb095248f307cc7bac7b0268ae4d5d3440f300).

Since it's an indirect dependency currently restricted to `^3.5.6` by `slevomat/coding-standard`, without restricting it directly, the repository would have to maintain compatibility with both 3.5 and 3.6. The solution is to restrict it temporarily until it's restricted by a direct dependency.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,6 +43,7 @@
         "jetbrains/phpstorm-stubs": "2020.2",
         "phpstan/phpstan": "0.12.81",
         "phpunit/phpunit": "^7.5.20|^8.5|9.5.0",
+        "squizlabs/php_codesniffer": "3.6.0",
         "symfony/console": "^2.0.5|^3.0|^4.0|^5.0",
         "vimeo/psalm": "4.6.4"
     },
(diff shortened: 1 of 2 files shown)


greg0ire (dismissed):

greg0ire:
Approving for the sake of fixing the build, but I really think referencing a rule in a file is enough motive to consider this a direct dependency, see https://github.com/doctrine/dbal/pull/4592#issuecomment-816473922

morozov:
> I really think referencing a rule in a file is enough motive to consider this a direct dependency

Valid point. Please see the update.

greg0ire (approved):

Issue #4591, opened 2021-04-08
2.13 `QueryBuilder::execute()` is no longer backwards compatible.


m50:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.13

#### Summary

The `QueryBuilder::execute()` function now returns `Doctrine\DBAL\ForwardCompatibility\Result` or `int`, which is no longer compatible with the previous `Doctrine\DBAL\Driver\Statement`, which causes anything requiring a `Doctrine\DBAL\Driver\Statement` to be passed in to break. This is not in the upgrade file, and it is not consistent in the API. Because of this, this broke our ORM built on top of DBAL with a minor version bump (simple composer update).

#### Previous behaviour

`QueryBuilder::execute()` returned `Doctrine\DBAL\Driver\Statement`.

#### Current behavior

`QueryBuilder::execute()` returns `Doctrine\DBAL\ForwardCompatibility\Result` or `int` which cannot be used in places where Statement is needed.

#### How to reproduce

Just require a `\Doctrine\DBAL\Driver\Statement` be passed into a function, and call `execute()` on a query builder instance, and pass it into the function. In 2.12.*, that will work, in 2.13.*, that is broken.

mdumoulin:
Right. We were mislead by documentation where breaking change was consecutively introduced :

Here for Doctrine\DBAL\Connection : https://github.com/doctrine/dbal/commit/df4343721d9981e58afd15eeaf18c885a2399766#diff-425f50e688c77962b97e5fd1f9803a6d9ffa9bf260ba12391f73b2aab14914eeL906

Here for Doctrine\DBAL\Query\QueryBuilder : https://github.com/doctrine/dbal/commit/68f0d99fcf270525b3df59294f277a6643facbf9#diff-829b37fd7ed65c7ead12002f3e689f3a672986dd88fd0f5d4d5f92321a61211dL201

I would say that following method should return an object implementing Doctrine\DBAL\Driver\Statement & Doctrine\DBAL\Result:
- Doctrine\DBAL\Connection::execute()
- ~~Doctrine\DBAL\Connection::executeCacheQuery()~~
- Doctrine\DBAL\Query\QueryBuilder::execute()

It can be achieved by making Doctrine\DBAL\ForwardCompatibility\Result implementing and wrap Doctrine\DBAL\Driver\Statement instead of Doctrine\DBAL\Driver\ResultStatement

Edit : 
[...]

PR #4596, opened 2021-04-09
Fix BC break on QueryBuilder::execute()

mdumoulin:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #4591

### Summary

This PR fix a BC break on return type for :
- Doctrine\DBAL\Connection::execute()
- Doctrine\DBAL\Query\QueryBuilder::execute()

Previously, this methods expected to return a Doctrine\DBAL\Driver\Statement object as result.

#### Situation in 2.12.x
- Connection::executeCacheQuery() return a Driver\ResultStatement
- Connection::executeQuery() return either a Driver\ResultStatement if cache is used, or a Driver\Statement if query is executed
- QueryBuilder::execute() return either a Driver\Statement when the query is a select, or an integer for any other statement

#### Situation in 2.13.0 :
- Connection::executeCacheQuery() return a Driver\ResultStatement & DBAL\Result 
- Connection::executeQuery() return a Driver\ResultStatement & DBAL\Result 
- QueryBuilder::execute() return a Driver\ResultStatement & DBAL\Result

#### Situation in this PR

Create 2 new interfaces ensuring forward compatibility :
- ForwardCompatibility\DriverResultStatement extends Driver\ResultStatement and DBAL\ResultStatement
- ForwardCompatibility\DriverStatement extends Driver\Statement and DBAL\Statement

Query execution result :
- Connection::executeCacheQuery() return a ForwardCompatibility\Result wrapping Driver\ResultStatement
- Connection::executeQuery() return a ForwardCompatibility\Result wrapping Driver\ResultStatement if cache is used, or a ForwardCompatibility\Result wrapping Driver\Statement if query is executed
- QueryBuilder::execute() return either a a ForwardCompatibility\Result wrapping Driver\Statement when the query is a select, or an integer for any other statement

ForwardCompatibility/Result now implements Driver\Statement interface. If a Driver\Statement object is provided to constructor, Driver\Statement features will be availables, if a Driver\ResultStatement object is provided an exception will be thrown when trying to use Driver\Statement features.

### DIFF
--- a/lib/Doctrine/DBAL/ForwardCompatibility/Result.php
+++ b/lib/Doctrine/DBAL/ForwardCompatibility/Result.php
@@ -2,10 +2,10 @@
 
 namespace Doctrine\DBAL\ForwardCompatibility;
 
-use Doctrine\DBAL\Driver\ResultStatement as DriverResultStatement;
+use Doctrine\DBAL\Driver;
 use Doctrine\DBAL\Exception;
 use Doctrine\DBAL\Exception\NoKeyValue;
-use Doctrine\DBAL\Result as BaseResult;
+use Doctrine\DBAL\ParameterType;
 use Doctrine\Deprecations\Deprecation;
 use IteratorAggregate;
 use PDO;
@@ -18,18 +18,27 @@ use function method_exists;
  * A wrapper around a Doctrine\DBAL\Driver\ResultStatement that adds 3.0 features
  * defined in Result interface
  */
-class Result implements IteratorAggregate, DriverResultStatement, BaseResult
+class Result implements IteratorAggregate, DriverStatement, DriverResultStatement
 {
-    /** @var DriverResultStatement */
+    /** @var Driver\ResultStatement */
     private $stmt;
 
-    public function __construct(DriverResultStatement $stmt)
+    public static function ensure(Driver\ResultStatement $stmt): Result
+    {
+        if ($stmt instanceof Result) {
+            return $stmt;
+        }
+
+        return new Result($stmt);
+    }
+
+    public function __construct(Driver\ResultStatement $stmt)
     {
         $this->stmt = $stmt;
     }
 
     /**
-     * @return DriverResultStatement
+     * @return Driver\ResultStatement
      */
     public function getIterator()
     {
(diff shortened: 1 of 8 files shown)


beberlei:
Its amazing that we can actually make this work, like a frankenstein of interfaces coming together for forward compatibility, thank you so much @mdumoulin for this amazing work :)

beberlei:
@mdumoulin what needs to be done for the remaining test failures?

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
```diff
@@ -356,7 +357,36 @@ public function testUserProvidedPDOConnection(): void
...
+    {
+        $result = $this->connection->executeQuery('SELECT 1');
+
+        self::assertInstanceOf(ForwardCompatibility\Result::class, $result);
+    }
+
+    public function testResultCompatibilityWhenExecutingQueryWithParams(): void
+    {
+        $result = $this->connection->executeQuery(
+            'SELECT 1, :param',
```
morozov: You'll need to use something like this to make your tests portable:

https://github.com/doctrine/dbal/blob/4297f5027011027529fdccfa3b1fbbb3090a7d25/tests/Doctrine/Tests/DBAL/Functional/ResultCacheTest.php#L266-L267

mdumoulin: Thanks ! I couldn't figure out how to make it work with a param, but It work with the basic query. Better than nothing I guess.

mdumoulin:
> Its amazing that we can actually make this work, like a frankenstein of interfaces coming together for forward compatibility, thank you so much @mdumoulin for this amazing work :)

Thanks, I'm glad if it can help to migrate on newer version of library.

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
morozov: It's the same: `getDummySelectSQL('1, :param')`.

mdumoulin: I did try this version, but I still got an error on postgres.

lib/Doctrine/DBAL/ForwardCompatibility/Result.php
```diff
@@ -18,7 +20,7 @@
  * A wrapper around a Doctrine\DBAL\Driver\ResultStatement that adds 3.0 features
  * defined in Result interface
  */
-class Result implements IteratorAggregate, DriverResultStatement, BaseResult
+final class Result implements IteratorAggregate, DriverStatement, BaseResult
```
morozov: Making it final would be a breaking change.

lib/Doctrine/DBAL/ForwardCompatibility/Result.php
```diff
@@ -301,4 +303,104 @@ private function ensureHasKeyValue(): void
...
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/4019',
+            'Result::bindValue() is deprecated, no replacement.'
+        );
+
+        if ($this->stmt instanceof DriverStatement) {
+            return $this->stmt->bindValue($param, $value, $type);
+        }
+
+        throw Exception::notSupported('bindValue');
```
morozov, line 331: I'm getting a bit lost: the forward-compatible result implements the statement interface and can act as a statement depending on the wrapped implementation. @beberlei could you review this please?

mdumoulin: In 2.12.x version :
- Doctrine\DBAL\Connection::executeCacheQuery() returned a Driver\ResultStatement object
- Doctrine\DBAL\Connection::executeQuery() returned a Driver\Statement object, except when using cache where a Driver\ResultStatement is returned instead

The solution makes both return a result implementing Driver\Statement, but Statement feature will only be available when using Doctrine\DBAL\Connection::executeQuery().

We can be more specific in methods @return documentation by adding following interfaces :
- ForwardCompatibility\DriverResultStatement implementing Driver\ResultStatement&DBAL\Result for executeCacheQuery()
- ForwardCompatibility\DriverStatement implementing Driver\Statement&DBAL\Result for executeQuery()

morozov: Thank you for the explanation. Makes sense. I think it's enough to leave the return type documented via an annotation with the intersection type, no need to introduce more interfaces.

mdumoulin: I can't find a way to fix documentation in QueryBuilder without adding an interface : https://github.com/doctrine/dbal/pull/4596/files#diff-829b37fd7ed65c7ead12002f3e689f3a672986dd88fd0f5d4d5f92321a61211dL204

Without interface, we will have to write something like :
```php
    /**
     * Executes this query using the bound parameters and their types.
     *
     * @return (ResultStatement&Doctrine\DBAL\Result)|int
     *
     * @throws Exception
     */
```

I test it with phpstan but it does not work.

mdumoulin: Also, I think we should to be more accurate on Doctrine\DBAL\Connection::executeQuery() with something like :

```php
    /**
     * ...
     * @return (Driver\ResultStatement&Doctrine\DBAL\Result)|(Driver\Statement&Doctrine\DBAL\Result) Return the executed result statement or cached statement if a cache profile is used
     *
     * @throws Exception
     */
```
Same issue here, I don't think phpstan can interpret complex type like that.

morozov: I see. Then the interfaces are necessary. The question is, is there a way to test the newly introduced types? Currently, the build passes but there are changes needed to satisfy the static analysis on the consumer side. That means that we're not analyzing all the cases ourselves. E.g. it might help to enable phpstan for some of the test cases that describe the consumers of the old API and make sure that such analysis passes.

mdumoulin: That's the meaning of tests added here : https://github.com/doctrine/dbal/pull/4596/files#diff-1056c8f8b7e1dbbb024efeed5bca59a3e9097603fafb63d0c7c7924531b2d63eR363-R381

But I still have to find a way to test executeQuery() with params.

lib/Doctrine/DBAL/Query/QueryBuilder.php
```diff
@@ -201,14 +201,16 @@ public function getState()
...
      *
      * @throws Exception
      */
     public function execute()
     {
         if ($this->type === self::SELECT) {
-            return $this->connection->executeQuery($this->getSQL(), $this->params, $this->paramTypes);
+            return ForwardCompatibility\Result::ensure(
+                $this->connection->executeQuery($this->getSQL(), $this->params, $this->paramTypes)
+            );
```
mdumoulin, line 213: Ensuring compatible result here has 2 advantages :
- We can be more accurate on return type documentation, otherwise we will have to document ForwardCompatibility\DriverStatement|ForwardCompatibility\DriverResultStatement|int
- If user has overridden Connection::executeQuery() the forward compatibility will be ensured

lib/Doctrine/DBAL/ForwardCompatibility/Result.php
beberlei: I think this is correct, it exposes the `bind*` api from the result, but since it is still acting as a statement in DBAL 2 this is correct and the deprecations trigger correctly. I believe we can roll with this.

beberlei (dismissed):
Good work!

SenseException (dismissed):

morozov (approved):

morozov:
Thanks, @mdumoulin!

PR #4598, opened 2021-04-10
[GH-4503] Make OracleSchemaManager::dropAutoincrement() protected

### LINKED ISSUE #4503: Inconsistent SchemaManager API across platforms

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

See #4503.

### DIFF
--- a/src/Schema/OracleSchemaManager.php
+++ b/src/Schema/OracleSchemaManager.php
@@ -265,11 +265,9 @@ class OracleSchemaManager extends AbstractSchemaManager
     }
 
     /**
-     * @internal The method should be only used from within the OracleSchemaManager class hierarchy.
-     *
      * @throws Exception
      */
-    public function dropAutoincrement(string $table): bool
+    protected function dropAutoincrement(string $table): bool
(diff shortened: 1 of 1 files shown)


src/Schema/OracleSchemaManager.php
```diff
@@ -265,11 +265,9 @@ public function createDatabase(string $database): void
     }
 
     /**
-     * @internal The method should be only used from within the OracleSchemaManager class hierarchy.
-     *
      * @throws Exception
      */
-    public function dropAutoincrement(string $table): bool
+    protected function dropAutoincrement(string $table): bool
```
greg0ire, line 270: Should it also be made final or is it meant to be overridden ?

morozov: I’d leave it as is for now. Making it final is out of scope, and there’s no problem if someone has it overridden. The issue is about public methods that require runtime assertions.

greg0ire (approved):

PR #4599, opened 2021-04-10
[GH-4553] Simplify error handling in the OCI8 driver

### LINKED ISSUE #4553: DBAL FETCH silently returns truncated results when database emits an error after initial prefetch

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

See #4553.

While reviewing #4581, I had an impression that the `oci8` extension only reports fetching errors via the PHP error mechanism and doesn't set the error status available via `oci_error()`. I did some debugging of the extension and it turned out the error handling works both ways: via the PHP errors and the error status on the statement.

This patch simplifies the original approach:
1. No longer need to set and reset the error handler.
2. No longer need to parse error messages.
3. No additional exception class.
4. I didn't check that yet but the resulting exception message should be closer to the one that PDO would produce.


morozov:
@amenning I did some digging into the extension and simplified your patch a bit. Could you take a look?

amenning (approved):
This is a great improvement.  Code looks good and I see no issues.  This covers the original example and will be far better for properly catching and reporting a wider set of related errors.

greg0ire (approved):
Nice simplification !

PR #4600, opened 2021-04-12
BUGFIX: sqlite has no knowledge about longblob

JeroenED:
|      Q       |   A
|------------- | -----------
| Type         | **bug**
| BC Break     | **no**
| Fixed issues | (none)

#### Summary

This fixes a bug where webapps, who are trying to create a longblob field in sqlite, are going wrong. The bugfix is based in the comment by @Neziak on a similar issue with MySQL in the Nextcloud server repository (https://github.com/nextcloud/server/issues/25748#issuecomment-784736577).
The bugfix has successfully been tested on my personal nextcloud instance.
I decided to not create a issue on github as why bother people if typing this message takes 10x more time than fixing the actual bug?

### DIFF
--- a/src/Platforms/SqlitePlatform.php
+++ b/src/Platforms/SqlitePlatform.php
@@ -567,6 +567,7 @@ class SqlitePlatform extends AbstractPlatform
             'image'            => 'string',
             'int'              => 'integer',
             'integer'          => 'integer',
+            'longblob'         => 'blob',
             'longtext'         => 'text',
             'longvarchar'      => 'string',
             'mediumint'        => 'integer',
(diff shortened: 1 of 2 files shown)


morozov:
> The bugfix has successfully been tested on my personal nextcloud instance.

@JeroenED please add a functional test to the test suite as well. You don't necessarily need to create a new test case, maybe you could extend an existing one that involves a BLOB column.

JeroenED:
Done. But it appears that some unrelated tests are failing.

morozov:
The build is green now but the test isn't functional and doesn't seem to cover the code change.

morozov:

PR #4620, opened 2021-04-27
Use `psr/cache` for result caching

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | N/A

#### Summary

The Doctrine Cache library has served us well, but it's currently being phased out. This PR proposes to rely on PSR-6 for caches. Symfony Cache is used for PHPUnit tests, but PSR-6 should allow us to plug in any application's favorite caching solution easily.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -6,6 +6,21 @@ awareness about deprecated code.
 - Use of our low-overhead runtime deprecation API, details:
   https://github.com/doctrine/deprecations/
 
+# Upgrade to 3.2
+
+## Introduction of PSR-6 for result caching
+
+Instead of relying on the deprecated `doctrine/cache` library, a PSR-6 cache
+can now be used for result caching. Please use the following new methods for
+this purpose:
+
+| class               | old method               | new method         |
+| ------------------- | ------------------------ | ------------------ |
+| `Configuration`     | `setResultCacheImpl()`   | `setResultCache()` |
+| `Configuration`     | `getResultCacheImpl()`   | `getResultCache()` |
+| `QueryCacheProfile` | `setResultCacheDriver()` | `setResultCache()` |
+| `QueryCacheProfile` | `getResultCacheDriver()` | `getResultCache()` |
+
 # Upgrade to 3.1
 
 ## Deprecated schema- and namespace-related methods
--- a/composer.json
+++ b/composer.json
@@ -33,9 +33,10 @@
     "require": {
         "php": "^7.3 || ^8.0",
         "composer/package-versions-deprecated": "^1.11.99",
-        "doctrine/cache": "^1.0",
+        "doctrine/cache": "^1.11",
         "doctrine/deprecations": "^0.5.3",
-        "doctrine/event-manager": "^1.0"
+        "doctrine/event-manager": "^1.0",
+        "psr/cache": "^1|^2|^3"
     },
     "require-dev": {
         "doctrine/coding-standard": "8.2.0",
@@ -45,6 +46,7 @@
         "phpunit/phpunit": "9.5.0",
         "psalm/plugin-phpunit": "0.13.0",
         "squizlabs/php_codesniffer": "3.6.0",
+        "symfony/cache": "^5.2",
         "symfony/console": "^2.0.5|^3.0|^4.0|^5.0",
         "vimeo/psalm": "4.6.4"
     },
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -106,5 +106,11 @@ parameters:
             message: '~^Instanceof between Doctrine\\DBAL\\Platforms\\Keywords\\KeywordList and Doctrine\\DBAL\\Platforms\\Keywords\\KeywordList will always evaluate to true\.~'
             paths:
                 - %currentWorkingDirectory%/src/Platforms/AbstractPlatform.php
+
+        # We're checking for invalid invalid input
+        -
+            message: "#^Strict comparison using \\!\\=\\= between null and null will always evaluate to false\\.$#"
+            count: 1
+            path: src/Cache/QueryCacheProfile.php
 includes:
     - vendor/phpstan/phpstan-strict-rules/rules.neon
--- a/src/Cache/QueryCacheProfile.php
+++ b/src/Cache/QueryCacheProfile.php
@@ -3,11 +3,17 @@
 namespace Doctrine\DBAL\Cache;
 
 use Doctrine\Common\Cache\Cache;
+use Doctrine\Common\Cache\Psr6\CacheAdapter;
+use Doctrine\Common\Cache\Psr6\DoctrineProvider;
 use Doctrine\DBAL\Types\Type;
+use Psr\Cache\CacheItemPoolInterface;
+use TypeError;
 
+use function get_class;
 use function hash;
 use function serialize;
 use function sha1;
+use function sprintf;
 
 /**
  * Query Cache Profile handles the data relevant for query caching.
@@ -16,8 +22,8 @@ use function sha1;
  */
 class QueryCacheProfile
 {
-    /** @var Cache|null */
-    private $resultCacheDriver;
+    /** @var CacheItemPoolInterface|null */
+    private $resultCache;
 
     /** @var int */
     private $lifetime = 0;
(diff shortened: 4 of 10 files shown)


src/Cache/QueryCacheProfile.php
```diff
@@ -26,22 +33,56 @@ class QueryCacheProfile
...
+    {
+        $this->lifetime = $lifetime;
+        $this->cacheKey = $cacheKey;
+        if ($resultCache instanceof CacheItemPoolInterface) {
+            $this->resultCache = $resultCache;
+        } elseif ($resultCache instanceof Cache) {
+            Deprecation::trigger(
+                'doctrine/dbal',
+                'https://github.com/doctrine/dbal/pull/4620',
+                'Passing an instance of %s as $resultCache is deprecated. Pass an instance of %s instead.',
```
greg0ire: Should `__METHOD__` be mentioned at some point?

src/Configuration.php
```diff
@@ -24,6 +28,15 @@ class Configuration
     /**
      * The cache driver implementation that is used for query result caching.
      *
+     * @var CacheItemPoolInterface|null
+     */
+    protected $resultCache;
```
greg0ire:
```suggestion
    private $resultCache;
```

greg0ire (commented):
Please document the deprecations in `UPGRADE.md`

src/Configuration.php
derrabus: I tried mimic the style of this class here, but it probably makes sense to make this property private.

greg0ire: Moreover, `$middlewares` is already private, so one could argue there is no style to mimic :)

derrabus:
I've addressed your comments. No idea why the OCI8 jobs are failing suddenly, but the failures look unrelated.

greg0ire:
Might be related to https://twitter.com/meShivamMathur/status/1386942165272567815

greg0ire:
Reported at https://github.com/shivammathur/setup-php/issues/449

composer.json
```diff
@@ -44,7 +44,9 @@
         "phpstan/phpstan-strict-rules": "^0.12.2",
         "phpunit/phpunit": "9.5.0",
         "psalm/plugin-phpunit": "0.13.0",
+        "psr/cache": "^1.0 || ^2.0 || ^3.0",
         "squizlabs/php_codesniffer": "3.6.0",
+        "symfony/cache": "^5.2",
```
morozov, line 49: Why is a dependency on a specific cache implementation needed? If it's for testing, should it be made a development dependency?

phpstan.neon.dist
```diff
@@ -106,5 +106,11 @@ parameters:
             message: '~^Instanceof between Doctrine\\DBAL\\Platforms\\Keywords\\KeywordList and Doctrine\\DBAL\\Platforms\\Keywords\\KeywordList will always evaluate to true\.~'
             paths:
                 - %currentWorkingDirectory%/src/Platforms/AbstractPlatform.php
+
+        # We're checking for invalid invalid input
```
morozov, line 110: It looks like this error shouldn't be manually suppressed given that we disable [`treatPhpDocTypesAsCertain`](https://phpstan.org/config-reference#treatphpdoctypesascertain): https://github.com/doctrine/dbal/blob/256da038ef3e5bb33a35e93520680661ef542f01/phpstan.neon.dist#L5

Is it really necessary?

composer.json
greg0ire: @morozov it is a development dependency, and now I'm wondering why `psr/cache` is one too :thinking:

alcaeus:
Given that ORM 2.9 won't support DBAL 3.x, are there any chances of backporting this into 2.x?

composer.json
derrabus: I can remove `psr/cache` indeed. But we need an implementation to test against, so either `symfony/cache` or any other PSR-6 implementation you want should remain in `require-dev`.

phpstan.neon.dist
derrabus: Yes PHPStan complained about it, which is why I added this entry. IIRC, PHPStan would also complain, if that ignore rule didn't match any errors. If you have a better idea how I could make PHPStan happy here, I'm all ears. 🙂

derrabus:
> Given that ORM 2.9 won't support DBAL 3.x, are there any chances of backporting this into 2.x?

It probably won't be hard to do this, but do we need to? Are DBAL and ORM sharing any caches?

composer.json
derrabus: The `psr/cache` dependency is gone now.

alcaeus:
> It probably won't be hard to do this, but do we need to? Are DBAL and ORM sharing any caches?

Depending on the timeline for DBAL 3 support in ORM 2, we don't need to. If that takes a while, I'd be in favour of backporting.

alcaeus:
On a different note, with DBAL 3 supporting PHP 7.3+ but ORM 2 supporting 7.1+, there will be a number of users using ORM 2 that can't use DBAL 3. This will slow down the process of getting everybody off doctrine/cache 1.x.

derrabus:
@alcaeus Doctrine Cache 2 on DBAL 2 is easily possible without backporting these changes: #4623

UPGRADE.md
```diff
@@ -6,6 +6,16 @@ awareness about deprecated code.
...
 
+# Upgrade to 3.2
+
+The usage of Doctrine Cache is deprecated in favor of PSR-6. The following
+methods related to Doctrine Cache have been replaced with PSR-6 counterparts:
+
+- `Configuration::setResultCacheImpl()` is deprecated, use `setResultCache()` instead.
+- `Configuration::getResultCacheImpl()` is deprecated, use `getResultCache()` instead.
+- `QueryCacheProfile::setResultCacheDriver()` is deprecated, use `setResultCache()` instead.
+- `QueryCacheProfile::getResultCacheDriver()` is deprecated, use `getResultCache()` instead.
```
greg0ire: If none of these are in a hot path, it's ok to add the deprecations now in a single PR. If not, maybe let's postpone the deprecations to a subsequent minor release.

derrabus: Those deprecations are all connected. `Connection::executeCacheQuery()` needs a result cache and pulls it either from the passed `QueryCacheProfile` or (if the profile does not contain a cache implementation) from its own `Configuration` instance. It does not really make sense to switch only one of the classes to PSR-6.

composer.json
greg0ire: What I meant to say is that there should probably be a dependency on `psr/cache`, but in `require`? Or is the reasoning that the dependency is optional. Since it's mentioned in `src`, it should appear in one of these sections IMO, sorry for the misunderstanding.

derrabus: No worries, I've added it to the `require` section now.

UPGRADE.md
greg0ire: I don't understand your comment. What I was saying is we don't want to get too many unfixable deprecations. So if some or all of these can happen a lot during a single HTTP request or process, then we should probably add the new methods in a new minor, and then the deprecations in a subsequent minor. I'm discussing the same issue for `doctrine/persistence` with @alcaeus , but FYI we had that very issue during the namespace move we did in that library, and IIRC we vowed to always proceed in 2 phases as described above.

derrabus: Okay, so I would basically remove all deprecations and just add a second way of configuring a cache? Yes, I could do that.

derrabus: One question through: We could already soft-deprecate those methods by adding the `@deprecated` annotations, right? I would only remove the triggers, correct?

derrabus: Done!

phpstan.neon.dist
morozov: > IIRC, PHPStan would also complain, if that ignore rule didn't match any errors.

Not in our setup: https://github.com/doctrine/dbal/blob/aefa1a97b274265b901ee8499c53387c10867700/phpstan.neon.dist#L6

> If you have a better idea how I could make PHPStan happy here, I'm all ears.

It's not only about making it happy. Usually, the process is that each suppressed issue is annotated either with a link to an existing issue in the DBAL or an issue reported at the PHPStan bug tracker. This way, there is an understanding of how to proceed with getting rid of this suppression.

This case looks like a bug that I cannot [reproduce](https://phpstan.org/r/48ba4d9e-05fb-4c82-8f67-ea8e0954b0e1) in their playground: the error is only reproducible when "Treat PHPDoc types as certain" is checked but in our case it's not.

morozov: Given that this is a temporary code, we can leave the suppression as is. I'll try to get to the bottom of the issue asynchronously.

derrabus:
Follow-up for the deprecation warnings: #4624

greg0ire:
> One question through: We could already soft-deprecate those methods by adding the @deprecated annotations, right? I would only remove the triggers, correct?

I don't have a religion on this, since it's probably only going to have an effect on the dev env. It would make sense to keep them so that people can already notice they should migrate something. @alcaeus WDYT?

alcaeus:
If we deprecate them, we should also emit deprecation notices. I'm fine with deferring the deprecation itself to a future release.

morozov:
What version is it planned to be released in? There's currently no `3.2.0` milestone but dependency changes and new deprecations don't get usually introduced in a patch. Somebody, please create a new milestone if necessary.

greg0ire:
Oh right 3.2.x does not even exist at this point, let me fix that.

greg0ire (approved):

derrabus:
> What version is it planned to be released in? There's currently no `3.2.0` milestone but dependency changes and new deprecations don't get usually introduced in a patch. Somebody, please create a new milestone if necessary.

Right. I've targeted the 3.1.x branch because there was no 3.2.x yet. This change definitely deserves a minor version bump.

morozov (approved):

greg0ire:
Thanks @derrabus !

src/Cache/QueryCacheProfile.php
```diff
@@ -26,22 +32,41 @@ class QueryCacheProfile
...
      */
-    public function __construct($lifetime = 0, $cacheKey = null, ?Cache $resultCache = null)
+    public function __construct($lifetime = 0, $cacheKey = null, ?object $resultCache = null)
+    {
+        $this->lifetime = $lifetime;
+        $this->cacheKey = $cacheKey;
+        if ($resultCache instanceof CacheItemPoolInterface) {
+            $this->resultCache = $resultCache;
+        } elseif ($resultCache instanceof Cache) {
+            $this->resultCache = CacheAdapter::wrap($resultCache);
```
greg0ire, line 46: @derrabus my team just stumbled upon this, this is a BC Break I'm afraid, if the cache key contains reserved characters.

derrabus: Can you elaborate? Or better, can you open an issue so we can discuss the BC break an possible solutions?

greg0ire: This might be fixed by using `rawurlencode` instead of throwing here: https://github.com/doctrine/cache/blob/331b4d5dbaeab3827976273e9356b3b453c300ce/lib/Doctrine/Common/Cache/Psr6/CacheAdapter.php#L265 in https://github.com/symfony/symfony/blob/0ac7aa230fe0177214415ee20fec113a44a261f1/src/Symfony/Component/Cache/DoctrineProvider.php#L60

greg0ire: @derrabus sorry, didn't notice your message. I will open an issue.

greg0ire: See https://github.com/doctrine/dbal/issues/5051

PR #4590, opened 2021-04-07
Fix user provided pdo connection

mdumoulin:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #4564 

#### Summary

When user provide a PDO connection, we get the following fatal error :

```text
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Doctrine\DBAL\Connection::ensureForwardCompatibilityStatement() must be an instance of Doctrine\DBAL\Driver\ResultStatement, instance of PDOStatement given, called in vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php on line 1313 and defined in vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php:1370
Stack trace:
#0 vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php(1313): Doctrine\DBAL\Connection->ensureForwardCompatibilityStatement(Object(PDOStatement))
#1 t.php(6): Doctrine\DBAL\Connection->executeQuery('SELECT 1;')
#2 {main}
  thrown in vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php on line 1370
```

It happen because the PDO connection is not wrapped with Driver/Connection Interface in that case. The proposed solution, wrap the PDOStatement with Driver\PDO\Statement.

#### How to reproduce

```php
<?php                                                                                                                   
require_once('vendor/autoload.php');                                                                                    
                                                                                                                        
$conn = Doctrine\DBAL\DriverManager::getConnection(['pdo' => new PDO('sqlite::memory:'), 'driver' => 'pdo_sqlite']);    
$conn->executeQuery('SELECT 1;');
```

PR #4621, opened 2021-04-27
Move fix for user provided pdo connection

### LINKED ISSUE #4619: ensureForwardCompatibilityStatement broke existing code
### Bug Report

This is likely a low priority. 

|    Q        |   A
|------------ | ------
| BC Break    | yes?
| Version     | 2.13.1, but it looks like 2.12.x branch has the issue too. 2.12.1 worked fine

#### Summary

Commit 4297f5027011027529fdccfa3b1fbbb3090a7d25 broke `Connection` when passing in `$params['pdo']`

https://github.com/doctrine/dbal/pull/4590 says it should have fixed the issue, but I'm using 2.13.1 and it still seems to be broken.

#### Current behaviour

If you build a `Connection` object using an already established `$params['pdo']` connection, commit 4297f5027011027529fdccfa3b1fbbb3090a7d25 caused breaking changes that no longer allow that behavior. Now you get the following error:

```
An exception occurred while executing 'SHOW FULL TABLES WHERE Table_type = 'BASE TABLE'':  Argument 1 passed to Doctrine\DBAL\Connection::ensureForwardCompatibilityStatement() must be an instance of Doctrine\DBAL\Driver\ResultStatement, instance of PDOStatement given, called in /var/www/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php on line 1314
```

#### How to reproduce

```php

use Doctrine\DBAL\Connection;
use Doctrine\DBAL\Driver\PDO\MySQL\Driver;

$params = [
    'pdo' => new \PDO('mysql:<whatever>'),
];
$driver = new Driver();

$connection = new Connection($params, $driver);
// This is a query that Doctrine Migrations tool does:
$connection->executeQuery("SHOW FULL TABLES WHERE Table_type = 'BASE TABLE'");
```

- Within the `executeQuery()` method, it calls `$connection = $this->getWrappedConnection();` which is returning the raw PDO connection. 
- Later in the method it calls `$stmt = $connection->prepare($sql);` which is a raw PDOStatement object.
- At the end of the method the referenced commit changed the code to `return $this->ensureForwardCompatibilityStatement($stmt);` and it expects `$stmt` to be an instance of `ResultStatement`, but of course it's the raw PDOStatement instead.

#### Expected behaviour

No error to be thrown

bizurkur:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | #4619

#### Summary

Commit 4297f5027011027529fdccfa3b1fbbb3090a7d25 broke `Connection` when passing in `$params['pdo']`

https://github.com/doctrine/dbal/pull/4590 says it should have fixed the issue, but I'm using 2.13.1 and it still seems to be broken.

I took the fix that was already applied in #4590 and moved it to the `Connection` class. This way it gets applied to all connections, not just connections built using the DriverManager.

If it's a good idea to keep the fix applied inside DriverManager, too, please let me know and I can add that back.

#### How to reproduce

```php

use Doctrine\DBAL\Connection;
use Doctrine\DBAL\Driver\PDO\MySQL\Driver;

$params = [
    'pdo' => new \PDO('mysql:<whatever>'),
];
$driver = new Driver();

$connection = new Connection($params, $driver);
// This is a query that Doctrine Migrations tool does:
$connection->executeQuery("SHOW FULL TABLES WHERE Table_type = 'BASE TABLE'");
```

-----

Also, I love Doctrine. You all are doing awesome work!

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -9,6 +9,7 @@ use Doctrine\DBAL\Cache\CacheException;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
 use Doctrine\DBAL\Cache\ResultCacheStatement;
 use Doctrine\DBAL\Driver\Connection as DriverConnection;
+use Doctrine\DBAL\Driver\PDO\Statement as PDODriverStatement;
 use Doctrine\DBAL\Driver\PingableConnection;
 use Doctrine\DBAL\Driver\ResultStatement;
 use Doctrine\DBAL\Driver\ServerInfoAwareConnection;
@@ -21,6 +22,7 @@ use Doctrine\DBAL\Query\QueryBuilder;
 use Doctrine\DBAL\Schema\AbstractSchemaManager;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\Deprecations\Deprecation;
+use PDO;
 use Throwable;
 use Traversable;
 
@@ -193,7 +195,18 @@ class Connection implements DriverConnection
         $this->params  = $params;
 
         if (isset($params['pdo'])) {
+            Deprecation::trigger(
+                'doctrine/dbal',
+                'https://github.com/doctrine/dbal/pull/3554',
+                'Passing a user provided PDO instance directly to Doctrine is deprecated.'
+            );
+
+            if (! $params['pdo'] instanceof PDO) {
+                throw Exception::invalidPdoInstance();
+            }
+
             $this->_conn = $params['pdo'];
+            $this->_conn->setAttribute(PDO::ATTR_STATEMENT_CLASS, [PDODriverStatement::class, []]);
             unset($this->params['pdo']);
         }
 
--- a/lib/Doctrine/DBAL/DriverManager.php
+++ b/lib/Doctrine/DBAL/DriverManager.php
@@ -8,10 +8,8 @@ use Doctrine\DBAL\Driver\IBMDB2;
 use Doctrine\DBAL\Driver\Mysqli;
 use Doctrine\DBAL\Driver\OCI8;
 use Doctrine\DBAL\Driver\PDO;
-use Doctrine\DBAL\Driver\PDO\Statement as PDODriverStatement;
 use Doctrine\DBAL\Driver\SQLAnywhere;
 use Doctrine\DBAL\Driver\SQLSrv;
-use Doctrine\Deprecations\Deprecation;
 
 use function array_keys;
 use function array_merge;
@@ -244,14 +242,7 @@ final class DriverManager
         }
 
         if (isset($params['pdo'])) {
-            Deprecation::trigger(
-                'doctrine/dbal',
-                'https://github.com/doctrine/dbal/pull/3554',
-                'Passing a user provided PDO instance directly to Doctrine is deprecated.'
-            );
-
             $params['pdo']->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_EXCEPTION);
-            $params['pdo']->setAttribute(\PDO::ATTR_STATEMENT_CLASS, [PDODriverStatement::class, []]);
             $params['driver'] = 'pdo_' . $params['pdo']->getAttribute(\PDO::ATTR_DRIVER_NAME);
         }
 
--- a/tests/Doctrine/Tests/DBAL/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/ConnectionTest.php
@@ -23,10 +23,9 @@ use Doctrine\DBAL\ParameterType;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Result;
 use Doctrine\Tests\DbalTestCase;
+use PDO;
 use PHPUnit\Framework\MockObject\MockObject;
 use stdClass;
-
-use function call_user_func_array;
 
 /**
  * @requires extension pdo_mysql
@@ -646,11 +645,22 @@ EOF
         self::assertSame($result, $conn->fetchAll($statement, $params, $types));
     }
 
+    public function testConnectionThrowsExceptionWhenInvalidPDOProvided(): void
+    {
+        $driverMock = $this->createMock(Driver::class);
+        $pdo        = $this->createMock(stdClass::class);
+
+        $this->expectException(Exception::class);
+
+        new Connection(['pdo' => $pdo], $driverMock);
+    }
+
     public function testConnectionDoesNotMaintainTwoReferencesToExternalPDO(): void
     {
         $driverMock = $this->createMock(Driver::class);
+        $pdo        = $this->createMock(PDO::class);
 
-        $conn = new Connection(['pdo' => new stdClass()], $driverMock);
+        $conn = new Connection(['pdo' => $pdo], $driverMock);
 
         self::assertArrayNotHasKey('pdo', $conn->getParams());
     }
@@ -658,22 +668,17 @@ EOF
     public function testPassingExternalPDOMeansConnectionIsConnected(): void
     {
         $driverMock = $this->createMock(Driver::class);
+        $pdo        = $this->createMock(PDO::class);
 
-        $conn = new Connection(['pdo' => new stdClass()], $driverMock);
+        $conn = new Connection(['pdo' => $pdo], $driverMock);
 
         self::assertTrue($conn->isConnected(), 'Connection is not connected after passing external PDO');
     }
 
     public function testCallingDeleteWithNoDeletionCriteriaResultsInInvalidArgumentException(): void
     {
-        $driver  = $this->createMock(Driver::class);
-        $pdoMock = $this->createMock(\Doctrine\DBAL\Driver\Connection::class);
-
-        // should never execute queries with invalid arguments
-        $pdoMock->expects($this->never())->method('exec');
-        $pdoMock->expects($this->never())->method('prepare');
-
-        $conn = new Connection(['pdo' => $pdoMock], $driver);
+        $driver = $this->createMock(Driver::class);
+        $conn   = new Connection([], $driver);
 
         $this->expectException(InvalidArgumentException::class);
         $conn->delete('kittens', []);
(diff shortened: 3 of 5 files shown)


bizurkur:
Oh, and I opened this PR against 2.12.x because it looked like the code had the same issue in that branch. Please let me know if I need to open a separate PR against 2.13.x or if 2.12.x will get merged into 2.13.x.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -195,6 +197,7 @@ public function __construct(
 
         if (isset($params['pdo'])) {
             $this->_conn = $params['pdo'];
+            $this->_conn->setAttribute(PDO::ATTR_STATEMENT_CLASS, [PDODriverStatement::class, []]);
```
mdumoulin, line 209: Maybe we should add some logic of the DriverManager here :
- Check that pdo is a PDO instance
- Trigger a deprecation

If we do so, I'm not sure that deprecation in DriverManager will still be useful. Someone using factory would get 2 deprecation.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -701,7 +704,7 @@ public static function dataCallConnectOnce(): iterable
     public function testCallConnectOnce(string $method, array $params): void
     {
         $driverMock   = $this->createMock(Driver::class);
-        $pdoMock      = $this->createMock(Connection::class);
+        $pdoMock      = $this->createMock(PDO::class);
         $platformMock = $this->createMock(AbstractPlatform::class);
         $stmtMock     = $this->createMock(Statement::class);
```
mdumoulin, line 714: Not sure that your tests modification actually test your modification. 

I would rather add a functional test where the connection is not built by DriverManager : https://github.com/doctrine/dbal/blob/1372a42083dedc56016051d61526507fb5521f9c/tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php#L354-L363

greg0ire:
2.12.x is no longer maintained, please target 2.13 instead.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
bizurkur: Added

lib/Doctrine/DBAL/Connection.php
bizurkur: - Moved the deprecation warning to trigger from Connection and removed it from DriverManager
- Added the PDO check

lib/Doctrine/DBAL/Connection.php
```diff
@@ -193,8 +195,20 @@ public function __construct(
         $this->_driver = $driver;
         $this->params  = $params;
 
+        // check for existing pdo object
+        if (isset($params['pdo']) && ! $params['pdo'] instanceof PDO) {
```
morozov: This condition partially duplicates the following one. Would it make sense to deduplicate them?
```php
if (isset(...)) {
    Deprecation::trigger()

    if (! instanceof PDO) {
        throw
    }
}
```

bizurkur: I'm all for changing it however it is wanted. The above code was a literal copy/paste from DriverManager. I wasn't sure if it was written like that to avoid nesting an `if` inside another `if` or what, so I kept it the same.

I swapped it to be like the above example, though.

lib/Doctrine/DBAL/Connection.php
```diff
@@ -193,8 +195,20 @@ public function __construct(
         $this->_driver = $driver;
         $this->params  = $params;
 
+        // check for existing pdo object
```
morozov: This comment is redundant. The code is self-explanatory.

lib/Doctrine/DBAL/DriverManager.php
```diff
@@ -220,14 +218,7 @@ public static function getConnection(
         }
 
         if (isset($params['pdo'])) {
-            Deprecation::trigger(
-                'doctrine/dbal',
-                'https://github.com/doctrine/dbal/pull/3554',
-                'Passing a user provided PDO instance directly to Doctrine is deprecated.'
-            );
-
             $params['pdo']->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_EXCEPTION);
```
morozov, line 245: Should this also be moved to `Connection`?

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -646,28 +647,40 @@ public function testFetchAll(): void
         self::assertSame($result, $conn->fetchAll($statement, $params, $types));
     }
 
+    public function testConnectionThrowsExceptionWhenInvalidPDOProvided(): void
+    {
+        $driverMock = $this->createMock(Driver::class);
+        $pdo        = $this->createMock(Connection::class);
```
morozov: Could it be an `stdClass` mock for more clarity?

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -646,28 +647,40 @@ public function testFetchAll(): void
...
+        $conn = new Connection(['pdo' => $pdo], $driverMock);
 
         self::assertTrue($conn->isConnected(), 'Connection is not connected after passing external PDO');
     }
 
     public function testCallingDeleteWithNoDeletionCriteriaResultsInInvalidArgumentException(): void
     {
         $driver  = $this->createMock(Driver::class);
-        $pdoMock = $this->createMock(\Doctrine\DBAL\Driver\Connection::class);
+        $pdoMock = $this->createMock(PDO::class);
```
morozov: The case is not PDO-specific but it's implemented via PDO as a way to inject a driver connection.

Please use the implementation from #3548: https://github.com/doctrine/dbal/blob/dcc3f2d4dcea0e7e82cfd3bfd4c1adbc925c746a/tests/Doctrine/Tests/DBAL/ConnectionTest.php#L688-L696

tests/Doctrine/Tests/DBAL/ConnectionTest.php
```diff
@@ -701,7 +714,7 @@ public static function dataCallConnectOnce(): iterable
     public function testCallConnectOnce(string $method, array $params): void
     {
         $driverMock   = $this->createMock(Driver::class);
-        $pdoMock      = $this->createMock(Connection::class);
+        $pdoMock      = $this->createMock(PDO::class);
```
morozov: Same as above, it likely doesn't have to be a PDO connection.

morozov:
> Property `Doctrine\DBAL\Connection::$_conn` (`Doctrine\DBAL\Driver\Connection|null`) does not accept `PDO`.

This error probably will have to be suppressed since the library violates its own API by allowing a PDO instance to be used instead of a driver connection.

morozov:
Please rebase on top of `2.13.x` to fix the MySQL build failures (see https://github.com/doctrine/dbal/pull/4625).

lib/Doctrine/DBAL/DriverManager.php
mdumoulin: Seems a bit risky to me. This will change the current behavior by overriding the error mode for someone providing a 'pdo' params without using DriverManager.

```php
$pdo = new PDO('sqlite::memory:');
$pdo->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_SILENT);

$conn = new \Doctrine\DBAL\Connection(['pdo' => $pdo], $driver);
```

tests/Doctrine/Tests/DBAL/ConnectionTest.php
bizurkur: I don't know how to change this, if it needs to be changed. The test is mocking `connect` and `connect` is the only method that sets `$_conn`, other than passing in a PDO object to the constructor. Which, now that it has the instanceof check, must be a PDO object.

I was able to change the other tests, just not this one.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
bizurkur: Updated

tests/Doctrine/Tests/DBAL/ConnectionTest.php
bizurkur: Yes. Changed.

lib/Doctrine/DBAL/Connection.php
bizurkur: Removed

bizurkur:
> Please rebase on top of `2.13.x` to fix the MySQL build failures (see #4625).

Rebase pushed

andrewkhan1:
Any luck getting this one pushed? It is preventing us from updating some of our dependencies. Thanks for all of your hard work everyone! :100:

morozov:
Somehow I missed all the updates after my last review. Will take a look. Thanks for the reminder, @andrewkhan1.

tests/Doctrine/Tests/DBAL/ConnectionTest.php
morozov: See the implementation from `3.x`: https://github.com/doctrine/dbal/blob/bd5fc0d72861caeed7aaef91756d3e35527aab67/tests/ConnectionTest.php#L587-L601

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
```diff
@@ -366,6 +367,21 @@ public function testUserProvidedPDOConnection(): void
...
+    public function testUserProvidedPDOConnectionWithoutDriverManager(): void
+    {
+        $connection = new Connection(
+            ['pdo' => new PDO('sqlite::memory:')],
+            new PDODriver()
+        );
+
+        $result = $connection->executeQuery('SELECT 1');
+
+        self::assertInstanceOf(ForwardCompatibility\Result::class, $result);
```
morozov: I believe the expected type should be one of these two: https://github.com/doctrine/dbal/blob/63d19ff4f3ac3eafe357e11b4fa11f13020b75a6/lib/Doctrine/DBAL/Connection.php#L1279

Otherwise, if the code starts returning just the `Result`, the method signature will be violated but the test will pass.

morozov (commented):
Looks good overall. Please see the comments and squash commits.

bizurkur:
Sorry, but I'm just gonna close this. The PR was meant to be an extremely simple fix to move the code from https://github.com/doctrine/dbal/pull/4590 into another class, nothing more. Yet it seems like each time I make the changes you guys ask for, you review it again and ask for different changes. I was only trying to move one line of code to a better location.

Feel free to re-open, copy/paste to a new PR, do whatever you want. I personally don't need this fix anymore so I've lost interest.

greg0ire:
Come on! I hate it when a contributor gets discouraged at 99% of the task! Let me do the remaining 1%, it would be a shame to waste your efforts so close to the finish line.

morozov (approved):

morozov:
Thanks, @bizurkur.

PR #4622, opened 2021-04-28
Add events for Transaction begin/commit/rollBack.

### LINKED ISSUE #3996: Events for commit and rollback
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

`doctrine/orm` has various events to listen to, for example – `postFlush`. There are use-cases where event for transaction commit is also required.

One example would be sending jobs to be processed to MQs. In case you send those instantly, it might very well be that worker would take the message even before your records are flushed into the database. If you'd send them after the flush (as to get the IDs for those records), there still might be issues in case you're inside the transaction and it's not committed yet.

Currently we configure other class for Connection, which has the following structure:

```php
<?php

declare(strict_types=1);

namespace Evp\Component\Doctrine\Connection;

use Doctrine\DBAL\Event\ConnectionEventArgs;
use Doctrine\DBAL\Connection;

class TransactionAwareConnection extends Connection
{
    public function commit()
    {
        $result = parent::commit();

        if ($this->getTransactionNestingLevel() == 0) {
            $this->getEventManager()->dispatchEvent(ConnectionEvents::onCommit, $this->getConnectEventArgs());
        }

        return $result;
    }

    public function rollBack()
    {
        parent::rollBack();
        if ($this->getTransactionNestingLevel() === 0) {
            $this->getEventManager()->dispatchEvent(ConnectionEvents::onRollback, $this->getConnectEventArgs());
        }
    }

    protected function getConnectEventArgs()
    {
        return new ConnectionEventArgs($this);
    }
}

```

This works well for us, but custom logic does not allow to build any open-source libraries that would depend on such events. Having such events in the core would allow to easily use them when needed.

Warxcell:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #3996

#### Summary

The use-case is the following: Got Entity which represents file. Got listener https://github.com/Warxcell/files/blob/master/src/EventListener/DoctrineORMListener.php#L49-L53 which deletes the actual file in case Entity is deleted. All good until it's all wrapped in transaction:

```php
$em->beginTransaction();
try {
  $em->delete($file);

  $em->flush();

  // do other stuffs
  
  $em->flush();

  $em->commit();
} catch(\Exception $ex) {
   $em->rollBack();
}
```

if exception is thrown after 1st flush - file will be deleted but Entity will stays.

### DIFF
--- a/docs/en/reference/events.rst
+++ b/docs/en/reference/events.rst
@@ -334,3 +334,84 @@ for event listeners.
 
 It allows you to access the table index definitions of the current database, table name, Platform and
 ``Doctrine\DBAL\Connection`` instance. Indexes, that are about to be added, are not listed.
+
+OnTransactionBegin Event
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+``Doctrine\DBAL\Events::onTransactionBegin`` is triggered when ``Doctrine\DBAL\Connection::beginTransaction()``
+is called. An instance of ``Doctrine\DBAL\Event\TransactionBeginEventArgs`` is injected as argument for event listeners.
+
+.. code-block:: php
+
+    <?php
+    class MyEventListener
+    {
+        public function onTransactionBegin(TransactionBeginEventArgs $event)
+        {
+            // Your EventListener code
+        }
+    }
+
+    $evm = new EventManager();
+    $evm->addEventListener(Events::onTransactionBegin, new MyEventListener());
+
+    $conn = DriverManager::getConnection($connectionParams, null, $evm);
+
+It allows you to access the ``Doctrine\DBAL\Connection`` instance.
+Please note that this event can be called multiple times, since transactions can be nested.
+
+OnTransactionCommit Event
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``Doctrine\DBAL\Events::onTransactionCommit`` is triggered when ``Doctrine\DBAL\Connection::commit()`` is called.
+An instance of ``Doctrine\DBAL\Event\TransactionCommitEventArgs`` is injected as argument for event listeners.
+
+.. code-block:: php
+
+    <?php
+    class MyEventListener
+    {
+        public function onTransactionCommit(TransactionCommitEventArgs $event)
+        {
+            // Your EventListener code
+        }
+    }
+
+    $evm = new EventManager();
+    $evm->addEventListener(Events::onTransactionCommit, new MyEventListener());
+
+    $conn = DriverManager::getConnection($connectionParams, null, $evm);
+
+It allows you to access the ``Doctrine\DBAL\Connection`` instance.
+Please note that this event can be called multiple times, since transactions can be nested.
+If you want to know if a transaction is actually committed, you should rely on
+``TransactionCommitEventArgs::getConnection()->getTransactionNestingLevel() === 0`` or
+``TransactionCommitEventArgs::getConnection()->isTransactionActive()``
+
+OnTransactionRollBack Event
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``Doctrine\DBAL\Events::onTransactionRollBack`` is triggered when ``Doctrine\DBAL\Connection::rollBack()`` is called.
+An instance of ``Doctrine\DBAL\Event\TransactionRollBackEventArgs`` is injected as argument for event listeners.
+
+.. code-block:: php
+
+    <?php
+    class MyEventListener
+    {
+        public function onTransactionRollBack(TransactionRollBackEventArgs $event)
+        {
+            // Your EventListener code
+        }
+    }
+
+    $evm = new EventManager();
+    $evm->addEventListener(Events::onTransactionRollBack, new MyEventListener());
+
+    $conn = DriverManager::getConnection($connectionParams, null, $evm);
+
+It allows you to access the ``Doctrine\DBAL\Connection`` instance.
+Please note that this event can be called multiple times, since transactions can be nested.
+If you want to know if a transaction is actually rolled back, you should rely on
+``TransactionCommitRollBackArgs::getConnection()->getTransactionNestingLevel() === 0`` or
+``TransactionCommitRollBackArgs::getConnection()->isTransactionActive()``
--- a/tests/ConnectionTest.php
+++ b/tests/ConnectionTest.php
@@ -11,6 +11,9 @@ use Doctrine\DBAL\Driver;
 use Doctrine\DBAL\Driver\Connection as DriverConnection;
 use Doctrine\DBAL\Driver\ServerInfoAwareConnection;
 use Doctrine\DBAL\DriverManager;
+use Doctrine\DBAL\Event\TransactionBeginEventArgs;
+use Doctrine\DBAL\Event\TransactionCommitEventArgs;
+use Doctrine\DBAL\Event\TransactionRollBackEventArgs;
 use Doctrine\DBAL\Events;
 use Doctrine\DBAL\Exception;
 use Doctrine\DBAL\Exception\InvalidArgumentException;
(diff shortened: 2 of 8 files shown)


greg0ire:
Will give a review if this gets accepted in principle. I have no objections, but maybe others will.

Warxcell:
> Will give a review if this gets accepted in principle. I have no objections, but maybe others will.

Sure. Added the docs aswell. I really hope it will be accepted. Otherwise I will be forced to use SQL Logger and check if SQL === "START TRANSACTION"; || "COMMIT" || "ROLLBACK;"

docs/en/reference/events.rst
```diff
@@ -334,3 +334,85 @@ for event listeners.
 
 It allows you to access the table index definitions of the current database, table name, Platform and
 ``Doctrine\DBAL\Connection`` instance. Indexes, that are about to be added, are not listed.
+
+OnTransactionBegin Event
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
morozov: Please clean up the formatting (also in other headers).

docs/en/reference/events.rst
```diff
@@ -334,3 +334,85 @@ for event listeners.
...
+            // Your EventListener code
+        }
+    }
+
+    $evm = new EventManager();
+    $evm->addEventListener(Events::onTransactionBegin, new MyEventListener());
+
+    $conn = DriverManager::getConnection($connectionParams, null, $evm);
+
+It allows you to access the``Doctrine\DBAL\Connection`` instance.
```
morozov: A space is missing.

docs/en/reference/events.rst
```diff
@@ -334,3 +334,85 @@ for event listeners.
...
+
+    $evm = new EventManager();
+    $evm->addEventListener(Events::onTransactionCommit, new MyEventListener());
+
+    $conn = DriverManager::getConnection($connectionParams, null, $evm);
+
+It allows you to access the``Doctrine\DBAL\Connection`` instance.
+Please note that this event can be called multiple times, since transactions can be nested.
+If you want to know if transaction is actually committed, you should rely on
+`TransactionCommitEventArgs->getConnection()->getTransactionNestingLevel() === 0`
```
morozov: It probably should be `TransactionCommitEventArgs::` instead of `->` since it's a class name.

docs/en/reference/events.rst
```diff
@@ -334,3 +334,85 @@ for event listeners.
...
+
+OnTransactionRollBack Event
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``Doctrine\DBAL\Events::onTransactionRollBack`` is triggered on `Connection::rollBack()` is called.
```
morozov: Should `Connection::rollBack()` be enclosed in double backticks as well (also, check other cases)?

docs/en/reference/events.rst
```diff
@@ -334,3 +334,85 @@ for event listeners.
...
+            // Your EventListener code
+        }
+    }
+
+    $evm = new EventManager();
+    $evm->addEventListener(Events::onTransactionRollBack, new MyEventListener());
+
+    $conn = DriverManager::getConnection($connectionParams, null, $evm);
+
+It allows you to access the``Doctrine\DBAL\Connection`` instance.
```
morozov: A space is missing.

tests/ConnectionTest.php
```diff
@@ -134,6 +137,114 @@ public function testConnectDispatchEvent(): void
...
     }
 
+    public function testTransactionBeginDispatchEvent(): void
+    {
+        $listenerMock = $this->createMock(TransactionBeginDispatchEventListener::class);
+        $listenerMock->expects(self::exactly(2))->method('onTransactionBegin');
+
+        $eventManager = new EventManager();
+        $eventManager->addEventListener([Events::onTransactionBegin], $listenerMock);
+        $listener = new class {
```
morozov: Since this is a connection unit test, I'd prefer to have the listener interface mocked instead of being implemented with additional code. See other tests for example:

https://github.com/doctrine/dbal/blob/c7201157f8e567fe4b61b0ffa0f1f04c0dc19992/tests/ConnectionTest.php#L124-L127

tests/ConnectionTest.php
```diff
@@ -134,6 +137,114 @@ public function testConnectDispatchEvent(): void
...
+            }
+        };
+        $eventManager->addEventListener([Events::onTransactionBegin], $listener);
+
+        $driverMock = $this->createMock(Driver::class);
+
+        $conn = new Connection([], $driverMock, new Configuration(), $eventManager);
+        $conn->beginTransaction();
+        self::assertSame($conn, $listener->getConnection());
+        self::assertEquals(1, $listener->getConnection()->getTransactionNestingLevel());
```
morozov: The transaction nesting level doesn't really look relevant to the logic being tested: the only responsibility of the connection is to call the listener and pass itself as an argument.

If you want to have a more complex scenario covered by a test, I believe it's better to be done in a separate test.

morozov:
> I have no objections, but maybe others will.

Looks reasonable to me.

tests/ConnectionTest.php
Warxcell: Yeah, I guess `self::assertSame($conn, $listener->getConnection());` is enough.

tests/ConnectionTest.php
Warxcell: How can I `self::assertSame($conn, $listener->getConnection());` with mock?

tests/ConnectionTest.php
morozov: It's better if the mock just asserts that it's called with the expected arguments (`$args->getConnection()`), it shouldn't contain any state.

Warxcell: Yes, but how do I assert that? Since it's called with `Event` class that wraps the connection. Is that possible with mock?

docs/en/reference/events.rst
```diff
@@ -334,3 +334,85 @@ for event listeners.
 
 It allows you to access the table index definitions of the current database, table name, Platform and
 ``Doctrine\DBAL\Connection`` instance. Indexes, that are about to be added, are not listed.
+
+OnTransactionBegin Event
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``Doctrine\DBAL\Events::onTransactionBegin`` is triggered on `Connection::beginTransaction()` is called.
```
greg0ire:
```suggestion
``Doctrine\DBAL\Events::onTransactionBegin`` is triggered when ``Connection::beginTransaction()`` is called.
```

docs/en/reference/events.rst
```diff
@@ -334,3 +334,85 @@ for event listeners.
...
+
+    $conn = DriverManager::getConnection($connectionParams, null, $evm);
+
+It allows you to access the``Doctrine\DBAL\Connection`` instance.
+Please note that this event can be called multiple times, since transactions can be nested.
+
+OnTransactionCommit Event
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``Doctrine\DBAL\Events::onTransactionCommit`` is triggered on `Connection::commit()` is called.
```
greg0ire:
```suggestion
``Doctrine\DBAL\Events::onTransactionCommit`` is triggered when ``Connection::commit()`` is called.
```

tests/ConnectionTest.php
Warxcell: I saw it... it have 
```php
$this->callback(function($subject){
  return is_callable([$subject, 'getName']) &&
         $subject->getName() == 'My subject';
})
```

docs/en/reference/events.rst
greg0ire:
```suggestion
``Doctrine\DBAL\Events::onTransactionRollBack`` is triggered when ``Connection::rollBack()`` is called.
```

tests/ConnectionTest.php
morozov: Yeah, you can do something like:
```php
$listener->expects(self::once())
    ->method('onTransactionBegin')
    ->with(self::callback(function (TransactionBeginEventArgs $args) use ($conn) {
        $this->assertSame($conn, $args->getConnection());
    }));
```

Warxcell: Yeah, refactored it like that:
```php
        $listenerMock = $this->createMock(TransactionRollBackDispatchEventListener::class);
        $listenerMock
            ->expects(self::exactly(1))
            ->method('onTransactionRollBack')
            ->with(
                self::callback(
                    static function (TransactionRollBackEventArgs $eventArgs) use ($conn): bool {
                        return $eventArgs->getConnection() === $conn;
                    }
                )
            );
```
It can be seen now. According to phpunit docs, it should return bool if argument is correct,

`The callback() constraint can be used for more complex argument verification. This constraint takes a PHP callback as its only argument. The PHP callback will receive the argument to be verified as its only argument and should return true if the argument passes verification and false otherwise.`

Warxcell:
Btw on #3996 there is proposal for
```php
ConnectionEvents::preTransactionCommit
ConnectionEvents::postTransactionCommit
ConnectionEvents::preTransactionRollback
ConnectionEvents::postTransactionRollback
ConnectionEvents::preTransactionBegin
ConnectionEvents::postTransactionBegin
```

What do you think about it? Or current implementation with single event which occurs after the action is done is enough?

tests/ConnectionTest.php
```diff
@@ -134,6 +137,85 @@ public function testConnectDispatchEvent(): void
...
+                self::callback(
+                    static function (TransactionBeginEventArgs $eventArgs) use ($conn): bool {
+                        return $eventArgs->getConnection() === $conn;
+                    }
+                )
+            );
+        $eventManager->addEventListener([Events::onTransactionBegin], $listenerMock);
+
+        $conn->beginTransaction();
+        $conn->beginTransaction();
```
morozov: Why is it necessary to test multiple calls? They are independent.

tests/ConnectionTest.php
```diff
@@ -134,6 +137,85 @@ public function testConnectDispatchEvent(): void
...
+                    }
+                )
+            );
+
+        $eventManager->addEventListener([Events::onTransactionRollBack], $listenerMock);
+
+        $conn->beginTransaction();
+        $conn->beginTransaction();
+
+        $conn->commit();
```
morozov: Why is it necessary to commit in order to test a rollback?

morozov (commented):
> What do you think about it? Or current implementation with single event which occurs after the action is done is enough?

I think the "after the action" events are sufficient. If a transaction wasn't committed successfully, there commit handler will be not invoked as if it wasn't committed at all, and so on.

tests/ConnectionTest.php
Warxcell: To keep from regression, for example if tomorrow somebody adds a check that will call `onBeginTransaction` single time when `nesting level === 0`.

morozov: What if tomorrow somebody decides to call the listener only when the nesting level is less than two? Currently, there's no logic that depends on the nesting level, so testing it doesn't make much sense.

Warxcell: ok, will delete these lines.

Warxcell:
Added one test that verifies commit event is not called if rollback is initiated.

morozov:
@Warxcell please address the test failures.

Warxcell:
Yeah, sorry. I cannot run tests locally, because I'm missing pdo_mysql. It should be fine now.

Warxcell:
Current failure seems unrelated: 

```
There was 1 failure:
1) Doctrine\DBAL\Tests\Functional\Platform\DefaultExpressionTest::testCurrentTimestamp
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'2021-05-10 06:00:34.473000'
+'2021-05-10 06:00:34.470000'
C:\projects\dbal\tests\Functional\Platform\DefaultExpressionTest.php:75
C:\projects\dbal\tests\Functional\Platform\DefaultExpressionTest.php:52
FAILURES!
````

**Review by morozov — DISMISSED** [review]

**Review by morozov — DISMISSED** [review]

**Review by SenseException — DISMISSED** [review]

File: docs/en/reference/events.rst [hunk 12]
```diff
@@ -334,3 +334,82 @@ for event listeners.
...
+    }
+
+    $evm = new EventManager();
+    $evm->addEventListener(Events::onTransactionCommit, new MyEventListener());
+
+    $conn = DriverManager::getConnection($connectionParams, null, $evm);
+
+It allows you to access the``Doctrine\DBAL\Connection`` instance.
+Please note that this event can be called multiple times, since transactions can be nested.
+If you want to know if transaction is actually committed, you should rely on
```
greg0ire:
```suggestion
If you want to know if a transaction is actually committed, you should rely on
```

docs/en/reference/events.rst
```diff
@@ -334,3 +334,82 @@ for event listeners.
...
+    }
+
+    $evm = new EventManager();
+    $evm->addEventListener(Events::onTransactionRollBack, new MyEventListener());
+
+    $conn = DriverManager::getConnection($connectionParams, null, $evm);
+
+It allows you to access the ``Doctrine\DBAL\Connection`` instance.
+Please note that this event can be called multiple times, since transactions can be nested.
+If you want to know if transaction is actually rollback, you should rely on
```
greg0ire:
```suggestion
If you want to know if a transaction is actually rolled back, you should rely on
```

greg0ire (dismissed):

greg0ire:

How to do that?

1. `git rebase -i origin/3.2.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

Warxcell:
> Please kindly squash your commits together.

Done. Also added note about ``TransacationCommitEventArgs::getConnection()->isTransactionActive()``. Didn't saw that before.

docs/en/reference/events.rst
```diff
@@ -334,3 +334,84 @@ for event listeners.
...
+    $evm = new EventManager();
+    $evm->addEventListener(Events::onTransactionRollBack, new MyEventListener());
+
+    $conn = DriverManager::getConnection($connectionParams, null, $evm);
+
+It allows you to access the ``Doctrine\DBAL\Connection`` instance.
+Please note that this event can be called multiple times, since transactions can be nested.
+If you want to know if a transaction is actually rolled back, you should rely on
+``TransactionCommitRollBackArgs::getConnection()->getTransactionNestingLevel() === 0`` or
+``TransacationCommitEventArgs::getConnection()->isTransactionActive()``
```
greg0ire:
```suggestion
``TransactionCommitEventArgs::getConnection()->isTransactionActive()``
```

docs/en/reference/events.rst
```diff
@@ -334,3 +334,84 @@ for event listeners.
...
+    $evm = new EventManager();
+    $evm->addEventListener(Events::onTransactionCommit, new MyEventListener());
+
+    $conn = DriverManager::getConnection($connectionParams, null, $evm);
+
+It allows you to access the ``Doctrine\DBAL\Connection`` instance.
+Please note that this event can be called multiple times, since transactions can be nested.
+If you want to know if a transaction is actually committed, you should rely on
+``TransactionCommitEventArgs::getConnection()->getTransactionNestingLevel() === 0`` or
+``TransacationCommitEventArgs::getConnection()->isTransactionActive()``
```
greg0ire:
```suggestion
``TransactionCommitEventArgs::getConnection()->isTransactionActive()``
```

morozov (approved):

morozov:
Thanks, @Warxcell!

PR #4614, opened 2021-04-20
[GH-4613] Use utf8mb4 instead of utf8 for testing connection charset

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

Closes #4613.

PR #4623, opened 2021-04-29
Support Doctrine Cache 2

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

This PR proposes to indicate compatibility with Doctrine Cache 2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
     "require": {
         "php": "^7.1 || ^8",
         "ext-pdo": "*",
-        "doctrine/cache": "^1.0",
+        "doctrine/cache": "^1.0|^2.0",
         "doctrine/deprecations": "^0.5.3",
         "doctrine/event-manager": "^1.0"
     },
@@ -44,6 +44,7 @@
         "phpstan/phpstan": "0.12.81",
         "phpunit/phpunit": "^7.5.20|^8.5|9.5.0",
         "squizlabs/php_codesniffer": "3.6.0",
+        "symfony/cache": "^4.4",
         "symfony/console": "^2.0.5|^3.0|^4.0|^5.0",
         "vimeo/psalm": "4.6.4"
     },
--- a/tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
@@ -3,6 +3,8 @@
 namespace Doctrine\Tests\DBAL\Functional;
 
 use Doctrine\Common\Cache\ArrayCache;
+use Doctrine\Common\Cache\Cache;
+use Doctrine\Common\Cache\Psr6\DoctrineProvider;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\ConnectionException;
@@ -19,8 +21,10 @@ use Error;
 use Exception;
 use PDO;
 use RuntimeException;
+use Symfony\Component\Cache\Adapter\ArrayAdapter;
 use Throwable;
 
+use function class_exists;
 use function file_exists;
 use function in_array;
 use function unlink;
@@ -389,7 +393,7 @@ class ConnectionTest extends DbalFunctionalTestCase
             $this->connection->getDatabasePlatform()->getDummySelectSQL(),
             [],
             [],
-            new QueryCacheProfile(1, 'cacheKey', new ArrayCache())
+            new QueryCacheProfile(1, 'cacheKey', $this->getArrayCache())
         );
 
         self::assertInstanceOf(Result::class, $result);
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -34,7 +34,7 @@
     "require": {
         "php": "^7.1 || ^8",
         "ext-pdo": "*",
-        "doctrine/cache": "^1.0",
+        "doctrine/cache": "^1.0|^2.0",
```
alcaeus, line 37: For code this is fine, but for tests we'll have to require `^1.11|^2.0` to avoid `--prefer-lowest` builds failing because 1.0 doesn't contain the PSR-6 adapters.

alcaeus (changes requested):
doctrine/cache dependency needs to be bumped at least for require-dev, but rest LGTM

composer.json
derrabus: If the adapter is not there, the tests fall back to `ArrayCache`, so we're good.

alcaeus: 🤦‍♂️ Of course, yes.

alcaeus (dismissed):

morozov:
To fix the build failure, we need to backport https://github.com/doctrine/dbal/pull/4614.

greg0ire (dismissed):

morozov (dismissed):

SenseException (commented):
The build is currently only loading doctrine/cache 1.x on composer install. It would be great if some of them are installing 2.x to show that the tests are also working with 2.x.

derrabus:
@SenseException I've tested with 2.0-dev locally. If you want v2 to show up in CI, someone would have to tag `2.0.0` on the Doctrine Cache repo or add a job to this repo that tests with `"minimum-stability": "dev"`.

greg0ire:
Let's wait for 2.0.0 to happen then?

alcaeus:
Done: https://github.com/doctrine/cache/releases/tag/2.0.0

derrabus:
I've added small fixes to make Psalm happy, now that Doctrine Cache 2 has been tagged.

greg0ire (dismissed):

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
```diff
@@ -402,10 +407,20 @@ public function testResultCompatibilityWhenExecutingCacheQuery(): void
...
         self::assertInstanceOf(Result::class, $result);
         self::assertInstanceOf(Driver\ResultStatement::class, $result);
     }
+
+    private function getArrayCache(): Cache
+    {
+        $cache = class_exists(DoctrineProvider::class)
+            ? DoctrineProvider::wrap(new ArrayAdapter())
+            : new ArrayCache();
+        assert($cache instanceof Cache);
```
morozov: Why is this assertion necessary? Shouldn't both versions implement the expected interface?

derrabus: Yes, but when Doctrine Cache 2 is installed, `ArrayCache` is undefined and Psalm is confused. We can remove the assertion on 3.2.x.

morozov: Shouldn't there be also a condition like `class_exists()`? From the static analysis PoV, the fact that the `DoctrineProvider` class doesn't exist doesn't mean that `ArrayCache` does.

So it could be something like:
```php
if (class_exists(DoctrineProvider::class)) {
    return DoctrineProvider::wrap(new ArrayAdapter());
}

if (class_exists(ArrayCache::class)) {
    return new ArrayCache();
}

self::markTestSkipped();
```

SenseException (dismissed):

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
derrabus: I'm using your code now. Psalm seems to be less confused that way. 🙂

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
```diff
@@ -402,10 +406,23 @@ public function testResultCompatibilityWhenExecutingCacheQuery(): void
...
+    {
+        if (class_exists(DoctrineProvider::class)) {
+            return DoctrineProvider::wrap(new ArrayAdapter());
+        }
+
+        if (class_exists(ArrayCache::class)) {
+            return new ArrayCache();
+        }
+
+        self::markTestSkipped();
```
morozov: As a second thought, I think failing (`self::fail('Cannot instantiate cache backend.')`) would be a safer option (like it would fail in the original implementation).

morozov (dismissed):

tests/Doctrine/Tests/DBAL/Functional/ConnectionTest.php
derrabus: Sure.

morozov (approved):

greg0ire:
Thanks @derrabus !

PR #4630, opened 2021-05-02
Suppress deprecation notices for ArrayCache

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

CI builds fail since the release of https://github.com/doctrine/cache/pull/355. The issue isn't reproducible on `2.13.x` because it suppresses all deprecation notices.


derrabus:
Merging 2.13.x up should also fix the issue.

derrabus:
See #4632

PR #4632, opened 2021-05-02
Merge 2.13.x into 3.1.x

derrabus:
This PR proposes to merge the latest changes of the 2.13.x branch into 3.1.x. While resolving conflicts, I've bumped Symfony Cache to 5.2 because we don't need to support PHP 7.1 anymore and to match the constraint on 3.2.x.

Replaces #4630.

PR #4633, opened 2021-05-02
Merge 2.13.x/3.1.x into 3.2.x

derrabus:
Follows #4632

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 # Doctrine DBAL
 
-| [4.0-dev][4.0] | [3.0][3.0] | [2.13][2.13] |
+| [4.0-dev][4.0] | [3.1][3.1] | [2.13][2.13] |
 |:----------------:|:----------:|:----------:|
-| [![GitHub Actions][GA 4.0 image]][GA 4.0] | [![GitHub Actions][GA 3.0 image]][GA 3.0] | [![GitHub Actions][GA 2.13 image]][GA 2.13] |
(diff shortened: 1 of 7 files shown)


morozov:
@derrabus what's the difference between `3.1.x` and `3.2.x` that requires fixing tests? It is possible that there is a BC break.

derrabus:
The `DoctrineProvider` class transforms the cache key. That's why I'm performing the assertion on the unwrapped adapter.

morozov:
A functional test cannot perform assertions on implementation details – that's why it has to be reworked when implementation details change creating a sense of a BC break. Since caching is by definition a non-functional requirement I'm curious if caching can be tested in a functional way at all.

We already had a similar discussion in https://github.com/doctrine/dbal/pull/4189#discussion_r464738865. Should we just get rid of this "functional" test?

derrabus:
Sorry for the late reply. My understanding of the test is "If this operation is performed, make sure that *something* has been written to the cache". I think, this is a valid thing to test. One could argue if there's a better way to do this, but for this PR (because it's just a merge from a lower branch), my main concern was to make the test work as it had before.

morozov:
> If this operation is performed, make sure that something has been written to the cache

This would be the responsibility of a unit test that would have changed as soon as the new implementation was introduced, not during a merge-up.

> my main concern was to make the test work as it had before

My main concern is that the caching piece of logic is barely maintainable (not your fault), so the rework of its internals could be used as an opportunity to improve the tests as well. @greg0ire please step in and review.

derrabus:
I've created #4639 to resolve the issue before we merge the branches. Let's continue the discussion there.

PR #4639, opened 2021-05-06
Add tests for wrapped PSR-6 caches

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

This PR adds tests that make sure the result cache also works with wrapped PSR-6 caches passed to `QueryCacheProfile`.

### DIFF
--- a/tests/Functional/ResultCacheTest.php
+++ b/tests/Functional/ResultCacheTest.php
@@ -3,6 +3,7 @@
 namespace Doctrine\DBAL\Tests\Functional;
 
 use Doctrine\Common\Cache\ArrayCache;
+use Doctrine\Common\Cache\Psr6\DoctrineProvider;
 use Doctrine\DBAL\Cache\QueryCacheProfile;
 use Doctrine\DBAL\Logging\DebugStack;
 use Doctrine\DBAL\Result;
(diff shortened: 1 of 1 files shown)


tests/Functional/ResultCacheTest.php
```diff
@@ -402,6 +432,37 @@ public function testChangeCacheImplLegacy(): void
...
+            [],
+            [],
+            new QueryCacheProfile(10, 'emptycachekey', $secondCache)
+        );
+
+        $this->hydrateViaIteration($stmt, static function (Result $result) {
+            return $result->fetchAssociative();
+        });
+
+        self::assertCount(2, $this->sqlLogger->queries, 'two hits');
```
greg0ire: This test is a bit hard to understand, especially because of this comment, which I know you only copied, but still… if I am correct this hits are database hits, not cache hits. If that's true, let's make it clear.
```suggestion
        self::assertCount(2, $this->sqlLogger->queries, 'two database hits');
```

derrabus: I have adjusted all `assetCounts()` calls on `$this->sqlLogger->queries` and added meaningful messages.

greg0ire (approved):

greg0ire:
Thanks @derrabus !

morozov:
@derrabus since this PR is merged, could you reopen and complete https://github.com/doctrine/dbal/pull/4633?

derrabus:
@morozov Reopened as #4664

PR #4644, opened 2021-05-11
Remove defaults for MySQL table charset, collation and engine

### LINKED ISSUE #4631: utf8mb4_0900_ai_ci should be the default collation for server_version: '8.0'
The default collation for MySQL is `utf8mb4_0900_ai_ci` since MySQL 8.0.1.

Doctrine is still using the older `utf8mb4_unicode_ci` to create new tables, when the server_version settings is set to 8.0

### Feature Request

Update the default collate to the newest version, to guarantee the best out-of-the-box configuration.

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

Closes #4631.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -8,6 +8,11 @@ awareness about deprecated code.
 
 # Upgrade to 4.0
 
+## Removed defaults for MySQL table charset, collation and engine
+
+The library no longer provides the default values for MySQL table charset, collation and engine.
+If omitted in the table definition, MySQL will derive the values from the database options.
+
 ## Removed `ReservedWordsCommand::setKeywordListClass()`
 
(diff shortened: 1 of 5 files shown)


greg0ire (approved):

greg0ire:
This probably made sense in the past: https://stackoverflow.com/questions/6769901/why-is-mysqls-default-collation-latin1-swedish-ci

I think it's now safe to assume MySQL uses sensible defaults 👍

morozov:
Even at the times when the defaults weren't that reasonable, the user was able to specify the options that would correspond to their use case. Overriding defaults in the DBAL didn't make sense even at that time.

Issue #4645, opened 2021-05-12
PHP warning on unknown database with Doctrine DBAL 2.13.x


mbrodala:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 2.13.1

#### Summary

Connection attempts to a missing database lead to a PHP warning.

#### Previous behaviour

No warning was shown.

#### Current behavior

A PHP warning is issued:
> PHP Warning:  mysqli::real_connect(): (HY000/1049): Unknown database ...

#### How to reproduce

1. Do a connection to a missing database.
2. With Doctrine DBAL 2.11.x no error occurs
3. With Doctrine DBAL 2.13.x a PHP warning occurs

Also see https://github.com/Nimut/testing-framework/issues/136 and https://github.com/TYPO3/testing-framework/issues/233 for more specific examples.

morozov:
Could be solved by backporting c9ce80acd7fc480c6c4585dc285741722919efe3. This will also solve the test failure on non-Linux operating systems:
```
1) Doctrine\Tests\DBAL\Driver\Mysqli\MysqliConnectionTest::testRestoresErrorHandlerOnException
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'Network is unreachable'
+'Permission denied'

tests/Doctrine/Tests/DBAL/Driver/Mysqli/MysqliConnectionTest.php:56
```

bmack:
@morozov Thank you for this super quick fix 🙇

morozov:
@bmack, you're welcome. Closed via #4647.

PR #4646, opened 2021-05-12
Properly catch lowlevel MySQLi connection errors again

mbrodala:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #4645

#### Summary

Restore the original error handling for MySQLi connections.

If a `set_error_handler()` callback returns `false`, the PHP error handler will be involved. This is a regression since the previous version implicitly returned `null` and thus skipped the PHP error handler.

See https://github.com/doctrine/dbal/commit/26c7adc7f09f408fb2138153c0031e462d33a973
And https://github.com/doctrine/dbal/commit/4a5be97953940a94a3cbe472e8eae450e1dcb1a3


morozov:
@mbrodala this must have been addressed in https://github.com/doctrine/dbal/pull/4647.

mbrodala:
@morozov Yes, I can confirm that your fix solves my problem. So my interim fix here can be skipped. :+1:

PR #4648, opened 2021-05-13
[GH-4643] Fix SQLServerPlatform::quoteIdentifier()

### LINKED ISSUE #4643: SQLServer platform quoteIdentifier wrong escaping
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | all

#### Summary

It looks like that escaping identifier for sql platform is broken since day one.

#### Current behaviour

This is in tests:
https://github.com/doctrine/dbal/blob/6611e6d24560944e0a299cfd619013597833efc2/tests/Platforms/AbstractSQLServerPlatformTestCase.php#L576
But if I try for example create table
```SQL
CREATE TABLE [test_escaping].[fo][o] ([col1] INT);
-- Failed to execute query. Error: Parse error at line: 1, column: 34: Incorrect syntax near '[o]'.
```

#### Expected behaviour

Expected output should be `[fo]]o]`

```SQL
CREATE TABLE [test_escaping].[fo]]o] ([col1] INT);
-- Query succeeded: Affected rows: 0
```

So the function should be:
```PHP
    public function quoteSingleIdentifier($str)
    {
        return '[' . str_replace(']', ']]', $str) . ']';
    }
```
Output is than same as from
```SQL
SELECT QUOTENAME('fo]o');
-- [fo]]o]
```
https://docs.microsoft.com/en-us/sql/t-sql/functions/quotename-transact-sql?view=sql-server-ver15

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

Closes #4643.

From the Oracle [documentation](https://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements008.htm):

> Quoted identifiers can contain any characters and punctuations marks as well as spaces. However, neither quoted nor nonquoted identifiers can contain double quotation marks or the null character (`\0`).

### DIFF
--- a/tests/Doctrine/Tests/DBAL/Functional/Platform/QuotingTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Platform/QuotingTest.php
@@ -2,7 +2,10 @@
 
 namespace Doctrine\Tests\DBAL\Functional\Platform;
 
+use Doctrine\DBAL\Platforms\OraclePlatform;
 use Doctrine\Tests\DbalFunctionalTestCase;
+
+use function key;
 
 class QuotingTest extends DbalFunctionalTestCase
 {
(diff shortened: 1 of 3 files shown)


Ocramius (dismissed):

greg0ire:
Shouldn't the commit message (and PR title) reference 4643 (as opposed to 4645?)

morozov:
> Shouldn't the commit message (and PR title) reference 4643 (as opposed to 4645?)

Yeah, they should. Thanks!

tests/Doctrine/Tests/DBAL/Functional/Platform/QuotingTest.php
```diff
@@ -29,4 +32,41 @@ public static function stringLiteralProvider(): iterable
...
+        );
+
+        $row = $this->connection->fetchAssociative($query);
+
+        self::assertNotFalse($row);
+        self::assertSame($identifier, key($row));
+    }
+
+    /**
+     * @return iterable<string,array{0:string}>
```
greg0ire, line 61: Let's omit numeric keys, shall we? Also, we should probably format this similarly to PHP code (with spaces after commas  and colons)

```suggestion
     * @return iterable<string, array{string}>
```

morozov: Unless it's part of the standard, there's no way to keep it consistent across the codebase. There are existing code fragments formatted like in my PR. Reformatting the code here will just increase the inconsistency. I would start with a coding standard proposal and disregard these issues until it's enforced by the standard.

greg0ire (approved):

PR #4650, opened 2021-05-13
PostgreSQL: Add support for NOW() and other functions as DEFAULT value for datetime columns

sustmi:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | Yes _? (It changes behavior of existing code, but it is actually a bugfix)_
| Fixed issues | _(Should I file a new issue too or is the PR sufficient?)_ 

#### Summary
When I define a new column in my entity as this:
```php
    /**
     * @var DateTime|null
     * @ORM\Column(type="datetime", nullable=true, options={"default": "NOW()"})
     */
    private $myColumn;
```
and I run:
```bash
php bin/console doctrine:migrations:diff
```
the newly generated migration contains the following `ALTER TABLE` query:
```sql
ALTER TABLE my_table ADD my_column TIMESTAMP(0) WITHOUT TIME ZONE DEFAULT 'NOW()'
```
Notice that the `NOW()` functions is quoted as a string.

While running an `INSERT` query without specifying `my_column` value afterwards, the column is filled with time when the `ALTER` query was executed (ie. it remains constant) and not the time when the `INSERT` query was executed.
I do not think that this is an expected behavior when setting `options={"default": "NOW()"}`.
I would expect the `my_column` to be filled with the current time when executing the `INSERT` query.

#### Additional notes
- There already is support for not [quoting](https://github.com/doctrine/dbal/blob/6611e6d24560944e0a299cfd619013597833efc2/src/Platforms/AbstractPlatform.php#L2316) `CURRENT_TIMESTAMP` function here: https://github.com/doctrine/dbal/blob/6611e6d24560944e0a299cfd619013597833efc2/src/Platforms/AbstractPlatform.php#L2300-L2302

- I only use PostgreSQL so I do not try to fix this on the `AbstractPlatform` level. I guess we could add support for `getNowExpression()`: https://github.com/doctrine/dbal/blob/6611e6d24560944e0a299cfd619013597833efc2/src/Platforms/AbstractPlatform.php#L841-L844
but I am not sure if all platforms support now expression as default value and for which column types.
Also, I am actually interested mainly in support of `NOW()` function and `PostgreSqlPlatform` overrides `getNowExpression()` method and returns `LOCALTIMESTAMP(0)` instead of `NOW()`. :neutral_face:  

- Additionally, note that `AbstractPlatform::getDefaultValueDeclarationSQL()` only supports upper-case variants (note the `===` comparison) of the functions (eg. `CURRENT_TIMESTAMP`): https://github.com/doctrine/dbal/blob/6611e6d24560944e0a299cfd619013597833efc2/src/Platforms/AbstractPlatform.php#L2300
while my PR adds case-insensitive support.

- I think this is my first contribution to `doctrine/dbal` so any feedback is welcome.

### DIFF
--- a/src/Platforms/PostgreSQL94Platform.php
+++ b/src/Platforms/PostgreSQL94Platform.php
@@ -13,6 +13,7 @@ use Doctrine\DBAL\Types\BigIntType;
 use Doctrine\DBAL\Types\BinaryType;
 use Doctrine\DBAL\Types\BlobType;
 use Doctrine\DBAL\Types\IntegerType;
+use Doctrine\DBAL\Types\PhpDateTimeMappingType;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\Deprecations\Deprecation;
 use UnexpectedValueException;
@@ -60,6 +61,18 @@ class PostgreSQL94Platform extends AbstractPlatform
             'off',
             '0',
         ],
+    ];
+
+    // Based on https://www.postgresql.org/docs/9.4/functions-datetime.html#FUNCTIONS-DATETIME-TABLE
+    private const ALLOWED_DATETIME_DEFAULT_EXPRESSIONS = [
+        'clock_timestamp()',
+        'current_timestamp',
+        'current_timestamp()',
+        'localtimestamp',
+        'localtimestamp()',
+        'now()',
+        'statement_timestamp()',
+        'transaction_timestamp()',
     ];
 
     /**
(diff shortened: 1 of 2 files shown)


morozov:
This will require support for default expressions (as opposed to default values). See https://github.com/doctrine/dbal/pull/3967#issuecomment-824215752.

morozov:

PR #4654, opened 2021-05-17
Backport changes from 4.x and 3.x to avoid memory leaks

tzkoshi:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues |  #4515

#### Summary

Backport changes from 3.x and 4.x.  to 2.13.x.
The main idea to copy the changes down to this version is that the official package for the ORM part does not support 3.x yet.
So this fixes the memory leak in the DBAL part.

I'm not very familiar with the codebase, so if there is something more that have to be done - please, let me know.

### DIFF
--- a/lib/Doctrine/DBAL/Connection.php
+++ b/lib/Doctrine/DBAL/Connection.php
@@ -146,13 +146,6 @@ class Connection implements DriverConnection
     private $platform;
 
     /**
-     * The schema manager.
-     *
-     * @var AbstractSchemaManager|null
-     */
-    protected $_schemaManager;
-
-    /**
      * The used DBAL driver.
      *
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Connection.php
```diff
@@ -218,8 +211,6 @@ public function __construct(
         $this->_config       = $config;
         $this->_eventManager = $eventManager;
 
-        $this->_expr = new Query\Expression\ExpressionBuilder($this);
```
morozov, line 221: This would be a BC break (see https://github.com/doctrine/dbal/pull/4516). That's why the issue is fixed only in the next major release.

PR #4655, opened 2021-05-18
Add bind() and binds() methods for more terse user code

aimeos:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Adds bind() and binds() methods to the query builder as alias for setParameter() and setParameters() for shorter user code.

### DIFF
--- a/docs/en/reference/query-builder.rst
+++ b/docs/en/reference/query-builder.rst
@@ -29,7 +29,8 @@ and are therefore subject to the possibility of SQL injection.
 To safely work with the QueryBuilder you should **NEVER** pass user
 input to any of the methods of the QueryBuilder and use the placeholder
 ``?`` or ``:name`` syntax in combination with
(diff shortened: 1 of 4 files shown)


greg0ire:
Hi! No thanks from me, IMO there is no point in just renaming if you don't also change the interface. Also, `binds` does not look like the plural of bind, but like bound conjugated in the 3rd person

aimeos:
The `bind()` method is also part of the PDO object and much less to write if you have a lot of parameters in the statement. The `binds()` method is an suggestion but I can remove it if you have doubts about the naming.

greg0ire:
I have strong doubts about that naming.

> if you have a lot of parameters in the statement

In that kind of case, you would be using `setParameters` (plural), wouldn't you be?

morozov:
Thank you for the patch, @aimeos. I don't think the library will benefit from this API change.

PR #3563, opened 2019-05-28
Allow Symfony 5

nicolas-grekas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

Sibling of https://github.com/doctrine/DoctrineBundle/pull/975
Not strictly required as it doesn't block the Symfony 5 CI. Yet will help ensure Symfony 5 is compatible.

PR #4656, opened 2021-05-19
Allow Symfony 6

jderusse:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Same as #3563
Still not strictly required as it doesn't block the Symfony 6 CI. Yet will help ensure Symfony 6 is compatible

Yes @greg0ire one more pipe

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,8 +45,8 @@
         "phpunit/phpunit": "9.5.0",
         "psalm/plugin-phpunit": "0.13.0",
         "squizlabs/php_codesniffer": "3.6.0",
-        "symfony/cache": "^5.2",
-        "symfony/console": "^2.0.5|^3.0|^4.0|^5.0",
+        "symfony/cache": "^5.2|^6.0",
+        "symfony/console": "^2.0.5|^3.0|^4.0|^5.0|^6.0",
         "vimeo/psalm": "4.6.4"
     },
     "suggest": {


composer.json
```diff
@@ -45,8 +45,8 @@
         "phpunit/phpunit": "9.5.0",
         "psalm/plugin-phpunit": "0.13.0",
         "squizlabs/php_codesniffer": "3.6.0",
-        "symfony/cache": "^5.2",
-        "symfony/console": "^2.0.5|^3.0|^4.0|^5.0",
+        "symfony/cache": "^5.2|^6.0",
+        "symfony/console": "^2.0.5|^3.0|^4.0|^5.0|^6.0",
```
stloyd, line 49:
```suggestion
        "symfony/console": "^3.4|^4.4|^5.2|^6.0",
```
There is no point to support the not maintained versions IMHO.

stloyd (changes requested):

composer.json
jderusse: I don't have opinion about this, but this is not the purpose of this PR, and think that should be discussed in a dedicated PR.

greg0ire (approved):

greg0ire:
I had a look at the CI, and symfony 6 does not seem to be actually installed.

EDIT: silly me that's completely normal since it's not released.

greg0ire:
Thanks @jderusse !

jderusse:
thanks

PR #4659, opened 2021-05-29
Make the comparator API platform aware

ausi:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #2663 #4551

This pull request adds platform awareness to the comparator API as discussed in https://github.com/doctrine/dbal/pull/4551#issuecomment-848495780

I think it works pretty well as shown by the tests. Some tests that previously used code like this:

```php
$diff = (new Comparator())->diffTable($table1, $table2);
self::assertNotFalse($diff);
self::assertEmpty($this->platform->getAlterTableSQL($diff));
```

Could be changed to the following code because the comparator now correctly detects these cases would end up in an empty ALTER TABLE:

```php
$diff = (new Comparator())->diffTable($table1, $table2);
self::assertFalse($diff);
```

One failing test remains: `ComparatorTest::testDiffColumnPlatformOptions` as I was not able to understand what exactly this test expects to happen. With the new comparator design it should be OK to report all platform option changes that have an effect in the resulting SQL code I think.

***

For backwards compatibility I would make the `$platform` argument defaulting to `null` execute the old code for `diffColumn` in this case and triggering a deprecation notice. This should make the upgrade path for users of this library pretty straight forward.


src/Schema/Comparator.php
```diff
@@ -432,95 +434,76 @@ public function diffForeignKey(ForeignKeyConstraint $key1, ForeignKeyConstraint
     /**
      * Returns the difference between the columns
      *
-     * If there are differences this method returns $field2, otherwise the
-     * boolean false.
+     * If there are differences this method returns the changed properties as a
+     * string array, otherwise an empty array gets returned.
```
morozov, line 550: Although this looks right, it's a breaking change which isn't acceptable in a minor release.

src/Schema/Comparator.php
```diff
@@ -432,95 +434,76 @@ public function diffForeignKey(ForeignKeyConstraint $key1, ForeignKeyConstraint
...
+        unset($properties1['name'], $properties2['name']);
 
-        if (get_class($properties1['type']) !== get_class($properties2['type'])) {
-            $changedProperties[] = 'type';
-        }
+        $changedProperties = [];
 
-        foreach (['notnull', 'unsigned', 'autoincrement'] as $property) {
-            if ($properties1[$property] === $properties2[$property]) {
+        foreach (array_unique(array_merge(array_keys($properties1), array_keys($properties2))) as $key) {
```
morozov: It might be simpler to do `array_keys(array_merge($properties1, $properties2))`. This way, there's no need for `array_unique()`.

src/Schema/Comparator.php
```diff
@@ -432,95 +434,76 @@ public function diffForeignKey(ForeignKeyConstraint $key1, ForeignKeyConstraint
...
-        $platformOptions2 = $column2->getPlatformOptions();
-
-        foreach (array_keys(array_intersect_key($platformOptions1, $platformOptions2)) as $key) {
-            if ($properties1[$key] === $properties2[$key]) {
-                continue;
-            }
-
-            $changedProperties[] = $key;
-        }
+        $sqlWithChange    = $platform->getCreateTableSQL(new Table('compare_table', [$column2]));
```
morozov: This looks like overkill. We shouldn't have to create a table to compare columns. If there's no platform API for generating a column SQL, it should be introduced.

src/Schema/Comparator.php
```diff
@@ -432,95 +434,76 @@ public function diffForeignKey(ForeignKeyConstraint $key1, ForeignKeyConstraint
...
+        } elseif ($column1->hasPlatformOption($property) || $column2->hasPlatformOption($property)) {
+            if (! $column1->hasPlatformOption($property)) {
+                $options = $column2WithoutChange->getPlatformOptions();
+                unset($options[$property]);
+                $column2WithoutChange->setPlatformOptions($options);
+            } else {
+                $column2WithoutChange->setPlatformOption($property, $column1->getPlatformOption($property));
             }
+        } else {
+            $column2WithoutChange->{'set' . $property}($column1->{'get' . $property}());
```
morozov: This needs to be reworked.

src/Schema/Comparator.php
```diff
@@ -432,95 +434,76 @@ public function diffForeignKey(ForeignKeyConstraint $key1, ForeignKeyConstraint
...
+        ));
+    }
 
-            if ($properties1['scale'] !== $properties2['scale']) {
-                $changedProperties[] = 'scale';
+    /**
+     * Detects if a changed property has an actual effect on the used platform
+     */
+    private function changedPropertyHasEffect(
+        string $property,
```
morozov: Why do we need to compare columns in the context of a single property? From the migration standpoint, we only need to know whether the columns are effectively equal or not. I believe we could remove some details from the resulting diff in order to not complicate the implementation.

morozov (commented):
Thanks for the patch, @ausi. API wise it looks good but logic-wise we need to work through some details. Please see the comments inline.

morozov:
> [...] what exactly this test expects to happen.

The test expects the comparator to identify a change because the current comparator isn't platform-aware. The new one is, and it understands that the "bar" option doesn't apply to any platform.

> With the new comparator design it should be OK to report all platform option changes that have an effect in the resulting SQL code I think.

Right. But the test needs to be reworked to test a real platform-specific option with a specific platform that supports it.

src/Schema/Comparator.php
ausi: The old method description was wrong, there is no `$field2` and it never returned `false`. I fixed the description, the behavior is the same as before.

src/Schema/Comparator.php
ausi: Done in 5dbb71531556225f189f5bc1cd26009b762d0c84

src/Schema/Comparator.php
ausi: Done in 5dbb71531556225f189f5bc1cd26009b762d0c84.

src/Schema/Comparator.php
ausi: Done in 5dbb71531556225f189f5bc1cd26009b762d0c84.

src/Schema/Comparator.php
ausi: Makes sense. I switched to comparing the SQL only once per column in 5dbb71531556225f189f5bc1cd26009b762d0c84 and it seems to work fine.

ausi:
> But the test needs to be reworked to test a real platform-specific option with a specific platform that supports it.

Done in 5dbb71531556225f189f5bc1cd26009b762d0c84.

> API wise it looks good but logic-wise we need to work through some details.

I think all comments should be addressed now.

Should I proceed with the deprecation layer to make the changed comparator API backwards compatible?

ausi:
> Should I proceed with the deprecation layer to make the changed comparator API backwards compatible?

Done in d2a53b5bbffe32f58db25c3b30e69c861941cc48.

src/Schema/Comparator.php
```diff
@@ -432,12 +455,57 @@ public function diffForeignKey(ForeignKeyConstraint $key1, ForeignKeyConstraint
...
+            );
+
+            return $this->diffColumnDeprecated($column1, $column2);
+        }
+
+        $properties1 = $column1->toArray();
+        $properties2 = $column2->toArray();
+
+        // Do not compare the column names
+        $properties1['name'] = $properties2['name'];
```
morozov: Instead of making the column names equal, you can use some fake (e.g. empty) name in https://github.com/doctrine/dbal/blob/49472c6ff30e795908c876c673708062c7895f97/src/Schema/Comparator.php#L603

src/Schema/Comparator.php
```diff
@@ -523,6 +591,24 @@ public function diffColumn(Column $column1, Column $column2)
         return array_unique($changedProperties);
     }
 
+    /**
+     * @param mixed[] $column
+     */
+    private function generateColumnSQL(array $column, AbstractPlatform $platform): string
```
morozov: Could this method accept a `Column` object instead? The `toArray()` transformation on a column just mixes all the properties together, so I'd avoid using it unless absolutely necessary.

src/Schema/Comparator.php
```diff
@@ -523,6 +591,24 @@ public function diffColumn(Column $column1, Column $column2)
...
+    {
+        if ($platform->isCommentedDoctrineType($column['type'])) {
+            $column['comment'] .= $platform->getDoctrineTypeComment($column['type']);
+        }
+
+        $sql = $platform->getColumnDeclarationSQL($column['name'], $column);
+
+        if (! $platform->supportsInlineColumnComments()) {
+            $sql .= ' -- ' . $column['comment'];
```
morozov: This whole method looks like a copy/paste from somewhere else. Even if not, this logic doesn't belong to the comparator. Please move it under the Platform API.

src/Schema/Comparator.php
```diff
@@ -432,12 +455,57 @@ public function diffForeignKey(ForeignKeyConstraint $key1, ForeignKeyConstraint
...
+                && array_key_exists($key, $properties2)
+                && $properties1[$key] === $properties2[$key]
+            ) {
+                continue;
+            }
+
+            $changedProperties[] = $key;
+        }
+
+        return $changedProperties;
```
morozov: I believe this should be deprecated: the DDL is produced because the two columns have different resulting SQL, not as a result of column property comparison. If the calling side is interested in a property-based diff, it can compare the two columns.

tests/Functional/Schema/ComparatorTest.php
```diff
@@ -39,7 +39,9 @@ public function testDefaultValueComparison(string $type, $value): void
 
         $onlineTable = $this->schemaManager->listTableDetails('default_value');
 
-        self::assertFalse($this->comparator->diffTable($table, $onlineTable));
+        self::assertFalse(
+            $this->comparator->diffTable($table, $onlineTable, $this->schemaManager->getDatabasePlatform())
```
morozov: Changes like this effectively stop testing the old implementation. We need to come up with a way of testing both. E.g. each test that should pass with both the new and the old implementations could be duplicated. Later, we can remove the old versions.

tests/Platforms/AbstractMySQLPlatformTestCase.php
```diff
@@ -808,6 +807,108 @@ protected function getQuotedAlterTableChangeColumnLengthSQL(): array
...
+        $fromTable->addColumn('blob1', 'blob', ['length' => MySQLPlatform::LENGTH_LIMIT_TINYBLOB]);
+        $fromTable->addColumn('blob2', 'blob', ['length' => MySQLPlatform::LENGTH_LIMIT_BLOB]);
+        $fromTable->addColumn('blob3', 'blob', ['length' => MySQLPlatform::LENGTH_LIMIT_MEDIUMBLOB]);
+        $fromTable->addColumn('blob4', 'blob', ['length' => MySQLPlatform::LENGTH_LIMIT_MEDIUMBLOB * 2]);
+        $fromTable->addColumn('blob5', 'blob');
+        $fromTable->addColumn('blob6', 'blob', ['length' => MySQLPlatform::LENGTH_LIMIT_TINYBLOB]);
+        $fromTable->addColumn('blob7', 'blob', ['length' => MySQLPlatform::LENGTH_LIMIT_BLOB]);
+        $fromTable->addColumn('blob8', 'blob', ['length' => MySQLPlatform::LENGTH_LIMIT_MEDIUMBLOB]);
+        $fromTable->addColumn('blob9', 'blob', ['length' => MySQLPlatform::LENGTH_LIMIT_MEDIUMBLOB * 2]);
+        $fromTable->addColumn('blob10', 'blob');
```
morozov: This will be really hard to maintain. As far as I understand, the test verifies that if a BLOB column length exceeds a certain threshold, the column type is changed. It could be done with a much simpler test and a data provider: i.e. declare one column of a given type, increase the length and assert that there is a diff. Repeat (via the data provider) for each size.

tests/Platforms/AbstractMySQLPlatformTestCase.php
```diff
@@ -808,6 +807,108 @@ protected function getQuotedAlterTableChangeColumnLengthSQL(): array
...
+        self::assertEquals(
+            $this->getAlterTableChangeBlobColumnLength(),
+            $this->platform->getAlterTableSQL($diff)
+        );
+    }
+
+    /**
+     * @return string[]
+     */
+    protected function getAlterTableChangeBlobColumnLength(): array
```
morozov: Why is a separate method needed?

tests/Platforms/AbstractMySQLPlatformTestCase.php
```diff
@@ -808,6 +807,108 @@ protected function getQuotedAlterTableChangeColumnLengthSQL(): array
...
+        $toTable->addColumn('blob3', 'blob', ['length' => MySQLPlatform::LENGTH_LIMIT_MEDIUMBLOB * 2]);
+        $toTable->addColumn('blob4', 'blob', ['length' => MySQLPlatform::LENGTH_LIMIT_TINYBLOB]);
+        $toTable->addColumn('blob5', 'blob', ['length' => MySQLPlatform::LENGTH_LIMIT_BLOB]);
+        $toTable->addColumn('blob6', 'blob', ['length' => MySQLPlatform::LENGTH_LIMIT_TINYBLOB]);
+        $toTable->addColumn('blob7', 'blob', ['length' => MySQLPlatform::LENGTH_LIMIT_BLOB]);
+        $toTable->addColumn('blob8', 'blob', ['length' => MySQLPlatform::LENGTH_LIMIT_MEDIUMBLOB]);
+        $toTable->addColumn('blob9', 'blob', ['length' => MySQLPlatform::LENGTH_LIMIT_MEDIUMBLOB * 2]);
+        $toTable->addColumn('blob10', 'blob');
+
+        $diff = (new Comparator())->diffTable($fromTable, $toTable, $this->platform);
```
morozov: It doesn't look like a platform test should involve the comparator. The only responsibility of the platform is to generate DDL.

tests/Schema/ComparatorMySQLTest.php
```diff
@@ -0,0 +1,107 @@
...
+class ComparatorMySQLTest extends ComparatorTest
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->platform = new MySQLPlatform();
+    }
+
+    public function testCompareChangedBlobColumn(): void
```
morozov, line 23: How is this test different from `testAlterTableChangeBlobColumnLength()`?

src/Schema/Comparator.php
```diff
@@ -199,8 +212,18 @@ public function diffSequence(Sequence $sequence1, Sequence $sequence2)
      *
      * @throws SchemaException
      */
-    public function diffTable(Table $fromTable, Table $toTable)
+    public function diffTable(Table $fromTable, Table $toTable, ?AbstractPlatform $platform = null)
```
morozov: I believe we could use this as an opportunity to introduce better method names. `diffTable()` could be renamed to `compareTables()`. This way, we won't have to introduce an optional parameter but could introduce the whole old method. The same applies to the rest of the methods except for `compareSchemas()` which already exists.

morozov:
@ausi, I linked a few pull requests that will be superseded by this one once it's done. I believe there are at least as many open issues (to be identified), so it appears to be a quite important improvement. No pressure, just FYI 😅

src/Schema/Comparator.php
trompette: Yesterday I openned https://github.com/doctrine/dbal/pull/4707 to introduce a more consistent Comparator API.

But as @ausi commented, it will make merging this PR harder, mainly because of the optional `$platform` parameter on all public methods.

After sleeping on it, I think injecting the platform as a constructor parameter when creating a Comparator object would help with avoiding conflcts and would make sense design wise: it seems comparing schemas depends on the platform, given all the linked pull requests.

Actually, it is precisely for enabling this type of API evolution that https://github.com/doctrine/dbal/pull/4707 deprecates static calls of `compareSchemas()`, as stated in https://github.com/doctrine/dbal/pull/4704#discussion_r667042163.

What do you think ?

morozov: > But as @ausi commented, it will make merging this PR harder, mainly because of the optional `$platform` parameter on all public methods.

Let's not worry about that. I can help resolve all the conflicts once they occur.

> I think injecting the platform as a constructor parameter when creating a Comparator object would help with avoiding conflcts and would make sense design wise: it seems comparing schemas depends on the platform, given all the linked pull requests.

Whether it's injected in the constructor or passed as an argument, it's the same design-wise: the comparator does depend on the platform. The question is, whether or not a given comparator instance needs be usable with different platforms, and the answer is "probably not". I think we can move the platform dependency a constructor argument as originally proposed.

tests/Platforms/AbstractMySQLPlatformTestCase.php
ausi: This is already the case in https://github.com/doctrine/dbal/blob/25496429ef4bba2352b0ced7daffe1b3dec83d13/tests/Platforms/AbstractPlatformTestCase.php#L965
We want to test if the comparator works correctly with every platform, so I think it is the right place here.

tests/Schema/ComparatorMySQLTest.php
ausi: It checks if the changedProperties get reported correctly. But as we now want to change this to only return `true` or `false` I could remove the `testAlterTableChangeBlobColumnLength()` from the platform test I think.

src/Schema/Comparator.php
ausi: Done in 518a43335354dc75d3bd470faf7f5fffa4f0fe3f

src/Schema/Comparator.php
ausi: Done, the method is not needed anymore.

src/Schema/Comparator.php
ausi: Done, added `getColumnSQL()` to `AbstractPlatform` in af6e02b13d8cc8b5c03baee4277727b7c0f5dbfb

src/Schema/Comparator.php
ausi: Done in 5cb69bcdae6348566f77824b692828e3b75f9461

tests/Functional/Schema/ComparatorTest.php
ausi: Done in a0e28045b36445bdfb14d21fb044718a9d2dd7a1

tests/Platforms/AbstractMySQLPlatformTestCase.php
ausi: I removed this tests now as they did not make much sense this way I think.

tests/Schema/ComparatorMySQLTest.php
ausi: Fixed now as the other tests got removed.

src/Schema/Comparator.php
ausi: The `compareColumns()` method now returns `bool` in this pull request. This would conflict with #4707 (but only if there would actually be a version release between the two pull requests).

ausi:
@morozov I think I addressed all comments now. This should be ready for another review 🎉

ausi:
Some tests seem to have problems with the `platformOptions` now if they come from the live database. E.g. https://github.com/doctrine/dbal/runs/3040652078?check_suite_focus=true#step:7:115

AFAIK this is due to charset and collation being reported even if they use the default.

For example, offline schema:

```
'platformOptions' => [],
```

online schema:

```
'platformOptions' => [
    'charset' => 'utf8',
    'collation' => 'utf8_unicode_ci',
]
```

I’m not sure how to best fix this with the concept of doctrine, as the `compareColumns()` method doesn’t know about the defaults of a specific table.

One possibility would be that the schema manager stores the information about which platformOptions don’t differ in the `Column` object. Like an array of platform options that match the default value. This would make it possible for the Comparator to compare the meaning of these options.

morozov:
Why did this issue surface only now? As far as I see from debugging the test, when the unexpected diff is produced, the columns are still compared on a per-property basis. This is how it already works.

> One possibility would be that the schema manager stores the information about which platformOptions don’t differ in the Column object.

I believe it's a bit risky. Omitted values on the defining side mean that the defaults on the database should kick in. If an online column has different properties, the comparison should result in a diff but it won't which is a bug.

What we could do instead is explicitly add the defaults to the DBAL version of the column before comparison. They are currently implemented in: https://github.com/doctrine/dbal/blob/a602f2baf74df3d9b17ea8b29fca5fc33786d68d/src/Platforms/MySQLPlatform.php#L480-L485

We'll also need to double-check how it will work when the defaults are omitted (see https://github.com/doctrine/dbal/pull/4644).

ausi:
> Why did this issue surface only now?

Currently doctrine only compares platform options that exist in both `Column` instances, see https://github.com/doctrine/dbal/blob/25496429ef4bba2352b0ced7daffe1b3dec83d13/src/Schema/Comparator.php#L512-L521

But this is wrong IMO as it would not detect changes from a different charset back to the default for example.

> > One possibility would be that the schema manager stores the information about which platformOptions don’t differ in the Column object.
> 
> I believe it's a bit risky. Omitted values on the defining side mean that the defaults on the database should kick in.

No, I meant explicitly storing information about which platform options are set to the default of the table. This way the comparator can compare the platform options and if a defining side option is missing it can check if the database side option is default or not.

> What we could do instead is explicitly add the defaults to the DBAL version of the column before comparison. They are currently implemented in:
> 
> https://github.com/doctrine/dbal/blob/a602f2baf74df3d9b17ea8b29fca5fc33786d68d/src/Platforms/MySQLPlatform.php#L480-L485

These are the table options, not the column options AFAIK. The charset/collation of the columns are set to the `TABLE_COLLATION` of the accociated table by default I think.

morozov:
> But this is wrong IMO as it would not detect changes from a different charset back to the default for example.

I agree. It would be nice to reproduce this issue separately by writing a test for it.

> No, I meant explicitly storing information about which platform options are set to the default of the table.

Where can we store this information? Keep in mind that the DBAL should be able to introspect and compare the objects created by other clients.

> These are the table options, not the column options AFAIK.

Yes, but eventually there will be the same problem: if we compare the DDL of the table generated from its definition in the DBAL and the DDL generated from the introspected data, there may be a mismatch due to some implied defaults. At least when the defaults are removed from the DBAL in https://github.com/doctrine/dbal/pull/4644.

> The charset/collation of the columns are set to the `TABLE_COLLATION` of the accociated table by default I think.

Yes, I think in order to compare the columns properly, we need to introspect the deployed table first. I.e.:
1. Introspect the needed properties of the table.
2. Augment the definition on the DBAL side with the defaults.
3. Compare. At this point, all properties should be explicitly set on both sides of the comparison.

ausi:
I tried out the concept now as you can see in the last two commits. It works pretty well but there is one major issue: the default for the `collation` depends on the `charset` option. So only changing the `charset` without specifying the `collation` would still result in wrongly detected diffs between the online and offline table.

> > But this is wrong IMO as it would not detect changes from a different charset back to the default for example.
> 
> I agree. It would be nice to reproduce this issue separately by writing a test for it.

Done in 4c061bbf7be53760abde1d0a44cd74e4788ac875

> > No, I meant explicitly storing information about which platform options are set to the default of the table.
> 
> Where can we store this information?

I stored it in the `Column` objects now in 8fad74365ef62218329fabd0b8918da620ff5f3a

***

I think it could make sense to let the abstraction gain knowledge about charset and collation to handle these cases better, instead of just putting them in the `platformOptions` array. But I don’t have experience in the other database systems, so I don’t know if there is a system that all databases agree on there.

src/Platforms/AbstractPlatform.php
```diff
@@ -2254,6 +2266,13 @@ public function getColumnDeclarationSQL($name, array $column)
         return $name . ' ' . $declaration;
     }
 
+    public function getColumnSQL(string $name, Column $column): string
```
morozov, line 2269: From the purpose perspective, how is it different from the existing `getColumnDeclarationSQL()`? I'd rather add the support for `Column $column` to the existing method.

src/Platforms/MySQLPlatform.php
```diff
@@ -361,7 +361,19 @@ public function getListTableColumnsSQL($table, $database = null)
...
-               'CHARACTER_SET_NAME AS CharacterSet, COLLATION_NAME AS Collation ' .
+               'CHARACTER_SET_NAME AS CharacterSet, COLLATION_NAME AS Collation, ' .
+               'IFNULL(( ' .
+                   'SELECT ( ' .
+                       'SELECT CHARACTER_SET_NAME ' .
+                       'FROM information_schema.COLLATION_CHARACTER_SET_APPLICABILITY ' .
+                       'WHERE COLLATION_NAME = TABLES.TABLE_COLLATION ' .
+                   ') FROM information_schema.TABLES ' .
+                   'WHERE TABLE_SCHEMA = ' . $database . ' AND TABLE_NAME = ' . $table .
+               '), @@character_set_database) AS DefaultCharacterSet, ' .
```
morozov, line 372: If I read it right, it means that if the charset is not specified on the column, the query will return the default database charset. Does the database work this way?

I.e. if I create a column and don't specify the charset, will it be introspected as NULL? And furthermore, if I change the default charset, will it affect all columns with the default charset? That would be weird. I'd expected it to be _copied_ to the column metadata during creation or maybe modification.

src/Schema/Column.php
```diff
@@ -44,6 +44,9 @@ class Column extends AbstractAsset
     /** @var mixed[] */
     protected $_platformOptions = [];
 
+    /** @var mixed[] */
+    protected $_platformDefaults = [];
```
morozov, line 48: It doesn't look like it belongs here. A column is a value object representing the specification defined by the user or introspected from the database. I'd say it's the responsibility of the platform to represent its defaults if needed for comparison.

src/Platforms/AbstractPlatform.php
```diff
@@ -1616,6 +1606,28 @@ public function getCreateTableSQL(Table $table, $createFlags = self::CREATE_INDE
         return array_merge($sql, $columnSql);
     }
 
+    /**
+     * @return mixed[] An associative array with the name of the properties
+     *                 of the column being declared as array indexes.
+     *                 See getColumnDeclarationSQL() for more information.
+     */
+    protected function _getColumnArrayFromColumn(Column $column): array
```
morozov, line 1614: It's better to declare it private. Also, please do not use the leading underscore for new methods.

src/Schema/Comparator.php
```diff
@@ -228,12 +248,28 @@ public function diffTable(Table $fromTable, Table $toTable)
...
+            if ($this->platform === null) {
+                $changedProperties = $this->diffColumn($column, $toTable->getColumn($columnName));
+            } else {
+                $properties1 = $column->toArray();
+                $properties2 = $toTable->getColumn($columnName)->toArray();
+
+                $changedProperties = [];
+
+                foreach (array_keys(array_merge($properties1, $properties2)) as $key) {
+                    if (($properties1[$key] ?? null) === ($properties2[$key] ?? null)) {
```
morozov, line 265: Why do we still need to do a property-based comparison while the platform is available? Note that the strict comparison alone will trigger false-positive diffs.

tests/Functional/Schema/ComparatorTest.php
```diff
@@ -22,7 +22,7 @@ protected function setUp(): void
         parent::setUp();
 
         $this->schemaManager = $this->connection->getSchemaManager();
-        $this->comparator    = new Comparator();
+        $this->comparator    = new Comparator($this->schemaManager->getDatabasePlatform());
```
morozov, line 25: Please do not modify existing tests. All existing tests should pass without changes.

src/Schema/Comparator.php
```diff
@@ -429,15 +465,95 @@ public function diffForeignKey(ForeignKeyConstraint $key1, ForeignKeyConstraint
...
-     * boolean false.
+     * If there are differences this method returns the changed properties as a
+     * string array, otherwise an empty array gets returned.
+     *
+     * @deprecated
      *
      * @return string[]
      */
-    public function diffColumn(Column $column1, Column $column2)
+    public function diffColumn(Column $column1, Column $column2): array
```
morozov, line 556: Enforcing the return type on a method is a breaking change, please revert.

src/Schema/Comparator.php
```diff
@@ -429,15 +465,95 @@ public function diffForeignKey(ForeignKeyConstraint $key1, ForeignKeyConstraint
...
+            );
+    }
+
+    /**
+     * Remove platform options from both columns if they match the default from
+     * the table or database.
+     *
+     * @return Column[]
+     */
+    private function stripDefaultValuesFromColumns(Column $column1, Column $column2): array
```
morozov, line 507: Stripping properties based on their value may be unnecessarily complex and error-prone. I'd rather use the opposite approach and augment the definition provided by the application with the database defaults.

morozov (commented):
> I think it could make sense to let the abstraction gain knowledge about charset and collation to handle these cases better, instead of just putting them in the `platformOptions` array.

I don't think we're there yet. I see quite some bits in the PR that I don't understand and I don't understand the overall design either. For instance, the property-based comparison in the presence of a specific platform goes again the entire idea of comparing SQL to SQL.

I don't think any platform-specific column parameters should be promoted to the DBAL: handling charsets is done differently in different platforms, and there's no way it could be properly described in an abstract API. It should be the opposite: we expose the platform to the comparator, so if any platform-specific comparison needs to be done, it should delegate it to the platform.

Please let me know if you need help rebasing the PR. It might help if you squash all the commits. This way, the least possible number of merge conflicts will have to be resolved.

src/Schema/Comparator.php
ausi: > augment the definition provided by the application with the database defaults

At which point would you do that? The column doesn’t store a reference to the table nor to the database connection.

morozov: Good question. Regardless of whether we strip the defaults or augment the definition, we'll need the DB connection to understand what the omitted options default to. But at the same time, we want to be able to compare pure definitions, w/o any runtime component.

Even if we keep the approach of stripping the defaults, false positives will be possible. E.g. the offline column has its charset explicitly set to the default value, but the online column has it stripped from the definition. Comparing the two will yield a diff, however, it shouldn't.

morozov: > But at the same time, we want to be able to compare pure definitions, w/o any runtime component.

It seems to be impossible by definition: whether or not the two definitions are different depends on the runtime. Instead of making the comparator dependent on the platform, it may need to depend on the schema manager which can use both the platform and the connection to normalize the offline column definition.

ausi: > Even if we keep the approach of stripping the defaults, false positives will be possible. E.g. the offline column has its charset explicitly set to the default value, but the online column has it stripped from the definition. Comparing the two will yield a diff, however, it shouldn't.

That case is already handled by my current implementation. If one or both columns have platform defaults, these defaults are stripped from both columns (online and offline). This way all comparisons should be possible (offline to offline, online to offline and online to online). If two online columns are compared and they have different platform defaults my current implementation does not strip the options.

morozov:
FWIW, I'm close to finishing https://github.com/doctrine/dbal/pull/4746 which takes the approach from here and reimplements this w/o introducing the undesired API changes. Just need to implement a comparator for SQL Server and decide on the way to rework unit tests which don't have access to a database connection (needed to determine the default database collation for SQL Server).

morozov:
Closing in favor of #4746.

Issue #4660, opened 2021-06-03
Array of parameters are not resolved properly in query parameters


VincentChalnot:
### Bug Report

See https://github.com/symfony/symfony/issues/41526 for further information.

|    Q        |   A
| ------------ | ------
| BC Break    | ?
| Version     | 2.13.1

#### Summary

When passing a parameter in the form of an array, there is no way that the values of the array could get resolved properly.

#### Current behaviour

During a query, when passing a parameter in the form of an array, `Doctrine\DBAL\SQLParserUtils::expandListParameters` expects the type to be an integer (`Doctrine\DBAL\Connection::PARAM_STR_ARRAY`) to properly expands the list of parameters for the connection layer BUT later on, `Doctrine\DBAL\Connection::getBindingInfo` expects the type to be an object to resolve values properly to their database representation.

So their is no way an array of parameters could get resolved properly because `Doctrine\DBAL\Types\Type::convertToDatabaseValue` will never be called.

#### How to reproduce

Super simple test case with native datetime type:
https://github.com/VincentChalnot/dbal/commit/77e91f56fa476b8c73b82c7ef601b1a92cedd0ad#diff-ff5f64e24b58f2f4b6feacbc55c20840fb0155b1d4db9f28823f4c49bd562ee9R97

An other way to reproduce this issue is to use Doctrine/ORM and a type that requires a database conversion like a Uuid:
https://github.com/symfony/symfony/issues/41526

#### Expected behaviour

`Doctrine\DBAL\Connection::getBindingInfo` should be called first, be able to deal with array of parameters and then only resolve the type to an integer for `Doctrine\DBAL\SQLParserUtils::expandListParameters`.

VincentChalnot:
Note that once this behavior is fixed, doctrine/orm will also require a fix for the example to work properly.

morozov:
@VincentChalnot please reproduce the issue using the DBAL APIs to make sure that it's indeed a DBAL issue.

VincentChalnot:
@morozov I'm writing a test case in a fork of doctrine/dbal in order to highlight this issue, however I'm not very familiar with the way doctrine/dbal works without doctrine/orm, for example:

- I'm writing a object value to the database using a custom type: check
- I'm fetching the value of this object using Connection::fetchOne but the convertToPHPValue is never called.

I'm guessing this is normal but it also seems weird to me.

It feels like custom types support was not really meant to work inside doctrine/dbal without doctrine/orm to do the other part of the job.

[...]

Issue #4602, opened 2021-04-13
SQLLogger does not log actual params used in query


Wirone:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.12.1

#### Summary

There are scenarios when params logged by `SQLLogger` are not the same as params actually used in query. It's because query, params and types are logged before preparing statement, in which params are bound (and while binding they can be typecasted which ends with different value than the one which was logged).

Initially reported in [Symfony](https://github.com/symfony/symfony/issues/40566), after discussion in [DoctrineBundle](https://github.com/doctrine/DoctrineBundle/issues/1312), then here after @ostrolucky's request.

#### Current behaviour

When using query builder with something like:
```
$qb->andWhere("foo.value IN (:foo_values)");
$qb->setParameter("foo_values", ['bar'], Connection::PARAM_INT_ARRAY);
```
Doctrine data collector in Symfony profiler shows "runnable query" with `foo.values IN ('bar')` but actually database query contains `foo.values IN (0)` because of `Connection::PARAM_INT_ARRAY` param type. It is very misleading, because looking at the profiler you may think the query is correct (even though the result is not). Profiler displays data from collector, which works on top of `SQLLogger`, so DBAL is the source of query, params and types available later in the profiler.

More context can be found in [`doctrine/DoctrineBundle#1312`](https://github.com/doctrine/DoctrineBundle/issues/1312#issuecomment-809225933) 

#### How to reproduce

Here's reproducer created by @Nyholm → https://github.com/Nyholm/sf-issue-40566

#### Expected behaviour

In the end user should know what params were actually used in query because it would allow to find bugs faster. Personally I've lost lot of time because I've seen in profiler wrong "runnable query", which was returning different set of values than actual query in the app. I had to enable and watch general log in MySQL to see what's actually executed and then was able to find source of the problem.

beberlei:
My opinion is this is a DoctrineBundle issue, because it wants to make sure the query is runnable and does something with the params. Yes it would reduce redudandcy to to it in DBAL, but it also causes overhead of converting when maybe the logger is not interested in this. As such i am :-1: here.

It would also be a BC break. Yes we don't change the signature, but we would change the expected behavior, maybe other loggers are already converting `$params` based on `$types`? This is a bug in DoctrineBundle.

ostrolucky:
[...]

PR #4661, opened 2021-06-05
Resolving array of parameters properly in query parameters

VincentChalnot:
|      Q       |   A
|------------- | -----------
| Type         | bugfix
| BC Break     | no
| Fixed issues | #4660 

#### Summary

Adding new ArrayOf type to handle array of parameters properly.

We also need a fix in Doctrine/ORM.

### DIFF
--- a/lib/Doctrine/DBAL/Types/ArrayOf.php
+++ b/lib/Doctrine/DBAL/Types/ArrayOf.php
@@ -1,0 +1,31 @@
+<?php
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Types;
+
+use Doctrine\DBAL\Connection;
+
+/**
+ * Type representing an array of an other type, not an actual type even though it should be but that would require to
+ * rewrite too much code because types are not allowed to rewrite the query and replace its parameters.
+ */
+class ArrayOf
+{
+    /** @var Type */
+    private $type;
+
+    public function __construct(Type $type)
+    {
+        $this->type = $type;
+    }
+
+    public static function requiresParametersExpansion($type): bool
+    {
+        return $type === Connection::PARAM_INT_ARRAY || $type === Connection::PARAM_STR_ARRAY || $type instanceof self;
+    }
+
+    public function getType(): Type
+    {
+        return $this->type;
+    }
+}
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -53,3 +53,9 @@ jobs:
         with:
           composer_require_dev: true
           args: --shepherd
+
+      - name: Psalm type inference tests
+        uses: docker://vimeo/psalm-github-actions:4.6.4
+        with:
+          composer_require_dev: true
+          args: --config=psalm-strict.xml
--- a/docs/en/reference/caching.rst
+++ b/docs/en/reference/caching.rst
@@ -1,10 +1,22 @@
 Caching
 =======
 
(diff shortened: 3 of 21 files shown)


lib/Doctrine/DBAL/Types/ArrayOf.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Doctrine\DBAL\Types;
+
+use Doctrine\DBAL\Connection;
+
+/**
+ * Type representing an array of an other type, not an actual type even though it should be but that would require to
+ * rewrite too much code because types are not allowed to rewrite the query and replace its parameters.
+ */
+class ArrayOf
```
morozov, line 12: I don't think it's a good idea. This way, the DBAL API still won't enforce the argument types at the language level. A better approach might be to make the `bind*` methods typed. See https://github.com/doctrine/dbal/issues/4602#issuecomment-826979381.

For this specific issue, we can just introduce `bindArray()` with the same signature as `bindValue()`:
https://github.com/doctrine/dbal/blob/de7f86d1ce98d32cb1c3cc36703b3182e44f74d7/lib/Doctrine/DBAL/Statement.php#L101

The only difference will be that `$value` will always have to be an array.

morozov:
Since this is a new feature, it will have to target `3.2.x`.

lib/Doctrine/DBAL/Types/ArrayOf.php
VincentChalnot: I agree with you, this should be done this way in a future release, typically the 3.2.x.
But I think my approach solves the problem by introducing absolutely no BC-break because it keeps the existing parameter expansion logic.

It could also be done without a new object, typically by just using a naming convention like `datetime[]` and testing the existence of `[]` at the end of the type but I think a new wrapper class is simple and better.

VincentChalnot: > Since this is a new feature, it will have to target `3.2.x`.

I'm sorry but I have to disagree with you, this is clearly a bugfix although it introduces a new "type" it allows to fix an issue that was here since the beginning of this project and introduces no BC-break.

That being said, my initial goal was to fix this issue in doctrine/orm but now that I saw the way the UnitOfWork deals with identifiers (imploding them like they're all strings) and more generally with types in queries, I must say I'm not eager to touch anything in this mess. So if you still want to rebase this in 3.2.x I'm fine with that.

I've already rolled back to not using custom object types in my projects anyway.

VincentChalnot:
> Since this is a new feature, it will have to target `3.2.x`.

I'm sorry but I have to disagree with you, this is clearly a bugfix although it introduces a new "type" it allows to fix an issue that was here since the beginning of this project and introduces no BC-break.

That being said, my initial goal was to fix this issue in doctrine/orm but now that I saw the way the UnitOfWork deals with identifiers (imploding them like they're all strings) and more generally with types in queries, I must say I'm not eager to touch anything in this mess. So if you still want to rebase this in 3.2.x I'm fine with that.

I've already rolled back to not using custom object types in my projects anyway.

morozov:
> I'm not eager to touch anything in this mess.

Thank you for contributing the PoC. Closing for now. At a later point in time, we may address this issue in a way I described earlier.

VincentChalnot:
Just to clarify, I was commenting the code of the UnitOfWork, I'm mostly fine with doctrine/dbal.
Il will try to create a new PR with the approach you suggested when I have some time to spare.

PR #4638, opened 2021-05-06
Fix #4637 by duplicating the type definition for `DriverManager::getConnection($args)` params

Ocramius:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #4637

Fixes #4637

As per #4637, `vimeo/psalm:4.6.4` cannot introduce templated parameters within a re-used imported
type: this is understandable, since imported types are copy-paste aided by the type checker, but do
not carry any template information with them (and it would be too complex to do so anyway).

Quoting the initial issue:

Minimal reproducer: https://psalm.dev/r/62a0a5854d

```php
<?php

class Connection {}
class MyConnection extends Connection{}

/**
 * @psalm-type Params = array{
 *     charset?: string,
 *     wrapperClass?: class-string<Connection>,
 * }
 */
final class DriverManager
{
    /**
     * @param array<string,mixed> $params
     * @param Configuration|null  $config       The configuration to use.
     * @param EventManager|null   $eventManager The event manager to use.
     *
     * @phpstan-param array<string,mixed> $params
     * @psalm-param Params $params
     * @psalm-return ($params is array{wrapperClass:mixed} ? T : Connection)
     * @template T of Connection
     */
    public static function getConnection(
        array $params
    ): Connection {
        throw new \Exception('irrelevant');
    }
}

function mkConnection(): MyConnection {
    return DriverManager::getConnection(['wrapperClass' => MyConnection::class]);
}
```

Which produces:

```
Psalm output (using commit aa854ae):

INFO: MixedInferredReturnType - 31:26 - Could not verify return type 'MyConnection' for mkConnection
```

The issue is that the `Param` alias type is not templated: this is potentially a psalm limitation, but also an understandable one, since there is no syntax to specify template arguments for an imported type (AFAIK).

Here is a potential fix:

https://psalm.dev/r/8ef9909bb9

```php
<?php

class Connection {}
class MyConnection extends Connection{}

/**
 * @psalm-type Params = array{
 *     charset?: string,
 *     wrapperClass?: class-string<Connection>,
 * }
 */
final class DriverManager
{
    /**
     * @psalm-param array{
     *     charset?: string,
     *     wrapperClass?: class-string<T>,
     * } $params
     * @psalm-return ($params is array{wrapperClass: mixed} ? T : Connection)
     * @template T of Connection
     */
    public static function getConnection(
        array $params
    ): Connection {
        throw new \Exception('irrelevant');
    }
}

[...]

PR #4666, opened 2021-06-05
Merge 2.13.x into 3.1.x

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -53,3 +53,9 @@ jobs:
         with:
           composer_require_dev: true
           args: --shepherd
+
+      - name: Psalm type inference tests
+        uses: docker://vimeo/psalm-github-actions:4.6.4
+        with:
+          composer_require_dev: true
+          args: --config=psalm-strict.xml
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         "doctrine/event-manager": "^1.0"
     },
     "require-dev": {
-        "doctrine/coding-standard": "8.2.0",
+        "doctrine/coding-standard": "9.0.0",
         "jetbrains/phpstorm-stubs": "2020.2",
         "phpstan/phpstan": "0.12.81",
         "phpunit/phpunit": "^7.5.20|^8.5|9.5.0",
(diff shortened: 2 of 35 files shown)


derrabus:
I don't know how to merge up #4638 properly since the directory structure inside the tests folder has changed.

cc @Ocramius

Ocramius:
Test dirs were added AFAIK: everything else should (in theory) have been
handled by `git mv`

derrabus:
In your PR, you've added a file `tests/Doctrine/StaticAnalysis/DBAL/driver-manager-retrieves-correct-connection-type.php`. In 3.0, the unit tests have been moved from `tests/Doctrine/Tests` to `tests`. That new file of yours felt a bit misplaced inside the `tests` folder now, so I've moved it into the `ci` folder.

greg0ire:
:thinking: I think `tests` would have been better actually… `ci` is for files that will only ever be useful in the CI, isn't it @morozov ? Maybe there should be a new directory called `static-analysis` or `sa`?

derrabus:
I don't have strong opinions here. Just tell me where to move that file. 🙂 

Alternatively, this could be done in a follow-up after the merge.

greg0ire:
Since the file is not suffixed with `Test`, it will not be picked up by PHPUnit. I think you can simply move them back to `tests`, and then in a follow-up we can rework the directory structure.

derrabus:
Yeah, but CodeSniffer will pick it up and complain when I move it there. I'll look into it later today.

morozov:
> 🤔 I think `tests` would have been better actually… `ci` is for files that will only ever be useful in the CI, isn't it @morozov ? Maybe there should be a new directory called `static-analysis` or `sa`?

`static-analysis` sounds good.

derrabus:
> `static-analysis` sounds good.

`static-analysis` it is then.

greg0ire (dismissed):

SenseException (dismissed):

morozov:
@derrabus, please redo. It looks like the merge commit doesn't include the current head of `3.1.x`.

derrabus:
Done.

morozov (approved):

morozov:
Thanks!

PR #4675, opened 2021-06-17
The sequence prevents saving with defined id

k0n3r:
The deleted line is executing the sequence therefore the subsequent code is no longer necessary,
If you look at line 534, the same line is deleted

this same error is in the previous version (2.13.1)

|      Q       |   A
|------------- | -----------
| BC Break     | yes/no
| Fixed issues |


morozov:
Thanks for the patch, @k0n3r. Please provide a functional test that reproduces the problem and fix the failing unit test.

morozov:

PR #4677, opened 2021-06-18
Create tables with unique constraints on PostgreSQL

hschletz:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

PostgreSQL94Platform::_getCreateTableSQL() ignored unique constraints.
Implementations for other platforms are OK.

### DIFF
--- a/src/Platforms/PostgreSQL94Platform.php
+++ b/src/Platforms/PostgreSQL94Platform.php
@@ -785,6 +785,12 @@ SQL
             }
         }
 
+        if (isset($options['uniqueConstraints'])) {
+            foreach ($options['uniqueConstraints'] as $uniqueConstraint) {
+                $sql[] = $this->getCreateConstraintSQL($uniqueConstraint, $name);
+            }
+        }
+
         if (isset($options['foreignKeys'])) {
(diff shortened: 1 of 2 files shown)


src/Platforms/PostgreSQL94Platform.php
```diff
@@ -785,6 +785,12 @@ protected function _getCreateTableSQL($name, array $columns, array $options = []
             }
         }
 
+        if (isset($options['uniqueConstraints']) && ! empty($options['uniqueConstraints'])) {
```
morozov: The `!empty()` condition is redundant. If it's empty, then the `foreach` will result in no-op which is fine.

morozov (commented):
Looks good. Just a tiny nitpick inline.

morozov:
The SQL Server build failure should be fixed after merging and up-merging https://github.com/doctrine/dbal/pull/4694. Otherwise, looks good.

morozov (approved):

morozov:
Thanks, @hschletz.

PR #4681, opened 2021-06-18
Deprecate TableGenerator

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

I propose to deprecate and later remove the `TableGenerator` component for the following reasons:
1. It doesn't seem to be used by open-source projects.
2. It's one of the components that use `Connection::getParams()` which I'd like to get rid of eventually (https://github.com/doctrine/dbal/issues/3593).
3. This code hasn't received any significant changes since its introduction in 2012 (https://github.com/doctrine/dbal/commit/468af759a58c88ee28fcf588fcbbd1bcae26f20e). This is either because it's perfect or because nobody uses it.

### DIFF
--- a/psalm.xml.dist
+++ b/psalm.xml.dist
@@ -51,6 +51,11 @@
                     is no longer supported.
                 -->
                 <file name="tests/Functional/ResultCacheTest.php"/>
+                <!--
+                    This suppression should be removed in 4.0.0.
+                -->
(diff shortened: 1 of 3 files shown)


greg0ire (approved):

beberlei (approved):

greg0ire:
Thanks @morozov !

PR #4682, opened 2021-06-19
Initial support for PHP 8.1

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

References:
1. https://wiki.php.net/rfc/deprecate_null_to_scalar_internal_arg
2. https://github.com/php/php-src/blob/578b67da49af51b2f796a48782e51ceb62860943/UPGRADING#L334-L341
3. https://github.com/php/php-src/blob/578b67da49af51b2f796a48782e51ceb62860943/UPGRADING#L123-L126

### DIFF
--- a/src/Schema/SqliteSchemaManager.php
+++ b/src/Schema/SqliteSchemaManager.php
@@ -419,7 +419,12 @@ class SqliteSchemaManager extends AbstractSchemaManager
                 ];
             }
 
-            $list[$name]['local'][]   = $value['from'];
+            $list[$name]['local'][] = $value['from'];
+
+            if ($value['to'] === null) {
+                continue;
+            }
+
             $list[$name]['foreign'][] = $value['to'];
         }
 
--- a/src/Schema/SqliteSchemaManager.php
+++ b/src/Schema/SqliteSchemaManager.php
@@ -196,7 +196,7 @@ class SqliteSchemaManager extends AbstractSchemaManager
         );
 
         foreach ($indexArray as $indexColumnRow) {
-            if ($indexColumnRow['pk'] === '0') {
+            if ($indexColumnRow['pk'] === 0 || $indexColumnRow['pk'] === '0') {
                 continue;
             }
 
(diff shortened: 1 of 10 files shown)


tests/Functional/BlobTest.php
```diff
@@ -144,7 +144,6 @@ public function testBindParamProcessesStream(): void
             "INSERT INTO blob_table(id, clobcolumn, blobcolumn) VALUES (1, 'ignored', ?)"
         );
 
-        $stream = null;
```
SenseException, line 147: `$stream` is not an instantiated variable now. Is that okay?

src/Schema/SqliteSchemaManager.php
```diff
@@ -419,7 +419,12 @@ protected function _getPortableTableForeignKeysList($tableForeignKeys)
                 ];
             }
 
-            $list[$name]['local'][]   = $value['from'];
+            $list[$name]['local'][] = $value['from'];
+
+            if ($value['to'] === null) {
```
SenseException, line 424: This changes the behavior now. Are `null` values even valid for `to`?

tests/Functional/BlobTest.php
morozov: Yes, it’s passed by reference on the next line.

src/Schema/SqliteSchemaManager.php
morozov: This specific change is a backport of 9ddd08a15a291bf1b4f6d7172307f3006859fb77 from `4.0.x`. The table manually declared in `SqliteSchemaManagerTest::testListForeignKeysFromExistingDatabase()` produces the NULL value of `to`. IIRC, it implies that the foreign key references columns with the same names in the referenced tables. Note that the original test expectation doesn't make sense in any case: https://github.com/doctrine/dbal/blob/e814cdc79b804b9ae40f46acbfe191bfc6731b54/tests/Functional/Schema/SqliteSchemaManagerTest.php#L75-L79

It might be a minor breaking change but I'd consider it a bugfix, even though the previous behavior was covered by a test as expected.

greg0ire (approved):

SenseException (approved):

PR #4685, opened 2021-06-21
Support for more drivers on PHP 8.1

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

### DIFF
--- a/src/Driver/Mysqli/Connection.php
+++ b/src/Driver/Mysqli/Connection.php
@@ -14,7 +14,10 @@ use mysqli;
 use function assert;
 use function floor;
 use function mysqli_init;
+use function mysqli_report;
 use function stripos;
+
+use const MYSQLI_REPORT_OFF;
 
 final class Connection implements ServerInfoAwareConnection
 {
@@ -45,6 +48,8 @@ final class Connection implements ServerInfoAwareConnection
         iterable $preInitializers = [],
         iterable $postInitializers = []
     ) {
+        mysqli_report(MYSQLI_REPORT_OFF);
+
         $connection = mysqli_init();
         assert($connection !== false);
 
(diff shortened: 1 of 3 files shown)


src/Driver/Mysqli/Connection.php
```diff
@@ -45,6 +48,8 @@ public function __construct(
         iterable $preInitializers = [],
         iterable $postInitializers = []
     ) {
+        mysqli_report(MYSQLI_REPORT_OFF);
```
greg0ire, line 51: Reference: https://php.watch/versions/8.1/mysqli-error-mode

This looks like a BC-break for situations where `mysqli_report` is called before this constructor. Should it be documented somewhere? Or is there just no way the DBAL could work with that setting set to anything but `MYSQLI_REPORT_OFF`?

morozov: > This looks like a BC-break for situations where `mysqli_report` is called before this constructor.

In its current state, the DBAL cannot properly work with any other mode than `MYSQLI_REPORT_OFF`. This code change enforces this expectation.

> Should it be documented somewhere?

What would the message and the purpose of the documentation be in this case? 

> Or is there just no way the DBAL could work with that setting set to anything but `MYSQLI_REPORT_OFF`?

The DBAL could learn to handle exceptions instead of the return value but I don't see much point in maintaining two mode-specific implementations. Instead, we can transition from one mode to another and adapt the code accordingly.

greg0ire: > What would the message and the purpose of the documentation be in this case?

The message would look like your first answer, and its purpose would be to inform people that would like to change that setting before calling the DBAL that they shouldn't. From what you tell me, it's orthogonal to this PR though, since the expectation already exists.

greg0ire (approved):

PR #4689, opened 2021-06-24
Normalize identifiers before filtering; fix getPrimaryKeyColumns not returning uppercased primary key

florisluiten:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | compatible with v2
| Fixed issues | None

I found out that my primary column "fooId" was not returned by Table::getPrimaryKeyColumns. After some investigation, it seems that the columnnames are normalized early on, translating the columname to lowercase. However in the filtering, the original name with upper-case characters is used. Since uppercase is not the same as lowercase, it wil not return the key, "loosing" the primary key.

This worked fine in the 2.x branch, but broke when we upgraded to version 3.x.

I've written a test to illustrate the problem, and applied normalization just before filtering. This will at least ensure that the column can be found; hoewever, I think it is still "wrong" to return the normalized version; changing that would probably mean a whole lot more work.

### DIFF
--- a/src/Schema/Table.php
+++ b/src/Schema/Table.php
@@ -735,6 +735,8 @@ class Table extends AbstractAsset
      */
     private function filterColumns(array $columnNames, bool $reverse = false): array
     {
+        $columnNames = array_map([$this, 'normalizeIdentifier'], $columnNames);
+
         return array_filter($this->_columns, static function (string $columnName) use ($columnNames, $reverse): bool {
             return in_array($columnName, $columnNames, true) !== $reverse;
         }, ARRAY_FILTER_USE_KEY);
--- a/tests/Schema/TableTest.php
+++ b/tests/Schema/TableTest.php
@@ -852,4 +852,18 @@ class TableTest extends TestCase
 
         $table->removeUniqueConstraint('unique_constraint');
     }
+
+    public function testPrimaryKeyWithCaptialization(): void
+    {
+        $table = new Table('foo');
+        $table->addColumn('fooID', 'integer');
+        $table->addColumn('bar_id', 'integer');
+        $table->setPrimaryKey(['bar_id', 'fooID']);
+
+        $primaryKey = $table->getPrimaryKeyColumns();
+
+        self::assertCount(2, $primaryKey);
+        self::assertArrayHasKey('fooid', $primaryKey);
+        self::assertArrayHasKey('bar_id', $primaryKey);
+    }
 }


tests/Schema/TableTest.php
```diff
@@ -852,4 +852,18 @@ public function testRemoveUniqueConstraintUnknownNameThrowsException(): void
...
+    {
+        $table = new Table('foo');
+        $table->addColumn('fooID', 'integer');
+        $table->addColumn('bar_id', 'integer');
+        $table->setPrimaryKey(['bar_id', 'fooID']);
+
+        $primaryKey = $table->getPrimaryKeyColumns();
+
+        self::assertCount(2, $primaryKey);
+        self::assertArrayHasKey('fooid', $primaryKey);
```
morozov, line 866: Why is it expected that the key is lower-cased? Column names are sensitive on some platforms and with some settings, so this might be the problem in the first place.

src/Schema/Table.php
```diff
@@ -735,6 +736,10 @@ public function getForeignKeyColumns()
      */
     private function filterColumns(array $columnNames, bool $reverse = false): array
     {
+        $columnNames = array_map(function (string $columnName): string {
```
morozov: Wouldn't it be enough to normalize the name during filtering? Why do we need a list of normalized names?

tests/Schema/TableTest.php
florisluiten: If you ask me, it's not expected :) 

The problem is that the columnames are already "normalized", meaning that they are stored in the lowercased form. The original name is not stored. 

I suppose it would be wise to do the normalization later, when needed, so the original columnnames are stored and getPrimaryKeyColumns can do simple lookup. That means that all other functions need to do the sanitazion on demand. Since this seems like a whole lot more work I do not want to do that unless I know that's the correct direction. Otherwise, it would be a waste of my time to do the change, and yours of checking it.

src/Schema/Table.php
florisluiten: You could normalize during the filtering, however, that would mean the sanitazion is done for every record, making it O(n), instead of doing it once where it is O(1). 

As said above, ideally we wouldn't require a list of sanitized names in the first place.

morozov: It's O(n) in both cases. Please move within `array_filter()`.

As for normalization, getting rid of it is part of a bigger challenge (see https://github.com/doctrine/dbal/issues/4357). We can consider it a requirement for now.

tests/Schema/TableTest.php
morozov: Agree. Let's leave the "normalization" in place for now.

morozov:

PR #4690, opened 2021-06-25
Merge 3.2.x into 4.0.x

morozov:


src/Schema/SqliteSchemaManager.php
```diff
@@ -154,7 +154,7 @@ public function listTableForeignKeys(string $table, ?string $database = null): a
                 $id = $value['id'];
 
                 $tableForeignKeys[$key] = array_merge($tableForeignKeys[$key], [
-                    'constraint_name' => isset($names[$id]) && $names[$id] !== '' ? $names[$id] : $id,
+                    'constraint_name' => isset($names[$id]) && $names[$id] !== '' ? $names[$id] : (string) $id,
```
morozov, line 157: This change doesn't come from the merge up but is necessary to fix the test failure:
```
1) Doctrine\DBAL\Tests\Functional\Schema\SqliteSchemaManagerTest::testListForeignKeysFromExistingDatabase
TypeError: Doctrine\DBAL\Schema\ForeignKeyConstraint::__construct(): Argument #4 ($name) must be of type string, int given, called in /home/runner/work/dbal/dbal/src/Schema/SqliteSchemaManager.php on line 426

src/Schema/ForeignKeyConstraint.php:65
src/Schema/SqliteSchemaManager.php:426
src/Schema/SqliteSchemaManager.php:164
tests/Functional/Schema/SqliteSchemaManagerTest.php:101
```
On the one hand, SQLite in PHP 8.1 returns integer values as PHP integers, and `$id` is sourced from an integer `id` column. On the other hand, the code in `4.0.x` expects the constraint name to be a string: https://github.com/doctrine/dbal/blob/08efef743dccdee2547b8aa2627fa74b766f9a56/src/Schema/ForeignKeyConstraint.php#L65-L69

Hence the failure.

A proper solution might be not to assign any name to an anonymous constraint but this approach was rejected in 912db7a28be30512a8dc631e06e5be3fb7906a37. This change was reverted as part of a bigger revert (#4255), so it wasn't necessarily wrong on its own.

For now, it's best to address the type issue and leave the logic as is.

greg0ire (approved):

PR #4692, opened 2021-06-26
[GH-4681] Remove TableGenerator

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

See #4681.


greg0ire (dismissed):
This deletion comes with a nice coverage increase :slightly_smiling_face: (+0.38%)

morozov:
Often, it's the runtime deprecation code that is not covered by tests. But in this case, it's in the constructors which are covered. Let's see if we can eventually get to 90% coverage.

greg0ire (approved):

PR #4693, opened 2021-06-26
Fix dropping PK on PostgreSQL

hschletz:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

On the PostgreSQL platform, the SQL statement generated by getDropIndexSQL() is not suitable for primary keys. In case of a PK, the call must be forwarded to getDropConstraintSQL().

This also fixes indirect usage, like changing a PK via alterTable().

The discussion on #2925 mentions the problem of passing indexes by name only, which cannot be solved easily. This PR  resolves the issues for Index objects.

The MySQL platform uses a similar approach.

### DIFF
--- a/src/Platforms/PostgreSQL94Platform.php
+++ b/src/Platforms/PostgreSQL94Platform.php
@@ -2,12 +2,14 @@
 
 namespace Doctrine\DBAL\Platforms;
 
+use Doctrine\DBAL\Exception\InvalidArgumentException;
 use Doctrine\DBAL\Schema\Column;
 use Doctrine\DBAL\Schema\ColumnDiff;
 use Doctrine\DBAL\Schema\ForeignKeyConstraint;
 use Doctrine\DBAL\Schema\Identifier;
 use Doctrine\DBAL\Schema\Index;
 use Doctrine\DBAL\Schema\Sequence;
+use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Schema\TableDiff;
 use Doctrine\DBAL\Types\BigIntType;
 use Doctrine\DBAL\Types\BinaryType;
@@ -753,6 +755,24 @@ SQL
     public function getCreateSchemaSQL($schemaName)
     {
         return 'CREATE SCHEMA ' . $schemaName;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getDropIndexSQL($index, $table = null)
+    {
+        if ($index instanceof Index && $index->isPrimary()) {
+            if (! ($table instanceof Table || (is_string($table) && $table !== ''))) {
+                throw new InvalidArgumentException(
+                    __METHOD__ . '() expects $table parameter to be nonempty string or ' . Table::class . '.'
+                );
+            }
+
+            return $this->getDropConstraintSQL($index, $table);
+        }
+
+        return parent::getDropIndexSQL($index, $table);
     }
 
     /**
--- a/src/Schema/AbstractSchemaManager.php
+++ b/src/Schema/AbstractSchemaManager.php
@@ -7,6 +7,7 @@ use Doctrine\DBAL\Event\SchemaColumnDefinitionEventArgs;
 use Doctrine\DBAL\Event\SchemaIndexDefinitionEventArgs;
 use Doctrine\DBAL\Events;
 use Doctrine\DBAL\Exception;
+use Doctrine\DBAL\Exception\InvalidArgumentException;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\Deprecations\Deprecation;
 use Throwable;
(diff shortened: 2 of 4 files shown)


src/Platforms/PostgreSQL94Platform.php
```diff
@@ -755,6 +757,28 @@ public function getCreateSchemaSQL($schemaName)
...
 
+    /**
+     * {@inheritDoc}
+     */
+    public function getDropIndexSQL($index, $table = null)
+    {
+        if ($index instanceof Index && $index->isPrimary()) {
+            if (! ($table instanceof Table || ($table !== null && $table !== ''))) {
+                throw new InvalidArgumentException(
+                    __METHOD__ . '() expects $table parameter to be nonempty string or ' . Table::class . '.'
```
greg0ire, line 768: The check above does not match this error message. For instance, you could pass an integer or an array here.

hschletz: Thank you for the hint. An explicit string type check also makes psalm happy.

greg0ire: Let's squash the commits together and pretend this never happened then :stuck_out_tongue: ?

How to do that?

1. `git rebase -i origin/3.1.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

src/Platforms/PostgreSQL94Platform.php
```diff
@@ -755,6 +757,24 @@ public function getCreateSchemaSQL($schemaName)
         return 'CREATE SCHEMA ' . $schemaName;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    public function getDropIndexSQL($index, $table = null)
+    {
+        if ($index instanceof Index && $index->isPrimary()) {
```
morozov, line 765: What if the PK is passed as a string (`table-name_pkey`) instead of an object?

morozov (commented):
Thanks for the patch, @hschletz. A change like this will require functional tests first of all. The unit ones are nice to have. Please see the [guidelines](https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/testing.html#integration-tests).

hschletz:
Thank you for pointing out functional tests. The platform-level fix turned out to be insufficient because dropIndex() connverted Index objects to strings. This is redundant because getDropIndexSQL() can and should handle the Index object, and passing the name as string suffers from the aforementioned limitation.

getDropIndexSQL() cannot reliably determine the index type from the name alone. I don't think this is even possible at platform level, because it would require querying the database schema. Passing the PK by name will cause incorrect SQL to be returned. This limitation existed before &ndash; I have not introduced a regression here. Only an Index object will guarantee correct output, which is not the case without the patch.

At schema manager level, this limitation does not apply. I added a fix to dropIndex() which will generate a proper Index object if only a name is given, and pass that to getDropIndexSQL(). This allows to pass the index as

- an Index object
- a string with the real index name
- the special value 'primary' which is used as a key in the index array. I'm not sure if this is intended behavior or just a side effect of the implementation. In the latter case, I would remove the test case.

Tests are skipped for SQLite because dropping a PK directly is not supported. Attempting to do so will result in invalid SQL syntax. As with some other operations, this would require re-creating the table without the PK. Is this already implemented somewhere?

hschletz:
Checks fail on OCI and DB2. From the backtrace, I conclude that those platform's getDropIndexSQL() implementation would need the same fix &ndash; branching out to getDropConstraintSQL() in case of a PK.
Unfortunately I don't have a test environment for those. How can I test that, other than blindly pushing in the vague hope that tests may succeed?

morozov:
> How can I test that, other than blindly pushing in the vague hope that tests may succeed?

You don't have to fix the test on Oracle and IBM DB2 if you don't want to. You can skip the failing tests and we'll address the remaining issues separately.

If you do want to fix it, you'll have to install the `oci8` and `ibm_db2` extensions into your PHP and use the corresponding Docker images to spin up a database server:
1. https://github.com/doctrine/dbal/blob/7dab37760854ff2f3358e8186af823f3c03fb07e/.github/workflows/continuous-integration.yml#L73-L74
2. https://github.com/doctrine/dbal/blob/7dab37760854ff2f3358e8186af823f3c03fb07e/.github/workflows/continuous-integration.yml#L432-L434

Those PHP extensions are not easy to install if you don't have experience (also `ibm_db2` doesn't work on macOS AFAIK):
1. For `oci8`, use a tutorial. E.g. [this one](https://gist.github.com/Yukibashiri/cebaeaccbe531665a5704b1b34a3498e) looks close to how I set it up.
2. For `ibm_db2`, look at: https://github.com/doctrine/dbal/blob/7dab37760854ff2f3358e8186af823f3c03fb07e/ci/github/ext/install-ibm_db2.sh#L5

morozov:

PR #4694, opened 2021-06-28
Update JetBrains PhpStorm stubs to 2021.1

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

1. Replace `microsoft/mssql-server-linux:2017-latest` in CI configuration with `mcr.microsoft.com/mssql/server:2017-latest`. Unrelated but necessary for the build to pass. Otherwise, it fails as:
   ```shell
   $ docker pull microsoft/mssql-server-linux:2019-latest
   Error response from daemon: manifest for microsoft/mssql-server-linux:2019-latest not found: manifest unknown: manifest unknown
   ```
2. Use PHP 8 for static analysis. The new version of stubs uses multi-line annotations which PHP 7.4 cannot parse. This may be by design (https://github.com/JetBrains/phpstorm-stubs/pull/1160).
3. Switch from the Psalm GitHub action to a generic implementation since there's no Psalm image built on top of PHP 8 (https://github.com/psalm/psalm-github-actions/issues/31).

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -407,7 +407,7 @@ jobs:
 
     services:
       mssql:
-        image: "microsoft/mssql-server-linux:2017-latest"
+        image: "mcr.microsoft.com/mssql/server:2017-latest"
         env:
           ACCEPT_EULA: "Y"
           SA_PASSWORD: "Doctrine2018"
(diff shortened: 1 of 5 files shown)


greg0ire (approved):
> Switch from the Psalm GitHub action to a generic implementation since there's no Psalm image built on top of PHP 8 

IMO this is better anyway: having a version number in the CI and one in composer.json is asking for trouble.

morozov:
> IMO this is better anyway: having a version number in the CI and one in composer.json is asking for trouble.

Let's keep it this way then. The generic configuration is pretty much straightforward and we use the same one for PHPUnit, PHPStan, and other tools anyways.

PR #4699, opened 2021-07-02
Update Psalm to 4.8.1

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | mostly no

#### Summary

While we did most of the argument names cleanup in https://github.com/doctrine/dbal/pull/4172, the updated Psalm flagged some more cases. Since we already support PHP 8, changing argument names could be technically called out as a BC break, I believe this change is acceptable:

1. The argument names are inconsistent across implementations, so the current state cannot be considered correct.
2. All changes are made in the methods that accept a single argument, so it's very unlikely that the calling side passes the argument by name.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
         "squizlabs/php_codesniffer": "3.6.0",
         "symfony/cache": "^4.4",
         "symfony/console": "^2.0.5|^3.0|^4.0|^5.0",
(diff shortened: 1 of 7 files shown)


SenseException:
The inconsistent naming was also a BC break, so this would fix some issues.

greg0ire (approved):

PR #4698, opened 2021-07-01
Add alterSchema() methods in AbstractSchemaManager

trompette:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

This is my first contribution to Doctrine DBAL! :tada: 

This PR introduces new convenient methods to apply a schema diff in one go.

I tried to follow the same template used in `alterTable()` methods.

The methods introduced in this PR would help me implement a [new feature on a small library](https://github.com/trompette/php-feature-toggles/issues/13) based on Doctrine DBAL I am maintaining.

What do you think? :thinking:

PR #4704, opened 2021-07-08
Use less verbose Comparator::compareSchema()

trompette:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

While working on #4698, I noticed a small inconsistency when comparing schemas with `Comparator`.

This PR uses an existing shortcut everywhere instead of a verbose alternative.

### DIFF
--- a/src/Schema/Schema.php
+++ b/src/Schema/Schema.php
@@ -437,10 +437,7 @@ class Schema extends AbstractAsset
      */
     public function getMigrateToSql(Schema $toSchema, AbstractPlatform $platform)
     {
-        $comparator = new Comparator();
-        $schemaDiff = $comparator->compare($this, $toSchema);
-
-        return $schemaDiff->toSql($platform);
+        return Comparator::compareSchemas($this, $toSchema)->toSql($platform);
     }
 
     /**
@@ -450,10 +447,7 @@ class Schema extends AbstractAsset
      */
     public function getMigrateFromSql(Schema $fromSchema, AbstractPlatform $platform)
     {
-        $comparator = new Comparator();
-        $schemaDiff = $comparator->compare($fromSchema, $this);
-
-        return $schemaDiff->toSql($platform);
+        return Comparator::compareSchemas($fromSchema, $this)->toSql($platform);
     }
 
     /**
--- a/src/Schema/Comparator.php
+++ b/src/Schema/Comparator.php
@@ -27,9 +27,7 @@ class Comparator
      */
     public static function compareSchemas(Schema $fromSchema, Schema $toSchema)
     {
-        $c = new self();
-
-        return $c->compare($fromSchema, $toSchema);
+        return (new self())->compare($fromSchema, $toSchema);
     }
 
     /**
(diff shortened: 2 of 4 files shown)


src/Schema/Schema.php
```diff
@@ -437,10 +437,7 @@ public function toDropSql(AbstractPlatform $platform)
      */
     public function getMigrateToSql(Schema $toSchema, AbstractPlatform $platform)
     {
-        $comparator = new Comparator();
-        $schemaDiff = $comparator->compare($this, $toSchema);
-
-        return $schemaDiff->toSql($platform);
+        return Comparator::compareSchemas($this, $toSchema)->toSql($platform);
```
morozov, line 440: This is the only static `Comparator` method, so I'd rather replace it with a non-static one for consistency with the rest of the API and deprecated the `compare()` method since its purpose is unclear from the name.

Please also see https://github.com/doctrine/dbal/pull/4659#discussion_r642103628.

trompette: Looking at the code and how it is used, it is quite odd the other static diffing methods are not present. Specially given the `Comparator` object does not have any state.

As mentionned in https://github.com/doctrine/dbal/pull/4659#discussion_r642103628, if the goal is to rename all methods anyways, maybe we could introduce a new `Differ` class with the following public API:

```php
class Differ
{
    public static function compareSchemas(Schema $s1, Schema $s2): SchemaDiff;
    public static function compareSequence(Sequence $s1, Sequence $s2): bool;
    public static function compareTable(Table $t1, Table $t2): TableDiff|false;
    public static function compareForeignKey(ForeignKeyConstraint $k1, ForeignKeyConstraint $k2): bool;
    public static function compareColumn(Column $c1, Column $c2): string[]; // could also be ColumnDiff
    public static function compareIndex(Index $i1, Index $i2): bool;
}
```

(Now that I see the API, maybe the methods returning a boolean should be renamed for clarity or moved to a new class.) 

What do you think ?

morozov: "Differ" is a bit odd name, I'd prefer the same class name: the comparator compares things and returns a diff – looks like a solid naming foundation to me. As for the specific method names, I think they should contain the plural form of the object (e.g. `compareColumns()`).

IIRC, we'll need to only migrate `compareSchemas()` from being static to non-static which doesn't require introduction of a new class. We can deprecate non-static calls to the method (which PHP allows) and make it non-static in the next major release.

Oh, and I don't think static methods is a good idea. This way, it will be impossible to evolve the API further, e.g. split it into multiple classes, define interfaces, pass it as a dependency, etc.

trompette: "Differ" sounded great from my suburb of Paris, France... :wink: 

I did not think about evolving the `Comparator` API further, you are quite right!

trompette:
Let's close this PR. I will open another one to make `Comparator` API more consistent.

src/Schema/Schema.php
morozov: > "Differ" sounded great from my suburb of Paris, France... 😉

I think that would be perfect in the Java world where they smash an _-er_ on top of any noun :-)

Issue #1004, opened 2014-12-10
DBAL-1069: datetimetz type needs to be a commented one by default


doctrinebot:
Jira issue originally created by user stof:

By default, the AbstractPlatform creates datetimetz fields as datetime field for platforms not defining a specific creation for fields with timezones.
This can be an acceptable fallback if you know that your application always send the value as UTC even when the database can accept any timezone as input thanks to the datetimetz type (database not supporting timezone would still have consistent data if you always send the same timezone).
However it creates an issue for the SchemaTool: the field will obviously be reported as datetime when inspecting the DB, not as datetimetz (since it does not have a datetimetz). So for platforms not supporting datetimetz natively, the special comment should be used to avoid useless updates.
The special Doctrine command should however not be used for platforms supporting the type natively (PostgreSQL, SQLServer2008, Oracle and SqlAnywhere12).

I have an idea about a fix, so I may send a PR for this in the following days.

morozov:
The schema introspection issue is a side effect of the DBAL silently dropping the timezone part in DDL and the values if the underlying platform doesn't timezones on the given column type:
https://github.com/doctrine/dbal/blob/25496429ef4bba2352b0ced7daffe1b3dec83d13/src/Platforms/AbstractPlatform.php#L3314-L3317
https://github.com/doctrine/dbal/blob/25496429ef4bba2352b0ced7daffe1b3dec83d13/src/Platforms/AbstractPlatform.php#L2979-L2982
Instead of that, it should explicitly throw an "unsupported" exception.

PR #4706, opened 2021-07-09
add comment to DateTimeTz if required

Khartir:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1004

#### Summary

Add database comment to DateTimeTzType if required.

### DIFF
--- a/src/Platforms/AbstractPlatform.php
+++ b/src/Platforms/AbstractPlatform.php
@@ -3275,6 +3275,16 @@ abstract class AbstractPlatform
     }
 
     /**
+     * Does this platform have native time type.
+     *
+     * @return bool
+     */
+    public function hasNativeTimezoneType()
+    {
+        return false;
+    }
+
+    /**
      * Whether this platform supports views.
      *
      * @return bool
--- a/src/Platforms/OraclePlatform.php
+++ b/src/Platforms/OraclePlatform.php
@@ -1206,4 +1206,12 @@ SQL
             $ownerCondition
         );
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function hasNativeTimezoneType()
+    {
+        return true;
+    }
 }
--- a/src/Platforms/PostgreSQL94Platform.php
+++ b/src/Platforms/PostgreSQL94Platform.php
@@ -1304,4 +1304,12 @@ SQL
             $this->quoteStringLiteral($table)
         );
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function hasNativeTimezoneType()
+    {
+        return true;
+    }
 }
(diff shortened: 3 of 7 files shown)


morozov:
@Khartir thanks for the patch. Since the code being added is related to the supported platforms, the patch needs to include functional tests as well. Please see the [testing guide](https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/testing.html#integration-tests) for more details.

Khartir:
Is this what you had in mind? I tried doing it similar to the tests for `JsonType` and `GuidType` which I used as guidelines for the fixes to the Type-class and I'm a bit confused.

morozov:
> Is this what you had in mind?

Not really. I was hoping, the test would reproduce the actual problem of storing/fetching data from the database.

After a closer look at the details of the original issue,

> This can be an acceptable fallback if you know that your application always send the value as UTC even when the database can accept any timezone as input thanks to the datetimetz type (database not supporting timezone would still have consistent data if you always send the same timezone).

It doesn't look like it's a proper solution to the problem. As far as I understand, it will allow DBAL to be part of a data loss if the platform doesn't support storing values with the timezone. Currently, it already silently drops the timezone part in DDL and the values:
https://github.com/doctrine/dbal/blob/25496429ef4bba2352b0ced7daffe1b3dec83d13/src/Platforms/AbstractPlatform.php#L3314-L3317
https://github.com/doctrine/dbal/blob/25496429ef4bba2352b0ced7daffe1b3dec83d13/src/Platforms/AbstractPlatform.php#L2979-L2982

If a user creates a `DateTimeTz` column on a platform that doesn't support it (say MySQL), stores a value with a non-default timezone and fetches it back, what will they get? This is what I was hoping the functional test would cover.

Khartir:
OK, I understood the original issue differently:
> However it creates an issue for the SchemaTool: the field will obviously be reported as datetime when inspecting the DB, not as datetimetz (since it does not have a datetimetz). So for platforms not supporting datetimetz natively, the special comment should be used to avoid useless updates.

Basically the problem is, if you create a table with a `datetimetz`-column on a platform that doesn't support it, the database type is simply `datetime`. If you then run the SchemaTool it reports a difference between the database and the defined schema and tries to create a migration that basically has the same code in the `up` and `down` part.
The issue of maybe loosing timezone information is not solved by this PR.

morozov:
I'll break it down into points:
> 1. [...] if you create a table with a datetimetz-column on a platform that doesn't support it, the database type is simply datetime.
> 2. If you then run the SchemaTool it reports a difference between the database and the defined schema and tries to create a migration that basically has the same code in the up and down part.

The behavior described in #​2 is only a problem if we agree that the behavior described in #​1 is valid. Which is not the case. The very fact of the DBAL simply dropping the timezone part is a bug.

> The issue of maybe loosing timezone information is not solved by this PR.

I don't think we should be solving a side effect of a known more severe problem in order to hide it.

Khartir:
OK in that case I have no idea on how to fix this. I'll close this PR.

PR #4707, opened 2021-07-09
Introduce consistent Comparator API

trompette:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no, deprecations only
| Fixed issues | none

#### Summary

This PR deprecates previous `Comparator` API and replaces it with a more consistent one.

For more information, see:
-  https://github.com/doctrine/dbal/pull/4659#discussion_r642103628
- https://github.com/doctrine/dbal/pull/4704#discussion_r666607812

### DIFF
--- a/psalm.xml.dist
+++ b/psalm.xml.dist
@@ -87,6 +87,11 @@
                 -->
                 <file name="src/Schema/AbstractSchemaManager.php" />
                 <file name="src/Schema/PostgreSQLSchemaManager.php" />
+                <!--
+                    See https://github.com/doctrine/dbal/pull/4707
+                    TODO: remove in 4.0.0
+                -->
+                <file name="src/Schema/Comparator.php" />
                 <directory name="tests" />
                 <referencedMethod name="Doctrine\DBAL\Statement::execute"/>
             </errorLevel>
--- a/src/Schema/AbstractSchemaManager.php
+++ b/src/Schema/AbstractSchemaManager.php
@@ -694,7 +694,9 @@ abstract class AbstractSchemaManager
      */
     public function migrateSchema(Schema $toSchema): void
     {
-        $this->alterSchema(Comparator::compareSchemas($this->createSchema(), $toSchema));
+        $schemaDiff = (new Comparator())->compareSchemas($this->createSchema(), $toSchema);
+
+        $this->alterSchema($schemaDiff);
     }
 
     /* alterTable() Methods */
--- a/src/Schema/Comparator.php
+++ b/src/Schema/Comparator.php
@@ -3,6 +3,7 @@
 namespace Doctrine\DBAL\Schema;
 
 use Doctrine\DBAL\Types;
+use Doctrine\Deprecations\Deprecation;
 
 use function array_intersect_key;
 use function array_key_exists;
@@ -21,30 +22,17 @@ use function strtolower;
 class Comparator
 {
     /**
+     * Returns a SchemaDiff object containing the differences between the schemas $fromSchema and $toSchema.
+     *
+     * This method should be called non-statically since it will be declared as non-static in the next major release.
+     *
      * @return SchemaDiff
      *
      * @throws SchemaException
      */
     public static function compareSchemas(Schema $fromSchema, Schema $toSchema)
     {
-        $c = new self();
-
-        return $c->compare($fromSchema, $toSchema);
-    }
-
-    /**
-     * Returns a SchemaDiff object containing the differences between the schemas $fromSchema and $toSchema.
-     *
-     * The returned differences are returned in such a way that they contain the
-     * operations to change the schema stored in $fromSchema to the schema that is
-     * stored in $toSchema.
-     *
-     * @return SchemaDiff
-     *
-     * @throws SchemaException
-     */
-    public function compare(Schema $fromSchema, Schema $toSchema)
-    {
+        $comparator       = new self();
         $diff             = new SchemaDiff();
         $diff->fromSchema = $fromSchema;
 
@@ -71,7 +59,7 @@ class Comparator
             if (! $fromSchema->hasTable($tableName)) {
                 $diff->newTables[$tableName] = $toSchema->getTable($tableName);
             } else {
-                $tableDifferences = $this->diffTable(
+                $tableDifferences = $comparator->diffTable(
                     $fromSchema->getTable($tableName),
                     $toSchema->getTable($tableName)
                 );
@@ -134,18 +122,18 @@ class Comparator
         foreach ($toSchema->getSequences() as $sequence) {
             $sequenceName = $sequence->getShortestName($toSchema->getName());
             if (! $fromSchema->hasSequence($sequenceName)) {
-                if (! $this->isAutoIncrementSequenceInSchema($fromSchema, $sequence)) {
+                if (! $comparator->isAutoIncrementSequenceInSchema($fromSchema, $sequence)) {
                     $diff->newSequences[] = $sequence;
                 }
             } else {
-                if ($this->diffSequence($sequence, $fromSchema->getSequence($sequenceName))) {
+                if ($comparator->diffSequence($sequence, $fromSchema->getSequence($sequenceName))) {
                     $diff->changedSequences[] = $toSchema->getSequence($sequenceName);
                 }
             }
         }
 
         foreach ($fromSchema->getSequences() as $sequence) {
-            if ($this->isAutoIncrementSequenceInSchema($toSchema, $sequence)) {
+            if ($comparator->isAutoIncrementSequenceInSchema($toSchema, $sequence)) {
                 continue;
             }
 
@@ -159,6 +147,24 @@ class Comparator
         }
 
         return $diff;
+    }
+
+    /**
+     * @deprecated Use non-static call to {@link compareSchemas()} instead.
+     *
+     * @return SchemaDiff
+     *
+     * @throws SchemaException
+     */
+    public function compare(Schema $fromSchema, Schema $toSchema)
+    {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pull/4707',
+            'Method compare() is deprecated. Use a non-static call to compareSchemas() instead.'
+        );
+
+        return $this->compareSchemas($fromSchema, $toSchema);
     }
 
     /**
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -8,6 +8,17 @@ awareness about deprecated code.
 
 # Upgrade to 3.2
 
+## Deprecated static calls to `Comparator::compareSchemas($fromSchema, $toSchema)`
+
+The usage of `Comparator::compareSchemas($fromSchema, $toSchema)` statically is
+deprecated in order to provide a more consistent API.
+
+## Deprecated `Comparator::compare($fromSchema, $toSchema)`
+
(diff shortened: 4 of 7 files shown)


ausi:
I’m not sure if I understand this change. Wouldn’t it make merging #4659 way harder if we have to keep backwards compatibility there with the new method names from this pull request?

src/Schema/AbstractSchemaManager.php
```diff
@@ -694,7 +694,10 @@ public function alterSchema(SchemaDiff $schemaDiff): void
      */
     public function migrateSchema(Schema $toSchema): void
     {
-        $this->alterSchema(Comparator::compareSchemas($this->createSchema(), $toSchema));
+        // @phpstan-ignore-next-line
```
morozov: We usually suppress errors in the configuration `phpstan.xml.dist`, please move this one there.

src/Schema/Comparator.php
```diff
@@ -21,23 +21,29 @@
 class Comparator
 {
     /**
+     * Returns a SchemaDiff object containing the differences between the schemas $fromSchema and $toSchema.
+     *
+     * The returned differences are returned in such a way that they contain the
+     * operations to change the schema stored in $fromSchema to the schema that is
+     * stored in $toSchema.
```
morozov: It looks like this paragraph belongs more to the `SchemaDiff` class itself rather than this method. Would it make sense to move it there?

src/Schema/Comparator.php
```diff
@@ -21,23 +21,29 @@
...
+     *
+     * This method is marked as static to avoid a BC break but should be called
+     * non-statically since it will not be static in the next major release.
```
morozov: For this, we can try adding a runtime deprecation based on `if (! isset($this))`. Also, please add runtime deprecations to the other methods.

src/Schema/Comparator.php
```diff
@@ -21,23 +21,29 @@
...
     }
 
     /**
-     * Returns a SchemaDiff object containing the differences between the schemas $fromSchema and $toSchema.
+     * This method is deprecated and will be removed in the next major release.
      *
-     * The returned differences are returned in such a way that they contain the
-     * operations to change the schema stored in $fromSchema to the schema that is
-     * stored in $toSchema.
+     * @deprecated Use non-static call to {@link compareSchemas()} instead.
```
morozov: Let's make sure the deprecated method doesn't contain any logic except for calling the recommended method.

src/Schema/Comparator.php
```diff
@@ -71,7 +77,7 @@ public function compare(Schema $fromSchema, Schema $toSchema)
             if (! $fromSchema->hasTable($tableName)) {
                 $diff->newTables[$tableName] = $toSchema->getTable($tableName);
             } else {
-                $tableDifferences = $this->diffTable(
+                $tableDifferences = $this->compareTables(
```
morozov: This has to remain unchanged. Somebody may use a customized version of the comparator with the overridden `diffTable()` implementation. If the library code stops calling their customized code, it may be considered a BC break. The library code is allowed to use its deprecated APIs.

src/Schema/Comparator.php
```diff
@@ -181,7 +187,7 @@ private function isAutoIncrementSequenceInSchema($schema, $sequence)
     /**
      * @return bool
      */
-    public function diffSequence(Sequence $sequence1, Sequence $sequence2)
+    public function compareSequences(Sequence $sequence1, Sequence $sequence2)
```
morozov: We can and should enforce the return type on all new methods.

src/Schema/Comparator.php
```diff
@@ -190,6 +196,18 @@ public function diffSequence(Sequence $sequence1, Sequence $sequence2)
         return $sequence1->getInitialValue() !== $sequence2->getInitialValue();
     }
 
+    /**
+     * This method is deprecated and will be removed in the next major release.
```
morozov: This usually describes what the method does, the users may see this in their IDEs, etc. The `@deprecated` annotation seems sufficient to me.

tests/Functional/Platform/NewPrimaryKeyWithNewAutoIncrementColumnTest.php
```diff
@@ -47,7 +47,7 @@ public function testAlterPrimaryKeyToAutoIncrementColumn(): void
         $newTable->dropPrimaryKey();
         $newTable->setPrimaryKey(['new_id']);
 
-        $diff = (new Comparator())->compare($schema, $newSchema);
+        $diff = (new Comparator())->compareSchemas($schema, $newSchema);
```
morozov: The tests should not change unless it's a bug or a breaking change. This way, we can make sure that the deprecated API still functions properly.

morozov (commented):
@trompette, the patch looks good overall. Please see a few suggestions inline.

src/Schema/Comparator.php
trompette: I tried the `if (! isset($this))` trick: it only works when the method is not marked as static.

But `ComparatorTest` fails when the `static` keyword is removed:
> Non-static method Doctrine\DBAL\Schema\Comparator::compareSchemas() should not be called statically

And this looks a lot like a BC break to me.

trompette: As a workaround, I guess we could make the method private and leverage `__call()` and `__callStatic()` to add the runtime deprecation, but this is not ideal.

What do you think?

ausi: `__call` and `__callStatic` is the only way possible to make both ways work AFAIK.

m-vo: Could this method be named differently to prevent this issue and possible confusion when refactoring to the new API?

ausi: Another option would be to name the new method `compareSchemata` as “schemata” is also a plural form of “schema”

morozov: A runtime deprecation isn't a must and it looks like it cannot be implemented here. We definitely cannot remove the `static` flag in a minor release, so let's leave it as is and just document the static call deprecation.

> `__call` and `__callStatic` is the only way possible to make both ways work AFAIK.

I wouldn't complicate it that much.

> Another option would be to name the new method `compareSchemata` as “schemata” is also a plural form of “schema”

While this is right, it would be inconsistent with the other method names. I'd prefer to keep the original name.

src/Schema/Comparator.php
```diff
@@ -20,31 +21,29 @@
...
-        return $c->compare($fromSchema, $toSchema);
-    }
-
     /**
      * Returns a SchemaDiff object containing the differences between the schemas $fromSchema and $toSchema.
      *
-     * The returned differences are returned in such a way that they contain the
-     * operations to change the schema stored in $fromSchema to the schema that is
-     * stored in $toSchema.
+     * This method is marked as static to avoid a BC break but should be called
```
morozov: Technically, this method has been marked as static since the very beginning, not in order to avoid a BC break. Let's skip this part and put it simply as:
> This method should be called non-statically since it will be declared as non-static in the next major release.

src/Schema/Comparator.php
```diff
@@ -20,31 +21,29 @@
...
+     * non-statically since it will not be static in the next major release.
      *
      * @return SchemaDiff
      *
      * @throws SchemaException
      */
-    public function compare(Schema $fromSchema, Schema $toSchema)
+    public static function compareSchemas(Schema $fromSchema, Schema $toSchema)
     {
+        /*
```
morozov: Please remove this. Apparently, `$this` is unavailable in a `static` method call even if it's called non-statically.

src/Schema/Comparator.php
```diff
@@ -20,31 +21,29 @@
  */
 class Comparator
 {
-    /**
-     * @return SchemaDiff
-     *
-     * @throws SchemaException
-     */
-    public static function compareSchemas(Schema $fromSchema, Schema $toSchema)
```
morozov, line 28: Would it be possible to retain the original placement of the methods? Otherwise, it takes extra effort to confirm that the method hasn't been actually removed but was put in a different place in the class.

trompette: The methods are in the same order before and after the PR: `compareSchemas()` first, then `compare()`.

It looks like Git shows the diff this way because the long body moved from `compare()` to `compareSchemas()`.

src/Schema/Comparator.php
```diff
@@ -178,10 +184,7 @@ private function isAutoIncrementSequenceInSchema($schema, $sequence)
         return false;
     }
 
-    /**
-     * @return bool
-     */
-    public function diffSequence(Sequence $sequence1, Sequence $sequence2)
+    public function compareSequences(Sequence $sequence1, Sequence $sequence2): bool
```
morozov: After looking at the full picture, I think that the new API is less intuitive than the existing one when it comes to the methods that return a boolean (I didn't realize we have some of those). A method `diff($x, $y): bool` is easy to interpret, it returns whether there is a difference between the values or not, but `compare($x, $y)` returning `false` when the values are equal is odd.

Additionally, the new methods would be nice to have to accommodate `$platform` as a required parameter in https://github.com/doctrine/dbal/pull/4659 but the current version accepts it in the constructor which is good as well. 

Let's keep the PR to its original scope and only deprecate the static method. We'll see if more improvements are necessary once the platform-aware patch lands. Sorry if it takes you more than originally planned.

trompette: > A method `diff($x, $y): bool` is easy to interpret, it returns whether there is a difference between the values or not, but `compare($x, $y)` returning `false` when the values are equal is odd.

I did not think of it while making the changes, but now you say it, the naming is quite odd indeed.

> Let's keep the PR to its original scope and only deprecate the static method.

OK, I will revert the changes except for `compareSchemas()` and `compare()`.

> Sorry if it takes you more than originally planned.

No worries! Small increments are the way to go.

psalm.xml.dist
```diff
@@ -60,11 +60,6 @@
         </DeprecatedClass>
         <DeprecatedMethod>
             <errorLevel type="suppress">
-                <!--
-                    This suppression should be removed after 2022
-                    See https://github.com/doctrine/dbal/pull/4317
-                -->
-                <file name="tests/Functional/LegacyAPITest.php"/>
```
morozov: I believe the `<directory name="tests" />` below comes from a merge conflict improperly resolved in [4532](https://github.com/doctrine/dbal/pull/4532/files#diff-2d20a98059390e83ae58234d4bda07fe9319b822009ce8c4479e0f3ae1943fbfR70).

In newer branches, I'd rather have all deprecations listed here explicitly in order to make sure we remove the BC layer in the next major release. Please revert the removal of the suppressions that don't belong to this patch, remove `<directory name="tests" />` and add the needed suppressions for this patch. If there are too many new issues reported, you can restore `<directory name="tests" />` after that. We can remove it separately.

trompette: > 68 errors found

Too many errors, I am restoring `<directory name="tests" />`.

morozov (dismissed):

morozov:
Sorry, one last thing. We'll need to add a note to `UPGRADE.md`. Otherwise, looks good.

morozov (approved):

morozov:
Thanks, @trompette!

morozov:
@trompette if you fancy, you can go ahead and file another PR that removes the deprecated APIs from `4.0.x` (will also need an note in `UPGRADE.md`).

trompette:
@morozov, done! :wink:

PR #4709, opened 2021-07-12
Avoid crash when getting details of table and they exist in different schemas

shemgp:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #950 

#### Summary

Previous [PR](https://github.com/doctrine/dbal/pull/4494) got lost due to wrong merges, so here's a new one:

This fix will use only the current schema when getting details of a table that has the same name in two or more schemas. Previously it would get the details of all the tables thus producing the exception.

### DIFF
--- a/src/Platforms/PostgreSQL94Platform.php
+++ b/src/Platforms/PostgreSQL94Platform.php
@@ -380,7 +380,7 @@ SQL
             [$schema, $table] = explode('.', $table);
             $schema           = $this->quoteStringLiteral($schema);
         } else {
-            $schema = "CURRENT_SCHEMA";
+            $schema = 'CURRENT_SCHEMA';
         }
 
         $table = new Identifier($table);
--- a/src/Platforms/PostgreSQL94Platform.php
+++ b/src/Platforms/PostgreSQL94Platform.php
@@ -380,7 +380,7 @@ SQL
             [$schema, $table] = explode('.', $table);
             $schema           = $this->quoteStringLiteral($schema);
         } else {
-            $schema = 'ANY(current_schemas(false))';
+            $schema = "ANY(CASE WHEN CURRENT_SCHEMA <> 'public' THEN string_to_array(CURRENT_SCHEMA, ',') ELSE current_schemas(false) END)";
         }
 
         $table = new Identifier($table);
--- a/src/Platforms/PostgreSQL94Platform.php
+++ b/src/Platforms/PostgreSQL94Platform.php
@@ -380,7 +380,7 @@ SQL
             [$schema, $table] = explode('.', $table);
             $schema           = $this->quoteStringLiteral($schema);
         } else {
-            $schema = "ANY(string_to_array(CURRENT_SCHEMA, ','))";
+            $schema = "CURRENT_SCHEMA";
         }
 
         $table = new Identifier($table);
--- a/src/Platforms/PostgreSQL94Platform.php
+++ b/src/Platforms/PostgreSQL94Platform.php
@@ -380,7 +380,7 @@ SQL
             [$schema, $table] = explode('.', $table);
             $schema           = $this->quoteStringLiteral($schema);
         } else {
-            $schema = "ANY(CASE WHEN CURRENT_SCHEMA <> 'public' THEN string_to_array(CURRENT_SCHEMA, ',') ELSE current_schemas(false) END)";
+            $schema = "ANY(string_to_array(CURRENT_SCHEMA, ','))";
         }
 
         $table = new Identifier($table);
--- a/tests/Functional/Schema/PostgreSQLSchemaManagerTest.php
+++ b/tests/Functional/Schema/PostgreSQLSchemaManagerTest.php
@@ -494,12 +494,12 @@ class PostgreSQLSchemaManagerTest extends SchemaManagerFunctionalTestCase
         self::assertTrue($tableFinal->getColumn('id')->getAutoincrement());
     }
 
-    public function testGetDetailsWithTwoSameTablesOnDifferentSchemas(): void
+    public function testListTableDetailsWithTwoSameTablesOnDifferentSchemas(): void
     {
-        $createTableSQL = 'CREATE TABLE migrations(a text);';
-        $this->connection->executeStatement($createTableSQL);
-        $createTableSQL = 'CREATE TABLE "001_test".migrations(a text);';
-        $this->connection->executeStatement($createTableSQL);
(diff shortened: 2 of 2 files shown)


tests/Functional/Schema/PostgreSQLSchemaManagerTest.php
```diff
@@ -494,6 +494,51 @@ public function testAlterTableAutoIncrementIntToBigInt(string $from, string $to,
...
+        $this->connection->executeStatement($createTableSQL);
+        $this->connection->executeStatement('SET search_path TO "001_test",public');
+
+        $table = new Table('migrations');
+        $table->addColumn('a', 'text');
+
+        try
+        {
+            $databaseTable = $this->schemaManager->listTableDetails('migrations');
+        } catch (\Exception $e) {
```
morozov: Why is an exception expected here?

tests/Functional/Schema/PostgreSQLSchemaManagerTest.php
```diff
@@ -494,6 +494,51 @@ public function testAlterTableAutoIncrementIntToBigInt(string $from, string $to,
...
+            $databaseTable = null;
+        }
+
+        $c    = new Comparator();
+        $diff = $c->diffTable($table, $databaseTable);
+
+        self::assertNotFalse($diff);
```
morozov: Is it the only requirement that the diff is not empty?

src/Platforms/PostgreSQL94Platform.php
```diff
@@ -380,7 +380,7 @@ private function getTableWhereClause($table, $classAlias = 'c', $namespaceAlias
             [$schema, $table] = explode('.', $table);
             $schema           = $this->quoteStringLiteral($schema);
         } else {
-            $schema = 'ANY(current_schemas(false))';
+            $schema = "CURRENT_SCHEMA";
```
morozov: This will work only if the table is found in the first schema in the search path. What if it's not?

tests/Functional/Schema/PostgreSQLSchemaManagerTest.php
```diff
@@ -494,6 +494,51 @@ public function testAlterTableAutoIncrementIntToBigInt(string $from, string $to,
...
+        $table->addColumn('a', 'text');
+
+        try
+        {
+            $databaseTable = $this->schemaManager->listTableDetails('migrations');
+        } catch (\Exception $e) {
+            $databaseTable = null;
+        }
+
+        $c    = new Comparator();
```
morozov: Given that the patch only affects schema introspection, why does the test need to involve schema comparison?

tests/Functional/Schema/PostgreSQLSchemaManagerTest.php
shemgp: Needed since when the patch is not applied, there is an exception so I catch it.

tests/Functional/Schema/PostgreSQLSchemaManagerTest.php
shemgp: I changed it so diff is not needed anymore.

src/Platforms/PostgreSQL94Platform.php
shemgp: I added tests to show that it still works with the table in the different schema.

tests/Functional/Schema/PostgreSQLSchemaManagerTest.php
shemgp: Yup. Removed the schema comparison in the tests.

tests/Functional/Schema/PostgreSQLSchemaManagerTest.php
morozov: This test is meant to fail without this patch. Please remove the try/catch.

src/Platforms/PostgreSQL94Platform.php
morozov: The `testGetDetailsStillWorkingWithOneTable` test only proves that the query doesn't fail but doesn't prove that it works. Adding a few assertions will prove that it doesn't work:
1. The introspected table name is `migrations` but it should be `001_test.migrations`.
2. It has no columns.

tests/Functional/Schema/PostgreSQLSchemaManagerTest.php
shemgp: removed the try catch

src/Platforms/PostgreSQL94Platform.php
shemgp: Added test for `001_test.migrations`. 

Of course it will not work if that table does not exist eg. if I change `migrations` to `"001_test".migrations` below:
```php
    public function testGetDetailsStillWorkingWithOneTable3(): void
    {
        $createTableSQL = 'CREATE TABLE IF NOT EXISTS migrations(a text);';
        $this->connection->executeStatement($createTableSQL);
        $createTableSQL = 'DROP TABLE IF EXISTS "001_test".migrations;';
        $this->connection->executeStatement($createTableSQL);
        $this->connection->executeStatement('SET search_path TO "001_test",public');

        $databaseTable = $this->schemaManager->listTableDetails('"001_test".migrations');

        self::assertNotNull($databaseTable);
    }
```

When the `migrations` table does not have columns it works for me. I added a test for it too. I thought it did not work, but my syntax was wrong: `CREATE TABLE IF NOT EXISTS migrations;`, since there's no `()` after `migrations`.

tests/Functional/Schema/PostgreSQLSchemaManagerTest.php
```diff
@@ -494,6 +494,88 @@ public function testAlterTableAutoIncrementIntToBigInt(string $from, string $to,
...
+    {
+        $sql = 'CREATE TABLE migrations(a text);';
+        $this->connection->executeStatement($sql);
+        $sql = 'CREATE TABLE "001_test".migrations(a text);';
+        $this->connection->executeStatement($sql);
+        $this->connection->executeStatement('SET search_path TO "001_test",public');
+
+        $databaseTable = $this->schemaManager->listTableDetails('migrations');
+
+        self::assertNotNull($databaseTable);
```
morozov, line 507: `$databaseTable` is never null: https://github.com/doctrine/dbal/blob/b4f49c050462878fb39423762054fe65276c1f99/src/Schema/AbstractSchemaManager.php#L319-L331

morozov (commented):
The new tests do not in fact cover the change. Please see the details inline.

Issue #4711, opened 2021-07-14
VERSION is out-of-date in 2.13.* release series


phil-davis:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.13.2

#### Summary
`lib/Doctrine/DBAL/Version.php` has out-of-date version.

#### Current behaviour
`public const VERSION = '2.12.2-DEV';`

#### How to reproduce

Look at the source code.

#### Expected behaviour
Should have been released as:
`public const VERSION = '2.13.2';`

And maybe in the branch it should now be:
`public const VERSION = '2.13.3-DEV';`

greg0ire:
That class was deprecated in https://github.com/doctrine/dbal/pull/3861

It was not stated, but I think one of the goals was to stop bumping it on every release. Also, since it's removed in 3.x, merging this will mean a conflict when merging up 2.x into 3.x.

phil-davis:
OK - feel free to close this, and PR #4712 if the decision is that the value of VERSION will no longer be incremented.

That will mean that consumers of this in 2.13.* will see changed behavior - the VERSION will not be reliably up-to-date. They will "see" more than just a deprecation annotation. (Whoever the consumers are that care - I don't ;) )

greg0ire:
Let's close then, and let's reopen if a consumer that cares pops up :)

morozov:
FWIW, the version of 2.x releases is increased manually by commits like adce7a954a1c2f14f85e94aed90c8489af204086 which is easy to forget (the process is described in https://github.com/doctrine/dbal/issues/3406).

The fact that this API is deprecated and bumping the release version will produce merge conflicts doesn't prevent us from maintaining the proper version for the next 2.x releases but there's nothing we can do about the already released ones.

phil-davis:
Someone please decide - do I reopen PR #4712 ?

By applying the change there, then even if nothing else happens, the next release will have `VERSION` `2.13.3-DEV`, which will be closer to reality.

greg0ire:
This has been labeled with "Can't fix", so to me, the decision is pretty clear :stuck_out_tongue:

PR #4712, opened 2021-07-14
Update VERSION constant in 2.13 series

phil-davis:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #4711 

#### Summary

Update the VERSION to be current.

### DIFF
--- a/lib/Doctrine/DBAL/Version.php
+++ b/lib/Doctrine/DBAL/Version.php
@@ -17,7 +17,7 @@ class Version
     /**
      * Current Doctrine Version.
      */
-    public const VERSION = '2.12.2-DEV';
+    public const VERSION = '2.13.2';
 
     /**
      * Compares a Doctrine version with the current one.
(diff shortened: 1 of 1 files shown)


phil-davis:
Static Analysis fails because of:
"RedundantCast - lib/Doctrine/DBAL/Version.php:31:48 - The call to strtolower is unnecessary (see https://psalm.dev/262)"

When the constant `VERSION` has only numbers and dots, then `strtolower()` is unnecessary. But as soon as someone adds `-DEV` on the end, `strtolower()` will be necessary!

I guess that the code has to be adjusted every time `-DEV` is added/removed from the string. That's a maintenance headache!

For now, I will also push a commit to move `VERSION` up to 2.13.3-DEV.

phil-davis:
Closing - see comments in issue. Can easily be re-opened if needed some time.

PR #4713, opened 2021-07-15
Handle lost connection during commit

### LINKED ISSUE #4522: Wrapper connection doesn't handle lost connection during commit
An existing test in fact reproduces the issue:
https://github.com/doctrine/dbal/blob/09eafbc7ecc384210e54146b07421b37f7b3869a/tests/Doctrine/Tests/DBAL/Functional/TransactionTest.php#L39-L45

The problems are:
1. The code doesn't throw the `ConnectionLost` exception that would let the wrapper reconnect.
2. The wrapper-level code throws a driver-level exception.

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

Fixes #4522.

The improvement relies on the breaking API changes introduced in https://github.com/doctrine/dbal/pull/3486 (`4.0.x`) and thus cannot be backported to earlier releases as is.

`TransactionTest` being introduced already exists in the older branches but it looks like it was mistakenly dropped during the merge up in https://github.com/doctrine/dbal/pull/3712.

### DIFF
--- a/tests/Functional/TransactionTest.php
+++ b/tests/Functional/TransactionTest.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Tests\Functional;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\PDO;
+use Doctrine\DBAL\Exception\ConnectionLost;
+use Doctrine\DBAL\Platforms\MySQLPlatform;
+use Doctrine\DBAL\Tests\FunctionalTestCase;
+
+use function sleep;
+
+use const E_ALL;
+use const E_WARNING;
+use const PHP_VERSION_ID;
+
+class TransactionTest extends FunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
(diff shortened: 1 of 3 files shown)


tests/Functional/TransactionTest.php
```diff
@@ -0,0 +1,50 @@
...
+        $this->connection->beginTransaction();
+
+        // during the sleep MySQL will close the connection
+        sleep(2);
+
+        // prevent the PHPUnit error handler from handling the "MySQL server has gone away" warning
+        $this->iniSet('error_reporting', (string) (E_ALL & ~E_WARNING));
+
+        $this->expectException(ConnectionLost::class);
+        $this->connection->commit();
```
greg0ire: How hard would it be to also cover the `rollBack` scenario? Wouldn't it be a simple as calling `rollBack` here?

morozov: Yea, I think so (assuming that the extensions report the disconnect). Let me update the test.

morozov: Apparently, not 🚀 . We need to fix https://github.com/doctrine/dbal/issues/4718 first.

greg0ire (approved):

PR #2711, opened 2017-05-04
Update ExpressionBuilder.php

rabihsyw:
parse array values as strings

PR #4716, opened 2021-07-15
Clarify ExpressionBuilder arguments

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

The current API documentation isn't clear enough. Specifically for the `in()` and `notIn()` methods, it says that the second argument is an array of _values_ while it's a list of arbitrary SQL expressions. See https://github.com/doctrine/dbal/pull/2711#issuecomment-876850494.

### DIFF
--- a/src/Query/Expression/ExpressionBuilder.php
+++ b/src/Query/Expression/ExpressionBuilder.php
@@ -256,8 +256,8 @@ class ExpressionBuilder
     /**
      * Creates a LIKE() comparison expression with the given arguments.
      *
-     * @param string $x Field in string format to be inspected by LIKE() comparison.
-     * @param mixed  $y Argument to be used in LIKE() comparison.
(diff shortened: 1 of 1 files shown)


src/Query/Expression/ExpressionBuilder.php
```diff
@@ -256,8 +256,8 @@ public function isNotNull($x)
     /**
      * Creates a LIKE() comparison expression with the given arguments.
      *
-     * @param string $x Field in string format to be inspected by LIKE() comparison.
-     * @param mixed  $y Argument to be used in LIKE() comparison.
+     * @param string $x The expression to be inspected by the NOT LIKE comparison
```
greg0ire:
```suggestion
     * @param string $x The expression to be inspected by the LIKE comparison
```

greg0ire (approved):

PR #4729, opened 2021-07-23
Test exception handling in wrapper Result

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

1. Add testing for exception handling in the wrapper Result. Still cannot find a way to trigger an exception by `rowCount()` or `columnCount()`, maybe the new DBAL API is too fool-proof.
2. Remove some code duplication and redundant error handling in the wrapper Result and Connection.

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -506,11 +506,7 @@ class Connection
      */
     public function fetchAssociative(string $query, array $params = [], array $types = [])
     {
-        try {
-            return $this->executeQuery($query, $params, $types)->fetchAssociative();
-        } catch (Driver\Exception $e) {
-            throw $this->convertExceptionDuringQuery($e, $query, $params, $types);
-        }
+        return $this->executeQuery($query, $params, $types)->fetchAssociative();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Connection.php
```diff
@@ -506,11 +506,7 @@ public function setAutoCommit($autoCommit)
      */
     public function fetchAssociative(string $query, array $params = [], array $types = [])
     {
-        try {
-            return $this->executeQuery($query, $params, $types)->fetchAssociative();
-        } catch (Driver\Exception $e) {
```
morozov, line 511: The above methods cannot throw a driver exception. There's an exception converter involved everywhere which will wrap the driver exception in a wrapper.

src/Connection.php
```diff
@@ -912,15 +888,7 @@ public function fetchFirstColumn(string $query, array $params = [], array $types
      */
     public function iterateNumeric(string $query, array $params = [], array $types = []): Traversable
     {
-        try {
-            $result = $this->executeQuery($query, $params, $types);
-
-            while (($row = $result->fetchNumeric()) !== false) {
```
morozov, line 918: We can call `iterateNumeric()` on the result instead of implementing iteration again.

greg0ire (approved):

PR #4733, opened 2021-07-26
Fix auto_increment implementation for Oracle

### LINKED ISSUE #4725: OraclePlatform::getCreateAutoincrementSql forces sequence value in autoincrement column
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.1.0

#### Summary

The trigger that OraclePlatform::getCreateAutoincrementSql() adds to the database when an autoincrement column is defined does not take into account values passed in directly in the INSERT statements and always overwrites them with the next value of the sequence. 

#### Current behaviour

Look at the trigger code:

https://github.com/doctrine/dbal/blob/a602f2baf74df3d9b17ea8b29fca5fc33786d68d/src/Platforms/OraclePlatform.php#L532-L536

In the very first statement, `sequence.nextval` always forces the next value sequence to the autoincrement column, no matter if there's a value defined already. Other databases allow forcing a value to the autoincrement column, with this code Oracle behaves differently. Weirdly, the rest of the trigger code seems to deal with the case of a value higher that the sequence value to get it up to match an higher value inserted directly; but that can not run, actually, because the first statement has already forced the sequence.nextval to the column to be inserted.

I think just removing that first statement will do.

#### How to reproduce

Create a table with an autoincrement column, try to INSERT a record with a value = 1000 in that column, select back => the value inserted has a value == 1.

#### Expected behaviour

Create a table with an autoincrement column, try to INSERT a record with a value = 1000 in that column, select back => the value inserted has a value == 1000; next insert operation with no value defined will determine a record with value == 1001.

mondrake:
Fixes #4725.

Obviously, different behavior per platform is not something very attractive, but at least with this MR we make Oracle behave like MySql and SQLite, and add tests for all the platforms (except MS Sql that does not allow arbitrary values in INSERT for identity columns).

### DIFF
--- a/src/Platforms/OraclePlatform.php
+++ b/src/Platforms/OraclePlatform.php
@@ -529,7 +529,6 @@ DECLARE
    last_Sequence NUMBER;
    last_InsertID NUMBER;
 BEGIN
-   SELECT ' . $sequenceName . '.NEXTVAL INTO :NEW.' . $quotedName . ' FROM DUAL;
    IF (:NEW.' . $quotedName . ' IS NULL OR :NEW.' . $quotedName . ' = 0) THEN
       SELECT ' . $sequenceName . '.NEXTVAL INTO :NEW.' . $quotedName . ' FROM DUAL;
    ELSE
@@ -540,6 +539,7 @@ BEGIN
       WHILE (last_InsertID > last_Sequence) LOOP
          SELECT ' . $sequenceName . '.NEXTVAL INTO last_Sequence FROM DUAL;
       END LOOP;
+      SELECT ' . $sequenceName . '.NEXTVAL INTO last_Sequence FROM DUAL;
    END IF;
 END;';
 
--- a/tests/Functional/AutoIncrementColumnTest.php
+++ b/tests/Functional/AutoIncrementColumnTest.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace Doctrine\DBAL\Tests\Functional;
+
+use Doctrine\DBAL\Platforms\SQLServer2012Platform;
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\Tests\FunctionalTestCase;
+
+class AutoIncrementColumnTest extends FunctionalTestCase
+{
+    /** @var bool */
+    private $shouldDisableIdentityInsert = false;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $table = new Table('auto_increment_table');
+        $table->addColumn('id', 'integer', ['autoincrement' => true]);
+        $table->setPrimaryKey(['id']);
+
+        $this->connection->createSchemaManager()
+            ->dropAndCreateTable($table);
+    }
+
+    protected function tearDown(): void
+    {
+        if ($this->shouldDisableIdentityInsert) {
+            $this->connection->executeStatement('SET IDENTITY_INSERT auto_increment_table OFF');
+        }
+
+        parent::tearDown();
+    }
+
+    public function testInsertIdentityValue(): void
+    {
+        if ($this->connection->getDatabasePlatform() instanceof SQLServer2012Platform) {
+            $this->connection->executeStatement('SET IDENTITY_INSERT auto_increment_table ON');
+            $this->shouldDisableIdentityInsert = true;
+        }
+
+        $this->connection->insert('auto_increment_table', ['id' => 2]);
+        self::assertEquals(2, $this->connection->fetchOne('SELECT id FROM auto_increment_table'));
+    }
+}
--- a/tests/Platforms/OraclePlatformTest.php
+++ b/tests/Platforms/OraclePlatformTest.php
@@ -358,8 +358,7 @@ class OraclePlatformTest extends AbstractPlatformTestCase
             sprintf(
                 'CREATE TRIGGER %s_AI_PK BEFORE INSERT ON %s FOR EACH ROW DECLARE last_Sequence NUMBER;'
                 . ' last_InsertID NUMBER;'
-                . ' BEGIN SELECT %s_SEQ.NEXTVAL'
-                . ' INTO :NEW.%s FROM DUAL;'
+                . ' BEGIN'
                 . ' IF (:NEW.%s IS NULL OR :NEW.%s = 0)'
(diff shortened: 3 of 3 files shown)


tests/Functional/AutoincrementColumnTest.php
```diff
@@ -0,0 +1,120 @@
...
+        parent::setUp();
+
+        if (! self::$tableCreated) {
+            $table = new Table('auto_increment_table');
+            $table->addColumn('id', 'integer', ['autoincrement' => true]);
+            $table->addColumn('test_int', 'integer');
+            $table->setPrimaryKey(['id']);
+
+            $sm = $this->connection->getSchemaManager();
+            $sm->createTable($table);
```
morozov: Please use `dropAndCreateTable()` and create the table unconditionally. This is effectively what truncation does.

tests/Functional/AutoincrementColumnTest.php
```diff
@@ -0,0 +1,120 @@
...
+    private static $tableCreated = false;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        if (! self::$tableCreated) {
+            $table = new Table('auto_increment_table');
+            $table->addColumn('id', 'integer', ['autoincrement' => true]);
+            $table->addColumn('test_int', 'integer');
```
morozov: Why is this column needed? The `id` column should be sufficient to test how autoincrement works.

tests/Functional/AutoincrementColumnTest.php
```diff
@@ -0,0 +1,120 @@
...
+    {
+        if ($this->connection->getDatabasePlatform()->getName() === 'mssql') {
+            self::markTestSkipped('Test does not work on MSSql.');
+        }
+
+        $this->connection->insert('auto_increment_table', ['test_int' => 1]);
+
+        $this->assertEquals([
+            ['id' => 1, 'test_int' => 1],
+        ], $this->lowerCaseResultColumnNames($this->connection->fetchAllAssociative(
```
morozov: `lowerCaseResultColumnNames()` isn't necessary. If a single column is selected, you can use `fetchOne()`, and we don't need to select all columns.

tests/Functional/AutoincrementColumnTest.php
```diff
@@ -0,0 +1,120 @@
...
+            'SELECT * FROM auto_increment_table ORDER BY id ASC'
+        )));
+
+        $this->connection->insert('auto_increment_table', ['id' => 1000, 'test_int' => 2]);
```
morozov: Why is the rest of the test needed? The above assertion verifies that the resulting value is the same as specified manually (could use a different value than `1` though).

tests/Functional/AutoincrementColumnTest.php
```diff
@@ -0,0 +1,120 @@
...
+        $this->assertEquals([
+            ['id' => 1, 'test_int' => 1],
+            ['id' => 1000, 'test_int' => 2],
+        ], $this->lowerCaseResultColumnNames($this->connection->fetchAllAssociative(
+            'SELECT * FROM auto_increment_table ORDER BY id ASC'
+        )));
+
+        $this->connection->insert('auto_increment_table', ['test_int' => 3]);
+
+        // DBMS autogenerated value after an arbitrary insert differs per platform.
```
morozov: If that's the case, why do we want to get it tested? Let's keep the test scoped to the original issue.

tests/Functional/AutoincrementColumnTest.php
mondrake: OK, thanks.

tests/Functional/AutoincrementColumnTest.php
mondrake: SQLite fails on `insert('autoincrement_table', [])` which would be the way to go here if we were just having a single autoincrement column in the table. Will change anyway so we do not get confused.

tests/Functional/AutoincrementColumnTest.php
mondrake: Changed to use `fetchFirstColumn()`.

tests/Functional/AutoincrementColumnTest.php
mondrake: IMHO we also need to test what happens next time a record is inserted _after_ an arbitrary value is inserted. This tests in Oracle, for instance, that the rest of the trigger code that aligns the sequence value with the max value inserted runs OK.

tests/Functional/AutoincrementColumnTest.php
mondrake: See above.

tests/Functional/AutoincrementColumnTest.php
morozov: Please try using `$platform->etEmptyIdentityInsertSQL()` instead:
https://github.com/doctrine/dbal/blob/891ec185b0a0740a0dfeede0a3eddd2f399bb009/src/Platforms/AbstractPlatform.php#L3421-L3424

tests/Functional/AutoincrementColumnTest.php
morozov: Why does it matter what happens if it's inconsistent across the platforms? The user cannot rely in this behavior anyways.

What does matter though is what happens if we create a new table, and then:
1. Insert an empty row (`1`)
2. Insert `2` manually. At this point, MySQL will bump `AUTO_INCREMENT` to `3` but Oracle will just skip the call to `seq.NEXTVAL`
3. Insert an empty row.

It doesn't matter what the resulting value is since it's inconsistent across platforms. It does matter whether it succeeds or fails.

tests/Functional/AutoincrementColumnTest.php
mondrake: Done, thanks

tests/Functional/AutoincrementColumnTest.php
mondrake: That scenario rules off the test for DB2 and PostgreSql, because the manual insertion in 2 does not update their sequence so 3 fails because it will try to insert a value '2' in the ID again.

morozov:
@mondrake do you mind if I update the test myself and push it to your branch? Otherwise, the review is taking too much time on my end.

mondrake:
Not at all, you’re welcome

morozov (approved):

morozov:
Thanks, @mondrake.

PR #4734, opened 2021-07-27
Add ReturnTypeWillChange to core interface implementations

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | N/A

#### Summary

PHP 8.1 triggers a deprecation, if we implement `Countable` or `IteratorAggregate` without return types. This PR proposes to suppress that warning because we cannot fix this without breaking BC.

### DIFF
--- a/lib/Doctrine/DBAL/Cache/ArrayStatement.php
+++ b/lib/Doctrine/DBAL/Cache/ArrayStatement.php
@@ -10,6 +10,7 @@ use Doctrine\DBAL\FetchMode;
 use InvalidArgumentException;
 use IteratorAggregate;
 use PDO;
+use ReturnTypeWillChange;
 
 use function array_merge;
 use function array_values;
@@ -95,6 +96,7 @@ class ArrayStatement implements IteratorAggregate, ResultStatement, Result
      *
      * @deprecated Use iterateNumeric(), iterateAssociative() or iterateColumn() instead.
      */
+    #[ReturnTypeWillChange]
     public function getIterator()
     {
         $data = $this->fetchAll();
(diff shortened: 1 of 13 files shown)


morozov:
I don't think we should be implementing the support for PHP 8.1 here. Please migrate to 3.x. If it's challenging, let's work on improving the upgrade path.

derrabus:
@morozov The blocker that remains is the ORM. Even if it's ready for DBAL 3 before PHP 8.1 is released, I doubt that the whole ecosystem around the ORM will be ready for it. As much as I would love to see DBAL 2 rot, I'd like to enable the PHP ecosystem to test PHP 8.1 and patching DBAL 2 would help a lot.

And there isn't really much to do, as far as I can tell. Two of the three PRs that I've opened this week are about fixing broken mocks without changing production code. I would volunteer to take care of the 2.13 branch, so you don't have to bother if you don't want to. And I would also volunteer to merge up and do the same work for DBAL 3 and 4. Deal?

morozov:
> I would volunteer to take care of the 2.13 branch, so you don't have to bother if you don't want to. And I would also volunteer to merge up and do the same work for DBAL 3 and 4. Deal?

I'll agree to this as an exception. The effort being put to support DBAL 2 should have been put to update ORM to support DBAL 3. DBAL 3 was released 8 months ago, so there was plenty of time to catch up. Once the support for PHP 8.1 is implemented (presumably, backported from the newer branches to avoid unnecessary merge conflicts), we'll have to have this discussion again, if some more changes like this are necessary in 2.x.

morozov (approved):

derrabus:
> I'll agree to this as an exception.

Thank you very much!

> The effort being put to support DBAL 2 should have been put to update ORM to support DBAL 3.

It's not like I haven't worked on that, see doctrine/orm#8780, doctrine/orm#8794, doctrine/orm#8796. I'd like to do more, but my time and my knowledge of ORM's internals are limited. 😞

morozov:
> It's not like I haven't worked on that [...]

I'm sorry if I came across that way. I didn't mean you personally, you're doing a great job as an external contributor. I meant the collective effort, primarily by the core team.

greg0ire:
Thanks @derrabus !

PR #4735, opened 2021-07-27
PHP 8.1 compatibility for 2.13

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | N/A

#### Summary

The methods `getDateTimeTzFormatString()` and `getDateTimeFormatString()` will never return `null` in real implementations, but this is what the mock created in that test currently does.

Their return value is passed to `DateTime::format()`. Passing `null` to that method triggers a deprecation warning in PHP 8.1. With this PR, I'd like to fix the broken mock.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -25,6 +25,7 @@ jobs:
           - "7.3"
           - "7.4"
           - "8.0"
+          - "8.1"
         dependencies:
           - "highest"
         include:
@@ -43,6 +44,10 @@ jobs:
           php-version: "${{ matrix.php-version }}"
           coverage: "pcov"
           ini-values: "zend.assertions=1"
+
+      - name: "Adjust dependencies"
+        if: "${{ matrix.php-version == '8.1' }}"
+        run: composer config platform.php 8.0.99
 
       - name: "Install dependencies with Composer"
         uses: "ramsey/composer-install@v1"
--- a/lib/Doctrine/DBAL/Schema/AbstractAsset.php
+++ b/lib/Doctrine/DBAL/Schema/AbstractAsset.php
@@ -24,7 +24,7 @@ use function substr;
 abstract class AbstractAsset
 {
     /** @var string */
-    protected $_name;
+    protected $_name = '';
 
     /**
      * Namespace of the asset. If none isset the default namespace is assumed.
--- a/tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
@@ -104,7 +104,7 @@ EOS
             new Schema\ForeignKeyConstraint(
                 ['log'],
                 'log',
-                [''],
+                [],
                 'FK_3',
                 ['onUpdate' => 'SET NULL', 'onDelete' => 'NO ACTION', 'deferrable' => false, 'deferred' => false]
             ),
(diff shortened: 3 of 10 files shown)


tests/Doctrine/Tests/DBAL/Types/VarDateTimeTest.php
```diff
@@ -20,7 +20,10 @@ class VarDateTimeTest extends DbalTestCase
     protected function setUp(): void
     {
         $this->platform = $this->createMock(AbstractPlatform::class);
-        $this->type     = new VarDateTimeType();
+        $this->platform->method('getDateTimeTzFormatString')->willReturn('Y-m-d H:i:s');
+        $this->platform->method('getDateTimeFormatString')->willReturn('Y-m-d H:i:s');
```
morozov: Please see if the changes from https://github.com/doctrine/dbal/pull/4682 could be backported instead.

morozov:
Note, DBAL 2.x is only supported in the part of improving the upgrade path to 3.x. I do not think we should be beating the dead horse and adding the support for PHP 8.1 thus increasing the supported version range from PHP 7.1 to 8.1.

derrabus:
See my comment https://github.com/doctrine/dbal/pull/4734#issuecomment-889822586

Let's continue the discussion there.

derrabus:
@morozov I have backported several changes from #4682 as you suggested, and enabled PHP 8.1 in CI.

.github/workflows/continuous-integration.yml
```diff
@@ -44,6 +45,10 @@ jobs:
           coverage: "pcov"
           ini-values: "zend.assertions=1"
 
+      - name: "Adjust dependencies"
+        if: "${{ matrix.php-version == '8.1' }}"
+        run: composer config platform.php 8.0.99
```
morozov, line 50: Is it temporary before all dependencies explicitly support 8.1?

morozov (approved):

lib/Doctrine/DBAL/Schema/AbstractAsset.php
```diff
@@ -24,7 +24,7 @@
 abstract class AbstractAsset
 {
     /** @var string */
-    protected $_name;
+    protected $_name = '';
```
morozov, line 27: Since it comes from a newer major release, I'd double check if it can cause any BC break. Do such changes belong to a patch release at all or the support for a new backwards-incompatible PHP version deserves a minor release?

tests/Doctrine/Tests/DBAL/Functional/Schema/SqliteSchemaManagerTest.php
```diff
@@ -104,7 +104,7 @@ public function testListForeignKeysFromExistingDatabase(): void
             new Schema\ForeignKeyConstraint(
                 ['log'],
                 'log',
-                [''],
+                [],
```
morozov, line 107: Since it comes from a newer major release, I'd double check if it can cause any BC break.

.github/workflows/continuous-integration.yml
derrabus: Yes. Right now Prophecy as a dependency of PHPUnit would block the installation. As soon as that is resolved, we can remove this block.

lib/Doctrine/DBAL/Schema/AbstractAsset.php
derrabus: All code that I could find that accesses this property already assumes that it is a string. In most cases, the value is used in a string concatenation operation which would implicitly cast it to an empty string if it were `null`.

morozov:
Thanks, @derrabus!

PR #4736, opened 2021-07-27
Don't use null as exception message

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | N/A

#### Summary

`null` is not a valid exception message. Before PHP 8.0, `null` was silently cast to an empty string. On PHP 8.1 however, passing `null` as `$message` to `Exception__construct()` will trigger  a deprecation warning.

This PR proposes to change the tests so that the empty string is used instead of `null`.


SenseException:
The values for `$message` are all the same. Is it necessary to keep them in the data provider?

derrabus:
There is also one data provider that emits actual error messages.

morozov:
Note, these changes are likely already implemented in newer branches. But as I said earlier, I do not think we should be implementing the support for PHP 8.1 in DBAL 2.x.

derrabus:
See my comment https://github.com/doctrine/dbal/pull/4734#issuecomment-889822586

Let's continue the discussion there.

morozov (approved):

greg0ire:
Thanks @derrabus !

PR #4741, opened 2021-07-31
Drop support for PHP 7.3

### LINKED ISSUE #4434: Target PHP 7.4 for 4.0.x
I can see that `3.0.x` currently targets PHP 7.3+; could we make 7.4 the minimum requirement?

Reasons:

- PHP 7.3's active support [ends in 20 days](https://www.php.net/supported-versions.php) (6 December)
- Targeting 7.4 would allow us to:
  - harden the codebase a bit using typed properties
  - make it a bit cleaner by using short closures in a few places.

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

Fixes #4434

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -68,44 +68,32 @@ class Connection
 
     /**
      * The wrapped driver connection.
-     *
-     * @var \Doctrine\DBAL\Driver\Connection|null
      */
-    protected $_conn;
+    protected ?DriverConnection $_conn = null;
 
-    /** @var Configuration */
-    protected $_config;
+    protected Configuration $_config;
 
-    /** @var EventManager */
-    protected $_eventManager;
+    protected EventManager $_eventManager;
 
     /**
      * The current auto-commit mode of this connection.
-     *
-     * @var bool
      */
-    private $autoCommit = true;
+    private bool $autoCommit = true;
 
     /**
      * The transaction nesting level.
-     *
-     * @var int
      */
-    private $transactionNestingLevel = 0;
+    private int $transactionNestingLevel = 0;
 
     /**
      * The currently active transaction isolation level or NULL before it has been determined.
-     *
-     * @var int|null
      */
-    private $transactionIsolationLevel;
+    private ?int $transactionIsolationLevel = null;
 
     /**
      * If nested transactions should use savepoints.
-     *
-     * @var bool
      */
-    private $nestTransactionsWithSavepoints = false;
+    private bool $nestTransactionsWithSavepoints = false;
 
     /**
      * The parameters used during creation of the Connection instance.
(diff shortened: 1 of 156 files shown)


derrabus (dismissed):

src/Connection.php
```diff
@@ -68,44 +68,32 @@ class Connection
 
     /**
      * The wrapped driver connection.
-     *
-     * @var \Doctrine\DBAL\Driver\Connection|null
      */
-    protected $_conn;
+    protected ?\Doctrine\DBAL\Driver\Connection $_conn = null;
```
SenseException: `use Doctrine\DBAL\Driver\Connection`

tests/Tools/Console/RunSqlCommandTest.php
```diff
@@ -16,10 +16,8 @@
 
 class RunSqlCommandTest extends TestCase
 {
-    /** @var CommandTester */
-    private $commandTester;
-    /** @var RunSqlCommand */
-    private $command;
+    private CommandTester $commandTester;
+    private RunSqlCommand $command;
```
SenseException, line 20:
```suggestion

    private RunSqlCommand $command;
```

SenseException (dismissed):

src/Connection.php
derrabus: … `as DriverConnection` (this file already declares a class of that name.)

greg0ire (approved):
Looks OK to me. I'm not to sure about doing such things now on this branch though: it causes a big divergence between branches, so merges up might become more difficult. By the time 4.0 is released, I think 7.3 will have been dropped from 3.x too, won't it?

derrabus:
@greg0ire We're applying pretty much the same changes to Symfony's 6.0 branch right now. Yes, they do cause merge conflicts, but they're easy to fix in most cases.

morozov:
> I'm not to sure about doing such things now on this branch though: it causes a big divergence between branches, so merges up might become more difficult.

Most of the changes in this PR are automated. So if there are new conflicts caused by the property types, they should be possible to resolve in favor of 3.x and then fixed by `phpcbf`. Doesn't look like a big deal to me.

> By the time 4.0 is released, I think 7.3 will have been dropped from 3.x too, won't it?

I don't see much point in doing it. Specifically, I don't see a point in introducing property types in `3.x`: compared to `4.0.x`, most of the methods don't have scalar argument/return types enforced, enforcing them at the property level will lead to a nightmare.

greg0ire:
Thanks to you both! What's wrong with the build?

morozov:
> What's wrong with the build?

I'll disable the PHP 7.3 jobs right before the merge. UPD: there aren't actually other PRs against 4.0.x, so we can safely disable them now.

PR #4746, opened 2021-08-15
Platform-aware schema comparison 

### LINKED ISSUE #2331: Can't convert JSON to LONGTEXT (or MEDIUMTEXT)
Using Laravel migrations and the change() method, one can convert a JSON datatype to a string, and in a subsequent migration to a LONGTEXT or MEDIUMTEXT, but one cannot change JSON to a LONGTEXT or MEDIUMTEXT directly.

### LINKED ISSUE #2411: Comparator miss-detects custom columns as changed
While working a little with custom types in Doctrine lately, I recently noticed a bug in the `Comparator`, although its roots are deeper down. Consider having a custom type like this:

```php
class CustomType extends Type
{
    {
        return $platform->getBinaryTypeDeclarationSQL([
            'length' => '16',
            'fixed' => true,
        ]);
    }

    {
        // …
    }

    {
        // …
    }

    public function requiresSQLCommentHint(AbstractPlatform $platform)
    {
        return true;
    }

    public function getName()
    {
        return 'CustomType';
    }
}
```

Now, when using that type in an entity, at first when generating the database SQL and inserting it, everything is fine. The column has the correct type (binary, fixed 16) and a comment denoting the "CustomType" type.

The problem now is when asking the schema-tool for a diff. Even though the column using that type wasn't changed, the schema-tool suggests an ALTER TABLE statement, changing the column to exactly what it already is.

I looked a bit into this, and it seems that the problem is that the `Schema` generated from the mapping files does not include the columns comment, length, fixed status and also that the type in the `Schema` coming from the DB has `Binary` as type, instead of `CustomType`. This lets the `Comparator` think it needs to update the column.

This problem looks rather complex to me, and I have no real idea on how to solve this, maybe someone else who has a better insight can analyse this further.

### LINKED ISSUE #2566: Column altering in migration from TEXT to LONGTEXT may not apply.

### LINKED ISSUE #2663: Comparator fails for BLOBS

morozov:
## State
The existing property-based column definition comparison has quite some flaws (see all the associated issues):
1. Certain properties, even if different, may yield the same column definition (e.g. the length of a blob column within the type limits).
2. The above is platform-specific, so if a certain type's property is ignored on one platform, it can cause a false-positive match on another platform.

## Proposal
The idea is to replace property-based comparison with the comparison of the SQL that the given column definition generates on the target platform. This way, the diff will be only generated if the column SQL representations on a given platform are different.

### API changes

This is quite an invasive change but it's meant to be safe for a minor release. The changes in the test logic identify the existing bugs rather than breaking changes. 

1. The `Comparator` class now takes an optional `AbstractPlatform` argument. If it's passed, column comparison will be performed in the context of the platform. Not passing a platform is now deprecated.
2. `AbstractPlatform::columnsEqual()` and `Comparator::columnsEqual()` have been added to compare column definitions in the context of the given platform. Depending on the platform, the comparison logic may be more complex than comparing the immediate column SQL. E.g. the default value on SQL Server is represented as a separate constraint.
3. `AbstractSchemaManager::createComparator()` has been added to create the comparator. Besides the schemas being compared and the target platform, the comparison logic may depend on the data available only at runtime (current database collation on SQL Server).
4. Constructors of the `Comparator` class and the subclasses are marked as internal to their schema managers. This will allow potentially improving the comparison logic without introducing breaking changes. The same is already implemented in driver-level connections and statements.

Platform-specific comparators clone the tables, if necessary, and normalize them before passing them to the base implementation. This implementation is far from perfect but it allows to avoid the introduction of a which may be not extensible enough and require a breaking change in the future.

### Testing

Most of the changes in the tests are temporary and are needed to guarantee the parity between the old and the new implementation. The data providers will be removed in the next major release when the dependency of `Comparator` on `AbstractPlatform` becomes mandatory.

1. All functional tests have been reworked to cover the default comparator and the one returned by `$schemaManager->createSchemaManager()`.
2. The unit tests that cover schema comparison in the context of a platform have been reworked to cover the default comparator and the one for the given platform (if it's different from the default).
3. The unit tests covering the `Comparator` outside of the context of a platform have been extended to cover all sub-classes.

## Future scope
1. Array representation of column definitions. Burn with fire. Some of the existing APIs expect `Column $column`, others expect `array<string,mixed> $column`. The latter is an archaism, relies on some magic defaults (see `AbstractPlatfrom::columnToArray()`) and requires unnecessary type conversion back and forth.
2. Deprecate and get rid of `Comparator::diffColumn()` and `ColumnDiff::$changedProperties`. In light of the above changes, the meaning and the purpose of the changed properties become ambiguous. If a property has changed but it doesn't result in a diff in the context of the current platform, should it be included in the changed properties?
3. Mark `SQLServer2012Platform::getDefaultConstraintDeclarationSQL()` and other similar methods internal and then protected. See #4503 for reference.
4. Use parametrized classes instead of the temporarily added assertions.

### DIFF
--- a/src/Driver/AbstractMySQLDriver.php
+++ b/src/Driver/AbstractMySQLDriver.php
@@ -14,6 +14,7 @@ use Doctrine\DBAL\Platforms\MySQLPlatform;
 use Doctrine\DBAL\Schema\MySQLSchemaManager;
 use Doctrine\DBAL\VersionAwarePlatformDriver;
 
+use function assert;
 use function preg_match;
 use function stripos;
 use function version_compare;
@@ -126,6 +127,8 @@ abstract class AbstractMySQLDriver implements VersionAwarePlatformDriver
      */
     public function getSchemaManager(Connection $conn, AbstractPlatform $platform)
     {
+        assert($platform instanceof MySQLPlatform);
+
         return new MySQLSchemaManager($conn, $platform);
     }
 
--- a/src/Schema/Comparator.php
+++ b/src/Schema/Comparator.php
@@ -2,6 +2,8 @@
 
 namespace Doctrine\DBAL\Schema;
 
+use Doctrine\DBAL\Exception;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types;
 use Doctrine\Deprecations\Deprecation;
 
@@ -21,6 +23,27 @@ use function strtolower;
  */
 class Comparator
 {
+    /** @var AbstractPlatform|null */
+    private $platform;
+
+    /**
+     * @internal The comparator can be only instantiated by a schema manager.
+     */
+    public function __construct(?AbstractPlatform $platform = null)
+    {
+        if ($platform === null) {
+            Deprecation::triggerIfCalledFromOutside(
+                'doctrine/dbal',
+                'https://github.com/doctrine/dbal/pull/4659',
+                'Not passing a $platform to %s is deprecated.'
+                    . ' Use AbstractSchemaManager::createComparator() to instantiate the comparator.',
+                __METHOD__
+            );
+        }
+
+        $this->platform = $platform;
+    }
+
     /**
      * Returns a SchemaDiff object containing the differences between the schemas $fromSchema and $toSchema.
      *
@@ -30,8 +53,10 @@ class Comparator
      *
      * @throws SchemaException
      */
-    public static function compareSchemas(Schema $fromSchema, Schema $toSchema)
-    {
+    public static function compareSchemas(
+        Schema $fromSchema,
+        Schema $toSchema
+    ) {
         $comparator       = new self();
         $diff             = new SchemaDiff();
         $diff->fromSchema = $fromSchema;
@@ -203,7 +228,7 @@ class Comparator
      *
      * @return TableDiff|false
      *
-     * @throws SchemaException
+     * @throws Exception
      */
     public function diffTable(Table $fromTable, Table $toTable)
     {
@@ -233,14 +258,20 @@ class Comparator
                 continue;
             }
 
-            // See if column has changed properties in "to" table.
-            $changedProperties = $this->diffColumn($column, $toTable->getColumn($columnName));
+            $toColumn = $toTable->getColumn($columnName);
 
-            if (count($changedProperties) === 0) {
+            // See if column has changed properties in "to" table.
+            $changedProperties = $this->diffColumn($column, $toColumn);
+
+            if ($this->platform !== null) {
+                if ($this->columnsEqual($column, $toColumn)) {
+                    continue;
+                }
+            } elseif (count($changedProperties) === 0) {
                 continue;
             }
 
-            $columnDiff = new ColumnDiff($column->getName(), $toTable->getColumn($columnName), $changedProperties);
+            $columnDiff = new ColumnDiff($column->getName(), $toColumn, $changedProperties);
 
             $columnDiff->fromColumn                               = $column;
             $tableDifferences->changedColumns[$column->getName()] = $columnDiff;
@@ -329,7 +360,7 @@ class Comparator
         $renameCandidates = [];
         foreach ($tableDifferences->addedColumns as $addedColumnName => $addedColumn) {
             foreach ($tableDifferences->removedColumns as $removedColumn) {
-                if (count($this->diffColumn($addedColumn, $removedColumn)) !== 0) {
+                if (! $this->columnsEqual($addedColumn, $removedColumn)) {
                     continue;
                 }
 
@@ -436,10 +467,24 @@ class Comparator
     }
 
     /**
+     * Compares the definitions of the given columns
+     *
+     * @throws Exception
+     */
+    public function columnsEqual(Column $column1, Column $column2): bool
+    {
+        if ($this->platform === null) {
+            return $this->diffColumn($column1, $column2) === [];
+        }
+
+        return $this->platform->columnsEqual($column1, $column2);
+    }
+
+    /**
      * Returns the difference between the columns
      *
-     * If there are differences this method returns $field2, otherwise the
-     * boolean false.
+     * If there are differences this method returns the changed properties as a
+     * string array, otherwise an empty array gets returned.
      *
      * @return string[]
      */
--- a/tests/Functional/Schema/MySQL/ComparatorTest.php
+++ b/tests/Functional/Schema/MySQL/ComparatorTest.php
@@ -1,0 +1,146 @@
+<?php
+
+namespace Doctrine\DBAL\Tests\Functional\Schema\MySQL;
+
+use Doctrine\DBAL\Exception;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Platforms\MySQLPlatform;
+use Doctrine\DBAL\Schema\AbstractSchemaManager;
+use Doctrine\DBAL\Schema\Column;
+use Doctrine\DBAL\Schema\Comparator;
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\Tests\Functional\Schema\ComparatorTestUtils;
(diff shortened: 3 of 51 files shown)


tests/Platforms/AbstractMySQLPlatformTestCase.php
```diff
@@ -769,10 +795,8 @@ public function testDoesNotPropagateDefaultValuesForUnsupportedColumnTypes(): vo
         $diffTable->changeColumn('def_blob', ['default' => null]);
         $diffTable->changeColumn('def_blob_null', ['default' => null]);
 
-        $diff = (new Comparator())->diffTable($table, $diffTable);
-        self::assertNotFalse($diff);
-
-        self::assertEmpty($this->platform->getAlterTableSQL($diff));
```
morozov, line 775: These two assertions cover the bug in the current implementation rather than the expected behavior. The original test and assertions read as follows:
1. Make some schema changes that don't change the resulting SQL.
2. Assert that the diff is not empty. This is the bug being fixed.
3. Assert that even though the diff is not empty, the resulting diff SQL is. This is a workaround of the bug.

tests/Platforms/OraclePlatformTest.php
```diff
@@ -575,13 +575,7 @@ public function testDoesNotPropagateUnnecessaryTableAlterationOnBinaryType(): vo
...
         $table2->addColumn('column_binary', 'binary');
 
-        $comparator = new Comparator();
-
-        // VARBINARY -> BINARY
-        // BINARY    -> VARBINARY
-        $diff = (new Comparator())->diffTable($table1, $table2);
-        self::assertNotFalse($diff);
-        self::assertEmpty($this->platform->getAlterTableSQL($diff));
+        self::assertFalse((new Comparator($this->platform))->diffTable($table1, $table2));
```
morozov, line 578: Same as above, the existing logic covers a bug, not the expected behavior.

src/Driver/AbstractMySQLDriver.php
```diff
@@ -126,6 +127,8 @@ public function getDatabasePlatform()
      */
     public function getSchemaManager(Connection $conn, AbstractPlatform $platform)
     {
+        assert($platform instanceof MySQLPlatform);
```
derrabus, line 130: Would it make sense to add the platform class as a `@template` to the `Driver` interface instead, like you did for `AbstractSchemaManager`?

src/Schema/Comparator.php
```diff
@@ -30,9 +53,12 @@ class Comparator
...
      */
-    public static function compareSchemas(Schema $fromSchema, Schema $toSchema)
-    {
-        $comparator       = new self();
+    public static function compareSchemas(
+        Schema $fromSchema,
+        Schema $toSchema,
+        ?AbstractPlatform $platform = null
+    ) {
+        $comparator       = new self($platform);
```
derrabus: Does this still work, now that we have `Comparator` implementations for the individual drivers? This constructor call will yield a generic `Comparator` and not the implementation specific to the passed platform.

tests/Functional/Schema/MySQL/ComparatorTest.php
```diff
@@ -0,0 +1,146 @@
...
+use Doctrine\DBAL\Tests\Functional\Schema\ComparatorTestUtils;
+use Doctrine\DBAL\Tests\FunctionalTestCase;
+use Doctrine\DBAL\Types\Types;
+
+final class ComparatorTest extends FunctionalTestCase
+{
+    /** @var AbstractPlatform */
+    private $platform;
+
+    /** @var AbstractSchemaManager */
```
derrabus, line 21:
```suggestion
    /** @var AbstractSchemaManager<MySQLPlatform> */
```

tests/Schema/ComparatorTest.php
```diff
@@ -14,13 +14,22 @@
...
 use function get_class;
 
 class ComparatorTest extends TestCase
 {
+    /** @var Comparator */
+    protected $comparator;
+
+    protected function setUp(): void
+    {
+        $this->comparator = new Comparator();
```
derrabus, line 30: You have deprecated not passing a platform to the constructor.

tests/Schema/ComparatorTest.php
```diff
@@ -42,7 +51,7 @@ public function testCompareSame1(): void
 
         $expected             = new SchemaDiff();
         $expected->fromSchema = $schema1;
-        self::assertEquals($expected, Comparator::compareSchemas($schema1, $schema2));
+        self::assertEquals($expected, $this->comparator->compareSchemas($schema1, $schema2));
```
derrabus, line 54: Why this change? The method is still static, isn't it?

src/Driver/AbstractMySQLDriver.php
morozov: It would. I tried that before wrapping up the patch but it looks like it's a rabbit hole (this is the current state: https://github.com/morozov/dbal/commit/8990107aae96b91d8a5b5362aa936c5347db245c).

A certain complexity arises when we parameterize the types like `Connection`, `Driver`, etc. that are used in connection parameters.

We can tackle this issue separately.

src/Schema/Comparator.php
morozov: Good point. The generic comparator supplied with a platform can cause false-positive diffs. It shouldn't be used this way. I'll also adjust the deprecation message in the comparator constructor accordingly.

tests/Schema/ComparatorTest.php
morozov: It is but now the comparator class is extended by platform-specific classes and I want to reuse the same test case to test all implementations. See the classes extending this one.

tests/Schema/ComparatorTest.php
morozov: Yes, but it still has to be covered with a test until the deprecated code is removed.

src/Driver/AbstractMySQLDriver.php
derrabus: Fair enough. 👍🏻

tests/Functional/Schema/MySQL/ComparatorTest.php
morozov: This will require `Connection::createSchemaManager()` and `Driver::createSchemaManager()` to be parametrized which will lead to the same issues as in https://github.com/doctrine/dbal/pull/4746#discussion_r695024202.

src/Schema/Comparator.php
morozov: I rolled this back and deprecated `Schema::getMigrateFromSql()` and `::getMigrateToSql()` that instantiate a comparator internally. Introducing a dependency of `Schema` on `SchemaManager` would be incorrect, and they are just convenience methods that can be replaced with their implementation.

greg0ire (dismissed):
Wow, looks like this is going to be a huge improvement given the number of linked issues :tada:

morozov:
It should be. I haven't check them individually yet but I want to before the merge.

morozov:
The remaining four linked issues are indeed fixed by this PR (the details are in the comments to each issue). I unlinked two since they appear to be unrelated and need to be addressed separately. One more is not reproducible even on `3.1.x` (closed).

greg0ire (approved):

src/Driver/AbstractMySQLDriver.php
morozov: For posterity: the issue of having to perform these assertions was addressed by https://github.com/doctrine/dbal/pull/5465. Now, even though it is possible to parameterize the `Driver` interface with the `AbtractPlatform` type, it doesn't solve any problem (there's no code that expects that a certain driver will return a certain platform).

PR #4750, opened 2021-08-20
Introduce platform family abstraction

### LINKED ISSUE #4749: RFC: deprecate AbstractPlatform::getName()
I should have asked when reviewing https://github.com/doctrine/orm/pull/8870/files#diff-3e426fe47b58d2d4b970b1b7a0cbf5999af950e4f84e48dce4bee5061b57ed31R657 (cc @derrabus ), but what good is a `getName()` method when you can use `instanceof`? Isn't it more fragile?

In the past I remember Symfony deprecating form names, and recommending avoidance of services IDs that are not FQCN, and IMO this has makes things more clear. Should we get rid of `AbstractPlatform::getName()` as well?

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #4749 

#### Summary

The possible return values of `AbstractPlatform::getName()` need to stay
very stable because consumers are relying on those in a lot of
scenarios, probably even more than they rely on `instanceof`, since
doing so for e.g. `PostgreSQLPlatform` no longer works with DBAL 3.
Introducing the possible values as constants of a new type will:
- make these strings explicitly part of our public API;
- help consumers understand better where these strings are defined, and
  hopefully make them add a dependency to doctrine/dbal when they need
  to;
- avoid bugs thanks to IDE auto-completion (less risk of typing
  posgresql instead of postgresql);
- avoid more bugs thanks to static analysis provided by the phpdoc.

Closes #4749

### DIFF
--- a/src/Platforms/AbstractPlatform.php
+++ b/src/Platforms/AbstractPlatform.php
@@ -337,7 +337,7 @@ abstract class AbstractPlatform
     /**
      * Gets the name of the platform.
      *
-     * @return string
+     * @return Family::*
      */
     abstract public function getName();
 
--- a/src/Platforms/DB2Platform.php
+++ b/src/Platforms/DB2Platform.php
@@ -135,7 +135,7 @@ class DB2Platform extends AbstractPlatform
      */
     public function getName()
     {
-        return 'db2';
+        return Family::DB2;
     }
 
     /**
--- a/src/Platforms/Family.php
+++ b/src/Platforms/Family.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Platforms;
+
+interface Family
+{
+    public const DB2        = 'db2';
+    public const MSSQL      = 'mssql';
+    public const MYSQL      = 'mysql';
+    public const ORACLE     = 'oracle';
+    public const POSTGRESQL = 'postgresql';
+    public const SQLITE     = 'sqlite';
+}
--- a/src/Platforms/MySQLPlatform.php
+++ b/src/Platforms/MySQLPlatform.php
@@ -1038,7 +1038,7 @@ SQL
      */
     public function getName()
     {
-        return 'mysql';
+        return Family::MYSQL;
     }
 
     /**
--- a/src/Platforms/OraclePlatform.php
+++ b/src/Platforms/OraclePlatform.php
@@ -1006,7 +1006,7 @@ SQL
      */
     public function getName()
     {
-        return 'oracle';
+        return Family::ORACLE;
     }
 
     /**
--- a/src/Platforms/PostgreSQL94Platform.php
+++ b/src/Platforms/PostgreSQL94Platform.php
@@ -1064,7 +1064,7 @@ SQL
      */
     public function getName()
     {
-        return 'postgresql';
+        return Family::POSTGRESQL;
     }
 
     /**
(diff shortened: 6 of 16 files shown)


SenseException (approved):
`getName` is really a lot used for unequal-, equal- or contain-checks.

morozov:
I'd rather get rid `getName()` and didn't introduce any new enums. Identifying the platform by its family name isn't different from doing `instanceof` which is already possible. Since a platform implements a certain strategy, if you will, it doesn't have to be identifiable, it should just implement its logic.

Although it's not (yet) applicable to the platforms specifically, adding names to strategies makes it harder to decorate them. E.g. if there's a wrapper Foo over the implementation that identifies itself as Bar, how should the wrapped implementation identify itself?

A solid "no" from me.

greg0ire:
> Identifying the platform by its family name isn't different from doing instanceof which is already possible.

It's no longer possible for `PostgreSQLPlatform`, maybe we should resurrect it as an interface?

src/Platforms/AbstractPlatform.php
```diff
@@ -337,7 +337,7 @@ abstract public function getBlobTypeDeclarationSQL(array $column);
     /**
      * Gets the name of the platform.
      *
-     * @return string
+     * @return Family::*
```
andrew-demb, line 340: This return type restriction makes impossible to implement any non-standard platform

morozov:
> It's no longer possible for `PostgreSQLPlatform`, maybe we should resurrect it as an interface?

There's no `PostgreSQLPlatform` class but it's still possible to determine the platform:
```php
if ($platform instanceof PostgreSQL94Platform) {
    // we're on PostrgreSQL
}
```
Why does it have to be named without a version? I don't think it would be a problem to introduce the names without the version for the base platform class. The version in the platform class just makes it clearer which versions it supports.

greg0ire:
It would be in order to rewrite that kind of code without having to care about the version of PG when you only really care about the family: https://github.com/symfony/symfony/pull/42011/files#diff-a508f4ce64bb6d2ab3c496c220fb866aade4e777a464817576f4be0bed306b4dR458 , in a way that would work regardless of the version of the DBAL.

morozov:
We can move all the code from `PostgreSQL94Platform` to `PostgreSQLPlatform` and have the former extend the latter.

derrabus:
> It's no longer possible for `PostgreSQLPlatform`, maybe we should resurrect it as an interface?

I do like that idea. If we had interfaces for all platform families, `instanceof` checks for those interfaces would become stable and the inheritance hierarchy would become an implementation detail for DBAL that nobody needs to rely on. I would be easier to remove versioned platform classes in the future.

morozov:
All platforms are effectively the same from the API standpoint. Why would there be an interface for each?

Note that currently `AbstractPlatform` exposes around 200 public methods. Declaring them all in an interface is insane. Without breaking it down into components, it will turn into a maintenance nightmare: if you need to add a new method (and you'll need to because currently, a platform doesn't have a responsibility, it's just a collection of APIs), it will be a breaking change or will have to be worked around and cause problems with static analysis.

Having interfaces for the APIs is a decent goal but getting there requires a tremendous amount of work.

greg0ire:
> Declaring them all in an interface is insane

It is, and nobody suggested that :stuck_out_tongue: I was just thinking of a marker interface (so without any methods).

> We can move all the code from PostgreSQL94Platform to PostgreSQLPlatform and have the former extend the latter.

This would work too, the end goal being to have a stable _type_ we can use in `instanceof` checks

morozov:
Marker interfaces are okay for exceptions where the `catch` block effectively performs `instanceof`. For APIs that actually implement some logic, marker interfaces are a hack.

For future reference, let's say there is an interface `A` and a marker interface `AExtended`. You want to implement a decorator for `A`. Should it implement `AExtended` as well?

PR #4751, opened 2021-08-24
Deprecate VersionAwarePlatformDriver and ServerInfoAwareConnection

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no

Currently, there are two ways the wrapper `Connection` can instantiate an `AbstractPlatform`:
1. If the driver is `VersionAwarePlatformDriver`, use: https://github.com/doctrine/dbal/blob/b38ee5ec26903e4799bde30acd82df92346b35f0/src/Connection.php#L357
2. Otherwise, fall back to: https://github.com/doctrine/dbal/blob/b38ee5ec26903e4799bde30acd82df92346b35f0/src/Connection.php#L360

Although these two approaches are mutually exclusive, version-aware drivers have to implement _both_ methods. Additionally, there is an assumption that the driver connection used by all `VersionAwarePlatformDriver`s is `ServerInfoAwareConnection` which isn't enforced statically and requires runtime type checks assertions.

Whether or not the driver instantiates a platform based on the server version should be considered an implementation detail which may change when the support of a new platform version is added or removed. For instance, as of https://github.com/doctrine/dbal/pull/3756, the SQL Server drivers do not use the server version to instantiate the platform but the connection is still declared as aware of the server version.

This detail doesn't need to be exposed outside of the driver and doesn't deserve an API.


src/Connection.php
```diff
@@ -441,6 +441,12 @@ private function getServerVersion()
             }
         }
 
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/pulls/4750',
+            'Driver connections that do not implement the ServerInfoAwareConnection interface are deprecated'
```
greg0ire:
```suggestion
            'Not implementing the ServerInfoAwareConnection interface in %s is deprecated',
            get_class($connection)
```

greg0ire (commented):
Are there easy ways to write tests for this?

greg0ire (approved):

PR #4753, opened 2021-08-24
Deprecate AbstractPlatform::getNowExpression()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no

The API is being deprecated for the following reasons:
1. It's not clear if it works with all supported data/time column types since it's not covered by any integration tests.
2. It's undocumented.
3. Its portability is unknown.
4. The clock on the database server may be not in-sync with the clock on the application server which may lead to all sorts of issues.
5. Usage of dates generation by the database makes the application logic harder to test.
6. This API isn't used within the DBAL and isn't used by the ORM.


greg0ire (dismissed):

src/Platforms/AbstractPlatform.php
```diff
@@ -993,10 +993,18 @@ public function getLocateExpression($str, $substr, $startPos = false)
...
      *
+     * @deprecated Generate dates within the application.
+     *
      * @return string
      */
     public function getNowExpression()
     {
+        Deprecation::trigger(
+            'doctrine/dbal',
+            'https://github.com/doctrine/dbal/issues/4753',
```
greg0ire: Is it deliberate to rely on the redirection from issue to PR? I see you didn't do that in https://github.com/doctrine/dbal/pull/4751

morozov: No, it needs to be fixed.

greg0ire (approved):

derrabus (approved):

PR #4755, opened 2021-08-24
Deprecate AbstractPlatform::getName()

### LINKED ISSUE #4749: RFC: deprecate AbstractPlatform::getName()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no

Closes https://github.com/doctrine/dbal/issues/4749.


tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -781,10 +779,12 @@ public function testRenameIndexUsedInForeignKeyConstraint(): void
...
     {
+        $platform = $this->connection->getDatabasePlatform();
+
         if (
-            ! $this->connection->getDatabasePlatform()->supportsInlineColumnComments() &&
-             ! $this->connection->getDatabasePlatform()->supportsCommentOnStatement() &&
-            $this->connection->getDatabasePlatform()->getName() !== 'mssql'
+            ! $platform->supportsInlineColumnComments() &&
+            ! $platform->supportsCommentOnStatement() &&
+            ! $platform instanceof SQLServer2012Platform
```
greg0ire: Should we (re)-introduce `SQLServerPlatform`, `PostgreSQLPlatform` before advising people to write instanceof calls? It would be great to have names stable across major DBAL versions, otherwise this is going to be a pain point I think.

morozov: Let's do.

greg0ire (dismissed):

greg0ire (approved):

tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
emodric: Hi,

I believe there's an issue with `PostgreSQLPlatform` class here. In 2.13, the name of the class is lowercase, `PostgreSqlPlatform`, so BC is not actually maintained between 2.13.x and 3.2.0

derrabus: Class names are case-insensitive in PHP. If this difference in casing creates an actual issue for you, please file a bug report.

emodric: I need to test a bit. PHP did complain about class not existing for me.

derrabus: Well, if you maintain a codebase that should be compatible with DBAL 2 _and_ 3 and you attempt to create instances of that class directly and you operate on a case-sensitive file system, you might need to trigger the autoloader on the old casing to make it work, e.g.:

https://github.com/doctrine/orm/blob/6857a2e8d4eba83444b679231260d6e87fcddfa0/tests/Doctrine/Tests/ORM/Query/SelectSqlGenerationTest.php#L37-L39

emodric: Alright, that helped, thanks!

Issue #3731, opened 2019-11-12
Decouple DDL generation into Platform SQL Builders


guilhermeblanco:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

A clean way to reduce cyclomatic complexity of Platforms is by decoupling DDL generation into a Builder concepts. 

We can discuss the main DDL definition mapping between the following options:

1- Coming from a mapped Table instance:

```
$table = new Table("foo");

$table->addColumn(new Column("id", ...));
$table->addOptions(["collation" => "utf8mb4", "charset" => "utf8mb4_unicode_ci"]);

$builder = new CreateTableBuilder($platform);

$sql = $builder->build($table);
```

2- Coming from a CreateTableBuilder:

```
$builder = new CreateTableBuilder($platform);

$builder->withName("foo");
$builder->withColumn(new Column("id", ...));
$builder->withOptions(["collation" => "utf8mb4", "charset" => "utf8mb4_unicode_ci"]);

$sql = $builder->build();
```

3- Coming from a Platform generated Builder:

```
$builder = $platform->createTableBuilder();

$builder->withName("foo");
$builder->withColumn(new Column("id", ...));
$builder->withOptions(["collation" => "utf8mb4", "charset" => "utf8mb4_unicode_ci"]);

$sql = $builder->build();
```

4- Coming from a CreateTable object:

```
$createTable = new CreateTable("foo");

$createTable->addColumn(new Column("id", ...));
$createTable->addOptions(["collation" => "utf8mb4", "charset" => "utf8mb4_unicode_ci"]);

$success = $platform->createTable($createTable);
```

5- Coming from Factory object:

```
$table = new Table("foo");

$table->addColumn(new Column("id", ...));
$table->addOptions(["collation" => "utf8mb4", "charset" => "utf8mb4_unicode_ci"]);

$createTable = CreateTable::fromTable($table);
$success = $platform->createTable($createTable);
```

Surely mix and match of procedures or ideas can be applied here. My personal take is that we should make the Schema data structures Immutable.
DDL generation should be simple as most RDBMs support standard SQL-99.

PR #4758, opened 2021-08-26
allow setting operator classes for indexes in postgresql

icewind1991:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | 

#### Summary

This allows setting [operator classes](https://www.postgresql.org/docs/current/indexes-opclass.html) on postgresql indexes, which is required for making the query usable in `LIKE` expressions when using a non `C` locale.

### DIFF
--- a/src/Platforms/AbstractPlatform.php
+++ b/src/Platforms/AbstractPlatform.php
@@ -3118,6 +3118,14 @@ abstract class AbstractPlatform
     }
 
     /**
+     * Whether the platform supports operator classes for indexes.
+     */
+    public function supportsIndexOperatorClasses(): bool
+    {
+        return false;
+    }
+
+    /**
      * Whether the platform supports altering tables.
      *
      * @return bool
--- a/src/Platforms/PostgreSQL94Platform.php
+++ b/src/Platforms/PostgreSQL94Platform.php
@@ -186,6 +186,11 @@ class PostgreSQL94Platform extends AbstractPlatform
         return true;
     }
 
+    public function supportsIndexOperatorClasses(): bool
+    {
+        return true;
+    }
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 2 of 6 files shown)


morozov:
> which is required for making the query usable in `LIKE` expressions when using a non `C` locale.

Is it already possible on other platforms? I'd start with an integration test that describes what this change gives to the API consumers.

icewind1991:
Other platforms can use indexes with `LIKE` queries (as long as the wildcard is at the end) as far as I know.

From my understanding pgsql defaults to creating an index that is setup for with locale aware collation rules for sorting and equality, by setting the operator class to `text_pattern_ops` pgsql instead makes an index for strict character by character sorting.

I'm not sure how to create an integration test for this as the only query still works without the opclass set, just without using the index

morozov:
> I'm not sure how to create an integration test for this as the only query still works without the opclass set, just without using the index

The test should reproduce the problem being solved. I'm not familiar with PostgreSQL well enough to suggest a scenario. Why do you want to use this feature that is currently unsupported? Does it improve performance or affects the query results? In the latter case, the test would be about making the query and making the assertion against the result.

icewind1991:
As with any index, it being used or not only impact the performance, without the correct index pgsql needs to do a full table scan, with a proper index it only needs to walk the index

src/Platforms/AbstractPlatform.php
```diff
@@ -3117,6 +3117,14 @@ public function supportsColumnLengthIndexes(): bool
         return false;
     }
 
+    /**
+     * Whether the platform supports operator classes for indexes.
+     */
+    public function supportsIndexOperatorClasses(): bool
```
morozov, line 3123: I'm really hesitant about accepting an API that requires a "supports" method and is applicable to a single platform. I'd rather see the schema API being extensible so that the consumers can implement the features specific to their platforms w/o introducing core API changes.

For instance, currently, all the features of `Index` are implemented as "options" which are `array<string,mixed>`. This API is far from perfect because if there's a typo in the option name, it will be silently ignored.

There are some ideas proposed to improve this in https://github.com/doctrine/dbal/issues/3731 and https://github.com/doctrine/dbal/issues/4772. Although those are rather long-term projects and they won't necessarily cover this case, they could be used as inspiration for a better API for this feature.

morozov:
Closing as stale.

Issue #3346, opened 2018-11-14
mssql 2016 SQLServer2012Platform::getConcatExpression() missing/wrong


Lacoi:
### Bug Report
Since SQL Server 2012 it is possible to use CONCAT() https://docs.microsoft.com/de-de/sql/t-sql/functions/concat-transact-sql?view=sql-server-2016
When i use `CONCAT(column1. '-', column2) LIKE :search)` it converts to `(f0_.column1 + '-' + f0_.column2)` which is not working in the where clauses.

|    Q        |   A
|------------ | ------
| Version     | 2.8.0

#### Solution
Add to Doctrine\DBAL\Platforms\SQLServer2012Platform
```php
   /**
     * {@inheritDoc}
     */
    public function getConcatExpression()
    {
        $args = func_get_args();

        return 'CONCAT(' . join(', ', (array) $args) . ')';
    }
```

Ocramius:
This requires a test case showing the wanted/valid SQL syntax being produced for SQL Server 2012+.

rogeriolino:
I'm trying to use `CONCAT` via DQL to show a period of years in my report page, but the result is the sum of fields.

Statement:

    CONCAT(fieldYear1, ' - ', fieldYear2)

Expected: 

    2010 - 2019

Got:

    4029

To fix it, I'd created a custom DQL `CONCAT` function that translate to SQL Server `CONCAT`.

PR #4773, opened 2021-09-07
[2.x] Allow support for CONCAT() in SQL Server 2012

yesdevnull:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3346

#### Summary

As per [SQL Server 2012](https://docs.microsoft.com/en-us/previous-versions/sql/sql-server-2012/hh231515(v=sql.110)) ([and onwards](https://docs.microsoft.com/en-us/sql/t-sql/functions/concat-transact-sql?view=sql-server-2016)), the `CONCAT()` function is available for use to generate valid T-SQL.

Backported this to 2.13.x as I'm still using 2.x in a project but upgrading to 3.x soon™️.

PR #4774, opened 2021-09-07
[3.x] Allow support for CONCAT() in SQL Server 2012

yesdevnull:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #3346

#### Summary

As per [SQL Server 2012](https://docs.microsoft.com/en-us/previous-versions/sql/sql-server-2012/hh231515(v=sql.110)) ([and onwards](https://docs.microsoft.com/en-us/sql/t-sql/functions/concat-transact-sql?view=sql-server-2016)), the `CONCAT()` function is available for use to generate valid T-SQL.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -4,12 +4,12 @@
     Procedure:
         1) Save a copy of this file with a name of your choosing. It doesn't matter
            where you place it as long as you know where it is.
-           i.e. "mysqlconf.xml" (It needs the ending .xml).
+           i.e. "mysqlconf.phpunit.xml" (It needs the ending .xml).
         2) Edit the file and fill in your settings (database name, type, username, etc.)
            Just change the "value"s, not the names of the var elements.
         3) To run the tests against the database type the following from within the
            tests/ folder: phpunit -c <filename> ...
-           Example: phpunit -c mysqlconf.xml
+           Example: phpunit -c mysqlconf.phpunit.xml
 -->
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
--- a/src/Platforms/SQLServer2012Platform.php
+++ b/src/Platforms/SQLServer2012Platform.php
@@ -1146,9 +1146,7 @@ class SQLServer2012Platform extends AbstractPlatform
      */
     public function getConcatExpression()
     {
-        $args = func_get_args();
-
-        return 'CONCAT(' . implode(', ', $args) . ')';
+        return sprintf('CONCAT(%s)', implode(', ', func_get_args()));
     }
 
     /**
--- a/src/Platforms/SQLServer2012Platform.php
+++ b/src/Platforms/SQLServer2012Platform.php
@@ -1148,7 +1148,7 @@ class SQLServer2012Platform extends AbstractPlatform
     {
         $args = func_get_args();
 
-        return '(' . implode(' + ', $args) . ')';
+        return 'CONCAT(' . implode(', ', $args) . ')';
     }
 
     /**
--- a/tests/Functional/Platform/ConcatTest.php
+++ b/tests/Functional/Platform/ConcatTest.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace Doctrine\DBAL\Tests\Functional\Platform;
+
+use Doctrine\DBAL\Platforms\MySQLPlatform;
+use Doctrine\DBAL\Platforms\SQLServer2012Platform;
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\Tests\FunctionalTestCase;
+
+class ConcatTest extends FunctionalTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        if ($this->connection->getDatabasePlatform() instanceof MySQLPlatform) {
+            return;
+        }
+
+        if ($this->connection->getDatabasePlatform() instanceof SQLServer2012Platform) {
+            return;
+        }
+
+        self::markTestSkipped('Restricted to MySQL and SQL Server.');
+    }
+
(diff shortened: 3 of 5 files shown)


SenseException:
Let's see what #4773 will bring that might affect this PR too. I'm not sure if an upmerge of any changes there is possible, so I keep this PR open.

morozov:
@yesdevnull please make sure that there's a functional test that reproduces the issue being fixed.

tests/Functional/Schema/SQLServerSchemaManagerTest.php
```diff
@@ -404,4 +404,11 @@ public function testPkOrdering(): void
         self::assertEquals('colB', $columns[0]);
         self::assertEquals('colA', $columns[1]);
     }
+
+    public function testConcat(): void
+    {
+        $result = $this->connection->fetchOne('SELECT CONCAT(1, 2, 3)');
+
+        self::assertEquals('123', $result);
+    }
```
yesdevnull: @morozov: I'm not that familiar with the Doctrine test suites, is this an appropriate place for the platform-specific test?

phpunit.xml.dist
```diff
@@ -4,12 +4,12 @@
     Procedure:
         1) Save a copy of this file with a name of your choosing. It doesn't matter
            where you place it as long as you know where it is.
-           i.e. "mysqlconf.xml" (It needs the ending .xml).
+           i.e. "mysqlconf.phpunit.xml" (It needs the ending .xml).
```
yesdevnull, line 7: I made these changes because `/*.phpunit.xml` is already in the `.gitignore` so others can avoid accidentally committing their custom PHPUnit config files.

tests/Functional/Schema/SQLServerSchemaManagerTest.php
morozov: It's sort of the right place but it would be better to create a separate test that would run on all supported platforms. Please see the ones located in [tests/Functional/Platform](https://github.com/doctrine/dbal/tree/3.1.x/tests/Functional/Platform) for example.

tests/Functional/Schema/SQLServerSchemaManagerTest.php
```diff
@@ -404,4 +404,11 @@ public function testPkOrdering(): void
         self::assertEquals('colB', $columns[0]);
         self::assertEquals('colA', $columns[1]);
     }
+
+    public function testConcat(): void
+    {
+        $result = $this->connection->fetchOne('SELECT CONCAT(1, 2, 3)');
```
morozov: This test doesn't cover the code being changed. It just performs an SQL query. BTW, why do you need to concatenate integers?

The test should do something like this but with your logic in place: https://github.com/doctrine/dbal/blob/e6add2943187cdf01599e198a77bb6d3672703ab/tests/Functional/Platform/QuotingTest.php#L17-L22

yesdevnull: Terrific, thanks for pointing out that other test. I'll get this updated.

(The concat of ints was a very trivial example, I'll change it to something more complex.)

yesdevnull: Moved to a [Concat-specific test](https://github.com/yesdevnull/dbal/blob/3.1.x/tests/Functional/Platform/ConcatTest.php#L27-L44). The concat is still fairly trivial however it does test whether `CONCAT()` will work or not. Hopefully this is much closer to what you're after 🙂

morozov: It doesn't have to be complex. It should be as trivial as your original query but the test should use the method that you're changing to build the query.

tests/Functional/Platform/ConcatTest.php
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        self::markTestSkipped('Restricted to MySQL and SQL Server.');
+    }
+
+    public function testConcat(): void
+    {
+        $table = new Table('concat_test');
+        $table->addColumn('name', 'string');
+        $this->connection->getSchemaManager()->dropAndCreateTable($table);
```
morozov, line 31: Why is it necessary? Concatenation in SQL can be tested without a table. Please see the example I pointed to earlier.

tests/Functional/Platform/ConcatTest.php
```diff
@@ -0,0 +1,45 @@
...
+
+        if ($this->connection->getDatabasePlatform() instanceof MySQLPlatform) {
+            return;
+        }
+
+        if ($this->connection->getDatabasePlatform() instanceof SQLServer2012Platform) {
+            return;
+        }
+
+        self::markTestSkipped('Restricted to MySQL and SQL Server.');
```
morozov, line 24: Doesn't concatenation work on all supported platforms?

tests/Functional/Platform/ConcatTest.php
```diff
@@ -0,0 +1,45 @@
...
+
+        $platform = $this->connection->getDatabasePlatform();
+        $query    = $this->connection->fetchOne(
+            \sprintf(
+                'SELECT name FROM concat_test WHERE name LIKE %s',
+                $platform->getConcatExpression("'%'", "'long'", "'%'")
+            )
+        );
+
+        self::assertEquals('A long string', $query);
```
morozov, line 43: This test doesn't seem to be testing concatenation in SQL.

src/Platforms/SQLServer2012Platform.php
```diff
@@ -1146,9 +1146,7 @@ public function getTrimExpression($str, $mode = TrimMode::UNSPECIFIED, $char = f
      */
     public function getConcatExpression()
```
mvorisek, line 1147:
```suggestion
    public function getConcatExpression(...$args)
```

morozov:
Closing in favor of https://github.com/doctrine/dbal/pull/4939.

PR #4781, opened 2021-09-09
SchemaDiff - remove orphan foreign keys before trying to drop related indexes

seferov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

Currently when save mode is enabled, SchemaDiff tries to drop indexes that are needed for foreign keys thus results in error.
With this change, SchemaDiff will only drop orphan foreign keys that their related indexes are going to be dropped.

### DIFF
--- a/lib/Doctrine/DBAL/Schema/SchemaDiff.php
+++ b/lib/Doctrine/DBAL/Schema/SchemaDiff.php
@@ -162,6 +162,15 @@ class SchemaDiff
         }
 
         foreach ($this->changedTables as $tableDiff) {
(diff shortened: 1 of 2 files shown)


morozov:
@seferov please add a functional test that reproduces the issue and retarget against `3.1.x`.

seferov:
@morozov 

>please add a functional test that reproduces the issue

sure

> and retarget against 3.1.x.

Is it going to be backported to 2.13.x?

morozov:
No. Only improvements to the upgrade path to DBAL 3 are accepted to the 2.x branches.

morozov:
Closing in favor of #4782.

PR #4782, opened 2021-09-09
SchemaDiff - remove orphan foreign keys before trying to drop related…

seferov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

Currently when save mode is enabled, SchemaDiff tries to drop indexes that are needed for foreign keys thus results in error.
With this change, SchemaDiff will only drop orphan foreign keys that their related indexes are going to be dropped.

### DIFF
--- a/src/Schema/SchemaDiff.php
+++ b/src/Schema/SchemaDiff.php
@@ -162,6 +162,17 @@ class SchemaDiff
         }
 
         foreach ($this->changedTables as $tableDiff) {
+            // make sure to drop related orphan foreign keys before dropping indexes that are related to FKs
+            if ($saveMode === true && $platform->supportsForeignKeyConstraints() && null !== $tableDiff->fromTable) {
+                foreach ($this->orphanedForeignKeys as $orphanedForeignKey) {
+                    if ($orphanedForeignKey->getLocalTable()->getName() !== $tableDiff->fromTable->getName()) {
+                        continue;
+                    }
+
+                    $sql[] = $platform->getDropForeignKeySQL($orphanedForeignKey, $orphanedForeignKey->getLocalTable());
+                }
+            }
+
             $sql = array_merge($sql, $platform->getAlterTableSQL($tableDiff));
         }
 
--- a/src/Schema/SchemaDiff.php
+++ b/src/Schema/SchemaDiff.php
@@ -163,7 +163,7 @@ class SchemaDiff
 
         foreach ($this->changedTables as $tableDiff) {
             // make sure to drop related orphan foreign keys before dropping indexes that are related to FKs
-            if ($saveMode === true && $platform->supportsForeignKeyConstraints() && null !== $tableDiff->fromTable) {
+            if ($saveMode === true && $platform->supportsForeignKeyConstraints() && $tableDiff->fromTable !== null) {
                 foreach ($this->orphanedForeignKeys as $orphanedForeignKey) {
                     if ($orphanedForeignKey->getLocalTable()->getName() !== $tableDiff->fromTable->getName()) {
                         continue;
--- a/tests/Functional/Schema/Db2SchemaManagerTest.php
+++ b/tests/Functional/Schema/Db2SchemaManagerTest.php
@@ -30,14 +30,4 @@ class Db2SchemaManagerTest extends SchemaManagerFunctionalTestCase
         self::assertNull($columns['bool']->getComment());
         self::assertSame("That's a comment", $columns['bool_commented']->getComment());
     }
-
-    public function testSaveModeHandleOrphanForeignKeysBeforeIndexDrop(): void
-    {
-        $diff = $this->diffSaveModeHandleOrphanForeignKeysBeforeIndexDrop();
-
-        $this->assertSame([
-            'ALTER TABLE test_save_mode_orphan_fk_foreign DROP FOREIGN KEY FK_52D644CEA81E660E',
-            'DROP INDEX UNIQ_52D644CEA81E660E',
-        ], $diff->toSaveSql($this->schemaManager->getDatabasePlatform()));
-    }
 }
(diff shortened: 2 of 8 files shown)


tests/Functional/Schema/Db2SchemaManagerTest.php
```diff
@@ -30,4 +30,14 @@ public function testGetBooleanColumn(): void
         self::assertNull($columns['bool']->getComment());
         self::assertSame("That's a comment", $columns['bool_commented']->getComment());
     }
+
+    public function testSaveModeHandleOrphanForeignKeysBeforeIndexDrop(): void
+    {
+        $diff = $this->diffSaveModeHandleOrphanForeignKeysBeforeIndexDrop();
+
+        $this->assertSame([
+            'ALTER TABLE test_save_mode_orphan_fk_foreign DROP FOREIGN KEY FK_52D644CEA81E660E',
```
morozov: These tests aren't really useful. Most likely you ran your code and copy/pasted the generated query as expected. The expected value is by no means a guarantee that the code will work as expected when it comes to interacting with a real database.

tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -1422,6 +1423,32 @@ public function testCommentInTable(): void
...
+        $foreignTable->addUniqueIndex(['fk']);
+        $foreignTable->addForeignKeyConstraint('test_save_mode_orphan_fk_primary', ['fk'], ['id']);
+
+        $toSchemaForeignTable = new Table('test_save_mode_orphan_fk_foreign');
+        $toSchemaForeignTable->addColumn('fk', 'integer');
+
+        $fromSchema = new Schema([$primaryTable, $foreignTable]);
+        $toSchema   = new Schema([$toSchemaForeignTable]);
+
+        return Comparator::compareSchemas($fromSchema, $toSchema);
```
morozov: This isn't a functional test, it doesn't interact with a real database. It should execute the SQL on a real database and confirm that the code works as expected by introspecting the resulting schema.

tests/Schema/SchemaDiffTest.php
```diff
@@ -105,7 +105,7 @@ private function createPlatform(bool $unsafe)
         $platform->expects(self::exactly(1))
                 ->method('supportsSequences')
                 ->will(self::returnValue(true));
-        $platform->expects(self::exactly(2))
+        $platform->expects(self::atLeastOnce())
```
morozov: Why does the test care how many times the `supportsForeignKeyConstraints()` method is invoked? If the change is necessary, please replace it with `self::any()`.

src/Schema/SchemaDiff.php
```diff
@@ -162,6 +162,17 @@ protected function _toSql(AbstractPlatform $platform, $saveMode = false)
         }
 
         foreach ($this->changedTables as $tableDiff) {
+            // make sure to drop related orphan foreign keys before dropping indexes that are related to FKs
```
morozov, line 165: I'm not sure how to interpret this documentation:

https://github.com/doctrine/dbal/blob/d6fd6957206e95a73ea7c77a0a7de8905d9fdc8a/src/Schema/SchemaDiff.php#L80-L84

Does it mean that foreign keys should not be dropped in this mode? As far as can tell from the code, that's the case: 

https://github.com/doctrine/dbal/blob/d6fd6957206e95a73ea7c77a0a7de8905d9fdc8a/src/Schema/SchemaDiff.php#L118-L122

If I replace `toSaveSql()` with `toSql()`, it doesn't fail with the original error (Cannot drop index '%s': needed in a foreign key constraint).

Should you just use `toSql()` instead? Otherwise, the patch seems to contradict the documentation.

src/Schema/SchemaDiff.php
```diff
@@ -162,6 +162,17 @@ protected function _toSql(AbstractPlatform $platform, $saveMode = false)
...
+            if ($saveMode === true && $platform->supportsForeignKeyConstraints() && $tableDiff->fromTable !== null) {
+                foreach ($this->orphanedForeignKeys as $orphanedForeignKey) {
+                    if ($orphanedForeignKey->getLocalTable()->getName() !== $tableDiff->fromTable->getName()) {
+                        continue;
```
morozov, line 169: If this line is commented out, the test will still pass (at least on MySQL). Is this condition tested?

src/Schema/SchemaDiff.php
seferov: > I'm not sure how to interpret this documentation:
> Does it mean that foreign keys should not be dropped in this mode? As far as can tell from the code, that's the case:


_"following things don't happen: 3. Foreign Keys which reference tables that would otherwise be deleted."_

What I understand from the documentation is that it doesn't delete orphan foreign keys unless its reference is in use.

> 
> 
> If I replace `toSaveSql()` with `toSql()`, it doesn't fail with the original error (Cannot drop index '%s': needed in a foreign key constraint).
> 
> Should you just use `toSql()` instead? Otherwise, the patch seems to contradict the documentation.

`toSql()` (which means don't save) drops unmapped tables as well. In this case, of course, there is no problem since it drops all related FK and indexes. So this is not an option for those who wants to **save**.

Currently with the save mode, in this particular case, the generated queries results in an error. Thus it doesn't only contradict with the documentation and the command itself as well.

src/Schema/SchemaDiff.php
seferov: added test for it in https://github.com/doctrine/dbal/pull/4782/commits/9dee9ae2a07ff16bbf9073ee0178648d070f7c01

src/Schema/SchemaDiff.php
morozov: Thanks for the details, @seferov. I need some time to digest this but I'm currently busy with other things. I'll try to get to this issue once I can.

seferov:
@derrabus why did you close the PR?
@morozov did you have a chance to digest it?

greg0ire:
It was accidental: we have removed 3.1.x, and we couldn't reopen your PR because your fork is apparently gone.

derrabus:
@seferov Can you open a new PR, targeting the 3.2.x branch? Sorry for that.

PR #4783, opened 2021-09-09
sqlite expression based index

recca0120:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

when use expression based index, index column name would be null, must parse column name from sqlite_master

### DIFF
--- a/tests/Functional/Schema/SqliteSchemaManagerTest.php
+++ b/tests/Functional/Schema/SqliteSchemaManagerTest.php
@@ -236,4 +236,23 @@ SQL;
             ->getColumn('col1')
             ->getComment());
     }
+    
+    public function testListTableIndexesWithIndexExpression(): void
+    {
+        $sql = <<<SQL
+CREATE TABLE company (
+    com_id int(4),
+    com_name text(15)
+);
+CREATE INDEX com_id_index ON company (com_id);
+CREATE INDEX com_name_5_index ON company (substr(com_name, 0, 5))
+SQL;
+
+        $this->connection->executeStatement($sql);
+
+        $indexes = $this->schemaManager->listTableIndexes('company');
+
+        self::assertArrayHasKey('com_id_index', $indexes);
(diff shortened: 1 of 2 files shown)


tests/Functional/Schema/SqliteSchemaManagerTest.php
```diff
@@ -236,4 +236,23 @@ public function testOnlyOwnCommentIsParsed(): void
...
+    
+    public function testListTableIndexesWithIndexExpression(): void
+    {
+        $sql = <<<SQL
+CREATE TABLE company (
+    com_id int(4),
+    com_name text(15)
+);
+CREATE INDEX com_id_index ON company (com_id);
+CREATE INDEX com_name_5_index ON company (substr(com_name, 0, 5))
```
morozov, line 248: Given that the test creates the table via plain SQL, I assume that the DBAL doesn't support such indices.

tests/Functional/Schema/SqliteSchemaManagerTest.php
```diff
@@ -236,4 +236,23 @@ public function testOnlyOwnCommentIsParsed(): void
...
+SQL;
+
+        $this->connection->executeStatement($sql);
+
+        $indexes = $this->schemaManager->listTableIndexes('company');
+
+        self::assertArrayHasKey('com_id_index', $indexes);
+        self::assertEquals('substr(com_name, 0, 5)', $indexes['com_name_5_index']->getColumns()[0]);
```
morozov, line 256: https://github.com/doctrine/dbal/blob/936e7e3350dd15a2d3de147821dfe5249169ed8b/src/Schema/Constraint.php#L22-L28

`substr(com_name, 0, 5)` isn't a column name. It looks like the `Index` abstraction in the DBAL wasn't designed for exposing column expressions.

morozov (commented):
The problem being solved doesn't look like a bug. If desired, the support for expression-based indexes should be designed at that API level first and implemented on all the platforms that support them.

morozov:
Closing as stale.

PR #4785, opened 2021-09-10
Deprecate not passing $fromColumn to ColumnDiff

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no

The `$fromColumn` parameter is optional since its introduction in https://github.com/doctrine/dbal/pull/220. It had to be declared as optional to avoid a BC break. The fact that not passing it wasn't immediately deprecated, requires the code that interprets the diff to interpret it in two modes: with and without the `$fromColumn`.

Given that the property-based comparison is deprecated (https://github.com/doctrine/dbal/pull/4746), the next step is to stop using `$changedProperties` and only rely on the actual column definitions. For that, we need to require `$fromColumn` to be always set.

Once we stop using `$changedProperties` in `4.0.x`, we can deprecate it.

### DIFF
--- a/src/Schema/ColumnDiff.php
+++ b/src/Schema/ColumnDiff.php
@@ -2,6 +2,8 @@
 
 namespace Doctrine\DBAL\Schema;
 
+use Doctrine\Deprecations\Deprecation;
+
 use function in_array;
 
 /**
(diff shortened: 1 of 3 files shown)


src/Schema/ColumnDiff.php
```diff
@@ -31,6 +34,15 @@ public function __construct(
         array $changedProperties = [],
         ?Column $fromColumn = null
     ) {
+        if (func_num_args() < 4) {
```
greg0ire: :thinking: so, passing `null` is not deprecated? That seems to contradict #4786 . Should this test be changed into a `null` or `instanceof Column` check?

morozov: Good point.

greg0ire (approved):

Issue #4772, opened 2021-09-03
RFC: Improve the APIs around handling SQL object names


morozov:
## Motivation

There's a bunch of issues around handling object names that exist due to the existing design assumptions (e.g. https://github.com/doctrine/dbal/issues/4357). The assumptions are:

1. Only the identifiers that are SQL keywords should be quoted. The list of keywords managed by the DBAL is always up to date.
2. Object names are case-insensitive on all platforms under any circumstances.
3. Unqualified object names cannot contain dots (`.`).
4. Object names cannot contain special characters (e.g. `` ` ``).

The above assumptions lead to the API where any name is represented as a string which then has to be _parsed_ to identify its meaning. E.g.:

1. The quotes are stripped from a quoted name but without considering the escaping rules.
2. A dot in the middle of the name is considered to be the delimiter between the name and its qualifier.

## Proposal

This design proposal attempts to replace parsing the string with representing names as domain objects.

### The name interface

The corner-stone interface describing object names will be `Name`:
```php
/**
 * SQL object name.
 */
interface Name
{
    /**
     * Returns object name representation as an SQL literal.
     */
    public function toLiteral(NameBuilder $builder): string;

    /**
     * Returns object name representation as an SQL identifier.
     */
    public function toIdentifier(NameBuilder $builder): string;
}
```
It will represent any name: qualified or unqualified; quoted or unquoted and will be used by all SQL objects like tables, views, columns, databases, etc.

Note, currently, there's no plan to expose the raw value of the name. This will make it easier to abuse it in the raw form which may be the source of all sorts of issues (e.g. SQL injections).

A name can be used in two forms in SQL, hence the methods:

1. **A literal**. The name will be represented as an SQL literal primarily during schema introspection. For instance:
   ```sql
   SELECT * FROM columns WHERE table = <literal>
   ```
2. **An identifier.** The name will be represented as an SQL identifier when the corresponding object is used an an SQL statement. For instance:
   ```sql
   SELECT * FROM <identifier>
   ```

### The SQL Builder interface family

The string representations of the name will be built by an SQL builder that is specific to the given database platform. For the sake of simplicity, we can consider the platform and the SQL builder synonymous for now.

```php
/**
 * SQL builder that builds SQL representing object names. Each database platform must provide an implementation
 * of this interface according to its logic of handling object names.
 */
interface NameBuilder
{
    /**
     * Builds an SQL literal of an unquoted object name.
     */
[...]

PR #4795, opened 2021-09-13
fix mysql identifier quoting fail

paulmhh:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | ?
| Fixed issues | #4357

#### Summary

\Doctrine\DBAL\Schema\AbstractAsset::getQuotedName()

does this:

        $keywords = $platform->getReservedKeywordsList();
        $parts    = explode('.', $this->getName());
            $parts[$k] = $this->_quoted || $keywords->isKeyword($v) ? $platform->quoteIdentifier($v) : $v;
        }

as mentioned in #4357 there is no api available if an identifier needs to be quoted or not. I just added this to the Keywords implementation. Not saying it should be done like that, but its a working workaround for me, and it should be added as an platform abstraction properly.

this definitely needs fixing! see also https://dev.mysql.com/doc/refman/5.6/en/identifiers.html for MySQL

### DIFF
--- a/lib/Doctrine/DBAL/Platforms/Keywords/MySQLKeywords.php
+++ b/lib/Doctrine/DBAL/Platforms/Keywords/MySQLKeywords.php
@@ -18,6 +18,15 @@ class MySQLKeywords extends KeywordList
     /**
      * {@inheritdoc}
      */
+    public function isKeyword($word)
+    {
+        return parent::isKeyword($word) ||  preg_match('/^\d+$/', $word) ||
+            preg_match('/[^\w$\x{0080}-\x{FFFF}]+/u', $word);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     protected function getKeywords()
     {
         return [


lib/Doctrine/DBAL/Platforms/Keywords/MySQLKeywords.php
```diff
@@ -15,6 +15,15 @@ public function getName()
         return 'MySQL';
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function isKeyword($word)
+    {
+        return parent::isKeyword($word) ||  preg_match('/^\d+$/', $word) ||
+            preg_match('/[^\w$\x{0080}-\x{FFFF}]+/u', $word);
```
morozov, line 24: The fact that an object name contains certain characters doesn't make it a keyword.

On MySQL, we can quote all identifiers. Whether or not it could be done without an API change, is a bigger question.

Please see https://github.com/doctrine/dbal/issues/4772 for more details.

morozov:
Closing as the change isn't acceptable in its current state.

PR #4796, opened 2021-09-15
Bump PHPStan and PHPUnit

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,8 +41,8 @@
     "require-dev": {
         "doctrine/coding-standard": "9.0.0",
         "jetbrains/phpstorm-stubs": "2021.1",
-        "phpstan/phpstan": "0.12.96",
-        "phpunit/phpunit": "^7.5.20|^8.5|9.5.5",
+        "phpstan/phpstan": "0.12.99",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -41,8 +41,8 @@
     "require-dev": {
         "doctrine/coding-standard": "9.0.0",
         "jetbrains/phpstorm-stubs": "2021.1",
-        "phpstan/phpstan": "0.12.96",
-        "phpunit/phpunit": "^7.5.20|^8.5|9.5.5",
+        "phpstan/phpstan": "0.12.99",
+        "phpunit/phpunit": "^7.5.20|^8.5|^9.5.5",
```
greg0ire: I think we really mean to pin tools to a specific version, IMO you should drop that caret.

derrabus: Oh okay. I did expect pinning to be necessary for PHPUnit, but why not. Pinned again.

greg0ire (approved):

PR #4797, opened 2021-09-17
Remove support for passing assets as names

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

#### Summary

The current Platform and Schema APIs allow passing object names as strings or as objects themselves (sub-classes of `Asset`). For instance, in order to drop a table, the caller can pass the table name or the `Table` object.

While it may look handly from the API consumer perspective, it violates the Interface Segregation Principle. In the example above, the `getDropTableSQL()` method has to support accepting an object although it only needs the name of the object.

Another problem is that this API is prone to being implemented inconsistently depending on the type of the actually passed argument. For instance:
1. MySQL will drop the PK using `ALTER TABLE tbl DROP PRIMARY KEY` only if the index is passed as an object since the information about whether the given index is the PK or not isn't encoded in the index name: https://github.com/doctrine/dbal/blob/27c26d2ccad78384534fd2db797b76c3c0e8d9af/src/Platforms/MySQLPlatform.php#L897-L901
2. The `Table` class will validate the columns of the foreign key against the referenced table only if the referenced table name is passed as an asset: https://github.com/doctrine/dbal/blob/f83121fa971abe481a5f3129821a28232f6bacd9/src/Schema/Table.php#L354-L360

Another practical problem is that it cripples the Events API: https://github.com/doctrine/dbal/blob/dc4ad2a0d7f72e9e55dd1f9e2a341c068fd2b89f/src/Event/SchemaDropTableEventArgs.php#L34-L37

The consumer of the `SchemaDropTableEventArgs` class has to support receiving the table name expressed as _either_ of the `string` and `Table`.

As part of prototyping https://github.com/doctrine/dbal/issues/4772, I'd like to convert all `string $name` arguments into `Name $name`. This is where the need to drop the support for `Asset|string` arguments arose.

**TODO**:
- [x] Deprecate the previous API (#4798).
   The deprecation should happen after the `Name`-based API is implemented or before 4.0.x is released (whatever happens first).

### DIFF
--- a/src/Event/SchemaDropTableEventArgs.php
+++ b/src/Event/SchemaDropTableEventArgs.php
@@ -5,36 +5,25 @@ declare(strict_types=1);
 namespace Doctrine\DBAL\Event;
 
 use Doctrine\DBAL\Platforms\AbstractPlatform;
-use Doctrine\DBAL\Schema\Table;
-use InvalidArgumentException;
 
 /**
  * Event Arguments used when the SQL query for dropping tables are generated inside {@link AbstractPlatform}.
  */
 class SchemaDropTableEventArgs extends SchemaEventArgs
 {
-    /** @var string|Table */
-    private $table;
+    private string $table;
 
     private AbstractPlatform $platform;
 
     private ?string $sql = null;
 
-    /**
-     * @param string|Table $table
-     *
-     * @throws InvalidArgumentException
-     */
-    public function __construct($table, AbstractPlatform $platform)
+    public function __construct(string $table, AbstractPlatform $platform)
     {
         $this->table    = $table;
         $this->platform = $platform;
     }
 
-    /**
-     * @return string|Table
-     */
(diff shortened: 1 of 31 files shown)


src/Event/SchemaDropTableEventArgs.php
```diff
@@ -5,34 +5,26 @@
...
-    public function __construct($table, AbstractPlatform $platform)
+    public function __construct(string $table, AbstractPlatform $platform)
     {
         $this->table    = $table;
         $this->platform = $platform;
     }
 
     /**
-     * @return string|Table
+     * @return string
```
morozov: To be rebased after https://github.com/doctrine/dbal/pull/4799 is merged.

greg0ire (approved):

mvorisek (commented):
Very good!

PR #4800, opened 2021-09-17
Add failing test for Schema::getTableNames() when using sqlite 

goetas:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | (unexpected change from dbal 2.x)
| Fixed issues | -

#### Summary

When using SQLite (or any other connection for which `\Doctrine\DBAL\Connection::getDatabase()` returns an empty string, the table names contains an invalid name.

This issue seems introduced by https://github.com/doctrine/dbal/commit/460f0a1fe2d41b1f03c929d62fc1783fe325755c

### DIFF
--- a/tests/Schema/SchemaTest.php
+++ b/tests/Schema/SchemaTest.php
@@ -2,6 +2,7 @@
 
 namespace Doctrine\DBAL\Tests\Schema;
 
+use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Schema\SchemaConfig;
 use Doctrine\DBAL\Schema\SchemaException;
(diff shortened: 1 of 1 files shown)


tests/Schema/SchemaTest.php
```diff
@@ -33,6 +34,17 @@ public function testAddTable(): void
         self::assertTrue($schema->hasTable($tableName));
     }
 
+    public function testGetTableNames(): void
+    {
+        $conn = DriverManager::getConnection(['driver' => 'pdo_sqlite', 'memory' => true]);
```
goetas: should I use here this instead to test it against the other database types?
```php
$conn = DriverManager::getConnection(TestUtil::getConnectionParams());
```

tests/Schema/SchemaTest.php
```diff
@@ -33,6 +35,17 @@ public function testAddTable(): void
...
 
+    public function testGetTableNames(): void
+    {
+        $conn = DriverManager::getConnection(TestUtil::getConnectionParams());
+
+        $schema = new Schema([], [], $conn->createSchemaManager()->createSchemaConfig());
+        $schema->createTable('foo');
+
+        $tableNames = $schema->getTableNames();
+        self::assertEquals(['foo'], $tableNames);
```
morozov, line 46: The same call to `getTableNames()` on `2.12.1` and `2.13.x` returns `public.foo` (which is pointless given that it's SQLite). Where does the expectation of the unqualified `foo` come from?

It looks like this API has been broken since the beginning.

goetas: I'm not sure if the `foo` expectations is correct, but to me feels that the current `.foo` is wrong (that is the current result).

it the return is `public.foo`, i'm 100% ok with it.

morozov: I believe the proper approach would be to iterate over the schema tables and get their names. The array keys are meant to be used for indexing the tables within the schema, not used as names.

I'm thinking of fixing and deprecating this method.

morozov: Fixed in https://github.com/doctrine/dbal/pull/4811.

PR #4801, opened 2021-09-19
Require $table in SQLServerPlatform::getDropIndexSQL()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Currently, SQLServer platform allows omitting table name when generating the `DROP INDEX` SQL statement and if the table name is omitted, generates an invalid statement:
```php
$connection->executeStatement(
    $connection->getDatabasePlatform()->getDropIndexSQL('foo')
);
// Must specify the table name and index name for the DROP INDEX statement
```

Additionally, the method implementation contains code unrelated to dropping indices not covered by tests. `DROP CONSTRAINT` statements should be generated by the corresponding method.

### DIFF
--- a/tests/Platforms/MySQLPlatformTest.php
+++ b/tests/Platforms/MySQLPlatformTest.php
@@ -5,6 +5,7 @@ namespace Doctrine\DBAL\Tests\Platforms;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Platforms\MySQLPlatform;
 use Doctrine\DBAL\TransactionIsolationLevel;
+use InvalidArgumentException;
 
 class MySQLPlatformTest extends AbstractMySQLPlatformTestCase
 {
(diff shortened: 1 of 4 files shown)


tests/Platforms/MySQLPlatformTest.php
```diff
@@ -20,4 +21,10 @@ public function testHasCorrectDefaultTransactionIsolationLevel(): void
             $this->platform->getDefaultTransactionIsolationLevel()
         );
     }
+
+    public function testDropIndexSQLRequiresTable(): void
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->platform->getDropIndexSQL('foo');
+    }
```
greg0ire, line 29: Picking MySQL looks arbitrary. Why do we not have a unit test for `AbstractPlatform`? Also, this looks like it should be in a separate commit.

morozov: > Picking MySQL looks arbitrary.

The logic of requiring the optional `$table` at runtime already exists in the MySQL platform but it wasn't covered with a test. I added it for symmetry.

> Why do we not have a unit test for `AbstractPlatform`?

Because the semantics of this method is inconsistent across the platforms: most of the platforms do not use `$table` since the index names there are scoped to the schema but MySQL and SQL Server use and require it because their index names are scoped the table to which the index applies.

A more consistent approach might be to make `$table` non-optional at the `AbstractPlatform` level but then it would require passing it while in most cases it wouldn't be used. So I decided not to change the API until/unless we come up with a better idea.

> Also, this looks like it should be in a separate commit.

Good idea. Done.

greg0ire (approved):

PR #4802, opened 2021-09-19
Platform API cleanup

morozov:
|      Q       |   A
|------------- | -----------
| BC Break     | no

The methods being reworked are defined in the `AbstractPlatform` sub-classes but do not belong to the Platform API:
1. `OraclePlatform::assertValidIdentifier()` is not used anywhere. The DBAL is not the source of truth about the validity of a given identifier.
2. `OraclePlatform::getCreateAutoincrementSql()` is only called from within the class and should be made protected.
3. `OraclePlatform::getDropAutoincrementSql()` is only called from within the schema manager. There's definitely some asymmetry with the above method. We can decide on how to improve this API later. For instance, get rid of the triggers in favor of identity columns.
4. The `SQLServerPlatform` methods in question are only called from within the class and should be made protected.
5. The SQLite user-defined functions don't need to be defined within the platform class.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -8,6 +8,10 @@ awareness about deprecated code.
 
 # Upgrade to 3.2
 
+## Deprecated `OraclePlatform::assertValidIdentifier()`
+
+The `OraclePlatform::assertValidIdentifier()` method has been deprecated.
+
 ## Deprecated features of `Table::getColumns()`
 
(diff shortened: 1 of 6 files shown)


derrabus (approved):

derrabus:
FIY: The ORM does not use any of those methods.

PR #4805, opened 2021-09-19
Deprecate AbstractPlatform::canEmulateSchemas()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no

The "schema emulation" feature as it's currently implemented in the DBAL doesn't belong to the DBAL. The emulation is effectively replacing the dot in the table names with a double underscore, assuming the dot is a fully qualified name separator (see https://github.com/doctrine/dbal/pull/4804), which isn't always true (see https://github.com/doctrine/dbal/pull/4708).

It is used by the ORM but since the ORM owns the table names, it can do the same replacement itself and emulate the schema on any platform. As such, this feature is really questionable. I wouldn't trust the results of such tests.

There isn't a single test in the DBAL that would cover this feature.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -8,6 +8,11 @@ awareness about deprecated code.
 
 # Upgrade to 3.2
 
+## Deprecated `AbstractPlatform::canEmulateSchemas()`.
+
+The `AbstractPlatform::canEmulateSchemas()` method and the schema emulation implemented in the SQLite platform
+have been deprecated.
+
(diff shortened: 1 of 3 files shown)


derrabus:
> It is used by the ORM

To me, this looks like the ORM uses this "feature" so Postgres functionality can be tested on SQLite. 🙈 

Yes, please remove it from DBAL.

derrabus (approved):

PR #4807, opened 2021-09-19
Generalize platform SQL

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

1. The `CREATE DATABASE` and `DROP DATABASE` statements are identical on all platforms that support databases.
2. The DBAL uses users on Oracle to emulate databases, so it should implement the "Create database" operation as `CREATE USER`. It already does the same with regards to dropping a database.
3. The IBM DB2 platform declares that it doesn't support databases: https://github.com/doctrine/dbal/blob/525eda78658b99b3118905b1058d272147c5d312/src/Platforms/DB2Platform.php#L440-L443  but its test attempt to cover the generation of the `CREATE DATABASE` and `DROP DATABASE` statements: https://github.com/doctrine/dbal/blob/11f0f94c474c7aa41b38b1defe1ab4c28ae0289d/tests/Platforms/DB2PlatformTest.php#L350-L351
4. The `CREATE SCHEMA` and `DROP SCHEMA` statements are identical on all platforms that support schemas.
5. The `CREATE VIEW` and `DROP VIEW` statements are identical on all platforms.
6. The is a common implementation of `DROP SEQUENCE` statements for all platforms that support sequences.

### DIFF
--- a/src/Platforms/AbstractPlatform.php
+++ b/src/Platforms/AbstractPlatform.php
@@ -1581,18 +1581,6 @@ abstract class AbstractPlatform
     }
 
     /**
-     * Returns the SQL snippet to drop an existing database.
-     *
-     * @param string $name The name of the database that should be dropped.
-     *
-     * @return string
-     */
-    public function getDropDatabaseSQL($name)
-    {
-        return 'DROP DATABASE ' . $name;
-    }
-
-    /**
      * Returns the SQL snippet to drop a schema.
      */
     public function getDropSchemaSQL(string $schemaName): string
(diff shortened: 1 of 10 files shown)


morozov:
I don't think it's worth covering the paths where the `NotSupported` exceptions are thrown with tests. This is a side effect of the `AbstractPlatform` being a god object and declaring the super-set of the APIs that its child classes implement.

The proper way to address this issue is to take the gigantic API apart.

greg0ire (approved):

PR #4811, opened 2021-09-22
Fix Schema::getTableNames()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

Closes #4800.

### DIFF
--- a/src/Schema/SqliteSchemaManager.php
+++ b/src/Schema/SqliteSchemaManager.php
@@ -568,4 +568,13 @@ SQL
 
         return $table;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getSchemaSearchPaths()
+    {
+        // SQLite does not support schemas or databases
+        return [];
+    }
 }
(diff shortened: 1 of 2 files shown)


goetas:
Two questions, 
1) shouldn't this branch contain also the failing test i've added? (eventually adjusted)
2) you were mentioning to deprecate `getTableNames` (that was looking a reasonable choice to me), did you change your mind about it? or it can be done in a separate PR?

morozov:
> shouldn't this branch contain also the failing test i've added? (eventually adjusted)

It should but the actual fully-qualified name is different on different platforms, so making the test pass requires effectively duplicating the code being tested. You can try it in your own branch to see the failures.

> you were mentioning to deprecate `getTableNames` (that was looking a reasonable choice to me), did you change your mind about it? or it can be done in a separate PR?

It's done in https://github.com/doctrine/dbal/pull/4812. The current PR goes into the next patch release as it's a bugfix, that one goes into the next minor as it's a deprecation.

greg0ire (approved):

PR #4814, opened 2021-09-26
Deprecate AbstractAsset::getFullQualifiedName()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | no

By definition, a fully qualified name is absolute and shouldn't depend on any input. The fully-qualified name of the assets is only used by `Schema` to index schema objects.

### DIFF
--- a/src/Schema/AbstractAsset.php
+++ b/src/Schema/AbstractAsset.php
@@ -3,6 +3,7 @@
 namespace Doctrine\DBAL\Schema;
 
 use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\Deprecations\Deprecation;
 
 use function array_map;
 use function crc32;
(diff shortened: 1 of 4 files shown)


src/Schema/Schema.php
```diff
@@ -200,6 +200,11 @@ private function getFullQualifiedAssetName($name)
         return strtolower($name);
     }
 
+    private function normalizeName(AbstractAsset $asset): string
+    {
+        return $asset->getFullQualifiedName($this->getName());
```
greg0ire, line 213: Why don't we switch to the recommended methods right now? That way we don't have to use `IfCalledFromOutside`, and we have an example of how to do the switch.

morozov: Because the deprecated method may be extended and contain custom logic. Not calling it is technically a BC break.

src/Schema/AbstractAsset.php
```diff
@@ -102,20 +103,29 @@ public function getShortestName($defaultNamespaceName)
     }
 
     /**
-     * The normalized name is full-qualified and lowerspaced. Lowerspacing is
+     * The normalized name is full-qualified and lower-cased. Lower-casing is
```
derrabus, line 106: I've backported this. 😇 #4827

derrabus (approved):

greg0ire (approved):

PR #4818, opened 2021-09-27
Cast decimals to string on SQLite

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | doctrine/orm#8963

#### Summary

On PHP 8.1, PDO's SQLite driver has "learnt" to convert integers and floats coming from the database to the corresponding native PHP types. Unfortunately, since SQLite is unable to differ between floats and decimals, we now get floats for decimals as well.

See the linked ORM issue for a more detailed discussion.

Since every driver has delivered DECIMALs as strings so far, `DecimalType::convertToPHPValue()` has not performed any type conversion yet. This PR proposes to change that by adding a `(string)` cast if we're on PHP 8.1 or higher and the value is a float. I believe that the caller of `DecimalType::convertToPHPValue()` would expect a string because otherwise they would've omitted the call and directly used the value as the driver returns it.

PHP should be able to optimize both checks, so the added condition should be fairly cheap. A float-to-string cast is unsafe on PHP 7 (because the result depends on the locale) which is another reason why we should have a version check here.

### DIFF
--- a/lib/Doctrine/DBAL/Types/DecimalType.php
+++ b/lib/Doctrine/DBAL/Types/DecimalType.php
@@ -4,6 +4,10 @@ namespace Doctrine\DBAL\Types;
 
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 
+use function is_float;
+
+use const PHP_VERSION_ID;
+
 /**
  * Type that maps an SQL DECIMAL to a PHP string.
  */
@@ -30,6 +34,10 @@ class DecimalType extends Type
      */
     public function convertToPHPValue($value, AbstractPlatform $platform)
     {
+        if (PHP_VERSION_ID >= 80100 && is_float($value)) {
+            return (string) $value;
+        }
+
         return $value;
     }
 }
--- a/tests/Doctrine/Tests/DBAL/Functional/Types/DecimalTest.php
+++ b/tests/Doctrine/Tests/DBAL/Functional/Types/DecimalTest.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Tests\DBAL\Functional\Types;
+
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\Types\Type;
+use Doctrine\DBAL\Types\Types;
+use Doctrine\Tests\DbalFunctionalTestCase;
+
+final class DecimalTest extends DbalFunctionalTestCase
+{
+    public function testInsertAndRetrieveDecimal(): void
+    {
+        $table = new Table('decimal_table');
+        $table->addColumn('val', Types::DECIMAL, ['precision' => 4, 'scale' => 2]);
+
+        $sm = $this->connection->getSchemaManager();
+        $sm->dropAndCreateTable($table);
+
+        $this->connection->insert(
+            'decimal_table',
+            ['val' => '13.37'],
+            ['val' => Types::DECIMAL]
+        );
+
+        $value = Type::getType(Types::DECIMAL)->convertToPHPValue(
+            $this->connection->fetchOne('SELECT val FROM decimal_table'),
+            $this->connection->getDatabasePlatform()
+        );
+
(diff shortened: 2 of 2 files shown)


tests/Doctrine/Tests/DBAL/Functional/Types/DecimalTest.php
```diff
@@ -0,0 +1,39 @@
...
+use Doctrine\DBAL\Types\Type;
+use Doctrine\DBAL\Types\Types;
+use Doctrine\Tests\DbalFunctionalTestCase;
+
+final class DecimalTest extends DbalFunctionalTestCase
+{
+    public function testInsertAndRetrieveDecimal(): void
+    {
+        $table = new Table('decimal_table');
+        $table->addColumn('id', Types::INTEGER);
```
greg0ire: Is there a point to having this column here? I don't think the `WHERE id = 1` part below is necessary, is it?

morozov:
I really think that instead of fixing the changes in the PHP internal on the DBAL/ORM side, we should embrace the reality and admit that the underlying database drivers are different and represent numbers using different PHP types.

Similar to not casting the values returned by any `fetch*()` or `lastInsertId()` methods, I think the DBAL should not be in charge of casting primitive PHP types. If the existing behavior of the PHP drivers is unacceptable from the end-user perspective, a concern should be raised with PHP instead.

> This change of behavior might break applications, for instance, if a string return type declaration is used for a getter method of that field.

I'd challenge this as a wrong expectation. Could you provide a concrete example of such an expectation?

derrabus:
> Similar to not casting the values returned by any `fetch*()` or `lastInsertId()` methods, I think the DBAL should not be in charge of casting primitive PHP types.

I agree that this should not happen for any low-level method like `fetchAssociative()` and friends. However, that is not the level on which this patch operates. What is the use case of the `Types` classes if not converting between database and PHP types?

> If the existing behavior of the PHP drivers is unacceptable from the end-user perspective, a concern should be raised with PHP instead.

I've done this already: https://bugs.php.net/81397

> > This change of behavior might break applications, for instance, if a string return type declaration is used for a getter method of that field.
> 
> I'd challenge this as a wrong expectation. Could you provide a concrete example of such an expectation?

Sure:

```php
/**
 * @Entity
 */
class MyEntity
{
    /**
     * @Column(type="integer")
     * @Id
     * @GeneratedValue
     */
    private $id;

    /**
     * @Column(type="decimal", precision=8, scale=2)
     */
    private $amount = '0.00';

    public function getId(): ?int
    {
        return $this->id;
    }

    public function getAmount(): string
    {
        return $this->amount;
    }

    public function setAmount(string $amount): void
    {
        $this->amount = $amount;
    }
}
```

tests/Doctrine/Tests/DBAL/Functional/Types/DecimalTest.php
derrabus: Fixed.

morozov (approved):

morozov:
Thanks for the details, @derrabus. I agree.

lib/Doctrine/DBAL/Types/DecimalType.php
```diff
@@ -30,6 +34,10 @@ public function getSQLDeclaration(array $column, AbstractPlatform $platform)
      */
     public function convertToPHPValue($value, AbstractPlatform $platform)
     {
+        if (PHP_VERSION_ID >= 80100 && is_float($value)) {
```
andrew-demb, line 37: Should it be commented with description - why is this `if` (with PHP version check) and casting added?

derrabus: Sure why not. Do you want to open a PR with a good comment? Maybe link to this PR.

andrew-demb: Yes, will do 
Added to today's schedule

PR #4819, opened 2021-09-27
PHPUnit 9.5.10

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A


SenseException (approved):
Do we need to have a fixed 9.5.x version?

derrabus:
I don't think we do. Shall I open a PR that drops the version pinning and we discuss this matter there?

PR #4823, opened 2021-09-28
Merge 2.13.x into 3.1.x

derrabus:


src/Driver/OCI8/Statement.php
```diff
@@ -101,7 +101,7 @@ public function bindParam($param, &$variable, $type = ParameterType::STRING, $le
 
             assert($lob !== false);
 
-            $lob->writetemporary($variable, OCI_TEMP_BLOB);
+            $lob->writeTemporary($variable, OCI_TEMP_BLOB);
```
derrabus, line 104: After the bump, PHPStan started to complain about this line. I've pushed a commit to make PHPStan happy again.

greg0ire (approved):

src/Driver/OCI8/Statement.php
morozov: I remember changing it a few times back and forth. This time, it should be related to https://github.com/php/php-src/pull/7405.

derrabus: Yes, probably. Thanks for the hint.

PR #4824, opened 2021-09-28
Remove pinned platform

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

I'd like to remove the `platform.php` setting from our composer.json file.

Without a lock file under version control, all that setting does is preventing the installation of newer packages on PHP 7.4 and higher (e.g. psr/log 2). I would argue that testing with different sets of dependencies for different PHP versions is somewhat desirable.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,9 +54,6 @@
     },
     "bin": ["bin/doctrine-dbal"],
     "config": {
-        "platform": {
-            "php": "7.3.0"
-        },
         "sort-packages": true
     },
     "autoload": {


greg0ire (approved):
It should indeed have been dropped when removing `composer.lock` from version control. My bad.

composer.json
```diff
@@ -54,9 +54,6 @@
     },
     "bin": ["bin/doctrine-dbal"],
     "config": {
-        "platform": {
-            "php": "7.3.0"
```
morozov, line 58: It was introduced to prevent locking the versions that are incompatible with the lowest supported PHP version. While the lock file is gone, it's less of an issue but without pinning the platform, it becomes possible to upgrade a dependency to such a version. Do we rely on CI for that?

derrabus: > without pinning the platform, it becomes possible to upgrade a dependency to such a version.

Yes. And I would argue that this is not a problem and might even be desirable.

> Do we rely on CI for that?

Well currently, the CI installs dependencies compatible with PHP 7.3.0, even on the PHP 8 jobs. For instance, `psr/log` is installed as version 1.1:

https://github.com/doctrine/dbal/runs/3733999925#step:4:59

If you look at the same job in this PR, you'll see that version 2.0 is installed instead.

derrabus: And if I may add: We do have a PHP 7.3 job that makes sure that the package version constraints we set in composer.json can be fulfilled on PHP 7.3. So we don't need the platform pinning for that matter either.

morozov: I agree. While this change makes the maintenance a bit more challenging, it's not the reason to disallow newer package versions.

morozov (approved):

PR #4833, opened 2021-09-30
Add explanation why we need to cast float to string in decimal type

andrew-demb:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Added explanation to code, why we need to cast a float to string only after PHP 8.1

refs: https://github.com/doctrine/dbal/pull/4818#discussion_r717958844
/cc @derrabus


derrabus (dismissed):

greg0ire (approved):

greg0ire:
Thanks @andrew-demb !

Issue #4816, opened 2021-09-27
Renaming columns with upper-cased letters doesn't work on case-sensitive platforms


morozov:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.1.2

Steps to reproduce: #4815 on SQL Server with Latin1_General_100_CS_AS ([build](https://github.com/doctrine/dbal/pull/4815/checks?check_run_id=3715083229)).

PR #4838, opened 2021-10-01
fixed using lower case column names

### LINKED ISSUE #4832: Columns are not found with 3.1.2
### Bug Report

Columns are not found. They are named in camelCase. 

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | 3.1.2

#### Summary

I am using doctrine/dbal in a lumen project. I recently upgraded dbal to 3.1.2 and now my tests with sqlite and mysql fail with the message that columns cannot be found when renamed or their type is changed. 

With version 2.13.3 and 3.1.1 it works. 

#### Current behaviour

-

#### How to reproduce

-

#### Expected behaviour

The columns should be found and renamed or their type should get altered.

eschricker:
Columns are not found when camelCase named columns are changed or renamed when using sqlite as database.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #4832

#### Summary

Recovered the behavior of version 3.1.1 to use lower case column names.

### DIFF
--- a/src/Platforms/SqlitePlatform.php
+++ b/src/Platforms/SqlitePlatform.php
@@ -920,7 +920,8 @@ class SqlitePlatform extends AbstractPlatform
                 continue;
             }
 
-            $columns = $this->replaceColumn($diff->name, $columns, $oldColumnName, $column);
+            $oldColumnName = strtolower($oldColumnName);
+            $columns       = $this->replaceColumn($diff->name, $columns, $oldColumnName, $column);
 
             if (! isset($newColumnNames[$oldColumnName])) {
                 continue;
@@ -934,7 +935,8 @@ class SqlitePlatform extends AbstractPlatform
                 continue;
             }
 
-            $columns = $this->replaceColumn($diff->name, $columns, $oldColumnName, $columnDiff->column);
+            $oldColumnName = strtolower($oldColumnName);
+            $columns       = $this->replaceColumn($diff->name, $columns, $oldColumnName, $columnDiff->column);
 
             if (! isset($newColumnNames[$oldColumnName])) {
                 continue;
@@ -1016,7 +1018,7 @@ class SqlitePlatform extends AbstractPlatform
     private function replaceColumn($tableName, array $columns, $columnName, Column $column): array
     {
         $keys  = array_keys($columns);
-        $index = array_search(strtolower($columnName), $keys, true);
+        $index = array_search($columnName, $keys, true);
 
         if ($index === false) {
             throw SchemaException::columnDoesNotExist($columnName, $tableName);
--- a/tests/Functional/Platform/RenameColumnTest.php
+++ b/tests/Functional/Platform/RenameColumnTest.php
@@ -13,9 +13,9 @@ class RenameColumnTest extends FunctionalTestCase
     /**
      * @dataProvider columnNameProvider
      */
-    public function testColumnPositionRetainedAfterRenaming(string $columnName): void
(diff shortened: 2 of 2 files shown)


derrabus:
Can you provide a test that fails without your changes?

eschricker:
@derrabus I have added a test, that fails without my changes

derrabus:
Thank you. Can you have a look at the failing CS check? You probably just need to run `./vendor/bin/phpcbf` to fix this.

derrabus (dismissed):

tests/Platforms/SqlitePlatformTest.php
```diff
@@ -360,6 +360,30 @@ public function testRenameNonExistingColumn(): void
         $this->platform->getAlterTableSQL($tableDiff);
     }
 
+    public function testRenameCamelCaseColumn(): void
```
morozov: @eschricker thank you for the patch. Instead of testing SQL, please test the behavior at the integration level. See `RenameColumnTest`.

src/Platforms/SqlitePlatform.php
```diff
@@ -920,7 +920,8 @@ public function getAlterTableSQL(TableDiff $diff)
                 continue;
             }
 
-            $columns = $this->replaceColumn($diff->name, $columns, $oldColumnName, $column);
+            $oldColumnName = strtolower($oldColumnName);
```
morozov, line 923: Note, the `replaceColumn()` method also converts the names to the lower case because this is how `$columns` are indexed. With this patch, it seems no longer necessary.

eschricker: I have removed the conversion to lower case in the `replaceColumn()` method

tests/Platforms/SqlitePlatformTest.php
eschricker: I have updated the RenameColumnTest to include in the dataprovider the desired name for renaming in addition to the column name. But due to the comment that it's not working on case-sensitive platforms, the extended tests should fail on them.

morozov: Please remove this test. Unlike the integration one, it doesn't prove that the implementation is correct, it only proves that it won't change accidentally.

morozov (commented):
Looks good. Please remove the new test and squash.

eschricker:
@morozov can you describe to me, how I can squash my commits?

derrabus:
The test failure on SQL server looks related.

greg0ire:
> can you describe to me, how I can squash my commits?

1. `git rebase -i origin/3.1.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

eschricker:
>     2\. A window will show up with many lines, replace `pick` with `fixup` on every line but the first one

Thanks @greg0ire it worked :)

morozov:
The failing test should be skipped on SQL Server due to https://github.com/doctrine/dbal/issues/4816. There's already the if-statement in the test.

eschricker:
@morozov is there anything else I should do or is it ok so far?

morozov:
@eschricker please fix the coding standard violation and disable the test on SQL Server.

morozov (approved):

morozov:
Thanks, @eschricker!

PR #4839, opened 2021-10-02
Deprecate the Constraint interface

morozov:
|      Q       |   A
|------------- | -----------
| Type         | deprecation, improvement
| BC Break     | no

#### Summary

The interface attempts to unify the API for different constraint classes but does it poorly.

1. Indeed, both indices and unique constraints are defined with a list of columns but the foreign key constraint is defined with _two sets_ of columns: referring (a.k.a. local) and referred (a.k.a. foreign). The `getColumns()` method on a `ForeignKeyConstraint` class doesn't make any sense (implies the only set of columns) but has to be implemented: https://github.com/doctrine/dbal/blob/916c90f43f3acb2bf27882688115395a5ca87915/src/Schema/ForeignKeyConstraint.php#L199-L204
2. There are other constraints in SQL like `CHECK`, `DEFAULT`, `NOT NULL`, etc. that do not use a list of columns in their definition.
3. The `AbstractPlatform::getCreateConstraintSQL()` method accepts a `Constraint` and by its signature is supposed to be capable of creating any constraint. But it only implements handling some specific sub-classes: https://github.com/doctrine/dbal/blob/f6f79552d33508e6d9df2938a84d333823d5feca/src/Platforms/AbstractPlatform.php#L1992-L2009 If a new class implementing the `Constraint` method is added, this method will have to be reworked which violates the Open-closed principle.

#### Additionally
The current implementation of `AbstractPlatform::getDropConstraintSQL()` doesn't work on MySQL older than 8.0.19 and MariaDB (see `SchemaManagerFunctionalTestCase::testDropAndCreateUniqueConstraint()`).

#### Proposed changes
1. Deprecate the `Constraint` interface and the corresponding methods.
2. Mark `AbstractPlatform::getDropConstraintSQL()` as `@internal` and convert to `protected` in the next major release. This method generates valid SQL and could be used by the methods that drop specific constraints.
3. Introduce drop/create methods for unique constraints in `AbstractPlatform` and `AbstractSchemaManager`.
4. Do not introduce `AbstractSchemaManager::dropAndCreateUniqueConstraint()`. It's hard to think of the usefulness of such "drop-and-create" methods outside of integration testing. These methods are primarily untested, do not allow proper static analysis, and suppress all underlying exceptions thrown by the "drop" part. We can deprecate them separately.
5. The proper implementation of `OraclePlatform::getCreatePrimaryKeySQL()` allows using `createIndex()` to create a primary key in the test instead of `createConstraint()`. This is still not the best API but at least it's consistent with the rest of the platforms.

### DIFF
--- a/src/Schema/ForeignKeyConstraint.php
+++ b/src/Schema/ForeignKeyConstraint.php
@@ -196,6 +196,8 @@ class ForeignKeyConstraint extends AbstractAsset implements Constraint
     /**
      * {@inheritdoc}
      *
+     * @deprecated Use {@link getLocalColumns()} instead.
+     *
      * @see getLocalColumns
      */
     public function getColumns()
@@ -210,6 +212,8 @@ class ForeignKeyConstraint extends AbstractAsset implements Constraint
      * But only if they were defined with one or the referencing table column name
      * is a keyword reserved by the platform.
      * Otherwise the plain unquoted value as inserted is returned.
+     *
+     * @deprecated Use {@link getQuotedLocalColumns()} instead.
      *
      * @see getQuotedLocalColumns
      *
--- a/tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
+++ b/tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
@@ -7,6 +7,7 @@ use Doctrine\DBAL\Events;
 use Doctrine\DBAL\Exception;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Platforms\OraclePlatform;
+use Doctrine\DBAL\Platforms\SqlitePlatform;
 use Doctrine\DBAL\Platforms\SQLServerPlatform;
 use Doctrine\DBAL\Schema\AbstractAsset;
 use Doctrine\DBAL\Schema\AbstractSchemaManager;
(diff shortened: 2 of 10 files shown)


tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -424,6 +426,34 @@ public function testDropAndCreateIndex(): void
         self::assertFalse($tableIndexes['test']->isPrimary());
     }
 
+    public function testDropAndCreateUniqueConstraint(): void
+    {
+        if ($this->connection->getDatabasePlatform() instanceof SqlitePlatform) {
+            self::markTestSkipped('SQLite does not support adding constrains to a table');
```
greg0ire:
```suggestion
            self::markTestSkipped('SQLite does not support adding constraints to a table');
```

src/Schema/ForeignKeyConstraint.php
```diff
@@ -196,6 +196,8 @@ public function getUnquotedForeignColumns()
     /**
      * {@inheritdoc}
      *
+     * @deprecated Use {@link getLocalColumns()} instead.
```
derrabus, line 199: Should we adjust the calls in those methods accordingly?

* `Doctrine\DBAL\Schema\SchemaException::namedForeignKeyRequired()`
* `Doctrine\DBAL\Schema\Table::_addForeignKeyConstraint()`

morozov: No, because this method may be overridden in a custom class extending `ForeignKeyConstraint`. Us no longer calling it may result in a BC break.

derrabus (approved):

PR #4843, opened 2021-10-02
Mark Statement::execute() private, remove QueryBuilder::execute()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

See https://github.com/doctrine/dbal/pull/4580.

### DIFF
--- a/psalm.xml.dist
+++ b/psalm.xml.dist
@@ -60,10 +60,6 @@
                 -->
                 <referencedMethod name="Doctrine\DBAL\Query\QueryBuilder::execute"/>
                 <referencedMethod name="Doctrine\DBAL\Statement::execute"/>
-                <!--
-                    TODO: remove in 4.0.0
-                -->
-                <referencedMethod name="Doctrine\DBAL\Schema\Schema::getMigrateToSql"/>
             </errorLevel>
         </DeprecatedMethod>
         <DocblockTypeContradiction>
--- a/psalm.xml.dist
+++ b/psalm.xml.dist
@@ -55,11 +55,6 @@
                     See https://github.com/doctrine/dbal/pull/4317
                 -->
                 <file name="tests/Functional/LegacyAPITest.php"/>
-                <!--
-                    These suppressions should be removed in 4.0.0
-                -->
-                <referencedMethod name="Doctrine\DBAL\Query\QueryBuilder::execute"/>
-                <referencedMethod name="Doctrine\DBAL\Statement::execute"/>
             </errorLevel>
         </DeprecatedMethod>
         <DocblockTypeContradiction>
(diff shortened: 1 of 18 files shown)


tests/Query/QueryBuilderTest.php
```diff
@@ -849,23 +849,6 @@ public function testJoinWithNonUniqueAliasThrowsException(): void
...
-            ->expects(self::any())
-            ->method('executeQuery')
-            ->willReturn($this->createMock(Result::class));
-
-        $result = $qb
-            ->select('id')
-            ->from('foo')
-            ->execute();
-
-        self::assertInstanceOf(Result::class, $result);
```
morozov, line 866: This expectation is now enforced at the API level.

psalm.xml.dist
```diff
@@ -55,15 +55,6 @@
...
                 <file name="tests/Functional/LegacyAPITest.php"/>
-                <!--
-                    These suppressions should be removed in 4.0.0
-                -->
-                <referencedMethod name="Doctrine\DBAL\Query\QueryBuilder::execute"/>
-                <referencedMethod name="Doctrine\DBAL\Statement::execute"/>
-                <!--
-                    TODO: remove in 4.0.0
-                -->
-                <referencedMethod name="Doctrine\DBAL\Schema\Schema::getMigrateToSql"/>
```
morozov, line 66: This one is a leftover from a previous deprecated API removal.

src/Statement.php
```diff
@@ -154,19 +153,11 @@ public function bindParam($param, &$variable, int $type = ParameterType::STRING,
     /**
      * {@inheritDoc}
      *
-     * @deprecated Statement::execute() is deprecated, use Statement::executeQuery() or executeStatement() instead
-     *
      * @throws Exception
      */
-    public function execute(?array $params = null): Result
+    private function execute(array $params = []): Result
```
derrabus:
```suggestion
    private function execute(array $params): Result
```

This private method is never called without parameters. Let's keep things simple. 🙂

src/Statement.php
```diff
@@ -154,19 +153,11 @@ public function bindParam($param, &$variable, int $type = ParameterType::STRING,
...
-        Deprecation::triggerIfCalledFromOutside(
-            'doctrine/dbal',
-            'https://github.com/doctrine/dbal/pull/4580',
-            'Statement::execute() is deprecated, use Statement::executeQuery() or Statement::executeStatement() instead'
-        );
-
-        if ($params !== null) {
+        if ($params !== []) {
             $this->params = $params;
         }
```
derrabus, line 162:
```suggestion
        $this->params = $params;
```

Assigning an empty array should be safe here, as it has always been.

src/Statement.php
derrabus: Oh, you've changed that already. 🚀

src/Statement.php
morozov: I believe the idea is that if you're calling `execute()` without parameters, it doesn't override the parameters bound via `bindParam()` or `bindValue()`. Not sure how sane this API is but I believe this is inherited from PDO.

derrabus: I see, thanks.

derrabus (approved):

PR #4844, opened 2021-10-02
Get view definition for SQL Server

kitloong:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

Currently, `SQLServer2012Platform::getListViewsSQL` does not export view definition.
This PR updated the SQL so that `Doctrine\DBAL\Schema\View::getSql()` will return view definition instead of empty string.

### DIFF
--- a/tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
+++ b/tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
@@ -106,14 +106,24 @@ abstract class SchemaManagerFunctionalTestCase extends FunctionalTestCase
      */
     private function hasElementWithName(array $items, string $name): bool
     {
-        $filteredList = array_filter(
+        $filteredList = $this->filterElementsByName($items, $name);
+
+        return count($filteredList) === 1;
+    }
+
+    /**
+     * @param AbstractAsset[] $items
+     *
+     * @return AbstractAsset[]
+     */
+    private function filterElementsByName(array $items, string $name): array
+    {
+        return array_filter(
             $items,
             static function (AbstractAsset $item) use ($name): bool {
                 return $item->getShortestName($item->getNamespaceName()) === $name;
             }
         );
-
-        return count($filteredList) === 1;
     }
 
     public function testListSequences(): void
@@ -688,7 +698,13 @@ abstract class SchemaManagerFunctionalTestCase extends FunctionalTestCase
 
         $this->schemaManager->dropAndCreateView($view);
 
-        self::assertTrue($this->hasElementWithName($this->schemaManager->listViews(), $name));
+        $views = $this->schemaManager->listViews();
+
+        $filtered = array_values($this->filterElementsByName($views, $name));
+        self::assertCount(1, $filtered);
+
+        $viewKey = strtolower($filtered[0]->getName());
+        self::assertStringContainsString('view_test_table', $views[$viewKey]->getSql());
     }
 
     public function testAutoincrementDetection(): void
--- a/src/Platforms/SQLServerPlatform.php
+++ b/src/Platforms/SQLServerPlatform.php
@@ -993,7 +993,9 @@ class SQLServerPlatform extends AbstractPlatform
      */
     public function getListViewsSQL($database)
     {
-        return "SELECT name FROM sysobjects WHERE type = 'V' ORDER BY name";
+        return "SELECT name, definition FROM sysobjects
+                    INNER JOIN sys.sql_modules ON sysobjects.id = sys.sql_modules.object_id
+                WHERE type = 'V' ORDER BY name";
     }
 
     /**
--- a/src/Schema/SQLServerSchemaManager.php
+++ b/src/Schema/SQLServerSchemaManager.php
@@ -248,7 +248,7 @@ SQL
     protected function _getPortableViewDefinition($view)
     {
         // @todo
-        return new View($view['name'], '');
+        return new View($view['name'], $view['definition']);
     }
 
     /**


src/Platforms/SQLServer2012Platform.php
```diff
@@ -1049,7 +1049,9 @@ public function getCreateViewSQL($name, $sql)
      */
     public function getListViewsSQL($database)
     {
-        return "SELECT name FROM sysobjects WHERE type = 'V' ORDER BY name";
+        return "SELECT name, definition FROM sysobjects
+                    INNER JOIN sys.sql_modules ON (sysobjects.id = sys.sql_modules.object_id)
```
morozov: Looks like parentheses are redundant here.

morozov (commented):
Thank you for the patch. Please add an integration test or reuse an existing one.

morozov:
Given that the DBAL doesn't use views itself, I'd rather classify this as an improvement. In this case, the patch should target `3.2.x`.

kitloong:
Hi @morozov , thank you for review.

I will add tests for this.

I am sorry, should I change the current PR base to `3.2.x`? Or checkout a new branch from `3.2.x` to commit the changes?

morozov:
> […] should I change the current PR base to `3.2.x`?

Yes, this was the first step. Now you need to rebase your changes on `3.2.x` and force-push the changes:
```shell
$ git checkout fix-sqlserver-view
$ git rebase HEAD~ --onto 3.2.x # this assumes you want to rebase one commit
$ git push -f # adjust according to your remote naming
```

tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -692,7 +708,8 @@ public function testCreateAndListViews(): void
 
         $this->schemaManager->dropAndCreateView($view);
 
-        self::assertTrue($this->hasElementWithName($this->schemaManager->listViews(), $name));
+        self::assertTrue($this->hasElementWithName($views = $this->schemaManager->listViews(), $name));
+        self::assertTrue($this->viewSqlContainsTableName($views, 'view_test_table'));
```
kitloong: Reusing `testCreateAndListViews`.

Added new assertion to check if `View::getSql()` always contains `view_test_table`, this also cover `SQLServerPlatform`.

kitloong:
Hello @morozov 

Following items completed.

- [x] Code update based on last reviewed.
- [x] Rebase 3.2.x.
- [x] Update integration test.

Please review.

tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
morozov: This assertion will make sure that at exactly one of the view definitions contains `view_test_table`. We need to make sure that specifically, the definition with the given name contains the SQL in question.

I assume that some platforms will expand `SELECT *` to an explicit set of columns. Would it make sense to change the query to something like `SELECT id ...` and assert that the entire query is available as part of the definition?

kitloong: > some platforms will expand SELECT * to an explicit set of columns

Thank you for this information. Has changed to `SELECT id ...`  in the new commit.

The tricky part is the different platforms return different format definitions.
For instance:
```
# MySQL
CREATE VIEW doctrine_test_view AS (select `db`.`view_test_table`.`id` AS `id` from `db`.`view_test_table`)

# PgSQL
CREATE VIEW doctrine_test_view AS SELECT view_test_table.id FROM view_test_table

# SQLServer
SELECT id FROM view_test_table
```

The new test a000089 assert `View::getSql()` contains SQL string with regex pattern`/\b(select|id|from|view_test_table)\b/i`
Should be better than the previous test.

morozov: Thanks for the details. In this case, the regex-based assertion looks too fragile, and it's better to just assert that the definition has the table name in it.

tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -120,6 +124,18 @@ static function (AbstractAsset $item) use ($name): bool {
         return count($filteredList) === 1;
     }
 
+    private function viewContainsSql(View $view, string $sql): bool
```
morozov: This logic is not reused anywhere in the test. Please revert to the original assertion and inline into the test method.

tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -686,13 +702,14 @@ public function testCreateAndListViews(): void
...
         $name = 'doctrine_test_view';
-        $sql  = 'SELECT * FROM view_test_table';
+        $sql  = 'SELECT id FROM view_test_table';
 
         $view = new View($name, $sql);
 
         $this->schemaManager->dropAndCreateView($view);
 
-        self::assertTrue($this->hasElementWithName($this->schemaManager->listViews(), $name));
+        self::assertTrue($this->hasElementWithName($views = $this->schemaManager->listViews(), $name));
```
morozov: Assignment in the method call is unreadable. Please extract it into a separate expression.

tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
kitloong: Thank you, has been updated to assert the definition contains the table name.

Extra note, `$views` key is view name, and again different platform has own format.
```
# MySQL
doctrine_test_view

# PgSQL
public.doctrine_test_view

# IBM DB2
DOCTRINE_TEST_VIEW
```

Has to use `View::getName()` and `strtolower` on key to pass test

tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -110,14 +110,24 @@ public function testDropAndCreateSequence(): void
...
+
+        return count($filteredList) === 1;
+    }
+
+    /**
+     * @param AbstractAsset[] $items
+     *
+     * @return AbstractAsset[]
+     */
+    private function filterElementByName(array $items, string $name): array
```
morozov: Should it be `filterElementsByName()`?

tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -686,13 +696,20 @@ public function testCreateAndListViews(): void
         $this->createTestTable('view_test_table');
 
         $name = 'doctrine_test_view';
-        $sql  = 'SELECT * FROM view_test_table';
+        $sql  = 'SELECT id FROM view_test_table';
```
morozov: This change is no longer necessary.

tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -686,13 +696,20 @@ public function testCreateAndListViews(): void
...
 
         $view = new View($name, $sql);
 
         $this->schemaManager->dropAndCreateView($view);
 
-        self::assertTrue($this->hasElementWithName($this->schemaManager->listViews(), $name));
+        $views = $this->schemaManager->listViews();
+
+        self::assertTrue($this->hasElementWithName($views, $name));
```
morozov: The two following lines perform the same assertion. Is this one needed?

morozov:
@kitloong looks good. Please squash again.

tests/Functional/Schema/SchemaManagerFunctionalTestCase.php
```diff
@@ -692,7 +702,12 @@ public function testCreateAndListViews(): void
 
         $this->schemaManager->dropAndCreateView($view);
 
-        self::assertTrue($this->hasElementWithName($this->schemaManager->listViews(), $name));
+        $views = $this->schemaManager->listViews();
+
+        $filtered = array_values($this->filterElementsByName($views, $name));
+        self::assertSame(1, count($filtered));
```
derrabus:
```suggestion
        self::assertCount(1, $filtered);
```

morozov (approved):

morozov:
Thanks, @kitloong!

PR #4846, opened 2021-10-04
Fix incorrect handling of transactions using deferred constraints

simPod:
_Closes #3424 as it's very long and outdated so I decided to start again here_

|      Q       |   A
|------------- | -----------
| Type         | bug/feature
| BC Break     | yes
| Fixed issues | #3423, https://github.com/doctrine/orm/issues/7545

#### Summary

Rollback is being called when outside of a transaction when using e.g. deferred constraints or when using ORM and database throws error like https://www.cockroachlabs.com/docs/v21.1/transaction-retry-error-reference.html#retry_write_too_old which is then effectively hidden after PDO's `There's no active transaction`.

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -919,14 +919,15 @@ class Connection implements ServerVersionProvider
         $this->beginTransaction();
         try {
             $res = $func($this);
-            $this->commit();
-
-            return $res;
         } catch (Throwable $e) {
             $this->rollBack();
 
             throw $e;
         }
+
+        $this->commit();
+
+        return $res;
     }
 
     /**
@@ -1005,17 +1006,26 @@ class Connection implements ServerVersionProvider
 
         $connection = $this->connect();
 
-        if ($this->transactionNestingLevel === 1) {
-            try {
-                $connection->commit();
-            } catch (Driver\Exception $e) {
-                throw $this->convertException($e);
+        try {
+            if ($this->transactionNestingLevel === 1) {
+                try {
+                    $connection->commit();
+                } catch (Driver\Exception $e) {
+                    throw $this->convertException($e);
+                }
+            } else {
+                $this->releaseSavepoint($this->_getNestedTransactionSavePointName());
             }
-        } else {
-            $this->releaseSavepoint($this->_getNestedTransactionSavePointName());
+        } finally {
+            $this->updateTransactionStateAfterCommit();
         }
+    }
 
-        --$this->transactionNestingLevel;
+    private function updateTransactionStateAfterCommit(): void
+    {
+        if ($this->transactionNestingLevel !== 0) {
+            --$this->transactionNestingLevel;
+        }
 
         if ($this->autoCommit !== false || $this->transactionNestingLevel !== 0) {
             return;
--- a/src/Driver/OCI8/Exception/Error.php
+++ b/src/Driver/OCI8/Exception/Error.php
@@ -7,7 +7,9 @@ namespace Doctrine\DBAL\Driver\OCI8\Exception;
 use Doctrine\DBAL\Driver\AbstractException;
 
 use function assert;
+use function explode;
 use function oci_error;
+use function str_replace;
 
 /**
  * @internal
@@ -16,12 +18,26 @@ use function oci_error;
  */
 final class Error extends AbstractException
 {
+    private const CODE_TRANSACTION_ROLLED_BACK = 2091;
+
     /** @param resource $resource */
     public static function new($resource): self
     {
         $error = oci_error($resource);
         assert($error !== false);
 
-        return new self($error['message'], null, $error['code']);
+        $code    = $error['code'];
+        $message = $error['message'];
+        if ($code === self::CODE_TRANSACTION_ROLLED_BACK) {
+            // There's no way this can be unit-tested as it's impossible to mock $resource
+            //ORA-02091: transaction rolled back
+            //ORA-00001: unique constraint (DOCTRINE.GH3423_UNIQUE) violated
+            [$firstMessage, $secondMessage] = explode("\n", $message, 2);
+
+            [$code, $message] = explode(': ', $secondMessage, 2);
+            $code             = (int) str_replace('ORA-', '', $code);
+        }
+
+        return new self($message, null, $code);
     }
 }
--- a/tests/Driver/PDO/ExceptionTest.php
+++ b/tests/Driver/PDO/ExceptionTest.php
@@ -56,4 +56,25 @@ class ExceptionTest extends TestCase
     {
         self::assertSame($this->wrappedException, $this->exception->getPrevious());
     }
+
+    public function testExposesUnderlyingErrorOnOracle(): void
+    {
+        $pdoException = new PDOException(<<<'TEXT'
+OCITransCommit: ORA-02091: transaction rolled back
+ORA-00001: unique constraint (DOCTRINE.C1_UNIQUE) violated
+ (/private/tmp/php-20211003-35441-1sggrmq/php-8.0.11/ext/pdo_oci/oci_driver.c:410)
+TEXT);
+
+        $pdoException->errorInfo = [self::SQLSTATE, 2091,
+
+        ];
+
+        $exception = Exception::new($pdoException);
+
+        self::assertSame(1, $exception->getCode());
+        self::assertStringContainsString(
+            'unique constraint (DOCTRINE.C1_UNIQUE) violated',
+            $exception->getMessage(),
+        );
+    }
 }
--- a/src/Driver/OCI8/Connection.php
+++ b/src/Driver/OCI8/Connection.php
@@ -95,7 +95,7 @@ final class Connection implements ConnectionInterface
 
     public function commit(): void
     {
-        if (! oci_commit($this->connection)) {
+        if (! @oci_commit($this->connection)) {
             throw Error::new($this->connection);
         }
 
--- a/src/Driver/PDO/Exception.php
+++ b/src/Driver/PDO/Exception.php
@@ -7,6 +7,9 @@ namespace Doctrine\DBAL\Driver\PDO;
 use Doctrine\DBAL\Driver\AbstractException;
 use PDOException;
 
+use function explode;
+use function str_replace;
+
 /**
  * @internal
  *
@@ -14,8 +17,12 @@ use PDOException;
  */
 final class Exception extends AbstractException
 {
+    private const CODE_TRANSACTION_ROLLED_BACK = 2091;
+
     public static function new(PDOException $exception): self
     {
+        $message = $exception->getMessage();
+
         if ($exception->errorInfo !== null) {
             [$sqlState, $code] = $exception->errorInfo;
 
@@ -25,6 +32,15 @@ final class Exception extends AbstractException
             $sqlState = null;
         }
 
-        return new self($exception->getMessage(), $sqlState, $code, $exception);
+        if ($code === self::CODE_TRANSACTION_ROLLED_BACK) {
+            //ORA-02091: transaction rolled back
+            //ORA-00001: unique constraint (DOCTRINE.GH3423_UNIQUE) violated
+            [$firstMessage, $secondMessage] = explode("\n", $message, 2);
+
+            [$code, $message] = explode(': ', $secondMessage, 2);
+            $code             = (int) str_replace('ORA-', '', $code);
+        }
+
+        return new self($message, $sqlState, $code, $exception);
     }
 }
--- a/tests/Functional/UniqueConstraintViolationsTest.php
+++ b/tests/Functional/UniqueConstraintViolationsTest.php
@@ -1,0 +1,249 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\DBAL\Tests\Functional;
+
+use Doctrine\DBAL\Connection;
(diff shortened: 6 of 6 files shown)


morozov:
@simPod thank you for the patch. I remember looking into this issue a couple of years ago. IIRC, it wasn't something trivial, so it may take some time before I post the review. But rest assured, it won't be ignored.

simPod:
Thank you, note taken. Take your time.

src/Connection.php
```diff
@@ -1336,17 +1343,22 @@ public function commit()
             }
         }
 
+        $this->updateTransactionStateAfterCommit();
+
+        return $result;
+    }
+
+    private function updateTransactionStateAfterCommit(): void
```
morozov, line 1024: Instead of declaring a method and calling it explicitly at different points of the `commit()` flow, should it be a `finally` block? I believe this is the intent.

src/Connection.php
```diff
@@ -1320,7 +1321,13 @@ public function commit()
                 $logger->startQuery('"COMMIT"');
             }
 
-            $result = $connection->commit();
+            try {
```
morozov: Right now, this block seems to be dealing with the APIs at two different levels of abstractions which warrants some cleanup in order to keep it comprehensible:
1. Move the body under the current `if`-statement in a private method like `doCommit()`.
2. Move the logging under the following `elseif`-statement to `releaseSavepoint()`. Otherwise, logging isn't done if that method is called from the outside. Should it even be public if it's used as part of `commit()` implementation?

Then the `finally` block suggested in the previous comment will fit nicely here.

src/Driver/PDO/Connection.php
```diff
@@ -140,7 +140,11 @@ public function beginTransaction()
      */
     public function commit()
     {
-        return $this->connection->commit();
+        try {
```
morozov: Should the same be applied to the `rollBack()`?

src/Driver/OCI8/Exception/Error.php
```diff
@@ -24,6 +28,18 @@ public static function new($resource): self
...
+        $code    = $error['code'];
+        $message = $error['message'];
+        if ($code === self::CODE_TRANSACTION_ROLLED_BACK) {
+            // There's no way this can be unit-tested as it's impossible to mock $resource
+            //ORA-02091: transaction rolled back
+            //ORA-00001: unique constraint (DOCTRINE.GH3423_UNIQUE) violated
+            [$firstMessage, $secondMessage] = explode("\n", $message, 2);
+
+            [$code, $message] = explode(': ', $secondMessage, 2);
+            $code             = (int) str_replace('ORA-', '', $code);
```
morozov, line 38: Why do we need to replace the error code obtained from the driver with the one parsed from the error message?

tests/Driver/PDO/ExceptionTest.php
```diff
@@ -61,4 +61,25 @@ public function testOriginalExceptionIsInChain(): void
     {
         self::assertSame($this->wrappedException, $this->exception->getPrevious());
     }
+
+    public function testExposesUnderlyingErrorOnOracle(): void
+    {
+        $pdoException = new PDOException(<<<TEXT
+OCITransCommit: ORA-02091: transaction rolled back
+ORA-00001: unique constraint (DOCTRINE.C1_UNIQUE) violated
+ (/private/tmp/php-20211003-35441-1sggrmq/php-8.0.11/ext/pdo_oci/oci_driver.c:410)
```
morozov, line 65: Please do not add unit tests for such runtime-specific cases. If necessary, this logic should be tested in an integration way across all relevant platforms.

tests/Functional/DeferrableConstraintsTest.php
```diff
@@ -0,0 +1,258 @@
...
+        parent::setUp();
+
+        $platform = $this->connection->getDatabasePlatform();
+
+        if ($platform instanceof OraclePlatform) {
+            $constraintName = 'C1_UNIQUE';
+        } elseif ($platform instanceof PostgreSQLPlatform) {
+            $constraintName = 'c1_unique';
+        } else {
+            $constraintName = 'c1_unique';
```
morozov: How is it different from the above? Is it possible to use `strtolower()` where necessary instead of branching based on the platform?

tests/Functional/DeferrableConstraintsTest.php
```diff
@@ -0,0 +1,258 @@
...
+            );
+        } elseif ($platform instanceof PostgreSQLPlatform) {
+            $createConstraint = sprintf(
+                'ALTER TABLE deferrable_constraints ' .
+                'ADD CONSTRAINT %s UNIQUE (unique_field) DEFERRABLE INITIALLY IMMEDIATE',
+                $constraintName
+            );
+        } elseif ($platform instanceof SqlitePlatform) {
+            $createConstraint = sprintf(
+                'CREATE UNIQUE INDEX %s ON deferrable_constraints(unique_field)',
```
morozov: So it looks like we're testing a feature that isn't supported by the DBAL. Does SQLite support deferrable constraints?

tests/Functional/DeferrableConstraintsTest.php
```diff
@@ -0,0 +1,258 @@
...
+use Doctrine\DBAL\Schema\UniqueConstraint;
+use Doctrine\DBAL\Tests\FunctionalTestCase;
+use Throwable;
+
+use function sprintf;
+
+final class DeferrableConstraintsTest extends FunctionalTestCase
+{
+    /** @var string */
+    private $constraintName = '';
```
morozov: Why should it be part of the test state? It's the test data that should be static.

morozov (commented):
I'm still wrapping my head around the changes but see a few comments below.

morozov:
Could you try rebasing so that all the required build jobs get executed?

tests/Functional/DeferrableConstraintsTest.php
simPod: No, it does not. But we wanted to run the test on all platforms https://github.com/doctrine/dbal/pull/3424#pullrequestreview-190518820

src/Driver/OCI8/Exception/Error.php
simPod: Because it's the underlying one that interests us (UC violation). OCI chains errors so there are two errors returned 
 in this case. `2091` and `1`.

But the code from the driver is related only to the first error (transaction roll back `2091`) while we are interested in the second one to get information about the actual underlying issue (uc violated `1`). We then know we can create UCViolation exception.

tests/Driver/PDO/ExceptionTest.php
simPod: It is tested in integration way in the main test here. 

Wanted to clarify and validate the custom logic so I isolated the error message parsing behaviour into unit tests as well to better document it. You asked exactly about it here https://github.com/doctrine/dbal/pull/4846#discussion_r733216076 and this tests covers it. But I can remove it before merging for sure.

src/Driver/OCI8/Exception/Error.php
morozov: The unique constraint violation error is of interest to us in this specific case but there may be other cases. I don't think that replacing the error reported by the driver based on a personal preference is a good idea.

Instead, such errors could be represented as a chain (see `Exception::$previous`) of exceptions. So it would be 1) a transaction exception caused by 2) a unique constraint violation.

tests/Functional/DeferrableConstraintsTest.php
morozov: In this case, it would make more sense to have two tests: each covers its own feature on the platforms that support it.

tests/Functional/DeferrableConstraintsTest.php
simPod: It's better to store constraint name in advance as it is not used for assertions only but in SQL statements as well

e.g. `SET CONSTRAINTS "%s" DEFERRED`

tests/Functional/DeferrableConstraintsTest.php
simPod: Should I convert it to method then? I need to know the name of constraint for each platform (https://github.com/doctrine/dbal/pull/4846#discussion_r753673594)

tests/Functional/DeferrableConstraintsTest.php
simPod: I renamed the test. The initial motive of previous author was to handle deferred constraints only in the original PR.

Then we realised unique constraints were not really tested so we've expanded the test. So this tests unique constraint violations while it also tests deferring the violations where eligible. So I don't really see how to split it but I've adapted the naming as it was misleading for sure.

src/Driver/OCI8/Exception/Error.php
simPod: ~I have to introduce custom TransactionRolledBack exception~

simPod: How do I make it work with ExceptionConverter then? It does not support chained errors from the driver.

This is what I have since I'm able to read multiple errors from the server now:

`Error - Transaction rolled back <- Error - UC violation`

--------------

This is what exception converter does right now

`Driver Exception (Error - Transaction rolled back <- Error - UC violation)`

---------------

We need to somehow inject `UniqueConstraintViolationException` into 👆🏾 but cannot simply use `previous` as OCI Error is used as `previous` now.

See 4f6487e

src/Driver/PDO/Connection.php
simPod: Added. It does not break any tests but I don't have a case for it either.

simPod:
.

morozov:
Closing as this pull request hasn't received significant updates in almost a year and is far from being done.

simPod:
@morozov I was waiting for the feedback

IIRC it was fully done and functional from my side.

morozov:
IIRC, it had a failing build for all this time. Didn't it? Let's reopen and double-check.

simPod:
I think it failed because of oracle platform or something similar. And it
was related to this IIRC
https://github.com/doctrine/dbal/pull/4846#discussion_r748900007

morozov:
If you plan to continue working on this patch, please ~~retarget it against 3.5.x,~~ rebase and address the build issues.

src/Connection.php
simPod: Addressed in https://github.com/doctrine/dbal/pull/5686

morozov:
It looks like the changes requested in https://github.com/doctrine/dbal/pull/4846#discussion_r733215344 got lost during the rebase.

I haven't started reviewing the patch yet. Let's get https://github.com/doctrine/dbal/pull/4846 done and this one rebased on top first.

simPod:
.

simPod:
D ys bs

simPod:
.

simPod:
Bot

simPod:
Fix urself

derrabus:
Can somebody fill me in what's up with this PR? It does not look like the kind of change we would accept on the 4.0.x branch.

simPod:
@derrabus see https://github.com/doctrine/orm/issues/7545 `Current behavior` section. The problem is that we're trying to rollback already rollbacked transaction and therefore the underlying issue (like unique constraint violation) is suppressed by `There's no active transaction`.

derrabus:
Understood. Let me take over the review then. But as I said, we don't accept this kind of change on the 4.0.x branch. You need to target 3.7.x currently if this is a bugfix and 3.8.x otherwise.

simPod:
Great. I have rebased this like trillion times over the years so I'm glad there's a promise of a movement now. I'll target 3.8.x I guess.

simPod:
@derrabus this is now working again.

src/Driver/OCI8/Exception/Error.php
```diff
@@ -16,12 +18,26 @@
...
     public static function new($resource): self
     {
         $error = oci_error($resource);
         assert($error !== false);
 
-        return new self($error['message'], null, $error['code']);
+        $code    = $error['code'];
+        $message = $error['message'];
+        if ($code === self::CODE_TRANSACTION_ROLLED_BACK) {
+            // There's no way this can be unit-tested as it's impossible to mock $resource
```
derrabus, line 32: This kind of comment does not belong into production code.

src/Driver/OCI8/Exception/Error.php
derrabus: > How do I make it work with ExceptionConverter then? It does not support chained errors from the driver.

That problem sounds solvable to me.

> Instead, such errors could be represented as a chain (see `Exception::$previous`) of exceptions. So it would be 1) a transaction exception caused by 2) a unique constraint violation.

This is the way.

src/Driver/OCI8/Exception/Error.php
```diff
@@ -16,12 +18,26 @@
...
     /** @param resource $resource */
     public static function new($resource): self
     {
         $error = oci_error($resource);
         assert($error !== false);
 
-        return new self($error['message'], null, $error['code']);
+        $code    = $error['code'];
+        $message = $error['message'];
+        if ($code === self::CODE_TRANSACTION_ROLLED_BACK) {
```
derrabus, line 31: I don't feel comfortable with making this exception class aware of specific error codes. That logic belongs into `ExceptionConverter`. Can we replace this logic with a generic detection of chained errors?

derrabus (changes requested):

PR #4848, opened 2021-10-07
Add Postgres 14 to the test matrix

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

Postgres 14 has been released and we should test against it.

https://www.postgresql.org/about/news/postgresql-14-released-2318/

I decided to keep Postgres 13 for now because it's still widely used. But we can probably remove it in a couple of months.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -186,9 +186,10 @@ jobs:
           - "9.2"
           - "9.4"
           - "13"
+          - "14"
         include:
           - php-version: "8.0"
-            postgres-version: "13"
+            postgres-version: "14"
 
     services:
       postgres:


derrabus:
We will need to adjust the branch protection rules. GHA is waiting for the old PHP 8 + Postgres 13 job.

greg0ire (approved):

morozov:
I removed "PHPUnit with PostgreSQL (8.0, 13)" from the list of "Status checks that are required" on 2.13.x ~~but it's still here. Should we just execute our Administrator privilege and merge it?~~ I recall removing the obsolete check worked immediately in the past.

Nevermind, I forgot to save the settings.

PR #4852, opened 2021-10-08
QueryCacheProfiler::generateCacheKey: Param password should be ignore in connection hash

janbarasek:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes

#### Summary

We use Doctrine in many containers for production and we change passwords every 3 days for better security.

When the password is changed, the old original cache is ignored and we have to re-generate everything very expensively. I think the password should not be part of the hash.

Snímek obrazovky 2021-10-08 v 15 39 39

### DIFF
--- a/src/Cache/QueryCacheProfile.php
+++ b/src/Cache/QueryCacheProfile.php
@@ -78,7 +78,7 @@ class QueryCacheProfile
      */
     public function generateCacheKeys($sql, $params, $types, array $connectionParams = [])
     {
-        if (count($connectionParams) > 1 && array_key_exists('password', $connectionParams)) {
+        if (isset($connectionParams['password'])) {
             unset($connectionParams['password']);
         }
         $realCacheKey = 'query=' . $sql .
--- a/src/Cache/QueryCacheProfile.php
+++ b/src/Cache/QueryCacheProfile.php
@@ -81,6 +81,7 @@ class QueryCacheProfile
         if (isset($connectionParams['password'])) {
             unset($connectionParams['password']);
         }
+
         $realCacheKey = 'query=' . $sql .
             '&params=' . serialize($params) .
             '&types=' . serialize($types) .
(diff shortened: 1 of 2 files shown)


src/Cache/QueryCacheProfile.php
```diff
@@ -78,6 +78,9 @@ public function getCacheKey()
      */
     public function generateCacheKeys($sql, $params, $types, array $connectionParams = [])
     {
+        if (count($connectionParams) > 1 && array_key_exists('password', $connectionParams)) {
```
stloyd: Doing just isset would be more efficient:
```suggestion
        if (isset($connectionParams['password'])) {
```

src/Cache/QueryCacheProfile.php
```diff
@@ -78,6 +78,9 @@ public function getCacheKey()
      */
     public function generateCacheKeys($sql, $params, $types, array $connectionParams = [])
     {
+        if (isset($connectionParams['password'])) {
+            unset($connectionParams['password']);
+        }
```
greg0ire, line 83: Can you please add a tests for this, that shows that the cache key does not depend on the password?

janbarasek: Added.

tests/Cache/QueryCacheProfileTest.php
```diff
@@ -135,4 +135,29 @@ public function testShouldGenerateSameKeysIfNoneOfTheParamsChanges(): void
...
+            $this->query,
+            $this->params,
+            $this->types,
+            [
+                'user'     => 'database_user',
+                'password' => 'second-password',
+            ]
+        );
+
+        self::assertEquals($firstRealCacheKey, $secondRealCacheKey, 'Cache keys for different password should be the same');
```
greg0ire: This should make the coding standard tool happy:
```suggestion
        self::assertEquals(
            $firstRealCacheKey,
            $secondRealCacheKey,
            'Cache keys for different password should be the same'
        );
```

greg0ire (approved):

greg0ire:
Thanks @janbarasek !

PR #4855, opened 2021-10-09
Test AbstractPlatform::getLengthExpression()

morozov:
This patch adds an integration test and improves the consistency of the `AbstractPlatform::getLengthExpression()` behavior across supported platforms.

1. Pass the "charset" from the PHPUnit configuration to the driver.
2. Add platform-specific implementations for MySQL and IBM DB2. This makes them return the length in characters instead of length in bytes. It's hardly a BC break since previously the expected behavior wasn't specified detailed enough.
3. Switch from SQL Server 2017 to 2019 on GitHub Actions (see [UTF-8 support](https://docs.microsoft.com/en-us/sql/relational-databases/collations/collation-and-unicode-support?view=sql-server-ver15#utf8)).
4. Explicitly configure UTF-8 connection charsets where required (MySQL < 8.0 and MariaDB < 10.5: `utf8mb4`, SQL Server: `*_UTF8`, Oracle: `AL32UTF8`).

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -369,18 +369,18 @@ jobs:
           - "sqlsrv"
           - "pdo_sqlsrv"
         collation:
-          - "Latin1_General_100_CI_AS"
+          - "Latin1_General_100_CI_AS_SC_UTF8"
         include:
-          - collation: "Latin1_General_100_CS_AS"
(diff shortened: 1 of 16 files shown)


morozov:
The pending builds should be removed from the branch configuration before the merge.

greg0ire (approved):

SenseException (approved):

PR #4763, opened 2021-08-27
Remove AbstractPlatform::getName()

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes

See https://github.com/doctrine/dbal/pull/4755.

PR #4860, opened 2021-10-11
Drop support for SQL Server 2016 and older

morozov:
The mainstream support has ended ([source](https://docs.microsoft.com/en-us/lifecycle/products/sql-server-2016)).


derrabus:
Shall we rename `SQLServer2012Platform` to `SQLServer2017Platform` then?

derrabus:
… or deprecate&drop `SQLServer2012Platform` completely? The name of the class is highly confusing, we only support 2017 and onwards.

morozov:
`SQLServer2012Platform` _was_ removed in https://github.com/doctrine/dbal/pull/4763. Now, `SQLServerPlatform` is explicitly documented as:

https://github.com/doctrine/dbal/blob/86322856e6802526c9f023820452bcc96a0eced7/src/Platforms/SQLServerPlatform.php#L46-L49

We decided not to have a version in the class name representing the oldest supported version. It will help avoid extra diffs when dropping old platforms, also https://github.com/doctrine/dbal/pull/4755#discussion_r696142079.

derrabus:
Oh, I didn't see that. 🙈

derrabus (approved):

PR #4867, opened 2021-10-13
Allow less specific version string

ihmels:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

If you specify the version manually (via connection DSN), you have to specify the exact version so far. It would be nicer if you could specify the version only up to the minor version.

### DIFF
--- a/src/Driver/AbstractMySQLDriver.php
+++ b/src/Driver/AbstractMySQLDriver.php
@@ -95,7 +95,7 @@ abstract class AbstractMySQLDriver implements VersionAwarePlatformDriver
     {
         if (
             preg_match(
-                '/^(?:5\.5\.5-)?(mariadb-)?(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)/i',
+                '/^(?:5\.5\.5-)?(mariadb-)?(?P<major>\d+)(?:\.(?P<minor>\d+)(?:\.(?P<patch>\d+))?)?/i',
                 $versionString,
                 $versionParts
             ) === 0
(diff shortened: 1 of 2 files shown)


morozov:
@ihmels if you know that the specific minor and patch version won't make a difference during the platform detection, why not pass something like `10.0.0`? This way, it's clear to the developer what version they are passing to the DBAL. If they are not happy with the result of detection, they could be more specific and pass say `10.2.7`. 

Passing just `10` puts the burden of the decision-making on the DBAL: should it be the earliest (e.g. `10.0.0`) or the latest `10.99.99`? What if the platform implemented for `10.0.0` (as derived from the `10` passed in the configuration) is incompatible with the actual server of the version say `10.5.6`? The developer passing `10` may rightfully expect that `10` should be compatible with any `10`, not only some specific `10`s.

I don't think this is a good idea.

ihmels:
@morozov Okay, you made some good points. Shouldn't the exact version always be specified then, because these problems also apply to other drivers?

https://github.com/doctrine/dbal/blob/d47c4877f050b24b1e522bbcbdd717a06524c01b/src/Driver/AbstractMySQLDriver.php#L60-L84

https://github.com/doctrine/dbal/blob/d47c4877f050b24b1e522bbcbdd717a06524c01b/src/Driver/AbstractPostgreSQLDriver.php#L26-L45

morozov:
> Shouldn't the exact version always be specified then, because these problems also apply to other drivers?

I believe it should. The logic in `getOracleMysqlVersionNumber()` looks invalid anyway, it may return the value without the patch version but with a trailing dot.

morozov:
Closing as the behavior of accepting an incomplete server version is not welcomed. Please feel free to file another pull request that would deprecate this behavior in the existing drivers and eventually remove it.

PR #4868, opened 2021-10-13
 Rework connect/disconnect routines in functional tests

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

The `FunctionalTestCase::tearDown()` method contains the connection management logic that must work correctly. Otherwise, connection resource leaks are possible (see https://github.com/doctrine/dbal/pull/4521). While some tests override the `tearDown()` method, there's no guarantee that they will call the parent. Having to keep an eye on the implementations is an additional development and maintenance burden.

Instead of implementing connection management in `setUp()` and `tearDown()`, we can use arbitrarily named `@before` and `@after` methods that the extending classes don't have to be aware of. It simplifies the tests by reducing the number of the necessary `parent::` calls.

**Change summary**:
1. Replace `setUp()` and `tearDown()` in `FunctionalTestCase` with `@before` and `@after`.
2. Remove the no longer necessary `parent::` calls.
3. (not directly related) Remove some `private static $xyz` shared state from the tests for better test isolation.

### DIFF
--- a/tests/Functional/LockMode/NoneTest.php
+++ b/tests/Functional/LockMode/NoneTest.php
@@ -59,14 +59,14 @@ class NoneTest extends FunctionalTestCase
 
     protected function tearDown(): void
     {
-        if ($this->connection2->isTransactionActive()) {
-            $this->connection2->rollBack();
-        }
-
         $this->connection2->close();
 
         if (! $this->connection->getDatabasePlatform() instanceof SQLServer2012Platform) {
             return;
+        }
+
+        if ($this->connection->isTransactionActive()) {
+            $this->connection->rollBack();
         }
 
         $db = $this->connection->getDatabase();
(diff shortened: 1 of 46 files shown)


tests/Functional/LockMode/NoneTest.php
```diff
@@ -59,18 +57,14 @@ public function setUp(): void
...
     {
-        parent::tearDown();
-
         if ($this->connection2->isTransactionActive()) {
             $this->connection2->rollBack();
         }
 
         $this->connection2->close();
 
-        $this->connection->getSchemaManager()->dropTable('users');
```
morozov, line 72: This `DROP TABLE` statement would implicitly commit the transaction started in the test method but the wrapper connection would still think that the transaction is active and fail to roll it back in `disconnect()`.

greg0ire (approved):

derrabus (approved):

Issue #4869, opened 2021-10-13
mysqli::real_connect may be called with null $flags


mondrake:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.1

#### Summary

In PHP 8.1 you get a deprecation when `mysqli::real_connect` is called with null $flags:
`
mysqli::real_connect(): Passing null to parameter #7 ($flags) of type int is deprecated
`

morozov:
Neither PHPStan nor Psalm in their current configuration catch this issue but PHPStan level 8 would:
```
$ phpstan a src/Driver/Mysqli/Connection.php
Note: Using configuration file phpstan.neon.dist.
 1/1 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ ----------------------------------------------------------------------------------- 
  Line   Connection.php                                                                     
 ------ ----------------------------------------------------------------------------------- 
  61     Parameter #7 $flags of method mysqli::real_connect() expects int, int|null given.  
 ------ ----------------------------------------------------------------------------------- 
                                                           
 [ERROR] Found 1 error

greg0ire:
That would be since https://php.watch/versions/8.1/internal-func-non-nullable-null-deprecation (so indeed PHP 8.1, more precisely in https://github.com/php/php-src/commit/b10416a652d26577a22fe0b183b2258b20c8bb86).

I pushed a fix in https://github.com/doctrine/dbal/pull/4879, which merges up #4873 into 3.1.x

iammati:
Is there maybe a patch/fix for `doctrine/dbal:^2.13.7` + PHP 8.1? Experiencing this at the very moment using a CMS (TYPO3 v11.5) which depends yet on dbal v2 (with deprecation exceptions reporting enabled)

derrabus:
Please open a PR against the 2.13.x branch.

PR #4873, opened 2021-10-15
Bump PHPStan level to 8

morozov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

See https://github.com/doctrine/dbal/issues/4869#issuecomment-943934836.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,5 +1,5 @@
 parameters:
-    level: 7
+    level: 8
     paths:
         - %currentWorkingDirectory%/lib
     reportUnmatchedIgnoredErrors: false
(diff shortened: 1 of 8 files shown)


phpstan.neon.dist
```diff
@@ -130,3 +130,21 @@ parameters:
             message: '~^Circular definition detected in type alias (Override)?Params\.$~'
             paths:
                 - %currentWorkingDirectory%/lib/Doctrine/DBAL/DriverManager.php
+
+        # TODO: remove in 4.0.0
+        -
+            message: '~^Parameter #2 \$table of class Doctrine\\DBAL\\Event\\SchemaIndexDefinitionEventArgs constructor expects string, string\|null given\.$~'
+            paths:
+                - %currentWorkingDirectory%/lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php
```
greg0ire, line 138: :thinking: Out of scope, but we should get rid of all these `%currentWorkingDirectory%` references, there useless since there is only one directory in which we have a `lib/Doctrine/DBAL/Schema/AbstractSchemaManager.php`.

greg0ire (approved):

derrabus (approved):

PR #4874, opened 2021-10-15
Get rid of current working directory variable

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

They are useless because there is only one directory from which this
configuration makes sense.


derrabus:
Rebase needed.

morozov (approved):

PR #4875, opened 2021-10-15
Run tests on PHP 8.1 with different database drivers

### LINKED ISSUE #4870: PHP 8.1: MySQLi: Default error mode set to exceptions
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.1

#### Summary

In PHP 8.1, MySQLi default error mode is set to exceptions. Prior to PHP 8.1, the default error reporting mode in MySQLi was to silent the errors.

If I understand correctly, now `mysqli_sql_exception` is not captured and standardized into DBAL exceptions, and rather bubble up in consuming code.

https://php.watch/versions/8.1/mysqli-error-mode

derrabus:
I've bumped all PHP 8.0 jobs for individual database drivers to PHP 8.1 to uncover issue that we might have there. If you think that testing on PHP 8.0 additionally provides extra value, I can re-add those jobs again. But for now, I did not want to bloat the test matrix further.

Fixes #4870

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -96,7 +96,7 @@ jobs:
       matrix:
         php-version:
           - "7.4"
-          - "8.0"
+          - "8.1"
 
     services:
       oracle:
@@ -139,7 +139,7 @@ jobs:
       matrix:
         php-version:
           - "7.4"
-          - "8.0"
+          - "8.1"
 
     services:
       oracle:
@@ -188,7 +188,7 @@ jobs:
           - "13"
           - "14"
         include:
-          - php-version: "8.0"
+          - php-version: "8.1"
             postgres-version: "14"
 
     services:
@@ -245,10 +245,10 @@ jobs:
           - "mysqli"
           - "pdo_mysql"
         include:
-          - php-version: "8.0"
+          - php-version: "8.1"
             mariadb-version: "10.5"
             extension: "mysqli"
-          - php-version: "8.0"
+          - php-version: "8.1"
             mariadb-version: "10.5"
             extension: "pdo_mysql"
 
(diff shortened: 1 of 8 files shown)


derrabus:
Marking as draft because I expect tests to fail.

.github/workflows/continuous-integration.yml
```diff
@@ -434,7 +433,7 @@ jobs:
           coverage: "pcov"
           ini-values: "zend.assertions=1"
           tools: "pecl"
-          extensions: "${{ matrix.extension }}-5.9.0preview1"
+          extensions: "${{ matrix.extension }}-5.10.0beta1"
```
derrabus, line 436: Earlier releases of the driver are not compatible with PHP 8.1.

derrabus:
Tests are green. As usual, we need to adjust the branch protection rules.

lib/Doctrine/DBAL/Driver/Mysqli/Exception/ConnectionError.php
```diff
@@ -18,4 +20,12 @@ public static function new(mysqli $connection): self
     {
         return new self($connection->error, $connection->sqlstate, $connection->errno);
     }
+
+    public static function upcast(mysqli_sql_exception $exception): self
+    {
+        $p = new ReflectionProperty(mysqli_sql_exception::class, 'sqlstate');
```
morozov, line 26: I'm wondering, what the idea behind the `mysqli_sql_exception` API design is 🤔

morozov (approved):
👍

lib/Doctrine/DBAL/Driver/Mysqli/Exception/ConnectionError.php
derrabus: Yeah. A protected property without any accessors is a really brilliant idea.

PR #4876, opened 2021-10-15
Fix XSD paths in PHPUnit XML config files

derrabus:


.gitignore
```diff
@@ -1,5 +1,5 @@
 /composer.lock
-/.phpunit.result.cache
+.phpunit.result.cache
```
greg0ire, line 2: What is this change about?

derrabus: If I run PHPUnit locally with any of the XML files, it will create the cache in the directory of the config file. I thought it would be best if I ignored the PHPUnit cache everywhere.

greg0ire: Ah right, you mean the ones in `ci/github/phpunit`?

greg0ire (approved):

PR #4879, opened 2021-10-15
Merge 2.13.x into 3.1.x

### LINKED ISSUE #4869: mysqli::real_connect may be called with null $flags

greg0ire:
Closes https://github.com/doctrine/dbal/issues/4869

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,5 +1,5 @@
 parameters:
-    level: 7
+    level: 8
     paths:
         - %currentWorkingDirectory%/lib
     reportUnmatchedIgnoredErrors: false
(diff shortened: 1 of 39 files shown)


greg0ire:
Note to self: port 429f746 to 2.13.x Cc @derrabus in case you want to beat me to it.

derrabus:
I intend to do that tomorrow, if that's not too late for you.

greg0ire:
@derrabus that's great, there is absolutely no hurry!

src/Schema/AbstractSchemaManager.php
```diff
@@ -156,6 +156,7 @@ public function listSequences($database = null)
     {
         if ($database === null) {
             $database = $this->_conn->getDatabase();
+            assert($database !== null);
```
greg0ire: I don't know how likely it is these assertions will actually fail. If there is a more appropriate exception I should throw, let me know.

phpstan.neon.dist
```diff
@@ -133,6 +133,30 @@ parameters:
...
+        # https://github.com/phpstan/phpstan/issues/4679
+        -
+            message: '~^Cannot call method writeTemporary\(\) on OCILob\|null\.$~'
+            paths:
+                - src/Driver/OCI8/Statement.php
+
+        # https://github.com/phpstan/phpstan/issues/5786
+        -
+            message: '~Parameter \#7 \$flags of method mysqli\:\:real_connect\(\) expects int, int\|null given\.~'
+            path: src/Driver/Mysqli/Connection.php
```
greg0ire: I added that last one, so please take a look at the linked PHPStan issue.

greg0ire:
Settings will have to be tweaked before pushing this.

phpstan.neon.dist
greg0ire: Ah I have to rework this actually, because of https://github.com/doctrine/dbal/issues/4869

src/Driver/Mysqli/Connection.php
```diff
@@ -42,10 +42,14 @@ public function __construct(
         ?string $database = null,
         ?int $port = null,
         ?string $socket = null,
-        ?int $flags = null,
+        ?int $flags = 0,
```
morozov: The constructor is internal. It won’t be a BC-break, if the parameter is made non-nullable.

src/Schema/AbstractSchemaManager.php
morozov: Probably, it should be an exception. Let’s suppress all errors with non-obvious fixes instead of fixing here.

derrabus (approved):

morozov (approved):

PR #4880, opened 2021-10-16
Catch MySQLi exceptions on PHP 8.1

### LINKED ISSUE #4870: PHP 8.1: MySQLi: Default error mode set to exceptions

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #4870

#### Summary

This is the MySQLi part of #4875, redone on the 3.1.x branch. The branches diverged in this area of the code so that I preferred to redo the changes to ease the merge-up.

### DIFF
--- a/src/Driver/Mysqli/Connection.php
+++ b/src/Driver/Mysqli/Connection.php
@@ -10,6 +10,7 @@ use Doctrine\DBAL\Driver\ServerInfoAwareConnection;
 use Doctrine\DBAL\Driver\Statement as DriverStatement;
 use Doctrine\DBAL\ParameterType;
 use mysqli;
+use mysqli_sql_exception;
 
 use function assert;
 use function floor;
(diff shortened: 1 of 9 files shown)


greg0ire (approved):

src/Driver/Mysqli/Connection.php
```diff
@@ -52,7 +53,13 @@ public function __construct(
             $initializer->initialize($connection);
         }
```
greg0ire, line 54: Yesterday, I had to move this block inside the `try` block for this to work… weird that you do not have to.

derrabus: I've adjusted the `Charset` initializer to catch the exception.

PR #4881, opened 2021-10-16
Add missing upcast

derrabus:
see https://github.com/doctrine/dbal/pull/4879#issuecomment-944703167


greg0ire:
This was caught by `tests/Functional/StatementTest.php` (`testFetchFromExecutedStatementWithFreedResult`). Maybe it's high level enough it can be ported with little to no modification?

derrabus:
It can be backported, however it'll be green even without my change. Also, `StatementTest` on 2.13 already contains tests that do the same thing through deprecated APIs. I would suggest to merge the PR as-is. Even if no test covers this try/catch statement, I'd leave it in place.

greg0ire (approved):

